Issue #863, opened 2025-08-05
Parquet Exporter better handling of adaptive OTAP Schema


albertlockett:
There are two ways that OTAP Arrow schema can change between record batches:

1. A dictionary builder may overflow, causing the type for this column to either get upgraded to a new dictionary key type, or the native array type
2. Optional fields may or may not be present

Currently the Parquet Exporter does not have the correct logic to handle these cases. It expects every record batch to have the same Schema (due to this assumption in the internal `ArrowWriter`).

### Fixes Needed:
There are multiple fixes needed to support adaptive schemas:

#### ~A. Handle compatible value sources for a single column (Dict8, Dict16, Native) when writing:~ ✅ _completed in #1006_ 
~We can get this "for free" by upgrading to arrow-rs 56.1, which has these changes: https://github.com/apache/arrow-rs/pull/8095~ 
 
#### ~B. Handle optional columns when writing:~ ✅ _completed in #1024_ 
~There are probably a few options for this:~
- ~1. In otel-arrow, add all optional the columns to the schema as entirely null columns. The downside of doing this is that we'll allocate memory for these all null arrays.~
- ~2. Try to support this in parquet-rs:~
  - ~While writing may be able to have the `ArrowRowGroupWriter` resolve the `ArrowColumnWriter` by name here, instead of going by schema position. We'd also need to be able to initialize new `ArrowColumnWriters` adaptively as new columns are added, and be able to give them null prefixes.~
https://github.com/apache/arrow-rs/blob/a9b6077b2d8d5b5aee0e97e7d4335878e8da1876/parquet/src/arrow/arrow_writer/mod.rs#L811-L820
  - ~We may need to have `ArrowWriter` keep track of a dynamic arrow schema, instead of using the one it's initialized with:~
~https://github.com/apache/arrow-~rs/blob/a9b6077b2d8d5b5aee0e97e7d4335878e8da1876/parquet/src/arrow/arrow_writer/mod.rs#L179~
- ~3. Have Otel-Arrow's parquet exporter managing the `ArrowColumnWriters` and dynamic schema itself, rather than trying to push all this code into `ArrowWriter`~

~I think option 2 would be ideal, but not sure if what we're trying to achieve is too special to be accepted into the default `ArrowWriter`. If so, we should go with option 3.~

#### C. Try to support easier read back
[...]

PR #1006, opened 2025-08-25
Upgrade to arrow 56.1 to improve adaptive schema handling in parquet exporter

albertlockett:
part of #863 

Improvements were made in the latest version of arrow-rs (https://github.com/apache/arrow-rs/pull/8095 and https://github.com/apache/arrow-rs/pull/8005) to improve parquet writer's ability to handle logically equivalent data types for the same column.

We need this b/c different OTAP batches might switch between Dictionary encoding (with various key sizes) and native encoding for the same record batch. The latest version of parquet is able to accept these different array types for the same column.

A test is added to ensure the behaviour works as expected.

### DIFF
--- a/rust/experimental/query_engine/engine-recordset/src/engine.rs
+++ b/rust/experimental/query_engine/engine-recordset/src/engine.rs
@@ -448,10 +448,18 @@ fn format_diagnostics(
         lines.push((line, Vec::new()));
     }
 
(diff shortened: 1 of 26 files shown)


lquerel (approved):
LGTM

jmacd (approved):

PR #1008, opened 2025-08-25
[otap-df-engine] Fix timer drift

utpilla:
## Changes

- Fix timer drift 

```diff
// Current implementation - causes drift
- let next_when = Instant::now() + duration;

// Better approach - maintains consistent intervals
+ let next_when = when + duration;
```

Problem:

```rust
// Say we want a 100ms recurring timer:
// Timer 1: Schedule for T+100ms
// Timer 1 fires at: T+102ms (2ms late due to system load)
// Timer 2: Schedule for (T+102ms) + 100ms = T+202ms  ← Should be T+200ms!
// Timer 2 fires at: T+205ms (3ms late)  
// Timer 3: Schedule for (T+205ms) + 100ms = T+305ms  ← Should be T+300ms!
//
// After 10 iterations, we could be 50ms+ off schedule!

// With the fix: let next_when = when + duration;
// Timer 1: Schedule for T+100ms
// Timer 1 fires at: T+102ms (2ms late, but we don't care)
// Timer 2: Schedule for (T+100ms) + 100ms = T+200ms  ← Correct!
// Timer 2 fires at: T+203ms (3ms late, but schedule stays correct)
// Timer 3: Schedule for (T+200ms) + 100ms = T+300ms  ← Still correct!
//
// The actual firing time can vary, but the schedule stays consistent
```

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
+++ b/rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
@@ -103,9 +103,6 @@ impl PipelineCtrlMsgManager {
                         if when > now {
                             tokio::time::sleep_until(when).await;
                         }
-                    } else {
-                        // No timers scheduled, wait indefinitely.
-                        futures::future::pending::<()>().await;
                     }
                 }, if next_expiry.is_some() => {
                     if let Some(Reverse((when, node_id))) = self.timers.pop() {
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -103,9 +103,6 @@ impl PipelineCtrlMsgManager {
                         if when > now {
                             tokio::time::sleep_until(when).await;
                         }
-                    } else {
```
utpilla, line 195: This is not related to the timer drift issue. The `else` block is unnecessary since we already have a guard for this select branch:
```rust
if next_expiry.is_some()
```

albertlockett (approved):

PR #1009, opened 2025-08-25
[query-engine] Allow accessor expressions in KQL group by expressions

CodeBlanch:
## Changes

* Improve the user experience when using KQL `summarize` by allowing accessor expressions in group-by clauses with auto resolution of names


drewrelmas (approved):
Thanks for improving this!

PR #1011, opened 2025-08-26
fix the parquet shutdown test flakiness again

albertlockett:
Hopefully this will finally fix the parquet shutdown test being flakey 🤞 

Instead of buffering up enough rows and _hoping_ that when we send the shutdown signal that the writer won't have enough time to flush them, we use an implementation of object store that forces there to be a delay before we write to the filesystem. That way when we call shutdown, the parquet should call flush_all on the writer manager, which will close the parquet file writers, which will hit this delay when they try to write the parquet footer, which will _hopefully_ cause a timeout and reliably this test will pass..

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/parquet_exporter.rs
+++ b/rust/otap-dataflow/crates/otap/src/parquet_exporter.rs
@@ -500,10 +500,10 @@ mod test {
         let temp_dir = tempfile::tempdir().unwrap();
         let base_dir: String = temp_dir.path().to_str().unwrap().into();
         let exporter = ParquetExporter::new(config::Config {
-            base_uri: base_dir.clone(),
+            base_uri: format!("testdelayed://{base_dir}?delay=500ms"),
             partitioning_strategies: None,
             writer_options: Some(WriterOptions {
-                target_rows_per_file: Some(5000),
+                target_rows_per_file: Some(50),
                 ..Default::default()
             }),
         });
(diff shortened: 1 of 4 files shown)


albertlockett:
I ran this test suite 5x times and it passed each time. I think this is decent evidence that the flake has resolved
https://github.com/open-telemetry/otel-arrow/actions/runs/17251595166/job/48960384799?pr=1011

rust/otap-dataflow/crates/otap/src/parquet_exporter.rs
```diff
@@ -500,10 +500,10 @@ mod test {
         let temp_dir = tempfile::tempdir().unwrap();
         let base_dir: String = temp_dir.path().to_str().unwrap().into();
         let exporter = ParquetExporter::new(config::Config {
-            base_uri: base_dir.clone(),
+            base_uri: format!("testdelayed://{base_dir}?delay=500ms"),
```
lquerel, line 503: Interesting

lquerel (approved):
LGTM.

rust/otap-dataflow/crates/otap/src/parquet_exporter.rs
albertlockett: I find using the URI like this a nice way to configure the object store. I think eventually when we add support to parquet for cloud object storage, we should use this pattern. e.g.
- AWS s3: `s3://my-bucket/some/path/`
- GCP cloud storage: `gs://...`
- Azure blob storage: `az://...`

This is what we did in LanceDB and it seemed to work reasonably well https://lancedb.github.io/lancedb/guides/storage/

That said, I'm not sure using parsing options from query string is a pattern we want/need for these cloud object object storage parameters. I think it probably makes sense to expose them through the config object directly because:
- 1. it's easier to validate,
- 2. users will find it easier to view the config struct definition to see which options are available

We can also document how it's possible to set  many of the config options through the environment variables:
https://docs.rs/object_store/latest/object_store/aws/struct.AmazonS3Builder.html#method.from_env

PR #1014, opened 2025-08-27
[query-engine] Concat & Join expressions

CodeBlanch:
## Changes

* Adds array concat, string concat, and string join expressions + recordset implementation
* Adds KQL support for `strcat`, `strcat_delim`, & `array_concat`

### DIFF
--- a/rust/experimental/query_engine/engine-recordset/src/primitives/resolved_value.rs
+++ b/rust/experimental/query_engine/engine-recordset/src/primitives/resolved_value.rs
@@ -506,6 +506,7 @@ impl ArrayValue for Sequence<'_> {
 
         let mut start = range.get_start_range_inclusize().unwrap_or(0);
         let mut end = range.get_end_range_exclusive().unwrap_or(len);
+        let mut index = 0;
 
         if end > len {
             panic!(
@@ -517,15 +518,24 @@ impl ArrayValue for Sequence<'_> {
         for v in &self.values {
             let len = usize::min(end, v.len());
             if len == 0 {
-                break;
+                continue;
             }
             if start < len {
                 let range = (start..len).into();
-                if !v.get_item_range(range, item_callback) {
+                if !v.get_item_range(
+                    range,
+                    &mut IndexValueClosureCallback::new(|_, v| {
+                        let r = item_callback.next(index, v);
+                        index += 1;
+                        r
+                    }),
+                ) {
                     return false;
                 }
+                start = 0;
+            } else {
+                start -= len;
             }
-            start += len;
             end -= len;
         }
 
@@ -874,13 +884,18 @@ mod tests {
     #[test]
     fn test_sequence_get_item_range() {
         fn run_test(v: &Sequence, range: ArrayRange, expected: &[Value]) {
-            v.get_item_range(
+            let mut items = 0;
+
+            assert!(v.get_item_range(
                 range,
                 &mut IndexValueClosureCallback::new(|i, v| {
                     assert_eq!(expected[i], v);
+                    items += 1;
                     true
                 }),
-            );
+            ));
+
+            assert_eq!(expected.len(), items);
         }
 
         let sequence = Sequence::new(vec![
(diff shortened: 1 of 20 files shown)


rust/experimental/query_engine/engine-recordset/src/primitives/resolved_value.rs
```diff
@@ -558,18 +679,31 @@ pub enum ResolvedArrayValue<'a> {
 
     /// A list of resolved values
     List(List<'a>),
+
+    /// A sequence of arrays
+    Sequence(Sequence<'a>),
```
drewrelmas, line 684: Part of me is skeptical about the ambiguous meanings between the terms `List`, `Collection`, and `Sequence` in the `engine-recordset` and `expressions` crates. I don't think I have a concrete recommendation for improvement, but it could be worth a second thought.

rust/experimental/query_engine/engine-recordset/src/primitives/resolved_value.rs
```diff
@@ -644,3 +876,129 @@ impl Display for ResolvedArrayValue<'_> {
...
+                OwnedValue::Integer(IntegerValueStorage::new(3)).to_value(),
+                OwnedValue::Integer(IntegerValueStorage::new(4)).to_value(),
+                OwnedValue::Integer(IntegerValueStorage::new(5)).to_value(),
+            ],
+        );
+        run_test(&sequence, (10..).into(), &[]);
+    }
+
+    #[test]
+    #[should_panic]
```
drewrelmas, line 977: Should we be testing with `#[should_panic]` instead of gracefully handling?

CodeBlanch: So in Rust we have this slice thing. You can index it by a range (ex `slice[start..end]`) and it will panic for some bad ranges. Not all, strangely[^1]. In our expressions we have some array-like structs which expose similar slice range indexing: Slice, Sequence, List. These structs have the same panic behavior which is what this test is validating. But if the user tries to take a slice of something with an invalid range they will get diagnostics/errors before it ever gets that far: https://github.com/open-telemetry/otel-arrow/blob/9859c859192f45e865c38fde9e616407e76c198b/rust/experimental/query_engine/engine-recordset/src/scalars/scalar_expressions.rs#L339-L345

There is also validation if you try to index into an array via a selector: https://github.com/open-telemetry/otel-arrow/blob/9859c859192f45e865c38fde9e616407e76c198b/rust/experimental/query_engine/engine-recordset/src/scalars/scalar_expressions.rs#L453-L478

I think this should be OK. Essentially we validate user input to prevent invalid access because the data structures themselves will panic if used incorrectly.

[^1]: IIRC `start..invalid_end` panics but `..invalid_end` does not panic it just gives you back all the valid data and ignores extras.

drewrelmas: The layering of validation before this step makes sense, thanks for confirming.

It is strange that `...invalid_end` doesn't panic! 😄

rust/experimental/query_engine/engine-recordset/src/primitives/resolved_value.rs
CodeBlanch: Ya naming is hard 🤣 Happy to rename anything if you have suggestions.

I feel pretty good about `Slice` because it matches the Rust concept. `Sequence` mirrors sort of .NET [ReadOnlySequence](https://learn.microsoft.com/dotnet/api/system.buffers.readonlysequence-1). List 🤷 List (`ListScalarExpression`) is really just an array built from resolved scalar values vs `ArrayScalarExpression` which is an array built from static values. We could rename `ArrayScalarExpression` -> `StaticArrayScalarExpression` and then do `ListScalarExpression` -> `ArrayScalarExpression` but there are a lot of similar structs which should also be renamed if we decide to do that. The only reason I came up with a new name ("List") was to avoid doing a big rename PR 😄

drewrelmas: I think I'm fine leaving it as-is for now! LGTM

PR #994, opened 2025-08-22
WIP: feat: add lower level support for batching OTAP data

msalib:

PR #1015, opened 2025-08-27
[otap-dataflow] otap pdata integration for otap_batch_processor

daviddahl:
Integration work for otap_batch_processor and otappdata batching capabilities

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
@@ -1,0 +1,1274 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! OTAP batch processor.
+//! Batches OtapPdata by count or timer; uses upstream OTAP batching for merge/split.
+
+use crate::OTAP_PROCESSOR_FACTORIES;
+use crate::pdata::OtapPdata;
+use async_trait::async_trait;
+use linkme::distributed_slice;
+use otap_df_config::error::Error as ConfigError;
+use otap_df_config::experimental::SignalType;
+use otap_df_config::node::NodeUserConfig;
+use otap_df_engine::config::ProcessorConfig;
+use otap_df_engine::error::Error as EngineError;
+use otap_df_engine::local::processor as local;
+use otap_df_engine::message::Message;
+use otap_df_engine::node::NodeId;
+use otap_df_engine::processor::ProcessorWrapper;
+use serde::Deserialize;
+use serde_json::Value;
+use std::num::NonZeroU64;
+use std::sync::Arc;
+use std::time::Duration;
+// For optional conversion during flush/partitioning
+use otel_arrow_rust::otap::OtapArrowRecords;
+use otel_arrow_rust::otap::batching::make_output_batches;
+
+/// URN for the OTAP batch processor
+pub const OTAP_BATCH_PROCESSOR_URN: &str = "urn:otap:processor:batch";
+
+/// Default configuration values (parity-aligned as we confirm Go defaults)
+pub const DEFAULT_SEND_BATCH_SIZE: usize = 8192;
+/// Default upper bound on batch size used to chunk oversized inputs (in number of items).
+/// See Config::send_batch_max_size for normalization semantics.
+pub const DEFAULT_SEND_BATCH_MAX_SIZE: usize = 0; // 0 means unlimited (OTLP parity)
+/// Timeout in milliseconds for periodic flush
+pub const DEFAULT_TIMEOUT_MS: u64 = 200;
+
+/// Semantic constants (avoid magic numbers)
+/// Minimum allowed send_batch_size
+pub const MIN_SEND_BATCH_SIZE: usize = 1;
+/// Sentinel meaning: no upper limit for max size (Go parity)
+pub const FOLLOW_SEND_BATCH_SIZE_SENTINEL: usize = 0;
+/// Minimum allowed metadata cardinality limit when specified
+pub const MIN_METADATA_CARDINALITY_LIMIT: usize = 1;
+/// Minimum increment to apply when counting items for size calculations
+pub const MIN_ITEM_INCREMENT: usize = 1;
+
+/// Log messages
+const LOG_MSG_SHUTTING_DOWN: &str = "OTAP batch processor shutting down";
+const LOG_MSG_DROP_EMPTY: &str = "OTAP batch processor: dropping empty OTAP record (0 rows)";
+const LOG_MSG_DROP_CONVERSION_FAILED: &str =
+    "OTAP batch processor: dropping message: OTAP conversion failed";
+const LOG_MSG_BATCHING_FAILED_PREFIX: &str = "OTAP batch processor: low-level batching failed for";
+const LOG_MSG_BATCHING_FAILED_SUFFIX: &str = "; dropping";
+
+/// Signal label constants (avoid repeated string literals)
+const SIG_LOGS: &str = "logs";
+const SIG_METRICS: &str = "metrics";
+const SIG_TRACES: &str = "traces";
+
+/// Configuration for the OTAP batch processor (parity with Go batchprocessor)
+#[derive(Debug, Clone, Deserialize)]
+pub struct Config {
+    /// Flush current batch when this count is reached.
+    /// When set to 0, the batch size is ignored and data will be sent immediately
+    /// subject only to send_batch_max_size (OTLP parity).
+    #[serde(default = "default_send_batch_size_opt")]
+    pub send_batch_size: Option<usize>,
+    /// Hard cap for splitting very large inputs.
+    /// Go behavior: 0 means no upper limit; missing defaults to 0 (unlimited).
+    #[serde(default = "default_send_batch_max_size")]
+    pub send_batch_max_size: usize,
+    /// Flush non-empty batches on this interval.
+    #[serde(with = "humantime_serde", default = "default_timeout_duration_opt")]
+    pub timeout: Option<Duration>,
+    /// Optional metadata partitioning keys (resource/scope/attribute names). Not yet supported.
+    /// ToDo: Support metadata-aware batching.
+    #[serde(default)]
+    pub metadata_keys: Vec<String>,
+    /// Optional limit on the number of distinct metadata-based groups this processor will track.
+    ///
+    /// Note: This is currently a no-op because grouping by metadata_keys has not yet been
+    /// implemented. When grouping lands, this will cap the number of concurrent groups and the
+    /// overflow strategy will be documented. Not yet supported.
+    /// ToDo: Support metadata-aware batching
+    #[serde(default)]
+    pub metadata_cardinality_limit: Option<usize>,
+}
+
+fn default_send_batch_size_opt() -> Option<usize> {
+    Some(DEFAULT_SEND_BATCH_SIZE)
+}
+
+fn default_send_batch_max_size() -> usize {
+    FOLLOW_SEND_BATCH_SIZE_SENTINEL // normalized in from_config
+}
+
+fn default_timeout_duration_opt() -> Option<Duration> {
+    Some(Duration::from_millis(DEFAULT_TIMEOUT_MS))
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            send_batch_size: default_send_batch_size_opt(),
+            send_batch_max_size: default_send_batch_max_size(),
+            timeout: default_timeout_duration_opt(),
+            metadata_keys: Vec::new(),
+            metadata_cardinality_limit: None,
+        }
+    }
+}
+
+/// Local (!Send) OTAP batch processor
+pub struct OtapBatchProcessor {
+    config: Config,
+    // Per-signal buffers of convertible OTAP records
+    current_logs: Vec<OtapArrowRecords>,
+    current_metrics: Vec<OtapArrowRecords>,
+    current_traces: Vec<OtapArrowRecords>,
+    // Running row counts per signal for size triggers
+    rows_logs: usize,
+    rows_metrics: usize,
+    rows_traces: usize,
+    // Tracks whether a size/max threshold was crossed since the last input for each signal.
+    // Used to decide whether a timer tick should flush that signal (OTLP parity semantics).
+    dirty_logs: bool,
+    dirty_metrics: bool,
+    dirty_traces: bool,
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+enum FlushReason {
+    Size,
+    Timer,
+    Shutdown,
+}
+
+/// Decide whether to request splitting from upstream batching.
+/// Returns Some(max) only when a split is needed and safe, otherwise None.
+fn requested_split_max(input: &[OtapArrowRecords], max_val: usize) -> Option<NonZeroU64> {
+    // Workaround: avoid splitting a single record until upstream bug is fixed.
+    // Also avoid requesting split when total rows fit within max.
+    let total_rows: usize = input.iter().map(|r| r.batch_length()).sum();
+    if total_rows <= max_val || input.len() <= 1 {
+        None
+    } else {
+        NonZeroU64::new(max_val as u64)
+    }
+}
+
+async fn log_batching_failed(
+    effect: &mut local::EffectHandler<OtapPdata>,
+    signal: &str,
+    err: &impl std::fmt::Display,
+) {
+    effect
+        .info(&format!(
+            "{LOG_MSG_BATCHING_FAILED_PREFIX} {signal}: {err}{LOG_MSG_BATCHING_FAILED_SUFFIX}"
+        ))
+        .await;
+}
+
+impl OtapBatchProcessor {
+    /// Construct a processor wrapper from a JSON configuration object and processor runtime config.
+    /// The JSON should mirror the Go collector batchprocessor shape. Missing fields fall back to
+    /// crate defaults. Invalid numeric values (e.g., zero) are normalized to minimal valid values.
+    pub fn from_config(
+        node: NodeId,
+        cfg: &Value,
+        proc_cfg: &ProcessorConfig,
+    ) -> Result<ProcessorWrapper<OtapPdata>, ConfigError> {
+        let mut config: Config =
+            serde_json::from_value(cfg.clone()).map_err(|e| ConfigError::InvalidUserConfig {
+                error: format!("invalid OTAP batch processor config: {e}"),
+            })?;
+
+        // Basic validation/normalization
+        // Allow send_batch_size = 0 to mean: ignore size threshold (send immediately),
+        // subject only to send_batch_max_size when non-zero (Go parity).
+        // Keep send_batch_max_size = 0 as unlimited (no upper bound).
+        // Validate: when both are non-zero, max must be >= size.
+        if config.send_batch_max_size != 0 {
+            if let Some(s) = config.send_batch_size {
+                if s != 0 && config.send_batch_max_size < s {
+                    return Err(ConfigError::InvalidUserConfig {
+                        error: format!(
+                            "invalid OTAP batch processor config: send_batch_max_size ({}) must be >= send_batch_size ({}) or 0 (unlimited)",
+                            config.send_batch_max_size, s
+                        ),
+                    });
+                }
+            }
+        }
+
+        if let Some(limit) = config.metadata_cardinality_limit {
+            if limit < MIN_METADATA_CARDINALITY_LIMIT {
+                config.metadata_cardinality_limit = Some(MIN_METADATA_CARDINALITY_LIMIT);
+            }
+        }
+        let user_config = Arc::new(NodeUserConfig::new_processor_config(
+            OTAP_BATCH_PROCESSOR_URN,
+        ));
+        let proc = OtapBatchProcessor {
+            config,
+            current_logs: Vec::new(),
+            current_metrics: Vec::new(),
+            current_traces: Vec::new(),
+            rows_logs: 0,
+            rows_metrics: 0,
+            rows_traces: 0,
+            dirty_logs: false,
+            dirty_metrics: false,
+            dirty_traces: false,
+        };
+        Ok(ProcessorWrapper::local(proc, node, user_config, proc_cfg))
+    }
+
+    /// Flush all per-signal buffers (logs, metrics, traces). Does nothing if all are empty.
+    async fn flush_current(
+        &mut self,
+        effect: &mut local::EffectHandler<OtapPdata>,
+        reason: FlushReason,
+    ) -> Result<(), EngineError> {
+        self.flush_logs(effect, reason).await?;
+        self.flush_metrics(effect, reason).await?;
+        self.flush_traces(effect, reason).await
+    }
+
+    async fn flush_logs(
+        &mut self,
+        effect: &mut local::EffectHandler<OtapPdata>,
+        reason: FlushReason,
+    ) -> Result<(), EngineError> {
+        // Only return early if there is nothing buffered for this signal at all
+        if self.current_logs.is_empty() {
+            return Ok(());
+        }
+        let input = std::mem::take(&mut self.current_logs);
+        if !input.is_empty() {
+            let max_val = self.config.send_batch_max_size;
+            if max_val <= MIN_SEND_BATCH_SIZE {
+                // Bypass upstream splitter for degenerate max; just forward each record
+                for records in input {
+                    let pdata: OtapPdata = records.into();
+                    effect.send_message(pdata).await?;
+                }
+                // Always reset counter in the degenerate path
+                self.rows_logs = 0;
+                if reason == FlushReason::Size {
+                    // Fully drained; no remainder -> not eligible for timer-based flush
+                    self.dirty_logs = false;
+                }
+            } else {
+                // Avoid upstream splitter where unnecessary or unsafe (see requested_split_max)
+                let max = requested_split_max(&input, max_val);
+                if let Some(max_nz) = max {
+                    let mut output_batches = match make_output_batches(Some(max_nz), input) {
+                        Ok(v) => v,
+                        Err(e) => {
+                            log_batching_failed(effect, SIG_LOGS, &e).await;
+                            Vec::new()
+                        }
+                    };
+
+                    // If size-triggered and we requested splitting (max is Some), re-buffer the last partial
+                    // output if it is smaller than the configured max. Timer/Shutdown flush everything.
+                    let mut rebuffered = false;
+                    if reason == FlushReason::Size && !output_batches.is_empty() {
+                        if let Some(last) = output_batches.last() {
+                            let last_rows = last.batch_length();
+                            if last_rows < max_val {
+                                let remainder = output_batches.pop().expect("last exists");
+                                self.rows_logs = last_rows;
+                                self.current_logs.push(remainder);
+                                rebuffered = true;
+                            }
+                        }
+                    }
+
+                    for records in output_batches {
+                        let pdata: OtapPdata = records.into();
+                        effect.send_message(pdata).await?;
+                    }
+
+                    if !rebuffered {
+                        self.rows_logs = 0;
+                    }
+                    if reason == FlushReason::Size {
+                        // Set timer eligibility based on whether a remainder was kept
+                        self.dirty_logs = rebuffered;
+                    }
+                } else {
+                    // No split requested (safe path)
+                    if input.len() > 1 {
+                        // Coalesce upstream only when there are multiple records to merge
+                        let output_batches = match make_output_batches(None, input) {
+                            Ok(v) => v,
+                            Err(e) => {
+                                log_batching_failed(effect, SIG_LOGS, &e).await;
+                                Vec::new()
+                            }
+                        };
+                        for records in output_batches {
+                            let pdata: OtapPdata = records.into();
+                            effect.send_message(pdata).await?;
+                        }
+                    } else {
+                        // Single record: forward as-is (avoids upstream edge-cases)
+                        for records in input {
+                            let pdata: OtapPdata = records.into();
+                            effect.send_message(pdata).await?;
+                        }
+                    }
+                    self.rows_logs = 0;
+                    if reason == FlushReason::Size {
+                        // Fully drained in no-split path
+                        self.dirty_logs = false;
+                    }
+                }
+            }
+        }
+        Ok(())
+    }
+
+    async fn flush_metrics(
+        &mut self,
+        effect: &mut local::EffectHandler<OtapPdata>,
+        reason: FlushReason,
+    ) -> Result<(), EngineError> {
+        if self.current_metrics.is_empty() {
+            return Ok(());
+        }
+        let input = std::mem::take(&mut self.current_metrics);
+        if !input.is_empty() {
+            let max_val = self.config.send_batch_max_size;
+            if max_val <= MIN_SEND_BATCH_SIZE {
+                for records in input {
+                    let pdata: OtapPdata = records.into();
+                    effect.send_message(pdata).await?;
+                }
+                self.rows_metrics = 0;
+                if reason == FlushReason::Size {
+                    self.dirty_metrics = false;
+                }
+            } else {
+                let max = requested_split_max(&input, max_val);
+                if let Some(max_nz) = max {
+                    let mut output_batches = match make_output_batches(Some(max_nz), input) {
+                        Ok(v) => v,
+                        Err(e) => {
+                            log_batching_failed(effect, SIG_METRICS, &e).await;
+                            Vec::new()
+                        }
+                    };
+
+                    let mut rebuffered = false;
+                    if reason == FlushReason::Size && !output_batches.is_empty() {
+                        if let Some(last) = output_batches.last() {
+                            let last_rows = last.batch_length();
+                            if last_rows < max_val {
+                                let remainder = output_batches.pop().expect("last exists");
+                                self.rows_metrics = last_rows;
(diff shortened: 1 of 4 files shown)


rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,698 @@
...
+            metadata_cardinality_limit: None,
+        }
+    }
+}
+
+/// Local (!Send) OTAP batch processor
+pub struct OtapBatchProcessor {
+    config: Config,
+    /// Single buffer of incoming pdata messages. We partition by signal type at flush.
+    current: Vec<OtapPdata>,
```
albertlockett: After reviewing a bit more carefully what how the code in https://github.com/open-telemetry/otel-arrow/pull/994 works under the hood, I _think_ there might be a change we need to make to how we manage the current batch.

If I'm not mistaken, we'll want to emit batches of logs, or metrics, or traces once we've reached the size threshold for each of these signal types, as opposed to calculating the total batch size across all the currently buffered pdatas.

If this is true, I think the net effect is that we'll want to keep some separate current state for each of these signal types.

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,698 @@
...
+        Ok(ProcessorWrapper::local(proc, node, user_config, proc_cfg))
+    }
+
+    /// Returns true if the buffered items reach or exceed the emission threshold.
+    ///
+    /// MVP: threshold is based on count (number of messages). In future, switch to real
+    /// row counts for OtapArrowRecords.
+    fn size_triggers_emission(&self) -> bool {
+        self.current.len() >= self.config.send_batch_size
+    }
```
albertlockett: I agree, long term threshold should use the total size of the batches instead of the count of batches. We should be able to do this now because in #994 a method was added to calculate the current batch size. This is what we should be using. 
https://github.com/open-telemetry/otel-arrow/blob/8c4473eefcd9f1107e49a56bb26d3b74d4c97ab0/rust/otel-arrow-rust/src/otap.rs#L96-L101

e.g. to get the size of the pdata, we convert it to OTAP then and then call this method:
```rs
// probably best to do this in `process`:
let otap: OtapArrowRecords = pdata.try_into()?;
let batch_len = otap.batch.length();
```

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,698 @@
...
+            }
+        }
+
+        // If nothing convertible, just emit passthrough items and return.
+        if convertible.is_empty() {
+            for item in passthrough {
+                effect.send_message(item).await?;
+            }
+            return Ok(());
+        }
```
albertlockett: I'm not sure we want to pass these messages along if there was an error converting them to OTAP. The best thing to do for now might be to return or log an error but eventually we'll want to emit a `Nack` message

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,698 @@
...
+                    .await;
+                convertible
+            }
+        };
+
+        // Emit converted (records) first, then passthrough
+        for records in output_batches {
+            let pdata: OtapPdata = records.into();
+            effect.send_message(pdata).await?;
+        }
```
albertlockett: If there's some "remainder batch" after splitting/merging whose size is less than the max batch length, do we want to forward it immediately here, or re-buffer it so it can be combined with following batches?

I'm honestly not sure what the right answer is here because I guess that batch could have rows that need to be flushed due to some timer tick?

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,698 @@
...
+        &mut self,
+        msg: Message<OtapPdata>,
+        effect: &mut local::EffectHandler<OtapPdata>,
+    ) -> Result<(), EngineError<OtapPdata>> {
+        match msg {
+            Message::Control(ctrl) => {
+                match ctrl {
+                    otap_df_engine::control::NodeControlMsg::TimerTick { .. } => {
+                        // Flush any buffered items on timer
+                        self.flush_current(effect).await
```
albertlockett: In the OTLP batch processor, we don't always flush on the timer tick. It looks like we flush if some threshold has been reached since the last time we saw an incoming batch
https://github.com/open-telemetry/otel-arrow/blob/8c4473eefcd9f1107e49a56bb26d3b74d4c97ab0/rust/otap-dataflow/crates/otlp/src/otlp_batch_processor.rs#L441-L463

Should we do something similar here?

daviddahl: agreed.

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: i updated the processor to operate more like the otlp batcher that handles various telemetry types separately

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: I have 3 current vecs for each telemetry type now along wuith seperate counts for each.

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+impl OtapBatchProcessor {
+    /// Construct a processor wrapper from a JSON configuration object and processor runtime config.
+    /// The JSON should mirror the Go collector batchprocessor shape. Missing fields fall back to
+    /// crate defaults. Invalid numeric values (e.g., zero) are normalized to minimal valid values.
+    pub fn from_config(
+        node: NodeId,
+        cfg: &Value,
+        proc_cfg: &ProcessorConfig,
+    ) -> Result<ProcessorWrapper<OtapPdata>, ConfigError> {
+        let mut config: Config = serde_json::from_value(cfg.clone()).unwrap_or_default();
```
lquerel: We shouldn't silently ignore errors during the configuration file parsing (see the `unwrap_or_default`). We should return an error.

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+            config.send_batch_max_size
+        };
+        config.send_batch_max_size = max;
+
+        if let Some(limit) = config.metadata_cardinality_limit {
+            if limit < MIN_METADATA_CARDINALITY_LIMIT {
+                config.metadata_cardinality_limit = Some(MIN_METADATA_CARDINALITY_LIMIT);
+            }
+        }
+        let user_config = Arc::new(NodeUserConfig::new_processor_config(
```
lquerel, line 203: The Arc looks suspect.

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+        create: |_: otap_df_engine::context::PipelineContext,
+                 node: NodeId,
+                 cfg: &Value,
+                 proc_cfg: &ProcessorConfig| {
+            OtapBatchProcessor::from_config(node, cfg, proc_cfg)
+        },
+    };
+
+/// Parses duration strings from Go-style configs (e.g., "200ms", "2s", "1m").
+/// If `s` is a plain number, it's treated as milliseconds for convenience.
```
lquerel: Old comment?

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+                    }
+                    otap_df_engine::control::NodeControlMsg::Ack { .. }
+                    | otap_df_engine::control::NodeControlMsg::Nack { .. }
+                    | otap_df_engine::control::NodeControlMsg::CollectTelemetry { .. } => Ok(()),
+                }
+            }
+            Message::PData(data) => {
+                let max = self.config.send_batch_max_size;
+                // Fallback item count for cases where conversion yields zero rows
+                let fallback_count = item_count(&data);
```
lquerel: Doesn't look useful

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+                match OtapArrowRecords::try_from(data.clone()) {
+                    Ok(rec) => {
+                        let rows = rec.batch_length();
+                        match &rec {
+                            OtapArrowRecords::Logs(_) => {
+                                let eff = if rows > 0 { rows } else { fallback_count };
```
lquerel: In my opinion, an empty batch could just be ignored by the batch processor

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+                                {
+                                    self.flush_logs(effect).await?;
+                                }
+                                // Append count and buffer appropriate storage
+                                self.rows_logs += eff;
+                                if rows > 0 {
+                                    self.current_logs.push(rec);
+                                } else {
+                                    // zero-row conversion -> keep original pdata for passthrough
+                                    self.passthrough.push(data.clone());
```
lquerel: See my previous comment

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+                                    self.flush_traces(effect).await
+                                } else {
+                                    Ok(())
+                                }
+                            }
+                            SignalType::Metrics => {
+                                // Metrics conversion from OTLP bytes not supported yet -> drop
+                                effect.info(LOG_MSG_DROP_NON_OTAP).await;
+                                Ok(())
+                            }
```
lquerel: No sure to understand the logic here. 
We probably need to 1) log the error, 2) use the passthrough mechanism to send the received pdata to the next downstream node.

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+use otap_df_engine::message::Message;
+use otap_df_engine::node::NodeId;
+use otap_df_engine::processor::ProcessorWrapper;
+use serde::Deserialize;
+use serde_json::Value;
+use std::num::NonZeroU64;
+use std::sync::Arc;
+// For optional conversion during flush/partitioning
+use otel_arrow_rust::otap::OtapArrowRecords;
+use otel_arrow_rust::otap::batching::make_output_batches as low_make_output_batches;
```
lquerel: nit: the low_ prefix is a bit confusing :-)

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: Ah yes I see that now. The state in the batch processor looks much better now. Thanks!

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+
+/// Local (!Send) OTAP batch processor
+pub struct OtapBatchProcessor {
+    config: Config,
+    // Per-signal buffers of convertible OTAP records
+    current_logs: Vec<OtapArrowRecords>,
+    current_metrics: Vec<OtapArrowRecords>,
+    current_traces: Vec<OtapArrowRecords>,
+    // For inputs that cannot be converted to OTAP (currently dropped with a log)
+    passthrough: Vec<OtapPdata>,
```
albertlockett: I wonder if this passthrough behvaiour is something that's needed long term? Is this something the go collector does when it fails to compute a batch?

I understand this would maybe be desirable for metrics for now, as currently converting from OTLP->OTAP for metrics will always fail (it's not implemented yet), so it's possible we just want to keep this passthrough as a temporary workaround until this is fixed (maybe with an error log as suggested by @lquerel here: https://github.com/open-telemetry/otel-arrow/pull/1015/files?diff=unified&w=0#r2307939743)

Generally though, I think we'd only expect conversion to OTAP to fail if the batch was actually not valid. For example, if the OTLP bytes that we convert from proto serialization, or the OTAP bytes we convert from arrow IPC, had garbage data. In these cases, I'm not entirely sure passing the data along is the correct approach

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: I added the same type of factory function that does use Arc to wrap NodeConfig  like the rest of the processors do, maybe there is a nuance I am missing?

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: removed

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+    ) -> Result<ProcessorWrapper<OtapPdata>, ConfigError> {
+        // Allow Go-style timeout strings; parse to milliseconds before deserialization.
+        let mut cfg_norm = cfg.clone();
+        if let Some(s) = cfg.get("timeout").and_then(|v| v.as_str()) {
+            if let Some(ms) = parse_duration_ms(s) {
+                if let Some(obj) = cfg_norm.as_object_mut() {
+                    let _ = obj.insert("timeout".to_string(), Value::from(ms));
+                }
+            }
+        }
```
albertlockett: Do we need to support both timeout as millisecond and parsing in human readable format? I think it might just be better to support parsing from human readable to simplify things.

You could change the definition of `timeout` in the config object to:
```rs
    /// Flush non-empty batches on this interval.
    #[serde(with = "humantime_serde")]
    pub timeout: Option<Duration>,
```
And then the value would automaticaly get parsed into a `std::time::Duration` when you call `serde_json` below.

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: Doesn't look like this is used anymore. Maybe we can just remove it from the struct?

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: looks like this was cleaned up, thanks @daviddahl

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+        match msg {
+            Message::Control(ctrl) => {
+                match ctrl {
+                    otap_df_engine::control::NodeControlMsg::TimerTick { .. } => {
+                        // Follow OTLP semantics: only flush on timer if a size threshold has been reached
+                        if self.size_triggers_emission() {
+                            self.flush_current(effect).await
+                        } else {
+                            Ok(())
+                        }
```
albertlockett: Is this following the OTLP semantics? This doesn't line up with the behaviour of the current OTLP processor which flushes on a timeout:
https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/otlp/src/otlp_batch_processor.rs#L432

FWIW, the golang processor uses timeout in the same way as our OTLP batch prcoessor, which will flush the batch of any size if the timeout is reached:
https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/batchprocessor
> timeout (default = 200ms): Time duration after which a batch will be sent regardless of size. If set to zero, send_batch_size is ignored as data will be sent immediately, subject to only send_batch_max_size.

I think we should follow this same behaviour in this processor. Otherwise, if we only flush once the size is reached, it may be a very long time before we flush if the batch size is high and the rate of telemetry signals is low.

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+            Message::PData(data) => {
+                let max = self.config.send_batch_max_size;
+                match OtapArrowRecords::try_from(data.clone()) {
+                    Ok(rec) => {
+                        let rows = rec.batch_length();
+                        // Per-signal policy: drop zero-row; pre-flush if exceeding max; append rows; flush on max or send_batch_size.
+                        match &rec {
+                            OtapArrowRecords::Logs(_) => {
+                                if rows == 0 {
+                                    effect.info(LOG_MSG_DROP_NON_OTAP).await;
```
albertlockett: This `LOG_MSG_DROP_NON_OTAP ` message is:
```
"OTAP batch processor: dropping non-OTAP message (conversion failed)"
```
But that's not actually what happened here, right? It's not that we failed to convert to OTAP, it's that the batch was  0 length.

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+            config.send_batch_size = MIN_SEND_BATCH_SIZE;
+        }
+        // Normalize 0 -> send_batch_size
+        let effective_sbs = config.send_batch_size;
+        let max = if config.send_batch_max_size == FOLLOW_SEND_BATCH_SIZE_SENTINEL {
+            effective_sbs
+        } else {
+            config.send_batch_max_size
+        };
+        config.send_batch_max_size = max;
```
albertlockett: I'm not sure the logic for setting these defaults is correct. Can you double check it?

FYI the analogous config options for the golang batch processor are described like:
I'm not sure this initialization is correct. The analogous options on the golang batch provessor are described like this:
> - `send_batch_size` (default = 8192): Number of spans, metric data points, or log records after which a batch will be sent regardless of the timeout. send_batch_size acts as a trigger and does not affect the size of the batch. If you need to enforce batch size limits sent to the next component in the pipeline see send_batch_max_size.
> - `send_batch_max_size` (default = 0): The upper limit of the batch size. 0 means no upper limit of the batch size. This property ensures that larger batches are split into smaller units. It must be greater than or equal to send_batch_size.

Also, note the comments on `SendBatchSize` here:
https://github.com/open-telemetry/opentelemetry-collector/blob/826dfcda3a85c5585bea2e7cd9ffccfe763bdd9e/processor/batchprocessor/config.go#L21C1-L24C55
 ```go
	// SendBatchSize is the size of a batch which after hit, will trigger it to be sent.
	// When this is set to zero, the batch size is ignored and data will be sent immediately
	// subject to only send_batch_max_size.
```

I think `send_batch_size` should be allowed to be 0. Probably what we want to do instead is make this field a type of `Option<usize>` on the `Config` struct, and implement the default initializer like:
```rs
fn default_send_batch_size() -> Option<usize> {
    Some(DEFAULT_SEND_BATCH_SIZE)
}
```

I don't think we should initialize `send_batch_max_size` to the value of `send_batch_size` if it's 0. Instead, I think we check if the value is non-zero, and if so and if `send_batch_size` is `Some`, then we should just check that `send_batch_max_size` is greater than `send_batch_size` and if not return `InvalidUserConfig` error.

WDYT? does that sound alright?

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,922 @@
...
+                                    self.rows_logs += rows;
+                                    self.current_logs.push(rec);
+                                    if (max > FOLLOW_SEND_BATCH_SIZE_SENTINEL
+                                        && self.rows_logs >= max)
+                                        || self.rows_logs >= self.config.send_batch_size
+                                    {
+                                        self.flush_logs(effect).await
+                                    } else {
+                                        Ok(())
+                                    }
```
albertlockett: This is kind of a continuation of this comment related to the semantics of `send_batch_size`. https://github.com/open-telemetry/otel-arrow/pull/1015/files#r2310873042

I don't think we need the second check `max > FOLLOW_SEND_BATCH_SIZE_SENTINEL && self.rows_logs >= max` because we've already done a check just above `max > FOLLOW_SEND_BATCH_SIZE_SENTINEL && self.rows_logs + rows > max` then added `rows` to `self.rows_logs`.

Also, if the user has explicitly set the batch size to zero, I don't think we want to flush here. So if/when config.send_batch_size type is changed to `Option<usize>`, we can probably simplify the second `if` block to:
```rs
if matches!(send_batch_size, Some(s) if rows > s) {
    self.flush_logs(effect).await
} else {
    Ok(())
}
```

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: Sounds good. Thank you for seeing this bug / incongruency

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: The first check is 

`flush if the incoming record would exceed max`

And the second check is:

 `flush when we hit the boundary exactly (>= max), and also honor send_batch_size as a trigger (0 => immediate).`

Does that make sense?

albertlockett: ah OK, I think I understand now. Thanks for clarifying

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: @daviddahl will you do this as part of this PR, or will it be a followup?

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: Thanks I see this is removed!

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: nice, looks like this is handled now. Thanks @daviddahl

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,1159 @@
...
+enum FlushReason {
+    Size,
+    Timer,
+    Shutdown,
+}
+
+/// Decide whether to request splitting from upstream batching.
+/// Returns Some(max) only when a split is needed and safe, otherwise None.
+fn requested_split_max(input: &[OtapArrowRecords], max_val: usize) -> Option<NonZeroU64> {
+    // Workaround: avoid splitting a single record until upstream bug is fixed.
```
albertlockett, line 144: I assume there's a bug in `make_output_batches`? Could be related to this https://github.com/open-telemetry/otel-arrow/issues/1027 ?

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: Yeah, the factory wasn't passing the full `Arc<NodeUserConfig>` as it should have been. This has been fixed in https://github.com/open-telemetry/otel-arrow/pull/1051

You'll you might need to change the signature of this method to:
```rs
    pub fn from_config(
        node: NodeId,
        node_config: Arc<NodeUserConfig>,
        proc_cfg: &ProcessorConfig,
    ) -> Result<ProcessorWrapper<OtapPdata>, ConfigError> {
```

You'll still have access to the `&serde_json::Value` by accessing `node_config.config`

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: Cool, I can see the behaviour in the timer tick has been corrected. Thanks for making this change @daviddahl

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: thanks for correcting the error message

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,1159 @@
...
+use otel_arrow_rust::otap::batching::make_output_batches;
+
+/// URN for the OTAP batch processor
+pub const OTAP_BATCH_PROCESSOR_URN: &str = "urn:otap:processor:batch";
+
+/// Default configuration values (parity-aligned as we confirm Go defaults)
+pub const DEFAULT_SEND_BATCH_SIZE: usize = 8192;
+/// Default upper bound on batch size used to chunk oversized inputs (in number of items).
+/// See Config::send_batch_max_size for normalization semantics.
+pub const DEFAULT_SEND_BATCH_MAX_SIZE: usize = 8192;
```
albertlockett: Should the default be zero?
> send_batch_max_size (default = 0): The upper limit of the batch size. 0 means no upper limit of the batch size. This property ensures that larger batches are split into smaller units. It must be greater than or equal to send_batch_size.

https://github.com/open-telemetry/opentelemetry-collector/tree/main/processor/batchprocessor

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: Nice, I see we've fixed the initialization logic of these values. Thanks @daviddahl !

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: i think so.

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: Instead of silent drop, we could send a Nack or route to a dead-letter/error sink? Not sure about this either? Do we go back to a passthrough mechanism but only for metrics until the conversion from OTLP -> OTAP is operational?

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: I added a `dirty` flag for each telemetry type and logic to make the flush happen when we have buffered telemetry

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: -  Zero-row batches now log LOG_MSG_DROP_EMPTY ("OTAP batch processor: dropping empty OTAP record (0 rows)") and are dropped.
- Conversion failures log LOG_MSG_DROP_CONVERSION_FAILED ("OTAP batch processor: dropping message: OTAP conversion failed") and are dropped.
- The old LOG_MSG_DROP_NON_OTAP is gone now

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: Updated the `default_send_batch_size` to be `Option<usize>` and follow the Golang implementation logic

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: yes. updating this now.

daviddahl: BTW, I seem to think there might have been conflicting documentations for these values.

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: i have updated to follow these semantics

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: updated to use NodeUserConfig correctly (now)

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: Just doing drop and log on error - there is still an open question about handling metrics with a `passthrough` mechanism, but for now it has been simplified to just log the errors

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
daviddahl: updated to use string parsing of this value like go impl

albertlockett: Looks good @daviddahl , thanks for making this change!

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: import name was fixed

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: I see you changed to 0! Thanks for the correction

Do you have a reference to the conflicting docs? Maybe we could fix it

It seems like in the code the default is documented as 0 as well:
https://github.com/open-telemetry/opentelemetry-collector/blob/5af9b6f62bc60230a84bce7d5d0154f944bf3121/processor/batchprocessor/config.go#L26-L29

Also when creating the default config, it gets set to 0 (not setting the value in golang should use the default value of 0)
https://github.com/open-telemetry/opentelemetry-collector/blob/5af9b6f62bc60230a84bce7d5d0154f944bf3121/processor/batchprocessor/factory.go#L38-L44

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,1272 @@
...
+    pub send_batch_size: Option<usize>,
+    /// Hard cap for splitting very large inputs.
+    /// Go behavior: 0 means no upper limit; missing defaults to 0 (unlimited).
+    #[serde(default = "default_send_batch_max_size")]
+    pub send_batch_max_size: usize,
+    /// Flush non-empty batches on this interval.
+    #[serde(with = "humantime_serde", default = "default_timeout_duration_opt")]
+    pub timeout: Option<Duration>,
+    /// Optional metadata partitioning keys (resource/scope/attribute names)
+    #[serde(default)]
```
lquerel:
```suggestion
    /// Optional metadata partitioning keys (resource/scope/attribute names). Not yet supported.
    /// ToDo: Support metadata-aware batching.
    #[serde(default)]
```

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,1272 @@
...
+    pub metadata_keys: Vec<String>,
+    /// Optional limit on the number of distinct metadata-based groups this processor will track.
+    ///
+    /// Note: This is currently a no-op because grouping by metadata_keys has not yet been
+    /// implemented. When grouping lands, this will cap the number of concurrent groups and the
+    /// overflow strategy will be documented.
+    #[serde(default)]
```
lquerel:
```suggestion
    /// overflow strategy will be documented. Not yet supported.
    /// ToDo: Support metadata-aware batching
    #[serde(default)]
```

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -0,0 +1,1272 @@
...
+    pub metadata_cardinality_limit: Option<usize>,
+}
+
+#[cfg(test)]
```
lquerel: Move all the tests at the end.

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: OK thanks David. I think what we have now looks good.!

In fact, upon further inspecting how the golang code works, I think what we had before was would have worked too, so I'm sorry for the confusion!.

I realize now that when I said "some threshold", I wasn't very clear, (again, sorry sorry for that). I was thinking about how in  `otlp_batch_processor` we keep these `last_update_...` `Instant`s 
https://github.com/open-telemetry/otel-arrow/blob/d25cb0dbcf66c1633451d848c97328c6b7ca02a0/rust/otap-dataflow/crates/otlp/src/otlp_batch_processor.rs#L352-L354
And we check them when we flush:
https://github.com/open-telemetry/otel-arrow/blob/d25cb0dbcf66c1633451d848c97328c6b7ca02a0/rust/otap-dataflow/crates/otlp/src/otlp_batch_processor.rs#L431-L453

However, It looks like the golang processor just works differently -- it runs a timer that will flush every `timeout` period regardless of when the batches were received, which is pretty much the same as what we're now doing:
https://github.com/open-telemetry/opentelemetry-collector/blob/5af9b6f62bc60230a84bce7d5d0154f944bf3121/processor/batchprocessor/batch_processor.go#L190-L258

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
albertlockett: Great thanks!

albertlockett (approved):
Thanks for responding to the feedback on this @daviddahl ! Looks good now

daviddahl:
@albertlockett looks like your git incantations have un-borked my branch. Thank you very much!

Issue #509, opened 2025-05-30
Retry Processor


lquerel:
The purpose of this processor is to generalize support for the retry mechanism for recoverable errors. This approach eliminates the need to implement retry logic within exporters and other components that require this functionality.

Benefits: a single configuration, one unified retry logic, less work for exporters.

This processor will utilize the concept of control messages of types `Ack` and `Nack`. An exporter that fails to process a message must emit a `Nack` control message (see details of the fields below). When message processing succeeds, the exporter must emit an `Ack` control message.

Definition of the Ack and Nack messages.

```
pub enum ControlMessage {
    Ack {
        msg_ids: Vec<String>,
    },
    Nack {
        msg_ids: Vec<String>,
        error: ErrorDetail,
        retry_policy: Option<RetryPolicy>,
    },
}

pub struct RetryPolicy {
    pub retry_delay_ms: u64,
}

pub struct ErrorDetail {
    pub recoverable: bool,
    pub reason: String,
}
```

daviddahl:
@lquerel 
I started working on a POC for the retryprocessor.  if you have some time, your additional details would really help. One thing I was wondering is - will this work for either OTLP or OTAP data from day one? I started with just a processor that will handle OTLP, but, am wondering if we need a trait that describes its API and 2 implementations for either OTLP or OTAP. 

What is the elegant way to begin?

PR #1017, opened 2025-08-27
Split Error and TypedError<T>; generalize NodeControlMsg<PData>

jmacd:
Part of #509 

While working to implement the rest of the retry_processor support, discovered a need to add `<PData>` to the `NodeControlMsg`. The `Nack` message now includes an optional `pdata` field.

At the same time, most uses of Error did not require a `<T>`, so this separates them and handles conversion to and from the `TypedError<T>` form. The type is erased and the error message is stored as a string.

No functional changes except w.r.t. error structure and a new `Option<Box<PData>>` in the Nack. While these changes are unrelated, they have approximately the same impact and combining them was relatively easy.


rust/otap-dataflow/crates/engine/src/control.rs
```diff
@@ -32,6 +32,12 @@ pub enum NodeControlMsg {
         id: u64,
         /// Human-readable reason for the NACK.
         reason: String,
+
+        /// Placeholder for optional return value, making it possible for the
+        /// retry sender to be stateless.
+        ///
+        /// TODO: should this be boxed to reduce the size of NodeControlMsg?
+        rvalue: Option<PData>,
```
lquerel: Any particular reason not naming this field pdata?

rust/otap-dataflow/crates/engine/src/error.rs
```diff
@@ -6,30 +6,26 @@
...
-pub enum Error<T> {
-    /// A wrapper for the config errors.
-    #[error("A config error occurred: {0}")]
-    ConfigError(#[from] Box<otap_df_config::error::Error>),
-
-    /// A wrapper for the channel errors.
-    #[error("A channel error occurred: {0}")]
-    ChannelRecvError(#[from] otap_df_channel::error::RecvError),
-
+pub enum ErrorT<T> {
```
lquerel: I don't have a better name to suggest for now, but if we find a better one we shouldn't hesitate :-)

rust/otap-dataflow/crates/engine/src/error.rs
```diff
@@ -38,9 +34,49 @@ pub enum Error<T> {
...
+    #[error("A config error occurred: {0}")]
+    ConfigError(#[from] Box<otap_df_config::error::Error>),
+
+    /// A wrapper for the channel errors.
+    #[error("A channel error occurred: {0}")]
+    ChannelRecvError(#[from] otap_df_channel::error::RecvError),
+
+    /// A wrapper for the channel errors.
+    #[error("A pipeline data channel error occurred")]
+    ChannelSendError,
```
lquerel: Erasing the `T` at this level is correct, but I think we should add a `reason` or `message` field of type string in which we would place the `to_string` of the initial error.

rust/otap-dataflow/crates/engine/src/error.rs
```diff
@@ -38,9 +34,49 @@ pub enum Error<T> {
...
+
+    /// A wrapper for the pipeline control message send errors.
+    #[error("A pipeline control channel error occurred")]
+    PipelineControlMsgError,
```
lquerel: The same is true here.

lquerel (approved):
LGTM. Made few minor suggestions

rust/otap-dataflow/crates/engine/src/control.rs
jmacd: Renamed `pdata`.

rust/otap-dataflow/crates/engine/src/error.rs
jmacd: Renamed `TypedError`.

rust/otap-dataflow/crates/engine/src/error.rs
jmacd: Done.

rust/otap-dataflow/crates/engine/src/error.rs
jmacd: Done.

albertlockett (approved):

PR #1022, opened 2025-08-28
[perf] Micro optimizations for Perf Exporter

utpilla:
## Changes
- Cache `1.0 - smoothing_factor` used in `update_average` method
- Use a single Fused Multiply Add operation instead of two arithmetic operations in `update_average` method
- Update `decode_timestamp` method to directly work on bytes instead of allocating strings and doing UTF-8 conversions

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/perf_exporter/exporter.rs
+++ b/rust/otap-dataflow/crates/otap/src/perf_exporter/exporter.rs
@@ -49,6 +49,7 @@ pub const OTAP_PERF_EXPORTER_URN: &str = "urn:otel:otap:perf:exporter";
 pub struct PerfExporter {
     config: Config,
     metrics: MetricSet<PerfExporterPdataMetrics>,
+    smoothing_complement: f64,
 }
 
 /// Declares the OTAP Perf exporter as a local exporter factory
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/otap/src/perf_exporter/exporter.rs
```diff
@@ -217,21 +224,38 @@ impl local::Exporter<OtapPdata> for PerfExporter {
...
+    smoothing_factor: f64,
+    smoothing_complement: f64,
+) -> f64 {
     // update the average using a exponential moving average which allows new data points to have a greater impact depending on the smoothing factor
-    smoothing_factor * new_value + (1.0 - smoothing_factor) * old_average
+    smoothing_factor.mul_add(new_value, smoothing_complement * old_average)
 }
 
 /// decodes the byte array from the timestamp header and gets the equivalent duration value
+/// Expects format: "{seconds}:{nanoseconds}" as ASCII digits
```
utpilla, line 239: Is it okay to expect that timestamp bytes will be valid?

lquerel:
I commented this part of the PerfExporter because we need to revisit the concept of header in the pdata (following the last evolutions around the pdata).

jmacd (approved):
Cool!

jmacd:
We can re-open -- the now commented out section of code can be upgraded in a few ways, overall the goal is to measure the latency of the pipeline-under-test.

PR #1023, opened 2025-08-28
OTLP and OTAP Exporter Metrics

lquerel:
This PR is a first iteration to progressively comply with this [RFC - Pipeline Component Telemetry](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md). 

The OTLP and OTAP exporters have been updated.

The following metric set has been used:
```rust
/// PData metrics for exporters.
/// Namespace grouped under `exporter.pdata`.
/// Uses a monotonic counter per outcome.
#[metric_set(name = "exporter.pdata")]
pub struct ExporterPDataMetrics {
    /// Number of pdata metrics consumed by this exporter.
    #[metric(unit = "{msg}")]
    pub metrics_consumed: Counter<u64>,

    /// Number of pdata metrics successfully exported.
    #[metric(unit = "{msg}")]
    pub metrics_exported: Counter<u64>,

    /// Number of pdata metrics that failed to be exported.
    #[metric(unit = "{msg}")]
    pub metrics_failed: Counter<u64>,

    /// Number of pdata logs consumed by this exporter.
    #[metric(unit = "{msg}")]
    pub logs_consumed: Counter<u64>,

    /// Number of pdata logs successfully exported.
    #[metric(unit = "{msg}")]
    pub logs_exported: Counter<u64>,

    /// Number of pdata logs that failed to be exported.
    #[metric(unit = "{msg}")]
    pub logs_failed: Counter<u64>,

    /// Number of pdata traces consumed by this exporter.
    #[metric(unit = "{msg}")]
    pub traces_consumed: Counter<u64>,

    /// Number of pdata traces successfully exported.
    #[metric(unit = "{msg}")]
    pub traces_exported: Counter<u64>,

    /// Number of pdata traces that failed to be exported.
    #[metric(unit = "{msg}")]
    pub traces_failed: Counter<u64>,
}
```


rust/otap-dataflow/crates/otap/src/otap_exporter.rs
```diff
@@ -109,6 +112,12 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
                 self.config.grpc_endpoint
             ))
             .await;
+
+        let exporter_id = effect_handler.exporter_id();
+        let _ = effect_handler
```
utpilla, line 117: Should we get hold of the `TelemetryTimerCancelHandle` returned here and use it to cancel telemetry timer when processing `NodeControlMsg::Shutdown`?

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
```diff
@@ -152,14 +161,25 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
...
                     break;
                 }
                 //send data
-                Message::PData(message) => {
-                    let message: OtapArrowBytes = message.try_into()?;
+                Message::PData(pdata) => {
+                    // Capture signal type before moving pdata into try_from
+                    let signal_type = pdata.signal_type();
+
+                    self.pdata_metrics.inc_consumed(signal_type);
```
utpilla, line 179: It looks like we are executing quite a bit of `match` expressions here which could be optimized:
- `pdata.signal_type()` would involve two `match` expressions
- `inc_consumed()` would involve an additional `match` expression

We anyway have to match on `message`, so we could simply use the signal specific consumed metric such as `metrics_consumed` instead of calling the helper method `inc_consumed()` which would require three `match` expressions.

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
```diff
@@ -174,21 +194,29 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
...
-                                    exporter: effect_handler.exporter_id(),
-                                    error: error.to_string(),
+                                .map_err(|e| {
+                                    self.pdata_metrics.metrics_failed.inc();
+                                    Error::ExporterError {
+                                        exporter: exporter_id.clone(),
+                                        error: e.to_string(),
+                                    }
                                 })?;
+                            self.pdata_metrics.metrics_exported.inc();
```
albertlockett, line 204: This metric is tracking the number of OTAP batches. Is that what we want, or should it be the number of logs (e.g. number of rows in the `Logs` record batch)?

lquerel: Here I want to compute the number of batches.

But interestingly the logic to compute the number of logs, metrics, spans which just after was entirely wrong... I misused the try_info and converted in the pdata into OtapArrowBytes instead of OtapArrowRecord and then used the len() method to get the number of signal. That was obviously wrong.

We probably need to make the pdata API a bit simpler.

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
lquerel: I made some changes this morning which not related to this comment but the result is much simpler.

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
lquerel: Sounds good. Fixed

rust/otap-dataflow/crates/otap/src/perf_exporter/exporter.rs
```diff
@@ -171,36 +168,40 @@ impl local::Exporter<OtapPdata> for PerfExporter {
...
                     // calculate average
-                    let mut decoder = Decoder::new();
-                    let header_list =
-                        decoder
-                            .decode(&batch.headers)
-                            .map_err(|_| Error::ExporterError {
-                                exporter: effect_handler.exporter_id(),
-                                error: "Failed to decode batch headers".to_owned(),
-                            })?;
+                    // ToDo Temporary disable latency calculation until we have a better way to add the timestamp header
```
jmacd, line 171: I wonder about using client-side measurements to convey the latency, instead. Maybe in an ideal world, the load generator clients write their latency metrics to the perf exporter via a metrics channel.

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
```diff
@@ -134,7 +143,7 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
                     error: error.to_string(),
                 })?;
 
-        if let Some(compression) = self.config.compression_method {
+        if let Some(ref compression) = self.config.compression_method {
```
jmacd, line 146: Ah! I have this change pending as well.

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
jmacd: Also/anyway I intend to continue refining the logic of translating success/failure into metrics. I am not familiar with the optimization potential / loss of match statements, I admit.

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
jmacd: Ideally, we are able to count (and limit) by at least four weights (requests, items, uncompressed bytes and compressed bytes). As a side-note, the "items" count is OTel's standard definition, it's a count of spans, logs records, metric data opints, and profiles.

albertlockett: @jmacd the node on terminology is interesting. We recently added a method to `OtapArrowRecords` called `batch_length` that returns exactly what you describe:
https://github.com/msalib/otel-arrow/blob/13c40695ea5747a1edc1c1bf6f473020c1dc6010/rust/otel-arrow-rust/src/otap.rs#L94-L96

Maybe this should be renamed to `items_count`?

I suppose we could expose this on `OtapPdata` to simplify things as @lquerel  suggested, but just note that we'd probably need to transparently convert the underlying representation to compute this, which would mean it takes a `&mut` receiver, and since the the conversion between pdata representations is fallible, it would need to return a result. It's not exactly the cleanest API imo, but maybe it's OK...
```rs
fn count_items(&mut self) -> Result<usize> {
```

rust/otap-dataflow/crates/otap/src/perf_exporter/exporter.rs
lquerel: Could be an option.

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
lquerel: @albertlockett yea, I'm still a bit undecided about the pdata API, but I feel that what we have is a little too complicated and somewhat error-prone. It's going to be an interesting exercise to discuss.

PR #1024, opened 2025-08-28
Parquet exporter handle optional fields

albertlockett:
part of #863 

Because some OTAP fields are optional, in a stream of record batches we may receive subsequent batches with different schemas. Parquet doesn't support having row groups with different sets of column chunks, which means we need to know the schema a-priori when the writer is created.

This PR adds code to normalize the schema of the record batch before writing by:
- putting all the fields in the same order
- creating all null/default value columns for any missing column

The missing columns should have a small overhead when written to disk, because parquet will either write an entirely empty column chunk for the null column (all null count, no data), or and for all default-value columns, parquet will use dictionary and RLE encoding by default, leading to a small column chunk with a single value value in dict & a single run for the key.

What's unfortunate is that we still materialize an all-null column before writing with the length of the record batch. This can be optimized when run-end encoded arrays are supported in parquet, because we could just create a run array with a single run of null/default value. The arrow community is currently working on adding support (see https://github.com/apache/arrow-rs/pull/7713 & https://github.com/apache/arrow-rs/pull/8069).

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/parquet_exporter/schema.rs
+++ b/rust/otap-dataflow/crates/otap/src/parquet_exporter/schema.rs
@@ -6,25 +6,25 @@
 //!
 //! The main reason we can't write the record batch verbatim is because OTAP considers some columns
 //! optional, but parquet spec requires that each row group contain a column chunk for every field
-//! in the schema. This means we can't receive two consecutive OTAP batches for some payload type
-//! and write them into the same writer.
+//! in the schema (and the column chunks must be in the correct order). 
+//! 
+//! This means we can't receive two consecutive OTAP batches for some payload type and write them 
+//! into the same writer. To  handle this, we insert all null columns for missing columns (or all 
+//! default-value where the column is not nullable), and also arrange the columns so they're always
+//! in the same order.
 //!
 //! Note that although we also switch between the Dictionary and Native encodings, we don't need to
 //! actually convert the existing columns to all be the same type. Parquet writer is able to accept
 //! logically compatible types, which includes compatibility between types like `T` (native array)
 //! and `Dictionary<K, T>` (for any `K`).
-//!
-//! TODO do we also need to handle the mixed-up column order?
-//!
-//! To handle this, we need to add all-null columns for all the optional columns.
 
+use std::iter::repeat_n;
 use std::sync::{Arc, LazyLock};
 
 use arrow::array::{
-    ArrayRef, BinaryArray, FixedSizeBinaryArray, Float32Array, Float64Array, Int64Array,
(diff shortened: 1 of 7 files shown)


jmacd (approved):

rust/otap-dataflow/crates/otap/src/parquet_exporter/schema.rs
```diff
@@ -0,0 +1,926 @@
...
+//! has a schema that's compatible with the Parquet writer.
+//!
+//! The main reason we can't write the record batch verbatim is because OTAP considers some columns
+//! optional, but parquet spec requires that each row group contain a column chunk for every field
+//! in the schema (and the column chunks must be in the correct order).
+//!
+//! This means we can't receive two consecutive OTAP batches for some payload type and write them
+//! into the same writer. To  handle this, we insert all null columns for missing columns (or all
+//! default-value where the column is not nullable), and also arrange the columns so they're always
+//! in the same order.
```
lquerel, line 14: We probably don't have any other simple option to implement to resolve this limitation. However, it's important to keep an eye on the implications this will have, particularly on memory consumption across the parquet -> arrow -> query engine path. It might have no effect, but we'll need to verify that.

lquerel (approved):
LGTM. See my comment.

rust/otap-dataflow/crates/otap/src/parquet_exporter/schema.rs
albertlockett: @lquerel I agree. Related to memory consumption, there's an optimization to switch the placeholders we create with a more efficient representation. We can do this once run-end encoding is supported in Parquet (work in progress).

Say we're adding a placeholder for a missing Int32 column, we'd need to add an all nulls Int32Array which still has a values buffer containing four bytes each row plus the nulls buffer:
```
values buf = [0, 0, ... ] // len = num rows
nulls buf  = [false, false ... ] // len = num rows
```
Eventually we'll change to use a RunArray, which would contain an Int32Array of length 1, plus a single run specifying to use the same value for the entire length:
```
values buf = [0]
nulls buf  = [false]
run ends   = [num_rows]
```
I think we could maybe even reuse the values & nulls buffer across multiple column replacements for the same datatypes.

I will create a followup issue to track this optimization

albertlockett: wrt to querying, I think the behvaiour will depend on the query engine but we can definitely verify. I imagine that generally the overhead of the extra metadata from these additional columns would be small compared to the size of the data. Depending on the query, we'd still probably have empty arrays in memory for these columns if the user selected them.

PR #1032, opened 2025-08-29
[otap-df-otap] Fix OTLP and OTAP receivers

utpilla:
## Problem with the current `tokio::select!` usage

By design, `tokio::select` returns whenever any branch completes and cancels the remaining branches.  We have placed `tokio::select!` within a `loop` and it has two branches: one for processing control messages and one for serving requests. The control message branch would complete anytime we finish processing any control message (for example, receiving shutdown, reporting telemetry, etc.).  The other branch for serving requests would usually not complete until something goes wrong with the `TcpListenerStream` itself - it's meant to run indefinitely.

This means that anytime we report telemetry (or process any other control message) we would return from `tokio::select!` and start the loop again. That leads to:
- Repeated creation of server and related service instances on every control message
- Improper handling of active connections when `tokio::select!` cancels the request serving branch
- Unnecessary resource overhead and potential connection drops

## Changes
- Fix the `tokio::select!` and `loop` construct by removing the outer loop
- Create the server and related service instances only once outside the select
- Update the control message processing branch by wrapping it in an async loop so that reporting telemetry etc. doesn't lead to recreating the server
- Properly handle control message results (shutdown vs errors) in the select branches

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/otap_receiver.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_receiver.rs
@@ -115,45 +115,50 @@ impl shared::Receiver<OtapPdata> for OTAPReceiver {
     ) -> Result<(), Error> {
         // create listener on addr provided from config
         let listener = effect_handler.tcp_listener(self.config.listening_addr)?;
-        let mut listener_stream = TcpListenerStream::new(listener);
+        let listener_stream = TcpListenerStream::new(listener);
 
-        //start event loop
-        loop {
-            //create services for the grpc server and clone the effect handler to pass message
-            let logs_service =
-                ArrowLogsServiceImpl::new(effect_handler.clone(), self.config.message_size);
-            let metrics_service =
-                ArrowMetricsServiceImpl::new(effect_handler.clone(), self.config.message_size);
-            let trace_service =
-                ArrowTracesServiceImpl::new(effect_handler.clone(), self.config.message_size);
+        //create services for the grpc server and clone the effect handler to pass message
+        let logs_service =
+            ArrowLogsServiceImpl::new(effect_handler.clone(), self.config.message_size);
+        let metrics_service =
+            ArrowMetricsServiceImpl::new(effect_handler.clone(), self.config.message_size);
+        let trace_service =
+            ArrowTracesServiceImpl::new(effect_handler.clone(), self.config.message_size);
 
-            let mut logs_service_server = ArrowLogsServiceServer::new(logs_service);
-            let mut metrics_service_server = ArrowMetricsServiceServer::new(metrics_service);
-            let mut trace_service_server = ArrowTracesServiceServer::new(trace_service);
+        let mut logs_service_server = ArrowLogsServiceServer::new(logs_service);
+        let mut metrics_service_server = ArrowMetricsServiceServer::new(metrics_service);
+        let mut trace_service_server = ArrowTracesServiceServer::new(trace_service);
 
-            // apply the tonic compression if it is set
-            if let Some(ref compression) = self.config.compression_method {
-                let encoding = compression.map_to_compression_encoding();
+        // apply the tonic compression if it is set
+        if let Some(ref compression) = self.config.compression_method {
+            let encoding = compression.map_to_compression_encoding();
 
-                logs_service_server = logs_service_server
-                    .send_compressed(encoding)
-                    .accept_compressed(encoding);
-                metrics_service_server = metrics_service_server
-                    .send_compressed(encoding)
-                    .accept_compressed(encoding);
-                trace_service_server = trace_service_server
-                    .send_compressed(encoding)
-                    .accept_compressed(encoding);
-            }
+            logs_service_server = logs_service_server
+                .send_compressed(encoding)
+                .accept_compressed(encoding);
+            metrics_service_server = metrics_service_server
+                .send_compressed(encoding)
+                .accept_compressed(encoding);
+            trace_service_server = trace_service_server
+                .send_compressed(encoding)
+                .accept_compressed(encoding);
+        }
 
-            tokio::select! {
-                biased; //prioritize ctrl_msg over all other blocks
-                // Process internal event
-                ctrl_msg = ctrl_msg_recv.recv() => {
-                    match ctrl_msg {
+        let server = Server::builder()
+            .add_service(logs_service_server)
+            .add_service(metrics_service_server)
+            .add_service(trace_service_server);
+
+        tokio::select! {
+            biased; //prioritize ctrl_msg over all other blocks
+
+            // Process internal events
+            ctrl_msg_result = async {
+                loop {
+                    match ctrl_msg_recv.recv().await {
                         Ok(NodeControlMsg::Shutdown {..}) => {
                             // ToDo: add proper deadline function
-                            break;
+                            return Ok(());
                         },
                         Err(e) => {
                             return Err(Error::ChannelRecvError(e));
@@ -163,20 +168,19 @@ impl shared::Receiver<OtapPdata> for OTAPReceiver {
                         }
                     }
                 }
-                // Poll the grpc server
-                result = Server::builder()
-                .add_service(logs_service_server)
-                .add_service(metrics_service_server)
-                .add_service(trace_service_server)
-                .serve_with_incoming(&mut listener_stream)=> {
-                    if let Err(error) = result {
-                        // Report receiver error
-                        return Err(Error::ReceiverError{receiver: effect_handler.receiver_id(), error: error.to_string()});
-                    }
+            } => {
+                return ctrl_msg_result;
+            },
+
+            // Run server
+            result = server.serve_with_incoming(listener_stream) => {
+                if let Err(error) = result {
+                    // Report receiver error
+                    return Err(Error::ReceiverError{receiver: effect_handler.receiver_id(), error: error.to_string()});
                 }
             }
         }
-        //Exit event loop
+
         Ok(())
     }
 }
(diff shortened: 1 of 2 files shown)


lquerel:
@utpilla The approach followed by the OTLP and OTAP receivers is indeed not ideal for the reasons you mentioned. However, I'm not convinced by the proposed remedy. The standard/recommended way to use `select!` is a loop wrapping the `select!`, with branches that properly support cancellation.

The proposal here is to push the event loop inside the branches, which in reality leads to having multiple event loops. The main problem with this approach is that the first branch to return `Ready` "wins" and prevents the other branch from making progress. If that branch is, in some sense, an infinite loop, then the other will never get a turn. That's why the loop is on the outside, with branches competing for the next event to process (control messages, or data arriving from a gRPC connection).

In my view, the way to fix the current approach is to move the creation of the gRPC server out (like you did), pin it, and then have the `select!` inside the loop.

```rust
let shutdown = CancellationToken::new();
let server_fut = Server::builder()
            .add_service(logs_service_server)
            .add_service(metrics_service_server)
            .add_service(trace_service_server)
            .serve_with_incoming_shutdown(incoming, shutdown.cancelled());

tokio::pin!(server_fut);

loop {
  biased; //prioritize ctrl_msg over all other blocks
  // Process internal event
  ctrl_msg = ctrl_msg_recv.recv() => match ctrl_msg { ... on Shutdown ctrl msg we can cancel on shutdown cancelation token ... },
  res = &mut server_fut => {
    // do something with the result
    break; // we are done
  }
}
```

utpilla:
> The proposal here is to push the event loop inside the branches, which in reality leads to having multiple event loops. The main problem with this approach is that the first branch to return `Ready` "wins" and prevents the other branch from making progress. If that branch is, in some sense, an infinite loop, then the other will never get a turn.

That's not the case. In my approach, I do have a `loop` within the `async` block but it will yield anytime there are no control messages in the channel. When that happens, `tokio::select` could poll the server branch. For example, if you were to run this test with `--no-capture`:

```rust
#[tokio::test]
async fn test_tokio_select() {
    tokio::select! {
        biased; //prioritize ctrl_msg over all other blocks

        // Process internal events
        _ = async {
            loop {
                tokio::time::sleep(Duration::from_millis(1000)).await;
                println!("In control message loop");
            }
        } => {},

        // Run server
        _ = async {
            loop {
                tokio::time::sleep(Duration::from_millis(1000)).await;
                println!("In server loop");
            }
        } => {},
    }
}
```

You would see the following output which confirms that the presence of a `loop` within a `tokio::select!` branch does not starve the other branches as long as it yields from time to time.

```shell
running 1 test
Starting test_tokio_select
In control message loop
In server loop
In control message loop
In server loop
In control message loop
In server loop
In control message loop
In server loop
In control message loop
In server loop
...
```

utpilla:
> In my view, the way to fix the current approach is to move the creation of the gRPC server out (like you did), pin it, and then have the `select!` inside the loop.
> 
> ```rust
> let shutdown = CancellationToken::new();
> let server_fut = Server::builder()
>             .add_service(logs_service_server)
>             .add_service(metrics_service_server)
>             .add_service(trace_service_server)
>             .serve_with_incoming_shutdown(incoming, shutdown.cancelled());
> 
> tokio::pin!(server_fut);
> 
> loop {
>   biased; //prioritize ctrl_msg over all other blocks
>   // Process internal event
>   ctrl_msg = ctrl_msg_recv.recv() => match ctrl_msg { ... on Shutdown ctrl msg we can cancel on shutdown cancelation token ... },
>   res = &mut server_fut => {
>     // do something with the result
>     break; // we are done
>   }
> }
> ```

This approach still has the issue of dropping active connections when processing a control message. `tokio::select!` would cancel the sever branch if the control message branch wins. We would still have the problem where reporting telemetry leads to dropping active connections.

More importantly, by reusing `server_fut`, we run this risk of polling a potentially cancelled future, which doesn't seem right,

utpilla:
> > @utpilla The approach followed by the OTLP and OTAP receivers is indeed not ideal for the reasons you mentioned. However, I'm not convinced by the proposed remedy. The standard/recommended way to use `select!` is a loop wrapping the `select!`, with branches that properly support cancellation.

I understand that in this PR we would not be using `tokio::select!` in the conventional way but it solves the problems with our receivers. It also seems better for performance as we don't have the overhead of repeated `tokio::select!` creation.

lquerel:
In my proposal, I'd be very surprised if the active gRPC connections were closed. If I'm wrong, my apologies and that would mean tonic isn't cancellation safe, which would surprise me, but who knows.

Generally speaking, just because a future is canceled in the second branch of the `select!` doesn’t mean the gRPC server's connections are closed. Similarly, just because the receiver side (recv method) of a channel is canceled doesn't mean the channel is closed. In fact, if these two futures are cancellation safe (as I believe they are) then they retain their state after cancellation and are ready to continue processing on the next loop iteration.

In the following example the socket is not close each time we iterate over the `select!` (and no data loss).

```rust
let mut interval = time::interval(time::Duration::from_secs(1));

let addr: SocketAddr = "0.0.0.0:5555".parse().unwrap();
let listener = TcpListener::bind(addr).await?;
eprintln!("Listening on {addr}");

loop {
  tokio::select! {
     _ = interval.tick() => {
          println!("ping");
      }

      Ok((stream, peer)) = listener.accept() => {
           tokio::spawn(async move {
                 if let Err(err) = handle_conn(stream, peer).await {
                        eprintln!("[{peer}] error: {err}");
                  }
            });
         }
    }
 }
 ```

Also AFAIK, the macro `select!` is mostly expanded as code without allocation on the heap so I don't believe you will notice a real overhead here. 

I should take more time to think through your approach and its implications, but what I do know is that it uses `select!` in a non "academic" way. Initially, I even thought it had a head-of-line blocking problem, which, according to the results of your test, doesn't seem to be the case, at least with this simple example.

**Comment by utpilla** [comment]
> In my proposal, I'd be very surprised if the active gRPC connections were closed. If I'm wrong, my apologies and that would mean tonic isn't cancellation safe, which would surprise me, but who knows.
> 
> Generally speaking, just because a future is canceled in the second branch of the `select!` doesn’t mean the gRPC server's connections are closed. Similarly, just because the receiver side (recv method) of a channel is canceled doesn't mean the channel is closed. In fact, if these two futures are cancellation safe (as I believe they are) then they retain their state after cancellation and are ready to continue processing on the next loop iteration.

I looked into it a bit more and I think your approach should not have those issues. The key thing to note is whether the futures provided in `tokio::select` take ownership of the data or borrow the data. In your approach, we are using `&mut server_fut`, which means dropping this reference shouldn't drop the underlying server, connections etc. It would still be good to confirm though.


> Also AFAIK, the macro `select!` is mostly expanded as code without allocation on the heap so I don't believe you will notice a real overhead here.

The overhead I suspect is not for allocation but about the repeated processing work that comes with each `tokio::select!` which is creating a state machine with the branches, registering the wakers, and selecting the winner branch and canceling/dropping the other one etc. When we execute `tokio::select!` in a loop, we have to do all of this repeatedly.

Having said that, I'm all in for following a pedantic usage of `tokio::select` if we think its maintainability benefits outweigh the performance gains (if any).

**Comment by jmacd** [comment]
I'm not sure I've followed completely, the nuances over tokio::select!

(Moderately related: while I was looking at the matter of Context and back-propagation of responses in the engine/otap crates, I discovered the use of  stream! for the OTAP stream producers, which yield single requests per stream. In the Golang OTAP exporter we keep a stream open and produce a stream of requests through it, but the Rust exporter is exporting one request per stream.)

**Comment by lquerel** [comment]
@utpilla @jmacd Generally speaking, a number of components, such as the OTLP/OTAP receivers/exporters, have served as a learning aid for Rust for some members of my team. There are definitely aspects that need improvement. I was planning to do a round of improvements after the current milestone, but I'm absolutely open to any improvements, including before this milestone.

@utpilla Now, regarding the conversation about `select!`, I'm not used to this way of using `select!`, but that doesn't mean it's bad :-) I'm okay with you trying to put this in place so we can see how it turns out. Ideally, it would be good to have some kind of benchmark with both approaches (yours and the one I proposed).

**Comment by albertlockett** [comment]
@utpilla @lquerel It looks like you guys are both circling around really similar solutions and the only difference is whether to put the control loop inside tokio select, or on the outside, and really I think the only difference is the overhead of invoking `tokio_select!` multiple times if the loop was on the outside.
 
I don't think using the loop inside the branch is unidiomatic. The `tokio_select!` docs use this pattern as well. Check out the second example in the Syntax section: https://tokio.rs/tokio/tutorial/select#syntax. So I think what @utpilla has suggested is OK

I created a benchmark to try to see if there was some overhead, and it turns out (as we suspected), that calling the loop inside the `tokio_select!` branch seems to be faster than calling `tokio_select!` inside an outer loop. 
```rs
use std::hint::black_box;

use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
use tokio::runtime::Runtime;

async fn do_work(i: i32) -> i32 {
    tokio::task::yield_now().await;
    i + 1
}

fn bench_compare(c: &mut Criterion) {
    let rt = Runtime::new().unwrap();
    let mut group = c.benchmark_group("async_channel");
    let _ = group.bench_function(BenchmarkId::new("select", "loop_inside_branch"), |b| {
        b.to_async(&rt).iter(|| async {
            tokio::select! {
                _ = async move {
                    let mut i = 0;
                    loop {
                        i = do_work(i).await;
                        if i > 1000 {
                            break
                        } else {
                            black_box(i);
                        }
                    }
                } => {}
            }
        });
    });

    let _ = group.bench_function(BenchmarkId::new("select", "loop_outside_select"), |b| {
        b.to_async(&rt).iter(|| async {
            let mut i = 0;
            loop {
                tokio::select! {
                    j = do_work(i) => {
                        i = j
                    }
                }
                if i > 1000 {
                    break;
                } else {
                    black_box(i);
                }
            }
        });
    });

    group.finish();
}

criterion_group!(benches, bench_compare);
criterion_main!(benches);
```

results:
```
async_channel/select/loop_inside_branch
                        time:   [12.477 µs 12.560 µs 12.644 µs]

async_channel/select/loop_outside_select
                        time:   [16.524 µs 16.598 µs 16.671 µs]
```

albertlockett (approved):

utpilla:
> I created a benchmark to try to see if there was some overhead, and it turns out (as we suspected), that calling the loop inside the `tokio_select!` branch seems to be faster than calling `tokio_select!` inside an outer loop.

Thank you for testing this @albertlockett!

PR #1033, opened 2025-08-29
Endpoints to expose the current overall state of the pipelines and to stop pipelines

lquerel:
This PR provides additional visibility into the current state of the system and pipelines. It is now also possible to stop pipelines through a shutdown endpoint.

List of admin endpoints added in this PR:
- Get the status of the specified pipeline: GET /pipeline-groups/{pipeline_group_id}/pipelines/{pipeline_id}/status
- Get the status of all the pipeline groups: GET /pipeline-groups/status
- Shutdown all pipeline groups: POST /pipeline-groups/shutdown

Heartbeat events are not yet implemented and will be added in a future PR.

Background context: This PR is the first step of the implementation of the reconciler concept.
A global observed state is now kept up to date by monitoring the changes made to/by the pipelines (and groups).
In future PRs, a desired state will be submitted to a reconciler in order to update the current configuration/execution of the pipelines. The ultimate goal is to support live reconfigurations and provide full observability of the pipelines' state.


rust/otap-dataflow/crates/admin/src/pipeline_group.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// Response body.
+#[derive(Serialize)]
+struct ShutdownResponse {
+    status: &'static str,
+}
+
+async fn shutdown_all_pipelines(
+    State(state): State<AppState>
```
albertlockett: Probably not needed for this PR, but at some point should this accept a timeout as part of the request body?

I imagine that this causes a `NodeControlMessage` (which has a timeout field) to get sent to each node, and some nodes require some time to shutdown (thinking parquet exporter will try to close all the file writers for example)

Issue #1027, opened 2025-08-28
[pdata batching] fix known splitting bugs and re-enable tests


albertlockett:
Two issues were found when testing the splitting code that was added in #994 . We should fix these issues as a followup and re-enable/complete the splitting unit tests.

Known issues:

- [ ] **Issue 1 - subtract with underflow when calculating splits for metrics** 

The first issue seems to be a subtract with underflow here:
https://github.com/msalib/otel-arrow/blob/13c40695ea5747a1edc1c1bf6f473020c1dc6010/rust/otel-arrow-rust/src/otap/groups.rs#L751-L758

This can be replicated using this test which is currently ignored:
https://github.com/msalib/otel-arrow/blob/13c40695ea5747a1edc1c1bf6f473020c1dc6010/rust/otel-arrow-rust/src/otap/groups.rs#L2128-L2137

- [x] **~Issue 2 - out of bounds array access when splitting non-metrics batches~** 
 ✅ _fixed in #1038_

~Splitting non metrics batches can hit an out of bound array access here:~
~https://github.com/msalib/otel-arrow/blob/13c40695ea5747a1edc1c1bf6f473020c1dc6010/rust/otel-arrow-rust/src/otap/groups.rs#L599~

~This can be replicated by this test, which is currently ignored:~
~https://github.com/msalib/otel-arrow/blob/13c40695ea5747a1edc1c1bf6f473020c1dc6010/rust/otel-arrow-rust/src/otap/groups.rs#L2067-L2073~

~I believe this issue is caused when child record batch does not has a parent ID that points identified by the range. For example, if we had something like:~
```
spans:
------
id: [0, 1, 2, 3]

span_links:
-----------
parent_id: [0, 1, 1, 1]

span_events:
-------------
parent_id: [2, 2, 3, 3]
```

lquerel:
@albertlockett Can we close this one now?

albertlockett:
@lquerel not yet, #1038 only fixed half of the issues

PR #1038, opened 2025-08-29
[otel-arrow-rust] fix a few bugs splitting traces/logs when creating resized OTAP batches

albertlockett:
part of #1027 

- fix out of bounds slice access when splitting child if child does not contain parent IDs in the final split of the parent batch
- stop creating empty record batches for child splits if the child has no parent IDs in the range of the parent split
- `take` child batch from the input arrays so we don't fail the assertion in `generic_split` that there's nothing left in the input `batches`

### DIFF
--- a/rust/otel-arrow-rust/src/otap/groups.rs
+++ b/rust/otel-arrow-rust/src/otap/groups.rs
@@ -340,7 +340,7 @@ fn generic_split<const N: usize>(
 // This is a recursive helper function; the depth of recursion is bounded by parent-child
 // relationships described in `child_payload_types` so we won't blow the stack.
 fn generic_split_helper<const N: usize>(
-    input: &[Option<RecordBatch>; N],
+    input: &mut [Option<RecordBatch>; N],
     payload: ArrowPayloadType,
     primary_payload: ArrowPayloadType,
     parent_ids: &IDSeqs,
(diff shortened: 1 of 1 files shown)


rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -351,12 +351,12 @@ fn generic_split_helper<const N: usize>(
     assert_ne!(payload, primary_payload);
 
     let payload_offset = POSITION_LOOKUP[payload as usize];
-    if let Some(table) = input[payload_offset].as_ref() {
+    if let Some(table) = std::mem::take(&mut input[payload_offset]) {
```
lquerel, line 354: Is there any kind of advantage to use here `std::mem::take` over the classic `Option::take` ?

lquerel (approved):
LGTM

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: I can't imagine it would make a big difference. I used `std::mem::take` out of habit.

They do basically the same thing . `std::mem::take` will just call `std::mem::replace` with `Default::default()`, which on Option (which returns `None`), 
https://doc.rust-lang.org/src/core/mem/mod.rs.html#805
https://doc.rust-lang.org/src/core/option.rs.html#2125

vs the `Option::take` does the same thing without invoking `default()`
https://doc.rust-lang.org/src/core/option.rs.html#1785

PR #1039, opened 2025-08-29
[otap-dataflow] Add metrics to fake signal receiver

clhain:
This PR adds counters for otlp items generated to the fake signal exporter, allowing like-for-like comparison of signals sent/received between it and perf-exporter.

```
    {
      "name": "fake_data_generator.receiver.metrics",
      "brief": "",
      "attributes": {
...
      },
      "metrics": [
        {
          "name": "logs",
          "unit": "{log}",
          "brief": "Number of logs generated.",
          "instrument": "counter",
          "value": 1000
        },
        {
          "name": "spans",
          "unit": "{span}",
          "brief": "Number of spans generated.",
          "instrument": "counter",
          "value": 0
        },
        {
          "name": "metrics",
          "unit": "{metric}",
          "brief": "Number of metrics generated.",
          "instrument": "counter",
          "value": 0
        }
      ]
    }
```

### DIFF
--- a/rust/otap-dataflow/crates/otlp/src/fake_signal_receiver/metrics.rs
+++ b/rust/otap-dataflow/crates/otlp/src/fake_signal_receiver/metrics.rs
@@ -1,0 +1,22 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Metrics for the OTLP Fake Signal Receiver node.
+
+use otap_df_telemetry::instrument::Counter;
+use otap_df_telemetry_macros::metric_set;
+
+/// Pdata-oriented metrics for the OTAP PerfExporter.
+#[metric_set(name = "fake_data_generator.receiver.metrics")]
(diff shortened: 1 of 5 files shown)


rust/otap-dataflow/crates/otlp/src/fake_signal_receiver/metrics.rs
```diff
@@ -0,0 +1,22 @@
...
+pub struct FakeSignalReceiverMetrics {
+    /// Number of logs generated.
+    #[metric(unit = "{log}")]
+    pub logs: Counter<u64>,
+    /// Number of spans generated.
+    #[metric(unit = "{span}")]
+    pub spans: Counter<u64>,
+    /// Number of metrics generated.
+    #[metric(unit = "{metric}")]
+    pub metrics: Counter<u64>,
```
lquerel: To follow the same naming conventions used in the exporter side (see also the collector RFC mentioned in the original PR), I suggest to add the suffix `_produced` to each these metrics.

PR #1041, opened 2025-08-30
[WIP] Draft back-propagation mechanism

jmacd:
**Tests do not compile!** 

retry_processor.rs major pieces in place (w/o delay scheduling).

Major pieces are:
- new Context, ReplyTo, ReplyState, Register (context.rs)
- new OtapPayload (former OtapPdata), OtapRequest (2-part container) (pdata.rs)
- new PipelineData trait has Context and Payload associated types (would be a separate PR) (engine lib.rs)
- new AckMsg, NackMsg; both contain Context propagating backwards (control.rs)
- new DeliverAck, DeliverNack pipeline controller messages (w/ node_id) (pipeline_ctrl.rs)
- new retry logic using two ReplyState registers (retry_processor.rs)
- modified effect_handler OTLP and OTAP exporters showing how to reply via pipeline controller!

Very drafty! Lots of `pub(crate)` and `pub` where it maybe does not belong. 😅


rust/otap-dataflow/crates/engine/src/control.rs
```diff
@@ -10,33 +10,94 @@ use crate::shared::message::{SharedReceiver, SharedSender};
 use otap_df_telemetry::reporter::MetricsReporter;
 use std::time::Duration;
 
+/// The ACK
+#[derive(Debug, Clone)]
+pub struct AckMsg<PData: crate::PipelineData> {
+    /// Reply-to information.
+    // Note! This Box<PData: crate::PipelineData> should be <Context> which can be
```
albertlockett: This is no longer a `Box`, is this an old comment?

rust/otap-dataflow/crates/engine/src/control.rs
```diff
@@ -10,33 +10,94 @@ use crate::shared::message::{SharedReceiver, SharedSender};
...
+    // accomplished by making PData bound by a trait and associated
+    // type i.e., we would use PData::Context. That will be a large
+    // change, for now we expect PData with empty data and returning
+    // context.
+    pub context: PData::Context,
+
+    /// Rejected items
+    pub rejected: Option<(i64, String)>,
```
albertlockett, line 25: nit: it's not super clear here what the contents of these tuples represent? Is `i64` a count and `String` a message? I wonder if using a struct here might make things more clear?
```rs
pub rejected: Option<AckMsgRejection>,

// ...

struct AckMsgRejection {
  count: i64,
  reason: String,
}
```

rust/otap-dataflow/crates/engine/src/control.rs
albertlockett: (same comment below in `NackMsg`)

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -213,34 +219,38 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
-                        }
+            self.try_send(node_id, msg).await;
                     }
                 }
             }
         }
         Ok(())
     }
+
+    async fn try_send(&mut self, node_id: usize, msg: NodeControlMsg<PData>) {
```
albertlockett: nit: I feel like functions beginning with `try_` generally return an `Result`. The signature here makes me feel like it might fail to send the message for some reason. Is the fact that it tries internally to use the fast path & ignores closed channels an implementation detail that should be abstracted away behind some more generic method name like `send_message_to_node`?

rust/otap-dataflow/crates/otap/src/context.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    /// Pushes new reply-to state.
+    pub(crate) fn reply_to(&mut self, node_id: usize, state: ReplyState) {
+        self.reply_to.push(ReplyTo { node_id, state });
+    }
+
+    /// Indicates the return destination by node_id index.
+    pub(crate) fn reply_node_id(&self) -> usize {
+        self.reply_to.last().expect("has_reply_state").node_id
+    }
```
albertlockett: Could you comment on why it's safe to expect here? It seems like someone could do something like:
```rs
let ctx = Context::default(); // reply_to will be the default empty vec
ctx.reply_node_id(); // will panic
```
Is the expectation here that users will always call `has_reply_state` first?

Maybe it would be safer to return an `Option`?
```suggestion
    /// Indicates the return destination by node_id index.
    pub(crate) fn reply_node_id(&self) -> Option<usize> {
        self.reply_to.last().map(|rs| rs.node_id)
    }
```

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
```diff
@@ -93,6 +95,78 @@ impl OTAPExporter {
...
+                            false,
+                            None,
+                            reply.take_reply_payload(),
+                        ))
+                    }),
+            )
+            .await
+    }
+
+    async fn grpc_status(
```
albertlockett: nit: looks like this only gets called for errors. Maybe call it `grpc_error_status` ?`status.code` could be `Code::Ok`, so otherwise someone might expect they could pass that status and send an `Ack`

rust/otap-dataflow/crates/otap/src/parquet_exporter.rs
```diff
@@ -174,7 +174,14 @@ impl Exporter<OtapPdata> for ParquetExporter {
                 }
 
                 Message::PData(pdata) => {
-                    let mut otap_batch: OtapArrowRecords = pdata.try_into()?;
+                    let mut request: OtapRequest = pdata.split_into();
+                    let payload = request.take_request_payload();
+                    let requested: OtapArrowRecords = payload.try_into()?;
+                    // TODO some kind of guard for reply
+                    let save_reply = OtapRequest::new_reply(request.take_context(), &requested);
+                    _ = save_reply;
```
albertlockett: I agree that we can probably do this in a followup for parquet exporter.  Figuring out when to send the `Ack` in parquet exporter might be a bit more involved. E.g., we'll need to consider whether it makes sense to send the `Ack` once it finishes writing the file (which might be a long time after the batch was received), or only when the write has been buffered. I'm still working on my mental model of the Ack/Nack/retry system, so not sure off-hand what the right answer is.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -211,56 +254,133 @@ impl OtapPdataHelpers for OtapArrowBytes {
...
+    /// Construct new OtapData with payload using default context.
+    pub fn new_default(payload: OtapPayload) -> Self {
+        Self {
+            context: Context::default(),
+            payload,
+        }
+    }
+
+    /// Disassembles the data into context and payload.
+    pub fn split_into(self) -> OtapRequest {
```
albertlockett: nit: should this be replaced with `impl From<OtapPdata> for OtapRequest`? I would expect `split_into` to be something more like:

```rs
/// Disassembles the data into context and payload.
pub fn split_into(self) -> (Context, OtapPayload) {
  (self.context, self.payload)
}
```

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -211,56 +254,133 @@ impl OtapPdataHelpers for OtapArrowBytes {
...
+
+    /// To send the request, take it.
+    pub fn take_request_payload(&mut self) -> OtapPayload {
+        self.payload.take().expect("called once")
+    }
+
+    /// Take the context.
+    pub fn take_context(&mut self) -> Context {
+        self.context.take().expect("called once")
+    }
```
albertlockett: Could you comment on why it's safe to expect that each of these can only been called once? If this gets called more than once it will panic

rust/otap-dataflow/crates/otap/src/context.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/// ReplyState is a the unit of return in a return chain.
+#[derive(Clone, Debug)]
+pub struct ReplyState {
+    /// First register
+    pub(crate) r0: Register,
+
+    /// Second register
+    pub(crate) r1: Register,
```
albertlockett: It seems like these only get used by the `RetryProcessor` where we try to turn this into a `RetryState`. Is there any reason that the types need to generic vs just using the types that `RetryState` expects directly?

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -290,52 +189,121 @@ impl Processor<OtapPdata> for RetryProcessor {
...
+                        retries: 0,
+                        last_ts: Instant::now(),
                     };
-
-                    let _previous = self.pending_messages.insert(id, pending);
-                    log::debug!("Added message {id} to retry queue");
-
-                    effect_handler.send_message(data).await?;
+                    data.context
+                        .reply_to(effect_handler.processor_id().index(), rstate.into());
```
jmacd, line 235: Here, I want a "Nack interest, no Ack interest" type of signal.
If No-Ack interest, then the pipeline controller can auto-propagate.

Issue #1025, opened 2025-08-28
[pdata batching] perf improvement, avoid flattening dictionaries


albertlockett:
in https://github.com/open-telemetry/otel-arrow/pull/994 we added the ability to split & concatenate batches of OTAP data, but a concession we made on the first pass was to flatten dictionaries into the native arrays. 

We do this for columns in the record bath here:
https://github.com/msalib/otel-arrow/blob/13c40695ea5747a1edc1c1bf6f473020c1dc6010/rust/otel-arrow-rust/src/otap/groups.rs#L1257-L1270

As well as for nested dictionaries within structs here:
https://github.com/msalib/otel-arrow/blob/13c40695ea5747a1edc1c1bf6f473020c1dc6010/rust/otel-arrow-rust/src/otap/groups.rs#L1493-L1499

We should revisit this implementation and avoid flattening the dictionaries.

PR #1042, opened 2025-09-01
Avoid flattening dictionaries when unifying record batches during batching

albertlockett:
Closes: #1025 

When concatenating OTAP batches, there are two issues that are a bit difficult WRT to dictionary columns:
- the combined record batch might overflow the dictionary due to more keys having more cardinality than the key type can support
- received record batches may have different data types for the same dictionary column (it depends on the size of the batch, and the cardinality of the column when it was constructed).

Before this, we were using a simler logic of flattening all dictionaries, however in doing this we lose the benefits of dictionary compression.

With this PR, we no longer flatten the dictionaries. Instead, we attempt to determine the correct key size to use and the cast the column for all record batches to be the correct data type. We try to make the decision about which key type to use in the most efficient way by considering the total batch size, the possible cardinality of the columns and the minimum dictionary key size.


rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1253,35 +1259,42 @@ fn unify<const N: usize>(batches: &mut [[Option<RecordBatch>; N]]) -> Result<()>
...
-                .unwrap_or(false)
-            {
-                if let Some(rb) = batches[payload_type_index].take() {
-                    let rb = flatten_dictionary_record_batch(rb)?;
-                    let _ = batches[payload_type_index].replace(rb);
-                }
+        for batches in batches.iter() {
+            if let Some(rb) = &batches[payload_type_index] {
+                try_count_dictionary_values(rb, &mut dict_fields)?;
+                try_count_struct_dictionary_fields(rb, &mut struct_fields)?;
```
jmacd: Code looks good -- 

I wonder if the function names try_{count,record}_{dictionary,struct}_{values,fields} could be renamed to improve the reader's experience. I'm feeling a bit confused because the names are all so similar. I couldn't tell why 3 of 4 end in "_fields" (one in "_values") and I couldn't tell if the "record" substring should be read as a verb or a noun. Possibly a high-level comment at this location to explain why we repeat two iterations over batches.iter() instead of fusing the four try_ into a single block, for example.

albertlockett: Good call @jmacd -- yeah I'll fix the naming. That's a good callout about the inconsistent naming (field/values) as well

albertlockett: @jmacd fixed up the method names and added that requested comment.

rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1253,35 +1259,50 @@ fn unify<const N: usize>(batches: &mut [[Option<RecordBatch>; N]]) -> Result<()>
             let _ = all_batch_indices.insert(batch_index);
         });
 
-        for batches in batches.iter_mut() {
-            // try to get the fields that should be in each struct column
+        // this will be used to initialize the dictionary key selector. There are certain
+        // optimizations that can be made when choosing the correct key sizes using this value
+        let total_batch_size = batches
```
msalib: Consider using `select` here.

rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1323,68 +1344,382 @@ fn unify<const N: usize>(batches: &mut [[Option<RecordBatch>; N]]) -> Result<()>
...
-            let value = *value.clone();
-            field.clone().with_data_type(value)
+impl UnifiedDictionaryKeySelector {
+    fn new(total_batch_size: usize, key_data_type: DataType) -> Self {
+        Self {
+            total_batch_size,
+            smallest_key_data_type: key_data_type,
+            total_values_visited: 0,
+            state: Default::default(),
+            dedup: BTreeSet::new(),
```
msalib: Would it make sense to use `with_capacity` picking either `2**8` or `2**16` based on `smallest_key_data_type`?

rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1323,68 +1344,382 @@ fn unify<const N: usize>(batches: &mut [[Option<RecordBatch>; N]]) -> Result<()>
...
+
+        // we know the values can fit into a dict keyed by u16 because of the batch size, but
+        // we also know there's no chance they'll fit into a u8, so the key decision is made
+        if self.total_batch_size <= u16::MAX as usize && self.dedup.len() > u8::MAX as usize {
+            return true;
+        }
+
+        // if this is true, we know we need to use the native array type
+        if self.dedup.len() > u16::MAX as usize {
+            return true;
```
msalib: Eh, should this be `false`?

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: Can you elaborate more on what you mean here?

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: `BTreeSet`/`BTreeMap` don't have `with_capacity`. 

afiak this data structure allocates an array  of values for each node in the tree (starting with single array for the root), and split/allocate child blocks when they need more capacity

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: It is correct to return `true` here. The purpose of this method is to try to determine as early as possible what type of array will be needed to hold all values. If we detect that the number of unique values is > u16::MAX, it means that they won't all fit into a u16 keyed dict and we need to use the native array (no dict encoding).

However, I now can see how the name of the method might be confusing. Maybe I should change the method name to something like `array_type_selected` instead? Do you think it'd help make things more clear?

rust/otel-arrow-rust/src/otap/groups.rs
msalib: I was thinking you could write
```rust
select(batches, payload_type_index).map(|rb|rb.num_rows()).sum()
```
instead....

rust/otel-arrow-rust/src/otap/groups.rs
msalib: Ah right!

rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1323,68 +1344,382 @@ fn unify<const N: usize>(batches: &mut [[Option<RecordBatch>; N]]) -> Result<()>
...
+///
+/// Next, it should visit the values, where it will try to calculate the total cardinality across
+/// the combined columns until it figures out what key type will fit all the values. To invoke this
+/// use the [`Self::visit_dictionary_values`].
+struct UnifiedDictionaryKeySelector {
+    smallest_key_data_type: DataType,
+    total_batch_size: usize,
+    total_values_visited: usize,
+    state: ahash::RandomState,
+    dedup: BTreeSet<u64>,
```
msalib: Can you write more about the rationale for using a `BTreeSet` of key hashes instead of an `AHashSet` of actual keys? I get that storing hashes saves you from having to deal with enum dispatch to generic boilerplate to deal with covering lots of different types, but it introduces the possibility of collisions. Given that some dictionary columns encode arbitrary strings, we're going to get collisions, right? What happens when we do?

Another benefit of just storing an `AHashSet` over keys is that you don't have to do any complicated math on how many duplicates there are; just get the length of the set and that's your cardinality. If you want to do early termination, the set length always tells you whether you've crossed the `u16::MAX` threshold.

rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1323,68 +1344,382 @@ fn unify<const N: usize>(batches: &mut [[Option<RecordBatch>; N]]) -> Result<()>
...
+
+        // if this is true, we know we need to use the native array type
+        if self.dedup.len() > u16::MAX as usize {
+            return true;
+        }
+
+        // calculate the max cardinality based on the total batch size and the number of duplicate
+        // values we've visited
+        let duplicates_visited = self.total_values_visited - self.dedup.len();
+        let possible_max_cardinality = self.total_batch_size - duplicates_visited;
```
msalib: This is what I mean above about complicated math. I think what you have here is correct...but I'd be more confident if there was no math and just `let candinality = self.all_keys.len()`.

rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1323,68 +1344,382 @@ fn unify<const N: usize>(batches: &mut [[Option<RecordBatch>; N]]) -> Result<()>
...
+    }
+
+    fn visit_dictionary_values(&mut self, maybe_dict_column_arr: &ArrayRef) -> Result<()> {
+        if self.key_selected() {
+            // we've already seen enough values that we can decide the dictionary key size
+            return Ok(());
+        }
+
+        // access the array containing the dictionary values ...
+        let values_arr = match maybe_dict_column_arr.data_type() {
```
msalib, line 1382: Minor nit: I think you can avoid the nested match expressions by just matching on cases like `DataType::Dictionary(DataType::UInt8, _)`.

rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1323,68 +1344,382 @@ fn unify<const N: usize>(batches: &mut [[Option<RecordBatch>; N]]) -> Result<()>
...
+                }
+            },
+            _ => {
+                // assume it's native column
+                maybe_dict_column_arr
+            }
+        };
+
+        // visit the dictionary values until we decide on the key type ..
+        match values_arr.data_type() {
```
msalib: You might be able to reduce code duplication here and in the `visit_*_dict_values` methods by converting from `Array` to `ArrayData` with `to_data()` and then calling the `buffer` method with either a 1-byte type, a 2-byte type, a 4-byte type or an 8-byte type, giving you a slice of elements. For cardinality estimation, we don't care about the precise values so much as their binary representation. You'd still need special cases for `Binary/Utf8/FixedSizeBinary`. Then your `visit_primitive_values` could just take a slice of `[u8; N]` so it would only be generic over the size of the byte array taken by an element.

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: ohhhh, yes this makes sense. Sorry, I had forgotten `select` was something we wrote 😅 .. I was searching for some new method on std::Iterator or something from `itertools` lol. made this change

rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1323,68 +1342,382 @@ fn unify<const N: usize>(batches: &mut [[Option<RecordBatch>; N]]) -> Result<()>
...
+            self.total_values_visited += 1;
+            if self.key_selected() {
+                break;
+            }
+        }
+    }
+
+    /// Choose the key type to use for the dictionary. If it returns `None` it means
+    /// use the native array type (not a dictionary array)
+    fn choose_key_type(&self) -> Option<DataType> {
```
msalib: A comment explaining how this code covers all the cases in `key_selected` would be helpful; honestly, I'd prefer a world in which this method was merged with `key_selected` and just returned an `Option<DataType>`.

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: Yeah that's a good callout about the collisions. On the off chance we did get a collision, we'd undercount the cardinality of the dictionary values and I guess it could push us into selecting a dictionary key that won't fit all the values which will panic when we concat the record batches together. It's probably worth fixing :) 

The disadvantage of storing the values is that it makes the code a bit more complicated.

We could get around the awkwardness of making the hashset generic over the value type by just having it contain bytes, and just always storing the bytes from the values buffer (we're already pulling these out [here](https://github.com/open-telemetry/otel-arrow/blob/268f54ba42c2f7c4abd75530dbbdbf7f0668e961/rust/otel-arrow-rust/src/otap/groups.rs#L1465-L1544)).

The thing that's more awkward about this is that, to avoid copying the values directly into the hashset, we'd ideally want to make it a `AHashSet<&[u8]>`, but it gets tricky with the lifetimes, especially where we eventually mutably borrow the slice of `Option<RecordBatch>` to replace the column while this `UnifiedDictionaryKeySelector` still exists. However, I think I could clone the `Arc` containing the array and make it a member of this struct which will hopefully appease the borrow checker .. Let me try this and I'll tell you how it goes!

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: Unfortunately at this place in the code, we haven't visited all the values yet, so the true cardinality of the column isn't `dedup.len()`.

`possible_max_cardinality` is what the cardinality _would be_ if all the values we haven't visited are unique.

I agree it's a bit more complicated, but the motivation for doing it this way is we don't need to visit the entire values array and hash/insert all the values

rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1323,68 +1342,382 @@ fn unify<const N: usize>(batches: &mut [[Option<RecordBatch>; N]]) -> Result<()>
...
+    record_batch: &RecordBatch,
+    total_batch_size: usize,
+    dictionary_fields: &mut BTreeMap<String, UnifiedDictionaryKeySelector>,
+) -> Result<()> {
+    let schema = record_batch.schema_ref();
+
+    for field in schema.fields() {
+        if let DataType::Dictionary(dict_key_type, _) = field.data_type() {
+            let field_name = field.name();
+            if dictionary_fields.contains_key(field.name()) {
```
msalib: Can we replace the `contains_key -> get_mut else insert` dance with the entry API?

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: I'm not sure that works does it? `DataType::Dictionary` doesn't contain the datatypes directly, they're `Box<DataType>`s 

https://github.com/apache/arrow-rs/blob/4506998155a5d915e7d70ffb8e0d511a24ada4ee/arrow-schema/src/datatype.rs#L359

rust/otel-arrow-rust/src/otap/groups.rs
msalib: Another option would be something like https://crates.io/crates/hyperloglogplus for estimating cardinality without having to store the values directly....

rust/otel-arrow-rust/src/otap/groups.rs
msalib: Ah right, boxed patterns are still unstable....

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: I refactored to use `AHashSet` in [69a4f7c](https://github.com/open-telemetry/otel-arrow/pull/1042/commits/69a4f7c2cc37f1af61a7694512b358ae4a15b700)

I did end up having to store the ArrayRef in this type, but that turned out to be kind of useful because it means we can look at them all at once when making the key decision, and which reduces the decision making happening across two different methods, which addresses the concern your raised here:
https://github.com/open-telemetry/otel-arrow/pull/1042#discussion_r2319546429

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: I think this is addressed now. See this comment https://github.com/open-telemetry/otel-arrow/pull/1042#discussion_r2320023769

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: FYI This has now been changed to:
https://github.com/open-telemetry/otel-arrow/blob/69a4f7c2cc37f1af61a7694512b358ae4a15b700/rust/otel-arrow-rust/src/otap/groups.rs#L1566-L1569

Can you give a concrete example of how you'd expect the dance to work?

If we call `.entry()`, we have to clone the field name because this method takes the key by value:
https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.entry

Also if we call get_mut(), I think we still have to call insert .. so this would just become:
```rs
            if dictionary_fields.get_mut(field.name()).is_none() {
               _ = dictionary_fields
                    .insert(field.name().clone(), UnifiedDictionaryTypeSelector::new()) 
            }
```
which is basically the same thing

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: Using `ArrayData` gets a bit awkward with lifetimes, because `to_data()` returns an owned reference to `ArrayData`. So if we want to insert the reference to the bytes into the `AHashSet` we're using to deduplicate the values, we can't create the `ArrayData` in the inner function we use to access the values.

I did refactor to try to remove the duplication, using a macro:
https://github.com/open-telemetry/otel-arrow/blob/69a4f7c2cc37f1af61a7694512b358ae4a15b700/rust/otel-arrow-rust/src/otap/groups.rs#L1482-L1530

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: FYI to avoid any confusion about this type only selecting dictionary keys (when in fact, it can also decide if the datatype should be native), in [69a4f7c](https://github.com/open-telemetry/otel-arrow/pull/1042/commits/69a4f7c2cc37f1af61a7694512b358ae4a15b700) I changed the type name to `UnifiedDictionaryTypeSelector`, changed the method signature to :
```rs
    fn select_column_data_type(&self) -> Result<UnifiedDictionaryType> {
```

Where the datatype is more explicit:
```rs
enum UnifiedDictionaryType {
    Dictionary(DataType),
    Native,
}
```

rust/otel-arrow-rust/src/otap/groups.rs
msalib: Yeah, this is one of those places where I'd be inclined to just copy bytes directly into the AHashSet and take the hit there since there never be more than 65K entries, but I appreciate that that might be a terrible idea!

msalib (approved):

rust/otel-arrow-rust/src/otap/groups.rs
albertlockett: ah yeah we don't need to copy. We still push the borrowed  u8 slice into the dedup HashSet. It's just that we need to have a function to get bytes that's generic over the primitive type instead of just using the non-generic ArrayData and having it be generic over the value size.

rust/otel-arrow-rust/src/otap/groups.rs
msalib: Ugh, yeah, I forgot about how the entry API doesn't do anything clever with `ToOwned` to generate copies on demand, sorry about that.

jmacd (approved):

Issue #1026, opened 2025-08-28
[pdata batching] perf improvment, sort using RowConverter


albertlockett:
As part of out batching and splitting algorithm, we sort the batches by their ID columns to optimize the remapping of various IDs. This sorting is done using the `lexsort_to_indices` kernel, but a well-known optimization for multi-column sort is to use the row converter:
https://github.com/msalib/otel-arrow/blob/13c40695ea5747a1edc1c1bf6f473020c1dc6010/rust/otel-arrow-rust/src/otap/groups.rs#L780-L836

Note that we have another function to sort record batches in the transform_optimize module. There might be an opportunity here to refactor this into something reusable and reuse it in both places:
https://github.com/msalib/otel-arrow/blob/13c40695ea5747a1edc1c1bf6f473020c1dc6010/rust/otel-arrow-rust/src/otap/transform/transport_optimize.rs#L408-L480

PR #1043, opened 2025-09-01
Use Arrow row format when sorting for batching if possible

albertlockett:
Closes: #1026 

Improve batching performance by using the arrow row format for multicolumn sort where possible.

We do this by creating a reusable function called `otap::transform::sort_to_indices` and using this both in the batching code and in the transport optimization encoding code (which is where this was refactored from).

### DIFF
--- a/rust/otel-arrow-rust/src/otap/transform.rs
+++ b/rust/otel-arrow-rust/src/otap/transform.rs
@@ -7,12 +7,14 @@ use std::sync::Arc;
 
 use arrow::array::{
     Array, ArrayRef, ArrowPrimitiveType, BooleanArray, DictionaryArray, NullBufferBuilder,
-    PrimitiveArray, PrimitiveBuilder, RecordBatch, StringArray,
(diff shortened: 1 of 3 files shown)


lquerel (approved):
LGTM

rust/otel-arrow-rust/src/otap/transform.rs
```diff
@@ -1464,6 +1466,51 @@ fn take_null_buffer_ranges(
     })
 }
 
+/// Helper function for sorting to indices. Encapsulates the logic of choosing whether to use row
+/// sort or single column sort
+pub(crate) fn sort_to_indices(
+    sort_columns: &[SortColumn],
+) -> std::result::Result<UInt32Array, ArrowError> {
```
msalib: Minor nit: use `arrow::error::Result<UInt32Array>` here instead.

albertlockett: Thanks @msalib! made this change in [2e26861](https://github.com/open-telemetry/otel-arrow/pull/1043/commits/2e268619a210d0fada09fcfb37a68b60ab32d088)

Issue #1047, opened 2025-09-02
[otap-dataflow] Processor runtime errors


clhain:
I'm seeing errors when running the attribute processor (I haven't tried others yet, Laurent mentioned this looks like an engine bug rather than issue with a specific processor):

```shell
$ ../../rust/otap-dataflow/target/release/df_engine --num-cores 1 --pipeline ../../rust/otap-dataflow/configs/otap-attr-perf.yaml 
Starting pipeline with 1 cores
[crates/controller/src/lib.rs:182:13] &results = [
    Err(
        PipelineRuntimeError {
            source: ProcessorError {
                processor: NodeId {
                    index: 1,
                    name: "attr",
                },
                error: "Expected a local sender for PData",
            },
        },
    ),
]
```

The config is:

```yaml
settings:
  default_control_channel_size: 100
  default_pdata_channel_size: 100

nodes:
  receiver:
    kind: receiver
    plugin_urn: "urn:otel:otap:receiver"
    out_ports:
      out_port:
        destinations:
          - attr
        dispatch_strategy: round_robin
    config:
      listening_addr: "0.0.0.0:1235"
      message_size: 100
      compression_method: Zstd

  attr:
    kind: processor
    plugin_urn: "urn:otap:processor:attributes_processor"
    out_ports:
      out_port:
        destinations:
          - perf
        dispatch_strategy: round_robin
    config:
      actions:
        - action: "rename"
          source_key: "http.method"
          destination_key: "rpc.method"       # Renames http.method to rpc.method
          apply_to: ["scope"]              # defaults to ["signal"]
  perf:
    kind: exporter
    plugin_urn: "urn:otel:otap:perf:exporter"
    config:
      frequency: 500
      cpu_usage: false
      mem_usage: false
      disk_usage: false
      io_usage: false

```

I see the same issue with otlp receiver / exporter configured.

PR #1051, opened 2025-09-02
fix building pipelines for proccesors

albertlockett:
closes: #1047 

- fix setting the receiver on local processors if upstream component is shared
- fixes attributes processors dropping the out ports from the configured `NodeUserConfig`


lquerel (approved):
LGTM

PR #1053, opened 2025-09-02
fix: remove orphan record.Schema() call in NewLogsMessage

MasslessParticle:
I'm playing with otel arrow and noticed this orphaned `record.Schema()` call in NewLogsMessage. This PR removes the call.


albertlockett (approved):
thanks, good catch @MasslessParticle!

PR #1054, opened 2025-09-02
Parameter --core-id-range

lquerel:
To isolate the different workloads in our benchmarking infrastructure, we need to be able to allocate subsets of CPU cores to different components of the infrastructure (traffic generator, system under test, perf collector).


rust/otap-dataflow/crates/config/src/pipeline_group.rs
```diff
@@ -81,6 +81,20 @@ pub struct Quota {
     /// If set to a value greater than 0, it will use that many cores.
     #[serde(default = "default_num_cores")]
     pub num_cores: usize,
+
+    /// Optional inclusive range of CPU core IDs to pin threads to, e.g., start..=end.
+    /// When specified, this takes precedence over `num_cores`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub core_id_range: Option<CoreIdRange>,
```
utpilla: It would be nice to enforce that only one of the two options can be present: either `num_cores` or `core_id_range` maybe using an `enum`), even though we enforce it at a command line level. It would make the intent clearer.

utpilla (approved):

rust/otap-dataflow/crates/config/src/pipeline_group.rs
lquerel: Good idea. I like this suggestion.

lquerel: https://github.com/open-telemetry/otel-arrow/pull/1054/commits/095ba60c4d8a215b50f7caf86d9bff82caeee4d8

albertlockett (approved):

albertlockett:
LGTM

PR #1055, opened 2025-09-03
Draft: otap batch processor internal telemetry

daviddahl:
Add minimal internal telemetry to the OTAP batch processor using telemetry-macros. We register a metric set and report on CollectTelemetry. Instrumentation covers:
•  Per-signal rows: received_rows_logs/metrics/traces
•  Flush reasons (aggregated): flushes_size, flushes_timer
•  Drops/errors (aggregated): dropped_empty_records, dropped_conversion, batching_errors
•  Split requests (aggregated): split_requests

This is branched off of the work in #1015 and will be rebased once 1015 lands


daviddahl:
Going to open a new PR with this diff reconstructed on `main` as the changes were messy

PR #1057, opened 2025-09-03
[otap-dataflow] Register the syslog ref receiver

c1ly:
Moved the syslog receiver inside the otap-df-otap and registered it to the receiver factory


lquerel:
@c1ly there are some issues in the benchmarks folder.

PR #1071, opened 2025-09-05
fix: metrics splitter in the batching subsystem

### LINKED ISSUE #1027: [pdata batching] fix known splitting bugs and re-enable tests

msalib:
Closes #1027.


lquerel (approved):
LGTM

PR #1076, opened 2025-09-05
[otap-dataflow] BatchProcessor internal telemetry

daviddahl:
Adds internal telemetry to the batch processor to observe volumes and flush decisions without changing behavior.

Key changes
•  Metrics: define OtapBatchProcessorMetrics (moved to crates/otap/src/otap_batch_processor/metrics.rs); update imports. Processor import path unchanged.
•  Wiring: add optional metrics field and from_config_with_metrics; register and inject MetricSet in the factory.
•  Instrumentation: received_rows_{logs,metrics,traces}; flushes_{size,timer,shutdown}; split_requests; dropped_{conversion,empty_records}; batching_errors; dirty_set/cleared per signal; timer_flush{performed,skipped} per signal.
•  Tests: internal telemetry collection test; small encoder test.

Notes
•  No functional changes to batching or flush logic; telemetry-only.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/encoder.rs
+++ b/rust/otap-dataflow/crates/otap/src/encoder.rs
@@ -3728,6 +3728,42 @@ mod test {
         _test_attributes_all_field_types_generic(RawLogsData::new(&logs_data_bytes));
     }
 
+    #[test]
+    fn test_encode_logs_batch_length_counts_rows() {
+        use otel_arrow_rust::otap::OtapArrowRecords;
+        use otel_arrow_rust::proto::opentelemetry::common::v1::{
+            AnyValue, InstrumentationScope, KeyValue,
+        };
+        use otel_arrow_rust::proto::opentelemetry::logs::v1::{
+            LogRecord, LogsData, ResourceLogs, ScopeLogs, SeverityNumber,
+        };
+        use otel_arrow_rust::proto::opentelemetry::resource::v1::Resource;
+
+        // Build logs with at least one attribute per record so log ids are set
+        let logs: Vec<LogRecord> = (0..3)
+            .map(|i| {
+                LogRecord::build(i as u64, SeverityNumber::Info, format!("log{i}"))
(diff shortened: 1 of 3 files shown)


daviddahl:
replaces #1055

rust/otap-dataflow/crates/otap/src/encoder.rs
```diff
@@ -3728,6 +3728,42 @@ mod test {
         _test_attributes_all_field_types_generic(RawLogsData::new(&logs_data_bytes));
     }
 
+    #[test]
```
daviddahl, line 3783: This test was added due to a bug in our test where we were trying to see `batch_length` for logs. It seems to be a timing issue, and this can be removed if thought superflous.

rust/otap-dataflow/crates/otap/src/otap_batch_processor/metrics.rs
```diff
@@ -0,0 +1,87 @@
...
+//! Metrics for the OTAP batch processor.
+
+use otap_df_telemetry::instrument::Counter;
+use otap_df_telemetry_macros::metric_set;
+
+/// Minimal, essential metrics for the OTAP batch processor, plus observability for
+/// dirty-flag behavior and timer flush decisions.
+#[metric_set(name = "otap.processor.batch")]
+#[derive(Debug, Default, Clone)]
+pub struct OtapBatchProcessorMetrics {
```
lquerel, line 13: For the naming of the metrics, the collector defined this RFC. We should follow their recommendations for now. Example instead of `received_rows_logs`, I thing they recommend `log_consumed`. I let you check that in details.

lquerel (dismissed):
See my comment

rust/otap-dataflow/crates/otap/src/otap_batch_processor/metrics.rs
daviddahl: ok, sounds good, I missed that.

Issue #1003, opened 2025-08-25
OtapPdata equivalence testing


jmacd:
The [go/pkg/otel/assert](https://github.com/open-telemetry/otel-arrow/tree/main/go/pkg/otel/assert) package is used to assert that two OTLP payloads are equivalent after surviving an OTLP->OTAP->OTLP roundtrip.

I am looking for similar support in Rust, and/or anyway the best practice for comparing an expectation and an actual `OtapPdata` value.

albertlockett:
@jmacd we have a few round-trip tests in `pdata.rs`, for example:
https://github.com/open-telemetry/otel-arrow/blob/110b3e365c85ae517ab944c9d93de5d9716a4d3e/rust/otap-dataflow/crates/otap/src/pdata.rs#L655

That said, I don't think these are comprehensive. We don't fill in absolutely every field, nor do we test round trip for all nullable/optional fields. Also we don't have tests for round-trip metrics yet because we haven't yet implemented converting OTLP Bytes -> OTAP for this signal type.

Issue #968, opened 2025-08-19
Pipeline "context" for nodes (& possible duplication)


jmacd:
PR #939 added engine::NodeId which is struct pairing a u16 with a Cow. The unique name in the NodeId is duplicative of the one in the component's engine::config struct.

See #946 also introduces PipelineContext. The new engine::NodeId can be replaced with this.

For example, in engine/config.rs
```
pub struct ReceiverConfig {
    /// Name of the receiver.
    pub name: NodeId,
...
```

(and `ProcessorConfig`, `ExporterConfig`). PR #939 introduced a new NodeId in many constructors and in some cases, the engine::NodeId's name does not equal the engine::config value presumably in places where it does not matter.

lquerel:
Should we close this issue?

Issue #338, opened 2025-04-10
Rust OTLP and OTAP pipeline data development


jmacd:
To build the OTAP pipeline and test it using OTLP data input and output, we need an easy way build and inspect pipeline data. The Golang collector has a "pdata" model, and we aim to do something similar for Rust.

This issue will tie together work to build, mutate, and access OTLP and OTAP data in a uniform way, especially for writing readable tests and benchmark basic protocol conversions.

Given the rust/otel-arrow-rust crate already uses both OTAP and OTLP protocols and has very few tests, this will present an opportunity to add tests using OTLP pipeline data. Similarly, the same crate iterates through OTAP data frames and generates OTLP pipeline data, working directly with prost::Message objects. The pipeline data approach covered here will improve readability and test coverage.

albertlockett:
Created issues for the oustranding OTLP -> OTAP conversion work:
- [ ] #384
- [ ] #385
- [ ] #387 
- [ ] #388 
- [ ] #389

PR #1078, opened 2025-09-05
Add Context to OtapPdata, introduce OtapPayload

jmacd:
The former enum becomes OtapPayload.
The new OtapPdata type carries a context, currently an empty placeholder.

Part of #968.
Part of #338.
Taken from #1041.
Part of #1098.

The OtapPdata and OtapPayload and its two helper impls now implement several common elements, including `num_items(&self)` which replaces a helper function in retry_processor.rs (associated with #1003).

The new OtapPdata type has a constructed used everywhere, new_default() which builds an OtapPdata converting from the underlying types into OtapPayload. It supports is_empty(), take_payload(), and take_context() as well.


rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -129,85 +136,267 @@ impl OtlpProtoBytes {
...
+    pub(crate) context: Context,
+    pub(crate) payload: OtapPayload,
+}
+
+/* -------- Signal type -------- */
+
 impl OtapPdata {
+    /// Construct new OtapData with payload using default context.
+    #[must_use]
+    pub fn new_default(payload: OtapPayload) -> Self {
```
lquerel, line 163: `new_default` sounds a bit odd. Why not just use the classic `new` in this case, and reserve `with_context` for when we need to create an `OtapPdata` with both a context and a payload?

rust/otap-dataflow/crates/otap/src/debug_processor.rs
```diff
@@ -164,9 +164,9 @@ impl local::Processor<OtapPdata> for DebugProcessor {
                 }
                 Ok(())
             }
-            Message::PData(pdata) => {
+            Message::PData(mut pdata) => {
                 // make a copy of the data and convert it to protobytes that we will later convert to the views
-                let otlp_bytes: OtlpProtoBytes = pdata.clone().try_into()?;
+                let otlp_bytes: OtlpProtoBytes = pdata.take_payload().try_into()?;
```
lquerel: If I understand correctly what the `pdata.take_payload` method does, it extracts the payload from the pdata structure and leaves the original pdata with an empty payload. In other words, the debug processor propagates an "empty" pdata to the next downstream nodes (which is not what you want). I think this interface is too error prone.

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -194,13 +194,13 @@ impl local::Processor<OtapPdata> for AttributesProcessor {
...
                 let signal = pdata.signal_type();
-                let mut records: OtapArrowRecords = pdata.try_into()?;
+                let mut records: OtapArrowRecords = pdata.take_payload().try_into()?;
 
                 // Apply transform across selected domains
                 apply_transform(&mut records, signal, &self.transform, &self.domains)?;
 
                 effect_handler
-                    .send_message(records.into())
+                    .send_message(OtapPdata::new_default(records.into()))
```
lquerel: What happens to the context of the initial pdata? It seems to me it's lost. We need to find an approach where this simply cannot happen.

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -194,13 +194,13 @@ impl local::Processor<OtapPdata> for AttributesProcessor {
                 }
 
                 let signal = pdata.signal_type();
-                let mut records: OtapArrowRecords = pdata.try_into()?;
+                let mut records: OtapArrowRecords = pdata.take_payload().try_into()?;
 
                 // Apply transform across selected domains
                 apply_transform(&mut records, signal, &self.transform, &self.domains)?;
```
lquerel: I feel there's something wrong here. I can't quite put my finger on it. But here's what bothers me:
- Arrow records are immutable; if `apply_transform` needs to modify attributes, it must recreate some of the arrow records. Otherwise, it does nothing. In both cases I don`t feel we should need a `mut` on pdata (we didn`t before).
- Ideally we`d like this processor to behave like a pass-through if there's no match on attributes, and we should propagate the context as is. That's not the case, see my next comment.

We have this dual state `OtlpBytes` vs `OtapArrowRecords` which adds a bit of complexity everywhere. Personally, I see this dual state as an internal optimization that shouldn't be visible to most nodes. @albertlockett, I mentioned this in another PR, but the more I see the impact of the current interface, the more I think we need to do something about it.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -129,85 +136,267 @@ impl OtlpProtoBytes {
...
+    /// Removes the payload from this request, leaving an empty request.
+    #[must_use]
+    pub fn take_payload(&mut self) -> OtapPayload {
+        self.payload.take_payload()
+    }
+
+    /// Take the context.
+    pub fn take_context(&mut self) -> Context {
+        std::mem::take(&mut self.context)
+    }
```
lquerel: I get the impression that exposing these two methods to the nodes is problematic, because we then no longer have any guarantee that the context will be propagated correctly.

lquerel (changes requested):
I've left several comments that make me think we don't yet have the ideal approach.

This isn't directly related to this PR, but I'm increasingly convinced that we should revisit the current pdata so that the internal states `OtlpBytes` and `OtapArrowRecords` remain an internal optimization. I'm bringing this up because I'm wondering whether this PR should be based on that revision of the current pdata to simplify the overall approach. @albertlockett , I'll try to find time early this week to discuss this specific point in more detail.

rust/otap-dataflow/crates/otap/src/debug_processor.rs
jmacd: I misunderstood this code and the clone() here. We were cloning it before, and my expectation was that we would remove the payload, convert it for debugging purposes, then put it back in the pdata.

rust/otap-dataflow/crates/otap/src/pdata.rs
jmacd: I wanted to call out the fact that we're setting a default context. I almost named this new_todo() since it's rarely correct to use a default context. To me, new() is misleading if it does not take both context and payload. I had such a method but I removed it before this PR because there were no legitimate uses constructing PData from context and payload. All the existing tests use new_default to spell out the fact that these requests have default context.

rust/otap-dataflow/crates/otap/src/debug_processor.rs
jmacd: Yes. "leaves the original pdata with an empty payload" is true--to me that's the nature of take, you are very-explicitly leaving behind a default object when you call a take method. However, I would say "leaves the original pdata with context and default payload". An alternative API that I considered was to force take() to remove both the context and the payload -- I had a type called `OtapRequest` that consisted of Option<Context> and Option<Payload>. The advantage of the approach I've taken here is it causes a fairly-small impact to the rest of the code. I've inserted new_default() and take_payload() in a lot of places but it's a simple change.

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
jmacd: Everywhere "new_default" appears in real code, it's a bug. I was trying to make a PR here that would be a small change so that we can start fixing the places where new_default does not make sense. I had experimented with allowing only a new(context, payload) method, in which case the change will have a lot of `OtapPdata::new(Context::todo(), payload)` expressions. After I had written that 15 times, I decided `new_default` was easier to review.

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -194,13 +194,13 @@ impl local::Processor<OtapPdata> for AttributesProcessor {
                 }
 
                 let signal = pdata.signal_type();
-                let mut records: OtapArrowRecords = pdata.try_into()?;
+                let mut records: OtapArrowRecords = pdata.take_payload().try_into()?;
```
jmacd: See -- the "pdata" object is consumed. I made the pdata mut so that I could take the payload and consume it, leaving me a valid context. On line 203 below I should be using that context.

jmacd: a8ff0a68 has the small fix for this, including the first use of `new(context, payload)` in the PR.

jmacd: https://github.com/open-telemetry/otel-arrow/pull/1078/commits/a8ff0a68718a75869b800b0c8ea05420d38567de

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
jmacd: https://github.com/open-telemetry/otel-arrow/pull/1078#discussion_r2330628330
https://github.com/open-telemetry/otel-arrow/pull/1078#discussion_r2330648301

rust/otap-dataflow/crates/otap/src/pdata.rs
jmacd: I think I agree that it will be nice if we can find a safer way to do this. I'm not very optimistic about it, but would love to hear what you and @albertlockett think.

In the Golang environment, the context.Context type has similar risks. It's easy to accidentally create a new background context or to propagate one that shouldn't. Developers need guidance, and code reviewers need to understand the API. For this, I considered a `todo()` context similar to Golang with the same concept in mind: https://pkg.go.dev/context#TODO

rust/otap-dataflow/crates/otap/src/debug_processor.rs
```diff
@@ -166,7 +166,7 @@ impl local::Processor<OtapPdata> for DebugProcessor {
             }
             Message::PData(pdata) => {
                 // make a copy of the data and convert it to protobytes that we will later convert to the views
-                let otlp_bytes: OtlpProtoBytes = pdata.clone().try_into()?;
+                let otlp_bytes: OtlpProtoBytes = pdata.clone_payload().try_into()?;
```
jmacd: See the fix in c79fbed6.

rust/otap-dataflow/crates/otap/src/debug_processor.rs
jmacd: https://github.com/open-telemetry/otel-arrow/pull/1078/files#r2330706572

rust/otap-dataflow/crates/otap/src/pdata.rs
albertlockett: I think a better interface could be:
```rs
pub fn into_parts(self) -> (OtapPayload, Context) {
    (self.payload, self.context)
}
```
This way, the caller takes ownership of the context, and after possibly modifying the `OtapPayload`, then can combine the payload w/ the original context back into an `OtapPdata`. This also avoids leaving the `OtapPdata` in a partially moved state without it being detectable by the compiler.

This kind of `into_parts` method for destructuring isn't so unusual. For example, arrow`s `RecordBatch` has something similar:
https://docs.rs/arrow/latest/arrow/array/struct.RecordBatch.html#method.into_parts

jmacd:
Summary of several commits responding to @lquerel.

I removed all the `Message::PData(mut pdata)` cases, replaced them with explicit notes about when context drops. I added a `take_apart()` method which consumes the pdata, but in hindsight I would suggest the name `split()`.

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
albertlockett: I think the reason this requires a `&mut` is because we modify the `OtapArrowRecords` (not the arrow records themselves)

`OtapArrowRecords` can be understood as simply a container of `RecordBatches`, where the `RecordBatches` can be accessed by the methods `get(&self, payload_type: ArrowPayloadtype)` and `set(&mut self, payload_type: ArrowPayloadType)`:
https://github.com/open-telemetry/otel-arrow/blob/c40ffb3cea32d7772039d2315ad969d237c7dd37/rust/otel-arrow-rust/src/otap.rs#L49-L70

So even though the arrow records are immutable, if you take the take the immutable `RecordBatch`, create a new one by transforming the attributes, you'll still need a mutable reference to `OtapArrowRecords` to be able to replace the record batch for the attributes payload type.

I agree that the current interface could be improved and we should think about how it should look.

But I don't think that needing the `&mut` reference in this particular line is a consequence of the `OtlpBytes` vs `OtapArrowRecords` representation. Even if we had only one representation containing `OtapArrowRecords`, I don't think it's necessarily uncouth for this representation to be some collection of `RecordBatches`, and while the records themselves may be immutable, the collection may not be. So to modify the collection the mut reference will be needed.

WRT to this processor being passthrough .. the function that eventually gets called here is is `otel_arrow_rust::otap::transform::transform_attributes`
https://github.com/open-telemetry/otel-arrow/blob/c40ffb3cea32d7772039d2315ad969d237c7dd37/rust/otel-arrow-rust/src/otap/transform.rs#L543-L546
Whether we actually modified the attributes is currently opaque to the caller -- no return value indicates whether or not any attributes were transformed (if there's no transformations to make, we basically just clone the `RecordBatch` which is not too expensive as it's mostly cloning `Arcs`). I guess we could have this function return an `Option::None` if there were no modifications made vs returning a clone? Again though, if the return value was `Option::Some(rb)`, you'd still need a mutable reference to the container of `RecordBatches` inside the pdata to replace the old record batch with the new one.

jmacd:
Based on the feedback, I reduced the surface area of the OtapPdata container and renamed two methods. The core API is payload() (test only), split(), and clone_payload().

Most processors split, manipulate, and reassemble. The debug processor uses clone_payload() instead. Most tests ignore context, thus just use payload(). I think I've removed all the extraneous `mut` modifiers at this point.

Also note that with today's changes, there are no uses of take_context(), which I have removed.

```
    /// Returns the payload from this request, consuming it.  This is
    /// only considered useful in testing.  Use split() to split an
    /// OtapPdata into (Context, OtapPayload).
    #[must_use]
    #[cfg(test)]
    pub fn payload(self) -> OtapPayload {
        self.payload
    }

    /// Splits the context and payload from this request, consuming it.
    #[must_use]
    pub fn split(self) -> (Context, OtapPayload) {
        (self.context, self.payload)
    }

    /// Clones the payload from this request.
    #[must_use]
    pub fn clone_payload(&self) -> OtapPayload {
        self.payload.clone()
    }
```

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -129,85 +137,283 @@ impl OtlpProtoBytes {
...
+                    .get(otel_arrow_rust::proto::opentelemetry::arrow::v1::ArrowPayloadType::Spans)
+                    .map_or(0, |batch| batch.num_rows())
+            }
+            Self::Metrics(_) => {
+                self.get(otel_arrow_rust::proto::opentelemetry::arrow::v1::ArrowPayloadType::UnivariateMetrics)
+                    .map_or(0, |batch| batch.num_rows())
+            }
+        }
+    }
+}
```
albertlockett: Should this just call `self.batch_length()`?
https://github.com/open-telemetry/otel-arrow/blob/1baa5f5be1526aaf2077ba0b58ff4ce94c4755f4/rust/otel-arrow-rust/src/otap.rs#L96-L102

Note - `batch_length` for metrics counts the datapoints, not the rows in the `UnivariateMetrics` `RecordBatch`:
https://github.com/open-telemetry/otel-arrow/blob/1baa5f5be1526aaf2077ba0b58ff4ce94c4755f4/rust/otel-arrow-rust/src/otap.rs#L368-L387

rust/otap-dataflow/crates/otap/src/pdata.rs
albertlockett: Looks like this change was made:
https://github.com/open-telemetry/otel-arrow/pull/1078#issuecomment-3267124754
Thanks @jmacd !

rust/otap-dataflow/crates/otap/src/pdata.rs
lquerel: Maybe we should have a `with_default_context` only accessible from the tests (or from the internal of the pipeline engine) and not from the node implementations. What do you think?

And introduce a `new(context, payload)` in a future PR like you said.

lquerel: Rereading your other comments, I think that is probably something we should do in a future PR (i.e. the accessibility part) and just do the renaming for this PR.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -75,12 +77,18 @@
...
-use otel_arrow_rust::otap::OtapArrowRecords;
+use otel_arrow_rust::otap::{OtapArrowRecords, OtapBatchStore};
 use otel_arrow_rust::otlp::{logs::logs_from, metrics::metrics_from, traces::traces_from};
 use prost::{EncodeError, Message};
 
 use crate::encoder::{encode_logs_otap_batch, encode_spans_otap_batch};
 
+/// Context for OTAP requests
+#[derive(Clone, Debug, Default)]
+pub struct Context {
```
lquerel, line 89: In a future PR, a node should not be able to create a context without updating an existing one or deriving it from an existing one.

lquerel: via `#[cfg(test)]` we could enable tests to create a context manually.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -129,85 +137,272 @@ impl OtlpProtoBytes {
...
+    #[must_use]
+    pub fn split(self) -> (Context, OtapPayload) {
+        (self.context, self.payload)
+    }
+
+    /// Clones the payload from this request.
+    #[must_use]
+    pub fn clone_payload(&self) -> OtapPayload {
+        self.payload.clone()
+    }
```
lquerel: For me, these methods are temporary and I don't think we should ultimately expose them at all for the nodes. I'll create a GitHub issue to describe what I have in mind at the pdata level to avoid this. But for now, I don't want to block you.

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
```diff
@@ -145,8 +145,11 @@ impl Exporter<OtapPdata> for OTLPExporter {
                     // Capture signal type before moving pdata into try_from
                     let signal_type = pdata.signal_type();
 
+                    // TODO: context is lost
+                    let (_context, payload) = pdata.split();
```
lquerel: @jmacd, @albertlockett - In a future PR, I have the impression that we could eventually simplify all this in the following way (it's a simplified implementation). It's at least an approach to explore in my opinion.

```rust
Message::PData(pdata) => {
  // Borrow-only conversion. No clone, no ownership changes yet.
  match pdata.as_otlp_proto_bytes() {
     Ok(service_req) => {
          // Attempt transport using borrowed bytes.
          match self.export(&service_req).await {
            Ok(()) => {
              effect_handler.reply_ack(pdata);  // pdata is still intact and available here
            }
            Err(e) => {
              // Just for this example, I choose to consider export issues transient
              effect_handler.reply_nack(pdata, RetryableError::Transient(e));
            }
          }
     }
     Err(e) => {
       // Conversion failure is categorically permanent => no clones needed.
       effect_handler.reply_nack(pdata, RetryableError::Permanent(e));
     }
  }
}
```

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -194,13 +194,15 @@ impl local::Processor<OtapPdata> for AttributesProcessor {
                 }
 
                 let signal = pdata.signal_type();
-                let mut records: OtapArrowRecords = pdata.try_into()?;
+                let (context, payload) = pdata.split();
```
lquerel: @jmacd , @albertlockett At some point, I'd like us to get to something as simple as this, which in my opinion is much simpler without sacrificing performance.

```rust
// If transfor_attributes do nothing then it returns self, otherwise it returns
//  a new pdata (including the context of consumed one).
let pdata = pdata.transform_attributes(&transforms);
effect_handler.send_message(pdata);
```

rust/otap-dataflow/crates/otap/src/debug_processor.rs
lquerel: @jmacd @albertlockett We could use the view mechanism to offer a navigable view of the data inside the pdata. Ideally, we'd have views with native support for both backends, `OtlpProtoBytes` and `OtapArrowRecords`. For now, we don't have the one for arrow records, I think, but in my opinion it's fundamentally not impossible.

Roughly that will be something like 

```rust
let signals = pdata.navigable_view();
// display the signals following the logic used in the current DebugProcessor
effect_handler.send_message(pdata); // pdata unchanged
```

rust/otap-dataflow/crates/otap/src/otap_batch_processor.rs
```diff
@@ -550,9 +550,14 @@ impl local::Processor<OtapPdata> for OtapBatchProcessor {
                     | otap_df_engine::control::NodeControlMsg::CollectTelemetry { .. } => Ok(()),
                 }
             }
-            Message::PData(data) => {
+            Message::PData(request) => {
                 let max = self.config.send_batch_max_size;
-                match OtapArrowRecords::try_from(data.clone()) {
+                let signal_type = request.signal_type();
```
lquerel, line 555: @jmacd  @albertlockett  At a high level, the batch processor code could look something like this.

```rust
self.pdata_msgs.push(pdata);

// ... decide when to flush ...

// conditions to flush are there
let pdata_msgs = std::mem::take(&mut self.pdata_msgs);
let (pdata, remainder_pdata) = make_output_batches(pdata_msgs);

if let Some(rem) = remainder_pdata {
  self.pdata_msgs.push(rem);
}
if let Some(batch) = pdata {
  effect_handler.send_message(batch);
}
```

lquerel (approved):
I've made a whole bunch of comments/proposals for future PRs. I think we can merge as is, but to me we're clearly in a transitional phase where (1) we must/can simplify the use of pdata, and (2) we can make OtapPdata’s internal state more opaque in the vast majority of cases (if not all the time). The ultimate goal is to have a simpler, less error-prone API while remaining at least as performant. I'm including @albertlockett , @msalib , and @utpilla for visibility on these proposals. I'll try to compile all this into a GitHub issue to make feedback easier and see how we can build a plan.

rust/otap-dataflow/crates/otap/src/pdata.rs
jmacd: I will use `into_parts`

rust/otap-dataflow/crates/otap/src/debug_processor.rs
jmacd: I removed clone_payload(), used instead clone().into_parts() and dropped the context.

jmacd:
I recorded the safety issue in #1098 will merge this ASAP.

PR #1082, opened 2025-09-05
[query-engine] KQL regex parsing and cleanup

CodeBlanch:
## Changes

* Adds regex parsing in KQL
* Cleans up the KQL pest a bit

### DIFF
--- a/rust/experimental/query_engine/kql-parser/src/kql.pest
+++ b/rust/experimental/query_engine/kql-parser/src/kql.pest
@@ -84,25 +84,19 @@ null_literal = {
     ("long(" ~ "null" ~ ")") |
     ("real(" ~ "null" ~ ")") |
     ("double(" ~ "null" ~ ")") |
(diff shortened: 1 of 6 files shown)


rust/experimental/query_engine/kql-parser/src/kql.pest
```diff
@@ -196,22 +213,19 @@ arithmetic_expression = {
     arithmetic_factor ~ ((plus_token | minus_token) ~ arithmetic_factor)*
 }
 
-// Modify scalar_expression to include arithmetic_expression as first option
+/* Note: Order is imporant here. Once Pest has matched something it won't go
+backwards. For example if integer_literal is defined before time_expression "1h"
+would be parsed as integer_literal(1) and the remaining "h" would be fed into
+the next rule. */
```
drewrelmas, line 219: Hopefully our existing test coverage on parsing would be enough to catch the sort of bug, but I understand there isn't a perfect way to do so.

PR #977, opened 2025-08-20
[WIP] Retry processor -> otap-df-otap

jmacd:
Part of #783 
Part of #888 

I am trying to fill in gaps with node control message delivery, but meanwhile have become confused about async and block_on.

PR #1083, opened 2025-09-06
[DRAFT] Back-propagation part 2 of 3: Add delayed data in pipeline and node control messages

jmacd:
This adds <PData> to the pipeline control message sender/receiver, which makes it wide.

Adds a new Delayed<> data type for sorting Box<PData> by timestamp to resume, node ID.

Taken from #1041.

Part of #977.


lquerel:
I'd like to separate the pipeline's internal control messages from the command messages between the admin and the pipeline engine. That will prevent propagation outside the pipeline engine when we make this kind of change on the control side. I'll take care of it today.

jmacd:
Closed in favor of #1181 after applying I believe all of @lquerel's suggestions in #1114.

Issue #1084, opened 2025-09-07
OTAP exporter support compression for Arrow IPC


albertlockett:
Currently the go collector's otel-arrow exporter supports two levels of compression: grpc payload and the arrow ipc payload. For example:
```yaml
exporters:
  otelarrow:
    compression: none
    arrow:
      payload_compression: zstd
```

However the rust's implementation only supports compression for the grpc payload:
https://github.com/open-telemetry/otel-arrow/blob/18996ac489c7ed98f315b803275a3dc112844fc5/rust/otap-dataflow/crates/otap/src/otap_exporter.rs#L50-L53

We should add support for compressiong the arrow IPC payload as well. To do:
- add new option to the otap_exporter's config with similar yaml structure as shown above
- modify `otel_arrow_rust::encode::Producer` 
  - to accept some parameters for whether to enable compression
  - to pass these into the `arrow_ipc::writer::StreamWriter` when it is created (see IpcWriterOptions https://docs.rs/arrow-ipc/latest/arrow_ipc/writer/struct.StreamWriter.html#method.try_new_with_options)
- use this compression setting when creating the `Producer` in the otap exporter.

Issue #1068, opened 2025-09-05
[otap-dataflow] Go collector zstd compression /w arrow fails


clhain:
Sending OTAP from the go collector to df_engine with Zstd compression results in errors. This is seen on the go collector logs:

```
2025-09-05T00:36:23.410Z        info    service@v0.133.0/service.go:187 Setting up own telemetry...     {"resource": {"service.instance.id": "7c1ecd2e-1bea-4fcb-8979-0a7fdd416545", "service.name": "otelarrowcol", "service.version": "0.42.0"}}
2025-09-05T00:36:23.416Z        info    service@v0.133.0/service.go:249 Starting otelarrowcol...        {"resource": {"service.instance.id": "7c1ecd2e-1bea-4fcb-8979-0a7fdd416545", "service.name": "otelarrowcol", "service.version": "0.42.0"}, "Version": "0.42.0", "NumCPU": 16}
2025-09-05T00:36:23.416Z        info    extensions/extensions.go:41     Starting extensions...  {"resource": {"service.instance.id": "7c1ecd2e-1bea-4fcb-8979-0a7fdd416545", "service.name": "otelarrowcol", "service.version": "0.42.0"}}
2025-09-05T00:36:23.422Z        info    otelarrowreceiver@v0.133.0/otelarrow.go:102     Starting GRPC server    {"resource": {"service.instance.id": "7c1ecd2e-1bea-4fcb-8979-0a7fdd416545", "service.name": "otelarrowcol", "service.version": "0.42.0"}, "otelcol.component.id": "otelarrow", "otelcol.component.kind": "receiver", "otelcol.signal": "logs", "endpoint": "0.0.0.0:4317"}
2025-09-05T00:36:23.422Z        info    service@v0.133.0/service.go:272 Everything is ready. Begin running and processing data. {"resource": {"service.instance.id": "7c1ecd2e-1bea-4fcb-8979-0a7fdd416545", "service.name": "otelarrowcol", "service.version": "0.42.0"}}
2025-09-05T00:36:23.433Z        info    arrow/stream.go:218     arrow is not supported  {"resource": {"service.instance.id": "7c1ecd2e-1bea-4fcb-8979-0a7fdd416545", "service.name": "otelarrowcol", "service.version": "0.42.0"}, "otelcol.component.id": "otelarrow", "otelcol.component.kind": "exporter", "otelcol.signal": "logs", "message": "Content is compressed with `zstdarrow5` which isn't supported"}
```

[...]

PR #1085, opened 2025-09-07
Add configurable arrow IPC compression to otap exporter & align default compression/IPC stream options

albertlockett:
Closes: #1084 

Adds the ability to set the Arrow IPC compression in the OTAP exporter, similar to what is configurable in configurable in the golang exporter:
```yaml
exporter:
  config:
    arrow:
      payload_compression: zstd / none
```
`ztd` is the default, same as golang collector:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/adbd65ed00ff447a96ce9bdce1320424e5bb420d/exporter/otelarrowexporter/internal/arrow/exporter.go#L62

This PR also changes the default gRPC compression to zstd, which is also what is configured by default in the golang collector:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/adbd65ed00ff447a96ce9bdce1320424e5bb420d/exporter/otelarrowexporter/factory.go#L68-L69

So by default, we have the "double compression".

This PR also turns on dictionary delta encoding in the IPC stream reader. This also aligns with what the go collector configures:
https://github.com/open-telemetry/otel-arrow/blob/18996ac489c7ed98f315b803275a3dc112844fc5/go/pkg/otel/arrow_record/producer.go#L403

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/compression.rs
+++ b/rust/otap-dataflow/crates/otap/src/compression.rs
@@ -12,10 +12,13 @@ use tonic::codec::CompressionEncoding;
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum CompressionMethod {
     /// Fastest compression
+    #[serde(alias = "zstd")]
     Zstd,
     /// Most compatible compression method
+    #[serde(alias = "gzip")]
     Gzip,
     /// Used for legacy systems
+    #[serde(alias = "deflate")]
     Deflate,
 }
 


rust/otap-dataflow/crates/otap/src/otap_exporter/config.rs
```diff
@@ -0,0 +1,111 @@
...
+#[derive(Debug, Deserialize)]
+pub struct ArrowConfig {
+    /// Compression to use for IPC serialized payloads within the BatchArrowMessages
+    ///
+    /// This defaults to "zstd", wherein arrow IPC stream will be compressed with IPC
+    /// independently of whatever compression gRPC may have been configured. This is on by
+    /// default, achieving "double compression" because:
+    /// (a) relatively cheap in CPU terms
+    /// (b) minor compression benefit
+    /// (c) helps stay under gRPC request size limits
```
albertlockett, line 38: This comment reuses the same justification for the double compression explained here, which I'm assuming is still valid
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/adbd65ed00ff447a96ce9bdce1320424e5bb420d/exporter/otelarrowexporter/internal/arrow/exporter.go#L54-L61

albertlockett:
Tested this with:

OTAP exporter config like this -- which defaults to zstd for grpc & zstd for arrow compression (double compression)
```yaml
  exporter:
    kind: exporter
    plugin_urn: "urn:otel:otap:exporter"
    config:
      grpc_endpoint: "http://127.0.0.1:4317"
```

and otel collector receiver configured like:
```yaml
receivers:
  otelarrow:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317

exporters:
  debug:
    verbosity: detailed
service:
  telemetry:
    logs:
      level: "debug"

  pipelines:
    logs/arrow:
      receivers: [otelarrow]
      exporters: [debug]
```

And this seems to work (not seeing failures)
```
$ curl -i -s "http://localhost:8087/telemetry/metrics?reset=false&format=prometheus" | grep -v "#" | grep logs
logs_produced{set="fake_data_generator.receiver.metrics",node_id="receiver",node_type="receiver",pipeline_id="default_pipeline",core_id="0",numa_node_id="0",process_instance_id="AGMSM257JB3MBKY6RY2NLDOW2Q",host_id="",container_id=""} 9 1757286240165
logs_consumed{set="exporter.pdata",node_id="exporter",node_type="exporter",pipeline_id="default_pipeline",core_id="0",numa_node_id="0",process_instance_id="AGMSM257JB3MBKY6RY2NLDOW2Q",host_id="",container_id=""} 10 1757286240165
logs_exported{set="exporter.pdata",node_id="exporter",node_type="exporter",pipeline_id="default_pipeline",core_id="0",numa_node_id="0",process_instance_id="AGMSM257JB3MBKY6RY2NLDOW2Q",host_id="",container_id=""} 10 1757286240165
logs_failed{set="exporter.pdata",node_id="exporter",node_type="exporter",pipeline_id="default_pipeline",core_id="0",numa_node_id="0",process_instance_id="AGMSM257JB3MBKY6RY2NLDOW2Q",host_id="",container_id=""} 0 1757286240165
```

jmacd:
The experimental data was pretty consistent on this point. Double compression saved 3-7% or so in space, without appearing expensive (ymmv).

jmacd (approved):

albertlockett:
> The experimental data was pretty consistent on this point. Double compression saved 3-7% or so in space, without appearing expensive (ymmv).

Great, thanks for confirming @jmacd !

rust/otap-dataflow/crates/otap/src/otap_exporter/config.rs
lquerel: I think we'll need to retest point (a), but at least it's consistent with the Go implementation, and we have the option to disable compression.

lquerel:
> Tested this with:
>
> OTAP exporter config like this -- which defaults to zstd for grpc & zstd for arrow compression (double compression)

The issue with zstdarrow[0-9] happens between the Go OTLP exporter and the Rust OTLP receiver (in this validation test it's the opposite). But anyway it's not something we can fix on the Rust side. So it's good for me.

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
```diff
@@ -169,25 +164,33 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
         let (traces_sender, traces_receiver) = tokio::sync::mpsc::channel(64);
         let (pdata_metrics_tx, mut pdata_metrics_rx) = tokio::sync::mpsc::channel(64);
         let (shutdown_tx, shutdown_rx) = tokio::sync::watch::channel(false);
+        let ipc_compression = matches!(
+            self.config.arrow.payload_compression,
+            Some(config::ArrowPayloadCompression::Zstd)
+        )
+        .then(|| arrow_ipc::CompressionType::ZSTD);
```
lquerel, line 171: So with this implementation, I get the impression that we can no longer disable compression. I think we need to introduce a `none` compression method like in the Go version or apply the same approach you used for the arrow compression method.

rust/otap-dataflow/crates/otap/src/otap_exporter/config.rs
```diff
@@ -0,0 +1,111 @@
...
+    pub payload_compression: Option<ArrowPayloadCompression>,
+}
+
+/// Compression options for arrow payloads
+#[derive(Debug, Deserialize)]
+pub enum ArrowPayloadCompression {
+    /// Zstd compression
+    #[serde(alias = "zstd")]
+    Zstd,
+}
```
lquerel, line 54: I like the use of the alias, we should use the same approach for CompressionMethod.

albertlockett:
@lquerel that's right -- there's an orthogonal issue to make the go collector's compression header compatible with tonic. Tracked here https://github.com/open-telemetry/otel-arrow/issues/1068

lquerel (dismissed):
I think we should be able to independently disable compression from the configuration at the gRPC level and at the Arrow payload level. Unless I'm mistaken, that's not the case at the gRPC level. Other than that, everything looks perfect.

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
albertlockett: We can disable compression - we can pass the string "none", same as in Go version.

```
  exporter:
    kind: exporter
    plugin_urn: "urn:otel:otap:exporter"
    config:
      grpc_endpoint: "http://127.0.0.1:4317"
      compression_method: "none"
      arrow:
        payload_compression: "none"
```

This will get deserialized as the `None` option using this special helper:
https://github.com/open-telemetry/otel-arrow/blob/775733780f6daa9e20579ea9bb5dbf05e95b3c0a/rust/otap-dataflow/crates/otap/src/otap_exporter/config.rs#L72-L93

There's also a test for this:
https://github.com/open-telemetry/otel-arrow/blob/775733780f6daa9e20579ea9bb5dbf05e95b3c0a/rust/otap-dataflow/crates/otap/src/otap_exporter.rs#L705-L730

rust/otap-dataflow/crates/otap/src/otap_exporter/config.rs
albertlockett: done in https://github.com/open-telemetry/otel-arrow/pull/1085/commits/db1473db0e68c2bb6d228605989b1df2d80a491f

jmacd:
Compression configuration is terribly complex! At least in Golang. I would be glad to see us simplify in the long term, but I also see how this helps with short-term compatibility. I think most users will be happy with the default level 6.

PR #1093, opened 2025-09-08
[otap-dataflow] add signal option to the debug processor config

c1ly:
Update the debug processor to allow the user to select which signals should get displayed


jmacd (approved):

jmacd:
@c1ly please take a look at the failed test, thanks.

c1ly:
> @c1ly please take a look at the failed test, thanks.

should be good now

PR #1094, opened 2025-09-08
Add tonic middleware to handle go exporter's grpc-encoding zstdarrow header value

albertlockett:
closes: #1068

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/otap_grpc/middleware/zstd_header.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_grpc/middleware/zstd_header.rs
@@ -1,0 +1,44 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Middlewares for adapting gRPC encoding request header
+
+use async_trait::async_trait;
+use http::{Request, Response};
+use tonic::body::Body;
+use tonic_middleware::{Middleware, ServiceBound};
+
+const ENCODING_HEADER: &str = "grpc-encoding";
+
+/// Tonic middleware implementation that will replace the `grpc-encoding` header that the golang
+/// exporter produces (which has the format zstdarrow[0-9]) with the value that tonic expects, 
+/// which is just "zstd"
+#[derive(Clone, Default)]
+pub struct ZstdRequestHeaderAdapter {}
+
+#[async_trait]
+impl<S> Middleware<S> for ZstdRequestHeaderAdapter
+where
+    S: ServiceBound,
+    S::Future: Send,
+{
+    async fn call(
+        &self,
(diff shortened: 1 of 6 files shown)


albertlockett:
Tested this using:
```yaml
nodes:
  receiver:
    kind: receiver
    
    plugin_urn: "urn:otel:otap:receiver"
    out_ports:
      out_port:
        destinations:
          - debug
    config:
      listening_addr: "0.0.0.0:1235"
      message_size: 100
      compression_method: Zstd
 # ...
```

Golang collector config:
```yaml
exporters:
  otelarrow:
    endpoint: '127.0.0.1:1235'
    compression: zstd
    tls:
      insecure: true
```

albertlockett:
We might only need this as a temporary workaround until https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/42536 is merged/released.. although we could also keep it for backward compatibility with go collectors that don't have that fix

rust/otap-dataflow/crates/otap/src/otap_grpc/middleware/zstd_header.rs
```diff
@@ -0,0 +1,44 @@
...
+        mut req: Request<Body>,
+        mut service: S,
+    ) -> Result<Response<Body>, S::Error> {
+        let headers = req.headers_mut();
+        if let Some(header_val) = headers.get(ENCODING_HEADER) {
+            let header_bytes = header_val.as_bytes();
+            if header_bytes.starts_with(b"zstdarrow") {
+                _ = headers.insert(
+                    ENCODING_HEADER,
+                    "zstd".parse().expect("can parse to header value"),
```
utpilla: I think we could pre-compute `HeaderName` and `HeaderValue` to avoid some repeated conversions and parsing.

```rust
static ENCODING_HEADER_NAME: HeaderName = HeaderName::from_static("grpc-encoding");
static ZSTD_HEADER_VALUE: HeaderValue = HeaderValue::from_static("zstd");

...
async fn call(
    &self,
    mut req: Request<Body>,
    mut service: S,
) -> Result<Response<Body>, S::Error> {
    let headers = req.headers_mut();
    if let Some(header_val) = headers.get(&ENCODING_HEADER_NAME) {
        let header_bytes = header_val.as_bytes();
        if header_bytes.starts_with(b"zstdarrow") {
            let _ = headers.insert(&ENCODING_HEADER_NAME, ZSTD_HEADER_VALUE.clone()); // // Use pre-computed static value to avoid parsing
        }
    }

    let result = service.call(req).await?;
    Ok(result)
}
```

albertlockett: Thanks @utpilla, made these changes

jmacd (approved):

PR #1097, opened 2025-09-09
[otap-dataflow] SignalTypeRouter: internal telemetry

daviddahl:
Summary:
•  Add internal telemetry to SignalTypeRouter:
•  Define SignalTypeRouterMetrics (signals_received_, signals_routed_named_, signals_routed_default_, signals_dropped_).
•  Register metrics via PipelineContext (with_pipeline_ctx) and handle CollectTelemetry to report snapshots.
•  Increment counters on every PData: detect named vs default port, record success/failure per signal type.
•  Add unit tests (scoped under tests::telemetry) covering logs/metrics/traces across named/default routing (success and forced failure) using MetricsSystem for collection.
•  Tests only affect test code; production changes are limited to the router’s telemetry instrumentation and collection path.


rust/otap-dataflow/crates/otap/src/signal_type_router.rs
```diff
@@ -35,6 +39,94 @@ pub const PORT_METRICS: &str = "metrics";
 /// Name of the out port used for log signals
 pub const PORT_LOGS: &str = "logs";
 
+/// Metrics for the SignalTypeRouter processor.
```
daviddahl, line 42: i left the metrics inside the STR module as it seems like a smaller set of values that may not grow too much.

jmacd (approved):

PR #1103, opened 2025-09-10
[query-engine] Rename transformations

CodeBlanch:
## Changes

* Adds `Move` and `RenameMapKeys` transformations
* Adds support for KQL `project-rename` expression

### DIFF
--- a/rust/experimental/query_engine/kql-parser/src/scalar_primitive_expressions.rs
+++ b/rust/experimental/query_engine/kql-parser/src/scalar_primitive_expressions.rs
@@ -567,7 +567,7 @@ pub(crate) fn parse_accessor_expression(
                                 format!(
                                     "Cannot access into key '{}' which is defined as a '{:?}' type",
                                     root_accessor_identity.get_value(),
-                                    key.get_value_type()
+                                    key.get_value_type().unwrap()
                                 ),
                             ));
                         }


rust/experimental/query_engine/engine-recordset/src/transform/transform_expressions.rs
```diff
@@ -20,6 +20,163 @@ pub fn execute_transform_expression<'a, TRecord: Record>(
...
+                                execution_context.add_diagnostic_if_enabled(
+                                    RecordSetEngineDiagnosticLevel::Verbose,
+                                    m,
+                                    || format!("Array index '{index}' could not be updated on target array: {e}"),
+                                );
+                            }
+                        }
+                    }
+                },
+                None => {
```
drewrelmas, line 175: To my eyes we'll just drop the `value` if Destination could not be resolved - we've already removed it from Source. Can you confirm? Wondering if better default behavior is leaving it on Source with this error message.

drewrelmas (approved):

rust/experimental/query_engine/expressions/src/transform_expressions.rs
```diff
@@ -27,30 +33,36 @@ impl TransformExpression {
...
         }
     }
 
     fn get_name(&self) -> &'static str {
         match self {
-            TransformExpression::Set(_) => "Transform(Set)",
-            TransformExpression::Remove(_) => "Transform(Set)",
+            TransformExpression::Move(_) => "Transform(Move)",
             TransformExpression::ReduceMap(r) => r.get_name(),
+            TransformExpression::Remove(_) => "Transform(Remove)",
```
drewrelmas, line 62: Must have missed this mistake on previous review, nice catch!

rust/experimental/query_engine/kql-parser/src/tabular_expressions.rs
```diff
@@ -263,6 +263,67 @@ pub(crate) fn parse_project_away_expression(
...
+    if expressions.len() == 1 {
+        let (location, source, destination) = expressions.drain(..).next().unwrap();
+
+        Ok(TransformExpression::Move(MoveTransformExpression::new(
+            location,
+            MutableValueExpression::Source(source),
+            MutableValueExpression::Source(destination),
+        )))
+    } else {
+        let mut keys = Vec::with_capacity(expressions.len());
```
drewrelmas, line 305: I think maybe important to note we're actually embellishing the KQL grammar here - nothing in [project-rename documentation](https://learn.microsoft.com/kusto/query/project-rename-operator?view=microsoft-fabric) actually suggests we can do this on subkeys.

```kql
print a='alpha', Attributes=dynamic({"b": 'bravo'})
| project-rename new_a = a, Attributes['new_b'] = Attributes['b']
```

The `Attributes['new_b'] = Attributes['b']` causes error:

> KS110 - Column rename assignment expected

I get that this is almost required for us because of the OpenTelemetry data model we want to support where users are very likely to rename things in the Attributes dictionary, but maybe worth leaving a comment about?

rust/experimental/query_engine/engine-recordset/src/transform/transform_expressions.rs
drewrelmas: This still wasn't addressed, could you please share your thoughts?

drewrelmas:
Resolves #1102

rust/experimental/query_engine/kql-parser/src/tabular_expressions.rs
CodeBlanch: For sure yes. We allow these accessor expressions a lot of places KQL doesn't. This is true in `project-rename` as well as `extend`, `project`, `project-keep`, and `project-away`.

It isn't necessarily due to Attributes though, or OTel. Query engine is running over complex objects and has a bunch of tools/magic for dealing with them. KQL is built for columns.

What these extensions to KQL are doing are allowing users to use the features of the engine to work with _any_ complex objects. Just so happens an OTel log is a complex object 😄

Like you could do `project-away Body.message.details` to remove whatever hangs off of `Body.message.details` while keep everything else.

For rename you can do stuff like `project-rename MessageDetails = Body.message.details` and it will remove the details value/tree from Body.message and promote it to a top-level attribute.

I'm happy to put comments but I don't know exactly where to do that 😄 It would be like _everywhere_. Probably the real need here is our own docs?

drewrelmas: That's fair, this is a meta thing unconnected to this actual PR. No real changes needed here, but we should do a doc round at some point.

drewrelmas: Opened #1110

rust/experimental/query_engine/engine-recordset/src/transform/transform_expressions.rs
CodeBlanch: Good catch! Yes this is an issue. It is kind of a good news/bad news situation.

## Bad news

We love the safety of Rust but it comes with some tradeoffs. One of its strict rules is you can't have simultaneous immutable and mutable references to something or multiple mutable references ever. This makes move/rename difficult!

The current structure is we first do the remove part. If the thing we are moving isn't found, we log a warning and complete. If the thing is found, we remove it and take ownership. Then we go to do the set with the owned value. If we can't find the thing or can't complete the write we log a warning, drop the value, and complete. This order makes Rust happy, basically.

If it is unacceptable to drop the value (which I'm not sure of), we could try to validate the destination first or rollback the state on failure but these things are expensive. Up-front validation will slow down happy path successful things. Rollback will take a lot of code and may also fail (although at first thought that seems unlikely).

I'm open to either but I'm leaning towards leaving as-is for now because there is...

## Good news

KQL parser has some schema knowledge + validation of queries.

In the "OTel mode" where we have "Attributes" which is the default map, the write into the destination will be unlikely to fail for simple renames. `source | new_thing = old_thing` is just a `HashMap` insert no reason for it to fail.

In strict mode, the parser will yield errors.

Let's say we define a schema with "B:String" and try to do `project-rename A = B` we will get: `KS109: The name 'A' does not refer to any known column, table, variable or function`.

Let's say we define a schema with "A:String, B:String" and try to do `project-rename A['invalid'] = B` we will get: `SyntaxError: Cannot access into key 'A' which is defined as a 'String' type`.

drewrelmas: I think this makes sense for now - I think the vast majority of the time project-rename will work in context with a default map as you mention.

PR #1105, opened 2025-09-10
feat: improve performance of dictionary unification code in batching

msalib:
We employ a few different tricks to improve performance here, mostly by removing expensive operations from our inner loop:

* Originally, we were performing type dispatch for each element. Now, we dispatch once per array.

* We used to check for null elements for every single index in the array; now we employ `NullBuffer::valid_slices()` to get a contiguous slice of valid elements; in the worst case (an alternating sequence of null and not-null values), this devolves into the same performance we had before, but in all other cases, this allows us to amortize null-handling over a much larger set of values.

* For arrays that don't have any nulls, either because their data-type is non-null or because there just happen to be no nulls present, we skip null-handling altogether and operate on slices directly.

* We size `HashSet` capacity based on the total number of elements so for small data sets, we should reduce allocation sizes.

* For small element types (those that fit within a `u32`), we skip the `HashSet` of byte slices and opt for a `RoaringBitmap`. [Roaring](https://arxiv.org/pdf/1603.06549) is a crate widely used in database implementations to compactly describe sets of integers by dynamically selecting a representation at runtime that includes dense bit vectors, sparse sequences of integers, and intervals. The result is fast and memory efficient. For anything `u32` sized or smaller, we just convert the bits into a `u32`, stash it in a `RoaringBitmap` and then rely on the bitmap's `len` method. The only caveat here is that the flexible representation means that `RoaringBitmap::len()` does some real work as opposed to `AHashSet::len()` which just reads a single value in memory, so instead of checking for early termination on every array value, we amortize it over every 256 array values.

This is not strictly a performance improvement, but I've also changed `total_batch_size` to refer to the number of non-null items in a sequence of columns. I think the math works but I'd appreciate extra review around that change.

Note that the unified diff for this PR looks really confusing, so you might benefit from switching to the side by side diff view.

### DIFF
--- a/rust/otap-dataflow/crates/config/src/engine.rs
+++ b/rust/otap-dataflow/crates/config/src/engine.rs
@@ -13,6 +13,7 @@ use std::collections::HashMap;
 /// Root configuration for the pipeline engine.
 /// Contains engine-level settings and all pipeline groups.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
 pub struct EngineConfig {
     /// Settings that apply to the entire engine instance.
     pub settings: EngineSettings,
@@ -23,6 +24,7 @@ pub struct EngineConfig {
 
 /// Global settings for the engine.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
 pub struct EngineSettings {
     /// Optional HTTP admin server configuration.
     pub http_admin: Option<HttpAdminSettings>,
@@ -33,6 +35,7 @@ pub struct EngineSettings {
 
 /// Configuration for the telemetry metrics system.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
 pub struct TelemetrySettings {
     /// The size of the reporting channel.
     #[serde(default = "default_reporting_channel_size")]
@@ -44,6 +47,7 @@ pub struct TelemetrySettings {
 
 /// Configuration for the HTTP admin endpoints.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
 pub struct HttpAdminSettings {
     /// The address to bind the HTTP server to (e.g., "127.0.0.1:8080").
     #[serde(default = "default_bind_address")]
--- a/rust/otap-dataflow/crates/config/src/node.rs
+++ b/rust/otap-dataflow/crates/config/src/node.rs
@@ -23,6 +23,7 @@ use std::collections::{HashMap, HashSet};
 /// - The hyper-edge configuration determines which downstream nodes are connected,
 ///   and how messages are routed (broadcast, round-robin, ...).
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
 pub struct NodeUserConfig {
     /// The kind of this node, which determines its role in the pipeline.
     /// 4 kinds are currently specified:
@@ -65,6 +66,7 @@ pub struct NodeUserConfig {
 /// Describes a hyper-edge from a node output port to one or more destination nodes,
 /// and defines the dispatching strategy for this port.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
 pub struct HyperEdgeConfig {
     /// List of downstream node IDs this port connects to.
     ///
@@ -182,5 +184,35 @@ impl NodeUserConfig {
     /// Sets the default output port name used by this node when no explicit port is specified.
     pub fn set_default_out_port<P: Into<PortName>>(&mut self, port: P) {
         self.default_out_port = Some(port.into());
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn node_user_config_unknown_field_is_rejected() {
+        // Includes an extra unknown field "unknown" that should cause an error
+        let json = r#"{
+            "kind": "receiver",
+            "plugin_urn": "urn:example:receiver",
+            "out_ports": {},
+            "config": {},
+            "unknown": 1
+        }"#;
+        assert!(serde_json::from_str::<NodeUserConfig>(json).is_err());
+    }
+
+    #[test]
+    fn node_user_config_minimal_valid() {
+        let json = r#"{
+            "kind": "receiver",
+            "plugin_urn": "urn:example:receiver",
+            "out_ports": {}
+        }"#;
+        let cfg: NodeUserConfig = serde_json::from_str(json).unwrap();
+        assert!(matches!(cfg.kind, NodeKind::Receiver));
+        assert!(cfg.out_ports.is_empty());
     }
 }
--- a/rust/otap-dataflow/crates/config/src/pipeline.rs
+++ b/rust/otap-dataflow/crates/config/src/pipeline.rs
@@ -24,6 +24,7 @@ use std::sync::Arc;
 ///
 /// This configuration defines the pipeline’s nodes, the interconnections (hyper-edges), and pipeline-level settings.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
 pub struct PipelineConfig {
     /// Type of the pipeline, which determines the type of PData it processes.
     ///
@@ -60,6 +61,7 @@ pub enum PipelineType {
 }
 /// A configuration for a pipeline.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
 pub struct PipelineSettings {
     /// The default size of the node control message channels.
     /// These channels are used for sending control messages by the pipeline engine to nodes.
--- a/rust/otap-dataflow/crates/config/src/pipeline_group.rs
+++ b/rust/otap-dataflow/crates/config/src/pipeline_group.rs
@@ -13,6 +13,7 @@ use std::collections::HashMap;
 /// Configuration for a single pipeline group.
 /// Contains group-specific settings and all its pipelines.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
 pub struct PipelineGroupConfig {
     /// All pipelines belonging to this pipeline group, keyed by pipeline ID.
     pub pipelines: HashMap<PipelineId, PipelineConfig>,
@@ -75,6 +76,7 @@ impl Default for PipelineGroupConfig {
 
 /// Pipeline group quota configuration.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Default)]
+#[serde(deny_unknown_fields)]
 pub struct Quota {
     /// CPU core allocation strategy for this pipeline group.
     #[serde(default)]
--- a/rust/otap-dataflow/crates/otap/src/compression.rs
+++ b/rust/otap-dataflow/crates/otap/src/compression.rs
@@ -10,15 +10,13 @@ use tonic::codec::CompressionEncoding;
 
 /// Enum to represent various compression methods
 #[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
 pub enum CompressionMethod {
     /// Fastest compression
-    #[serde(alias = "zstd")]
     Zstd,
     /// Most compatible compression method
-    #[serde(alias = "gzip")]
     Gzip,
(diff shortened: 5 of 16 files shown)


rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1431,144 +1430,386 @@ impl UnifiedDictionaryTypeSelector {
...
+    fn new(arrays: &'parent [ArrayRef], smallest_key_type: &'parent DataType) -> Option<Self> {
+        let total_elements: usize = arrays.iter().map(|v| v.len() - v.null_count()).sum();
+        let elements_seen = 0;
+        let bitmap = roaring::RoaringBitmap::new();
+
+        let element_width_bytes = bitmap_data_type_width(arrays.first()?.data_type())?;
+        // Because we're using RoaringBitmap, we're limited to `u32` and smaller types....
+        if element_width_bytes > 4 {
+            return None;
+        }
```
albertlockett: Would it be slightly faster here to check the datatype before counting the valid elements? I guess this would only make a big difference if there were a large number of arrays

rust/otel-arrow-rust/Cargo.toml
```diff
@@ -40,6 +40,8 @@ tonic-prost = "0.14"
 tokio = { version = "1.43.0", features = ["rt", "time", "net", "io-util", "sync", "macros", "rt-multi-thread", "process"] }
 itertools = "0.14.0"
 smallvec = { version = "1.15.1", features = ["union"] }
+roaring = "0.11.2"
+zerocopy = "0.8.27"
```
albertlockett: is `zerocopy` used anywhere? I didn't see it when scanning the diff

jmacd:
```suggestion
```

jmacd (approved):

rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1431,144 +1430,386 @@ impl UnifiedDictionaryTypeSelector {
...
-        let offset = index * size;
-        &values_buffer[offset..offset + size]
-    })
+impl<'parent> LargeCardinalityEstimator<'parent> {
+    fn new(arrays: &'parent [ArrayRef], smallest_key_type: &'parent DataType) -> Self {
+        let total_elements: usize = arrays.iter().map(|v| v.len() - v.null_count()).sum();
+        let elements_seen = 0;
+
+        let dedup = AHashSet::with_capacity(if total_elements <= u8::MAX as usize {
+            u8::MAX as usize
```
utpilla, line 1668: Why not just use `total_elements` as the capacity?

```suggestion
            total_elements
```

rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1431,144 +1430,386 @@ impl UnifiedDictionaryTypeSelector {
...
+        } else {
+            // TODO, is this too big?
+            u16::MAX as usize
```
utpilla, line 1671: This does seem like over-allocation for a lot of cases. (65,535 for anything greater than 255)  I think a more balanced approach could be to start with a reasonable allocation and let the `Hashmap` grow in size as needed:

```rust
let dedup = AHashSet::with_capacity(if total_elements <= u8::MAX as usize {
    total_elements
} else if total_elements <= u16::MAX as usize {
    // Use a smaller initial capacity but allow for growth
    // Start with 1024 (reasonable default) or sqrt of total elements
    // as a cardinality estimate (assumes some duplicates)
    1024.min(total_elements)
} else {
    // For very large datasets, cap at a reasonable size
    // The HashMap will grow as needed
    4096
});
```

rust/otel-arrow-rust/src/otap/groups.rs
```diff
@@ -1431,144 +1430,386 @@ impl UnifiedDictionaryTypeSelector {
...
+    // these two are immutable inputs
+    arrays: &'parent [ArrayRef],
+    smallest_key_type: &'parent DataType,
+
+    // these are immutable after construction time
+    total_elements: usize, // Note: we only consider non-null items!
+
+    // and these are mutable state variables
+    elements_seen: usize,
+    dedup: AHashSet<&'parent [u8]>,
```
utpilla, line 1659: I don't completely understand all the Arrow batch manipulation details here, but if we don't need exact cardinality (which seems to be the case since we're just determining dictionary encoding thresholds), we should consider using HyperLogLog.

The current `LargeCardinalityEstimator` uses `AHashSet<&'parent [u8]>` which requires `O(n)` memory for `n` unique values. HyperLogLog would use constant memory regardless of cardinality, with a reasonable error rate (like ~2%) that should be fine for the encoding decisions being made here (`UInt8 ≤255`, `UInt16 ≤65,535`, or native).

This could significantly improve memory usage and performance, especially for high-cardinality columns that end up using native encoding anyway. We could even reuse a single HyperLogLog instance across multiple estimations by clearing it between uses.

rust/otel-arrow-rust/src/otap/groups.rs
msalib: For sure! Good catch!

rust/otel-arrow-rust/Cargo.toml
msalib: Ooops, this was left over from an aborted experiment; it turns out we were already transitively depending on `zerocopy` but I'll remove the explicit dependency!

rust/otel-arrow-rust/src/otap/groups.rs
msalib: Great minds think alike: I considered HyperLogLog and looked at a few implementations! Unfortunately, I don't think this use case is one where we can tolerate the error associated with a cardinality approximation, even though it is low. My reasoning is that if we're wrong about cardinality (and with HLL, we will be wrong sometimes), arrow gives us no useful information when it fails to concatenate `RecordBatch`es; we get an opaque error indicating that there was a dictionary overflow, but no indication at all as to which column that occurred in. Which means that we'd have to write all of this code anyway to handle those failure cases. That might still be a worthwhile optimization but since we have to write this code anyway, I thought it better to get the explicit storage approach working and tested first and then look at more real world data to see whether further optimization makes sense.

My only caveat here is that I might not understand HLL's error model well enough -- is there a variant of HLL that guarantees that it never undercounts but may still overcount (sort of like how bloom filters have false positives but never have false negatives)? If such a thing existed, we could apply it here!

rust/otel-arrow-rust/src/otap/groups.rs
msalib: That's an interesting idea! I have a slight preference for the existing code because it reduces the number of allocation classes making it easier to reuse a previous allocation reducing fragmentation, but maybe reducing total bytes allocated trumps that?

If we took that approach, I'd go with `total_elements.min(u16::MAX as usize)` since we shouldn't have more than 2**16 elements.

One consideration that's been driving my thinking is that this is a very short lived allocation but one that we need to make repeatedly -- we build this `HashSet` and throw it away quickly, so it won't be hanging around for a long time. That suggests that reducing reallocations and reducing fragmentation are both helpful.

rust/otel-arrow-rust/src/otap/groups.rs
msalib: Where are you getting the square root from?

utpilla: We want a reasonable estimate for the number of unique values. We do not want to use the total number of elements, as that could lead to excessive memory allocation. At the same time, starting with the default capacity might be too low given our expectation of a large number of items. So, we need a balanced approximation. It doesn’t have to be based on a square root. We could assume a certain percentage of the total elements are unique or simply use a fixed default like 1024 to avoid both under- and over-allocation.

rust/otel-arrow-rust/src/otap/groups.rs
utpilla: > If we took that approach, I'd go with total_elements.min(u16::MAX as usize) since we shouldn't have more than 2**16 elements.

I was mainly referring to the first if check: `if total_elements <= u8::MAX as usize`. In that case, `total_elements` is already `≤ 255`. So, if it's, say, `100`, we could just allocate `100` instead of defaulting to `255`.

> One consideration that's been driving my thinking is that this is a very short lived allocation but one that we need to make repeatedly -- we build this HashSet and throw it away quickly, so it won't be hanging around for a long time. That suggests that reducing reallocations and reducing fragmentation are both helpful.

That's a good point. I hadn't considered efficiency. It might be worth validating this with allocator stats in a simulated scenario. Something to explore in the future, perhaps.

rust/otel-arrow-rust/src/otap/groups.rs
utpilla: Thanks for the detailed explanation! That makes a lot of sense.

There are a few ways to mitigate HLL’s undercounting issue. One approach is to make the estimate more conservative by incorporating confidence intervals. Since we control the error rate when constructing the HLL, we can add multiple standard errors to the estimate to ensure we (almost) never undercount. For example:

```rust
fn conservative_cardinality_estimate(hll: &HyperLogLog, configured_error: f64) -> usize {
    let estimate = hll.count();
    let standard_error = estimate * configured_error;

    // Add 4 standard errors for ~99.99% confidence of not undercounting
    let conservative_estimate = estimate + (4.0 * standard_error);
    conservative_estimate.round() as usize
}

// Usage:
let error_rate = 0.01; // 1% error
let mut hll = HLL::new(error_rate);
// ... insert elements ...
let conservative_count = conservative_cardinality_estimate(&hll, error_rate);
```

We can revisit this idea later if it turns out to be a bottleneck or worth optimizing.

lquerel:
@msalib I haven't fully reviewed the code in detail yet, so it's possible that what follows is already in place (my apologies in advance if that's the case).
Fundamentally, when merging two columns that are dictionary encoded, the lower bound is `max(col_a.values().len(), col_a_prime.values().len())` and the upper bound is `col_a.values().len() + col_a_prime.values().len()`. The upper bound already gives us an idea (at no cost) of whether we can keep the larger of the two dictionary key/index sizes for the merged column. In that case, we don't need to do any estimation with a hashmap or a bitset. This case may not be rare and would thus save us extra work. Another complementary point to consider is an approach I'd call optimistic. If `dict_key_size(col_a) = u8` and `dict_key_size(col_a_prime) = u16`, we can optimistically concatenate these two columns into a column using a u16 dictionary key size, and if the concatenation operation fails, promote to the next larger key size. This approach won’t be efficient if failures happen often, but otherwise it will be faster on average. Perhaps there's a way to keep track of previous concatenations and, depending on the failures, decide ahead of time which strategy to apply (optimistic concatenation vs estimating the cardinality of the final dictionary).
I think the first point is something that can be applied without risk if it isn't already in place. The second point is more of a consideration relative to the current approach.

Adding @albertlockett @utpilla @jmacd for visibility.

jmacd:
I see now that Roaring is used only for this estimation step.

> we can optimistically concatenate these two columns into a column using a u16 dictionary key size, and if the concatenation operation fails, promote to the next larger key size. 

I think I agree that a simpler approach that almost always works would be fine. The cost saved for the times it works could make up for the times it doesn't work?

albertlockett:
> Fundamentally, when merging two columns that are dictionary encoded, the lower bound is `max(col_a.values().len(), col_a_prime.values().len())` and the upper bound is `col_a.values().len() + col_a_prime.values().len()`. The upper bound already gives us an idea (at no cost) of whether we can keep the larger of the two dictionary key/index sizes for the merged column. In that case, we don't need to do any estimation with a hashmap or a bitset. This case may not be rare and would thus save us extra work.

We already have an optimization similar to this ... 
https://github.com/open-telemetry/otel-arrow/blob/1c00e0c1fa3e3fb5f74e91ad7cbbae734e418050/rust/otel-arrow-rust/src/otap/groups.rs#L1414-L1431
`total_batch_size` is the total amount of the non-null values across all the types we're concatenating together. Analogous to `col_a.values().len() + col_a_prime.values().len()` in your example. Basically we check if it would fit into a u8, and return early. 

Then we check if the smallest dict size across any column is u16  and return early if the values would fit in that size dict. In this logic is the implicit assumption that there's no point checking if we can fit the values in a u8 dict b/c maybe u16 is the min dict size for this column (we configure some columns like this, like https://github.com/open-telemetry/otel-arrow/blob/9ddb78ee31c5182d4d145a3ba5f72d9da2fe024e/go/pkg/otel/logs/arrow/logs.go#L46)).

> we can optimistically concatenate these two columns into a column

I agree this would be a good optimization, but one caveat is that we're currently concatenating batches using`BatchCoalescer` which just returns an opaque error that doesn't tell us which column overflowed the dictionary. Ideally we'd fix arrow's `BatchCoalescer` to give us some more helpful feedback when this occurs. Otherwise our options are to a) do the slow-path to calculate the correct dict size for all columns or b) change how we're coalescing batches so we know which column is overflowing the dict.

edit: I created an issue https://github.com/apache/arrow-rs/issues/8339 to improve the `BatchCoalescer` error

PR #1108, opened 2025-09-10
[otap-dataflow] Add OutputMode to the DebugProcessor

c1ly:
OutputMode determines if the debug processor should output by batches or by signal

In the `batch` mode the debug processor will generate a report given the entire batch and then output it to the console
in the `signal` mode the debug processor will generate a report for each individual signal and then output it to the console

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/debug_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/debug_processor.rs
@@ -8,7 +8,7 @@
 //! ToDo: Implement proper deadline function for Shutdown ctrl msg
 //! ToDo: Use OTLP Views instead of the OTLP Request structs
 
-use self::config::{Config, Verbosity};
+use self::config::{Config, OutputMode, SignalActive, Verbosity};
 use self::detailed_marshaler::DetailedViewMarshaler;
 use self::marshaler::ViewMarshaler;
 use self::normal_marshaler::NormalViewMarshaler;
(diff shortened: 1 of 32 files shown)


rust/otap-dataflow/crates/otap/src/debug_processor.rs
```diff
@@ -245,9 +245,11 @@ async fn push_metric(
     let resource_metrics = metric_request.resource_metrics.len();
     let mut data_points = 0;
     let mut metrics = 0;
+    let mut metric_signals = vec![];
     for resource_metrics in &metric_request.resource_metrics {
         for scope_metrics in &resource_metrics.scope_metrics {
             metrics += scope_metrics.metrics.len();
+            metric_signals.append(&mut scope_metrics.metrics.clone());
```
albertlockett: nit: it looks like this `metrics_signal` `vec` is only used if the OutputMode is `Signal`, right? Maybe we should avoid allocating/populating it if it's not used.

```suggestion
    let mut metric_signals = matches!(mode, OutputMode::Signal).then(|| vec![]);
    for resource_metrics in &metric_request.resource_metrics {
        for scope_metrics in &resource_metrics.scope_metrics {
            metrics += scope_metrics.metrics.len();
            if let Some(metric_signals)  = metric_signals {
                metric_signals.append(&mut scope_metrics.metrics.clone());
            }
```

And the below on 295:
```diff
        OutputMode::Signal => {
+           // safety: this should be initialized above if the mode is Signal        
+           let metric_signals = metric_signals.expect("metric_signals not None");
            for (index, metric) in metric_signals.iter().enumerate() {
```

albertlockett: If you think this change is worth making, we should probably make similar change below in `push_trace` and `push_log` functions

Issue #1095, opened 2025-09-09
Support direct conversion OTAP -> OTLP


lquerel:
Today, when the incoming pdata is OtapArrowRecords, the OTLP exporter first converts the Arrow records into struct-based Protobuf messages, then serializes those structs into Protobuf bytes. As on the receiver side, we should be able to bypass creating many small in-memory structs.

PR #1109, opened 2025-09-11
Encode OTAP -> OTLP proto bytes directly

albertlockett:
part of #1095 
closes https://github.com/open-telemetry/otel-arrow/issues/1123

This PR adds an optimized mechanism for encoding OTAP -> OTLP proto bytes directly. This implements the encoding only for Logs. Traces & Metrics can happen in followup PRs. This PR also integrates this optimized mechanism into the OTLP exporter.

In order to do the encoding in a single pass, we allocate a fixed sized length encoding for some fields and zero-pad the varint. This is not optimal from a storage perspective, and we could probably try to optimize the size we guess in the future by inspecting the arrow data.

This PR also doesn't implement direct encoding for Map/Slice type attributes, which are CBOR encoded. This can be done in a followup PR. For now, we decode the CBOR to the prost struct and encode that.

Some quick performance testing was performed on a pipeline like following:
- loadgen = 500,000 logs/sec
- batch size 8000
- single CPU core per pipeline
- machine = M4 macbook pro (my laptop)
```
                                                                
      ┌──────────────────────────────────┐                      
      │ load gen      core 1             │                      
      │                                  │                      
      │ ┌────────────┐     ┌───────────┐ │                      
      │ │ fake data  ┼────►│OTLP       ┼─┼───────┐              
      │ │ generator  │     │exporter   │ │       │              
      │ └────────────┘     └───────────┘ │       │              
      └──────────────────────────────────┘       │              
                                                 │              
                                                 │              
  ┌──────────────────────────────────────────────┘              
  │                                                             
  │  ┌──────────────────────────────────────────────┐           
  │  │    system under test          core 2         │           
  │  │                                              │           
  │  │┌────────┐    ┌────────────┐     ┌───────────┐│           
  │  ││OTLP    │    │  attribute │     │  OTLP     ││           
  └─►►│Receiver┼───►│  processor ┼─────►  Exporter ┼┼─────► ... 
     │└────────┘    └────────────┘     └───────────┘│           
     │                                              │           
     └──────────────────────────────────────────────┘           
                                                                
```
Results - CPU average utilization:
- `main` 65.1%
- `albert/1095`44.3%
- ~20% decrease

### DIFF
--- a/rust/otap-dataflow/crates/config/Cargo.toml
+++ b/rust/otap-dataflow/crates/config/Cargo.toml
@@ -20,4 +20,4 @@ serde_yaml = { workspace = true }
 miette = {workspace = true}
 urn = "0.7"
 
-schemars = { version="0.9.0" }
+schemars = { version="1.0.0" }
--- a/rust/otel-arrow-rust/Cargo.toml
+++ b/rust/otel-arrow-rust/Cargo.toml
@@ -40,6 +40,7 @@ tonic-prost = "0.14"
 tokio = { version = "1.43.0", features = ["rt", "time", "net", "io-util", "sync", "macros", "rt-multi-thread", "process"] }
 itertools = "0.14.0"
 smallvec = { version = "1.15.1", features = ["union"] }
+roaring = "0.11.2"
 
 [[bench]]
 name = "attribute_transform"
--- a/rust/otel-arrow-rust/src/otap/groups.rs
+++ b/rust/otel-arrow-rust/src/otap/groups.rs
@@ -6,22 +6,21 @@ use std::{
     collections::{BTreeMap, HashMap, HashSet},
     iter::{once, repeat, repeat_n},
     num::NonZeroU64,
-    ops::{Add, Range, RangeFrom, RangeInclusive},
+    ops::{Add, ControlFlow, Range, RangeFrom, RangeInclusive},
     sync::Arc,
 };
 
 use ahash::AHashSet;
 use arrow::{
     array::{
-        Array, ArrayRef, ArrowPrimitiveType, BinaryArray, DictionaryArray, FixedSizeBinaryArray,
-        GenericByteArray, PrimitiveArray, RecordBatch, StringArray, StructArray, UInt16Array,
-        UInt32Array,
+        Array, ArrayRef, ArrowPrimitiveType, DictionaryArray, PrimitiveArray, RecordBatch,
+        StructArray, UInt16Array, UInt32Array,
     },
     buffer::NullBuffer,
     compute::cast,
     datatypes::{
-        ArrowNativeTypeOp, ByteArrayType, DataType, Field, Fields, Float32Type, Float64Type,
-        Int32Type, Int64Type, Schema, SchemaBuilder, UInt8Type, UInt16Type, UInt32Type, UInt64Type,
+        ArrowNativeType, ArrowNativeTypeOp, DataType, Field, Fields, Float64Type,
+        GenericBinaryType, Int64Type, Schema, SchemaBuilder, UInt8Type, UInt16Type, UInt64Type,
     },
 };
 use itertools::Itertools;
@@ -1352,7 +1351,7 @@ impl UnifiedDictionaryTypeSelector {
     }
 
     fn visit_dict_array(&mut self, maybe_dict_column_arr: &ArrayRef) -> Result<()> {
-        self.total_batch_size += maybe_dict_column_arr.len();
+        self.total_batch_size += maybe_dict_column_arr.len() - maybe_dict_column_arr.null_count();
 
         // access the array containing the dictionary values ...
         let values_arr = match maybe_dict_column_arr.data_type() {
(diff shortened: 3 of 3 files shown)


rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
```diff
@@ -148,14 +155,26 @@ impl Exporter<OtapPdata> for OTLPExporter {
...
-                    _ = match service_req {
-                        OtlpProtoBytes::ExportLogsRequest(bytes) => {
+                    match (signal_type, payload) {
+                        // use optimized direct encoding OTAP -> OTLP directly
+                        // TODO currently this is only implemented for logs, but eventually we'll do
+                        // do this for other signal types as well
+                        (SignalType::Logs, OtapPayload::OtapArrowRecords(mut otap_batch)) => {
+                            proto_buffer.clear();
+                            logs_encoder
+                                .encode(&mut otap_batch, &mut proto_buffer)
```
utpilla, line 167: It looks like you have already implemented `TryFrom<OtapArrowRecords> for OtlpProtoBytes`. It would probably be better to simply call `try_into` here instead of explicitly encoding the data here. That way we would have the encoding logic in one place which would be the `try_into` implementation.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -429,11 +434,17 @@ impl TryFrom<OtapArrowRecords> for OtlpProtoBytes {
...
+                // and the buffer, a these structures can be used between requests
+                let mut logs_encoder = LogsProtoBytesEncoder::new();
+                let mut buffer = ProtoBuffer::new();
+
+                // Fast path direct OTAP -> OTLP for Logs. Eventually we'll implement this for
+                // other signal types
+                logs_encoder
+                    .encode(&mut value, &mut buffer)
+                    .map_err(map_otlp_conversion_error)?;
+                let bytes = buffer.as_ref().to_vec();
```
utpilla, line 447: I think we could provide the `vec` used by `ProtoBuffer` struct instead of creating a new one.

```rust
impl ProtoBuffer {
    pub fn into_vec(self) -> Vec<u8> {
        self.buffer
    }
}
```

We can then use it here:

```diff
logs_encoder
    .encode(&mut value, &mut buffer)
    .map_err(map_otlp_conversion_error)?;
- let bytes = buffer.as_ref().to_vec();
+ let bytes = buffer.into_vec();
Ok(Self::ExportLogsRequest(bytes))
```

rust/otel-arrow-rust/src/otlp/common.rs
```diff
@@ -171,3 +268,542 @@ impl fmt::Display for AnyValue {
...
+
+/// mutable buffer for encoding protobuf bytes
+#[derive(Debug, Default)]
+pub struct ProtoBuffer {
+    buffer: Vec<u8>,
+}
+
+impl ProtoBuffer {
+    #[must_use]
+    pub fn new() -> Self {
```
utpilla: `len()` and `is_empty()` can also be made `const`.

```suggestion
    pub const fn new() -> Self {
```

rust/otel-arrow-rust/src/otlp/attributes.rs
```diff
@@ -1,7 +1,163 @@
...
+            if let Some(attr_double) = &attr_arrays.attr_double {
+                if let Some(val) = attr_double.value_at(index) {
+                    result_buf.encode_field_tag(ANY_VALUE_DOUBLE_VALUE, wire_types::FIXED64);
+                    result_buf.extend_from_slice(&val.to_le_bytes());
+                }
+            }
+        }
+        AttributeValueType::Bytes => {
+            if let Some(attr_bytes) = &attr_arrays.attr_bytes {
+                if let Some(val) = attr_bytes.value_at(index) {
```
utpilla: nit: We don't need to own the bytes. We could use a reference to make the intent clear.

```suggestion
                if let Some(val) = &attr_bytes.value_at(index) {
```

rust/otel-arrow-rust/src/otlp/common.rs
```diff
@@ -171,3 +268,542 @@ impl fmt::Display for AnyValue {
...
+///
+/// The workaround is that we set aside a fixed length number of bytes, and create a zero-padded
+/// varint. For example, the varint 5, encoded in 8 bytes would be. Observe that in all bytes, the
+/// continuation bit (msb) is set:
+/// ```text
+/// 0x85 0x80 0x80 0x80 0x80 0x80 0x80 0x00
+/// ```
+///
+/// Note: this is less efficient from a space perspective, so there's a tradeoff being made here
+/// between encoded size and CPU needed to compute the size ofr the length.
```
utpilla:
```suggestion
/// between encoded size and CPU needed to compute the size of the length.
```

rust/otel-arrow-rust/src/otlp/common.rs
```diff
@@ -171,3 +268,542 @@ impl fmt::Display for AnyValue {
...
+///
+/// TODO: currently we're always allocating 8 byte. This is clearly too much, but we over-allocate
```
utpilla: I think we could at least reduce it to `4` bytes. With `4` bytes, we should be able to represent data upto `~268 MB` in size. That should be good enough.

rust/otel-arrow-rust/src/otlp/common.rs
```diff
@@ -171,3 +268,542 @@ impl fmt::Display for AnyValue {
...
+        let len = $buf.len() - len_start_pos - num_bytes;
+        $crate::otlp::common::patch_len_placeholder($buf, num_bytes, len, len_start_pos);
+    }};
+}
+
+pub(crate) fn encode_len_placeholder(buf: &mut ProtoBuffer, num_bytes: usize) {
+    for _ in 0..num_bytes - 1 {
+        buf.buffer.push(0x80)
+    }
+    buf.buffer.push(0x00);
```
utpilla: I think we could use a static array instead of using a `for` loop. The compiler might be smart enough to optimize it anyway though.

```rust
   buf.buffer.extend_from_slice(&[0x80, 0x80, 0x80, 0x00]); // This is assuming we use 4 bytes for encoding length
```

rust/otel-arrow-rust/src/otlp/common.rs
```diff
@@ -171,3 +268,542 @@ impl fmt::Display for AnyValue {
...
+    }
+
+    /// This helper function exists so we can the heap allocation of the `rows` vec associated with
+    /// this [`BatchSorter`]. We effectively just shuffle the vec between having the lifetime of
+    /// the borrowed rows while we're sorting, and the static lifetime when we're done
+    fn reuse_rows_vec<'a, 'b>(mut v: Vec<(usize, Row<'a>)>) -> Vec<(usize, Row<'b>)> {
+        v.clear();
+        // there's a compiler optimization that happens here where it will just recognize that the
+        // vec we're creating has the same type/alignment as the source, and reuse the allocation
+        v.into_iter().map(|_| unreachable!()).collect()
```
utpilla, line 556: Wow! This is such a cool optimization.

It'd be good to write a test to ensure that:
- The input and output vec have the same capacity
- Or the input and output vec point to the same location (`as_ptr()`)

rust/otel-arrow-rust/src/otlp/common.rs
```diff
@@ -171,3 +268,542 @@ impl fmt::Display for AnyValue {
...
+
+    /// Reset the cursor. Typically this would be called before the cursor is then reinitialized.
+    pub fn reset(&mut self) {
+        self.sorted_indices.clear();
+        self.curr_index = 0;
+    }
+
+    /// Get the current index to visit
+    pub fn curr_index(&self) -> usize {
+        self.sorted_indices[self.curr_index]
```
utpilla: This could panic if the index is out of bounds. Is it something to worry about? We could look at updating this method signature to return `Option<usize>` instead so that it can return `None` for out of bounds index. Or maybe we could ensure that `advance()` doesn't increment the `curr_index` to a value higher than `sorted_indices.len()`.

rust/otel-arrow-rust/src/otlp/common.rs
albertlockett: Yeah that seems reasonable, thanks @utpilla! Made this change

rust/otel-arrow-rust/src/otlp/common.rs
albertlockett: Made this change, thanks @utpilla 

I was curious about the compiler optimization too! After checking w/ `cargo asm` it turns out the compiler was only smart enough to unroll 8/4 loop iterations & push the bytes one at a time.

If we use the static array as you suggested, it's smart enough to just push a 32 bit literal, which I guess is another good reason to use 4 bytes for the fixed size len here
```asm
mov w10, #32896          // low 16 bits = 0x8080
movk w10, #128, lsl #16  // high 16 bits = 0x0080
// final: w10 = 0x00808080
str w10, [x9, x8] // x9 = vec ptr, x8 = vec offset
```

rust/otel-arrow-rust/src/otlp/common.rs
albertlockett: Thanks! I found out about it from this video from slack earlier this week day https://cloud-native.slack.com/archives/C08RRSJR7FD/p1757143706388599

Test has been added in 06742129

rust/otel-arrow-rust/src/otlp/common.rs
albertlockett: Made this change in 9922e1c4

rust/otel-arrow-rust/src/otlp/attributes.rs
albertlockett: Thanks @utpilla. Actually as I was looking at this, I realized it probably didn't make sense for us to copy the bytes into an owned `Vec` when calling `value_at` here. I added a new method called `slice_at` to this type, that can be used to get a shared ref to the byte slice. Hopefully that's an OK solution

dbbcd70a

rust/otel-arrow-rust/src/otlp/common.rs
albertlockett: Yeah that's a good point. I guess some bad logic around how `advance()` gets called could cause a panic, so it would be good to guard against that

I think returning `Option` makes the most sense. The tricky thing about limiting how far we advance is we wouldn't know if the cursor is at the last element, or past it.

Made this change in ccdb326a

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
albertlockett: The reason this doesn't call `try_into` is because I wanted to reuse the `LogsProtoBytesEncoder` and `ProtoBuffer` for different batches. Internally, these use a handful of `Vec` and I didn't want to throw away those allocations for every batch.

Do you think it'd make sense to implement something like this?
```rs
impl
    TryFrom<(
        OtapArrowRecords,
        &mut LogsProtoBytesEncoder,
        &mut ProtoBuffer,
    )> for OtlpProtoBytes
{
  // ...
```

rust/otap-dataflow/crates/otap/src/pdata.rs
albertlockett: I can make the change to allow converting between `ProtoBuffer` and `vec` as you suggested, but where would we provide the `vec` from in this case? 

Similar to what I asked [here](https://github.com/open-telemetry/otel-arrow/pull/1109/files#r2344286175), are you thinking that we should implement `TryFrom` in a way that allows us to pass this vec? 

```rs
impl TryFrom<(OtapArrowRecords, Vec<u8>)> for OtlpProtoBytes {
    fn try_from(value: (OtapArrowRecords, Vec<u8>)) -> Result<Self> {
        let mut batch = value.0;
        let mut buffer = ProtoBuffer::from_vec(value.1);
        
        let mut logs_encoder = LogsProtoBytesEncoder::new(); // (aside: as I mentioned in the other comment, it'd be nice to reuse this somehow)
        logs_encoder
            .encode(&mut batch, &mut buffer)
            .map_err(map_otlp_conversion_error)?;

        let bytes = buffer.into_vec(); 
        Ok(Self::ExportLogsRequest(bytes))
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/arrow_records_encoder.rs
```diff
@@ -178,6 +182,15 @@ mod tests {
         );
     }
 
+    fn otlp_logs_to_otap(mut logs_otap_batch: OtapArrowRecords) -> ExportLogsServiceRequest {
```
jmacd, line 185: should this be otap_logs_to_otlp?

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
```diff
@@ -148,14 +155,26 @@ impl Exporter<OtapPdata> for OTLPExporter {
...
+                                .map_err(|e| {
+                                    self.pdata_metrics.logs_failed.inc();
+                                    Error::ExporterError {
+                                        exporter: exporter_id.clone(),
+                                        error: e.to_string(),
+                                    }
+                                })?;
+                            // TODO we should try to change the client interfaces to accept a slice
```
jmacd, line 175: When I came to understand this code, I wondered why Tonic would need to take ownership of the Vec and not a reference. For the retry_sender case, we need to hold a copy while the request exports, and it seems we end up with a clone, so yes.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -429,11 +434,17 @@ impl TryFrom<OtapArrowRecords> for OtlpProtoBytes {
 
         match value {
             OtapArrowRecords::Logs(_) => {
-                let export_logs_svc_req = logs_from(value).map_err(map_otlp_conversion_error)?;
-                let mut bytes = vec![];
-                export_logs_svc_req
-                    .encode(&mut bytes)
-                    .map_err(map_prost_encode_error)?;
+                // TODO it'd be nice to expose a better API where we can make it easier to pass the encoder
+                // and the buffer, a these structures can be used between requests
```
jmacd, line 438: Sounds good.

rust/otel-arrow-rust/src/otlp/common.rs
```diff
@@ -171,3 +268,602 @@ impl fmt::Display for AnyValue {
...
+/// ```text
+/// 0x85 0x80 0x80 0x00
+/// ```
+///
+/// Note: this is less efficient from a space perspective, so there's a tradeoff being made here
+/// between encoded size and CPU needed to compute the size of the length.
+///
+/// TODO: currently we're always allocating 4 byte. This may often be too much but we over-allocate
+/// to be safe. Eventually we should maybe allow a size hint here and allocate fewer bytes.
+///
```
jmacd, line 442: My sense is that CPU efficiencies combined with compression will make up for the extra space.

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
utpilla: Makes sense.

rust/otap-dataflow/crates/otap/src/pdata.rs
utpilla: > I can make the change to allow converting between ProtoBuffer and vec as you suggested, but where would we provide the vec from in this case?

In the current `TryFrom` implementation, we are allocating a `Vec<u8>` twice for each conversion:
``` rust
let mut buffer = ProtoBuffer::new(); // First allocation
...
let bytes = buffer.as_ref().to_vec(); // Second allocation
```

My suggestion of using a `into_vec` method was mainly to avoid the second allocation.


> Similar to what I asked [here](https://github.com/open-telemetry/otel-arrow/pull/1109/files#r2344286175), are you thinking that we should implement TryFrom in a way that allows us to pass this vec?

Yeah, we could explore something like that. It would achieve two things:
- Allow reuse of `LogsProtoBytesEncoder` and `Protobuffer`
- Keep the encoding logic limited to a central place

albertlockett: > My suggestion of using a into_vec method was mainly to avoid the second allocation.

ohhh, I'm sorry @utpilla , I misunderstood! My suggestion of using a into_vec method was mainly to avoid the second allocation. Yes it would have been trivial to avoid the second allocation. I'll clean this up in a future PR

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/arrow_records_encoder.rs
albertlockett: yeah you're right. I'll make this change in a future PR

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
albertlockett: Digging into this, looks like Tonic does expect to take ownership of whatever it is passed:
https://github.com/hyperium/tonic/issues/783

That said, I think if we made the internal representation of pdata an `Arc<Vec<u8>>` and I think that'll work with tonic as well.

PR #1111, opened 2025-09-11
[otel-arrow-rust] Simplify OtapBatchStore

utpilla:
## Changes
- Simplify `OtapBatchStore` to use `Self::COUNT` in the return type for `into_batches` method
- Remove `shrink` method as it's no longer required

### DIFF
--- a/rust/experimental/query_engine/engine-recordset/src/transform/mod.rs
+++ b/rust/experimental/query_engine/engine-recordset/src/transform/mod.rs
@@ -2,4 +2,5 @@
 // SPDX-License-Identifier: Apache-2.0
 
 pub(crate) mod reduce_map_transform_expression;
+pub(crate) mod rename_map_keys_transform_expression;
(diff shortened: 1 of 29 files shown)


rust/otel-arrow-rust/src/otap.rs
```diff
@@ -133,14 +133,17 @@ impl OtapArrowRecords {
...
+pub trait OtapBatchStore: Default + Clone {
     /// Internally, implementers should use a bitmask for the types they support.
     /// The offsets in the bitmask should correspond to the ArrowPayloadType enum values.
     const TYPE_MASK: u64;
 
     /// The number of `RecordBatch`es needed for this kind of telemetry.
     const COUNT: usize;
 
+    /// The type of array used to store the record batches.
+    type BatchArray;
```
albertlockett, line 145: Cool! Good addition @utpilla

Issue #891, opened 2025-08-07
KQL attribute renaming component


jmacd:
We have been developing a KQL implementation with the intention to integrate with the OTAP engine in the future.

We have an effort to optimize the renaming of attributes in an OTAP frame, https://github.com/open-telemetry/otel-arrow/issues/813.

This issue is to demonstrate an application of KQL for simple renaming which will execute efficiently over OTAP.

Consider this a stretch goal for the milestone!

lquerel:
@jmacd do you have an ETA for this one? I'm trying to figure out if we will have a baby version of this processor in more or less one week. If not it's not a big deal :-)

drewrelmas:
I have a mostly-working version of this on a local branch (mapping `extend` -> `rename` and `project-away` -> `delete`). This isn't exactly correct, realized kql-parser is missing `project-rename` functionality, tracking that in #1102. That is the grammar we should be using equivalent to `rename`.

I plan to bring a PR by the next SIG meeting on Sept 18 at the latest.

PR #1112, opened 2025-09-11
[DRAFT] Add KqlAttributesProcessor variant of AttributesProcessor

drewrelmas:
Addresses #891

This PR is the first attempt at combining the work happening in `rust/experimental/query-engine` with `otap-dataflow`. This is not a final state, but wanted to publish this draft early to start conversation.

Remaining:
- [x] Add KQL mapping for `project-rename` expression to `rename` action once #1103 merges
- [x] Abstract shared `attributes_processor` code (`process()` implementation, `Action` enum declaration, telemetry setup, test harness) to a base `common.rs` used by both `kql_attributes_processor` and `attributes_processor` (eliminate code duplication)
- [ ] Consider implications of the added dependency in `otap-dataflow` on `rust/experimental/query-engine` crates

We should decide if `kql_attributes_processor` should be part of the 'core' `otap-dataflow` or should fall into something like a `Contrib` package.

The `query-engine` crates have a lot more functionality/grammar support than the current operations supported by `attributes_processor`. Should a minimal subset be copied into `otap-dataflow` while development continues in `experimental`?

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/attributes_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/attributes_processor.rs
@@ -35,7 +35,6 @@ use crate::{OTAP_PROCESSOR_FACTORIES, pdata::OtapPdata};
 use async_trait::async_trait;
 use linkme::distributed_slice;
 use otap_df_config::error::Error as ConfigError;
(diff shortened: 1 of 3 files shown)


jmacd:
Feel free to re-open.

PR #1114, opened 2025-09-12
[DRAFT] Back-propagation part 1 of 3: PData with_reply_to, deadline, take_reply logic, AckMsg/NackMsg

jmacd:
This replaces my earlier draft #1041 with a smaller subset following #1078. Replaces #1080.

The OtapPdata Context gains a Vec<Frame> for context, each Frame contains Interest, NodeID, timestamp and optional deadline managed by the module, and ContextData containing a placeholder `usize` register, currently used only by the `retry_processor` for its `retries` count.

Features of this implementation:

A push/pop for the per-node state and relevant details at each level of the Context stack. On the forward path, use push_reply_to prior to send_message:

```
    /// Reply with certain interest to this node_id (the usize index)
    /// to and node-specific context data.
    pub fn push_reply_to(
        &mut self,
        interest: Interest,
        node_id: usize,
        data: ContextData,
    ) -> Result<(), error::Error>
```

on the return path, via AckMsg or NackMsg, use take_reply:

```
    /// Reply state returns a context
    pub fn take_reply(&mut self) -> Option<Frame> {
        self.context.stack.pop()
    }
```

Deadline is automatically enforced in push_reply_to, hence the Result. When the attempt to push_reply_to fails because of deadline in retry_processor, it Nacks to the caller. 

Important helper methods used by retry_processor:

```
    /// The deadline on this request.
    pub fn deadline(&self) -> Option<Instant> {
        self.context.deadline()
    }

    /// Return the origin node_id for routing Ack/Nack.
    pub fn return_node_id(&self) -> Option<usize> {
        self.context.return_node_id()
    }
```

The deadline() function is used in retry_processor to determine if the request will timeout before the delay, to fail fast. The return_node_id() function is used to route Ack/Nack since the engine crate doesn't know about the actual data type.

Leaves me with an option question about how failures in send_message(), send_delayed(), reply() at the effect handler, which will fail when a channel is full, should be handled. In order to retain an OtapPdata that would otherwise drop, so that it can be returned to a retry_sender, the Error returned from `process()` will need to change to `TypedError<PData>` and the engine will have to have a way to recover typed fields from full channel errors.

Leaves several TODOs for future PRs

1. New EffectHandler::reply_ack() and reply_nack() calls the same methods in core, but are placeholders. These will instead deliver pipeline control messages for routing the Ack/Nack messages.
2. New EffectHandler::send_delayed() for delaying a message before it is resent. One open question is the way this delay currently stands. The delay-setter knows its own node_id but not the node_id of the destination it is sending to. Therefore, when it sets a delay for a message, the timer will come back to it, where it can then try the send again. We could imagine a way for send_delayed to indicate the destination channel or node_id, but this seems to require a larger architectural change.


rust/otap-dataflow/crates/engine/src/control.rs
```diff
@@ -14,6 +14,71 @@ use otap_df_telemetry::reporter::MetricsReporter;
...
+pub struct AckMsg<PData> {
+    /// Reply-to information.
+    /// Placeholder: will become Box<PData> containing context and empty payload.
+    pub msgid: u64,
+
+    /// Placeholder: we will store Box<PData> and remove this field
+    _data: std::marker::PhantomData<PData>,
+
+    /// Rejected items, part of OTLP protocol responses.
+    pub rejected: Option<RejectedItems>,
```
lquerel, line 29: I believe you already told me about this before, but I no longer understand why we can have rejected items in an AckMsg.

rust/otap-dataflow/crates/engine/src/effect_handler.rs
```diff
@@ -200,6 +200,44 @@ impl EffectHandlerCore {
             pipeline_ctrl_msg_sender,
         })
     }
+
+    /// Reply in failure
+    pub async fn reply_nack<PData>(
```
lquerel: To minimize the possibilities of misuse of the effect handler, I think we could just pass the NackMsg to this method. The `node_id` could be part of the effect handler's context (if it isn't already).

rust/otap-dataflow/crates/engine/src/effect_handler.rs
```diff
@@ -200,6 +200,44 @@ impl EffectHandlerCore {
...
+    ) -> Result<(), Error> {
+        // TODO Future PR: Need to add <PData> to this struct and PipelineCtrlMsgSender first.
+        // Note node_id is Option<_> to avoid repetitive error checking.
+        Ok(())
+    }
+
+    /// Reply in success
+    pub async fn reply_ack<PData>(
+        &self,
+        _node_id: Option<usize>,
```
lquerel, line 218: Same comment

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -86,13 +86,88 @@ use otel_arrow_rust::otlp::ProtoBuffer;
...
+impl Context {
+    /// The deadline is the optional point at which this request has
+    /// exceeded its lifetime.
+    pub fn deadline(&self) -> Option<Instant> {
+        let last = self.stack.last();
+        last.map(|f| f.deadline()).flatten()
+    }
+
+    /// Return NodeID is the upstream component ID.
+    pub fn return_node_id(&self) -> Option<usize> {
```
lquerel, line 169: I don't understand this method. I was expecting to see an internal method returning the highest `node_id` in the stack for a specific interest. Having this method exposed will be, in my opinion, source of errors. The routing could stay inside the engine.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -86,13 +86,88 @@ use otel_arrow_rust::otlp::ProtoBuffer;
...
+/// Interest allows pipeline controller to route by Ack or Nack.
+#[derive(Clone, Debug)]
+pub enum Interest {
+    /// Only cares for Ack responses (???).
+    AckOnly,
+    /// Only cares for Nack responses (for retry).
+    NackOnly,
+    /// Cares about any response.
+    AckOrNack,
+}
```
lquerel, line 124: If we have more than two options (Ack, Nack), it'll become complicated to represent all the combinations. I think we should use an approach (bit-set) that can easily support every combination. A crate like [bitflags](https://crates.io/crates/bitflags) could be used here, starting with a `u8` backend for now.

```rust
bitflags::bitflags! {
    #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
    pub struct Interests: u8 {  // or bigger if/when needed
        const ACKS   = 1 << 0;
        const NACKS  = 1 << 1;
        // ...
    }
}
```

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -86,13 +86,88 @@ use otel_arrow_rust::otlp::ProtoBuffer;
...
+
+/// Frame of interest.
+#[derive(Clone, Debug)]
+pub struct Frame {
+    /// Which responses matter.  TODO: Used in a future PR when
+    /// pipeline_ctrl.rs actually routes Ack/Nacks.
+    _interest: Interest,
```
lquerel, line 131: I suggest to use the plural form for this field as this field will represent all the interests of a specific `node_id`.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -86,13 +86,88 @@ use otel_arrow_rust::otlp::ProtoBuffer;
...
+
+    /// Which node created this frame.
+    node_id: usize,
+
+    /// When the frame was initialized.
+    when: Instant,
```
lquerel, line 137:
```suggestion
    /// When the frame was initialized.
    /// TODO: check perf impact of having a std Instant versus something like https://crates.io/crates/quanta
    when: Instant,
```

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -216,6 +299,48 @@ impl OtapPdata {
         (self.context, self.payload)
     }
 
+    /// Reply state returns a context
+    pub fn take_reply(&mut self) -> Option<Frame> {
```
lquerel, line 305: In what context do we need this method? Don't we need to know the interest in order to know to whom to send a message?

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -216,6 +299,48 @@ impl OtapPdata {
...
+        self.context.stack.pop()
+    }
+
+    /// Reply with certain interest to this node_id (the usize index)
+    /// to and node-specific context data.
+    pub fn push_reply_to(
```
lquerel, line 311: Why not `subscribe_to(...)` ?

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -86,13 +86,88 @@ use otel_arrow_rust::otlp::ProtoBuffer;
 use otel_arrow_rust::otlp::logs::LogsProtoBytesEncoder;
 use otel_arrow_rust::otlp::{metrics::metrics_from, traces::traces_from};
 use prost::{EncodeError, Message};
+use std::time::{Duration, Instant};
 
 use crate::encoder::{encode_logs_otap_batch, encode_spans_otap_batch};
 
+/// Context data per node
```
lquerel, line 95: I'm not entirely clear on the definition of this struct. What is a register in this context?

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -6,11 +6,50 @@
...
+//! ## RetryState Handler Responsibility
+//!
+//! The retry processor uses a **stateless design** where retry information is stored in the message
+//! context stack rather than in the processor itself. This enables horizontal scaling and fault tolerance.
+//!
+//! ### Forward Path (Data Processing)
+//! 1. When a PData message arrives, the processor checks if it has a return node ID
+//! 2. If it does, the processor **pushes** a `RetryState` onto the context stack before forwarding
+//! 3. The `RetryState` contains the current retry count (initially 0)
+//! 4. The message is then sent downstream with this state attached
```
lquerel: We could point out that this mechanism is generic and will be used not only by the `RetryProcessor` but also potentially by other processors, and more specifically by receivers that want to return Ack/Nack messages to signal producers.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -6,11 +6,50 @@
...
+//! ```text
+//! Request Flow:
+//! [Upstream] --PData--> [RetryProcessor: PUSH RetryState] --PData+State--> [Downstream]
+//!
+//! Success Flow:
+//! [Upstream] <--ACK-- [RetryProcessor: PEEK+POP State] <--ACK+State-- [Downstream]
+//!
+//! Retry Flow:
+//! [Upstream] <--Delayed-- [RetryProcessor: PEEK State, increment retries] <--NACK+State-- [Downstream]
+//!                         [RetryProcessor: PUSH updated State] --PData+State--> [Downstream]
```
lquerel: I'm not yet completely clear on how this delay will be defined. The delay itself should be specified by the retry policy, since it is the one that has the retry policy configuration (e.g. exponential backoff).

rust/otap-dataflow/crates/otap/src/pdata.rs
lquerel: The `Ack/NAck` messages rely on the fields `when` and `dead`. Will these fields be important for every context in other types of control? Should we have these fields wrapped in `Option<_>` (like for `dead`), or even pushed into `ContextData`?

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -86,13 +86,88 @@ use otel_arrow_rust::otlp::ProtoBuffer;
...
+    node_id: usize,
+
+    /// When the frame was initialized.
+    when: Instant,
+
+    /// Remaining time, relative to when.
+    dead: Option<Duration>,
+
+    /// Data for this frame.
+    pub data: ContextData,
```
lquerel, line 143: I'm thinking we could use a `SmallVec<[RawVal; 2]>` or `SmallVec<[RawVal; 3]>` to encode the `ContextData` because (1) it's efficient, (2) it's always smaller than an empty HashMap, (3) it spills to the heap if the user node needs more than 3 slots. We could, for example, put `when` (position 0) and `dead` (position 1) in the context data (if we consider these 2 fields optional). With the `bytemuck` crate (widely used and probably already in Arrow's dependencies), we can convert `RawVal` to `u64`, `i64`, `f64` at zero cost. That would give something like:

```rust
use bytemuck::{Pod, Zeroable};

#[repr(transparent)]
#[derive(Clone, Copy, Debug, Pod, Zeroable)]
pub struct RawVal([u8; 8]);

impl RawVal {
    pub fn from<T: Pod>(v: T) -> Self {
        Self(bytemuck::cast(v))
    }
    pub fn into<T: Pod>(self) -> T {
        bytemuck::cast(self.0)
    }
}

#[derive(Clone, Debug)]
pub struct Frame {
  interests: Interests,
  node_id: usize,
  data: SmallVec<[RawVal, 3]>,
}
```

I think this works well for all fields that only make sense for the node that made the subscription.

rust/otap-dataflow/crates/engine/src/local/processor.rs
```diff
@@ -205,6 +214,30 @@ impl<PData> EffectHandler<PData> {
         self.core.start_periodic_timer(duration).await
     }
 
+    /// Reply in failure
+    pub async fn reply_nack(
+        &self,
+        node_id: Option<usize>,
```
lquerel, line 220: I probably don't understand something here. Why `node_id` here and why is it an option?

rust/otap-dataflow/crates/engine/src/local/processor.rs
```diff
@@ -205,6 +214,30 @@ impl<PData> EffectHandler<PData> {
...
+        nack: NackMsg<PData>,
+    ) -> Result<(), Error> {
+        self.core.reply_nack(node_id, nack).await
+    }
+
+    /// Reply in success
+    pub async fn reply_ack(&self, node_id: Option<usize>, ack: AckMsg<PData>) -> Result<(), Error> {
```
lquerel, line 227: Same kind of interrogation here.

rust/otap-dataflow/crates/engine/src/local/processor.rs
```diff
@@ -205,6 +214,30 @@ impl<PData> EffectHandler<PData> {
...
+        self.core.reply_ack(node_id, ack).await
+    }
+
+    /// Delay a message
+    pub async fn send_delayed(
+        &self,
+        sender_id: usize,
```
lquerel, line 234: Same here

rust/otap-dataflow/crates/otap/src/pdata.rs
lquerel: I proposed an alternative, see few comments below.

rust/otap-dataflow/crates/otap/src/pdata.rs
lquerel: If you like `subscribe_to` then we should replace `reply_ack` and `reply_nack` with `notify_ack` and `notify_nack`.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -290,52 +228,123 @@ impl Processor<OtapPdata> for RetryProcessor {
...
-                    let _previous = self.pending_messages.insert(id, pending);
-                    log::debug!("Added message {id} to retry queue");
-
-                    effect_handler.send_message(data).await?;
+            Message::PData(mut data) => {
+                match data.push_reply_to(
+                    Interest::NackOnly,
+                    effect_handler.processor_id().index(),
+                    ContextData::new(0),
+                ) {
```
lquerel: To better hide the details and avoid errors, I suggest reversing the way things are done here.

```rust
effect_handler.subscribe_to(Interests::Nack, &mut data)
```

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -290,52 +228,123 @@ impl Processor<OtapPdata> for RetryProcessor {
...
+                        effect_handler.send_message(data).await?;
+                        Ok(())
+                    }
+                    Err(err) => {
+                        // Reply_to not pushed, e.g., deadline already exceeded.
+                        effect_handler
+                            .reply_nack(
+                                data.return_node_id(),
+                                NackMsg::new(data, err.to_string(), false, None),
+                            )
```
lquerel: In the same spirit:

```rust
effect_handler.notify_nack(NackMsg::new(data, err.to_string(), false, None))
```

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -290,52 +228,123 @@ impl Processor<OtapPdata> for RetryProcessor {
...
+                    // If the error is permanent or too many retries.
+                    // If the payload is empty: the effect is also
+                    // permanent, as by intentionaly failing to return
+                    // the data.
+                    let has_failed = if nack.permanent {
+                        nack.reason = format!("cannot retry permanent: {}", nack.reason);
+                        true
+                    } else if rstate.retries >= self.config.max_retries {
+                        nack.reason = format!("max retries reached: {}", nack.reason);
+                        true
```
lquerel: Why not using a code here?

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -290,52 +228,123 @@ impl Processor<OtapPdata> for RetryProcessor {
...
+                    } else if nack.refused.is_empty() {
+                        nack.reason = format!("retry internal error: {}", nack.reason);
+                        nack.permanent = true;
+                        nack.code = Some(Code::Internal as i32);
+                        true
+                    } else {
+                        false
+                    };
+                    let return_node_id = nack.refused.return_node_id();
```
lquerel: IMO, knowing where to send the nack message next should not be the responsibility of the node.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -290,52 +228,123 @@ impl Processor<OtapPdata> for RetryProcessor {
...
+
+                    if has_failed {
+                        // Permanent failure: forward NACK upstream (originator will POP state)
+                        effect_handler.reply_nack(return_node_id, nack).await?;
```
lquerel: I will suggest something like:

```rust
effect_handler.notify_nack(nack).await?;
```

and the effect_handler will unstack the vec of frames to determine where to send the nack next.

rust/otap-dataflow/crates/otap/src/pdata.rs
lquerel: See one of my next comments, I suggest to only expose this method in the effect handler.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
lquerel: Ok got it, this method is call by the retry processor on its own effect handler. That makes sense now.

lquerel (changes requested):
Sorry for the delay. It took me a bit of time because it's inherently complicated, and since it's so central I wanted to really absorb the details before replying. I've made suggestions that I think stay fairly close to your proposal but that, in my view, either simplify the API or make it less error-prone. At a high level, I propose the following principles:
- Routing that is, determining where to send a control message, should only be handled at the engine level. Nodes shouldn't have to pop the stack and decide where to send control messages.
- I'm leaning toward a subscription/notification concept. It's more of a renaming than anything else, but it seems more understandable to me.
- Fields in the context that only make sense for a particular node should end up in the context's "opaque" part.

Much more details in the comments.

jmacd:
Closed in favor of #1181 after applying I believe all of @lquerel's suggestions.

PR #1137, opened 2025-09-15
[query-engine] Extend KQL parser schema into child maps

CodeBlanch:
## Changes

* Extend schema support in KQL parser into child maps

### DIFF
--- a/rust/experimental/query_engine/kql-parser/src/scalar_primitive_expressions.rs
+++ b/rust/experimental/query_engine/kql-parser/src/scalar_primitive_expressions.rs
@@ -548,28 +548,34 @@ pub(crate) fn parse_accessor_expression(
             }
         }
 
+        let mut resolved_value_type = None;
+
         if let Some(schema) = scope.get_source_schema() {
             match schema.get_schema_for_key(root_accessor_identity.get_value()) {
                 Some(key) => {
                     if value_accessor.has_selectors() {
-                        // Note: If we are selecting a well-defined key on the
-                        // source we can do some validation. If there are child
-                        // selectors they are only valid for maps and arrays.
-                        // This logic could be improved further to inspect the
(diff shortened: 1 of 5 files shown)


rust/experimental/query_engine/kql-parser/src/scalar_primitive_expressions.rs
```diff
@@ -548,28 +548,36 @@ pub(crate) fn parse_accessor_expression(
...
+                            ParserMapKeySchema::Map(inner_schema) => {
+                                if let Some(schema) = inner_schema {
+                                    resolved_value_type = schema.try_resolve_value_type(
+                                        value_accessor.get_selectors_mut(),
+                                        &scope.get_pipeline().get_resolution_scope(),
+                                    )?;
+                                }
+                            }
+                            ParserMapKeySchema::Array | ParserMapKeySchema::Any => {
+                                // todo: Implement validation for arrays
```
drewrelmas, line 567: Could you expand a little bit more on this todo? Does this mean checking that every item in the array is a specific type?

CodeBlanch: Not sure exactly what we should do here that's why I didn't do anything 🤣 I could see different scenarios\modes. Should be pretty easy to support validation of homogenous arrays where every element should match some known schema. Harder things: Values matching some known set of schemas (one_of) or expecting specific schema at specific indices (🤮)

I'm not planning to work on this anytime soon, I just put the comment there to indicate "lack of array support was intentional" (more or less). We could just wait and see what scenarios\asks pop up?

drewrelmas: Sound good to me, LGTM

Issue #1119, opened 2025-09-12
Direct Conversion OTAP -> OTLP proto bytes for Traces


albertlockett:
in #1109 we added a direct conversion between OTAP and OTLP proto bytes for logs to optimize CPU utilization in the OTLP exporter.

We should make the same improvement for Traces

PR #1138, opened 2025-09-15
Encode OTAP -> OTLP proto bytes directly for Traces

albertlockett:
closes: #1119 

Continues the work started in #1109 traces


rust/otel-arrow-rust/src/otlp/common.rs
```diff
@@ -369,6 +366,11 @@ impl ProtoBuffer {
         Self { buffer: Vec::new() }
     }
 
+    #[must_use]
+    pub fn to_vec(self) -> Vec<u8> {
```
utpilla: I think we should use the prefix `into` instead of `to` here since the input struct is being consumed.

https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

Also, we could probably use `into_bytes` as the method name instead of `into_vec`. Similar to the std library method [`String::into_bytes()`](https://doc.rust-lang.org/std/string/struct.String.html#method.into_bytes)

```suggestion
    pub fn into_bytes(self) -> Vec<u8> {
```

rust/otel-arrow-rust/src/otlp/common.rs
```diff
@@ -634,19 +651,55 @@ impl BatchSorter {
...
+                        .map(|i| ids.value_at(i).unwrap_or_default())
+                        .enumerate(),
+                );
+            }
+        }
 
         if ids.null_count() == 0 {
             // fast path, no null IDs
-            self.u16_ids.sort_unstable_by(|(_, a), (_, b)| a.cmp(b));
+            sort_ids_tmp.sort_unstable_by(|(_, a), (_, b)| a.cmp(b));
```
utpilla: It might be faster to use `sort_unstable_by_key` for these simple key-based comparisons.

```suggestion
            sort_ids_tmp.sort_unstable_by_key(|&(_, value)| value);
```

jmacd (approved):

Issue #1120, opened 2025-09-12
Direct conversion OTAP -> OTLP proto bytes for Metrics


albertlockett:
in https://github.com/open-telemetry/otel-arrow/pull/1109 we added a direct conversion between OTAP and OTLP proto bytes for logs to optimize CPU utilization in the OTLP exporter.

We should make the same improvement for Metrics

PR #1143, opened 2025-09-17
Encode OTAP -> OTLP proto bytes directly for Metrics

albertlockett:
closes: #1120 

Continues the work started in https://github.com/open-telemetry/otel-arrow/pull/1109 metrics

This also 
- removes the last of the code that computed OTLP structs directly from OTAP, including the `otel_arrow_rust::otlp::store::AttributeStore`.
- fixes a bug in materializing parent IDs where we didn't handle dict encoded parent IDs, which can be present in some attributes where parent_id is u32

### DIFF
--- a/rust/otel-arrow-rust/src/otap/transform.rs
+++ b/rust/otel-arrow-rust/src/otap/transform.rs
@@ -188,7 +188,6 @@ where
     let val_bytes_arr = record_batch.column_by_name(consts::ATTRIBUTE_BYTES);
 
     // downcast parent ID into an array of the primitive type
-    // let parent_id_arr = T::get_parent_id_column(record_batch)?;
     let parent_id_arr = MaybeDictArrayAccessor::<PrimitiveArray<T::ArrayType>>::try_new(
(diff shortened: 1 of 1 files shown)


rust/otel-arrow-rust/src/otlp/common.rs
```diff
@@ -384,6 +375,12 @@ impl ProtoBuffer {
         self.buffer.push(value as u8);
     }
 
+    /// encodes the signed varint type (e.g. sint32, sint64, etc.) using zig-zag encoding
+    /// https://protobuf.dev/programming-guides/encoding/#signed-ints
+    pub fn encode_sint32(&mut self, value: i32) {
```
lquerel:
```suggestion
    /// encodes the signed varint type (e.g. sint32, sint64, etc.) using zig-zag encoding
    /// https://protobuf.dev/programming-guides/encoding/#signed-ints
    #[inline]
    pub fn encode_sint32(&mut self, value: i32) {
```

Issue #1121, opened 2025-09-12
Direct conversion for OTAP Map/Slice attributes to proto bytes


albertlockett:
In #1109 we added the capability to directly convert between OTAP attributes and OTLP proto representation of AnyValue, however for Map & Slice type attributes which are using CBOR representation we still go through the intermediate prost structs:
https://github.com/open-telemetry/otel-arrow/blob/1b556dce0a2f981d8c1b13d5115763e78cce7d70/rust/otel-arrow-rust/src/otlp/attributes.rs#L133-L155

We should implement a direct translation of CBOR to OTLP proto representation for `AnyValue` message and use it for these attribute types

PR #1147, opened 2025-09-17
Encode OTAP -> OTLP proto bytes directly for Slice/Map AnyValues

albertlockett:
closes: #1121 

Stop going to the intermediate prost struct when decoding Map / Slice AnyValues

### DIFF
--- a/rust/otel-arrow-rust/src/otlp/attributes/cbor.rs
+++ b/rust/otel-arrow-rust/src/otlp/attributes/cbor.rs
@@ -34,18 +34,17 @@ fn proto_encode_cbor_value(value: &ciborium::Value, result_buf: &mut ProtoBuffer
             result_buf.encode_varint(*bool_val as u64);
         }
         ciborium::Value::Bytes(bytes_val) => {
-            result_buf.encode_bytes(ANY_VALUE_BYTES_VALUE, &bytes_val);
+            result_buf.encode_bytes(ANY_VALUE_BYTES_VALUE, bytes_val);
         }
         ciborium::Value::Float(float_val) => {
             result_buf.encode_field_tag(ANY_VALUE_DOUBLE_VALUE, wire_types::FIXED64);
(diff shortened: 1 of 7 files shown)


jmacd (approved):

rust/otel-arrow-rust/src/otlp/attributes/cbor.rs
```diff
@@ -1,96 +1,135 @@
...
-impl TryFrom<Vec<ciborium::Value>> for Value {
-    type Error = Error;
+fn proto_encode_array_values(
+    list_val: &[ciborium::Value],
+    result_buf: &mut ProtoBuffer,
+) -> Result<()> {
+    for v in list_val {
+        proto_encode_len_delimited_unknown_size!(
+            ARRAY_VALUE_VALUES,
+            proto_encode_cbor_value(v, result_buf)?,
```
utpilla, line 85: We should consider enforcing some depth limit for the recursion. Otherwise, it might be exploited by a bad actor. A deeply nested malicious payload may lead to StackOverflow/ DoS.

utpilla: Something to check for the OTLP proto bytes -> OTAP conversion as well:

https://github.com/open-telemetry/otel-arrow/blob/eaf81aabee7eaa80d4ffe78a59d9bfd0922345d0/rust/otap-dataflow/crates/otap/src/encoder/cbor.rs#L72-L89

lquerel: It's a general problem that we need to address at different levels, namely protecting the engine against crafted messages intended to cause a crash or any kind of malfunction. This is true for the case mentioned by Utkarsh, but it is also true for very long syslog | otlp | otap messages/batches.

lquerel: See https://github.com/open-telemetry/otel-arrow/issues/1150 for tracking.

PR #1154, opened 2025-09-18
[perf][otel-arrow-rust] Add fast path optimization for varint encoding

utpilla:
## Changes
- Introduce fast-path for one-byte (`0-127`) and two-byte values (`128-16383`) which should be the most common cases
- For the two-byte values, we now use `extend_from_slice` method which should be better than doing two vec push operations
- The method is now also marked `#[inline]`

## Performance
I observed an approximate **~35%** improvement (12.35 µs vs 19.08 µs) in varint encoding performance for two-byte values, based on benchmarks conducted on my machine. The test involved 10,000 iterations of the encoding operation. It would be helpful if others could run similar benchmarks on their systems to confirm whether they observe comparable results.

<details>
<summary>Benchmark code</summary>

```rust
//! Minimal benchmark for testing varint encoding performance for two-byte values.
use criterion::{criterion_group, criterion_main, Criterion};
use otel_arrow_rust::otlp::ProtoBuffer;
use std::hint::black_box;

/// Reference implementation (old version) for comparison
struct OldProtoBuffer {
    buffer: Vec<u8>,
}

impl OldProtoBuffer {
    fn new() -> Self {
        Self { buffer: Vec::new() }
    }
    
    fn clear(&mut self) {
        self.buffer.clear();
    }
    
    #[inline]
    fn encode_varint_old(&mut self, mut value: u64) {
        while value >= 0x80 {
            self.buffer.push(((value as u8) & 0x7F) | 0x80);
            value >>= 7;
        }
        self.buffer.push(value as u8);
    }
}

fn bench_varint_10000(c: &mut Criterion) {
    let value = 10000u64; // Two byte value for testing the fast path
    const ITERATIONS: usize = 10_000; // Number of encoding operations per benchmark iteration
    
    let _ = c.bench_function("new_encode_varint_10000", |b| {
        let mut buf = ProtoBuffer::new();
        b.iter(|| {
            for _ in 0..ITERATIONS {
                buf.encode_varint(black_box(value));
                buf.clear();
            }
            let _ = black_box(&buf);
        });
    });
    
    let _ = c.bench_function("old_encode_varint_10000", |b| {
        let mut buf = OldProtoBuffer::new();
        b.iter(|| {
            for _ in 0..ITERATIONS {
                buf.encode_varint_old(black_box(value));
                buf.clear();
            }
            let _ = black_box(&buf);
        });
    });
}

criterion_group!(benches, bench_varint_10000);
criterion_main!(benches);
```
</details>

### DIFF
--- a/rust/otel-arrow-rust/src/otlp/common.rs
+++ b/rust/otel-arrow-rust/src/otlp/common.rs
@@ -324,12 +324,29 @@ impl ProtoBuffer {
         self.encode_varint(key);
     }
 
-    pub fn encode_varint(&mut self, mut value: u64) {
-        while value >= 0x80 {
-            self.buffer.push(((value as u8) & 0x7F) | 0x80);
-            value >>= 7;
+    #[inline]
+    pub fn encode_varint(&mut self, value: u64) {
+        // Fast path for single byte (very common)
+        if value < 0x80 {
+            self.buffer.push(value as u8);
+            return;
         }
-        self.buffer.push(value as u8);
+        
+        // Fast path for two bytes (common)
+        if value < 0x4000 {
+            self.buffer.extend_from_slice(&[
+                ((value & 0x7F) | 0x80) as u8,
+                (value >> 7) as u8,
+            ]);
+            return;
+        }
+        
+        let mut v = value;
+        while v >= 0x80 {
+            self.buffer.push(((v & 0x7F) | 0x80) as u8);
+            v >>= 7;
+        }
+        self.buffer.push(v as u8);
     }
 
     /// encodes the signed varint type (e.g. sint32, sint64, etc.) using zig-zag encoding


albertlockett (approved):
cool, LGTM @utpilla

albertlockett:
@utpilla I ran the benchmark as well and got good results from your fix!

```
new_encode_varint_10000 time:   [17.602 µs 17.643 µs 17.685 µs]

old_encode_varint_10000 time:   [10.636 µs 10.714 µs 10.792 µs]
```

utpilla:
> @utpilla I ran the benchmark as well and got good results from your fix!
> 
> ```
> new_encode_varint_10000 time:   [17.602 µs 17.643 µs 17.685 µs]
> 
> old_encode_varint_10000 time:   [10.636 µs 10.714 µs 10.792 µs]
> ```

@albertlockett From the results you shared, it looks like old encoding is performing better than the new one. Did it worsen things? 😄

albertlockett:
> > @utpilla I ran the benchmark as well and got good results from your fix!
> > ```
> > new_encode_varint_10000 time:   [17.602 µs 17.643 µs 17.685 µs]
> > 
> > old_encode_varint_10000 time:   [10.636 µs 10.714 µs 10.792 µs]
> > ```
> 
> @albertlockett From the results you shared, it looks like old encoding is performing better than the new one. Did it worsen things? 😄

lol I copied the wrong lines from my terminal 🤦 ... those results above, I had run your benchmark without your code checked out. These are the actual results:
```
new_encode_varint_10000 time:   [5.6579 µs 5.6743 µs 5.6906 µs]

old_encode_varint_10000 time:   [10.699 µs 10.889 µs 11.160 µs]
```
Sorry for the confusion!

lquerel (approved):
Nice.

Issue #1075, opened 2025-09-05
DebugProcessor - Round 2 Enhancements


lquerel:
This debugging/troubleshooting capability has proven very useful since we turned it from an exporter into a processor. We can now integrate it at various points in a pipeline to see the before and after of certain operations.

In this second round of improvements (most likely multiple PRs), I propose the following enhancements:
- Batch vs. signal mode
  - Batch mode: the processor outputs the entire batch.
  - Signal mode: the processor outputs each signal individually (filtered signals are not shown).
- Filtering (inclusion/exclusion)
  - Inclusion filter: only outputs batches or signals that match a predicate.
    - In batch mode: if at least one signal matches, the entire batch is shown.
    -  In signal mode: only matching signals are shown.
  - Exclusion filter: suppresses from output any batches or signals that match a predicate.
  - **Note: The definition of predicates (syntax, supported fields such as attributes or other record fields) is still open for discussion and needs to be defined.**
- Signal type selection
  - Option to restrict which signal types (logs, spans, metrics) are considered for debug output.
- Sampling
  - Multiple sampling strategies should be considered, starting with the “zap” sampling currently supported by the debug exporter.
- Output modes
  - Output can be directed either to the console or to an out-port.

Example Configuration (YAML)

```yaml
debug_processor:
  # Signal types to consider for debug output
  signals: [logs, spans, metrics]

  # Output mode: console or out_port
  output: console

  # Mode: batch or signal
  mode: signal

  # Sampling strategy
  sampling:
    type: zap
    sampling_initial: 2
    sampling_thereafter: 10

  # Filtering rules
  filter:
    predicate: attributes["service.name"] == "payment"
    mode: include   # or exclude
```

kenarora:
Looks like a promising direction!  Some specific comments for consideration:

1.  _Subjective Word Choice_: I like "batch" mode, but the complement/opposite of batch mode might be renamed to "singleton" or "single" or "individual".
2. _Filtering_
- Are multiple filters allowed?  If so, they should be ordered (or some other strategy used to determine priorities among conflicting filters).  For example, if Filter "A" says exclude "severity < WARN" and Filter B says include "event_type = SECURITY", which wins for a SECURITY event with severity INFO?
- It might be more streamlined/one less field, if signals (the type of signal to consider) were just another predicate.  For example, "signals: [logs,spans,metrics]" could be a filter with predicate such as signal:[logs,spans,metrics] with mode: include.
[...]

PR #1156, opened 2025-09-18
[otap-dataflow] add filtering to debug processor

c1ly:
Part of https://github.com/open-telemetry/otel-arrow/issues/1075

Add a filter setting to the debug processor allowing users to define what fields to filter against. The filter setting will check against the signal objects and will update the msg depending on whether you choose to exclude or include this will affect the output that is displayed

The first version currently supports attributes but allows for future fields to be supported via the SignalField enum

Below is an example config 

``` yaml
verbosity: basic
mode: batch
signals:
- metrics
- spans
- logs
filters:
- predicate:
    field: attribute
    value:
    - key: service.name
      value: fdafsda
  mode: exclude
```
this filter would look for the service.name attribute and with a specific value. Value can be of many types from string, int, double, and key value

Below is the current rust definitions

``` rust
pub struct Config {
    verbosity: Verbosity,
    mode: OutputMode,
    signals: HashSet<SignalActive>,
    filters: Vec<FilterRules>,
}

#[serde(rename_all = "snake_case")]
pub enum FilterMode {
    Include,
    Exclude,
}

pub struct FilterRules  {
    predicate: Predicate,
    mode: FilterMode,
}


pub struct Predicate {
    field: SignalField,
    value: MatchValue,
}

#[serde(rename_all = "snake_case")]
pub enum SignalField {
    Attribute
}


#[serde(untagged)]
pub enum MatchValue {
    String(String),
    Int(i64),
    Double(f64),
    Boolean(bool),
    Array(Vec<MatchValue>),
    KeyValue(Vec<KeyValue>)
}

pub struct KeyValue {
    key: String,
    value: MatchValue
}
```


rust/otap-dataflow/crates/otap/src/debug_processor.rs
```diff
@@ -338,14 +342,25 @@ async fn push_metric(
...
         OutputMode::Signal => {
-            // safety: this should be initialized above if the mode is Signal
-            let metric_signals = metric_signals.expect("metric_signals not None");
+            let metric_signals: Vec<Metric> = metric_request
+                .resource_metrics
+                .into_iter()
+                .flat_map(|resource| resource.scope_metrics)
+                .flat_map(|scope| scope.metrics)
+                .collect();
             for (index, metric) in metric_signals.iter().enumerate() {
```
albertlockett: nit: I don't think we'd need to collect these into a vec before iterating it in the for loop.. we could probably do something like:
```suggestion
            let metric_signals = metric_request
                .resource_metrics
                .into_iter()
                .flat_map(|resource| resource.scope_metrics)
                .flat_map(|scope| scope.metrics);
            for (index, metric) in metrics_signals.enumerate()  {
```

albertlockett: we could make the same change for spans and logs below as well

c1ly: ahh I see, we keep the iter

lquerel (approved):
LGTM

PR #1160, opened 2025-09-19
[otap-df-otap] Minor simplification to OTAP Exporter

utpilla:
## Changes
- With this change we `match` on `SignalType` only once
- Earlier we would `match` on `SignalType` twice: once when calling `self.pdata_metrics.inc_consumed(signal_type);` and the second time for finding the right sender

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/otap_exporter.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_exporter.rs
@@ -9,7 +9,7 @@
 
 use crate::OTAP_EXPORTER_FACTORIES;
 use crate::metrics::ExporterPDataMetrics;
-use crate::pdata::OtapPdata;
+use crate::pdata::{OtapPayload, OtapPdata};
 use async_stream::stream;
 use async_trait::async_trait;
 use linkme::distributed_slice;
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/otap/src/otap_exporter.rs
```diff
@@ -218,22 +218,31 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
                     }
                     //send data
                     Message::PData(pdata) => {
-                        // Capture signal type before moving pdata into try_from
-                        let signal_type = pdata.signal_type();
-
-                        self.pdata_metrics.inc_consumed(signal_type);
```
utpilla, line 224: This is where the additional `match` occurs.

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
```diff
@@ -218,22 +218,31 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
...
+                        // TODO(#1098): Note context is dropped.
                         let (_context, payload) = pdata.into_parts();
 
-                        // TODO(#1098): Note context is dropped.
-                        let message: OtapArrowRecords = payload
-                            .try_into()
-                            .inspect_err(|_| self.pdata_metrics.inc_failed(signal_type))?;
```
albertlockett, line 230: I think we need to keep this call to try_into right? because if the payload was in otlp format, we want to convert it to OTAP when we call try_into here

utpilla: Good point! I could potentially add dedicated branches for all combinations of signal types and OtlpProtoBytes but the code doesn't look as good if do that. Not sure if all that is worth avoiding the cost of an additional match statement,  I'm going to close this PR for now. We could revisit this later.

PR #1172, opened 2025-09-22
[perf][otap-df-otap] Avoid redundant allocations for validating AttributeTransform config

utpilla:
## Changes
- The transforms for `AttributesProcessor` are not changed after initialization
- Currently we are `AttributesTransform::validate()` in the hot-path. This method allocates a `BTreeSet` when called
- In this PR, I have updated the code to perform validation of `AttributesTransform` just one in the `new` method


lquerel (approved):
LGTM

PR #1177, opened 2025-09-23
[perf] Avoid allocations in transform plan creation hot-path

utpilla:
## Changes
- Reference the precomputed bytes representation of the transforms instead of collecting them into `vec`s repeatedly
- Updated the transform benchmarks to use `black_box` around the input and to make the measurement more targeted

### DIFF
--- a/rust/otel-arrow-rust/benches/attribute_transform.rs
+++ b/rust/otel-arrow-rust/benches/attribute_transform.rs
@@ -10,7 +10,9 @@ use std::collections::{BTreeMap, BTreeSet};
 use std::hint::black_box;
 use std::sync::Arc;
 
-use otel_arrow_rust::otap::transform::{transform_attributes, AttributesTransform, DeleteTransform, RenameTransform};
+use otel_arrow_rust::otap::transform::{
+    AttributesTransform, DeleteTransform, RenameTransform, transform_attributes,
+};
 use otel_arrow_rust::schema::consts;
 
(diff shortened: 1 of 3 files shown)


rust/otel-arrow-rust/src/otap/transform.rs
```diff
@@ -1478,7 +1518,7 @@ struct KeyReplacementPlan<'a> {
     ranges: Vec<(usize, usize, usize)>,
 
     /// this contains the bytes to replace each value in each of `ranges`
-    replacement_bytes: Vec<&'a [u8]>,
+    replacement_bytes: &'a [Vec<u8>],
```
utpilla, line 1523: Updated the data type to use slice to utilize the precomputed bytes.

rust/otel-arrow-rust/benches/attribute_transform.rs
```diff
@@ -61,11 +61,42 @@ fn generate_dict_keys_attribute_batch(
 }
 
 fn bench_transform_attributes(c: &mut Criterion) {
+    // Pre-create AttributesTransform instances to avoid measuring their creation cost
+    let single_replace_no_delete = AttributesTransform {
```
utpilla, line 67: I had to extract the `AttributeTransform` instances outside of the benchmarks. Since, we now precompute the bytes during `AttributesTransform` creation, we need to exclude that time from the measurements otherwise we would have misleading numbers.

PR #1180, opened 2025-09-24
[perf] Avoid unnecessary vec allocation for merged transforms

utpilla:
## Changes
- In the current implementation, we always allocate a `vec` when merging transform ranges (even if there is no merge required for example if there are either only replacement ranges or only delete ranges)
- With this PR, we append the `KeyTransformRangeType` data from the get-go and store them in `KeyReplacementPlan` and `KeyDeletePlan`. This allows us to simply refer to these fully self-contained ranges in the `merge_transform_ranges` method if there is no merge required.

### DIFF
--- a/rust/otel-arrow-rust/src/otap/transform.rs
+++ b/rust/otel-arrow-rust/src/otap/transform.rs
@@ -1011,9 +1011,9 @@ fn transform_keys(
     let mut last_end_offset = 0;
 
     // iterate over the transform ranges, copying unmodified ranges and replacing renamed keys
-    for (start_idx, end_idx, transform_idx, range_type) in transform_ranges.iter() {
+    for (start_idx, end_idx, transform_idx, range_type) in transform_ranges.iter().copied() {
         // directly copy all the bytes of the values that were not replaced
-        let start_offset = offsets[*start_idx] as usize;
+        let start_offset = offsets[start_idx] as usize;
         new_values.extend_from_slice(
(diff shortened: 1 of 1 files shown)


rust/otel-arrow-rust/src/otap/transform.rs
```diff
@@ -1472,43 +1473,35 @@ fn merge_transform_ranges(
...
             }
 
-            result
+            Cow::Owned(result)
         }
-        (Some(replacement_plan), None) => replacement_plan
-            .ranges
-            .iter()
-            .map(|(start, end, idx)| (*start, *end, *idx, KeyTransformRangeType::Replace))
-            .collect(),
```
utpilla, line 1505: This is the allocation that we now avoid.

rust/otel-arrow-rust/src/otap/transform.rs
```diff
@@ -1675,18 +1681,18 @@ fn find_matching_key_ranges(
...
         // add the final trailing range
         if let Some(s) = eq_range_start {
-            ranges.push((s, array_len, target_idx))
+            ranges.push((s, array_len, target_idx, range_type));
         }
     }
 
     // Sort the ranges to replace by start_index (first element in contained tuple)
-    ranges.sort();
+    ranges.sort_unstable_by_key(|r| r.0);
```
utpilla, line 1695: Using unstable sort here.

albertlockett (approved):
LGTM, thanks @utpilla

PR #1181, opened 2025-09-25
Ack/Nack support [WIP]

jmacd:
Replaces #1114, #1083.

This PR demonstrates: 

- New AckMsg, NackMsg types
- Interests is a bitflags! of ACKS and NACKS
- CtxVal is [u8;8] with bytemuck::cast
- CallData is SmallVec of 2 CtxVal
- Context is a Vec<Frame>
- Frame is { Interests, CallData, NodeId }
- ProducerEffectHandlerExtension: subscribe_to for Interests w/ call data
- ConsumerEffectHandlerExtension: notify_nack and notify_ack invoke the delivery path
- DelayedData is a placeholder NodeControlMsg for data returning after a retry_processor attempt
- OTLP exporter proof-of-concept returns Ack and Nack: TODOs about cloning of data
- OTLP receiver has comments indicating needs like a way to block the RPC awaiting Ack/Nack
- EffectHandler wrappers for core delay_data(), route_ack(), and route_nack() methods

No testing.
Extraneous: renames OtapPayload::payload() to take_payload(), eases minor confusion. See how we "dump" data, and the TODOs it creates; we're losing information about num_items though we've retained signal type here.

### DIFF
--- a/rust/otap-dataflow/Cargo.toml
+++ b/rust/otap-dataflow/Cargo.toml
@@ -88,6 +88,9 @@ tonic = { version = "0.14", default-features = false, features = [
 tonic-middleware = "0.4.0"
 tonic-prost = "0.14"
 prost = "0.14"
+bitflags = "2.9"
+bytemuck = "1.23"
+smallvec = "1.15"
 
 [features]
 default = []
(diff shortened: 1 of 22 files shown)


lquerel (approved):
I think in its current state, it's heading in the right direction.
Thanks

jmacd:
Replaced by #1225

PR #1185, opened 2025-09-26
[query-engine] Remove top-level fields specified in attributes schema in otlp bridge

CodeBlanch:
## Changes

* Add special handling for top-level fields being specified as attribute schema in OTLP bridge

### DIFF
--- a/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
+++ b/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
@@ -39,7 +39,7 @@ pub fn parse_kql_query_into_pipeline(
 
     if let Some(mut attributes_schema) = options.and_then(|mut v| v.take_attributes_schema()) {
         let schema = attributes_schema.get_schema_mut();
-        for top_level_key in log_record_schema.get_schema().keys() {
+        for (top_level_key, top_level_key_schema) in log_record_schema.get_schema() {
             // Note: If any top-level fields are duplicated on Attributes Schema
             // they get removed automatically. This is done for two purposes.
             // The first is to make it easy for callers to pass in something
@@ -51,7 +51,14 @@ pub fn parse_kql_query_into_pipeline(
             // present in Attributes users might query with ambiguous naming.
             // For example: source | extend Body = 'something' will write to the
             // top-level field and not Attributes.
-            schema.remove(top_level_key);
+            if let Some(removed) = schema.remove(top_level_key)
+                && &removed != top_level_key_schema
+            {
+                return Err(vec![ParserError::SchemaError(format!(
+                    "'{top_level_key}' key cannot be declared as '{}' type",
+                    &removed
+                ))]);
+            }
         }
 
         parser_options = parser_options.with_summary_map_schema(attributes_schema.clone());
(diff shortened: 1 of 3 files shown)


rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
```diff
@@ -534,7 +549,8 @@ mod tests {
                 Some(
                     BridgeOptions::new().with_attributes_schema(
                         ParserMapSchema::new()
-                            .with_key_definition("int_value", ParserMapKeySchema::Double),
+                            .with_key_definition("Body", ParserMapKeySchema::Double)
```
drewrelmas: Confirming my own interpretation here - since Body is an `Any` value, there is no enforcement here of having a `Double` value?

rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
```diff
@@ -25,28 +25,47 @@ pub fn parse_kql_query_into_pipeline(
...
+            // The first is to make it easy for callers to pass in something
+            // like a table schema. Many backends flatten log records into
+            // columns. This feature is essentially a convenience thing so
+            // callers with table schema don't need to map columns back to the
+            // log record schema. The second reason is to prevent
+            // accidental\confusing query results. If for example "Body" is
+            // present in Attributes users might query with ambiguous naming.
+            // For example: source | extend Body = 'something' will write to the
+            // top-level field and not Attributes.
+            schema.remove(top_level_key);
```
drewrelmas: Could it be worth leaving a diagnostic message here saying this happened?

CodeBlanch: We could but there is a bit of a challenge. This method for query validation doesn't currently return diagnostics, it only returns errors. We can't return (at the moment) warnings or infos. We could expand it for that. But then this particular thing we don't really want to display to users it is really for the dev doing the implementation. If it got in front of the end-user that would kind of suck.

Thoughts?

rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
CodeBlanch: In this case "Body: Double" gets dropped so it doesn't really matter what the type is. We could do _something_ with this. One option would be return an error saying "Hey you tried to declare 'Body' as a 'Double' which isn't supported". If then it gets switched to "Map" we would just drop it and continue along.

rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
drewrelmas: I agree that this shouldn't be user-facing. I'm not entirely satisfied with silently doing this but there isn't a clear way to notify. At the very least, this should be part of the crate documentation.

rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
drewrelmas: I'm going back and forth on the answer I gave [here](https://github.com/open-telemetry/otel-arrow/pull/1185#discussion_r2383226963).

I think it could be better to return an error here - otherwise we're essentially ignoring customer's request in public contract.

CodeBlanch: Added

PR #1187, opened 2025-09-26
Configuration validation errors should prevent startup

jmacd:
It looks like there was a regression causing configuration validation logic not to fail or print at startup.

I think this is the desired behavior.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/lib.rs
+++ b/rust/otap-dataflow/crates/otap/src/lib.rs
@@ -72,3 +72,61 @@ pub mod otlp_grpc;
 /// Factory for OTAP-based pipeline
 #[pipeline_factory(OTAP, OtapPdata)]
 pub static OTAP_PIPELINE_FACTORY: PipelineFactory<OtapPdata> = build_factory();
+
+#[cfg(test)]
+mod tests {
+    use otap_df_config::{
+        engine::HttpAdminSettings,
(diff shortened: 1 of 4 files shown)


rust/otap-dataflow/crates/otap/src/lib.rs
```diff
@@ -72,3 +72,61 @@ pub mod otlp_grpc;
 /// Factory for OTAP-based pipeline
 #[pipeline_factory(OTAP, OtapPdata)]
 pub static OTAP_PIPELINE_FACTORY: PipelineFactory<OtapPdata> = build_factory();
+
+#[cfg(test)]
+mod tests {
+    use otap_df_config::{
```
albertlockett, line 78: nit: should this test be in the otap_df_controller crate?

It looks like we're only importing `otap_df_controller` crate in `otap_df_otap` for this test, but avoiding that will reserve for us the option to import `otap_df_otap` in the controller crate if we ever need to, without having a circular dependency

jmacd: I can move this to the controller crate, and let that crate's tests depend on otap-df-otap.

jmacd:
I noticed that the process now refuses to exit when I send ^C.

There's something not right about the change itself.

jmacd:
See #1207

PR #1188, opened 2025-09-26
Add support for KQL not()

drewrelmas:
Noticed we have the `LogicalExpression` support for `not()` (used in conjunction with `where`) but never added this standalone in the KQL grammar

### DIFF
--- a/rust/experimental/query_engine/kql-parser/src/kql.pest
+++ b/rust/experimental/query_engine/kql-parser/src/kql.pest
@@ -255,10 +255,8 @@ scalar_list_expression = {
     "(" ~ scalar_expression ~ ("," ~ scalar_expression)* ~ ")"
 }
 
-not_expression = { "not" ~ "(" ~ scalar_expression ~ ")" }
 logical_expression = {
     scalar_expression
-    not_expression
 }
 
 assignment_expression = { accessor_expression ~ "=" ~ scalar_expression }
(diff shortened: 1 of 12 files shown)


rust/experimental/query_engine/kql-parser/src/kql.pest
```diff
@@ -212,6 +212,11 @@ temporal_expressions = {
     now_expression
 }
 
+not_expression = { "not" ~ "(" ~ logical_expression ~ ")" }
+logical_function_expressions = {
```
CodeBlanch: How do you feel about renaming this `logical_function_expressions` -> `logical_expressions`? That is more consistent with the other "groups" we have. I'll admit it is kind of weird to have `logical_expressions` and `logical_expression` but I can't think of anything much better.

I suppose we could rename them all to a pattern like...

```
scalar_unary_expression = {
    type_unary_expressions
    | conditional_unary_expressions
    | conversion_unary_expressions
    | string_unary_expressions
    | array_unary_expressions
    | math_unary_expressions
    | temporal_unary_expressions
    | logical_unary_expressions
    ...
}
```

drewrelmas: Followed that pattern for `*_unary_expressions`

jmacd (approved):

PR #1196, opened 2025-09-29
Use optimized `append_..._n` methods in syslog receiver when constructing OTAP record

### LINKED ISSUE #879: Add `append_slice_n` to FixedSizeBinaryBuilder and use optimized method in SyslogReceiver
Currently, syslog receiver is doing many loops to append `None` to the adaptive array builder for FixedSizeBinary.

It's a bit tricky to add optimized methods that call `append_nulls` using the current pattern, which is to accept an optional value and `n: usize` values to insert, because there's no `append_slice_n` on `FixedSizeBinaryBuilder` yet

current pattern:
https://github.com/open-telemetry/otel-arrow/blob/e119596b3e8abd447a9bd19a975aec430591ec36/rust/otel-arrow-rust/src/encode/record/logs.rs#L159-L165

We should implement this method, expose the `append_<field>_n` on the log builder for the fields where it's not currently exposed, and use these optimized methods in the SyslogReceiver

see related discussion:
https://github.com/open-telemetry/otel-arrow/pull/861/files#r2257954281

albertlockett:
Closes #879 

Use the "append_..._n" methods in the syslog receiver when constructing the arrow records. These methods have some nice performance optimizations:
- when appending a non-null value if the underlying builder is a dictionary array, the builder is able to append the value multiple times without doing multiple key lookups. For example:
https://github.com/apache/arrow-rs/blob/0737c61e76057b127312dd8058887649ece702b8/arrow-array/src/builder/generic_bytes_dictionary_builder.rs#L304-L316

- when the value is null, we can use the optimized code path to append multiple nulls to the null buffer:
https://github.com/apache/arrow-rs/blob/0737c61e76057b127312dd8058887649ece702b8/arrow-buffer/src/builder/boolean.rs#L193

Benchmark results:
```
arrow_batch_creation/rfc3164_arrow_batch_100_msgs
                        time:   [82.608 µs 82.774 µs 82.950 µs]
                        change: [−3.9462% −3.4305% −2.9199%] (p = 0.00 < 0.05)
                        Performance has improved.
                        
arrow_batch_creation/rfc5424_arrow_batch_100_msgs
                        time:   [46.139 µs 46.238 µs 46.338 µs]
                        change: [−4.2893% −3.9924% −3.7179%] (p = 0.00 < 0.05)
                        Performance has improved.
                        
arrow_batch_creation/cef_arrow_batch_100_msgs
                        time:   [40.973 µs 41.141 µs 41.308 µs]
                        change: [−3.8351% −3.3433% −2.8615%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/rust/otel-arrow-rust/src/encode/record/array/fixed_size_binary.rs
+++ b/rust/otel-arrow-rust/src/encode/record/array/fixed_size_binary.rs
@@ -193,11 +193,22 @@ mod test {
         CheckedArrayAppend::append_value(&mut fsb_builder, &b"1234".to_vec()).unwrap();
         CheckedArrayAppend::append_value(&mut fsb_builder, &b"5678".to_vec()).unwrap();
         CheckedArrayAppend::append_value(&mut fsb_builder, &b"9012".to_vec()).unwrap();
+        CheckedArrayAppendSlice::append_slice(&mut fsb_builder, b"4180").unwrap();
+        CheckedArrayAppendSlice::append_slice_n(&mut fsb_builder, b"5140", 2).unwrap();
+
         let result = ArrayBuilder::finish(&mut fsb_builder);
         assert_eq!(result.data_type(), &DataType::FixedSizeBinary(4));
 
         let expected = FixedSizeBinaryArray::try_from_iter(
-            [b"1234".to_vec(), b"5678".to_vec(), b"9012".to_vec()].iter(),
+            [
+                b"1234".to_vec(),
+                b"5678".to_vec(),
+                b"9012".to_vec(),
+                b"4180".to_vec(),
+                b"5140".to_vec(),
+                b"5140".to_vec(),
+            ]
+            .iter(),
         )
         .unwrap();
 
@@ -222,6 +233,11 @@ mod test {
         let index =
             CheckedDictionaryArrayAppend::append_value(&mut dict_builder, &b"b".to_vec()).unwrap();
         assert_eq!(index, 1);
+        let index = CheckedDictionaryAppendSlice::append_slice(&mut dict_builder, b"c").unwrap();
+        assert_eq!(index, 2);
+        let index =
+            CheckedDictionaryAppendSlice::append_slice_n(&mut dict_builder, b"d", 2).unwrap();
+        assert_eq!(index, 3);
 
         let result = DictionaryBuilder::finish(&mut dict_builder);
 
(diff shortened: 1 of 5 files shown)


rust/otel-arrow-rust/src/encode/record/array/fixed_size_binary.rs
```diff
@@ -114,6 +122,21 @@ where
...
         }
     }
+
+    fn append_slice_n(
+        &mut self,
+        val: &[Self::Native],
+        n: usize,
+    ) -> super::dictionary::checked::Result<usize> {
+        // TODO use optimized method once we upgrade to latest arrow:
+        // https://github.com/apache/arrow-rs/pull/8498
```
albertlockett, line 132: https://github.com/apache/arrow-rs/issues/8497

rust/otel-arrow-rust/src/encode/record/logs.rs
```diff
@@ -142,6 +142,12 @@ impl LogsRecordBatchBuilder {
         self.observed_time_unix_nano.append_value(&val);
     }
 
+    /// append a value to the `observed_time_unix_nano` array `n` times
+    pub fn append_observed_time_unix_nano_n(&mut self, val: Option<i64>, n: usize) {
+        let val = val.unwrap_or(0);
```
utpilla: Could we simply use `i64` as the value argument type instead of using `Option<i64>`?

If we are using `Option<i64>` because observed timestamp is a nullable field, then maybe we should map `None` values to null?

rust/otel-arrow-rust/src/encode/record/logs.rs
```diff
@@ -142,6 +142,12 @@ impl LogsRecordBatchBuilder {
         self.observed_time_unix_nano.append_value(&val);
     }
 
+    /// append a value to the `observed_time_unix_nano` array `n` times
+    pub fn append_observed_time_unix_nano_n(&mut self, val: Option<i64>, n: usize) {
```
utpilla: nit: I think we could mark these methods with `#[inline]` attribute.

albertlockett: made this change

rust/otel-arrow-rust/src/encode/record/logs.rs
albertlockett: Yeah, we can probably use `i64` as the type. Made this change.

This field is non-nullable, where `0` represents the null value:
https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L141-L159

So originally we were passing `None` here and encapsulating the "null = 0" semantics in the builder, but I think it probably makes more sense to have pass `i64` as the field shouldn't take a null value and now the args make this clear

lquerel (approved):
LGTM

PR #1199, opened 2025-09-29
[otap-dataflow] syslog cef receiver internal telemetry

daviddahl:
- Add internal telemetry collection
- Add tests for internal telemetry

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
+++ b/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
@@ -177,7 +177,7 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
                                     let metrics = Rc::clone(&self.metrics);
 
                                     // Spawn a task to handle the connection.
-                                    // TODO: Use the JoinHandle for graceful shutdown when wiring stop signals.
+                                    // ToDo should this be abstracted and exposed a method in the effect handler?
                                     _ = tokio::task::spawn_local(async move {
                                         let mut reader = BufReader::new(socket);
                                         let mut line_bytes = Vec::new();


rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -202,39 +240,36 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+                                                                    continue;
                                                                 }
                                                             };
 
-                                                            arrow_records_builder.append_syslog(parsed_message);
-
                                                             // Clear the bytes for the next iteration
                                                             line_bytes.clear();
 
-                                                            if arrow_records_builder.len() >= MAX_BATCH_SIZE {
```
utpilla, line 218: Why is this being removed?

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -349,6 +415,34 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+/// RFC-aligned metrics for Syslog CEF receiver.
+///
+/// Key fields follow the component universal telemetry RFC for receivers:
+/// received.items.<signal>.<outcome>
+#[metric_set(name = "syslog_cef.receiver.metrics")]
+#[derive(Debug, Default, Clone)]
+pub struct SyslogCefReceiverMetrics {
+    /// Number of log records successfully received and forwarded
+    #[metric(unit = "{item}")]
+    pub received_items_logs_success: Counter<u64>,
```
utpilla: Using both `items_logs` sounds wordy. I think we can use more concise name for metrics.
```suggestion
    pub received_logs_success: Counter<u64>,
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -349,6 +415,34 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+    #[metric(unit = "{item}")]
+    pub received_items_logs_refused: Counter<u64>,
+
+    /// Number of TCP connections accepted
+    #[metric(unit = "{conn}")]
+    pub tcp_connections_accepted: Counter<u64>,
+
+    /// Number of batches built and sent downstream
+    #[metric(unit = "{batch}")]
+    pub batches_built: Counter<u64>,
```
utpilla: What's the utility of this metric? It looks like we are already tracking the number of logs successfully received or sent to downstream components.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -349,6 +415,34 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+    #[metric(unit = "{item}")]
+    pub received_items_logs_failure: Counter<u64>,
+
+    /// Number of log records refused by downstream (backpressure/unavailable)
+    #[metric(unit = "{item}")]
+    pub received_items_logs_refused: Counter<u64>,
+
+    /// Number of TCP connections accepted
+    #[metric(unit = "{conn}")]
+    pub tcp_connections_accepted: Counter<u64>,
```
utpilla: Just knowing the number of accepted connections isn't super helpful. I think a better metric would be to track the number of active connections. We could use an `UpDownCounter<u64>` to track it.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -110,10 +119,14 @@ pub static SYSLOG_CEF_RECEIVER: ReceiverFactory<OtapPdata> = ReceiverFactory {
 #[async_trait(?Send)]
 impl local::Receiver<OtapPdata> for SyslogCefReceiver {
     async fn start(
-        self: Box<Self>,
+        mut self: Box<Self>,
         mut ctrl_chan: local::ControlChannel<OtapPdata>,
         effect_handler: local::EffectHandler<OtapPdata>,
     ) -> Result<(), Error> {
+        // Start periodic telemetry collection (1s), similar to other nodes
+        let _ = effect_handler
```
utpilla: We want to cancel the telemetry collection timer when being shut down. Instead of discarding the result here, obtain a reference to the handle and call `cancel()` on it when processing `NodeControlMsg::Shutdown`.

Check #1155 for reference.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -142,10 +155,12 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
                         accept_result = listener.accept() => {
                             match accept_result {
                                 Ok((socket, _peer_addr)) => {
-                                    // Clone the effect handler so the spawned task can send messages.
+                                    // Count accepted connections
+                                    self.metrics.tcp_connections_accepted.inc();
+                                    // Clone the effect handler and metrics so the spawned task can send messages and record telemetry.
                                     let effect_handler = effect_handler.clone();
+                                    let mut metrics = self.metrics.clone();
```
utpilla: Does the clone point to the same `MetricSet` that get reported? I suspect this would effectively lead to this task working on its own copy of the `MetricSet` which would not get reported.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -349,6 +415,34 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
 }
 
+/// RFC-aligned metrics for Syslog CEF receiver.
+///
+/// Key fields follow the component universal telemetry RFC for receivers:
+/// received.items.<signal>.<outcome>
+#[metric_set(name = "syslog_cef.receiver.metrics")]
+#[derive(Debug, Default, Clone)]
+pub struct SyslogCefReceiverMetrics {
+    /// Number of log records successfully received and forwarded
```
utpilla: This is not correct. This metric only tracks the number of logs forwarded. The received logs could be higher.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -349,6 +415,34 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+    #[metric(unit = "{item}")]
+    pub received_items_logs_success: Counter<u64>,
+
+    /// Number of log records that failed to be received (parse errors, etc.)
```
utpilla, line 542: This is not correct. This metric is only tracking the number of logs that couldn't be parsed. The number of received logs could be higher. I think we need a dedicated metric for received logs which would track the count of logs received at the socket level even we before attempt to parse.

rust/otap-dataflow/crates/engine/src/testing/receiver.rs
```diff
@@ -92,6 +92,23 @@ impl<PData> TestContext<PData> {
             })
     }
 
+    /// Requests the node under test to collect and report its telemetry metrics.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the message could not be sent.
+    pub async fn send_collect_telemetry(
```
utpilla: It doesn't look like you're using this method. Could we remove it?

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
daviddahl: It was not intended. I was tinkering and forgot to restore it.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
daviddahl: batches_built was intended to show how often the receiver flushed a batch downstream, regardless of why. I have replaced it with:

batches.built.size: incremented when MAX_BATCH_SIZE forces a flush
batches.built.time: incremented on interval-based flush

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
daviddahl: ah. yes. I have removed the clone and am using a channel based approach to observe correct ownership

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
daviddahl: added a dedicated pre-parse counter to get the correct received numbers

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
utpilla: Could you please help me understand how `batches.built.size` and `batches.built.time` provide actionable insights or inform decision-making? I share a similar concern regarding `tcp_connections_accepted` Its practical impact isn't clear to me. We already track `tcp_connections_active`, which seems to cover the necessary TCP connection status.

I’m inclined to avoid emitting metrics that don’t contribute meaningful value or influence operational decisions.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
utpilla: The channel approach based doesn't seem well suited for the thread-per-core shared-nothing architecture of the engine.  We should be able to avoid performance costs due to thread synchronization (Arc etc.) and even channels.

Since all the tasks that are spawned to handle TCP connections are executed on the same thread, we should be able to get this done with some `Rc<RefCell<T>>` construct where `T` is `MetricSet<SyslogCefReceiverMetrics>`.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
daviddahl: I thought perhaps operators might want to tune MAX_BATCH_SIZE or the pipeline based on the size versus time triggers of sending downstream. Since this is just noise, I will remove them.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -110,43 +123,63 @@ pub static SYSLOG_CEF_RECEIVER: ReceiverFactory<OtapPdata> = ReceiverFactory {
...
-        self: Box<Self>,
+        mut self: Box<Self>,
         mut ctrl_chan: local::ControlChannel<OtapPdata>,
         effect_handler: local::EffectHandler<OtapPdata>,
     ) -> Result<(), Error> {
+        // Start periodic telemetry collection (1s), similar to other nodes
+        let handle = effect_handler
+            .start_periodic_telemetry(std::time::Duration::from_secs(1))
+            .await?;
+        let mut telemetry_timer = Some(handle);
```
utpilla: Why are we wrapping it with an `Option`? I don't think it's needed. You can use the handle directly.

```rust
let timer_cancel_handle = effect_handler
    .await?;

...

Ok(NodeControlMsg::Shutdown { .. }) => {
    _ = timer_cancel_handle.cancel().await;
    break;
}
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -110,43 +123,63 @@ pub static SYSLOG_CEF_RECEIVER: ReceiverFactory<OtapPdata> = ReceiverFactory {
...
-                                    // Clone the effect handler so the spawned task can send messages.
+                                    // Track active connections
+                                    {
+                                        let mut m = self.metrics.borrow_mut();
+                                        m.tcp_connections_active.inc();
+                                    }
+
+                                    // Clone the effect handler and metrics so the spawned task can send messages and record telemetry.
                                     let effect_handler = effect_handler.clone();
+                                    let metrics = self.metrics.clone();
```
utpilla: We don't want to clone the `MetricSet`.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -110,43 +123,63 @@ pub static SYSLOG_CEF_RECEIVER: ReceiverFactory<OtapPdata> = ReceiverFactory {
...
+
                                     // Spawn a task to handle the connection.
-                                    // ToDo should this be abstracted and exposed a method in the effect handler?
-                                    _ = tokio::task::spawn_local(async move {
+                                    drop(tokio::task::spawn_local(async move {
```
utpilla: I think it's best to leave this as is. We want to keep the TODO comment and intentionally use `_ = tokio::task::spawn_local(async move { ... })` instead of dropping the handle. I plan to use that handle later when working on the graceful shutdown feature.

In general, I'd recommend keeping changes focused on the purpose of this PR and avoiding unrelated modifications.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
daviddahl: i have updated this to be obvious we are cloning the Rc pointer not the MetricSet itself

utpilla: Ah that's my bad! I mistakenly assumed that it's cloning the `MetricSet`.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -245,28 +359,48 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+                                                            .await
+                                                        {
+                                                            Ok(_) => {
+                                                                let mut m = metrics.borrow_mut();
+                                                                m.received_logs_success.add(items);
+                                                            }
+                                                            Err(_) => {
+                                                                let mut m = metrics.borrow_mut();
+                                                                m.received_logs_refused.add(items);
+                                                                m.tcp_connections_active.dec();
```
utpilla: The TCP connection would still be active. It's closed only on EOF or socket read error.

```suggestion
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -947,3 +1167,239 @@ mod tests {
...
+            // Trigger telemetry snapshot
+            let _ = ctrl_tx.send(NodeControlMsg::CollectTelemetry {
+                metrics_reporter: reporter.clone(),
+            });
+            sleep(Duration::from_millis(50)).await;
+
+            // Inspect metrics
+            let map = collect_syslog_metrics_map(&registry);
+            let get = |k: &str| map.get(k).copied().unwrap_or(0);
+            assert!(get("received.logs.success") >= 1, "success >= 1");
```
utpilla: Let's use stricter assertions to check for the exact value instead of `>=`.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -202,39 +270,85 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+                                                                    continue;
                                                                 }
                                                             };
 
-                                                            arrow_records_builder.append_syslog(parsed_message);
-
                                                             // Clear the bytes for the next iteration
                                                             line_bytes.clear();
 
+                                                            // Flush when batch size threshold is reached (match UDP behavior and main)
```
utpilla:
```suggestion
                                                            // Flush when batch size threshold is reached
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -110,42 +123,61 @@ pub static SYSLOG_CEF_RECEIVER: ReceiverFactory<OtapPdata> = ReceiverFactory {
...
                 loop {
                     tokio::select! {
-                        biased; //Prioritize control messages over data
+                        biased; // Prioritize control messages over data
 
                         // Process incoming control messages.
                         ctrl_msg = ctrl_chan.recv() => {
                             match ctrl_msg {
-                                Ok(NodeControlMsg::Shutdown {..}) => {
-                                // ToDo: Add proper deadline function
```
utpilla, line 129: Don't remove this. This is not done.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -110,42 +123,61 @@ pub static SYSLOG_CEF_RECEIVER: ReceiverFactory<OtapPdata> = ReceiverFactory {
...
+                                    }
+
+                                    // Clone the effect handler so the spawned task can send messages and record telemetry.
                                     let effect_handler = effect_handler.clone();
+                                    // Clone the Rc pointer (not the inner MetricSet) so all tasks update the same metrics instance.
+                                    let metrics = Rc::clone(&self.metrics);
+
                                     // Spawn a task to handle the connection.
-                                    // ToDo should this be abstracted and exposed a method in the effect handler?
+                                    // TODO: Use the JoinHandle for graceful shutdown when wiring stop signals.
```
utpilla: This seems wrong. The Todo statement is completely changed. We don't want to remove any ToDo until they are actually done.
```suggestion
                                    // ToDo should this be abstracted and exposed a method in the effect handler?
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -333,22 +496,66 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+/// RFC-aligned metrics for Syslog CEF receiver.
+///
+/// Key fields follow the component universal telemetry RFC for receivers:
+/// received.items.<signal>.<outcome>
+#[metric_set(name = "syslog_cef.receiver.metrics")]
+#[derive(Debug, Default, Clone)]
+pub struct SyslogCefReceiverMetrics {
+    /// Number of log records successfully forwarded downstream
+    #[metric(unit = "{item}")]
+    pub received_logs_success: Counter<u64>,
```
utpilla, line 540: `received_logs_success` sounds the same as `received_logs_total`. We should use a name that conveys the intention clearly.

```suggestion
    pub received_logs_forwarded: Counter<u64>,
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -333,22 +496,66 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+
+    /// Number of log records that failed to be received (parse errors, etc.)
+    #[metric(unit = "{item}")]
+    pub received_logs_failure: Counter<u64>,
```
utpilla, line 544: `received_logs_invalid` is a better name for the metric

```suggestion
    pub received_logs_invalid: Counter<u64>,
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -333,22 +496,66 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+
+    /// Number of log records refused by downstream (backpressure/unavailable)
+    #[metric(unit = "{item}")]
+    pub received_logs_refused: Counter<u64>,
```
utpilla, line 548: `received_logs_forward_failed` removes any ambiguity about who refused the logs and why.

```suggestion
    pub received_logs_forward_failed: Counter<u64>,
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -947,3 +1165,241 @@ mod tests {
...
+            // Trigger telemetry snapshot
+            let _ = ctrl_tx.send(NodeControlMsg::CollectTelemetry {
+                metrics_reporter: reporter.clone(),
+            });
+            sleep(Duration::from_millis(50)).await;
+
+            // Inspect metrics
+            let map = collect_syslog_metrics_map(&registry);
+            let get = |k: &str| map.get(k).copied().unwrap_or(0);
+            assert!(get("received.logs.refused") >= 1, "refused >= 1");
```
utpilla, line 1387: Same as https://github.com/open-telemetry/otel-arrow/pull/1199#discussion_r2412289925

utpilla:
@daviddahl I have made a few more suggestions. The important ones are around:
- Not making any unrelated changes in this PR. Please check the diff to avoid that.
- Metric naming suggestions. I think we can use better names to ease the diagnosability of these metrics.

The PR would look good after these changes.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
lquerel: @daviddahl could you add back those lines. Thanks

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
lquerel: I agree.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
lquerel: I agree

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
lquerel: I agree

lquerel:
@daviddahl Please take into account the last suggestions and then we can merge. Thanks

daviddahl:
I gave up on rebasing this PR, started over here: #1309

Issue #986, opened 2025-08-21
Add `unsafe-optimizations` feature


lquerel:
Use `unsafe-optimizations` as the opt‑in "enable all" feature, keep specific unsafe knobs granular and well‑documented (`unchecked-index`, `intrinsics`, `simd-*`, ...), and make the safe implementation the default.

Examples of granular unsafe optimization features:
- `unchecked-index`: Replaces bounds-checked indexing with get_unchecked, requiring the caller to ensure indices are always valid.
- `unchecked-arithmetic`: Removes overflow checks from integer operations, requiring the caller to guarantee no overflow occurs.
- `intrinsics`: Uses low-level `core::intrinsics` for operations, assuming caller guarantees input invariants (e.g. alignment, non-null pointers).

PR #1203, opened 2025-09-30
[perf] optimize allocations and UTF-8 decoding of strings for otlp_bytes -> views -> OTAP

### LINKED ISSUE #751: Optimize allocations and UTF-8 decoding of strings for otlp_bytes -> views -> OTAP
The deserialization process from `otlp_bytes` to OTAP makes extensive use of the `std::String::from_utf8()` method, which checks the validity of a string in UTF8 format and allocates memory for the validated string. This validation and allocation process is relatively costly, and the number of strings in OTLP batches is generally high. It is also important to note that there is a lot of redundancy in the deserialized strings (for example, attribute names are often very recurrent).

Ideally, we would like to validate strings with the same content only once per batch. Additionally, the simdutf8 crate is an accelerated implementation of this validation process, which can be applied to a single string or a collection of strings.

To optimize the number of UTF8 validations, we can rely on the following observation: a BinaryArray initialized with the binary version of a UTF8 string collection is structurally equal to a StringArray initialized with the same strings. The optimization then consists, at the construction level, of replacing all our StringArrays using dictionary encoding with a BinaryArray using dictionary encoding. This BinaryArray will be populated by the byte slices read from the otlp\_bytes slices. At the end of the process, we have a BinaryArray deduplicating all the strings (not yet validated as UTF8). From there, we can validate only the deduplicated entries of the BinaryArray (which should, on average, massively reduce the number of UTF8 validations), and if everything is valid, we can transform the BinaryArray into a StringArray without having to recreate the data (zero cost). It is even possible to consider using a special mode of simdutf8 to validate all binary entries representing strings in the dictionary in a single call.

albertlockett:
Closes #751 

Use BinaryArray builders where possible to lazily validate the UTF-8 Decoding when converting from OTLP Bytes -> OTAP. This is a nice optimization because we don't validate strings in the views, multiple times when inserting them in dictionary values. Instead, the UTF-8 Validation happens when we convert the Binary array to a String array.

~5.5% perf improvement encoding OTLP bytes -> OTAP
```
encode_otap_logs_using_views/proto_bytes->views->OTAP/resources=1,scopes=1,logs=1000
                        time:   [542.11 µs 542.41 µs 542.73 µs]
                        change: [−5.5094% −5.4312% −5.3581%] (p = 0.00 < 0.05)

encode_otap_logs_using_views/proto_bytes->views->OTAP/resources=1,scopes=1,logs=5000
                        time:   [2.6414 ms 2.6429 ms 2.6443 ms]
                        change: [−5.4051% −5.3303% −5.2524%] (p = 0.00 < 0.05)

encode_otap_logs_using_views/proto_bytes->views->OTAP/resources=1,scopes=1,logs=10000
                        time:   [5.2606 ms 5.2647 ms 5.2691 ms]
                        change: [−5.1650% −5.0651% −4.9688%] (p = 0.00 < 0.05)
```

Some improvement building OTAP from syslog as well
```
arrow_batch_creation/rfc3164_arrow_batch_100_msgs
                        time:   [73.360 µs 73.489 µs 73.627 µs]
                        change: [−2.0723% −1.7579% −1.4337%] (p = 0.00 < 0.05)
                        Performance has improved.

arrow_batch_creation/rfc5424_arrow_batch_100_msgs
                        time:   [42.795 µs 42.850 µs 42.904 µs]
                        change: [−0.1449% +0.0844% +0.2942%] (p = 0.45 > 0.05)
                        No change in performance detected.

arrow_batch_creation/cef_arrow_batch_100_msgs
                        time:   [35.831 µs 35.957 µs 36.093 µs]
                        change: [−5.6171% −5.2694% −4.9296%] (p = 0.00 < 0.05)
                        Performance has improved.
```


rust/otel-arrow-rust/src/encode/record/array.rs
```diff
@@ -681,11 +620,79 @@ pub type TimestampNanosecondArrayBuilder = PrimitiveArrayBuilder<TimestampNanose
...
+pub fn binary_to_utf8_array(src: &ArrayRef) -> Result<ArrayRef, ArrowError> {
+    let src_data_type = src.data_type();
+
+    if *src_data_type == DataType::Binary {
+        // safety: we've just checked the data type, so expecting on this cast is safe
+        let binary_arr = src
+            .as_any()
+            .downcast_ref::<BinaryArray>()
+            .expect("is binary array");
+        return StringArray::try_from_binary(binary_arr.clone())
```
lquerel, line 703: Do you know if StringArray::try_from_binary using simdutf8 behind the scene?

rust/otap-dataflow/crates/otap/src/encoder/cbor.rs
```diff
@@ -49,8 +49,10 @@ where
 {
     match source.value_type() {
         ValueType::String => {
-            let s = source.as_string().expect("expected string");
-            serializer.serialize_str(s)
+            let s_bytes = source.as_string().expect("expected string");
+            let s_str = str::from_utf8(s_bytes)
```
lquerel: 2 questions:
- do we need to validate utf8 here?
- did you try to use simdutf8 instead of str::from_utf8?

lquerel (approved):
LGTM. Nice perf improvements.
I left 2-3 comments for suggestions/questions

rust/otel-arrow-rust/src/encode/record/array.rs
albertlockett: Not at this time. It's using `std::str::from_utf8`:
https://github.com/apache/arrow-rs/blob/570af8a1b2af45bf80ea6468396316affe58bc91/arrow-array/src/types.rs#L1605-L1609

There has been some discussion about changing to use simdutf8 globally, so maybe this will change in the future
https://github.com/apache/arrow-rs/issues/7014#issuecomment-2613251782

lquerel: I will curious to see the perf improvement if it's easy to check with a patch.

rust/otap-dataflow/crates/otap/src/encoder/cbor.rs
albertlockett: > do we need to validate utf8 here?

It's probably safest if we validate, otherwise we could end up with invalid utf8 in our cbor serialized values.

Were you thinking we'd maybe try to do the validation somewhere else?


> did you try to use simdutf8 instead of str::from_utf8?
Using simdutf8 does give us an extra performance boost. Added in 94b1af51 

```
encode_otap_logs_using_views/proto_bytes->views->OTAP/resources=1,scopes=1,logs=1000
                        time:   [526.31 µs 526.84 µs 527.38 µs]
                        change: [−2.9872% −2.8702% −2.7468%] (p = 0.00 < 0.05)
                        Performance has improved.

encode_otap_logs_using_views/proto_bytes->views->OTAP/resources=1,scopes=1,logs=5000
                        time:   [2.5542 ms 2.5563 ms 2.5583 ms]
                        change: [−3.3857% −3.2760% −3.1800%] (p = 0.00 < 0.05)
                        Performance has improved.

encode_otap_logs_using_views/proto_bytes->views->OTAP/resources=1,scopes=1,logs=10000
                        time:   [5.0761 ms 5.0801 ms 5.0840 ms]
                        change: [−3.6149% −3.5066% −3.4000%] (p = 0.00 < 0.05)
                        Performance has improved.
```

rust/otel-arrow-rust/src/encode/record/array.rs
albertlockett: It doesn't seem to make a huge difference, at least in our current otap_encoding benchmarks.

I tried patching it in like this, using unsafe to reconstruct the StringArray after using simdutf8 to validate https://github.com/open-telemetry/otel-arrow/commit/544a23396bf2da820ceedbb52e380b7428a669f9
That didn't really make any difference in performance

But it seems like we don't really spent much time validating utf8, which might be a function of our test data. See here, even in this function (which is a very small fraction of the overall profile), we only spent a small amount of time validating utf8

image

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/parser/parsed_message.rs
```diff
@@ -9,30 +9,30 @@ use otel_arrow_rust::encode::record::attributes::AttributesRecordBatchBuilder;
 use std::borrow::Cow;
 
 // Common attribute key constants for both RFC5424 and RFC3164 messages
-const SYSLOG_FACILITY: &str = "syslog.facility";
-const SYSLOG_SEVERITY: &str = "syslog.severity";
-const SYSLOG_HOST_NAME: &str = "syslog.host_name";
-const SYSLOG_MESSAGE: &str = "syslog.message";
+const SYSLOG_FACILITY: &[u8] = b"syslog.facility";
```
utpilla: These were const str slices which were already UTF-8 validated. I might be wrong here, but it seems like by updating them to bytes slices we would end up having to UTF-8 validate them each time we build the attributes record batch.

For these kinds of cases, it might be better to offer another `AttributesRecordBatchBuilder` struct which has `keys` type defined as `StringArrayBuilder` instead of `BinaryArrayBuilder`.

rust/otel-arrow-rust/src/encode/record/array.rs
lquerel: Thanks for testing. Better to stay with the official version.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/parser/parsed_message.rs
albertlockett: You're right @utpilla , thanks for calling this out. The `key` = `StringArrayBuilder` variant of `AttributesRecordBatchBuilder` was the solution that jumped to mind of me for this as well. That should be straight forward to add

albertlockett: @utpilla made this change. 

As part of this work, I ended up refactoring `AttributesRecordBatchBuilder` to extract out the builders for the attribute value columns into a new type. This is nice because we can reuse this in `otel_arrow_rust::encode::record::logs::LogsBodyBuilder` as well.

rust/otap-dataflow/crates/otap/src/encoder/cbor.rs
```diff
@@ -49,8 +49,10 @@ where
 {
     match source.value_type() {
         ValueType::String => {
-            let s = source.as_string().expect("expected string");
-            serializer.serialize_str(s)
+            let s_bytes = source.as_string().expect("expected string");
```
jmacd, line 52: Tangent: how is it that we call as_string() and get s_bytes?

context for my question:
https://github.com/open-telemetry/oteps/pull/257

The question being asked here, and in this evergreen issue https://github.com/open-telemetry/opentelemetry-specification/issues/3421, is whether users have a legitimate option and/or methods of recourse for invalid utf8.

**Emphasis** the Golang gRPC collector interface is not validating UTF-8 for performance reasons and it will now be a breaking change to fix. As a vendor I've had to introduce a processor for this issue, hence the issue report. I also had spicy conversations with my former company's Rust team about the philosophical merits. I forget how Tonic acts, in this regard.

But, one easy way to side-step all of the above is to let strings be bytes and pass them to the user, who then faces the question themselves.

jmacd (approved):

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/parser/parsed_message.rs
```diff
@@ -121,41 +121,49 @@ impl ParsedSyslogMessage<'_> {
...
+                        .append_str(hostname);
                     attributes_count += 1;
                 }
 
                 if let Some(appname) = msg.app_name {
                     log_attributes_arrow_records.append_key(SYSLOG_APP_NAME);
                     log_attributes_arrow_records
-                        .append_str(std::str::from_utf8(appname).unwrap_or_default());
+                        .any_values_builder
+                        .append_str(appname);
```
utpilla, line 160: What happens in the builder's `finish()` method logic if a certain `&[u8]` in the binary array is not valid UTF-8? Does the value of the row for that field become null or default? 

Or does the entire binary array to string array conversion fail?

rust/otap-dataflow/crates/otap/src/encoder/cbor.rs
albertlockett: > Tangent: how is it that we call as_string() and get s_bytes?

wrt to `as_string` here .. `string` is meant to represent the field on the `AnyValue` moreso than the return type of the medho. e.g. the idea behind the trait method we're calling (`source`, which is an instance of `AnyValueView` trait) is to get the value of the `AnyValue` as if it were a type of string -- so from the proto perspective, we're getting the bytes representation of the `string_value` field:
https://github.com/open-telemetry/opentelemetry-proto/blob/b8054e9abdba27c8c9b51dcb3f1d1884906d8572/opentelemetry/proto/common/v1/common.proto#L28-L32

We could rename the trait method to something like `string_value` to make this more clear?

> I forget how Tonic acts, in this regard.

wrt to how tonic acts -- tonic doesn't do any utf8 validation itself, but prost will do it internally when we call `prost::Message::decode` (which we're currently avoiding by converting directly between proto bytes & OTAP)
https://github.com/tokio-rs/prost/blob/812358c9ac93aae198caae08fdc98cc288158695/prost/src/encoding.rs#L596

It's true that we _could_ avoid validating UTF-8 at all, it is possible, but we'd probably need to use some unsafe code to make it work with the libraries we're using. For example, in this location, the `Serializer` trait expects a `str` & doesn't give us access to the underlying buffer, which we'd need to write the cbor tag = 3 for type string followed by the bytes.. e.g. we want to do this: 
https://github.com/pyfisch/cbor/blob/347a3f096d5a1e8e380b9968e74838c9037a5b14/src/ser.rs#L353-L358
but we can't, so we'd need to do an unsafe conversion from `&[u8]` to `str` to get the correct args for the method.

Similarly, we'd probably want to keep the arrow array types for our string arrays `Utf8`, but StringArray always validates that the values buffer is `utf-8` unless you use the unsafe method `new_unchecked`.
https://docs.rs/arrow/latest/arrow/array/type.StringArray.html#method.new_unchecked

I'm not against unsafe code personally, but there has been some discussions around this and I think where it landed was that we should have a feature flag for the use of unsafe in the codebase https://github.com/open-telemetry/otel-arrow/issues/986. I guess one consideration is, does it make sense to have this `unsafe` feature also control utf-8 validation (_if_ we choose not the validate at all)?

WRT to whether or not we _should_ validate at all.. I'd like to hear more about the philosophical merits ;)

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/parser/parsed_message.rs
albertlockett: It will fail with an error. Added unit test in [3cc53f2](https://github.com/open-telemetry/otel-arrow/pull/1203/commits/3cc53f261cfcaf04a6876e71fa2262455f6b388c) showing the behaviour

albertlockett: If the desired behaviour here is to silently insert "" for invalid utf-8 sequences here, maybe we could implement another version of the BinaryArray -> StringArray conversion that does this in a followup PR?

utpilla: Okay. We definitely want to avoid a single malformed message with invalid UTF-8 in any field could to cause the entire batch to fail.

We can implement the mitigation in a follow-up PR. We should probably choose to do lossy UTF-8 conversion using `String::from_utf8_lossy()` to replace invalid UTF-8 sequences with the replacement character (�) instead of using an empty string. We should try to retain as much information as we can.

albertlockett: Sounds good. Our of curiosity, do you have any sense for how often we might receive invalid utf-8 sequences from syslog?

I wonder if we could do something like validate utf-8 for the entire values buffer (which arrow is currently doing) as the fast path, and if that fails, we'll do something following this as the slow path when we get an invalid sequence:
```rs
let string_arr_builder = StringBuilder::new(); // or w/ capacity ...
for bytes in binary_arr {
  string_arr_builder.append_value(from_utf8_lossy(bytes))
}
// ...
```

utpilla: > Our of curiosity, do you have any sense for how often we might receive invalid utf-8 sequences from syslog?

Ideally, we shouldn't receive any invalid UTF-8 sequences. There might be a few cases where some device is erroneously sending invalid UTF-8 sequences but my main motivation to fix this is to avoid security issues. From a security standpoint, this issue would probably apply to OTLP data as well. We don't want any bad actor to exploit this vulnerability and cause denial of service.

> I wonder if we could do something like validate utf-8 for the entire values buffer (which arrow is currently doing) as the fast path, and if that fails, we'll do something following this as the slow path when we get an invalid sequence:

I don't know the details of `SringArrayBuilder` creation which might influence our decision here, but I think we could simply call `String::from_utf8_lossy()` and be done. The method works for both invalid and valid UTF-8 sequences. It returns a `Cow<'a, str>`. For valid UTF-8 sequences, there is no allocation, so it returns a borrowed string. For invalid UTF-8 sequences, it has to replace the characters so it would return an owned string.

albertlockett: sounds good, yeah that can work. thanks @utpilla

Issue #1129, opened 2025-09-13
[otap-dataflow] Arrow IPC compression misbehaving


clhain:
Related to #1085. cc @albertlockett 

I think the additional arrow compression stuff may not be behaving quite as expected. I was running the demo earlier this week on an image prior to this change. Testing today from main I noticed the cpu use went through the roof on any arrow exporter, and I think it's due to the default enabled zstd compression introduced in 1085. See screenshot below.

The lower cpu use is with what was previously the default behavior (payload_compression: none). On the right, everything is the same except removing that statement.

<img width="1213" height="469" alt="Image" src="https://github.com/user-attachments/assets/95489f6d-6e6f-4f98-b566-99d27ee1e966" />

albertlockett:
@clhain thanks for the heads up. I'll have a look this week

clhain:
If it helps, I just added https://github.com/open-telemetry/otel-arrow/pull/1140 with the configs I'm using to validate this (running as-is will be in the 'misbehaving' state). 

You can comment out the first test here: https://github.com/open-telemetry/otel-arrow/pull/1140/files#diff-d5d625e8bd93fe7bffb1476b7eda652d0da70d95c9d950d66242c9ac85617200R74 and then run that file and it'll start with one of the broken otap scenarios. In the output you'd be looking for lower values in the df_engine cpu rows of this table (expecting around .2 to .3 on this particular system instead of ~.9):

## Table: component_resource_metrics
  
**Description**: CPU and Memory metrics for the components in the Test Suite during the test.
  
| component_name   | metric_name            |   min_value |   avg_value |   max_value |
|:-----------------|:-----------------------|------------:|------------:|------------:|
| load-generator   | container.memory.usage | 1.43e+08    | 2.6461e+08  | 5.19098e+08 |
| load-generator   | container.cpu.usage    | 0.168579    | 0.177284    | 0.196819    |
| df-engine        | container.cpu.usage    | 0.878826    | 0.909728    | 0.926941    |
| df-engine        | container.memory.usage | 3.77979e+07 | 3.84094e+07 | 4.02473e+07 |
| backend-service  | container.cpu.usage    | 0.0794287   | 0.0910984   | 0.116861    |
| backend-service  | container.memory.usage | 2.51453e+07 | 2.51994e+07 | 2.5346e+07  |

albertlockett:
@clhain have made some progress on this.

[...]

PR #1204, opened 2025-09-30
[PerfTest] temporarily remove arrow payload compression in benchmark suites

clhain:
Until the Arrow-rs patch mentioned in https://github.com/open-telemetry/otel-arrow/issues/1129 is merged and in-use here, we need to disable the payload compression in the perf test benchmarks.


albertlockett (approved):
:+1: good call

PR #1207, opened 2025-10-01
Consolidate errors in the observed state store

lquerel:
Consolidate errors in the observed state store.


jmacd:
Is this related to https://github.com/open-telemetry/otel-arrow/pull/1187?

That is: (a) errors in configuration do not print, (b) errors in configuration do not result in failure.

I couldn't figure out how to fix both without breaking the function of CTRL-C to shutdown.

rust/otap-dataflow/configs/fake-otap.yaml
```diff
@@ -0,0 +1,26 @@
+settings:
```
lquerel, line 1: Not really related to this PR, but this config was missing for my manual tests.

lquerel:
@jmacd 

> #1187

Errors will now be reported both via the status endpoint and on the console (see GH issues created to export those error events via OTLP or OTAP).

Configuration errors are not also reported and change the pipeline status.

I think we'll need another iteration to finalize everything, because in principle we must be able to dynamically support both multiple pipeline configurations and updates to those configurations, which complicates the system shutdown logic.

rust/otap-dataflow/crates/admin/src/pipeline.rs
```diff
@@ -39,3 +52,44 @@ pub async fn show_status(
...
+/// Used by the kubelet livenessProbe to decide whether to restart the container.
+///
+/// Typical use cases:
+/// - Detect deadlocks or stuck event loops.
+/// - Force a restart when the app can't recover by itself.
+/// - Should be cheap and internal (not dependent on external systems).
+///
+/// ToDo Implement heartbeat checks.
+async fn liveness(
+    Path((pipeline_group_id, pipeline_id)): Path<(String, String)>,
```
jmacd, line 65: Didn't realize parameters could be destructured 😂

rust/otap-dataflow/crates/admin/src/pipeline.rs
```diff
@@ -39,3 +52,44 @@ pub async fn show_status(
     let pipeline_status = state.observed_state_store.pipeline_status(&key);
     Ok(Json(pipeline_status))
 }
+
+/// Used by the kubelet livenessProbe to decide whether to restart the container.
```
albertlockett, line 56: Is mentioning that this is specifically for kubelet might be too implementation specific as a description of the endpoints? E.g. other services could make use of these as well. If someone were to run this on ECS, or behind an LB that does active health checking for example.

Also, how is this intended to be used by kube? Will the `pipeline_group_id` and `pipeline_id` be known a-priori? and isn't there multiple pipelines per instance of the application? kube only allows us to define one liveness probe as far as I know. I guess we'd configure it like this?
```yaml
# pod spec:
spec:
  containers:
  - name: otap-dataflow
    # .....
    livnessProbe:
      httpGet:
        path: /pipeline-groups/known_pipeline_group_id/pipelines/known_pipeline_id/livez
        port: liveness-port
```

rust/otap-dataflow/crates/state/src/pipeline_rt_status.rs
```diff
@@ -0,0 +1,411 @@
...
+
+    // Centralizes entering Deleting and resets flags.
+    fn go_to_deleting(&mut self, mode: DeletionMode) -> ApplyOutcome {
+        self.delete_pending = false;
+        self.goto(PipelinePhase::Deleting(mode))
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
albertlockett, line 305: good test suite!

rust/otap-dataflow/crates/state/src/pipeline_status.rs
```diff
@@ -0,0 +1,365 @@
...
+        let top_fail = count
+            .failed_reasons
+            .iter()
+            .max_by_key(|(_, n)| *n)
+            .map(|(r, _)| *r);
+        let top_reject = count
+            .rejected_reasons
+            .iter()
+            .max_by_key(|(_, n)| *n)
+            .map(|(r, _)| *r);
```
albertlockett, line 98: nit: it might be a slightly better performance to compute these where we need them? e.g. we only need `top_fail` in the block below near 112 where we're setting `self.phase = PipelineAggPhase::Failed`

rust/otap-dataflow/crates/state/src/store.rs
```diff
@@ -144,80 +101,47 @@ impl ObservedStateStore {
...
+            EventType::Request(_) | EventType::Error(_) => {
+                eprintln!("Observed event: {observed_event:?}")
             }
+            EventType::Success(_) => { /* no console output for success events */ }
         }
+
+        let mut pipelines = self.pipelines.lock().unwrap_or_else(|poisoned| {
+            log::warn!(
+                "ObservedStateStore mutex was poisoned; continuing with possibly inconsistent state"
+            );
```
albertlockett, line 122: I see we're using `log` in a few places in this PR. Should we add a tracing subscriber like `env_logger` in  `main.rs`?

jmacd (approved):

rust/otap-dataflow/crates/state/src/store.rs
jmacd: Yes, please. I see this is linked with #1237.

jmacd:
```
    #[allow(
        clippy::print_stderr,
        reason = "Use `eprintln!` while waiting for https://github.com/open-telemetry/otel-arrow/issues/1237."
    )]
```

Here, I mean.

Issue #1096, opened 2025-09-09
[otap-dataflow] pipelines with OTAP & OTLP Exporter fail to run if initial client connection fails


albertlockett:
If I start a pipeline like this:
```                            
 ┌──────────────────────────────────┐                      
 │ pipeline                         │                      
 │                                  │                      
 │ ┌────────────┐     ┌───────────┐ │                      
 │ │  upstream  ┼────►│OTLP/OTAP  ┼─┼───────► gRPC Service 
 │ │  component │     │exporter   │ │                      
 │ └────────────┘     └───────────┘ │                      
 └──────────────────────────────────┘
```

And for some reason we cannot initiate the connection to the remote gRPC service when the the pipeline is initialized (e.g. when the `Exporter` implementation's `start` method is called), the start method will exit and the pipeline will never run.

https://github.com/open-telemetry/otel-arrow/blob/c03a3ea488414d4f432a047f4084d5208775dce0/rust/otap-dataflow/crates/otap/src/otap_exporter.rs#L127-L133

This is especially noticeable when testing a chain of collectors/df-engines -- if they're not all started in the correct sequence, the df-engine with the exporter will never produce any telemetry. It will be important to fix this for deployment in a k8s environment as well, where we shouldn't try to rely on guarantees around the order that various pods start up.

We should probably have some kind of behaviour to handle this. One good option would be to retry the initial client connection periodically until it can be established.

PR #1211, opened 2025-10-01
handle grpc server not ready in OTLP exporter

albertlockett:
part of #1096 

This change fixes an issue where if the grpc server could not be reached when the pipeline started, the OTLP exporter's start method would exit early with no feedback. Now, we wait to try the connection until we need it. This change adds test to ensure the exporter behaves as expected in the case of the server not being ready when the pipeline starts and if the server restarts (e.g. the connection is closed).

If this is OK as a solution, I'll add the same solution to OTAP exporter as a followup

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
+++ b/rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
@@ -506,6 +506,8 @@ mod tests {
         let (server_start_ack_sender, server_start_ack_receiver) = tokio::sync::mpsc::channel(1);
         let (shutdown_sender1, shutdown_signal1) = tokio::sync::oneshot::channel();
         let (shutdown_sender2, shutdown_signal2) = tokio::sync::oneshot::channel();
+        let (server_shutdown_ack_sender, server_shutdown_ack_receiver) =
+            tokio::sync::mpsc::channel(1);
         let (req_sender, req_receiver) = tokio::sync::mpsc::channel(32);
 
         async fn start_exporter(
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
```diff
@@ -105,25 +106,17 @@ impl Exporter<OtapPdata> for OTLPExporter {
             .start_periodic_telemetry(Duration::from_secs(1))
             .await?;
 
+        let channel = Channel::from_shared(self.config.grpc_endpoint.clone())
+            .map_err(|e| Error::ExporterError {
+                exporter: exporter_id.clone(),
+                error: format!("grpc channel error {e}"),
+            })?
+            .connect_lazy();
```
utpilla, line 119: This definitely looks better but I was wondering if this would solve the problem completely. The current export code would still break out of the loop and return from method when it encounters an error.

```rust
 _ = logs_client.export(bytes).await.map_err(|e| {
    self.pdata_metrics.logs_failed.inc();
    Error::ExporterError {
        exporter: exporter_id.clone(),
        error: e.to_string(),
    }
})?;
```

So even if we don't return from the method initially if the server is not available, we might still return from the method later on if there is an error during the actual usage of the channel.

albertlockett: Thanks @utpilla! Originally I was planning on handling the failed requests in a followup, but after thinking about it more, I agree it makes sense to fix this all in one PR. I added some changes to gracefully handle the failed requests. It just updating the failed metric for now. Eventually we'll probably have some better behaviour like nacking the message and/or logging the error.

PR #1217, opened 2025-10-02
[perf][otel-arrow-rust] Optimize `produce_bar` method used in OTAP Exporter

utpilla:
## Changes
- Remove `HashMap` overhead- no need of repeated `HashMap` lookups and retain operations
- Since the 28 `ArrowPayloadTypes` are known-beforehand, we can use a fixed array of 28 entries to track the `StreamProducer`s

### DIFF
--- a/rust/otel-arrow-rust/src/encode/producer.rs
+++ b/rust/otel-arrow-rust/src/encode/producer.rs
@@ -5,7 +5,7 @@
 //!
 //! `BatchArrowRecords` is the protobuf type that contains the Arrow IPC serialized messages.
 
-use std::{collections::HashMap, io::Cursor};
+use std::io::Cursor;
 
 use arrow::array::RecordBatch;
 use arrow::datatypes::SchemaRef;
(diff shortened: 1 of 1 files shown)


rust/otel-arrow-rust/src/encode/producer.rs
```diff
@@ -57,11 +54,61 @@ impl StreamProducer {
...
+    schema_id: String,
+    producer: StreamProducer,
+}
+
 /// Produces OTAP `BatchArrowRecords` from OTAP Batches
 pub struct Producer {
     next_batch_id: i64,
     next_schema_id: i64,
-    stream_producers: HashMap<String, StreamProducer>,
+    stream_producers: [Option<ProducerEntry>; PAYLOAD_TYPE_COUNT],
```
utpilla, line 113: The perf benefit of having `O(1)` access does come with increased memory usage. For reference, the size of this array is around `5.4KB`. We could reduce it further by having dedicated `Producers` for each signal. Currently, we use the same producer for all signals which forces us to keep 28 items in the array.

PR #1219, opened 2025-10-02
Parquet exporter: add internal telemetry

daviddahl:
- Add pdata counters (consumed/exported/failed) and IO counters (files created/closed, rows written, flush reasons, requeues, age checks).
- Start a 1s telemetry timer that triggers non-blocking metrics collection via CollectTelemetry
- Return WriteStats from writer and plumb into exporter to record metrics
- Add tests for telemetry timer and metrics reporting

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/otap_exporter.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_exporter.rs
@@ -114,34 +114,22 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
             ))
             .await;
 
+        let exporter_id = effect_handler.exporter_id();
+        let channel = Channel::from_shared(self.config.grpc_endpoint.clone())
+            .map_err(|e| Error::ExporterError {
+                exporter: exporter_id,
+                error: format!("grpc channel error {e}"),
+            })?
+            .connect_lazy();
+
         let timer_cancel_handle = effect_handler
             .start_periodic_telemetry(Duration::from_secs(1))
             .await?;
 
         // start a grpc client and connect to the server
-        let mut arrow_metrics_client =
-            ArrowMetricsServiceClient::connect(self.config.grpc_endpoint.clone())
-                .await
-                .map_err(|error| Error::ExporterError {
-                    exporter: effect_handler.exporter_id(),
-                    error: error.to_string(),
-                })?;
-
-        let mut arrow_logs_client =
-            ArrowLogsServiceClient::connect(self.config.grpc_endpoint.clone())
-                .await
-                .map_err(|error| Error::ExporterError {
-                    exporter: effect_handler.exporter_id(),
-                    error: error.to_string(),
-                })?;
-
-        let mut arrow_traces_client =
-            ArrowTracesServiceClient::connect(self.config.grpc_endpoint.clone())
-                .await
-                .map_err(|error| Error::ExporterError {
-                    exporter: effect_handler.exporter_id(),
-                    error: error.to_string(),
-                })?;
+        let mut arrow_metrics_client = ArrowMetricsServiceClient::new(channel.clone());
+        let mut arrow_logs_client = ArrowLogsServiceClient::new(channel.clone());
+        let mut arrow_traces_client = ArrowTracesServiceClient::new(channel.clone());
 
         if let Some(ref compression) = self.config.compression_method {
             let encoding = compression.map_to_compression_encoding();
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/otap/src/parquet_exporter/metrics.rs
```diff
@@ -0,0 +1,41 @@
...
+    #[metric(unit = "{file}")]
+    pub flush_scheduled_max_age: Counter<u64>,
+
+    /// Files whose flush was requeued because child payloads still had unflushed rows.
+    #[metric(unit = "{file}")]
+    pub flush_requeued_due_to_children: Counter<u64>,
+
+    /// Number of periodic age checks executed (timer ticks).
+    #[metric(unit = "{check}")]
+    pub flush_age_checks: Counter<u64>,
```
albertlockett: Most of the metrics look reasonable!

I'm not sure `flush_requeued_due_to_children` and `flush_age_checks` are real interesting stats. What would an end user do with this information? `flush_requeued_due_to_children` seems like more of an implementation detail, and `flush_age_checks` would just be uptime / `config.flush_age_check_interval`, right?

rust/otap-dataflow/crates/otap/src/parquet_exporter/writer.rs
```diff
@@ -76,47 +99,58 @@ impl WriterManager {
...
+        let rows_written_per_writer = self
             .curr_writer_for_prefix
             .values_mut()
             .map(|fw| fw.write_scheduled())
             .collect::<FuturesUnordered<_>>()
             .try_collect::<Vec<_>>()
             .await?;
 
-        Ok(())
+        Ok(rows_written_per_writer)
```
albertlockett: I think one thing that might be nice here would be to return the total rows written instead of a vec of rows written per writer.

The reason being that, it looks like we only ever use the result of this call to sum up the rows written. We could avoid a heap allocated vector by computing the size instead by doing something like:

```suggestion
   async fn write_scheduled(&mut self) -> Result<usize, ParquetError> {
        let total_rows_written = self
            .curr_writer_for_prefix
            .values_mut()
            .map(|fw| fw.write_scheduled())
            .collect::<FuturesUnordered<_>>()
            .try_fold(0usize, |acc, n| async move {
                Ok(acc + n)
            })
            .await?;

      Ok(total_rows_written)

File: rust/otap-dataflow/crates/otap/src/parquet_exporter/writer.rs [hunk 3]
```diff
@@ -76,47 +99,58 @@ impl WriterManager {
...
     }
 
     fn schedule_write_batch(
```
albertlockett, line 162: nit: we could maybe add a comment to this about what the return type of bool represents. Something like:

```suggestion
    /// Schedules the record batch for a given signal to be written. If there's not already a file writer for
    /// the given partition/payload type, a new one will be created
    ///
    /// Returns boolean true/false indicating whether a new file writer was created
    fn schedule_write_batch(
```

rust/otap-dataflow/crates/otap/src/parquet_exporter/writer.rs
```diff
@@ -427,7 +504,7 @@ mod test {
         let otap_batch = to_logs_record_batch(create_simple_logs_arrow_record_batches(
             SimpleDataGenOptions::default(),
         ));
-        writer
+        let _ = writer
```
albertlockett: There's a handful of places here in the `test` module we ignore the returned writer stats. Do you think it'd be worthwhile to assert on the stats here?

rust/otap-dataflow/crates/otap/src/parquet_exporter.rs
```diff
@@ -167,6 +224,8 @@ impl Exporter<OtapPdata> for ParquetExporter {
                     let mut timeout = Delay::new(deadline).fuse();
                     let flush_all = writer.flush_all().fuse();
                     pin_mut!(flush_all);
+                    // Stop telemetry loop
+                    let _ = telemetry_cancel_handle.cancel().await;
```
albertlockett: I'm not sure we actually need to wait for this to happen before we shut down, do we? I'd think we'd want to start trying to flush immediately, while racing against the shutdown deadline and let this telemetry_cancel also try to happen simultaneously. Maybe we should do something like this:

```suggestion
                    let cancel_fut = async {
                        _ = telemetry_cancel_handle.cancel().await;
                        futures::future::pending::<()>().await
                    }.fuse();
                    pin_mut!(cancel_fut);

                    return futures::select_biased! {
                        _ = cancel_fut => unreachable!(),
                        _timeout = timeout => Err(Error::IoError {
                                node: effect_handler.exporter_id(),
                                error: std::io::Error::from(ErrorKind::TimedOut)
                            }),
                        _ = flush_all => Ok(()),
                    };
```

rust/otap-dataflow/crates/otap/src/parquet_exporter/metrics.rs
daviddahl: no worries, if its noise, lets jettison it.

rust/otap-dataflow/crates/otap/src/parquet_exporter.rs
jmacd: Had to read this twice. Tricky. If this catches on, I will recommend a helper method in the telemetry crate to avoid writing this very many times.

albertlockett (approved):
Looks good to me @daviddahl , approved!

Thanks for addressing those few feedbacks!

rust/otap-dataflow/crates/otap/src/parquet_exporter/writer.rs
albertlockett: actually come to think of it, I guess there are tests in parquet_exporter covering the metrics, so that probably covers that we get the correct `WriterStats` result here as well.

daviddahl:
@albertlockett thanks for the great review!

PR #1221, opened 2025-10-02
[otel-arrow-rust] Micro optimizations and minor fixes to SchemaBuilder

utpilla:
## Changes
- Use `sort_unstable_by_key` instead of `sort_by` for higher efficiency
- Optimize iterator usage by using `extend` with a `Range` instead of `enumerate` and repeated `push` operations
- Truncate `sort_buf` when writing `Union` types
- Fix unit test


albertlockett (approved):
LGTM, thanks again @utpilla

Issue #498, opened 2025-05-30
[otap-dataflow] emit control message from pipeline components using effect handler


albertlockett:
We should add the capability to emit control messages using the various effect handlers.

There's a few important places where this would be used:
- register a timer interval with the pipeline engine https://github.com/open-telemetry/otel-arrow/issues/497
- send `Ack` / `Nack` control messages.
  - For example, `Nack` messages could be emitted for retryable errors which could be handled by the retry processor https://github.com/open-telemetry/otel-arrow/issues/368

PR #1225, opened 2025-10-03
Ack/Nack and EffectHandler integration

jmacd:
Part of #498 
Part of #1244

Adds a new testing component named "error_exporter", similar to "noop_exporter" but returns an error via NACK.

Engine:

- crate::Interests is bitflags of u8: ACKS, NACKS
- control::{AckMsg, NackMsg}
  - these return PData and caller's CallData
  - replace the existing placeholder, no more msg_id: u64
- control::{CallData, Context8u8}
  - caller's data is two 8-byte values
  - components implement actual type to/from CallData
- crate::ProducerEffectHandlerExtension for PData-specific effect handlers
  - subscribe_to(Interests, CallData, &mut PData) API
  - for receivers and processors
  - shared and local impls
- crate::ConsumerEffectHandlerExtension for PData-specific effect handlers
  - notify_ack(AckMsg), notify_nack(NackMsg) API
  - for processors and exporters
  - shared and local impls
- effect_handler::EffectHandlerCore prepare CallData and send for routing
  - route_ack, route_nack take PData-specific FnOnce
  - e.g., FnOnce(AckMsg<PData>) -> Option<(usize, Ack<PData>)>
  - with shared and local impls
  - placeholders: these return Err(_) b/c control messages do not exist yet

OTAP:
- noop_exporter responds with ACK
- new error_exporter component response with NACK
- Frame is { NodeId, Interests, CallData }
- Context is Vec<Frame>
  - subscribe_to() push to the context
  - next_ack, next_nack functions passed to EffectHandlerCore pop from the context
- Consumer/Producer effect handler extensions impls
  - PData-specific effect handler behavior
  - with shared and local impls
- retry_processor:
  - placeholder fixes, retains existing logic
  - implements CallData with msg_id: u64 to retain existing tests
  - future work will continue here, we can avoid internal state
  - removes this benchmark; we can add one later after the rewrite to use CallData.


rust/otap-dataflow/crates/engine/src/control.rs
```diff
@@ -9,12 +9,91 @@ use crate::error::{Error, TypedError};
...
+/// size is arbitrary, but shouldn't be larger than needed by
+/// callers. For example: retry count, sequence and generation
+/// numbers, etc.
+pub type CallData = SmallVec<[Context8u8; 2]>;
+
+/// The ACK message.
+#[derive(Debug, Clone)]
+pub struct AckMsg<PData> {
+    /// Accepted pdata being returned.
+    pub accepted: Box<PData>,
```
lquerel, line 64: Why do we need to keep the entire pdata for Acks? That means we keep the whole message in memory until the Ack has been routed to its final destination. Is that really desirable?

rust/otap-dataflow/crates/engine/src/effect_handler.rs
```diff
@@ -200,6 +201,48 @@ impl<PData> EffectHandlerCore<PData> {
             pipeline_ctrl_msg_sender,
         })
     }
+
+    /// Send a Ack to a node of known-interest.
+    pub async fn route_ack<F>(
+        &self,
+        ack_in: AckMsg<PData>,
+        cxf: F,
```
lquerel: Documenting the purpose of cxf will be nice for future users.

rust/otap-dataflow/crates/engine/src/lib.rs
```diff
@@ -157,6 +159,36 @@ where
     })
 }
 
+bitflags::bitflags! {
+/// An 8-bit flags struct intended to store various intents describing
+/// callers in in a pipeline, e.g., detail about whether Ack and/or
```
lquerel:
```suggestion
/// callers in a pipeline, e.g., detail about whether Ack and/or
```

rust/otap-dataflow/crates/otap/src/error_exporter.rs
```diff
@@ -0,0 +1,112 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
```
lquerel, line 3: A module-level documentation is missing here.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -92,7 +98,67 @@ use crate::encoder::{encode_logs_otap_batch, encode_spans_otap_batch};
...
+    }
+
+    /// Consume frames to locate the most recent subscriber with ACKS.
+    #[must_use]
+    pub fn next_ack(mut ack: AckMsg<OtapPdata>) -> Option<(usize, AckMsg<OtapPdata>)> {
+        ack.accepted
+            .context
+            .next_with_interest(Interests::ACKS)
+            .map(|frame| {
+                ack.calldata = Some(frame.calldata);
```
lquerel: Based on the code, we can't have a frame without calldata so can we have a ack/nack message without calldata?

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -355,7 +386,9 @@ impl RetryProcessor {
         }
     }
 
-    fn acknowledge(&mut self, id: u64) -> bool {
+    fn acknowledge(&mut self, ack: AckMsg<OtapPdata>) -> bool {
+        let rstate: RetryState = ack.calldata.expect("has context").into();
```
lquerel: See my previous comment. Looks like we expect a `calldata` if we don't have one `expect` will panic which is not ideal. Except if we have good ready, I will suggest to remove the `Option<_>` from the `calldata` field.

lquerel (changes requested):
My only main concern is the presence of PData in the `Ack` message. Maybe there is a good reason to have it for `Ack` but I didn't see it.

Otherwise LGTM.

rust/otap-dataflow/crates/engine/src/control.rs
```diff
@@ -9,12 +9,91 @@ use crate::error::{Error, TypedError};
...
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug)]
+pub struct Context8u8([u8; 8]);
+
+impl<T: Pod> From<T> for Context8u8 {
+    /// From T to Context8u8
+    fn from(v: T) -> Self {
+        Self(bytemuck::cast(v))
+    }
+}
```
albertlockett, line 34: This can panic if `T` isn't 8 bytes long https://docs.rs/bytemuck/1.24.0/bytemuck/fn.cast.html

> This is like [try_cast](https://docs.rs/bytemuck/1.24.0/bytemuck/fn.try_cast.html), but will panic on a size mismatch.

For example:
```rs
        let x = 5u64;
        let y = Context8u8::from(x); // <-- this is OK
        println!("{:?}", y);

        let x2 = 5u16;
        let y2 = Context8u8::from(x2); // <-- this will panic
        println!("{:?}", y2);
```

Maybe we could add a compile time assertion that the length of `T` is correct?
```suggestion
impl<T: Pod> From<T> for Context8u8 {
    /// From T to Context8u8
    fn from(v: T) -> Self {
        const {
            assert!(size_of::<T>() == 8);
        }
        Self(bytemuck::cast(v))
    }
}
```

rust/otap-dataflow/crates/otap/src/error_exporter.rs
albertlockett: Looks like it's documented in lib.rs -- 

```rs
/// An error-exporter returns a static error.
pub mod error_exporter;
```

should we prefer the docs in the inner module instead?
```suggestion
//! An error-exporter returns a static error.
```

rust/otap-dataflow/crates/otap/src/pdata.rs
albertlockett: I think we can have Ack w/out calldata briefly.

If I'm understanding correctly, when we call `Ack::new` the ack data is initialized without the `calldata`, and it gets populated here from the pop'd frame before it's returned to what called this function

rust/otap-dataflow/crates/engine/src/effect_handler.rs
```diff
@@ -200,6 +201,48 @@ impl<PData> EffectHandlerCore<PData> {
...
+    ) -> Result<(), TypedError<AckMsg<PData>>>
+    where
+        F: FnOnce(AckMsg<PData>) -> Option<(usize, AckMsg<PData>)>,
+    {
+        // TODO: Placeholder: This returns a SendError::Closed as there
+        // is no appropriate message to send yet.
```
albertlockett, line 219: Just trying to understand -- so instead of returning this error, eventually the effect handler would emit a message to the correct subscriber on the `node_id`?

Then that subscriber would handle the ack, and possibly call `next_ack`?

rust/otap-dataflow/crates/engine/src/control.rs
jmacd: Discussed in a 1:1. I will extend the `Interests` bitflag in a future PR to control whether data is actually propagated backwards.

lquerel (approved):
Agree with the concept of data subscription idea.

rust/otap-dataflow/crates/otap/src/pdata.rs
albertlockett: I guess we could initialise it with empty vec

rust/otap-dataflow/crates/engine/src/control.rs
jmacd: #1244

rust/otap-dataflow/crates/engine/src/effect_handler.rs
jmacd:
```
    /// Send a AckMsg using a context-transfer function.  The context
    /// transfer function applies PData-specific logic to discover the
    /// next recipient in the chain of Acks, if any.  When there is a
    /// recipient, this returns its node_id and the AckMsg prepared for
    /// delivery with the recipient's calldata.
    pub async fn route_ack<Transfer>(
        &self,
        ack_in: AckMsg<PData>,
        transfer: Transfer,
    ) -> Result<(), TypedError<AckMsg<PData>>>
    where
        Transfer: FnOnce(AckMsg<PData>) -> Option<(usize, AckMsg<PData>)>,
```

rust/otap-dataflow/crates/engine/src/effect_handler.rs
jmacd: I've updated the docs,
```
    /// Send a AckMsg using a context-transfer function.  The context
    /// transfer function applies PData-specific logic to discover the
    /// next recipient in the chain of Acks, if any.  When there is a
    /// recipient, this returns its node_id and the AckMsg prepared for
    /// delivery with the recipient's calldata.
    pub async fn route_ack<Transfer>(
        &self,
        ack_in: AckMsg<PData>,
        transfer: Transfer,
    ) -> Result<(), TypedError<AckMsg<PData>>>
    where
        Transfer: FnOnce(AckMsg<PData>) -> Option<(usize, AckMsg<PData>)>,
```

the effect handler calls transfer() to prepare `AckMsg::calldata` for the recipient by `node_id`.

The recipient will call the same function that got us here, notify_ack() on their effect handler, to deliver to the next recipient backwards in the pipeline.

Only the core effect handler applies these "transfer" functions i.e., next_ack and next_nack, but it's in the engine crate so it doesn't know the real data type.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
jmacd: I plan to completely replace this code, so I was only focused on making it build and pass its old tests. I agree we should not panic even for internal errors in the Ack/Nack routing logic.

rust/otap-dataflow/crates/otap/src/pdata.rs
jmacd: I see, so not an Option::None but a smallvec![] to start?
This OK with me, maybe slightly more defensive to leave the Option<>.

jmacd: Done!

rust/otap-dataflow/crates/otap/src/retry_processor.rs
jmacd: (Fixed.)

jmacd:
About codecov: the shared components are not really implemented for us to test, we don't have every one of (shared, local) x (E, P, R).

Issue #888, opened 2025-08-07
Rate Limiter Design


lquerel:
TBD

See #919 for some pointers.

PR #783, opened 2025-07-25
Feature retry processor (v2)

daviddahl:
(This is a refactored version after the EffectHandler, etc changes in #532)

Add RetryProcessor with Testing and Benchmarks

Summary

Adds a new RetryProcessor component that provides reliable message delivery with exponential backoff retry logic for the OTAP dataflow engine.

Changes

•  New RetryProcessor: Handles message retries with configurable policies, ACK/NACK feedback, and graceful shutdown
•  9 comprehensive unit tests: Cover all major functionality including retry logic, backpressure, and edge cases  
•  Performance benchmarks: Microbenchmarks showing ~150-230ns latency for core operations

Key Features

•  Exponential backoff with configurable limits (default: 3 retries, 1s initial delay)
•  ACK/NACK message tracking with HashMap-based O(1) operations
•  Queue limits for backpressure handling
•  Dynamic configuration updates
•  Graceful shutdown with pending message flushing

PR #1234, opened 2025-10-06
Pipeline controller: Delayed data support

jmacd:
Part of #888 

Part of #783 

Part of #1245 

Adds a "delay data" pipeline control message.
Adds a delayed-data heap in the pipeline controller. 
Adds a "delayed data" node control message.
Tests delayed-data delivery.


rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -186,14 +226,19 @@ impl<PData> PipelineCtrlMsgManager<PData> {
                         PipelineControlMsg::CancelTelemetryTimer { node_id, _temp } => {
                             self.telemetry_timers.cancel(node_id);
                         }
+            PipelineControlMsg::DelayData { node_id, when, data } => {
+                let delayed = Delayed { node_id, when, data };
+                self.delayed_data.push(delayed);
+            }
```
albertlockett: `cargo fmt` not correcting the indentation inside the macro automatically

```suggestion
                    PipelineControlMsg::DelayData { node_id, when, data } => {
                        let delayed = Delayed { node_id, when, data };
                        self.delayed_data.push(delayed);
                    }
```

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -214,13 +259,19 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
+                    }
+
                     // Deliver all accumulated control messages (best-effort)
                     for (node_id, msg) in to_send {
                         if let Some(sender) = self.control_senders.get(node_id) {
                             // Use try_send as a fast path:
                             // - avoids allocating/awaiting a future when the channel has capacity
                             // - keeps the event loop responsive and reduces timer jitter
-                            // - isolates backpressure to congested channels (only await on Full)
+                // - isolates backpressure to congested channels (only await on Full)
```
albertlockett:
```suggestion
                            // - isolates backpressure to congested channels (only await on Full)
````

jmacd:
@albertlockett It's case (a). The node can only delay a request to itself. I considered the other case: components don't have a node_id corresponding with each output channel. As a matter of back-pressure working, I think we want to return to the node with the delay to choose which output channel and to block on the pipeline.

jmacd:
@lquerel asked about memory control. See #1245.

PR #1240, opened 2025-10-07
Fix map schema to id

### LINKED ISSUE #1223: Is the Map type being used correctly when producing BatchArrowRecords?
While working on #1221, I noticed that `schema_id` generation logic for `Map` types seems incorrect:

https://github.com/open-telemetry/otel-arrow/blob/55aaa48e214407da73f940f6ca29357a3bd876fa/rust/otel-arrow-rust/src/otap/schema.rs#L124-L130

We are writing the same data type twice. There is a test `test_all_field_types` which tests the `Map` type like this:

```rust
Field::new(
    "map",
    DataType::Map(Arc::new(Field::new("item", DataType::Utf8, true)), true),
    true,
),
```

However, the arrow [doc](https://docs.rs/arrow/latest/arrow/datatypes/enum.DataType.html#variant.Map) says that:

> A Map is a logical nested type that is represented as List<entries: Struct<key: K, value: V>>

Shouldn't we be testing it with something like this instead?

```rust
Field::new(
    "map",
    DataType::Map(
        Arc::new(Field::new(
            "entries",
            DataType::Struct(
                vec![
                    Field::new("key", DataType::Utf8, false),
                    Field::new("value", DataType::Utf8, true),
                ]
                .into(),
            ),
            true,
        )),
        true,
    ),
    true,
),
```

danielthank:
Fix #1223 

The corresponding go logic is
https://github.com/open-telemetry/otel-arrow/blob/452821e455b16e9a397a09d299340e197eb91571/go/pkg/arrow/schema.go#L259-L264

However, there is no equivalent `KeyType` and `ItemType` function in rust so pattern matching them manually.

Arrow rust docs: https://docs.rs/arrow/latest/arrow/datatypes/enum.DataType.html#variant.Map

### DIFF
--- a/rust/otel-arrow-rust/src/otap/schema.rs
+++ b/rust/otel-arrow-rust/src/otap/schema.rs
@@ -115,10 +115,27 @@ impl SchemaIdBuilder {
             }
 
             Map(field, _) => {
+                let [key_field, value_field] = match field.data_type() {
+                    Struct(fields) => {
+                        // Try the common case first: two fields named "key" and "value"
+                        match fields.iter().as_slice() {
+                            [k, v] if k.name() == "key" && v.name() == "value" => [k, v],
+                            [v, k] if k.name() == "key" && v.name() == "value" => [k, v],
+                            _ => {
+                                let key = fields.iter().find(|f| f.name() == "key")
+                                    .expect("Map field must contain a field named 'key'");
+                                let value = fields.iter().find(|f| f.name() == "value")
+                                    .expect("Map field must contain a field named 'value'");
+                                [key, value]
+                            }
+                        }
+                    },
+                    _ => panic!("Map field must be a struct"),
+                };
                 self.out.push_str("Map<");
-                self.write_data_type(field.data_type());
+                self.write_data_type(key_field.data_type());
                 self.out.push(',');
-                self.write_data_type(field.data_type());
+                self.write_data_type(value_field.data_type());
                 self.out.push('>');
             }
 
@@ -163,7 +180,7 @@ impl Default for SchemaIdBuilder {
 #[cfg(test)]
 mod test {
     use arrow::datatypes::{DataType, Field, Schema, TimeUnit, UnionFields, UnionMode};
-    use std::sync::Arc;
+    use std::{sync::Arc, vec};
 
     use crate::otap::schema::SchemaIdBuilder;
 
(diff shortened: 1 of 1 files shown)


rust/otel-arrow-rust/src/otap/schema.rs
```diff
@@ -115,10 +115,31 @@ impl SchemaIdBuilder {
...
+                                    .expect("Map field must contain a field named 'key'");
+                                let value = fields
+                                    .iter()
+                                    .find(|f| f.name() == "value")
+                                    .expect("Map field must contain a field named 'value'");
+                                [key, value]
+                            }
+                        }
+                    }
+                    _ => panic!("Map field must be a struct"),
```
albertlockett: I'm not sure we'll want to panic here on the off chance that the fields do not contain these keys. The arrow-rs map docs state: https://docs.rs/arrow/latest/arrow/datatypes/enum.DataType.html#variant.Map

> In a field with Map type, the field has a child Struct field, which then has two children: key type and the second the value type. The names of the child fields may be respectively “entries”, “key”, and “value”, _but this is not enforced._

Since it's not enforced, I think we'll want to handle this gracefully. Maybe this `match` could return an `Option<[Field, Field]>`, and if the option is `None` we use the existing logic as a fallback. Does that sound OK?

rust/otel-arrow-rust/src/otap/schema.rs
```diff
@@ -198,7 +219,83 @@ mod test {
             ),
             Field::new(
                 "map",
-                DataType::Map(Arc::new(Field::new("item", DataType::Utf8, true)), true),
+                DataType::Map(
```
albertlockett, line 208: great tests!

rust/otel-arrow-rust/src/otap/schema.rs
```diff
@@ -163,7 +184,7 @@ impl Default for SchemaIdBuilder {
 #[cfg(test)]
 mod test {
     use arrow::datatypes::{DataType, Field, Schema, TimeUnit, UnionFields, UnionMode};
-    use std::sync::Arc;
+    use std::{sync::Arc, vec};
```
albertlockett: nit: `vec` is part of the standard prelude, so we don't need to explicitly import it

```suggestion
 use std::sync::Arc;
```

jmacd (approved):
Thanks!

rust/otel-arrow-rust/src/otap/schema.rs
utpilla: Or we could do something like this. If it's not a valid struct, we would simply serialize `Map<>`.

```rust
Map(field, _) => {
    self.out.push_str("Map<");
    if let Struct(fields) = field.data_type() {
        if fields.len() == 2 {
            self.write_data_type(fields[0].data_type()); // key type
            self.out.push(',');
            self.write_data_type(fields[1].data_type()); // value type
        }
    }
    self.out.push('>');
}
```

albertlockett: Yeah that could work too @utpilla -- it simplifies the code and it's sort of similar where it assumes the key is field[0] and value is field[1]:

https://github.com/apache/arrow-go/blob/3ae84281674622d33b4617c878e099d13d4a1113/arrow/datatype_nested.go#L593-L596

rust/otel-arrow-rust/src/otap/schema.rs
danielthank: Thanks!

rust/otel-arrow-rust/src/otap/schema.rs
```diff
@@ -198,7 +205,65 @@ mod test {
...
+                            .into(),
+                        ),
+                        true,
+                    )),
+                    true,
+                ),
+                true,
+            ),
+            Field::new(
+                "map_extra",
```
utpilla: nit: Rename this to make the intent clear.
```suggestion
                "map_invalid",
```

PR #1241, opened 2025-10-07
Update repo codecov to allow for fluctuation

drewrelmas:
Currently the codecov status check tends to transiently mark builds red due to very small fluctuation in coverage percentages.

The best examples of this are in non-code PRs like updating github workflow hash versions or release version updates:

- https://app.codecov.io/gh/open-telemetry/otel-arrow/pull/1131/indirect-changes
- https://app.codecov.io/gh/open-telemetry/otel-arrow/pull/1202/indirect-changes
- https://app.codecov.io/gh/open-telemetry/otel-arrow/pull/1192/indirect-changes

It seems there may be some flakiness based on fake data generation - it is shown under 'indirect changes' for these PRs:

image

This change:

- allows `0.1%` fluctuation in overall repo coverage to avoid the small +- percentage. 
- fixes a small misconfiguration in the `component_management` section (`ignore` isn't a valid key there for path scoping)

### DIFF
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -3,7 +3,7 @@ coverage:
     project: # Overall repo coverage
       default:
         target: auto # Increase
-        threshold: 0.5% # Allow small fluctuation
+        threshold: 0.1% # Allow small fluctuation
 
 comment:
   layout: "header, diff, flags, components"


jmacd (approved):
Yes!! Thank you.

jmacd:
I think we should probably file issues for "big" individual contributions when we see them, like for fake_data_generator here. We should be able to improve coverage for probabilistic tests, but the chance of non-coverage never disappears. This allowance helps.

PR #1246, opened 2025-10-07
[WIP] Receiver accessory slots::State<UData> for Ack/Nack correlation

jmacd:
For discussion in weekly SIG.

This originally appeared in #1197. The core datastructure is a pair of Context8u8-compatible values for storing receiver state in the data structs of #1225. In OTLP receiver, this would be a oneshot::Channel, in OTAP receiver, this would be a pair like { tx, batch_id }.


jmacd:
Replaced by #1268

PR #1247, opened 2025-10-07
Add Interests::RETURN_DATA

### LINKED ISSUE #1244: Add component-level subscription Interests::RETURN_DATA
This is an interest that is checked whenever an Ack or Nack first enters the return path. 

If the recipient of an ACKS or NACKS subscription also sets RETURN_DATA, the original payload will be preserved.

This specific interest will be inherited by successive elements in the context stack, by default, i.e., it will auto-copy from frame to frame so that we have a O(1) lookup.

Without this interest set, the first notify_ack() or notify_nack() will take the payload and drop it. This operation will retain the original structure of the request, which indicates the correct signal type but has 0 bytes or 0 rows of data.

Presently, `OtapPayload` has a method for this, though I'm not sure it's properly named:

```
    /// Removes the payload from this request, leaving an empty request.
    #[must_use]
    pub fn payload(&mut self) -> Self {
        match self {
            Self::OtlpBytes(value) => Self::OtlpBytes(value.payload()),
            Self::OtapArrowRecords(value) => Self::OtapArrowRecords(value.payload()),
        }
    }
```

jmacd:
Fixes #1244

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/pdata.rs
+++ b/rust/otap-dataflow/crates/otap/src/pdata.rs
@@ -707,6 +707,7 @@ impl ConsumerEffectHandlerExtension<OtapPdata>
 #[cfg(test)]
 mod test {
     use super::*;
+    use crate::testing::TestCallData;
     use otel_arrow_rust::{
         otap::OtapArrowRecords,
         proto::opentelemetry::{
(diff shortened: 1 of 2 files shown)


jmacd:
Will re-open

PR #1249, opened 2025-10-08
Complete Ack/Nack control message routing and delivery

### LINKED ISSUE #498: [otap-dataflow] emit control message from pipeline components using effect handler

### LINKED ISSUE #1244: Add component-level subscription Interests::RETURN_DATA

jmacd:
This finishes the basic integration of Ack/Nack delivery into the engine. Future work will involve modifying all components with an Ack/Nack interest to subscribe-to and route these correctly.

Fixes #498 
Fixes #1244.

Adds DeliverAck, DeliverNack messages to the pipeline controller.

Finishes EffectHandlerCore route_ack, route_nack. Modifies error_exporter and noop_exporter tests for Nack and Ack delivery.

Incorporates https://github.com/open-telemetry/otel-arrow/pull/1247; tests RETURN_DATA integration with Ack/Nack delivery.

Noteworthy simplification: route_ack and route_nack return untyped Errors, making error handling much easier. We can revisit this, in case we want to handle the data that would otherwise drop when it can't be returned.


rust/otap-dataflow/crates/engine/src/effect_handler.rs
```diff
@@ -149,6 +148,17 @@ impl<PData> EffectHandlerCore<PData> {
         UdpSocket::from_std(sock.into()).map_err(into_engine_error)
     }
 
+    // Re-usable function
+    async fn pipeline_msg_sender(
```
lquerel: I find the name of this method confusing. I suggest `send_pipeline_ctrl_msg`.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -267,30 +273,33 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
-
+                        self.try_send(node_id, msg).await;
                     }
                 }
             }
         }
         Ok(())
     }
+
+    async fn try_send(&mut self, node_id: usize, msg: NodeControlMsg<PData>) {
```
lquerel: I'wondering if we shouldn't call this method `send`.

rust/otap-dataflow/crates/engine/src/effect_handler.rs
jmacd: Done

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
jmacd: Done

PR #1250, opened 2025-10-08
[otap-dataflow] Add Sampling (based on zap) to DebugProcessor

c1ly:
Added the SamplingConfig struct to allow the user to configure their sampling settings

```rust
#[derive(Debug, Clone, Copy, Deserialize)]
pub struct SamplingConfig {
    // number of samples to log initially within the sampling interval
    pub sampling_initial: u64,
    // the sampling rate
    pub sampling_thereafter: u64,
    // the interval that we sample in, unit is seconds
    pub sampling_interval: u64,
}
```

Added the Sampler struct to handle the sampling logic for the debug processor

```rust
/// The sampler keeps track of the current state, number of msgs seen and next interval
pub struct Sampler {
    // sampling settings
    sampling_config: SamplingConfig,
    // counter for msgs seen
    msgs_current_interval: u64,
    next_interval: Instant,
}
```

uses the sample function that accepts a closure (in this closure is the call to output the msgs either to console or outport depending on configuration) to determine whether to log or drop a msg

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/debug_processor/sampling.rs
+++ b/rust/otap-dataflow/crates/otap/src/debug_processor/sampling.rs
@@ -11,6 +11,7 @@ use tokio::time::{Duration, Instant};
 #[serde(tag = "type", rename_all = "snake_case")]
 pub enum SamplingConfig {
     NoSampling,
+    // ToDo: Come up with a different name for this?
     ZapSampling {
         sampling_initial: u64,
         sampling_thereafter: u64,


rust/otap-dataflow/crates/otap/src/debug_processor/README.md
```diff
@@ -98,6 +102,43 @@ In the config setting we tell the debug-processor to use `logging_port`
...
+
+You can control how often msgs are sent out via the sampling settings
+
+#### Sampling Config
+
+```yaml
+   sampling:
+      sampling_initial: 1
+      sampling_thereafter: 1
+      sampling_interval: 1
```
lquerel:
```suggestion
      sampling_initial: 2
      sampling_thereafter: 5
      sampling_interval: 2
```

lquerel: I suggest to combine the config example with the next one to simplify.

rust/otap-dataflow/crates/otap/src/debug_processor/README.md
```diff
@@ -98,6 +102,43 @@ In the config setting we tell the debug-processor to use `logging_port`
...
+
+#### Sampling Example
+
+```yaml
+   sampling:
+      sampling_initial: 2
+      sampling_thereafter: 5
+      sampling_interval: 2
+```
+
```
lquerel: I suggest to reuse the previous updated example.

rust/otap-dataflow/crates/otap/src/debug_processor/sampling.rs
```diff
@@ -0,0 +1,72 @@
...
+
+//! Implementation of the Sampler for the DebugProcessor
+
+use otap_df_engine::error::Error;
+use serde::Deserialize;
+use tokio::time::{Duration, Instant};
+
+/// Sampling Settings
+#[derive(Debug, Clone, Copy, Deserialize)]
+pub struct SamplingConfig {
```
lquerel: I suggest to transform this struct as an enum so in the future we can support multiple sampling stragegy.
For now, we could define two variants:
- NoSampling
- ZapSampling

NoSampling could be the default one, so when this field is not specified in the DebugProcessor config there is no sampling.

jmacd: The connection with "Zap" across language boundaries is mildly strange. I suggest a name like TimeSampling.

As you may see https://github.com/open-telemetry/opentelemetry-rust/issues/3195 I have tinkered in Rust sampling for OTel purposes, I would be nice to re-use a SDK component here, maybe.

c1ly: Left it as ZapSampling for now, added a note to consider a different name as we add multiple sampling strategies in the future

PR #1262, opened 2025-10-09
[query-engine] Relax restrictions in KQL summarize expressions

### LINKED ISSUE #1243: [kql-parser] Flexibility in bin(<Timestamp>, <Timespan>) handling
Currently, parsing the KQL snippet for `bin(<Timestamp>, <Timespan>)` will fail because there is no assignment on the left-hand side of the bin to hint at the key to be used in the result of that operation.

```kql
// This is valid
source | summarize by DestinationTimestamp=bin(SourceTimestamp, 1m)

// This fails
// Parser should re-use identifier 'SourceTimestamp' by default
source | summarize by bin(SourceTimestamp, 1m)
```

The second example is valid in KQL, and should be supported in our parser.

CodeBlanch:
## Changes

* Allow aggregate and group by expression without explicit assignment in KQL summarize expressions

## Details

Today you can do:
* `summarize by Timestamp = bin(Timestamp, 1m)`
* `summarize Count = count()`

But you will get an error if you do:
* `summarize by bin(Timestamp, 1m)`
* `summarize count()`

This PR makes the second two examples work. In the first case the name `Timestamp` will be used and in the second case `count`.

This is done by giving the KQL parser some logic to determine the identifier for any given scalar expression. Note: We could probably use this in extend\project* cases too but I plan to look at that in a follow-up.

Fixes #1243


rust/experimental/query_engine/kql-parser/src/tabular_expressions.rs
```diff
@@ -373,64 +372,42 @@ pub(crate) fn parse_summarize_expression(
...
+                                    ));
+                                }
+
+                                group_by_expressions.insert(full_identifier.into(), scalar);
                             }
                             _ => {
                                 return Err(ParserError::SyntaxError(
                                     group_by_first_rule_location,
-                                    "Could not determine the source and/or name for summary group-by expression. Try using assignment syntax instead (Name = [expression]).".into(),
+                                    "Could not determine the name for summary group-by expression. Try using assignment syntax instead (Name = [expression]).".into(),
```
drewrelmas: This might need changed in a few tests as well, but wondering if we can make it clearer?

```suggestion
                                    "Could not determine the name for summary group-by expression. Try using assignment syntax instead (new_identifier = [expression]).".into(),
```

CodeBlanch: Updated

Issue #1073, opened 2025-09-05
[otap-dataflow] syslog-cef receiver should expect syslog headers with CEF


clhain:
The syslog-CEF receiver currently doesn't seem to parse syslog messages with CEF message bodies. Sending an encoded CEF format message directly without the syslog header fields does parse correctly, but as I read it, this seems like incorrect behavior. Specifically:

(from [the most official looking spec doc I can find](https://www.microfocus.com/documentation/arcsight/arcsight-smartconnectors-8.3/cef-implementation-standard/Content/CEF/Chapter%201%20What%20is%20CEF.htm))

> ## Header Information
> 
> CEF uses Syslog as a transport mechanism. It uses the following format that contains a Syslog prefix, a header, and an extension:
> 
> Jan 18 11:07:53 host CEF:Version|Device Vendor|Device Product|Device Version|Device Event Class ID|Name|Severity|[Extension]
> 
> 

> ## Using CEF Without Syslog
> Syslog applies a syslog prefix to each message, no matter which device it arrives from, that contains the date and hostname in the following example:
> 
> Jan 18 11:07:53 host CEF:Version|…
> 
> Even if an event producer is unable to write Syslog messages, it is possible to write the events to a file by performing the following steps:
> 
> Discard the syslog prefix (Jan 18 11:07:53 host).
> 
> Begin the message with the following format:
> 
> CEF:Version|Device Vendor|Device Product|Device Version|Device Event Class ID|Name|Severity|[Extension]

I'm assuming the long term vision for this is something closer to what the go collector does where we can specify the expected syslog format and then apply various parsers to the message body (cef being one of many). For now this just prevents us from doing a CEF-based syslog performance comparison with identical messages here and against the go collector.

utpilla:
@clhain Thanks for creating this issue!

> Sending an encoded CEF format message directly without the syslog header fields does parse correctly, but as I read it, this seems like incorrect behavior. 

I don't think this is incorrect. I think we are expected to parse CEF messages that don't have a Syslog header. However, not being able to parse CEF messages with a Syslog header looks incorrect.

> I'm assuming the long term vision for this is something closer to what the go collector does where we can specify the expected syslog format and then apply various parsers to the message body (cef being one of many).

[...]

PR #1264, opened 2025-10-10
[otap-df-otap] Add support for CEF messages with Syslog headers

utpilla:
Address #1073

## Changes
- Add support for parsing CEF messages with Syslog header
- Note that we still support parsing raw CEF messages

#### Raw CEF
```CEF:0|Security|threatmanager|1.0|100|worm successfully stopped|10|src=10.0.0.1 dst=2.1.2.2 spt=1232```

#### CEF with Syslog header (This is a partial header with RFC 3164, example is from the [spec](https://www.microfocus.com/documentation/arcsight/arcsight-smartconnectors-8.3/cef-implementation-standard/Content/CEF/Chapter%201%20What%20is%20CEF.htm))
```Sep 29 08:26:10 host CEF:1|Security|threatmanager|1.0|100|worm successfully stopped|10|src=10.0.0.1 dst=2.1.2.2 spt=1232```

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/arrow_records_encoder.rs
+++ b/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/arrow_records_encoder.rs
@@ -2643,7 +2643,10 @@ mod tests {
             log1_attrs.get("cef.device_version"),
             Some(&"1.0".to_string())
         );
-        assert_eq!(log1_attrs.get("cef.device_event_class_id"), Some(&"100".to_string()));
+        assert_eq!(
+            log1_attrs.get("cef.device_event_class_id"),
+            Some(&"100".to_string())
+        );
         assert_eq!(
             log1_attrs.get("cef.name"),
             Some(&"worm successfully stopped".to_string())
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/parser/mod.rs [hunk 1]
```diff
@@ -40,22 +40,79 @@ pub enum ParseError {
...
     /// Error parsing UTF-8 strings
     InvalidUtf8,
+    /// Error indicating that the format of the syslog message is unknown
+    UnknownFormat,
 }
 
 /// Parse a syslog message from bytes, automatically detecting the format
 pub(super) fn parse(input: &[u8]) -> Result<ParsedSyslogMessage<'_>, ParseError> {
-    // Check if it's a CEF message first
+    // Try pure CEF first - it's the simplest check
```
jmacd, line 49: I was going to suggest updating the docs, but now I realize there aren't any docs. Would you add a section in crates/otap/README.md explaining the logic that determines which Syslog/CEF form will be identified?

utpilla: Yeah, I'll be updating the README describing the receiver's behavior in terms of the message formats accepted (with examples) and also the attributes that the receiver would add when converting them to a `LogRecord`.

Issue #1139, opened 2025-09-16
Shutdown should gracefully drain remaining metric updates


clhain:
Right now, we allow telemetry for nodes to continue exporting after shutdown has been called - however we don't flush any remaining accumulation of metrics that occur between the last metric flush interval prior to shutdown and the shutdown call itself. This causes, for example, the  Fake Signal generator to report fewer 'total signals sent' than were actually sent, which impacts loss calculations in the perf tests.

So the problem looks like:
```
Signal generated
T0 - Signal generated
T1 - Metric flushed and available on /telemetry/metrics - e.g. signals_sent = 1
T2 - Signal generated
T3 - Metric flushed and available on /telemetry/metrics - e.g. signals_sent = 2
T4 - Signal generated
T5 - Shutdown Called, completed
T6+ - /telemetry/metrics still reports signals_sent = 2 instead of the correct 3
```

PR #1273, opened 2025-10-13
Improve graceful shutdown mechanism to ensure full metrics flushing and orderly pipeline termination

lquerel:
**Note: Many files were modified in this PR, but most of the changes are signature updates.**

This PR refines the graceful shutdown process of the pipeline engine to make node termination deterministic and data-safe (in particular for the metrics).

The following diagram describes the process.
image

Other changes:
- The deadline field in the Shutdown control message now represents an actual deadline instead of a duration.
- Local channels now expose an `is_empty` method, consistent with other channel implementations.
- Simplified some unit tests.
- Improved TestRuntime to make `otap_exporter` testing more robust.
	
Closes: #1139

### DIFF
--- a/rust/otap-dataflow/Cargo.toml
+++ b/rust/otap-dataflow/Cargo.toml
@@ -39,6 +39,25 @@ clap.workspace = true
 mimalloc-rust.workspace = true
 
 [workspace.dependencies]
+
+fluke-hpack = "0.3.1"
+futures-channel = "0.3"
+local-sync = "0.1.1"
+async-unsync = "0.3.0"
+unsync = "0.1.2"
+portpicker = "0.1.1"
+tokio-stream = "0.1.17"
+
+tower-service = "0.3"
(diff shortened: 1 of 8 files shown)


clhain:
Checked with the benchmark framework and this is working as expected now. Thanks!!

rust/otap-dataflow/crates/admin/src/pipeline_group.rs
```diff
@@ -62,8 +63,9 @@ async fn shutdown_all_pipelines(State(state): State<AppState>) -> impl IntoRespo
         .ctrl_msg_senders
         .iter()
         .filter_map(|sender| {
+            let deadline = Instant::now() + Duration::from_secs(10);
```
jmacd, line 67:
```suggestion
            // ToDo configurable shutdown timeout
            let deadline = Instant::now() + Duration::from_secs(10);
```

Issue #1213, opened 2025-10-01
Filter Processor


lquerel:
In the long run, we'll have a KQL|OTTL processor capable of supporting a wide range of use cases, but in the meantime it's probably useful to support a simple filtering processor.

```yaml
    filter/filter-1:
        logs:
            include:
                match_type: regexp
                record_attributes:
                    - key: <name>
                      value: <value>
        metrics:
            include:
                expressions:
                    - Label("xyz") == "platform"
                    - Label("xyz") == "hardware"
                    - Label("xyz") == "datapath"
                match_type: expr
        ...
```

Signals to support:
- Metrics (higher priority)
- Logs
- Spans (lower priority)

For more details see: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/filterprocessor

jmacd:
Similar to our existing attributes_processor, the "filter" processor is a basic component that filters without modifying data.

c1ly:
ToDo in follow up PR to https://github.com/open-telemetry/otel-arrow/pull/1279

- Improve build_id() use the booleanbuilder to build the filter based on whether the id seen is in the id_set + combine use of booleanbuilder with current implementation and switch between the two based on the input size 

see -> https://github.com/open-telemetry/otel-arrow/pull/1279#discussion_r2453094683
also see -> https://github.com/open-telemetry/otel-arrow/pull/1279#discussion_r2433517168

- Handle optional record batches (log attributes and resource attributes) 
see -> https://github.com/open-telemetry/otel-arrow/pull/1279#discussion_r2441032269
also see -> https://github.com/open-telemetry/otel-arrow/pull/1279#discussion_r2441159073


- Handle string column stored as DictionaryArray 
see -> https://github.com/open-telemetry/otel-arrow/pull/1279#discussion_r2441128853

PR #1279, opened 2025-10-13
[otap-dataflow]Basic filter processor - log support

c1ly:
Add a basic filter processor that is based on the golang version minus the OTTL support

Add support for log filtering

Defined to following rust structs that will allow us to filter on 
- resource attributes
- log record attributes
- log record bodies
- log record severity text
- log record severity number (optional)

```rust
/// struct that describes the overall requirements to use in order to filter logs
pub struct LogFilter {
    // Include match properties describe logs that should be included in the Collector Service pipeline,
    // all other logs should be dropped from further processing.
    include: LogMatchProperties,
    // Exclude match properties describe logs that should be excluded from the Collector Service pipeline,
    // all other logs should be included.
    exclude: LogMatchProperties,

    // LogConditions is a list of OTTL conditions for an ottllog context.
    // If any condition resolves to true, the log event will be dropped.
    // Supports `and`, `or`, and `()`
    #[allow(dead_code)]
    log_record: Vec<String>,
}

/// LogMatchProperties specifies the set of properties in a log to match against and the type of string pattern matching to use.
pub struct LogMatchProperties {
    // LogMatchType specifies the type of matching desired
    match_type: LogMatchType,

    // ResourceAttributes defines a list of possible resource attributes to match logs against.

    // RecordAttributes defines a list of possible record attributes to match logs against.
    record_attributes: Vec<KeyValue>,

    // SeverityTexts is a list of strings that the LogRecord's severity text field must match
    // against.
    severity_texts: Vec<String>,

    // SeverityNumberProperties defines how to match against a log record's SeverityNumber, if defined.
    severity_number: Option<LogSeverityNumberMatchProperties>,

    // LogBodies is a list of values that the LogRecord's body field must match
    // against.
    bodies: Vec<AnyValue>,
}

/// LogSeverityNumberMatchProperties specifies the requirements needed to match on the log severity field
pub struct LogSeverityNumberMatchProperties {
    // Min is the minimum severity needed for the log record to match.
    // This corresponds to the short names specified here:
    // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#displaying-severity
    // this field is case-insensitive ("INFO" == "info")
    min: i32,
    // MatchUndefined lets logs records with "unknown" severity match.
    // If MinSeverity is not set, this field is ignored, as fields are not matched based on severity.
    match_undefined: bool,
}
```

Followup PR
1. Update Regexp, handle dictionary array 
2. Handle edge case where resource attributes and log attribute record batches are not present since they are optional
3. Utilize BooleanBuilder to optimize building the filter

Future PR
- add trace and metric support

Part of #1213


rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
```diff
@@ -0,0 +1,290 @@
...
+	fn get_log_record_filter(&self, logs_payload: &OtapArrowRecord) -> Result<BooleanArray, Error> {
+		let log_records = logs_payload.get(ArrowPayLoadType::Logs).map_err()?;
+
+		// create filter for severity texts
+		let severity_texts_column = log_records.column_by_name(self::SEVERITY_TEXT);
+		let mut severity_texts_filter;
+		for severity_text in self.serverity_texts {
+			let severity_text_scaler = StringArray::new_scaler(severity_text).map_err()?;
+			let severity_text_filter = arrow::compute::kernels::cmp::eq(&severity_texts_column, &severity_text_scaler).map_err()?;
+			severity_texts_filter = arrow::compute::or_kleene(&severity_text_filter, &severity_text_filter).map_err()?;
```
lquerel: The parameters seem to be repetitive.

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
```diff
@@ -0,0 +1,381 @@
...
+
+impl LogFilter {
+	pub fn filter(&self, logs_payload: OtapArrowRecord) -> Result<OtapArrowRecords, Error>{
+		let (include_resource_attr_filter, include_log_record_filter, include_log_attr_filter) = self.include.create_filters(logs_payload, false)?;
+		let (exclude_resource_attr_filter, exclude_log_record_filter, exclude_log_attr_filter) = self.exclude.create_filters(logs_payload, true)?;
+
+		// combine the include and exclude filters
+		let resource_attr_filter = arrow::compute::or_kleene(&include_resource_attr_filter, &exclude_resource_attr_filter).map_err()?;
+		let log_record_filter = arrow::compute::or_kleene(&include_log_record_filter, &exclude_log_record_filter).map_err()?;
+		let log_attr_filter = arrow::compute::or_kleene(&include_log_attr_filter, &exclude_log_attr_filter).map_err()?;
```
albertlockett: Should this be `and` instead of `or`?

In the config, it says that if both exclude and include are defined, include filtering occurs first. 

If I'm understanding correctly, say we had filter `BooleanArray`s like:
```
include = T, T, F, T
exclude = T, F, T, T
```
The `include` filter drops the 3rd element, and the `exclude` filter drops the 2nd element, so we'd want to keep only the 1st and 4th elements. But if we `or` these together we'd get `T,T,T,T`, which keeps all the elements, and that's not what we want, right?

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
```diff
@@ -0,0 +1,381 @@
...
+        // HERE WE CLEAN UP THE TABLE WE EXTRACT THE IDS OF RECORDS WE ARE REMOVING AND MAKE SURE TO SYNC THESE REMOVALS WITH THE OTHER TABLES
+
+        // we start with the resource_attr we get the id of the resource_attr that are being removed and use that to update the log_records filter
+        // we want to remove the log_records the matching parent_id as the resource_attr that have been removed
+
+
+        // create filters that we will use to get the rows that are getting removed
+        // use these filters to get the ids of rows that are getting removed from the resource_attr
+        let id_resource_attr_filter = arrow::compute::not(&resource_attr_filter).map_err()?;
+        let resource_id_column = resource_attrs.column_by_name("parent_id").map_err()?;
```
albertlockett: We have a helper function that might be helpful here .. it can map the column not being found into an `otel_arrow_rust::error::Error::ColumnNotFound` error.

Also, `otel_arrow_rust` defines a bunch of constants for each column name that we typically use.

```rust
use crate::arrays::get_required_array;
use crate::schema::consts;
// ...
let resource_id_column = get_required_array(resource_attrs, consts::PARENT_ID)?;
```

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
```diff
@@ -0,0 +1,381 @@
+// Copyright The OpenTelemetry Authors
```
albertlockett: For these types of modules that contain utility functions for operating on OTAP Record batches, we've typically kept them inside `otel_arrow_rust::otap` module. 
https://github.com/open-telemetry/otel-arrow/tree/main/rust/otel-arrow-rust/src/otap

Do you think you could move this to something like `otel-arrow-rust/src/otap/filter.rs` maybe?

If for no other reason, it keeps all our OTAP transformation type code in one place. 

The other kind of nice reason for doing this this is: you could have the methods here return an a result with the error type `otel_arrow_rust::error::Error`, which many of `otel_arrow_rusts`'s utility functions return. That way, you can just use the functions with the `?` without having to map the error into say, a `otap_df_engine::error::Error::ProcessorError` or something

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
```diff
@@ -0,0 +1,381 @@
...
+        let resource_id_column = resource_attrs.column_by_name("parent_id").map_err()?;
+        // we have a have a array contain all the parent_ids that will be removed
+
+        // get ids that we being kept
+        let resource_ids_filtered = arrow::compute::filter(&resource_id_column, resource_attr_filter).map_err()?;
+        // get ids being removed
+        let resource_ids_removed = arrow::compute::filter(&resource_id_column, id_resource_attr_filter).map_err()?;
+
+        // update the log_records_filter to remove records that contain the parent_ids we found above
+        let log_record_resource_id_column = log_records.column_by_name("resource_id").map_err()?;
```
albertlockett: I think the resource ID column is actually embedded within a `StructArray` on the log record. It can be accessed like:
```rs
        let struct_array = get_required_array(log_records, consts::RESOURCE)?;
        let struct_array = struct_array
            .as_any()
            .downcast_ref::<StructArray>()
            .with_context(|| error::ColumnDataTypeMismatchSnafu {
                name: consts::RESOURCE,
                actual: struct_array.data_type().clone(),
                expect: Self::data_type().clone(),
            })?;
```

albertlockett: FYI the best reference for the OTAP schema (imo) is the golang code:
https://github.com/open-telemetry/otel-arrow/blob/877b2cb3dddef09143c7cb20164fd0ba789dd210/go/pkg/otel/logs/arrow/logs.go#L25-L52

https://github.com/open-telemetry/otel-arrow/blob/877b2cb3dddef09143c7cb20164fd0ba789dd210/go/pkg/otel/common/arrow/resource.go#L21-L40

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
```diff
@@ -0,0 +1,381 @@
...
+        // init booleanarray here
+        let log_record_resouce_id_filter;
+
+
+        // TODO: we remove overlapping ids to only get the unique ids that are being removed, 
+        //we can move this logic to the get...filter functions to make sure the ids all stick together 
+        //that is parent_id with 0 should not both get removed or kept should be xor. In the get...filter functions
+        // after we get a filter that extracts all the attributes we should get the ids and use that to build our final filter
+        let resource_ids = resource_ids_removed - resource_ids_filtered;
```
albertlockett: I'm not sure I understand the logic here. 

`resource_ids` is intended to be the unique IDs of resources we want to keep right?

But above, we compute:
```rs
        // get ids that we being kept
        let resource_ids_filtered = arrow::compute::filter(&resource_id_column, resource_attr_filter).map_err()?;
        // get ids being removed
        let resource_ids_removed = arrow::compute::filter(&resource_id_column, id_resource_attr_filter).map_err()?;
```

Isn't `resource_ids_filtered` the complement of `resource_ids_removed`?

I think what we'd really want here is the unique values in `resource_ids_filtered`, right?

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
```diff
@@ -0,0 +1,381 @@
...
+
+        // these are the ids we need to remove from the log_attr table
+        let ids = arrow::compute::filter(&log_record_id_column, &id_log_records_filter);
+        let log_attr_id_column = log_attr.column_by_name("parent_id");
+        let log_attr_parent_id_filter;
+        for id in ids() {
+            let id_scaler = UInt16Array::new_scaler(id).map_err()?;
+            let id_filter = arrow::compute::kernels::eq(&log_attr_id_column, &id_scaler).map_err()?;
+            log_attr_parent_id_filter = arrow::compute::or_kleene(&log_attr_parent_id_filter, &id_filter).map_err()?;
+        }
```
albertlockett: I think this might end up being slightly inefficient. We should probably find a more optimal way to do this, either now or in a fast followup to this PR.

For each iteration of this loop, I think we'll end up allocating a new boolean array. That might be OK above for resource_ids because chances could be that we only have a handful of resources for a batch of logs (that might still be a bad assumption), but in this loop every log will have a unique ID.

A few suggestions that could help, as we're thinking about optimizing this:

1. Using `BooleanBuilder` to create the mask. This type allows you to gradually build up a boolean array element by element:
```rs
use arrow::array::BooleanBuilder;
let mut log_attr_parent_id_filter = BooleanBuilder::with_capacity(log_attr_id_column.len());

// then you can do:
log_attr_parent_id_filter.append(bool);

// and finally
let log_attr_parent_id_filter = log_attr_parent_id_filter.finish(); // returns a BooleanArray
```

2.  It could be interesting could be to see if there's a way to get the contiguous ranges of IDs to be removed (or kept, whatever is easier to compute). The reason I think this might help is, when we create OTAP data from OTLP, we have a sequence of IDs for the resource, and a sequence of IDs for the logs. So it would be common to have OTAP data like:
```
// logs OTAP batch
resource.id = [1, 1, 1, 1, 2, 2, 2, 2, 3, ... 
id =          [0, 1, 2, 3, 4, 5, 6, 7, 8, ...
```

Let's say we filtered out resource `2`, then your range of filtering log IDs would be `[3, 8)`. Then we could do something like:
```rs
for id in ids {
  let mut remove = false;
  for (start, end) in remove_ranges {
    if id >= start and id < end {
       remove = true;
       break
    }
  }
  log_attr_parent_id_filter.append(remove);
}
let log_attr_parent_id_filter = log_attr_parent_id_filter.finish();
```
There might even be further optimizations we could make to this eventually. For example binary searching in the ranges, or we could eventually make it even faster if there were some assumptions we could make about the sort order of the columns

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
```diff
@@ -0,0 +1,381 @@
...
+    const BODY_STR: &str = "body_str";
+    const BODY_INT: &str = "body_int";
+    const BODY_DOUBLE: &str = "body_double";
+    const BODY_BOOL: &str = "body_bool";
+    const BODY_BYTES: &str = "body_bytes";
+    const BODY_SER: &str = "body_ser";
+
+    const KEY: &str = "key";
+    const SEVERITY_NUMBER: &str = "severity_number";
+    const SEVERITY_TEXT: &str = "severity_text";
```
albertlockett: We can use the constants defined in `otel_arrow_rust::schema::consts`

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
```diff
@@ -0,0 +1,381 @@
...
+			let key_filter = arrow::compute::kernels::cmp::eq(&key_column, &key_scaler).map_err()?;
+			let value_filter = match attribute_filter.value {
+				MatchValue::String(value) => {
+					// get string column
+					let string_column = resource_attrs.column_by_name(self::STR)?;
+					let value_scaler = StringArray::new_scaler(value).map_err()?;
+					arrow::compute::kernels::cmp::eq(&string_column, &value_scaler).map_err()?
+				}
+				MatchValue::Int(value) => {
+					let int_column = resource_attrs.column_by_name(self::INT)?;
```
albertlockett: One thing to handle here -- some of these columns on attribute record batches are optional. So for example, if we receive a batch of OTel data where there are no attributes of type int, we omit the `int` column.

c1ly: so this call would just return an error since there is no int column?

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
c1ly: I believe I was trying to get the complement of the unique values of `resource_ids_filtered`

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
c1ly: yes this should be `and`

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
c1ly: nvm this returns an option here

albertlockett: I don't think we want to return error, but how we handle it depends on the context.

In our docs we have, 
```
	// ResourceAttributes defines a list of possible resource attributes to match logs against.

	// RecordAttributes defines a list of possible record attributes to match logs against.
    record_attributes: Vec<KeyValue>,
```

So if we see there's no `int` column in `get_resource_attr_filter`, then we know no resource can have a matching attribute and we can probably just set the entire filter to `false` and return early because nothing matches this attribute filter.

However, if we're in the `get_log_attr_filter`, I think we just ignore the match criteria e.g. which would leave the `filter` as is without flipping elements to `true`

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
albertlockett: > the complement of the unique values of resource_ids_filtered

Would that be the unique IDs in `resource_ids_removed`?

c1ly: yup

rust/otap-dataflow/crates/otap/src/filter_processor/filter_logs.rs
c1ly: why don't we also ignore missing columns in `get_resource_attr_filter`?

albertlockett: because according to our docs:
```
 	// ResourceAttributes defines a list of possible resource attributes to match logs against.
		                                          ^^^^^^^^^^^^^^^^^^^^^^^
```
So if the column isn't there, the field can't have the attribute so it can't match.

This is different than what we've specificed for or the records attributes:
```
	// RecordAttributes defines a list of possible record attributes to match logs against.
	                                          ^^^^^^^^^^^^^^^^^^^^
```

That said, maybe we want to actually unify this behaviour in our attributes processor and handle the attributes in the same way. I realize we're inheriting this behaviour from the go imlpementation, and I'm not sure why they don't have a consistent behaviour when filtering by attributes. @jmacd / @lquerel what do you think?

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,492 @@
...
+    // RecordAttributes defines a list of possible record attributes to match logs against.
+    // A match occurs if any record attribute matches at least one expression in this given list.
+    record_attributes: Vec<KeyValue>,
+
+    // SeverityTexts is a list of strings that the LogRecord's severity text field must match
+    // against.
+    severity_texts: Vec<String>,
+
+    // SeverityNumberProperties defines how to match against a log record's SeverityNumber, if defined.
+    severity_number: Option<LogServerityNumberMatchProperties>,
```
albertlockett:
```suggestion
    severity_number: Option<LogSeverityNumberMatchProperties>,
```

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,492 @@
...
+        }
+
+        Ok((resource_attr_filter, log_record_filter, log_attr_filter))
+    }
+
+    fn get_resource_attr_filter(&self, logs_payload: &OtapArrowRecords) -> Result<BooleanArray> {
+        // get resource_attrs record batch
+        let resource_attrs = logs_payload
+            .get(ArrowPayloadType::ResourceAttrs)
+            .context(error::LogRecordNotFoundSnafu)?;
```
albertlockett: We discussed on Teams earlier, but just documenting again here so the TODO doesn't get lost. `ResourceAttrs` is an optional `RecordBatch`, so if this record batch doesn't exist, it means that no resources could match the filters. 

I think we could handle this by returning a `BooleanArray` filter mask completely containing false?

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,492 @@
...
+        let num_rows = resource_attrs.num_rows();
+
+        let mut attributes_filter = BooleanArray::new_null(num_rows);
+        let key_column = get_required_array(resource_attrs, consts::ATTRIBUTE_KEY)?;
+        // generate the filter for this record_batch
+        for attribute in &self.resource_attributes {
+            let key_scalar = StringArray::new_scalar(attribute.key.clone());
+
+            let key_filter = arrow::compute::kernels::cmp::eq(&key_column, &key_scalar)
+                .expect("columns should have equal length");
```
albertlockett: two nits:
- typically when we call `expect`, we add a comment specifying why the operation is safe
- I don't think the `expect` message here is making sense because `key_scalar` is s Scalar value because the 

```suggestion
            // safety: this operation will only fail if the columns are not the same length or if the types
            // are not valid for comparision, neither of which are the case here
            let key_filter = arrow::compute::kernels::cmp::eq(&key_column, &key_scalar)
                .expect("can compare string key column with string scalar");
```

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,492 @@
...
+            let value_filter = match &attribute.value {
+                AnyValue::String(value) => {
+                    // get string column
+                    let string_column = get_required_array(resource_attrs, consts::ATTRIBUTE_STR)?;
+                    match self.match_type {
+                        LogMatchType::Regexp => {
+                            let string_column = string_column
+                                .as_any()
+                                .downcast_ref::<StringArray>()
+                                .expect("array can be downcast to StringArray");
```
albertlockett, line 449: Unfortunately the `string_column` here might not be a `StringArray`, so this conversion isn't safe.

Many of our columns are dictionary encoded in a way that the type can either be dictionary encoded, or it could be a the native array type.

For dictionary encoding, what it means is that we keep a dictionary one array which are the possible values that can be in the array (called "keys") and another array which contains the indices of the values array, called the "keys". For example to encode the strings `["a", "a", "b", "a", "c", "c"]`, our dictionary encoded arrays would be like:
```
values: ["a", "b", "c"],
keys: [0, 0, 1, 0, 2, 2]
```

This encoding is encapsulated in a type of arrow array called `DictionaryArray`  https://docs.rs/arrow/latest/arrow/array/struct.DictionaryArray.html, which is dynamic over both the type of key, and the values. 

Different types of keys can be used depending on the size of the dictionary. e.g. `DictionaryArray<UInt8, String>` can hold 255 values, where `DictionaryArray<UInt8, String>` can hold 2^16 values. In fact, we dynamically choose between the different dictionary keys depending on how many values there are in the array. If there are more than 2^16 values, we use just use a `StringArray`.

All this to say that we need to handle this here, which adds a bit of complication so handling regex matches could do in a followup. The reason I say that it's complicated is because unfortunately `regexp_is_match_scalar` won't accept a `DictionaryArray` event if it's values are `Strings`.

So what we'd need to do is something like:
```rs
match string_column.data_type() {
  DataType::Utf8 => { current logic },
  DataType::Dictionary(_, val) => {
    match **key, **val {
       (DataType::UInt8, DataType::Utf8) => {
           // downcast to dictionary array
           // safety: we've checked the datatype already
           let dict_arr = string_column.as_any().downcast_ref::<DictionaryArray<UInt8Type>>().expect("can cast to type");

           // get string from values
           // safety: we've checked the type
           let string_values = dict_arr.values().as_any().downcast_ref::<StringArray>().expect("can cast to type");
           let val_filt = regexp_is_match_scalar(string_values, "a*", None).unwrap();

           let key_filt = BooleanBuilder::with_capacity(dict_arr.len());
           for key in dict_arr.keys() {
              if let Some(k) =  key {
                   key_filter.append_value(val_filt.value(k as usize));
               }
           }
           key_filt.finish()
    }
    // .. also need to handle the key = uint16type here
  }
}
```

Note, since we need to handle it for multiple dict key types, it might be good to extract that logic to a function that's generic over the dictionary key type like:
```
fn regex_is_match_for_dict<K: ArrowDictionaryKeyType>(
    src: &ArrayRef,
    regex: &str,
) -> Result<BooleanArray, ArrowError> {
```
([here's](https://github.com/open-telemetry/otel-arrow/blob/8b3120c2ccb0af05d8a7d872192753ee5c93c694/rust/otel-arrow-rust/src/encode/record/array.rs#L739) an example of a similar function)

rust/otel-arrow-rust/src/otap/filter.rs
```diff
@@ -0,0 +1,309 @@
...
+use serde::Deserialize;
+use snafu::OptionExt;
+use std::collections::HashSet;
+use std::sync::Arc;
+
+pub mod logs;
+
+/// struct that describes the overall requirements to use in order to filter logs
+#[derive(Debug, Clone, Deserialize)]
+pub struct LogFilter {
```
albertlockett: nit: maybe since this is prefixed by `Logs` it should be moved to ./filter/logs.rs

rust/otel-arrow-rust/src/otap/filter.rs
```diff
@@ -0,0 +1,309 @@
...
+        &self,
+        logs_payload: &OtapArrowRecords,
+        mut resource_attr_filter: BooleanArray,
+        mut log_record_filter: BooleanArray,
+        mut log_attr_filter: BooleanArray,
+    ) -> Result<(BooleanArray, BooleanArray, BooleanArray, BooleanArray)> {
+        // get the record batches we are going to filter
+        let resource_attrs = logs_payload
+            .get(ArrowPayloadType::ResourceAttrs)
+            .context(error::LogRecordNotFoundSnafu)?;
```
albertlockett: These attribute record batches are optional, so at some point we'll need to handle the case here where they don't exist

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,788 @@
...
+                .context(error::LogRecordNotFoundSnafu)?;
+            let filtered_scope_attrs = arrow::compute::filter_record_batch(scope_attrs, &filter)
+                .expect("can apply predicate on record batch with equal row length");
+            logs_payload.set(ArrowPayloadType::ScopeAttrs, filtered_scope_attrs);
+        }
+
+        Ok(logs_payload)
+    }
+
+    fn sync_up_filters(
```
jmacd, line 170: Please comment this function and all of the large function bodies below. Thanks!

rust/otap-dataflow/crates/otap/src/filter_processor.rs
```diff
@@ -0,0 +1,457 @@
...
+mod config;
+/// The URN for the filter processor
+pub const FILTER_PROCESSOR_URN: &str = "urn:otel:filter:processor";
+
+/// processor that outputs all data received to stdout
+pub struct FilterProcessor {
+    config: Config,
+}
+
+/// Factory function to create an FilterProcessor.
```
lquerel:
```suggestion
/// Factory function to create a FilterProcessor.
```

rust/otap-dataflow/crates/otap/src/filter_processor/README.md
```diff
@@ -0,0 +1,56 @@
...
+        - checkout started
+    log_record: []
+```
+
+Currently we only support log filter
+
+### Logs
+
+To filter logs you can choose to define logs to `include` or `exclude`.
+You can also choose to define both, if both are defined then then result
```
lquerel:
```suggestion
You can also choose to define both, if both are defined then the result
```

rust/otap-dataflow/crates/otap/src/filter_processor/README.md
```diff
@@ -0,0 +1,56 @@
+# Filter Proccessor
+
+Status: **WIP**
+
+This crate will contain the implementation of the filter processor.
```
lquerel, line 5: May be add a link to the filter processor documentation of the Go implementation for reference.

rust/otel-arrow-rust/src/otap/filter.rs
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/// Finds all nulls and converts them to false values
+/// null values affect the filter computation as when
+/// we perform the not operation nothing happens to
+/// the null values
+#[must_use]
+pub fn nulls_to_false(a: &BooleanArray) -> BooleanArray {
+    let valid = arrow::compute::is_not_null(a).expect("is_not_null doesn't error"); // BooleanArray with no nulls
+    arrow::compute::and_kleene(a, &valid).expect("can combine two columns with equal length") // nulls become false; trues stay true
```
lquerel, line 61: Can these two "expect" statements be triggered by a misconfiguration of the `FilterProcessor`?
If so, we should replace them with one or two error returns.
We should only use `expect` in critical cases where the error indicates an internal logic problem.

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,788 @@
...
+    include: LogMatchProperties,
+    // Exclude match properties describe logs that should be excluded from the Collector Service pipeline,
+    // all other logs should be included.
+    // If both Include and Exclude are specified, Include filtering occurs first.
+    exclude: LogMatchProperties,
+
+    // LogConditions is a list of OTTL conditions for an ottllog context.
+    // If any condition resolves to true, the log event will be dropped.
+    // Supports `and`, `or`, and `()`
+    #[allow(dead_code)]
```
lquerel, line 41: Why dead_code? If it's for a future development then you should add a ToDo.

rust/otel-arrow-rust/src/otap/filter/logs.rs
lquerel: Yes, documentation - and maybe also a name that better reflects what this method actually does :-)

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,788 @@
...
+    fn get_ids(
+        &self,
+        id_column: &Arc<dyn Array>,
+        filter: &BooleanArray,
+        column_type: &str,
+    ) -> Result<HashSet<u16>> {
+        // get ids being removed
+        // error out herre
+        let filtered_ids = arrow::compute::filter(id_column, filter)
+            .expect("can apply predicate on column with equal length");
```
lquerel: If this `expect` can be triggered by a misconfiguration or by malformed data, then it should be turned into an error.
However, if checks have already been performed upstream and logically this `expect` should never be triggered, then it’s fine to keep it.

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,788 @@
...
+            } else {
+                arrow::compute::kernels::cmp::neq(id_column, &id_scalar)
+                    .expect("can compare uint16 id column with uint16 scalar")
+            };
+            combined_id_filter = arrow::compute::or_kleene(&combined_id_filter, &id_filter)
+                .expect("can combine two boolean arrays with equal length");
+        }
+        combined_id_filter = nulls_to_false(&combined_id_filter);
+        // inverse because these are the ids we want to remove
+        arrow::compute::not(&combined_id_filter).expect("not doesn't fail")
```
lquerel: Probably ok, but see my previous comments on `expect`. In this specific context, it's possible that is ok but I prefer to double check.

lquerel (changes requested):
Great work! I've added a few comments and recommendations, especially regarding the use of `expect`, which should only be used under certain conditions (see my comments for more details).

rust/otel-arrow-rust/src/otap/filter.rs
c1ly: not possible, is_not_null never errors (https://docs.rs/arrow/latest/arrow/compute/fn.is_not_null.html) and will return a boolean array that is the same length. so we can guarantee and `a` and `valid` have equal length in the second line. I will add a comment here stating such

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,848 @@
...
+    /// get_ids() takes the id_column from a record batch and the corresponding filter
+    /// and applies it to extract all ids that match and then returns the set of ids.
+    fn get_ids(
+        &self,
+        id_column: &Arc<dyn Array>,
+        filter: &BooleanArray,
+        column_type: &str,
+    ) -> Result<HashSet<u16>> {
+        // get ids being removed
+        // error out herre
```
albertlockett:
```suggestion
        // error out here
```

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,848 @@
...
+                    .expect("can compare uint16 id column with uint16 scalar")
+            };
+            combined_id_filter = arrow::compute::or_kleene(&combined_id_filter, &id_filter)
+                .context(error::ColumnLengthMismatchSnafu)?;
+        }
+        // make sure there are no null values before we invert
+        combined_id_filter = nulls_to_false(&combined_id_filter);
+        // inverse because these are the ids we want to remove
+        // not is safe please see https://docs.rs/arrow/latest/arrow/compute/fn.not.html
+        Ok(arrow::compute::not(&combined_id_filter).expect("not doesn't fail"))
```
albertlockett, line 356: @c1ly this is interesting because on paper this is `O(id_column.len() * id_set.len())` because we're iterating the `id_column` for each id in the `id_set` to check if it's equal to `id_scalar`. 

So typically I'd think it's faster to iterate the `id_column`, check if each one is in the `id_set`, and fill in a boolean array using `BooleanBuilder`:
```rs
    let col = id_column
        .as_any()
        .downcast_ref::<UInt16Array>()
        .expect("expected UInt16Array");

    let mut builder = BooleanBuilder::with_capacity(col.len());

    for i in 0..id_column.len() {
        if col.is_null(i) {
            builder.append_null();
            continue;
        }
        let id = col.value(i);
        let keep = id_set.contains(&id) == match_id;
        builder.append_value(keep);
    }
    return not(builder.finish());
```

Although if the batch size (e.g. `id_column.len()`) is large and the number of IDs in the set is small (e.g. if the filter wasn't very selective) , using the `eq`/`neq` kernels like this is actually faster.

I was interested in this, so I wrote quick benchmark to test:
https://github.com/open-telemetry/otel-arrow/commit/b498c1446100fcb008f13bfbc807b61e72c2c86b

num_ids | batch_size | current_impl (µs) | bool_builder_impl (µs) | delta
-- | -- | -- | -- | --
5 | 128 | 2.2999 | 0.96764 | -1.33226
50 | 128 | 16.75 | 0.95873 | -15.79127
100 | 128 | 24.552 | 0.97949 | -23.57251
**5** | **1024** | 2.8131 | 6.406 | **3.5929**
50 | 1024 | 22.571 | 7.8465 | -14.7245
100 | 1024 | 45.283 | 6.6644 | -38.6186
500 | 1024 | 186.58 | 7.1438 | -179.4362
**5** | **8192** | 6.5974 | 50.539 | **43.9416**
**50** | **8192** | 62.783 | 70.643 | **7.86**
100 | 8192 | 125.18 | 68.564 | -56.616
500 | 8192 | 610.52 | 59.595 | -550.925

Do you think maybe we should have another implementation of this method that using the `BooleanBuilder` approach from that benchmark, and we have some heuristic like to choose between the implementations? E.g. if there are more if id_column.len() is large (say > 2000) and `id_set.len()` is small (say <0 10% of `id_column.len()`) we use the current approach.

Feel free to do this in a followup as well btw. I think maybe this could be part of a broader piece of followup work where we implement benchmarks for this filtering code and optimize any hotspots.

albertlockett (approved):
After discussing w/ @c1ly on standup, we've decided some of the comments I made that can be fixed in followups, so I am approving! Thanks for all your work on this @c1ly looks good

rust/otap-dataflow/crates/otap/src/filter_processor.rs
```diff
@@ -0,0 +1,457 @@
...
+                let signal = pdata.signal_type();
+                // convert to arrow records
+                let (context, payload) = pdata.into_parts();
+
+                let arrow_records: OtapArrowRecords = payload.try_into()?;
+
+                let filtered_arrow_records: OtapArrowRecords = match signal {
+                    SignalType::Metrics => {
+                        // ToDo: Add support for metrics
+                        return Ok(());
```
utpilla, line 113: For metrics and traces, should we let the data pass through instead of dropping them entirely until we have implemented filtering support for them?

```suggestion
                        // Pass through unfiltered until metrics filtering is implemented
                        arrow_records
```

c1ly:
Please see PR #1341 which adds the following fixes

- fixed edge case where optional record batches are not present
- fixed edge case where logs with no resource attribute/record attributes are not being filtered out
- fixed string column not guaranteed to be a string array when downcast to apply regex filter
- utilized booleanbuilder to optimize building the id filter

Close this PR when #1341 is approved, this PR should only be used for reference

rust/otap-dataflow/crates/otap/src/filter_processor.rs
c1ly: fixed in my followup pr #1341

rust/otel-arrow-rust/src/otap/filter/logs.rs
jmacd: Nice

utpilla: That's a very good benchmark test. It highlights the two main factors contributing to the compute costs:
- `id_set` length (or `num_ids`) which determines the number of times we allocate a new `BooleanArray` since we call `or_kleene` operation as many times as the `id_set` length
- `id_column_length` (or `batch_size`), I believe arrow's compute kernels are able to utilize SIMD operations here, but the manual iteration does not.

I think the heuristic we want would be influenced by our preferences for memory utilization vs pure compute performance. Maybe a certain kind of allocator is better for the current approach of repeated intermediate `BooleanArray` allocations.

PR #1281, opened 2025-10-14
[query-engine] Add support for defining map schema while allowing undefined keys

CodeBlanch:
Requested by @drewrelmas

## Changes

* Add support in KQL Parser for defining map schema while still allowing undefined keys

## Details

Currently if you define schema for a map then any unknown key will yield "key-not-found" errors. @drewrelmas has a user scenario where he wants to define schema for known keys but still allow custom keys to be added/removed/modified. This PR essentially adds a flag `allow_undefined_keys` to enable this behavior.

Note: There will be a follow-up PR exposing this in the bridge options.

### DIFF
--- a/rust/experimental/query_engine/kql-parser/src/kql_parser.rs
+++ b/rust/experimental/query_engine/kql-parser/src/kql_parser.rs
@@ -26,7 +26,7 @@ pub(crate) fn map_kql_errors(error: ParserError) -> ParserError {
     match error {
         ParserError::KeyNotFound { location, key } => ParserError::QueryLanguageDiagnostic {
             location,
-            diagnostic_id: "KS109",
+            diagnostic_id: "KS142",
             message: format!(
                 "The name '{key}' does not refer to any known column, table, variable or function"
             ),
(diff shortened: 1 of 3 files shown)


rust/experimental/query_engine/kql-parser/src/kql_parser.rs
```diff
@@ -22,6 +22,19 @@ impl Parser for KqlParser {
     }
 }
 
+pub(crate) fn map_kql_errors(error: ParserError) -> ParserError {
+    match error {
+        ParserError::KeyNotFound { location, key } => ParserError::QueryLanguageDiagnostic {
+            location,
+            diagnostic_id: "KS109",
+            message: format!(
+                "The name '{key}' does not refer to any known column, table, variable or function"
```
drewrelmas, line 31: nit: in [previous PRs](https://github.com/open-telemetry/otel-arrow/pull/1262#discussion_r2421256753) we talked about `name` versus `identifier`. Do you think this should be identifier too? Is this string coming directly from a `KS109` in KQL? I searched and didn't find any reference to it.

drewrelmas (approved):

rust/experimental/query_engine/kql-parser/src/kql_parser.rs
CodeBlanch: I got it from the Azure Data Explorer UI:

image

I had it as KS109 looks like it should actually be KS142. I made that change. Maybe that is why you couldn't find it?

drewrelmas: Ok if we're copying a direct error from ADX, happy to keep it

PR #1285, opened 2025-10-15
Update continuous benchmark results page

clhain:
Note: This is for the "benchmarks" branch, not main.

This PR updates the index.html page at https://open-telemetry.github.io/otel-arrow/benchmarks/continuous/ to better support the 4-test suite (e.g. all 4 results for the same metric displayed in a single chart - more similar to the go collector output https://opentelemetry.io/docs/collector/benchmarks/).

### DIFF
--- a/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
+++ b/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
@@ -614,6 +614,30 @@ mod tests {
     }
 
     #[test]
+    fn test_parse_kql_query_into_pipeline_with_attributes_schema_and_allow_undefined_keys() {
+        let run_test_success = |query: &str| {
+            parse_kql_query_into_pipeline(
+                query,
+                Some(
+                    BridgeOptions::new().with_attributes_schema(
+                        ParserMapSchema::new()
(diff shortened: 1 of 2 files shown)


lquerel (approved):
LGTM

lquerel:
@drewrelmas could you take a look at the CI issue here because we don't know what to do. @clhain can give you more details. Thanks

drewrelmas:
This is strange, nothing stands out immediately to me. Problem is in setting up `cargo-llvm-cov` and `cargo-nextest` [here](https://github.com/open-telemetry/otel-arrow/blob/877b2cb3dddef09143c7cb20164fd0ba789dd210/.github/workflows/rust-ci.yml#L43-L46). Still looking.

EDIT:

I am just now noticing the target branch for this PR. Could we consider updating the branch `benchmarks` with latest `main` content? It is possible that some of the old action versions used are no longer functional.

clhain:
Easy CLA seems to not love these co-authored changes. Per convo with Drew, the best path forward appears to be:

1. Close this out
2. Open a new PR that removes CI checks from this branch
3. Open a new PR with the html changes against this branch
4. Figure out what files are actually needed on this branch and strip it down to the minimal set (possibly just the docs folder).

Issue #1291, opened 2025-10-15
Fan-out processor component


jmacd:
The OpenTelemetry Collector model is that of receivers connected with processor-sequences connected with exporters, as we understand:

- Receivers automatically fan-out to each pipeline
- The pipelines are called in sequence
- Internally, the MutatesData flag is consulted to isolate the pipelines from each other

See e.g., https://github.com/open-telemetry/opentelemetry-collector/blob/28a5087030b5d5db895fc0ee04ce6ca39040c866/internal/fanoutconsumer/logs.go#L43

We need similar logic, or at least a configurable fan-out processor that can apply to its downstream in sequence safely cloning data.

Note: It is possible with our RETURN_DATA Interest bit, that we could arrange for a clone-free sequencing of exporters, such that each exporter will ack/nack including the data so that the fanout consumer applies without cloning. For this to be successful, we will have to prevent mutation downstream, as the model calls for these to be isolated. This could be accomplished using an Interests::MUTATION which would prevent the fanout processor from working without the use of cloning.

jmacd:
We think we can "hide" this detail in the implementation of hyper-edges and the dispatch strategy.

jmacd:
Related: https://github.com/open-telemetry/otel-arrow/pull/1295#discussion_r2457309517

In the OpenTelemetry Collector, note that fanout consumers in the Collector are "invisible", they do not make metrics IIUC. When a fanout consumer sits between a receiver and processor(s), the "produced" count of the receiver will equal the "consumed" count of each pipeline separately.

lalitb:
I would be interested on working on this @jmacd.

lalitb:
I’m still getting up to speed with the codebase and design choices, so please let me know if this approach doesn’t align with the project direction or if I’ve misunderstood anything. To keep the PR focused and easier to review, I propose to implement this in phases.

### Phase 1: Mutation-Aware Fanout Processor

  Features:
  - Mutation-aware cloning via Interests::MUTATION flag
  - OTel Collector ordering (mutators first, readonly last)
  - Last mutator optimization (saves one clone)

  ```yaml
   # Without fanout (load balancing) - this already exists (?)
   receiver → [exp1, exp2, exp3]
   # Result: exp1 gets msg1, exp2 gets msg2, exp3 gets msg3

   # With fanout (broadcasting) - Phase 1
   receiver → fanout → [exp1, exp2, exp3]
   # Result: all exporters get all messages
  ```

[...]

PR #1295, opened 2025-10-16
Retry processor (stateless) Ack/Nack

jmacd:
Replaces stateful retry processor with new stateless Ack/Nack support.

The component has been restructured with exactly same configuration used by the OTel Collector retry configuration, with 3 duration fields (in seconds) instead of a retry limit.

New tests written. Metrics updated.

Note: stores the num_items calculated from the initial data in a 3-word calldata, as a temporary measure. In the 10/21/2025 SIG  meeting we discussed placing num_items in another field of the context, potentially, that could be a separate change.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/retry_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/retry_processor.rs
@@ -12,6 +12,8 @@
 //! - Backoff multiplier
 //! ```
 
+// ToDo: Consider adding a jitter mechanism.
+
 use crate::pdata::OtapPdata;
 
 use async_trait::async_trait;
@@ -40,22 +42,39 @@ use std::time::{Duration, Instant, SystemTime};
 /// URN for the RetryProcessor processor
 pub const RETRY_PROCESSOR_URN: &str = "urn:otel:retry:processor";
 
-/// Configuration for the retry processor. Modeled on
+/// Configuration for the retry processor. Modeled exactly on
 /// https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md#retry-on-failure.
+///
+/// The calculated delay is:
+///
+///   min(max_interval, initial_interval * multiplier.pow(retry_number))
+///
+/// Retries will be attempted until max_elapsed_time has passed
+/// from the initial attempt.
 #[serde_as]
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct RetryConfig {
-    /// Initial retry interval in seconds.
+    /// Initial retry interval in seconds. This is how long the
+    /// first delay will be following the first NACK response.
+    /// This interval is multiplied by the multiplier on subsequent
+    /// retries, until it exceeds max_interval.
     #[serde_as(as = "DurationSecondsWithFrac<f64, Flexible>")]
     #[serde(default = "default_initial_interval")]
     pub initial_interval: Duration,
 
-    /// Maximum retry interval in seconds.
+    /// Maximum retry interval in seconds. This is a limit on
+    /// individual delays in the retry processor following a single
+    /// NACK failure. Prevents exponential growth when the initial
+    /// interval times the exponentiated multiplier reaches this
+    /// value.
     #[serde_as(as = "DurationSecondsWithFrac<f64, Flexible>")]
     #[serde(default = "default_max_interval")]
     pub max_interval: Duration,
 
-    /// Maximum elapsed time in seconds.
+    /// Maximum elapsed time in seconds.  This is the maximum elapsed
+    /// wall time for the entire request, beginning when the retry
+    /// processor first sees it. Retries will not be scheduled if they
+    /// would begin after this many seconds from the start.
     #[serde_as(as = "DurationSecondsWithFrac<f64, Flexible>")]
     #[serde(default = "default_max_elapsed_time")]
     pub max_elapsed_time: Duration,
@@ -368,7 +387,10 @@ struct RetryState {
     /// Number of retry attempts so far (0 = first attempt, 1+ = retries).
     retries: u64,
 
-    /// Deadline for the retry operation.
+    /// Deadline for the retry operation.  Note this is an f64 because
+    /// it's the only 64-bit value we can get that is lossless. The
+    /// SystemTime::as_millis() and other APIs for integer fractional
+    /// seconds return u128.
     deadline: f64,
 
     /// Item count


rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -323,208 +234,206 @@ pub fn create_retry_processor(
...
+
+    /// Item count
+    num_items: u64,
 }
 
 impl RetryState {
-    #[cfg(test)]
-    fn new(id: u64) -> Self {
-        Self { id }
+    fn new(num_items: u64, deadline: f64) -> Self {
```
utpilla:
```suggestion
    const fn new(num_items: u64, deadline: f64) -> Self {
```

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -323,208 +234,206 @@ pub fn create_retry_processor(
...
+            rstate.into(),
+            &mut rereq,
+        );
 
-        let mut removed = 0u64;
-        for id in expired_ids {
-            if self.pending_messages.remove(&id).is_some() {
-                log::warn!("Removed expired message with ID: {id}");
-                removed += 1;
+        self.metrics.add_retry_attempts(signal, num_items);
```
utpilla: Should this be `1` and not `num_items`?

jmacd: I removed the parameter from add_retry_attempts.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -225,53 +182,17 @@ impl RetryProcessorMetrics {
...
-    fn default() -> Self {
-        Self {
-            max_retries: 3,
-            initial_retry_delay_ms: 1000,
-            max_retry_delay_ms: 30000,
-            backoff_multiplier: 2.0,
-            max_pending_messages: 10000,
-            cleanup_interval_secs: 60,
+    /// Increment retry_attempts for the given signal by 1.
+    pub fn add_retry_attempts(&mut self, st: SignalType) {
```
utpilla: Maybe use the term `increment` instead of `add` now that we only add `1` to the value? Or we could also just let it be to stay consistent with the other metric update method names.

```suggestion
    pub fn increment_retry_attempts(&mut self, st: SignalType) {
```

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -323,208 +234,212 @@ pub fn create_retry_processor(
...
-                    id,
-                    self.config.max_retries,
-                    pending.last_error
-                );
-                if let Some(m) = self.metrics.as_mut() {
-                    m.msgs_dropped_exceeded_retries.inc();
-                }
+        effect_handler.notify_ack(ack).await?;
+
+        // Forward the ACK upstream to complete the request chain
```
utpilla: The comment should be above the `notify_ack` call.

```suggestion
        // Forward the ACK upstream to complete the request chain
        effect_handler.notify_ack(ack).await?;

```

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -323,208 +234,212 @@ pub fn create_retry_processor(
...
+        match calldata.try_into() {
+            Err(_err) => {
+                // Malformed context error: we don't know what this is.
             }
-        } else {
-            log::warn!("Attempted to handle nack for non-existent message with ID: {id}");
-        }
+            Ok(RetryState { num_items, .. }) => {
+                // The producer and consumer both succeed.
+                self.metrics.add_consumed_success(signal, num_items);
```
utpilla: Shouldn't this metric be updated in the `process` method when you receive `Message::PData`?

Maybe I don't understand these metrics correctly. It seems off that both `consumed_success` and `produced_success` would always have the same value.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -323,208 +234,212 @@ pub fn create_retry_processor(
...
             }
+            Ok(retry) => retry,
+        };
+
+        // Regardless of the next step, this NACK counts as a producer
+        // refusal.
+        self.metrics.add_produced_refused(signal, rstate.num_items);
+
+        // Check for missing payload, we won't retry an empty request.
+        if nack.refused.is_empty() {
```
utpilla, line 501: Do we expect NACK messages without a payload given that we subscribe to `Interests::RETURN_DATA` in the `process` method?

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -323,208 +234,212 @@ pub fn create_retry_processor(
...
-                }
-                Err(e) => {
-                    if let Some(m) = self.metrics.as_mut() {
-                        m.add_produced_refused(signal, items);
-                    }
-                    return Err(e.into());
-                }
-            }
+        // Retry attempt: calculate delayed retry and increment
+        let delay_secs = (self.config.initial_interval.as_secs_f64()
```
utpilla: We could have the `RetryProcessor` struct pre-calculate the delay intervals into a `vec` typed field which could be indexed by the retry count to avoid the repeated floating-point calculations. We would only have to calculate entries up to the maximum interval.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
jmacd: We do not expect NACK messages without payload. It would be an internal error condition.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -323,208 +234,212 @@ pub fn create_retry_processor(
...
+            * self.config.multiplier.powi(rstate.retries as i32))
+        .min(self.config.max_interval.as_secs_f64());
+
+        // Compute the delay.
+        let now_s = SystemTime::now();
```
utpilla: It might be cleaner to simply use the convenient `now_f64()` method instead of creating a `SystemTime` instance.

Something like this:
```rust
// Compute the delay.
let now_i = Instant::now();
let next_retry_time_i = now_i + Duration::from_secs_f64(delay_secs);

if rstate.deadline <= now_f64() + delay_secs {
    nack.reason = format!("final retry: {}", nack.reason);
    effect_handler.notify_nack(nack).await?;
    // The caller has refused as much as we'll let them.
    self.metrics.add_consumed_refused(signal, rstate.num_items);
    return Ok(());
}
```

rust/otap-dataflow/crates/otap/src/retry_processor.rs
jmacd: My understanding of the Collector metrics RFC is that `consumed_success` counts the number we consumed that eventually succeeded, and `produce_success` for the number we produced. These will increment at the same count, but `produced_rejected` will increment for each time there is a retry, so they contribute to different rates.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
utpilla: Do we need this check then?

rust/otap-dataflow/crates/otap/src/retry_processor.rs
utpilla: But if `produced_success` is always going to be equal to `consumed_success`, then maybe we just need one metric here. I don't see any additional value being added by tracking a duplicate metric in the form of `consumed_sucess`.

utpilla: If we want to keep both metrics, I think they at least need to be updated at different places in the code?

```rust
self.metrics.add_produced_success(signal, num_items); // Should be updated before forwarding the ACK upstream

effect_handler.notify_ack(ack).await?;

...
self.metrics.add_consumed_success(signal, num_items); // Should be updated after forwarding the ACK
```

utpilla (approved):
Left a suggestion: https://github.com/open-telemetry/otel-arrow/pull/1295#discussion_r2453568264

rust/otap-dataflow/crates/otap/src/retry_processor.rs
jmacd: Done! Thank you, that makes sense.

Another way I was thinking about this, is that if this were rewritten with "success", "failure", and "refused" as attributes instead of part in the metric name, then it would look a little more ordinary to have two metrics incremented with the same attribute values in one location. Either form, the attribute form or the metrics-looking-redundant form, is meant to facilitate a straightforward query experience where to define the success-rate of a consumer (ignoring retries) you use the consumer-side metrics (e.g., consumer{outcome=success}/consumer{outcome=.}). To evaluate the producer (considering retries), you use the producer-side metrics (e.g., producer{outcome=success}/producer{outcome=.}).

rust/otap-dataflow/crates/otap/src/retry_processor.rs
jmacd: Not sure if I follow -- just because something shouldn't happen won't prevent it from, in case of bugs or what not. I was checking this because we checked the same condition when the data first arrived. If the same is not still true, the payload was lost, I would say through "mis-handling". If a caller wants to use RETURN_DATA, they have to prepare for the case where the data does not return intact anyway.

jmacd: #1339

rust/otap-dataflow/crates/otap/src/retry_processor.rs
jmacd: Done! I worried a bit about memory, so I have also added more thorough config validation and a hard limit of 1000 distinct values before max_interval kicks in.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
jmacd: Done!

rust/otap-dataflow/crates/otap/src/retry_processor.rs
utpilla: > just because something shouldn't happen won't prevent it from, in case of bugs or what not.
 
Okay so basically, we should be prepared to receive empty payload here. That's what I wanted to know when I asked if we should expect this. I guess "expect" was not the best word to be used here 😅

I was mainly trying to figure out if it was a redundant check.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -73,15 +33,150 @@ use otap_df_telemetry::instrument::Counter;
...
+
+    /// Maximum retry interval in seconds.
+    #[serde_as(as = "DurationSecondsWithFrac<f64, Flexible>")]
+    #[serde(default = "default_max_interval")]
+    pub max_interval: Duration,
+
+    /// Maximum elapsed time in seconds.
+    #[serde_as(as = "DurationSecondsWithFrac<f64, Flexible>")]
+    #[serde(default = "default_max_elapsed_time")]
+    pub max_elapsed_time: Duration,
```
lquerel: Getting clarification on the difference between these two parameters would be good.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -73,15 +33,150 @@ use otap_df_telemetry::instrument::Counter;
...
+            initial_interval: default_initial_interval(),
+            max_elapsed_time: default_max_elapsed_time(),
+            multiplier: default_multiplier(),
+        }
+    }
+}
+
+impl RetryConfig {
+    /// Computes the maximum retry count by simulation. The configuration
+    /// combines exponential growth with a limit, making it difficult tnnnno
```
lquerel:
```suggestion
    /// combines exponential growth with a limit, making it difficult to
```

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -73,15 +33,154 @@ use otap_df_telemetry::instrument::Counter;
...
```
utpilla:
```suggestion
    /// combines exponential growth with a limit, making it difficult to
```

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -73,15 +33,150 @@ use otap_df_telemetry::instrument::Counter;
...
 
 /// URN for the RetryProcessor processor
-pub const RETRY_PROCESSOR_URN: &str = "urn:otap:processor:retry_processor";
+pub const RETRY_PROCESSOR_URN: &str = "urn:otel:retry:processor";
+
+/// Configuration for the retry processor. Modeled on
+/// https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md#retry-on-failure.
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct RetryConfig {
```
lquerel, line 56: Not for this PR, but eventually, shouldn't we introduce a jitter mechanism?

rust/otap-dataflow/crates/otap/src/retry_processor.rs
```diff
@@ -302,229 +334,241 @@ pub fn create_retry_processor(
...
+
+/// State tracking for retry attempts, sized for Context8u8.
+#[derive(Debug, Clone, Serialize, Deserialize)]
 struct RetryState {
-    id: u64,
+    /// Number of retry attempts so far (0 = first attempt, 1+ = retries).
+    retries: u64,
+
+    /// Deadline for the retry operation.
+    deadline: f64,
```
lquerel, line 394: What is the rational behind a f64 instead of u64 for the deadline?

rust/otap-dataflow/crates/otap/src/retry_processor.rs
jmacd: #1291 is related.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
jmacd: /// Deadline for the retry operation.  Note this is an f64 because
    /// it's the only 64-bit value we can get that is lossless. The
    /// SystemTime::as_millis() and other APIs for integer fractional
    /// seconds return u128.

I didn't want to round-up to the nearest integer second, and I didn't want to deal with u128 values being truncated to u64 values, and I didn't want to think about splitting u128 values into two Context8u8 slots.

I would be glad to see this improve, just not sure what's cleanest/best.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
jmacd: Yes. Added a ToDo.

rust/otap-dataflow/crates/otap/src/retry_processor.rs
jmacd: Added a bunch of comments.

```
/// The calculated delay is:
///
///   min(max_interval, initial_interval * multiplier.pow(retry_number))
///
/// Retries will be attempted until max_elapsed_time has passed
/// from the initial attempt.
```

PR #1303, opened 2025-10-16
Batch processor metrics set: non-optional

jmacd:
This is a cleanup. Makes the MetricSet non-optional using the default registry, which removes a bunch of `if let Some(_)` phrases.

Part of #1304

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,6 +250,7 @@ repository](https://github.com/open-telemetry/community/blob/main/community-memb
 ### Approvers
 
 - [Lei Huang](https://github.com/v0y4g3r), Greptime
+- [Utkarsh Umesan Pillai](https://github.com/utpilla), Microsoft
 
 For more information about the approver role, see the [community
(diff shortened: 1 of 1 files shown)


lquerel (approved):
LGTM

rust/otap-dataflow/crates/telemetry/src/lib.rs
```diff
@@ -38,6 +38,10 @@ pub mod registry;
 pub mod reporter;
 pub mod semconv;
 
+// Note! I can't get #[cfg(test)] to work here!
+// #[cfg(test)]
+pub mod testing;
```
jmacd: I think I see -- can't use workspace dependencies for siblings without test features, somehow.

jmacd: Filed #1304.

jmacd:
I'm surprised that these changes are not very well covered, because there are so many tests!
Since this is a mechanical change, ignoring coverage.

Issue #1302, opened 2025-10-16
[Dataflow Engine] Investigate missing metrics


clhain:
Please assign to me.

I'm looking into operational metrics exposed by the telemetry endpoint and am noticing some missing items that 'should' be there based on the config and according to the telemetry/live-schema endpoint.

The setup:
```
                 +-------------------+      +-------------------+
                 | df_engine (OTLP)  |      | df_engine (OTAP)  |
                 |  Load Generator   |      |  Load Generator   |
                 +--------+----------+      +--------+----------+
                          |                          |
                          |                          |
                          v                          v
                      +---------------------------------------------+
                      |               df_engine                     |
                      |       (Receiver + Proc + Exporters)         |
                      |                                             |
                      |   +----------------+    +----------------+  |
                      |   | OTLP Receiver  |    | OTAP Receiver  |  |
                      |   +-------+--------+    +--------+-------+  |
                      |           |                     |          |
                      |           v                     v          |
                      |   +----------------+    +----------------+  |
                      |   | OTLP Attr Proc |    | OTAP Attr Proc |  |
                      |   +-------+--------+    +--------+-------+  |
                      |           |                     |          |
                      |           v                     v          |
                      |   +----------------+    +----------------+  |
                      |   | OTLP Exporter  |    | OTAP Exporter  |  |
                      |   +-------+--------+    +--------+-------+  |
                      +-----------|---------------------|-----------+
                                  |                     |
                                  v                     v
                 +--------------------------+     +--------------------------+
                 | df_engine Backend (OTLP) |     | df_engine Backend (OTAP) |
                 +--------------------------+     +--------------------------+
```

* I see metrics for both fake-signal generator (in the loadgen) and perf_exporter (in the backend), we've been using these in benchmarks and they look fine.
* I see set=exporter.pdata metrics for the main df_engine process for both the otap and otlp exporters (I also see them for the perf exporter in the backend and the exporters in the loadgen), these look good.
* I do not see any set=otlp.receiver.metrics despite the presence of a functioning otlp receiver that's definitely getting data in
* I don't see comparable otap.receiver.metrics registered in the telemetry/live-schema endpoint, they may not exist yet, I haven't looked.
* I do not see any set=attributes.processor.metrics despite the presence of 2 attribute processors in the main df_engine process that are in-path and definitely seeing data.
* Separately I'm investigating getting the batch processor working...  it may or may not be suffering the same issues.

[...]

Issue #1305, opened 2025-10-17
Periodic telemetry API improvement


clhain:
Related to #1302 

Currently, receiver/processor/exporter implementations are required to remember to call effect_handler.start_periodic_telemetry(), and if not done, there are no metrics surfaced until the pipeline is shutdown. This is something that can/should probably be handled at a lower level in the engine. 

This issue is to investigate how to handle this correctly.

The API will likely need to support use-cases like perf exporter where the collection interval is specified as part of the node's config:
https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/otap/src/perf_exporter/exporter.rs#L135

PR #1306, opened 2025-10-17
add start_periodic_telemetry calls in various processors and receivers

clhain:
Fixes https://github.com/open-telemetry/otel-arrow/issues/1302

This PR adds missing calls to start_periodic_telemetry in nodes which have an existing metrics implementation, allowing those metrics to be exported prior to shutdown.

df_engine::shared::receiver and df_engine::local::processor were missing the start_periodic_telemetry method definition, so it was added to those locations.

https://github.com/open-telemetry/otel-arrow/issues/1305 will further revamp the api for this. Since it's likely that effort will affect how the timers are canceled on shutdown, and given the current inconsistency in handling the shutdown signal between node implementations, I ignored this cleanup for now (but happy to add it in if it makes sense).

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/lib.rs
+++ b/rust/otap-dataflow/crates/engine/src/lib.rs
@@ -372,7 +372,6 @@ impl<PData: 'static + Clone + Debug> PipelineFactory<PData> {
                             .map(|id| pipeline.get_node(id.index).expect("ok"))
                     })
                     .ok_or_else(|| Error::UnknownNode { node: name.clone() })?;
-
                 dest_nodes.push(node);
             }
 


rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -282,6 +285,11 @@ impl local::Processor<OtapPdata> for AttributesProcessor {
         msg: Message<OtapPdata>,
         effect_handler: &mut local::EffectHandler<OtapPdata>,
     ) -> Result<(), EngineError> {
+        // Start periodic telemetry collection
+        let _ = effect_handler
+            .start_periodic_telemetry(Duration::from_secs(1))
+            .await?;
+
```
lquerel: The `process` method is called for every PData message, so this statement schedules a recurring timer for each message. There is definitely an issue in the current `Processor` interface (and to a lesser extent in the `Receiver`/`Exporter` interfaces) that makes the scheduling of this timer error-prone. I’ll work on it as soon as possible.
In the meantime, I think we should introduce a flag to prevent scheduling a new recurring timer for every received message.

clhain:
Per discussion, removed the timer starts in the processors themselves and added it temporarily to the ProcessorWrapper start method.

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
```diff
@@ -131,6 +131,11 @@ impl shared::Receiver<OtapPdata> for OTLPReceiver {
             .add_service(metrics_service_server)
             .add_service(trace_service_server);
 
+        // Start periodic telemetry collection
+        let _ = effect_handler
```
utpilla: It'd be good to cancel the telemetry collection timer when being shut down. Instead of discarding the result here, we could obtain a reference to the handle and call `cancel()` on it when processing `NodeControlMsg::Shutdown`.

Check #1155 for reference.

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
clhain: This is resolved by moving the start calls to the wrapper before it enters the loop.

rust/otap-dataflow/crates/engine/src/processor.rs
```diff
@@ -238,6 +239,12 @@ impl<PData> ProcessorWrapper<PData> {
                 effect_handler
                     .core
                     .set_pipeline_ctrl_msg_sender(pipeline_ctrl_msg_tx);
+
+                // Start periodic telemetry collection
+                let _ = effect_handler
```
jmacd: Should this case save the cancel handler, like for otlp_receiver?

lquerel (approved):
LGTM

PR #1321, opened 2025-10-20
OTLP receiver: new Settings with support for wait_for_result

### LINKED ISSUE #1311: OTLP receiver Ack/Nack harms benchmarks too severely
Something is wrong. We will disable backpressure while investigating.
https://open-telemetry.github.io/otel-arrow/benchmarks/continuous/

jmacd:
Introduces `wait_for_result` in the OTLP receiver.

Fixes #1311 

As I begun to understand this code, I could see ways to shorten it while adding new configuration support.

The new struct `Settings` follows collector practice where there is a widely-used non-configuration object used to pass startup conditions, this follows that convention.

Comments added: it was not clear at first why we construct Grpc<.> and OtapBatchService<.> in each request.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
@@ -411,11 +411,10 @@ impl tower_service::Service<Request<Body>> for MetricsServiceServer {
     type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;
 
     fn call(&mut self, req: Request<Body>) -> Self::Future {
-        let common = self.common.clone();
         match req.uri().path() {
             super::METRICS_SERVICE_EXPORT_PATH => {
                 let common = self.common.clone();
-                let mut grpc = new_grpc(SignalType::Logs, common.settings);
+                let mut grpc = new_grpc(SignalType::Metrics, common.settings);
                 let service = OtapBatchService::new(common.effect_handler, common.state);
                 Box::pin(async move { Ok(grpc.unary(service, req).await) })
             }
@@ -455,11 +454,10 @@ impl tower_service::Service<Request<Body>> for TraceServiceServer {
     type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;
 
     fn call(&mut self, req: Request<Body>) -> Self::Future {
-        let common = self.common.clone();
         match req.uri().path() {
             super::TRACE_SERVICE_EXPORT_PATH => {
                 let common = self.common.clone();
-                let mut grpc = new_grpc(SignalType::Logs, common.settings);
+                let mut grpc = new_grpc(SignalType::Traces, common.settings);
                 let service = OtapBatchService::new(common.effect_handler, common.state);
                 Box::pin(async move { Ok(grpc.unary(service, req).await) })
             }
--- a/rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
+++ b/rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
@@ -601,7 +601,7 @@ mod tests {
     }
 
     #[test]
-    fn test_otlp_receiver() {
+    fn test_otlp_receiver_ack() {
         let test_runtime = TestRuntime::new();
 
         let grpc_addr = "127.0.0.1";


rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
```diff
@@ -83,6 +85,19 @@ impl SharedState {
     }
 }
 
+/// Common settings for OTLP receivers.
+#[derive(Clone, Debug)]
+pub struct Settings {
+    /// Maximum concurrency
+    pub max_concurrent_requests: usize,
```
utpilla, line 92: I believe this is the max concurrent requests per core instance of the receiver and not across all the receiver instances. It would be good to be clear about (preferably in the name of the setting).

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
```diff
@@ -83,6 +85,19 @@ impl SharedState {
...
+    /// Whether the receiver should wait.
```
utpilla, line 93: We’ve agreed, based on the Slack discussion, to use the name `wait_for_result` since it aligns with the existing setting and is familiar to users. However, it would be helpful to provide a more explicit explanation of its behavior and expectations.

Something like this maybe?
```
When enabled, the receiver will not send a response 
until the immediate downstream component has acknowledged receipt of the data. 
This does not guarantee that data has been fully processed or successfully 
exported to the final destination.
```

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
```diff
@@ -40,18 +42,33 @@ pub const OTLP_RECEIVER_URN: &str = "urn:otel:otlp:receiver";
 #[derive(Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct Config {
+    /// The endpoint details: protocol, name, port.
     listening_addr: SocketAddr,
+
+    /// Compression methods accepted
+    /// TODO: this should be Vec<CompressionMethod>, with separate settings
```
utpilla: Or we could use a tuple or an array of size 2 to semantically convey that it would only ever have two values?
```suggestion
    /// TODO: this should be (CompressionMethod, CompressionMethod), with separate settings
```

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
```diff
@@ -40,18 +42,33 @@ pub const OTLP_RECEIVER_URN: &str = "urn:otel:otlp:receiver";
...
+    /// (as tonic supports) for request and response compression.
     compression_method: Option<CompressionMethod>,
 
     /// Maximum number of concurrent (in-flight) requests (default: 1000)
     #[serde(default = "default_max_concurrent_requests")]
     max_concurrent_requests: usize,
+
+    /// Whether to wait for the result (default: true)
+    /// Note when wait_for_result=false, it is impossible to
+    /// see a failure, errors are efffectively suppressed.
```
utpilla:
```suggestion
    /// see a failure, errors are effectively suppressed.
```

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
lquerel: It is indeed a limit that applies per core.

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
```diff
@@ -360,25 +368,16 @@ impl tower_service::Service<Request<Body>> for LogsServiceServer {
     type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;
 
     fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let common = self.common.clone();
```
utpilla: Since we are cloning `ServerCommon` on the hot path, we could consider changing the type to be an `Arc<ServerCommonInner>` with `ServerCommonInner` having the individual fields. That way we can reduce the cloning cost.

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
lquerel: That's not how our retry/ack/nack mechanism works. Ack/Nack messages are propagated back to the receivers not by the immediate downstream node, but by any downstream node that participates in the retry logic. A retry processor can very well be at the end of the pipeline and not directly the successor of the receiver in the DAG.

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
utpilla: We should consider doing something similar for `EffectHandler` as well.

jmacd: I see. I have this diff:

```
index d2fa4ea9..13a1ca7e 100644
--- a/rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
@@ -349,7 +349,7 @@ impl ServerCommon {
 #[derive(Clone)]
 pub struct LogsServiceServer {
     /// common support for OTLP servers
-    pub common: ServerCommon,
+    pub common: Arc<ServerCommon>,
 }
 
 impl LogsServiceServer {
@@ -357,7 +357,7 @@ impl LogsServiceServer {
     #[must_use]
     pub fn new(effect_handler: EffectHandler<OtapPdata>, settings: &Settings) -> Self {
         Self {
-            common: ServerCommon::new(effect_handler, settings),
+            common: Arc::new(ServerCommon::new(effect_handler, settings)),
         }
     }
 }
@@ -371,9 +371,9 @@ impl tower_service::Service<Request<Body>> for LogsServiceServer {
         let common = self.common.clone();
         match req.uri().path() {
             super::LOGS_SERVICE_EXPORT_PATH => Box::pin(async move {
-                Ok(new_grpc(SignalType::Logs, common.settings)
+                Ok(new_grpc(SignalType::Logs, common.settings.clone())
                     .unary(
-                        OtapBatchService::new(common.effect_handler, common.state),
+                        OtapBatchService::new(common.effect_handler.clone(), common.state.clone()),
                         req,
                     )
                     .await)
@@ -395,7 +395,7 @@ impl NamedService for LogsServiceServer {
 #[derive(Clone)]
 pub struct MetricsServiceServer {
     /// common support for OTLP servers
-    pub common: ServerCommon,
+    pub common: Arc<ServerCommon>,
 }
 
 impl MetricsServiceServer {
@@ -403,7 +403,7 @@ impl MetricsServiceServer {
     #[must_use]
     pub fn new(effect_handler: EffectHandler<OtapPdata>, settings: &Settings) -> Self {
         Self {
-            common: ServerCommon::new(effect_handler, settings),
+            common: Arc::new(ServerCommon::new(effect_handler, settings)),
         }
     }
 }
@@ -417,9 +417,9 @@ impl tower_service::Service<Request<Body>> for MetricsServiceServer {
         let common = self.common.clone();
         match req.uri().path() {
             super::METRICS_SERVICE_EXPORT_PATH => Box::pin(async move {
-                Ok(new_grpc(SignalType::Metrics, common.settings)
+                Ok(new_grpc(SignalType::Metrics, common.settings.clone())
                     .unary(
-                        OtapBatchService::new(common.effect_handler, common.state),
+                        OtapBatchService::new(common.effect_handler.clone(), common.state.clone()),
                         req,
                     )
                     .await)
@@ -441,7 +441,7 @@ impl NamedService for MetricsServiceServer {
 #[derive(Clone)]
 pub struct TraceServiceServer {
     /// common support for OTLP servers
-    pub common: ServerCommon,
+    pub common: Arc<ServerCommon>,
 }
 
 impl TraceServiceServer {
@@ -449,7 +449,7 @@ impl TraceServiceServer {
     #[must_use]
     pub fn new(effect_handler: EffectHandler<OtapPdata>, settings: &Settings) -> Self {
         Self {
-            common: ServerCommon::new(effect_handler, settings),
+            common: Arc::new(ServerCommon::new(effect_handler, settings)),
         }
     }
 }
@@ -463,9 +463,9 @@ impl tower_service::Service<Request<Body>> for TraceServiceServer {
         let common = self.common.clone();
         match req.uri().path() {
             super::TRACE_SERVICE_EXPORT_PATH => Box::pin(async move {
-                Ok(new_grpc(SignalType::Traces, common.settings)
+                Ok(new_grpc(SignalType::Traces, common.settings.clone())
                     .unary(
-                        OtapBatchService::new(common.effect_handler, common.state),
+                        OtapBatchService::new(common.effect_handler.clone(), common.state.clone()),
                         req,
                     )
                     .await)
```

and I realize what you mean about effect_handler now. If I only wrap the ServerCommon in an Arc<>, then I end up cloning all the same things. I think we should defer this change and do it more consistently across the codebase.

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
jmacd: added "/// Maximum concurrent requests per receiver instance (per core).".

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
utpilla: > I think we should defer this change and do it more consistently across the codebase.

I'm totally okay with that.

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
utpilla: You might have missed pushing the commit. It doesn't show the updated doc comment.

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
```diff
@@ -360,25 +368,16 @@ impl tower_service::Service<Request<Body>> for LogsServiceServer {
...
         match req.uri().path() {
-            super::LOGS_SERVICE_EXPORT_PATH => {
-                let effect_handler = self.effect_handler.clone();
-                let state = self.state.clone();
-                let accept_compression_encodings = self.accept_compression_encodings;
-                let send_compression_encodings = self.send_compression_encodings;
-                Box::pin(async move {
-                    let res = handle_service_request(
+            super::LOGS_SERVICE_EXPORT_PATH => Box::pin(async move {
+                Ok(new_grpc(SignalType::Logs, common.settings)
```
utpilla: Since we are boxing the async block, we can reduce its heap allocation by not having it capture `ServerCommon`. We could create `grpc` and `service` outside of the async block to do that.

Something like this:

```rust
fn call(&mut self, req: Request<Body>) -> Self::Future {
    match req.uri().path() {
        super::LOGS_SERVICE_EXPORT_PATH => {
            let common = self.common.clone();
            let mut grpc = new_grpc(SignalType::Logs, common.settings);
            let service = OtapBatchService::new(common.effect_handler, common.state);
            Box::pin(async move {
                Ok(grpc.unary(service, req).await)
            })
        },
        _ => Box::pin(async move { Ok(unimplemented_resp()) }),
    }
}
```

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
jmacd: /// When enabled, the receiver will not send a response until the
    /// immediate downstream component has acknowledged receipt of the
    /// data.  This does not guarantee that data has been fully
    /// processed or successfully exported to the final destination,
    /// since components are able acknowledge early.

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
jmacd: Nice!

lquerel (approved):
LGTM.
For now, I don't see a better approach around `new_grpc`. I think it would be good to see the impact of this approach with a profiler when sending a high volume of traffic.

PR #1330, opened 2025-10-22
DelayData implementation details & processor test support

jmacd:
These are the parts of #1295 which are not part of the `retry_processor` specifically.
These are the other changes I needed to complete the retry processor, including:

- EffectHandlerCore and {local,shared}::processor::EffectHandler have a delay_data() API
- SendError<T> has an inner() method to get the data back so that delay_data() can return unsent data
- CallData support for f64 fields, retry_processor stores SystemTime::as_f64_secs()
- CallData widened to 3 words (from 2); the retry_processor will store { retry_count: u64; expiration: f64; num_items: u64 } in support of correct metrics on the return path
- engine/src/testing/processor.rs support for setting the effect handler
- OtapPdata method for current_context(); for retry_processor wants to peek at its own call data in the DelayedData handler


utpilla (approved):

jmacd:
All of this code is test-covered in #1295.

PR #1341, opened 2025-10-23
[otap-dataflow]basic filter processor logs followup

c1ly:
Add a basic filter processor that is based on the golang version minus the OTTL support

Update to #1279 
- fixed edge case where optional record batches are not present
- fixed edge case where logs with no resource attribute/record attributes are not being filtered out
- fixed string column not guaranteed to be a string array when downcast to apply regex filter
- utilized booleanbuilder to optimize building the id filter

Add support for log filtering

Defined to following rust structs that will allow us to filter on 
- resource attributes
- log record attributes
- log record bodies
- log record severity text
- log record severity number (optional)

```rust
/// struct that describes the overall requirements to use in order to filter logs
pub struct LogFilter {
    // Include match properties describe logs that should be included in the Collector Service pipeline,
    // all other logs should be dropped from further processing.
    include: LogMatchProperties,
    // Exclude match properties describe logs that should be excluded from the Collector Service pipeline,
    // all other logs should be included.
    exclude: LogMatchProperties,

    // LogConditions is a list of OTTL conditions for an ottllog context.
    // If any condition resolves to true, the log event will be dropped.
    // Supports `and`, `or`, and `()`
    #[allow(dead_code)]
    log_record: Vec<String>,
}

/// LogMatchProperties specifies the set of properties in a log to match against and the type of string pattern matching to use.
pub struct LogMatchProperties {
    // LogMatchType specifies the type of matching desired
    match_type: LogMatchType,

    // ResourceAttributes defines a list of possible resource attributes to match logs against.

    // RecordAttributes defines a list of possible record attributes to match logs against.
    record_attributes: Vec<KeyValue>,

    // SeverityTexts is a list of strings that the LogRecord's severity text field must match
    // against.
    severity_texts: Vec<String>,

    // SeverityNumberProperties defines how to match against a log record's SeverityNumber, if defined.
    severity_number: Option<LogSeverityNumberMatchProperties>,

    // LogBodies is a list of values that the LogRecord's body field must match
    // against.
    bodies: Vec<AnyValue>,
}

/// LogSeverityNumberMatchProperties specifies the requirements needed to match on the log severity field
pub struct LogSeverityNumberMatchProperties {
    // Min is the minimum severity needed for the log record to match.
    // This corresponds to the short names specified here:
    // https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#displaying-severity
    // this field is case-insensitive ("INFO" == "info")
    min: i32,
    // MatchUndefined lets logs records with "unknown" severity match.
    // If MinSeverity is not set, this field is ignored, as fields are not matched based on severity.
    match_undefined: bool,
}
```

Future PR
- add trace and metric support

Part of #1213

### DIFF
--- a/rust/otel-arrow-rust/src/otap/filter.rs
+++ b/rust/otel-arrow-rust/src/otap/filter.rs
@@ -35,7 +35,7 @@ pub enum MatchType {
     Regexp,
 }
 
-fn default_match_type() -> MatchType {
+const fn default_match_type() -> MatchType {
     MatchType::Strict
 }
 


rust/otel-arrow-rust/src/otap/filter.rs
```diff
@@ -0,0 +1,271 @@
...
+) -> Result<BooleanArray> {
+    if (id_column.len() >= ID_COLUMN_LENGTH_MIN_THRESHOLD)
+        && ((id_set.len() as f64 / id_column.len() as f64) <= IDS_PERCENTAGE_MAX_THRESHOLD)
+    {
+        let mut combined_id_filter = BooleanArray::new_null(id_column.len());
+        // build id filter using the id hashset
+        for id in id_set {
+            let id_scalar = UInt16Array::new_scalar(id);
+            let id_filter = arrow::compute::kernels::cmp::eq(id_column, &id_scalar)
+                .expect("can compare uint16 id column with uint16 scalar");
```
lquerel, line 201: In general, it would be good to add a comment above or next to each `expect` to explain why it's safe to use `expect` in this context.

lquerel (approved):
Made a comment around `expect`. Otherwise LGTM.
@albertlockett when you have time, could you review this PR. Thanks.

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,813 @@
...
+        // the ids should show up self.resource_attr.len() times otherwise they don't have all the required attributes
+        let ids = arrow::compute::filter(&parent_id_column, &attributes_filter)
+            .context(error::ColumnLengthMismatchSnafu)?;
+        // extract correct ids
+        let ids = ids
+            .as_any()
+            .downcast_ref::<UInt16Array>()
+            .expect("array can be downcast to UInt16Array");
+        // remove null values
+        let ids: Vec<u16> = ids.iter().flatten().collect();
```
utpilla: We can avoid this allocation and also initialize the HashMap with the right capacity,

```rust
let mut ids_counted: HashMap<u16, usize> = HashMap::with_capacity(ids.len());
// since we require that all the resource attributes match we use the count of the ids extracted to determine a full match
// a full match should meant that the amount of times a id appears is equal the number of resource attributes we want to match on
for id in ids.iter().flatten() {
    *ids_counted.entry(id).or_default() += 1;
}
```

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,813 @@
...
+    /// Creates a booleanarray that will filter a log record record batch based on the
+    /// defined severity_text, log body, and severity_number (if definied). A log record should match all
+    /// defined requirements.
+    fn get_log_record_filter(&self, logs_payload: &OtapArrowRecords) -> Result<BooleanArray> {
+        let log_records = logs_payload
+            .get(ArrowPayloadType::Logs)
+            .context(error::LogRecordNotFoundSnafu)?;
+        let num_rows = log_records.num_rows();
+        // create filter for severity texts
+        let mut filter: BooleanArray = vec![true; num_rows].into();
```
utpilla: There are several places where we create a `Vec<bool>` (either all `true` or all false) only to convert it into a `BooleanArray`. I think it should be more efficient to use arrow's `BooleanBuffer` here.

`BooleanBuffer::new_set` creates a buffer with all values as `true` and `BooleanBuffer::new_unset` creates a buffer with all values as `false`.

```diff
- vec![true; record_batch.num_rows()].into();

+ let boolean_buffer = BooleanBuffer::new_set(record_batch.num_rows());
+ let boolean_array = BooleanArray::from(boolean_buffer);
```

rust/otel-arrow-rust/src/otap/filter.rs
```diff
@@ -0,0 +1,277 @@
...
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum MatchType {
+    /// match on the string values exactly how they are defined
+    Strict,
+    /// apply string values as a regexp
+    Regexp,
+}
+
+fn default_match_type() -> MatchType {
```
utpilla:
```suggestion
const fn default_match_type() -> MatchType {
```

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,820 @@
...
+                    &severity_numbers_filter,
+                    &unknown_severity_number_filter,
+                )
+                .context(error::ColumnLengthMismatchSnafu)?;
+            }
+            // combine severity number filter to the log record filter
+            filter = arrow::compute::and_kleene(&filter, &severity_numbers_filter)
+                .context(error::ColumnLengthMismatchSnafu)?;
+        }
+        Ok(nulls_to_false(&filter))
```
utpilla, line 685: Do we not need to call this in `get_resource_attr_filter` and `get_log_attr_filter` methods?

rust/otel-arrow-rust/src/otap/filter.rs
```diff
@@ -0,0 +1,277 @@
...
+        && ((id_set.len() as f64 / id_column.len() as f64) <= IDS_PERCENTAGE_MAX_THRESHOLD)
+    {
+        let mut combined_id_filter = BooleanArray::new_null(id_column.len());
+        // build id filter using the id hashset
+        for id in id_set {
+            let id_scalar = UInt16Array::new_scalar(id);
+            // since we use a scalar here we don't have to worry a column length mismatch when we compare
+            let id_filter = arrow::compute::kernels::cmp::eq(id_column, &id_scalar)
+                .expect("can compare uint16 id column with uint16 scalar");
+            combined_id_filter = arrow::compute::or_kleene(&combined_id_filter, &id_filter)
```
utpilla, line 202: If I understand correctly, each `or_kleene` and `and_kleene` call would lead to an allocation of a `BooleanArray`. It might be better to create a `BooleanBuilder` instance and append values to it. This investigation is not necessarily needed for this PR.

albertlockett: Is that not what the `else` block below here is doing?

I did some investigation into which was faster and it s dependant on the batch size and how selective the filter is https://github.com/open-telemetry/otel-arrow/pull/1279#discussion_r2453094683

albertlockett (approved):
LGTM @c1ly

rust/otel-arrow-rust/src/otap/filter.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// keyvalue type
+    KeyValue(Vec<KeyValue>),
+}
+
+/// struct that represents attributes and other key value pairs
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct KeyValue {
+    key: String,
+    value: AnyValue,
+}
```
jmacd, line 65: Feeling wary, because this type is likely to exist somewhere else or will in many places if we're not careful. I expect there to be a common KeyValue type, not to block this PR -- we can sort this later.

rust/otel-arrow-rust/src/otap/filter/logs.rs
```diff
@@ -0,0 +1,820 @@
...
+    /// this function takes the filters for each record batch and makes sure that incomplete
+    /// returns the cleaned up filters that can be immediately applied on the record batches
+    /// ToDo: RecordBatches that are optional and not present will result in the corresponding filter being returned as None
+    /// ToDo: Handle edge case where LogRecords don't have attributes are getting through when we want to filter on attributes as well.
+    fn sync_up_filters(
+        &self,
+        logs_payload: &OtapArrowRecords,
+        resource_attr_filter: BooleanArray,
+        mut log_record_filter: BooleanArray,
+        log_attr_filter: BooleanArray,
```
jmacd, line 166: It looks like scopes are missing from this list. Will it be possible to filter by scope name? That's a fairly common scenario, I imagine.

rust/otel-arrow-rust/src/otap/filter.rs
utpilla: Thanks for sharing that! I responded to the original post with my thoughts.

PR #1343, opened 2025-10-23
[PerfTest] Add nightly benchmarks suite and CI workflows

clhain:
This PR adds additional nightly perf benchmarks. 2 additional suites are defined:

* Syslog - various syslog -> otlp/otap with and without CEF formatting
* Backpressure - same as continuous benchmark suite but with wait_for_result set to true (otlp tests enabled now, otap tests defined and ready to enable when the config flag is added to otap receiver)

Both are run as part of a new CI workflow executing at 10AM UTC nightly on the bare metal runner.

Directory structure for integration test configs was refactored slightly to better support the additional setups.

Small fix to the orchestrator loadgen execution strategy to support static message bodies


tools/pipeline_perf_test/test_suites/integration/templates/configs/engine/backpressure/otap-attr-otap.yaml
```diff
@@ -0,0 +1,43 @@
...
+    out_ports:
+      out_port:
+        destinations:
+          - attr
+        dispatch_strategy: round_robin
+    config:
+      listening_addr: "0.0.0.0:4317"
+      message_size: 100
+      compression_method: zstd
+      wait_for_result: true
```
jmacd, line 19: I was looking for a wait_for_result=false test. 

We made wait_for_result=false the default: should this test the other way?

#1328

clhain: Right now the continuous benchmarks use the default (wait_for_result=false) - the nightly backpressure tests are doing wait_for_result=true. I'm totally down for flipping these 2 around if we want backpressure running more frequently (but lets wait till all 4 are functional to do that imo)

jmacd: Agree, thanks.

Issue #1327, opened 2025-10-21
Support Tonic timeout configuration (client and server)


jmacd:
The Tonic library supports automatic cancellation of async RPC tasks after configurable timeouts. We should at a minimum enable this support.

https://docs.rs/tonic/latest/tonic/transport/struct.Endpoint.html#method.timeout

https://docs.rs/tonic/latest/tonic/transport/struct.Server.html#method.timeout

etc

lalitb:
I will be interested to pick this.

PR #1356, opened 2025-10-28
Add tonic timeout config

lalitb:
fixes: #1327

## Summary
Adds optional timeout configuration field to all gRPC clients (exporters) and servers (receivers) using Tonic's built-in timeout support.

## Changes

- Added optional `timeout` field to OTAP Exporter, OTLP Exporter, OTLP Receiver, and OTAP Receiver configs
- Applied timeout via `Endpoint::timeout()` for clients and `Server::builder().timeout()` for servers
- Uses `humantime-serde` for human-readable duration format (e.g., `"30s"`, `"5m"`, `"500ms"`)
- Added config parsing tests to verify timeout deserialization

## Default behavior

- When timeout is not specified, no timeout is applied (requests can run indefinitely). This matches Tonic's default behavior and maintains backward compatibility with existing configurations.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/otap_exporter.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_exporter.rs
@@ -120,7 +120,7 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
             Channel::from_shared(self.config.grpc_endpoint.clone()).map_err(|e| {
                 let source_detail = format_error_sources(&e);
                 Error::ExporterError {
-                    exporter: exporter_id.clone(),
+                    exporter: exporter_id,
                     kind: ExporterErrorKind::Connect,
                     error: format!("grpc channel error {e}"),
                     source_detail,


lquerel (approved):
LGTM

albertlockett (approved):

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
```diff
@@ -116,17 +116,23 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
...
 
         let exporter_id = effect_handler.exporter_id();
-        let channel = Channel::from_shared(self.config.grpc_endpoint.clone())
-            .map_err(|e| {
+        let mut endpoint =
+            Channel::from_shared(self.config.grpc_endpoint.clone()).map_err(|e| {
                 let source_detail = format_error_sources(&e);
                 Error::ExporterError {
-                    exporter: exporter_id,
+                    exporter: exporter_id.clone(),
```
albertlockett: nit: this clone is unnecessary isn't it?
```suggestion
                    exporter: exporter_id,
```

rust/otap-dataflow/configs/otap-otap.yaml
```diff
@@ -14,8 +14,11 @@ nodes:
         dispatch_strategy: round_robin
     config:
       listening_addr: "127.0.0.1:4317"
+      message_size: 256 # Required: channel buffer capacity (number of messages)
```
jmacd, line 17: Tangent: I realize I have no idea what this field is, I suspect it could have a better name, but I see you didn't add this field.

rust/otap-dataflow/crates/otap/src/otap_exporter.rs
jmacd: That stuck my eye too. Wouldn't a clippy rule catch it?

rust/otap-dataflow/configs/otap-otap.yaml
utpilla: This is used by OTAP receiver as the capacity of channel that acts as the response stream for the gRPC client. I found out about it in #1358 and yes that name is odd. Maybe we could use a name like `response_stream_channel_size`?

Issue #1324, opened 2025-10-21
OTAP receiver Ack/Nack handling with concurrency


jmacd:
The OTAP Receiver component does not yet support Ack and Nack.

When doing so, there are three steps in the lifetime of a request:

1. Receiving from the gRPC stream, Arrow IPC readers create OTAP from the received `BatchArrowRecords`. At some point during this step, we must "gain" admission to the pipeline. If we can't, we have to block (wait) or fail (fast). Point is, we block the sender if we can't gain admission.
2. Receiver continues to receive another request. Data is in the pipeline, it's memory or its concurrency slot is acquired, awaiting the Ack/Nack when memory is released.
3. Responding to the Ack/Nack. When the Ack or Nack is received, the gRPC stream `Send()` method delivers a `BatchStatus` message.

The current code immediately Acks the request. We can add `wait_for_result` support as with #1321 (for the OTLP receiver) to retain the current behavior.

PR #1358, opened 2025-10-29
[otap-df-otap] Add ACK and NACK support for OTAP Receiver

utpilla:
Towards #1324 

## Changes
- Add ACK and NACK support for OTAP Receiver


rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
```diff
@@ -34,10 +34,12 @@ use tonic::server::{Grpc, NamedService, UnaryService};
 /// generally optional depending on wait_for_result: true, we do not
 /// create or use the state when ack/nack is not required.
 #[derive(Clone)]
-pub struct SharedState(Arc<Mutex<SlotsState<oneshot::Sender<Result<(), NackMsg<OtapPdata>>>>>>);
+pub struct SharedState(
```
utpilla, line 37: In this PR, I made them visible to OTAP Receiver so that it can use them. These structs could be moved to a shared location as they are used by both OTLP and OTAP receiver. We can do that in another PR. I wanted to keep the diff smaller for this PR.

albertlockett (approved):
LGTM @utpilla!

rust/otap-dataflow/crates/otap/src/otap_receiver.rs
```diff
@@ -321,56 +561,344 @@ mod tests {
...
         }
     }
 
+    /// Test scenario for NACK functionality - expects error responses for all signals
+    fn nack_scenario(
+        grpc_endpoint: String,
+    ) -> impl FnOnce(TestContext<OtapPdata>) -> Pin<Box<dyn Future<Output = ()>>> {
+        move |ctx| {
+            Box::pin(async move {
+                // Test NACK with metrics
```
jmacd, line 572: I think it would be nice to avoid repeating ~70 lines of code for each signal. All the arrow streams use the same concrete data types, so this should be pretty easy.

utpilla: Updated in https://github.com/open-telemetry/otel-arrow/pull/1358/commits/bba2b42ff35d65f88884070d6ccd6783b2e528c4

PR #1361, opened 2025-10-29
[PerfTest] add workflow to manually trigger perf-tests for a given pr

clhain:
This workflow will allow folks with the correct permissions to run a text-output-only version of the benchmark suite on the same bare metal runner used on commits to main / nightly. This is needed to evaluate performance of some PRs prior to merge.


lquerel (approved):
LGTM

PR #1367, opened 2025-10-30
Fix continuous benchmarks for OTAP Receiver

utpilla:
Follow-up to #1365 

## Changes
- Update the continuous benchmark config files for OTAP receiver to use the updated setting name


lquerel (approved):
Thanks.

PR #1381, opened 2025-11-04
[PipelinePerf] setup verification of manual workflow runs for PRs

clhain:
These changes allow automated verification of a manually triggered perf test workflow for a given PR number. Because of how github actions work for this weird case, it's a 2 step thing:

- The new workflow file below (pipeline-perf-verify-pr.yaml) inserts a placeholder status into PRs on create/update that will remain "pending" indefinitely.
- The original manual workflow now updates that placeholder status with the result of it's own run (for the specified PR number).

So the flow is:
- New PR created, new job inserts the pending status result
- Maintainer triggers the manual run with target PR number
- Manual job runs and sets the previously pending result to pass/fail

New commits will clear the result back to pending.

### DIFF
--- a/.github/workflows/pipeline-perf-test-manual-pr.yaml
+++ b/.github/workflows/pipeline-perf-test-manual-pr.yaml
@@ -1,7 +1,11 @@
 # Manual version of the pipeline perf test that can be triggered for a given PR.
+# This workflow writes to / expects to find a placeholder status on the PR ref since we aren't
+# able to directly insert a state into it from 'outside' of that context.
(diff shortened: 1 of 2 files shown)


clhain:
> Nice way to get around GitHub Actions limitations :-)
> 
> In the future, do you think it would be feasible to have the benchmark job add a summary of the results to the PR under test thread? (similar to how the coverage tool inserts its visual summary?)

Yes, the manual job should have sufficient permissions to write a comment - I can definitely do text output, but I think we can figure out a way to have it generate charts also

jmacd (approved):

PR #1357, opened 2025-10-29
[WIP] Set of attempts to optimize OTLP receiver/exporter

lquerel:
Not ready yet.

PR #1382, opened 2025-11-04
Unify OTLP/OTAP gRPC Configuration and Introduce Experimental Non-Tonic Receiver

lquerel:
## Align OTLP/OTAP receivers on shared gRPC config

This PR is a subset of #1357  and focuses on the OTLP and OTAP receivers.
- Both gRPC-based receivers now share the same configuration structure (GrpcServerSettings) to ensure greater consistency. The configuration has also been significantly extended to allow for more precise tuning. Parameters representing byte quantities can now be expressed either as raw byte counts or using units such as MB, MiB, etc.
- Adds flexible compression deserialization (single value, list, or none) and reuses it across request/response compression settings. By default, response compression is no longer enabled since responses are typically very small. Both zstd and gzip are enabled for requests.
- Pushes max-concurrent-request tuning into both receivers (and exposes a helper) so that downstream capacity directly drives the gRPC concurrency clamp.
- Updates the OTLP/OTAP servers to honor the new settings: compression preferences, adaptive windows, and message-size limits are applied once when the services are built.

Key changes for the OTAP receiver
- Response streaming is now driven by an async state machine instead of spawning a new task per request. This removes the extra mpsc hop and preserves backpressure.
- Ack/Nack correlation slots are protected by a parking_lot::Mutex, providing fast, non-poisoning locking in async contexts where a poisoned std::sync::Mutex would otherwise stall the Tokio worker.
- Compression preferences, concurrency limits, and middleware (such as zstd header handling) are applied once per service construction so that hot-path processing remains lean.

## New experimental receiver (non-Tonic)

This PR also introduces an experimental OTAP receiver (otel_receiver) that does not rely on Tonic. The intent is to:
- Better align our gRPC-based receivers with the thread-per-core design of the OTAP pipeline engine
- Improve control over the internal mechanics of these receivers. A first version of the admission controller is included; a future PR will extend it to protect the engine and keep CPU/memory usage under control
- Improve overall performance
- Support OTAP and OTLP on the same port

This experimental receiver does not support OTLP yet, but that is coming soon.

The following diagram describes the overall design to ease the review process.

image

### DIFF
--- a/rust/otap-dataflow/Cargo.toml
+++ b/rust/otap-dataflow/Cargo.toml
@@ -38,6 +38,7 @@ serde_json.workspace = true
 clap.workspace = true
 mimalloc-rust.workspace = true
 env_logger.workspace = true
+jemallocator = { workspace = true, optional = true }
 
 [workspace.dependencies]
 otap-df-pdata-otlp-macros = { path = "./crates/pdata/src/otlp/macros"}
@@ -78,7 +79,9 @@ local-sync = "0.1.1"
 log = "0.4"
 miette = { version="7.6.0", features = ["fancy"] }
 mimalloc-rust = "0.2.1"
-nix = { version = "0.30.0", features = ["process", "signal"] }
+jemalloc-ctl = "0.5.4"
+jemallocator = "0.5.4"
+nix = { version = "0.30.0", features = ["process", "signal", "resource", "feature"] }
 num_enum = "0.7"
 object_store = "0.12.3"
 once_cell = "1.20.2"
@@ -135,7 +138,7 @@ weaver_resolved_schema = { git = "https://github.com/open-telemetry/weaver.git",
 weaver_resolver = { git = "https://github.com/open-telemetry/weaver.git", tag = "v0.17.0"}
 weaver_semconv = { git = "https://github.com/open-telemetry/weaver.git", tag = "v0.17.0"}
 zip = "=4.2.0"
-byte-unit = "5.1.6"
+byte-unit = "5.2.0"
 bytes = "1.10.1"
 env_logger = "0.11"
 h2 = "0.4.7"
@@ -149,6 +152,8 @@ unchecked-arithmetic = []
 # Experimental exporters (opt-in)
 experimental-exporters = ["otap-df-otap/experimental-exporters"]
 geneva-exporter = ["otap-df-otap/geneva-exporter"]
+jemalloc-metrics = ["otap-df-otap/jemalloc-metrics"]
+jemalloc-allocator = ["dep:jemallocator"]
 
 [workspace.lints.rust]
 # General compatibility lints
--- a/rust/otap-dataflow/crates/otap/Cargo.toml
+++ b/rust/otap-dataflow/crates/otap/Cargo.toml
@@ -45,6 +45,8 @@ bitflags = { workspace = true }
 bytes = { workspace = true }
 parking_lot = { workspace = true }
 flate2 = { workspace = true }
+nix = { workspace = true }
+jemalloc-ctl = { workspace = true, optional = true }
 
 otap-df-engine = { path = "../engine" }
 otap-df-engine-macros = { path = "../engine-macros" }
@@ -79,6 +81,7 @@ experimental-exporters = []
 geneva-exporter = ["experimental-exporters", "dep:geneva-uploader", "dep:opentelemetry-proto"]
 # Opt-in unsafe micro-optimizations used in the OTEL receiver hot path.
 unsafe-optimizations = []
+jemalloc-metrics = ["dep:jemalloc-ctl"]
 
 [dev-dependencies]
 flume = { workspace = true }
--- a/rust/otap-dataflow/crates/otap/src/otap_grpc.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_grpc.rs
@@ -40,12 +40,14 @@ use crate::{
     pdata::{Context, OtapPdata},
 };
 
+pub mod client_settings;
 pub mod common;
 pub mod middleware;
 pub mod otlp;
-pub mod receiver_settings;
+pub mod server_settings;
 
-pub use receiver_settings::GrpcServerSettings;
+pub use client_settings::GrpcClientSettings;
+pub use server_settings::GrpcServerSettings;
 
 /// Common settings for OTLP receivers.
 #[derive(Clone, Debug)]
--- a/rust/otap-dataflow/crates/otap/src/otap_grpc/server_settings.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_grpc/server_settings.rs
@@ -24,9 +24,7 @@ pub struct GrpcServerSettings {
     /// and deflate (in that preference order).
     #[serde(
         default,
-        deserialize_with = "compression::deserialize_compression_methods",
-        alias = "compression_method",
-        alias = "request_compression_method"
+        deserialize_with = "compression::deserialize_compression_methods"
     )]
     pub request_compression: Option<Vec<CompressionMethod>>,
 
@@ -262,15 +260,15 @@ const fn default_tcp_nodelay() -> bool {
     true
 }
 
-fn default_tcp_keepalive() -> Option<Duration> {
+const fn default_tcp_keepalive() -> Option<Duration> {
     Some(Duration::from_secs(45))
 }
 
-fn default_tcp_keepalive_interval() -> Option<Duration> {
+const fn default_tcp_keepalive_interval() -> Option<Duration> {
     Some(Duration::from_secs(15))
 }
 
-fn default_tcp_keepalive_retries() -> Option<u32> {
+const fn default_tcp_keepalive_retries() -> Option<u32> {
     Some(5)
 }
 
@@ -278,27 +276,27 @@ const fn default_load_shed() -> bool {
     true
 }
 
-fn default_initial_stream_window_size() -> Option<u32> {
+const fn default_initial_stream_window_size() -> Option<u32> {
     Some(8 * 1024 * 1024)
 }
 
-fn default_initial_connection_window_size() -> Option<u32> {
+const fn default_initial_connection_window_size() -> Option<u32> {
     Some(24 * 1024 * 1024)
 }
 
-fn default_max_frame_size() -> Option<u32> {
+const fn default_max_frame_size() -> Option<u32> {
     Some(16 * 1024)
 }
 
-fn default_max_decoding_message_size() -> Option<u32> {
+const fn default_max_decoding_message_size() -> Option<u32> {
     Some(4 * 1024 * 1024)
 }
 
-fn default_http2_keepalive_interval() -> Option<Duration> {
+const fn default_http2_keepalive_interval() -> Option<Duration> {
     Some(Duration::from_secs(30))
 }
 
-fn default_http2_keepalive_timeout() -> Option<Duration> {
+const fn default_http2_keepalive_timeout() -> Option<Duration> {
     Some(Duration::from_secs(10))
 }
 
@@ -306,7 +304,7 @@ const fn default_http2_adaptive_window() -> bool {
     false
 }
 
-fn default_http2_handshake_timeout() -> Duration {
+const fn default_http2_handshake_timeout() -> Duration {
     Duration::from_secs(5)
 }
 
@@ -341,5 +339,62 @@ impl Default for GrpcServerSettings {
             wait_for_result: default_wait_for_result(),
             timeout: None,
         }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::compression::{CompressionMethod, DEFAULT_COMPRESSION_METHODS};
+    use tonic::codec::CompressionEncoding;
+
+    #[test]
+    fn defaults_match_expected_compression() {
+        let settings = GrpcServerSettings::default();
+
+        assert_eq!(
+            settings.request_compression_methods(),
+            DEFAULT_COMPRESSION_METHODS.to_vec()
+        );
+        assert!(settings.response_compression_methods().is_empty());
+        assert_eq!(settings.preferred_response_compression(), None);
+    }
+
+    #[test]
+    fn response_compression_prefers_first_entry() {
+        let settings = GrpcServerSettings {
+            response_compression: Some(vec![CompressionMethod::Gzip, CompressionMethod::Zstd]),
+            ..Default::default()
+        };
+
+        assert_eq!(
+            settings.preferred_response_compression(),
+            Some(CompressionMethod::Gzip)
+        );
+        assert_eq!(
+            settings.response_compression_methods(),
+            vec![CompressionMethod::Gzip, CompressionMethod::Zstd]
+        );
+    }
+
+    #[test]
+    fn build_settings_carries_core_limits_and_compression() {
+        let settings = GrpcServerSettings {
+            max_concurrent_requests: 42,
+            wait_for_result: true,
+            max_decoding_message_size: Some(8 * 1024 * 1024),
+            request_compression: Some(vec![CompressionMethod::Deflate]),
+            response_compression: Some(vec![CompressionMethod::Deflate]),
+            ..Default::default()
+        };
+
+        let built = settings.build_settings();
+        assert_eq!(built.max_concurrent_requests, 42);
+        assert!(built.wait_for_result);
+        assert_eq!(built.max_decoding_message_size, Some(8 * 1024 * 1024));
+
+        let (req, resp) = settings.compression_encodings();
+        assert!(req.is_enabled(CompressionEncoding::Deflate));
+        assert!(resp.is_enabled(CompressionEncoding::Deflate));
     }
 }
--- a/rust/otap-dataflow/src/main.rs
+++ b/rust/otap-dataflow/src/main.rs
@@ -4,6 +4,9 @@
 //! Create and run a multi-core pipeline
 
 use clap::Parser;
+#[cfg(feature = "jemalloc-allocator")]
+use jemallocator::Jemalloc;
+#[cfg(not(feature = "jemalloc-allocator"))]
 use mimalloc_rust::*;
 use otap_df_config::pipeline::PipelineConfig;
 use otap_df_config::pipeline_group::{CoreAllocation, CoreRange, Quota};
@@ -12,8 +15,13 @@ use otap_df_controller::Controller;
 use otap_df_otap::OTAP_PIPELINE_FACTORY;
 use std::path::PathBuf;
 
+#[cfg(feature = "jemalloc-allocator")]
 #[global_allocator]
-static GLOBAL_MIMALLOC: GlobalMiMalloc = GlobalMiMalloc;
+static GLOBAL_ALLOC: Jemalloc = Jemalloc;
+
+#[cfg(not(feature = "jemalloc-allocator"))]
+#[global_allocator]
+static GLOBAL_ALLOC: GlobalMiMalloc = GlobalMiMalloc;
 
 #[derive(Parser)]
 #[command(
--- a/rust/otap-dataflow/configs/otlp-perf.yaml
+++ b/rust/otap-dataflow/configs/otlp-perf.yaml
@@ -22,7 +22,7 @@ nodes:
     config:
       frequency: 1000
       display_throughput: true
-      cpu_usage: false
-      mem_usage: false
+      cpu_usage: true
+      mem_usage: true
       disk_usage: false
       io_usage: false
--- a/rust/otap-dataflow/crates/otap/src/metrics.rs
+++ b/rust/otap-dataflow/crates/otap/src/metrics.rs
@@ -55,26 +55,47 @@ pub struct ExporterPDataMetrics {
 
 impl ExporterPDataMetrics {
     pub fn inc_consumed(&mut self, st: SignalType) {
-        match st {
-            SignalType::Metrics => self.metrics_consumed.inc(),
-            SignalType::Logs => self.logs_consumed.inc(),
-            SignalType::Traces => self.traces_consumed.inc(),
-        }
+        self.add_consumed(st, 1);
     }
 
     pub fn inc_exported(&mut self, st: SignalType) {
-        match st {
-            SignalType::Metrics => self.metrics_exported.inc(),
-            SignalType::Logs => self.logs_exported.inc(),
-            SignalType::Traces => self.traces_exported.inc(),
-        }
+        self.add_exported(st, 1);
     }
 
     pub fn inc_failed(&mut self, st: SignalType) {
+        self.add_failed(st, 1);
+    }
+
+    pub fn add_consumed(&mut self, st: SignalType, count: u64) {
+        if count == 0 {
+            return;
+        }
         match st {
-            SignalType::Metrics => self.metrics_failed.inc(),
-            SignalType::Logs => self.logs_failed.inc(),
-            SignalType::Traces => self.traces_failed.inc(),
+            SignalType::Metrics => self.metrics_consumed.add(count),
+            SignalType::Logs => self.logs_consumed.add(count),
+            SignalType::Traces => self.traces_consumed.add(count),
+        }
+    }
+
+    pub fn add_exported(&mut self, st: SignalType, count: u64) {
+        if count == 0 {
+            return;
+        }
+        match st {
+            SignalType::Metrics => self.metrics_exported.add(count),
+            SignalType::Logs => self.logs_exported.add(count),
+            SignalType::Traces => self.traces_exported.add(count),
+        }
+    }
+
+    pub fn add_failed(&mut self, st: SignalType, count: u64) {
+        if count == 0 {
+            return;
+        }
+        match st {
+            SignalType::Metrics => self.metrics_failed.add(count),
+            SignalType::Logs => self.logs_failed.add(count),
+            SignalType::Traces => self.traces_failed.add(count),
         }
     }
 }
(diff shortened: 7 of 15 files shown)


rust/otap-dataflow/crates/otap/src/otap_grpc.rs
```diff
@@ -224,162 +168,186 @@ impl ArrowTracesService for ArrowTracesServiceImpl {
...
+                            batch_id,
+                            status_code: StatusCode::Unavailable as i32,
+                            status_message: format!("Pipeline processing failed: {}", nack.reason),
+                        });
+                    }
+                    Err(_) => {
+                        log::error!("Response channel closed unexpectedly");
+                        return BatchProcessingResult::Terminate;
+                    }
+                }
```
jmacd: cc @utpilla in #1324 described a feature request to implement asynchronous requests, which the `otelarrowreceiver` does from Phase 1.

Do you object to concurrent execution, meaning to allow the stream to continue reading with some number of outstanding responses pending?

rust/otap-dataflow/crates/otap/src/otap_grpc.rs
```diff
@@ -224,162 +168,186 @@ impl ArrowTracesService for ArrowTracesServiceImpl {
...
-                    log::error!("Error sending BatchStatus response: {e:?}");
-                })?;
-
-                return Ok(());
-            }
-            Err(_) => {
-                log::error!("Response channel closed unexpectedly");
-                return Err(());
-            }
+    let stream = stream::unfold(state, |mut state| async move {
```
jmacd: 💥 Cool

rust/otap-dataflow/crates/config/src/byte_units.rs
```diff
@@ -0,0 +1,126 @@
...
+            (parsed.as_u64() as u128, text)
+        }
+    };
+
+    if bytes > u32::MAX as u128 {
+        return Err(DeError::custom(format!(
+            "byte size {} exceeds u32::MAX ({} bytes)",
+            repr,
+            u32::MAX
+        )));
```
utpilla, line 42:
```suggestion
        return Err(DeError::custom(format!(
            "byte size '{}' ({} bytes) exceeds u32::MAX ({} bytes)",
            repr,
            bytes,
            u32::MAX
        )));
```

rust/otap-dataflow/crates/otap/src/otap_grpc.rs
lquerel: No objection.

rust/otap-dataflow/crates/engine/src/local/receiver.rs
```diff
@@ -121,10 +122,14 @@ impl<PData> ControlChannel<PData> {
 }
 
 /// A `!Send` implementation of the EffectHandler.
+/// Note: This handler MUST be cloneable with low overhead.
 #[derive(Clone)]
 pub struct EffectHandler<PData> {
     core: EffectHandlerCore<PData>,
+    senders: Rc<EffectHandlerSenders<PData>>,
```
utpilla: Why not just wrap everything behind `Rc`/`Arc`?

```rust
#[derive(Clone)]
    inner: Rc<EffectHandlerInner<PData>>,
}

struct EffectHandlerInner<PData> {
    core: EffectHandlerCore<PData>,
    msg_senders: HashMap<PortName, LocalSender<PData>>,
    default_sender: Option<LocalSender<PData>>,
}
```

rust/otap-dataflow/crates/otap/src/compression.rs
```diff
@@ -26,16 +25,73 @@ impl CompressionMethod {
...
+        CompressionConfigValue::List(methods) => methods,
+        CompressionConfigValue::NoneKeyword(CompressionNone::None) => Vec::new(),
+    };
+
+    let mut deduped = Vec::with_capacity(methods.len());
+    for method in methods {
+        if !deduped.contains(&method) {
+            deduped.push(method);
+        }
+    }
```
utpilla, line 92: This logic is only needed for `CompressionConfigValue::List`. We could move it inside the match statement to make the code simpler.

rust/otap-dataflow/crates/otap/Cargo.toml
```diff
@@ -42,6 +42,8 @@ tonic-prost = { workspace = true }
 prost = { workspace = true }
 smallvec = { workspace = true }
 bitflags = { workspace = true }
+bytes = { workspace = true }
+parking_lot = { workspace = true }
```
utpilla, line 46: I don't think we need to add this dependency. We won't have any thread contention so there shouldn't be any performance improvement coming from this.

rust/otap-dataflow/crates/engine/src/local/receiver.rs
lquerel: It's definitely an option. It's just that I noticed the presence of the hashmap and wanted to encapsulate it without thinking too much about it. I’ll do the proposed refactoring.

rust/otap-dataflow/crates/otap/Cargo.toml
lquerel: Receivers are unfortunately Send because of Tonic limitation. Don't understand the feedback.

rust/otap-dataflow/crates/otap/src/otap_grpc.rs
```diff
@@ -224,162 +168,186 @@ impl ArrowTracesService for ArrowTracesServiceImpl {
...
+                    BatchProcessingResult::Emit(status) => StreamStep::Yield(Ok(status)),
+                    BatchProcessingResult::Terminate => StreamStep::Done,
+                },
+                Ok(None) => {
+                    self.finished = true;
+                    StreamStep::Done
+                }
+                Err(status) => {
+                    self.finished = true;
+                    StreamStep::Yield(Err(status))
```
utpilla: Do we always want to terminate the stream on any error? Not something for this PR, but maybe we should inspect the error to check if it's a fatal error or not?

cc @jmacd

rust/otap-dataflow/crates/otap/Cargo.toml
utpilla: My question is why use `parking_lot::Mutex`at all? We can continue to use `std::sync::Mutex` for satisfying the `Send` requirement. That way we don't have to add "a new dependency" to the project.

Also, since there is no thread contention, there shouldn't be any perf improvements either by using `parking_lot::Mutex`.

rust/otap-dataflow/crates/otap/src/otap_grpc.rs
utpilla: If I understand correctly, this will result in the server processing incoming telemetry only when the client pulls responses. That would create a client-driven processing model which is problematic. We want the server's processing to be independent of the client's response consumption. The task spawning + channel based approached lets us achieve that.

rust/otap-dataflow/crates/otap/src/otap_grpc/server_settings.rs
```diff
@@ -0,0 +1,334 @@
...
+    #[serde(default = "default_http2_keepalive_timeout", with = "humantime_serde")]
+    pub http2_keepalive_timeout: Option<Duration>,
+
+    /// Upper bound on concurrently active HTTP/2 streams per connection.
+    /// By default this tracks the effective `max_concurrent_requests`, keeping logical and transport
+    /// concurrency aligned. Lower values improve fairness between chatty clients. Higher values
+    /// matter only if you also raise `max_concurrent_requests`. Set to `0` to inherit the derived
+    /// default.
+    #[serde(default)]
+    pub max_concurrent_streams: Option<u32>,
```
utpilla, line 158: @jmacd if I remember correctly, your suggestion was to always have this set to `1`, right? Could you please share the reasons for doing that here as well? One thing that would do is force the client to open a new connection for each stream which also helps `SO_REUSEPORT` option to balance the load more effectively.

rust/otap-dataflow/crates/otap/src/otap_grpc/receiver_settings.rs
```diff
@@ -0,0 +1,334 @@
...
+
+const fn default_max_concurrent_requests() -> usize {
+    0
+}
+
+const fn default_tcp_nodelay() -> bool {
+    true
+}
+
+fn default_tcp_keepalive() -> Option<Duration> {
```
utpilla: nit: All of these methods could be made `const`.

rust/otap-dataflow/crates/otap/src/otap_receiver.rs
```diff
@@ -840,86 +775,153 @@ mod tests {
...
-        assert!(!receiver.config.wait_for_result);
-        assert!(receiver.config.compression_method.is_none());
-        assert!(receiver.config.timeout.is_none());
-
-        // Test with minimal required fields, max_concurrent_requests defaults to 1000, wait_for_result defaults to false
+        assert_eq!(
+            receiver.config.grpc.listening_addr.to_string(),
+            "127.0.0.1:4317"
+        );
+        assert_eq!(receiver.config.grpc.max_concurrent_requests, 5000);
```
utpilla, line 790: For one of these tests, we should also verify that `tune_max_concurrent_requests` method works correctly.

utpilla: It looks like we have it added for OTLP Receiver but not for OTAP Receiver.

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
```diff
@@ -442,36 +392,203 @@ mod tests {
...
+        // Defaults clamp to downstream capacity.
+        let config_default = json!({
+            "listening_addr": "127.0.0.1:4317"
+        });
+        let mut receiver =
+            OTLPReceiver::from_config(pipeline_ctx.clone(), &config_default).unwrap();
+        receiver.tune_max_concurrent_requests(128);
+        assert_eq!(receiver.config.grpc.max_concurrent_requests, 128);
+
+        // User provided smaller value is preserved.
```
utpilla, line 575: We should also test for user providing a larger value than the downstream capacity.

rust/otap-dataflow/crates/otap/Cargo.toml
jmacd: I read in the PR description that this might be about lock-poisoning, which parking_lot doesn't do, it eliminates an async corner case I guess.

rust/otap-dataflow/crates/otap/src/otap_grpc.rs
jmacd: This was a question in the Phase 1 otelarrowreceiver where I didn't feel like I knew the right answer, but would agree that some errors are permanent and should terminate the stream. I would probably add INVALID_ARGUMENT and AUTHORIZATION_FAILED or similar codes to the conditions that might be permitted, but a default of terminating the stream is a good start.

rust/otap-dataflow/crates/otap/src/otap_grpc.rs
jmacd: Hmm, I may not understand unfold(). (This sounds like a valid concern.)

rust/otap-dataflow/crates/otap/src/otap_grpc/server_settings.rs
jmacd: This is just a standard setting that all gRPC and HTTP/2 servers provide, I've come to appreciate how the option has the same name everywhere. In the 2024 blog post, we recommended setting this to 1 because it provides maximum independence between streams--they cannot block each other, compete for buffer space, and so on. The observation is that if a gRPC server sets this field to 1, then the upstream load balanacer (e.g., an Envoy) will open new connections for each stream, which gives better performance/reliability/scale. It's still a good option for users, but I can't see reasons why it would be useful for an OTAP stream. Possibly this is more reasonable to set with OTLP-unary requests.

rust/otap-dataflow/crates/otap/Cargo.toml
utpilla: Regarding poisoning vs non-poisoning, the choice depends on our expectations:
- If we expect panics might occur while holding the lock, then poisoning helps detect potential data corruption
- If we expect panics to never occur, then we can use `.expect("lock should never be poisoned")` with `std::sync::Mutex`, making the poisoning case explicit and self-documenting

Is the slightly cleaner API (no usage of `.expect()`) worth adding the dependency?

rust/otap-dataflow/crates/otap/src/otap_grpc.rs
lquerel: That's a valid concern. I'm currently revising my draft and improving both inter- and extra-connection concurrency, as well as the overall admission control mechanism.

rust/otap-dataflow/crates/otap/Cargo.toml
lquerel: The experimental OTEL receiver (combining OTLP and OTAP) no longer uses a Mutex, so this issue is gone.

rust/otap-dataflow/crates/otap/src/compression.rs
```diff
@@ -26,16 +25,73 @@ impl CompressionMethod {
...
+    Single(CompressionMethod),
+    List(Vec<CompressionMethod>),
+    NoneKeyword(CompressionNone),
+}
+
+#[derive(Deserialize)]
+#[serde(rename_all = "snake_case")]
+enum CompressionNone {
+    None,
+}
```
albertlockett, line 66: Interesting -- this is `CompressionNone` enum to trick serde to deserializing "none" correctly?

Not sure if it's helpful here, but we do have this helper which is used in other places:
https://github.com/open-telemetry/otel-arrow/blob/8664c2fe36221ca8b1f33445749ff54cda69ec01/rust/otap-dataflow/crates/otap/src/otap_exporter/config.rs#L80-L101

rust/otap-dataflow/crates/otap/src/fake_data_generator/fake_signal.rs
```diff
@@ -136,6 +137,11 @@ fn spans(signal_count: usize, registry: &ResolvedRegistry) -> Vec<Span> {
                 )
                 .kind(otel_span_kind(group.span_kind.as_ref()))
                 .end_time_unix_nano(end_time)
+                // Note: without status set, the OTAP encoder fails at runtime
+                .status(otap_df_pdata::proto::opentelemetry::trace::v1::Status::new(
+                    StatusCode::Ok,
+                    "ok",
+                ))
```
albertlockett, line 144: possibly fixed by @jmacd in https://github.com/open-telemetry/otel-arrow/pull/1436 ?

lquerel: Yes I think so

rust/otap-dataflow/src/main.rs
```diff
@@ -84,6 +84,7 @@ fn parse_core_id_range(s: &str) -> Result<CoreRange, String> {
 }
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
+    init_logging();
```
albertlockett, line 95: I'm glad to see this being added 🔥 

Should pipeline components still use effect_handler.info(), or can we now use log::info() ?

rust/otap-dataflow/crates/otap/src/otel_receiver.rs
```diff
@@ -0,0 +1,994 @@
...
+    async fn start(
+        mut self: Box<Self>,
+        mut ctrl_msg_recv: local::ControlChannel<OtapPdata>,
+        effect_handler: local::EffectHandler<OtapPdata>,
+    ) -> Result<TerminalState, Error> {
+        let admitter = Admitter::new(
+            100000,
+            self.config.grpc.max_concurrent_streams.unwrap_or(100),
+            100000,
+        );
```
albertlockett, line 161: Do we want these hard-coded? If so, maybe we could comment on why we chose these values? And if not, maybe let's add a TODO to make them configurable?

lquerel: Yes I will add a todo, because a future PR will focus on the admiter and these parameters

rust/otap-dataflow/crates/otap/src/otel_receiver.rs
```diff
@@ -0,0 +1,994 @@
...
+                                            log::debug!("H2 connection ended with error: {err}");
+                                        }
+                                    }
+                                });
+                            }
+                            AdmitDecision::Busy => {
+                                // Soft backpressure: do not spawn; let the kernel backlog absorb.
+                                if log::log_enabled!(log::Level::Trace) {
+                                    log::trace!("Connection admission busy; pausing accepts briefly");
+                                }
```
albertlockett: nit: I'm not sure we need to check if the log level is enabled before calling `log::trace!` here do we? Especially since we're not computing anything that would be expensive to log, and I think the log level check also happens internally in `log::trace!`

rust/otap-dataflow/crates/otap/src/otel_receiver.rs
```diff
@@ -0,0 +1,994 @@
...
+                (ping_pong, interval, timeout)
+            }
+            _ => return None,
+        };
+        Some(Self {
+            ping_pong,
+            interval,
+            timeout,
+            sleep: None,
+        })
```
albertlockett, line 704: nit: could this be simplified as follows?
```suggestion
        let interval = interval?;
        if interval.is_zero() {
            return None
        }

        Some(Self {
            ping_pong: ping_pong?,
            interval,
            timeout: timeout?,
            sleep: None,
        })
```

rust/otap-dataflow/crates/otap/src/otel_receiver.rs
```diff
@@ -0,0 +1,994 @@
...
+impl GrpcRequestRouter {
+    async fn route_grpc_request(
+        self: Rc<Self>,
+        request: Request<h2::RecvStream>,
+        respond: SendResponse<Bytes>,
+    ) -> Result<(), Status> {
+        let path = request.uri().path();
+        match path {
+            ARROW_LOGS_SERVICE => {
+                // log::info!("Handling ArrowLogs stream");
```
albertlockett: nit: maybe we should remove all the places we have commented log lines in this `impl` block? Or alternatively change them to debug! or trace! and uncomment them

rust/otap-dataflow/crates/otap/src/fake_data_generator/fake_signal.rs
jmacd, line 144:
```suggestion
                .status(otap_df_pdata::proto::opentelemetry::trace::v1::Status::new(
                    StatusCode::Ok,
                    "ok",
                ))
```

Maybe fixed in #1436

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server.rs
```diff
@@ -23,6 +23,7 @@ use otap_df_pdata::OtlpProtoBytes;
 use otap_df_pdata::proto::opentelemetry::collector::logs::v1::ExportLogsServiceResponse;
 use otap_df_pdata::proto::opentelemetry::collector::metrics::v1::ExportMetricsServiceResponse;
 use otap_df_pdata::proto::opentelemetry::collector::trace::v1::ExportTraceServiceResponse;
+use parking_lot::Mutex;
```
jmacd, line 26: 👍 TIL

rust/otap-dataflow/crates/otap/src/otap_grpc/common.rs
```diff
@@ -0,0 +1,138 @@
...
+/// the receiver to the rest of the pipeline).
+pub fn tune_max_concurrent_requests(config: &mut GrpcServerSettings, downstream_capacity: usize) {
+    // Fall back to the downstream channel capacity when it is tighter than the user setting.
+    let safe_capacity = downstream_capacity.max(1);
+    if config.max_concurrent_requests == 0 || config.max_concurrent_requests > safe_capacity {
+        config.max_concurrent_requests = safe_capacity;
+    }
+}
+
+/// Applies the shared server tuning options to a tonic server builder.
```
jmacd, line 103:
```suggestion
/// Applies the shared server tuning options to a server builder.
```

rust/otap-dataflow/crates/otap/src/otel_receiver/ack.rs
```diff
@@ -0,0 +1,370 @@
...
+
+/// We pre-allocate a boxed slice of slots so indices remain stable and
+/// `AckToken` can cheaply refer to one.  `free_stack` is a simple LIFO of
+/// currently unused slot indices: allocate = `pop()`, complete/cancel =
+/// `push()`.  Because the registry is tiny and single-threaded, this gives us
+/// O(1) operations with minimal bookkeeping.
+struct AckRegistryInner {
+    slots: Box<[AckSlot]>,
+    free_stack: Vec<usize>,
+}
```
jmacd, line 70: I'm a tiny bit confused. Is this the new experimental code intended to replace something else, or have you already replaced something with this? This looks like the data type in crates/otap/src/accessory which use `slotmap`, which I see still referenced from otap_grpc/otlp/server.rs. This looks like an alternative to slotmap?

rust/otap-dataflow/Cargo.toml
```diff
@@ -133,6 +135,11 @@ weaver_resolved_schema = { git = "https://github.com/open-telemetry/weaver.git",
 weaver_resolver = { git = "https://github.com/open-telemetry/weaver.git", tag = "v0.17.0"}
 weaver_semconv = { git = "https://github.com/open-telemetry/weaver.git", tag = "v0.17.0"}
 zip = "=4.2.0"
+byte-unit = "5.1.6"
+bytes = "1.10.1"
+env_logger = "0.11"
+h2 = "0.4.7"
+zstd = "0.13"
```
utpilla: The dependencies above are listed in alphabetical order. Rearrange these as well?

rust/otap-dataflow/crates/otap/src/otel_receiver/grpc.rs
```diff
@@ -0,0 +1,1368 @@
...
+                    _ = sleep.as_mut() => Err(()),
+                    result = future.as_mut() => {
+                        self.reset();
+                        Ok(result)
+                    }
+                }
+            }
+        }
+    }
+}
```
jmacd, line 1008: You can write your own gRPC implementation, but it better be 1000 lines or less!

rust/otap-dataflow/crates/otap/src/otap_grpc.rs
```diff
@@ -224,162 +206,282 @@ impl ArrowTracesService for ArrowTracesServiceImpl {
...
+fn build_status_stream<S, T, F>(
+    state: ArrowBatchStreamState<S, T, F>,
+) -> impl Stream<Item = Result<BatchStatus, Status>>
+where
+    S: ArrowRequestStream + Send,
+    T: OtapBatchStore + 'static,
+    F: Fn(T) -> OtapArrowRecords + Send + Copy + 'static,
+{
+    stream::unfold(state, |mut state| async move {
+        match state.next_item().await {
```
utpilla, line 275: It seems the same problem persists. The receiver requires the client to poll the output stream. Even though `fill_inflight()` reads data eagerly from the input stream, it won’t run until the client polls the output stream, right?

```
client polls the output stream -> `ArrowBatchStreamState::next_item()` -> `ArrowBatchStreamState::fill_inflight()`
```

rust/otap-dataflow/crates/otap/src/otel_receiver/otap_tests.rs
```diff
@@ -0,0 +1,1127 @@
...
+                }
+            }
+            let expected_traces: Vec<_> = plan.traces.into_iter().map(|(_, msg)| msg).collect();
+            assert_equivalent(&expected_traces, &actual_traces);
+        })
+    }
+}
+
+#[test]
+#[ignore = "temporarily disabled while investigating produce_bar failure"]
```
albertlockett, line 293: @lquerel since #1447 and #1446 are merged this should pass now. I tested locally by merging main into your branch and it seems to work

rust/otap-dataflow/crates/otap/src/otel_receiver/grpc.rs
albertlockett: lol yeah we gotta delete all the comments to get under the line limit

utpilla (commented):
Thanks for the effort on this! I wanted to share a few thoughts and concerns. To be upfront, I don’t have deep expertise in HTTP/2 or gRPC internals, so I can’t fully evaluate this implementation and that’s part of why I’m raising these points.

I understand the motivation behind avoiding `Send` bounds for a thread-per-core design, and that seems like a valid goal. My concern is more about long-term maintainability than the immediate code. HTTP/2 and gRPC have subtle requirements around flow control, connection state, error handling, deadline propagation, and header compression. I guess my question is less about whether this works now but whether it will still work correctly down the road when someone encounters a weird client, a specific network condition, or a new gRPC feature. Tonic has years of production exposure and an active community finding and fixing edge cases. With a custom implementation, we'd be building up that production experience from scratch. I’m not sure how many of us (myself included) feel comfortable debugging HTTP/2 frame-level issues if they arise later. Just something to consider on the maintenance side.

Maybe we could feature flag this implementation rather than trying to make it default? This would let teams who need the thread-per-core performance benefit opt in, while keeping Tonic as the battle-tested default. It would also give time to harden the h2 implementation with early adopters before broader rollout.

lquerel:
@utpilla We are on the same page. I plan to keep both the OTLP and OTAP receivers for quite some time, in addition to this experimental receiver.

lquerel:
For now, it has been decided to keep Tonic.

PR #1383, opened 2025-11-04
[PipelinePerf] Try cleanup cache to fix disk space issues in runners

cijothomas:
Trying to see if this can unblock CI immediately. Can move the cleanup to a common script etc. once this is confirmed to be helping.


clhain:
Looks good to me - we can probably add a check for free space < X so we can still take advantage of the caching in the long-term fix

Issue #387, opened 2025-05-07
[otel-arrow-rust] encode Metrics


albertlockett:
Implement encoding OTLP Metrics as OTAP

The scope of this should include both univariate and multivariate metrics

albertlockett:
We've implemented encoding from OTLP prost structs, but we still need to implement OTLP proto bytes. Similar to what was done for traces in: https://github.com/open-telemetry/otel-arrow/pull/983

We can use this issue to track this work since #768 (which also covered this work), is now closed.

PR #1386, opened 2025-11-05
Initial support for directly encoding OTAP from OTLP bytes

albertlockett:
part of #387 

Handles `MetricsData`, `ResourceMetrics`, `ScopeMetrics` and `Metrics` proto messages. Metrics data will be handled in immediate followup PRs.

Also fixes an issue where metrics metadata was not getting decoded correctly when converting from OTAP to OTLP Bytes.


rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/metrics.rs
```diff
@@ -0,0 +1,433 @@
...
+            let range = match to_nonzero_range(start, end) {
+                Some(range) => range,
+                None => return,
+            };
+
+            match field_num {
+                METRIC_NAME => self.name.set(Some(range)),
+                METRIC_DESCRIPTION => self.description.set(Some(range)),
+                METRIC_UNIT => self.unit.set(Some(range)),
+                METRIC_GAUGE | METRIC_SUM | METRIC_HISTOGRAM | METRIC_EXPONENTIAL_HISTOGRAM => {
```
utpilla: Include `METRIC_SUMMARY` as well?

albertlockett: Thanks @utpilla you're right. Made this change

Issue #1377, opened 2025-11-03
Consolidate otel-arrow-rust modules into otap-dataflow crates


jmacd:
Part of #1376 
Part of #1340 

While working on the batch processor, I wanted to use the pdata/views abstraction as part of a test, but I had difficulty because of the separation of the otap-dataflow and otel-arrow-rust crates.

It is time to consolidate these!

Issue #867, opened 2025-08-05
Consolidate all the Rust code into a single repo


lquerel:
Consolidate all the existing Rust repo into the existing `otap-dataflow` folder.

Note: This is a strecht goal for the milestone demo

PR #1389, opened 2025-11-05
[otel-arrow-rust] Replace Snafu with thiserror

jmacd:
Part of #1377 #867 
Replaces snafu with thiserror.

Here's my retrospective on Snafu vs. Thiserror:

Snafu has a little more sugar, maybe too much magic. E.g., options and results both have context extensions, with thiserror you choose ok_or_else() or map_err() explicitly. While I admire Snafu for its automation, I prefer the more sparing approach taken by thiserror.

### DIFF
--- a/.github/workflows/pipeline-perf-test-continuous.yml
+++ b/.github/workflows/pipeline-perf-test-continuous.yml
@@ -37,19 +37,8 @@ jobs:
           run: |
             set -euxo pipefail
 
-            # Check disk space before operations
-            echo "=== Disk space check ==="
-            df -h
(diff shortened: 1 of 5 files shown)


rust/otel-arrow-rust/src/pdata/mod.rs
```diff
@@ -86,114 +83,3 @@ impl From<[u8; 8]> for SpanID {
         SpanID(sid)
     }
 }
-
```
jmacd, line 89: Unrelated dead code.

PR #1393, opened 2025-11-06
OTLP bytes -> OTAP encoding for Metric data: Gauge & NumberDataPoint

albertlockett:
part of #387

Support direct encoding OTLP Bytes -> OTAP for Metric data of type Gauge including the NumberDataPoints.

Will continue to support additional data types and other metric messages in followup PRs.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/metrics.rs
+++ b/rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/metrics.rs
@@ -164,8 +164,8 @@ pub struct MetricFieldRanges {
     unit: Cell<Option<(NonZeroUsize, NonZeroUsize)>>,
     first_metadata: Cell<Option<(NonZeroUsize, NonZeroUsize)>>,
 
-    // since this is a oneof, we also keep the field_num alongside the range. 
-    // That way when `get_field_range` is called, we can avoid returning the 
+    // since this is a oneof, we also keep the field_num alongside the range.
+    // That way when `get_field_range` is called, we can avoid returning the
     // data range for the wrong variant of the oneof field
     data: Cell<Option<((NonZeroUsize, NonZeroUsize), u64)>>,
 }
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/metrics.rs
```diff
@@ -214,7 +228,148 @@ impl FieldRanges for MetricFieldRanges {
...
+}
+
+/// Implementation of [`GaugeView`]
+pub struct RawGauge<'a> {
+    byte_parser: ProtoBytesParser<'a, GaugeFieldRanges>,
+}
+
+/// Known field ranges for fields on Gauge messages
+pub struct GaugeFieldRanges {
+    first_data_point: Cell<Option<(NonZeroUsize, NonZeroUsize)>>,
```
jmacd, line 244: Curious reader question: why are you using Cell<_> here and elsewhere in this file?

albertlockett: The whole idea behind this is that we only want to pass over the buffer that contains the serialized message once, regardless of the order in which a user of the view trait accesses the fields. We also parse buffer lazily, the idea being that some hypothetical caller might not actually access all the fields.

We have this [`ProtoBytesParser`](https://github.com/albertlockett/otel-arrow/blob/53464ed23e17449c5a8122f72dd8ab692a44b746/rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/decode.rs#L95) which will take a single pass over the buffer and fill in the `FieldRanges` impl as it encounters the offsets of the fields as it's parsing. That way if a caller uses the view to access a field we've seen before, we can just return the range without any additional parsing. 

There's also the [`RepeatedFieldProtoBytesParser`](https://github.com/albertlockett/otel-arrow/blob/53464ed23e17449c5a8122f72dd8ab692a44b746/rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/decode.rs#L206) which acts as an iterator of byte slices that contain some repeated field.

For example: if we look up above at `RawMetricsData`, we can see that it contains a ProtoBytesParser which will fill in the ranges for `MetricFieldRanges` for fields contained in a `Metric` proto message some user calls methods on the `MetricView`. However, if the caller wants to access the attributes, which are a repeated field in the same buffer which contains the serialized `Metric` message, then we still end up parsing the same buffer searching for the repeated `KeyValue`s, but this is done using [`KeyValueIter`](https://github.com/albertlockett/otel-arrow/blob/53464ed23e17449c5a8122f72dd8ab692a44b746/rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/common.rs#L195) (which contains the `RepeatedFieldProtoBytesParser`).

So we have multiple proto bytes parsers that are parsing the same buffer, ideally they'll share the same `FieldRanges` instance. That avoids each one having to parse the buffer itself -- they can share the ranges of the fields they encounter.

Effectively, both parsers share the same instance of `FieldRanges`, but rust doesn't allow us to have both these own these `FieldRanges`. Instead, each one takes an `Rc`, from which we can only get a shared (non-`mut` reference). But crucially, we can still modify the value of these `Cells` if we only have a shared ref and so that's why they are used!

PR #1395, opened 2025-11-06
Move otel-arrow-rust into otap-dataflow

### LINKED ISSUE #1377: Consolidate otel-arrow-rust modules into otap-dataflow crates

jmacd:
Part of #867 
Fixes #1377

This is a move, no real code changes were required, only to replace `otel_arrow_rust::` with `crate::` (in crates/pdata) or `otap-df-pdata::` (in crates/otap). Top-level module files were merged with the existing, so several files were manually combined, but there were no name conflicts.

All relative paths change from otel-arrow-rust/${x} to otap-dataflow/crates/pdata/${x}.

It is a pleasure to finally combine these two codebases. Thanks to @v0y4g3r for the original code structure here. A lot has changed since the first commit!

This crate had a `#[allow(missing_docs)]` declaration. I've removed it from the crate-level and applied it in dozens of locations instead. I will file a new issue reminding us that there are a lot of missing docs in this crate; I think it will be easier to finish this documentation now that it's more solidified.

### DIFF
--- a/rust/README.md
+++ b/rust/README.md
@@ -25,68 +25,6 @@ elements in the OpenTelemetry data model, by signal. The OTAP pipeline
 also supports passing through OTLP bytes as literal data, with
 **direct conversion** between the OTAP records and OTLP bytes models.
 
-## OTel-Arrow Rust
-
-**[Sub-project README](./otel-arrow-rust/README.md)**
-
-The `otel-arrow-rust` folder contains the project's Rust reference
-implementation for OTel-Arrow, similar in nature to the [OTel-Arrow
-Golang library](../go/README.md) used by the project's Golang
-collector components.  This library translates between the following
-representations of OpenTelemetry:
-
-- OTAP records: represented using [Apache Arrow (arrow-rs)][ARROW_RS]
-  record batches
-- OTLP records: represented using [Prost][PROST_RS] message objects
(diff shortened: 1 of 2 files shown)


rust/otap-dataflow/crates/otap/src/encoder.rs
```diff
@@ -623,11 +624,8 @@ where
     ndp.append_parent_id(*curr_metric_id);
     ndp.append_start_time_unix_nano(Some(ndp_view.start_time_unix_nano() as i64));
     ndp.append_time_unix_nano(ndp_view.time_unix_nano() as i64);
-    let (double, integer) = match ndp_view.value() {
-        Some(metrics::Value::Double(val)) => (Some(val), None),
-        Some(metrics::Value::Integer(val)) => (None, Some(val)),
-        None => (None, None),
-    };
+    let double = ndp_view.value().and_then(|v| v.as_double());
+    let integer = ndp_view.value().and_then(|v| v.as_integer());
```
jmacd: I have a bad merge

jmacd: Manually fixed/checked.

PR #1402, opened 2025-11-07
Move OTAP encoding functions, OtapPayload struct & helpers to otap_df_pdata

jmacd:
I am working on testing the OTAP batching functionality and I want to use the encoding functions from otap_df_pdata, which is where they belong.

The otap_df_otap::pdata::{Context, OtapPdata} types remain, but about half of it moves to otap_df_pdata::payload.

Minor test cleanups, part of otap_df_otap::testing moves with it, some of the same moves to the one location it was used.

#1377

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/error.rs
+++ b/rust/otap-dataflow/crates/engine/src/error.rs
@@ -382,19 +382,17 @@ pub enum Error {
     /// Note: this is not a specific type such as otap_df_pdata::error::Error
     /// because this crate does not specifically depend on that crate.  We
     /// could use a dyn Error, maybe.
-    #[error("Pipeline data error: {}", error)]
+    #[error("Pipeline data error: {}", reason)]
     PDataError {
-        /// otap_df_pdata error
-        #[from]
-        error: otap_df_pdata::error::Error,
+        /// otap_df_pdata error string
+        reason: String,
     },
 
     /// Error from the prost encoder.
-    #[error("Prost encode error: {}", error)]
-    ProstEncodeError {
-        /// Prost error
-        #[from]
-        error: prost::EncodeError,
+    #[error("Prost encode error: {}", reason)]
+    ProtoEncodeError {
+        /// Prost error string
+        reason: String,
     },
 }
 
@@ -420,7 +418,7 @@ impl Error {
             Error::PipelineControlMsgError { .. } => "PipelineControlMsgError",
             Error::ProcessorAlreadyExists { .. } => "ProcessorAlreadyExists",
             Error::ProcessorError { .. } => "ProcessorError",
-            Error::ProstEncodeError { .. } => "ProstEncodeError",
+            Error::ProtoEncodeError { .. } => "ProtoEncodeError",
             Error::ReceiverAlreadyExists { .. } => "ReceiverAlreadyExists",
             Error::ReceiverError { .. } => "ReceiverError",
             Error::SpmcSharedNotSupported { .. } => "SpmcSharedNotSupported",
@@ -480,5 +478,21 @@ pub fn error_summary_from(err: &Error) -> ErrorSummary {
             message: err.to_string(),
             source: None,
         },
+    }
+}
+
+impl From<prost::EncodeError> for Error {
+    fn from(e: prost::EncodeError) -> Self {
+        Self::ProtoEncodeError {
+            reason: e.to_string(),
+        }
+    }
+}
+
+impl From<otap_df_pdata::error::Error> for Error {
+    fn from(e: otap_df_pdata::error::Error) -> Self {
+        Self::PDataError {
+            reason: e.to_string(),
+        }
     }
 }
--- a/rust/otap-dataflow/crates/pdata/src/payload.rs
+++ b/rust/otap-dataflow/crates/pdata/src/payload.rs
@@ -366,8 +366,9 @@ mod test {
             metrics::v1::{
                 AggregationTemporality, ExponentialHistogram, ExponentialHistogramDataPoint, Gauge,
                 Histogram, HistogramDataPoint, Metric, NumberDataPoint, ResourceMetrics,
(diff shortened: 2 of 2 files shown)


rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -316,7 +316,8 @@ impl local::Processor<OtapPdata> for AttributesProcessor {
                 let signal = pdata.signal_type();
                 let (context, payload) = pdata.into_parts();
 
-                let mut records: OtapArrowRecords = payload.try_into()?;
+                let mut records: OtapArrowRecords =
+                    payload.try_into().map_err(crate::pdata_to_engine_error)?;
```
albertlockett: I think if we implement the following in `otap_df_engine`, we won't need the `map_err` here. We will be able to continue using `?`
```rs
impl From<otap_df_pdata::error::Error> for crate::error::Error {
  fn from(e: otap_df_pdata::error::Error) -> Self {
     Self::PDataError {
        message: e.to_string(),
     }
  }
}
```

rust/otap-dataflow/crates/otap/src/fake_data_generator.rs
```diff
@@ -393,27 +394,35 @@ async fn generate_signal(
...
-                error: format!("error encoding protobuf: {e}"),
-            })
-        };
         let mut bytes = vec![];
         Ok(match value {
             OTLPSignal::Logs(logs_data) => {
-                logs_data.encode(&mut bytes).map_err(map_error)?;
+                logs_data
+                    .encode(&mut bytes)
+                    .map_err(encode_to_engine_error)?;
```
albertlockett: Similar to my other comment, I think if we implement the following in `otap_df_engine`, we won't need to use `map_err` everywhere here

```rs
impl From<prost::EncodeError> for crate::error::Error {
  fn from (e: prost::EncodeError) -> Self {
    Self::PDataError {
        message: e.to_string(),
    }
  }
}
```

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
jmacd: I wasn't sure about this because the engine does not depend on otap_df_pdata.

rust/otap-dataflow/crates/otap/src/fake_data_generator.rs
jmacd: ... and it does not depend on prost, either.

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
jmacd: I believe the thiserror `#[from]` attribute will automatically derive the From impl for you, as well. I will make this change.

rust/otap-dataflow/crates/engine/src/error.rs
```diff
@@ -377,38 +377,59 @@ pub enum Error {
...
+        #[from]
+        error: otap_df_pdata::error::Error,
+    },
+
+    /// Error from the prost encoder.
+    #[error("Prost encode error: {}", error)]
+    ProstEncodeError {
+        /// Prost error
+        #[from]
+        error: prost::EncodeError,
```
lquerel: I recommend avoiding leaking errors from external libraries in our own errors. The idea is to preserve as much flexibility as possible to change our dependencies with minimal impact.

I suggest to implement the trait `From<prost::EncodeError> for Error` (local) so you can keep the `?`.

jmacd: To be clear you don't mind the engine crate depending on the pdata crate, only the Prost one is a problem?

jmacd: > I suggest to implement the trait From<prost::EncodeError> for Error (local) so you can keep the ?.

I'm pretty confused! See also

https://github.com/open-telemetry/otel-arrow/pull/1402#discussion_r2505132091
https://github.com/open-telemetry/otel-arrow/pull/1402#discussion_r2505146094

I can't define this in the crate where it's used: neither of these types is from the same crate. 

I had a `.map_err(...)?` with a function defined in the create, no From<_> impls.

albertlockett: I think maybe @lquerel  was thinking we could do something like this -- e.g. not leak the prost::EncodeError, but just encapsulate the stringified message in an error type of our own creatin:
```rs
    /// Error from the proto encoder.
    #[error("Proto encode error: {reason}")]
    ProtoEncodeError {
        /// Proto error
        reason: String
    },
}

impl From<prost::EncodeError> for Error {
    fn from(e: prost::EncodeError) -> Self {
        Self::ProtoEncodeError { reason: e.to_string() }
    }
}
```

jmacd: And, to be clear, we can add a dependency on prost in the engine only for this reason?
May we consider Into<otap_df_engine::Error> for prost::EncodeError, implemented in otap_df_otap?

Likewise, why not create an Into<otap_df_engine::Error> alongside `otap_df_pdata::error::Error`, to avoid the engine depending on the pdata crate?

albertlockett: Unfortunately, rust won't let us do that because it could violate [coherence rules](https://doc.rust-lang.org/reference/items/implementations.html#trait-implementation-coherence)

> a trait implementation is only allowed if either the trait or at least one of the types in the implementation is defined in the current crate. It prevents conflicting trait implementations across different crates and is key to ensuring coherence.

image

If we wanted to avoid having otap_df_engine depend on prost, I wonder if we could implement `From<prost::EncodeError> for otap_df_pdata:error::Error` since we already have `From<otap_df_pdata::error::Error> for otap_df_engine::error::Error` (which produces a PdataError)?

jmacd:
I've added the explicit From<_> method approach, with `reason: String` fields to erase the type of the error.

The engine now depends on prost and otap_df_pdata for this reason alone.

PR #1403, opened 2025-11-07
OTLP Bytes -> OTAP encoding for Metric Exponential Histogram Data

albertlockett:
part of #387


rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/decode.rs
```diff
@@ -411,6 +440,13 @@ pub fn read_varint(buf: &[u8], mut pos: usize) -> Option<(u64, usize)> {
     None
 }
 
+/// Decode 32 bit zigzag encoding
+#[inline]
+#[must_use]
+pub fn decode_sint32(val: i32) -> i32 {
```
utpilla:
```suggestion
pub const fn decode_sint32(val: i32) -> i32 {
```

PR #1412, opened 2025-11-10
Use GITHUB_STEP_SUMMARY on release action dry_run

drewrelmas:
Inspired by usage in other repositories - nicer to avoid digging through detailed dry run logs to ensure release looks as expected.


lquerel (approved):
LGTM

Issue #1352, opened 2025-10-27
Filtering: optional include/exclude `LogMatchProperties`


albertlockett:
I was trying to use the new filtering code (https://github.com/open-telemetry/otel-arrow/pull/1341) to create filters that only include criteria for what I want to include (e.g., nothing should be excluded).

For example:
```rs
        let include = LogMatchProperties::new(
            MatchType::Strict,
            // include logs with any resource attributes
            vec![],
            // include all logs with any attributes
            vec![],

            // only include logs with severity_level = "ERROR"
            vec![
                "ERROR".into()
            ],
            None, // include all severity numbers
            vec![] // include all bodies
        );

        // don't exclude anything ...
        let exclude = LogMatchProperties::new(
            MatchType::Strict,
            vec![],
            vec![],
            vec![],
            None,
            vec![]
        );

        let filter = LogFilter::new(include, exclude, vec![]);
```

Unfortunately, this doesn't seem to give the desired behaviour. I think we have an opportunity to make the API easier to use, and also maybe improve performance.

I was thinking that `LogFilter::new` signature could be augmented to take `Option<LogMatchProperties>` for both include & exclude, then we could compute filters optionally for `inlcude` & `exclude` only if the the `LogMatchProperties` is `Some`. If they are both `Some`, only then do we need to `and` the filters together:
https://github.com/open-telemetry/otel-arrow/blob/36d4a6b782d67ae44d7324f6815064a5bce42731/rust/otel-arrow-rust/src/otap/filter/logs.rs#L108-L126

albertlockett:
@c1ly maybe you could have a look when you have some time?

albertlockett:
closed as completed in #1413

PR #1413, opened 2025-11-10
[otap-dataflow] allow optional exclude/include in the filter processor

c1ly:
Changed the include and exclude fields to be optional so that both don't have to be defined at the same time.

fixes issue described in https://github.com/open-telemetry/otel-arrow/issues/1352


lquerel (approved):
LGTM

jmacd (approved):

PR #1414, opened 2025-11-10
[otap-dataflow] remove redundant quorum enum

clhain:
The Quorum enum for configuring healthchecks has both an ::All and a ::Percent(u8) variant. These are redundant, and I'm running into some serialization errors. This removes the All variant and replaces all instances of it with Percent(100).


lquerel (approved):
LGTM

PR #1418, opened 2025-11-11
Remove duplicated documentation for OTAP pdata & document pdata module

albertlockett:
Looks like somehow in our refactoring of otel_arrow_rust / otap pdata we duplicated the documentation for the pdata module. The documentation this PR removes is also in [pdata.rs](https://github.com/open-telemetry/otel-arrow/blob/430a4835157fdbe9cbf98b75970ff971476a39af/rust/otap-dataflow/crates/otap/src/pdata.rs#L4) and we probably don't need it in both places and adds documentation for the pdata module

Part of #1422.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/pdata.rs
+++ b/rust/otap-dataflow/crates/otap/src/pdata.rs
@@ -1,6 +1,18 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+//! This module manages how telemetry pipeline data (pdata) flows through pipeline
+//! components (receivers, processors, exporters) and how components receive
+//! acknowledgments (Ack) and negative acknowledgments (Nack) about this pdata.
+//!
+//! This includes:
+//! - Context: a stack of pipeline component subscriptions
+//! - Interests: allows components to declare whether they subscribe to Ack/Nack
+//!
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/pdata/src/payload.rs
```diff
@@ -1,86 +1,6 @@
...
-//!                ],
-//!            ),
-//!        ],
-//!    ),
-//!  ]);
-//! let mut buf = Vec::new();
-//! otlp_service_req.encode(&mut buf).unwrap();
-//!
-//! // Create a new OtapPayload from OTLP bytes
-//! let payload: OtapPayload = OtlpProtoBytes::ExportLogsRequest(buf).into();
```
utpilla, line 52: This documentation is mainly targeting the `OtapPayload` enum which is defined in this file. Maybe we should remove the documentation from `pdata.rs` and not here?

albertlockett: ah you're right. let me make that change -- good catch @utpilla

albertlockett: @utpilla fixed!

albertlockett:
oops looks like this is now failing b/c the pdata module does not have documentation. I can try to add some documentation for this module tomorrow.

PR #1420, opened 2025-11-12
[otap-dataflow] support more advanced CPU allocation syntax

### LINKED ISSUE #1135: Support more advanced cpu allocation syntax
We should be able to specify more complex cpu range such as "0-3,5,7-9"

mx-psi:
Supports more advanced CPU allocation syntax.

For example, it now supports: `0..3,5,7-8` to specify the set `0,1,2,3,5,7,8`.

Fixes #1135

### DIFF
--- a/rust/otap-dataflow/crates/config/src/node.rs
+++ b/rust/otap-dataflow/crates/config/src/node.rs
@@ -23,7 +23,6 @@ use std::collections::{HashMap, HashSet};
 /// - The hyper-edge configuration determines which downstream nodes are connected,
 ///   and how messages are routed (broadcast, round-robin, ...).
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
-#[serde(deny_unknown_fields)]
 pub struct NodeUserConfig {
     /// The kind of this node, which determines its role in the pipeline.
     /// 4 kinds are currently specified:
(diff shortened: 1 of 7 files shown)


jmacd (approved):
Looks good! Thank you.

I briefly considered whether a Vec<Range> struct could be replaced by maybe a BTreeSet<usize> by expanding the range, but came around to thinking this is a more readable approach for the operator. 

I also asked myself whether select_cores_for_quota(), which is the first time where an invalid input will be flagged, could be called sooner, before we're in a function called run_forever(). However, that's not an issue with this PR, just a point of confusion about startup: I recall that component validation happens once per core, hence invalid component configuration also is first flagged somewhere underneath run_forever. It means we're being consistent on this point of confusion.

mx-psi:
Thanks for the review!

> I also asked myself whether select_cores_for_quota(), which is the first time where an invalid input will be flagged, could be called sooner, before we're in a function called run_forever(). However, that's not an issue with this PR, just a point of confusion about startup: I recall that component validation happens once per core, hence invalid component configuration also is first flagged somewhere underneath run_forever. It means we're being consistent on this point of confusion.

I considered this and came to the same conclusion :) I'd be happy to refactor that if that is of interest (right now I am just trying to learn more about the Arrow pipeline in general :smile:)

Issue #1399, opened 2025-11-06
Correct support for packed/expanded protobuf fields


albertlockett:
There are a handful of fields in OTLP protobuf that are repeated primitive types. HistograpDataPoint `bucket_counts` and `explicit_bounds` are examples.

In protobuf encoding, these can be either "packed" or "expanded". See the documentation here: https://protobuf.dev/programming-guides/encoding/#repeated

Prost, correctly, uses the packed encoding for these fields (which is the default in proto3). However, our direct OTAP -> OTLP bytes conversion does not use a packed encoding:
https://github.com/open-telemetry/otel-arrow/blob/99deb083032f33365657140d51358363e1b3bc83/rust/otap-dataflow/crates/pdata/src/otlp/metrics/data_points/histogram.rs#L193-L221

However in our OTLP bytes -> OTAP conversion, we support packed only (as of https://github.com/open-telemetry/otel-arrow/pull/1398). Also, the interpretation of packed values does not handle a particular edge case: where there are multiple packed values in the same message. The docs says this needs to be handled:
> Note that although there’s usually no reason to encode more than one key-value pair for a packed repeated field, parsers must be prepared to accept multiple key-value pairs. In this case, the payloads should be concatenated. Each pair must contain a whole number of elements. The following is a valid encoding of the same message above that parsers must accept:
```
6: {3 270}
6: {86942}
```

The docs also states:
> Protocol buffer parsers must be able to parse repeated fields that were compiled as packed as if they were not packed, and vice versa. This permits adding [packed=true] to existing fields in a forward- and backward-compatible way.

Clearly we need to correct this. To fix:
1. **[Most critical]** OTAP -> OTLP bytes should be made to use the packed encoding.
   - if there's only one value, we _could_ optionally use the expanded form to save 1 byte per field (by not encoding the length).
   - this is the most critical thing to fix b/c currently our proto decoder will not correctly read what is produced by the proto encoder
2. OTLP bytes -> OTAP should support:
   - decoding the expanded encoding
   - If multiple sequences of packed values are present.

The docs are not clear on whether proto encoders can mix packed & expanded in the same message, so our proto decoder probably does not need to handle this.

albertlockett:
This also applies to packed varints, such as ExponentialHistogramDatapoint's Bucket bucket_counts field

PR #1421, opened 2025-11-12
fix repeated primitive fields using different encodings when converting between OTLP bytes & OTAP

albertlockett:
part of #1399 

This fixes an issue caused by a mismatch encodings used repeated primitive proto fields. There are three fields of this type:
- HistogramDataPoint bucket_counts (repeated fixed64) & explicit_bounds (repeated double)
- ExpHistogramDataPoint Bucket bucket_counts (repeated uint64)

When we encoded OTAP -> OTLP bytes, we were writing the "expanded" encoding, but when decoding OTLP Bytes -> OTAP, we were expecting the "packed" encoding (for more details on the encodings see https://protobuf.dev/programming-guides/encoding/#repeated).

This corrects the encoding of OTAP -> OTLP bytes to use "packed", which is the default since proto Edition 2023 & in proto3 (it is also what prost uses).

In a followup PR, I will add correct the logic in our decoder to handle the expanded decoding and additional edge cases for packed encoding.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otlp/metrics/data_points/histogram.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otlp/metrics/data_points/histogram.rs
@@ -190,8 +190,6 @@ pub(crate) fn proto_encode_histogram_data_point(
         }
     }
 
-    // TODO should there be unit tests to ensure we write these using packed encoding?
-
     if let Some(bucket_counts) = &hist_dp_arrays.histogram_bucket_counts {
         if bucket_counts.list.is_valid(index) {
             let value_offsets = bucket_counts.list.value_offsets();


rust/otap-dataflow/crates/pdata/src/payload.rs
```diff
@@ -415,59 +415,77 @@ mod test {
...
         };
 
         let result = ExportMetricsServiceRequest::decode(bytes.as_ref()).unwrap();
         assert_eq!(otlp_service_request, result);
+
+        // check that we can also re-decode the OTLP proto bytes that we encode directly
+        // from OTAP, that we get the same result
+        let pdata: OtapPayload = otlp_bytes.into();
+        let otap_batch2: OtapArrowRecords = pdata.try_into().unwrap();
+        assert_eq!(otap_batch, otap_batch2);
```
albertlockett, line 488: This test was failing before the code was fixed.

I added the same assertions to the other signal types as well for good measure.

jmacd (approved):

PR #1424, opened 2025-11-12
Support expanded and packed encoding for repeated primitive fields when decoding protobuf

### LINKED ISSUE #1399: Correct support for packed/expanded protobuf fields

albertlockett:
closes #1399 

According to the [protobuf encoding docs](https://protobuf.dev/programming-guides/encoding/), there's two cases decoders need to handle:

> Protocol buffer parsers must be able to parse repeated fields that were compiled as packed as if they were not packed, and vice versa.

Also, if the encoding is "packed", we need to handle if there are more than one segment of the buffer containing packed values
> Note that although there’s usually no reason to encode more than one key-value pair for a packed repeated field, parsers must be prepared to accept multiple key-value pairs. In this case, the payloads should be concatenated. Each pair must contain a whole number of elements. The following is a valid encoding of the same message above that parsers must accept

This PR brings our decoding logic in line with the recommendations for the few fields we have that could have this encoding. Before this change, we were assuming the encoding was always packed. 

This change implements a new iterator for these types called `RepeatedPrimitiveIter` which encapsulates the logic of determining how to iterate the buffer and produce values correctly for both types of encodings. Internally, it use a combination of the `RepeatedFieldProtoBytesParser` and the packed field iterators which were already implemented before this change.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/metrics.rs
+++ b/rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/metrics.rs
@@ -910,8 +910,8 @@ impl FieldRanges for BucketsFieldRanges {
                 }
             }
             EXP_HISTOGRAM_BUCKET_BUCKET_COUNTS => {
-                if (wire_type == wire_types::LEN
-                    || wire_type == wire_types::VARINT) && self.first_bucket_count.get().is_none()
+                if (wire_type == wire_types::LEN || wire_type == wire_types::VARINT)
+                    && self.first_bucket_count.get().is_none()
                 {
                     let packed = wire_type == wire_types::LEN;
                     self.first_bucket_count.set(Some((range, packed)))


rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/decode.rs
```diff
@@ -417,6 +449,184 @@ impl<'a> Iterator for PackedVarintIter<'a> {
...
+    pos: Rc<Cell<usize>>,
+    field_ranges: Rc<T>,
+
+    // expected wire type for the repeated field .. e.g. if the field is type `repeated double`
+    // this would be wire_types::FIXED64
+    repeated_wire_type: u64,
+
+    // this will be used to iterate over either:
+    // - the repeated values (in case that the encoding is expanded)
+    // - the segments containing the packed fields (in case that the encoding si packed)
```
jmacd:
```suggestion
    // - the segments containing the packed fields (in case that the encoding is packed)
```

rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/decode.rs
```diff
@@ -417,6 +449,184 @@ impl<'a> Iterator for PackedVarintIter<'a> {
...
+    field_iter: Option<RepeatedFieldProtoBytesParser<'a, T>>,
+
+    // this will be initialized lazily if the encoding is packed. one instance of the packed
+    // encoder (type P) wil be created for each segment of the buffer containing packed values
+    packed_iter: Option<P>,
+
+    // flag for whether the field is packed or expanded. this will be initialized lazily
+    // as well once the encoding is determined
+    packed: bool,
```
jmacd, line 489: Interesting. Is the encoding ever mixed, I wonder? I would imagine this could be obtained by concatenating the bytes of protobuf messages that were encoded with different encodings. (It sounds awful, wouldn't mind letting errors happen if it does.)

jmacd (approved):

rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/decode.rs
albertlockett: The docs don't explicitly say whether it's allowed to be mixed. They only give examples using one encoding or the other https://protobuf.dev/programming-guides/encoding/#repeated

However, I get the impression based on how it's configured that mixing the encodings wouldn't be allowed. The packed/expanded encoding is configured at the level of the file (with defaults depending on the syntax/edition) and can be overridden for individual fields.
https://protobuf.dev/editions/features/#repeated_field_encoding

rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/metrics.rs
```diff
@@ -880,15 +910,30 @@ impl FieldRanges for BucketsFieldRanges {
                 }
             }
             EXP_HISTOGRAM_BUCKET_BUCKET_COUNTS => {
-                if wire_type == wire_types::LEN {
-                    self.bucket_counts.set(Some(range))
+                if wire_type == wire_types::LEN
+                    || wire_type == wire_types::VARINT && self.first_bucket_count.get().is_none()
```
utpilla: I think we need to put the `wire_type` checks in parentheses for the desired evaluation:

```suggestion
                if (wire_type == wire_types::LEN
                    || wire_type == wire_types::VARINT) && self.first_bucket_count.get().is_none()
```

rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/decode.rs
```diff
@@ -326,27 +326,48 @@ where
...
-/// packed encoding for these types of fields
-/// https://protobuf.dev/editions/features/#repeated_field_encoding
+/// can be encoded as a fixed64. e.g. `repeated double` or `repeated fixed64` and the field
+/// is encoded using packed encoding.
+///
+/// Note this is for internal use only. The proto documentation states that decoders should
+/// have the flexibility to accept both packed and expanded encodings. Unless we're sure that
+/// the field is packed encoded, it's safer to use [`RepeatedFixed64Iter`] which automatically
+/// handles both encodings.
 pub struct PackedFixed64Iter<'a, V: FromFixed64> {
```
utpilla, line 347: Not for this PR, but we could consider implementing [`ExactSizeIterator`](https://doc.rust-lang.org/std/iter/trait.ExactSizeIterator.html) trait for this struct to make the `collect` calls from the consumer of this struct more efficient.

PR #1426, opened 2025-11-13
[otap-dataflow] add trace support to the filter processor

c1ly:
Added trace support

Defined two new structs that will help define and create filters for trace data

```rust
/// struct that describes the overall requirements to use in order to filter traces
pub struct TraceFilter {
    // Include match properties describe traces that should be included in the Collector Service pipeline,
    // all other traces should be dropped from further processing.
    include: Option<TraceMatchProperties>,
    // Exclude match properties describe traces that should be excluded from the Collector Service pipeline,
    // all other traces should be included.
    exclude: Option<TraceMatchProperties>,
    // ToDo: Add ottl support -> see golang version https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/8558294afa723b9ed917ed5d6bb6c656bb096a49/processor/filterprocessor/config.go#L90
}

/// TraceMatchProperties specifies the set of properties in a trace to match against and the type of string pattern matching to use.
pub struct TraceMatchProperties {
    // MatchType specifies the type of matching desired
    #[serde(default = "default_match_type")]
    match_type: MatchType,

    // ResourceAttributes defines a list of possible resource attributes to match traces against.
    #[serde(default)]

    // SpanAttributes defines a list of possible record attributes to match traces against.
    #[serde(default)]
    span_attributes: Vec<KeyValue>,

    // SpanNames is a list of span names that the span's names field must match against.
    span_names: Vec<String>,

    // EventNames is a list of event names that the span's events field must match against.
    #[serde(default)]
    event_names: Vec<String>,

    // EventsAttributes is a list of possible event attributes to match span events against.
    // A match occurs of any event attribute matches at least one expression in this given list.
    #[serde(default)]
    event_attributes: Vec<KeyValue>,

    // LinkAttributes is a list of possible link attributes to match span links against.
    // A match occurs of any link attribute matches at least one expression in this given list.
    #[serde(default)]
    link_attributes: Vec<KeyValue>,
}
```

When filtering traces we can filter based off `resource_attributes`, `span_attributes`, `span_names`, `event_names`, `event_attributes` and `link_attributes`

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/testing/mod.rs
+++ b/rust/otap-dataflow/crates/otap/src/testing/mod.rs
@@ -61,7 +61,8 @@ impl TryFrom<CallData> for TestCallData {
 /// Create minimal test pdata
 #[must_use]
 pub fn create_test_pdata() -> OtapPdata {
-    let otlp_service_req = otap_df_pdata::testing::create_test_logs();
+    // Note this has to be one log record for existing tests.
+    let otlp_service_req = otap_df_pdata::testing::fixtures::log_with_no_scope();
     let mut otlp_bytes = vec![];
     otlp_service_req.encode(&mut otlp_bytes).unwrap();
 
--- a/rust/otap-dataflow/crates/pdata/src/encode/mod.rs
+++ b/rust/otap-dataflow/crates/pdata/src/encode/mod.rs
@@ -64,6 +64,8 @@ where
     let mut events = EventsRecordBatchBuilder::new();
     let mut links = LinksRecordBatchBuilder::new();
 
+    let mut total_span_count = 0;
+
     // First, we traverse the view collecting the trace data into our RecordBatch builders.
 
     #[allow(clippy::explicit_counter_loop)]
@@ -95,7 +97,9 @@ where
         }
 
         for scope_spans in resource_spans.scopes() {
-            spans.append_schema_url_n(scope_spans.schema_url(), scope_spans.spans().count());
+            let span_count = scope_spans.spans().count();
+            total_span_count += span_count;
+            spans.append_schema_url_n(scope_spans.schema_url(), span_count);
             if let Some(scope) = scope_spans.scope() {
                 for kv in scope.attributes() {
                     scope_attrs.append_parent_id(&curr_scope_id);
@@ -145,12 +149,14 @@ where
                 spans.append_dropped_events_count(Some(span.dropped_events_count()));
                 spans.append_dropped_links_count(Some(span.dropped_links_count()));
 
-                spans
-                    .status
-                    .append_code(span.status().map(|s| s.status_code()));
-
-                if let Some(status) = span.status() {
-                    spans.status.append_status_message(status.message());
+                if let Some(status) = &span.status() {
+                    let code = status.status_code();
+                    let message = status.message();
+                    spans.status.append_code((code != 0).then_some(code));
+                    spans.status.append_status_message(message);
+                } else {
+                    spans.status.append_code(None);
+                    spans.status.append_status_message(None)
                 }
 
                 for event in span.events() {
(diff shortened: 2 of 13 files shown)


lquerel (approved):
LGTM.

There are lint issues to fix.

rust/otap-dataflow/crates/pdata/src/otap/filter/traces.rs
```diff
@@ -0,0 +1,1094 @@
...
+            span_attr_filter,
+            span_event_filter,
+            span_event_attr_filter,
+            span_link_attr_filter,
+        ))
+    }
+
+    /// Creates a booleanarray that will filter a resource_attribute record batch based on the
+    /// defined resource attributes we want to match
+    fn get_resource_attr_filter(&self, traces_payload: &OtapArrowRecords) -> Result<BooleanArray> {
```
albertlockett: I think the code in this method is duplicated from `otap::filter::logs::LogMatchProperties::get_filter_resources`, right?

Maybe we could extract this code into a shared function to make any future maintenance easier.

Similary, could we do the same thing for `get_span_attr_filter` below and `otap::filter::logs::LogMatchProperties::get_log_attr_filter` ?

rust/otap-dataflow/crates/pdata/src/otap/filter/traces.rs
```diff
@@ -0,0 +1,1094 @@
...
+                        NO_RECORD_BATCH_FILTER_SIZE,
+                    )));
+                } else {
+                    // if we did match on record attributes then there are no attributes to match
+                    return Ok(BooleanArray::from(BooleanBuffer::new_unset(
+                        NO_RECORD_BATCH_FILTER_SIZE,
+                    )));
+                }
+            }
+        };
```
albertlockett, line 642: I have two suggestions here:

1. Optimize creating the "placeholder" we create when there is no filtering to do

In the event that we don't need to do any filtering, it looks like we're still allocating a boolean buffer of size one. It would be nice if we could optimize what we return as a placeholder. 

The simplest option, which would involve minimal refactoring would be to change `NO_RECORD_BATCH_FILTER_SIZE` to 0 (unless it needs to be 1 for some reason?). This avoids a heap allocation because internally, `BooleanBuffer::new_unset(len)` creates a MutableBuffer which won't reserve any heap space if the length is zero:
https://github.com/apache/arrow-rs/blob/9d7dde2722a8392996a1f5a978e989b2fea4ed0e/arrow-buffer/src/buffer/mutable.rs#L83-L109

However, a better option could also be to return  `Result<Option<BooleanArray>>`, and here. Returning `Ok(None)` makes the intention clear that there's no actual filtering to be done for filters applied to this payload type.

2. Regardless of if/how we choose to optimize how the placeholder is generated, I think we can simplify this code. For example:
```rs
if self.event_names.is_empty() {
   // return no filter placeholder, e.g.
   return Ok(BooleanArray::from(BooleanBuffer::new_set(
      NO_RECORD_BATCH_FILTER_SIZE,
    ))); // or Ok(None)
}

let span_events = match traces_payload.get(ArrowPayloadType::SpanEvents) {
    Some(rb) => rb,
    None => {
        // return no filter placeholder ...
    }
};
```

c1ly: right now we just set it to one to a handle two scenarios
1. record batch doesn't exist and the user tries to filter on non existent record batch -> then we know that everything will end up being a false boolean array
2. record batch doesn't exist and the user doesn't intend to filter on non existent record batch -> then we ignore this record batch

albertlockett: oh I got it -- sorry I didn't notice the difference between new_set and new_unset here. Thanks for the clarification

albertlockett (approved):

PR #1428, opened 2025-11-13
Batch processor testing, comments

jmacd:
This contains several edge-case improvements in the crates/pdata batching logic.

Adds new batch round-trip testing framework and a small number of new tests. More testing with synthetic data and specific test fixtures will be necessary.

Adds otap_df_pdata::testing::DataGenerator as a placeholder for future work. This enables the batching tests to obtain multiple payloads with distinct timestamps, presently. This should be expanded to cover more variation in the data, maybe following the approach used in Phase 1's go/pkg/datagen, maybe using OTel-Weaver.

Modifies the public grouping API to take SignalType and restrict to one signal per call. This was already being handled in the crates/otap batch_processor. The code maintains the same structure, just returns an error if there is mixed telemetry data. We can revisit this later, otherwise the batching logic was treating the signals as logically separate anyway, so the wider API was not providing any benefit.

Note: This PR previously included refactoring in the batch processor, but it was not required. The component is modified only for the changed API: one #[ignore] test was removed (had mixed signals), one test was added to maintain coverage (for metrics).

The changes in groups.rs mainly address a degenerate case in logs (no ID b/c no LogsAttributes) and defects in metric batching:

- generic_split(): handle missing ID column gracefully; no ID means no generic_split_helper (renamed parent_child_split)
- split_metric_batches(): extensive comments, minor reorganization, more maintainable, limitations documented
- generic_concatenate(): refactor to simplify, comments
- reindex_record_batch(): handle missing ID column

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap/groups.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/groups.rs
@@ -1030,7 +1030,7 @@ fn generic_schemaless_concatenate<const N: usize>(
         }
     }
 
-    assert_all_empty(&batches);
+    assert_all_empty(batches);
     Ok(result)
 }
 
--- a/rust/otap-dataflow/crates/pdata/src/testing/fixtures.rs
+++ b/rust/otap-dataflow/crates/pdata/src/testing/fixtures.rs
@@ -785,6 +785,7 @@ pub struct DataGenerator {
 
 impl DataGenerator {
     /// Generate N 'limit' number of items
+    #[must_use]
     pub fn new(limit: usize) -> Self {
         Self {
             limit,


rust/otap-dataflow/crates/pdata/src/otap/groups.rs
```diff
@@ -792,14 +908,24 @@ fn sort_record_batch(rb: RecordBatch, how: HowToSort) -> Result<RecordBatch> {
...
                     values: id_values,
                     options,
                 }]
             }
-            _ => unreachable!(),
+            (_, _, None) => {
+                // No id or parent_id columns, no sorting required.
+                return RecordBatch::try_new(schema, columns)
+                    .map_err(|e| Error::Batching { source: e });
+            }
```
jmacd, line 928: I am suspicious of the changes here. I had uncovered a way to hit the unreachable! below, and after making these changes find myself a bit confused by the comment reproduced below (from above).

rust/otap-dataflow/crates/pdata/src/otap/groups.rs
jmacd, line 1418: Dubious comment here.

rust/otap-dataflow/crates/pdata/src/testing/fixtures.rs
```diff
@@ -766,3 +766,204 @@ pub fn metrics_multiple_sums_no_resource() -> MetricsData {
...
+                        .finish(),
+                ],
+            )],
+        )])
+    }
+
+    /// Generate test OTLP metrics data at a timestamp offset
+    #[must_use]
+    pub fn generate_metrics(&mut self) -> MetricsData {
+        // TODO: @@@
```
jmacd, line 907: This is where I'm not finished. I was working to exercise the point-count-over-limit case, which isn't tested but clearly needs it. I would be glad to finish the testing fixture work here.

rust/otap-dataflow/crates/pdata/src/otap/groups.rs
lquerel: Not sure to understand the comment neither.

lquerel (approved):
I did a quick review. I know you know this area inside out and I don't want to slow you down, so I'm approving it.

PR #1430, opened 2025-11-13
Create OtlpProtoMessage type; add equivalence testing for OTLP messages

jmacd:
Moves the OTLPSignal type from fake_data_generator, give it a name following the otap-df-pdata pattern, an enum named OtlpProtoMessage with variants for {Logs,Metrics,Traces}Data by signal type.

Adds a OtlpProtoMessage::batch_length() definition, equivalent to OtapArrowRecords::batch_length() used in the batching logic. This is like the Collector's definition for item count. Gives it a signal_type().

The equivalence tests are modeled after `go/pkg/otel/assert/equiv.go` except instead of using JSON encoding, we use Protobuf encoding. The algorithm is:

- Split each data into single-item message objects
- Canonicalize in place, recursively
- To canonicalize message objects: serialize to [u8], sort, deserialize
- Sort non-semantic-ordered fields (attributes, links, events)

The sibling PR #1428 demonstrates how the new test functions are used, including calls to batch_length() and signal_type().


rust/otap-dataflow/crates/pdata/src/proto/mod.rs
```diff
@@ -83,3 +83,223 @@ pub mod opentelemetry {
...
+/// objects are required. OtlpProtoBytes has an efficient translation
+/// into OtapArrowRecords, this type does not.
+///
+///
+/// Note this could be considered for #[cfg(test)], however we is are
+/// aware of uses in otap-df-otap's fake_signal_generator and
+/// debug_processor.
+#[derive(Clone, Debug)]
+pub enum OtlpProtoMessage {
+    /// Llogs data.  This is equivalent to ExportLogsServiceRequest.
```
utpilla:
```suggestion
    /// Logs data. This is equivalent to ExportLogsServiceRequest.
```

rust/otap-dataflow/crates/pdata/src/proto/mod.rs
```diff
@@ -83,3 +83,223 @@ pub mod opentelemetry {
...
+
+/// Protocol message data of some type.
+///
+/// Generally, callers should to use of OtlpProtoBytes type defined in
+/// crate::otlp instead, this is only useful where proto::Message
+/// objects are required. OtlpProtoBytes has an efficient translation
+/// into OtapArrowRecords, this type does not.
+///
+///
+/// Note this could be considered for #[cfg(test)], however we is are
```
utpilla:
```suggestion
/// Note this could be considered for #[cfg(test)], however we are
```

rust/otap-dataflow/crates/pdata/src/proto/mod.rs
```diff
@@ -83,3 +83,223 @@ pub mod opentelemetry {
         pub mod v1;
     }
 }
+
+/// Protocol message data of some type.
+///
+/// Generally, callers should to use of OtlpProtoBytes type defined in
```
utpilla:
```suggestion
/// Generally, callers should use the OtlpProtoBytes type defined in
```

utpilla (approved):

PR #1431, opened 2025-11-14
OtapArrowRecords: rename tag to signal_type

jmacd:
Simple cleanup.
This tag() is confusingly named since we renamed OtapArrowRecordTag to the consolidated SignalType #1397.
I gave it a signal_type() method, it's something all of OtapArrowRecords, OtlpProtoBytes, etc implement.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap.rs
@@ -95,7 +95,7 @@ impl OtapArrowRecords {
 
     /// The signal type of this data.
     #[must_use]
-    pub const fn signal_type(&self) -> SignalType {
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/pdata/src/otap.rs
```diff
@@ -93,8 +93,9 @@ impl OtapArrowRecords {
         }
     }
 
+    /// The signal type of this data.
     #[must_use]
-    const fn tag(&self) -> SignalType {
+    pub const fn signal_type(&self) -> SignalType {
```
utpilla:
```suggestion
    const fn signal_type(&self) -> SignalType {
```

PR #1436, opened 2025-11-14
Improve edge-case handling for OTLP-to-OTAP with missing resource/scope/attributes

jmacd:
This uses pieces of #1428 plus #1430, corrects the handling of missing resource/scope tables primarily.

Adds a new otap_df_pdata::testing round-trip testing facility
Adds a new testing::fixtures for edge-case data.

Tests round trips with optional scope and resource columns. Observed two small bugs with trace::Status objects, fixed (a fieldname typo and an array-length mismatch).

Tests that we drop "empty envelopes", see https://github.com/open-telemetry/opentelemetry-proto/issues/598, otherwise these would cause various failures as we build length=0 record batches.

(There are many tests that have lengthy, hard-coded tests like these. It would benefit us to consolidate these further, see payload.rs, retry_processor.rs, etc. I resisted the urge to continue refactoring tests, more to come!)

### DIFF
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -61,8 +61,8 @@ jobs:
     - uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
       with:
         go-version: "1.25.4"
-    - uses: github/codeql-action/init@0499de31b99561a6d14a36a5f662c2a54f91beee # v4.31.2
+    - uses: github/codeql-action/init@014f16e7ab1402f30e7c3329d33797e7948572db # v4.31.3
       with:
         languages: go
-    - uses: github/codeql-action/analyze@0499de31b99561a6d14a36a5f662c2a54f91beee # v4.31.2
+    - uses: github/codeql-action/analyze@014f16e7ab1402f30e7c3329d33797e7948572db # v4.31.3
       timeout-minutes: 60
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@0499de31b99561a6d14a36a5f662c2a54f91beee # v4.31.2
+        uses: github/codeql-action/upload-sarif@014f16e7ab1402f30e7c3329d33797e7948572db # v4.31.3
         with:
           sarif_file: results.sarif
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -40,7 +40,7 @@ jobs:
         with:
           toolchain: stable
       - name: install cargo-llvm-cov and nextest
-        uses: taiki-e/install-action@44c6d64aa62cd779e873306675c7a58e86d6d532 # v2.62.49
+        uses: taiki-e/install-action@537c30d2b45cc3aa3fb35e2bbcfb61ef93fd6f02 # v2.62.52
         with:
           tool: cargo-llvm-cov,cargo-nextest
       - name: Build the test collector
--- a/rust/experimental/query_engine/engine-recordset-otlp-bridge/tests/otlp_kql_recordset.rs
+++ b/rust/experimental/query_engine/engine-recordset-otlp-bridge/tests/otlp_kql_recordset.rs
@@ -418,6 +418,51 @@ fn test_strlen_function() {
 }
 
 #[test]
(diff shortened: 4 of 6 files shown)


jmacd:
cc/ https://github.com/open-telemetry/opentelemetry-proto/issues/598

rust/otap-dataflow/crates/pdata/src/encode/mod.rs
```diff
@@ -145,12 +149,14 @@ where
...
-                    .status
-                    .append_code(span.status().map(|s| s.status_code()));
-
-                if let Some(status) = span.status() {
-                    spans.status.append_status_message(status.message());
+                if let Some(status) = &span.status() {
+                    let code = status.status_code();
+                    let message = status.message();
+                    spans.status.append_code((code != 0).then_some(code));
+                    spans.status.append_status_message(message);
```
utpilla, line 156: If we are appending `None` for status code when it's zero, then should we also append `None` as the message in that case? We ideally shouldn't receive such kind of message though.

jmacd: I wasn't sure what the "right" way to approach this field is. I think that Arrow supports the validity bit for structs, so that we could an optional struct with non-nullable fields.

But, (and probably we should document this position), I think our aim in OTel-Arrow has been to be faithful to the OTLP protocol messages themselves. We're not in the business of validating OTLP data, we just want to round-trip it and yield equivalent protocol data.

Google protobufs have flipped on how to treat optional message fields more than once, and `proto3` dictates that the option bit is semantically required, meaning you should be able to tell the difference between an absent Status and a present Status with no fields set.

jmacd: (To be clear: I do not feel strongly about this proto3 optional struct behavior. I do not think it makes a difference to OpenTelemetry consumers.)

albertlockett: > proto3 dictates that the option bit is semantically required,

Is this _option bit_ something in the serialized protobuf message? I can't find any documentation that.

Interestingly the docs make it seem like the field is optional, but I'm not sure off-hand if this means the message could just be excluded from the serialized protobuf payload?
https://github.com/open-telemetry/opentelemetry-proto/blob/a8951735f7801e8adfaec5c0ace9262771cfec6e/opentelemetry/proto/trace/v1/trace.proto#L301-L303 

I'm moreso just asking for my own edification. I'm fine with how this works. Having the struct be nullable in OTAP I don't think actually gains us anything. We'd just have an extra null buffer for the validity bitmap for the struct and if the struct is null we're semantically just supposed to treat it has having the default values, which we're setting anyway.

rust/otap-dataflow/crates/pdata/src/otlp/common.rs
```diff
@@ -90,7 +89,19 @@ impl<'a> TryFrom<&'a RecordBatch> for ResourceArrays<'a> {
     type Error = Error;
 
     fn try_from(rb: &'a RecordBatch) -> Result<Self> {
-        let struct_array = get_required_array(rb, consts::RESOURCE)?;
+        // Resource column is nullable - if it's missing, return all None values
```
albertlockett:
```suggestion
        // Resource column is optional - if it's missing, return all None values
```

rust/otap-dataflow/crates/pdata/src/otlp/common.rs
```diff
@@ -141,12 +152,25 @@ impl<'a> TryFrom<&'a RecordBatch> for ScopeArrays<'a> {
     type Error = Error;
 
     fn try_from(rb: &'a RecordBatch) -> Result<Self> {
-        let struct_array = get_required_array(rb, consts::SCOPE)?;
+        // Scope column is nullable - if it's missing, return all None values
```
albertlockett:
```suggestion
        // Scope column is optional - if it's missing, return all None values
```

rust/otap-dataflow/crates/pdata/src/encode/mod.rs
jmacd: We're in a rabbit hole, for sure. 

https://protobuf.dev/programming-guides/proto3/

has a line about this ("In proto3, message-type fields already have field presence. Because of this, adding the optional modifier doesn’t change the field presence for the field."), linking to

https://protobuf.dev/programming-guides/field_presence/

OTLP has a few explicit `optional` fields, like histogram Sum/Min/Max fields; for these fields the proto encoder is expected to encode 0-values as a presence indicator, where otherwise they would skip encoding.

Before I added the workaround in the Status equiv checker, there were two extra bytes on one side, the Status Tag byte and varint 0 byte (length) indicating explicit presence. While this is meaningful in the strict proto3 sense, it's not meaningful to OpenTelemetry, so this sort of exception can live in our equivalence checker IMO.

PR #1444, opened 2025-11-18
Nit: Don't fail CI when perf label is missing

cijothomas:
The pipeline perf workflow currently fails when the label isn't present, making every new PR show a red check by default. This feels "less-welcoming" for new contributors who haven't done anything wrong, and might confuse them.

Changed the label-check job to skip gracefully instead of failing when the label is missing. The perf tests still only run when the label is present, but PRs no longer appear broken by default.

(If this was a intentional design, feel free to close this PR)

### DIFF
--- a/rust/otap-dataflow/src/README.md
+++ b/rust/otap-dataflow/src/README.md
@@ -12,13 +12,13 @@ cargo build --release
 cargo run --release
 
 # Run with custom configuration file
-cargo run --release -- -p configs/otlp-otlp.json
+cargo run --release -- -p configs/otlp-otlp.yaml
 
 # Run with custom core count
 cargo run --release -- --num-cores 4
 
 # Run with both custom config and cores
-cargo run --release -- -p configs/otlp-otlp.json --num-cores 8
+cargo run --release -- -p configs/otlp-otlp.yaml --num-cores 8
 
 # Get help
 cargo run --release -- --help


albertlockett (approved):

lquerel:
[if I misunderstood the purpose of this change, I apologize. I just want to make sure we have the right safeguards in place to avoid unpleasant surprises]

Are we once again in a situation where we'll see a performance regression after a merge into main?
I'm not sure there's a better solution than using a red check. If there is, I'd be interested, but to me the cost of discovering a performance regression late, after several merges, is far higher than the slightly unwelcoming aspect of a red check.

clhain:
The flaky marker / placeholder "pending" thing was all I could come up with - but it was causing too many problems. I'm open to other solutions too - but agree a stronger visual queue is needed.

FWIW it looks like only 3 of the last weeks worth of PRs got tagged, even with the red check - including a few with e.g. config schema changes.

cijothomas:
> The flaky marker / placeholder "pending" thing was all I could come up with - but it was causing too many problems. I'm open to other solutions too - but agree a stronger visual queue is needed.
> 
> FWIW it looks like only 3 of the last weeks worth of PRs got tagged, even with the red check - including a few with e.g. config schema changes.

Can we rely on maintainers to trigger this if the changes are potentially perf impacting before hitting merge button. Or we can auto-trigger this whenever code (.rs) files or engine directory is modified.

(I might be missing some context. Totally okay to revert this while we decide on something better.)

PR #1446, opened 2025-11-18
handle dict encoded attribute parent IDs when adding transport optimized encoding

### LINKED ISSUE #1442: `apply_transport_optimized_encodings` does not handle correctly `Dict` encoded ids
There should be a case here to handle `Dict` encoded IDs and parent ID columns

https://github.com/open-telemetry/otel-arrow/blob/4aeb8c9e44698e18994657b8ab4f0c5e79cc5e03/rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize.rs#L823-L837

albertlockett:
fixes #1442 

Fixes a bug in the code that adds the transport optimized encoding to the parent_id column of attributes record batches where the parent_id is a u32 type. For attributes, this column may be dictionary encoded which was not being handled by the code which adds the transport optimized encoding.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize.rs
@@ -1156,7 +1156,6 @@ where
     let val_bool_arr = record_batch.column_by_name(consts::ATTRIBUTE_BOOL);
     let val_bytes_arr = record_batch.column_by_name(consts::ATTRIBUTE_BYTES);
 
-    // let mut encoded_parent_ids = PrimitiveArray::<T>::builder(record_batch.num_rows());
     let mut encoded_parent_ids =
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize.rs
```diff
@@ -1087,7 +1156,9 @@ where
     let val_bool_arr = record_batch.column_by_name(consts::ATTRIBUTE_BOOL);
     let val_bytes_arr = record_batch.column_by_name(consts::ATTRIBUTE_BYTES);
 
-    let mut encoded_parent_ids = PrimitiveArray::<T>::builder(record_batch.num_rows());
+    // let mut encoded_parent_ids = PrimitiveArray::<T>::builder(record_batch.num_rows());
```
jmacd:
```suggestion
```

Issue #1445, opened 2025-11-18
Roundtrip conversion of `AnyValue`s not correct if attribute values are all default value


albertlockett:
For example, if I create a set of `LogRecord`s whose body are all empty string, this will get encoded to an OTAP Record batch as:
```
+----+----------+--------------------------------------------+----------------------------+----------------------------+-----------+------------+
| id | resource | scope                                      | time_unix_nano             | observed_time_unix_nano    | body      | event_name |
+----+----------+--------------------------------------------+----------------------------+----------------------------+-----------+------------+
| 0  | {id: 0}  | {id: 0, name: fake_signal, version: 1.0.0} | 2025-11-18T13:30:34.749050 | 2025-11-18T13:30:34.749050 | {type: 1} | app.log    |
| 1  | {id: 0}  | {id: 0, name: fake_signal, version: 1.0.0} | 2025-11-18T13:30:34.749052 | 2025-11-18T13:30:34.749052 | {type: 1} | app.log    |
+----+----------+--------------------------------------------+----------------------------+----------------------------+-----------+------------+
```

Note that the `body` column does correctly identify the `type` as `1`, which is a `string`. But the `str` column is omitted from this struct record batch because its contents would be entirely the default value of `""`. In OTAP, we omit optional columns if their contents are entirely null or default value.

The conversion back to OTLP in this case is not correct -- the body gets set to `null` but it should get set to `""`

jmacd:
It's important to note that AnyValue contains a `oneof` field, and oneof fields require "explicit presence" to carry type information even when they have a default value.

PR #1447, opened 2025-11-18
Support protobuf encoding default `AnyValue`s from missing optional columns

albertlockett:
fixes: #1445 

We have a handful of columns that that represent `AnyValue`'s value that are optional columns. If the column is full of default values, we omit these columns. However there was a bug in the encoding of these fields into OTLP protobuf, we treated the missing column as a null value which is not correct if the value_type column is not `ValueType::Empty`.

These values must be encoded in OTLP, despite their default value, because they are protobuf `oneof` fields.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otlp/attributes.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otlp/attributes.rs
@@ -120,6 +120,8 @@ pub(crate) fn encode_any_value(
             result_buf.encode_string(ANY_VALUE_STRING_VALUE, val);
         }
         AttributeValueType::Bool => {
+            // TODO handle case when bool column is missing we correct the default value handling
+            // https://github.com/open-telemetry/otel-arrow/issues/1449
             if let Some(attr_bool) = &attr_arrays.attr_bool {
                 if let Some(val) = attr_bool.value_at(index) {
                     result_buf.encode_field_tag(ANY_VALUE_BOOL_VALUE, wire_types::VARINT);
@@ -190,6 +192,9 @@ mod test {
         rb_builder.append_bytes(b"");
         rb_builder.append_double(0.0);
         rb_builder.append_int(0);
+
+        // TODO include test cases for bool once we've corrected the default value handling:
+        // https://github.com/open-telemetry/otel-arrow/issues/1449
 
         let mut fields = vec![];
         let mut columns = vec![];


rust/otap-dataflow/crates/pdata/src/otlp/attributes.rs
```diff
@@ -127,27 +128,30 @@ pub(crate) fn encode_any_value(
...
-                }
-            }
+            let val = attr_arrays
+                .attr_bytes
+                .as_ref()
+                .and_then(|col| col.slice_at(index))
+                .unwrap_or_default();
+            result_buf.encode_bytes(ANY_VALUE_BYTES_VALUE, val);
         }
         AttributeValueType::Map | AttributeValueType::Slice => {
```
utpilla, line 158: Do we not need to do the same for these types?

rust/otap-dataflow/crates/pdata/src/otlp/attributes.rs
```diff
@@ -163,3 +167,59 @@ pub(crate) fn encode_any_value(
...
+        }
+
+        let results = ArrayValue::decode(protobuf.as_ref()).unwrap().values;
+        let expected = vec![
+            AnyValue::new_string(""),
+            AnyValue::new_bytes(b""),
+            AnyValue::new_double(0.0),
+            AnyValue::new_int(0),
+        ];
+
```
utpilla, line 245: Maybe also add assertions to confirm that the columns are still omitted?

```rust
assert!(rb.column_by_name("attr_str").is_none(), "String column should be omitted");
assert!(rb.column_by_name("attr_bytes").is_none(), "Bytes column should be omitted");
...
```

rust/otap-dataflow/crates/pdata/src/otlp/attributes.rs
```diff
@@ -163,3 +167,59 @@ pub(crate) fn encode_any_value(
...
+    use crate::proto::opentelemetry::common::v1::{AnyValue, ArrayValue};
+
+    #[test]
+    fn test_default_anyvalue_encoded_when_column_missing() {
+        // append a bunch of "default" values
+        let mut rb_builder = AnyValuesRecordsBuilder::new();
+        rb_builder.append_str(b"");
+        rb_builder.append_bytes(b"");
+        rb_builder.append_double(0.0);
+        rb_builder.append_int(0);
```
utpilla, line 194: Should we add test for `bool` as well since it's treated differently?

rust/otap-dataflow/crates/pdata/src/otlp/attributes.rs
jmacd: I suspect yes.

I would say this follows from https://protobuf.dev/programming-guides/proto3/
because AnyValue is a oneof field

> If you set a oneof field to the default value (such as setting an int32 oneof field to 0), the “case” of that oneof field will be set, and the value will be serialized on the wire.

albertlockett: I don't think we want to handle this the same way we do for the primitive types.

This attr_ser column is an optional binary column, which OTAP encoding will omit if either all values are null, or they're all zero-length byte arrays.

But for maps and lists, what ends up in the `attr_ser` array is cbor encoded. Even a default (empty) map or list will still be serialized as a non zero length value with this encoding scheme.

For example:
```rs
let log_rec = LogsData {
    resource_logs: vec![ResourceLogs {
        scope_logs: vec![ScopeLogs {
            log_records: vec![
                LogRecord {
                    body: Some(AnyValue::new_array(ArrayValue::default())),
                    ..Default::default()
                },
                LogRecord {
                    body: Some(AnyValue::new_kvlist(KeyValueList::default())),
                    ..Default::default()
                }
            ],
            ..Default::default()
        }],
        ..Default::default()
    }],
    ..Default::default()
};

let otap_batch = encode_logs(&log_rec);
let logs_rb = otap_batch.get(ArrowPayloadType::Logs).unwrap();
arrow::util::pretty::print_batches(&[logs_rb.clone()]).unwrap();
```

prints:
```
+----------+---------+---------------------+-------------------------+----------------------+
| resource | scope   | time_unix_nano      | observed_time_unix_nano | body                 |
+----------+---------+---------------------+-------------------------+----------------------+
| {id: 0}  | {id: 0} | 1970-01-01T00:00:00 | 1970-01-01T00:00:00     | {type: 6, ser: 9fff} |
| {id: 0}  | {id: 0} | 1970-01-01T00:00:00 | 1970-01-01T00:00:00     | {type: 5, ser: bfff} |
+----------+---------+---------------------+-------------------------+----------------------+
```

It would actually be unusual if the attribute type column indicated the the value type should be map/list and then the column wasn't present in the attributes record batch, so I think we should just keep the current logic for these types and let it return null if it happens.

jmacd (approved):

rust/otap-dataflow/crates/pdata/src/otlp/attributes.rs
albertlockett: thanks @utpilla , made this change

rust/otap-dataflow/crates/pdata/src/otlp/attributes.rs
albertlockett: I think we should, but I found an issue in the boolean array builder that if there are all default/null values it actually doesn't omit the optional column. I created an issue for this and added comments to this code for where to come back and fix the code/test.

PR #1448, opened 2025-11-18
[otap-dataflow benchmark] filter processor benchmarks + internal telemetry

c1ly:
Add filter processor scenarios to the nightly benchmark suite
Collect internal metrics inside the filter processor tracking, number of signals before and after the filtering

```rust
/// Pdata-oriented metrics for the OTAP FilterProcessor
#[metric_set(name = "filter.processor.pdata.metrics")]
pub struct FilterPdataMetrics {
    /// Number of log signals consumed
    #[metric(unit = "{log}")]
    pub log_signals_consumed: Counter<u64>,
    /// Number of span signals consumed
    #[metric(unit = "{span}")]
    pub span_signals_consumed: Counter<u64>,

    /// Number of log signals sent
    #[metric(unit = "{log}")]
    pub log_signals_sent: Counter<u64>,
    /// Number of span signals sent
    #[metric(unit = "{span}")]
    pub span_signals_sent: Counter<u64>,
}

```


tools/pipeline_perf_test/test_suites/integration/continuous/100klrps-docker.yaml
clhain, line 1: imo we should probably just do these on the nightly runs.

clhain:
Good start! Ran this locally and saw 0 logs on the backend and assumed it broke - but I guess we must just be filtering 100% of signals. A few high level things:

- Probably we should have an explicit counter for items filtered. Doing 'consumed - sent' is maybe ok if there's no chance it ever fails to send for reasons other than 'filtered' (like 'errors'), but generally if I have a filter processor I want a metric that tells me how much it filtered so I don't have to math.
- We should have some sort of affirmation that this is working other than just 'no logs on the backend' - either we need to incorporate the metrics from the filter processor into the final report (probably best), or we could e.g. figure out how to make it filter something like half the traffic only so we can see data flowing through.

The metrics for whatever we generate should already be getting collected, so it'll just be a matter of rewriting the reporting sql to do whatever we decide on the above. I can help with that when you get there.

c1ly:
> Good start! Ran this locally and saw 0 logs on the backend and assumed it broke - but I guess we must just be filtering 100% of signals. A few high level things:
> 
> * Probably we should have an explicit counter for items filtered. Doing 'consumed - sent' is maybe ok if there's no chance it ever fails to send for reasons other than 'filtered' (like 'errors'), but generally if I have a filter processor I want a metric that tells me how much it filtered so I don't have to math.
> * We should have some sort of affirmation that this is working other than just 'no logs on the backend' - either we need to incorporate the metrics from the filter processor into the final report (probably best), or we could e.g. figure out how to make it filter something like half the traffic only so we can see data flowing through.
> 
> The metrics for whatever we generate should already be getting collected, so it'll just be a matter of rewriting the reporting sql to do whatever we decide on the above. I can help with that when you get there.

updated the metrics so it emits the number of filtered signals, what is the process of getting these metrics onto the final report

clhain:
Benchmark setup looks good to me now, good stuff. There are still issues with the wait_for_result in otlp->otap tests so I suggest we disable that setting in all of them for now unless others feel strongly about it.

tools/pipeline_perf_test/test_suites/integration/configs/integration_report_logs_filtered.yaml
```diff
@@ -0,0 +1,323 @@
+queries:
+  - name: Filter superflous metadata
+    sql: |
+      CREATE VIEW metadata_filtered AS
+      SELECT *
+      FROM metadata
+      WHERE Attribute NOT LIKE 'test.ctx.%';
+
+  - name: Calculate observation window for this test
```
jmacd, line 9: I'm not sure how to review this file. Should it have a README? I wonder if there's something re-usable in here, or a pattern I haven't spotted yet.

jmacd (approved):

lquerel:
Didn't observe any perf regression. See -> https://github.com/open-telemetry/otel-arrow/actions/runs/19656637468/job/56294519094

PR #1457, opened 2025-11-21
Add github workflow for automatic Tidy on Renovate PRs

### LINKED ISSUE #1456: Renovate Bot can't handle multi-step Golang tidy
As seen in many PRs like #1440, Renovate is not able to cleanly handle the multi-step build process we need in otel-arrow when updating Go modules.

Not only do we need to update in `go/` subdirectory, but oftentimes we need to execute Golang collector builder with `make genotelarrowcol` for dependency upgrades to reflect in `collector/cmd/otelarrowcol`.

While looking through the main Collector repo, I came across [this workflow used for tidying up after Renovate PRs](https://github.com/open-telemetry/opentelemetry-collector/blob/main/.github/workflows/tidy-dependencies.yml) using the `otelbot` identity (which we already use for Release processes, see #737).

We can try to copy this and hopefully solve our problems with painful Golang updates requiring manual touches.

For a look at this process in action, check out [opentelemetry-collector #14181](https://github.com/open-telemetry/opentelemetry-collector/pull/14181/commits).

drewrelmas:
Fixes #1456.

Adds a workflow to run `make genotelarrowcol` on top of any Renovate dependency upgrades.

If it works properly, I thought it would be best to remove the bunching together of all Golang upgrades together which currently exists in our Renovate config. It will make it easier to quickly merge non-breaking upgrades and isolating upgrades that require further manual intervention.

### DIFF
--- a/.github/workflows/tidy.yml
+++ b/.github/workflows/tidy.yml
@@ -38,4 +38,3 @@ jobs:
 
           echo "git diff --exit-code || (git add . && git commit -m \"make genotelarrowcol\" && git push)"
           git diff --exit-code || (git add . && git commit -m "make genotelarrowcol" && git push)
-      


.github/renovate.json5
```diff
@@ -47,7 +47,7 @@
       "groupName": "github workflow dependencies",
       "matchFileNames": [".github/workflows/*.yml", ".github/workflows/*.yaml"],
       "schedule": [
-        "before 8am on Monday"
+        "before 8am on the first day of the month"
```
drewrelmas, line 50: Mostly unrelated, just thought on a whim that we also don't need as many Github Workflow hash updates. We can use same configuration as image manifest upgrades.

.github/renovate.json5
```diff
@@ -59,18 +59,6 @@
...
-      // note that these changes often cannot be taken as-is
-      // need to run genotelarrowcol again manually
-      // to properly update collector/cmd/otelarrowcol
-      // grouping them together reduces pain of this manual work
-      "groupName": "golang module dependencies",
-      "matchFileNames": ["go/go.mod"],
-      "schedule": [
-        "before 8am on Monday"
-      ]
-    },
```
drewrelmas, line 73: In #624 I added this section to instruct Renovate to group together all Golang updates since they required manual intervention anyway to get proper `tidy` and `genotelarrowcol` results. If the new workflow works as intended, it is actually better for them to come independently again to easily merge those with non-breaking changes without being blocked by other upgrades that do have breaking changes.

jmacd: Nice! Now it makes sense. Thanks.

PR #1462, opened 2025-11-22
Batch processor refactoring, testing

jmacd:
This is such an extensive change that we may as well call it a rewrite, although the basic structure remains.

I have modified the Config slightly to match the original Collector batch processor, with three fields (batch_size, max_batch_size, timeout):

- Uses Option<NonZeroUsize> values instead of more-elaborate logic around 0 values disabling features.
- Only one size is required, it can be either the lower or upper bound. This component will only invoke the otap_df_pdata logic after it reaches the smaller of the two sizes. The max-size (if set) is passed to the lower-level utility for splitting.

I made the logic signal-generic; there is only one call to send_message().

I included a fairly-large hack for timing: as the signals are independent, and as the effect handler has no "one-shot" timer support, I used the DelayData signal with an empty request of the signal that set the timer. Previously the code was incomplete: the "TimerTick" message was being used to flush, but no timers were set. Now, we set the timer whenever a transition from 0-items happens; however we also reset this and have no way to cancel, so this is a major TODO.

Modifies the internal metrics to match the new code structure: primary focus is on producer/consumer items/batches so we can monitor performance and item-count correctness.

Completely replaces the tests following the new DataGenerator; however does not modify/extend/improve the test fixture support which remains incomplete following #1428.

Adds a demo config.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/batch_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/batch_processor.rs
@@ -80,7 +80,7 @@ const fn default_send_batch_max_size() -> Option<NonZeroUsize> {
     None
 }
 
-fn default_timeout_duration() -> Duration {
+const fn default_timeout_duration() -> Duration {
     Duration::from_millis(DEFAULT_TIMEOUT_MS)
 }
 


rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -5,294 +5,263 @@
...
+    /// Optionally limit batch sizes to an upper bound. Measured in
+    /// items, as described for send_batch_size.
     #[serde(default = "default_send_batch_max_size")]
-    pub send_batch_max_size: usize,
-    /// Flush non-empty batches on this interval.
-    #[serde(with = "humantime_serde", default = "default_timeout_duration_opt")]
+    pub send_batch_max_size: Option<NonZeroUsize>,
+
+    /// Flush non-empty batches on this interval, which may be 0 for
+    /// immediate flush or None for no timeout.
```
utpilla, line 69: Do we want to allow `timeout` to be `None`? We could run into a latency and staleness issues if we allow a BatchProcessor to only have size-based flushing?

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -5,294 +5,263 @@
...
+        config.validate()?;
 
-        if let Some(limit) = config.metadata_cardinality_limit {
-            if limit < MIN_METADATA_CARDINALITY_LIMIT {
-                config.metadata_cardinality_limit = Some(MIN_METADATA_CARDINALITY_LIMIT);
-            }
-        }
+        let lower_limit = config
+            .send_batch_max_size
+            .or(config.send_batch_size)
```
utpilla: The order should be switched to calculate the lower limit correctly.

```suggestion
            .config.send_batch_size
            .or(send_batch_max_size)
```

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -313,338 +282,184 @@ impl OtapBatchProcessor {
...
+            FlushReason::Size => self.metrics.flushes_size.inc(),
+            FlushReason::Timer => self.metrics.flushes_timer.inc(),
+            FlushReason::Shutdown => {}
+        }
+
+        let limit = self.config.send_batch_max_size;
+        let mut output_batches = match make_output_batches(signal, nzu_to_nz64(limit), input) {
+            Ok(v) => v,
+            Err(e) => {
+                self.metrics.batching_errors.inc();
```
utpilla: It's probably better to track the number of batches that we couldn't process.

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -313,338 +282,184 @@ impl OtapBatchProcessor {
...
-                                rebuffered = true;
-                            }
-                        }
-                    }
 
-                    for records in output_batches {
-                        let pdata = OtapPdata::new_todo_context(records.into());
-                        effect.send_message(pdata).await?;
-                    }
+        let input = std::mem::take(&mut buffer.pending);
```
utpilla: This would replace the `pending` buffer with an empty vec. It would be better to use `drain` here to retain the `pending` buffer's capacity.

```suggestion
        let input = buffer.pending.drain(..).collect::<Vec<_>>();
```

rust/otap-dataflow/crates/otap/src/batch_processor.rs
jmacd: https://github.com/open-telemetry/opentelemetry-collector/pull/13763

I interpret zero timeout as requesting immediate flush. It's a useful configuration in my experience when you are [splitting on behalf of a metrics SDK](https://github.com/lightstep/otel-launcher-go/blob/6356dc37469229c0a80e4b9831b0d8ca753b0fdb/lightstep/sdk/metric/exporters/otlp/otelcol/client.go#L73).

rust/otap-dataflow/crates/otap/src/batch_processor.rs
jmacd: I raised the counter by input.len() to record the number of batches. I considered extending metrics with consumed_batches_{signal}_failed, like the retry_processor does, but would held this idea back for a larger redesign of component-level metrics.

rust/otap-dataflow/crates/otap/src/batch_processor.rs
jmacd: Fixed.

rust/otap-dataflow/crates/otap/src/batch_processor.rs
jmacd: Now implemented/tested this way, too.

utpilla: I was referring to `timeout` being `None` and not zero.

Setting the timeout to `None` sounds like there wouldn't be any timeout set meaning batch processor would flush items only if the batch size exceeds a certain number. There won't be any time-based flush trigger.

> I interpret zero timeout as requesting immediate flush. It's a useful configuration in my experience when you are [splitting on behalf of a metrics SDK](https://github.com/lightstep/otel-launcher-go/blob/6356dc37469229c0a80e4b9831b0d8ca753b0fdb/lightstep/sdk/metric/exporters/otlp/otelcol/client.go#L73).

That's okay but then we should be using [`Duration::ZERO`](https://doc.rust-lang.org/std/time/struct.Duration.html#associatedconstant.ZERO) to represent that and not `None`.

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -5,294 +5,278 @@
...
+
+    /// Flush non-empty batches on this interval, which may be 0 for
+    /// immediate flush or None for no timeout.
+    #[serde(with = "humantime_serde", default = "default_timeout_duration")]
+    pub timeout: Duration,
 }
 
-fn default_send_batch_size_opt() -> Option<usize> {
-    Some(DEFAULT_SEND_BATCH_SIZE)
+fn default_send_batch_size() -> Option<NonZeroUsize> {
```
utpilla:
```suggestion
const fn default_send_batch_size() -> Option<NonZeroUsize> {
```

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -5,294 +5,278 @@
...
+    NonZeroUsize::new(DEFAULT_SEND_BATCH_SIZE)
 }
 
-fn default_send_batch_max_size() -> usize {
-    FOLLOW_SEND_BATCH_SIZE_SENTINEL // normalized in from_config
+fn default_send_batch_max_size() -> Option<NonZeroUsize> {
```
utpilla:
```suggestion
const fn default_send_batch_max_size() -> Option<NonZeroUsize> {
```

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -5,294 +5,278 @@
...
+    // No upper-bound
+    None
 }
 
-fn default_timeout_duration_opt() -> Option<Duration> {
-    Some(Duration::from_millis(DEFAULT_TIMEOUT_MS))
+fn default_timeout_duration() -> Duration {
```
utpilla:
```suggestion
const fn default_timeout_duration() -> Duration {
```

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -5,294 +5,278 @@
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+/// Per-signal state
+#[derive(Default)]
+struct BatchSignals {
```
utpilla, line 141: nit: Maybe `SignalBatches` is a better name.

rust/otap-dataflow/crates/otap/src/batch_processor.rs
jmacd: fixed.

rust/otap-dataflow/crates/otap/src/batch_processor.rs
jmacd: Done!

Issue #1441, opened 2025-11-17
[otap-df-pdata] OTLP view backend for OTAP records


jmacd:
We have experimental exporters now, [Geneva](https://github.com/open-telemetry/otel-arrow/pull/1355) merged and [LogAnalytics](https://github.com/open-telemetry/otel-arrow/pull/1429) pending. Both of these were written based on the OTLP data model and have existing code that works with Prost message objects in the opentelemetry-proto crate, a part of the OTel-Rust SDK. These exporters are serializing to OTLP bytes, parsing as OTLP message objects, then converting into another format. I think these components should instead consume a PData view of the OTAP records by forming the OTLP view without a copy.

This means being able to iterate over the distinct resources, then each distinct scope of each resource, and so on recursively.

See also this comment, on the same topic.
https://github.com/open-telemetry/otel-arrow/pull/1429#discussion_r2534693549

lalitb:
Just to update - I am looking into this as this is required to avoid double transformation in Geneva exporter.

PR #1467, opened 2025-11-24
  [otap-df-pdata]Add zero-copy view for OTAP logs format

lalitb:
Fixes: #1441 

Implements `OtapLogsView` - a zero-copy view over Arrow RecordBatch data for OTAP logs.

## Use Cases
This view is useful for  exporters that need to iterate over the OTAP log data:

- Geneva Exporter: Enables the exporter to iterate directly over Arrow data to generate custom payloads, avoiding the costly round-trip of converting to OTLP bytes and decoding them back to objects.
- Log Analytics: Allows efficient, zero-copy access to log fields for serialization to JSON.

- General Data Access: Provides a standard, efficient way for any component to read the hierarchical OTLP structure (Resource → Scope → Log) without needing to understand the underlying Arrow complexity.

Without views: OTAP → OTLP bytes → decode → iterate (slower) 
With views: OTAP → view → iterate (faster, zero-copy)

 ## Design

- Stores column references and accessors (fixed overhead)
- Pre-computes indices for resource/scope/log hierarchy
- Actual telemetry data remains in Arrow RecordBatch (no copying)

## Benchmark Results

  System: Apple M4 Pro, 12 cores, 24 GB RAM

  | Logs | Zero-Copy View | OTAP→OTLP→Process |
  |------|----------------|-------------------|
  |   10 |      1.48 µs   |      6.43 µs      |
  |  100 |      9.86 µs   |     40.40 µs      |
  | 1000 |    121.42 µs   |    383.11 µs      |

  OTAP→OTLP→Process = full cost without views (OTAP→OTLP conversion + OTLP decode + iterate)

  ## Changes

 - Added `OtapLogsView` with hierarchical iteration (Resource → Scope → LogRecord → Attributes)
 - Added column caching to avoid repeated schema lookups
 - Updated Geneva exporter:
    - Added example view-based code (commented out) to demonstrate intended usage
    - Currently uses OTAP→OTLP fallback conversion until geneva-uploader 0.4.0+ supports view API
    - All pipelines (OTLP bytes and OTAP Arrow) continue to work

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
@@ -503,13 +503,13 @@ mod tests {
         UInt16Array, UInt32Array,
     };
     use arrow::datatypes::{DataType, Field, Schema, TimeUnit};
-    use otap_df_pdata::otap::OtapArrowRecords;
-    use otap_df_pdata::proto::opentelemetry::arrow::v1::ArrowPayloadType;
     use std::sync::Arc;
 
-    // View imports for future zero-copy path tests
-    use otap_df_pdata::views::logs::{LogsDataView, ResourceLogsView, ScopeLogsView};
-    use otap_df_pdata::views::otap::OtapLogsView;
+    // TODO: Re-enable these imports when zero-copy view tests are uncommented
+    // use otap_df_pdata::otap::OtapArrowRecords;
+    // use otap_df_pdata::proto::opentelemetry::arrow::v1::ArrowPayloadType;
+    // use otap_df_pdata::views::logs::{LogsDataView, ResourceLogsView, ScopeLogsView};
+    // use otap_df_pdata::views::otap::OtapLogsView;
 
     // TODO: Re-enable when zero-copy view tests are uncommented
     /// Helper to create a simple OTAP logs RecordBatch for testing Geneva exporter


rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
```diff
@@ -38,6 +38,9 @@ use otap_df_engine::message::{Message, MessageChannel};
 use otap_df_engine::node::NodeId;
 use otap_df_engine::terminal_state::TerminalState;
 use otap_df_pdata::otlp::OtlpProtoBytes;
+// TODO: Uncomment when geneva-uploader supports pdata-views (after 0.3.0)
+// use otap_df_pdata::views::otap::OtapLogsView;
```
lalitb, line 42: The integration code using `OtapLogsView` is currently commented out because the geneva-uploader crate requires an update to support iteration over this new view type. I have included the commented-out code solely to demonstrate the intended usage pattern. This will be uncommented and enabled once a compatible version of geneva-uploader is published. Please feel free to ignore the commented block for this review.

jmacd (approved):
Excellent!

I have one thought, non-blocking feedback looking for additional ways to test this.

I was thinking that since the former releases of geneva_uploader were based on OTLP protocol message objects, that I'd see OTLP test message objects used as input and corresponding test expectations written for the Geneva output we expect.

Possibly, when the geneva_uploader component releases with view support, we could see a few tests with input OlpProtoMessages and output Geneva structs added, so that we're testing an end-to-end without hand-crafted OtapArrowRecords.

rust/otap-dataflow/crates/pdata/src/views/otap/logs.rs
```diff
@@ -0,0 +1,1993 @@
...
+    // Without these, we'd need to scan the entire logs batch for each resource/scope
+    // to find matching rows, and scan all attributes for each log record.
+    //
+    // Memory allocation: Stores IDs and row indices only, not actual data.
+    // Example: 1000 logs with 4 log attributes each allocates ~50 KB for these
+    // index structures (resource/scope attributes add negligibly since there are
+    // typically only 1-10 resources/scopes). The actual attribute strings (~200 KB)
+    // remain in the Arrow RecordBatch (zero-copy).
+    //
+    // TODO: Consider HashMap instead of BTreeMap for log_attrs_map when log count
```
lquerel, line 103: Maybe explain why you start with a BTreeMap instead of using a HashMap right away.

lalitb: Good suggestion! I've expanded the comment focusing on the memory vs speed trade-off and the need to profile before switching.

rust/otap-dataflow/crates/pdata/src/views/otap/logs.rs
```diff
@@ -0,0 +1,2004 @@
...
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.current_idx >= self.matching_rows.len() {
+            return None;
+        }
+
+        let attr_row_idx = self.matching_rows[self.current_idx];
+        self.current_idx += 1;
+
+        let log_attrs_cols = self.log_attrs?;
```
lquerel: Should we increment this index before knowing if `log_attrs` or `get_attribute_key` are None?

rust/otap-dataflow/crates/pdata/src/views/otap/logs.rs
```diff
@@ -0,0 +1,2004 @@
...
+}
+
+#[allow(dead_code)]
+fn get_schema_url_for_scope<'a>(_batch: &'a RecordBatch, _scope_id: u16) -> Option<Str<'a>> {
+    // TODO: Implement schema_url lookup for scope
+    // Would need to look up in a separate scope table or nested struct
+    None
+}
+
+#[cfg(test)]
```
lquerel, line 1204: Did you test scenarios with log batches that contain no resource attributes or fields, or no scope attributes/fields?

lquerel (commented):
Overall looks good, made few comments.

I’m wondering whether all cases have been tested, especially those where no attributes are present.

@albertlockett could take a look at this PR.

rust/otap-dataflow/crates/pdata/src/views/otap/logs.rs
lalitb: @lquerel  - Incrementing before these checks is safe here. When `log_attrs` is None, matching_rows is empty so we never reach the increment. For `get_attribute_key` returning None (malformed data), stopping iteration is the correct behavior.

I have now added test `test_missing_attributes_iterator` to validate this.

rust/otap-dataflow/crates/pdata/src/views/otap/logs.rs
lalitb: I have now added `test_missing_attributes_iterator`  to cover this. It tests log batches with no resource attributes, no scope attributes, and no log attributes (all passed as `None`). The logs batch contains resource and scope ID fields per the OTAP schema, but has zero attributes.

lalitb:
> Possibly, when the geneva_uploader component releases with view support, we could see a few tests with input OlpProtoMessages and output Geneva structs added, so that we're testing an end-to-end without hand-crafted OtapArrowRecords.

Good suggestion!  I'll look into adding OTLP → Geneva struct tests that validate the encoding without needing backend connectivity. That would give us better test coverage of the full conversion path.

rust/otap-dataflow/crates/pdata/src/views/otap/logs.rs
albertlockett: Thanks for adding the test @lalitb 

I think we should update the test to exclude those ID columns. These columns are optional and if there are no attributes, we don't don't include them in the Logs `RecordBatch`.

rust/otap-dataflow/crates/pdata/src/views/otap/logs.rs
```diff
@@ -0,0 +1,2050 @@
...
+    observed_time_unix_nano: Option<&'a TimestampNanosecondArray>,
+    severity_number: Option<Int32ArrayAccessor<'a>>,
+    severity_text: Option<StringArrayAccessor<'a>>,
+    body_columns: Option<OtapBodyColumns<'a>>,
+    dropped_attributes_count: Option<&'a UInt32Array>,
+    flags: Option<&'a UInt32Array>,
+    trace_id: Option<&'a FixedSizeBinaryArray>,
+    span_id: Option<&'a FixedSizeBinaryArray>,
+    event_name: Option<StringArrayAccessor<'a>>,
+}
```
albertlockett: We have a similar struct that already exists here. Maybe we should use that to avoid duplicating the code?
https://github.com/open-telemetry/otel-arrow/blob/2d494069407b06e36238ea0fb6c7812924859cd0/rust/otap-dataflow/crates/pdata/src/otlp/logs.rs#L35-L48

albertlockett: Similarly, we have these structs to represent logs body and attributes which we might be able to use in place of `OtapAttributeColumns` `OtapBodyColumns`?

https://github.com/open-telemetry/otel-arrow/blob/2d494069407b06e36238ea0fb6c7812924859cd0/rust/otap-dataflow/crates/pdata/src/otlp/logs.rs#L119-L122

https://github.com/open-telemetry/otel-arrow/blob/2d494069407b06e36238ea0fb6c7812924859cd0/rust/otap-dataflow/crates/pdata/src/otlp/attributes.rs#L44-L51

rust/otap-dataflow/crates/pdata/src/views/otap/logs.rs
```diff
@@ -0,0 +1,2050 @@
...
+
+        // Extract value with type information
+        let value = get_attribute_value(log_attrs_cols, attr_row_idx);
+
+        Some(OtapAttributeView { key, value })
+    }
+}
+
+/// Iterator over attributes for a resource
+pub struct OtapResourceAttributeIter<'a> {
```
albertlockett: It seems like `OtapResourceAttributeIter`, `OtapScopeAttributeIter` and `OtapAttributeIter` all have a similar implementation. Is there any reason we couldn't combine these into one type?

albertlockett:
overall looks really good @lalitb! Thanks a lot for taking this on.

I had a few minor comments about where we could reuse some code to avoid duplication. I'm happy to approve as-is if these suggestions aren't sensible so let me know what you think :)

rust/otap-dataflow/crates/pdata/src/views/otap/logs.rs
```diff
@@ -0,0 +1,2050 @@
...
+        OtapScopeAttributeIter {
+            scope_attrs: self.view.scope_attrs.as_ref(),
+            matching_rows,
+            current_idx: 0,
+        }
+    }
+
+    #[inline]
+    fn dropped_attributes_count(&self) -> u32 {
+        0 // TODO: implement if stored in OTAP
```
albertlockett, line 629: I created an issue to track the methods we don't currently implement: https://github.com/open-telemetry/otel-arrow/issues/1475

lalitb:
> I had a few minor comments about where we could reuse some code to avoid duplication. I'm happy to approve as-is if these suggestions aren't sensible so let me know what you think :)

  @albertlockett - All valid points, thanks! I've addressed the code duplication you raised. Additionally, I've improved the attribute iterator handling, added dictionary encoding support for ID columns, and expanded test coverage.

 I would be doing similar implementation for traces once this is merged.

lquerel:
@albertlockett are you ok to merge?

albertlockett (approved):
approved! Thanks again @lalitb for the contribution. @lquerel good to merge it

PR #1474, opened 2025-11-25
OTLP Exporter Optimizations

lquerel:
Support for multiple simultaneous client connections to improve throughput and the ack/nack system.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
+++ b/rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
@@ -398,7 +398,7 @@ fn prepare_otap_export<Enc: ProtoBytesEncoder>(
     encoder: &mut Enc,
     exporter: &NodeId,
     signal_type: SignalType,
-) -> Result<EncodedExport, EncodingFailure> {
+) -> Result<EncodedExport, Box<EncodingFailure>> {
     proto_buffer.clear();
     if let Err(e) = encoder.encode(&mut otap_batch, proto_buffer) {
         let error = Error::ExporterError {
@@ -413,11 +413,11 @@ fn prepare_otap_export<Enc: ProtoBytesEncoder>(
         }
         let saved_payload: OtapPayload = otap_batch.into();
 
-        return Err(EncodingFailure {
+        return Err(Box::new(EncodingFailure {
             error,
             context,
             saved_payload,
-        });
+        }));
     }
 
     // Maintain the buffer's capacity across repeated calls.
@@ -495,18 +495,18 @@ async fn dispatch_otap_export<Enc, Fut, MakeFuture>(
 }
 
 async fn notify_prepare_error(
-    error: EncodingFailure,
+    error: Box<EncodingFailure>,
     effect_handler: &EffectHandler<OtapPdata>,
 ) -> Result<(), Error> {
     let EncodingFailure {
         error,
         context,
         saved_payload,
-    } = error;
+    } = *error;
 
     effect_handler
         .notify_nack(NackMsg::new(
-            &error.to_string(),
+            error.to_string(),
             OtapPdata::new(context, saved_payload),
         ))
         .await?;
--- a/rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
@@ -18,7 +18,7 @@ pub struct GrpcClientSettings {
     pub grpc_endpoint: String,
 
     /// Compression method to use for outbound requests. Defaults to no compression.
-    #[serde(default)]
+    #[serde(default, alias = "compression_method")]
     pub compression: Option<CompressionMethod>,
 
     /// Maximum number of concurrent in-flight requests allowed by the transport stack.


rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
```diff
@@ -114,52 +129,103 @@ impl Exporter<OtapPdata> for OTLPExporter {
...
+            let msg = if let Some(msg) = pending_msg.take() {
+                msg
+            } else if inflight_exports.is_empty() {
+                msg_chan.recv().await?
+            } else {
+                let completion_fut = inflight_exports.next_completion().fuse();
+                let recv_fut = msg_chan.recv().fuse();
+                futures::pin_mut!(completion_fut, recv_fut);
+
+                futures::select_biased! {
```
jmacd, line 198: I am having trouble reading this, other than to make this select_biased magic work you need the other three lines of magic above. I'm trying to decipher how to read "fuse" in relation to the two futures.

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
```diff
@@ -169,105 +235,101 @@ impl Exporter<OtapPdata> for OTLPExporter {
                     _ = metrics_reporter.report(&mut self.pdata_metrics);
                 }
                 Message::PData(pdata) => {
-                    // Capture signal type before moving pdata into try_from
-                    let signal_type = pdata.signal_type();
+                    if inflight_exports.len() >= max_in_flight {
+                        pending_msg = Some(Message::PData(pdata));
```
jmacd, line 242: Can we `debug_assert!(pending_msg.is_none())`?
I see it's from the `take()` above.

The control flow is somehow hard to follow. After the continue below we will enter a loop and the first if-statement is guaranteed true, which does a thing and then continues the loop. It leaves me wondering if the block of code at the top of the loop (`if let Some(completed) = inflight_exports.next_completion().await {`) couldn't just execute here, with one less `continue` statement maybe.

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
```diff
@@ -312,73 +374,351 @@ async fn handle_export_result<T>(
...
+
+        return Err(EncodingFailure {
+            error,
+            context,
+            saved_payload,
+        });
+    }
+
+    let (bytes, next_capacity) = proto_buffer.take_into_bytes();
+    proto_buffer.ensure_capacity(next_capacity);
```
jmacd, line 425:
```suggestion
    // Maintain the buffer's capacity across repeated calls.
    let (bytes, next_capacity) = proto_buffer.take_into_bytes();
    proto_buffer.ensure_capacity(next_capacity);
```

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
```diff
@@ -312,73 +374,351 @@ async fn handle_export_result<T>(
...
+        }
+
+        let traces_cap = self.traces.capacity();
+        for _ in 0..traces_cap {
+            self.traces.push(self.make_traces_client());
+        }
+    }
+
+    fn take_logs(&mut self) -> LogsServiceClient<Channel> {
+        self.logs.pop().unwrap_or_else(|| self.make_logs_client())
```
jmacd: What will cause the `unwrap_or_else()` case to happen? I'm wondering if it has to do with failing to release, or if this may (intentionally) allow the pool to become over-capacity.

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
lquerel: I could use a `select!` with the `biased` statement to simplify and avoid the fuse. The only reason I did that what to be a bit more independent of Tokio (`select!` is a tokio thing). In terms of performance it's probably equivalent.

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
lquerel: I agree, I need to cleanup a bit the code here and better document.

jmacd (approved):

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
lquerel: I can safely replace these `unwrap_or_else` calls with `expect("...")`. The exporter guarantees that `inflight.len()` stays below `max_in_flight` before pushing a new export. After each completion the client is returned through `release`, so the steady state pool size always remains at or below `max_in_flight` per signal. The `unwrap_or_else` was acting as a safety net in case the pool wasn't prepopulated or had been drained, but with the current flow it never grows unbounded. I'm switching to `expect` with a clear message: if it ever triggers, it means there is a design issue, which I don’t believe is the case.

lquerel: Done

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
lquerel: Added some clarifications

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
```diff
@@ -312,73 +377,358 @@ async fn handle_export_result<T>(
...
-    T2::ResponseBody: Send + 'static,
-    <T2::ResponseBody as tonic::transport::Body>::Error:
-        Into<Box<dyn std::error::Error + Send + Sync>> + Send,
-    S: crate::otap_grpc::otlp::client::ServiceDescriptor,
-    Resp: prost::Message + Default + Send + 'static,
-{
-    // Encode OTAP records to protobuf
+    exporter: &NodeId,
+    signal_type: SignalType,
+) -> Result<EncodedExport, Box<EncodingFailure>> {
```
utpilla, line 401: We could remove the `Box` here.

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
```diff
@@ -114,52 +129,106 @@ impl Exporter<OtapPdata> for OTLPExporter {
...
+        grpc_clients.prepopulate_clients();
 
+        let mut inflight_exports = InFlightExports::new();
+        let mut pending_msg: Option<Message<OtapPdata>> = None;
+
+        // Main loop: 1) finish ready completions, 2) biased wait for either a completion
+        // or the next message, 3) dispatch work while respecting the in-flight budget.
         loop {
-            match msg_chan.recv().await? {
+            // Backpressure guard: when full and a message is parked, only drain completions.
```
utpilla, line 165: How about we place the code for finishing ready completions at the start of the loop?

```rust
// Opportunistically drain completions before we park on a recv.
while let Some(completed) = inflight_exports.next_completion().now_or_never().flatten()
{
    ...
}

// Backpressure guard: when full and a message is parked, only drain completions.
if inflight_exports.len() >= max_in_flight && pending_msg.is_some() {
    ...
}

// Prefer completions if any are ready, otherwise biased select between completion and recv.
...
```

I think starting the loop with an opportunistic attempt at draining already ready completions will lead to more efficient polling of the futures. This way we are maximizing our chances of not having to wait on `inflight_exports.next_completion().await` when running on capacity with a pending message.

rust/otap-dataflow/crates/otap/src/otlp_exporter.rs
```diff
@@ -40,14 +53,16 @@ pub const OTLP_EXPORTER_URN: &str = "urn:otel:otlp:exporter";
...
-    /// Timeout for RPC requests. If not specified, no timeout is applied.
-    /// Format: humantime format (e.g., "30s", "5m", "1h", "500ms")
-    #[serde(default, with = "humantime_serde")]
-    pub timeout: Option<Duration>,
+    /// Shared gRPC client settings reused across OTLP exports.
+    #[serde(flatten)]
+    pub grpc: GrpcClientSettings,
+    /// Maximum number of concurrent in-flight export RPCs.
+    #[serde(default = "default_max_in_flight")]
+    pub max_in_flight: usize,
```
utpilla, line 61: Maybe we should the name of the field or the documentation reflect that this is a per-core setting?

PR #1476, opened 2025-11-25
[query-engine] Simply lifetimes on RecordSetEngine ExecutionContext struct

CodeBlanch:
## Changes

* Removes `'c` lifetime from `ExecutionContext` struct

## Details

`ExecutionContext` has `'a` lifetime for tracking the `PipelineExpression` and `'c` lifetime for tracking diagnostics associated with expressions in that pipeline. Really `'a` and `'c` are the same lifetime so what this PR does is just remove `'c` in favor of `'a`. Net gain is just simpler code.

### DIFF
--- a/rust/experimental/query_engine/expressions/src/primitives/array_value.rs
+++ b/rust/experimental/query_engine/expressions/src/primitives/array_value.rs
@@ -37,22 +37,6 @@ pub trait ArrayValue: Debug {
     }
 }
 
-impl AsStaticValue for dyn ArrayValue + 'static {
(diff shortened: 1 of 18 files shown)


rust/experimental/query_engine/expressions/src/primitives/array_value.rs
```diff
@@ -37,22 +37,6 @@ pub trait ArrayValue: Debug {
     }
 }
 
-impl AsStaticValue for dyn ArrayValue + 'static {
```
CodeBlanch, line 40: Unrelated change. Just removing some uncalled code that snuck in.

PR #1477, opened 2025-11-25
Columnar query engine filter processor initial implementation

### LINKED ISSUE #1410: Custom `ExecutionPlan` impl for filtering by attributes
In #1342 , filtering by attributes are supported, including statements like:
```kql
logs | where attributes["x"] == "y"
logs | where attributes["x"] == "y" and attributes["x2"] == ["y2"]
logs | where attributes["x"] == "y" or resource.attributes["x2"] != ["y2"]
// etc.
```

However, in OTAP attributes are stored in a separate record batch which must be _joined_. That means the three plans above are actually implemented like:
```rs
// logs | where attributes["x"] == "y"
HashJoinExec: mode=CollectLeft, join_type=LeftSemi, on=[(id@0, parent_id@0)]
  DataSourceExec: partitions=1, partition_sizes=[1] // Logs
  FilterExec: key@1 = x AND str@2 = Y, projection=[parent_id@0]
    DataSourceExec: partitions=1, partition_sizes=[1] // LogAttrs

// logs | where attributes["x"] == "y" and attributes["x2"] == ["y2"]
HashJoinExec: mode=CollectLeft, join_type=LeftSemi, on=[(id@0, parent_id@0)]
  DataSourceExec: partitions=1, partition_sizes=[1] // Logs
    HashJoinExec: mode=CollectLeft, join_type=LeftSemi, on=[(id@0, parent_id@0)], projection=[_row_number@1]
      HashJoinExec: mode=CollectLeft, join_type=LeftSemi, on=[(id@0, parent_id@0)]
          DataSourceExec: partitions=1, partition_sizes=[1]
          FilterExec: key@1 = x AND str@2 = Y, projection=[parent_id@0]
            DataSourceExec: partitions=1, partition_sizes=[1] // LogAttrs
    FilterExec: key@1 = x2 AND str@2 = Y2, projection=[parent_id@0]
      DataSourceExec: partitions=1, partition_sizes=[1] // LogAttrs

// logs | where attributes["x"] == "y" or attributes["x2"] != ["y2"]
HashJoinExec: mode=CollectLeft, join_type=LeftSemi, on=[(id@8, parent_id@0)]
  DataSourceExec: partitions=1, partition_sizes=[1] // Logs
  AggregateExec: mode=Partial, gby=[_row_number@0 as _row_number], aggr=[first_value(_row_number)]
    UnionExec:
      FilterExec: key@1 = x AND str@2 = Y, projection=[parent_id@0]
        DataSourceExec: partitions=1, partition_sizes=[1] // LogAttrs
      FilterExec: key@1 = x2 AND str@2 != Y2, projection=[parent_id@0]
        DataSourceExec: partitions=1, partition_sizes=[1] // ResourceAttrs
```

In all of these cases, which are somewhat simple filters, the plans are quite complicated. Furthermore, although datafusion's optimizer is quite good, it's not a panacea and doesn't really do much to optimize these plans.

There's an opportunity here to simplify things, and get better performance.

Once we've implemented https://github.com/open-telemetry/otel-arrow/issues/1409 we'll have the opportunity do build a custom filtering `ExecutionPlan` implementation that has access to both the root record batch (Logs) and the various child batches (`LogAttrs`, `ResourceAttrs`). It would be better to use this do create a filter that:
- filters the attributes according to the predicate
- creates a unique list of parent IDs from the filtered child record batch
- uses this list to create a mask for selected rows from the parent batch
- filters the parent batch
This is actually similar to what we do in the current Filter Processor implementation that was added in https://github.com/open-telemetry/otel-arrow/pull/1341

I added a partial implementation of this here https://github.com/albertlockett/otel-arrow/blob/840293f189d2226a0e64d1662d4aeb5b24f083be/rust/experimental/query_engine/engine-columnar/src/attributes/filter.rs#L250-L310 and was able to get ~30-40% improvement on query performance filtering by a single attribute compared with using `HashJoinExec`

albertlockett:
closes #1410 

Adds implementation of `PipelineStage` that performs filtering called `FilterPipelineStage`.

The implementation is a hybrid approach between using DataFusion and apache arrow compute functions. Each binary expression in the filter predicate is implemented as a type called `FilterExec`, which implements filtering by:
- applies to the root `RecordBatch` DataFusion's [`PhysicalExpr`](https://docs.rs/datafusion-physical-expr/latest/datafusion_physical_expr/trait.PhysicalExpr.html) which returns a selection vector (a `BooleanArray` which has value `true` for rows that pass the filter).
- invoking `AttributeFilterExec` which, applies `PhysicalExpr` to an attributes record batch, and returns a bitmap of which parent_ids are selected that I've called an "id mask" in the code. This is then combined with selection vector for the root using arrow compute functions.

Boolean combinations of filter expressions (`and`, `or` and `not`) are planned and executed by a generic enum called `Composite<T>` which has capability to combine the selection vectors and id_masks for each `FilterExec` / `AttributeFilterExec` contained within it.

This has some advantages over what was described in https://github.com/open-telemetry/otel-arrow/issues/1410 and/or just using a regular DataFusion `ExecutionPlan` to do the filtering:
- It's easier to plan, especially because the attribute filtering is not driven by Joins and Distinct Unions
- Each step in the filtering not need to materialize the entire arrow record batch -- only the selection vectors & id masks are required which can be easily combined using arrow compute functions.

A basic `PipelinePlanner` is implemented that is able to produce `FilterPipelineStage` from the AST in `data_engine_expressions::PipelineExpression`.

Many types of filters are already supported:
```kql
// simple filtering by fields
logs | where severity_text == "ERROR"

// simple filtering by attributes
logs | where attributes["x"] == "y"

// comparing numbers
logs | where attributes["z"] > 10 // also supports ">=", "<" & "<=" numeric comparisons

// filtering by resource & scope
logs | where instrumentation_scope.name == "otel-sdk"
logs | where resource.schema_url = "opentelemetry.io/schema/v1"
logs | where instrumentation_scope.attributes["x"] == "y"
logs | where resource.attributes["x"] == "y"

// putting the literal on the left
logs | where "ERROR" == severity_text

// logical combinations of any of the above using "and", "or" and "not"
logs | where not(severity_text == "ERROR" or severity_text == "INFO") and event_name == "x" // etc.
```

There are still many TODOs in the code which I will make issues to resolve. Notably, what's not supported/not working:
- Empty batches (missing root `RecordBatch`)
- Missing attribute `RecordBatch`s
- Updating the plan when schemas changing
- Comparing two fields (`logs | where severity_text == event_name` or `logs | where attributes["x"] == attributes["z"]`)
- Comparing two literals (`logs | where 1 == 2`)
- Filtering by integers that aren't type i64
- Comparing text using "contains" or regex matches
- Metrics & Traces are not supported
- Any kind of arithmetic or function calls in the filter expressions

However before stepping into the list above the next direction I'd like to take this is to:
- Add benchmarks
- Work on optimizing the query plan. Notably, I think there's places we could combine the logical expressions in `FilterPlan` and `AttributesFilterPlan` into a single `PhysicalExpr` which might give us better query performance.

I realize this PR is quite long (although a lot of the volume it is tests). I'm happy to break it up if anyone reviewing would prefer.

### DIFF
--- a/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
+++ b/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
@@ -844,7 +844,7 @@ mod test {
         LogRecord, LogsData, ResourceLogs, ScopeLogs,
     };
     use otap_df_pdata::proto::opentelemetry::resource::v1::Resource;
-    use otap_df_pdata::testing::round_trip::encode_otlp;
+    use otap_df_pdata::testing::round_trip::otlp_to_otap;
     use otap_df_pdata::{OtapPayload, OtlpProtoBytes};
     use prost::Message;
 
@@ -858,7 +858,7 @@ mod test {
     }
 
     pub async fn exec_logs_pipeline(kql_expr: &str, logs_data: LogsData) -> LogsData {
-        let otap_batch = encode_otlp(&OtlpProtoMessage::Logs(logs_data));
+        let otap_batch = otlp_to_otap(&OtlpProtoMessage::Logs(logs_data));
         let pipeline_expr = KqlParser::parse(kql_expr).unwrap();
         let mut pipeline = Pipeline::new(pipeline_expr);
         let result = pipeline.execute(otap_batch.clone()).await.unwrap();


albertlockett:
converting to draft for now until I get CI fixed

rust/experimental/query_engine/engine-columnar/src/consts.rs
```diff
@@ -0,0 +1,6 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+pub(crate) const ATTRIBUTES_FIELD_NAME: &str = "attributes";
+pub(crate) const RESOURCES_FIELD_NAME: &str = "resource";
+pub(crate) const SCOPE_FIELD_NAME: &str = "instrumentation_scope";
```
CodeBlanch, line 6: We should probably discuss a bit about naming. Just FYI in otlp bridge/recordset existing stuff we went with "Attributes", "Resource", and "Scope" \ "InstrumentationScope" (both are allowed). Reason being, that is what they are called on the OTel spec. Tried to not leak protobuf\implementation details into the queries, basically.

albertlockett: Oh that's interesting. Yeah it would be good to get alignment on this. I was going with the convention that's used in OTTL where everything is lowercase and snake cased.

https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/filterprocessor
https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor

lquerel: OpenTelemetry uses snake_case consistently across the ecosystem, whether in YAML configuration files or OTTL. I think we should stick to this convention to minimize friction with the rest of the community.

lquerel: OpenTelemetry uses snake_case consistently across the ecosystem, whether in YAML collector configuration files, semantic convention files, or OTTL. I think we should stick to this convention to minimize friction with the rest of the community.

rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
```diff
@@ -0,0 +1,1751 @@
...
+        session_ctx: &SessionContext,
+        otap_batch: &OtapArrowRecords,
+    ) -> Result<Self::ExecutablePlan> {
+        let root_rb = match otap_batch.root_record_batch() {
+            Some(rb) => rb,
+            None => {
+                // TODO eventually we need to handle this case. We should possibly handle this
+                // somewhere up the call-stack. For example in Pipeline, which lazily plans the
+                // execution when it receives the first batch, it should not do it's planning until
+                // it receives a non-empty batch.
```
jmacd, line 315: (Batch processor will quietly acknowledge empty requests, for the record, they are empty containers and this is considered correct.)

rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
```diff
@@ -0,0 +1,1751 @@
...
+                        signal_type
+                    ),
+                });
+            }
+        };
+
+        Ok(otap_batch)
+    }
+}
+
```
jmacd, line 828: Really neat, this code!

PR #1478, opened 2025-11-25
[query-engine] Expand expressions to support user-defined functions

CodeBlanch:
Relates to #1479

## Changes

* Make it possible to declare and invoke user-defined functions in query expression tree

## Details

Implementation and KQL parsing will come as follow-ups.

### DIFF
--- a/rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
+++ b/rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
@@ -1532,9 +1532,26 @@ impl InvokeFunctionScalarExpression {
             ));
         }
 
-        // If function is implemented by expressions, and it has a single return
-        // statement, and that return statement is static, then we can elide the
-        // invocation of the function completely
+        /*
+        Note:
+
+        If a function...
+          * Has an implementation supplied by expressions
+          * Has only immutable arguments
+          * Has a single return statement
+          * Has a static return value
+        ...then we can elide the invocation of the function completely.
+
+        The immutable argument rule is there to make sure side-effects are preserved. Consider a function like...
+
+        my_func(m: Mutable(Map)) {
+            m['some_attr'] = 1;
+            return 1;
+        }
+
+        ...in that case the return is a constant (Long(1)) but we can't elide the function because the map
+        argument "m" still needs to be mutated.
+        */
         if let PipelineFunctionImplementation::Expressions(expressions) =
             function.get_implementation()
             && !function.get_parameters().iter().any(|v| {


rust/experimental/query_engine/expressions/src/pipeline_expression.rs
```diff
@@ -267,6 +317,210 @@ impl AsRef<PipelineExpression> for PipelineExpressionBuilder {
...
+        }
+    }
+
+    pub fn new_external(
+        name: &str,
+        parameters: Vec<PipelineFunctionParameter>,
+        return_value_type: Option<ValueType>,
+    ) -> PipelineFunction {
+        Self {
+            query_location: QueryLocation::new(0, 0, 1, 1).unwrap(),
```
drewrelmas, line 413: Should this not be the location the external func is mentioned in the query?

rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
```diff
@@ -3144,4 +3383,109 @@ mod tests {
...
+                    ScalarExpression::Static(StaticScalarExpression::String(
+                        StringScalarExpression::new(QueryLocation::new_fake(), "hello world"),
+                    )),
+                )],
+            ),
+            vec![],
+            Some("hello world".into()),
+        );
+
+        // Note: If a mutable parameter is present return value won't be folded
```
drewrelmas, line 3464: Why is this exactly? The expressions used by the function don't depend on param `a` - it only `Return`s a static value?

rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
```diff
@@ -1393,6 +1416,222 @@ impl Expression for SelectScalarExpression {
...
+        // invocation of the function completely
+        if let PipelineFunctionImplementation::Expressions(expressions) =
+            function.get_implementation()
+            && !function.get_parameters().iter().any(|v| {
+                matches!(
+                    v.get_parameter_type(),
+                    PipelineFunctionParameterType::MutableValue(_)
+                )
+            })
+        {
```
drewrelmas: Related to comment below - mention in comments why this conditional block depends on no parameter being a `MutableValue`.

CodeBlanch: Pushed an update to explain the mutable argument rule.

rust/experimental/query_engine/expressions/src/pipeline_expression.rs
CodeBlanch: This is the definition of an external function. It doesn't exist in the query so there isn't really a decent `QueryLocation` for it. I just defaulted to the "first character" of query. Where it is invoked that will be `ScalarExpression::InvokeFunction` and that will be tied to the real query location where invocation occurs.

drewrelmas: Ah ok, I misunderstood, all good.

PR #1480, opened 2025-11-26
OTLP receiver improvements/optimizations

lquerel:
Key changes:
- Use the new GrpcServerSettings config
- Tune gRPC Tonic settings
- New AckRegistry
- New global concurrency limiter
- Reduce copy/allocation on the hot path
- Clean up the control loop

Notes: 
- In order to keep the OTAP Receiver unchanged, I had to duplicate some data structures (e.g. Settings vs NewSettings). These duplications will be removed in the next PR.
- Perf improvements are not visible with the current continuous benchmark because the level of pressure is too low. However on high-load, the number of memory allocations is significantly smaller leading to better performance.

Manual pipeline perf test results: https://github.com/open-telemetry/otel-arrow/actions/runs/19776412483/job/56669638225
No regression observed.


rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server_new.rs
```diff
@@ -0,0 +1,630 @@
...
+/// Common settings for OTLP receivers.
+/// Per-signal server settings derived from user configuration and shared with the services.
+#[derive(Clone, Debug)]
+pub struct OtlpServerSettings {
+    /// Maximum concurrent requests per receiver instance (per core).
+    pub max_concurrent_requests: usize,
+    /// Whether the receiver should wait.
+    pub wait_for_result: bool,
+    /// Maximum size for inbound gRPC messages.
+    pub max_decoding_message_size: Option<usize>,
```
jmacd, line 106:
```suggestion
    // Note the Collector calls this max_recv_msg_size_mib
    pub max_decoding_message_size: Option<usize>,
```
I think we might consider ways to support human-readable sizes, also (not to block this PR!) to rename max_receive_message_size, maybe. https://github.com/open-telemetry/opentelemetry-collector/blob/152042ebfa9d67731b23ae3cb5b23f585e13d2a2/config/configgrpc/configgrpc.go#L183

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server_new.rs
```diff
@@ -0,0 +1,630 @@
...
+                        partial_success: None,
+                    }
+                    .encoded_len(),
+                );
+                ExportLogsServiceResponse {
+                    partial_success: None,
+                }
+                .encode(&mut buf)
+                .expect("encode logs response");
+                Bytes::from(buf)
```
jmacd, line 133: I'll suggest applying this suggestion in a future PR: https://github.com/lquerel/otel-arrow/pull/5

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server_new.rs
```diff
@@ -0,0 +1,630 @@
...
+        let mut guard = self.metrics.lock();
+        guard.request_bytes.add(len as u64);
+        let result = match self.signal {
+            SignalType::Logs => OtlpProtoBytes::ExportLogsRequest(bytes),
+            SignalType::Metrics => OtlpProtoBytes::ExportMetricsRequest(bytes),
+            SignalType::Traces => OtlpProtoBytes::ExportTracesRequest(bytes),
+        };
+        let context = if self.preallocate_frame {
+            // Pre-reserve a single frame since wait_for_result uses one slot.
+            Context::with_capacity(1)
```
jmacd, line 287: I wonder if Context should use a smallvec of frames to avoid this allocation?

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server_new.rs
```diff
@@ -0,0 +1,630 @@
...
+        } else {
+            RouteResponse::Expired
+        }
+    }
+}
+
+/// Common settings for OTLP receivers.
+/// Per-signal server settings derived from user configuration and shared with the services.
+#[derive(Clone, Debug)]
+pub struct OtlpServerSettings {
```
jmacd, line 100: I'm confused about this compared with the struct defined in server_settings.rs?

PR #1486, opened 2025-11-26
Batch processor Ack/Nack support

### LINKED ISSUE #1326: OTAP batch processor Ack/Nack handling
This can be added as an option via `wait_for_result`.

When waiting for the result is requested, the batch processor will retain enough state to Ack/Nack all the inputs.

This was previously explored in https://github.com/open-telemetry/otel-arrow/blob/main/collector/processor/concurrentbatchprocessor/README.md.

The batch processor will need a mechanism for this correlation as it will have to maintain the output of multiple concurrent batch-exports independently awaiting Ack/Nack. As shown in the concurrent batch processor code base (now deleted), this can be achieved by counting how many items are pending a response joining errors as needed. In the CBP, we awaited a complete response, meaning every item had an outcome, although we could imagine other policy configurations with regards to failure scenarios and partial responses.

jmacd:
Adds routing state for inbound and outbound batches, calls notify_ack and notify_nack appropriately. This depends on behavior referred to as "in-line" delivery, which requires that points are not re-ordered by the batcher; this property is required by the present algorithm to align Ack and Nack responses appropriately. This is tested lightly, now, in this PR. This requires more substantive testing in the lower-level library! This only tests logs and traces, because metric batching has known current defects.

The batch processor automatically tracks inbound and outbound context only as needed, considering the whether the arriving data `has_subscribers()`, which will be determined by `wait_for_result: true` in the receiver.

New testing revealed a few more cases where protocol buffer form and OTAP-records-converted forms have insignificant differences: TraceID, SpanID, Resource, and scope presence information is lost, default values are filled in. The equivalence tests now canonicalize these.

Fixes #1326

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/batch_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/batch_processor.rs
@@ -782,6 +782,7 @@ impl SignalBuffer {
         output_batches: &mut Vec<OtapArrowRecords>,
         last_items: usize,
     ) {
+        // SAFETY: protected by output_batches.len() > 1.
         let remaining = output_batches.pop().expect("has last");
         let last_input = from_inputs.context.last().expect("has last");
         let new_part = BatchPortion::new(last_input.inkey, last_items);


rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -542,7 +683,154 @@ impl local::Processor<OtapPdata> for BatchProcessor {
...
+    fn take_pending(&mut self) -> Vec<OtapArrowRecords> {
+        std::mem::take(&mut self.pending)
+    }
+
+    fn take_context(&mut self) -> MultiContext {
+        MultiContext::new(std::mem::take(&mut self.context))
+    }
+}
+
 impl SignalBuffer {
```
albertlockett, line 765: nit: it might be nice to have some brief comments about what the new methods we've implemented are doing (`take_remaining`, `drain_context` and `handle_partial_responses`).

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -542,7 +683,154 @@ impl local::Processor<OtapPdata> for BatchProcessor {
...
+        }
+    }
+
+    fn take_remaining(
+        &mut self,
+        from_inputs: &mut Inputs,
+        output_batches: &mut Vec<OtapArrowRecords>,
+        last_items: usize,
+    ) {
+        let remaining = output_batches.pop().expect("has last");
```
albertlockett, line 786: Is it always to expect here? We'll always call this wiht a non-empty `output_batches`?

albertlockett:
@jmacd does this close https://github.com/open-telemetry/otel-arrow/issues/1326 ?

rust/otap-dataflow/crates/otap/src/batch_processor.rs
jmacd: // SAFETY: protected by output_batches.len() > 1.

Issue #1484, opened 2025-11-26
`LogFilter::filter` method returns error if there are no log attributes


albertlockett:
I noticed that the `LogFilter::filter` method used by the Filter Processor returns an error if there are no log attributes in the OTAP batch.

Note - we exclude the ID column from the logs records batch if there are no log attributes as this column is null for rows that have no attributes and if the entire column is null then we exclude the column from the record batch entirely.

```rs
use otap_df_pdata::otap::filter::logs::{LogFilter, LogMatchProperties};
use otap_df_pdata::otap::filter::MatchType;
use otap_df_pdata::proto::opentelemetry::logs::v1::{LogRecord, LogsData, ResourceLogs, ScopeLogs};
use otap_df_pdata::testing::round_trip::otlp_to_otap;

fn main() {
    // Filter only for WARN logs
    let include = LogMatchProperties::new(
        MatchType::Strict,
        Vec::new(),
        Vec::new(),
        vec!["WARN".into()],
        None,
        Vec::new(),
    );

    let filter = LogFilter::new(
        Some(include),
        None,
        Vec::new()
    );

    let logs_data = LogsData {
        resource_logs: vec![
            ResourceLogs {
                scope_logs: vec![
                    ScopeLogs {
                        log_records: vec![
                            LogRecord::build().severity_text("WARN").finish(),
                            LogRecord::build().severity_text("WARN").finish(),
                            LogRecord::build().severity_text("INFO").finish(),
                            LogRecord::build().severity_text("INFO").finish(),
                        ],
                        ..Default::default()
                    }
                ],
                ..Default::default()
            }
        ]
    };

    let input = otlp_to_otap(&otap_df_pdata::proto::OtlpProtoMessage::Logs(logs_data));

    // this will panic with: "called `Result::unwrap()` on an `Err` value: ColumnNotFound { name: "id" }"
    let _ = filter.filter(input).unwrap();
}
```

PR #1489, opened 2025-11-28
Fix error when filtering logs if there are no attributes

albertlockett:
closes: #1484 

I noticed when testing some of the filtering code that if we have a Logs OTAP batch with no attributes, an error will be thrown when filtering b/c the ID column is not found. This column is optional if there are no attributes. I corrected the log filtering code and added a test.

I also added a test for the trace filtering code for good measure, even though it isn't afflicted with the same behaviour. It seems that for traces, we treat the ID column as non-optional.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap/filter/logs.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/filter/logs.rs
@@ -529,7 +529,6 @@ mod test {
     use super::*;
     use crate::otap::filter::MatchType;
     use crate::proto::OtlpProtoMessage;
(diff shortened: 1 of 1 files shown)


lalitb:
Thanks, the changes look good. I hit a similar issue earlier when I tried sending logs without attributes through the`otlp-receiver` -> `batch-processor` -> `debug/geneva exporter` pipeline. I assumed the batching and reindexing in the batch processor wasn’t handling the ID column correctly, but I never dug into it. I’ll recheck that flow once this PR is merged.

jmacd (approved):

Issue #487, opened 2025-05-28
Infrastructure for pipeline instrumentation


daviddahl:
The pipeline and its components must be instrumented to provide an external observability system with an accurate representation of the system’s internal state, enabling effective monitoring, debugging, and troubleshooting.

The pipeline engine adopts a thread-per-core, share-nothing architecture to minimize synchronization primitives, optimize the usage of various cache levels, and maximize data locality.

Consequently, the instrumentation framework must:
- Be optimized for thread-per-core operation
- Minimize the use of synchronization primitives to the absolute necessary. Synchronization MUST NOT occur each time a metric is updated or a new event is generated.
- Optimize memory usage (e.g. update metrics in-place).
- Conform to Semantic Convention. Medium term, leverage Weaver to generate our code for signals emission.

Possible approach: Each component (receivers, processors, exporters) will maintain their metrics/events locally, the pipeline engine will scraped these telemetry objects to emit them on a schedule. Each pipeline engine instance will report the collected signals (e.g. every 5s) on a channel locally shared across pipelines. The consumer of this channel could leverage the existing Rust Client SDK.

Other approach to explore: If the existing Rust Client SDK compatible with the requirements mentioned above, we could used it directly from the pipeline instances.

Here is an example of metrics currently used in the Go collector: https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/batchprocessor/documentation.md

lquerel:
I made some changes to the issue description to clarify the requirements that matter most to me.

@jmacd, @cijothomas, @utpilla, could you please review the issue, provide feedback, and potentially add further details based on your knowledge of the client SDK? Thanks!

utpilla:
> Other approach to explore: If the existing Rust Client SDK compatible with the requirements mentioned above, we could used it directly from the pipeline instances.

OpenTelemetry Rust SDK for Metrics does NOT follow a shared-nothing model. The hot path for recording measurements involves looking up a `RwLock` protected `Hashmap` and atomic operations to make the actual update to the metric trackers (for Histogram updates, the SDK uses a `Mutex` as it's not feasible to use atomic operations for it).

[...]

PR #1490, opened 2025-11-29
azure monitor exporter gzip batcher and performance improvements.

gouslu:
- async exports
- ack/nack support
- refactoring for better readability
- other performance improvements leading to 25-30k events per second based on 1kb completely random highly uncompressible events

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
@@ -304,11 +304,9 @@ impl AzureMonitorExporter {
                 return self.queue_pending_batch(effect_handler).await;
             }
             Ok(FinalizeResult::Empty) => Ok(()),
-            Err(e) => {
-                Err(Error::InternalError {
-                    message: format!("Failed to finalize batch: {:?}", e),
-                })
-            },
+            Err(e) => Err(Error::InternalError {
+                message: format!("Failed to finalize batch: {:?}", e),
+            }),
         }
     }
 
@@ -570,7 +568,11 @@ exports | in_flight={} stats_time={:?}
 mod tests {
     use super::super::config::{ApiConfig, AuthConfig, SchemaConfig};
     use super::*;
+    use crate::pdata::Context;
     use azure_core::time::OffsetDateTime;
+    use otap_df_engine::local::exporter::EffectHandler;
+    use otap_df_engine::node::NodeId;
+    use otap_df_telemetry::reporter::MetricsReporter;
     use std::collections::HashMap;
 
     fn create_test_config() -> Config {
@@ -642,5 +644,88 @@ mod tests {
             "Expected at least 29s, got {}s",
             duration_until_refresh.as_secs()
         );
+    }
+
+    #[tokio::test]
+    async fn test_handle_export_success() {
+        let config = create_test_config();
+        let mut exporter = AzureMonitorExporter::new(config).unwrap();
+
+        let (_, reporter) = MetricsReporter::create_new_and_receiver(10);
+        let node_id = NodeId {
+            index: 0,
+            name: "test_exporter".to_string().into(),
+        };
+        let effect_handler = EffectHandler::new(node_id, reporter);
+
+        let batch_id = 1;
+        let msg_id = 100;
+        let context = Context::default();
+        let bytes = Bytes::from("test");
+
+        exporter
+            .state
+            .add_msg_to_data(msg_id, context.clone(), bytes.clone());
+        exporter.state.add_batch_msg_relationship(batch_id, msg_id);
+
+        // This might fail due to missing sender in effect_handler, but state should be updated
+        let _ = exporter
+            .handle_export_success(
+                &effect_handler,
+                batch_id,
+                10.0,
+                std::time::Duration::from_secs(1),
+            )
+            .await;
+
+        // Verify stats
+        assert_eq!(exporter.stats.successful_batch_count(), 1.0);
+        assert_eq!(exporter.stats.successful_msg_count(), 1.0);
+        assert_eq!(exporter.stats.successful_row_count(), 10.0);
+
+        // Verify state cleared
+        assert!(exporter.state.batch_to_msg.is_empty());
+        assert!(exporter.state.msg_to_data.is_empty());
+    }
+
+    #[tokio::test]
+    async fn test_handle_export_failure() {
+        let config = create_test_config();
+        let mut exporter = AzureMonitorExporter::new(config).unwrap();
+
+        let (_, reporter) = MetricsReporter::create_new_and_receiver(10);
+        let node_id = NodeId {
+            index: 0,
+            name: "test_exporter".to_string().into(),
+        };
+        let effect_handler = EffectHandler::new(node_id, reporter);
+
+        let batch_id = 1;
+        let msg_id = 100;
+        let context = Context::default();
+        let bytes = Bytes::from("test");
+
+        exporter
+            .state
+            .add_msg_to_data(msg_id, context.clone(), bytes.clone());
+        exporter.state.add_batch_msg_relationship(batch_id, msg_id);
+
+        let _ = exporter
+            .handle_export_failure(
+                &effect_handler,
+                batch_id,
+                10.0,
+                "Simulated error".to_string(),
+            )
+            .await;
+
+        // Verify stats
+        assert_eq!(exporter.stats.failed_batch_count(), 1.0);
+        assert_eq!(exporter.stats.failed_msg_count(), 1.0);
+        assert_eq!(exporter.stats.failed_row_count(), 10.0);
+
+        // Verify state cleared
+        assert!(exporter.state.batch_to_msg.is_empty());
+        assert!(exporter.state.msg_to_data.is_empty());
     }
 }
--- a/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/gzip_batcher.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/gzip_batcher.rs
@@ -86,9 +86,7 @@ impl GzipBatcher {
 
         if is_first_entry {
             self.batch_id += 1;
-            self.buf
-                .write_all(b"[")
-                .map_err(|e| e.to_string())?;
+            self.buf.write_all(b"[").map_err(|e| e.to_string())?;
         }
 
         // Update calculation to use the constant
@@ -96,8 +94,7 @@ impl GzipBatcher {
         let must_flush = next_size > self.remaining_size;
 
         if must_flush {
-            self.buf.flush()
-                .map_err(|e| e.to_string())?;
+            self.buf.flush().map_err(|e| e.to_string())?;
 
             self.flush_count += 1;
             let compressed_size = self.buf.get_ref().len();
@@ -129,15 +126,11 @@ impl GzipBatcher {
             }
         } else {
             if !is_first_entry {
-                self.buf
-                    .write_all(b",")
-                    .map_err(|e| e.to_string())?;
+                self.buf.write_all(b",").map_err(|e| e.to_string())?;
                 self.total_uncompressed_size += 1;
                 self.uncompressed_size += 1;
             }
-            self.buf
-                .write_all(data)
-                .map_err(|e| e.to_string())?;
+            self.buf.write_all(data).map_err(|e| e.to_string())?;
             self.uncompressed_size += data.len();
             self.total_uncompressed_size += data.len();
             self.row_count += 1.0;
@@ -151,9 +144,7 @@ impl GzipBatcher {
             return Ok(FinalizeResult::Empty);
         }
 
-        self.buf
-            .write_all(b"]")
-            .map_err(|e| e.to_string())?;
+        self.buf.write_all(b"]").map_err(|e| e.to_string())?;
 
         let old_buf = std::mem::replace(&mut self.buf, Self::new_encoder());
 


rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
```diff
@@ -0,0 +1,342 @@
...
+        self.cached_token = AccessToken {
+            token: "".into(),
+            expires_on: OffsetDateTime::now_utc(),
+        };
+    }
+
+    #[allow(clippy::print_stdout)]
+    fn create_credential(auth_config: &AuthConfig) -> Result<Arc<dyn TokenCredential>, String> {
+        match auth_config.method {
+            AuthMethod::ManagedIdentity => {
```
lalitb, line 93: Auth now supports managed identity and dev creds but not Azure Arc MSI; I understand this would be add later, but please add a TODO (and also create issue) tracking Azure Arc MSI support so the exporter can run on edge/Arc devices.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/gzip_batcher.rs
```diff
@@ -0,0 +1,375 @@
...
+        // This algorithm is still very good up to 100KB per entry, which
+        // can be considered quite abnormal for log entries.
+        if data.len() > (ONE_MB - 2) {
+            return PushResult::TooLarge;
+        }
+
+        if self.total_uncompressed_size == 0 {
+            self.buf
+                .write_all(b"[")
+                .expect("write to memory buffer failed");
```
lalitb: With `expect, any unexpected gzip/memory error crashes the exporter instead of surfacing an error.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
```diff
@@ -156,276 +149,198 @@ impl LogsIngestionClient {
...
-        let token = token_response.token.secret();
+    pub async fn send(&mut self, body: Vec<u8>) -> Result<(), String> {
+        // Ensure we have a valid token (fast path when cached)
+        self.ensure_valid_token().await?;
 
-        // Serialize to JSON
-        let json_bytes =
-            serde_json::to_vec(&body).map_err(|e| format!("Failed to serialize to JSON: {e}"))?;
+        // Clone static headers and add the auth header
+        let mut headers = self.static_headers.clone();
```
lalitb: these are static headers - I believe we should be able to avoid cloning in hot path.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/gzip_batcher.rs
```diff
@@ -0,0 +1,375 @@
...
+            self.total_uncompressed_size += 1;
+            self.current_uncompressed_size += 1;
+        } else {
+            self.buf
+                .write_all(b",")
```
lalitb: we append `,` here, but later when the size check triggers a flush, the batch is emitted as `[... ,]` without the attempted element. Trailing comma will make the JSON invalid, and most probably fail at ingestion.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
```diff
@@ -0,0 +1,373 @@
...
+#[allow(clippy::print_stdout)]
+pub struct Auth {
+    credential: Arc<dyn TokenCredential>,
+    scope: String,
+    // Thread-safe shared token cache
+    cached_token: Arc<RwLock<Option<AccessToken>>>,
+}
+
+// TODO: Remove print_stdout after logging is set up
+#[allow(clippy::print_stdout)]
```
jmacd, line 28: I will stake out a position and ask @cijothomas for an opinion too about plain-old-logging APIs with message formatting like `println!("string {variable}")` where `variable` is borrowed if the statement executes.

As I've worked now with Arrow and Datafusion, here's what I find:

1. Not much logging in Arrow itself (Arrow Flight the RPC framework uses `tracing_log`)
2. DataFusion is heavily instrumented with the `log` package with `env_logger` (I know I can set RUST_LOG=debug and learn about its internals)

I like the `log` API surface. 

Now, I'm asking @lquerel for an opinion too. How would you like to see a logging SDK adopted? I think `env_logger` is great in development settings.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
```diff
@@ -0,0 +1,373 @@
...
+#[derive(Clone)]
+// TODO - Remove print statements
+#[allow(clippy::print_stdout)]
+pub struct Auth {
+    credential: Arc<dyn TokenCredential>,
+    scope: String,
+    // Thread-safe shared token cache
+    cached_token: Arc<RwLock<Option<AccessToken>>>,
+}
+
```
jmacd:
```suggestion
// TODO - Consolidate with crates/otap/src/{cloud_auth,object_store)/azure.rs
#[allow(clippy::print_stdout)]
pub struct Auth {
    credential: Arc<dyn TokenCredential>,
    scope: String,
    // Thread-safe shared token cache
    cached_token: Arc<RwLock<Option<AccessToken>>>,
}

```

I recognize this is using things already committed in crates/otap/src/experimental/azure_monitor_exporter/config.rs. 

Have a look at rust/otap-dataflow/crates/otap/src/cloud_auth/azure.rs and rust/otap-dataflow/crates/otap/src/object_store/azure.rs, added subsequently in https://github.com/open-telemetry/otel-arrow/pull/1517. The struct here looks similar to the object_store struct, and there are two similar Auth configs now. It will be nice to see less code and more re-use as we move forward, not to block this PR.

(@lalitb please review. I believe our position should be that Azure auth code/config belongs in an extension, the extension used by parquet_exporter for object_store and by Azure Monitor here.)

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
```diff
@@ -1,31 +1,152 @@
...
+    last_message_received_at: Option<tokio::time::Instant>,
+    idle_duration: tokio::time::Duration,
+    successful_row_count: f64,
+    successful_batch_count: f64,
+    successful_msg_count: f64,
+    failed_row_count: f64,
+    failed_batch_count: f64,
+    failed_msg_count: f64,
+    average_client_latency_secs: f64,
+}
```
jmacd: Have look at a few of the components in the crates/otap/src folder, such as retry_processor.rs, and how there is an existing metrics APIs covering at the Counters here.

As for histogram measurements, I would prefer to hold back. Again, asking @lquerel for opinions: recording histogram instruments could be a thread-local histogram data structure or a message-passing of latency measurements (or both), at which point it may as well be a span message aggregated in the OTel SDK or one of our own pipelines for a latency metric.

We shouldn't reinvent this stuff, see https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/metrics/internal/exponential_histogram.rs.

 I see this as a question of routing histogram measurements (per core) to the central instrumentation collector

@gouslu I understand that some of this is for your own performance studies. Latency and counters aside, can we remove processing_started_at, last_message_received_at, idle_duration, and average_client_latency_secs? 

As for the counters, we aim to standardize pipeline metrics, the topic in #487 and a model RFC in the collector RFC https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md. The counters here are fine until we have a more-standard solution.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
```diff
@@ -34,113 +155,301 @@ impl AzureMonitorExporter {
...
+
+        self.last_batch_queued_at = tokio::time::Instant::now();
+
+        Ok(())
+    }
+
+    async fn handle_pdata(
+        &mut self,
+        effect_handler: &EffectHandler<OtapPdata>,
+        request: ExportLogsServiceRequest,
```
jmacd, line 212: It will be a relatively large performance improvement when we avoid constructing ExportLogsServiceRequest objects and use the view instead. @lalitb please help with pointer and/or an example?

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
```diff
@@ -154,28 +463,118 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
...
+
+                msg = msg_chan.recv().fuse() => {
+                    self.stats.message_received();
+
+                    match msg {
+                        Ok(Message::Control(NodeControlMsg::Shutdown { deadline, .. })) => {
+                            self.handle_shutdown(&effect_handler).await?;
+                            return Ok(TerminalState::new(
+                                deadline,
+                                std::iter::empty::<otap_df_telemetry::metrics::MetricSetSnapshot>(),
```
jmacd, line 483: I mentioned in another comment, we should follow other crates/otap component examples of the metrics integration with crates/telemetry: then you'll have a proper MetricsSet at this point.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/in_flight_exports.rs
```diff
@@ -0,0 +1,100 @@
...
+
+pub struct CompletedExport {
+    pub batch_id: u64,
+    pub client: LogsIngestionClient,
+    pub result: Result<Duration, String>,
+    pub row_count: f64,
+}
+
+pub struct InFlightExports {
+    futures: FuturesUnordered<BoxFuture<'static, CompletedExport>>,
```
jmacd, line 20: :+1:

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
```diff
@@ -154,28 +463,118 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
...
+
+                        println!(
+                            "\n\
+─────────────── AzureMonitorExporter ──────────────────────────
+ perf    │ th/s={:.2}  avg_lat={:.2}ms
+ success │ rows={:.0}  batches={:.0}  msgs={:.0}         
+ fail    │ rows={:.0}  batches={:.0}  msgs={:.0}       
+ time    │ elapsed={:.1}s  active={:.1}s  idle={:.1}s
+ state   | batch_to_msg={}  msg_to_batch={}  msg_to_data={}
+───────────────────────────────────────────────────────────────\n",
```
jmacd, line 539: FYI @cijothomas I've recommended that this component use the crates/telemetry framework, which would mean we could compute performance measurements using the continuous benchmarks. OTOH we would need a server that accepts gzip-compressed-json for the testing. Either way, I've recommended to use the Counter<_> and existing metrics support for now (for counters); asked questions here about logging and histogram measurements.

jmacd:
All of my _real_ questions have to do with instrumentation and the potential for re-use of Azure auth structs. The code looks good!

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
cijothomas: `tracing` macros are what we use in OTel-Rust. It's zero cost when not-enabled. And we can setup subscribers (listeners that routes logs to destination like otel or stdout or file etc..) per thread.

(`tracing` has a `log` bridge too, so the subscriber we use for `tracing` will be able to get `log` events as well)

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/global/internal_logging.rs
OTel Rust wrote wrapper macros, so we can easily switch from `tracing` to something else, if needed.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
lalitb: +1, @gouslu - you can find the example, and the benchmark here - [OtapLogsView]( https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/benchmarks/benches/otap_logs_view/main.rs)

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
lalitb: @jmacd That's correct. Instead of each exporter (Azure Monitor, Parquet/object_store, etc.) implementing its own auth config and credential creation logic, they should all reference this shared extension - and any common requirements should be met by extending this module rather than adding parallel implementations.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
```diff
@@ -66,366 +111,461 @@ impl LogsIngestionClient {
     pub fn new(config: &Config) -> Result<Self, String> {
         let http_client = Client::builder()
             .timeout(Duration::from_secs(30))
+            .http2_prior_knowledge()  // Use HTTP/2 directly (faster, no upgrade negotiation)
```
lalitb: **A note on HTTP/2**: Since Azure Monitor uses HTTPS, ALPN negotiates HTTP/2 as part of the SSL handshake - so there's no extra round trip. Explicitly adding this configuration provides no benefit.

I'd also recommend benchmarking with both HTTP/1.1 and HTTP/2. With HTTP/2, the Geneva backend restricts the client from creating new connections, meaning all requests are multiplexed over a single connection. When payload sizes are large, this single connection can become a bottleneck and lead to timeouts. I ran into this exact issue with the Geneva exporter and had to switch to HTTP/1.1 to resolve it.

Not a blocker for this PR, but good to do some benchmark.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
gouslu: thanks for pointing it out, as you mentioned I am not done with the auth quite yet. I am tracking my work via internal tasks, do you think creating an issue on github on top of what I already have is needed?

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/gzip_batcher.rs
gouslu: this should never happen. I understand the concern, but if this happens it is an indicator of a acritical bug.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/gzip_batcher.rs
gouslu: the API handles this just fine, but you are right that this is not a valid json format, so I will fix that as you suggested.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
gouslu: I need these for myself until I am done. Can put TODO to change this later.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
gouslu: I can look into this in future updates. currently focused on optimizing the performance and refactoring the code.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
gouslu: created a TODO for tracking metrics, will work with cijo for perf testing.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
gouslu: I was considering this as well, switched back to 1.1. I did 't have any perf differences, but chose 1.1 because it is better for creating 1 http client per API client, which is what helps improve perf a lot in my case.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/gzip_batcher.rs
andborja: Agree with @lalitb. Let's avoid expect usage (panic)

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
jmacd: :+1:

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
jmacd: It's not a priority IMO to post publish issues about azure-specific development.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
```diff
@@ -154,30 +424,223 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
...
+                        other => {
+                            self.handle_message(&effect_handler, other, &mut msg_id).await?;
+                        }
                     }
                 }
-                _ => {
-                    // Ignore other message types
+
+                _ = tokio::time::sleep_until(next_stats_print) => {
+                    next_stats_print = tokio::time::Instant::now() + tokio::time::Duration::from_secs(STATS_PRINT_INTERVAL);
```
jmacd, line 493: OK for experimental. We'll want to use the built-in metrics, and we'll want to make the continuous benchmark support this exporter as we've discussed.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
jmacd: >  I ran into this exact issue with the Geneva exporter and had to switch to HTTP/1.1 to resolve it.

I've seen this addressed by installing "multi-connection" resolvers in gRPC where a single IP address is replicated w/ different metadata to trick the exporter into using multiple connections. I was at one point being asked to try contributing a multi-conn feature to OTel Collector, but my team and I pushed back successfully: the better solution (we advised & implemented) is for load-balanced frontends to expose multiple IP addresses so exporters just scale.

PR #1495, opened 2025-12-01
[otap-df-quiver] Scaffolding for otap-df-quiver crate

AaronRM:
Introduces the experimental `otap-df-quiver` crate into the main `otap-dataflow` Rust workspace. Changes include configuration scaffolding, placeholder APIs, and benchmarking stubs, but do not yet implement actual data persistence. Integration is opt-in via the new `quiver-persistence` Cargo feature, which is disabled by default to keep the experimental code path out of release builds.

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/engine.rs
+++ b/rust/otap-dataflow/crates/quiver/src/engine.rs
@@ -94,10 +94,10 @@ mod tests {
 
         fn payload(&self, slot: SlotId) -> Option<PayloadRef<'_>> {
             if slot == SlotId::new(0) {
-                    Some(PayloadRef {
-                        schema_fingerprint: [0; 32],
-                        batch: &self.batch,
-                    })
+                Some(PayloadRef {
+                    schema_fingerprint: [0; 32],
+                    batch: &self.batch,
+                })
             } else {
                 None
             }


rust/otap-dataflow/crates/quiver/src/record_bundle.rs
```diff
@@ -0,0 +1,142 @@
...
+//! Minimal traits that adapters implement to feed data into Quiver.
+
+use std::borrow::Cow;
+use std::time::SystemTime;
+
+use arrow_array::RecordBatch;
+
+/// Logical identifier for a payload slot inside a bundle.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct SlotId(pub u16);
```
jmacd, line 13: Suggest having a look at the `slotmap` type, see rust/otap-dataflow/crates/otap/src/accessory/slots.rs!
(Not to block this PR.)

AaronRM: Thanks for the pointer! I think we still would want this as its own type in quiver since we are trying to keep quiver from having a dependency on otap-dataflow.

rust/otap-dataflow/crates/quiver/src/record_bundle.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    /// Looks up the descriptor for a slot id, if present.
+    #[must_use]
+    pub fn get(&self, id: SlotId) -> Option<&SlotDescriptor> {
+        self.slots.iter().find(|slot| slot.id == id)
+    }
+}
+
+/// Stable fingerprint for a payload schema (currently 128 bits).
+pub type SchemaFingerprint = [u8; 16];
```
lalitb: nit - this seems to be inconsistent with the Architecture doc - 
```
  - Slot bitmap (currently a single `u64`) followed by `slot_count` metadata
    blocks. Each block contains `payload_type_id: u16`,
    `schema_fingerprint: [u8;32]`, `row_count: u32`, and
    `payload_len: u32`.
```

AaronRM: Good catch, thanks!

PR #1496, opened 2025-12-01
Fix cargo build warnings

AaronRM:
Fix cargo build warnings:

1. Use correct syntax for deprecated `serde_yaml` dependency
2. Move bench profile settings to workspace level `Cargo.toml`


rust/otap-dataflow/Cargo.toml
```diff
@@ -97,7 +97,7 @@ serde = { version = "1.0.219", features = ["derive", "rc"] }
 serde_cbor = "0.11.2"
 serde_json = { version = "1.0.142" }
 serde_with = { version = "3.14.1", features = ["std", "macros", "json"] }
-serde_yaml = "0.9.34+deprecated"        # Deprecated, but no good alternative yet
+serde_yaml = "0.9.34-deprecated"        # Deprecated, but no good alternative yet
```
jmacd, line 103: This is mildly confusing -- what's the difference between + and - in this setting and/or what was the warning?

AaronRM: With "+deprecated", cargo build issues this error about the '+deprecated' SemVer metadata being ignored.

```
otel-arrow/rust/otap-dataflow/crates/telemetry/Cargo.toml: version requirement `0.9.34+deprecated` for dependency `serde_yaml` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
```

The actual versions on crates.io appear to be identical whether you refer to `0.9.34` or `0.9.34+deprecated` or `0.9.34-deprecated`. (The more relevant issue is going to be getting off this crate altogether.)

PR #1488, opened 2025-11-27
Optimize function used to filter child record batches with u16 parent_ids

albertlockett:
part of #1390 

When benchmarking/profiling filtering for the columnar query engine (#1477 / #1342, the poc), I noticed we were spending a lot of time in `build_uint16_id_filter`. 

This function is shared with the filter processor, and it is used after we filter the parent record batch to remove rows in children that point to rows that were filtered out. To do this, we take the set of IDs from the parent, and then build a selection vector for the child record batch.

This PR tries to optimize this function. I added a benchmark that swept across:
- batch sizes
- number of child records per parent
- filter selectivities (e.g. what fraction of the IDs to filter)
- sort order
   - sorted by parent ID
   - "transport optimized" sort order (where the parent IDs are mixed up b/c we sorted by attribute keys/values first)

I then used the results of this benchmark to make the following optmizations:

**1. Use the optimized `BooleanBuilder::append_n` when building up the result.**
Turns out this is faster than just appending values one at a time: https://github.com/apache/arrow-rs/blob/1c90efe717c1a2f99a0a97d6e9c93cd913eaca3c/arrow-buffer/src/builder/boolean.rs#L164-L196

**2. Use a bitmap for the ID set instead of a HashSet**
This saves us some cycles because we don't need to hash everything in the ID column when doing lookups in the ID set.

**3. Be more restrictive when using the "simd" code path**
When the ID column is very long, but there are relatively few records in the ID set, we determined that it was faster to use this algorithm that invokes the arrow `eq` and `or_kleene` compute kernels. See the discussion [here](https://github.com/open-telemetry/otel-arrow/pull/1279#discussion_r2453094683).

There are some heuristics in place for when to choose this code path: the ID column must be sufficiently long and the ID set must be short in comparison.

However, if the ID column is very very long, the ID set could still be short enough in comparison to the ID column, but still be so long that we iterate this code path many times. This is slow!

I did some testing and figured out that, at least on my laptop, this code path tends to be slower after about 20 iterations even if the ID column is very very long. So now we check this as well.

---
We should probably do the same for the u32 version of this function, which can happen in a followup.

PR #1497, opened 2025-12-02
columnar query engine filtering handle adaptive OTAP schemas

### LINKED ISSUE #1494: Columnar query engine filtering - adaptive schema planning improvements
In #1477 we've implemented the filtering `PipelineStage` for the columnar query engine, but there are a handful of situations that are not yet handled which all related adaptive schemas. All of the following can change between subsequent record batches:
- [ ] Presence of optional payload type's `RecordBatch`
- [ ] Presence of optional columns
- [ ] Column data types (dict<u8/u16> / native)
- [ ] Column order

We need to add support for each of these in such a way that we can efficiently detect the situation, and make minimal modification to either the query plan or the record batch so the filter pipeline stage will not throw an error and will produce correct results.

albertlockett:
closes #1494 

Adds code to the `FilterPipelineStage` for the columnar query engine to handle all the interesting ways that OTAP batches might change change from one batch to the next including:
- Presence of optional payload type's RecordBatch
- Presence of optional columns
- Column data types (dict / native)
- Column order rearranged in a given record batch

A new type called `AdaptivePhysicalExprExec` is added to encapsulate the logic around all of this.

To handle optional `RecordBatch`s for some payloads, `AdaptivePhysicalExprExec` lazily converts the logical `Expr` into the `PhysicalExpr` only when a record batch for some type is present, which means we don't need all payload types to be present in the first batch we see.

To handle optional columns and columns changing order, `AdaptivePhysicalExprExec` determine which columns are referenced in the DataFusion logical `Expr` that defines the filter by visiting it using a new type called`ProjectedSchemaExprVisitor`. The result of the expression traversal is then captured in a type called `FilterProjection`. Before applying the filter, we attempt do a lightweight projection to create a new record batch with only the referenced columns in the correct order. If the columns are missing, this projection returns `None` and we treat it as no rows passed the filter (will correct this logic in a future PR when we handle predicates like `something == null`).

Tests are added to cover the new functionality and the tests that were previously ignored in this module are now passing

### DIFF
--- a/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
+++ b/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
@@ -620,11 +620,9 @@ impl Composite<AttributeFilterExec> {
 }
 
 /// This is responsible for evaluating a  [`PhysicalExpr`](datafusion::physical_expr::PhysicalExpr)
-/// while adapting to "schema" changes that may be encountered between evaluations.
+/// while adapting to schema changes that may be encountered between evaluations.
 ///
-/// OTAP data might have minor changes the "schema" between batches.
-/// - Optional payload type's [`RecordBatch`]s may change presence
-/// - Optional columns within a [`RecordBatch`] may change presence
+/// A given payload type's [`RecordBatch`] might have minor changes the schema between batches
 /// - The type of a column may change between Dict<u8, V>, Dict<16, V>, and the native array type
 /// - The order of columns may change
 ///


rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
```diff
@@ -734,7 +994,7 @@ impl FilterPipelineStage {
 
         // build the selection vector for the child record batch. This uses common code shared
         // with the filter processor
-        let id_mask: HashSet<u16> = id_col.iter().flatten().collect();
+        let id_mask = id_col.iter().flatten().map(|i| i as u32).collect();
```
albertlockett, line 998: This corrects the type as `build_uint16_id_filter` since https://github.com/open-telemetry/otel-arrow/pull/1488

rust/experimental/query_engine/engine-columnar/profile.json.gz
jmacd, line 1: Is this intentional?

rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
albertlockett: also fixed in #1501

rust/experimental/query_engine/engine-columnar/profile.json.gz
albertlockett: committing it in the first place was not intentional, but deleting it in this PR is intentional :D

albertlockett:
> The code is very readable! I may be losing track of the overall design as it evolves, however. I understand that a SessionContext is the primary Datafusion object we are programming with, and take it that FilterStagePipeline is an assembly of logic that builds physical plans. I'm missing somehow what the session_ctx is now doing as it passes through. Are there table providers, in-memory tables, etc carried there that I should know about, that sort of question.

Hey @jmacd ! I think that's a pretty good understanding.

Basically the [`SessionContext`](https://docs.rs/datafusion/latest/datafusion/execution/context/struct.SessionContext.html) is a thin wrapper around the [`SessionState`](https://docs.rs/datafusion/latest/datafusion/execution/session_state/struct.SessionState.html).

The `SessionContext` provides APIs for executing queries, whereas the internal `SessionState` provides everything needed for query planning. This includes the `TableProvider`s you mentioned, as well as planners, optimizers registered functions, and the [`SessionConfig`](https://docs.rs/datafusion/latest/datafusion/execution/context/struct.SessionConfig.html) which contains information for _how_ to plan queries (options for batch size, partitioning, predicate pushdown, etc.).

One of my goals in this implementation is to have these readily available for various pipeline stages and whatever unique and interesting ways they made need to invoke datafusino.

From the otel-arrow side, we have the `PipelineStage` trait which is supposed to represent step in the columnar query engine's pipeline that transforms an `OtelArrowRecords`. The `FilterPipelineStage` is the implementation of this trait that performs filtering.

Below that, my mental model of the hierarchy of types is something like this:
```rs
FilterPipelineStage {
  // Composite here represents some logical and/or/not combination of filters that will be applied
  filter_exec: Composite<FilterExec>
}

// performs filtering on the root record batch by executing the predicate
// and can also filter by attributes by invoking the attribute filter
FilterExec {
  predicate: Option<AdaptivePhysicalExprExec>
  attributes_filter: Option<Composite<AttributeFilterExec>>
}

// performs filtering on attributes, returning the parent_IDs that pass the filter
AttributeFilterExec {
  filter: AdaptivePhysicalExprExec
}

// PhysicalExpr is a predicate that evaluates to a boolean, so this will generate a selection vector
// for which rows pass the predicate
AdaptivePhysicalExprExec {
    /// The original logical plan used to produce the [`PhysicalExpr`]
   logical_expr: Expr,

   /// initialized lazily
   physical_expr: Option<datafusion::physical_expr::PhysicalExpr>,
}
```

In `AdaptivePhysicalExprExec`, it creates the datafusion `PhysicalExpr` using: the logical `Expr`, the `SessionState` (which we get from the `SessionContext` and, recall, has everything needed to plan queries), and the schema of the incoming batch (which has the types for the columns). This logical -> physical planning can't happen until we have the schema, but now knowing if the `RecordBatch` will be available in any given batch, we must do the planning lazily. This is why why we now pass the SessionContext all the way down from `FilterPipelineStage` to `AdaptivePhysicalExprExec` alongside the record batches so it's available for this lazy planning.

Hope that clears things up, and let me know if you have any other questions!

PR #1498, opened 2025-12-02
chore: Add fossa badges

thompson-tomo:
This adds the fossa badges to the readme to improve the clo monitor score.


trask:
@thompson-tomo please create an issue in https://github.com/open-telemetry/sig-security to discuss first before sending these PRs so we can agree on a plan and we can then link these PRs to that plan in order to provide more context to maintainers. Thanks!

PR #1504, opened 2025-12-02
[otap-dataflow] fix exclude filter returning empty batch #1483

### LINKED ISSUE #1483: `LogFilter` with only exclude critieria unexpectedly returns empty batch
I was benchmarking the LogFilter code (which gets used in our filter processor) against what was added in #1477 and I noticed that if I create a `LogFilter` with only exclude `LogMatchProperties`, the `filter` method always seems to return an empty batch. I wasn't expecting this, and I'm wondering if this is a bug or if I'm just not supplying the `LogMatchProperties` correctly.

```rs
use otap_df_pdata::otap::filter::logs::{LogFilter, LogMatchProperties};
use otap_df_pdata::otap::filter::MatchType;
use otap_df_pdata::proto::opentelemetry::arrow::v1::ArrowPayloadType;
use otap_df_pdata::proto::opentelemetry::common::v1::{AnyValue, KeyValue};
use otap_df_pdata::proto::opentelemetry::logs::v1::{LogRecord, LogsData, ResourceLogs, ScopeLogs};
use otap_df_pdata::testing::round_trip::otlp_to_otap;

fn main() {
    let exclude = LogMatchProperties::new(
        MatchType::Strict,
        Vec::new(), // don't exclude any resource attrs
        Vec::new(), // don't exclude any log attrs,
        vec!["WARN".into()], // exclude severity_text == "WARN",
        None, // don't exclude any severity_numbers
        Vec::new(), // don't exclude any bodies
    );

    let filter = LogFilter::new(
        None, // no filters for what to include
        Some(exclude),
        Vec::new() // no ottl filters
    );

    let logs_data = LogsData {
        resource_logs: vec![
            ResourceLogs {
                scope_logs: vec![
                    ScopeLogs {
                        log_records: vec![
                            // WARN rows should be excluded
                            LogRecord::build().severity_text("WARN").finish(),
                            LogRecord::build().severity_text("WARN").finish(),
                            // INFO rows should not be excluded
                            LogRecord::build().severity_text("INFO").finish(),
                            LogRecord::build()
                                .severity_text("INFO")
                                .attributes(vec![KeyValue::new("x", AnyValue::new_string("y"))])
                                .finish(),
                        ],
                        ..Default::default()
                    }
                ],
                ..Default::default()
            }
        ]
    };

    let input = otlp_to_otap(&otap_df_pdata::proto::OtlpProtoMessage::Logs(logs_data));

    let (result, _, _) = filter.filter(input).unwrap();

    let logs = result.get(ArrowPayloadType::Logs).unwrap();

    // expect two rows, but there are none
    println!("there are {:?} rows", logs.num_rows());
}
```

c1ly:
Fixes #1483

Adds check for fields where an empty vec provided so nothing should be excluded


rust/otap-dataflow/crates/otap/src/filter_processor.rs
```diff
@@ -1074,6 +1072,15 @@ mod tests {
...
 
+        //         let exclude_props = LogMatchProperties::new(
+        //     MatchType::Strict,
+        //     Vec::new(), // don't exclude any resource attrs
+        //     Vec::new(), // don't exclude any log attrs,
+        //     vec!["WARN".into()], // exclude severity_text == "WARN",
+        //     None, // don't exclude any severity_numbers
+        //     Vec::new(), // don't exclude any bodies
+        // );
+
```
albertlockett: We can remove the commented code, right?
```suggestion
```

c1ly: yup. removed

albertlockett:
Great @c1ly!

Would you mind adding a simple test case in both `otap::filter::logs` and `otap::filter::traces` to confirm the behaviour?

They can be similar to the ones we have already, but instead just passing the match properties as `exclude` instead of `include`
https://github.com/open-telemetry/otel-arrow/blob/f9aec883865d4a3bff625554bcf6d04e1737559e/rust/otap-dataflow/crates/pdata/src/otap/filter/logs.rs#L538-L585
https://github.com/open-telemetry/otel-arrow/blob/f9aec883865d4a3bff625554bcf6d04e1737559e/rust/otap-dataflow/crates/pdata/src/otap/filter/traces.rs#L692-L739

PR #1510, opened 2025-12-02
  Add TLS to OTLP/OTAP receivers (server-side, with reload)

lalitb:
This PR adds server-side TLS for receivers, following the Go collector's approach. Plan is to submit changes incrementally to make the review easier and to get feedback on the implementation direction before raising further PRs.

  **TL;DR:** This enables TLS for receivers. mTLS and exporter client-side TLS are not part of this PR.

  ### What's in this PR
  **Server-side TLS for receivers:**
  - OTLP and OTAP receivers can now accept TLS connections
  - Certificate and key configuration via `cert_file`/`key_file` or `cert_pem`/`key_pem`
  - Async certificate hot-reload without service restart or connection drops

  **Certificate reload behavior:**
  - Similar to the Go collector's implementation
  - File-based certificates are monitored via mtime
  - Reload triggered asynchronously when files change and `reload_interval` has elapsed
  - Active connections continue using the current certificate while reload happens in background
  - Zero downtime during rotation assumes the new certificate is deployed before the old certificate expires, creating an overlap window where both are valid.
  The `reload_interval` should be set appropriately based on your certificate rotation schedule (e.g., if certificates are rotated daily, use `reload_interval:
  "5m"` to ensure timely pickup).

  **Implementation notes:**
  - Uses `rustls` for TLS
  - Certificate reload runs in spawned async tasks
  - Compare-exchange used to ensure single reload task at a time
  - **NUMA note:** Each receiver creates its own TLS resolver instance (per-core design). Arc-wrapped atomics are used for local coordination between the resolver and its async reload task. No cross-NUMA sharing.

  ### What's NOT in this PR

  **Deferred to follow-up PRs:**
  - mTLS (mutual TLS / client certificate verification)
  - Client-side TLS for exporters
  - Comprehensive benchmarks (planned after full TLS implementation)

  ### Feature flag

  Currently gated behind `experimental-tls` feature flag. This will be removed once we're confident the implementation is stable and production-ready.

  ### Testing

  Manual E2E testing completed:
  - Basic TLS handshake and data transmission
  - Certificate hot-reload during active connections  
  - Plaintext rejection when TLS is enabled
  - Certificate chain validation
  - Backward compatibility (plaintext mode works without feature flag)

  Automated tests included:
  - Unit tests for reload logic (`tls_utils::tests`)
  - Integration test for end-to-end certificate rotation (`tests/tls_reload.rs`)

  ---

  **Configuration example:**

  ```yaml
  receivers:
    otlp:
      config:
        listening_addr: "0.0.0.0:4319"
        tls:
          cert_file: "/path/to/cert.pem"
          key_file: "/path/to/key.pem"
          reload_interval: "5m"  # Optional, defaults to 5 minutes

 
```

### DIFF
--- a/rust/experimental/query_engine/engine-columnar/benches/filter.rs
+++ b/rust/experimental/query_engine/engine-columnar/benches/filter.rs
@@ -30,8 +30,9 @@ fn bench_log_pipeline(
         let _ = group.bench_with_input(benchmark_id, &batch_size, |b, batch_size| {
             b.iter_custom(|iters| {
                 let batch = generate_logs_batch(**batch_size);
-                let query = KqlParser::parse(bench_pipeline_kql).expect("can parse pipeline");
-                let mut pipeline = Pipeline::new(query);
+                let parser_result =
+                    KqlParser::parse(bench_pipeline_kql).expect("can parse pipeline");
+                let mut pipeline = Pipeline::new(parser_result.pipeline);
                 rt.block_on(async move {
                     // execute the query once to initiate planning
                     pipeline.execute(batch.clone()).await.unwrap();
--- a/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
+++ b/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
@@ -1177,8 +1177,8 @@ mod test {
 
     pub async fn exec_logs_pipeline(kql_expr: &str, logs_data: LogsData) -> LogsData {
         let otap_batch = otlp_to_otap(&OtlpProtoMessage::Logs(logs_data));
-        let pipeline_expr = KqlParser::parse(kql_expr).unwrap();
-        let mut pipeline = Pipeline::new(pipeline_expr);
+        let parser_result = KqlParser::parse(kql_expr).unwrap();
+        let mut pipeline = Pipeline::new(parser_result.pipeline);
         let result = pipeline.execute(otap_batch.clone()).await.unwrap();
         otap_to_logs_data(result)
     }
@@ -1200,8 +1200,8 @@ mod test {
                 .finish(),
         ]);
 
-        let pipeline_expr = KqlParser::parse("logs | where severity_text == \"ERROR\"").unwrap();
-        let mut pipeline = Pipeline::new(pipeline_expr);
+        let parser_result = KqlParser::parse("logs | where severity_text == \"ERROR\"").unwrap();
+        let mut pipeline = Pipeline::new(parser_result.pipeline);
         let result = pipeline.execute(otap_batch.clone()).await.unwrap();
         let expected = to_otap(vec![
             LogRecord::build()
@@ -1212,8 +1212,8 @@ mod test {
         assert_eq!(result, expected);
 
         // test same filter where the literal is on the left and column name on the right
-        let pipeline_expr = KqlParser::parse("logs | where \"ERROR\" == severity_text").unwrap();
-        let mut pipeline = Pipeline::new(pipeline_expr);
+        let parser_result = KqlParser::parse("logs | where \"ERROR\" == severity_text").unwrap();
+        let mut pipeline = Pipeline::new(parser_result.pipeline);
         let result = pipeline.execute(otap_batch.clone()).await.unwrap();
         assert_eq!(result, expected);
     }
@@ -1242,8 +1242,8 @@ mod test {
                 .finish(),
         ];
 
-        let pipeline_expr = KqlParser::parse("logs | where attributes[\"x\"] == \"b\"").unwrap();
-        let mut pipeline = Pipeline::new(pipeline_expr);
+        let parser_result = KqlParser::parse("logs | where attributes[\"x\"] == \"b\"").unwrap();
+        let mut pipeline = Pipeline::new(parser_result.pipeline);
         let result = pipeline.execute(otap_batch.clone()).await.unwrap();
         let result_otlp = otap_to_logs_data(result);
         pretty_assertions::assert_eq!(
@@ -1252,8 +1252,8 @@ mod test {
         );
 
         // test same filter where the literal is on the left and the attribute is on the right
-        let pipeline_expr = KqlParser::parse("logs | where \"b\" == attributes[\"x\"]").unwrap();
-        let mut pipeline = Pipeline::new(pipeline_expr);
+        let parser_result = KqlParser::parse("logs | where \"b\" == attributes[\"x\"]").unwrap();
+        let mut pipeline = Pipeline::new(parser_result.pipeline);
         let result = pipeline.execute(otap_batch.clone()).await.unwrap();
         let result_otlp = otap_to_logs_data(result);
         pretty_assertions::assert_eq!(
@@ -1449,10 +1449,10 @@ mod test {
         let otap_batch = to_otap(log_records.clone());
 
         // check simple filter "and" properties
-        let pipeline_expr =
+        let parser_result =
             KqlParser::parse("logs | where severity_text == \"ERROR\" and event_name == \"2\"")
                 .unwrap();
-        let mut pipeline = Pipeline::new(pipeline_expr);
+        let mut pipeline = Pipeline::new(parser_result.pipeline);
         let result = pipeline.execute(otap_batch.clone()).await.unwrap();
         let result_otlp = otap_to_logs_data(result);
         pretty_assertions::assert_eq!(
@@ -1461,11 +1461,11 @@ mod test {
         );
 
         // check simple filter "and" with attributes
-        let pipeline_expr = KqlParser::parse(
+        let parser_result = KqlParser::parse(
             "logs | where severity_text == \"ERROR\" and attributes[\"x\"] == \"c\"",
         )
         .unwrap();
-        let mut pipeline = Pipeline::new(pipeline_expr);
+        let mut pipeline = Pipeline::new(parser_result.pipeline);
         let result = pipeline.execute(otap_batch.clone()).await.unwrap();
         let result_otlp = otap_to_logs_data(result);
         pretty_assertions::assert_eq!(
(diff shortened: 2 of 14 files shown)

### CODE REVIEW

File: rust/otap-dataflow/crates/otap/src/tls_utils.rs [hunk 1]
```diff
@@ -0,0 +1,696 @@
...
+/// - Mtime checks for the certificate and key files are throttled by a configurable interval (`check_interval_secs`)
+///   to minimize filesystem operations. This means the filesystem is not polled on every handshake, but only after
+///   the specified interval has elapsed since the last check.
+/// - If a file change is detected during a TLS handshake (i.e., the mtime has changed since the last check),
+///   the certificate and key are reloaded asynchronously. This ensures that new certificates are picked up without
+///   requiring a server restart, while avoiding excessive filesystem access.
+/// - Reloads happen in the context of TLS handshakes. Certificate updates are applied to subsequent connections after
+///   the async reload completes, which typically happens within a short time after file modification is detected.
+#[derive(Debug)]
+pub struct LazyReloadableCertResolver {
```
lalitb, line 125: Each receiver creates its own TLS resolver instance (per-core by design). Arc-wrapped atomics are used to coordinate between the resolver and its async reload task on the same core. No cross-NUMA sharing.

rust/otap-dataflow/crates/config/src/tls.rs
```diff
@@ -0,0 +1,78 @@
...
+    // pub cipher_suites: Option<Vec<String>>,
+
+    // /// Trusted platform module configuration
+    // pub tpm_config: Option<TpmConfig>,
+    /// Minimum interval between certificate reload checks.
+    /// Certificates are only reloaded if file modification time has changed.
+    /// Defaults to 5 minutes ("5m").
+    /// Format: Standard duration string (e.g., "30s", "5m", "1h").
+    #[serde(default = "default_reload_interval")]
+    pub reload_interval: Option<String>,
```
jmacd: I would like us to use std::time::Duration and the humantime serde-with feature, like

https://github.com/open-telemetry/otel-arrow/pull/1511#discussion_r2586266994

this has been inconsistent in the code, this one preferred in my opinion.

rust/otap-dataflow/crates/otap/src/otap_receiver.rs
```diff
@@ -318,7 +338,23 @@ impl shared::Receiver<OtapPdata> for OTAPReceiver {
             },
 
             // Run server
-            result = server.serve_with_incoming(listener_stream) => {
+            // Note: Unlike otlp_receiver.rs, this uses an inline match because the server
+            // has a middleware layer applied which changes the Router type, making it
+            // incompatible with a shared helper function without complex generic bounds.
+            result = async {
```
jmacd, line 344: 👍

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
```diff
@@ -71,6 +75,10 @@ pub struct Config {
     /// Format: humantime format (e.g., "30s", "5m", "1h", "500ms")
     #[serde(default, with = "humantime_serde")]
```
jmacd: (e.g.,)

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -0,0 +1,696 @@
...
+
+    /// Last known modification times (Arc for sharing with async reload tasks)
+    cert_mtime: Arc<AtomicU64>,
+    key_mtime: Arc<AtomicU64>,
+
+    /// Last time we checked mtime (unix timestamp in seconds)
+    last_check_time: AtomicU64,
+
+    /// Minimum interval between mtime checks (seconds)
+    check_interval_secs: u64,
```
jmacd, line 141: I would prefer std::time::Duration to be consistent across the code.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -0,0 +1,696 @@
...
+        key_path: PathBuf,
+        check_interval: Option<Duration>,
+    ) -> Result<Self, io::Error> {
+        let cert_key = load_certified_key_sync(&cert_path, &key_path)?;
+        let cert_mtime = get_mtime(&cert_path)?;
+        let key_mtime = get_mtime(&key_path)?;
+        let now = current_timestamp();
+
+        Ok(Self {
+            cert_key: Arc::new(ArcSwap::from_pointee(cert_key)),
```
jmacd, line 160: TIL: arc_swap meant for this kind of long-lived periodically updated runtime and/or configuration struct. https://docs.rs/arc-swap/latest/arc_swap/

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -0,0 +1,696 @@
...
+
+        // Spawn async reload - doesn't block the current handshake
+        // Fire-and-forget: we intentionally don't await the task.
+        // Using drop() to explicitly ignore the JoinHandle and satisfy clippy::let_underscore_future.
+        drop(tokio::spawn(async move {
+            match load_certified_key_async(&cert_path, &key_path).await {
+                Ok(new_cert) => {
+                    cert_key.store(Arc::new(new_cert));
+                    cert_mtime.store(current_cert_mtime, Ordering::Relaxed);
+                    key_mtime.store(current_key_mtime, Ordering::Relaxed);
```
jmacd, line 248: 👍 👍

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -0,0 +1,696 @@
...
+fn get_mtime(path: &PathBuf) -> Result<u64, io::Error> {
+    std::fs::metadata(path)?
+        .modified()?
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .map(|d| d.as_secs())
+        .map_err(io::Error::other)
+}
+
+/// Async version for background reloads - doesn't block handshakes
+async fn load_certified_key_async(
```
jmacd: (Wish I could answer for myself, and/or I didn't have to reveal such ignorance!!!)

I see three functions with parallel implementations synchonous and asynchronous, one used during start and one during reload. Is there a way to block-on the asynchronous form at start, to share the code?

rust/otap-dataflow/crates/otap/tests/tls_stream.rs
```diff
@@ -0,0 +1,276 @@
...
+    use std::process::Command;
+    use std::sync::Arc;
+    use std::time::Duration;
+    use tempfile::TempDir;
+    use tokio::io::{AsyncReadExt, AsyncWriteExt};
+    use tokio::net::{TcpListener, TcpStream};
+    use tokio_rustls::TlsAcceptor;
+    use tokio_stream::wrappers::TcpListenerStream;
+
+    fn generate_ca(dir: &std::path::Path, name: &str, cn: &str) {
```
jmacd, line 21: Could this test helper be shared? (I see a similar copy above.)

rust/otap-dataflow/crates/otap/tests/tls_stream.rs
```diff
@@ -0,0 +1,276 @@
...
+            ])
+            .current_dir(dir)
+            .output()
+            .expect("Failed to generate CA");
+        if !status.status.success() {
+            panic!("CA gen failed: {}", String::from_utf8_lossy(&status.stderr));
+        }
+    }
+
+    fn generate_server_cert(dir: &std::path::Path, name: &str, ca_name: &str, cn: &str) {
```
jmacd, line 50: Could this test helper be shared? (I see a similar copy above.)

rust/otap-dataflow/crates/config/src/tls.rs
lalitb: Yes, agree. Done.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
lalitb: I have updated the public API to use `std::time::Duration`, but it would be better to keep the internal representation as u64 for implementation efficiency.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
lalitb: Exactly! It's perfect for this scenario where reads are extremely frequent (every TLS handshake) but writes are rare (only on certificate rotation).

rust/otap-dataflow/crates/otap/src/tls_utils.rs
lalitb: this is good comment.

>  Is there a way to block-on the asynchronous form at start, to share the code?

We could technically use `futures::executor::block_on(async_function())` inside `LazyReloadableCertResolver::new()`, but it's dangerous. If `new()` is ever called from within an existing Tokio runtime (e.g. inside `#[tokio::main]`), it will panic because we'd be trying to start a blocking runtime inside an async one. Keeping them separate avoids this risk.

Though we can keep both the functions, and extract the common parsing logic in separate private function:

```
Sync function: Reads files from disk (sync) -> calls parse_helper.
Async function: Reads files from disk (async) -> calls parse_helper.
```

I have made changes as above.

rust/otap-dataflow/crates/otap/tests/tls_stream.rs
lalitb: I looked into sharing them, but it's a bit tricky because `tls_utils.rs` is a unit test (inside `src/`) while `tls_stream.rs` is an integration test (inside `tests/`). Sharing code between them would require exposing these test helpers in the `public`.

rust/otap-dataflow/crates/otap/tests/tls_stream.rs
lalitb: As mentioned [here](https://github.com/open-telemetry/otel-arrow/pull/1510#discussion_r2587037425)

jmacd:
Thanks for explaining!

PR #1511, opened 2025-12-02
feat: Integrate Opentelemetry Rust SDK for internal metrics (phase 1)

andborja:
Integrates Opentelemetry Rust SDK for internal metrics (phase 1).

### DIFF
--- a/rust/otap-dataflow/Cargo.toml
+++ b/rust/otap-dataflow/Cargo.toml
@@ -38,6 +38,7 @@ quiver = { workspace = true, optional = true }
 serde_json.workspace = true
 clap.workspace = true
 mimalloc-rust.workspace = true
+rustls = { workspace = true, optional = true }
 
 [workspace.dependencies]
 otap-df-pdata-otlp-macros = { path = "./crates/pdata/src/otlp/macros"}
@@ -48,6 +49,7 @@ otap-df-otap = { path = "crates/otap" }
 quiver = { package = "otap-df-quiver", path = "crates/quiver" }
 
 ahash = "0.8.11"
+arc-swap = "1.7"
 arrayvec = "0.7.6"
 arrow = { version = "57.0", features=["prettyprint"] }
 arrow-ipc = { version = "57.0", features=["zstd"] }
@@ -57,6 +59,7 @@ async-stream = "0.3.6"
 async-trait = "0.1.88"
 async-unsync = "0.3.0"
 axum = "0.8.4"
+base64 = "0.22.1"
 bitflags = "2.10"
 bytes = "1.11"
 bytemuck = "1.24"
@@ -98,6 +101,10 @@ prost = "0.14"
 quote = "1.0"
 rand = "0.9.2"
 roaring = "0.11.2"
+rustls = { version = "0.23.22", default-features = false, features = ["ring", "std", "tls12"] }
+rustls-pemfile = "2.0"
+tokio-rustls = { version = "0.26", default-features = false, features = ["ring", "tls12"] }
+rustls-native-certs = "0.8.2"
 schemars = { version = "1.0.0" }
 serde = { version = "1.0.228", features = ["derive", "rc"] }
 serde_cbor = "0.11.2"
@@ -157,6 +164,8 @@ unchecked-index = []
 unchecked-arithmetic = []
 # Experimental persistence prototype (opt-in)
 quiver-persistence = ["quiver"]
+# Experimental features
+experimental-tls = ["otap-df-otap/experimental-tls", "dep:rustls"]
 # Experimental exporters (opt-in)
 experimental-exporters = ["otap-df-otap/experimental-exporters"]
 geneva-exporter = ["otap-df-otap/geneva-exporter"]
--- a/rust/otap-dataflow/crates/config/src/lib.rs
+++ b/rust/otap-dataflow/crates/config/src/lib.rs
@@ -23,6 +23,8 @@ pub mod node;
 pub mod observed_state;
 pub mod pipeline;
 pub mod pipeline_group;
+/// TLS configuration.
+pub mod tls;
 pub mod urn;
 
 /// Signal types
--- a/rust/otap-dataflow/crates/config/src/tls.rs
+++ b/rust/otap-dataflow/crates/config/src/tls.rs
@@ -1,0 +1,80 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use std::path::PathBuf;
+use std::time::Duration;
+
+/// Common TLS configuration for both client and server.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, JsonSchema, Default)]
+pub struct TlsConfig {
+    /// Path to the TLS cert to use for TLS required connections.
+    pub cert_file: Option<PathBuf>,
+
+    /// In memory PEM encoded TLS cert to use for TLS required connections.
+    pub cert_pem: Option<String>,
+
+    /// Path to the TLS key to use for TLS required connections.
+    pub key_file: Option<PathBuf>,
+
+    /// In memory PEM encoded TLS key to use for TLS required connections.
+    pub key_pem: Option<String>,
+    // TODO: Implement these fields
+    // /// MinVersion sets the minimum TLS version that is acceptable.
+    // /// If not set, TLS 1.2 will be used.
+    // pub min_version: Option<String>,
+
+    // /// MaxVersion sets the maximum TLS version that is acceptable.
+    // pub max_version: Option<String>,
+
+    // /// CipherSuites is a list of TLS cipher suites that the TLS transport can use.
+    // pub cipher_suites: Option<Vec<String>>,
+
+    // /// Trusted platform module configuration
+    // pub tpm_config: Option<TpmConfig>,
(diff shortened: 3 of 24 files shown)


rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
```diff
@@ -0,0 +1,113 @@
...
+            sdk_resource_builder =
+                sdk_resource_builder.with_attribute(KeyValue::new(k.clone(), v.clone()));
+        }
+        let sdk_resource = sdk_resource_builder.build();
+        sdk_meter_builder = sdk_meter_builder.with_resource(sdk_resource);
+    }
+
+    let sdk_meter_provider = sdk_meter_builder.build();
+
+    global::set_meter_provider(sdk_meter_provider);
```
cijothomas: if we need to flush/shutdown, need to hold on to a clone of the meter_provider.
https://github.com/open-telemetry/opentelemetry-rust/blob/main/examples/metrics-basic/src/main.rs#L20

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
```diff
@@ -0,0 +1,113 @@
...
+    use std::time::Duration;
+
+    #[test]
+    fn test_configure_minimal_otel_sdk() {
+        let config = TelemetryConfig::default();
+        configure_otel_sdk(&config);
+        let meter = global::meter("test-meter");
+
+        let counter = meter.u64_counter("test-counter").build();
+        counter.add(1, &[]);
```
cijothomas, line 105: nothing tested here...?

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
```diff
@@ -0,0 +1,113 @@
...
+            metrics: metrics_config,
+            resource,
+        };
+        configure_otel_sdk(&config);
+        let meter = global::meter("test-meter");
+
+        let counter = meter.u64_counter("test-counter").build();
+        counter.add(1, &[]);
+
+        // Sleep to allow the periodic reader to export metrics
```
cijothomas: flush or shutdown, instead of sleeps in tests.

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -612,11 +622,207 @@ impl Default for PipelineConfigBuilder {
...
+                }
+                Err(serde::de::Error::custom(
+                    "Expected either 'console' or 'otlp' exporter",
+                ))
+            }
+        }
+
+        deserializer.deserialize_map(MetricsPeriodicExporterConfigVisitor)
+    }
+}
```
lalitb, line 801: nit - Is there plan to eventually use the [opentelemetry-config](https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-config) crate once it is published?

andborja: Yes. That's the initial idea.

cijothomas: we can publish it anytime. No need to wait for usual release cycle.

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -612,11 +622,207 @@ impl Default for PipelineConfigBuilder {
...
+
+/// Telemetry backend configuration.
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct TelemetryConfig {
+    /// The size of the reporting channel.
+    #[serde(default = "default_reporting_channel_size")]
+    pub reporting_channel_size: usize,
+    /// The interval at which metrics are flushed and aggregated by the collector.
+    #[serde(default = "default_flush_interval")]
+    pub flush_interval: std::time::Duration,
```
jmacd: This looks out of place. The type and defaults are good.

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -612,11 +622,207 @@ impl Default for PipelineConfigBuilder {
...
+
+/// Opentelemetry Metrics Reader configuration.
+#[derive(Debug, Clone, Serialize, JsonSchema)]
+#[serde(rename_all = "lowercase")]
+pub enum MetricsReaderConfig {
+    /// Periodic reader that exports metrics at regular intervals.
+    Periodic(MetricsReaderPeriodicConfig),
+    /// Pull reader that allows on-demand metric collection.
+    Pull {
+        //TODO: Add specific configuration for supported pull readers.
```
jmacd: Note: The reason I say flush_interval is out of place at the higher level is that Pull metrics readers will not have a flush interval so to speak.

rust/otap-dataflow/crates/config/src/pipeline.rs
jmacd: For this copied section, a few changes I'd like (assuming `use std::time::Duration`):

```suggestion
    /// The size of the reporting channel, measured in the number of internal metric
    /// events shared across all cores.
    #[serde(default = "default_reporting_channel_size")]
    pub reporting_channel_size: usize,
    /// The interval at which metrics are flushed and aggregated by the collector.
    #[serde(with = "humantime_serde", default = "default_reporting_interval")]
    pub reporting_interval: Duration,
```

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -612,11 +622,207 @@ impl Default for PipelineConfigBuilder {
...
+    }
+}
+
+fn default_reporting_channel_size() -> usize {
+    100
+}
+
+fn default_flush_interval() -> std::time::Duration {
+    std::time::Duration::from_secs(1)
+}
```
jmacd:
```suggestion
fn default_reporting_interval() -> Duration {
    Duration::from_secs(1)
}
```

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -612,11 +622,207 @@ impl Default for PipelineConfigBuilder {
...
+
+/// Opentelemetry Metrics Periodic Reader configuration.
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct MetricsReaderPeriodicConfig {
+    /// The metrics exporter to use.
+    pub exporter: MetricsPeriodicExporterConfig,
+    /// The interval (in milliseconds) at which metrics are exported.
+    #[serde(default = "default_interval")]
+    pub interval: u64,
```
jmacd:
```suggestion
    /// The interval at which metrics are periodically exported.
    #[serde(with = "humantime_serde", default = "default_push_interval")]
    pub interval: Duration,
```

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -612,11 +622,207 @@ impl Default for PipelineConfigBuilder {
...
+}
+
+fn default_interval() -> u64 {
+    6000
+}
```
jmacd:
```suggestion
fn default_periodic_interval() -> Duration {
    Duration::from_secs(60)
 }
```

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
jmacd: Is the use of a global meter provider necessary? I would probably avoid this so that flush/shutdown work. What thread/actor performs the periodic export ("flush"), by the way?

rust/otap-dataflow/crates/telemetry/Cargo.toml
```diff
@@ -27,3 +28,6 @@ thiserror = { workspace = true }
 slotmap = { workspace = true }
 parking_lot = { workspace = true }
 serde = { workspace = true }
+opentelemetry_sdk = "0.31.0"
+opentelemetry-stdout = "0.31.0"
+opentelemetry = "0.31.0"
```
jmacd: Please add these in the workspace and use a { workspace = true } value here.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
cijothomas: global is only necessary, if components want to obtain a meter by calling `global::meter`. The alternative is to get meter directly from the `meter_provider`.

In OTel Rust, it creates a single std::thread per provider, which does the periodic exporting part. (And as of today, it's not controllable which core that thread runs. We should be able to add such support in OTel Rust, if its required for Arrow telemetry needs)

rust/otap-dataflow/crates/config/src/pipeline.rs
jmacd, line 11: (Asking the group: what does JsonSchema do for us? It was added a long time ago.)

andborja: It actually added some complexity to the "duration" serialization in serde..

rust/otap-dataflow/crates/config/src/pipeline.rs
andborja: Updated

rust/otap-dataflow/crates/config/src/pipeline.rs
jmacd: I see. We have at least three ways to treat human-readable durations via serde at this time. This will be cleaned up later, thanks.

Issue #501, opened 2025-05-30
[otap-dataflow] parquet exporter support cloud object stores


albertlockett:
In https://github.com/open-telemetry/otel-arrow/pull/488 we added parquet exporter but currently it only supports writing to a local filesystem.

We should add the capability to support remote object stores: S3, GCS and Azure Storage.

We should document which permissions are need for each of these. This can probably live in the Rust docs for now, although it might also make sense in a README somewhere within the repo.

---
Configuration

Probably we can parse the object_store implementation to use from the `base_uri`:
- `s3://bucket-name/`
- `gcs://bucket-name/`
- `az://storage-container/`

For each of these users should be able to provide the path as well .. e.g. `s3://bucket-name/path/in-bucket/`.

Note that for each of these object store implementations, there are many configuration options.
- https://docs.rs/object_store/0.12.0/object_store/aws/struct.AmazonS3Builder.html
- https://docs.rs/object_store/0.12.0/object_store/gcp/struct.GoogleCloudStorageBuilder.html
- https://docs.rs/object_store/0.12.0/object_store/azure/struct.MicrosoftAzureBuilder.html

Although these can be configured through the environment, it would also be nice if users could set these values through config passed directly to configuration of this pipeline node as well. This would allow dynamic configuration of these values w/out modifying the global environment, and it would allow users to have different configurations for different exporter instances w/in the same process.

JakeDern:
I'm thinking about this a little bit because I want to get some azure and better auth support merged for this component, but I think the scope of this may expand past this specific exporter. It probably also matters to think about how plugins will work (in tree vs out of tree), but I'm just ignoring that for now 😅. I think if there were some `contrib` repo or something for this stuff to live in that would work just fine to.

The obvious short path to configuring storage for parquet_exporter is to use the existing `object_store` builders which come with a [configuration setting enum](https://docs.rs/object_store/0.12.0/object_store/azure/enum.AzureConfigKey.html) that has `serde` support and an associated [with_config api](https://docs.rs/object_store/0.12.0/object_store/azure/struct.MicrosoftAzureBuilder.html#method.with_config).  Included in this are all the auth configuration options like `FederatedTokenFile` or `UseAzureCli`, because a while back they [decided not to use SDKs](https://github.com/apache/arrow-rs-object-store/issues/252) anymore.

[...]

PR #1517, opened 2025-12-03
Azure support for parquet exporter and cloud_auth module

JakeDern:
The primary motivation for this PR is adding support for azure backends for the parquet exporter. This work seems potentially re-usable for anything that interacts with object storage and/or anything that needs an azure token credential, so I also did a bit of module shuffling. Major changes:

- Added a `cloud_auth` module with support for an initial set of azure token credential types
- Moved the `object_store` module up and added support for azure token credential types
- Added a new `azure` feature flag that gates all things azure across both `cloud_auth` and `object_store`
- Fixed up tests where appropriate

Something that I did not do at this stage was explore moving all of this auth stuff into some kind of `identity` extension. That might be the next step and I think much of the code here could be easily folded into such an extension.

Related to: https://github.com/open-telemetry/otel-arrow/issues/501

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/cloud_auth/azure.rs
+++ b/rust/otap-dataflow/crates/otap/src/cloud_auth/azure.rs
@@ -187,7 +187,7 @@ mod test {
 
     fn test_auth_method(json: &str, expected: AuthMethod) {
         let method: AuthMethod =
-            serde_json::from_str(&json).expect("Failed to deserialize AuthMethod");
+            serde_json::from_str(json).expect("Failed to deserialize AuthMethod");
         assert_eq!(method, expected);
 
         assert!(from_auth_method(method).is_ok());


rust/otap-dataflow/crates/otap/src/object_store/azure.rs
```diff
@@ -0,0 +1,81 @@
...
+
+#[derive(Debug)]
+struct TokenProviderState {
+    /// The last obtained object store credential
+    current_access_token: AccessToken,
+    /// The object store representation of [Self::current_access_token]
+    current_object_store_cred: Arc<AzureCredential>,
+}
+
+impl AzureTokenCredentialProvider {
```
JakeDern, line 39: This is a general adapter for using any azure `TokenCredential` with the object store's `MicrosoftAzureBuilder`. 

I delegated the caching to the `TokenCredential` because those types cache internally. Most of the machinery here is to avoid allocating a String every time this is called (every request that object_store makes). To do that I needed a cache and internal mutability via a mutex. If it doesn't seem worth it for the complexity, we can chop this implementation down a lot.

rust/otap-dataflow/crates/otap/src/cloud_auth/azure.rs
```diff
@@ -0,0 +1,109 @@
...
+};
+
+use azure_core::credentials::{AccessToken, TokenCredential};
+use azure_identity::AzureCliCredential;
+use object_store::{CredentialProvider, azure::AzureCredential};
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(tag = "type")]
+pub enum AuthMethod {
```
JakeDern, line 16: This module is the general machinery for azure token credential types. It's de-coupled from the object_store provider implementation over in the `object_store` package.

I was originally thinking of putting this code in the `experimental` folder, but that code doesn't seem linked into the crate structure in a way that is usable. Happy to move it wherever make sense

rust/otap-dataflow/configs/fake-parquet.yaml
```diff
@@ -22,7 +22,13 @@ nodes:
...
     plugin_urn: "urn:otel:otap:parquet:exporter"
     config:
-      base_uri: /tmp
+      storage:
+        # file:
+        #   base_uri: /home/jakedern/local-fabric-test
+        azure:
+          base_uri: https://barquettest.blob.core.windows.net/barquet1
+          auth:
+            type: AzureCli
```
albertlockett: Config looks good! This seems like a more extensible way to configure the the storage instead of trying to infer it from the path as suggested in #501

albertlockett: Looks like the `auth` block could be reused in other component configuration too with minimal refactoring, so that's another bonus

JakeDern: Two notes on this!

1. object_store is doing some inference based on the uri still. For example, if I put `abfs://` vs `adl://` vs `az://`, those are different storage backends in azure and there is special handling for those: https://github.com/apache/arrow-rs-object-store/blob/1a9758be2282f85f135d1076a6979af2878c7376/src/azure/builder.rs#L531C3-L558C12
2. The reason I think it makes sense to first differentiate `azure:` /`gcp:`/`aws:`/`file:` is that the auth options are only valid per backend (and are also not relevant at all for file storage!). Differentiating which storage backend you want to target first will let us fail at deserialization if someone tries to do something like add auth to file storage, or configure an aws based auth mechanism for azure.

rust/otap-dataflow/crates/otap/src/cloud_auth/azure.rs
albertlockett: Yeah I think it's probably fine to keep this in otap-dataflow.

A good option could be to keep this in the `otap-dataflow/crates/otap` crate, and create a new module like `otap-df-otap::cloud_auth`.

There's some precident for this -- some of the utilities that have a cross-cutting functionality that are used by multiple components are also in modules in `otap-dataflow/crates/otap/src`. A few examples of this are the `otap_grpc` module which contains some common code used by grpc receivers and exporters. Another is the `accessory` module.

rust/otap-dataflow/configs/fake-parquet.yaml
albertlockett: ohh yes that's right, thanks @JakeDern

rust/otap-dataflow/crates/otap/src/cloud_auth/azure.rs
JakeDern: Yeah, I really like the idea of moving `parquet_exporter::cloud_auth` and even the `parquet_exporter::object_store` module (which now contains adapters for auth) up next to `otap_grpc`.

I think it will at least convey the intent of both to be somewhat generic modules and raise the visibility of them for other contributors. If that sounds good to you, I'll introduce that change.

albertlockett (approved):
This is awesome! Thanks @JakeDern

Issue #1479, opened 2025-11-26
[query-engine] User-defined functions


CodeBlanch:
# KQL Analysis

KQL has a pretty narrow set of features w.r.t. functions:

## Scalar functions:

```
let scalar = (a:int, b:string="default") {
   let c = 1; // Let statement support
   strcat(a, b, c) // Must return a scalar
};
print scalar(1); // output: 1default1
```

## Tabular functions:

```
let tabular1 = (T:(*)) { // Wildcard schema for T
   let c = 1; // Let statement support
   T | extend a = c // Must end in a tabular statement. Multiple tabular statements not supported
};

let tabular2 = (T:(a:int)) { // Defined schema for T
  T | project-keep a
};

source | invoke tabular1(); // Implicit pass of table from current context
tabular2((range a from 1 to 10 step 1)); // Explicit pass of table
```

## Not supported

There doesn't seem to be any kind of control flow in KQL. No if/then/else/for/break/continue type of things.

# Goal with query engine

* Support user-defined functions in query expression tree
  * Inline functions (described fully by expressions)
  * External functions (callbacks registered with engine)
* Support immutable and mutable passing of arguments
* Support returning immutable values
* Support control flow (probably won't be part of initial work but planning for it to be easy to add later)

The goals with query engine are bigger than what KQL can do. In order to have really useful functions control flow is important. But given this feature set it should be possible to express what KQL supports.

# Example

The KQL scalar from above:

```
let scalar = (a:int, b:string="default") {
   let c = 1; // Let statement support
   strcat(a, b, c) // Must return a scalar
};
```

Becomes...

```
Function
 Name=scalar
 ReturnType=string
 Parameters
  a=Immutable(int)
  b=Immutable(string)
 Implementation(Expressions)
  Set(Variable(c), Static(Long(1)))
  Return(Concat(Argument(a), Argument(b), Variable(c))
```

The KQL tabular from above:

```
let tabular1 = (T:(*)) { // Wildcard schema for T
   let c = 1; // Let statement support
   T | extend a = c // Must end in a tabular statement. Multiple tabular statements not supported
};
```

```
Function
 Name=tabular1 
 ReturnType=None
 Parameters
  T=Mutable(Map)
 Implementation(Expressions)
  Set(Variable(c), Static(Long(1)))
  Transform(Set(Argument(T), Variable(c))
```

The KQL tabular with schema...

```
let tabular2 = (T:(a:int)) { // Defined schema for T
[...]

PR #1522, opened 2025-12-04
[query-engine] Implement argument scalar in recordset engine

CodeBlanch:
Relates to #1479

## Changes

* Implements `ScalarExpression::Argument` in RecordSet engine

### DIFF
--- a/rust/experimental/query_engine/engine-recordset/src/scalars/scalar_expressions.rs
+++ b/rust/experimental/query_engine/engine-recordset/src/scalars/scalar_expressions.rs
@@ -35,24 +35,7 @@ where
 {
     let value = match scalar_expression {
         ScalarExpression::Source(s) => {
-            if let Some(record) = execution_context.get_record() {
-                let mut selectors = s.get_value_accessor().get_selectors().iter();
-
-                select_from_borrowed_value(
-                    execution_context,
-                    BorrowSource::Source,
-                    record.borrow(),
-                    scalar_expression,
-                    &mut selectors,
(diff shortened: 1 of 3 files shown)


rust/experimental/query_engine/engine-recordset/src/scalars/scalar_expressions.rs
```diff
@@ -35,24 +35,7 @@ where
...
-                )?
-            } else {
-                execution_context.add_diagnostic_if_enabled(
-                    RecordSetEngineDiagnosticLevel::Warn,
-                    scalar_expression,
-                    || "Source could not be found".into(),
-                );
-                ResolvedValue::Computed(OwnedValue::Null)
-            }
+            execute_source_scalar_expression(execution_context, scalar_expression, s)?
```
jmacd, line 38: As a reader, how should I read "source" in this function name? I see the other name in this position is "variable", maybe "source" means the expression is static, as in the source code.

jmacd (approved):
I see the connection to #1479.

rust/experimental/query_engine/engine-recordset/src/scalars/scalar_expressions.rs
CodeBlanch: "source" essentially means the current record being processed. Could benefit from a rename 😄

In comes from this type of thing:

```
source
| project-away source.Attributes['some_key']
```

At the moment "source" is how you refer to the record when you want to fully-qualify access.

PR #1523, opened 2025-12-04
Columnar query engine add short-circuit optimization to composite and/or filter execs

### LINKED ISSUE #1506: Columnar query engine optimization for composite filtering
There are a few places in the columnar query engine code where we execute both sides of some logical `and`/`or` filtering clause and then `and`/`or` the respective results together:
https://github.com/open-telemetry/otel-arrow/blob/f9aec883865d4a3bff625554bcf6d04e1737559e/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs#L497-L506
https://github.com/open-telemetry/otel-arrow/blob/f9aec883865d4a3bff625554bcf6d04e1737559e/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs#L585-L603

However, there are some shortcuts we can take to avoid executing one side. If it's an `and` and the left side returns all `false`, we don't need to execute the right. Similarly if it's an `or` and the left side returns all `true`, we don't need to execute the right.

We should make these optimizations

albertlockett:
closes #1506 

Adds an optimization to the columnar query engine's filtering code where we can short-circuit the execution of and/or filters if depending on the result of one side of the evaluation. For example:
- `x1 == y1 AND x2 == y2`, if `x1 == y1` matches zero rows, there is no point in checking `x2 == y2`
- similarly `x1 == y1 OR x2 == y2`, if `x1 == y1` matches every row, , there is no point checking `x2 == y2`

This adds this optimization to `Composite<FilterExec>` for the `and` and `or` case, and to `Composite<AttributeFilterExec>` for the `and` case only. 

The `or` case for attributes isn't done b/c we would need to check if the predicate passes for every unique parent_id, which is so straight forward so it's not clear if in some cases would introduce more overhead than just executing the other side of the filter. Requires further investigation.

Selected benchmark results showing the performance improvement of this short-circuit optimzation:
```
and_attrs_short_circuit/batch_size/32
                        time:   [1.8433 µs 1.8522 µs 1.8600 µs]
                        change: [−49.152% −48.864% −48.549%] (p = 0.00 < 0.05)
                        Performance has improved.
and_attrs_short_circuit/batch_size/1024
                        time:   [4.5644 µs 4.5717 µs 4.5786 µs]
                        change: [−68.020% −67.941% −67.864%] (p = 0.00 < 0.05)
                        Performance has improved.
and_attrs_short_circuit/batch_size/8192
                        time:   [24.917 µs 24.956 µs 24.994 µs]
                        change: [−74.971% −74.895% −74.827%] (p = 0.00 < 0.05)
                        Performance has improved.

and_short_circuit/batch_size/32
                        time:   [824.82 ns 830.06 ns 835.07 ns]
                        change: [−71.132% −70.962% −70.795%] (p = 0.00 < 0.05)
                        Performance has improved.
and_short_circuit/batch_size/1024
                        time:   [1.3861 µs 1.3899 µs 1.3933 µs]
                        change: [−90.697% −90.664% −90.632%] (p = 0.00 < 0.05)
                        Performance has improved.
and_short_circuit/batch_size/8192
                        time:   [5.4423 µs 5.4488 µs 5.4551 µs]
                        change: [−95.432% −95.416% −95.401%] (p = 0.00 < 0.05)
                        Performance has improved.


or_short_circuit/batch_size/32
                        time:   [2.3195 µs 2.3265 µs 2.3330 µs]
                        change: [−44.904% −44.594% −44.270%] (p = 0.00 < 0.05)
                        Performance has improved.
or_short_circuit/batch_size/1024
                        time:   [23.992 µs 24.059 µs 24.124 µs]
                        change: [−18.687% −18.371% −18.042%] (p = 0.00 < 0.05)
                        Performance has improved.
or_short_circuit/batch_size/8192
                        time:   [127.00 µs 127.23 µs 127.49 µs]
                        change: [−22.200% −21.999% −21.791%] (p = 0.00 < 0.05)
                        Performance has improved.
  ```

As part of the implementation this PR also introduces an `IdMask` num type as the return value from `AttributeFilterExec::execute` method, with explicit variants for selecting `All`, `None`, `Some` IDs and a `NotSome` representing IDs not to select. The reasons for this are:
- it's a clear signal to the caller when `All` or `None` IDs should be selected (an empty or full RoaringBitmap could have been used, but [allocating a full roaring bitmap is expensive](https://docs.rs/roaring/latest/src/roaring/bitmap/inherent.rs.html#36))
- the `NotSome` variant allows us to  invert the selection of IDs without having to do a 2nd pass over the parent ID column (like we were doing [here](https://github.com/open-telemetry/otel-arrow/blob/f3080ac9bfc5673cf4fc8ac24776fd096f353546/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs#L620-L626))

I will probably do a followup PR that tries to take better advantage of this `IdMask` to further optimize filter performance for attribute filtering.

### DIFF
--- a/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
+++ b/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
@@ -612,7 +612,7 @@ impl Composite<FilterExec> {
 }
 
 /// This represents which IDs have been selected by some filter operation.
-/// 
+///
 /// For example it can be used as the return type from filtering attributes to represent
 /// values from the parent_id column matched some filter that was applied.
 #[derive(Clone, Debug, PartialEq)]

### CODE REVIEW

File: rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs [hunk 1]
```diff
@@ -468,14 +478,14 @@ impl FilterExec {
             }
         };
 
-        let mut selection_vec = match self.predicate.as_mut() {
-            Some(predicate) => predicate.evaluate_filter(root_rb, session_ctx)?,
-
-            // TODO -- we might be able to optimize this method by not allocating this here
-            // and instead returning None
```
albertlockett, line 475: I also cleaned up this TODO. It didn't have a huge impact on performance, but I guess it's still just good practice not to allocate this BooleanArray when we don't need it.

PR #1528, opened 2025-12-04
Add benchmark overview doc

cijothomas:
This docs is an attempt at the schema for our Phase 2 performance summary, when phase 2 is completed. It defines the key scenarios (Idle, 100k Load, Saturation) and the comparative analysis with OTLP/Collector. I've put TBD for actual numbers, as this is just attempting to finalize what we want to have in an easy to consume format. Actual numbers will be filled in later. This can also be used to see if there are gaps in the perf test suites that we want to add.

The existing pages like https://open-telemetry.github.io/otel-arrow/benchmarks/nightly/backpressure/ are still retained. This doc will have distilled information from them.


docs/benchmarks-summary.md
```diff
@@ -0,0 +1,250 @@
...
+following specifications:
+
+- **CPU**: 64 cores (x86-64 architecture)
+- **Memory**: 512 GB RAM
+- **Platform**: Oracle Bare Metal Instance
+- **OS**: Oracle Linux 8
+
+This consistent, high-performance environment ensures reproducible results and
+allows for comprehensive testing across various CPU core configurations (1, 4,
+and 8 cores etc.) by constraining the df-engine to specific core allocations.
```
gouslu: One problem with this is that, if there are any components that spawn background threads, which is possible to do, then this could not truly be contained to the specific number of cores. In order to truly constraint the number of cores we need to run the program inside a constrained environment, for example a container or different sized machine if you have to do it bare metal.

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,250 @@
...
+
+### Performance Metrics
+
+#### Idle State Performance
+
+Baseline resource consumption with no active telemetry traffic.
```
gouslu: baseline resource consumption may change over time if there are any, for example, memory leaks. I know load testing is not necessarily soak testing but clarifying what kind of baseline would be good here. Is it the initial baseline? Baseline over time with periodic breaks, etc.?

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,250 @@
...
+| Memory Usage | TBD |
+
+These baseline metrics validate that the engine maintains minimal resource
+footprint when idle, ensuring efficient operation in environments with variable
+telemetry loads.
+
+#### Standard Load Performance
+
+Resource utilization at 100,000 log records per second (100K logs/sec). Tests
+are conducted with three different batch sizes to demonstrate the impact of
```
gouslu: nit: I see 4 different batch sizes below

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,250 @@
...
+
+**Test Parameters:**
+
+- Total input load: 100,000 log records/second
+- Average log record size: 1 KB
+- Batch sizes tested: 10, 100, 1000, and 10000 records per request
+
+This wide range of batch sizes evaluates performance across diverse deployment
+scenarios. Small batches (10-100) represent edge collectors or real-time
+streaming requirements, while large batches (1000-10000) represent gateway
```
gouslu: do we have any hard limitations or guidelines on batch sizes? if not, maybe we can use this to create guidelines.

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,250 @@
...
+The dataflow engine (df-engine), implemented in Rust, provides predictable
+performance characteristics and efficient resource utilization across varying
+load conditions. This columnar approach is expected to offer substantial
+advantages over traditional row-oriented telemetry pipelines in terms of CPU
+efficiency, memory usage, and throughput.
+
+This document presents key performance metrics across different load scenarios
+and test configurations.
+
+### Test Environment
```
gouslu, line 72: where is data generated and how is it delivered to the collector? is the collector running in bare metal and we are sending the logs from somewhere outside over public/private network?

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,250 @@
...
+
+For detailed technical documentation, see the [OTAP Dataflow Engine
+Documentation](../rust/otap-dataflow/README.md) and [Phase 2
+Design](phase2-design.md).
+
+---
+
+## Comparative Analysis: OTel Arrow vs OpenTelemetry Collector
+
+### Methodology
```
gouslu: how is the EPS measurement done? at the source? at the target? at the exporter via metrics? where exactly?

lquerel (changes requested):
I really like this document.

I think we should also include the OTLP to OTLP scenario in the different sections since it will be one of the most common scenarios, at least in the beginning.

I also think we should add the wait_for_result mode in the otel-arrow section because it provides a true end to end unified ack/nack mechanism, which I believe is not fully supported by the Go collector.

docs/benchmarks-summary.md
cijothomas: everything in bare metal (there is a single machine) and communicates via local network.

docs/benchmarks-summary.md
cijothomas: if components misbehave by creating threads randomly, then yes. We should not have such components in this repo. (there exists some temporarily like OTel sdk. But I dont think we plan to keep them that way for long).

We can use docker limits to restrict the cores too. (We do this for OTel Collector today as that is the only way to restrict it)

docs/benchmarks-summary.md
cijothomas: good point, will try to clarify more.

docs/benchmarks-summary.md
cijothomas: not to my knowledge! Yes, we know the impact of higher batch size, and this doc can cover that so users know.

docs/benchmarks-summary.md
cijothomas: both at source (load-generator) and destination (fake-endpoint). There is a metric we track called "logs lost%", to count for missing logs in the pipeline.

albertlockett (approved):

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,272 @@
...
+Baseline resource consumption with no active telemetry traffic, measured after
+startup stabilization over a 60-second period. These metrics represent
+post-initialization idle state and validate minimal resource footprint. Note
+that longer-duration soak testing for memory leak detection is outside the scope
+of this benchmark summary.
+
+| Configuration | CPU Usage | Memory Usage |
+|---------------|-----------|--------------|
+| Single Core   | 0.06%     | 27 MB        |
+| All Cores (128) | 2.5%    | 600 MB       |
```
reyang: Two things to consider:
1. What's the memory usage on different CPU architecture (ARM64, AMD64, etc.)?
2. What's the trend as the number of cores increase? I guess it is C + N * R where C is a constant, N is the number of cores and R is the per core memory?

cijothomas: for 2, added benchmarks test to confirm it.

```txt
Found 6 idle state result directories
Found: 16 core(s) -> 97.08 MiB
Found: 1 core(s) -> 27.60 MiB
Found: 2 core(s) -> 29.52 MiB
Found: 32 core(s) -> 161.8 MiB
Found: 4 core(s) -> 38.14 MiB
Found: 8 core(s) -> 57.16 MiB
Benchmark JSON written to: /home/opc/actions-runner/_work/otel-arrow/otel-arrow/tools/pipeline_perf_test/results/idle-memory-scaling.json

================================================================================
IDLE STATE MEMORY SCALING ANALYSIS
================================================================================

Goal: Verify linear memory scaling (Memory = C + N × R)

  C (Constant Overhead):   22.15 MiB
  R (Per-Core Overhead):   4.420 MiB/core
  R² (Fit Quality):        0.9980

Formula: Memory (MiB) ≈ 22.1 + 4.42 × N

--------------------------------------------------------------------------------
Cores    Actual (MiB)       Predicted (MiB)    Error (%)    Status    
--------------------------------------------------------------------------------
1        27.60              26.57              3.8%         ✅
2        29.52              30.99              5.0%         ✅
4        38.14              39.83              4.4%         ✅
8        57.16              57.51              0.6%         ✅
16       97.08              92.87              4.3%         ✅
32       161.8              163.6              1.1%         ✅
--------------------------------------------------------------------------------

SUMMARY:
  • Each additional core adds ~4.420 MiB of memory overhead
  • Base memory footprint (shared infrastructure): ~22.15 MiB
  • Memory range: 27.60 MiB (1 core) → 161.8 MiB (32 cores)

✅ EXCELLENT: Near-perfect linear fit (R² ≥ 0.99).
   Memory scaling follows the share-nothing model precisely.
```

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,272 @@
...
+significant network bandwidth savings. Phase 2 expands this foundation by
+implementing the entire in-process pipeline using Apache Arrow's columnar
+format, targeting substantial improvements in data processing efficiency while
+maintaining the network efficiency gains from Phase 1.
+
+The OTel Arrow dataflow engine, implemented in Rust, provides predictable
+performance characteristics and efficient resource utilization across varying
+load conditions. The engine uses a [thread-per-core
+architecture](#thread-per-core-design) where resource consumption scales with
+the number of configured cores.
```
reyang: > resource consumption scales with the number of configured cores

I found this a bit hard to interpret. Do you mean "the throughput scales with the number of configured CPU cores, almost in a linear fashion?"

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,272 @@
...
+performance tests (continuous, nightly, saturation, idle state, and binary size
+benchmarks), see [Detailed Benchmark
+Results](benchmarks.md#current-performance-results).
+
+### Test Environment
+
+All performance tests are executed on bare-metal compute instance with the
+following specifications:
+
+- **CPU**: 64 physical cores / 128 logical cores (x86-64 architecture)
```
reyang: Consider calling out the number of NUMA groups?

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,272 @@
...
+
+This document presents a curated set of key performance metrics across different
+load scenarios and test configurations. For the complete set of automated
+performance tests (continuous, nightly, saturation, idle state, and binary size
+benchmarks), see [Detailed Benchmark
+Results](benchmarks.md#current-performance-results).
+
+### Test Environment
+
+All performance tests are executed on bare-metal compute instance with the
```
reyang:
```suggestion
All performance tests are executed on a dedicated bare-metal compute instance with the
```

reyang: Not sure if this is a shared resource or dedicated, consider calling it out.

cijothomas: it is dedicated. Will update

docs/benchmarks-summary.md
reyang:
```suggestion
| Configuration   | CPU Usage | Memory Usage |
| --------------- | --------- | ------------ |
| Single Core     | 0.06%     | 27 MB        |
| All Cores (128) | 2.5%      | 600 MB       |
```

docs/benchmarks-summary.md
cijothomas: ya. We just confirmed that the CNCF machine has 2 sockets
https://github.com/open-telemetry/otel-arrow/actions/runs/23278418373#summary-67686308469

So far no tests were run with engine running on more than 32 cores (and they were all in same node). I have to see if we can actually do that, given load-gen and fake-backend also needs cores to run on.

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,272 @@
...
+of this benchmark summary.
+
+| Configuration   | CPU Usage | Memory Usage |
+| --------------- | --------- | ------------ |
+| Single Core     | 0.06%     | 27 MB        |
+| All Cores (128) | 2.5%      | 600 MB       |
+
+*Note: CPU usage is normalized (percentage of total system capacity). Memory
+usage scales with core count due to the [thread-per-core
+architecture](#thread-per-core-design).*
```
reyang: Memory usage could be confusing (cached, shared, non-paged pool, virtual memory vs. physical memory, ...), consider aligning and pointing to the OTel System metrics semantic conventions https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md.

docs/benchmarks-summary.md
cijothomas: ya it reads weird.. I'll update with a better wording
(I meant throughput scales linearly, but so does memory consumption)

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,272 @@
...
+| 10/batch | TBD | TBD | TBD | TBD |
+| 100/batch | TBD | TBD | TBD | TBD |
+| 1000/batch | 17% | 46 MB | 718 KB/s | 750 KB/s |
+| 5000/batch | 7% | 50 MB | 390 KB/s | 422 KB/s |
+| 10000/batch | 5% | 58 MB | 350 KB/s | 383 KB/s |
+
+This represents the optimal scenario where the dataflow engine operates with its
+native protocol end-to-end, eliminating protocol conversion overhead.
+
+##### Standard Load - OTLP -> OTLP (Standard Protocol)
```
reyang: Which OTLP? (gRPC, proto via HTTP 1.1, JSON, TLS enabled vs. not)

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Comparative Analysis: OTel Arrow vs OpenTelemetry Collector
+
+### Methodology
+
+To provide a fair and meaningful comparison between the OTel Arrow dataflow
+engine and the OpenTelemetry Collector, we use **Syslog (UDP/TCP)** as the
+ingress protocol for both systems.
+
+#### Rationale for Syslog-Based Comparison
```
reyang:
```suggestion
#### Rationale for Syslog-based Comparison
```

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,272 @@
...
+| 4 Cores   | ~653K logs/sec | ~70%*     | 69%               | ~80 MB       |
+| 8 Cores   | ~1.47M logs/sec | ~82%*    | 78%               | ~178 MB      |
+| 16 Cores  | ~2.37M logs/sec | ~78%*    | 62%               | ~288 MB      |
+| 24 Cores  | ~3.67M logs/sec | ~77%*    | 64%               | ~461 MB      |
+
+TODO: Use more load-generator so saturate CPU at higher cores.
+
+Scaling Efficiency = (Throughput at N cores) / (N * Single-core throughput)
+
+### Architecture
```
reyang: It is a bit weird to have an architecture section in the benchmark document (unless it is talking about the benchmarking environment's own architecture).

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,272 @@
...
+#### Standard Load Performance (Single Core)
+
+Resource utilization at 100,000 log records per second (100K logs/sec) on a
+single CPU core. Tests are conducted with four different batch sizes to
+demonstrate the impact of batching on performance.
+
+**Test Parameters:**
+
+- Total input load: 100,000 log records/second
+- Average log record size: 1 KB
```
cijothomas: @noahfalk shared a feedback about the unrealistic compression we are getting: 100K logs/sec, each with 1KB, but still, network output is 1-2 MB/sec, so achieving 50 compression ratio.

https://github.com/open-telemetry/otel-arrow/issues/2540 is tracking addressing this unrealism.

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,189 @@
...
+post-initialization idle state and validate minimal resource footprint. Note
+that longer-duration soak testing for memory leak detection is outside the scope
+of this benchmark summary.
+
+| Configuration   | CPU Usage | Memory Usage |
+| --------------- | --------- | ------------ |
+| Single Core     | 0.1%      | 27 MB        |
+| All Cores (128) | 2.5%      | 600 MB       |
+
+*Note: CPU usage is normalized (percentage of total system capacity). Memory
```
reyang: Instead of using and explaining the CPU usage, consider:
1. Call it "CPU utilization"
2. Use the same semantic and unit as https://github.com/open-telemetry/semantic-conventions/blob/main/docs/system/system-metrics.md#metric-systemcpuutilization (pick the process utilization rather than the system one)
3. Provide a link to the semantic convention, so we don't have to explain it here

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,189 @@
...
+performance tests (continuous, nightly, saturation, idle state, and binary size
+benchmarks), see [Detailed Benchmark
+Results](benchmarks.md#current-performance-results).
+
+### Test Environment
+
+All performance tests are executed on a dedicated bare-metal compute instance with the
+following specifications:
+
+- **CPU**: 64 physical cores / 128 logical cores (x86-64, 2 NUMA nodes)
```
reyang:
```suggestion
- **CPU**: 64 physical cores / 128 logical cores (x86-64, 2 [NUMA](https://en.wikipedia.org/wiki/Non-uniform_memory_access) nodes)
```

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,189 @@
...
+#### Standard Load Performance (Single Core)
+
+Resource utilization at 100,000 log records per second (100K logs/sec) on a
+single CPU core. Tests are conducted with different batch sizes to demonstrate
+the impact of batching on performance.
+
+**Test Parameters:**
+
+- Total input load: 100,000 log records/second
+- Average log record size: 1 KB
```
reyang: An alternative way based on what we've discussed in the meeting:

* ingress: xyz bytes/second (log records / second, log record size distribution)
* egress: xyz bytes/second (batch size, etc.)
* test duration

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,189 @@
...
+*Note: Only 512/batch has been tested for OTAP at standard load so far. OTAP
+batch size variants for small batches are blocked by a known drain-timeout issue
+and will be added once resolved.*
+
+This represents the optimal scenario where the dataflow engine operates with its
+native protocol end-to-end, eliminating protocol conversion overhead.
+
+##### Standard Load - OTLP -> OTLP (Standard Protocol)
+
+*OTLP is gRPC with Protobuf encoding, no TLS.*
```
reyang: link to https://opentelemetry.io/docs/specs/otlp/#otlpgrpc

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,189 @@
...
+| 1 Core    | ~238K logs/sec | ~99%      | 100% (baseline)   | ~38 MB       |
+| 2 Cores   | ~414K logs/sec | ~91%      | 87%               | ~55 MB       |
+| 4 Cores   | ~653K logs/sec | ~70%*     | 69%               | ~80 MB       |
+| 8 Cores   | ~1.47M logs/sec | ~82%*    | 78%               | ~178 MB      |
+| 16 Cores  | ~2.37M logs/sec | ~78%*    | 62%               | ~288 MB      |
+| 24 Cores  | ~3.67M logs/sec | ~77%*    | 64%               | ~461 MB      |
+
+*\* At higher core counts, CPU utilization drops below 100% because the
+load-generator cannot fully saturate all engine cores. Adding more load-generator
+cores would increase both throughput and CPU utilization.*
```
reyang: Is the plan to publish this as-is, or to fix the load generator throughput issue?

cijothomas: We will fix the issue before merging. Almost there! 
https://github.com/open-telemetry/otel-arrow/pull/2997 fixed a lot more underlying issues and we now have near-linear scaling for upto 16 cores. Ran out of cores to test more :(
Each SUT needs multiple load-gen cores to saturate it.  I'll try more complex payloads so we can use less loadgen cores to saturate a SUT.

jmacd (approved):

docs/benchmarks-summary.md
```diff
@@ -0,0 +1,241 @@
...
+#### Saturation Performance (Single Core)
+
+Maximum throughput a single CPU core can sustain at full utilization. This
+establishes the baseline "unit of capacity" for capacity planning and exposes
+the raw efficiency of the engine and the wire protocol.
+
+**Test Parameters:**
+
+- Payload: realistic ~300-byte log records (the OpenTelemetry semantic
+  conventions data shape), identical to the standard-load tests above for
```
ThomsonTan, line 160: This 300-bytes log records is inconsistent the the above standard-load tests which says average 1KB record size.

cijothomas: good catch. I'll fix the consistency issues.

PR #1529, opened 2025-12-04
Run Rust CI only if rust directory is modified

cijothomas:
Avoid running CI checks unnecessarily. Need to validate if this causes any issue with "required check". Hopefully not, as this is a pattern used in other OTel repos too.

Later to be expanded to Go also.

### DIFF
--- a/rust/experimental/query_engine/Cargo.toml
+++ b/rust/experimental/query_engine/Cargo.toml
@@ -26,7 +26,7 @@ datafusion = { version = "51.0.0", default-features = false }
 bytes = "1.10.1"
 caseless = "0.2.2"
 chrono = "0.4.41"
-criterion = "0.7.0"
+criterion = "0.8.0"
 futures-core = "0.3"
 hex = "0.4.3"
 opentelemetry-proto = "0.31.0"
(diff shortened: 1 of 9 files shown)


drewrelmas:
Unfortunately, my intuition is that this will break due to required checks. Do you have an example from other opentelemetry repos showing that is not the case?

cijothomas:
> Unfortunately, my intuition is that this will break due to required checks. Do you have an example from other opentelemetry repos showing that is not the case?

It should not break. https://github.com/open-telemetry/opentelemetry-dotnet/pull/5023 is when we tried this in .NET by @CodeBlanch !

jmacd:
(We'll see if it breaks!)

albertlockett (approved):

drewrelmas:
@cijothomas The Rust-CI workflow fails in merge queue on the new `check-changes` job: https://github.com/open-telemetry/otel-arrow/actions/runs/20029073546/job/57433396112

cijothomas:
> @cijothomas The Rust-CI workflow fails in merge queue on the new `check-changes` job: https://github.com/open-telemetry/otel-arrow/actions/runs/20029073546/job/57433396112

Pushed a fix. If this does not work, then I'll come back to this a bit later. Please add this to merge-queue once more.

.github/workflows/rust-ci.yml
```diff
@@ -20,7 +20,25 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  check-changes:
+    runs-on: ubuntu-latest
+    outputs:
+      rust: ${{ steps.filter.outputs.rust }}
```
CodeBlanch, line 26: @cijothomas It has been a while since I worked on this in .NET repo but the issue may be here. What the check-changes outputs gives you is an array of strings IIRC.

Try changing this line to like...

```yaml
rust: ${{ contains(steps.filter.outputs.changes, 'rust') }}
```

PR #1537, opened 2025-12-05
[otap-df-quiver] Add Write-Ahead-Log (WAL) implementation to Quiver

AaronRM:
This pull request introduces an implementation of the Quiver write-ahead log (WAL). The most significant change from the initial spec includes a rewrite and clarification of the WAL file rotation and checkpointing mechanism. Documentation has been updated to reflect the new design.

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/wal/writer.rs
+++ b/rust/otap-dataflow/crates/quiver/src/wal/writer.rs
@@ -464,7 +464,11 @@ impl WalWriter {
     ///
     /// Writes the slot header (id, fingerprint, row_count, payload_len) followed by
     /// the Arrow IPC-encoded payload bytes.
-    fn encode_slot_into_buffer(&mut self, slot_id: SlotId, payload: PayloadRef<'_>) -> WalResult<()> {
+    fn encode_slot_into_buffer(
+        &mut self,
+        slot_id: SlotId,
+        payload: PayloadRef<'_>,
+    ) -> WalResult<()> {
         let row_count = u32::try_from(payload.batch.num_rows())
             .map_err(|_| WalError::RowCountOverflow(payload.batch.num_rows()))?;
         let payload_bytes = encode_record_batch(payload.batch)?;


rust/otap-dataflow/deny.toml
```diff
@@ -103,6 +103,7 @@ allow = [
     "MIT-0",
     "Apache-2.0",
     "Unicode-3.0",
+    "BSD-2-Clause",
```
AaronRM, line 106: Required due to the `array-ref` transitive dependency (from `blake3`) using this license. `BSD-2-Clause` is even more permissive than `BSD-3-Clause` (which is already allowed here), so I don't believe this should be a concern.

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
```diff
@@ -0,0 +1,1174 @@
...
+
+        let mut writer = Self {
+            active_file,
+            coordinator,
+            next_sequence: 0,
+            #[cfg(test)]
+            test_crashed: false,
+        };
+        writer.next_sequence = writer.coordinator.detect_next_sequence()?;
+        Ok(writer)
```
lalitb: Is this an issue if the collector crashes mid-write?  The WAL file could have a partial/corrupt record at the end. After restart, `WalWriter::open()` will seek to physical EOF, and subsequent writes happening after corrupted data could result in:

```
Before crash:  [Batch 1] [Batch 2] [Batch 3] [partial...]
After restart: [Batch 1] [Batch 2] [Batch 3] [garbage] [Batch 4] [Batch 5]
```

Maybe this is more theoretical. Either way, would be good to have a test that simulates a partial write and verifies recovery behavior (whether that's truncation, detection, or something else).

rust/otap-dataflow/crates/quiver/src/wal/reader.rs
```diff
@@ -0,0 +1,456 @@
...
+            }
+            Ok(ReadStatus::Filled) => {}
+            Err(err) => {
+                self.finished = true;
+                return Some(Err(err));
+            }
+        }
+
+        let entry_len = u32::from_le_bytes(len_buf) as usize;
+        self.buffer.resize(entry_len, 0);
```
lalitb, line 177: Here, the reader is trusting the length got from the file, and doing the allocation. In case of the WAL file is corrupted/or malicious attack, the length is large enough ( say 0xFFFF..), the reader will try allocate that size. While the 4 bytes will limit the allocation to 4GB, all df_instance doing this allocation can result in OOM crash. Should we have some kind of max limit check (say WAL size won't be more than 64MB) ?

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -52,14 +67,50 @@ impl QuiverEngine {
...
+    };
+    let options = WalWriterOptions::new(wal_path, segment_cfg_hash(config), flush_policy)
+        .with_max_wal_size(config.wal.max_size_bytes.get())
+        .with_max_rotated_files(config.wal.max_rotated_files as usize)
+        .with_rotation_target(config.wal.rotation_target_bytes.get());
+    Ok(WalWriter::open(options)?)
+}
+
+fn wal_path(config: &QuiverConfig) -> PathBuf {
+    config.data_dir.join("wal").join("quiver.wal")
```
lalitb, line 86: Just confirming - each df_engine instance has a separate `data_dir`, right? Otherwise multiple instances would conflict on this path.

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
```diff
@@ -0,0 +1,1174 @@
...
+
+impl WalWriterOptions {
+    pub fn new(path: PathBuf, segment_cfg_hash: [u8; 16], flush_policy: FlushPolicy) -> Self {
+        Self {
+            path,
+            segment_cfg_hash,
+            flush_policy,
+            max_wal_size: u64::MAX,
+            max_rotated_files: 8,
+            rotation_target_bytes: 64 * 1024 * 1024,
```
lalitb: Both above two magic numbers seems to be important operational defaults that affect capacity behavior. Good to define them as named constants to make them more discoverable along with doc comments to explain the rational behind using these default values.

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
```diff
@@ -0,0 +1,1174 @@
...
+    pub rotation_target_bytes: u64,
+}
+
+impl WalWriterOptions {
+    pub fn new(path: PathBuf, segment_cfg_hash: [u8; 16], flush_policy: FlushPolicy) -> Self {
+        Self {
+            path,
+            segment_cfg_hash,
+            flush_policy,
+            max_wal_size: u64::MAX,
```
lalitb: More of a question - what will happen if we have reached the limit of 8 wal rotated files of 64MB each, while this max_wal_size limit is still not reached?

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
```diff
@@ -0,0 +1,1174 @@
...
+        for rotated_file in &self.rotated_files {
+            if let Some(seq) = self.scan_file_last_sequence(&rotated_file.path)? {
+                highest = Some(match highest {
+                    Some(current) => current.max(seq),
+                    None => seq,
+                });
+            }
+        }
+        Ok(highest.map_or(0, |seq| seq.wrapping_add(1)))
+    }
```
lalitb, line 908: `detect_next_sequence()` scans all entries in all WAL files on startup. For the default ( 64MB * 8 = 512MB ) capacity this is fine, but would it make sense to persist the last sequence in the checkpoint sidecar for faster recovery?

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
AaronRM: Yes, good catch and not theoretical! Detection of partial/corrupt records like this should be happening during `WalWriter::open()`. I will address (and ensure we have tests to simulate such cases).

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
AaronRM: Better yet, `detect_next_sequence()` doesn't need to scan all the rotated WAL files. It should only be looking at the active file (or the single most recent rotated file if the active file doesn't exist). Measuring this locally and time is on the order of ~30ms with a 64MB file, so doesn't seem worth putting in the sidecar.

rust/otap-dataflow/crates/quiver/src/engine.rs
AaronRM: Yes. There is a single `WalWriter` per `QuiverEngine`. Each `QuiverEngine` would have its own `data_dir`.

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -52,14 +67,50 @@ impl QuiverEngine {
...
+}
+
+fn segment_cfg_hash(config: &QuiverConfig) -> [u8; 16] {
+    // Placeholder fingerprint derived from segment configuration; later phases will
+    // mix in adapter-specific layout metadata once available.
+    let mut hasher = Hasher::new();
+    let _ = hasher.update(&config.segment.target_size_bytes.get().to_le_bytes());
+    let _ = hasher.update(&config.segment.max_stream_count.to_le_bytes());
+    let _ = hasher.update(&config.segment.max_open_duration.as_nanos().to_le_bytes());
```
albertlockett: Are these not the operational knobs that we consider out of scope for the segment config hash?

https://github.com/open-telemetry/otel-arrow/blob/d6096d2435f5f30d3425ef8cec0cdeca22949c62/rust/otap-dataflow/crates/quiver/ARCHITECTURE.md?plain=1#L99-L102

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
AaronRM: There are two different size enforcement mechanisms:

1. `rotation_target_bytes` x `max_rotated_files`
2. `max_wal_size`

If we reach the limit for (1), then [we return a `WalAtCapacity` error](https://github.com/AaronRM/otel-arrow/blob/d6096d2435f5f30d3425ef8cec0cdeca22949c62/rust/otap-dataflow/crates/quiver/src/wal/writer.rs#L646) and (2) is not enforced.  (1) is for controlling rotation behavior and limiting the amount of file descriptors used. (2) is for global size enforcement.  (We could potentially simplify and eliminate (2), but will leave in for now.)

See tests: `wal_writer_errors_when_rotated_file_cap_reached` and `wal_writer_preflight_rejects_when_rotated_file_cap_hit`.

rust/otap-dataflow/crates/quiver/src/wal/checkpoint_sidecar.rs
```diff
@@ -0,0 +1,211 @@
...
+            file.write_all(&encoded)?;
+            file.flush()?;
+            file.sync_data()?;
+        }
+        #[cfg(test)]
+        if writer_test_support::take_crash(CrashInjection::BeforeSidecarRename) {
+            return Err(WalError::InjectedCrash(
+                "crash injected before checkpoint sidecar rename",
+            ));
+        }
```
albertlockett, line 143: 🤩 this is a cool way to test this

rust/otap-dataflow/crates/quiver/src/engine.rs
AaronRM: Yep, that's correct. The hash should be determined by the OTAP adapter layer (which doesn't exist yet), so will use a hardcoded placeholder value for now and revisit later.

rust/otap-dataflow/crates/quiver/src/wal/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+pub(crate) use reader::{DecodedWalSlot, WalConsumerCheckpoint, WalReader, WalRecordBundle};
+// Writer is used broadly soon; suppress warnings while integration lands.
+#[allow(unused_imports)]
+pub(crate) use writer::{FlushPolicy, WalOffset, WalWriter, WalWriterOptions};
+
+pub(crate) const WAL_MAGIC: &[u8; 10] = b"QUIVER\0WAL";
+pub(crate) const ENTRY_TYPE_RECORD_BUNDLE: u8 = 0;
+pub(crate) const ENTRY_HEADER_LEN: usize = 1 + 8 + 8 + 8;
+pub(crate) const SCHEMA_FINGERPRINT_LEN: usize = 32;
+pub(crate) const SLOT_HEADER_LEN: usize = 2 + SCHEMA_FINGERPRINT_LEN + 4 + 4;
```
albertlockett, line 113: nit: it might be helpful to have a comment here for anyone who comes across these consts and wonders why they are set to these particular values. Maybe we could just linking to the relevant section of the architecture doc
https://github.com/open-telemetry/otel-arrow/blob/d6096d2435f5f30d3425ef8cec0cdeca22949c62/rust/otap-dataflow/crates/quiver/ARCHITECTURE.md#wal-file-format--rotation

rust/otap-dataflow/crates/quiver/src/engine.rs
albertlockett: got it, that makes sense. Thanks @AaronRM

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
lalitb: > Measuring this locally and time is on the order of ~30ms with a 64MB file, so doesn't seem worth putting in the sidecar.

Agreed on 30ms for typical cases, though edge devices with slower storage could see higher times. For now, should be good to scan only the active file starting from the checkpoint offset (as we read sidecar before wal) rather than from the beginning. We can revisit persisting the sequence number if startup time becomes an issue in production.

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
```diff
@@ -0,0 +1,1215 @@
...
+                return Err(WalError::SlotOutOfRange(slot.id));
+            }
+            if let Some(payload) = bundle.payload(slot.id) {
+                slot_bitmap |= 1u64 << slot_index;
+                let encoded_slot = self.prepare_slot(slot.id, payload)?;
+                let slot_size = encoded_slot.serialized_size();
+                total_payload_bytes = total_payload_bytes
+                    .checked_add(slot_size)
+                    .ok_or(WalError::EntryTooLarge(slot_size))?;
+                encoded_slots.push(encoded_slot);
```
albertlockett: This might be a minor/insubstantial optimization, but wouldn't we be able to avoid the heap allocation above of `let mut encoded_slots = Vec::new();` if we just wrote into `self.active_file.payload_buffer` directly here?

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
AaronRM: Added

rust/otap-dataflow/crates/quiver/src/wal/reader.rs
AaronRM: Yes, good point. The default rotation target size for a single file is 64MB. I added an upper bound of 256MB for the rotation target file size which is also used by the reader as an upper bound for an entry size.

Additionally, when we are reading the file, I added validation to ensure `entry_len` doesn't exceed the remaining file length to guard against this corruption/attack scenario.

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
```diff
@@ -0,0 +1,1215 @@
...
+fn encode_record_batch(batch: &RecordBatch) -> WalResult<Vec<u8>> {
+    let schema = batch.schema();
+    let mut buffer = Vec::new();
+    {
+        let mut writer = StreamWriter::try_new(&mut buffer, &schema).map_err(WalError::Arrow)?;
+        writer.write(batch).map_err(WalError::Arrow)?;
+        writer.finish().map_err(WalError::Arrow)?;
+    }
+    Ok(buffer)
+}
```
albertlockett, line 987: In the future, would we want to reuse the `StreamWriter` between record batches for a given slot? 

This could improve write performance and reduce the file size slightly. Having a `StreamWriter` per batch means we incur some extra overhead including initializing the StreamWriter, serializing/writing the arrow schema, and writing end of stream message when we close it.

I realize doing this would significantly complicate both the reader and the writer, so it's probably not worth doing unless this becomes a real bottleneck

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
AaronRM: I think that will become fairly substantial once we start stressing this code. Pushed a commit to remove `encoded_slots` altogether and use `payload_buffer` directly.

In thinking about this more, realized there is also an issue here where a single large RecordBundle may end up permanently bloating the `payload_buffer` capacity. (Will address that in a follow up PR.)

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
AaronRM: Will keep this in mind and come back for a possible future optimization (esp. if profiling shows a bottleneck here). Not addressing with this initial PR.

Issue #1508, opened 2025-12-02
Columnar query engine filtering additional feature support


albertlockett:
In #1477 we added the initial implementation of the the filtering `PipelineStage` for the columnar query engine.

There are many filtering patterns/features that are not supported which we should implement:
- [x] ~property/attribute is null (e.g. `severity_text == null` or `attributes["x"] == null`)~ https://github.com/open-telemetry/otel-arrow/pull/1538
  - [x] ~ensure we check the inverse of this, otherwise implement as special case~ https://github.com/open-telemetry/otel-arrow/pull/1551
- [x] ~Additional signal types:~ https://github.com/open-telemetry/otel-arrow/pull/1560
  - [x] ~metrics~
  - [x] ~traces~
- [x] function calls in the logical expression
- [x] arithmetic in the logical expression
- [x] ~Text comparison~ https://github.com/open-telemetry/otel-arrow/pull/1591
   - [x] ~contains~
   - [x] ~regex match~
- [x] ~Filtering by integer types that are not int64~ https://github.com/open-telemetry/otel-arrow/pull/1627
  - e.g. `logs | where dropped_attributes_count > 0` does not work
- [x] Comparing two fields:
  - [x] `logs | where severity_text == event_name`
  - [x] `logs | where attributes["x"] == attributes["z"]`
  - [x] `logs | where attributes["x"] == severity_text`
- [x] Compare literals e.g. `logs | where 1 == 2`
  - [x] ~Comparing literal to null `logs | where 1 == null` or `null == null`~ https://github.com/open-telemetry/otel-arrow/pull/1538
  - [x] Filter by a static literal. e.g. `logs | where true` (which expressions above get folded into)

lquerel:
All done

PR #1538, opened 2025-12-05
Columnar query engine filter handle `== null` filters

albertlockett:
part of #1508 

Handles filter predicates like `severity_text == null` or `attributes["x"] == null` (e.g. doesn't exist) when filtering using the columnar query engine.

It handles all the cases, including when the optional column is not present, when then the ID column (used to join attributes) is not present or null (meaning no attributes), and when attribute batches are entirely absent.

### DIFF
--- a/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
+++ b/rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
@@ -562,6 +562,7 @@ impl From<AdaptivePhysicalExprExec> for FilterExec {
         Self {
             predicate: Some(predicate),
             attributes_filter: None,
+            missing_attrs_pass: false,
         }
     }
 }
@@ -3080,6 +3081,7 @@ mod test {
                 projection: FilterProjection {
                     schema: vec![ProjectedSchemaColumn::Root("x".into())],
                 },
+                missing_data_passes: false,
             }),
         );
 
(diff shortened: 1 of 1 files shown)


rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
```diff
@@ -199,6 +199,19 @@ impl FilterPlan {
...
+            return Err(Error::InvalidPipelineError {
+                cause: format!(
+                    "cannot compare null using operator {}. only == is allowed",
+                    binary_op
+                ),
+                query_location: None,
+            });
+        }
+
         // TODO there are several branches below which are not yet supported supported
```
jmacd:
```suggestion
        // TODO there are several branches below which are not yet supported
```

rust/experimental/query_engine/engine-columnar/src/pipeline/filter.rs
```diff
@@ -360,9 +429,22 @@ impl ToExec for FilterPlan {
             .map(|attr_filter| attr_filter.to_exec(session_ctx, otap_batch))
             .transpose()?;
 
+        // compute how ot handle missing attributes. If the attrs filter is not(attr exists), then
```
jmacd:
```suggestion
        // compute how to handle missing attributes. If the attrs filter is not(attr exists), then
```

jmacd (approved):

PR #1539, opened 2025-12-05
Add mTLS client authentication for OTLP/OTAP receivers

lalitb:
Adds mutual TLS (mTLS) client certificate authentication to OTLP/OTAP receivers with parallel handshake processing and timeout protection.

  ## What's New

  ### 1. mTLS Client Certificate Verification
  - Client CA loading from file, PEM, or system pool
  - Standards-compliant X.509 validation via `WebPkiClientVerifier`
  - Mandatory authentication when CA is configured

  ### 2. Parallel Handshake Processing
  ```rust
  .buffer_unordered(MAX_CONCURRENT_HANDSHAKES)  // 64 concurrent handshakes
  Prevents head-of-line blocking from slow/malicious clients.
```

### 3. Handshake timeout
```rust
  tokio::time::timeout(handshake_timeout, acceptor.accept(conn))
  Default 10s, configurable. DoS protection against slowloris-style attacks.
```

### Enable mTLS

```yaml
  receivers:
    otlp:
      tls:
        cert_file: /path/to/server.crt
        key_file: /path/to/server.key
        client_ca_file: /path/to/client-ca.crt  # NEW
        handshake_timeout: "10s"  # NEW (default: 10s)
```

###  System CA Pool

```yaml
  receivers:
    otlp:
      tls:
        include_system_ca_certs_pool: true  # NEW
```

##   What's Deferred

  - Client CA certificate reload (separate PR)
  - Client-side TLS for exporters (separate PR)
  - TLS version/cipher configuration

### DIFF
--- a/rust/otap-dataflow/crates/telemetry/src/metrics/dispatcher.rs
+++ b/rust/otap-dataflow/crates/telemetry/src/metrics/dispatcher.rs
@@ -56,7 +56,7 @@ impl MetricsDispatcher {
 
     fn dispatch_metrics(&self) -> Result<(), Error> {
         self.metrics_handler
-            .visit_metrics_and_reset(|descriptor, attributes, metrics_iter| {
+            .visit_current_metrics(|descriptor, attributes, metrics_iter| {
                 let meter = global::meter(descriptor.name);
 
                 // There is no support for metric level attributes currently. Attaching resource level attributes to every metric.


rust/otap-dataflow/crates/otap/Cargo.toml
```diff
@@ -79,11 +79,11 @@ rustls-native-certs = { workspace = true, optional = true }
 base64 = { workspace = true, optional = true }
 arc-swap = { workspace = true, optional = true }
 rustls = { workspace = true, optional = true }
-rustls-pemfile = { workspace = true, optional = true }
+rustls-pki-types = { workspace = true, optional = true }
```
lalitb, line 82: `rustls-pemfile` is no longer maintained, so replaced by `rustls-pki-types`, which is maintained by `rustls` project.

rust/otap-dataflow/crates/otap/src/otap_receiver.rs
```diff
@@ -345,7 +348,7 @@ impl shared::Receiver<OtapPdata> for OTAPReceiver {
                 #[cfg(feature = "experimental-tls")]
                 match maybe_tls_acceptor {
                     Some(tls_acceptor) => {
-                        let tls_stream = create_tls_stream(listener_stream, tls_acceptor);
+                        let tls_stream = create_tls_stream(listener_stream, tls_acceptor, handshake_timeout);
```
lalitb, line 351: `handshake_timeout` is added to protect against slow/malicious clients that could hold connection slots indefinitely by not completing the TLS handshake.

Without a timeout, a malicious client could:

- Open a TCP connection
- Never send TLS ClientHello (or send it very slowly)
- Hold the connection slot forever, preventing legitimate clients from connecting
- Repeat this to exhaust server resources (DoS attack)

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -18,6 +23,14 @@ use tonic::transport::{Identity, ServerTlsConfig};
 /// while preventing potential OOM issues from loading extremely large files.
 const MAX_TLS_FILE_SIZE: u64 = 4 * 1024 * 1024; // 4MB
 
+/// Maximum number of concurrent TLS handshakes per receiver instance.
+///
+/// This is a conservative default that balances concurrency with resource usage:
+/// - Allows concurrent handshakes to prevent slow clients from blocking others
+/// - Limits memory overhead for pending handshake state
+/// - May need adjustment based on actual workload characteristics
+const MAX_CONCURRENT_HANDSHAKES: usize = 64;
```
lalitb, line 32: Not mTLS-specific - this is a general TLS hardening to prevent DoS via connection flooding. Limits concurrent handshakes to 64, applying backpressure when exceeded. Without this limit, an attacker could overwhelm the server by initiating thousands of handshakes simultaneously.

rust/otap-dataflow/crates/otap/Cargo.toml
cijothomas: https://github.com/open-telemetry/otel-arrow/pull/1536/files can be reverted with this.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -362,8 +391,79 @@ pub async fn build_reloadable_server_config(
...
+    } else {
+        // If system CAs were requested but no certs were loaded (and no user CA provided),
+        // warn the user that mTLS is effectively disabled or might not work as expected.
+        // Note: client_ca_pem is empty here.
+        if config.include_system_ca_certs_pool == Some(true) {
+            log::warn!(
+                "include_system_ca_certs_pool is true, but no CA certificates were loaded. mTLS will be disabled."
+            );
+        }
+        builder.with_no_client_auth()
```
albertlockett, line 466: Would it make sense to return an error here? It seems like it would be better from a security perspective not to fail open like this

rust/otap-dataflow/crates/otap/Cargo.toml
albertlockett: yes, we can either revert that as part of this PR, or I can do it in a followup after this is merged. Thanks for updating the unmaintained dependency @lalitb!

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -77,36 +90,59 @@ pub async fn load_server_tls_config(
...
+                        // Try TLS handshake
+                        let handshake_future = acceptor.accept(conn);
+                        let timeout_duration = handshake_timeout.unwrap_or(Duration::from_secs(10));
+
+                        match tokio::time::timeout(timeout_duration, handshake_future).await {
+                            Ok(Ok(stream)) => Some(Ok::<_, io::Error>(stream)),
+                            Ok(Err(e)) => {
+                                // TLS handshake failed - log and continue
+                                log::debug!("TLS handshake failed: {}", e);
+                                None
```
albertlockett, line 127: Do you think this is important enough to log at `warn` level? Or would that just create too much noise in the event that there were many failed handshakes?

rust/otap-dataflow/crates/otap/tests/mtls_tests.rs
```diff
@@ -0,0 +1,418 @@
...
+        );
+    }
+}
+
+#[tokio::test]
+async fn test_mtls_client_cert_verification() {
+    let _ = rustls::crypto::ring::default_provider().install_default();
+    if skip_if_no_openssl() {
+        return;
+    }
```
albertlockett: These tests look great!

Could we avoid skipping them if we used something like rcgen instead of openssl? https://docs.rs/rcgen/latest/rcgen/

I'm thinking it'd nice to ensure that these tests could always run. That way we can rely on them to catch regressions w/out worrying about if openssl accidentally got removed from some test runner.

If this is feasible, we can make the change a future PR (I don't want to block this PR from getting merged for this)

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -77,36 +90,59 @@ pub async fn load_server_tls_config(
...
+                        // Transport-level listener error - propagate to terminate server
+                        Some(Err(e))
+                    }
                 }
             }
-        }
-    })
+        })
+        // Allow concurrent handshakes to prevent slow/malicious clients from blocking others
+        .buffer_unordered(MAX_CONCURRENT_HANDSHAKES)
```
jmacd, line 143: Cool, how this works.

rust/otap-dataflow/crates/otap/tests/mtls_tests.rs
lalitb: That's a good suggestion. Have modified test to use `rcgen` now.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
lalitb: Yes, it's better to log `warn`, have changed accordingly.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
lalitb: Very good point. We should fail here. Updated accordingly.

rust/otap-dataflow/crates/otap/Cargo.toml
lalitb: I have reverted the #1536.

Issue #1340, opened 2025-10-23
Release otap-dataflow on crates.io


jmacd:
It would be convenient for users at this point, I think, to have crates released on crates.io that they can kick around.

See related issue https://github.com/open-telemetry/otel-arrow/issues/737

jmacd:
I discussed prerequisites with @lquerel. As a primary blocker, we need to integrate and rationalize the layout of our `otap-dataflow` crates. The `otel-arrow-rust` sibling crate will have to move in a coherent way into `otap-dataflow/crates/pdata`.

gouslu:
Can we also keep the crate version up-to-date as part of the release process? rust/otap-dataflow/Cargo.toml shows 0.1.0.

cijothomas:
> Can we also keep the crate version up-to-date as part of the release process? rust/otap-dataflow/Cargo.toml shows 0.1.0.

We are not publishing/releasing any Rust component from this repo yet.

gouslu:
> > Can we also keep the crate version up-to-date as part of the release process? rust/otap-dataflow/Cargo.toml shows 0.1.0.
> 
> We are not publishing/releasing any Rust component from this repo yet.

What I meant was that we should version the df_engine releases and we should do it such that the versioning is also reflected in Cargo.toml so that we can have access to the version of the application in the runtime using "CARGO_PKG_VERSION".

drewrelmas:
Marking this as completed as of Release `v0.55.0`! We now have process in place to publish additional crates when necessary.

PR #1540, opened 2025-12-06
Add more options to df_engine docker build, slim down features for some packages

JakeDern:
This PR addresses a few problems that I ran into getting `df_engine` running in my kubernetes environment. The end result of all this is that you can now run a docker build like this: ` docker build --build-arg FEATURES="azure" --build-arg RUSTFLAGS="-C target-feature=-gfni" --build-arg DEBUG=1 --build-context otel-arrow=../../ -f Dockerfile -t df_engine .`

# TLS

Some of the Azure SDK crates + reqwest have a lot of default features enabled that we are not using and which rely on linking with openssl for TLS. This was giving me trouble with our musl build.

I've disabled default features for these packages at the workspace level and opted for using `rustls-tls` with reqwest in the `otap` crate.

# Configuring the docker build

- Added the `FEATURES` build arg to the dockerfile and the cross platform build script so that you can turn on features like `azure` when building
- Added the ability to specify `RUSTFLAGS` for the docker build so that we can turn some features on or off when compiling which brings me to my next item...

# Adding some optional debug utils to the image

When running on k8s I was getting very mysterious crashes after a couple of seconds with 0 logs to stdout. Remoting into the container and running the binary gave me the "Illegal instruction (core dumped)" message, which was a start.

After checking the obvious vector extensions like SSE/AVX/AVX512, I was stumped. So, I added gdb to the container and found the exact instruction which is the longest mnemonic I have ever seen (comes from arrow's buffer crate):

```
(gdb)  x/i $pc
=> 0x7ffff745fb4f <new+463>:    vgf2p8affineqb $0x0,%ymm3,%ymm5,%ymm5
```

Turns out my laptop has support for a specific feature called `gfni`, which the nodes with Intel Xeon processors that I was running did not. 

I think this will not be the last time someone needs to debug a native code/cross-compilation problem, so I added an option to the build called `DEBUG` that will include `gdb` (and maybe other debug utilities in the future) in the image.

### DIFF
--- a/rust/otap-dataflow/Dockerfile
+++ b/rust/otap-dataflow/Dockerfile
@@ -24,7 +24,7 @@ COPY . /build/rust/dataflow/.
 
 # RUSTFLAGS can be used to pass argument to rustc e.g. 
 # --build-arg RUSTFLAGS="-C target-feature=-gfni"
-ARG RUSTFLAGS
+ARG RUSTFLAGS="-C target-cpu=native"
 ENV RUSTFLAGS=${RUSTFLAGS}
 
 # These are used to enable features and are passed to carg's --features argument


rust/otap-dataflow/crates/otap/Cargo.toml
```diff
@@ -70,7 +70,7 @@ geneva-uploader = { version = "0.3.0", optional = true }
 # Azure Monitor Exporter dependencies
 azure_identity = { workspace = true, optional = true }
 flate2 = { workspace = true, optional = true }
-reqwest = { workspace = true, optional = true }
+reqwest = { workspace = true, optional = true, default-features = false, features = ["rustls-tls"] }
```
lalitb: Couple of comments here -

- No need to repeat `[default-features = false]` in the crate if the workspace already sets it.
- It’s better to let the crate manage its own reqwest features instead of forcing them at the workspace level.
- Pick one of `rustls-tls` or `rustls-tls-native-roots`. They both pull in their own CA bundles, so using both at once doesn’t help.

lalitb:
One thing to keep in mind: the PR switches the TLS backend from `native-tls` (OpenSSL or BoringSSL) to `rustls` with `ring`. That solves the musl build issue, but ring isn’t `FIPS` 140-2 or 140-3 certified. If `FIPS` compliance becomes important later, we might want to look at `rustls-tls-aws-lc-rs` instead since aws-lc is FIPS validated, or potentially move back to `native-tls`. Not a blocker, just something to be aware of.

rust/otap-dataflow/crates/otap/Cargo.toml
JakeDern: Thank you for the review!

> No need to repeat [default-features = false] in the crate if the workspace already sets it.

Good to know! Removed it in the otap crate.

> It’s better to let the crate manage its own reqwest features instead of forcing them at the workspace level. 

Makes perfect sense, removed at the workspace level.

> Pick one of rustls-tls or rustls-tls-native-roots. They both pull in their own CA bundles, so using both at once doesn’t help.

Good catch, I meant to pick `rustls-tls` not `rustls-tls-native-roots` and forgot to update both places. Removed the workspace level feature anyways now, so we just have rustls-tls in the otap crate.

JakeDern:
> One thing to keep in mind: the PR switches the TLS backend from `native-tls` (OpenSSL or BoringSSL) to `rustls` with `ring`. That solves the musl build issue, but ring isn’t `FIPS` 140-2 or 140-3 certified. If `FIPS` compliance becomes important later, we might want to look at `rustls-tls-aws-lc-rs` instead since aws-lc is FIPS validated, or potentially move back to `native-tls`. Not a blocker, just something to be aware of.

Yeah absolutely, my goal here is just to try to get things working in the context of the existing docker build. We actually have a couple of other challenges on the Microsoft side related to crypto with this project, because `rustls-tls-aws-lc-rs` is not an approved library internally either, despite the FIPS compliance, and there's no plan for it to be.

I started trying to get arrow-rs-object-store to allow for pluggable crypto a few months ago rather than just swap from `ring` to `rustls-tls-aws-lc-rs`, but I haven't had time to revisit the work: https://github.com/apache/arrow-rs-object-store/pull/462.

rust/otap-dataflow/Dockerfile
```diff
@@ -22,12 +22,27 @@ COPY --from=otel-arrow /proto/opentelemetry-proto /build/proto/opentelemetry-pro
 
 COPY . /build/rust/dataflow/.
 
+# RUSTFLAGS can be used to pass argument to rustc e.g. 
+# --build-arg RUSTFLAGS="-C target-feature=-gfni"
+ARG RUSTFLAGS
+ENV RUSTFLAGS=${RUSTFLAGS}
```
albertlockett, line 28: I really like that we have the ability to set the rust flags for the docker build! 

However, I think the impact of this is that we'll override the default override the default rustflags from `./.cargo/config.toml` even if `--build-arg RUSTFLAGS...` isn't set. 

FYI, we set `-C target-cpu=native` in https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/.cargo/config.toml which is probably why you're seeing the AVX512 instructions which your build machine supports.

You'll see the rust flags getting set in the build if you add a `-v` to the `cargo build` step in `rust/otap-dataflow/cross-arch-build.sh` and run the docker build with `--progress plain`. On `main`, we'll see output like this:
```
#18 15.90      Running `/usr/local/rustup/toolchains/stable-aarch64-unknown-linux-gnu/bin/rustc <...snip...> -C target-cpu=native
```
On this PR's branch, we don't see `-C target-cpu=native`.

Without `target-cpu=native`, we wouldn't get micro-architecture specific instructions in the final binary, which we would probably want for best performance in many cases.

That said, the way we're setting the `target-cpu` in `./cargo/config.toml` is probably not correct if we eventually want to publish otap-dataflow on crates.io (which I think we eventually do https://github.com/open-telemetry/otel-arrow/issues/1340). I think this will cause our library compiled w/ native instructions, which users who depend on it might not expect.

One thing we could do is:
- a) remove the rustflags configuration from `./.cargo/config.toml`
- b) set the default value for the RUSTFLAGS docker ARG to `-C target-cpu=native`

Does that sound OK?

albertlockett: @clhain just want to confirm something -- for the performance tests in this repo, we're always using the docker images right? e.g. we're not building using `cross-arch-build.sh` directly (outside of docker) docker and then just running the resulting binary?

JakeDern: Thanks for the context, I had no idea we already had default args in `.cargo/config.toml`! 

That totally works for me, but what do you think about keeping the configuration in `.cargo/config.toml` so that folks will still get best performance by default when building/benchmarking locally and then still add `-C target-cpu=native` as the default for the docker build as well?

albertlockett: Oh yeah good point -- I hadn't thought about folks doing local benchmarking.

OK,  sounds good. Let's leave `.cargo/config.toml` as is for now, and just set `-C target-cpu=native` as the default for the docker build.

clhain: > @clhain just want to confirm something -- for the performance tests in this repo, we're always using the docker images right? e.g. we're not building using `cross-arch-build.sh` directly (outside of docker) docker and then just running the resulting binary?

Correct, always running in docker - not the binary directly.

PR #1547, opened 2025-12-08
Update contributing pre-req to include Rust

cijothomas:


cijothomas:
https://github.com/open-telemetry/otel-arrow/pull/1529 might have broken the CI required check.. If I don't fix it in next 1 hour, I'll revert the change to unblock and get back to it in a future date.

cijothomas:
> #1529 might have broken the CI required check.. If I don't fix it in next 1 hour, I'll revert the change to unblock and get back to it in a future date.

https://github.com/open-telemetry/otel-arrow/pull/1548 is done. Some more work is required before we can enable this. Better to come back to it later.

PR #1554, opened 2025-12-09
[otap-df-quiver] WAL: Shrink payload buffer after large bundle spikes

AaronRM:
Follow up for [an issue noted during the main WAL PR](https://github.com/open-telemetry/otel-arrow/pull/1537#discussion_r2599941542). Add memory reclamation for the WAL writer's payload buffer. Previously, after serializing a large RecordBundle, the buffer's capacity would remain elevated indefinitely since `Vec::clear()` preserves capacity.

Introduce a high-water mark with configurable decay to track typical usage. After each append, the mark decays by a configurable fraction (default 15/16 ≈ 6% per append). When capacity exceeds 2× the decayed high-water mark and is above 64 KB, shrink_to() reclaims excess memory.

- Add `buffer_decay_rate` option to `WalWriterOptions` with validation
- Add `InvalidConfig` error variant for configuration validation
- Add tests for decay rate validation and shrinking behavior
- Add (ignored) test to observe RSS behavior

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/wal/tests.rs
+++ b/rust/otap-dataflow/crates/quiver/src/wal/tests.rs
@@ -2482,7 +2482,12 @@ fn wal_memory_after_large_bundle_spike() {
 
     fn print_rss(label: &str) {
         if let Some(rss) = get_rss_kb() {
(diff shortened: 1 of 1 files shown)


jmacd:
:+1:

Issue #1520, opened 2025-12-04
Perf Test - have nightly run for zero load


cijothomas:
We don't really measure the CPU/Memory needed by the engine when it is serving no/minimal load. This is an important thing to measure with nightly perf tests.

cijothomas:
https://github.com/open-telemetry/otel-arrow/blob/main/docs/benchmarks.md#8-idle-state exists and this can be closed.

PR #1555, opened 2025-12-09
InternalMetrics- conditionally start dispatch thread if there are any readers

cijothomas:
~Alternate to https://github.com/open-telemetry/otel-arrow/pull/1552 
Since perf tests don't create service/telemetry, this will also unblock perf tests.~

Simple optimization- avoid the thread creation if there are no readers.

### DIFF
--- a/rust/otap-dataflow/crates/config/src/pipeline.rs
+++ b/rust/otap-dataflow/crates/config/src/pipeline.rs
@@ -680,6 +680,7 @@ pub struct MetricsConfig {
 
 impl MetricsConfig {
     /// Returns `true` if there are any metric readers configured.
+    #[must_use]
     pub fn has_readers(&self) -> bool {
         !self.readers.is_empty()
     }


rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -662,6 +662,17 @@ impl Default for TelemetryConfig {
     }
 }
 
+impl TelemetryConfig {
+    /// Returns `true` if there are any metric readers configured.
+    ///
+    /// This can be used to conditionally initialize OpenTelemetry SDK components
+    /// only when telemetry export is actually configured.
+    #[must_use]
+    pub fn has_metric_readers(&self) -> bool {
```
andborja: Make it more generic pub(crate) fn configured_metrics

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -75,9 +75,21 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
         // Initialize metrics system and observed event store.
         // ToDo A hierarchical metrics system will be implemented to better support hardware with multiple NUMA nodes.
         let telemetry_config = &pipeline.service().telemetry;
-        let opentelemetry_client = OpentelemetryClient::new(telemetry_config);
+        let has_metric_readers = telemetry_config.has_metric_readers();
+
+        // Only create the OpenTelemetry client and dispatcher if there are metric readers configured.
```
andborja: Is it worth optimizing it? Does it consume resources if no exporters are provided?

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -75,9 +75,21 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
...
+        let opentelemetry_client = if has_metric_readers {
```
andborja: It should not be optional based on metrics only. It will have logs and traces in the future. It should always exist.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -91,11 +103,15 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
                 metrics_system.run(cancellation_token)
             })?;
 
-        // Start the metrics dispatcher
-        let metrics_dispatcher_handle =
-            spawn_thread_local_task("metrics-dispatcher", move |cancellation_token| {
-                metrics_dispatcher.run_dispatch_loop(cancellation_token)
-            })?;
+        // Start the metrics dispatcher only if there are metric readers configured.
+        let metrics_dispatcher_handle = if let Some(dispatcher) = metrics_dispatcher {
```
andborja: I think the only check should be here. No metrics are configured, do not start the thread.

rust/otap-dataflow/crates/controller/src/lib.rs
cijothomas: Not really. It'd do nearly nothing.

Once we have https://github.com/open-telemetry/otel-arrow/issues/1520, we'd know if the internal telemetry pipeline would be even measurable here.

rust/otap-dataflow/crates/controller/src/lib.rs
cijothomas: Agree

rust/otap-dataflow/crates/controller/src/lib.rs
cijothomas: good point. fixed.

PR #1556, opened 2025-12-09
Minor readme improvement for dataflow configs

cijothomas:
The readme.md of the pipeline's config directory was outdated. Updated it to match reality. I am also not super sure if we need this or its better to just remove the list of all configs..

### DIFF
--- a/.github/workflows/pipeline-perf-test-nightly.yml
+++ b/.github/workflows/pipeline-perf-test-nightly.yml
@@ -71,6 +71,10 @@ jobs:
             docker build --build-context otel-arrow=../../ -f Dockerfile -t df_engine .
             cd ../..
 
+        - name: Build otelarrowcol
+          run: |
+            docker build . -t otelarrowcol
+
         - name: Install dependencies
           run: |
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/configs/fake-perf.yaml
```diff
@@ -14,7 +14,7 @@ nodes:
         dispatch_strategy: round_robin
     config:
       traffic_config:
-        max_signal_count: 1000
+        # max_signal_count: 1000
```
cijothomas, line 19: commented out. otherwise, the fake gen just finishes its job in 1st iteration itself. Now it'll continue producing fake logs non-stop.

jmacd:
```suggestion
        # For testing with a finite stream, use max_signal_count. Otherwise
        # this component outputs a continuous signal.
        # max_signal_count: 1000
```

albertlockett (approved):

lquerel (approved):

PR #1558, opened 2025-12-09
add filter to nightly html page

cijothomas:
**Note: This for for the benchmarks branch, not *main***

To add filter results to https://open-telemetry.github.io/otel-arrow/benchmarks/nightly/


clhain (approved):
Looks good, we still need to update the filter index.html to do all the aggregation of the different test runs - I'll ask Chanly to have a look at that.

Issue #1422, opened 2025-11-12
Remove #[allow(missing_docs)] from rust/otap-dataflow/crates/pdata


jmacd:
There was a crate-level #[allow(missing_docs)]. It was replaced by dozens of struct, method and field-level annotations, but they persist.

This issue should block #1340 (and I don't see other blockers).

See #1395.

jmacd:
I realize it now, but looking back, the issue description needed more detail. The task is to remove #[allow(missing_docs)] throughout the entire codebase (mainly as it persists in crates/pdata, explained in the issue), by writing comments for anything with a #[allow(missing_docs)].

jmacd:
Fixed?!?

PR #1562, opened 2025-12-09
Remove #[allow(missing_docs)] from rust/otap-dataflow/crates/pdata

SAROJJV:


jmacd:
As part of #1422, we actually want comments written 😁
I will close this for now.

PR #1563, opened 2025-12-09
Add batch Sizer; configure inbound/outbound

### LINKED ISSUE #1569: Batch processor: sizer support
The Collector exporterhelper's [batch configuration](https://github.com/open-telemetry/opentelemetry-collector/blob/8f51a1792addcfc6beb636fd4c57b13941af78d3/exporter/exporterhelper/internal/queuebatch/config.go#L18) struct has four fields:
 
- min_size
- max_size
- sizer
- flush_interval 

where `sizer` is one of "requests", "items", or "bytes". We should aim to use the same configuration. It is permitted to return an error for an unsupported sizer, but "items" should always be supported. 

https://github.com/open-telemetry/otel-arrow/pull/1563 modifies the configuration with a new sizer, setting it to "items"; in the future, we can add limits based on request count or even estimated byte count, this way compatible with the Collector.

jmacd:
Batch configuration "modernization". The configuration had been copied from the Collector batch_processor which has 3 fields (batch_size, batch_size_max, timeout). The newer BatchConfig has "min_size", "max_size", "sizer" (an enum for items, requests, bytes), and "flush_timeout".

Fixes #1569 

Initial settings for inbound/outboudn limits were arbitrary at 1000/1000. I figure outbound are likely to be larger than inbound; also I realized we have an 8192 default so, 1024 and 512 seemed more natural than 1000/1000. 

This brings the batch processor in line with the newer sending_queue::batch configuration model and addresses a related TODO; configuration now extends to the number of inbound and outbound slots.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/batch_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/batch_processor.rs
@@ -64,7 +64,7 @@ use std::time::{Duration, Instant};
 pub const OTAP_BATCH_PROCESSOR_URN: &str = "urn:otel:batch:processor";
 
 /// Default configuration values (parity-aligned as we confirm Go defaults)
-pub const DEFAULT_MIN_SIZE: usize = 8192;
+pub const DEFAULT_MIN_SIZE_ITEMS: usize = 8192;
 
 /// Timeout in milliseconds for periodic flush
 pub const DEFAULT_TIMEOUT_MS: u64 = 200;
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -75,32 +75,60 @@ const LOG_MSG_DROP_CONVERSION_FAILED: &str =
...
 /// Batch processor configuration.
 #[derive(Debug, Clone, Deserialize)]
 pub struct Config {
     /// Flush current batch when this count is reached, as a
     /// minimum. Measures the number of items (logs: records, traces:
     /// spans, metrics: data points). When pending data reaches this
     /// threshold a new batch will form and be sent.
-    #[serde(default = "default_send_batch_size")]
-    pub send_batch_size: Option<NonZeroUsize>,
+    #[serde(default = "default_min_size")]
```
cijothomas: the doc comment for this should be updated to reflect that `Sizer` determines whether size refers to count of items or size or request..

jmacd: Re-commented, thanks!

lalitb:
LGTM. One minor observation regarding the default values, just to ensure the behavior change is intentional - The hardcoded outbound slot limit was previously 1000. The new default outbound_request_limit is 512, this effectively halves the default outbound buffer size, This is likely fine (and safer), but it's worth noting that the default behavior has become slightly more restrictive for outbound requests.

jmacd:
Added:

> Initial settings for inbound/outboudn limits were arbitrary at 1000/1000. I figure outbound are likely to be larger than inbound; also I realized we have an 8192 default so, 1024 and 512 seemed more natural than 1000/1000.

lalitb:
> Added:

My bad, should have read the PR description properly. :)

lalitb (approved):

PR #1564, opened 2025-12-09
Naming consistency for batching logic

jmacd:
Relates to #1563 which adds a `sizer`. Existing batching logic is entirely based on item counts, therefore to add clarity:

Renames the batch_length() function to num_items().

Renames the max_output_batch parameter to max_items.


albertlockett (approved):
nice good improvment

albertlockett:
failing CI fixed in https://github.com/open-telemetry/otel-arrow/pull/1568

PR #1566, opened 2025-12-10
feat: Add OTLP export support for internal telemetry opentelemetry

andborja:
Add OTLP export support for internal telemetry opentelemetry integration.


rust/otap-dataflow/configs/fake-debug-noop-telemetry.yaml
```diff
@@ -41,5 +41,20 @@ service:
         - periodic:
             exporter:
               console:
+        - periodic:
```
cijothomas, line 41: I think just one is enough for this config. No need to show all possible options.

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -728,39 +729,75 @@ impl<'de> Deserialize<'de> for MetricsReaderConfig {
...
+    /// Use HTTP with JSON encoding for communication.
+    HttpJson,
+}
+
+/// The temporality of the metrics to be exported.
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Default, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum Temporality {
+    /// Delta temporality means that the metric values are the difference between the current and previous values.
+    #[default]
```
cijothomas: nit: Cumulative is the default in OTel world.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
```diff
@@ -3,74 +3,54 @@
...
-use crate::error::Error;
+use opentelemetry::KeyValue;
+use opentelemetry_sdk::{Resource, metrics::SdkMeterProvider};
+use otap_df_config::pipeline::TelemetryConfig;
+
+use crate::{error::Error, opentelemetry_client::meter_provider::MeterProvider};
 
 /// Client for the OpenTelemetry SDK.
 pub struct OpentelemetryClient {
+    /// The tokio runtime used to run the OpenTelemetry SDK.
```
cijothomas: tokio runtime is a requirement very specific to `tonic` gRPC exporter. Lets try to keep/use it only when tonic+gRPC OTLP Exporter is configured, and nowhere else.

andborja: Correct. It is instantiated only if used at least once.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
```diff
@@ -3,74 +3,54 @@
...
+            meter_provider,
+        })
+    }
 
-        Self {
-            meter_provider: sdk_meter_provider,
+    fn configure_resource(
+        resource_attributes: &std::collections::HashMap<String, String>,
+    ) -> Resource {
+        let mut sdk_resource_builder = Resource::builder();
```
cijothomas, line 47: using `builder()` comes prefilled with few resource detectors. Not sure if we need that, else use `builder_empty`

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider.rs
```diff
@@ -0,0 +1,301 @@
...
+                    .with_tonic()
+                    .with_endpoint(&otlp_config.endpoint)
+                    .with_temporality(Self::to_sdk_temporality(&otlp_config.temporality))
+                    .build()
+            })
+            .map_err(|e| Error::ConfigurationError(e.to_string()))?;
+        Ok((exporter, Some(tokio_runtime)))
+    }
+
+    fn to_sdk_temporality(config: &Temporality) -> opentelemetry_sdk::metrics::Temporality {
```
cijothomas, line 146: maybe implement `From`/ `Into` for these things, as that feels more Rust idiomatic.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
andborja: Can you elaborate? I'm not sure I'm following.  The resource attributes part is not implemented yet. We need to make it not only strings.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider.rs
andborja: As the objects are from different crates, they cannot be implemented

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider.rs
```diff
@@ -0,0 +1,300 @@
...
+        global::set_meter_provider(sdk_meter_provider.clone());
+
+        Ok(MeterProvider {
+            sdk_meter_provider,
+            runtime,
+        })
+    }
+
+    /// Consume the MeterProvider and return its components.
+    pub fn deconstruct(self) -> (SdkMeterProvider, Option<tokio::runtime::Runtime>) {
```
jmacd: The Rust idiom for this function is `into_parts()`, I recently learned. ([for example](https://arrow.apache.org/rust/arrow_array/array/struct.PrimitiveArray.html#method.into_parts), we have also `OtapPdata::into_parts` for the pair of Context and Payload).

```suggestion
    pub fn into_parts(self) -> (SdkMeterProvider, Option<tokio::runtime::Runtime>) {
```

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider.rs
```diff
@@ -0,0 +1,300 @@
...
+        match &otlp_config.protocol {
+            OtlpProtocol::Grpc => {
+                (exporter, runtime) = Self::configure_grpc_otlp_exporter(otlp_config, runtime)?
+            }
+            OtlpProtocol::HttpBinary => {
+                exporter = Self::configure_http_exporter(otlp_config, Protocol::HttpBinary)?
+            }
+            OtlpProtocol::HttpJson => {
+                exporter = Self::configure_http_exporter(otlp_config, Protocol::HttpJson)?
+            }
```
jmacd, line 106: Can HTTP/JSON be an optional feature, or even left out in this case?

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
cijothomas: Please modify the comment also to make it clear its a OTLP exporter specific requirement, not SDK.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider.rs
```diff
@@ -0,0 +1,300 @@
...
+        (
+            opentelemetry_otlp::MetricExporter,
+            Option<tokio::runtime::Runtime>,
+        ),
+        Error,
+    > {
+        // If there is a tokio runtime already, use it. Otherwise, create a new one.
+        let tokio_runtime = match runtime {
+            Some(rt) => rt,
+            None => tokio::runtime::Runtime::new()
```
cijothomas, line 130: This part , I am not sure what is the right thing to do. If we just create a runtime like above, will it create new thread(s)? Or should we use the current-thread runtime? I am unsure how this would affect rest of df_engine.

andborja: It basically creates the runtime "lazy/on-demand". It will be created only if needed, and only once.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
andborja: Sure. Updated.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider.rs
cijothomas: I am referring to which thread(s) the runtime will run on? The df_engine has some logic to ensure it runs on the threads (and cpu cores) requested by user. This is probably not an issue, but I don't have enough understand of the engine/controller to confirm.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider.rs
andborja: The user is not forced to configure it... only used if configured. It is not a separate library.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider.rs
lalitb: This Tokio runtime is created once in the main/controller thread, before the per-core pipeline threads are spawned. There are per-core runtimes, which are different.

cijothomas: I also opened https://github.com/open-telemetry/otel-arrow/issues/1573 to discuss this more. Maybe there is no action required, but something to track.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
cijothomas: I was asking if you should use builder() or builder_empty().
https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/resource/mod.rs#L61-L62

If using builder(), OTel SDK adds some built-in attributes, which may be not really needed in the df-engine. 
Not a blocker for this PR!

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -728,39 +729,75 @@ impl<'de> Deserialize<'de> for MetricsReaderConfig {
...
+pub struct OtlpExporterConfig {
+    /// The Otlp communication protocol to use when exporting data.
+    #[serde(default)]
+    pub protocol: OtlpProtocol,
+    /// The endpoint to which the Otlp exporter will send data.
+    pub endpoint: String,
+
+    /// The temporality of the metrics to be exported.
+    #[serde(default)]
+    pub temporality: Temporality,
```
lalitb: For future iteration - good to add header support - for auth headers (bearer token, API-KEYs etc). otel-otlp supports adding headers in builder, so should be feasible to extend.

lalitb:
Changes looks good. Thanks :)

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
andborja: Got it. I'll consider it when implementing that part (next PR).

lalitb (approved):

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider.rs
jmacd: Thanks for the issue. This is an important question. My best understanding is that we are likely to be using a https://docs.rs/tokio/latest/tokio/runtime/struct.LocalRuntime.html in each of the pipelines so we need to change into a threaded runtime, and this is probably OK. @lquerel please see #1573

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider.rs
jmacd: I see, so really the question is whether the OTel SDK has a feature flag to avoid building all three exporter modes.

PR #1577, opened 2025-12-10
feat: Add mTLS client CA hot-reload with file watching

lalitb:
## What This Does

  Adds automatic hot-reload of mTLS **client CA certificates** for OTLP/OTAP receivers. When the CA file changes, certificates reload automatically without restarting the service or dropping connections.

  **Important:** This PR is specifically for **client CA reload (mTLS)**. Server certificate reload uses a different mechanism (covered separately). And it follows the pattern used in Go based otel collector.

  ## Reload Mechanisms Summary

| Certificate Type | Configuration          | Reload Method      | When Checked                                      |
|------------------|------------------------|--------------------|---------------------------------------------------|
| Server Cert      | `reload_interval`      | Lazy polling       | During TLS handshake (if interval expired)        |
| Client CA        | `watch_client_ca: true`| File watching      | Immediate (~50-500ms after file change)           |
| Client CA        | `watch_client_ca: false`| Background polling | Every `reload_interval` (dedicated thread)        |

  - **Client CA**: Uses `watch_client_ca` to choose between file watcher (notify) or interval polling
  - **Server Cert**: Always uses lazy polling on handshake, ignores `watch_client_ca` flag

  ## Why We Need This

  Modern cloud deployments rotate certificates frequently:
  - **SPIFFE/SPIRE** issues 1-hour certificates
  - **cert-manager** rotates daily/weekly
  - **Zero-trust** architectures require continuous rotation

  Without this, every cert rotation means restarting receivers and losing telemetry data.

  ## How It Works

  ### Hot Path (TLS Handshakes)
  The verification path is wait-free - just an atomic pointer load. No filesystem access, no blocking.

  ```rust
  fn verify_client_cert(...) -> Result<...> {
      let verifier = self.inner.load();  // Fast atomic load
      verifier.verify_client_cert(...)    // Delegate
  }

```

## Reload Path (File Changes)

  Two reload modes are supported for client CA certificates:

  ### File Watching Mode (watch_client_ca: true):
  - Uses OS-native file notifications (inotify/kqueue)
  - Detects changes within 50-500ms
  - No polling overhead

  ### Polling Mode (watch_client_ca: false):
  - Checks file periodically based on reload_interval
  - More compatible with exotic filesystems
  - Less responsive but simpler

  **When a change is detected:**

  1. **Filter** - Check if it's actually our CA file
  2. **Debounce** - Wait 1 second to avoid reload storms
  3. **Verify** - Check file identity changed (inode)
  4. **Reload** - Load new CA and atomic swap
  5. **Graceful fallback** - Keep old CA if reload fails

  The reload happens in the file watcher's OS thread, not the async runtime, so it doesn't block other connections.

  ## Design Decision: Why Not Use Async Channels?

  We could bridge the file watcher to a tokio worker task via channels, which would make all I/O async. However:

  - The blocking I/O (checking file metadata) happens in the watcher's dedicated OS thread, not the tokio runtime
  - Certificate reloads are rare (minutes/hours apart)
  - The hot path (TLS handshakes) is already wait-free
  - A channel adds complexity for negligible benefit here

  We documented this trade-off in the code for future reference. If we need to refactor later, the channel pattern is a viable option. Also the current

 ## Configuration

  ### File Watching Mode (Recommended for Client CA)

  Immediate reload when CA file changes (50-500ms detection):

```yaml
  receivers:
    otlp:
      protocols:
        grpc:
          tls:
            cert_file: /etc/certs/server.crt       # Server cert (lazy polling)
            key_file: /etc/certs/server.key
            client_ca_file: /etc/certs/ca.crt      # Client CA (file watching)
            watch_client_ca: true                   # Enable OS file watching for CA
            reload_interval: 5m                     # Used for server cert polling
```

  ### Polling Mode (Fallback for Client CA)

  Periodic checks based on interval (better filesystem compatibility):

```yaml
  receivers:
    otlp:
      protocols:
        grpc:
          tls:
            cert_file: /etc/certs/server.crt       # Server cert (lazy polling)
            key_file: /etc/certs/server.key
            client_ca_file: /etc/certs/ca.crt      # Client CA (polling)
            watch_client_ca: false                  # Use polling for CA
            reload_interval: 1m                     # Check CA every 1 min (default: 5m)
```

  **Clarifications:**
  - `watch_client_ca` only affects client CA reload, not server certificates
  - `reload_interval` is used for:
    - Server cert lazy polling (always)
    - Client CA polling (only when `watch_client_ca: false`)

  ## Performance Impact

  
**Hot Path (TLS Handshakes):**
- Hot-reload overhead: ~3-5 nanoseconds (single atomic pointer load)
- No file I/O or blocking operations during handshake
- Lock-free - no contention between connections
- Total verification time dominated by crypto operations, not reload mechanism

**Memory:**
- ReloadableClientCaVerifier struct: ~100 bytes
- Inner CA store: Depends on number of CA certificates loaded
- Shared across all connections - not duplicated per connection

**Threads:**
- +1 OS thread per receiver (notify crate's watcher thread)
  - `watch_client_ca: true` → blocked in epoll/kqueue (near-zero CPU)
  - `watch_client_ca: false` → polling on interval (minimal CPU)

**CPU:**
- File watcher: Event-driven (inotify/kqueue) - idles when no changes
- Reload: ~1-10ms when CA changes (file read + PEM parse + cert validation)
- Frequency: Typically hours/days between reloads


rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -318,6 +374,590 @@ impl ResolvesServerCert for LazyReloadableCertResolver {
...
+/// only affect the watcher thread, which exists solely to monitor file changes.
+///
+/// Since CA reloads are rare (minutes/hours apart), the performance impact is
+/// negligible. The TLS handshake path remains wait-free - just an atomic pointer load.
+///
+/// Alternative: A channel-based bridge to a tokio worker task would eliminate blocking
+/// entirely, but adds complexity for minimal benefit in this use case.
+struct CaWatcherState {
+    /// The verifier to update on reload
+    inner: Arc<ArcSwap<Arc<dyn ClientCertVerifier>>>,
```
jmacd, line 393: This construction Arc<ArcSwap<Arc<dyn T>>>>-- could use more comments!

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -318,6 +374,590 @@ impl ResolvesServerCert for LazyReloadableCertResolver {
...
+/// chain validation.
+impl ClientCertVerifier for ReloadableClientCaVerifier {
+    fn root_hint_subjects(&self) -> &[DistinguishedName] {
+        // This is called during handshake to send CertificateRequest.
+        // We need to return a static slice, but our verifier can change.
+        // The safest approach is to not send hints (empty slice), which
+        // is valid per TLS spec and is what Envoy does for dynamic CAs.
+        //
+        // Note: Some clients may not like this. If needed, we can cache
+        // hints and accept that they may be slightly stale.
```
jmacd: I can't quite follow, it sounds dangerous. Maybe add 1 sentence explaining what hints are?

jmacd (approved):

rust/otap-dataflow/crates/otap/src/tls_utils.rs
lalitb: Agree, bit complex - good part is only inner Arc is in the hot-path. Have added the comment for easy understanding.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
lalitb: The wording did sound dangerous. Have rephrased it.  Root hints are the list of trusted CA Distinguished Names that the server sends to clients during the TLS handshake to help them choose which certificate to present when they have multiple options. We return an empty list here because this method requires a static reference but our CA verifier can change at runtime due to hot-reload, and returning empty is TLS-spec compliant with no practical impact since most clients only have one certificate anyway.

gouslu (approved):

PR #1579, opened 2025-12-10
Fix collector build

cijothomas:
Collector build failed: https://github.com/open-telemetry/otel-arrow/actions/runs/20094832773/job/57650543120#logs

I was able to repro the failure locally, and the fix is working locally now.

Changes:
Updated the builder version from v0.89.0 to v0.140.0 to match the component versions in otelarrowcol-build.yaml
Changed the build to run from within otelarrowcol where the go.mod exists
Added go mod tidy before building to resolve dependencies

(liberally used GenAI help, as my Collector building experience is minimal.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,14 +14,14 @@ COPY . .
 ENV CGO_ENABLED=0
 
 # Note the version MUST MATCH otelarrowcol-build.yaml
-RUN go install go.opentelemetry.io/collector/cmd/builder@v0.89.0
+RUN go install go.opentelemetry.io/collector/cmd/builder@v0.140.0
 
 # This command generates main.go, go.mod but does not update deps.
 RUN builder --skip-compilation --skip-get-modules --config=collector/otelarrowcol-build.yaml
 
-# This build will update the go.mod, using the checked-in go.work file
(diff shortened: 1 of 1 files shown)


cijothomas:
@c1ly @clhain PTAL

Dockerfile
```diff
@@ -14,14 +14,14 @@ COPY . .
 ENV CGO_ENABLED=0
 
 # Note the version MUST MATCH otelarrowcol-build.yaml
-RUN go install go.opentelemetry.io/collector/cmd/builder@v0.89.0
+RUN go install go.opentelemetry.io/collector/cmd/builder@v0.140.0
```
jmacd, line 17: Ah! This is correct in the sense that the collector/otelarrowcol-build.yaml already refers to a v0.140.x series of components. However, it means a previous change modified the component versions but not the cmd/builder version, and these genuinely need to match (i.e., the "MUST MATCH" comment above was not followed!).

There is also a `make` rule that I suppose you could invoke (i.e., `make otelarrowcol`) which checks that the code base is up-to-date with respect to the commands on line 22-24. If the source code is up-to-date we shouldn't need these steps.

PR #1580, opened 2025-12-10
move columnar query engine into otap-dataflow

albertlockett:
Moves the columnar query engine from experimental into otap-dataflow.

The primary reason I'd like to move this is for other projects that might depend on otap-dataflow pdata crate. For example, if a crate were to depend on otap-dataflow and locate it through git, like:
```
otap-df-pdata = { git = "https://github.com/open-telemetry/otel-arrow.git", package = "otap-df-pdata", #...
```
Cargo will try to clone the git repo and scan through it to find a crate called `otap-df-pdata`. However, cargo may encounter query_engine/engine-columnar/Cargo.toml which depends on `otap-df-pdata` by a relative path, and get confused, resulting in an error like:
```
error: no matching package named `otap-df-pdata` found
```
In summary, I think this probably would have needed to happen anyway before we publish on crates.io https://github.com/open-telemetry/otel-arrow/issues/1340

If the reason above is not convincing, we can additionally justify moving this into otap-dataflow on the following merits:
1. the implementation is quite specific to OTAP, in that the pipeline stages operate on and produce `OtapArrowRecords`
2. the experimental stage of this implementation was acheived in the POC https://github.com/open-telemetry/otel-arrow/pull/1342, and currently we're trying to build a columnar query engine that is of comparable quality/maturity with the other otap-dataflow crates.

I realize the README and rust docs are currently light on content, but I will happily update those in a followup PR if we're OK with relocating this code.

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/Cargo.toml
+++ b/rust/otap-dataflow/crates/query-engine/Cargo.toml
@@ -17,7 +17,6 @@ parking_lot = { workspace = true }
 roaring = { workspace = true }
 thiserror = { workspace = true }
 data_engine_expressions = { path = "../../../experimental/query_engine/expressions" }
-data_engine_kql_parser = { path = "../../../experimental/query_engine/kql-parser" }
 
 otap-df-pdata = { path = "../pdata" }
 
@@ -26,6 +25,7 @@ criterion = { workspace = true, features = ["async_tokio"] }
 pretty_assertions = { workspace = true }
 prost = { workspace = true }
 tokio = { workspace = true }
+data_engine_kql_parser = { path = "../../../experimental/query_engine/kql-parser" }
 
 [[bench]]
 name = "filter"


rust/otap-dataflow/crates/query-engine/Cargo.toml
```diff
@@ -16,10 +16,10 @@ futures-core = { workspace = true }
 parking_lot = { workspace = true }
 roaring = { workspace = true }
 thiserror = { workspace = true }
+data_engine_expressions = { path = "../../../experimental/query_engine/expressions" }
+data_engine_kql_parser = { path = "../../../experimental/query_engine/kql-parser" }
```
jmacd: I am curious about these two dependencies, and how they don't create the same problem you mentioned in the PR description? 

I was wondering if kql-parser was a dev-dependency or a dependency?

I take it we can move these two into a non-experimental location eventually.

jmacd (approved):

rust/otap-dataflow/crates/query-engine/Cargo.toml
albertlockett: > I am curious about these two dependencies, and how they don't create the same problem you mentioned in the PR description?

This is a good callout -- they might if some project was depending on these crates within the `experimental` folder. Do you know of any projects that are doing that?  (cc @CodeBlanch for this question as well).

If we merge this and it becomes an issue for building those projects, one thing I could try would be to change the location of these dependencies here from `path=...` to `git=https://github.com/open-telemetry/otel-arrow.git`.

> I was wondering if kql-parser was a dev-dependency or a dependency?

Moved kql-parser to a dev dependency. good catch!

> I take it we can move these two into a non-experimental location eventually.

:+1: I think we probably could

.github/workflows/rust-ci.yml
```diff
@@ -39,6 +39,39 @@ jobs:
...
+          sudo rm -rf /usr/local/share/chromium
+          # Remove Microsoft/Edge and Google Chrome builds
+          sudo rm -rf /opt/microsoft /opt/google
+          # Remove Azure CLI
+          sudo rm -rf /opt/az
+          # Remove PowerShell
+          sudo rm -rf /usr/local/share/powershell
+
+          echo "disk usage after"
+          df -h
```
albertlockett, line 73: @drewrelmas does this look OK to you?

Trying to fix this job running out of disk space:
https://github.com/open-telemetry/otel-arrow/actions/runs/20113439435/job/57716769401

drewrelmas: I don't think I have any problem with this.

cijothomas: we have to do similar cleanup in other repos too! This is fine.

albertlockett:
> Nit - please update [labeler config](https://github.com/open-telemetry/otel-arrow/blob/main/.github/labeler.yml) to reflect the move, as well as [this readme](https://github.com/open-telemetry/otel-arrow/tree/main/rust/experimental/query_engine).

yep! I'll do ASAP in a followup

Issue #1334, opened 2025-10-22
otel-arrow-rust batching logic panic


jmacd:
otap_batch_processor.rs has an ignored test:

```
    // Test constants for batching smoke tests
    #[test]
    #[ignore = "upstream-batching-bug"]
    // NOTE: Validates cross-signal partitioning and expected row totals.
    fn test_make_output_batches_partitions_and_splits() {
```

I do not believe we have recorded the underlying bug. Indeed, when you unignore this test there is a panic in the underlying otel-arrow-rust groups.rs.

jmacd:
There are several test cases commented out in https://github.com/open-telemetry/otel-arrow/blob/95471b6fb7ad8df2a79298148267f64fee6422aa/rust/otap-dataflow/crates/pdata/src/otap/batching_tests.rs#L105

These are known to crash the OTAP batching logic for metrics.

jmacd:
In #1861 when I connect the ITS mode with the the file `rust/otap-dataflow/configs/internal-telemetry.yaml`, with an ITS `batch_processor` with `format: otap` to exercise the OTAP batching logic, that the code panics like:

thread 'internal-pipeline' (378332) panicked at /home/jmacd/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/arrow-array-57.0.0/src/array/struct_array.rs:91:46:
called `Result::unwrap()` on an `Err` value: InvalidArgumentError("Incorrect datatype for StructArray field \"str\", expected Dictionary(UInt8, Utf8) got Dictionary(UInt16, Utf8)")

jmacd:
The ITS configuration like:

```
# Internal telemetry pipeline - separate from main pipeline
# Uses hardcoded settings: single thread, no admin server
internal:
  telemetry:
    kind: receiver
    plugin_urn: "urn:otel:otap:internal_telemetry:receiver"
    out_ports:
      out_port:
        destinations:
          - batch
        dispatch_strategy: round_robin
    config: {}
  batch:
    kind: processor
    plugin_urn: "urn:otel:batch:processor"
    out_ports:
      out_port:
        destinations:
          - console
        dispatch_strategy: round_robin
    config:
      otap:
        min_size: 1000
        sizer: items
      format: otap
      flush_timeout: 3s
  console:
    kind: exporter
    plugin_urn: "urn:otel:console:exporter"
    config: {}

service:
  telemetry:
    logs:
      level: "debug"
      providers:
        global: its
        engine: its
[...]

PR #1583, opened 2025-12-10
otap_df_pdata::otap::batching metric batching tests & minor fixes

jmacd:
Part of #1334.

Adds configurability to the DataGenerator interface, letting the test script various configurations of input data for batching/equivalence testing.

The original issue described in #1334 has been addressed; the batching_tests module is more comprehensive, and another test `#[ignore]` has been removed. Overall, the metrics-batching logic is more correct.

However, this method of testing has uncovered at least three more defects, so #1334 may remain open. The tests are commented out.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -28,7 +28,10 @@ jobs:
           - otap-dataflow
           - experimental/query_abstraction
           - experimental/query_engine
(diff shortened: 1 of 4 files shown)


albertlockett (approved):
Nice tests @jmacd  -- and good that we were able to find identify a few issues

Issue #1576, opened 2025-12-10
OTAP dataflow logging SDK requirements


jmacd:
From [a review thread about debug and info-level logging](https://github.com/open-telemetry/otel-arrow/pull/1490#discussion_r2604108454).

We need to adopt a logging SDK for the OTAP dataflow engine to make our code base more accessible.

As an OpenTelemetry project, I think we should defer to the best practice recommended by OpenTelemetry-Rust.

While we have justified the use of a special metric SDK internally, the justification used ("multivariate", in particular) do not apply to logs and spans. We need to plan for emitting logs and/or spans in this code base, and this is an issue to discuss requirements.

The tokio `tracing` crate is recommended by OpenTelemetry-Rust. We believe it is compatible with our thread-per-core design, so long as the SDK is configured in a per-thread way.

Are there requirements to avoid `Sync` and `Send` traits?

This raises questions about contextual enrichment the way an OpenTelemetry logging SDK might do. While the otap-dataflow engine has a Context type where we could propagate the OTel span context, this would likely have to be introduced in the `otap_df_otap::pdata::Context` before we could benefit from pipeline-level tracing.

cijothomas:
To get us started with something (instead of println! statements!), I propose to do the following:

1. Introduce logging macros, similar to what OTel Rust does. (https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry/src/global/internal_logging.rs)
2. This will use `tracing` to begin with, but free to change as we learn more and get more feedback.
3. Engine can read a simple config for logging (min_severity_level and destination - only stdout to start with), and use this to setup a `fmt` subscriber from `tracing` that outputs logs/events to stdout. To begin with, I'll set this up as `global` affecting entire process. In a future iteration, we can try doing this on a per thread basis (after measuring the impact on perf/contention etc.)

This will give us a basic picture of how logging would work. In a future iteration, we can discuss leveraging OTel SDK and output the logs via OTLP or elsewhere (like OS_native facilities like etw/linux trace points).

I'll send a PR with the above shortly to facilitate easier discussion. 

(Hoping to get some basic logging in, before folks start holidays!)

lquerel:
What matters most to me, in order of priority:
- minimal impact on the engine's performance and adherence to the thread-per-core and NUMA-aware design
- options to configure the log and trace reporting level and to disable them entirely if needed, ideally either globally or per pipeline instance

cijothomas:
[...]

PR #1584, opened 2025-12-10
Add internal logging support

cijothomas:
Attempt at https://github.com/open-telemetry/otel-arrow/issues/1576#issuecomment-3638738082

Haven't done the configuration part in this PR yet. That can be added after I get a green-light to proceed with this approach.

This adds the logging macros. Supports EventName, key-value pairs.
Telemetry crate also has helper to setup `tracing:fmt` to print logs to stdout.
Controller initializes this.
To demonstrate how logging looks like, added couple of simple logs.

Note: This is now *off by default*, so there won't be any perf hit.

Sample output:

```txt
Running `target/release/df_engine -p configs/fake-perf.yaml --num-cores 1`
Starting pipeline with 1 cores
2025-12-10T23:28:21.057469Z  INFO otap-df-controller:  name="controller_start" num_nodes=2 pdata_channel_size=100
2025-12-10T23:28:21.057736Z  INFO opentelemetry:  name="MeterProvider.GlobalSet" Global meter provider is set. Meters can now be created using global::meter() or global::meter_with_scope().
2025-12-10T23:28:21.060217Z  INFO otap-df-admin:  name="admin_server_listening" endpoint="127.0.0.1:8080" Admin HTTP server listening
Starting Perf Exporter
2025-12-10T23:28:23.175542Z  INFO otap-df-otap:  name="fake_data_generator_receiver_start" signals_per_second=1000 max_batch_size=1000 Fake data generator receiver started
```

(OpenTelemetry Rust logs also shows up, as it is instrumented with tracing too. It is extremely easy to control which logs we want vs now. Not showing it here, but if needed, I can expand the description to show them too. Also, `tracing-log` can be enabled and it'll capture logs from crates using `log` crate too!)


rust/otap-dataflow/crates/telemetry/src/internal_events.rs
```diff
@@ -0,0 +1,120 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
```
cijothomas, line 2: Copied from OTel Rust project, originally written by @lalitb

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -686,6 +690,39 @@ impl MetricsConfig {
...
+/// Log level for internal engine logs.
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Default)]
+#[serde(rename_all = "lowercase")]
+pub enum LogLevel {
+    /// Logging is completely disabled.
+    Off,
+    /// Debug level logging.
+    Debug,
+    /// Info level logging.
+    #[default]
```
cijothomas: Given the fact the current implementation does have contention, I think it is better to use `Off` as the default so as to keep the engine perf not affected.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -321,11 +329,25 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
         pipeline_ctrl_msg_tx: PipelineCtrlMsgSender<PData>,
         pipeline_ctrl_msg_rx: PipelineCtrlMsgReceiver<PData>,
     ) -> Result<Vec<()>, Error> {
+        // Create a tracing span for this pipeline thread
```
cijothomas, line 332: @lquerel @jmacd Showing how we can contextual enrichment by specifying these info once, and it'll auto-inherit for all subsequent logs.

If needed, these can be converted to proper OTel spans as well. (using a 3rd party crate named `tracing-opentelemetry`)

rust/otap-dataflow/crates/telemetry/src/lib.rs
```diff
@@ -34,12 +34,88 @@ pub mod collector;
...
+        LogLevel::Warn => LevelFilter::WARN,
+        LogLevel::Error => LevelFilter::ERROR,
+    };
+
+    // If RUST_LOG is set, use it for fine-grained control.
+    // Otherwise, fall back to the config level with some noisy dependencies silenced.
+    // Users can override by setting RUST_LOG explicitly.
+    let filter = EnvFilter::try_from_default_env().unwrap_or_else(|_| {
+        // Default filter: use config level, but silence known noisy HTTP dependencies
+        EnvFilter::new(format!("{level},h2=off,hyper=off"))
```
cijothomas, line 110: hyper/h2 are heavily instrumented with `tracing` macros, so I am disabling it here temporarily. 

The right way is to leverage the existing syntax, shown below for an example.

```
RUST_LOG=info,otap=debug,opentelemetry=debug,h2=off,hyper=off cargo run --release -- -p configs/fake-perf.yaml --num-cores 1

```

The above would enable debug logs from otap* and opentelemetry*, turn off h2 and hyper, and info from everything else.

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
lalitb: Ah yes, my finest work... that I have zero memory of writing  😅 :)

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -72,9 +72,17 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
         quota: Quota,
         admin_settings: HttpAdminSettings,
     ) -> Result<(), Error> {
-        // Initialize metrics system and observed event store.
         // ToDo A hierarchical metrics system will be implemented to better support hardware with multiple NUMA nodes.
         let telemetry_config = &pipeline.service().telemetry;
+
+        // Initialize internal logging.
+        init_logging(&telemetry_config.logs);
```
lalitb, line 79: Will it make sense to add a NUMA-aware note here, as queue is allocated here in main thread's NUMA node, and all pipelines push to this queue.  This is somewhat different from metrics, which uses thread local for aggregation.

cijothomas: that method has a TODO Comment for this. I'll mention NUMA-awrrenes there.

rust/otap-dataflow/crates/admin/src/lib.rs
```diff
@@ -75,6 +76,17 @@ pub async fn run(
             details: format!("{e}"),
         })?;
 
+    otel_info!(
```
lquerel, line 79: Since the event name is required, could we rewrite this macro to use a more regular syntax (see the proposal below)?

```rust
otel_info!(
  "admin_server_listening",
  endpoint: addr.to_string(),
  message: "Admin HTTP server listening"
);

```

Or use "=" syntax only if my suggestion can't be implemented for whatever reason.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -321,11 +329,25 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
...
         pipeline_ctrl_msg_rx: PipelineCtrlMsgReceiver<PData>,
     ) -> Result<Vec<()>, Error> {
+        // Create a tracing span for this pipeline thread
+        // so that all logs within this scope include pipeline context.
+        let span = otel_info_span!(
+            "pipeline_thread",
+            pipeline_group = %pipeline_key.pipeline_group_id,
+            pipeline = %pipeline_key.pipeline_id,
+            core_id = core_id.id,
+        );
```
lquerel, line 339: This `otel_info_span!` follows more or less the suggestion I made before. I will suggest to replace the `=` by `:` in order to be more aligned with the field syntax used for structs.

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
```diff
@@ -0,0 +1,120 @@
...
+
+#[doc(hidden)]
+pub mod _private {
+    pub use tracing::{debug, error, info, warn};
+}
+
+/// Macro for logging informational messages.
+///
+/// # Fields:
+/// - `name` (required): The OpenTelemetry Event name identifying the log event.
```
lquerel, line 21: See my previous comment.

cijothomas:
Closed for https://github.com/open-telemetry/otel-arrow/pull/1601

rust/otap-dataflow/crates/admin/src/lib.rs
lquerel: Also, what happens to the `endpoint` parameter if the log level `info` is not enabled? Is `to_string` still evaluated in that case or not? I assume it is not, but I just want to be sure.

rust/otap-dataflow/crates/telemetry/src/lib.rs
```diff
@@ -34,12 +34,88 @@ pub mod collector;
...
+///    ```ignore
+///    // start of thread
+///    let _guard = tracing::subscriber::set_default(subscriber);
+///    // now, with this thread, all tracing calls will go to this subscriber
+///    // eliminating contention.
+///    // end of thread
+///    ```
+///
+/// 2. Use custom subscriber that batches logs in thread-local buffer, and
+///    flushes them periodically.
```
lquerel, line 88: Definitively the most important aspect for me.

lquerel (approved):
I made few comments and suggestions that I believe will improve the dev  experience. Thanks

rust/otap-dataflow/crates/admin/src/lib.rs
cijothomas: > Also, what happens to the `endpoint` parameter if the log level `info` is not enabled? Is `to_string` still evaluated in that case or not? I assume it is not, but I just want to be sure.

it is not per my understanding also. I'll be sure to double/triple check these. My priority is also - if something is not enabled, it should cost as close to zero as possible.

lalitb: > Is to_string still evaluated in that case or not? I assume it is not, but I just want to be sure.

It is not evaluated if no subscriber is attached, or level is not enabled.

PR #1591, opened 2025-12-11
columnar query engine filtering support for text contains and regex match

albertlockett:
part of #1508 

Adds support for filtering where text contains a substring or if text matches a regex.

For the text contains, datafusion's [built-in contains function](https://github.com/apache/datafusion/blob/e914935660875bc415887ade3fea1f5e59bed5e1/datafusion/functions/src/string/contains.rs#L101-L148) does not support dictionary arrays, so I added a version that extends it to add support in a new module `pipeline::functions`. There is also a child_module of this called `expr_fn` which exposes functions that produce logical exprs that invoke the UDF (this is similar to how it works in upstream datafusion).

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/functions/contains.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/functions/contains.rs
@@ -20,8 +20,8 @@ use datafusion::scalar::ScalarValue;
 /// dictionary encoded arrays.
 ///
 // TODO some of this code is lifted & adapted from [`datafusion::functions::string::contains`] and
-// eventually we could contribute it back upstream to add support upstream for dictionary arrays,
-// then upstream contains function directly.
+// eventually we could contribute it back upstream to add support for dictionary arrays, then use
+// contains function directly.
 #[derive(Default, Debug, PartialEq, Eq, Hash)]
 pub struct ExtendedContainsFunc {
     inner: ContainsFunc,


rust/experimental/query_engine/engine-columnar/src/pipeline/functions/contains.rs
```diff
@@ -0,0 +1,468 @@
...
+    fn invoke_with_args(&self, scalar_func_args: ScalarFunctionArgs) -> Result<ColumnarValue> {
+        let args = scalar_func_args.args;
+        if args.len() != 2 {
+            return exec_err!("invalid number of args. received {} expected 2", args.len());
+        }
+
+        // take the args
+        // safety: we've already checked the length of the args vec
+        let mut iter = args.into_iter();
+        let left = iter.next().expect("there are wo args");
```
jmacd:
```suggestion
        let left = iter.next().expect("there are two args");
```

rust/otap-dataflow/crates/query-engine/src/pipeline/functions/contains.rs
```diff
@@ -0,0 +1,468 @@
...
+                coerced_scalar
+            );
+        }
+    };
+
+    Ok(scalar_str)
+}
+
+// this is lifted somewhat verbatim from [`datafusion::functions::string::contains`] which is why
+// we should eventually contribute this back upstream
```
jmacd, line 150: :+1:

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
```diff
@@ -402,17 +407,164 @@ impl TryFrom<&LogicalExpression> for Composite<FilterPlan> {
...
+            BinaryArg::Column(left_column) => {
+                let mut attrs = None;
+                let left_expr = match left_column {
+                    ColumnAccessor::ColumnName(left_col_name) => col(left_col_name),
+                    ColumnAccessor::StructCol(left_struct_name, left_struct_field) => {
+                        col(left_struct_name).field(left_struct_field)
+                    }
+                    ColumnAccessor::Attributes(attrs_identifier, attrs_key) => {
+                        attrs = Some((attrs_identifier, attrs_key));
+                        // for now we assume that text contains is always applied to the str column
```
jmacd: Got it -- we're coercing Arrow types above, not OpenTelemetry types.

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
```diff
@@ -402,17 +407,164 @@ impl TryFrom<&LogicalExpression> for Composite<FilterPlan> {
...
+            BinaryArg::Literal(left_lit) => {
+                let left_expr = try_static_scalar_to_literal(&left_lit)?;
+
+                let mut attrs = None;
+                let right_expr = match right_arg {
+                    BinaryArg::Column(right_column) => match right_column {
+                        ColumnAccessor::ColumnName(right_col_name) => col(right_col_name),
+                        ColumnAccessor::StructCol(right_struct_name, right_struct_field) => {
+                            col(right_struct_name).field(right_struct_field)
+                        }
```
jmacd: I think this repeats the handling for left_column above, it might help to add a new function.

albertlockett: Good calll @jmacd , made this change

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
albertlockett: Yeah exactly -- the coercion that happens in the contains function is for the arrow types.

The type coercion that's happening there is more general than what we need for OTAP specifically (because we don't use `LargeUtf` or `UtfView` types), but I was tentatively planning to try to contribute that code back to upstream datafusion in the future so I wanted to keep the same coercion logic that the contains function currently uses.

PR #1594, opened 2025-12-11
Few fixes to make Windows CI better

cijothomas:
No need to run code coverage.
Attempt to fix a test that is failing in Windows CI only.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/object_store.rs
+++ b/rust/otap-dataflow/crates/otap/src/object_store.rs
@@ -209,6 +209,7 @@ mod test {
         }
     }
 
+    // Skipping on Windows: https://github.com/open-telemetry/otel-arrow/issues/1614
     #[test]
     #[cfg(not(windows))]
     fn test_get_testdelayed_file_storage() {
@@ -218,6 +219,7 @@ mod test {
         assert!(from_storage_type(&storage).is_ok());
     }
 
+    // Skipping on Windows: https://github.com/open-telemetry/otel-arrow/issues/1614
     #[test]
     #[cfg(not(windows))]
     fn test_get_file_storage() {
--- a/rust/otap-dataflow/crates/otap/src/otap_exporter.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_exporter.rs
@@ -782,7 +782,7 @@ mod tests {
     }
 
     // Skipping on Windows due to flakiness: https://github.com/open-telemetry/otel-arrow/issues/1611
-    #[cfg_attr(target_os = "windows", ignore)]
+    #[cfg(not(windows))]
     #[test]
     fn test_receiver_not_ready_on_start() {
         let grpc_addr = "127.0.0.1";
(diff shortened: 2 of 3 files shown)


cijothomas:
Windows CI still fails (fix coming later), but it improved time now. 14mins to 11mins now after removing code coverage.

rust/otap-dataflow/crates/otap/src/object_store.rs
```diff
@@ -211,16 +211,17 @@ mod test {
...
     #[test]
     fn test_get_testdelayed_file_storage() {
-        let storage = StorageType::File {
-            base_uri: "testdelayed:///tmp".to_string(),
-        };
+        let temp_dir = std::env::temp_dir();
+        let base_uri = format!("testdelayed://{}", temp_dir.display());
+        let storage = StorageType::File { base_uri };
         assert!(from_storage_type(&storage).is_ok());
     }
```
albertlockett, line 220: @cijothomas this is failing on windows. Does std::env::temp_dir() return something like `C:\Users\YourUsername\AppData\Local\Temp`?

I think the reason this test could be failing is that we try to parse this as a url in tests if it starts with `testdelayed`
https://github.com/open-telemetry/otel-arrow/blob/b70ddc7bff3d04142f4a248a985f5557a0bfdbc0/rust/otap-dataflow/crates/otap/src/object_store.rs#L110-L133

We could probably clean up this hack eventually, but in the interim I think it might fix this test if we just strip off the `C:` prefix:

```suggestion
    #[test]
    fn test_get_testdelayed_file_storage() {
        let mut temp_dir = std::env::temp_dir();
        if temp_dir.starts_with("C:") {
            temp_dir = base_uri.strip_prefix("C:").expect("has prefix");
        }
        let base_uri = format!("testdelayed://{}", temp_dir.display());
        let storage = StorageType::File { base_uri };
        assert!(from_storage_type(&storage).is_ok());
    }
```

jmacd (approved):

rust/otap-dataflow/crates/otap/src/object_store.rs
cijothomas: I disabled test in windows for now to make CI green. Will come back to fix all windows failures in future.

albertlockett:
@cijothomas looks like the test is also failing on windows.
```
    thread 'otap_exporter::tests::test_receiver_not_ready_on_start' (8840) panicked at crates\otap\src\otap_exporter.rs:882:13:
    assertion `left == right` failed
   ```

I created an issue for it https://github.com/open-telemetry/otel-arrow/issues/1611

Could you add this 
```rs
    #[test]
    #[cfg(not(windows))] // TODO https://github.com/open-telemetry/otel-arrow/issues/1611
    fn test_receiver_not_ready_on_start() {
```
to:
https://github.com/open-telemetry/otel-arrow/blob/785f92edfcbf9f2f9797933c9236a8a35a7d5369/rust/otap-dataflow/crates/otap/src/otap_exporter.rs#L784-L785

PR #1600, opened 2025-12-11
feat: Add resource attributes support for different types of data

andborja:
Add resource attributes support for different types of data

### DIFF
--- a/rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
+++ b/rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
@@ -130,6 +130,10 @@ impl<'de> Deserialize<'de> for AttributeValue {
             where
                 E: de::Error,
             {
+                if value > i64::MAX as u64 {
+                    let message = format!("value {} out of range (max {})", value, i64::MAX);
+                    return Err(de::Error::custom(message));
+                }
                 Ok(AttributeValue::I64(value as i64))
             }
 
@@ -401,6 +405,19 @@ mod tests {
             );
         } else {
             panic!("Expected deserialization to fail for mixed type array");
+        }
+    }
+
+    #[test]
+    fn test_attribute_value_u64_out_of_range_error() {
+        let yaml_str = r#"
+            out_of_range_value: 9223372036854775808
+            "#;
+        let result: Result<HashMap<String, AttributeValue>, _> = serde_yaml::from_str(yaml_str);
+        if let Err(err) = &result {
+            assert!(err.to_string().contains("out of range"));
+        } else {
+            panic!("Expected deserialization to fail for out of range u64 value");
         }
     }
 }


rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
```diff
@@ -50,6 +49,162 @@ fn default_reporting_interval() -> Duration {
     Duration::from_secs(1)
 }
 
+/// Attribute value types for telemetry resource attributes.
```
cijothomas, line 84: is this specific to Resource only? or for Log/metric attributes as well? 

Also, could we use the ones from `opentelemetry` crate instead?

cijothomas: @lalitb WDYT? Should this just take `opentelemetry` types as it has to be anyway converted to that format...

andborja: Yes. Metric attributes are transformed from the internal telemetry attributes.
The opentelemetry objects are not serializable, so they cannot be referenced directly.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
```diff
@@ -50,6 +49,162 @@ fn default_reporting_interval() -> Duration {
...
+    String(String),
+    /// Boolean type attribute value.
+    Bool(bool),
+    /// 64-bit integer type attribute value.
+    I64(i64),
+    /// 64-bit floating point type attribute value.
+    F64(f64),
+    /// Array type attribute value.
+    Array(AttributeValueArray),
+}
```
jmacd, line 97: crates/telemetry/src/attributes has a very similar type, could we use it? If the Array type is missing (it is) and you need it, could we add it there instead?

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
```diff
@@ -50,6 +49,162 @@ fn default_reporting_interval() -> Duration {
...
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, JsonSchema)]
+pub enum AttributeValueArray {
+    /// Array of bools
+    Bool(Vec<bool>),
+    /// Array of integers
+    I64(Vec<i64>),
+    /// Array of floats
+    F64(Vec<f64>),
+    /// Array of strings
+    String(Vec<String>),
```
jmacd, line 241: Can we use a Box<Vec<AttributeValue>> ? or are you intentionally limiting this to 1 dimension of string array?

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
cijothomas: Or we can just support simple String or basic types, and come back in future to add arrays.

jmacd (approved):

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
andborja: Yes, the spec is only 1 dimension for resource attributes.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
andborja: The dependency is to the opposite side. It should be created in the config crate if we want to reuse it.
The user of the "telemetry" crate is "admin" crate, and it adds a few helper methods (not only serialization).
What we can do later is to migrate the one in the telemetry crate to this one, after evaluating what will happen with the admin crate.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
```diff
@@ -42,16 +44,43 @@ impl OpentelemetryClient {
     }
 
     fn configure_resource(
-        resource_attributes: &std::collections::HashMap<String, String>,
+        resource_attributes: &std::collections::HashMap<String, AttributeValue>,
     ) -> Resource {
-        let mut sdk_resource_builder = Resource::builder();
+        let mut sdk_resource_builder = Resource::builder_empty();
```
lalitb, line 49: `builder_empty()` is going to drop the default attributes ('telemetry.sdk' etc). Is this intended - Or should we be keeping these attributes, and then layer config-provided attributes on top.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
```diff
@@ -82,6 +81,162 @@ pub enum LogLevel {
...
+                E: de::Error,
+            {
+                Ok(AttributeValue::I64(value))
+            }
+
+            fn visit_u64<E>(self, value: u64) -> Result<Self::Value, E>
+            where
+                E: de::Error,
+            {
+                Ok(AttributeValue::I64(value as i64))
```
lalitb, line 137: This seems to be silently conveting `u64` to `i64`, which mayn't always be safe/correct. ` > i64::max` will get wrapped into a negative number. Better to error on overflow, or else log a warning and drop the attribute.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
lalitb: > The dependency is to the opposite side. It should be created in the config crate if we want to reuse it.

Agree on this. If we really want to reuse, we can later think to have a shared attribute crate, but I believe it should be simple to keep them separate for now.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
lalitb: > Should this just take opentelemetry types as it has to be anyway converted to that format...

It would have been possible, if we could implement `Deserialize / JsonSchema` for `opentelemetry::Value` in `otap-df-config`. Unfortunately, rust `orphan rule` doesn't allow that.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
```diff
@@ -82,6 +81,162 @@ pub enum LogLevel {
...
+    Bool(bool),
+    /// 64-bit integer type attribute value.
+    I64(i64),
+    /// 64-bit floating point type attribute value.
+    F64(f64),
+    /// Array type attribute value.
+    Array(AttributeValueArray),
+}
+
+impl<'de> Deserialize<'de> for AttributeValue {
```
lalitb, line 99: nit comment - we have custom deserializer, how ever we rely on the default serializer. This is fine for df_engine reading configs, but round-tripping (YAML/JSON -> struct -> YAML/JSON) and any schema-based tooling/docs may be confusing or fail. Not urgent, just flagging that aligning serialize/schema with the accepted input format would avoid surprises.

lalitb:
LGTM with couple of comments, which can be handled later.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
jmacd: We have a proliferation-of-identical types problem. I don't want to force us to address it at this time, and I don't want it to last forever. We are also duplicating opentelmetry-proto with otap_df_pdata::proto and we have slight differences to resolve relating to TraceID/SpanID handling and derive macros.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
jmacd: That's sort of changing but let's not worry about it. 
https://opentelemetry.io/blog/2025/complex-attribute-types/

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
jmacd: Thanks. This is an area that worries me, but also I don't to see this code base slide into becoming a configuration project.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
jmacd: I would prefer error, myself.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
jmacd: I would leave these in by default.

jmacd (approved):

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry.rs
andborja: Validation added

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
andborja: This is intended. The user defines its own attributes.

PR #1605, opened 2025-12-12
Tweak internal logging macros to be more dev friendly

cijothomas:
Address part of https://github.com/open-telemetry/otel-arrow/pull/1584/changes#r2612281448

Before:
```rust
otel_info!(
            name: "Receiver.Start",
            signals_per_second = signals_per_second.unwrap_or(0),
            max_batch_size = max_batch_size,
            message = "Fake data generator receiver started"
        );
```

After: 
```rust
otel_info!(
            "Receiver.Start",
            signals_per_second = signals_per_second.unwrap_or(0),
            max_batch_size = max_batch_size,
            message = "Fake data generator receiver started"
        );
```

The "=" syntax for attribute is still kept... I think `tracing` users are quite used to that pattern is very popular. We can re-adjust if there are strong opinion on that. Laurent has ":" [preference](https://github.com/open-telemetry/otel-arrow/pull/1584/changes#r2612321520)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -10,9 +10,10 @@ ci-repo:
 query-engine:
 - changed-files:
   - any-glob-to-any-file: 'rust/experimental/query_engine/**/*'
+  - any-glob-to-any-file: 'rust/otap-dataflow/crates/query-engine/**/*'
 query-engine-columnar:
 - changed-files:
-  - any-glob-to-any-file: 'rust/experimental/query_engine/engine-columnar*/**/*'
+  - any-glob-to-any-file: 'rust/otap-dataflow/crates/query-engine/**/*'
 query-engine-recordset:
 - changed-files:
   - any-glob-to-any-file: 'rust/experimental/query_engine/engine-recordset*/**/*'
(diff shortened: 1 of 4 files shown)


jmacd (approved):
🙈

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -327,11 +327,19 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
         pipeline_ctrl_msg_tx: PipelineCtrlMsgSender<PData>,
         pipeline_ctrl_msg_rx: PipelineCtrlMsgReceiver<PData>,
     ) -> Result<Vec<()>, Error> {
+        // Create a tracing span for this pipeline thread
+        // so that all logs within this scope include pipeline context.
+        let span = otel_info_span!("pipeline_thread", core.id = core_id.id);
+        let _guard = span.enter();
```
lquerel, line 333: Q: How does it work for the guard when the info level is disabled? Does the guard effectively become a no-op?

cijothomas: yes that is correct. `span.disabled()` will be true then.

cijothomas:
> Thanks for this change!
> 
> In a future PR, I think we should set up `weaver registry live-check` in our CI so we can start ensuring that we follow the semantic conventions for naming spans, attributes, metrics, and so on.

https://github.com/open-telemetry/otel-arrow/issues/1613 to track this.

PR #1606, opened 2025-12-12
Telemetry: Support f64 metrics, fix gauge accumulation, enrich pipeline-oriented monitoring

lquerel:
PR summary:
- [most important/interesting change, all the rest is relatively mechanical] New pipeline-oriented metrics, see the comments in the `pipeline_metrics.rs` for more details. **It's now possible to measure the cpu and memory usage per pipeline instance!**
- Use jemalloc instead of mimalloc to capture per-thread memory metrics, I didn't observe any performance regression under load. When jemalloc is not used the memory-oriented metrics are not reported.
- Telemetry core now supports f64 values end‑to‑end (all metric types).
- Registry accumulation fixed to respect instrument semantics (fixed bug for Gauge).

Future work: add pipeline-oriented `network_io_received` and `network_io_sent` metrics.


rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
```diff
@@ -0,0 +1,580 @@
...
+//!   The per-interval increase (`cpu_time_delta`) is the CPU budget actually used.
+//!
+//! - `cpu_utilization` (`Gauge<f64>`, `{1}`):
+//!   Ratio of `cpu_time_delta / wall_time_delta` over the last interval,
+//!   clamped to `[0, 1]`. Interpreted as "how CPU-bound this pipeline thread is".
+//!   On a core shared by multiple pipeline specs, this reflects the pipeline's
+//!   share of that core; summing `cpu_utilization` across co-pinned pipelines
+//!   approximates total core load.
+//!
+//! - `memory_usage` (`Gauge<u64>`, `{By}`; jemalloc only):
```
jmacd, line 33: (semantics) Should be UpDownCounter.

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
```diff
@@ -0,0 +1,580 @@
...
+//!   Current heap bytes in use by the pipeline thread:
+//!   `memory_allocated - memory_freed`. A rising long-term trend indicates a leak
+//!   or persistent buffering growth inside this pipeline.
+//!
+//! - `memory_allocated` / `memory_freed` (`Gauge<u64>`, `{By}`; jemalloc only):
```
jmacd, line 38: UpDownCounter.

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
```diff
@@ -0,0 +1,580 @@
...
+//!   Cumulative heap bytes allocated/freed by this pipeline since start.
+//!   These are baselines for long-term trends and sanity checks.
+//!
+//! - `memory_allocated_delta` / `memory_freed_delta` (`Counter<u64>`, `{By}`; jemalloc only):
+//!   Per-interval heap bytes allocated/freed. They capture allocation churn
+//!   even when `memory_usage` is stable. High churn increases allocator and cache
+//!   pressure and often correlates with latency variance.
```
jmacd, line 45: :+1:

Technically these two can be used to which is equal to `memory_allocated` (i.e., Counter - Counter = UpDownCounter). We'd expect a metrics query engine to do this sort of thing!

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
```diff
@@ -0,0 +1,580 @@
...
+//!   Minor faults indicate working-set or locality churn. Spikes can also occur
+//!   during first-touch of large buffers, and can highlight uneven sharding across
+//!   NUMA nodes. Major faults indicate disk I/O due to swapping and almost always
+//!   imply severe performance degradation.
+//!
+//! **Derived signals (not emitted explicitly)**
+//! - `net_allocated_delta = memory_allocated_delta - memory_freed_delta` `{By}`
+//! - `allocation_rate = memory_allocated_delta / interval_s` `{By/s}`
+//! - `deallocation_rate = memory_freed_delta / interval_s` `{By/s}`
+//! - `net_allocation_rate = net_allocated_delta / interval_s` `{By/s}`
```
jmacd, line 73: Yes, we'd expect a metrics query engine to do this sort of thing!

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
```diff
@@ -0,0 +1,580 @@
...
+    #[metric(unit = "{By}")]
+    pub memory_usage: Gauge<u64>,
+
+    /// Memory allocated to the heap by the pipeline instance since start.
+    #[metric(unit = "{By}")]
+    pub memory_allocated: Gauge<u64>,
+
+    /// Memory freed to the heap by the pipeline instance since start.
+    #[metric(unit = "{By}")]
+    pub memory_freed: Gauge<u64>,
```
jmacd, line 172: Three UpDownCounters.

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
```diff
@@ -0,0 +1,580 @@
...
+/// - deallocation_rate   = memory_freed_delta / interval_s             [{By/s}]
+/// - net_allocation_rate = net_allocated_delta / interval_s            [{By/s}]
+/// - cpu_utilization_ratio = (cpu_time_delta_s / wall_time_delta_s)    [{1}]
+///   where cpu_time_delta_s is the per-interval increase in cpu_time.
+#[metric_set(name = "pipeline.metrics")]
+#[derive(Debug, Default, Clone)]
+pub struct PipelineMetrics {
+    /// The time the pipeline instance has been running.
+    #[metric(unit = "{s}")]
+    pub uptime: Gauge<f64>,
```
jmacd, line 160: Here, the semantics are sort of an open question. I say UpDownCounter, you could argue for Counter, but I would say definitely not Gauge. I suggest UpDownCounter because this is a value that never descends but does reset. I suggest maybe Counter becasue it is non-descending. I suggest not Gauge because we are interested in the rate of uptime, which approximates the Prometheus metric named `up`.

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
```diff
@@ -0,0 +1,580 @@
...
+    pub memory_freed_delta: Counter<u64>,
+
+    /// Total CPU seconds used by the pipeline since start.
+    #[metric(unit = "{s}")]
+    pub cpu_time: Counter<f64>,
+
+    /// Difference in pipeline cpu time since the last measurement, divided by the elapsed time.
+    /// Reported as a ratio in the range [0, 1].
+    #[metric(unit = "{1}")]
+    pub cpu_utilization: Gauge<f64>,
```
jmacd, line 189: (Curiosity question: what made you decide to place a field here, instead of deriving it as the ratio of cpu_time / real time? Alternatively, the ratio of cpu_time/uptime as discussed above?)

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
lquerel: Our OTel semantic conventions define `process.uptime` as a Gauge :-)
 
https://opentelemetry.io/docs/specs/semconv/system/process-metrics/#metric-processuptime

However, I would like to better understand why you think it has to be an UpDownCounter.

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
lquerel: It was mostly for convenience that I added `cpu_utilization`, so it can be observed directly on our metrics HTTP endpoint without having to do any calculations. I use that endpoint a lot during development. 
I can remove it if you prefer.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -293,6 +303,15 @@ impl<PData> PipelineCtrlMsgManager<PData> {
                         to_send.push((delayed.node_id, NodeControlMsg::DelayedData { when: delayed.when, data: delayed.data }));
                     }
 
+                    // Update and report pipeline metrics
+                    pipeline_metrics_monitor.update_metrics();
+                    if let Err(err) =
+                        self.metrics_reporter.report(pipeline_metrics_monitor.metrics_mut())
+                    {
+                        // Metric reporting failures are non-fatal; continue loop.
+                        let _ = err;
```
albertlockett, line 312: It might not be fatal, but is the error something a user might want to know about? Should we consider printing a debug log here?

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
```diff
@@ -0,0 +1,580 @@
...
+//!
+//! - *Allocator churn*:
+//!   stable `memory_usage` but high `memory_allocated_delta` and
+//!   `memory_freed_delta` rates. Example: repeatedly rebuilding large buffers
+//!   per batch. Often manifests as higher p99 latency despite stable mean.
+//!
+//! - *Memory leak or unbounded buffering*:
+//!   `memory_usage` trending upward over minutes/hours, with positive
+//!   `net_allocated_delta` each interval. Helps attribute leaks to a specific
+//!   pipeline rather than the whole process.
```
albertlockett, line 110: what's nice about switching to jemalloc is that, if this happens, we could use jemalloc_pprof to profile where the memory was allocated :)
https://crates.io/crates/jemalloc_pprof

rust/otap-dataflow/benchmarks/benches/exporter/main.rs
```diff
@@ -67,13 +67,12 @@ use otap_df_telemetry::MetricsSystem;
...
 use std::sync::Arc;
+use tikv_jemallocator::Jemalloc;
 use tokio_stream::Stream;
 use tokio_stream::wrappers::ReceiverStream;
 
-use mimalloc_rust::GlobalMiMalloc;
-
 #[global_allocator]
-static GLOBAL: GlobalMiMalloc = GlobalMiMalloc;
+static GLOBAL: Jemalloc = Jemalloc;
```
lalitb, line 75: Small thought (non-blocking): since df_engine is switching from mimalloc to jemalloc for per-thread memory metrics, have we considered feature-gating the allocator choice? Users who preferred mimalloc (or want the system allocator) could opt for those, accepting that per-thread metrics are only available with jemalloc.

albertlockett (approved):

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
```diff
@@ -0,0 +1,580 @@
...
+    #[metric(unit = "{s}")]
+    pub uptime: Gauge<f64>,
+
+    /// The amount of heap memory in use by the pipeline instance running on a specific core.
+    #[metric(unit = "{By}")]
+    pub memory_usage: Gauge<u64>,
+
+    /// Memory allocated to the heap by the pipeline instance since start.
+    #[metric(unit = "{By}")]
+    pub memory_allocated: Gauge<u64>,
```
cijothomas, line 168: shouldn't this be a Counter?

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
cijothomas: why updown? This is the total since start, so it should be counter.

rust/otap-dataflow/crates/telemetry/src/metrics/dispatcher.rs
```diff
@@ -114,61 +115,125 @@ impl MetricsDispatcher {
...
+                };
+                counter.add(v, attributes);
+            }
+            MetricValueType::F64 => {
+                let counter = meter
+                    .f64_counter(field.name)
+                    .with_description(field.brief)
+                    .with_unit(field.unit)
+                    .build();
+                counter.add(value.to_f64(), attributes);
```
lalitb, line 141: Not blocking, as this could be more of correctness issue in existing code (if I understand it correctly) - The MetricsRegistry maintains a cumulative total, but counter.add() expects a delta (increment). Passing the total would result in double-counting on every flush.

rust/otap-dataflow/crates/telemetry/src/metrics/dispatcher.rs
```diff
@@ -114,61 +115,125 @@ impl MetricsDispatcher {
...
-            .build();
-        counter.add(value, attributes);
+        match field.value_type {
+            MetricValueType::U64 => {
+                let counter = meter
+                    .u64_counter(field.name)
+                    .with_description(field.brief)
+                    .with_unit(field.unit)
+                    .build();
+                let v = match value {
```
cijothomas, line 129: in this match block, we already know its U64. Why match again?

rust/otap-dataflow/benchmarks/benches/exporter/main.rs
lquerel: This PR already contains this capability :-)
I added `jemalloc` and `mimalloc` features in the main `Cargo.toml` file (default on `jemalloc`).

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
lquerel: Here we have three people who have been contributing to and using OpenTelemetry for quite some time, and we are arriving at three different conclusions (Joshua: UpDownCounter, Laurent: Gauge, Cijo: Counter). That is probably a sign that something is wrong with the OpenTelemetry metrics model. 

The semantic conventions folks seem to have decided that it should be an UpDownCounter: 
https://opentelemetry.io/docs/specs/semconv/system/process-metrics/#metric-processmemoryusage

For now, I will go with an `UpDownCounter` since a consensus seems to be emerging, and I will add the concept of temporality in the annotation on the field representing this metric to specify that, in this specific case, I do not want to report a delta but a cumulative value since the start of the measurement. In the longer term, I would really like us to have a metrics API that is simply obvious and self-descriptive for users.

rust/otap-dataflow/benchmarks/benches/exporter/main.rs
lalitb: Yes, however `rust/otap-dataflow/src/main.rs` hardcodes `jemalloc` as the global allocator, which makes the allocator choice non-configurable and requires code changes to switch.

lquerel: Haha interesting. A commit of last minute that I didn't push!

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
lquerel: See the previous discussion. That will be changed into an UpDownCounter to follow the semantic convention reco.

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
cijothomas: I am not sure why there is a confusion. The linked one from OTel semantic conventions is using `UpDownCounter` because it is "**The amount of physical memory in use.**", which can go up or down.

The one I commented about is **"Memory allocated to the heap by the pipeline instance _since start_"**, which can never go down, so it should be counter.
OTel convention also uses `Counter` for such metrics like context_switches_since_start etc.

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
cijothomas: Replied there too https://github.com/open-telemetry/otel-arrow/pull/1606#discussion_r2615829732

rust/otap-dataflow/crates/telemetry-macros/src/lib.rs
```diff
@@ -119,27 +120,53 @@ pub fn derive_metric_set_handler(input: TokenStream) -> TokenStream {
...
+                                    }
+                                    Some(syn::GenericArgument::Type(syn::Type::Path(p)))
+                                        if p.path.is_ident("f64") =>
+                                    {
+                                        quote!(otap_df_telemetry::descriptor::MetricValueType::F64)
+                                    }
+                                    _ => {
+                                        return syn::Error::new(
+                                            seg.ident.span(),
+                                            "Metric field type must be one of Counter<u64>, Counter<f64>, UpDownCounter<u64|f64>, Gauge<u64|f64>",
```
cijothomas, line 153:
```suggestion
                                            "Metric field type must be one of Counter<u64|f64>, UpDownCounter<u64|f64>, Gauge<u64|f64>",
```

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
```diff
@@ -0,0 +1,580 @@
...
+    /// Reported as a ratio in the range [0, 1].
+    #[metric(unit = "{1}")]
+    pub cpu_utilization: Gauge<f64>,
+
+    /// Number of times the pipeline thread voluntarily yielded the CPU since the last update.
+    ///
+    /// A rising rate can indicate scheduling pressure due to blocking work
+    /// (I/O, locks, backpressure) and helps distinguish "waiting" from "CPU-bound".
+    #[metric(unit = "{1}")]
+    pub context_switches_voluntary_delta: Counter<u64>,
```
cijothomas, line 196: unsure why we report "delta" here... We can report the absolute number since the beginning...Having "delta" in metric names is unusual in OTel world from what I've observed.

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
lquerel: The code originally selected by Joshua contained three metrics (`memory_usage`, `memory_allocated`, and `memory_freed`), which is where the confusion came from. I thought you were referring to `memory_usage`, which can go up or down. If you were referring only to the latter two, then I agree with you that they should be counters.

And my general comment remains the same: three people, three opinions, including for the last two metrics.

lquerel:
@lalitb , @cijothomas , @albertlockett , @jmacd FYI, I was in the process of taking everyone's feedback into account. The PR was merged by mistake, so I will carry over my changes and improvements in a follow up PR.

PR #1608, opened 2025-12-12
Use temp_dir() in tests as an OS-agnostic path

cbandy:
Tests are failing on Windows because `LocalFileSystem` requires an existing path, but `/tmp` does not exist there.

See: https://github.com/open-telemetry/otel-arrow/pull/1517

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/object_store.rs
+++ b/rust/otap-dataflow/crates/otap/src/object_store.rs
@@ -211,16 +211,18 @@ mod test {
 
     #[test]
     fn test_get_testdelayed_file_storage() {
+        let tmp = std::env::temp_dir().display().to_string();
         let storage = StorageType::File {
-            base_uri: "testdelayed:///tmp".to_string(),
+            base_uri: format!("testdelayed://{tmp}").to_string(),
         };
         assert!(from_storage_type(&storage).is_ok());
     }
 
     #[test]
     fn test_get_file_storage() {
+        let tmp = std::env::temp_dir().display().to_string();
         let storage = StorageType::File {
-            base_uri: "/tmp".to_string(),
+            base_uri: tmp,
         };
         assert!(from_storage_type(&storage).is_ok());
     }


cbandy:
I don't have a Windows machine to test this. 🤞🏻

rust/otap-dataflow/crates/otap/src/object_store.rs
```diff
@@ -211,16 +211,18 @@ mod test {
 
     #[test]
     fn test_get_testdelayed_file_storage() {
+        let tmp = std::env::temp_dir().display().to_string();
         let storage = StorageType::File {
-            base_uri: "testdelayed:///tmp".to_string(),
+            base_uri: format!("testdelayed://{tmp}").to_string(),
         };
         assert!(from_storage_type(&storage).is_ok());
     }
```
albertlockett, line 219: Hey @cbandy  -- thanks for helping us get windows CI fixed!

I don't think this fix will work for this test

We actually tried a similar solution in https://github.com/open-telemetry/otel-arrow/pull/1594 (see this [24e7811b](https://github.com/open-telemetry/otel-arrow/pull/1594/commits/24e7811b5ebdea82e5eddb13f527b43c6c9826f3)) and the test still failed:
https://github.com/open-telemetry/otel-arrow/actions/runs/20145683833/job/57825280423

I think the reason this still fails is is that the base_uri for the "delayed" object store gets parsed as a URL:
https://github.com/open-telemetry/otel-arrow/blob/5b181894a6a397460b94a40490da5950fc9e58c0/rust/otap-dataflow/crates/otap/src/object_store.rs#L110-L133 and `std::env::temp_dir()` returns something that doesn't parse correctly. 

Unforuntately, I don't have a windows machine either, so I can't definitively confirm this.

Admittedly, parsing the base_uri as a URL was a bit of a hack that we can probably do away with now since the object store config was refactored in https://github.com/open-telemetry/otel-arrow/pull/1517/files 

Anyway, in https://github.com/open-telemetry/otel-arrow/pull/1594 we just decided to ignore these tests on windows for now.

cbandy: Excellent! Thanks!

PR #1609, opened 2025-12-12
Update docs and labeler config since columnar query engine relocated

albertlockett:
updates the documentation for columnar query engine since it was relocated in #1580 and updates the relabeler config


albertlockett:
closing as I used the wrong source branch

PR #1616, opened 2025-12-12
[otap-df-quiver] WAL Refactoring

AaronRM:
- Updated terminology from "checkpoint" to "cursor" throughout the WAL writer implementation for clarity.
- Update WAL Header and Cursor (formerly 'checkpoint') file formats to use variable length headers for better forward compatibility

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/wal/cursor_sidecar.rs
+++ b/rust/otap-dataflow/crates/quiver/src/wal/cursor_sidecar.rs
@@ -66,8 +66,7 @@ impl CursorSidecar {
         let mut buf = vec![0u8; size];
         let mut cursor = 0;
 
-        buf[cursor..cursor + CURSOR_SIDECAR_MAGIC.len()]
-            .copy_from_slice(CURSOR_SIDECAR_MAGIC);
+        buf[cursor..cursor + CURSOR_SIDECAR_MAGIC.len()].copy_from_slice(CURSOR_SIDECAR_MAGIC);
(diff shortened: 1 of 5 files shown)


rust/otap-dataflow/crates/quiver/src/wal/cursor_sidecar.rs
```diff
@@ -0,0 +1,346 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Crash-safe WAL cursor persistence.
```
jmacd, line 4: (For next time!) I suggest renaming this with "git mv", it'll show the diffs more clearly.

rust/otap-dataflow/crates/quiver/src/wal/cursor_sidecar.rs
```diff
@@ -0,0 +1,346 @@
...
+//!
+//! The cursor sidecar is a tiny file (`quiver.wal.cursor`) that tracks how
+//! much of the WAL has been durably consumed. It survives crashes so restarts
+//! can resume from the last known safe position without rescanning the entire log.
+//!
+//! # Format (v1, 24 bytes, variable-width support)
+//!
+//! ```text
+//! ┌────────────┬─────────┬──────────┬─────────────────┬──────────┐
+//! │ magic (8)  │ ver (2) │ size (2) │ wal_position (8)│ crc (4)  │
```
jmacd, line 14: Added size sounds like a safe plan for the future, thanks.

PR #1617, opened 2025-12-12
Use #[cfg(unix)] for Unix-like OS detection

AaronRM:
Update the platform-specific configuration for enabling the `set_reuse_port` socket option from only Linux and macOS to all Unix-like systems, which improves portability.


rust/otap-dataflow/crates/engine/src/effect_handler.rs
```diff
@@ -104,7 +104,7 @@ impl<PData> EffectHandlerCore<PData> {
         // (load balancing).
         // Goal: Load balancing incoming connections.
         // TODO: Investigate adding set_reuse_port support for Windows.
-        #[cfg(any(target_os = "linux", target_os = "macos"))]
+        #[cfg(unix)]
```
lalitb, line 107: Nit - the SO_REUSEPORT behavior isn’t consistent across all Unix systems, particularly around load balancing. One FreeBSD, the semantics is  different - https://gitlab.isc.org/isc-projects/bind9/-/merge_requests/3365

But then, the issue existed even before we added linux and macOS condition, so not blocking :)

albertlockett (approved):

rust/otap-dataflow/crates/engine/src/effect_handler.rs
jmacd: When users are asking for FreeBSD support, we'll have a nice problem!

PR #1619, opened 2025-12-13
Try speed up benchmark prep steps

cijothomas:
3.5 minutes spend on preparing the dedicated machine. Attempting to optimize this.
If this helps, I'll move this to a common script and re-use in all workflows. (we have 4 now, but likely to add more)


cijothomas:
@jmacd can you add "pipelineperf" label, so we can verify this works!

cijothomas:
saving 2-3 mins per run now!

PR #1626, opened 2025-12-15
feat:  Add TLS/mTLS support for OTLP and OTAP exporters

lalitb:
Adds client-side TLS and mTLS support for OTLP and OTAP exporters, enabling secure communication with downstream collectors. 

## Changes:

  - Extended existing `experimental-tls` feature flag to support exporters (previously only covered receivers)
  - Implemented `build_endpoint_with_tls()` in `GrpcClientSettings` to handle client-side TLS configuration
  - “Added `load_client_tls_config()` in `tls_utils.rs` to construct client TLS/mTLS config (trust roots + client identity) with Go-aligned `insecure` semantics; `insecure_skip_verify=true` is rejected (fail-fast) as of now.”
  - Updated both OTLP and OTAP exporters to build TLS-enabled endpoints

 ## Configuration

TLS is scheme-driven by default: `https://` enables TLS (system roots), `http://` is plaintext when `tls` is omitted. The optional `tls` block adds trust roots and client identity (mTLS). Note: `insecure: true` with no custom CA results in no explicit TLS config; the endpoint scheme still decides TLS vs plaintext.

```yaml
  exporter:
    config:
      grpc_endpoint: "https://backend:4317"
      tls:
        ca_file: "/path/to/ca.crt"              # Server trust
        cert_file: "/path/to/client.crt"       # Client identity (mTLS)
        key_file: "/path/to/client.key"
        include_system_ca_certs_pool: true     # Default: true
```

 ## Testing

  - Added integration tests in `crates/otap/tests/otlp_exporter_tls.rs`

## Known Limitations

Certificate hot reload is not included in this PR. This applies to:

- Client Identity: Client certificate and key (mTLS).
- Trust Anchors: Custom CA certificates (ca_file) and System CA certificates.

If any of these files change (e.g., certificate rotation or updating the CA bundle), the process must be restarted to pick up the changes.

While the Go OpenTelemetry Collector's exporters support hot reload for client certificates via periodic polling, implementing this for our Rust exporters is complex:
  - Requires either recreating the gRPC channel (may disrupt in-flight requests) or implementing a custom TLS connector with lazy certificate loading.
  - Unlike receivers which use LazyReloadableCertResolver, exporters would need significant integration work with tonic's transport layer.

  See detailed explanation in `crates/otap/src/tls_utils.rs:115-130`. Will create a follow-up issue to track hot reload support, and can be implemented if this becomes an operational requirement.

## Client-side TLS config (OTLP exporter) — Go vs Rust

| Field | Config key | Go behavior | Rust behavior | Notes |
|-------|------------|-------------|---------------|-------|
| Plaintext vs TLS | (scheme-driven) | `https://` → TLS, `http://` → plaintext | Same | gRPC dialing; scheme decides |
| `insecure` | `insecure` | If true + no CA, TLS config is nil; scheme still decides | Same | Go's intended semantics — "insecure" isn't a guaranteed plaintext switch |
| `insecure` + custom CA | `insecure` + `ca_file`/`ca_pem` | TLS still built when CA exists | Same | Surprises people |
| Skip cert verification | `insecure_skip_verify` | Supported | Not supported (fails fast) | **Major difference** |
| SNI override | `server_name_override` | Sets TLS ServerName | Same (stored as `server_name` internally) | Same key |
| Custom root CAs | `ca_file`, `ca_pem` | Adds trust roots | Same | — |
| Client cert (mTLS) | `cert_file`, `key_file` | Presents client cert | Same | — |
| No TLS block | — | Scheme-driven | Same | gRPC dialing; `https://` → TLS, `http://` → plaintext |

**TL;DR:** Behavior matches except `insecure_skip_verify` — Go supports it, Rust currently fails fast if set.

### DIFF
--- a/.github/workflows/pipeline-perf-test-continuous.yml
+++ b/.github/workflows/pipeline-perf-test-continuous.yml
@@ -7,11 +7,6 @@ on:
         branches: [ main ]
     workflow_dispatch:
 
-# Cancel in-progress runs on new commits to same PR
-concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
-
 jobs:
     pipeline-perf-test:
         runs-on: oracle-bare-metal-64cpu-512gb-x86-64
@@ -139,33 +134,11 @@ jobs:
 
           - name: Consolidate pipeline benchmark data
             run: |
-              echo "Consolidating pipeline benchmark JSON files..."
-              find results-pipeline -name "*.json" -type f | while read file; do
-                echo "Processing: $file"
-                cat "$file"
-                echo
-              done
-
-              # Combine all benchmark JSON files into a single output (find them recursively)
-              find results-pipeline -name "*.json" -type f -exec cat {} \; | jq -s 'map(.[])' > output-pipeline.json
-
-              echo "Consolidated pipeline benchmark data:"
-              cat output-pipeline.json
+              bash ./.github/workflows/scripts/consolidate-benchmarks.sh results-pipeline output-pipeline.json
 
           - name: Consolidate saturation benchmark data
             run: |
-              echo "Consolidating saturation benchmark JSON files..."
-              find results-saturation -name "*.json" -type f | while read file; do
-                echo "Processing: $file"
-                cat "$file"
-                echo
-              done
-
-              # Combine all benchmark JSON files into a single output (find them recursively)
-              find results-saturation -name "*.json" -type f -exec cat {} \; | jq -s 'map(.[])' > output-saturation.json
-
-              echo "Consolidated saturation benchmark data:"
-              cat output-saturation.json
+              bash ./.github/workflows/scripts/consolidate-benchmarks.sh results-saturation output-saturation.json
 
           - name: Update pipeline benchmark data and deploy to GitHub Pages
             uses: benchmark-action/github-action-benchmark@4bdcce38c94cec68da58d012ac24b7b1155efe8b # v1.20.7
--- a/.github/workflows/pipeline-perf-test-nightly.yml
+++ b/.github/workflows/pipeline-perf-test-nightly.yml
@@ -7,11 +7,6 @@ on:
       - cron: '0 10 * * *'  # runs every day at 10 AM UTC
     workflow_dispatch:
 
-# Cancel in-progress runs on new commits to same PR
-concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
-
 jobs:
     pipeline-perf-test:
         runs-on: oracle-bare-metal-64cpu-512gb-x86-64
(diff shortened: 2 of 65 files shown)


rust/otap-dataflow/crates/config/src/tls.rs
```diff
@@ -53,7 +53,26 @@ fn default_reload_interval() -> Option<Duration> {
...
+    /// Controls whether system CA certificates are loaded for server certificate verification.
+    ///
+    /// Defaults to true when omitted.
+    pub include_system_ca_certs_pool: Option<bool>,
+
+    /// Override the server name used for TLS SNI and certificate verification.
+    ///
+    /// This is typically needed when connecting by IP address or when the endpoint hostname
+    /// does not match the certificate.
+    pub server_name_override: Option<String>,
```
jmacd: Please review the corresponding Collector configuration struct:

https://github.com/open-telemetry/opentelemetry-collector/blob/b31ddb9382988b767c6c7226878c1e3504e4e3e1/config/configtls/configtls.go#L103

and please make sure for each of these fields, we use the identical name and meaning.

I think `server_name_override` should be just `server_name`.

The most important flag I would ask for is `insecure`. I think the system should not look at "https://" vs "http://`, would prefer it follow the collector as secure by default--it means requiring `insecure: true` in local development.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -85,6 +94,16 @@ fn convert_native_certs_to_pem(cert_res: &rustls_native_certs::CertificateResult
...
 
+/// Returns true if the given URI uses the HTTPS scheme.
+///
+/// Trims leading whitespace before checking to handle common configuration errors.
+/// Performs case-insensitive comparison per RFC 3986 (scheme is case-insensitive).
+pub(crate) fn is_https_endpoint(uri: &str) -> bool {
+    uri.trim_start()
+        .get(..8)
+        .is_some_and(|s| s.eq_ignore_ascii_case("https://"))
+}
```
jmacd: Asking to implement `insecure: bool` instead, default false. The user has to explicitly state that they do not want a secure connection. https://github.com/open-telemetry/opentelemetry-collector/blob/b31ddb9382988b767c6c7226878c1e3504e4e3e1/config/configtls/configtls.go#L103

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -4,12 +4,25 @@
...
 use tonic::codec::CompressionEncoding;
 use tonic::transport::Endpoint;
 
+/// Checks if a URI string starts with "https://" (case-insensitive, per RFC 3986).
+fn is_https_endpoint(uri: &str) -> bool {
+    uri.trim_start()
+        .get(..8)
+        .is_some_and(|s| s.eq_ignore_ascii_case("https://"))
+}
+
```
jmacd: I think this is duplicated, I've seen two of them, moreover I do not want to infer security settings from the URL scheme. (I know this is something OTel SDKs do! This is not how the OTel Collector operates, nor is it part of the OTel declarative configuration as far as I know.)

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -199,9 +271,16 @@ const fn default_keep_alive_while_idle() -> bool {
 }
 
 #[cfg(test)]
+#[allow(missing_docs)]
 mod tests {
     use super::*;
 
+    #[cfg(feature = "experimental-tls")]
```
jmacd, line 277: I have a sense that users will always want TLS and that it might cost us in development velocity to maintain it as a feature. Should we include TLS immediately? (I think yes!)

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -199,9 +271,16 @@ const fn default_keep_alive_while_idle() -> bool {
...
+    use otap_df_config::tls::{TlsClientConfig, TlsConfig};
+
+    #[cfg(feature = "experimental-tls")]
+    use tempfile::NamedTempFile;
```
jmacd: Please comment why we're using a named temp file w/ security implications.

rust/otap-dataflow/crates/otap/tests/otlp_exporter_tls.rs
```diff
@@ -0,0 +1,349 @@
...
+    let _ = client.export(Bytes::from(buf)).await.unwrap();
+
+    // Server should observe a message.
+    let observed = tokio::time::timeout(std::time::Duration::from_secs(5), rx.recv())
+        .await
+        .unwrap();
+    assert!(observed.is_some());
+
+    server.abort();
+}
```
jmacd, line 345: Good tests 👍

rust/otap-dataflow/crates/config/src/tls.rs
lalitb: Thanks, valid comments - reviewed the Go reference (`configtls.ClientConfig`) and updated the Rust OTLP exporter/client TLS config to align on names and (as much as possible) semantics.

Where we landed for **gRPC client/exporter**, to match Go collector behavior:

- Default (no `insecure` / `insecure: false`, no `tls: block`): scheme-driven. `https://` → TLS with system roots; `http://` → plaintext.
- `tls:` block present (no `insecure`): TLS is enabled via explicit config, regardless of scheme.
- `insecure: true` (no CA configured): no explicit TLS config is applied; scheme still decides transport. This matches Go’s intended behavior - insecure is not a guaranteed “force plaintext” switch.
- `insecure: true` + custom CA (`ca_file`/`ca_pem`): TLS is still built using the configured CA; `insecure` does not override a provided CA.
- `insecure_skip_verify`: Go supports it and will skip server certificate verification when set. Rust does not support this today (tonic `ClientTlsConfig` can’t inject a custom rustls verifier). If `insecure_skip_verify: true` is set in Rust, startup fails with a configuration error (it will not silently proceed with normal verification).
- `server_name_override`, `ca_file`/`ca_pem`, `cert_file`/`key_file`: same intent and config keys as Go (Rust stores `server_name_override` internally as `server_name`).

Added tests to validate these, and also updated the PR desc.

If the goal is: “treat endpoints as TLS by default even when the URL is http:// (and require insecure: true to allow plaintext for local dev)”, that is a different behavior than the Go collector’s gRPC dialing today, which is scheme-driven (https:// → TLS, http:// → plaintext). Rust can be changed to that stricter default, but it would no longer be an “identical meaning” match with the collector’s current behavior.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
lalitb: Valid point, implemented as mentioned [here](https://github.com/open-telemetry/otel-arrow/pull/1626#discussion_r2622080968).

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
lalitb: this is no more duplicate now. Thanks,

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
lalitb: I agree.  I can take this on as a separate PR, since this applies to receivers as well (and we should make a deliberate, consistent choice across both client + server).

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
lalitb: This is used only in tests to create a real on-disk file so we can validate the *_file code paths and size limits. It has no production impact.

jmacd (approved):

PR #1632, opened 2025-12-16
Pipeline metrics - Follow up of #1606

lquerel:
So far, we had hot-path support for **Counter**, **UpDownCounter**, and **Gauge** metrics. These metrics are defined using macros that specify the name, description, and unit, which lets us automatically generate a semantic convention registry for our engine and for all pipeline components (receiver, processor, exporter).

Some of these metrics are fundamentally **cumulative counters** or **cumulative up-down counters** (for example, memory usage). In this PR, I therefore split our `Counter` and `UpDownCounter` types into:

* **DeltaCounter / DeltaUpDownCounter**
* **ObserveCounter / ObserveUpDownCounter**

With the following API:

* delta counters and delta up-down counters expose an `add` method
* observe counters and observe up-down counters expose an `observe` method
* gauges expose a `set` method

This, I believe, aligns with the OTel intrumentation API model.

Now I'm hitting a problem: I haven't found a way to automatically represent these **cumulative** metrics without going through an SDK exporter configuration file. I put a temporary workaround in `dispatcher.rs`, but I’d like a better and more accurate solution to connect the Rust Client SDK with these internal metrics.

### DIFF
--- a/rust/experimental/query_engine/engine-recordset/src/engine.rs
+++ b/rust/experimental/query_engine/engine-recordset/src/engine.rs
@@ -22,6 +22,8 @@ use crate::{
 pub struct RecordSetEngineOptions {
     pub(crate) diagnostic_level: RecordSetEngineDiagnosticLevel,
     pub(crate) summary_cardinality_limit: usize,
+    pub(crate) external_function_implementations:
+        HashMap<Box<str>, Box<dyn RecordSetEngineFunctionCallback>>,
 }
 
 impl Default for RecordSetEngineOptions {
@@ -35,6 +37,7 @@ impl RecordSetEngineOptions {
         Self {
             diagnostic_level: RecordSetEngineDiagnosticLevel::Warn,
             summary_cardinality_limit: 8192,
+            external_function_implementations: HashMap::new(),
         }
     }
 
@@ -53,11 +56,22 @@ impl RecordSetEngineOptions {
         self.summary_cardinality_limit = summary_cardinality_limit;
         self
     }
+
+    pub fn with_external_function_implementation<F: RecordSetEngineFunctionCallback + 'static>(
+        mut self,
+        name: &str,
+        callback: F,
+    ) -> RecordSetEngineOptions {
+        self.external_function_implementations
+            .insert(name.into(), Box::new(callback));
+        self
+    }
 }
 
 pub struct RecordSetEngine {
     diagnostic_level: RecordSetEngineDiagnosticLevel,
     summary_cardinality_limit: usize,
+    external_function_implementations: HashMap<Box<str>, Box<dyn RecordSetEngineFunctionCallback>>,
 }
 
 impl Default for RecordSetEngine {
@@ -75,6 +89,7 @@ impl RecordSetEngine {
         Self {
             diagnostic_level: options.diagnostic_level,
             summary_cardinality_limit: options.summary_cardinality_limit,
+            external_function_implementations: options.external_function_implementations,
         }
     }
 
@@ -121,6 +136,7 @@ impl<'a, 'b, TRecord: Record + 'static> RecordSetEngineBatch<'a, 'b, TRecord> {
 
         let execution_context = ExecutionContext::<TRecord>::new(
             self.engine.diagnostic_level.clone(),
+            &self.engine.external_function_implementations,
             self.pipeline,
             &self.global_variables,
             &self.summaries,
@@ -186,6 +202,7 @@ impl<'a, 'b, TRecord: Record + 'static> RecordSetEngineBatch<'a, 'b, TRecord> {
             self.diagnostics,
             process_summaries(
                 self.engine.diagnostic_level.clone(),
+                &self.engine.external_function_implementations,
                 &self.global_variables,
                 self.pipeline,
                 &self.summaries,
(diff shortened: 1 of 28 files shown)


lalitb:
Not entirely related to this PR, but seems there is a bug in code affecting `DeltaCounter` and `DeltaUpDownCounter` metrics - `dispatcher.rs::visit_current_metric()` never resets the registry. Since the registry accumulates the delta values, the dispatcher exports ever growing totals instead of per-interval delta.

lquerel:
> Not entirely related to this PR, but seems there is a bug in code affecting `DeltaCounter` and `DeltaUpDownCounter` metrics - `dispatcher.rs::visit_current_metric()` never resets the registry. Since the registry accumulates the delta values, the dispatcher exports ever growing totals instead of per-interval delta.

@lalitb Calling `visit_metrics_and_reset` instead of `visit_current_metric` in the dispatcher.rs code should fix the issue. I will make this change.

rust/otap-dataflow/benchmarks/Cargo.toml
```diff
@@ -39,6 +38,9 @@ unsync.workspace = true
 portpicker.workspace = true
 tokio-stream.workspace = true
 
+[target.'cfg(not(windows))'.dev-dependencies]
```
jmacd, line 41: TIL

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
```diff
@@ -18,7 +18,7 @@
 //!   Wall-clock time since this pipeline instance started. Useful for detecting
 //!   restarts, warm-up effects, and aligning deltas to a known lifetime.
 //!
-//! - `cpu_time` (`Counter<f64>`, `{s}`):
+//! - `cpu_time` (`DeltaCounter<f64>`, `{s}`):
```
jmacd: For the record (not to block or change) I think you could keep `Counter` as meaning delta counter and use `ObserveCounter` as the cumulative form, likewise for UpDown forms. I think we can have this added in the OTel specification at some point in the future, and these names reflect my interpretation of the OTel convention. Whereas you Add() (and Sub()) a Counter (and UpDownCounter) you always Set() an Observe(UpDown)Counter, like the synchronous Gauge.

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
```diff
@@ -273,28 +276,24 @@ impl PipelineMetricsMonitor {
...
+        };
 
-        let (
-            rusage_thread_supported,
-            last_voluntary_context_switches,
-            last_involuntary_context_switches,
-            last_minor_page_faults,
-            last_major_page_faults,
-        ) = Self::init_rusage_baseline();
+        #[cfg(windows)]
```
jmacd, line 282: Thank you for doing this work.

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
jmacd: However, I think it would reduce the size of the PR and generally impact on other open branches, maybe, to return to Counter and UpDownCounter, removing "Delta". Our goal is that the method name corresponds with the semantic intention: delta instruments only support Add(), cumulative instruments only support Set().

jmacd (approved):

lquerel:
@jmacd I agree, I will remove the prefix Delta to reduce the size of this PR and reduce friction with other ongoing PRs.

rust/otap-dataflow/crates/engine/src/pipeline_metrics.rs
lquerel: I removed the `Delta` prefix for counters and up-down counters.

rust/otap-dataflow/crates/telemetry/src/metrics/dispatcher.rs
```diff
@@ -57,7 +57,7 @@ impl MetricsDispatcher {
 
     fn dispatch_metrics(&self) -> Result<(), Error> {
         self.metrics_handler
-            .visit_current_metrics(|descriptor, attributes, metrics_iter| {
+            .visit_metrics_and_reset(|descriptor, attributes, metrics_iter| {
```
cijothomas, line 60: This will break the existing admin/prometheus metrics...

https://github.com/open-telemetry/otel-arrow/pull/1552

cijothomas: Due to OTel not enabled in perf test (https://github.com/open-telemetry/otel-arrow/pull/1555/changes), this should be okay.
I'll open an issue to track fixing this - as of now, the admin/prometheus and otel-metrics both cannot be enabled at same time without interference.

lquerel: I've just seen your comment after this PR was merged. I'll open another PR to take your feedback into account.
However, I'm still not sure I fully understand how to use the client SDK with "observe" counters and up-down counters. Do you have any recommendations on how to approach this?

jmacd: If this turns out to be difficult, we should use Gauge points for now! Call the object ObserveCounter and ObserveUpDownCounter though :-)

PR #1633, opened 2025-12-16
[batch_processor] Support bytes-based batching via new `format = [otap|otlp|preserve]`

### LINKED ISSUE #1570: Batch processor: payload format support
The batch processor currently forces all of its inputs into the OtapArrowRecords format, batches in terms of OTAP structures, and outputs the same format. For users sending OTLP inputs, batching, and sending OTLP outputs, this will be inefficient. While Arrow batching may yield more compressed batches, this is not the simplest way to batch OTLP data.

Since we have an OTLP bytes representation, we should support batching in the OTLP bytes representation. 

Proposing a new `format` field in the batch_processor configuration, with values:

- "otap": current behavior batches in OTAP format (default)
- "preserve": new behavior batches in the arriving format
- "otlp": new behavior batches in OTLP format

The new options will enable batching OTLP bytes directly. Fortunately, batching OTLP bytes is trivial, we simply concatenate the arrays. In the OTLP batching mode we will not support splitting, only merging.

jmacd:
Fixes #1570.

Adds dual format configuration to batch processor, with separate `FormatConfig` structs for each payload format.
This supports forcing payload into one or the other format, or allowing both to be preserved.

The new bytes-based batching routines operate by scanning through top-level fields. Unlike the items-based batching mode, this may produce batches that are less than the limit; like that mode, it can also produce outputs greater than the limit.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/batch_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/batch_processor.rs
@@ -341,7 +341,7 @@ impl FormatConfig {
             // indefinite delay.
             if self.min_size.is_some() {
                 return Err(ConfigError::InvalidUserConfig {
-                    error: format!("min_size set requires flush_timeout is set"),
+                    error: "min_size set requires flush_timeout is set".into(),
                 });
             }
             // If max_size is unset, we're doing nothing with a batch processor,
@@ -1555,7 +1555,7 @@ mod tests {
             format: BatchingFormat::Preserve,
             ..Default::default()
         };
-        assert!(!cfg.validate().is_ok());
+        assert!(cfg.validate().is_err());
 
         // Both None: ERROR
         let cfg = Config {


rust/otap-dataflow/configs/fake-batch-debug-noop.yaml
```diff
@@ -31,7 +31,9 @@ nodes:
     config:
       # At the rate configured above, this will print a batch
       # every 10 seconds.
-      min_size: 1000
+      otap:
```
cijothomas, line 32: the comment right above this does not make sense.. is that to be fixed?

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -176,29 +310,33 @@ impl Config {
         }
 
         // Presently we have only the OTAP mode of batching, which supports only Items.
-        if self.sizer != Sizer::Items {
+        let (expect_sizer, with_msg) = match format {
+            SignalFormat::OtapRecords => (Sizer::Items, "OTAP batch sizer: must be items"),
+            SignalFormat::OtlpBytes => (Sizer::Bytes, "OTLP batch sizer: must be bytes"),
+        };
```
cbandy: ⛏️ This comment looks outdated; there are multiple modes now.

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -176,29 +310,33 @@ impl Config {
...
                     ),
                 });
             }
         }
 
-        // Zero-timeout is a valid split-only configuration.
-        if self.timeout == Duration::ZERO {
-            if let Some(batch_size) = self.min_size {
+        // Zero-timeout is a valid split-only configuration, but must have
+        if no_timeout {
```
cbandy: ⛏️ Did part of this comment get lost? It is unclear to me what the remainder may be.

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -1149,69 +1497,86 @@ mod tests {
...
-        // Only batch size: OK
+        // Only batch size: OTLP OK (OTLP invalid OK)
         let cfg = Config {
-            min_size: NonZeroUsize::new(100),
-            max_size: None,
-            timeout: Duration::from_millis(100),
+            otap: FormatConfig::new_items(100, 0),
+            otlp: FormatConfig::new_items(0, 0),
+            flush_timeout: Duration::from_millis(100),
+            format: BatchingFormat::Otap,
```
cbandy: ⛏️ I don't understand the comment on this test. The `otlp` field appears incorrect (because items), but this situation is okay?

rust/otap-dataflow/crates/pdata/src/otlp/batching_tests.rs
```diff
@@ -0,0 +1,161 @@
...
+    let inputs_otlp: Vec<_> = inputs_otlp.collect();
+    let signal_type = inputs_otlp.get(0).expect("ok").signal_type();
+
+    let inputs_bytes: Vec<OtlpProtoBytes> = inputs_otlp.iter().map(otlp_message_to_bytes).collect();
+
+    let total_input_bytes: usize = inputs_bytes.iter().map(|b| b.num_bytes()).sum();
+
+    // Run a single equivalence test
+    let test_config = |limit: Option<NonZeroU64>, label: &str| {
+        let outputs = make_bytes_batches(signal_type, limit, inputs_bytes.clone()).expect("ok");
```
cbandy, line 36: ⛏️ I don't see an assertion that this list contains multiple values. Is it possible to pass the expected number of batches to this closure?

- no limit &rarr; 1 batch
- limit == size &rarr; 1 batch
- 10% &rarr; 10 batches?
- 50% &rarr; 2 batches?
- limit == 1 &rarr; `size` batches?

rust/otap-dataflow/crates/pdata/src/otlp/batching.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+impl Position {
+    fn new(input_idx: usize, offset: usize) -> Self {
+        Self { input_idx, offset }
+    }
+}
+
+/// Definition of the batch size in bytes
+fn batch_bytes(start_pos: &Position, end_pos: &Position, inputs: &[OtlpProtoBytes]) -> usize {
```
gouslu, line 25: can use #[inline] on this one (?)

rust/otap-dataflow/crates/pdata/src/otlp/batching.rs
```diff
@@ -0,0 +1,209 @@
...
+                    } else if current_size != 0 {
+                        // Flush the pending batch b/c the new field
+                        // does not fit.
+                        finalize_batch(signal, &inputs, &batch_start, &field_start, &mut batches);
+                        batch_start = field_start;
+                        current_size = 0;
+                    }
+
+                    // Increment size counter
+                    current_size += field_size;
```
gouslu, line 195: this code seems like it'll allow a maximum of 1 oversized field and no more, which means that max_size is not a strictly enforced limit. if field size is larger than max size that is allowed, it'll create a single item batch with a 1 oversized item. is this intentional? is there a test that validates it? if it is not intentional, then we should try to make sure that we return an error or something like that if a field size is larger than the max size.

rust/otap-dataflow/crates/pdata/src/otap/batching.rs
```diff
@@ -2,16 +2,14 @@
...
 use super::{OtapArrowRecords, error::Result, groups::RecordsGroup};
 use otap_df_config::SignalType;
 use std::num::NonZeroU64;
 
-/// Rebatch records to the appropriate size in a single pass.
-/// Returns error if not the same signal type.
-pub fn make_output_batches(
+/// Rebatch records to the appropriate size in a single pass, measured
+/// in items.  Requires all inputs have the same signal type.
+pub fn make_item_batches(
```
gouslu, line 12: no changes here from you, but I am still curious why arrow batching is done based on item count vs otlp based on bytes?

rust/otap-dataflow/crates/pdata/src/testing/round_trip.rs
```diff
@@ -27,6 +27,38 @@ pub fn otlp_to_otap(msg: &OtlpProtoMessage) -> OtapArrowRecords {
...
+    msg.encode(&mut buf).expect("encoding should not fail");
+
+    match msg {
+        OtlpProtoMessage::Logs(_) => OtlpProtoBytes::ExportLogsRequest(buf.into()),
+        OtlpProtoMessage::Metrics(_) => OtlpProtoBytes::ExportMetricsRequest(buf.into()),
+        OtlpProtoMessage::Traces(_) => OtlpProtoBytes::ExportTracesRequest(buf.into()),
+    }
+}
+
+/// Transcode a protocol message object to OTLP bytes.
```
cbandy: ⛏️ copy pasta, maybe.


```suggestion
/// Transcode OTLP bytes to a protocol message object.
```

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -152,22 +221,87 @@ const fn default_outbound_request_limit() -> NonZeroUsize {
     NonZeroUsize::new(512).expect("ok") // default_min_size/16
 }
 
+const fn default_batching_format() -> BatchingFormat {
+    BatchingFormat::Otap
```
albertlockett: I wonder if this should be `Preserve`? 

If batching OTLP Bytes is more efficient than converting to OTAP and batching arrow records, then having this as `Preserve` would mean users doing `OTLP Receiver -> Batch Processor -> OTLP Exporter` would not have to opt-in to the most efficient pipeline

albertlockett: After reading through the code, I now realize that this behaviour could cause confusion because incoming telemetry would get grouped separately based on it's format, which is somewhat unexpected if users are used to how the go collector works. Feel free to ignore this suggestion if it doesn't make sense.

rust/otap-dataflow/crates/pdata/src/otlp/batching.rs
```diff
@@ -0,0 +1,209 @@
...
+            let buf = inputs[end_pos.input_idx].as_bytes();
+            batch.extend_from_slice(&buf[..end_pos.offset]);
+        }
+    }
+
+    outputs.push(OtlpProtoBytes::new_from_bytes(signal, batch))
+}
+
+/// Combines OTLP content by concatenation of bytes. Because we have a
+/// top-level repeated field, this is precisely correct.
```
albertlockett, line 80: Makes sense. In the future, would we support stepping into the nested Resource<signal> -> Scope<Signal> -> Signal to either to get batch sizes that are more closely aligned to the max size? 

I realize that makes this significantly more complicated, but once we had that logic in place, we could also use it batch OTLP bytes for items sizer (at least for logs & traces).

rust/otap-dataflow/configs/fake-batch-debug-noop.yaml
jmacd: I moved it and added some clarity. This is about how 100 signals/sec with a batch size of 1000 equals 10s, how a 3s flush timeout actually yields 300/batch.

rust/otap-dataflow/crates/otap/src/batch_processor.rs
jmacd: Fixed, thanks!

rust/otap-dataflow/crates/pdata/src/testing/round_trip.rs
jmacd: Fixed, thanks.

rust/otap-dataflow/crates/pdata/src/otlp/batching.rs
jmacd: Right!

rust/otap-dataflow/crates/otap/src/batch_processor.rs
jmacd: I changed this to Preserve. I think users will adjust, if it's confusing we can see about how to message this in the startup logs.

rust/otap-dataflow/crates/pdata/src/otlp/batching_tests.rs
jmacd: Fixed, somewhat. I added this test but also removed the 1/10 limit which made estimating the number of batches difficult because it tests with variable top-level batch sizes.

rust/otap-dataflow/crates/otap/src/batch_processor.rs
jmacd: Sorry, replacing this with:

        // The OTLP configuration is invalid, but that's OK because
        // format is OTAP.

rust/otap-dataflow/crates/otap/src/batch_processor.rs
jmacd: Fixed!

rust/otap-dataflow/crates/pdata/src/otlp/batching.rs
jmacd: This is intentional, and I believe this is tested. This was mentioned in my PR description: the metrics batching logic for OTAP also returrns single items that are larger than the limit. This is also the question that Albert was asking about the potential to split items, as this logic only batches at the top level of the hierarchy. If we could split we could control size better, but this is the way it is in order to be simple.

rust/otap-dataflow/crates/pdata/src/otap/batching.rs
jmacd: The logic is very different in each case. There is nothing to say we couldn't use the other sizer in both cases, i.e., Arrow batch by bytes, OTLP batch by items, but this is simply more code. Even for batching OTLP bytes as we have here this is simple in the sense that it never splits a top-level item.

PR #1638, opened 2025-12-16
Adds OTAP Dataflow processor that uses columnar query engine

### LINKED ISSUE #1628: Integrate columnar query engine into otap-dataflow processor
build a processor that uses the columnar query engine to process OTAP batches

albertlockett:
closes #1628 

Adds a processor implementation to OTAP Dataflow that uses the columnar query engine to transform telemetry data.

Example config:
```yaml
nodes:
  transform:
    kind: processor
    plugin_urn: urn:otel:transform:processor
    out_ports:
      out_port:
        destinations:
        - exporter
    config:
      query: logs | where event_name == "gen_ai.system.message"
```

**I'm flexible on the name of this plugin and open to suggestions**. I hesitated between some alternative names:
- "\<_language_\> processor" - where "_language_" = KQL/OTTL/ anything that can be transformed into our expression AST. I was thinking this could eventually accept different ways to express the transform so pinning it to one language seemed too specific.
- "transform processor" - this is a good name that expresses the purpose of the processor, but it's not really API compatible with Go collector's [transform processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor) and I didn't want to cause confusion.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/transform_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/transform_processor.rs
@@ -192,7 +192,6 @@ impl Processor<OtapPdata> for TransformProcessor {
                                 processor: effect_handler.processor_id(),
                                 kind: ProcessorErrorKind::Other,
                                 error: format!("Error executing query engine pipeline {e}"),
-,
                                 source_detail: e.to_string(),
                             });
                         }
--- a/rust/otap-dataflow/crates/otap/src/transform_processor/metrics.rs
+++ b/rust/otap-dataflow/crates/otap/src/transform_processor/metrics.rs
@@ -1,7 +1,7 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-//! Metrics for the AttributesProcessor node.
+//! Metrics for the TransformProcessor node.
 
 use otap_df_telemetry::instrument::Counter;
 use otap_df_telemetry_macros::metric_set;


rust/otap-dataflow/configs/fake-queryengine-debug-noop.yaml
```diff
@@ -0,0 +1,58 @@
...
+  query_engine:
+    kind: processor
+    plugin_urn: urn:otel:queryengine:processor
+    out_ports:
+      out_port:
+        destinations:
+        - debug2
+        dispatch_strategy: round_robin
+    config:
+      program: logs | where event_name == "gen_ai.system.message"
```
cijothomas: "program" feels odd to read to me... maybe "query"?

rust/otap-dataflow/configs/fake-queryengine-debug-noop.yaml
```diff
@@ -0,0 +1,58 @@
+# Shows the batch processor.
```
cijothomas: nit: is this relevant comment?

rust/otap-dataflow/crates/otap/src/query_engine_processor/config.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::Deserialize;
+
+/// Configuration for the [`QueryEngineProcessor`](super::QueryEngineProcessor)
+#[derive(Debug, Deserialize)]
+pub struct Config {
+    /// the program that defines the transformation to be applied
+    pub program: String,
```
cijothomas: perhaps another enum config to determine the query language supported? like Enum {KQL, SQL, etc...}

cijothomas:
> "transform processor" - this is a good name that expresses the purpose of the processor, but it's not really API compatible with Go collector's [transform processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor) and I didn't want to cause confusion.

I think "transform processor" is the most apt name. Would we eventually support OTTL, and then it'd be compatible with Go transform? We can support multiple query languages, and when used in OTTL mode, it'll be compatible with Go transform processor..

rust/otap-dataflow/configs/fake-queryengine-debug-noop.yaml
albertlockett: Yeah you're right, made this change, thanks @cijothomas

rust/otap-dataflow/crates/otap/src/query_engine_processor/config.rs
albertlockett: Yes, I was planning on adding something like this when we add support for other languages. 

At present, I'm not sure a language enum would be all that's needed. OTTL for example isn't as expressive as KQL, and it also needs some identifier from outside the statement's syntax to identify the context. We might need some additional config around the OTTL statements for the processor to know what to do. Given this  ambiguity, I figured I'd defer adding this type of config until it's needed.

jmacd:
Related to https://github.com/open-telemetry/otel-arrow/pull/1642

rust/otap-dataflow/crates/otap/src/transform_processor.rs
```diff
@@ -0,0 +1,492 @@
...
+                        Ok(otap_batch) => {
+                            self.metrics.msgs_transformed.inc();
+                            otap_batch.into()
+                        }
+                        Err(e) => {
+                            self.metrics.msgs_transform_failed.inc();
+                            return Err(EngineError::ProcessorError {
+                                processor: effect_handler.processor_id(),
+                                kind: ProcessorErrorKind::Other,
+                                error: "Error executing query engine pipeline {e}".into(),
```
lalitb:
```suggestion
                                error: format!("Error executing query engine pipeline {e}"),
,
```

rust/otap-dataflow/crates/otap/src/query_engine_processor.rs
```diff
@@ -0,0 +1,492 @@
...
+                                processor: effect_handler.processor_id(),
+                                kind: ProcessorErrorKind::Other,
+                                error: "Error executing query engine pipeline {e}".into(),
+                                source_detail: e.to_string(),
+                            });
+                        }
+                    }
+                };
+
+                // TODO Ack/Nack?
```
jmacd: I would say there is only a nack requirement for cases where the request is not passed to the downstream i.e., when a new error is returned, otherwise no special Ack handling will be required here.

rust/otap-dataflow/crates/otap/src/transform_processor.rs
```diff
@@ -0,0 +1,492 @@
...
+                    payload
+                } else {
+                    let mut otap_batch: OtapArrowRecords = payload.try_into()?;
+                    otap_batch.decode_transport_optimized_ids()?;
+                    match self.pipeline.execute(otap_batch).await {
+                        Ok(otap_batch) => {
+                            self.metrics.msgs_transformed.inc();
+                            otap_batch.into()
+                        }
+                        Err(e) => {
```
lalitb, line 189: When filter matches 0 rows, does query engine error-outs and get propagated as ProcessorError? 0 matches should probably be non-error.

rust/otap-dataflow/crates/otap/src/transform_processor/metrics.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Metrics for the AttributesProcessor node.
```
lalitb: nit - copy-paste error?

albertlockett: fixed, thanks!

rust/otap-dataflow/crates/otap/src/transform_processor.rs
albertlockett: It won't produce an error if there are no matches. It will produce an empty batch. A downstream component could Ack the batch, for example as BatchProcessor does [here](https://github.com/open-telemetry/otel-arrow/blob/70e2d4934f8e0acf880b5429dcfe98433323d14d/rust/otap-dataflow/crates/otap/src/batch_processor.rs#L435-L441)

albertlockett:
@jmacd @cijothomas renamed to `transformprocessor` as suggested

rust/otap-dataflow/crates/otap/src/transform_processor.rs
lalitb: Ok thanks. My confusion is because query-engine [errors](https://github.com/open-telemetry/otel-arrow/blob/6dbec074678d2265b953223691466e41ffa0ce4b/rust/otap-dataflow/crates/query-engine/src/pipeline.rs#L121) on empty result, and this processor propagates that as processor error. But probably I am missing something. Please ignore :)

lalitb (approved):

rust/otap-dataflow/crates/otap/src/transform_processor.rs
albertlockett: oh that's a good call out! 

I should probably remove that `DataFusionPipelineStage` as it's no longer used and was mostly just there for illustrative purposes.

The reason it's not used is: it turns out that it's somewhat difficult to implement single datafusion `ExecutionPlan` to transform OTAP data (which is what `DataFusionPipelineStage` does). There are multiple arrow `RecordBatch`s associated with each OTAP batch, and processing them all together in a `ExecutionPlan` involves a lot of complicated joins. The second difficulty comes from the dynamic nature of OTAP schemas, where some columns or even certain `RecordBatch`s are optional (`ExecutionPlan` expects a more/less stable schema).

For filtering (which is currently the only operation the columnar query engine supports), we do something a bit bespoke where we use datafusion's `PhysicalExpr` to do the filtering, however choosing when/which `RecordBatch` to apply the expressions to and projecting the correct columns is all done outside of datafusion. All the code for it lives here: https://github.com/open-telemetry/otel-arrow/blob/6dbec074678d2265b953223691466e41ffa0ce4b/rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs

And to tie this back to the original discussion, we emit an empty batch if all the rows are filtered out :) 
https://github.com/open-telemetry/otel-arrow/blob/6dbec074678d2265b953223691466e41ffa0ce4b/rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs#L1605-L1613

lalitb: Thanks for the detailed explanation - that makes sense now. I was looking at the wrong code path. Appreciate the clarification.

PR #1642, opened 2025-12-16
[kql_processor] Experimental KQL processor

jmacd:
This component invokes the OTLP-bytes form of KQL query based on the `recordset` engine.

Part of #891

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/kql_processor/README.md
+++ b/rust/otap-dataflow/crates/otap/src/experimental/kql_processor/README.md
@@ -1,7 +1,18 @@
-# KQL Processor
+# EXPERIMENTAL KQL "RecordSet" Processor
 
-An experimental otap-dataflow processor that filters and transforms
-OpenTelemetry data using Kusto Query Language (KQL) expressions.
+An OTAP-Dataflow processor that filters and transforms OpenTelemetry
+data using Kusto Query Language (KQL) expressions over OpenTelemetry data
+in an opinionated way.
+
+## Caveats: Notes!
+
+The underlying engine does not use Apache Arrow or DataFusion. This
+is not an optimized code path in the OTAP Dataflow engine.
+
+The underlying engine implements a data model that is not perfectly
+aligned with OpenTelemetry. This component is capable of emitting
+arbitrary data that, while being expressed in an OTLP bytes encoding,
+is not well-specified OpenTelemetry data.
 
 ## Overview
 
@@ -15,7 +26,7 @@ and production quality however not an optimized implementation.
 
 - **Filters**: Use KQL `where` clauses to filter logs, traces, and metrics
 - **Transformations**: Apply KQL `extend` and `project` operations to modify data
-- **Aggregations**: Perform KQL `summarize` operations for data aggregation
+- **Aggregations**: Perform KQL `summarize` operations for data aggregation.
 
 ## Configuration
 


rust/otap-dataflow/Cargo.toml
```diff
@@ -171,6 +171,11 @@ reqwest = { version = "0.12.24", default-features = false }
 time = "0.3.44"
 wiremock = "0.6.5"
 
+# KQL processor
+data_engine_recordset_otlp_bridge = { path = "../experimental/query_engine/engine-recordset-otlp-bridge" }
+data_engine_recordset = { path = "../experimental/query_engine/engine-recordset" }
+data_engine_expressions = { path = "../experimental/query_engine/expressions" }
```
CodeBlanch, line 177: There are re-exports so you shouldn't need to manually reference recordset + expressions to use the bridge. Probably OK to have them if you want, just pointing that out 😄

https://github.com/open-telemetry/otel-arrow/blob/6dbec074678d2265b953223691466e41ffa0ce4b/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/lib.rs#L18-L21

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
```diff
@@ -0,0 +1,458 @@
...
+
+    async fn process_logs(
+        &mut self,
+        bytes: bytes::Bytes,
+        signal: SignalType,
+    ) -> Result<OtlpProtoBytes, Error> {
+        let (included_records, _) =
+            process_protobuf_otlp_export_logs_service_request_using_registered_pipeline(
+                self.kql_query_id,
+                RecordSetEngineDiagnosticLevel::Warn,
```
CodeBlanch, line 184: Just FYI if there are diagnostics emitted they currently get stuffed into an attribute:

https://github.com/open-telemetry/otel-arrow/blob/6dbec074678d2265b953223691466e41ffa0ce4b/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs#L289-L294

Not sure if that is fine or we may want to pull them out and push into logging?

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
```diff
@@ -0,0 +1,458 @@
...
+/// Factory function to create a KQL processor
+pub fn create_kql_processor(
+    pipeline_ctx: PipelineContext,
+    node: NodeId,
+    node_config: Arc<NodeUserConfig>,
+    processor_config: &ProcessorConfig,
+) -> Result<ProcessorWrapper<OtapPdata>, ConfigError> {
+    let config: KqlConfig = serde_json::from_value(node_config.config.clone()).map_err(|e| {
+        ConfigError::InvalidUserConfig {
+            error: format!("Failed to parse KQL configuration: {e}"),
```
andborja, line 66: nit: KQL processor configuration
(It sounds as it is validating here the KQL content)

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
```diff
@@ -0,0 +1,458 @@
...
+        let config = KqlConfig {
+            query: query.to_string(),
+        };
+
+        let processor = KqlProcessor::with_pipeline_ctx(pipeline_ctx, config);
+        assert!(
+            processor.is_ok(),
+            "Body enrichment with vendor links should parse successfully"
+        );
+    }
```
andborja, line 457: Can you add a test with the result of the processor after applied?

andborja:
LGTM

andborja (approved):

rust/otap-dataflow/crates/engine/src/processor.rs
```diff
@@ -246,7 +246,12 @@ impl<PData> ProcessorWrapper<PData> {
                     .await?;
 
                 while let Ok(msg) = message_channel.recv().await {
-                    processor.process(msg, &mut effect_handler).await?;
+                    if let Err(error) = processor.process(msg, &mut effect_handler).await {
+                        otap_df_telemetry::otel_error!(
+                            "Procsesor.Error",
+                            message = error.to_string(),
+                        );
+                    }
```
lquerel, line 254: This code no longer propagate the errors to pipeline controller and it's not great because all the pipeline status infrastructure relies on it. IMO, we should keep the code as is and then log errors at this centralize place.

rust/otap-dataflow/crates/engine/src/processor.rs
```diff
@@ -273,7 +278,12 @@ impl<PData> ProcessorWrapper<PData> {
...
```
lquerel, line 286: See my previous comment.

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/README.md
```diff
@@ -0,0 +1,69 @@
...
+An experimental otap-dataflow processor that filters and transforms
+OpenTelemetry data using Kusto Query Language (KQL) expressions.
+
+## Overview
+
+This processor integrates the experimental KQL "recordset" engine from
+`rust/experimental/query_engine` to enable powerful data
+transformations within OTAP pipelines.  This was developed as a prototype
+as we prepare for a direct column-oriented implementation, it is functional
+and production quality however not an optimized implementation.
```
lquerel, line 23: I would feel more comfortable if we explicitly stated that this processor does not use Apache Arrow or DataFusion and, as a consequence, should not be used to benchmark the OTAP engine. I do not want people to draw incorrect conclusions about our engine because they included this processor, which does not take advantage of OTAP.

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/README.md
```diff
@@ -0,0 +1,69 @@
...
+- **Filters**: Use KQL `where` clauses to filter logs, traces, and metrics
+- **Transformations**: Apply KQL `extend` and `project` operations to modify data
+- **Aggregations**: Perform KQL `summarize` operations for data aggregation
+
+## Configuration
+
+```yaml
+processors:
+  kql:
+    query: "source | where SeverityText == 'ERROR' | extend processed_time = now()"
```
lquerel, line 36: What is this `processed_time`? Is it an OTel attribute? How is it exported?

lquerel: This leads me to wonder what the goal of this processor is in the context of this project. At a minimum, a processor should consume and produce valid OTel signals, or in other words, in our engine, valid OTAP messages as input and output. That does not seem obvious to me when looking at this example.

If the goal is to produce a stream of objects that are not OTel, then instead of a processor we should have a specialized exporter using KQL to perform a transformation toward an external consumer that does not expect a pure OTel data model.

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/README.md
```diff
@@ -0,0 +1,69 @@
...
+## Examples
+
+### Filter logs by severity
+```yaml
+query: "source | where SeverityNumber >= 17"  # ERROR and above
+```
+
+### Add computed fields
+```yaml
+query: "source | extend hour = bin(Timestamp, 1h)"
```
lquerel, line 48: See my previous comment: this KQL query does not produce a valid OTel signal. It therefore should not be an OTel processor, but at best be integrated into an exporter.

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
```diff
@@ -0,0 +1,458 @@
...
+    name: KQL_PROCESSOR_URN,
+    create: create_kql_processor,
+};
+
+/// KQL processor that applies KQL queries to telemetry data
+pub struct KqlProcessor {
+    config: KqlConfig,
+
+    // Note! library uses a static for registration so all we store is a usize,
+    // TODO reconcile, otherwise memory management is unclear.
```
lquerel, line 53: Could you elaborate on the comment "memory management is unclear"? I am not sure whether the issue is related to the framework or is specific to KQL.

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/README.md
lquerel: If `processed_time` magically becomes an attribute, how do we distinguish between an attribute and a root-level signal field? How do we handle name collisions between attributes and root-level fields?

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
```diff
@@ -0,0 +1,458 @@
...
+            Message::Control(control_msg) => {
+                use otap_df_engine::control::NodeControlMsg;
+                match control_msg {
+                    NodeControlMsg::Config { config } => {
+                        if let Ok(new_config) = serde_json::from_value::<KqlConfig>(config) {
+                            // Re-parse the pipeline (if changed)
+                            if new_config.query != self.config.query {
+                                match register_pipeline_for_kql_query(&new_config.query, None) {
+                                    Ok(kql_query_id) => {
+                                        otap_df_telemetry::otel_error!(
```
gouslu, line 217: is this supposed to be error, or should it be info?

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
```diff
@@ -0,0 +1,458 @@
...
+                    "Processor.Failure",
+                    processor = "kql",
+                    input_items = input_items,
+                    message = message,
+                );
+
+                effect_handler
+                    .notify_nack(NackMsg::new(
+                        message,
+                        OtapPdata::new(ctx, OtapPayload::empty(SignalType::Logs)),
```
gouslu, line 168: is there no situation where we would need to notify_nack with data?

gouslu, line 168: Wouldn't the handling of err need to send nacks with the proper signal type? I know traces and metrics is not implemented today, but I thought it would still make sense to return appropriate signal type with the nack?

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
```diff
@@ -0,0 +1,458 @@
...
+                    ))
+                    .await?;
+                Err(e)
+            }
+        }
+    }
+
+    async fn process_logs(
```
gouslu, line 176: could this make use of unit tests?

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
```diff
@@ -0,0 +1,458 @@
...
+                                            processor = "kql",
+                                        );
+
+                                        self.kql_query_id = kql_query_id;
+                                        self.config = new_config;
+                                    }
+                                    Err(errors) => {
+                                        let message =
+                                            format!("Failed to parse updated query: {:?}", errors);
+                                        otap_df_telemetry::otel_error!(
```
gouslu, line 228: should this code path return err response?

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
gouslu, line 176: async doesn't seem to be needed.

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
```diff
@@ -0,0 +1,458 @@
...
+                self.kql_query_id,
+                RecordSetEngineDiagnosticLevel::Warn,
+                &bytes,
+            )
+            .map_err(|e| self.map_bridge_error(e, signal))?;
+
+        Ok(OtlpProtoBytes::ExportLogsRequest(included_records.into()))
+    }
+
+    fn map_bridge_error(&mut self, error: BridgeError, signal: SignalType) -> Error {
```
gouslu, line 192: can be made static

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
```diff
@@ -0,0 +1,458 @@
...
+        node,
+        node_config,
+        processor_config,
+    ))
+}
+
+impl KqlProcessor {
+    /// Creates a new KQL processor
+    pub fn with_pipeline_ctx(
+        _pipeline_ctx: PipelineContext,
```
gouslu, line 83: _pipeline_ctx is not used, is it something that you keep for the future?

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/README.md
CodeBlanch: In this case `extend processed_time = now()` is treated as though user wrote `extend source.Attributes['processed_time'] = now()`. Meaning, `processed_time` is defined on `Attributes`. The bridge create will _always_ yield valid OTel logs.

This code is using the bridge and the OTLP schema is defined in the bridge create itself:

* Special default map: https://github.com/open-telemetry/otel-arrow/blob/0b67b7783d5c2621fbffb5cbb4b74200a834d2c9/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs#L341
* Top-level fields: https://github.com/open-telemetry/otel-arrow/blob/0b67b7783d5c2621fbffb5cbb4b74200a834d2c9/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs#L342-L350

The "default map" tells the parser to use the "source.Attributes" map for anything which does not match something defined on schema. This fuzzy matching using "default map" is a feature of the KQL parser. We could add options to turn it off for this processor if it isn't desired.

PS: I say this is a parser feature because the final/actual expression tree ends up explicit with execution over (essentially) `Scalar::Source(["Attributes", "processed_time"])`. There is nothing fuzzy at runtime only when the KQL query is parsed.

rust/otap-dataflow/crates/engine/src/processor.rs
jmacd: I should not have included this piece. I am a bit confused about how to handle errors, that is all.

rust/otap-dataflow/crates/engine/src/processor.rs
jmacd: Reverting unrelated pieces in rust/otap-dataflow/crates/engine/src/processor.rs

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
jmacd: We / OpenTelemetry endorse the `tracing` crate for use with OpenTelemetry, generally, so I would recommend `tracing::debug!()` statements.

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
jmacd: Ah, the tests are definitely incomplete! Nice catch.

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/README.md
jmacd: I would say something like

```
The underlying engine does not use Apache Arrow or DataFusion. This
is not an optimized code path in the OTAP Dataflow engine.

The underlying engine implements a data model that is not perfectly
aligned with OpenTelemetry. This component is capable of emitting
arbitrary data that, while being expressed in an OTLP bytes encoding,
is not well-specified OpenTelemetry data.
```

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/README.md
jmacd: I tried to capture this sentiment in
https://github.com/open-telemetry/otel-arrow/pull/1642/commits/7f6f3c016516821d367b1598dcb3bc2a805f51e0

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
jmacd: There is a static variable in the recordset engine, and the way queries are cached is not I would say fully developed. This component is, I would say, experimental!

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/README.md
CodeBlanch: See my comment above. This is fake news. Bridge always produces valid OTel.

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
jmacd: not sure :)

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
jmacd: I wouldn't say no situation -- I was trying to avoid something unnecessayr. The exporterr components have more responsibilities forr rrerturning data in case of retries, wherreas we don't really expect to rretrry beforre the KQL processor.

jmacd: agree!

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
jmacd: I'm feeling uncertain about this question myself.

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/processor.rs
jmacd: yes! sorry for the  bare-bones take on this processor. We would use the pipeline_ctx in case we were emitting metrics.

rust/otap-dataflow/crates/otap/src/experimental/kql_processor/README.md
lquerel: I don’t know whether this is fake news... From my perspective, this kind of implicit substitution (trace_id → trace_id field, but trace.id → attributes["trace.id"]) is a perfect recipe for subtle bugs that will end up being detected and fixed by users of this language. As far as OPL is concerned, we do not want to reuse this kind of implicit transformation.

PR #1643, opened 2025-12-16
[otap-df-quiver] Segment file reader/writer for Quiver durable storage layer.

AaronRM:
Implements Segment file reader/writer for Quiver durable storage layer

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/segment/open_segment.rs
+++ b/rust/otap-dataflow/crates/quiver/src/segment/open_segment.rs
@@ -560,4 +560,37 @@ mod tests {
         let seg = OpenSegment::default();
         assert!(seg.is_empty());
     }
+
+    #[test]
+    fn append_exceeds_max_slots_per_bundle_fails() {
+        use crate::record_bundle::SlotDescriptor;
+
+        let mut seg = OpenSegment::new();
+        let schema = test_schema();
+        let fp = [0x77u8; 32];
+
+        // Create a bundle with MAX_SLOTS_PER_BUNDLE + 1 slots
+        let slot_count = MAX_SLOTS_PER_BUNDLE + 1;
+        let slots: Vec<SlotDescriptor> = (0..slot_count)
+            .map(|i| SlotDescriptor::new(SlotId::new(i as u16), format!("Slot{}", i)))
+            .collect();
+
+        let batch = make_batch(&schema, &[1], &["a"]);
+        let bundle = TestBundle::new(slots).with_payload(SlotId::new(0), fp, batch);
+
+        let result = seg.append(&bundle);
+        assert!(result.is_err());
+
+        let err = result.unwrap_err();
+        match err {
+            SegmentError::InvalidFormat { message } => {
+                assert!(
+                    message.contains("slots") && message.contains("exceeds limit"),
+                    "unexpected error message: {}",
+                    message
+                );
+            }
+            other => panic!("expected InvalidFormat, got {:?}", other),
+        }
+    }
 }


rust/otap-dataflow/crates/quiver/ARCHITECTURE.md
```diff
@@ -488,15 +531,160 @@ graph TD
...
+|---------|-----------|-------------|
+| 1       | 0         | 0           |
+| 2       | 1         | 0           |
+| 30      | 2         | 0           |
+| 31      | 3         | 0           |
+
+Using Arrow's nested types avoids string parsing and leverages the existing
+IPC decoder. The struct field types use the `ArrowPrimitive` trait to ensure
+type synchronization between the Rust newtypes (`SlotId`, `StreamId`,
+`ChunkIndex`) and their Arrow schema representation.
```
jmacd, line 649: Nice.

rust/otap-dataflow/crates/quiver/ARCHITECTURE.md
```diff
@@ -517,18 +705,23 @@ sequenceDiagram
...
+  could widen key types when cardinality exceeds the original type's capacity.
+  For example, if batches arrive with `DictionaryArray<UInt8>` but the unified
+  vocabulary exceeds 255 values, unification would produce `DictionaryArray<UInt16>`.
+  This breaks round-trip schema guarantees, which is unacceptable for a persistence
+  layer whose job is faithful reproduction.
+- **Trade-offs**:
+  - *Pro*: Exact schema preservation - readers get back what writers sent
+  - *Pro*: Each batch is self-contained and independently readable
+  - *Con*: Larger file sizes due to duplicate dictionary values, which also
+    increases memory consumption when segments are memory-mapped for reading
```
jmacd, line 722: Sounds good to me, thanks for all this detail.

rust/otap-dataflow/crates/quiver/src/record_bundle.rs
```diff
@@ -7,6 +7,134 @@ use std::borrow::Cow;
...
+/// schemas with the correct `DataType` for each field. Without this trait,
+/// the schema definition and the actual Rust type could drift apart:
+///
+/// ```ignore
+/// // Problem: SlotId is u16, but if someone changes it to u32,
+/// // this schema definition becomes wrong:
+/// Field::new("slot_id", DataType::UInt16, false)  // Hardcoded!
+/// ```
+///
+/// With `ArrowPrimitive`, the schema always derives from the Rust type:
```
jmacd, line 36: 🥇

rust/otap-dataflow/crates/quiver/src/config.rs
```diff
@@ -161,14 +161,55 @@ impl Default for WalConfig {
...
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct SegmentConfig {
-    /// Target payload size for a finalized segment.
+    /// Target payload size (in bytes) for a finalized segment.
+    ///
+    /// When the open segment's estimated size reaches this threshold, finalization
+    /// is triggered. Larger values improve compression and reduce per-segment
+    /// overhead, but increase memory usage and latency to visibility.
+    ///
+    /// Default: 32 MB. Recommended range: 8-128 MB.
```
jmacd, line 197: Are there guarantees about being over-or-under the target size? 
Curious since the batch processor has to address this in two ways:
- try not to exceed the limit, preferring to be under the limit instead
- admit there exist single-item requests that can exceed the limit, in this case let them but make them be singletons.

rust/otap-dataflow/crates/quiver/src/segment/writer.rs
```diff
@@ -0,0 +1,828 @@
...
+    /// - chunk_count: UInt32
+    fn encode_stream_directory(&self, streams: &[StreamMetadata]) -> Result<Vec<u8>, SegmentError> {
+        let schema = Arc::new(Schema::new(vec![
+            Field::new("stream_id", StreamId::arrow_data_type(), false),
+            Field::new("slot_id", SlotId::arrow_data_type(), false),
+            Field::new("schema_fingerprint", DataType::FixedSizeBinary(32), false),
+            Field::new("byte_offset", DataType::UInt64, false),
+            Field::new("byte_length", DataType::UInt64, false),
+            Field::new("row_count", DataType::UInt64, false),
+            Field::new("chunk_count", ChunkIndex::arrow_data_type(), false),
```
jmacd, line 355: Curiosity question: I see this is the schema of the stream directory, and I suspect there's a struct somewhere with matching field widths. Since I've seen other mechanisms (record_bundle's ArrowPrimitive) for aligning these, I'm curious if this struct and corresponding schema should have similar protections. Just curious how you think about the it--not blocking.

rust/otap-dataflow/crates/quiver/src/segment/reader.rs
```diff
@@ -0,0 +1,1489 @@
...
+    /// Opens a segment file with memory mapping for zero-copy access.
+    ///
+    /// The returned `SegmentReader` and any [`ReconstructedBundle`]s read from
+    /// it will reference the memory-mapped file directly. The file is unmapped
+    /// when all references are dropped.
+    ///
+    /// # Safety
+    ///
+    /// Memory mapping is inherently unsafe because the underlying file could
+    /// be modified or truncated while mapped, leading to undefined behavior.
```
lalitb, line 362: I believe this is critical issue, as it can lead to pipeline instance crash. Good to have a tracking issue, or else have a TODO in the code.

rust/otap-dataflow/crates/quiver/src/segment/writer.rs
```diff
@@ -0,0 +1,828 @@
...
+    /// Serializes each stream directly to the writer, avoiding the need to
+    /// buffer all IPC bytes in memory before writing.
+    fn write_streaming<W: Write>(
+        &self,
+        writer: &mut W,
+        accumulators: Vec<super::StreamAccumulator>,
+        manifest: Vec<ManifestEntry>,
+        path: &Path,
+    ) -> Result<(u64, u32), SegmentError> {
+        // Validate limits before writing to prevent creating unreadable segments
```
lalitb, line 219: Do we also need to validate MAX_SLOTS_PER_BUNDLE, and others, as done in reader.rs ?

rust/otap-dataflow/crates/quiver/src/segment/open_segment.rs
```diff
@@ -0,0 +1,551 @@
...
+    /// This is the sum of `get_array_memory_size()` for all accumulated
+    /// `RecordBatch`es across all streams. It represents the actual Arrow
+    /// buffer memory consumption, suitable for triggering size-based
+    /// finalization decisions.
+    ///
+    /// Note: This does not include manifest overhead or IPC encoding overhead,
+    /// so the final segment file size may differ slightly.
+    #[must_use]
+    pub fn estimated_size_bytes(&self) -> usize {
+        self.streams.values().map(|acc| acc.buffer_size()).sum()
```
lalitb, line 109: More of a question  - does this estimate decides when to close the current segment - How accurate is this estimate, or should it be complemented by a hard limit ?

lalitb:
Overall this is a solid PR, good to merge with understanding that there would be further incremental optimization before declaring stable/production-ready.

rust/otap-dataflow/crates/quiver/src/config.rs
AaronRM: No, the current segment writing implementation is the latter approach, a target/estimate which can be slightly exceeded. After we append a `RecordBundle` to the segment accumulator, we check for several trigger conditions for finalization one of which is the size estimate exceeding the target size.

I think this is generally acceptable because it can be tricky to accurately know the precise file size before actually writing it out (think padding, manifests, etc...) and end users aren't going to care about files being close to a target vs exactly a precise size.

However, I am _very_ aware that having a configurable hard upper bound on total disk utilization is quite important, so this is likely a topic we'll come back to later in that context.

albertlockett (approved):

rust/otap-dataflow/crates/quiver/src/segment/writer.rs
AaronRM: In fact, we can use the same mapping here. Pushed a commit to address.

rust/otap-dataflow/crates/quiver/src/segment/open_segment.rs
AaronRM: [Discussed here](https://github.com/open-telemetry/otel-arrow/pull/1643#discussion_r2628718412)

rust/otap-dataflow/crates/quiver/src/segment/writer.rs
AaronRM: Yep, good call. Added a check for this when appending to an open segment with a test to validate.

rust/otap-dataflow/crates/quiver/src/segment/reader.rs
AaronRM: This should be a very unusual case. Segment files are designed to be immutable after creation—we set filesystem permissions to read-only and never modify them in place. FWIW, `open_mmap` is hidden behind a `mmap` cargo feature for now. The default behavior is to read a copy of the bytes into memory using standard file reading APIs.

However, we should switch the mmap code to use `map_copy_read_only` which creates a private (copy-on-write) mapping. I'll create a follow up PR for this. If another process somehow modifies the file content, our view remains stable. The kernel provides a private copy of the original data.

This doesn't protect against truncation (which could cause `SIGBUS`), but memory mapping is *the same model used for loading system binaries and shared libraries*. As long as segment files are deleted rather than truncated—which is the normal lifecycle—existing mappings remain valid because the inode persists until all references are dropped.

AaronRM: PR: [[otap-quiver-df] Enhance safety of memory mapping in SegmentReader by using copy-on-write mapping](https://github.com/open-telemetry/otel-arrow/pull/1649)

PR #1645, opened 2025-12-17
Use item count instead of message count in transform proc

cijothomas:
Opening for discussion... I am wondering if the metrics should count the number of items, rather than the count of messages.

A transform with query `logs | where severity_text == "ERROR"` processing 1000 logs (assume batching_processor made a batch of 1000) will now show 1000 items received, 100 items sent (if only 100 were errors), making it easy to known 900 were dropped/filtered out....

OR we need both?

### DIFF
--- a/.github/workflows/pipeline-perf-on-label.yaml
+++ b/.github/workflows/pipeline-perf-on-label.yaml
@@ -133,7 +133,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
         with:
-          python-version: '3.14'
+          python-version: '3.11'
 
       - name: Build dataflow_engine
         run: |
--- a/rust/otap-dataflow/configs/fake-batch-debug-noop.yaml
+++ b/rust/otap-dataflow/configs/fake-batch-debug-noop.yaml
@@ -29,11 +29,13 @@ nodes:
           - debug
         dispatch_strategy: round_robin
     config:
-      # At the rate configured above, this will print a batch
-      # every 10 seconds.
-      min_size: 1000
-      # This interrupts it, so we expect around 300.
-      timeout: 3s
+      otap:
+        # At the rate set above (100/s), min_size=1000 would emit a batch
+        # every 10 seconds.
+        min_size: 1000
+        sizer: items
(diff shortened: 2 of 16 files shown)


cijothomas:
@albertlockett I opened to have a discussion (I am not fully familiar yet with all internals!)... Let me know your thoughts on this...

cijothomas:
> However for now, maybe we should just omit items_transformed or keep it in units of messages to avoid confusion?

Yes that makes sense (the overall explanation too!). I'll get back to this PR later.

jmacd:
I believe our goal should be to follow the Collector's standard pipeline metrics plan, documented here:

This question, the one raised here, is addressed in that proposal by having separate consumer- and producer-oriented counters so we can see when a component is filtering. 

Another answer, related to #1633 is that it makes sense to recognize three sizes for all pipeline data: Requests, Items, and Bytes. We do not currently have a bytes metric for OTAP records payloads, but we can come up with an estimate to count how much memory is used by a payload.

albertlockett:
> I believe our goal should be to follow the Collector's standard pipeline metrics plan, documented here: https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md
> 
> This question, the one raised here, is addressed in that proposal by having separate consumer- and producer-oriented counters so we can see when a component is filtering.
> 

@jmacd sounds good. So would the suggestion here be we remove these metrics entirely as they're generic enough and can be calculated by an instrumented channel? And after, if/when we have the columnar query engine emitting metrics, _those_ are what we'd expose as metrics for this component?

> We do not currently have a bytes metric for OTAP records payloads, but we can come up with an estimate to count how much memory is used by a payload.

I think this'd be straight forward. We could just add up the size of all the `RecordBatch`s for every payload type using https://docs.rs/arrow/latest/arrow/array/struct.RecordBatch.html#method.get_array_memory_size

jmacd:
@albertlockett yes, that sounds like my meaning. I'm glad to learn that Arrow has a method for computing byte size, so yes I would like to see three separate counts for producer and consumer sides of the component.

Therefore, I propose we accept @cijothomas's PR and later when we have something better/more comprehensive in place. Essentially, to say that these metrics are for the benchmarks.

cijothomas:
Closing now. Will comeback to this later.

Issue #1639, opened 2025-12-16
Columnar query engine support `RenameMapKeys` and `RemoveMapKeys` data expressions


albertlockett:
Add the capability to the columnar query engine to rename attributes and delete attributes.

We have already implemented the code needed to do this in the attributes processor, so writing a `PipelineStage` implementation that invokes this transformation code should be straight forward.

Issue #1634, opened 2025-12-16
Columnar query engine eager pipeline validation


albertlockett:
Currently the planning of the executable pipeline happens lazily when the first batch is seen. This is done because some `PipelineStage`s need the arrow schema in order to create datafusion physical plans/expressions. 

However, this is not great for when we need to integrate this into an OTAP Dataflow processor, since the user won't know they've written a pipeline expression with unsupported syntax until the telemetry starts flowing.

The query engine _should_ be able to inspect the pipeline expression early and at least tell if there are expressions used that are invalid or not supported. This would allow us to give an error at configuration time instead of runtime, which is a better experience for users.

albertlockett:
There are at least two `PipelineStage` implementations for which we need to implement this:

One is the `FilterPipelineStage`, which has many kinds of filters that are valid syntax but not supported by the engine.

The other is the `AttributeTransformPipelineStage`, which for renaming enforces the validation logic specified by `AttributesTransform` to ensure there is not ambiguous behaviour when bulk renaming attributes:
https://github.com/open-telemetry/otel-arrow/blob/17d4378b7e7e992edf4f77f8436f17c8958e0c82/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L535-L545

ThomsonTan:
This is still relevant, removing the stale label so it doesn't auto-close.

To move this forward, it would help to settle the design first, since the current lazy planning exists specifically because some stages (e.g. those building `DataFusion` physical plans) need the Arrow schema.

PR #1648, opened 2025-12-17
Support renaming attributes in columnar query engine

albertlockett:
part of #1639

Add support for renaming attributes in columnar query engine. This is done by using the `otap_df_pdata::otap::transform::transform_attributes` function to service some variants of the `Move` and `RenameMapKeys` `TranformExpression` from our pipeline AST.

For example, we now support queries like this to rename attributes:
```kql
logs | project-rename attributes["x"] = attributes["y"], attributes["a"] = attributes["b"]
```

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
@@ -247,7 +247,7 @@ mod test {
     }
 
     #[tokio::test]
-    async fn test_filtering_when_no_attrs_batch_present() {
+    async fn test_rename_when_no_attrs_batch_present() {
         let input = vec![LogRecord::build().event_name("test").finish()];
         let result = exec_logs_pipeline(
             "logs | 


rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -532,6 +533,20 @@ pub struct AttributesTransform {
...
 
 impl AttributesTransform {
+    /// Set the rename transform
+    #[must_use]
+    pub fn with_rename(mut self, rename: RenameTransform) -> Self {
+        self.rename = Some(rename);
+        self
+    }
+
+    /// Set the rename transform
```
cijothomas: stale comment?

rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
```diff
@@ -0,0 +1,283 @@
...
+                        .finish(),
+                ],
+            )],
+        )])
+    }
+
+    #[tokio::test]
+    async fn test_rename_single_attributes() {
+        let result = exec_logs_pipeline(
+            "logs | project-rename attributes[\"y\"] = attributes[\"x\"]",
```
cijothomas, line 124: this is to be read as "rename attribute "y" to attribute "x" ? or vice-versa? 

The test is validating the presence of "y", so I assume the vice-versa is true.

albertlockett: This is renaming x -> y. The way I think about it is, "take the value of x and assign it to y".

rust/otap-dataflow/crates/query-engine/src/pipeline/planner.rs
```diff
@@ -106,6 +122,141 @@ impl PipelinePlanner {
...
+                    // currently the only type of move transform supported is renaming attributes
+                    (
+                        ColumnAccessor::Attributes(src_attrs_id, src_key),
+                        ColumnAccessor::Attributes(dest_attrs_id, dest_key),
+                    ) => {
+                        // the attributes being renamed must be in the same map. for example doing
+                        // `attributes["x"] = resource.attributes["y"]` is not yet supported
+                        if src_attrs_id != dest_attrs_id {
+                            Err(Error::NotYetSupportedError {
+                                message: "renaming attributes from different maps".into(),
```
lalitb: nit - as this is end-user facing error message, should we make it more explicit say "attribute key rename only supports renaming within the same attributes map; found ... ", similar for others.

lalitb:
LGTM, with nit comment. Thanks,

rust/otap-dataflow/crates/query-engine/src/pipeline/planner.rs
albertlockett: thanks @lalitb, made this change

rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
```diff
@@ -0,0 +1,283 @@
...
+                .attributes,
+            &[
+                KeyValue::new("ys1", AnyValue::new_string("a")),
+                KeyValue::new("xs2", AnyValue::new_string("a")),
+            ]
+        );
+    }
+
+    #[tokio::test]
+    async fn test_filtering_when_no_attrs_batch_present() {
```
cijothomas:
```suggestion
    async fn test_rename_when_no_attrs_batch_present() {
```

rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
```diff
@@ -0,0 +1,283 @@
...
+            to_logs_data(input.clone()),
+        )
+        .await;
+
+        assert_eq!(result.resource_logs[0].scope_logs[0].log_records, input);
+    }
+
+    #[tokio::test]
+    async fn test_invalid_renames_are_errors() {
+        let invalid_renames = [
```
cijothomas, line 266: is the behavior same, when the query itself is valid, but it is renaming to an attribute which is already present in the log-record? We could add one test to target that scenario.

Also, a question on this - the ones being tested in this test - they seem to be detectable as invalid while parsing the query itself. Are we planning to detect such invalid queries much earlier?

rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
albertlockett: 😆 thanks again @cijothomas

rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
albertlockett: > is the behavior same, when the query itself is valid, but it is renaming to an attribute which is already present in the log-record? We could add one test to target that scenario.

Good catch @cijothomas -- I just checked and in this case it would create a duplicate attribute. This behaviour probably isn't desirable.

I created an issue to fix this in the utility function this uses for renaming, which is also used by the AttributesProcessor:
https://github.com/open-telemetry/otel-arrow/issues/1650

You're also correct that we should be able to detect this before the pipeline is executed. I'll address that in 
https://github.com/open-telemetry/otel-arrow/issues/1634

rust/otap-dataflow/crates/query-engine/src/pipeline.rs
```diff
@@ -329,6 +333,35 @@ mod test {
...
+        let otlp_bytes: OtlpProtoBytes = otap_payload.try_into().unwrap();
+        TracesData::decode(otlp_bytes.as_bytes()).unwrap()
+    }
+
+    /// helper function for converting [`OtapArrowRecords`] to [`MetricsData`]
+    pub fn otap_to_metrics_data(otap_batch: OtapArrowRecords) -> MetricsData {
+        let otap_payload: OtapPayload = otap_batch.into();
+        let otlp_bytes: OtlpProtoBytes = otap_payload.try_into().unwrap();
+        MetricsData::decode(otlp_bytes.as_bytes()).unwrap()
+    }
```
jmacd, line 355: non-blocking: Good candidate for a common helper method somewhere in crates/pdata/..

jmacd (approved):

PR #1651, opened 2025-12-18
PerfResults - use PR title only in charts

cijothomas:
image
Since we (we here refers to GH Actions Benchmarks) use PR Title + PR Description of each commit, it's quite hard to read the actual value when viewing the charts.

This PR modifies to just use the PR Title. There is no built-in way to do this, so had to do split on "\n". While at it, I also moved the logic to a shared script as well.


clhain:
oh man, beautiful!

albertlockett (approved):

PR #1652, opened 2025-12-18
Add validation for CPU core allocation arguments

cijothomas:
Validates `--num-cores` and `--core-id-range` arguments against available system cores. Previously, invalid core allocations were silently adjusted or failed with unclear errors. 

- Rejects `--num-cores` exceeding available cores (Previously we'd silently run on available ones...)
- Rejects core IDs in ranges that exceed system limits
- Rejects invalid ranges where start > end


cijothomas:
> LGTM - do we need to validate that the core ranges aren't overlapping? (this could happen in a followup, this is already a great addition)

https://github.com/open-telemetry/otel-arrow/issues/1658 opened to track. The current PR may be merged as-is.

PR #1653, opened 2025-12-18
[Geneva Exporter] Implement concurrent uploads and add metrics

lalitb:
Refactor GenevaExporter to support concurrent batch uploads, to improve throughput for high-volume telemetry - allowing multiple in-flight upload requests per exporter instance (controlled by max_concurrent_uploads).

 It also introduces internal metrics to track the health and performance of the export process and ensures proper Ack/Nack propagation. Also added tests to ensure  that the exporter correctly parses the user's YAML configuration, ensuring all fields and authentication methods (like certificates or managed identities) are deserialized properly.


rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
```diff
@@ -233,13 +274,19 @@ impl GenevaExporter {
...
+        payload: OtapPayload,
         effect_handler: &EffectHandler<OtapPdata>,
-    ) -> Result<(), String> {
-        // Split pdata into context and payload
-        let (_context, payload) = pdata.into_parts();
+    ) -> Result<usize, String> {
+        if payload.is_empty() {
+            // Empty payloads are a no-op but should still be acked.
+            // Avoids unnecessary encoding/upload work.
+            effect_handler
```
cijothomas: maybe switch to otel_info! since they are available now..

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
```diff
@@ -284,22 +331,17 @@ impl GenevaExporter {
...
-                        // Consider revisiting to implementing concurrent uploads
-                        // Upload each batch
-                        for batch in &batches {
-                            self.geneva_client
-                                .upload_batch(batch)
-                                .await
-                                .map_err(|e| format!("Failed to upload log batch: {}", e))?;
-                        }
+                        let batches_encoded =
+                            self.upload_batches_concurrent(&batches, "log").await?;
```
cijothomas: nit: could we use enum instead of str for the "log"/signal-type

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
```diff
@@ -137,17 +144,23 @@ pub enum AuthConfig {
...
-    // - bytes_sent: Counter<u64>
+    /// Total number of compressed batches produced by the encoder.
+    #[metric(unit = "{batch}")]
+    pub batches_encoded: Counter<u64>,
+
+    /// Total number of batches considered successfully uploaded.
+    #[metric(unit = "{batch}")]
+    pub batches_uploaded: Counter<u64>,
+
+    /// Total number of export attempts that failed (per message).
```
cijothomas: what does "per message" mean?

lalitb: It's actually per batch - which can contain `n` number of log-records. Have clarified better now
`/// Total number of export attempts that failed (per PData message).`

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
lalitb: Good point, done.

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
lalitb: Agree, done.

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
```diff
@@ -224,6 +241,32 @@ impl GenevaExporter {
...
+    ) -> Result<usize, String> {
+        let batches_encoded = batches.len();
+        self.metrics.batches_encoded.add(batches_encoded as u64);
+
+        let max_concurrent = self.config.max_concurrent_uploads.max(1);
+        let client = &self.geneva_client;
+
+        futures::stream::iter(batches.iter())
+            .map(Ok::<_, String>)
+            .try_for_each_concurrent(max_concurrent, |batch| async move {
```
utpilla, line 258: `try_for_each_concurrent` would return immediately on the first error it encounters and not run the remaining `upload_batch` operations to completion.

You might want to use [`for_each_concurrent`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.for_each_concurrent) here.

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
```diff
@@ -137,17 +146,23 @@ pub enum AuthConfig {
...
+    #[metric(unit = "{batch}")]
+    pub batches_encoded: Counter<u64>,
+
+    /// Total number of batches considered successfully uploaded.
+    #[metric(unit = "{batch}")]
+    pub batches_uploaded: Counter<u64>,
+
+    /// Total number of export attempts that failed (per PData message).
+    #[metric(unit = "{msg}")]
+    pub exports_failed: Counter<u64>,
```
utpilla, line 159: This seems redundant. You're already tracking that with `self.pdata_metrics.inc_failed(signal_type);`.

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
```diff
@@ -284,30 +335,26 @@ impl GenevaExporter {
...
-                                .await
-                                .map_err(|e| format!("Failed to upload log batch: {}", e))?;
-                        }
+                        let batches_encoded = self
+                            .upload_batches_concurrent(&batches, SignalType::Logs)
+                            .await?;
+
+                        otel_info!(
+                            "GenevaExporter.Upload",
+                            count = batches.len(),
```
utpilla, line 344: We should be checking the result of `upload_batches_concurrent` method to determine this value.

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
```diff
@@ -355,9 +398,10 @@ impl GenevaExporter {
             OtapPayload::OtlpBytes(otlp_bytes) => {
                 match otlp_bytes {
                     OtlpProtoBytes::ExportLogsRequest(bytes) => {
-                        effect_handler
-                            .info("Uploading logs to Geneva using OTLP fallback path")
-                            .await;
+                        otel_info!(
+                            "GenevaExporter.Upload",
+                            message = "Uploading logs to Geneva using OTLP fallback path"
```
utpilla, line 403: This isn't fallback path. It's the regular path for scenarios where there is no arrow conversion needed, right?

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
```diff
@@ -457,32 +491,67 @@ impl Exporter<OtapPdata> for GenevaExporter {
...
-
-                    return Ok(TerminalState::new(deadline, [self.metrics]));
+                    otel_info!(
+                        "GenevaExporter.Shutdown",
+                        message = "Geneva exporter shutting down"
+                    );
+
+                    return Ok(TerminalState::new(
+                        deadline,
+                        vec![self.pdata_metrics.snapshot(), self.metrics.snapshot()],
```
utpilla, line 513: nit: We could use an array instead of `vec!` here.

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
lalitb: @utpilla  Good point. This was done intentionally, but open for discussion. Currently, we send failure/nack for the complete batch, not for individual failures. So
- **With fail-fast** (`try_for_each_concurrent`):                                                                                                                                                      
  - Batch 1 fails -> stop immediately                                                                                                                                                             
  - `NACK` the message                                                                                                                                                                             
  - Upstream retries all batches                                                                                                                                                                 
  - No duplicates, no wasted work                                                                                                                                                                
                                                                                                                                                                                                 
- **With best-effort** (for_each_concurrent):                                                                                                                                                        
  - Batch 1 fails, but we continue uploading 2, 3, 4                                                                                                                                             
  - We still `NACK` (because batch 1 failed)                                                                                                                                                       
  - Upstream retries all batches                                                                                                                                                                 
  - Now batches 2, 3, 4 are duplicates                                                                                                                                                           
  - Wasted bandwidth and resources

utpilla: The problem of duplicate uploads would apply to fail-fast (`try_for_each_concurrent`) as well, right? What if only the last batch fails, you would still NACK the entire message which would lead to retrying all the batches?

lalitb: Right, duplicates can happen in both cases since we NACK the whole message. The difference is _when_ we keep doing work.
With **best-effort**, once a batch has already failed, we still continue uploading later batches even though we already know the message will be retried and those uploads will be duplicated. That’s wasted work.
The **fail-fast** approach avoids that by stopping as soon as failure is known, so we don’t keep uploading batches that are guaranteed to be retried anyway.

utpilla: There's another consideration: **latency**. With `try_for_each_concurrent`, if an early batch fails, we cancel all remaining uploads and NACK the entire message. This means batches that would have succeeded now wait for the full retry cycle before being delivered.

I believe getting logs delivered quickly (even if duplicated) is often more valuable than optimizing for efficiency, but we could have different opinions here.

That said, the ideal solution would be **per-batch success tracking** - using `for_each_concurrent` while tracking which specific batches failed and only retrying those. This would eliminate both the latency penalty and duplicates but requires changes to how we handle Acks/Nacks upstream. @jmacd Worth considering for a future enhancement?

I wanted to surface this trade-off in case it affects how you're thinking about resource efficiency vs. latency for this exporter. Let me know what you think. Either way, we should probably document this decision and its implications.

lalitb: On latency - fair point, but the tradeoff is: one retry cycle of delay vs guaranteed duplicate uploads.  I'd rather delay slightly than send data we know will be sent again anyway.

On per-batch tracking - agree that's the ideal long-term solution. It can't be done in the exporter alone though - AckMsg/NackMsg operate on the full PData with no partial ACK support. Would need engine-level changes.  Worth tracking as an eventual goal. cc @jmacd

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
lalitb: Good catch. This is just the aggregate of the per-signal-type counters in pdata_metrics - removing in separate PR.

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
lalitb: Good catch. Fixing in separate PR.

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
lalitb: Good point. Renaming in separate PR.

rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/mod.rs
lalitb: fixed in separate PR.

PR #1654, opened 2025-12-18
Support deleting attributes in columnar query engine

### LINKED ISSUE #1639: Columnar query engine support `RenameMapKeys` and `RemoveMapKeys` data expressions

albertlockett:
closes #1639 

Adds the ability to delete attributes to the columnar query engine.

We now support queries like:
```kql
logs | project-away attributes["x"], attributes["y"]
```


rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
```diff
@@ -280,4 +280,70 @@ mod test {
...
     }
+
+    #[tokio::test]
+    async fn test_delete_attributes() {
+        let result = exec_logs_pipeline(
+            "logs | project-away attributes[\"x\"]",
+            generate_logs_test_data(),
+        )
+        .await;
+        assert_eq!(
```
cijothomas, line 291: nit: Maybe asserting that the result does not contain the deleted attribute is more easier to read the tests (in future by others!)

rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
```diff
@@ -280,4 +280,70 @@ mod test {
...
+                .unwrap()
+                .attributes,
+            &[KeyValue::new("xs2", AnyValue::new_string("a")),]
+        );
+    }
+
+    #[tokio::test]
+    async fn test_delete_when_no_attrs_batch_present() {
+        let input = vec![LogRecord::build().event_name("test").finish()];
+        let result = exec_logs_pipeline(
```
cijothomas, line 348: nit: Add test coverage for similar situation (i.e attribute not present) for resource.attr, scope.attr.

albertlockett: done in 9f68a06d

rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
```diff
@@ -280,4 +280,91 @@ mod test {
...
+            input.resource_logs[0].resource,
+        );
+        assert_eq!(
+            result.resource_logs[0].scope_logs[0].scope,
+            input.resource_logs[0].scope_logs[0].scope
+        );
+        assert_eq!(
+            result.resource_logs[0].scope_logs[0].log_records,
+            input.resource_logs[0].scope_logs[0].log_records
+        );
```
lalitb, line 368: Nice work, just some test gaps, if they really make sense
   - Deleting a non-existent attributes ( I believe it would be no-op)
   - Deleting all attributes from a log record
   - Deleting a non-map column ( this should be error ?)
   - Mixing rename and delete in same query.

PR #1659, opened 2025-12-18
Fix perf result uploading

cijothomas:
https://github.com/open-telemetry/otel-arrow/pull/1651 didn't work well, as the GH Action is adding the commit info late, and we don't have a way to interfere easily. Will check for some other option, but until then, this PR must be merged to get back broken perf uploading!


cijothomas:
@lquerel Please merge quickly to unblock perf results :)

lquerel:
@cijothomas PR in the merge queue!

PR #1661, opened 2025-12-18
Tokio metrics

lquerel:
This PR adds Tokio-related multivariate metrics. Some of these metrics are only reported when the DF engine is compiled with `RUSTFLAGS="--cfg tokio_unstable"`.

It is also now possible to enable or disable pipeline-related metrics per pipeline configuration. By default, all of these metrics are enabled.


lquerel:
> Looks good. I think ideally we will use some kind of configuration knob with levels for metric detail. Some of these metrics will be detailed-level, IMO, something we can figure out later.

Note that I added two knobs in the pipeline configuration to enable/disable pipeline metrics and tokio metrics. It's probably not granular enough but we can already disable tokio metrics for a specific pipeline !

PR #1663, opened 2025-12-19
feat: Add internal logs receiver

andborja:
Add internal logs sdk exporter sending telemetry to an internal logs receiver.

Opening as draft initially to discuss the general model and decide on the proper objects to send through the channel.

This PR presents the bridge to receive the internal telemetry. It also demonstrate how to implement the receiver to be included in a pipeline.

### DIFF
--- a/rust/otap-dataflow/crates/telemetry/Cargo.toml
+++ b/rust/otap-dataflow/crates/telemetry/Cargo.toml
@@ -42,9 +42,6 @@ tracing = { workspace = true }
 tracing-subscriber = { workspace = true, features = ["env-filter","registry", "std", "fmt"] }
 prost = { workspace = true }
 opentelemetry-prometheus = { workspace = true }
-opentelemetry-appender-tracing = { workspace = true }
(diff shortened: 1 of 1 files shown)


jmacd:
Thank you @andborja.

Comparing and contrasting with the draft I'm working on, #1672, your approach configures the OTel SDK with its various processor and exporter options from the configuration. I agree that we want an option to configure the OTel SDK in all of the ways implied by the declarative configuration model, and at the same time I'm worried about relying on the SDK's data representation in this code base. 

In this code base, I think it makes more sense to focus on the OTLP bytes approach outlined in #1672 which is to optimize for the self-hosted internal telemetry code path at the expense of making the use 
OpenTelemetry SDK processors/exporters somewhat more costly. The argument is that OTAP Dataflow has a batch processor and OTLP exporter for OTLP bytes that bypasses the use of OTLP protobuf objects. The OTAP Dataflow code path will be less expensive than the OTel SDK code path because it avoids OTLP protobuf objects.

PR #1665, opened 2025-12-19
Another windows and macOS test to ignore

cijothomas:
Hopefully, this'd make it back to green!

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/README.md
+++ b/rust/otap-dataflow/crates/otap/src/experimental/geneva_exporter/README.md
@@ -31,6 +31,12 @@ You should see `urn:otel:geneva:exporter` in the Exporters list.
 ./target/release/df_engine --pipeline config.yaml --num-cores 4
 ```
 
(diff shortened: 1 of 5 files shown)


rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -490,6 +490,10 @@ mod tests {
 
     #[cfg(feature = "experimental-tls")]
     #[tokio::test]
+    #[cfg_attr(
+        any(target_os = "windows", target_os = "macos"),
+        ignore = "Skipping on Windows and macOS due to flakiness"
```
lalitb, line 495: I implemented this on macOS, so should have worked. Maybe something different in Github runner env, can have a look into it.

cijothomas: its just flaky. Works fine on my Mac too!

jmacd (approved):

albertlockett (approved):

Issue #1110, opened 2025-09-11
[query-engine] Documentation round of KQL grammar choices / supported grammar


drewrelmas:

drewrelmas:
I will throw together a quick README.md showing the available KQL grammar

PR #1668, opened 2025-12-19
[query-engine] KQL function parsing + type conversion

CodeBlanch:
Relates to #1479

## Changes

* Adds support for function definition and invocation in KQL parser
* Implements automatic type conversion for functions in RecordSet engine

### DIFF
--- a/rust/experimental/query_engine/kql-parser/src/shared_expressions.rs
+++ b/rust/experimental/query_engine/kql-parser/src/shared_expressions.rs
@@ -107,7 +107,9 @@ pub(crate) fn parse_variable_definition_expression(
         return Err(ParserError::QueryLanguageDiagnostic {
             location: to_query_location(&identifier_rule).clone(),
             diagnostic_id: "KS201",
-            message: format!("A variable or function with the name '{name}' has already been declared"),
+            message: format!(
+                "A variable or function with the name '{name}' has already been declared"
+            ),
         });
     }
 


rust/experimental/query_engine/kql-parser/src/query_expression.rs
```diff
@@ -59,58 +61,230 @@ pub(crate) fn parse_query(
...
+
+                let identifier_rule = udf_rules.next().unwrap();
+
+                let name = identifier_rule.as_str().trim();
+                if state.is_well_defined_identifier(name) {
+                    errors.push(ParserError::QueryLanguageDiagnostic {
+                        location: to_query_location(&identifier_rule).clone(),
+                        diagnostic_id: "KS201",
+                        message: format!(
+                            "A variable with the name '{name}' has already been declared"
```
drewrelmas: Non-blocking comment - I was on the fence again about this one, for a user it may be easier to see `A function with the name....` but I know we've been deferring to the existing KQL grammar diagnostics - I tried it myself in Azure Data Explorer.

drewrelmas (approved):
As the `recordset` engine reaches its initial maturity, we should get back to #1110 and add some docs on supported grammar. The function support especially could use some instructions.

rust/experimental/query_engine/kql-parser/src/query_expression.rs
CodeBlanch: I switched it to `A variable or function with the name '{name}' has already been declared`

PR #1670, opened 2025-12-20
PipelinePerfTest - Try scaling tests for engine

cijothomas:


tools/pipeline_perf_test/test_suites/integration/continuous/saturation-cores-template.yaml.j2
```diff
@@ -0,0 +1,81 @@
...
+          - "--http-admin-bind"
+          - "0.0.0.0:8080"
+    monitoring:
+      docker_component:
+        allocated_cores: {{backend_cores}}
+      prometheus:
+        endpoint: http://localhost:8087/telemetry/metrics?format=prometheus&reset=false
+
+tests:
+  - name: OTLP-ATTR-OTLP
```
cijothomas, line 70: only starting with OTLP-OTLP for now. Once the test parameters are polished to hit proper saturation, we can add more scenarios involving OTAP.

cijothomas:
This should show results in the same charts as before with each plot representing core count like below.

"Continuous - Saturation - 1 Core(s) - OTLP-ATTR-OTLP"
"Continuous - Saturation - 2 Core(s) - OTLP-ATTR-OTLP"

I think once this is polished, I can try to run some derived numbers like scaling-efficiency etc. and plot them, so we know instantly how much linear scaling we get. 


Also, I started with 3LoadGen perf engine-core, and 200 batch-size per engine-core. Will tweak these based on actual runs on the dedicated machine!

cijothomas:
> This should show results in the same charts as before with each plot representing core count like below.
> 
> "Continuous - Saturation - 1 Core(s) - OTLP-ATTR-OTLP" "Continuous - Saturation - 2 Core(s) - OTLP-ATTR-OTLP"
> 
> I think once this is polished, I can try to run some derived numbers like scaling-efficiency etc. and plot them, so we know instantly how much linear scaling we get.
> 
> Also, I started with 3LoadGen perf engine-core, and 200 batch-size per engine-core. Will tweak these based on actual runs on the dedicated machine!

Something like below: (shown for just 1-2 cores)
image

clhain:
Shouldn't we hold the max batch size at a constant value for each of these so we're not changing 2 things at once? Otherwise looks great!

cijothomas:
> Shouldn't we hold the max batch size at a constant value for each of these so we're not changing 2 things at once? Otherwise looks great!

Good point. I was trying to get the CPU to max and ran out of cores to run load-gens, so only other variable I had was the batch-size! I'll monitor the results in the dedicated runner, and adjust them after 1st few runs.

I agree to keep batch_size steady, and vary just the load-generator instance count..

.github/workflows/pipeline-perf-test-continuous.yml
```diff
@@ -88,7 +88,10 @@ jobs:
         - name: Run pipeline performance test log suite - Saturation
           run: |
             cd tools/pipeline_perf_test
-            python orchestrator/run_orchestrator.py --debug --config test_suites/integration/continuous/saturation-docker.yaml
+            for config in test_suites/integration/continuous/saturation-*.yaml; do
```
lalitb, line 91: This will also include running the [saturation_docker.yaml](https://github.com/open-telemetry/otel-arrow/blob/main/tools/pipeline_perf_test/test_suites/integration/continuous/saturation-docker.yaml), which contains 1 core test similar to saturation_1core.yaml. Is this intended - as this seems to be duplicate work.

tools/pipeline_perf_test/test_suites/integration/continuous/saturation-16cores.yaml
```diff
@@ -0,0 +1,11 @@
+# Saturation test with 16 core df_engine
```
lalitb, line 1: I was going to comment whether Github runners will have the 64 cores as required by this test, but then I noticed we use the dedicate bare metal Oracle machine - just curious if you know where it is hosted?

tools/pipeline_perf_test/test_suites/integration/continuous/saturation-16cores.yaml
```diff
@@ -0,0 +1,11 @@
...
+from_template:
+  path: test_suites/integration/continuous/saturation-cores-template.yaml.j2
+  variables:
+    num_cores: 16
+    engine_core_range: "1-16"
+    loadgen_cores: 48
+    loadgen_core_range: "17-64"
```
lalitb: I believe the cores are numbered `0-63` in 64-CPU machine, should this be `17-63` ?

tools/pipeline_perf_test/test_suites/integration/continuous/saturation-16cores.yaml
```diff
@@ -0,0 +1,11 @@
...
+    backend_cores: 1
```
lalitb: With the 16-core engine test, will the single-core backend become a bottleneck, or is the perf exporter lightweight enough to keep up with engine throughput as it scales?

lalitb:
For higher core tests (8-core, 16-core) - 
 - At input side, would the single exporter used with loadgen become bottleneck - one gRPC connection to push enough data to saturate 8-16 engine cores?
 - at output side, would the single-core backend become a bottleneck?

Not a blocker - just something to validate once we have results.

tools/pipeline_perf_test/test_suites/integration/continuous/saturation-16cores.yaml
cijothomas: you are right. I'll fix.
(https://github.com/open-telemetry/otel-arrow/pull/1652 was added recently to catch and fail on such situations!)

tools/pipeline_perf_test/test_suites/integration/continuous/saturation-16cores.yaml
cijothomas: to be seen. Will need to run few to find the right balance.

cijothomas:
> For higher core tests (8-core, 16-core) -
> 
> * At input side, would the single exporter used with loadgen become bottleneck - one gRPC connection to push enough data to saturate 8-16 engine cores?
> * at output side, would the single-core backend become a bottleneck?
> 
> Not a blocker - just something to validate once we have results.

for input-side, we are increasing the count of loadgen cores so as to hit saturation. (output side is to be seen.)

tools/pipeline_perf_test/test_suites/integration/continuous/saturation-16cores.yaml
cijothomas: It's a CNCF provided machine for OTel perf testing. Its quite powerful, see spec below:

System Information:
  Available CPU cores: 128
  Available memory: 491.73 GB / 503.27 GB

cijothomas: https://github.com/open-telemetry/community/blob/main/assets.md?plain=1#L68

.github/workflows/pipeline-perf-test-continuous.yml
cijothomas: removed the existing one

tools/pipeline_perf_test/test_suites/integration/continuous/saturation-16cores.yaml
cijothomas: modified backend to match to also increase proportionately.

cijothomas:
> Shouldn't we hold the max batch size at a constant value for each of these so we're not changing 2 things at once? Otherwise looks great!

Modifies to use fixed batch_size. (Will monitor and adjust in subsequent PRs.)

PR #1672, opened 2025-12-20
Internal telemetry routing ** WIP DRAFT **

jmacd:
Related to #1663.

This is a proposal to support an internal route for telemetry that lets us process OTAP-Dataflow telemetry using our own pipeline support. This requires special protections from self-induced telemetry, and it requires options to route telemetry in many ways to remain compatible with OpenTelemetry and keep our options. The proposal to accompany this PR is in the ARCHITECTURE.md draft.

### DIFF
--- a/rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
+++ b/rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
@@ -2,15 +2,15 @@
 
 ## Architecture
 
-OTAP-Dataflow uses a configurable internal telemetry data plane.  We
-support alternatives to enable a range of observability requirements.
 The internal telemetry SDK is designed for the engine to safely
 consume its own telemetry, and we intend for the self-hosted telemetry
-pipeline to be the standard configuration.
+pipeline to be the standard configuration for all OpenTelemetry
+signals.
 
-Consuming self-generated ("telemetry presents a potential a kind of
+Consuming self-generated telemetry presents a potential a kind of
 feedback loop, situations where a telemetry pipeline creates pressure
-on itself and must not explode. 
+on itself. We have designed for the OTAP dataflow engine to remain
+reliable even with this kind of dependency on itself.
 
 ## Internal telemetry receiver
 
@@ -28,10 +28,10 @@ in this exchange.
 
 The internal telemetry receiver is the SDK's counterpart, making it
 second party as it is responsible for routing internal telemetry. The
-ITR cannot use the internal telemetry SDK itself, an invisible member
-of the pipeline. The ITR can be instrumented using third-party
-instrumentation (e.g., `tracing`, `log` crates) provided it can
-guarantee there is no potential for feedback (e.g., a single
+ITR cannot use the internal telemetry SDK itself, making it an
+invisible member of the pipeline. The ITR can be instrumented using
+third-party instrumentation (e.g., `tracing`, `log` crates) provided
+it can guarantee there is no potential for feedback (e.g., a single
 `tracing::info()` statement at startup).
 
 ## Pitfall avoidance
@@ -49,15 +49,17 @@ telemetry pitfalls, as follows:
   telemetry processes that use third-party instrumentation.
 - A thread-local variable is used to redirect third-party
   instrumentation in dedicated internal telemetry threads. Internal
-  telemetry threads automatically configure a safe configuration.
+  telemetry threads automatically configure a safe configuration
+  that drop third-party instrumentation instead of creating feedback.
 - Components under observation (non-ITR components) have internal
-  telemetry events routed queues in the OTAP-Dataflow pipeline on the
-  same core, this avoids blocking the engine. First-party
+  telemetry events routed to queues in the OTAP-Dataflow pipeline on
+  the same core, this avoids blocking the engine. First-party
   instrumentation will be handled on the CPU core that produced the
   telemetry under normal circumstances. This isolates cores that are
   able to process their own internal telemetry.
-- Option to fall back to no-op, a non-blocking global provider, and/or
-  raw logging.
+- Option to configure internal telemetry multiple ways, including the
+  no-op implementation, multi-threaded subscriber, routing to the
+  same-core ITR, and/or raw logging.
 
 ## OTLP-bytes first
 
@@ -70,15 +72,24 @@ Event][TOKIOEVENT] handler to produce OTLP bytes before dispatching to
 an internal pipeline, used (in different configurations) for first and
 third-party instrumentation.
 
+We use an intermediate representation in which the dynamic elements of
+the `tracing` event are encoded while primtive fields and metadata
+remain in structured form. These are encoded to the OTLP `LogRecord`
+protocol.
+
 [TOKIOEVENT]: https://docs.rs/tracing/latest/tracing/struct.Event.html
 
 ## Raw logging
 
 We support formatting events for direct printing to the console from
-OTLP bytes, based on `otap_df_pdata::views::logs::LogsDataView` and
-associated types, a zero-copy approach. We refer to this most-basic
-form of printing to the console as raw logging because it is a safe
-configuration early in the lifetime of a process.
+OTLP bytes. For the dynamic encoding, these are consumed using
+`otap_df_pdata::views::logs::LogsDataView` (which is forbidden from
+using Tokio `tracing`), our zero-copy accessor. We refer to this
+most-basic form of printing to the console as raw logging because it
+is a safe configuration early in the lifetime of a process.
+
+This configuration is meant for development purposes, it is likely to
+introduce contention over the console.
 
 ## Routing
 
@@ -100,47 +111,24 @@ The two internal logs data paths are:
 Each of the items below is relatively small, estimated at 300-500
 lines of new code plus new tests.
 
-### TracingLogRecord: Tokio tracing Event and Metadata to LogRecordView
+### LogRecord: Tokio tracing Event and Metadata to LogRecordView
 
 When we receive a Tokio tracing event whether through a
 `tracing::info!` macro (or similar) or through a dedicated
 `EffectHandler`-based API, the same happens:
 
-Create a `TracingLogRecord`, a struct derived from `tracing::Event`
-and `tracing::Metadata`, containing raw LogRecord fields extracted
-from the tracing macro layer. The `otap_df_pdata::views::logs::LogRecordView` is
-implemented for `TracingLogRecord` making it the `TracingLogRecord` something
-we can transcode into OTel-Arrow batches.
+Create a `LogRecord`, a struct derived from `tracing::Event` and
+`tracing::Metadata`, containing raw LogRecord fields extracted from
+the tracing macro layer plus a fresh timestamp. Log record attributes
+and the log event body are encoded as the "attributes and body bytes"
+field of `LogRecord`, the other fields are copied.
 
-The `otap_df_pdata` crate currently has no OTLP bytes encoder for
-directly accepting `otap_df_pdata::views::*` inputs (note the
-OTAP-records-to-OTLP-bytes function bypasses the views and encodes
-bytes directly). Therefore, this project implies we extend or refactor
-`otap_df_ptdata` with an OTLP bytes encoder for its views interfaces.
+With this record, we can defer formatting or encoding the entire
+record until later. We can:
 
-Then, `TracingLogRecord` implements the log record view, we will encode
-the reocrd as OTLP bytes by encoding the view.
-
-### Stateful OTLP bytes encoder for repeated LogRecordViews
-
-We can avoid sending a log record through a channel every time an event
-happens by buffering log records. We will buffer them as OTLP bytes. Each 
-receiver of events from `TracingLogRecord` OTLP bytes will use one stateful
-encoder that is:
-
-- Preconfigured with the process-level OpenTelemetry `Resource` value
-- Remembers the OpenTelemetry `InstrumentationScope.Name` that was previously used
-- Remembers the starting position of the current `ResourceLogs` and `ScopeLogs` of a 
-  single OTLP bytes payload.
-  
-Whether a global logging collector thread or an effect handler thread
-processing internal telemetry, we will enter the stateful encoder and
-append a `LogRecordView` with its effective
-`InstrumentationScope`. The stateful encoder will append the log
-record correctly, recognizing change of scope and a limited buffer
-size.  This re-uses the `ProtoBuf` object from the existing
-OTAP-records-to-OTLP-bytes code path for easy protobuf generation
-(1-pass encoder with length placeholders).
+- For raw logging, format directly for the console
+- Finish the full OTLP bytes encoding for the `LogRecord`
+- Sort and filter before combining into a `LogsData`.
 
 ### OTLP-bytes console logging handler
 
@@ -152,7 +140,7 @@ format human-readable text for the console directly from OTLP bytes.
 This OTLP-bytes-to-human-readable logic will be used to implement raw
 logging.
 
-### Global logs collection thread 
+### Global logs collection thread
 
 An OTAP-Dataflow engine will run at least one global logs collection
 thread. These threads receive encoded (OTLP bytes) log events from
@@ -195,16 +183,16 @@ service:
       level: info
       internal_collection:
         enabled: true
-        
+
         # Per-thread buffer
         buffer_size_bytes: 65536
-        
+
         # Individual record size limit
         max_record_bytes: 16384
-        
+
         # Bounded channel capacity
         max_record_count: 10
-        
+
         # Timer-based flush interval
         flush_interval: "1s"
 ```


rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,298 @@
...
+which means it is designed to safely consume its own telemetry. While
+this requires care and attention, the engine contains within
+`otap-df-otap` all the building blocks to "be" an OpenTelemetry SDK.
+
+Consuming internal telemetry presents a potential for self-induced
+telemetry, a harmful form of self-dependency. The OTAP-Dataflow
+internal telemetry pipeline is rigorously safeguarded against these
+pitfalls through:
+
+- OTAP-Dataflow components downstream of an ITR cannot be configured
```
andborja: ITR - first time usage. Define.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,298 @@
...
+configuration is called in the OpenTelemetry Collector][TELSETTINGS].
+
+[TELSETTINGS]: https://github.com/open-telemetry/opentelemetry-collector/blob/bf28fa76882d0d6e40457db8bfffb86a4efcdfbf/component/telemetry.go#L14
+
+TelemetrySettings will be configurable, allowing fine-grain control
+over logging behavior in the components.  The router is configurable
+with:
+
+- No-op logging
+- Raw logging
```
andborja: If "Raw logging" means console output, It should be through the SDK. It is yet another exporter, and should not be considered by separate.

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver/mod.rs
```diff
@@ -0,0 +1,72 @@
+// Copyright The OpenTelemetry Authors
```
andborja, line 1: avoid the "mod.rs" format. Use a file internal_telemetry_receiver.rs at the same level as lib.rs instead.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
cijothomas: for stdout/console, we cannot (and should not) use OTel. Need to use `tracing::fmt` or develop customer.

jmacd:
I apologize-- the code in this PR is not acceptable for review, we may view it as a feasibility study to accompany the ARCHITECTURE.md document.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: Yes the OTel exporter is not specified as a stable output format by OTel.

I see that it is a JSON-line oriented exporter, and I would welcome such a thing (@gouslu and I discussed deriving such a thing from the AzMon exporter code, which is similar to JSON-line).

In this PR as proof of concept I replaced `tracing::fmt` and developed an OTLP-bytes-to-console exporter, that's essentially what you're saying I think @cijothomas.

Actually, I've seen logging SDKs emit directly to protobuf bytes before! One reason we can't use `tracing::fmt` over the OTLP bytes representation in this case (AFAICT) is that the tracing Event struct does not contain a timestamp, there's no way to format a log statement recorded in the past. This is not the case for OTel SDK, why in this proposal we're able to reconstruct an OTLP SDK log event and process/export as an alternative to the ITR.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -18,6 +19,75 @@ pub struct LogsConfig {
...
+    pub buffer_size_bytes: usize,
+    
+    /// Maximum size in bytes for a single log record.
+    /// Records exceeding this size are dropped with a counter increment.
+    /// This limit enables encoder optimizations (2-byte length placeholders for 14-bit sizes).
+    #[serde(default = "default_max_record_bytes")]
+    pub max_record_bytes: usize,
+    
+    /// Maximum number of records in the bounded channel.
+    /// When full, new records fall back to raw console logger.
```
cijothomas, line 55: why fallback here, instead of keeping count of the drops?

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -18,6 +19,75 @@ pub struct LogsConfig {
...
+    #[serde(default = "default_buffer_size_bytes")]
+    pub buffer_size_bytes: usize,
+    
+    /// Maximum size in bytes for a single log record.
+    /// Records exceeding this size are dropped with a counter increment.
+    /// This limit enables encoder optimizations (2-byte length placeholders for 14-bit sizes).
+    #[serde(default = "default_max_record_bytes")]
+    pub max_record_bytes: usize,
+    
+    /// Maximum number of records in the bounded channel.
```
cijothomas, line 54: "bounded" channel feels like internal implementation details; so we should avoid exposing them to public config..

rust/otap-dataflow/crates/engine/src/effect_handler.rs
```diff
@@ -6,13 +6,60 @@
...
 use tokio::net::{TcpListener, UdpSocket};
+use tokio::sync::mpsc;
+
+/// Telemetry buffer state for local (!Send) components.
+pub(crate) struct LocalTelemetryState {
+    pub(crate) encoder: StatefulOtlpEncoder,
+    pub(crate) resource_bytes: OtlpProtoBytes,
+    pub(crate) scope_name: String,
+    pub(crate) flush_threshold_bytes: usize,
+    pub(crate) overflow_sender: mpsc::UnboundedSender<Bytes>,
```
cijothomas, line 28: is it okay to use a unbounded sender for overflow purposes?

lquerel: This PR is a draft, so it is probably a temporary shortcut. The rule I personally follow is to systematically use bounded channels together with a policy that specifies what should happen in case of overflow, that is, drop the incoming message or block the sender. This policy can be configurable, or chosen directly in the code depending on whether it is worth making configurable.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -18,6 +19,75 @@ pub struct LogsConfig {
     /// The list of log processors to configure.
     #[serde(default)]
     pub processors: Vec<processors::LogProcessorConfig>,
+    
+    /// Internal collection for component-level logs.
+    /// 
+    /// When enabled, component logs (otel_info!, otel_warn!, etc.) are routed through
+    /// an internal telemetry receiver in the OTAP pipeline, allowing use of built-in
+    /// batch processors, retry, and exporters (console, OTLP, etc.).
```
lquerel, line 27: I'm a big fan of this approach.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -18,6 +19,75 @@ pub struct LogsConfig {
...
+    /// 
+    /// When disabled (default), component logs are routed to the OpenTelemetry SDK,
+    /// using the same export path as 3rd party logs from tokio-tracing-rs.
+    #[serde(default)]
+    pub internal_collection: InternalCollectionConfig,
```
lquerel, line 32: Long term, I think this approach will apply not only to logs, but also to metrics and traces. At some point, we might promote this configuration to a more general level.
For now, I suggest that this field be:
* an option, so we can remove the `enabled` field from `InternalCollectionConfig`, which I think would simplify things
* renamed to something more explicit. I'm not fully satisfied with my proposal `otap_pipeline`; there is probably a better name to find

rust/otap-dataflow/crates/engine/src/local/processor.rs
```diff
@@ -252,6 +258,43 @@ impl<PData> EffectHandler<PData> {
...
 
+    /// Log a telemetry event from this component.
+    ///
+    /// This method encodes the log record into the component's telemetry buffer.
+    /// On overflow (buffer exceeds threshold), it automatically flushes to the
+    /// configured destination (ITR or SDK fallback) via non-blocking channel send.
+    ///
+    /// This method never fails - errors are silently dropped to prevent recursion.
+    /// If the telemetry buffer is not configured, this is a no-op.
+    pub fn log_event(&self, log_record: &impl otap_df_pdata::views::logs::LogRecordView) {
```
lquerel, line 269: We need to find a way to reuse the `NodeAttributeSet` that we already use for metrics. That will let every log emitted by a node share a common context with the metrics.

lquerel (commented):
I mainly focused on the architecture document for now, which is very detailed and informative. A few general remarks:

* In the long term, we should be able to define `TelemetrySettings` at a global level, at the engine level, with the possibility to override them in a specific pipeline configuration.
* In the case of a tight integration with the otap-engine, I think we need to find a way to reuse the `AttributeSet` concept already used for metrics. This offers several advantages:

  1. we get a common language and definition for attributes,
  2. they are populated by the engine and registered only once, which avoids redefining them on every call. We only need to provide an attribute set ID, which is essentially just a number. Additional "dynamic" attributes could be added when needed.
* We should be able to compile the engine in a way that eliminates all overhead for a given logging level, essentially the same behavior we have today with the current macros. I believe this is already the plan, but I wanted to make this rule explicit.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+pipeline to be the standard configuration.
+
+Consuming self-generated ("telemetry presents a potential a kind of
+feedback loop, situations where a telemetry pipeline creates pressure
+on itself and must not explode. 
+
+## Internal telemetry receiver
+
+The Internal Telemetry Receiver or "ITR" is an OTAP-Dataflow receiver
+component that produces telemetry from internal sources. An internal
```
andborja, line 18: "receives" instead of "produces"

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+telemetry pipeline consists of one or more ITR components and any of
+the connected processor and exporter components reachable from ITR
+source nodes.
+
+To begin with, every OTAP-Dataflow comonent is configured with an
```
andborja, line 23: *component

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+internal telemetry SDK meant for primary instrumentation of that
+component. Components are required to exclusively use the internal
+telemetry SDK for self-diagnostics, as they are considered first party
+in this exchange.
+
+The internal telemetry receiver is the SDK's counterpart, making it
+second party as it is responsible for routing internal telemetry. The
+ITR cannot use the internal telemetry SDK itself, an invisible member
+of the pipeline. The ITR can be instrumented using third-party
+instrumentation (e.g., `tracing`, `log` crates) provided it can
```
andborja: why? Can we leave 3p instrumentation mechanisms out? That could go out of control pretty easily

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+## Pitfall avoidance
+
+The OTAP-Dataflow engine is safeguarded against many self-induced
+telemetry pitfalls, as follows:
+
+- OTAP-Dataflow components reachable from an ITR cannot be configured
+  to send to an ITR node. This avoids a direct feedback cycle for
+  internal telemetry because the components cannot reach
+  themselves. For example, ITR and downstream components may be
+  configured for raw logging, no metrics, etc.
```
andborja, line 46: Can you elaborate on "raw logging"? They are just another component, so they might be "configured" to produce telemetry. What the framework cannot do is to route its telemetry to an ITR.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+- ITR instances share access to one or more threads with associated
+  async runtime. They use these dedicated threads to isolate internal
+  telemetry processes that use third-party instrumentation.
```
andborja, line 49: I don't follow this part. The ITR is just another receiver, so it has its own threads as any other receiver.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+- A thread-local variable is used to redirect third-party
+  instrumentation in dedicated internal telemetry threads. Internal
+  telemetry threads automatically configure a safe configuration.
+- Components under observation (non-ITR components) have internal
+  telemetry events routed queues in the OTAP-Dataflow pipeline on the
```
andborja: When you say "queues", are you talking about channels? => flume?

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+  same core, this avoids blocking the engine. First-party
+  instrumentation will be handled on the CPU core that produced the
+  telemetry under normal circumstances. This isolates cores that are
+  able to process their own internal telemetry.
```
andborja, line 59: I'm not following this part. Can you elaborate?

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+- Option to fall back to no-op, a non-blocking global provider, and/or
+  raw logging.
```
andborja: Same here. Can you elaborate?

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+
+## OTLP-bytes first
+
+As a key design decision, the OTAP-Dataflow internal telemetry data
```
andborja, line 66: This also implies that we will create and send one message (OTLP) per event (will not batch them).

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+these bytes, we will build a custom [Tokio `tracing`
+Event][TOKIOEVENT] handler to produce OTLP bytes before dispatching to
+an internal pipeline, used (in different configurations) for first and
+third-party instrumentation.
+
+[TOKIOEVENT]: https://docs.rs/tracing/latest/tracing/struct.Event.html
+
+## Raw logging
+
+We support formatting events for direct printing to the console from
```
andborja, line 84: Why? This should be configurable if the user does not want verbosity in the output or does not have any reliable mechanism to capture it (Kubernetes scenario)

andborja: In general, console should be considered "yet another exporter"

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+OTLP bytes, based on `otap_df_pdata::views::logs::LogsDataView` and
+associated types, a zero-copy approach. We refer to this most-basic
+form of printing to the console as raw logging because it is a safe
+configuration early in the lifetime of a process.
+
+## Routing
+
+The two internal logs data paths are:
+
+- Third-party: Tokio `tracing` global subscriber: third-party log
```
andborja, line 98: Why not getting rid of the global subscriber? We can have more control over it. We can pass the telemetry objects wherever it is needed (from the thread variable?)

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+we can transcode into OTel-Arrow batches.
+
+The `otap_df_pdata` crate currently has no OTLP bytes encoder for
+directly accepting `otap_df_pdata::views::*` inputs (note the
+OTAP-records-to-OTLP-bytes function bypasses the views and encodes
+bytes directly). Therefore, this project implies we extend or refactor
+`otap_df_ptdata` with an OTLP bytes encoder for its views interfaces.
+
+Then, `TracingLogRecord` implements the log record view, we will encode
+the reocrd as OTLP bytes by encoding the view.
```
andborja: *record

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+
+### Stateful OTLP bytes encoder for repeated LogRecordViews
```
andborja: Which component does this? The ITR? Is it just another processor added after it? Is it during production? (the macro) => this might not be the same thread

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+### Global logs collection thread 
+
+An OTAP-Dataflow engine will run at least one global logs collection
+thread. These threads receive encoded (OTLP bytes) log events from
+various locations in the process. The global logs collection thread is
+special because it sets a special anti-recursion bit in the
+thread-local state to prevent logging in its own export path
+
+The global logs collection thread is configured as one (or more, if
+needed) instances consuming logs from the global Tokio `tracing`
```
andborja, line 152: What if another library is used by the 3p component? I think those should be left out so we can have more control over the telemetry.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+
+## Example configuration
+
+```yaml
+service:
+  telemetry:
+    logs:
+      level: info
+      internal_collection:
+        enabled: true
```
andborja, line 185: It is preferred at configuration level if present, it is enabled (instead of the flag).

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,210 @@
...
+blocking the effect handler.
+
+When a telemetry event is routed directly, as in this case and
+`send_message()` succeeds, it means there was queue space to accept
+the log record on the same core. When this fails, the configurable
+telemetry router will support options to use global logs collection
+thread, a raw logger, or do nothing (dropping the internal log
+record).
+
+## Example configuration
```
andborja, line 177: This might be more component level, right? (the telemetry producer)

jmacd:
I've read and appreciated all the feedback. 
A big portion of this was broken into #1735 and #1741.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: It would be in the effect handler.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: In #1741 I emphasized the two different diagnostic paths, one for 3rd party instrumentation and one for components to use directly. Definitely agree.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: Moving to #1735.

Issue #1575, opened 2025-12-10
Perftest - track binary size of df_engine


cijothomas:
A simple thing to incorporate into perf runs. When we add more components, we need to keep an eye on the binary size. (for various architectures)

ThomsonTan:
@cijothomas could this be closed by #1643 which had been merged?

cijothomas:
https://github.com/open-telemetry/otel-arrow/blob/main/docs/benchmarks.md#9-binary-size yes we have this and can be closed.

PR #1673, opened 2025-12-20
DF Engine binary size tracking

cijothomas:
https://github.com/open-telemetry/otel-arrow/issues/1575

### DIFF
--- a/collector/cmd/otelarrowcol/go.mod
+++ b/collector/cmd/otelarrowcol/go.mod
@@ -216,7 +216,7 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20250825161204-c5933d9347a5 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20250825161204-c5933d9347a5 // indirect
 	google.golang.org/grpc v1.76.0 // indirect
-	google.golang.org/protobuf v1.36.10 // indirect
+	google.golang.org/protobuf v1.36.11 // indirect
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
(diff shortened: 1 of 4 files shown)


cijothomas:
## Dataflow Engine Binary Size
| Platform | Size |
|----------|------|
| linux-amd64 | 510MiB |
| linux-arm64 | 504MiB |

works fine! Will expand the matrix to include with and without "--all-features" in a future iteration.

lalitb:
Nice work! Have couple of suggestions:

- the current `Dockerfile` [defaults](https://github.com/open-telemetry/otel-arrow/blob/1efee73c2ef92e2d7c2957928945dd5ece06820d/rust/otap-dataflow/Dockerfile#L28) to `target-cpu=native`, which optimizes the binary for the specific CPU's instruction set. And since GitHub runners are heterogeneous within each architecture  - we may get Intel Xeon one day, AMD EPYC another day for x86_64 runners - this can produces different `native` binaries (so different size) even if the code hasn’t changed. The fix  can be to override `target-cpu` to a stable baseline (e.g., `x86-64` for x86_64 runner, and `generic` for arm64 runner) via the CI YAML build arguments. 
- Also, good to pass build arguments to strip the binary to ensure we measure the actual runtime size, as currently the binary has [debug symbols](https://github.com/open-telemetry/otel-arrow/blob/1efee73c2ef92e2d7c2957928945dd5ece06820d/rust/otap-dataflow/Cargo.toml#L258) for profiling.

cijothomas:
> Nice work! Have couple of suggestions:
> 
> * the current `Dockerfile` [defaults](https://github.com/open-telemetry/otel-arrow/blob/1efee73c2ef92e2d7c2957928945dd5ece06820d/rust/otap-dataflow/Dockerfile#L28) to `target-cpu=native`, which optimizes the binary for the specific CPU's instruction set. And since GitHub runners are heterogeneous within each architecture  - we may get Intel Xeon one day, AMD EPYC another day for x86_64 runners - this can produces different `native` binaries (so different size) even if the code hasn’t changed. The fix  can be to override `target-cpu` to a stable baseline (e.g., `x86-64` for x86_64 runner, and `generic` for arm64 runner) via the CI YAML build arguments.
> * Also, good to pass build arguments to strip the binary to ensure we measure the actual runtime size, as currently the binary has [debug symbols](https://github.com/open-telemetry/otel-arrow/blob/1efee73c2ef92e2d7c2957928945dd5ece06820d/rust/otap-dataflow/Cargo.toml#L258) for profiling.


After removing debug symbols.

## Dataflow Engine Binary Size
| Platform | Size |
|----------|------|
| linux-amd64 | 80MiB |
| linux-arm64 | 68MiB |

cijothomas:
> Is there any specific reason this isn't just another job under [pipeline-perf-test-nightly](https://github.com/open-telemetry/otel-arrow/blob/main/.github/workflows/pipeline-perf-test-nightly.yml)?

We can merge it with it after we polish this a bit further (to cover windows, + feature-flags etc.) Nightly perf tests are run on fixed architecture/hardware, and didn't want to mix this up with that.

Issue #1678, opened 2025-12-22
Add HTTP proxy support for OTAP and OTLP gRPC exporters


lalitb:
Add support for routing OTAP and OTLP gRPC exporter traffic through **HTTP proxies** using the `HTTP CONNECT` tunneling method. This enables telemetry export from environments behind corporate proxies or firewalls.

This also bring parity with the proxy support in opentelemetry-collector - [OTel Collector Exporter Proxy Support](https://github.com/open-telemetry/opentelemetry-collector/blob/v0.142.0/exporter/README.md#proxy-support).   In the Go collector, proxy support is provided automatically through Go's [`net/http`](https://pkg.go.dev/net/http) package, which handles `HTTP_PROXY` environment variables transparently. However, the Rust gRPC library ([`tonic`](https://github.com/hyperium/tonic)) does not include built-in proxy support, requiring explicit implementation of the HTTP CONNECT protocol for tunneling gRPC traffic through proxies.

PR #1679, opened 2025-12-22
feat: Add HTTP proxy support for OTLP/OTAP gRPC exporters

lalitb:
fixes: #1678

  ## Summary

  Implements HTTP CONNECT proxy support for OTLP/OTAP gRPC exporters:

  - `HTTP/1.1 CONNECT` tunneling to the proxy for both `http://` targets (gRPC over h2c inside the tunnel) and `https://` targets (TLS inside the tunnel).
  - Environment variable support (`HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY`, `ALL_PROXY`)
  - NO_PROXY with CIDR notation (e.g., `192.168.0.0/16`, `10.0.0.0/8`)
  - Configures TCP socket options (`TCP_NODELAY` + `keepalive`). When using a proxy, these are applied to the TCP connection to the proxy (the same connection that carries the CONNECT tunnel).
  - Explicit rejection of `https://` proxy URLs with helpful error messages

  ## Why OTLP/OTAP exporters need custom proxy implementation

  **Azure Monitor and Geneva exporters** use the `reqwest` HTTP client, which provides built-in proxy support via `reqwest::Proxy::all()` - no custom code needed.

  **OTLP/OTAP exporters** use gRPC (`tonic`), which:
  - Has no built-in proxy support
  - Requires custom TCP connectors with `tower::service_fn`
  - Needs manual HTTP CONNECT tunnel implementation for proxy traversal

  This PR implements the missing proxy infrastructure for gRPC-based exporters.

  ## Changes

  - `proxy.rs`: HTTP/1.1 CONNECT tunnel implementation with socket option handling
  - `client_settings.rs`: Integration with tonic via custom connector
  - Dependencies: Added `socket2` and `ipnet` crates

## How HTTP CONNECT Tunneling Works

**Step 1: THE HANDSHAKE**
The Exporter creates a TCP connection to the Proxy and sends a plaintext request. The Proxy establishes a TCP leg to the Backend.

```
┌───────────┐  TCP + HTTP/1.1 CONNECT backend:PORT       ┌───────────┐
│ Exporter  │ ─────────────────────────────────────────>.│   Proxy   │
└───────────┘ <─────── 200 Connection Established ────── └───────────┘
```

**Step 2: THE DATA TUNNEL**

Once the 200 is received, the Exporter uses the socket for the actual OTLP data. The Proxy merely moves bytes back and forth.

```
┌───────────┐       ┌───────────┐       ┌───────────────┐
│ OTLP/OTAP │  TCP  │   Proxy   │  TCP  │    Backend    │
│ Exporter  │══════>│ (relays)  │══════>│               │
└───────────┘       └───────────┘       └───────────────┘
          ║                                     ║
          ╚═════════════════════════════════════╝
     On-the-wire protocol inside the tunnel (opaque to proxy):
     - Case 1 (TLS): TLS + gRPC over HTTP/2 (HTTP/2 negotiated via ALPN)
     - Case 2 (h2c): gRPC over HTTP/2 cleartext (HTTP/2 prior knowledge)
```

## Test Setup:

 The implementation was verified via a manual end-to-end test setup (not included in the PR). The architecture below was used to validate proxy traversal for both h2c and TLS traffic:

  ```
  ┌─────────────────┐
  │ Fake Data Gen   │ Generates test telemetry (logs/traces/metrics)
  └────────┬────────┘
           │
           ▼
  ┌─────────────────┐
  │ OTAP Exporter   │ gRPC client with proxy configuration
  │ (df_engine)     │ • HTTP_PROXY=http://localhost:8080
  │                 │ • grpc_endpoint=http://remote:4317
  │                 │ • admin_port: 8081 (changed to avoid proxy conflict)
  └────────┬────────┘
           │
           │ HTTP CONNECT tunneling
           ▼
  ┌─────────────────┐
  │   mitmproxy     │ Intercepts and logs proxy traffic
  │   :8080         │ (validates CONNECT requests are working)
  └────────┬────────┘
           │
           ▼
  ┌─────────────────┐
  │ OTel Collector  │ Receives telemetry via gRPC
  │   :4317         │
  └────────┬────────┘
           │
           ▼
  ┌─────────────────┐
  │ Debug Exporter  │ Displays received data
  └─────────────────┘
  ```


rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -196,6 +223,123 @@ impl GrpcClientSettings {
...
+        // Capture settings at creation time.
+        // Note: If GrpcClientSettings are modified after this connector is created,
+        // those changes will NOT be reflected in the connector.
+        let tcp_nodelay = self.tcp_nodelay;
+        let tcp_keepalive = self.tcp_keepalive;
+        let tcp_keepalive_interval = self.tcp_keepalive_interval;
+        let tcp_keepalive_retries = self.tcp_keepalive_retries;
+
+        service_fn(move |uri: http::Uri| {
+            let proxy = Arc::clone(&proxy);
```
lalitb, line 282: We use `Arc` because tonic/tower requires `Send + 'static` for connectors.

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -196,6 +223,123 @@ impl GrpcClientSettings {
...
+    + 'static {
+        // Capture settings at creation time.
+        // Note: If GrpcClientSettings are modified after this connector is created,
+        // those changes will NOT be reflected in the connector.
+        let tcp_nodelay = self.tcp_nodelay;
+        let tcp_keepalive = self.tcp_keepalive;
+        let tcp_keepalive_interval = self.tcp_keepalive_interval;
+        let tcp_keepalive_retries = self.tcp_keepalive_retries;
+
+        service_fn(move |uri: http::Uri| {
```
lalitb, line 281: This closure runs per TCP connection, not per RPC. With HTTP/2 multiplexing, connections are long-lived (startup + reconnects only), so the cost here is negligible. The cost here involved are: 
- `Arc::clone` for proxy object, 
- `get_proxy_for_uri` lookup
- string operations in `should_bypass` (have TODO in code to optimize this later).

Also this code path is only invoked when a proxy is configured. Without a proxy, tonic uses its default connector directly, so above overhead is not there.

lalitb:
this is ready for review now.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+    /// Fallback proxy URL for all protocols.
+    /// If not set, reads from ALL_PROXY/all_proxy environment variable.
+    #[serde(default)]
+    pub all_proxy: Option<String>,
+
+    /// Comma-separated list of hosts that should bypass the proxy.
+    /// Supports wildcards (e.g., "*.local,localhost,127.0.0.1").
+    /// If not set, reads from NO_PROXY/no_proxy environment variable.
+    #[serde(default)]
+    pub no_proxy: Option<String>,
```
lquerel, line 168: Non-blocking for this PR since URL redaction during logging is already handled, but it would be good to use the Newtype pattern to properly manage sensitive data. For example, this configuration could declare:

```rust
#[derive(Debug, Clone, Default, Deserialize)]
#[serde(deny_unknown_fields)]
#[doc(hidden)]
pub struct ProxyConfig {
    #[serde(default)]
    pub http_proxy: Option<SensitiveUrl>, // Changed from String

    #[serde(default)]
    pub https_proxy: Option<SensitiveUrl>, // Changed from String

    #[serde(default)]
    pub all_proxy: Option<SensitiveUrl>, // Changed from String

    #[serde(default)]
    pub no_proxy: Option<String>, // Can stay String as it's a host list
}
```

This would give us a unified, default approach for URL sanitization. The `SensitiveUrl` type could implement `Display` and `Debug` with redaction by default. An explicit `expose` method could be added for cases where the original value needs to be accessed.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+    }
+}
+
+/// Parses a proxy URL and returns (host, port, auth_header_value).
+fn parse_proxy_url(proxy_url: &str) -> Result<(String, u16, Option<String>), ProxyError> {
+    let uri: Uri = proxy_url
+        .parse()
+        .map_err(|_| ProxyError::InvalidProxyUrl(proxy_url.to_string()))?;
+
+    // Reject https:// proxy URLs - we don't support TLS to the proxy server itself
```
lquerel, line 403: Why this limitation, if the target audience for this HTTP CONNECT based tunneling is large enterprises?

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+    /// - "example.com" - exact hostname matching
+    /// - "192.168.1.1" - exact IP matching
+    /// - "192.168.0.0/16" - CIDR notation for IP ranges
+    fn should_bypass(&self, host: &str) -> bool {
+        let no_proxy = match &self.no_proxy {
+            Some(np) => np,
+            None => return false,
+        };
+
+        // TODO(perf): Pre-parse NO_PROXY rules to avoid allocations in hot path
```
lquerel, line 317: It would be good to create an issue for tracking, if it hasn't already been done.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+        .map_err(|_| ProxyError::InvalidResponse("failed to reunite stream".to_string()))?;
+    Ok(stream)
+}
+
+/// Applies TCP socket options (nodelay, keepalive) to a stream.
+///
+/// This function performs a series of conversions (tokio -> std -> socket2 -> std -> tokio)
+/// to apply socket options that are not directly exposed by tokio's TcpStream.
+/// Specifically, `socket2` is required to set detailed keepalive parameters (interval, retries).
+fn apply_socket_options(
```
lquerel, line 582: You could replace this chain of conversions `tokio -> std -> socket2 -> std -> tokio` by creating a socket2 first, configuring it, and then converting it to a Tokio TcpStream. This would simplify the code and would most likely be more efficient.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+        format!("[{}]", target_host)
+    } else {
+        target_host.to_string()
+    };
+
+    let mut connect_request = format!(
+        "CONNECT {formatted_target}:{target_port} HTTP/1.1\r\n\
+         Host: {formatted_target}:{target_port}\r\n\
+         Connection: Keep-Alive\r\n"
+    );
```
lquerel, line 478: Adding a User-Agent header will make this code more universal. Some proxies may block requests that don't have a user agent defined.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+pub struct ProxyConfig {
+    /// HTTP proxy URL (e.g., "http://proxy.example.com:3128")
+    /// If not set, reads from HTTP_PROXY/http_proxy environment variable.
+    #[serde(default)]
+    pub http_proxy: Option<String>,
+
+    /// HTTPS proxy URL (e.g., "http://proxy.example.com:3128")
+    /// If not set, reads from HTTPS_PROXY/https_proxy environment variable.
+    #[serde(default)]
+    pub https_proxy: Option<String>,
```
lquerel: To remove any ambiguity, it would be helpful to clarify that `https_proxy` does not represent an HTTPS connection to the proxy, but an HTTPS connection inside the tunnel.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+                    } else {
+                        url.to_string()
+                    }
+                } else {
+                    url.to_string()
+                };
+                otap_df_telemetry::otel_debug!(
+                    "Proxy.Using",
+                    proxy = safe_url,
+                    target = uri.to_string()
```
lquerel, line 252: Could that be a potential security leak? See my other comment on using a Newtype pattern to reduce this type of risk.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+    connect_request.push_str("\r\n");
+
+    if proxy_auth.is_some() {
+        otap_df_telemetry::otel_debug!(
+            "Proxy.ConnectRequest",
+            target = format!("{formatted_target}:{target_port}"),
+            has_auth = true
+        );
+    } else {
+        otap_df_telemetry::otel_debug!("Proxy.ConnectRequest", request = connect_request.trim());
```
lquerel: Could there be a security risk in reporting `connect_request`?

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+        let _ = config.has_proxy();
+    }
+
+    #[test]
+    fn test_no_proxy_bypass() {
+        let config = ProxyConfig {
+            http_proxy: Some("http://proxy:3128".to_string()),
+            https_proxy: None,
+            all_proxy: None,
+            no_proxy: Some("localhost,*.local,127.0.0.1,.example.com".to_string()),
```
lquerel, line 714: You should add an example with host:port to validate this option which I'm not sure is properly supported.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+    /// - "192.168.1.1" - exact IP matching
+    /// - "192.168.0.0/16" - CIDR notation for IP ranges
+    fn should_bypass(&self, host: &str) -> bool {
+        let no_proxy = match &self.no_proxy {
+            Some(np) => np,
+            None => return false,
+        };
+
+        // TODO(perf): Pre-parse NO_PROXY rules to avoid allocations in hot path
+        let host_lower = host.to_lowercase();
```
lquerel: You should consider adding `.trim_end_matches('.')` as `example.com.` is a valid domain.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+
+    otap_df_telemetry::otel_debug!("Proxy.ConnectResponse", status_line = status_line.trim());
+
+    // Parse "HTTP/1.1 200 Connection established"
+    let parts: Vec<&str> = status_line.trim().splitn(3, ' ').collect();
+    if parts.len() < 2 {
+        return Err(ProxyError::InvalidResponse(format!(
+            "invalid status line: {status_line}"
+        )));
+    }
```
lquerel: For a more robust parsing method, you probably should consider that multiple ASCII spaces/tabs can separate those words.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+    pub fn has_proxy(&self) -> bool {
+        self.http_proxy.is_some() || self.https_proxy.is_some() || self.all_proxy.is_some()
+    }
+}
+
+/// Parses a proxy URL and returns (host, port, auth_header_value).
+fn parse_proxy_url(proxy_url: &str) -> Result<(String, u16, Option<String>), ProxyError> {
+    let uri: Uri = proxy_url
+        .parse()
+        .map_err(|_| ProxyError::InvalidProxyUrl(proxy_url.to_string()))?;
```
lquerel: proxy_url could leak in the error logs

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+
+    // Reject https:// proxy URLs - we don't support TLS to the proxy server itself
+    if uri.scheme_str() == Some("https") {
+        return Err(ProxyError::InvalidProxyUrl(format!(
+            "https:// proxy URLs are not supported (proxy URL: {}). \
+Use http:// instead - the CONNECT tunnel will still encrypt \
+traffic to the final destination for https:// targets.",
+            proxy_url
+        )));
```
lquerel, line 410: proxy_url could leak in the error logs

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -0,0 +1,939 @@
...
+        let stream = TcpStream::connect((proxy_host.as_str(), proxy_port))
+            .await
+            .map_err(|e| {
+                otap_df_telemetry::otel_warn!(
+                    "Proxy.ConnectFailed",
+                    host = proxy_host,
+                    port = proxy_port,
+                    error = e.to_string()
+                );
+                ProxyError::ProxyConnectionFailed(e)
```
lquerel, line 665: Not sure but depending of what is reported in io::Error we might here leak some sensitive data.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Agreed, this could be limitation for enterprises -  created issue #1710 to track this. The Go-based collector doesn’t support it today, so the current PR is intended to stay feature-consistent with that behavior.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Good point, added #1711 to track this.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Good suggestion! You're right that creating the `socket2` first, configuring it, and then converting to tokio would eliminate the conversion chain.

  The main reason I didn't go that route is the async connect handling - since `socket2` is synchronous, we'd need to properly handle non-blocking connect (dealing with `EINPROGRESS`/`WouldBlock`, waiting for writability, checking `SO_ERROR`, etc.). The current approach lets Tokio handle the async connect, and we just use `socket2` to apply the keepalive options on an already-established socket.

  Since this only runs on connection establishment (not per-RPC), I felt the simpler approach is fine for now. Happy to track it as a follow-up issue if you think it's worth revisiting.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Great point - the naming is a bit misleading. In our implementation `https_proxy` means “the proxy to use when the target is HTTPS”; we still connect to the proxy over plain HTTP and use CONNECT to create a tunnel, then do TLS to the target inside that tunnel. Will update it better.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Good catch! I'll update it to always log structured fields (target + has_auth flag) rather than the raw request string.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Good point — `example.com`. is a valid hostname. I’ll trim trailing `.` during NO_PROXY matching so example.com and example.com. both match.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Good call. Will switch from `splitn` to `split_ascii_whitespace` to handle this.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Good point. `io::Error::to_string()` can include environment details, will change to avoid printing the full error string and instead log only `ErrorKind + raw_os_error` (host/port are already logged).

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Done. I've introduced a `SensitiveUrl` struct that handles redaction in its `Debug` and `Display` implementations, and updated `ProxyConfig` to use it for the proxy URL fields.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Added `User-Agent: otap-dataflow` to the CONNECT request headers.  And included a TODO to make it configurable in the future.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Updated the documentation to explicitly clarify that https_proxy refers to the proxy used for HTTPS targets, and that the connection to the proxy itself is established over plain HTTP.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Fixed by using the `SensitiveUrl` type, which now automatically redacts credentials when logged.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: You were right, it wasn't properly supported. I've updated the logic to correctly handle `host:port` entries (including IPv6) and added a test case to verify it.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Done.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Fixed - now using `SensitiveUrl` to redact the proxy URL before including it in the error message.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Fixed - now using `SensitiveUrl` to redact the proxy URL before including it in the error message.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Done as above,

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Added `.trim_end_matches('.')` to normalize hostnames during NO_PROXY matching, ensuring that FQDNs with trailing dots are handled correctly.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: Done as above.

lalitb:
> It's really great to add this support to both our OTLP and OTAP exporters. I made a few suggestions, mainly to make this functionality more secure and more robust with certain proxies.
> 
> I think it would be a good idea to add complete documentation in a markdown file describing this tunneling capability. For now, we have a very detailed PR description, but unfortunately it won't be seen by future users.

 Thanks for the thorough review and great suggestions! I have tried incorporating them in code and with tracking issues. Also copied the architecture details  from PR desc to `./docs/proxy-support.md`.

lquerel (approved):
LGTM

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
jmacd: Also non-blocking we may consider using https://github.com/microsoft/oxidizer/blob/main/crates/data_privacy/README.md in the future

PR #1682, opened 2025-12-22
Upgrade Collector dependencies to 0.142.0 and misc Go modules

drewrelmas:
Supersedes #1677, #1676, and #1675.

Touched manually since already upgrading Collector dependencies.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ COPY . .
 ENV CGO_ENABLED=0
 
 # Note the version MUST MATCH otelarrowcol-build.yaml
-RUN go install go.opentelemetry.io/collector/cmd/builder@v0.140.0
+RUN go install go.opentelemetry.io/collector/cmd/builder@v0.142.0
 
 # This command generates main.go, go.mod but does not update deps.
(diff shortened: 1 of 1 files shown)


collector/otelarrowcol-build.yaml
```diff
@@ -27,29 +27,28 @@ dist:
 
 exporters:
   # This is the core OpenTelemetry Protocol with Apache Arrow exporter
-  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/otelarrowexporter v0.140.1
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/exporter/otelarrowexporter v0.142.0
```
cijothomas, line 30: https://github.com/open-telemetry/otel-arrow/blob/main/Dockerfile#L16-L17 This also need a bump to match.
(we need find some way to make it easy for them to be in sync in future)

drewrelmas: We should be able to just update the `make builder` command to do this - have a change in draft locally.

PR #1683, opened 2025-12-22
fix: Small (but anoying) fixes to tests and code

andborja:
Small (but anoying) fixes to tests and code


rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -272,6 +272,7 @@ const fn default_keep_alive_while_idle() -> bool {
 #[allow(missing_docs)]
 mod tests {
     use super::*;
+    #[cfg(feature = "experimental-tls")]
```
lalitb, line 275: Thanks, I am going to remove this feature flag in subsequent PR, so the TLS is enabled as default.

Issue #1667, opened 2025-12-19
Support conditionally applying pipeline to some data in columnar query engine


albertlockett:
Imagine a pipeline wishing to express: 

```kql
logs | 
    if (severity == ERROR) 
    then (attributes["important"] = true | routeto important_out_port)
```
(above is pseudocode, might not be the actual syntax we'll use).

Needed:
- support this kind of conditional application of pipeline to data
- parser support

PR #1684, opened 2025-12-22
Columnar query engine support conditionally executing pipeline stages

albertlockett:
part of #1667 

Adds a new `ConditionalDataExpression` to the transformation expression AST for applying transformation `PipelineStages` to some subset of rows that match a condition. This is used to implement a `ConditionalPipelineStage`, which operates like `if/else if/else` control flow.

For example, imagine we had a hypothetical syntax like:
```kql
logs |
     set attributes["important"] = "very" | set attributes["triggers_alarm"] = true
  } else if (severity_text == "WARN) {
     set attributes["important"] = "somewhat"
  } else {
     set attributes["important"] = "no"
  }
```

This could be modeled using our conditional expression like:
```rs
// this is pesudocode to illustrate what each field represents
Conditional {
  branches: [
     ConditionalBranch {
       condition: "severity_text == \"ERROR\"",
       expressions: [ 
         "set attributes[\"important\"] = \"very\"",
         "set attributes[\"triggers_alarm\"] = true"  
      ],
     },
     ConditionalBranch {
       condition: "severity_text == \"WARN\"",
       expressions: [
        "set attributes[\"important\"] = \"somewhat\"
      ],
     },
  ],
  default_branch: Some([
    "set attributes["important"] = \"no\""
  ])
}
```

Note there is currently no parser support for a language syntax that creates this variant of `DataExpression`. That will happen in a future PR


rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs
```diff
@@ -0,0 +1,507 @@
...
+                        .execute(
+                            default_branch_batch,
+                            session_ctx,
+                            config_options,
+                            task_context.clone(),
+                        )
+                        .await?;
+                }
+            }
+            branch_results.push(default_branch_batch);
```
lalitb, line 174: If I understand the code properly. the logic here will change the order of the records - the records would now be grouped based on the branch they matched with. This should be fine I believe for most cases, but should also add a quick note in the code mentioning the behavior ?

lalitb:
LGTM! Really liked the `attributes["important"]` example in the PR description - made it super easy to understand the if/else semantics at a glance.

rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs
albertlockett: Thanks @lalitb. You're right about what the resulting order will be. I added a note simply stating that the order might not be preserved 5a18a400. I don't want want the resulting order by branch match to be part of the explicit contract.

Issue #1482, opened 2025-11-26
Reducing Golang dependencies


jmacd:
We see alerts like https://github.com/open-telemetry/otel-arrow/pull/1452; these come from checking-in the `otelarrowcol` package which is generated with a set of OTel Collector dependencies.

We can reduce this exposure by removing the collector/cmd/otelarrowcol directory; we can test using a Collector-Contrib distribution (Docker image), instead of building it each time in the repository, to avoid many of the dependency issues found in collector/cmd/otelarrowcol. 

The go/go.mod dependency set presents a number test-only dependencies that we should aim to gradually replace with Phase 2 equivalents. Whereas we have used many analysis libraries (e.g., hyperloglog, hdrhistogram) that could be replaced with OTAP dataflow equivalents. Our goal should be that the OTel Collector-Contrib repository does not indirectly depend on this stuff from its dependency on otel-arrow/go.

drewrelmas:
@jmacd I would like to take a pulse on this again from you as I look through my backlog and notice I still haven't addressed #1469. That work is required for us to properly ensure Go updates make it into the generated collector in our repo. If we instead prefer to take up the removal you propose, then we don't need to do #1469.

drewrelmas:
This should be closed

PR #1685, opened 2025-12-22
Update Makefile builder command to replace in build.yaml and Dockerfile

drewrelmas:
Following up on https://github.com/open-telemetry/otel-arrow/pull/1682#discussion_r2640864598

In 99% of cases, the same version should be used for:
* Builder
* Collector/Collector-Contrib references in `otelarrowcol-build.yaml`
* Main `Dockerfile`

This addition to `builder` command would be invoked every time `genotelarrowcol` is invoked but could be brittle if more references are added elsewhere in the repo.

#1482 is probably the correct long-term approach to stop building our own collector in this repo.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,8 @@ WORKDIR /otel-arrow
 COPY . .
 ENV CGO_ENABLED=0
 
-# Note the version MUST MATCH otelarrowcol-build.yaml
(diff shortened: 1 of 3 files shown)


albertlockett (approved):
👍 thanks again @drewrelmas

PR #1687, opened 2025-12-23
Fix bugs transforming attributes when all attrs deleted

### LINKED ISSUE #1681: `transform_attributes` produces error when all attributes are deleted
When specifying an attribute transformation that deletes all attributes (either by the attributes processor, or the columnar query engine), the underlying function that gets called will produce an error: `error transforming attributes Failed to write record batch`.

The source of the error is here:
https://github.com/open-telemetry/otel-arrow/blob/1efee73c2ef92e2d7c2957928945dd5ece06820d/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L1747

and it's because `concat` needs at least one slice. 

We should fix this to correctly produce an empty array or otherwise handle properly. In fact, the proper handling is probably to remove the attributes record batch from the OTAP batch entirely, so we should add tests that the situation is handled this effect.

Also, we should include the source in the error string:
https://github.com/open-telemetry/otel-arrow/blob/1efee73c2ef92e2d7c2957928945dd5ece06820d/rust/otap-dataflow/crates/pdata/src/error.rs#L109-L113

albertlockett:
closes #1681 

- Fixes a bug where calling `transform_attributes` would produce an error when attribute keys are dictionary encoded and all attributes are deleted.
- Adds proper handling of the situation where this is called (in `AttributesProcessor` and `AttributesTransformFilterStage`), which is that the `RecordBatch` for the attribute's payload type should be removed from the OTAP batch.
- Consolidates the transform attributes into a single function. Before this, we had two functions `transform_attributes` and `transform_attributes_with_stats` that had almost identical logic (the differences just being stats returned, and comments)
- Fixes the to_string method on various `otap_df_pdata::error::Error` variants that wrapped `ArrowError` but didn't expose the underlying error in their error message.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/attributes_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/attributes_processor.rs
@@ -1005,7 +1005,6 @@ mod tests {
             vec![],
             vec![
                 KeyValue::new("a", AnyValue::new_string("1")),
-                KeyValue::new("a", AnyValue::new_string("2")),
                 KeyValue::new("b", AnyValue::new_string("1")),
             ],
         );


rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -993,6 +997,67 @@ mod tests {
...
     }
+
+    #[test]
+    fn test_delete_all_attributes() {
+        let input = build_logs_with_attrs(
+            vec![],
+            vec![],
+            vec![
+                KeyValue::new("a", AnyValue::new_string("1")),
+                KeyValue::new("a", AnyValue::new_string("2")),
```
drewrelmas: Unrelated to the actual PR, but was just surprised to see 2 attributes with the same key - what would happen if we filtered on `where attributes["a"] == "2"`?

drewrelmas (approved):

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
albertlockett: oh good catch. I'll remove the duplicate attribute just to avoid any confusion it might cause.

If we were to filter on that, we'd just get a single row for the telemetry item that had the attribute. It's not as if we materialize a log row for every attribute that matches the filter.

PR #1656, opened 2025-12-18
[PipelinePerfTest] - Collect logs from the engine

cijothomas:
Adds automatic collection and display of Docker container logs during perf test.
Logs are collected and printed at the end, separated by each container to avoid interfering with test log itself.

The previous behavior of printing logs in  "--debug" mode during container shutdown is retained. (I'll send a follow up to remove --debug option from the CI runs to avoid too much noise)

Also, enabled engine configs in perf test to do INFO level logs. (as of today, engine logs are OFF by default)

Example from local run:

```text
================================================================================
Component: df-engine
Container ID: dd2985efabfe53324ffb410f7bbd18a6584ccb804701464e9ff47543ced149e8
================================================================================
Starting pipeline on core ID set 1-2
2025-12-18T20:59:42.948630Z  INFO pipeline-core-1 pipeline_thread{core.id=1}: otap-df-otap:  name="Exporter.Start" grpc_endpoint="http://backend-service:1235" Starting OTLP Exporter
2025-12-18T20:59:42.948629Z  INFO pipeline-core-2 pipeline_thread{core.id=2}: otap-df-otap:  name="Exporter.Start" grpc_endpoint="http://backend-service:1235" Starting OTLP Exporter
2025-12-18T20:59:42.948744Z  INFO pipeline-core-2 pipeline_thread{core.id=2}: otap-df-otap:  name="Receiver.Start" Starting OTLP Receiver
2025-12-18T20:59:42.948744Z  INFO pipeline-core-1 pipeline_thread{core.id=1}: otap-df-otap:  name="Receiver.Start" Starting OTLP Receiver

================================================================================
```

We don't really have much logs in the engine, but we are adding them, and they are badly needed to investigate data loss issues like https://github.com/open-telemetry/otel-arrow/issues/1637

PR #1690, opened 2025-12-23
PipelinePerfTest - remove running with debug flags

cijothomas:
https://github.com/open-telemetry/otel-arrow/pull/1656 added ability to show logs from the containers. The "--debug" flag is also doing it, which feels quite noisy. This PR removes the "--debug" flag from all perf tests.


cijothomas:
9mins for the runs, vs 12 mins, so this is good!

PR #1691, opened 2025-12-23
feat: Add internal telemetry prometheus exporter

andborja:
Add internal telemetry configurable prometheus exporter.

### DIFF
--- a/rust/otap-dataflow/crates/config/README.md
+++ b/rust/otap-dataflow/crates/config/README.md
@@ -64,7 +64,9 @@ The telemetry configuration includes:
 
 - **Metrics**: OpenTelemetry metrics for pipeline observability
   - **Readers**: Periodic or pull-based metric readers
-  - **Exporters**: Console, OTLP (gRPC/HTTP), or custom exporters
+    - **Periodic Readers**: Export metrics at regular intervals
+    - **Pull Readers**: Expose metrics via HTTP endpoint for scraping (e.g., Prometheus)
+  - **Exporters**: Console, OTLP (gRPC/HTTP), or Prometheus
   - **Views**: Metric aggregation and transformation rules
   - **Temporality**: Delta or cumulative aggregation
 - **Logs**: Internal logging configuration
@@ -89,11 +91,20 @@ service:
       deployment.environment: "production"
     metrics:
       readers:
+        # Periodic reader - pushes metrics to OTLP endpoint
         - periodic:
             exporter:
               otlp:
                 endpoint: "http://localhost:4318"
                 protocol: "grpc/protobuf"
+            interval: "60s"
+        # Pull reader - exposes metrics for Prometheus scraping
+        - pull:
+            exporter:
+              prometheus:
+                host: "0.0.0.0"
+                port: 9090
+                path: "/metrics"
       views:
         - selector:
             instrument_name: "logs.produced"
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/metrics/readers/pull.rs
```diff
@@ -0,0 +1,166 @@
...
+        deserializer.deserialize_map(MetricsPullExporterConfigVisitor)
+    }
+}
+
+/// Prometheus Exporter configuration.
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq)]
+#[serde(deny_unknown_fields)]
+pub struct PrometheusExporterConfig {
+    /// The host address where the Prometheus exporter will expose metrics.
+    pub host: String,
```
cijothomas, line 62: might be good to have defaults for host and port

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider/prometheus_exporter_provider.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        Ok(())
+    }
+
+    /// Define the routes for the Prometheus exporter.
+    fn routes(path: &str) -> Router<Registry> {
+        Router::new().route(path, get(Self::get_metrics))
+    }
+
+    /// Handler for the `/metrics` endpoint.
```
cijothomas: nit: endpoint could be anything, not just /metrics.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider/prometheus_exporter_provider.rs
```diff
@@ -0,0 +1,244 @@
...
+        let encoder = TextEncoder::new();
+        let metric_families = registry.gather();
+        let mut buffer = Vec::new();
+        encoder
+            .encode(&metric_families, &mut buffer)
+            .map_err(|_| StatusCode::INTERNAL_SERVER_ERROR)?;
+
+        let response = Response::builder()
+            .status(StatusCode::OK)
+            .header("Content-Type", "text/plain; version=0.0.4; charset=utf-8")
```
cijothomas, line 122: we may have to support openmetrics format in the future. Not something for this PR.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client/meter_provider/prometheus_exporter_provider.rs
```diff
@@ -0,0 +1,244 @@
...
+            _ => panic!("Expected ConfigurationError for invalid path"),
+        }
+    }
+
+    #[tokio::test]
+    async fn test_prometheus_get_metrics_handler() {
+        let registry = Registry::new();
+        registry
+            .register(Box::new(
+                prometheus::Counter::new("test_counter", "A test counter").unwrap(),
```
cijothomas, line 191: lets use OpenTelemetry counter to properly test the integration. Otherwise, we are just testing `prometheus` crate mostly here...

andborja: This is testing the handler as a unit, that receives the registry as parameter. It is not an integration test.

andborja:
> How does this PR align with Joshua’s PR #1672?

It is a separate use case. This is just another exporter of the SDK integration. Internal telemetry routing is a different approach on how to route internal telemetry to a pipeline. That PR is more aligned with (second phase) https://github.com/open-telemetry/otel-arrow/pull/1663/

lquerel (approved):
Thanks for the work on this PR. I want to clarify that I'm approving it **with reservations**.

I am strongly in favor of using our pipeline engine to report our internal telemetry (more aligned with #1672 ). I see multiple advantages in this approach:

* it provides strong motivation to develop fundamental exporters, such as an OTAP-based Prometheus exporter,
* it enables native support for OTAP and, in the future, experimentation with protocol extensions to natively support multivariate metrics, which we already support internally,
* it makes it possible to leverage our future processing language, OPL, on our own internal telemetry.

**Important**: integrating Prometheus support via the SDK must not lead to the removal of the existing `/metrics` endpoint. F5 currently uses this endpoint (JSON and compact JSON formats) because of its optimized support for multivariate metrics. It is fine to remove the Prometheus format from this endpoint if desired.

Looking ahead, I think this type of development should be discussed upstream to avoid this kind of duplication. I'm fully aware that I've been absent from the last two to three SIG meetings, so I take my share of responsibility for that.

@jmacd @drewrelmas @albertlockett  for visibility

Issue #1435, opened 2025-11-14
OTAP transform function to Condense attributes


drewrelmas:
## Context

#1149 references a potential improvement for the `syslog-cef-receiver` component

> Add support for a parsing configuration that allows selective inclusion of attributes and optional renaming. Attributes not explicitly listed may either be excluded from the output or aggregated into a single string under a designated fallback attribute.

This point may be split into two different 'features':
- Attribute renaming (supported by `attributes_processor`)
- Attribute condensing (currently unsupported)

This issue proposes an alternate solution to support this behavior, focusing on the second feature.

The [current implementation](https://github.com/open-telemetry/otel-arrow/blob/1477cd3640ddf45c083cf9170288010ce185bb45/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/parser/parsed_message.rs#L289-L341) of the receiver creates individual attributes for all `extensions` (key-value pairs present in the CEF message outside of certain header values).

```rust
let mut extensions_iter = msg.parse_extensions();
while let Some((key, value)) = extensions_iter.next_extension() {
    log_attributes_arrow_records.append_key(std::str::from_utf8(key).unwrap_or_default());
    log_attributes_arrow_records
        .any_values_builder
        .append_str(value);
    attributes_count += 1;
}
```

## Use case

Various vendor-specific destinations supporting CEF data may map certain 'well-known' `extension` keys to certain output labels, like [Microsoft's `CommonSecurityLog` table](https://learn.microsoft.com/en-us/azure/sentinel/cef-name-mapping):

| CEF extension key | CommonSecurityLog field |
|-------------------|-------------------------|
| act | DeviceAction |
| app | ApplicationProtocol |
| ... | ... |

However, it is common to also have vendor-specific CEF extension key/values produced by a client device that may not have a well-known mapping. In the case of `CommonSecurityLog`, these are all condensed into an overflow `AdditionalExtensions` field.

For example, a CEF message with the following extensions:
```json
{
  "act": "DeviceActionValue",
  "app": "ApplicationProtocolValue",
  "vendorSpecificExtension1": "VendorSpecificExtension1Value",
  "vendorSpecificExtension2": "VendorSpecificExtension2Value"
}
```
should be transformed in the following way:
```json
{
  "DeviceAction": "DeviceActionValue",
  "ApplicationProtocol": "ApplicationProtocolValue",
[...]

Issue #1035, opened 2025-08-29
OTAP transform function to insert attributes


albertlockett:
In https://github.com/open-telemetry/otel-arrow/issues/813 the capability was added to rename and delete attributes. We should enhance this capability to insert attributes as well. This will eventually be used in the attributes processor.

There are three sources for the value of attributes the caller might want to insert. See [here](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/attributesprocessor) for additional details).
- [ ] static value
- [ ] value of a different attribute (e.g. from a different row in the attributes Record Batch)
- [ ] from context (e.g. from metadata, or auth information)
  - this will require some changes to OtapPdata as we'll need to keep context (metadata) from things like the gRPC request headers, BatchArrowRecord headers, information. We'll also need to modify the the gRPC receivers to populate this metadata.

PR #1695, opened 2025-12-26
feat: Initial Condense processor implementation

### LINKED ISSUE #1693: Initial implementation of Condense in experimental processor

drewrelmas:
Related to #1435

Fixes #1693

This is a basic implementation of the `Condense` behavior from the above issue that works for `LogAttrs` payload types.

This iteration currently builds an entirely new `RecordBatch` during execution. As mentioned in comments, once #1035 is completed, working in-place on the existing `RecordBatch` would be more efficient especially with respect to persisted attributes.

With a debug pipeline configuration composed of:
* `syslog_cef_receiver`
* `attributes_processor` (doing various renames and deletes)
* `condense_attributes_processor`

Sending the CEF message:
> <134>Dec 29 17:28:13 securityhost CEF:0|Security|threatmanager|1.0|100|worm successfully stopped|10|src=10.0.0.1 dst=2.1.2.2 spt=1232 dpt=80 proto=tcp act=blocked vendorspecificext1=value1 vendorspecificext2=value2

Results in the following LogRecord:
```
LogRecord #0:
   -> ObservedTimestamp: 1767029293753398998
   -> Timestamp: 1767029293000000000
   -> SeverityText: INFO
   -> SeverityNumber: 9
   -> Body: <134>Dec 29 17:28:13 securityhost CEF:0|Security|threatmanager|1.0|100|worm successfully stopped|10|src=10.0.0.1 dst=2.1.2.2 spt=1232 dpt=80 proto=tcp act=blocked vendorspecificext1=value1 vendorspecificext2=value2
   -> Attributes:
      -> AdditionalExtensions: vendorspecificext1=value1|vendorspecificext2=value2
      -> Computer: securityhost
      -> DeviceVendor: Security
      -> DeviceProduct: threatmanager
      -> DeviceVersion: 1.0
      -> DeviceEventClassId: 100
      -> Activity: worm successfully stopped
      -> LogSeverity: 10
      -> SourceIP: 10.0.0.1
      -> DestinationIP: 2.1.2.2
      -> SourcePort: 1232
      -> DestinationPort: 80
      -> Protocol: tcp
      -> DeviceAction: blocked
   -> Trace ID:
   -> Span ID:
   -> Flags: 0 
```

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
@@ -1193,7 +1193,7 @@ mod config_tests {
         }
     }
 
-        #[test]
+    #[test]
     fn test_config_parsing_destination_key_in_source_keys() {
         let cfg = json!({
             "destination_key": "condensed",


rust/otap-dataflow/Cargo.toml
```diff
@@ -196,6 +196,9 @@ experimental-tls = ["otap-df-otap/experimental-tls", "dep:rustls"]
 experimental-exporters = ["otap-df-otap/experimental-exporters"]
 geneva-exporter = ["otap-df-otap/geneva-exporter"]
 azure-monitor-exporter = ["otap-df-otap/azure-monitor-exporter"]
+# Experimental processors (opt-in)
+experimental-processors = ["otap-df-otap/experimental-processors"]
+condese-attributes-processor = ["otap-df-otap/condense-attributes-processor"]
```
lalitb:
```suggestion
condense-attributes-processor = ["otap-df-otap/condense-attributes-processor"]
```

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
```diff
@@ -0,0 +1,1068 @@
...
+                        arr.value(index)
+                    })
+                    .map(|v| v.to_string())
+                }),
+                // If needed, add handling for Map, Slice, and Bytes?
+                _ => None,
+            }
+        };
+
+        let mut parent_to_attrs: HashMap<u16, Vec<(&str, String)>> = HashMap::new();
```
lalitb, line 340: With use of HashMap, same logical data could produce different strings. Eg,
```
 Log A: "x=1;y=2"
 Log B: "y=2;x=1" 

```
 Backend searches for condensed values will miss results. Is this Intentional?

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
```diff
@@ -0,0 +1,1068 @@
...
+            .ok_or_else(|| ConfigError::InvalidUserConfig {
+                error: "destination_key is required and must be a string".to_string(),
+            })?
+            .to_string();
+
+        let delimiter = config
+            .get("delimiter")
+            .and_then(|v| v.as_str())
+            .and_then(|s| s.chars().next())
+            .ok_or_else(|| ConfigError::InvalidUserConfig {
```
lalitb, line 125: should we restrict "=" as delimiter as the string output could be confusing with `k1=v1=k2=v2`

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
drewrelmas: I don't think it is an explicit requirement for `Condense` to produce a certain order. The use case I had in mind was an 'overflow' field on whatever backend is being used. If an end user wants to do some further manipulation, I would expect them to leverage individual keys directly either with a `parse` or a simple `contains` over the `condensed` value.

i.e.
```kql
// parse specific value out
| parse condensed with * "x=" value:int ";" *
| where value == 5

// basic filter
| where condensed contains "x=5"
```

I don't really foresee a use case where it is required for multiple keys to be in specific order together inside the `condensed` value.

We could always produce a reliable order by sorting, but that incurs performance penalty.

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
drewrelmas: Great point! Pushed

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
lalitb: Thanks for clarifying the overflow field use case - that makes total sense! Should we add a code comment and/or update README.md explaining the non-deterministic behavior is intentional.

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
```diff
@@ -0,0 +1,1112 @@
...
+                    })
+                    .map(|v| v.to_string())
+                }),
+                // If needed, add handling for Map, Slice, and Bytes?
+                _ => None,
+            }
+        };
+
+        let mut parent_to_attrs: HashMap<u16, Vec<(&str, String)>> = HashMap::new();
+        let mut rows_to_preserve: Vec<usize> = Vec::new();
```
lalitb: More of the optimization suggestion (not blocker) - `parent_to_attrs` and `rows_to_preserve` are allocated on every `condense()` call. We can consider moving them to the processor struct and using `.clear()` between batches instead of reallocating

lalitb:
LGTM with the comments - nothing blocking as such.

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
drewrelmas: Thinking about this - it is easily done for `rows_to_preserve` but I'm going to leave `parent_to_attrs` alone for the moment. It currently borrows attribute key `&str` strings from the batch being processed, so if I'm thinking about lifetimes properly a reusable version would require creating owned `Strings` instead, hurting performance.

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
drewrelmas: Good idea, added a note to the Readme.

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
lalitb: Yes i thought about this too, probably 'String' would still have better performance. But right to leave it for now as this would need comparing perf results before deciding.

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/README.md
```diff
@@ -0,0 +1,123 @@
...
+cargo build --features condense-attributes-processor
+```
+
+## Configuration
+
+The processor accepts the following configuration options:
+
+| Field | Type | Required | Description |
+| ----- | ---- | -------- | ----------- |
+| `destination_key` | string | **Yes** | The name of the attribute that will contain the condensed string |
```
cijothomas, line 39: nit: Document the behavior if this name is already an attribute name.

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/README.md
```diff
@@ -0,0 +1,123 @@
+# Condense Attributes Processor
+
+**Status:** WIP
+
+## Overview
+
+The Condense Attributes Processor is an experimental OTAP processor that
+condenses multiple log attributes into a single string attribute. This is a
+niche operation useful when working with firmly-structured data shapes that
+include a specific overflow field for non-critical attributes.
```
cijothomas: Is the term "overflow" apt here?

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
```diff
@@ -0,0 +1,1119 @@
...
+use crate::pdata::OtapPdata;
+
+/// URN identifier for the Condense Attributes processor
+pub const CONDENSE_ATTRIBUTES_PROCESSOR_URN: &str = "urn:otel:condense_attributes:processor";
+
+/// Configuration for the Condense Attributes processor
+///
+/// `destination_key`: The key under which the condensed attributes will be stored.
+/// `delimiter`: The character used to separate individual attribute values in the condensed string.
+///                Note: Cannot be '=' as it would create ambiguous output (e.g., k1=v1=k2=v2).
```
cijothomas, line 48: nit: does providing "=" creates error at runtime? or we can detect it at startup?

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
```diff
@@ -0,0 +1,1119 @@
...
+            Message::PData(pdata) => {
+                let signal = pdata.signal_type();
+                let (context, payload) = pdata.into_parts();
+
+                let mut records: OtapArrowRecords = payload.try_into()?;
+
+                let condensed_records: OtapArrowRecords =
+                    match signal {
+                        SignalType::Logs => match self.condense(&mut records) {
+                            Ok(_condensed) => records,
```
cijothomas, line 544: wondering why the returned `_condensed` is ignored?

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/README.md
drewrelmas: This was an unhandled edge case - for the moment I've disallowed `destination_key` being included in `source_keys` and `exclude_keys` and simply skip existing attribute of the same name. Left a TODO to properly instrument notice for the user when this happens.

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/README.md
drewrelmas: Swapped to use the term `catch-all` - do you think this brings more clarity?

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
drewrelmas: Yes, we detect at startup and return `InvalidUserConfig`.

rust/otap-dataflow/crates/otap/src/experimental/condense_attributes_processor/mod.rs
drewrelmas: Left a TODO marker on this as well, it is the number of attributes 'condensed', it should be used for proper instrumentation.

PR #1696, opened 2025-12-26
[otap-dataflow] Add validation process for verifying otlp requests after encoding and decoding

c1ly:
Add the validation process based on [https://github.com/open-telemetry/otel-arrow/blob/main/docs/validation_process.md](https://github.com/open-telemetry/otel-arrow/blob/main/docs/validation_process.md) to verify that otlp requests are equivalent after encoding and decoding.


jmacd:
I'm glad to see the use of Weaver to generate synthetic data!

lquerel (approved):
LGTM. It's a good first step to buy us confidence on the encoding/decoding process.

PR #1718, opened 2026-01-05
Improve performance producing stream per batch `BatchArrowRecord`s

### LINKED ISSUE #1669: Add method to `otap_df_pdata::Producer` to close steams
It's possible that users might wish to serialize OTAP batches and distribute them in a way where deserialization cannot always be guaranteed to be handled by the same reader. For example, pushing messages onto a remote queue where several readers are opportunistically pulling the messages. 

This is a problem because each payload in the OTAP stream uses an Arrow IPC stream, and readers that don't see the first batch are missing the IPC header, which contains the schemas, dictionaries, etc.

For these cases, it would be nice if we offered a mode where the `Producer` could produce single batch streams for each `BatchArrowRecord` it produces.

The workaround today is to just create a new `Producer` for each OTAP batch, but some of the state on it's `StreamWriter`s are expensive to instantiate - notably the zstd context.

My proposal is that we should expose a method on the `Producer` that can be used to reset its internal `StreamWriter`s between batches that users wishing to produce these single-record streams could use.

When this is implemented, we should write a test using multiple consumers to ensure the solution solves the stated problem.

albertlockett:
closes #1669 

It's conceivable that there would be a need for an exporter that will produce serialized OTAP batches that may be consumed opportunistically by downstream consumers which do not coordinate to ensure each consumer is reading from a single Arrow IPC stream.

Example:
```
    exporter ───────> queue ─┰───> consumer1 (pulls next BAR)
(produces BARs)              └───> consumer2 (pulls next BAR)
```

Within each `BatchArrowRecord` (BAR), `otap_df_pdata::Producer::serialize_batch` will produce a single Arrow IPC stream for each payload type. This means that if `consumer1` receives the first BAR, and `consumer2` receives the 2nd BAR, the second BAR will only data messages in the Arrow IPC stream, but the stream header (which contains the schema) will be missing. This will cause an error when consuming the BAR.

One solution is to create a new Arrow IPC stream for every batch. Currently, doing this would mean creating a new `Producer` for every batch. There's some overhead in doing this because the `Producer`s contain a `HashMap` of `arrow::ipc::StreamWriters`, which contain a ztd context, all of which needs to be reinitialized.

This PR adds a method called `reset_streams` to the `Producer`. Calling this will mean the next serialized batch begins a new Arrow IPC stream for each payload type. This can be called after each batch to create a new Arrow IPC stream per batch, with less overhead than creating new `Producer`s.

---

This PR also moves a few test utility functions for creating `OtelArrowRecord`s from proto structs from the query-engine crate into the pdata crate.

### DIFF
--- a/rust/otap-dataflow/deny.toml
+++ b/rust/otap-dataflow/deny.toml
@@ -78,6 +78,7 @@ ignore = [
     { id="RUSTSEC-2021-0127", reason="`serde_cbor` is archived, but current implementation meets needs" },
     { id="RUSTSEC-2025-0021", reason="security vulnurability from `gix-features` crate" }, # remove after next weaver release
     { id="RUSTSEC-2023-0018", reason="security vulnurability from `remove_dir_all` crate" }, # remove after next weaver release
+    { id="RUSTSEC-2025-0140", reason="security vulnerability from `gix-date` crate" }, # remove after next weaver release
     { id="RUSTSEC-2023-0028", reason = "unmaintained `buf_redux` crate that is a dependency of weaver_semconv"}, # is a warning level in weaver deny.toml
     { id="RUSTSEC-2023-0050", reason = "unmaintained `multipart` crate that is a dependency of weaver_semconv and weaver_forge"}, # is a warning level in weaver deny.toml
     { id="RUSTSEC-2023-0081", reason = "unmaintained `safemem` crate that is a dependency of weaver_semconv"}, # is a warning level in weaver deny.toml


albertlockett:
Benchmark result:

| new producer (µs) | reset streams (µs) | IPC compression | improvement |
|---|---|---|---|
|46.227|40.968|None| 11.3% |
|216.89|193.61| zstd | 10.7% |

Benchmark code:
```rs
use arrow::ipc::CompressionType;
use otap_df_pdata::encode::producer::ProducerOptions;
use otap_df_pdata::proto::opentelemetry::common::v1::{AnyValue, KeyValue};
use otap_df_pdata::proto::opentelemetry::logs::v1::LogRecord;

use criterion::{BatchSize, BenchmarkId, Criterion, criterion_group, criterion_main};
use otap_df_pdata::testing::round_trip::to_otap_logs;
use otap_df_pdata::{OtapArrowRecords, Producer};

fn do_benches(c: &mut Criterion) {
    let mut group = c.benchmark_group("1669_logs_bench");

    fn produce_otap_batch() -> OtapArrowRecords {
        let mut log_records = vec![];
        for i in 0..2000 {
            log_records.push(LogRecord::build()
                .time_unix_nano(1767571308000u64)
                .observed_time_unix_nano(1767571308000u64)
                .event_name(&format!("event_{}", i))
                .severity_number(10)
                .severity_text("WARN")
                .body(AnyValue::new_string(&format!("body_{}", i)))
                .attributes(vec![KeyValue::new(
                    "attribute_key",
                    AnyValue::new_string(&format!("attribute_value_{}", i)),
                )])
                .finish());
        }
        let mut batch = to_otap_logs(log_records);
        batch.encode_transport_optimized().unwrap();
        batch
    }

    // let ipc_compression = Some(CompressionType::ZSTD);
    let ipc_compression = None;

    // Add benchmarks here, for example:
    _ = group.bench_function(BenchmarkId::new("new_producer", 1), |b| {
        b.iter_batched(
            || { produce_otap_batch() },
            |batch| { 
                let mut producer = Producer::new_with_options(ProducerOptions {
                    ipc_compression
                });
                let bar = producer.produce_bar(&mut batch.clone()).expect("working");
                _ = std::hint::black_box(bar);
             },
            BatchSize::SmallInput,
        )
    });

    let mut producer = Producer::new_with_options(ProducerOptions { 
        ipc_compression
    });

    _ = group.bench_function(BenchmarkId::new("reset_producer", 1), |b| {
        b.iter_batched(
            || {
                produce_otap_batch()
            },
            |mut batch| {
                let bar = producer.produce_bar(&mut batch).expect("working");
                producer.reset_streams().expect("resetting streams works");
                _ = std::hint::black_box(bar);
            },
            BatchSize::SmallInput,
        )
    });

    group.finish();
}

#[allow(missing_docs)]
mod benches {
    use super::*;
    criterion_group!(benches, do_benches);
}

criterion_main!(benches::benches);
```

Benchmark machine: Apple M4 Pro

rust/otap-dataflow/crates/pdata/src/encode/producer.rs
```diff
@@ -9,48 +9,139 @@ use std::io::Cursor;
...
+    /// Reset the stream to the initial state. After calling this, the byte sequence returned by
+    /// `serialize_batch` will start a new arrow IPC stream.
+    fn reset_stream(&mut self, ipc_write_options: &IpcWriteOptions) -> Result<()> {
+        // reset buffer
+        self.buffer.set_position(0);
+
+        // reset dictionary tracker ...
+        // TODO we should eventually add a method to the dictionary tracker to reset it so we can
+        // reuse the existing instance instead of creating a new one and avoid extra allocations
+        self.dictionary_tracker = DictionaryTracker::new(false);
```
lquerel, line 139: That would be a good source of optimization. To prevent the dictionaries from growing indefinitely, we could introduce a full reset every X iterations that make changes, or something along those lines. I think we should create a GitHub issue to track this.

lquerel (approved):
LGTM.

Another potential future source of optimization, which is not directly related to the reset but is linked to zstd compression: for small batches, it would be beneficial to train a zstd dictionary to improve the compression of small messages.

PR #1722, opened 2026-01-05
Add `if`/`else if`/`else` expression to KQL parser

albertlockett:
part of #1667 

Adds parsing support for an `if`/`else if`/`else` expression that gets parsed into the `ConditionalDataExpression` that was added in https://github.com/open-telemetry/otel-arrow/pull/1684

The syntax looks like this.
```kql
logs |
    extend attributes["important"] = "very" | extend attributes["triggers_alarm"] = "true"
} else if (severity_text == "WARN") {
    extend attributes["important"] = "somewhat"
} else if (severity_text == "INFO") {
    extend attributes["important"] = "rarely"
} else {
    extend attributes["important"] = "no"
}
```

`else if` and `else` are optional, so the following expressions are also supported:

```kql
logs | 
    extend attributes["is_error"] = true
} else {
    extend attributes["is_not_error"] = true
}
```
```kql
logs | 
    extend attributes["triggers_alarm"] = "true"
}
```

If we're OK with the syntax and how this is implemented, I'll go back and clean up the tests that we're added in #1684 in a followup PR. Those tests are manually building the expression AST, but it might be easier to maintain them if we parsed the query.

### DIFF
--- a/rust/experimental/query_engine/kql-parser/src/if_else_expression.rs
+++ b/rust/experimental/query_engine/kql-parser/src/if_else_expression.rs
@@ -2,8 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use data_engine_expressions::{
-    ConditionalDataExpression, ConditionalDataExpressionBranch, DataExpression, LogicalExpression,
-    QueryLocation, TransformExpression,
+    ConditionalDataExpression, ConditionalDataExpressionBranch, DataExpression, Expression,
+    LogicalExpression, QueryLocation,
 };
 use data_engine_parser_abstractions::{ParserError, ParserScope, to_query_location};
 use pest::iterators::Pair;
@@ -21,9 +21,9 @@ pub fn parse_if_else_expression(
     let mut conditional_expr = ConditionalDataExpression::new(query_location);
 
     // keep track of the location of the current branch
-    let mut branch_loc_start = 0;
-    let mut branch_loc_line = 0;
-    let mut branch_log_col = 0;
+    let mut branch_location_start = 0;
+    let mut branch_location_line = 0;
+    let mut branch_location_col = 0;
 
     let mut next_condition: Option<LogicalExpression> = None;
     let mut next_branch: Vec<DataExpression> = Vec::new();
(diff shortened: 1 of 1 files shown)


drewrelmas:
@albertlockett One FYI note about the usage of the `Conditional` expressions for assignments, single statement executions could use already supported `iif` and `case`. Ignoring the double assignment, a sample statement:

```kql
    extend attributes["important"] = "very" // ignoring | extend attributes["triggers_alarm"] = "true"
} else if (severity_text == "WARN") {
    extend attributes["important"] = "somewhat"
} else if (severity_text == "INFO") {
    extend attributes["important"] = "rarely"
} else {
    extend attributes["important"] = "no"
}
```

could also be written as:

```kql
logs
| extend attributes["important"] = case(severity_text == "ERROR", "very", severity_text == "WARN", "somewhat", severity_text == "INFO", "rarely", "no"
```

That said, I think the actual important part of this PR here is actually executing multiple statements in a branch (to get the `triggers_alarm` attribute) and I recall your intention potentially being routing between nodes - that part cannot be represented easily in pure KQL.

@CodeBlanch what other _significant_ departures from pure KQL do we currently have? Off the top of my head, I recall our map accessing is a bit different.

I wonder if we can have a superset of the grammar in a separate `.pest` file that inherits the base for these significant divergences from KQL with user config specifying what 'flavor' of grammar to support? See [load multiple grammars](https://pest.rs/book/grammars/grammars.html#load-multiple-grammars) from the Pest docs. I think I do see the value in driving this direction in supported expression tree, I'm still musing on how mixed it should be in the parser... 🤔

albertlockett:
Hey @drewrelmas thanks for taking a look!

> @albertlockett One FYI note about the usage of the `Conditional` expressions for assignments, single statement executions could use already supported `iif` and `case`. Ignoring the double assignment, a sample statement:
> 
> ```kusto
> if (severity_text == "ERROR") { 
>     extend attributes["important"] = "very" // ignoring | extend attributes["triggers_alarm"] = "true"
> } else if (severity_text == "WARN") {
>     extend attributes["important"] = "somewhat"
> } else if (severity_text == "INFO") {
>     extend attributes["important"] = "rarely"
> } else {
>     extend attributes["important"] = "no"
> }
> ```
> 
> could also be written as:
> 
> ```kusto
> logs
> | extend attributes["important"] = case(severity_text == "ERROR", "very", severity_text == "WARN", "somewhat", severity_text == "INFO", "rarely", "no"
> ```
> 
> That said, I think the actual important part of this PR here is actually executing multiple statements in a branch (to get the `triggers_alarm` attribute) and I recall your intention potentially being routing between nodes - that part cannot be represented easily in pure KQL.
> 

Yeah exactly. We'd like to add some additional conditional capabilities beyond just assignment, and routing would be part of that. I can imagine us wanting to do something like:

```kusto
logs |
    extend attributes["triggers_alarm"] = "true" | route_to "high_severity_out_port"
} else {
    route_to "low_severity_out_port"
}
```

> @CodeBlanch what other _significant_ departures from pure KQL do we currently have? Off the top of my head, I recall our map accessing is a bit different.
> 
> I wonder if we can have a superset of the grammar in a separate `.pest` file that inherits the base for these significant divergences from KQL with user config specifying what 'flavor' of grammar to support? See [load multiple grammars](https://pest.rs/book/grammars/grammars.html#load-multiple-grammars) from the Pest docs. I think I do see the value in driving this direction in supported expression tree, I'm still musing on how mixed it should be in the parser... 🤔

Thanks for sharing that! I didn't know about the multiple grammars, but I think a that's exactly what we'll need: a base grammar, and parsers that inherit the base, plus various extensions.

I created an issue to track this work https://github.com/open-telemetry/otel-arrow/issues/1728 and I'll spend some time today investigating how to make this work. In terms of getting this PR merged, I'll leave it up to you whether you're comfortable merging as is, or whether it'd be better to implement a solution for #1728 first and rebase these changes on top of that.

I foresee a few other divergences from standard KQL we'd like to support in the columnar query engine. These include operators `route_to`, `fork` and possibly `drop`. From the parser's perspective, these would all be variants of `tabular_expression`, like `if_else_expression`.

albertlockett:
@drewrelmas wrt to:

> I wonder if we can have a superset of the grammar in a separate .pest file that inherits the base for these significant divergences from KQL with user config specifying what 'flavor' of grammar to support? See [load multiple grammars](https://pest.rs/book/grammars/grammars.html#load-multiple-grammars) from the Pest docs. I think I do see the value in driving this direction in supported expression tree, I'm still musing on how mixed it should be in the parser... 🤔

I added a solution for this in https://github.com/open-telemetry/otel-arrow/pull/1734 Can you (and maybe @CodeBlanch) take a look and let me know what you think?

The changes there will obviously conflict with the changes in this PR, but I'm open to having them merged in either order.

albertlockett:
Closing this for now. Will open a different PR moving this code to OPL Parser that was added in #1752

PR #1727, opened 2026-01-06
Internal Telemetry Guidelines

lquerel:
This PR defines a set of guidelines for our internal telemetry and for describing how we can establish a telemetry by design process.

Once this PR is merged, I will follow up with a series of PRs to align the existing instrumentation with these recommendations.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/transform_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/transform_processor.rs
@@ -345,17 +345,13 @@ mod test {
                 // Allow the collector to pull from the channel
                 tokio::time::sleep(std::time::Duration::from_millis(50)).await;
 
-                let mut msgs_consumed = 0;
-                let mut msgs_forwarded = 0;
                 let mut msgs_transformed = 0;
                 let mut msgs_transform_failed = 0;
                 registry.visit_current_metrics(|desc, _attrs, iter| {
-                    if desc.name == "transform.processor.metrics" {
+                    if desc.name == "transform.processor" {
                         for (field, v) in iter {
                             let val = v.to_u64_lossy();
                             match field.name {
-                                "msgs.consumed" => msgs_consumed += val,
-                                "msgs.forwarded" => msgs_forwarded += val,
                                 "msgs.transformed" => msgs_transformed += val,
                                 "msgs.transform.failed" => msgs_transform_failed += val,
                                 _ => {}
@@ -364,8 +360,6 @@ mod test {
                     }
                 });
 
-                assert_eq!(msgs_consumed, 1);
-                assert_eq!(msgs_forwarded, 1);
                 assert_eq!(msgs_transformed, 1);
                 assert_eq!(msgs_transform_failed, 0)
             });


rust/otap-dataflow/crates/otap/src/attributes_processor/metrics.rs
```diff
@@ -10,14 +10,6 @@ use otap_df_telemetry_macros::metric_set;
...
 #[derive(Debug, Default, Clone)]
 pub struct AttributesProcessorMetrics {
-    /// PData messages consumed by this processor.
-    #[metric(unit = "{msg}")]
-    pub msgs_consumed: Counter<u64>,
-
-    /// PData messages forwarded by this processor.
-    #[metric(unit = "{msg}")]
-    pub msgs_forwarded: Counter<u64>,
-
```
lquerel, line 20: Removed because redundant with the channel metrics.

rust/otap-dataflow/crates/otap/src/transform_processor/metrics.rs
```diff
@@ -7,17 +7,9 @@ use otap_df_telemetry::instrument::Counter;
...
 #[derive(Debug, Default, Clone)]
 pub struct Metrics {
-    /// PData messages consumed by this processor.
-    #[metric(unit = "{msg}")]
-    pub msgs_consumed: Counter<u64>,
-
-    /// PData messages forwarded by this processor.
-    #[metric(unit = "{msg}")]
-    pub msgs_forwarded: Counter<u64>,
-
```
lquerel, line 20: Removed because redundant with the channel metrics.

rust/otap-dataflow/docs/telemetry/README.md
```diff
@@ -0,0 +1,275 @@
...
+
+This project plans to integrate with **Weaver** to:
+
+* validate emitted telemetry against the versioned semantic convention registry,
+* perform registry compliance checks in CI,
+* execute live checks during tests to ensure that expected signals are actually
+  produced,
+* generate authoritative documentation in Markdown or HTML from the registry.
+
+An administrative endpoint exposes the live, resolved schema at runtime to
```
andborja, line 191: This should be configurable, not always on

lquerel: This is partially covered in `security-privacy-guide.md` (link below the sentence you commented), but I've just added a more explicit point on this aspect (see https://github.com/open-telemetry/otel-arrow/pull/1727/commits/4fe4bb3a771c6e01933f350cb3e845b3cf220813).

rust/otap-dataflow/docs/telemetry/attributes-guide.md
```diff
@@ -0,0 +1,158 @@
...
+- [stability-compatibility-guide.md](stability-compatibility-guide.md) for
+  evolution rules
+- [security-privacy-guide.md](security-privacy-guide.md) for sensitive-data
+  constraints
+
+## Attribute categories
+
+Attributes fall into three categories.
+
+### 1) Resource attributes
```
andborja, line 23: I think the resource attributes should be configured at startup time, by the user (in the configuration). This is how it is done in the go collector, and it allows you to define any resource level attribute that you consider for your application.
Depending on where and how the engine will run, a resource attribute might make sense or not. For example, container id only makes sense if the engine is running in isolation in a container. If it is part of a side car, I might want to reference something different.
If it runs as a component/part of an external application, that might be not even a service (think on an agent), I would use other resource attributes. Things like service might not make sense.

andborja:
This is great!

rust/otap-dataflow/docs/telemetry/attributes-guide.md
```diff
@@ -0,0 +1,158 @@
...
+- Do not redefine upstream attributes with different meaning.
+
+### Project-defined namespace
+
+Project-defined entity attributes MUST be namespaced to avoid collisions with
+upstream conventions.
+
+Policy:
+
+- Use `otelcol.*` for project-defined attributes.
```
jmacd, line 66:
```suggestion
- Use OpenTelemetry-specified semantic conventions where applicable. For pipeline-related metrics with [existing names and definitions in the OpenTelemetry Collector](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md), specifically, use the `otelcol.*` metric namespace.
```

I'm open to using `otapcol.*` as well!

rust/otap-dataflow/docs/telemetry/README.md
```diff
@@ -0,0 +1,275 @@
...
+Multivariate metrics are treated as a fundamental modeling capability rather
+than a post-processing optimization.
+
+Note: OTLP and OTAP protocols do not yet have first-class support for
+multivariate metrics. The SDK and exporters handle the necessary translation and
+encoding. We plan to contribute multivariate support to OpenTelemetry protocols
+in the future. In the meantime, this project serves as a proving ground for the
+concept.
+
+### 7. Tooling-driven validation and documentation with Weaver
```
jmacd, line 178: 🚀

rust/otap-dataflow/docs/telemetry/README.md
```diff
@@ -0,0 +1,272 @@
...
+### 5. First-class support for multivariate metrics
+
+The internal telemetry model and SDK natively support **multivariate metric
+sets**.
+
+This enables:
+
+* efficient sharing of attribute tuples,
+* coherent modeling of related measurements,
+* reduced duplication and cardinality explosion compared to naive univariate
```
cijothomas: nit: not sure how it avoids cardinality explosions?

rust/otap-dataflow/docs/telemetry/README.md
```diff
@@ -0,0 +1,272 @@
...
+Security and deployment guidance for this endpoint is in the
+[Security and Privacy Guide](security-privacy-guide.md).
+
+Registry compliance checks and live checks are not yet enforced in CI. See
+[Implementation Gaps](implementation-gaps.md).
+
+### 7. Automated client SDK generation (longer term)
+
+In the longer term, the custom semantic convention registry will be used to
+generate **type-safe Rust client SDKs** via Weaver.
```
cijothomas: nit: we are not generated while client SDKs right? just the API used to instrument, which gets plugged into an existing SDK.
```suggestion
generate **type-safe Rust instrumentation APIs** via Weaver.
```

rust/otap-dataflow/docs/telemetry/README.md
```diff
@@ -0,0 +1,272 @@
...
+Refer to [Stability and Compatibility Guide](stability-compatibility-guide.md).
+
+For items that are documented but not yet implemented or enforced, see
+[Implementation Gaps](implementation-gaps.md).
+
+## Runtime safety and failure behavior
+
+Telemetry MUST be non-fatal and bounded:
+
+- Export failures MUST NOT break the dataflow engine.
```
cijothomas: Export failure not breaking engine -- this is not really related to telemetry, but more broader principle..

rust/otap-dataflow/docs/telemetry/README.md
```diff
@@ -0,0 +1,272 @@
...
+- Telemetry pipelines MUST use bounded buffers.
+- Under pressure, the default behavior SHOULD be to drop telemetry rather than
+  block critical work.
+- Drops SHOULD be observable via counters (by drop reason) and optionally debug
```
cijothomas: :100: love this part!

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
```diff
@@ -0,0 +1,130 @@
...
+
+If in doubt, do not emit it.
+
+## Allowed data with constraints
+
+The following are generally allowed if they remain bounded and non-sensitive:
+
+- stable internal IDs for entities (pipelines, nodes, channels), as defined in
+  the entity model
+- categorical outcomes and reasons (closed enums), for example drop reasons
```
cijothomas, line 43: closed enums - would likely mean we need to anticipate all the possible failures upfront. Possible, but might be hard. Usually I see "Other" as a catch-all for any un-anticipated outcomes!

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
```diff
@@ -0,0 +1,130 @@
...
+## Principles
+
+- Telemetry MUST NOT include secrets, credentials, or sensitive personal data.
+- Telemetry SHOULD include only what is required for operations and debugging.
+- Telemetry MUST be safe under failure and safe under adversarial inputs.
+- Access to telemetry SHOULD follow least privilege.
+
+## Data that must never appear in telemetry
+
+The following MUST NOT be recorded in metrics, events, traces, or attributes:
```
cijothomas, line 20: do we expect to introduce a special debug mode/similar, where these rules can be bent? of course based on explicit opt-in...?

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
```diff
@@ -0,0 +1,130 @@
...
+- URL path -> route template
+- SQL -> normalized fingerprint
+- IP address -> prefix or bucket
+- error message -> error class or error type
+
+Do not emit raw content that can include secrets or user identifiers.
+
+## Exceptions and stack traces
+
+Exceptions often include sensitive data. Rules:
```
cijothomas, line 61: nit: It may be good to call out that exception.message, and stack-trace are likely sensitive, but exception-type is not.

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
```diff
@@ -0,0 +1,130 @@
...
+- Use structured exception attributes (`exception.type`, `exception.message`)
+  when needed.
+- `exception.stacktrace` SHOULD be gated behind:
+    - debug severity, or
+    - an explicit configuration flag
+- Stack traces MUST NOT be emitted on hot paths by default.
+
+## Events, body size, and spill risk
+
+Events are often exported as logs.
```
cijothomas:
```suggestion
Events are exported as logs.
```
always right?

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
```diff
@@ -0,0 +1,130 @@
...
+- Keep event bodies small and bounded.
+- When details are required, emit a stable error type in attributes and keep the
+  long detail behind debug-level gating.
+
+## Trace correlation
+
+When exporting events as logs and trace context exists:
+
+- include trace correlation (trace id and span id) so operators can pivot
+- do not copy trace ids into custom attributes unless required by tooling
```
cijothomas, line 89: nit: since Otel Logs Datamodel has well known place for storing trace correlation, I don't think this ever occurs.

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
```diff
@@ -0,0 +1,130 @@
...
+
+Trace ids are not secrets but they can be used to join information across
```
cijothomas:
```suggestion
Trace and Span ids are not secrets but they can be used to join information across
```

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
```diff
@@ -0,0 +1,130 @@
...
+- It MUST NOT expose secrets or raw configuration values.
+- Treat the endpoint as sensitive because it can reveal topology and
+  identifiers.
+
+## Metrics and diagnostic endpoints (/metrics, /status)
+
+If the system exposes metrics scrape endpoints (for example Prometheus-style) or
+diagnostic endpoints:
+
+- They SHOULD be protected by authentication and authorization, or limited to
```
cijothomas, line 113: isn't this upto the folks operating the arrow-df-engine to enforce these?

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
```diff
@@ -0,0 +1,130 @@
...
+  trusted network boundaries.
+- They MUST NOT expose secrets or raw configuration values.
+- They SHOULD be designed to avoid unbounded responses (for example unbounded
+  label sets or dumping full topology on every request).
+- If an endpoint includes topology identifiers (pipelines, nodes, channels),
+  treat it as sensitive.
+
+## Access controls and retention
+
+- Access to telemetry backends SHOULD be restricted to operators and trusted
```
cijothomas: not so clear what are we referring to here...

rust/otap-dataflow/docs/telemetry/semantic-conventions-guide.md
```diff
@@ -0,0 +1,272 @@
...
+follow in this project** when defining **metric names**, **units**,
+**attributes**, and **event metadata**.
+
+Its goal is to **ease contributor work** by providing a clear, concise, and
+opinionated reference tailored to this project.
+
+**Important**
+All rules and conventions described here are **derived directly from the
+official OpenTelemetry Semantic Conventions**.
+The OpenTelemetry specifications are the **ultimate source of truth**, and this
```
cijothomas: does OTel spec has any rule here? its only the semantic conventions that holds the truth.

rust/otap-dataflow/docs/telemetry/semantic-conventions-guide.md
```diff
@@ -0,0 +1,272 @@
...
+
+```
+http.request.start
+otelcol.pipeline.config.apply
+connection.close
+```
+
+---
+
+### Event attributes
```
cijothomas, line 209: mention something about - removing/renaming attributes for a given Eventname is considered breaking change.

rust/otap-dataflow/docs/telemetry/semantic-conventions-guide.md
```diff
@@ -0,0 +1,272 @@
...
+Note:
+In OTLP, events are represented as LogRecords with the `event_name` field set.
+The `event.name` attribute is deprecated and should not be used for new
+telemetry.
+
+Examples:
+
+```
+http.request.start
+otelcol.pipeline.config.apply
```
cijothomas, line 203: I am still unsure of whether EventName should be fully qualifying or InstrumentationScope+EventName is what is fully qualifying...

(Josh's PR https://github.com/open-telemetry/otel-arrow/pull/1735#discussion_r2670304468 uses "target::event-name". (Target is same as scope), so that becomes fully qualifying.

Not a blocker, something we can sort out while adding events.

rust/otap-dataflow/docs/telemetry/stability-compatibility-guide.md
```diff
@@ -0,0 +1,209 @@
...
+
+Breaking changes include:
+
+- renaming metrics, metric sets, event names, span names, or attribute keys
+- changing units
+- changing instrument semantic meaning (counter vs gauge semantics,
+  monotonicity, temporality assumptions)
+- removing a metric, event, span, or attribute
+- changing the meaning of an attribute value
+- widening attribute cardinality such that aggregations change meaning or cost
```
cijothomas, line 82: anything to mention about Event's Body?

rust/otap-dataflow/docs/telemetry/stability-compatibility-guide.md
```diff
@@ -0,0 +1,209 @@
...
+- **dual emission**
+    - emit old and new signals together temporarily
+- **alias and translate at export**
+    - if exporter can map old key to new key without losing meaning
+- **side-by-side dashboards**
+    - validate new telemetry before switching alerts
+
+Avoid:
+
+- silent renames
```
cijothomas, line 203: hopefully, can use weave live-check in CI to police this!

rust/otap-dataflow/docs/telemetry/attributes-guide.md
```diff
@@ -0,0 +1,158 @@
...
+### 1) Resource attributes
+
+Describe the producing service and runtime environment.
+
+- MUST be attached at the resource level
+- MUST NOT be duplicated on every signal
+- SHOULD reuse upstream semantic conventions (`service.*`, `host.*`,
+  `process.*`, `container.*`)
+
+### 2) Entity attributes
```
cijothomas, line 32: is this anyway related to OTel Entities?

rust/otap-dataflow/docs/telemetry/attributes-guide.md
```diff
@@ -0,0 +1,158 @@
...
+
+Identify stable in-process entities (pipelines, nodes, channels, runtime
+threads).
+
+- MUST be attached/translated as scope attributes in OTLP exports
+- MUST NOT be duplicated on every signal
```
cijothomas, line 38: the line above says it MUST be added as scope attributes - so it'd normally get duplicated for each signal.. Not sure if I misunderstood the wording.

rust/otap-dataflow/docs/telemetry/attributes-guide.md
```diff
@@ -0,0 +1,158 @@
...
+
+- If I aggregate across this attribute, does the result still make sense?
+- Is the value space bounded and known at entity creation time?
+
+If not, the attribute is mis-modeled for core system metrics.
+
+### Prohibited by default in core system metrics
+
+The following are prohibited as metric attributes unless explicitly approved and
+normalized:
```
cijothomas, line 106: if explicitly opting in, do we need to normalize? For example, a company can opt-in to add user-id as a metric attribute, to provide/measures user level SLOs/failure rates etc.

rust/otap-dataflow/docs/telemetry/events-guide.md
```diff
@@ -0,0 +1,160 @@
...
+`event.name` as an attribute.
+
+Use events to record:
+
+- Controller/Pipeline actions (config reload, shutdown, ack, timer ticks).
+- State transitions (batch flush, backpressure, queue full).
+- Exceptional outcomes (errors, retries, drops).
+
+If the signal is high-volume or needs aggregation, prefer metrics. If the
+event is part of a dataflow trace, record it as a span event.
```
cijothomas: lets avoid span events completely, and just use events - they anyway get correlated to parent active span.
(OTel is actively moving away from SpanEvents..)

rust/otap-dataflow/docs/telemetry/events-guide.md
```diff
@@ -0,0 +1,160 @@
...
+
+Exception rule (traces):
+
+- If you are recording an actual exception on a span, the span event name MUST
```
cijothomas: lets report exception as Events, not SpanEvents.

rust/otap-dataflow/docs/telemetry/events-guide.md
```diff
@@ -0,0 +1,160 @@
...
+## Severity and placement
+
+When events are exported as logs, set an appropriate severity. When they are
+attached to traces, use span events with the same name and attributes.
+
+Regarding severity, some events may be logged at different levels depending on
+their severity or impact. For example, a `node.shutdown` event may be logged at
+INFO level during a graceful shutdown, but at ERROR level if the shutdown is due
+to a critical failure. When exporting events as logs, choose the log level that
+best reflects the significance of the event.
```
cijothomas, line 102: interesting. That would mean the same Event can have different severities. Could we instead pick different event names ?
eg: node.shutdown for normal shutdown.
node.shutdown.failure for critical failure shutdown.

rust/otap-dataflow/docs/telemetry/attributes-guide.md
lquerel: I had this approach in mind for **groups of metrics** (our `metric_set`s) that all refer to the same entity and can therefore be placed in the same scope, since they are a group of signals of the same type.
In a similar way, one can imagine several events emitted by the same entity and thus grouped in the same event scope corresponding to that entity. This case is probably less frequent than for metrics, but I think it remains an approach worth exploring.

rust/otap-dataflow/docs/telemetry/README.md
lquerel: I did get a bit carried away on the cardinality aspect indeed. That said, the redundancy of attributes is still greatly reduced, inversely reduced relative to the number of metrics in the set.

Regarding cardinality, I had in mind what the Collector does, for example with memory-related metrics using the `state` attribute (same for `mode`). I had the impression that `state` is a pattern often used to represent a category of multivariate metrics.

In any case, I’m going to remove the word "cardinality" from that sentence.

rust/otap-dataflow/docs/telemetry/README.md
lquerel: Today, in the short term, yes, we are focusing on the instrumentation side and reusing an existing SDK. In a second phase, what Joshua is preparing is to integrate our OTAP pipeline to report our internal telemetry. In absolute terms, this will allow us to report our telemetry via OTLP, OTAP, and so on, without going through an existing SDK. In the long term, I was thinking of reusing all this infrastructure to generate a new kind of schema-driven and type-safe SDKs.

rust/otap-dataflow/docs/telemetry/README.md
lquerel: By this sentence, I meant for example that our instrumentation API must not be blocked, in one way or another, by a defective or congested SDK export function. If you have a rewording in mind that you think would be more accurate, feel free to suggest it.

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
lquerel: We do not need to anticipate all possible values of an enum. To my knowledge, adding a new variant is not considered a breaking change, so in a schema-driven approach I think we can avoid the catch-all `other` variant 🙂

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
lquerel: I no longer remember where I mentioned debug-level gating for certain attributes, but yes, indeed, we should allow a more permissive behavior in that mode. I will definitely clarify this point.

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
lquerel: I will clarify this point.

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
lquerel: I had span events in mind. But if we choose not using them then I'm ok.

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
lquerel: I’ve seen cases in some applications where people put all kinds of things into attributes… :-)

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
lquerel: Yes, but I was saying that it does not hurt to restate it in this security document, and it also forces us not to forget that these endpoints need to be configurable and protectable.

rust/otap-dataflow/docs/telemetry/security-privacy-guide.md
lquerel: It may indeed be unnecessary to mention this point in this document. I will remove it.

rust/otap-dataflow/docs/telemetry/semantic-conventions-guide.md
lquerel: Yes, it’s not clear. By "OpenTelemetry specifications", I meant to refer to the OTel Semantic Conventions. This paragraph deserves to be simplified and improved. Thanks

rust/otap-dataflow/docs/telemetry/semantic-conventions-guide.md
lquerel: It's mentioned in the stability/compatibility doc but that doesn't hurt to mention this important point also here.

rust/otap-dataflow/docs/telemetry/semantic-conventions-guide.md
lquerel: Something to discuss indeed.

rust/otap-dataflow/docs/telemetry/stability-compatibility-guide.md
lquerel: Probably. Need to think about it. Any idea?

rust/otap-dataflow/docs/telemetry/stability-compatibility-guide.md
lquerel: This is definitely one of our objectives with introducing Weaver in our CI.

rust/otap-dataflow/docs/telemetry/attributes-guide.md
lquerel: Yes

rust/otap-dataflow/docs/telemetry/attributes-guide.md
lquerel: In the context of our internal df-engine telemetry, where would this user-id come from?

rust/otap-dataflow/docs/telemetry/events-guide.md
lquerel: Ah, I didn't know that span events were no longer being promoted. I thought the recent developments were more about aligning the semantic conventions between regular events and span events, so that the definition of an event could be used in either of these two contexts. But I may be completely wrong on that point.

rust/otap-dataflow/docs/telemetry/events-guide.md
lquerel: Something to discuss.

rust/otap-dataflow/docs/telemetry/events-guide.md
cijothomas: https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/4430-span-event-api-deprecation-plan.md

There are active PRs right now in spec/conventions to make the above happen!

rust/otap-dataflow/docs/telemetry/stability-compatibility-guide.md
cijothomas: One option is to *not* use Body at all, as recommended (newly) by [OTel semantic conventions](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/events.md#general-event-semantics)

Or state that the "shape" of body will not change.

Not blocker, we can come back to this as we work through instrumentation.

rust/otap-dataflow/docs/telemetry/README.md
jmacd: This sounds similar to https://opentelemetry.io/docs/specs/otel/error-handling/

rust/otap-dataflow/docs/telemetry/stability-compatibility-guide.md
jmacd: Tangentially. I like the use of Body for arbitrary structured data, provided we have a way for it to be self-descriptive meaning we have a way to observe the schema.

For OTel, we have no way to represent a structured message format, a deferred print statement with user-defined formatting, with recursively structured arguments. I notice that the Rust standard library has the same issue, which I wrote about in #1735 as the inability to destructure the std::fmt::Debug trait or the standard implementations.

when I write log::info!("this {a} and that {b}: {c}"), I intend for the body to be some kind of structure containing "this {a} and that {b}: {c}" and a map of OTLP AnyValue encodings for the three variables and whatever ("self-descriptive") type information we decide on.

(Ask me my opinion?! Ha. I think the reason the OTel dialog about Events is in disarray is that we need to solve the "printf problem" first. Instead, we're) told we shouldn't use any Body and shouldn't expect formatted messages. 🤣)

cijothomas: String formatting is support in tracing/otel-rust:

```rust
error!(name: "another-event", target: "my-system", attr1= 10, attr2=20, "foo is {} and {}", 42, true);
```
(After all kv-pairs, the last one is specially treated by tracing and Otel)

the above will give the following outcome in OTel Rust:
```txt
EventName: "another-event"
         Target (Scope): "my-system"
         Observed Timestamp: 2026-01-08 17:39:04.254108
         SeverityText: "ERROR"
         SeverityNumber: Error
         Body: String(Owned("foo is 42 and true"))
         Attributes:
                 ->  attr1: Int(10)
                 ->  attr2: Int(20)
```

`fmt` gives the following:

```txt
2026-01-08T17:39:04.254134Z ERROR main my-system: foo is 42 and true attr1=10 attr2=20
```

rust/otap-dataflow/docs/telemetry/events-guide.md
jmacd: Why should severity and event_name be 1:1? I want to hear the argument, at least. I also want us to have a concept that is "statement identity".

Outside of this project, I think it should be the developer's choice whether to use one or two statements, and if they choose one statement and vary the severity level, I don't see any problems. If they are monitoring these events, they'll query by statement identity and group by level all under one event name.

rust/otap-dataflow/docs/telemetry/semantic-conventions-guide.md
jmacd: @cijothomas 
I have at times thought of the Tokio target and name as like "scope" but I also like @lquerel's position that scope variables should be under the application's control. What we are talking about here are static properties of the call site. I want us to have the concept of a statement identifier to capture static properties of each instrumentation point (target/name/file/line/etc), these are the components that can stand in for an event name when the user doesn't provide one. I understand the sense that they are like scope attributes, but I think of them more like associated definitions.

rust/otap-dataflow/docs/telemetry/attributes-guide.md
cijothomas: one theoretical example:

the engine could add support for Baggage, and users sending telemetry via engine sends "userid=foo" in Baggage. Now there can be a processor that retrieves userid from Baggage content and attaches it to metrics/logs etc.

(I might be imagining too far ahead..?)

rust/otap-dataflow/crates/otap/src/attributes_processor/metrics.rs
cijothomas: suggestion: Its easier from review standpoint, to keep PRs more focused. Since this PR is adding telemetry guidelines doc, lets stick with that. Cleaning up metrics can be own PR.

rust/otap-dataflow/docs/telemetry/semantic-conventions-guide.md
jmacd: And they have static Level in the tokio tracing framework, which I agree with. @cijothomas to your other discussion point about event name, I would say that Callsite and Level are 1:1. It's OK for users to re-use event names but it should be impossible to duplicate a callsite so users will always be able to break down event_name by level into distinct statements.

cijothomas: > for an event name when the user doesn't provide one.

The macros we defined (otel_info! etc.) mandates event name. :)
`tracing` defaults to filename:linenumber for event-name by default, if omitted. and it defaults to module name for `target`.

rust/otap-dataflow/docs/telemetry/attributes-guide.md
lquerel: I agree this something to keep in mind. I will add a comment in the doc for future exploration.

lquerel:
Thanks @cijothomas @jmacd @andborja and @albertlockett for all your feedback. This commit contains all my edits https://github.com/open-telemetry/otel-arrow/pull/1727/commits/57e4f67dc9138d3012ea69d4bd04b4beb64417b9

I will merge this PR soon when all those linter issues will be fixed.

rust/otap-dataflow/docs/telemetry/attributes-guide.md
andborja: @lquerel bumping this one as I think it is important.

PR #1725, opened 2026-01-06
Add support for schema key aliases in query engine Parsers

drewrelmas:
Draft PR to open discussion - The current `otlp-bridge` for the `recordset` engine uses the OpenTelemetry [log data model spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md) for its initial schema keys (`Attributes`, `Timestamp`, `ObservedTimestamp`, `SeverityText`, etc).

However, many well-versed in the OpenTelemetry space may be more used to the snake case representation (`attributes`, `time_unix_nano`, `observed_time_unix_nano`, `severity_text`, etc) from the [proto](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/pdata/src/views/otlp/proto/logs.rs) representation.

Do we have any significant risks if we plan to support both? Inspired by `severity_text` reference in #1722, been on the back of my mind for a while.

This is still somewhat incomplete, could need more wiring for user-provided aliases in bridge, but for the moment just doing it for known OpenTelemetry fields.

PR #1730, opened 2026-01-06
[kql_processor] Experimental KQL recordset processor

CodeBlanch:
Relates to #1642

## Changes

* Adds a processor which executes KQL query using the RecordSet engine (OTLP-bytes form)

## Details

@drewrelmas took @jmacd's original work and added tests + config. I cleaned it up a bit and improved the bridge API to allow the processor to own the pipeline memory instead of storing it in a static. The static path is in place for callers needing to invoke things using FFI (from non-Rust platforms).

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
@@ -199,9 +199,18 @@ impl Processor<OtapPdata> for RecordsetKqlProcessor {
                                 || new_config.bridge_options != self.config.bridge_options
                             {
                                 let bridge_options =
-                                    Self::parse_bridge_options(&new_config.bridge_options)
-                                        .ok()
-                                        .flatten();
+                                    match Self::parse_bridge_options(&new_config.bridge_options) {
+                                        Err(e) => {
+                                            otap_df_telemetry::otel_warn!(
+                                                "Processor.ReconfigureError",
+                                                processor = "kql",
+                                                message = format!("{e}")
+                                            );
+                                            None
+                                        }
+                                        Ok(v) => v,
+                                    };
+
                                 match parse_kql_query_into_pipeline(
                                     &new_config.query,
                                     bridge_options,


rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
```diff
@@ -0,0 +1,1061 @@
...
+                                let bridge_options =
+                                    Self::parse_bridge_options(&new_config.bridge_options)
+                                        .ok()
+                                        .flatten();
+                                match parse_kql_query_into_pipeline(
+                                    &new_config.query,
+                                    bridge_options,
+                                ) {
+                                    Ok(pipeline) => {
+                                        otap_df_telemetry::otel_error!(
```
gouslu: why is this an error?

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/README.md
```diff
@@ -0,0 +1,83 @@
...
+### Filter logs by severity
+
+```yaml
+query: "source | where SeverityNumber >= 17"  # ERROR and above
+```
+
+### Add computed fields
+
+```yaml
+query: "source | extend hour = bin(Timestamp, 1h)"
```
drewrelmas, line 140: Can you please include some explanation of implicit `Attributes` handling in the Readme? It may help reduce confusion.

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
```diff
@@ -0,0 +1,1061 @@
...
+                        message,
+                        OtapPdata::new(ctx, OtapPayload::empty(SignalType::Logs)),
+                    ))
+                    .await?;
+                Err(e)
+            }
+        }
+    }
+
+    async fn process_logs(
```
gouslu: there's nothing being awaited in this method, any reason to still keep the async?

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
```diff
@@ -0,0 +1,1061 @@
...
+                &self.pipeline,
+                RecordSetEngineDiagnosticLevel::Warn,
+                &bytes,
+            )
+            .map_err(|e| self.map_bridge_error(e, signal))?;
+
+        Ok(OtlpProtoBytes::ExportLogsRequest(included_records.into()))
+    }
+
+    fn map_bridge_error(&mut self, error: BridgeError, signal: SignalType) -> Error {
```
gouslu: can be made static?

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/README.md
```diff
@@ -0,0 +1,83 @@
...
+- **Transformations**: Apply KQL `extend` and `project` operations to modify
+  data
+- **Aggregations**: Perform KQL `summarize` operations for data aggregation
+
+## Configuration
+
+```yaml
+processors:
+  kql:
+    query: "source | where SeverityText == 'ERROR' | extend processed_time = now()"
```
drewrelmas, line 33: The Readme should probably document the available predefined OpenTelemetry keys (SeverityText, Timestamp, etc). Reminder to myself, if we proceed with #1725, I should update this Readme as well.

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
```diff
@@ -0,0 +1,1061 @@
...
+                    "Processor.Failure",
+                    processor = "recordset_kql",
+                    input_items = input_items,
+                    message = message,
+                );
+
+                effect_handler
+                    .notify_nack(NackMsg::new(
+                        message,
+                        OtapPdata::new(ctx, OtapPayload::empty(SignalType::Logs)),
```
gouslu, line 150: should nack not send the data back if the previous component expects it?

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
```diff
@@ -0,0 +1,1061 @@
...
+                        if let Ok(new_config) =
+                            serde_json::from_value::<RecordsetKqlProcessorConfig>(config)
+                        {
+                            // Re-parse the pipeline (if changed)
+                            if new_config.query != self.config.query
+                                || new_config.bridge_options != self.config.bridge_options
+                            {
+                                let bridge_options =
+                                    Self::parse_bridge_options(&new_config.bridge_options)
+                                        .ok()
```
gouslu: should this at least emit a warning or something instead of fully discarding the error?

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
CodeBlanch: Updated

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
CodeBlanch: Updated

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
CodeBlanch: I switched it to `otel_info`. @drewrelmas @jmacd does "info" work? Wasn't sure if it should be "debug" for something like this.

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/README.md
CodeBlanch: Updated

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/README.md
CodeBlanch: Updated

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
drewrelmas: I think we'll do another pass over these components for instrumentation as #1727 is finalized. For now, fine with `info`. Agree that it shouldn't be `error`.

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
CodeBlanch: I updated to emit a warning if the config doesn't parse.

rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
drewrelmas: Based on my understanding of `notify_nack` and `notify_ack`, yes, we would normally want to return original payload along with the `nack`. However, payload is being consumed in the conversion to `OtlpProtoBytes`...

jmacd: The design of the Ack/Nack system allows for simple components that pass through the Context to do nothing and it will just work.

There is not much to worry about with dropping the input data, unless we want to use the current retry processor _before_ the KQL processor. As long as we retry after KQL processor, not a problem.

PR #1731, opened 2026-01-06
perf(azure-monitor-exporter): Azure monitor exporter log transformer optimizations

gouslu:
- Reuse buffer across log records
- Schema pre-parsing to avoid parsing for each transformation
- Use enum matching instead of string comparisons for log record fields
- Use Cow<str> for HashMap lookups and eq_ignore_ascii_case to avoid temporary strings
- Remove "disable_schema_mapping", not needed for customers

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/README.md
+++ b/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/README.md
@@ -61,7 +61,6 @@ nodes:
             "span_id": "SpanId"
             "attributes":
               "message": "ParsedMessage"
-          disable_schema_mapping: false
 
       # Authentication configuration (uses Azure SDK defaults)
       auth:


rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/transformer.rs
```diff
@@ -8,244 +8,272 @@ use otap_df_pdata::views::{
...
+    SeverityText,
+    Body,
+}
 
-    #[serde(flatten)]
-    record: serde_json::Map<String, Value>,
+impl LogRecordField {
+    /// Parse a field name string into a LogRecordField enum
+    fn from_str(s: &str) -> Option<Self> {
+        if s.eq_ignore_ascii_case("time_unix_nano") {
```
drewrelmas, line 45: Not a blocker, but depending on how #1725 turns out, maybe we want to support aliases for the OpenTelemetry [log data model spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md) naming convention. You're already partially flexible thanks to the case insensitivity.

drewrelmas (approved):
You mention perf improvements - do you have benchmarks of before and after? I wonder if we should add these experimental components into [benchmarks](https://github.com/open-telemetry/otel-arrow/tree/main/rust/otap-dataflow/benchmarks) like the rest of primary components.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/transformer.rs
```diff
@@ -8,244 +8,272 @@ use otap_df_pdata::views::{
...
+    source: LogRecordField,
+    /// The destination field name in the output JSON
+    dest: String,
+}
 
-    #[serde(flatten)]
-    scope: &'a serde_json::Map<String, Value>,
+/// Enum representing known log record fields for fast matching
+#[derive(Debug, Clone, Copy)]
+enum LogRecordField {
```
cijothomas, line 30: EventName missing

lalitb:
> Remove "disable_schema_mapping", not needed for customers

Nit - This is still there in readme and yaml.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/transformer.rs
gouslu: done

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/transformer.rs
gouslu: I didn't bench it particularly, but I saw an about 3%-5% throughput increase in my local tests!

gouslu:
addressed the comments, ready to merge.

PR #1734, opened 2026-01-07
Refactor KQL parser into common base and language extensions

### LINKED ISSUE #1728: KQL Parsing support different 'flavors' of language when parsing
In https://github.com/open-telemetry/otel-arrow/pull/1722 we added support for an "if/else" type statement to the KQL, and there are more custom extensions we'd like to add to the grammar.

However, there might be some contexts where we're parsing the query for execution on a query-engine that doesn't support the statement. For example, record-set engine does not support the ConditionalDataExpression that was added in #1684 

Ideally, we would have multiple flavors of the pest grammar, that can share the same base but extend it with additional syntax.

albertlockett:
closes #1728 

In https://github.com/open-telemetry/otel-arrow/pull/1722 we added an `if/else` expression to the KQL parser that may not be supported by all engines that execute some KQL query.

Seeing as this is a departure from standard KQL, and there may be other new expressions added in the future (such as `route_ro`), a need was identified to have a common base grammar with the ability to add/parse language extensions. This PR implements this capability.

The pest grammar is split into `base.pest` and `kql.pest`, and the KQL parser now uses both these grammar files using pest's "[load multiple grammars](https://pest.rs/book/grammars/grammars.html#load-multiple-grammars)" feature.

A second parser is added for a hypothetical, KQL-inpisred query-language that will support the `if/else` expression and other future extensions. This is added as a new crate under `rust/otal-dataflow/crates/opl`. It also uses `base.pest`, and another grammar file called `opl.pest` where future language extensions will be added.

**Making the parser functions generic:**

There are many parser utility functions in the `kql-parser` crate that ideally all KQL derived parses could use. Inside these functions, there are many checks for which variant of the `Rule` enum (derived by `pest_derive::Parser`) is being handled. For example:
https://github.com/open-telemetry/otel-arrow/blob/5d4801241809c9e08da610997e8f0685894a653a/rust/experimental/query_engine/kql-parser/src/scalar_expression.rs#L353-L361

One challenge in making these functions generic is that they need to operate on a concrete enum type, and `pest_derive::Parser` proc macro generates a different `Rule` enum for every parser.

The solution in this PR is to derive a base `pest_derive::Parser` (and hence, the associated `Rule` enum), for the `base.pest` grammar. This is done in `kql-parser/src/base_parser.rs`. The parser utility functions for each expression are then made generic over a type of pest `Rule` that can be converted into `base_parser::Rule`. A trait is provided called `TryAsBaseRule` that encapsulates the conversion logic, so in this PR we see many functions changed to take a generic `R` where `pest::iterators::Pair<'_, R>` implements `TryAsBaseRule`.

Implementing the `TryInto` for some derived `Rule` into `base_parser::Rule` would be somewhat tedious to do by hand, because we'd need to write a match with a branch for every variant of the enum, and update these conversion functions each time we add a new rule to base.pest. To avoid this, a procedural macro is created to generate this conversion code. The macro lives in `kql-parser/src/macros`, and parsers simply need to use this proc macro to make their rules compatible with the generic parser functions:
```rs
#[derive(Parser, BaseRuleCompatible)]
#[grammar = "base.pest"]
#[grammar = "kql.pest"]
pub(crate) struct KqlPestParser;
```

One additional challenge related to parser functions generic is that the `scalar_expression::parse_scalar_expression` uses a `PrattParser`, which takes the `Pair` as its argument. Pest doesn't provide any simple way convert a `Pair<'_, R>` to `Pair<'_, base_parser::Rule>`, which means that the generic `parse_scalar_expression` function also needs a way to generically create a `PrattParser` that accepts `Pair<'_, R>`.  

To handle that challenge, the `BaseRuleCompatible` procedural macro also creates a `PrattParser` for the derived `Rule`, and implements a trait for the `Rule` enum that can be used to access the `PrattParser`. The trait is called `ScalarExprRules`, and so this trait bound is also added to the generic parser function signatures.

### DIFF
--- a/rust/experimental/query_engine/parser-abstractions/src/parser_error.rs
+++ b/rust/experimental/query_engine/parser-abstractions/src/parser_error.rs
@@ -31,7 +31,7 @@ pub enum ParserError {
     },
 
     #[error("{0}")]
-    RuleConversionError(String)
+    RuleConversionError(String),
 }
 
 impl From<&ExpressionError> for ParserError {


albertlockett:
Converting to draft after feedback from Jan 7th SIG meeting - this needs some refactoring

albertlockett:
On the Jan 7th SIG call, we came to the conclusion that sharing the Pest grammar was untenable and it would be best if an OPL parser had its own Pest grammar.

There was thinking at the time that maybe we could still share the parser code, which translates the Pest rules into our expression AST. On the surface, this seems desirable because there's a lot of code in the kql-parser crate that would need to be duplicated otherwise.

However as I dig into this more, I'm realizing that without a shared grammar, sharing this parser code is probably not the best approach, and OPL should probably just implement its own parser. In the paragraphs below I'll explain why.

In the shared parser code model, I imagine we'll make our parser utilities accept derived `Rules` from either the KQL or OPL parser (as was implemented in this PR). Effectively, this would mean that we implement `TryInto<kql_parser::Rule> for opl_parser::Rule`, which simply converts one `Rule` to the other for the enum variant with the same name. Our parser functions are generic over `R: TryInto<kql_parser::Rule>`.

What's not captured anywhere in this scheme is the hierarchical relationship between the rules and how the parser handles them, and this could result in subtle bugs.

For example, consider how kql-parser parses null literals. The rule hierarchy looks like `scalar_expression` -> `scalar_unary_expression` -> `type_unary_expression` -> `null_literal`. Accordingly, when parsing we call `scalar_expression::parse_scalar_unary_expression` which matches the rule to `type_unary_expression` which then calls `scalar_primitive_expressions::parse_type_unary_expression`, which then handles each typed scalar variant:
https://github.com/open-telemetry/otel-arrow/blob/9b0aa84cfc1ef5df2f4434a3290bab3abbd7299c/rust/experimental/query_engine/kql-parser/src/scalar_expression.rs#L69-L75

https://github.com/open-telemetry/otel-arrow/blob/9b0aa84cfc1ef5df2f4434a3290bab3abbd7299c/rust/experimental/query_engine/kql-parser/src/scalar_primitive_expressions.rs#L13-L31

Let's say for some reason that KQL needs needs to reorganize its grammar, and `null_literal` becomes a child of `scalar_unary_expression`:
```diff
diff --git a/rust/experimental/query_engine/kql-parser/src/kql.pest b/rust/experimental/query_engine/kql-parser/src/kql.pest
index c2bc4ced..9d65cd1f 100644
--- a/rust/experimental/query_engine/kql-parser/src/kql.pest
+++ b/rust/experimental/query_engine/kql-parser/src/kql.pest
@@ -134,8 +134,7 @@ dynamic_map_expression = { "{" ~ (dynamic_map_item_expression ~ ("," ~ dynamic_m
 dynamic_inner_expression = _{ dynamic_array_expression|dynamic_map_expression|type_unary_expressions }
 dynamic_expression = { "dynamic" ~ "(" ~ dynamic_inner_expression ~ ")" }
 type_unary_expressions = {
-    null_literal
-    | real_expression
+    real_expression
     | datetime_expression
     | time_expression
     | regex_expression
@@ -237,7 +236,8 @@ backwards. For example if integer_literal is defined before time_expression "1h"
 would be parsed as integer_literal(1) and the remaining "h" would be fed into
 the next rule. */
 scalar_unary_expression = {
-    type_unary_expressions
+    null_literal
+    | type_unary_expressions
     | get_type_expression
     | conditional_unary_expressions
     | conversion_unary_expressions
diff --git a/rust/experimental/query_engine/kql-parser/src/scalar_expression.rs b/rust/experimental/query_engine/kql-parser/src/scalar_expression.rs
index 0f45a327..36d56325 100644
--- a/rust/experimental/query_engine/kql-parser/src/scalar_expression.rs
+++ b/rust/experimental/query_engine/kql-parser/src/scalar_expression.rs
@@ -296,6 +296,9 @@ pub(crate) fn parse_scalar_unary_expression(
     let rule = scalar_unary_expression_rule.into_inner().next().unwrap();
 
     Ok(match rule.as_rule() {
+        Rule::null_literal => {
+            ScalarExpression::Static(parse_standard_null_literal(rule))
+        }
         Rule::type_unary_expressions => {
             ScalarExpression::Static(parse_type_unary_expressions(rule)?)
         }
diff --git a/rust/experimental/query_engine/kql-parser/src/scalar_primitive_expressions.rs b/rust/experimental/query_engine/kql-parser/src/scalar_primitive_expressions.rs
index 58db3601..6a33990a 100644
--- a/rust/experimental/query_engine/kql-parser/src/scalar_primitive_expressions.rs
+++ b/rust/experimental/query_engine/kql-parser/src/scalar_primitive_expressions.rs
@@ -16,7 +16,6 @@ pub(crate) fn parse_type_unary_expressions(
     let rule = type_unary_expressions_rule.into_inner().next().unwrap();
 
     Ok(match rule.as_rule() {
-        Rule::null_literal => parse_standard_null_literal(rule),
         Rule::real_expression => parse_real_expression(rule)?,
         Rule::datetime_expression => parse_datetime_expression(rule)?,
         Rule::time_expression => parse_timespan_expression(rule)?,
```
This works, and all the kql-parser tests will pass. However, if OPL had been using the same organization of its grammar rules, the parsing using the shared code would fail for OPL unless it also made the same adjustment to its grammar (not only would it fail, it would panic at scalar_primitive_expressions.rs::28).

This brings me to my first point, which is that without a shared grammar either:
- a) the organization of the rules becomes an immutable contract between kql-parser and crates that share its parser code. We'd probably consider this is untenable as it places undue restrictions on kql-parser's ability to adapt its own grammar
- b) OPL parser needs to have its own set of parser tests to catch these issues, which means that all the test cases from kql-parser get duplicated anyway even if we share the parser code.

The second difficulty I see in sharing this parser code is that OPL may wish to make modifications to expressions relatively deep within the expression tree. For example, OPL might wish to support untyped null literal (whereas KQL requires parsing null as `string(null)`), or string interpolation..

When parsing strings, for example, we wind up in a call stack like:
```
...
scalar_expression::parse_scalar_expression
scalar_expression::parse_scalar_unary_expression
scalar_primitive_expressions::parse_type_unary_expressions
scalar_primitive_expressions::parse_string_literal
```
and at the bottom of this call stack, we need to call some custom OPL specific string parsing code. To accommodate this `parse_string_literal` could becomes generic over the `Rule` which implements some trait for string parsing. From the perspective of kql-parser crate this adds a complexity, especially as more trait methods are added for custom parsing behaviour. Note that the more custom parsing behaviour that is introduced, the more complex it becomes to support and the more dubious are the benefits of sharing the parser code in the first place. At the point it stops being worth it, the complexity actually makes it harder back out.

**TL;DR** - without a shared grammar, sharing the parser code would lead to a brittle parser implementation OPL unless it implements its own test suite (which means half the code from kql-parser kind of gets duplicated anyway) and also introduces extra complexity into the kql-parser to support custom behaviour for certain types. Given these drawbacks, and the benefits/desire to be masters of our own destiny, I propose OPL just implement its own parser.

PR #1735, opened 2026-01-07
Internal logging code path: Raw logger support

jmacd:
Implements new internal logging configuration option.

~Changes the default logging configuration to use internal logging at level INFO. Previously, default logging was disabled.~

Implements a lightweight Tokio tracing layer to construct partially-encoded OTLP bytes from the Event, forming a struct that can be passed through a channel to a global subscriber.

As the first step, implements "raw logging" directly to the console using simple write! macros and the view object for LogRecord to interpret the partial encoding and print it. The raw logging limits console message size to 4KiB.

~Adds a new `configs/internal-telemetry.yaml` to demonstrate this configuration.~

Adds benchmarks showing good performance, in the 50-200ns range to encode or encode/format:

```
encode/0_attrs/100_events
                        time:   [5.5326 µs 5.5691 µs 5.6054 µs]
                        change: [−7.3098% −4.0342% −1.9226%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
encode/3_attrs/100_events
                        time:   [8.5902 µs 8.6810 µs 8.7775 µs]
                        change: [−5.7968% −3.2559% −1.1958%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  2 (2.00%) low mild
  2 (2.00%) high mild
  3 (3.00%) high severe
encode/10_attrs/100_events
                        time:   [19.583 µs 19.764 µs 19.944 µs]
                        change: [−1.5682% +0.0078% +1.3193%] (p = 0.99 > 0.05)
                        No change in performance detected.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
encode/0_attrs/1000_events
                        time:   [53.424 µs 53.874 µs 54.289 µs]
                        change: [−2.8602% −1.8582% −0.9413%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high severe
encode/3_attrs/1000_events
                        time:   [84.768 µs 85.161 µs 85.562 µs]
                        change: [−3.3406% −2.4035% −1.5473%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  4 (4.00%) high mild
encode/10_attrs/1000_events
                        time:   [193.04 µs 194.07 µs 195.13 µs]
                        change: [−1.8940% −0.1358% +1.7994%] (p = 0.89 > 0.05)
                        No change in performance detected.
Found 7 outliers among 100 measurements (7.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  2 (2.00%) high mild
  3 (3.00%) high severe

format/0_attrs/100_events
                        time:   [26.281 µs 26.451 µs 26.633 µs]
                        change: [−16.944% −14.312% −10.992%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild
  4 (4.00%) high severe
format/3_attrs/100_events
                        time:   [38.813 µs 39.180 µs 39.603 µs]
                        change: [−8.0880% −6.7812% −5.5109%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  4 (4.00%) high mild
  2 (2.00%) high severe
format/10_attrs/100_events
                        time:   [70.655 µs 71.176 µs 71.752 µs]
                        change: [−4.8840% −3.9457% −3.0096%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  4 (4.00%) high mild
format/0_attrs/1000_events
                        time:   [295.80 µs 310.56 µs 325.75 µs]
                        change: [−3.2629% −0.5673% +2.4337%] (p = 0.71 > 0.05)
                        No change in performance detected.
Found 10 outliers among 100 measurements (10.00%)
  3 (3.00%) high mild
  7 (7.00%) high severe
format/3_attrs/1000_events
                        time:   [422.93 µs 430.42 µs 439.21 µs]
                        change: [−1.3953% +0.8886% +3.3330%] (p = 0.46 > 0.05)
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  5 (5.00%) high mild
format/10_attrs/1000_events
                        time:   [720.96 µs 725.68 µs 730.81 µs]
                        change: [−15.540% −13.383% −11.371%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 9 outliers among 100 measurements (9.00%)
  1 (1.00%) low mild
  5 (5.00%) high mild
  3 (3.00%) high severe

encode_and_format/0_attrs/100_events
                        time:   [32.698 µs 32.914 µs 33.147 µs]
                        change: [−9.4066% −7.8944% −6.3427%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  2 (2.00%) low mild
  3 (3.00%) high mild
  3 (3.00%) high severe
encode_and_format/3_attrs/100_events
                        time:   [48.927 µs 49.498 µs 50.133 µs]
                        change: [−7.2473% −5.1069% −2.7211%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  3 (3.00%) high mild
  7 (7.00%) high severe
encode_and_format/10_attrs/100_events
                        time:   [95.328 µs 96.088 µs 96.970 µs]
                        change: [−6.3169% −4.9414% −3.6501%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
encode_and_format/0_attrs/1000_events
                        time:   [326.65 µs 328.86 µs 331.27 µs]
                        change: [−41.188% −39.915% −38.764%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 7 outliers among 100 measurements (7.00%)
  6 (6.00%) high mild
  1 (1.00%) high severe
encode_and_format/3_attrs/1000_events
                        time:   [500.59 µs 504.82 µs 509.33 µs]
                        change: [−50.787% −48.877% −47.483%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
encode_and_format/10_attrs/1000_events
                        time:   [944.34 µs 951.79 µs 960.38 µs]
                        change: [−55.389% −54.741% −54.065%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
```

### DIFF
--- a/rust/otap-dataflow/benchmarks/benches/self_tracing/main.rs
+++ b/rust/otap-dataflow/benchmarks/benches/self_tracing/main.rs
@@ -125,7 +125,7 @@ macro_rules! emit_log {
             attr_str3 = "string3",
             attr_bool2 = false,
             attr_float2 = 5.678,
-            attr_int2 =7 100u64,
+            attr_int2 = 100u64,
             attr_str4 = "string4",
             "benchmark message"
         )


rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
```diff
@@ -0,0 +1,639 @@
...
+    }
+}
+
+impl<S> TracingLayer<S> for RawLayer
+where
+    S: Subscriber + for<'a> LookupSpan<'a>,
+{
+    // Allocates a buffer on the stack, formats the event to a LogRecord
+    // with partial OTLP bytes.
+    fn on_event(&self, event: &Event<'_>, _ctx: Context<'_, S>) {
```
cijothomas, line 325: `fmt` layer takes care of the "span" attributes too, so we can get more contextual information in logs, without having to explicitly provide them. Let's add a TODO here, to bring that support to our custom layer as well. (I am not sure if that'll be super easy or not...better tackle it in follow up)

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
```diff
@@ -23,26 +27,100 @@ pub struct OpentelemetryClient {
...
+            let sdk_layer = opentelemetry_appender_tracing::layer::OpenTelemetryTracingBridge::new(
+                &logger_provider,
+            );
+
+            // Try to initialize the global subscriber. In tests, this may fail if already set,
+            // which is acceptable as we're only validating the configuration works.
+            let _ = tracing_setup.with(fmt_layer).with(sdk_layer).try_init();
+            (Some(logger_provider), runtime)
+        } else {
+            let writer = if std::env::var("NO_COLOR").is_ok() {
```
cijothomas: wondering why we use env variable for color enable feature?

rust/otap-dataflow/crates/telemetry/src/self_tracing/mod.rs
```diff
@@ -0,0 +1,100 @@
...
+use bytes::Bytes;
+use encoder::DirectFieldVisitor;
+use otap_df_pdata::otlp::ProtoBuffer;
+use std::time::{SystemTime, UNIX_EPOCH};
+use tracing::Event;
+use tracing::callsite::Identifier;
+use tracing::{Level, Metadata};
+
+pub use encoder::DirectLogRecordEncoder;
+pub use formatter::{ConsoleWriter, RawLayer as RawLoggingLayer};
```
cijothomas: why not use `RawLoggingLayer` always?

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
```diff
@@ -23,26 +27,100 @@ pub struct OpentelemetryClient {
...
+                ConsoleWriter::no_color()
+            } else {
+                ConsoleWriter::color()
+            };
+            // See comment above.
+            let _ = tracing_setup.with(RawLoggingLayer::new(writer)).try_init();
```
cijothomas: lets do a console.write at least when this fails, so it won't be totally silent! its expected to occur in tests only, so there won't be any noise.

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
```diff
@@ -0,0 +1,100 @@
+// Copyright The OpenTelemetry Authors
```
andborja, line 1: use the module name as file name (self_tracing.rs) syntax instead of mod.rs

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
jmacd: 🤷 it's a thing? https://crates.io/crates/no_color https://no-color.org/

cijothomas: have to learn something new every day :)

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
jmacd: I didn't realize this was a thing people have feels about! Done.

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
```diff
@@ -0,0 +1,639 @@
...
+        // TODO: there are allocations implied here that we would prefer
+        // to avoid, it will be an extensive change in the ProtoBuffer to
+        // stack-allocate this temporary.
+        let record = LogRecord::new(event);
+        let callsite = SavedCallsite::new(event.metadata());
```
cijothomas, line 330: nit: Don't think we need to store metadata() into own `SavedCallsite`? Could we just store the `Metadata` as-is?

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
```diff
@@ -0,0 +1,639 @@
...
+
+        let mut buf = [0u8; LOG_BUFFER_SIZE];
+        let len = self.writer.write_log_record(&mut buf, &record, &callsite);
+        self.writer.write_line(callsite.level, &buf[..len]);
```
cijothomas: mm.. If I read this correctly, this looks like we are going to write to stdout/err in hot path of every event... won't the stdout mutex cause contention? We don't have much INFO lines, so not noticeable today in perf tests, but ....

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
```diff
@@ -0,0 +1,639 @@
...
+    // with partial OTLP bytes.
+    fn on_event(&self, event: &Event<'_>, _ctx: Context<'_, S>) {
+        // TODO: there are allocations implied here that we would prefer
+        // to avoid, it will be an extensive change in the ProtoBuffer to
+        // stack-allocate this temporary.
+        let record = LogRecord::new(event);
+        let callsite = SavedCallsite::new(event.metadata());
+
+        let mut buf = [0u8; LOG_BUFFER_SIZE];
+        let len = self.writer.write_log_record(&mut buf, &record, &callsite);
```
cijothomas, line 333: what is the behavior when log exceeds LOG_BUFFER_SIZE?

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
```diff
@@ -0,0 +1,639 @@
...
+        }
+
+        // Write attributes if present
+        let mut attrs = record.attributes().peekable();
+        if attrs.peek().is_some() {
+            let _ = w.write_all(b" [");
+            let mut first = true;
+            for attr in attrs {
+                if Self::is_full(w) {
+                    break;
```
cijothomas, line 212: instead, we can determine full to be "trimmed...".length() smaller than max_size, and then fill in "trimmed...", so readers know something got trimmed at the end..

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
```diff
@@ -23,26 +27,100 @@ pub struct OpentelemetryClient {
...
-        // Extract the meter provider and runtime by consuming the MeterProvider
-        let (meter_provider, runtime) = meter_provider.into_parts();
-
-        let logger_provider = LoggerProvider::configure(sdk_resource, &config.logs, runtime)?;
-
-        let (logger_provider, runtime) = logger_provider.into_parts();
+        let (meter_provider, runtime) =
+            MeterProvider::configure(sdk_resource.clone(), &config.metrics, runtime)?.into_parts();
+
+        let tracing_setup = tracing_subscriber::registry().with(get_env_filter(config.logs.level));
```
andborja: Can we think on something decoupled from the tracing framework or other libraries? As we are implementing the macros, can we just implement something that publishes the messages to the channel (or not) without passing through an external library?
That is global, so we might not be able to shut off the path for the internal telemetry pipeline telemetry

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
```diff
@@ -0,0 +1,639 @@
...
+}
+
+impl<S> TracingLayer<S> for RawLoggingLayer
+where
+    S: Subscriber + for<'a> LookupSpan<'a>,
+{
+    // Allocates a buffer on the stack, formats the event to a LogRecord
+    // with partial OTLP bytes.
+    fn on_event(&self, event: &Event<'_>, _ctx: Context<'_, S>) {
+        // TODO: there are allocations implied here that we would prefer
```
cijothomas, line 326: for future consideration: could we leverage thread-local buffer and re-use it?

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
```diff
@@ -0,0 +1,639 @@
...
+        let _ = w.write_all(b"\n");
+
+        w.position() as usize
+    }
+
+    /// Write callsite details as event_name to buffer.
+    #[inline]
+    fn write_event_name(w: &mut BufWriter<'_>, callsite: &SavedCallsite) {
+        let _ = w.write_all(callsite.target.as_bytes());
+        let _ = w.write_all(b"::");
```
cijothomas, line 156: wow! "target::name". `fmt` subscriber does not print event name, so we worked around it by duplicating it as an attribute named "name" 
https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/telemetry/src/internal_events.rs#L38

(https://github.com/tokio-rs/tracing/pull/3358 PR in fmt to add it!)

rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
```diff
@@ -0,0 +1,305 @@
...
+    pub fn new(buf: &'buf mut ProtoBuffer) -> Self {
+        Self { buf }
+    }
+
+    /// Reset the underlying buffer.
+    pub fn clear(&mut self) {
+        self.buf.clear();
+    }
+
+    /// Encode a tracing Event as a complete LogRecord message.
```
cijothomas, line 32: why we need it? we print timestamp,eventname etc directly, so only body/attributes need to be encoded?

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: Fixed, thanks.

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: Yes. Thanks for pointing this out -- I've come around to thinking that our decision earlier to set default logging to None was not good for the long term. I want to get us logging, especially for developers, by default. I use "raw" logging to be suggestive that in this mode we try to avoid complexities, though we do allocate in this code path. Still, in production I would not choose raw logging. 

(Aside: I learned things about `tracing`. It's inconvenient that we can't reconfigure the global subscriber, to say the least, otherwise I would install raw logging early in the process and reconfigure it with a production-grade logging object later, whatever it is. Since we can't do that, I just wanted to wire this up any way I could, later it can be an option. I'll write this up!)

I think we should guide the components away from logging anything at INFO or above that would happen during a perf test, and maybe we should monitor how many logs are happening somehow.

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: For one thing I left a TODO about how the protocol buffer is not limited in size. When the write fails, some amount of the log is lost. I sort of don't care because this is raw logging, but I'd prefer if we made a note about truncation (there's a TODO).

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: I made it a TODO in this PR, otherwise yes. I thought I had already written this TODO somewhere, now added.

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
jmacd: Yes. We will likely still want to use the env_logger filter, since it's so well known to developers. (I appreciate how the `log` package used by DataFusion also uses the same filter, for example.)

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: Yes! See https://github.com/open-telemetry/otel-arrow/pull/1741. I would have the thread-local variable accumulate more than one log record in a `LogsData`, that's why the DirectLogsProtoEncoder takes a mutable reference to the ProtoBuffer, and it will be possible to sort by Scope before we do.

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: Interesting. I would follow that syntax when it lands.

rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
jmacd: This will let us form the complete data envelope and send to an internal pipeline, i.e., make OtlpProtoBytes and consume.

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
cijothomas: The original decision to set logging default to None, was until we explore a way to avoid "global" tracing subscriber... I'd expect the default logging would be INFO in production once we tackle it.. 

> Still, in production I would not choose raw logging.

If this raw_logger is not meant for production, then we can just use `fmt` layer and accept its perf hit, if any. (Btw, fmt is used in production by many, and I think they use some channel/buffer instead of writing directly to stdout)

> I think we should guide the components away from logging anything at INFO or above that would happen during a perf test, and maybe we should monitor how many logs are happening somehow.

In happy path, I dont expect much logs at INFO/above. But we'd be wanting to stress test the engine in poor scenarios (like disk failing, network failing etc.) and there'd be logs, and it'd start contention due to global lock in stdout.writes..

rust/otap-dataflow/crates/telemetry/src/opentelemetry_client.rs
jmacd: Done

rust/otap-dataflow/crates/telemetry/src/self_tracing/mod.rs
jmacd: Done

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: Noted. 

Secretly (haha!) I am interested in a stateless approach where we emit one event per event, then a query engine can sort out span duration :)

cijothomas: The usage of `span` is not only for duration calculation or defining hierarrchy. I am mostly interested in the ability to do in-process-contextual-enrichment.
Eg: https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/controller/src/lib.rs#L385-L388

Every logging statement occurring within that context will get core-id for *free*. (there is no some perf cost here...)

jmacd:
From https://github.com/open-telemetry/otel-arrow/pull/1672#discussion_r2662754289

Have discussed with @andborja and I realize that the `internal::enabled` setting is not well placed. Andres suggested having something at a higher level, not to conflict with the OTel declarative configuration model, and to have more of a mode selection like "noop", "raw", "global", "threaded", "opentelemetry", each with different behavior.

jmacd:
I removed the changes in crates/config and opentelemetry_client.rs, leaving only the encoding logic, formatting logic used and benchmarks for now.

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: See #1746

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: See #1746

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: I see that we need more nuance even from the start, so I removed the contentious changes from this PR leaving only the new encoder/formatter/layer.

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
cijothomas: thanks!

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: We all want more context! The way I read #1727 we want more context than only the variables provided in a span; there are other ways we can instrument core_id, and I'd like to leave this open. I think we should sort out a few details about how we will represent component identity, today we have a MetricKey for each metric event, I think we want to use that for Logs (and I believe that includes the core ID).

rust/otap-dataflow/crates/telemetry/src/self_tracing/README.md
```diff
@@ -0,0 +1,29 @@
...
+OTAP-Dataflow engine.
+
+## OTLP bytes first
+
+This module currently implements encoding support for OTLP bytes, in
+two forms:
+
+- Partial: The `LogRecord` type encodes the dynamic arguments from the event
+  along with a timestamp, yielding a representation that can be passed into
+  an internal pipeline because it is already encoded nas bytes. This representation
```
lquerel:
```suggestion
  an internal pipeline because it is already encoded as bytes. This representation
```

rust/otap-dataflow/crates/telemetry/src/self_tracing/README.md
```diff
@@ -0,0 +1,29 @@
...
+  allows sorting and filtering records before encoding full OTLP messages.
+- Full: The `DirectLogRecordEncoder` type supports appending the OTLP bytes
+  representation for the complete LogRecrd (without Scope and Resource wrappers).
```
lquerel:
```suggestion
  representation for the complete LogRecord (without Scope and Resource wrappers).
```

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
andborja: It is encouraged to use the "new" format:
https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
```diff
@@ -0,0 +1,114 @@
...
+
+pub mod encoder;
+pub mod formatter;
+
+use bytes::Bytes;
+use encoder::DirectFieldVisitor;
+use otap_df_pdata::otlp::ProtoBuffer;
+use std::time::{SystemTime, UNIX_EPOCH};
+use tracing::callsite::Identifier;
+use tracing::{Event, Level, Metadata};
```
andborja, line 18: Why not our own objects to break this dependency?

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
```diff
@@ -0,0 +1,114 @@
+// Copyright The OpenTelemetry Authors
```
andborja, line 1: self_tracing or internal_tracing? The concept used as of now is "internal", not "self"

rust/otap-dataflow/benchmarks/benches/self_tracing/main.rs
```diff
@@ -0,0 +1,199 @@
...
+#[cfg(not(windows))]
+use tikv_jemallocator::Jemalloc;
+
+#[cfg(not(windows))]
+#[global_allocator]
+static GLOBAL: Jemalloc = Jemalloc;
+
+/// The operation to perform on each event within the layer.
+#[derive(Clone, Copy)]
+enum BenchOp {
```
cijothomas, line 33: https://github.com/open-telemetry/opentelemetry-rust/pull/3307/changes

I was adding similar benches in OTel Rust to measure the cost of conversions!

rust/otap-dataflow/benchmarks/benches/self_tracing/main.rs
```diff
@@ -0,0 +1,199 @@
...
+}
+
+impl<S> Layer<S> for BenchLayer
+where
+    S: Subscriber + for<'a> LookupSpan<'a>,
+{
+    fn on_event(&self, event: &Event<'_>, _ctx: tracing_subscriber::layer::Context<'_, S>) {
+        match self.op {
+            BenchOp::Encode => {
+                for _ in 0..self.iterations {
```
cijothomas, line 63: any advantage doing the iteration ourselves, than doing just once, and let criterion do the repeated execution itself?

rust/otap-dataflow/crates/telemetry/src/self_tracing/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Raw logging handler
+
+This package supports a `RawLoggingLayer` tracing subscriber that
+prints colorized or uncolorized messages on the console. The
+formatting code path in this module is safe for logging in critical
+regions and to be used as a fallback inside other logging handlers.
+It uses no synchronization and depends only on the console.
```
cijothomas, line 25: console writing internally uses locks.. not sure if that is worth mentioning here.

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
jmacd: I think of self-tracing as including this "raw" logging support for low-level logging about the process itself.
I think of "internal" logging as logging that passes through an internal pipeline starting at ITR.

I'm not against moving combining these two directories, if we decide it makes sense.

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
jmacd: I am not trying to avoid use of Tokio tracing, and I don't think we should. Even when our custom macros use the effect handler directly, they are likely to be Tokio tracing macros.

rust/otap-dataflow/benchmarks/benches/self_tracing/main.rs
jmacd: I did this I suppose to be certain that overhead from the subscriber isn't being picked up. I'm pleased when I see 100 executions at 8.6xxx and 1000 executions at 85.xxx.

rust/otap-dataflow/crates/telemetry/src/self_tracing/README.md
jmacd: I think the meaning is clear for the users of this low-level interface. If there's user-level configuration that allows configuring "raw" logging of this nature, then it will be important to emphasize this. Can't the OS buffer stdout any way it likes? I assume this is console and OS dependent.

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
andborja: This layer of the architecture should not be coupled to external libraries, especially if they are just simple data structures. This is independent of the decision or not to couple another layer to tokio tracing.

PR #996, opened 2025-08-22
[otap-df-engine] Simplify timer control

utpilla:
## Changes
- Use only one HashMap to track the timer state instead of three different ones
- Instead of 3-4 HashMap lookups per operation we now only have one lookup
- This also offers better cache locality

PR #1737, opened 2026-01-07
feat: support attribute insertion in OTAP transform

### LINKED ISSUE #1035: OTAP transform function to insert attributes

ThomsonTan:
Fix #1035

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/error.rs
+++ b/rust/otap-dataflow/crates/engine/src/error.rs
@@ -394,6 +394,24 @@ pub enum Error {
         /// Prost error string
         reason: String,
     },
+
+    /// No default output port is configured when multiple ports are connected.
+    #[error(
+        "Ambiguous default out port for node {node}: multiple ports connected and no default configured"
+    )]
+    NoDefaultOutPort {
+        /// The node that has no default port configured.
+        node: NodeId,
+    },
+
+    /// An unknown output port was specified.
+    #[error("Unknown out port '{port}' for node {node}")]
+    UnknownOutPort {
+        /// The node where the unknown port was referenced.
+        node: NodeId,
+        /// The name of the unknown port.
+        port: PortName,
+    },
 }
 
 impl Error {
@@ -410,6 +428,7 @@ impl Error {
             Error::InvalidHyperEdge { .. } => "InvalidHyperEdge",
             Error::IoError { .. } => "IoError",
             Error::JoinTaskError { .. } => "JoinTaskError",
+            Error::NoDefaultOutPort { .. } => "NoDefaultPort",
             Error::NodeControlMsgSendError { .. } => "NodeControlMsgSendError",
             Error::PDataError { .. } => "PDataError",
             Error::PdataConversionError { .. } => "PdataConversionError",
@@ -425,6 +444,7 @@ impl Error {
             Error::TooManyNodes {} => "TooManyNodes",
             Error::UnknownExporter { .. } => "UnknownExporter",
             Error::UnknownNode { .. } => "UnknownNode",
+            Error::UnknownOutPort { .. } => "UnknownPort",
             Error::UnknownProcessor { .. } => "UnknownProcessor",
             Error::UnknownReceiver { .. } => "UnknownReceiver",
             Error::UnsupportedNodeKind { .. } => "UnsupportedNodeKind",
--- a/rust/otap-dataflow/crates/engine/src/local/processor.rs
+++ b/rust/otap-dataflow/crates/engine/src/local/processor.rs
@@ -34,7 +34,7 @@
 
 use crate::control::{AckMsg, NackMsg};
 use crate::effect_handler::{EffectHandlerCore, TelemetryTimerCancelHandle, TimerCancelHandle};
-use crate::error::{Error, ProcessorErrorKind, TypedError};
+use crate::error::{Error, TypedError};
 use crate::local::message::LocalSender;
 use crate::message::Message;
 use crate::node::NodeId;
@@ -155,13 +155,28 @@ impl<PData> EffectHandler<PData> {
                 .send(data)
                 .await
                 .map_err(TypedError::ChannelSendError),
-            None => Err(TypedError::Error(Error::ProcessorError {
-                processor: self.processor_id(),
-                kind: ProcessorErrorKind::Configuration,
-                error:
-                    "Ambiguous default out port: multiple ports connected and no default configured"
-                        .to_string(),
-                source_detail: String::new(),
+            None => Err(TypedError::Error(Error::NoDefaultOutPort {
+                node: self.processor_id(),
+            })),
+        }
+    }
+
+    /// Attempts to send a message without awaiting.
+    ///
+    /// Unlike `send_message`, this method returns immediately if the downstream
+    /// channel is full, allowing the caller to handle backpressure without awaiting.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`TypedError::ChannelSendError`] containing [`SendError::Full`] if the
+    /// channel is full, or [`SendError::Closed`] if the channel is closed.
+    /// Returns a [`TypedError::Error`] if no default port is configured.
+    #[inline]
+    pub fn try_send_message(&self, data: PData) -> Result<(), TypedError<PData>> {
+        match &self.default_sender {
+            Some(sender) => sender.try_send(data).map_err(TypedError::ChannelSendError),
+            None => Err(TypedError::Error(Error::NoDefaultOutPort {
+                node: self.processor_id(),
             })),
         }
     }
@@ -183,14 +198,34 @@ impl<PData> EffectHandler<PData> {
                 .send(data)
                 .await
                 .map_err(TypedError::ChannelSendError),
-            None => Err(TypedError::Error(Error::ProcessorError {
-                processor: self.processor_id(),
-                kind: ProcessorErrorKind::Configuration,
-                error: format!(
-                    "Unknown out port '{port_name}' for node {}",
-                    self.processor_id()
-                ),
-                source_detail: String::new(),
+            None => Err(TypedError::Error(Error::UnknownOutPort {
+                node: self.processor_id(),
+                port: port_name,
+            })),
+        }
+    }
+
+    /// Attempts to send a message to a specific named out port without awaiting.
+    ///
+    /// Unlike `send_message_to`, this method returns immediately if the downstream
+    /// channel is full, allowing the caller to handle backpressure without awaiting.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`TypedError::ChannelSendError`] containing [`SendError::Full`] if the
+    /// channel is full, or [`SendError::Closed`] if the channel is closed.
+    /// Returns a [`TypedError::Error`] if the port does not exist.
+    #[inline]
+    pub fn try_send_message_to<P>(&self, port: P, data: PData) -> Result<(), TypedError<PData>>
+    where
+        P: Into<PortName>,
+    {
+        let port_name: PortName = port.into();
+        match self.msg_senders.get(&port_name) {
+            Some(sender) => sender.try_send(data).map_err(TypedError::ChannelSendError),
+            None => Err(TypedError::Error(Error::UnknownOutPort {
+                node: self.processor_id(),
+                port: port_name,
             })),
         }
     }
@@ -261,6 +296,7 @@ mod tests {
     use super::*;
     use crate::local::message::LocalSender;
     use crate::testing::test_node;
+    use otap_df_channel::error::SendError;
     use otap_df_channel::mpsc;
     use std::borrow::Cow;
     use std::collections::{HashMap, HashSet};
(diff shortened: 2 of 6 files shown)


rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -746,6 +783,415 @@ mod tests {
...
+                };
+                let decoded = ExportLogsServiceRequest::decode(bytes.as_ref()).expect("decode");
+
+                let res_attrs = &decoded.resource_logs[0]
+                    .resource
+                    .as_ref()
+                    .unwrap()
+                    .attributes;
+
+                println!("Resource attrs: {:?}", res_attrs);
```
albertlockett: I think we can get rid of this println right? Assuming it's leftover debugging code
```suggestion
```

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -746,6 +783,415 @@ mod tests {
...
+
+                assert!(res_attrs.iter().any(|kv| kv.key == "c"));
+                assert!(res_attrs.iter().any(|kv| kv.key == "r"));
+
+                // Scope 'c' should remain
```
albertlockett: I'm not sure I understand this comment. Something like this might make it more clear:
```suggestion
                // Scope 'c' should not be inserted; 'a' should remain
```

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -746,6 +783,415 @@ mod tests {
...
+                let scope_attrs = &decoded.resource_logs[0].scope_logs[0]
+                    .scope
+                    .as_ref()
+                    .unwrap()
+                    .attributes;
+                assert!(!scope_attrs.iter().any(|kv| kv.key == "c"));
+                assert!(scope_attrs.iter().any(|kv| kv.key == "a"));
+
+                // Log 'c' should be deleted; 'b' should remain
```
albertlockett: I think it would be more accurate to say the attribute 'c' was not inserted vs that it was `deleted`
```suggestion
                // Log 'c' should not be inserted; 'b' should remain
```

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -152,6 +163,25 @@ impl AttributesProcessor {
...
+                            if let Some(i) = n.as_i64() {
+                                Some(LiteralValue::Int(i))
+                            } else {
+                                n.as_f64().map(LiteralValue::Double)
+                            }
+                        }
+                        Value::Bool(b) => Some(LiteralValue::Bool(b)),
+                        // Ignore null, object, array as not supported by LiteralValue
+                        _ => None,
+                    };
```
albertlockett: We should probably return an error here rather than silently ignoring if the user supplies a type that's not supported.

We might be able to get this behaviour automatically if we define the `Insert` variant of `Action` as:
```rs
    /// Insert a new attribute.
    Insert {
        /// The attribute key to insert.
        key: String,
        /// The value to insert.
        value: LiteralValue,
    },
```
And define `otap_df_pdata::otap::transform::LiteralValue` as:
```rs
use serde::Deserialize

#[derive(Debug, Clone, PartialEq, Deserialize)]
#[serde(untagged)]
pub enum LiteralValue {
    Str(String),
    Int(i64),
    Double(f64),
    Bool(bool),
}
```

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -746,6 +783,415 @@ mod tests {
...
+
+                let otlp_bytes: OtlpProtoBytes = first.try_into().expect("convert to otlp");
+                let bytes = match otlp_bytes {
+                    OtlpProtoBytes::ExportLogsRequest(b) => b,
+                    _ => panic!("unexpected otlp variant"),
+                };
+                let decoded = ExportLogsServiceRequest::decode(bytes.as_ref()).expect("decode");
+
+                let log_attrs = &decoded.resource_logs[0].scope_logs[0].log_records[0].attributes;
+                assert!(log_attrs.iter().any(|kv| kv.key == "count"));
```
albertlockett, line 912: In this test (and the others that have been added in this module), should we also assert on the value of the attribute that was inserted?

```suggestion
                assert!(
                    log_attrs
                         .iter()
                         .any(|kv| kv.key == "count" && kv.value == Some(AnyValue::new_int(42)));
```

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -480,6 +484,25 @@ fn create_next_eq_array_for_array<T: Array>(arr: T) -> BooleanArray {
...
+pub struct InsertTransform {
+    pub(super) entries: Vec<(String, LiteralValue)>,
+}
+
+impl InsertTransform {
+    #[must_use]
+    pub fn new(entries: Vec<(String, LiteralValue)>) -> Self {
+        Self { entries }
+    }
+}
```
albertlockett, line 507: Instead of using `Vec<(String, LiteralValue)>` here, would it make sense to use `BTreeMap<String, LiteralValue>` like we do for `RenameTransform`? That way it wouldn't be possible to create an `InsertTransform` with invalid duplicate entries.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -585,6 +618,17 @@ impl AttributesTransform {
...
+        if let Some(insert) = &self.insert {
+            let mut insert_keys = BTreeSet::new();
+            for (key, _) in &insert.entries {
+                if !insert_keys.insert(key) {
+                    return Err(Error::InvalidAttributeTransform {
+                        reason: format!("Duplicate key in insert: {key}"),
+                    });
+                }
+            }
+        }
```
albertlockett, line 632: Definitiely it makes sense to validate that there are no duplicate keys being inserted, but I think it also makes sense to validate that the keys of the attributes we're inserting are not the same keys as the attributes being deleted or replaced.

For example, if we had:
```
replace: [("x", "y")]
insert: [("x", String("a"))]
```
There's an ambiguity about what the final result will be.  If `rename` gets applied first, attributes "x" will have the value "a". But if `insert` gets applied before `rename`, "x" will have the original value. To avoid all this ambiguity, currently the logic in this method says that no keys can be duplicated.
```suggestion
        if let Some(insert) = &self.insert {
            for (key, _) in &insert.entries {
                if ! all_keys.insert(key) {
                    return Err(Error::InvalidAttributeTransform {
                        reason: format!("Duplicate key in insert: {key}"),
                    });
                }
            }
        }
```

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3610,3 +3733,1094 @@ mod test {
         assert_eq!(with_stats, plain);
     }
 }
+
+/// Helper to extract string key from a record batch key column at a given index.
+/// Handles both plain Utf8 and Dictionary-encoded columns.
+fn get_key_at_index(key_col: &ArrayRef, idx: usize) -> Option<String> {
```
albertlockett: There's a helper type that can encapsulate this logic: `otap_df_pdata::arrays::StringArrayAccessor`
https://github.com/open-telemetry/otel-arrow/blob/2c3976c9672536835e94dae07a4cc7f26333276e/rust/otap-dataflow/crates/pdata/src/arrays.rs#L641

Instead of having a method defined here specially for this, below in the `create_inserted_batch` you could do
```rs
let key_col = current_batch
            .column_by_name(consts::ATTRIBUTE_KEY)
            .map(StringArrayAccessor::try_new)
            .transpose()?;
```
Then you could probably return an error if the `key_col` ends up being `None`.

This type provides the `str_at` method which returns an `Option<&str>`

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3610,3 +3733,1094 @@ mod test {
...
+    // Build Key column
+    let key_col_idx =
+        schema
+            .index_of(consts::ATTRIBUTE_KEY)
+            .map_err(|_| Error::ColumnNotFound {
+                name: consts::ATTRIBUTE_KEY.into(),
+            })?;
+    let key_type = schema.field(key_col_idx).data_type();
+
+    let new_keys: ArrayRef = match key_type {
```
albertlockett: Instead of using the regular arrow builders and having to write these match statements to append values to arrays that could be dict/native type, we have a set of helper types that can encapsulate this logic. 

For example, `StringArrayBuilder`:
https://github.com/open-telemetry/otel-arrow/blob/2c3976c9672536835e94dae07a4cc7f26333276e/rust/otap-dataflow/crates/pdata/src/encode/record/array.rs#L650-L656

This type also exposes a `append_str_n` method which can be used to append the same string multiple times. This is usually faster than appending the values one at a time. So an optimization we could make here, if we're appending the same key multiple times, would be to use this method. 

There are similar builders for the other types we're inserting below as values (int, double, bool)

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3610,3 +3733,1094 @@ mod test {
...
+        .as_any()
+        .downcast_ref::<PrimitiveArray<UInt16Type>>()
+        .ok_or_else(|| Error::ColumnDataTypeMismatch {
+            name: consts::PARENT_ID.into(),
+            expect: DataType::UInt16,
+            actual: parent_ids.data_type().clone(),
+        })?;
+
+    // Build a set of (parent_id, key) pairs that already exist
+    let mut existing_keys: BTreeMap<u16, BTreeSet<String>> = BTreeMap::new();
```
albertlockett: I'm wondering if there are any ways we could optimize building up the set of attributes we're going to insert.

For example, using a BTreeMap here, and a BTreeSet below for `unique_parents` means that we'll be hashing every parent_id multiple times. It might be faster to use a `RoaringBitmap` for the unique_parent_ids, and maybe we could have a `RoaringBitmap` for each insert entry corresponding to whether the row w/ some parent_id contains the attribute.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3610,3 +3733,1094 @@ mod test {
...
+    schema: &arrow::datatypes::Schema,
+) -> Result<(RecordBatch, usize)> {
+    let parent_ids_arr = parent_ids
+        .as_any()
+        .downcast_ref::<PrimitiveArray<UInt16Type>>()
+        .ok_or_else(|| Error::ColumnDataTypeMismatch {
+            name: consts::PARENT_ID.into(),
+            expect: DataType::UInt16,
+            actual: parent_ids.data_type().clone(),
+        })?;
```
albertlockett: The parent_ids won't always be u16. For example, for metrics datapoint attributes and attributes for span link and span event, the parent ID type can be u32. Also, for u32 IDs, the parent_id_arr can be dictionary encoded (e.g. it's not always a `PrimitiveArray`).

To handle this, we might need to make the function generic over `T` where:
```rs
    T: ParentId,
    <T as ParentId>::ArrayType: ArrowPrimitiveType,
```
(You'll see we do something similar in this file for [`materialize_parent_id_for_attributes`](https://github.com/open-telemetry/otel-arrow/blob/2c3976c9672536835e94dae07a4cc7f26333276e/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L146-L150)).

Then we can get the parent_ids as:
```suggestion
        let parent_ids_arr = MaybeDictArrayAccessor::<PrimitiveArray<T::ArrayType>>::try_new(
            get_required_array(record_batch, consts::PARENT_ID)?,
        )?;
```

To ensure we hande u32 parent IDs correctly, it probably also makes sense to add a test for this somewhere.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3610,3 +3733,1094 @@ mod test {
...
+                expect_oneof: vec![DataType::Utf8],
+                actual: key_type.clone(),
+            });
+        }
+    };
+
+    // We collect columns into a map or vec matching schema order.
+    let mut columns = Vec::with_capacity(schema.fields().len());
+
+    for field in schema.fields() {
```
albertlockett, line 4217: There's a case that the logic here might not handle correctly, which is if we're inserting a type of attribute and the original schema did not previously contain any attributes of this type.

All the `ATTRIBUTE_*` columns are optional in OTAP. For example, if some attribute `RecordBatch` had no values of type `int`, the `ATTRIBUTE_INT` column would be omitted. If we encountered such a record batch and we were inserting an integer attribute, it would not be included in the original batch with this logic.

We should add a test for this and handle it. We might need to write some custom batch concatenation logic for this, rather than relying on arrow`s `concat_batches` compute kernel, or make some modifications to the original batch before we invoke this function

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -785,17 +852,41 @@ pub fn transform_attributes_with_stats(
...
+    // - The transformed batch (rb) to check which (parent_id, key) pairs already exist
+    if let Some(insert) = &transform.insert {
+        if let Some(original_parent_ids) = attrs_record_batch.column_by_name(consts::PARENT_ID) {
+            let (new_rows, count) =
+                create_inserted_batch(&rb, original_parent_ids, insert, rb.schema().as_ref())?;
+            if count > 0 {
+                let combined = arrow::compute::concat_batches(&rb.schema(), &[rb, new_rows])
+                    .map_err(|e| Error::Format {
+                        error: e.to_string(),
+                    })?;
```
albertlockett: I mentioned in https://github.com/open-telemetry/otel-arrow/pull/1737/files#r2704902249 that we might need to be careful about how we invoke this `concat_batches` function due to the case where we're inserting a column that was not contained in the original batch.

There's another case we probably need to handle as well which is, if the column is dictionary encoded and inserting the new value would cause the dictionary to overflow, then we need to either expand the key type (e.g. convert from a Dict<u8> to a Dict<u16>) or convert from Dict<u16> to non dict encoded array.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -653,16 +713,18 @@ pub fn transform_attributes_with_stats(
...
             // Possibly remove any delta-encoding on the parent ID column. If there were any
             // deletes, it could cause issues if the parent_ids are using the transport optimized
             // quasi-delta encoding. This is because subsequent runs of key-value pairs may be
             // joined deleted segments, meaning the delta encoding will change.
+            // Note: if insert_needed was true, parent IDs are already materialized so this is noop.
             let any_rows_deleted = keys_transform_result.keep_ranges.is_some();
-            let should_materialize_parent_ids =
-                any_rows_deleted && schema.column_with_name(consts::PARENT_ID).is_some();
+            let should_materialize_parent_ids = (any_rows_deleted || insert_needed)
+                && schema.column_with_name(consts::PARENT_ID).is_some();
```
albertlockett: FYI in https://github.com/open-telemetry/otel-arrow/pull/1824 we made a change so we always materialize the parent_ids. This was the most expedient way to fix a subtle bug (#996)

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
ThomsonTan: Used Vec to keep the insertion order, but seems the order is not required by the spec. Changed to BTreeMap.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
ThomsonTan: I think this has been fixed. The necessary metadata is added to schema in extend_schema_for_inserts before the insertion.

Issue #1736, opened 2026-01-07
Internal logging bridge, internral logging receiver


jmacd:
This issue covers the creation of an internal logging path which receives (a) Tokio tracing events from 3rd party instrumentation and (b) direct EffectHandler-aware diagnostic logging for individual components. In both cases, we expect to deliver internal logs safely to an internal pipeline.

This requires avoidance of feedback from the consumer-to-producer for internal logging pipelines.

The concept of an Internal Telemetry Receiver is a node in the OTAP Dataflow engine which receives internal telemetry and forwards to the configured pipeline.

See #1663 #1672 #1727 #1735.

PR #1741, opened 2026-01-08
Internal logs architecture document

jmacd:
Document the approach we will take for routing internal logs, see #1736.

### DIFF
--- a/rust/otap-dataflow/docs/self_tracing_architecture.md
+++ b/rust/otap-dataflow/docs/self_tracing_architecture.md
@@ -177,7 +177,7 @@ flowchart LR
 
     UL -->|LogPayload::Singleton| CH
     BL -->|periodic flush<br/>LogPayload::Batch| CH
-    NL -.->|discarded| X[∅]
+    NL -.->|discarded| X[null]
 
     CH --> LC
     LC -->|raw format| CON
@@ -244,7 +244,7 @@ flowchart LR
 
     UL -->|LogPayload::Singleton| CH
     BL -->|periodic flush<br/>LogPayload::Batch| CH
-    NL -.->|discarded| X[∅]
+    NL -.->|discarded| X[null]
 
     CH --> ITR
     ITR -->|OtapPayload<br/>ExportLogsRequest| PROC


rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+[TOKIOEVENT]: https://docs.rs/tracing/latest/tracing/struct.Event.html
+
+## Raw logging
+
+We support formatting events for direct printing to the console from
+OTLP bytes. For the dynamic encoding, these are consumed using
+`otap_df_pdata::views::logs::LogsDataView`, our zero-copy accessor. We
+refer to this most-basic form of printing to the console as raw
+logging because it is a safe configuration early in the lifetime of a
+process. Note that the views implementation
```
lalitb: nit - not related to the proposed architecture, but this sentence seems incomplete :)

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+occurs on the `EffectHandler` thread, it will be routed using thread-local
+state so that event is immediately encoded and stored or flushed, without
+blocking the effect handler.
+
+When a telemetry event is routed directly, as in this case and
+`send_message()` succeeds, it means there was queue space to accept
+the log record on the same core. When this fails, the configurable
+telemetry router will support options to use global logs collection
+thread, a raw logger, or do nothing (dropping the internal log
+record).
```
lalitb: What should be the default for production, if the user doesn't select any of these options. Good to mention that.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+  telemetry under normal circumstances. This isolates cores that are
+  able to process their own internal telemetry.
+- Option to configure internal telemetry multiple ways, including the
+  no-op implementation, multi-threaded subscriber, routing to the
+  same-core ITR, and/or raw logging.
+
+## OTLP-bytes first
+
+As a key design decision, the OTAP-Dataflow internal telemetry data
+path produces OTLP-bytes first. Because OTLP bytes is one of the
```
lalitb: "OTLP-byte first" approach would need OTLP byte encoding of the message - will this be done in the hot-path? As this can be issue given the df_engine’s per-core, non-blocking requirement.

lalitb: I believe I need to review #1735 first, to understand this better :)

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,198 @@
+# Internal Telemetry Collection Architecture & Development Plan
+
+## Architecture
+
+The internal telemetry SDK is designed for the engine to safely
+consume its own telemetry, and we intend for the self-hosted telemetry
+pipeline to be the standard configuration for all OpenTelemetry
+signals.
+
+Consuming self-generated telemetry presents a potential a kind of
```
lquerel:
```suggestion
Consuming self-generated telemetry presents a potential
```

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+
+## Internal telemetry receiver
+
+The Internal Telemetry Receiver or "ITR" is an OTAP-Dataflow receiver
+component that produces telemetry from internal sources. An internal
+telemetry pipeline consists of one or more ITR components and any of
+the connected processor and exporter components reachable from ITR
+source nodes.
+
+To begin with, every OTAP-Dataflow comonent is configured with an
```
lquerel:
```suggestion
To begin with, every OTAP-Dataflow component is configured with an
```

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+path produces OTLP-bytes first. Because OTLP bytes is one of the
+builtin `OtapPayload` formats, once we have the OTLP bytes encoding of
+an event we are able to send to an OTAP-Dataflow pipeline. To obtain
+these bytes, we will build a custom [Tokio `tracing`
+Event][TOKIOEVENT] handler to produce OTLP bytes before dispatching to
+an internal pipeline, used (in different configurations) for first and
+third-party instrumentation.
+
+We use an intermediate representation in which the dynamic elements of
+the `tracing` event are encoded while primtive fields and metadata
```
lquerel:
```suggestion
the `tracing` event are encoded while primitive fields and metadata
```

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+  that drop third-party instrumentation instead of creating feedback.
+- Components under observation (non-ITR components) have internal
+  telemetry events routed to queues in the OTAP-Dataflow pipeline on
+  the same core, this avoids blocking the engine. First-party
+  instrumentation will be handled on the CPU core that produced the
+  telemetry under normal circumstances. This isolates cores that are
+  able to process their own internal telemetry.
+- Option to configure internal telemetry multiple ways, including the
+  no-op implementation, multi-threaded subscriber, routing to the
+  same-core ITR, and/or raw logging.
```
lquerel: Multi-threaded subscriber? Or multi-channel subscriber?

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,198 @@
...
+  telemetry events routed to queues in the OTAP-Dataflow pipeline on
+  the same core, this avoids blocking the engine. First-party
+  instrumentation will be handled on the CPU core that produced the
+  telemetry under normal circumstances. This isolates cores that are
+  able to process their own internal telemetry.
+- Option to configure internal telemetry multiple ways, including the
+  no-op implementation, multi-threaded subscriber, routing to the
+  same-core ITR, and/or raw logging.
+
+## OTLP-bytes first
```
lquerel, line 61: In our recent telemetry guidelines, for events we distinguish between **entity-based attributes** (stable context) and other **event-specific attributes** (dynamic context). Using this terminology, attributes belonging to the stable context do not need to be emitted with every event instance; instead, they are identified by a unique numeric ID that references the attributes in an attribute registry.

The "dynamic" attributes are the ones that should travel as OTLP bytes from the hot path to the cold path. If I recall our recent conversation correctly, you were arguing that building this dynamic map would take roughly the same amount of time in a classic representation as in OTLP bytes form. That seems plausible to me, provided we are careful with this micro-serialization and keep attribute values simple.

In any case, I think we should run a few benchmarks to validate all of this.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+
+- Third-party: Tokio `tracing` global subscriber: third-party log
+  events, instrumentation in code without access to an OTAP-Dataflow
+  `EffectHandler`. These are handled in a dedicated internal telemetry
+  thread.
+- First-party: components with a local or shared `EffectHandler` use
+  dedicated macros (e.g., `otel_info!(effect, "interesting thing")`),
+  these use the configured internal telemetry SDK and for ordinary
+  components (not ITR-downstream) these are routed through the ITR the
+  same core. These are always non-blocking APIs, the internal SDK must
```
lquerel: Or to a core in the same numa node than the core reporting the telemetry.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+process. Note that the views implementation 
+
+This configuration is meant for development purposes, it is likely to
+introduce contention over the console.
+
+## Routing
+
+The two internal logs data paths are:
+
+- Third-party: Tokio `tracing` global subscriber: third-party log
```
lquerel: If I understand well, this approach will be limited to third party library on which we have no control. Right?

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+  events, instrumentation in code without access to an OTAP-Dataflow
+  `EffectHandler`. These are handled in a dedicated internal telemetry
+  thread.
+- First-party: components with a local or shared `EffectHandler` use
+  dedicated macros (e.g., `otel_info!(effect, "interesting thing")`),
+  these use the configured internal telemetry SDK and for ordinary
+  components (not ITR-downstream) these are routed through the ITR the
+  same core. These are always non-blocking APIs, the internal SDK must
+  drop logs instead of blocking the pipeline.
```
lquerel: and keep track of the number of logs dropped.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+the tracing macro layer plus a fresh timestamp. Log record attributes
+and the log event body are encoded as the "attributes and body bytes"
+field of `LogRecord`, the other fields are copied.
+
+With this record, we can defer formatting or encoding the entire
+record until later. We can:
+
+- For raw logging, format directly for the console
+- Finish the full OTLP bytes encoding for the `LogRecord`
+- Sort and filter before combining into a `LogsData`.
```
lquerel: The id of the attribute set corresponding to the entite-based attributes will not be serialized I guess because it needs to be interpreted into the ITR. Also in many situations, I believe the dynamic attribute set will be an empty set so the OTLP-bytes will be more or less the timestamp serialized and the severity number.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+
+We require a way to print OTLP bytes as human-readable log lines. We
+cannot easily re-use the Tokio `tracing` format layer for this,
+however we can use the `LogsDataView` trait with `RawLogsData` to
+format human-readable text for the console directly from OTLP bytes.
+
+This OTLP-bytes-to-human-readable logic will be used to implement raw
+logging.
+
+### Global logs collection thread
```
lquerel: Rather than using the term *global logs collection*, I would find it more explicit to refer to our **internal telemetry system** or **internal telemetry pipeline**. This internal telemetry system can be deployed primarily in two modes:

- **One thread per process** (as today)
- **One thread per NUMA node**, to eliminate any inter-NUMA-node communication

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+needed) instances consuming logs from the global Tokio `tracing`
+subscriber. In this thread, we'll configure the OpenTelemetry SDK or a
+dedicated OTAP-Dataflow pipeline (by configuration) for logs export.
+
+Because global logs collection threads are used as a fallback for
+`EffectHandler`-level logs and because third-party libraries generally
+could call Tokio `tracing` APIs, we arrange to explicitly disallow
+these threads from logging. The macros are disabled from executing.
+
+### Global and Per-core Event Router
```
lquerel: I think a small diagram would help clarify all of this.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
```diff
@@ -0,0 +1,198 @@
...
+record).
+
+## Example configuration
+
+```yaml
+service:
+  telemetry:
+    logs:
+      level: info
+      internal_collection:
```
lquerel: I think that some (if not all) of these parameters should also be overridable at the level of each deployed pipeline. That would make it a truly multi-tenant system.

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: I'd like to take this one step at a time. Presently, the internal logging design is concerned with the mechanics for handling and encoding events without considering entity-based attributes. We will be able to add dynamic context using the appropriate numeric ID to represent "scope" in a future PR.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: Fixed

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: Yes-- the OTLP bytes are coded directly, it's not bad! :)

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: Fixed.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: See the new proposed configuration. The previous document was unclear, I've updated to match a more-clear configuration model, with provider modes and an output mode.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: Yes! Added "NUMA-regional" as a future step and linked to the README for more motivation.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: I've updated this to explain we have one provider mode for third-party logging and one provider mod for engine threads (as well as one for internal telemetry components).

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: Yes, agree. Added.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: I've use "global or regional" and explain this can be one or more threads.

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: Added

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: Please see the new configuration proposal in this PR.

jmacd (commented):
I heavily

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -5,74 +5,180 @@
...
-                    console:
-            "#;
-        let config: LogsConfig = serde_yaml::from_str(yaml_str).unwrap();
-        assert_eq!(config.level, LogLevel::Info);
-        assert_eq!(config.processors.len(), 1);
+/// Logging providers for different execution contexts.
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct LoggingProviders {
+    /// Provider mode for non-engine threads. This defines the global Tokio
+    /// `tracing` subsriber. Default is Unbuffered. Note that Buffered
```
andborja: *subscriber

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,248 @@
...
+
+Consuming self-generated telemetry presents a potential
+feedback loop, situations where a telemetry pipeline creates pressure
+on itself. We have designed for the OTAP dataflow engine to remain
+reliable even with this kind of dependency on itself.
+
+## Internal telemetry receiver
+
+The Internal Telemetry Receiver or "ITR" is an OTAP-Dataflow receiver
+component that produces telemetry from internal sources. An internal
```
andborja: It does not "produces" but "receives"

rust/otap-dataflow/crates/telemetry/ARCHITECTURE.md
jmacd: New default I added new text stating the default would be an asynchronous raw console logger, like:

```
service:
  telemetry:
    logs:
      level: info
      strategies:
        global: unbuffered
        engine: buffered
        internal: noop
      output: raw
```

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,248 @@
...
+components reachable from ITR source nodes.
+
+## Logs instrumentation
+
+The OTAP Dataflow engine has dedicated macros, and every component is
+configured with an internal telemetry SDK meant for primary
+instrumentation. Using the `otel_info!(effect, name, args...)` macro
+requires access the component EffectHandler. This is considered
+first-party internal logging, and other uses of Tokio `tracing` are
+considered third-party internal logging.
```
andborja, line 38: I think we need to design and implement the internal logging and explicitly exclude/ignore third-party logging.

jmacd: The `providers::internal` setting does just that, and it defaults to Noop to prevent internal telemetry pipeline components for logging.

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,248 @@
...
+- Non-blocking interfaces. We prefer to drop and count dropped
+  internal log events than to block the pipeline.
+- Option to configure internal telemetry multiple ways, including the
+  no-op implementation, global or regional logs consumers, buffered and 
+  unbuffered.
+
+## OTLP-bytes first
+
+As a key design decision, the OTAP-Dataflow internal telemetry data
+path produces a partially encoded OTLP-bytes representation first.
```
andborja, line 64: It should be OtapPayload first, no? It will be generated initially from OTLP bytes, but the "transport" object should be OtapPayload in my opinion.

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,248 @@
...
+as raw logging because it is a safe configuration that avoids feedback
+for internal logging.
+
+Note: Raw logging is likely to introduce contention over the console.
+
+In cases where internal logging code is forced to handle its own
+errors, the `otap_df_telemetry::raw_error!` macro is meant for
+emergency use, to report about failures to log.
+
+## Logging provider modes
```
andborja, line 93: Where is this configured? Per component?

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,248 @@
...
+
+The logging configuration supports multiple distinct provider mode
+settings:
+
+- Global: The default Tokio subscriber, this will apply in threads
```
andborja, line 98: I insist that we should not consider this case (at least not initially)

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,248 @@
...
+  that do not belong to an OTAP dataflow engine core.
+- Engine: This is the default configuration for engine core threads.
+- Internal: This is the default configuration for internal telemetry
+  pipeline components.
+  
+Provider mode values are:
+
+- Noop: Ignore these producers
+- Unbuffered: Use a non-blocking write to the internal logs channel.
```
andborja, line 107: By definition the channel has some buffer. How to avoid this at producer level?

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,248 @@
...
+  Unbuffered is the default for the global provider.
+- Buffered: Use a thread-local buffer, requires managed flushing. 
+  The global provider is not supported for buffered logging.
+- OpenTelemetry: Use the OpenTelemetry SDK. This option has the most
```
andborja, line 111: Are we not mixing the concepts here? I see opentelemetry SDK more in the "consumer/destination" side of the internal telemetry

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: For internal pipelines, OtapPayload::OtlpBytes is the fully formed OTLP encoding, yes. The full encoding includes resource and scope, whereas the previously merged `self_tracing` package (in #1735) supports only directly encoding single LogRecords at this time. 

There is still an intermediate representation, called self_tracing::LogRecord which defers encoding the whole OTLP payload immediately. Mainly, for the bytes that must be copied because of lifetime, requiring ownership, we encode OTLP bytes immediately. In the case of raw logging, we skip the whole payload representation.

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: The `logs.rs` draft included in this PR includes a `providers` section to allow logging providers to be configured base on the context.

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: Please explain what you are aiming for? Use of Tokio tracing is a widespread standard for diagnostics that is well understood, already useful, and recommended by OpenTelemetry. We have to make a choice about the Tokio subscriber that we install, and we are going to have a setting that controls the global behavior. Since you seem to want this disabled, you can set `providers::global: noop` for that effect.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -5,74 +5,180 @@
...
+
+/// Logs producer: how log events are captured and routed.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, JsonSchema, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
+pub enum ProviderMode {
+    /// Log events are silently ignored.
+    Noop,
+
+    /// Place into a thread-local buffer.
+    Buffered,
```
andborja: I think this should not be an option at production time. If someone wishes to buffer and batch events, that should be done in an internal telemetry pipeline (not at telemetry production time).

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: The Unbuffered and Buffered modes are associated with a channel, yes. The Noop and Raw mode both avoid the use of a channel, and for this we have distinct providers we can configure. The present proposal include `providers::internal: noop` as the default to avoid telemetry from internal components entirely. You could also choose raw logging to avoid the channel, for example, or configure OpenTelemetry as the provider.

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,248 @@
...
+
+    G -->|tracing event| UL
+    E -->|tracing event| BL
+    I -->|tracing event| NL
+
+    UL -->|LogPayload::Singleton| CH
+    BL -->|periodic flush<br/>LogPayload::Batch| CH
+    NL -.->|discarded| X[∅]
+
+    CH --> LC
```
andborja, line 182: There is one component missing here. There might be some "subscription manager" or some component in charge of pulling the objects from the channel and sending it to the configured destination (console or an internal telemetry receiver channel)

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -5,74 +5,180 @@
 
 pub mod processors;
 
+use crate::error::Error;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
+/// Internal Telemetry Receiver node URN for internal logging using OTLP bytes.
+pub const INTERNAL_TELEMETRY_RECEIVER_URN: &str = "urn:otel:otlp:telemetry:receiver";
```
andborja: why here?

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -5,74 +5,180 @@
 
 pub mod processors;
 
+use crate::error::Error;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
+/// Internal Telemetry Receiver node URN for internal logging using OTLP bytes.
```
andborja: Suggestion: Do not include any source code changes as part of this PR unless it is a complete POC.
I'm not clear how this changes are aligned with the design, and there might be problems when starting the SDK threads when some objects are present.

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: There is a point that I haven't made clear, connected with yours. OpenTelemetry is a provider because it expects to be invoked in context, it's a provider of an API not exactly a destination for telemetry. 

For my proposed configuration here, the `OutputMode` type captures the behavior of the collector, and what you're proposing is that we could make OpenTelemetry be an output, the destination for telemetry _after_ going through the channel? I think the answer is that that is not how OpenTelemetry APIs work, they're designed to be providers. It's not unreasonable to offer OpenTelemetry as a destination after the channel, but it would require translating from OTLP bytes into the OTel structured object since we would have bypassed the OpenTelemetry/Tokio bridge.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
jmacd: The internal telemetry pipeline **is the buffer**. The internal telemetry pipeline starts with a buffer in each engine thread.

rust/otap-dataflow/docs/self_tracing_architecture.md
andborja: We are basically intercepting it/not using it for the telemetry of the engine. We should not even configure it or depend on it at this level. If an external library uses it, it will not be captured as it is out of control from the engine. External libraries could also use other tracing libraries.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
jmacd: If you and other reviewers feel that this configuration is OK, I'll be glad to remove this change and include it in the PR with implementation.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
jmacd: The controller or engine at startup need to know which node is the ITR, and it's implemented in the crates/otap area, this just needs to be somewhere both can depend on, and this is sort of a type of configuration. It could go in the engine, controller, or other low-level crate.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
jmacd: Fixed.

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: Fixed.

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: This "LC" is the logs collector. The design documented in crates/telemetry/README.md calls for this to be NUMA-regional eventually, for now it's a global/single thread. This is one of the dedicated threads described for internal telemetry above, and it pulls from the channel and does what the OutputMode prescribes, meaning it prints to the console or sends to the internal pipeline.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -5,74 +5,178 @@
...
+pub const INTERNAL_TELEMETRY_RECEIVER_URN: &str = "urn:otel:otlp:telemetry:receiver";
+
 /// Internal logs configuration.
-#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Default)]
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct LogsConfig {
```
andborja, line 13: This configuration is more "pipeline" level than engine level. That's why I don't see this structure configured here.
I would expect the configuration to be defined per component, with defaults for the type of component. For example, the internal telemetry receiver and subsequent nodes should be Noop or console only.

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,252 @@
...
+In this configuration, the `InternalTelemetryReceiver` node consumes
+from the channel and emits `OtapPayload::ExportLogsRequest` into the
+pipeline.
+
+```yaml
+service:
+  telemetry:
+    logs:
+      level: info
+      strategies:
```
andborja: This is more for the existing SDK configuration (service/telemetry).
We should configure this block in a separate, first level block not coupled to the SDK configuration.

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,252 @@
...
+        global: unbuffered
+        engine: buffered
+        internal: noop
```
andborja, line 200: This configuration should not be global to the engine, but local to a pipeline (node for our case?)

jmacd: In the accompanying document, I tried to capture this concept that the Internal Telemetry Receiver and all of its downstream nodes are considered an Internal Telemetry pipeline. All node in the internal telemetry pipeline would have the Noop configuration in this example. This setting is meant to apply to all of those internal telemetry components.

I'm aware of a desire to control the telemetry provider in more fine-grain ways. For example, we might imagine overriding LogLevel at the component level. We can imagine varying the ProviderMode at that level too, it's just more-detailed configuration.

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: Sorry! I had renamed this field "providers" and the type "LoggingProviders" to capture the three top-level default contexts that we can configure, which are global, engine, and internal.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
lquerel: In my opinion, this URN is ambiguous. We should have a URN that clearly expresses that this is a receiver used by our internal telemetry. I would suggest something like `urn:otelcol:internal:receiver`, or something along those lines.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
lquerel, line 13: In the future, I expect a lot of changes at this level. Right now we are talking about `LogConfig`, but in the long run this will be a configuration for the entire internal telemetry pipeline/system and for all internally used signal types (metric set, events, ...). 

I can see this being acceptable as an intermediate step for now, but as I mentioned, I don’t think this configuration is 1) general enough, or 2) in the right place.

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,252 @@
...
+Consuming self-generated telemetry presents a potential
+feedback loop, situations where a telemetry pipeline creates pressure
+on itself. We have designed for the OTAP dataflow engine to remain
+reliable even with this kind of dependency on itself.
+
+## Internal telemetry receiver
+
+The Internal Telemetry Receiver or "ITR" is an OTAP-Dataflow receiver
+component that receives telemetry from internal sources and sends
+sends to an internal telemetry pipeline. An internal
```
lquerel:
```suggestion
component that receives telemetry from internal sources and sends
to an internal telemetry pipeline. An internal
```

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,252 @@
...
+consume its own telemetry, and we intend for the self-hosted telemetry
+pipeline to be the standard configuration for all OpenTelemetry
+signals.
+
+Consuming self-generated telemetry presents a potential
+feedback loop, situations where a telemetry pipeline creates pressure
+on itself. We have designed for the OTAP dataflow engine to remain
+reliable even with this kind of dependency on itself.
+
+## Internal telemetry receiver
```
lquerel:
```suggestion
## Internal Telemetry Receiver (ITR)
```

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,252 @@
...
+
+The Internal Telemetry Receiver or "ITR" is an OTAP-Dataflow receiver
+component that receives telemetry from internal sources and sends
+sends to an internal telemetry pipeline. An internal
+telemetry pipeline consists of one (global) or more (NUMA-regional)
+ITR components and any of the connected processor and exporter
+components reachable from ITR source nodes.
```
lquerel, line 29: IMO, the ITR(s) are part of the Internal Telemetry System or ITS which is the combination of registries (entity and metric set) and an iternal otap-based telemetry pipeline.

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,252 @@
+# Internal Telemetry Logging Pipeline
```
lquerel, line 1: In the near future, I hope that this internal telemetry logging pipeline will become more general and integrate metrics and other types of signals as well.

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,252 @@
...
+    logs:
+      level: info
+      providers:
+        global: unbuffered
+        engine: buffered
+        internal: noop
+      output: raw
+```
+
+```mermaid
```
andborja, line 151: I'm thinking on something a little simpler:
https://github.com/open-telemetry/otel-arrow/pull/1769

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
jmacd: Note that @lquerel also doesn't think the new configuration belongs here. I'm not sure where it belongs? This feels OK to me.

I do see this as configuration for the main function, so the `global`, `engine`, and `internal` aspects inside `providers` are for global-level, engine-level, and internal-pipeline logging configuration. As mentioned https://github.com/open-telemetry/otel-arrow/pull/1741#discussion_r2683909495 this is something we can configure component level, per-tenant, etc.

jmacd: @lquerel this position in the YAML is the same as the OTel Collector at `service::telemetry::logs` with the `level` field. I'm open to placing it anywhere you or @andborja could agree. Definitely the resource used for logs and metrics; I would expect other signal's internal configuration to mirror this.

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: Definitely!

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: Thank you @andborja. 

For the record, I think these visions are compatible!

The part you have as "Event manager (configurable)" in the diagram, that's the Internal Telemetry Receiver component. There is one instance in the configuration. As the engine starts, we will have one (global) or one per NUMA region of these, and they can have any configuration we all agree. I have no configuration so far, and (as my next PR will show) in my current draft the ITR uses send_message to the default out-port.

The part where you have multiple internal pipelines. That's already part of the OTAP-Dataflow graph structure, the idea that any component including ITR can have multiple output destinations and policies allowing various behavior in the pipeline-level configuration. So, if the ITR configuration for "event manager" as you describe it can name the out-ports that it wants to send to, I think we'll have essentially the same diagram.

@lquerel please see if this aligns with your thinking, the idea of a single ITR in the OTAP-dataflow pipeline configuration with some sort of policy that it uses to send to one or more out-ports in the config. The ITR configuration is a mapping from something to the out-port name, that is.

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -0,0 +1,252 @@
...
+telemetry pipeline consists of one (global) or more (NUMA-regional)
+ITR components and any of the connected processor and exporter
+components reachable from ITR source nodes.
+
+## Logs instrumentation
+
+The OTAP Dataflow engine has dedicated macros, and every component is
+configured with an internal telemetry SDK meant for primary
+instrumentation. Using the `otel_info!(effect, name, args...)` macro
+requires access the component EffectHandler. This is considered
```
andborja, line 36: I think we need a separate object (not the effecthandler) to hold the logger. The effect handler is instantiated per component, and it is even different data structure depending on the component type (receiver effect handler is different from the processor effect handler).
I think we need something like a "request context", that is the same instance through the whole pipeline path, and different per request.
That is the one that should keep the instance of the logger (and later the TelemetrySettings).

rust/otap-dataflow/docs/self_tracing_architecture.md
andborja: What I think is different is the hard/specific separation and decoupling of the production of the internal telemetry and its routing and consumption.
The ITR is not the same event manager. The event manager is part of the reception and routing of the messages produced by any component in the engine. The ITR is yet another receiver, so it is part of the "consumption" block.
A view with this separation simplify several downstream definitions including the configuration of the different components, and even the need or not of a ITR id in the "production" side.
The channels used here are not the same as the graph queues, so the out-ports and other concepts don't have the same meaning. The ITR is not a processor, it is a receiver, so it does not read the incoming messages from a "port". They have their own flume channel.

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: This sounds possible, and we already have a request-level Context object use for Ack/Nack handling. This is a step above the kind of protection we need here, however, so I would evaluate that objective on its own. Why should we want per-request-context telemetry settings, instead of per-component settings with isolated internal pipelines using entirely different choices w/o runtime dispatch?

andborja: Does this request context available at receiver level (where the message is received/produced)? It should be something that is received since the beginning by the first receiver.
Technically it should be configured per "pipeline", this is a sequence of components for the internal telemetry. This concept does not exist today. We have a graph of components that you don't really know if they are part of the same "pipeline".
As we have nodes, that are linked in sequence, we need to set the logger at the ITR, and then the rest of the nodes in the same request should use the same configuration.
For example, if the node graph is:
ITR->Processor1
OTLPReceiver->Processor1
It would be like 2 pipelines pointing to the same processor, one internal.
The request coming from the ITR to the Processor1 should use Noop (by default), and the other should use the default logger (sending telemetry through the channel).
In general, the logger to use depends on the request, if it is part of an internal telemetry pipeline or not.

PR #1744, opened 2026-01-08
PerfTest - Idle run perf numbers for all available cores

cijothomas:
We were running with core-id=0, so df_engine only spun one thread. 
Adding one more test where we don't restrict cores, so it'll run on all cores (creating as many threads) and show memory/cpu.

(the lab perf machine has 128 cores)

### DIFF
--- a/rust/otap-dataflow/README.md
+++ b/rust/otap-dataflow/README.md
@@ -1,4 +1,4 @@
-# OTAP Dataflow Library
+# OTAP Dataflow Engine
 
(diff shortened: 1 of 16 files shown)


cijothomas:
> LGTM - if my interpretation is correct, namely that this scenario is additive and that the single-core results are still present.

yes both.

lalitb:
nit - Should the [idle_state_report.yaml](https://github.com/open-telemetry/otel-arrow/blob/b524eb10a8f31a1eff89878cbf2ce6fd9fde7d5b/tools/pipeline_perf_test/test_suites/integration/configs/idle_state_report.yaml#L36C1-L37C1) now also include the allocated CPUs, as the number of cores can differ based on the CI (though it is fixed to 128 for now)? Currently, seems we can’t tell how many cores were available when reading the results.

lalitb (approved):

PR #1747, opened 2026-01-09
Add example config for syslog receiver and add info log

cijothomas:


rust/otap-dataflow/configs/README.md
```diff
@@ -111,6 +111,29 @@ OTLP receiver with performance metrics:
...
 
+### `syslog-perf.yaml`
+
+Syslog/CEF receiver with performance metrics:
+
+- Receives syslog messages on UDP `0.0.0.0:5140`
+- Measures and exports performance metrics
+- View metrics at: `http://127.0.0.1:8080/telemetry/metrics?format=prometheus&reset=false`
+
+To send test syslog messages:
```
cijothomas, line 122: thought of adding a very easy to test syslog.. for advanced/heavy-load, the existing perf tests can be used, but this is just for quick developer setting.

PR #1751, opened 2026-01-09
add pipeline support to validation tests

c1ly:
- Moved the validations tests outside of otap-dataflow to directly under the rust directory
- Added pipeline support to run pipelines and validate the output 
- Allows for multiple pipelines to be defined and added to the validation test, currently only supports non transformative processors


albertlockett:
going to try closing and reopening to trigger CI

edit: this didn't work

albertlockett:
@drewrelmas any idea why this PR is not running the rust CI job?

I assume we have some checks in place that this needs to run for any changes under the rust/ directory.

It looks like this should run for any pull request targeting main, so I'm confused about why this isn't running.
https://github.com/open-telemetry/otel-arrow/blob/31f3a173fc5f67cb21445b4afc042111cc545fc2/.github/workflows/rust-ci.yml#L2-L12

.github/workflows/rust-ci.yml
```diff
@@ -130,6 +130,28 @@ jobs:
         with:
           fail_ci_if_error: true
 
+  validation_test:
+    strategy:
+      fail-fast: false
+      fail-fast: false
+      matrix:
```
drewrelmas: Does this need to be a matrix strategy if only processing one folder? Also, there are duplicate `fail-fast` definitions here - that could be why the Rust-CI didn't run.

drewrelmas: @albertlockett See the comment here about `fail-fast`, I think this is the issue with the CI.

albertlockett: great catch, thanks @drewrelmas!

c1ly:
closing this as progress has moved to #1879

PR #1752, opened 2026-01-10
Initial OPL Parser implementation

### LINKED ISSUE #1743: OPL Parser
implement parser for OPL (separate from KQL).

Context: https://github.com/open-telemetry/otel-arrow/pull/1734#issuecomment-3724179634

albertlockett:
closes #1743 

Initial implementation of OPL parser.

This has just enough functionality to support the filtering capabilities we've already implemented in the columnar query engine. I will integrate this with the test suite in that crate in a followup PR, then continue implementing the parser for various other pipeline stages such as attribute modification and if/else statement (continuing from work in #1722). I can fill in more documentation in future PRs as well.

### DIFF
--- a/rust/otap-dataflow/crates/opl/src/opl.pest
+++ b/rust/otap-dataflow/crates/opl/src/opl.pest
@@ -92,7 +92,7 @@ additive_op = _{
 }
 
 additive_expression = {
-    multiplicative_expression ~ (additive_op ~ additive_expression)*
+    multiplicative_expression ~ (additive_op ~ multiplicative_expression)*
 }
 
 multiplicative_op_mul = @{ "*" }
@@ -103,7 +103,7 @@ multiplicative_op = _{
 }
 
 multiplicative_expression = {
-    unary_expression ~ (multiplicative_op ~ multiplicative_expression)*
+    unary_expression ~ (multiplicative_op ~ unary_expression)*
 }
 
 rel_op_eq = @{ "==" }
--- a/rust/otap-dataflow/crates/opl/src/parser/expression.rs
+++ b/rust/otap-dataflow/crates/opl/src/parser/expression.rs
@@ -1,6 +1,8 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
+use std::sync::LazyLock;
+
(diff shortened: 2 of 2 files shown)


rust/otap-dataflow/crates/opl/src/opl.pest
```diff
@@ -0,0 +1,155 @@
...
+
+multiplicative_op_mul = @{ "*" }
+multiplicative_op_div = @{ "/" }
+multiplicative_op_mod = @{ "%" }
+multiplicative_op = _{
+    multiplicative_op_mul | multiplicative_op_div | multiplicative_op_mod
+}
+
+multiplicative_expression = {
+    additive_expression ~ (multiplicative_op ~ multiplicative_expression)*
```
lalitb: This looks like it gives `addition` higher precedence than `multiplication` - so `2 + 3 * 4` would parse as `(2 + 3) * 4` instead of `2 + (3 * 4)`. Is that intentional?

rust/otap-dataflow/crates/opl/src/opl.pest
```diff
@@ -0,0 +1,155 @@
...
+
+additive_op_add = @{ "+" }
+additive_op_sub = @{ "-" }
+additive_op = _{
+    additive_op_add | additive_op_sub
+}
+
+additive_expression = {
+    unary_expression ~ (additive_op ~ additive_expression)*
+}
```
lalitb, line 96: The has right-recursion for addition, which means  `10 - 3 - 1` parse as `10 - (3 - 1)` -> 8 instead of `(10 - 3) - 1` -> 6. Is this intentional, as standard arithmetic operators are left-associative ? Same pattern exists for * / % just below.

rust/otap-dataflow/crates/opl/src/opl.pest
albertlockett: good catch, thanks @lalitb . Fixed in 3b277653

rust/otap-dataflow/crates/opl/src/opl.pest
albertlockett: Good catch, thanks @lalitb . Fixed in af579352

PR #1764, opened 2026-01-12
[otap-df-quiver] Quiver Subscriber API; quiver-e2e test tool

AaronRM:
- Subscriber Registry & Lifecycle Management — Added a subscriber system to enable multiple independent consumers to track progress through the segment stream. Includes registration, activation/deactivation, and bundle claiming with RAII handles (BundleHandle) for ack/reject/defer semantics.
- Durable Progress File Format — Implemented a versioned binary format (`quiver.sub.<id>`) with CRC32 validation for crash-safe progress persistence. Uses atomic write-fsync-rename for durability. Supports per-bundle ack tracking via bitmaps for out-of-order delivery scenarios.
- Disk Budget & Backpressure — Added `DiskBudget` for enforcing storage caps with two retention policies: `Backpressure` (slow down ingestion) and `DropOldest` (force-complete old segments). Supports reserved headroom for WAL rotation and segment finalization.
- Engine API Unification — Extended `QuiverEngine` to be the entry point for ingestion, subscription, and maintenance. Added `maintain()` method for periodic progress flush + segment cleanup, builder pattern for configuration, and blocking `next_bundle_blocking()` with condvar-based wakeup.
- Add a `quiver-e2e` crate for stress testing the persistence flow. Features concurrent ingest/consume, support for multiple Quiver engine instances, TUI dashboard with real-time metrics, configurable disk budgets, jemalloc memory tracking, and subscriber delay simulation.

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/budget.rs
+++ b/rust/otap-dataflow/crates/quiver/src/budget.rs
@@ -148,6 +148,7 @@ impl std::fmt::Debug for DiskBudget {
         f.debug_struct("DiskBudget")
             .field("cap", &self.cap)
             .field("used", &self.used.load(Ordering::Relaxed))
+            .field("reserved_headroom", &self.reserved_headroom)
             .field("policy", &self.policy)
             .finish_non_exhaustive()
     }
@@ -263,7 +264,7 @@ impl DiskBudget {
     ///
     /// See [`for_engine`](Self::for_engine) for rationale.
     #[must_use]
-    pub fn calculate_headroom(segment_size_bytes: u64, wal_max_size_bytes: u64) -> u64 {
+    pub const fn calculate_headroom(segment_size_bytes: u64, wal_max_size_bytes: u64) -> u64 {
         segment_size_bytes + (wal_max_size_bytes / 4)
     }
 
@@ -271,14 +272,14 @@ impl DiskBudget {
     ///
     /// This is `reserved_headroom + segment_size` (need room for at least one segment).
     #[must_use]
-    pub fn minimum_cap(segment_size_bytes: u64, wal_max_size_bytes: u64) -> u64 {
+    pub const fn minimum_cap(segment_size_bytes: u64, wal_max_size_bytes: u64) -> u64 {
         let reserved = Self::calculate_headroom(segment_size_bytes, wal_max_size_bytes);
         reserved + segment_size_bytes
     }
 
     /// Returns the configured cap.
     #[must_use]
-    pub fn cap(&self) -> u64 {
+    pub const fn cap(&self) -> u64 {
         self.cap
     }
 
@@ -296,7 +297,7 @@ impl DiskBudget {
 
     /// Returns the reserved headroom for internal operations.
     #[must_use]
-    pub fn reserved_headroom(&self) -> u64 {
+    pub const fn reserved_headroom(&self) -> u64 {
         self.reserved_headroom
     }
 
@@ -448,7 +449,7 @@ impl DiskBudget {
     /// Called during startup to account for files from previous runs.
     /// This can exceed the cap (to accurately reflect reality).
     pub fn record_existing(&self, bytes: u64) {
-        let _ = self.used.fetch_add(bytes, Ordering::Relaxed);
+        let _ = self.used.fetch_add(bytes, Ordering::Release);
     }
 
     /// Releases bytes when files are deleted.
@@ -458,7 +459,7 @@ impl DiskBudget {
         // Saturating sub to avoid underflow if accounting is slightly off
         let _ = self
             .used
-            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current| {
+            .fetch_update(Ordering::Release, Ordering::Relaxed, |current| {
                 Some(current.saturating_sub(bytes))
             });
     }
@@ -477,7 +478,7 @@ pub struct PendingWrite {
 impl PendingWrite {
     /// Returns the number of bytes reserved.
     #[must_use]
-    pub fn reserved(&self) -> u64 {
+    pub const fn reserved(&self) -> u64 {
         self.reserved
     }
 


rust/otap-dataflow/crates/quiver/src/budget.rs
```diff
@@ -0,0 +1,825 @@
...
+                            cap: self.cap,
+                        });
+                    }
+                    RetentionPolicy::DropOldest => {
+                        // Try to reclaim space
+                        let needed = new_used - self.cap;
+                        if let Some(callback) = self.reclaim_callback.lock().as_ref() {
+                            callback(needed);
+                            // Retry after reclaim attempt
+                            continue;
```
lalitb: This seems to be looping forever if the callback fails to free the space eg. no droppable segment, or delete failed. Should there be bail-out mechanism?

rust/otap-dataflow/crates/quiver/src/segment_store.rs
```diff
@@ -0,0 +1,450 @@
...
+                let mut perms = metadata.permissions();
+                #[allow(clippy::permissions_set_readonly_false)]
+                perms.set_readonly(false);
+                let _ = std::fs::set_permissions(&path, perms);
+            }
+            std::fs::remove_file(&path)?;
+
+            // Release bytes from budget after successful deletion
+            if let (Some(budget), Some(size)) = (&self.budget, file_size) {
+                budget.release(size);
```
lalitb: There seems to be some budget leak here - we only release when the file exists. If the file was already removed (manual rm/crash), we drop the handle and return Ok but never release the budget,

rust/otap-dataflow/crates/quiver/src/budget.rs
AaronRM: Yes, good catch. Changing the callback to include the amount of bytes freed/reclaimed with a fallback to Backpressure.

rust/otap-dataflow/crates/quiver/src/config.rs
```diff
@@ -8,17 +8,51 @@ use std::path::{Path, PathBuf};
...
+    ///
+    /// On crash, the entire open segment is lost (potentially thousands of bundles).
+    /// Provides ~3x higher throughput than WAL mode.
+    ///
+    /// Use this when:
+    /// - Throughput is more important than durability
+    /// - Data can be re-fetched from upstream on crash
+    /// - You have other durability guarantees (e.g., upstream acknowledgement)
+    SegmentOnly,
+}
```
lalitb, line 40: Not blocking for this  PR - but given the number of tuning parameters involved, we should eventually provide an operations guide for production. It would be helpful to document recommended settings for common scenarios - covering durability modes, segment sizes, and WAL max etc. One option could be to provide the  profiles for common scenarios (eg throughput-first profile, durability-first profile), and let the user select one to get the parameters tuned accordingly.

rust/otap-dataflow/crates/quiver/src/budget.rs
AaronRM: Resolved in latest update.

rust/otap-dataflow/crates/quiver/src/segment_store.rs
AaronRM: Resolved in latest update.

rust/otap-dataflow/crates/quiver/src/config.rs
AaronRM: Agreed that we should have a quiver-specific tuning guide, but we should be very deliberate about exposed settings in a `otap-dataflow` processor that uses quiver. Only start with reasonable defaults and a bare-minimum set of parameters initially and expose more as necessary.

rust/otap-dataflow/crates/quiver-e2e/README.md
```diff
@@ -0,0 +1,64 @@
...
+| `--keep-temp` | Keep temp directory after test | false |
+
+## Architecture
+
+The harness spawns multiple threads:
+
+- **Ingest Thread**: Generates and ingests synthetic bundles
+- **Consumer Threads**: Subscribe and consume bundles from segments
+- **Cleanup Thread**: Periodically removes fully-consumed segments
+- **Stats Thread**: Collects memory and disk metrics
```
jmacd, line 64: Screenshot?🥇

rust/otap-dataflow/crates/quiver/src/subscriber/registry.rs
```diff
@@ -0,0 +1,1400 @@
...
+    /// Registers a new subscriber.
+    ///
+    /// If the subscriber was previously registered (state exists from progress file),
+    /// this reactivates the existing state. Otherwise, creates new state
+    /// starting from the latest segment.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the subscriber ID is invalid.
+    pub fn register(&self, id: SubscriberId) -> Result<()> {
```
lalitb, line 201: not sure if I understand the flow here, but should the register of new subscriber also mark them dirty - to ensure in case of any crash before ack, the subscriber is forgotten on restart.

lalitb:
Looks good overall - big change set, so it’s great to have extra eyes on it.

rust/otap-dataflow/crates/quiver/src/subscriber/registry.rs
AaronRM: Subscribers don't have meaning to quiver aside from progress tracking, so we shouldn't have to worry about a new subscriber not getting a progress file persisted (in the case of a crash). In that case, the consumer should end up re-registering the subscriber as new on process restart.

I don't think a change is necessary. Lifetime will be clearer once we integrate with a persistence processor. We can always revisit as part of those changes if needed.

jmacd (approved):

rust/otap-dataflow/crates/quiver-e2e/README.md
AaronRM: image

jmacd: Nice!

I want this as an internal telemetry pipeline exporter where each panel is a configurable metrics aggregate expression.  Then run our dataflow engines inside tmux.

rust/otap-dataflow/crates/quiver-e2e/README.md
```diff
@@ -0,0 +1,64 @@
...
+ingestion -> WAL -> segment finalization -> subscriber consumption -> cleanup.
+
+## Features
+
+- **Steady-State Testing**: Continuous ingest/consume loop with configurable throughput
+- **Memory Leak Detection**: Tracks heap allocations via jemalloc to detect leaks
+- **Disk Usage Monitoring**: Monitors segment and WAL growth over time
+- **TUI Dashboard**: Real-time visualization with sparklines and metrics
+- **Configurable Durability**: Test with full WAL protection or high-throughput
+                               segment-only mode
```
lquerel, line 18: This is really cool!

rust/otap-dataflow/crates/quiver/src/budget.rs
```diff
@@ -0,0 +1,867 @@
...
+    cleanup_callback: Mutex<Option<CleanupCallback>>,
+}
+
+impl std::fmt::Debug for DiskBudget {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // Note: reclaim_callback and cleanup_callback are omitted because
+        // function pointers / closures don't implement Debug
+        f.debug_struct("DiskBudget")
+            .field("cap", &self.cap)
+            .field("used", &self.used.load(Ordering::Relaxed))
```
utpilla, line 150:
```suggestion
            .field("used", &self.used.load(Ordering::Relaxed))
            .field("reserved_headroom", &self.reserved_headroom)
```

rust/otap-dataflow/crates/quiver/src/budget.rs
```diff
@@ -0,0 +1,867 @@
...
+            return Ok(PendingWrite {
+                budget: self.clone(),
+                reserved: 0,
+                committed: AtomicBool::new(false),
+            });
+        }
+
+        // Try to atomically reserve the space
+        loop {
+            let current = self.used.load(Ordering::Acquire);
```
utpilla, line 376: There seems to be a memory ordering inconsistency which makes it a bit confusing. `Acquire` loads are expected to synchronize with `Release` stores.

In this case, `try_reserve()` uses `Acquire` for `self.used` loads, however, both `release()` and `record_existing()` use `Relaxed` ordering for stores, which doesn't form a proper synchronizes-with relationship.

I think we should keep the intent clear. We should either:
- Use `Relaxed` for both loads and stores of `self.used` thereby conveying the intent that the weak ordering is safe to use here
- Or make the stores use `Release` ordering to match the general expectation of using `Acquire/Release` together

rust/otap-dataflow/crates/quiver/src/budget.rs
```diff
@@ -0,0 +1,867 @@
...
+        // function pointers / closures don't implement Debug
+        f.debug_struct("DiskBudget")
+            .field("cap", &self.cap)
+            .field("used", &self.used.load(Ordering::Relaxed))
+            .field("policy", &self.policy)
+            .finish_non_exhaustive()
+    }
+}
+
+impl DiskBudget {
```
utpilla, line 157: nit:

A lot of methods for `DiskBudget` could be made `const`

- `new()`
- `with_reserved_headroom()`
- `unlimited()`
- `for_engine()`
- `calculate_headroom()`
- `minimum_cap()`
- `cap()`
- `reserved_headroom()`

rust/otap-dataflow/crates/quiver/src/budget.rs
```diff
@@ -0,0 +1,867 @@
...
+pub struct PendingWrite {
+    budget: Arc<DiskBudget>,
+    reserved: u64,
+    committed: AtomicBool,
+}
+
+impl PendingWrite {
+    /// Returns the number of bytes reserved.
+    #[must_use]
+    pub fn reserved(&self) -> u64 {
```
utpilla:
```suggestion
    pub const fn reserved(&self) -> u64 {
```

rust/otap-dataflow/crates/quiver/src/budget.rs
AaronRM: Good observation, thanks Utkarsh! I'll push a commit shortly to use `Acquire`/`Release` consistently.

rust/otap-dataflow/crates/quiver/src/budget.rs
AaronRM: The constructor(s) initialize a new Mutex and can't be made const. Will add to the other functions where possible.

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -211,28 +587,280 @@ impl QuiverEngine {
...
+        // Delete the segment files, counting bundles before deletion
+        let mut deleted = 0;
+        let mut bundles_dropped: u64 = 0;
+        for seq in &to_drop {
+            // Count bundles before deleting
+            if let Ok(count) = self.segment_store.bundle_count(*seq) {
+                bundles_dropped += count as u64;
+            }
+            if let Err(e) = self.segment_store.delete_segment(*seq) {
+                tracing::warn!(segment = seq.raw(), error = %e, "Failed to delete force-dropped segment");
```
utpilla, line 803: Shouldn't `bundles_dropped` be decremented if the deletion fails?

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -211,28 +587,280 @@ impl QuiverEngine {
...
+        let _ = self
+            .force_dropped_segments
+            .fetch_add(deleted as u64, Ordering::Relaxed);
+        let _ = self
+            .force_dropped_bundles
+            .fetch_add(bundles_dropped, Ordering::Relaxed);
+
+        // Clean up registry internal state
+        if let Some(&max_dropped) = to_drop.iter().max() {
+            self.registry.cleanup_segments_before(max_dropped.next());
```
utpilla, line 823: I'm wondering, if `delete_segment` fails for some segment in `to_drop` but succeeds for others, we would still clean up registry state for all segments up to `max_dropped.next()`. Could that be a concern with orphaned files with no progress info consuming disk space but not being counted in budget?  (Deletion failures at physical level would be rare so probably not?)

rust/otap-dataflow/crates/quiver/src/engine.rs
AaronRM: I think this should be kept as-is. "Drop" is a logical operation, not a physical one. We are logically dropping it and still removing it from the registry. Deletion fails at the physical level (i.e. some filesystem error that prevented a delete for example) are expected to be a rare case which would at least be captured by warnings.  We could consider adding a counter for "files successfully deleted" (whether via force drop or standard cleanup) if we decide that's a valuable metric.

rust/otap-dataflow/crates/quiver/src/engine.rs
AaronRM: The question is really getting at "what happens if we orphan files on disk because we failed to delete them". This should be a rare case, but this could lead to a "leak" of sorts. If we're hitting this repeatedly it suggests something very wrong is happening that would require manual intervention.

I'll file an issue to track as a follow up.

rust/otap-dataflow/crates/quiver/src/subscriber/state.rs
```diff
@@ -0,0 +1,706 @@
...
+            .iter()
+            .map(|(&seg_seq, progress)| {
+                SegmentProgressEntry::from_bitmap(
+                    seg_seq,
+                    progress.bundle_count(),
+                    progress.resolved_bitmap(),
+                )
+            })
+            .collect()
+    }
```
utpilla, line 412: Not for this PR, but I think we could get rid of the vec allocation here. The consumers of this method could work with `ExactSizeIterator`. There would be some changes required at call sites of this method as well.

```suggestion
    pub fn to_progress_entries(&self) -> impl ExactSizeIterator<Item = SegmentProgressEntry> + '_ {
        self.segments
            .iter()
            .map(|(&seg_seq, progress)| {
                SegmentProgressEntry::from_bitmap(
                    seg_seq,
                    progress.bundle_count(),
                    progress.resolved_bitmap(),
                )
            })
    }
```

rust/otap-dataflow/crates/quiver/src/engine.rs
AaronRM: Opened #1792 to track.

rust/otap-dataflow/crates/quiver/src/subscriber/state.rs
AaronRM: Will require a more substantial change to the callsite at `SubscriberRegistry::flush_progress`. Tracking with #1793

Issue #1761, opened 2026-01-12
OPL Parser: extend (assign) expression


albertlockett:
We should support syntax for assigning properties and attributes.

```kql
extend severity_text = "WARN"
extend attributex["x"] = "y"
extend instrumentation_scope.name = "sdk"
// etc.
```

This should get parsed into our AST expression `DataExpression::Transform(TransformExpression::Set(...))`

PR #1766, opened 2026-01-12
OPL Parser: support `extend`/`set` operator call

albertlockett:
closes: #1761 

Adds parser support for the `extend` operator to the OPL Parser.

Example syntax:
```kql
extend severity_text = "ERROR"
extend attributes["x"] = "y"
extend instrumentation_scope.name = "sdk"
```

`set` is a keyword that can be used as an alias for `extend`. The thinking is, this is a bit easier to understand for users w/out a kql background or users who might be more familiar with the [`set` function from OTTL transform processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/transformprocessor#basic-config)


albertlockett:
closing as it targets wrong remote

PR #1770, opened 2026-01-12
retry auth forever if it fails instead of crashing the exporter

gouslu:
- prevents auth failures from crashing the exporter, and should assist with self healing during, for example a potential outage.
- logging, metrics and thiserror changes are not part of this PR, it will be done in the future as improvements.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
@@ -336,10 +336,7 @@ impl AzureMonitorExporter {
             let delay_secs = (capped_delay_secs + jitter).max(1.0);
             let delay = tokio::time::Duration::from_secs_f64(delay_secs);
 
-            println!(
-                "[AzureMonitorExporter] Retrying in {:.1}s...",
-                delay_secs
-            );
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
```diff
@@ -432,11 +474,8 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
                 biased;
 
                 _ = tokio::time::sleep_until(next_token_refresh) => {
-                    let token = auth.get_token()
```
jmacd, line 435: Please add a TokenRefreshError like

    #[error("Failed to refresh token: {}", error)]
    TokenRefreshError {
      error: String,
     }

and use map_err(|e| Error::TokenRefreshError{ error: e.to_string() }) in two locations.

gouslu: methods in these blocks do not return any errors! and in other parts of the project I'll do it with a future PR

jmacd: My mistake I was reading the old version.

jmacd (approved):

PR #1771, opened 2026-01-12
Internal logging configuration model, background thread(s)

jmacd:
**Too large for review!**

Complete draft configuration from #1741. This establishes a default logging provider configuration that supports 4 choices in 3 different contexts. Choices:

- Noop
- Raw (synchronous to console)
- Immediate (asynchronous to a channel)
- OpenTelemetry

Contexts:

- Global tokio subscriber
- Engine threads of main pipeline
- Internal telemetry pipeline threads

There are major caveats. Two critical new configuration fields named "providers" and "output" are added in the current service::telemetry::logs area. This area is off-spec for the OpenTelemetry declarative configuration, so these belong elsewhere. However, the code needs substantial refactoring and it gets in the way of this intermediate step to refactor more.

The major pieces of the changeset are largely independent:

- The two new configuration fields "providers" and "output", associated LoggingProviders struct, ProviderMode, OutputMode enums (crates/config/src/service/telemetry/logs.rs)
- Add `pub struct PipelineNodes(HashMap<NodeId, Arc<NodeUserConfig>>)`; add a second set of node for the internal pipeline `nodes` (existing) and `internal` (new); refactoring to re-use validation logic (crates/config/src/pipeline.rs)
- Two new internal logging setups in the controller "direct" for simple console printing and "internal" to use the internal telemetry pipeline; this is either a new thread to run simple console printing or a new thread to run_forever() the internal pipeline
- Consolidation of the Tokio tracing setup formerly part of `opentelemetry_client.rs` into a new `telemetry_runtime.rs`; here the 4 provider modes are supported, with a helper to install Tokio subscribers for certain threads (crates/telemetry/src/logs.rs); definition of a flume bounded channel identical to / parallel to the observed state store. (TODO: future consolidation with observed_state); "direct" collector support for asynchronous console logging
- Various scattered eprintln!, log::info!, log::error!, println! and other diagnostics changed to exercise new logging paths
- Internal telemetry receiver component: when output mode is "internal" this collects log records from the channel; special setup required to inject the flume channel to this component as the internal telemetry pipeline is built.
- [Unrelated debugging]: #1334 ~makes it easy to produce panics, setting the internal telemetry pipeline to use OTAP batching created a nice test for what happens when the internal logging path panics or fails; needs separate investigation, setting internal provider to "raw" helpfully shows what happens.~
- OTAP console exporter; re-uses the raw logging support to print OTLP logs data including resource and scope data; new support for encoding OTLP bytes directly from the HashMap of resource values from service::telemetry::resource.
- `raw_error!` macro that is safe to use everywhere including inside Tokio subscribers, for logging about internal telemetry failures as an emergency
- Setup a nice logger early in main.rs so that df_engine prints consistent logs format.

Console logs look like:

```
2026-01-15T18:11:39.467Z  RESOURCE  v1.Resource:  [service.name=test, service.id=1234]
2026-01-15T18:11:39.467Z  │ SCOPE    v1.InstrumentationScope:
2026-01-15T18:11:39.467Z  │ └─ DEBUG otap-df-engine::pipeline.build.complete (crates/engine/src/lib.rs:532):  [name=pipeline.build.complete, pipeline_group_id=default_pipeline_group, pipeline_id=default_pipeline, core_id=0]
```

We expect to fill in scope attributes soon). I had tested with batching format=otap to see if the OTAP representation would naturally group single statements into the resource value. In this configuration, it now prints:

```
2026-01-15T18:18:10.138Z  ERROR otap-df-otap::Processor.BatchingError (crates/otap/src/batch_processor.rs:844):  [name=Processor.BatchingError, signal=Logs, error=Failed to batch OTAP data: Schema error: target schema is not superset of current schema target=Field { "parent_id": UInt16, metadata: {"encoding": "plain"} }, Field { "key": Dictionary(UInt8, Utf8) }, Field { "type": UInt8 }, Field { "str": nullable Dictionary(UInt16, Utf8) }, Field { "int": nullable Dictionary(UInt16, Int64) }, Field { "body": nullable Struct("type": UInt8) }, Field { "resource": nullable Struct("id": nullable UInt16, metadata: {"encoding": "plain"}) }, Field { "scope": nullable Struct("id": nullable UInt16, metadata: {"encoding": "plain"}) } current=Field { "parent_id": UInt16, metadata: {"encoding": "plain"} }, Field { "key": Dictionary(UInt8, Utf8) }, Field { "type": UInt8 }, Field { "str": nullable Dictionary(UInt16, Utf8) }, Field { "body": nullable Struct("type": UInt8) }, Field { "resource": nullable Struct("id": nullable UInt16, metadata: {"encoding": "plain"}) }, Field { "scope": nullable Struct("id": nullable UInt16, metadata: {"encoding": "plain"}) }, Field { "int": nullable Dictionary(UInt16, Int64) }]
```
This is a raw error printed by the internal telemetry pipeline.

That's the error after modifying batch_processor.rs briefly in this PR, otherwise it panics. Since I've modified the batch processor not to return an error, we can no longer see the situation where the internal telemetry pipeline panics. When producers write to the channel after a panic, the send fails resulting in a `raw_error!` which tells the user on the console when logging service dies.

```
2026-01-15T18:23:03.953Z  ERROR otap_df_telemetry::logs::failed to send log (crates/telemetry/src/logs.rs:105):  [err=sending on a closed channel]
```

### DIFF
--- a/rust/otap-dataflow/configs/internal-telemetry.yaml
+++ b/rust/otap-dataflow/configs/internal-telemetry.yaml
@@ -14,7 +14,7 @@ nodes:
         dispatch_strategy: round_robin
     config:
       traffic_config:
-        max_signal_count: 1000
+        max_signal_count: 100000
         max_batch_size: 1000
         signals_per_second: 1000
         log_weight: 100
@@ -59,7 +59,7 @@ internal:
         min_size: 1000
         sizer: items
       flush_timeout: 3s
-      format: otlp
+      format: preserve
   console:
     kind: exporter
     plugin_urn: "urn:otel:console:exporter"
--- a/rust/otap-dataflow/crates/otap/src/batch_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/batch_processor.rs
@@ -73,6 +73,10 @@ pub const DEFAULT_OTLP_MIN_SIZE_BYTES: usize = 262144;
 /// Timeout in milliseconds for periodic flush
 pub const DEFAULT_TIMEOUT_MS: u64 = 200;
 
+/// Log messages
+const LOG_MSG_BATCHING_FAILED_PREFIX: &str = "OTAP batch processor: low-level batching failed for";
+const LOG_MSG_BATCHING_FAILED_SUFFIX: &str = "; dropping";
+
 /// How to size a batch.
 ///
 /// Note: these are not always supported. In the present code, the only
@@ -510,6 +514,18 @@ fn nzu_to_nz64(nz: Option<NonZeroUsize>) -> Option<NonZeroU64> {
     nz.map(|nz| NonZeroU64::new(nz.get() as u64).expect("nonzero"))
 }
 
+async fn log_batching_failed(
+    effect: &mut local::EffectHandler<OtapPdata>,
+    signal: SignalType,
+    err: &impl std::fmt::Display,
+) {
+    effect
+        .info(&format!(
+            "{LOG_MSG_BATCHING_FAILED_PREFIX} {signal:?}: {err}{LOG_MSG_BATCHING_FAILED_SUFFIX}"
+        ))
+        .await;
+}
+
 impl BatchProcessor {
     /// Parse JSON config and build the processor instance with the provided metrics set.
     /// This function does not wrap the processor into a ProcessorWrapper so callers can
@@ -841,21 +857,14 @@ where
                 Ok(v) => v,
                 Err(e) => {
                     self.metrics.batching_errors.add(count as u64);
-                    otap_df_telemetry::otel_error!(
-                        "Processor.BatchingError",
-                        signal = format!("{:?}", self.signal),
-                        error = e.to_string(),
-                    );
-
+                    log_batching_failed(effect, self.signal, &e).await;
                     let str = e.to_string();
-                    let res = Err(str);
+                    let res = Err(str.clone());
                     // In this case, we are sending failure to all the pending inputs.
                     self.buffer
                         .handle_partial_responses(self.signal, effect, &res, inputs.context)
                         .await?;
-                    // Log and drop instead of returning an error to avoid crashing the pipeline.
-                    // The error has been logged and subscribers notified; continue processing.
-                    return Ok(());
+                    return Err(EngineError::InternalError { message: str });
                 }
             };
 
--- a/rust/otap-dataflow/crates/pdata/src/lib.rs
+++ b/rust/otap-dataflow/crates/pdata/src/lib.rs
@@ -38,9 +38,6 @@ mod validation;
 pub use decode::decoder::Consumer;
 pub use encode::producer::Producer;
 
-/// Re-export prost for proto message encoding/decoding.
-pub use prost;
-
 /// TraceID identifier of a Trace
 #[derive(Eq, PartialEq, Clone, Copy, Debug, Default)]
 pub struct TraceID([u8; 16]);
--- a/rust/otap-dataflow/crates/pdata/src/otap/groups.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/groups.rs
@@ -1513,7 +1513,7 @@ impl UnifiedDictionaryTypeSelector {
         Self {
             total_batch_size: 0,
             values_arrays: Vec::new(),
-            smallest_key_type: DataType::UInt8, // Start with smallest, upgrade as needed
+            smallest_key_type: DataType::UInt16,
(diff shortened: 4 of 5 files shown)


rust/otap-dataflow/configs/fake-batch-debug-noop.yaml
```diff
@@ -51,3 +51,12 @@ nodes:
     kind: exporter
     plugin_urn: "urn:otel:noop:exporter"
     config:
+
+service:
+  telemetry:
```
andborja, line 56: This is more for the existing SDK configuration, in alignment with the otel collector style.
I would configure it in /telemetry/logs instead if required.

rust/otap-dataflow/configs/fake-batch-debug-noop.yaml
```diff
@@ -51,3 +51,12 @@ nodes:
...
+    logs:
+      level: "debug"
+      providers:
+        global: immediate
+        engine: immediate
+      output: direct
```
andborja, line 62: What does this mean?

rust/otap-dataflow/configs/internal-telemetry.yaml
```diff
@@ -0,0 +1,79 @@
...
+    config:
+      verbosity: basic
+  noop:
+    kind: exporter
+    plugin_urn: "urn:otel:noop:exporter"
+    config: {}
+
+# Internal telemetry pipeline - separate from main pipeline
+# Uses hardcoded settings: single thread, no admin server
+internal:
```
andborja, line 39: Suggestion: We can make it multi-graph with the same effort and risk.
Something like:
```yaml
graphs: # array of graphs
   - internal: # name of the graph
     # other graph level properties, including the logger type
     nodes: # copied from the original graph structure
       telemetry:  # same block that is configured here
        ...
```
We can keep the existing configuration of single graph as is. This would be more an extension of the configuration.

rust/otap-dataflow/configs/internal-telemetry.yaml
```diff
@@ -0,0 +1,79 @@
...
+    config: {}
+
+service:
+  telemetry:
+    logs:
+      level: "debug"
+      providers:
+        global: immediate
+        engine: immediate
+        internal: raw  # Avoid feedback in internal pipeline
```
andborja, line 75: This should be configured at the "internal" block, not really here

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -68,6 +76,184 @@ pub enum PipelineType {
     /// OpenTelemetry with Apache Arrow Protocol (OTAP) pipeline.
     Otap,
 }
+
+/// A collection of nodes forming a pipeline graph (hyper-DAG). One of
+/// these is the main pipeline, and one is the internal telemetry pipeline.
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Default)]
+#[serde(transparent)]
+pub struct PipelineNodes(HashMap<NodeId, Arc<NodeUserConfig>>);
```
andborja, line 84: nit: We can do something more "structured" using named attributes instead of a tuple

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -379,54 +512,137 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
...
+        pipeline_ctrl_msg_rx: PipelineCtrlMsgReceiver<PData>,
+        startup_tx: std_mpsc::Sender<Result<(), Error>>,
+    ) -> Result<Vec<()>, Error> {
+        // Run with the appropriate tracing subscriber for this pipeline.
+        telemetry_setup.with_subscriber(log_level, || {
+            // Create a tracing span for this pipeline thread
+            let span = otel_info_span!("internal_pipeline_thread", core.id = core_id.id);
+            let _guard = span.enter();
+
+            // No core pinning for internal pipeline - it's lightweight
```
lquerel, line 601: I propose changing this comment and turning it into a TODO. In the long run, we will also introduce thread pinning for this internal pipeline, since an instance of our Internal Telemetry System will be deployed on each NUMA region. We will therefore need one thread per internal telemetry pipeline for each NUMA node, with thread pinning to one of the cores in the corresponding region.

jmacd:
Closing in favor of #1841 which is the final 1/2 of this large PR.
The second half of this large PR would be the ITR, the console exporter, and setup for the "ITS" internal telemetry system.

PR #1772, opened 2026-01-12
Split MetricsRegistry into an EntityRegistry and a MetricSetRegistry

lquerel:
This PR is primarily a mechanical refactoring of the `MetricsRegistry`, with the goal of clearly separating the `TelemetryRegistry` general concept from the `EntityRegistry` and `MetricSetRegistry` sub-registries (please check the updated architecture diagram below).

The intent is to enable reuse of entity-based attributes across metric sets and event-based instrumentation.

With this change, events can be reported using an attribute set ID that represents the stable attributes of the emitting entity. This does not imply support for dynamic attributes at this level. Dynamic attributes will continue to be reported using the approach Joshua is currently implementing, namely an OTLP-bytes representation for dynamic attributes.

image

This PR implements the bottom left part of the diagram. The bottom right part is provided for context only, the ITRs and the internal pipeline are currently being implemented by Joshua.

### DIFF
--- a/rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
+++ b/rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
@@ -339,6 +339,10 @@ impl SourceScalarExpression {
         &mut self.accessor
     }
 
+    pub fn into_value_accessor(self) -> ValueAccessor {
+        self.accessor
+    }
+
     pub fn get_value_type(&self) -> Option<ValueType> {
         self.value_type.clone()
     }
--- a/rust/experimental/query_engine/expressions/src/value_accessor.rs
+++ b/rust/experimental/query_engine/expressions/src/value_accessor.rs
@@ -40,6 +40,10 @@ impl ValueAccessor {
         &self.selectors
     }
 
+    pub fn into_selectors(self) -> Vec<ScalarExpression> {
+        self.selectors
+    }
+
     pub fn get_selectors_mut(&mut self) -> &mut [ScalarExpression] {
         &mut self.selectors
     }
--- a/rust/otap-dataflow/crates/opl/src/opl.pest
+++ b/rust/otap-dataflow/crates/opl/src/opl.pest
@@ -49,13 +49,13 @@ index_expression = {
 }
 
 attribute_selection_expression = {
-    identifier_expression ~ ("." ~ identifier_expression)+
+    identifier_expression ~ ("." ~ member_expression)+
 }
 
 member_expression = {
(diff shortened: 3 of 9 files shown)


jmacd (approved):

rust/otap-dataflow/crates/telemetry/src/entity.rs
```diff
@@ -0,0 +1,389 @@
...
+            descriptor,
+            values,
+            sorted_indices: indices.into(),
+        }
+    }
+}
+
+impl PartialEq for EntityAttributeSet {
+    fn eq(&self, other: &Self) -> bool {
+        // Defensive: in release builds, return false on length mismatch to avoid OOB access.
```
albertlockett, line 68: Why does this comment specify "in release builds"? Wouldn't this return false in all builds when the lengths don't match?

rust/otap-dataflow/crates/telemetry/src/entity.rs
```diff
@@ -0,0 +1,389 @@
...
+
+        self.sorted_indices.len().hash(state);
+        for idx in self.sorted_indices.iter() {
+            let field = &self.descriptor.fields[*idx];
+            field.key.hash(state);
+            attribute_value_type_rank(field.r#type).hash(state);
+            let value = &self.values[*idx];
+            hash_attribute_value(value, state);
+        }
+    }
```
albertlockett, line 115: I'm not sure I understand the hash implementation here. We hash the length of values & descriptor, and if they don't match, we don't hash any of the fields?

jmacd: (Good question.)

rust/otap-dataflow/crates/telemetry/src/entity.rs
lquerel: It's mostly because the invariant is checked in the constructor with a debug assert, but it's technically possible that someone builds in release mode without observing this invariant. That's why I have this defensive code, even though most of the time the corresponding potential issue will be caught during the development of a new attribute set implementation.

rust/otap-dataflow/crates/telemetry/src/entity.rs
lquerel: The lenght are hashed first. If they mismatch we return early to avoid OOB indexiing lines 109 and 112, and PartialEq will return false for mismatched lengths anyway. Collisions are always possible with any hash, even if 2 mismatched lenghts collided, so it's fine because Eq still rejects them , which is all Hashmap required :-)

PR #1778, opened 2026-01-13
OPL Parser support `if`/`else` statement

albertlockett:
closes: #1777 
relates to: #1667 

This is followup from #1722 , where we decided to not implement this feature in KQL parser and instead have OPL implement it's own parser.

Adds parsing support for an if/else if/else expression that gets parsed into the ConditionalDataExpression that was added in https://github.com/open-telemetry/otel-arrow/pull/1684

The syntax looks like this.
```
logs |
if (severity_text == "ERROR") { 
    extend attributes["important"] = "very" | extend attributes["triggers_alarm"] = "true"
} else if (severity_text == "WARN") {
    extend attributes["important"] = "somewhat"
} else if (severity_text == "INFO") {
    extend attributes["important"] = "rarely"
} else {
    extend attributes["important"] = "no"
}
```

`else if` and `else` are optional, so the following expressions are also supported:
```
logs | 
if (severity_text == "ERROR") { 
    extend attributes["is_error"] = true
} else {
    extend attributes["is_not_error"] = true
}
```
```
logs | 
if (severity_text == "ERROR") { 
    extend attributes["triggers_alarm"] = "true"
}
```

In a future PR, I'll go back and use `OplParser` in columnar query engine's [`pipeline/conditional.rs` unit tests](https://github.com/open-telemetry/otel-arrow/blob/cf43a4c0ddc170aa1b8ec243f8bb28f0bce591ba/rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs#L227-L237). This can happen once [#1762](https://github.com/open-telemetry/otel-arrow/issues/1762) is also complete, as those tests use `project-rename`, so support for this must also be added to OPL Parser.

PR #1779, opened 2026-01-13
OPL Parser: add support for `project-rename` operator

### LINKED ISSUE #1762: OPL Parser: rename map keys expression
We should support syntax for renaming map keys. rename and project-rename will be synonms
```kql
rename attributes["x"] = attributes["y"]
project-rename attributes["x"] = attributes["y"]
```

This should get parsed into `DataExpression::Transform(TransformExpression::RenameMapKeys(..))` from our expressions AST

albertlockett:
Closes #1762 

Add support for the `project-rename` operator which can rename attributes:
```kql
project-rename attributes["x"] = attributes["y"]
project-rename resource.attributes["x"] = resource.attributes["y"]
```

`rename` is an alias for `project-rename` to aid users who are not familiar with this KQL inspired operator.

The `parse_assignment_expression` is refactored to return `(SourceScalarExpression, ScalarExpression)` instead of `SetTransformExpression` to make it easier to reuse the result of this function for both `extend` and `project-rename` operator calls (KQL parser does [something similar](https://github.com/open-telemetry/otel-arrow/blob/b3214230ca5d394d11891414c2021fc4c72f8225/rust/experimental/query_engine/kql-parser/src/shared_expressions.rs#L37-L40)).

This PR also updates the test in columnar query engine to use the `OplParser` for some test cases of `pipeline::attributes::AttributeTransformPipelineStage` and `pipeline::conditional::ConditionalPipelineStage` for tests using supported syntax.

Finally, there was some PR feedback on https://github.com/open-telemetry/otel-arrow/pull/1778 that was added after the PR merged, so this PR includes those small changes.

### DIFF
--- a/rust/otap-dataflow/crates/opl/src/parser/operator.rs
+++ b/rust/otap-dataflow/crates/opl/src/parser/operator.rs
@@ -97,7 +97,7 @@ pub(crate) fn parse_rename_operator_call(
                 return Err(ParserError::SyntaxNotSupported(
                     query_location,
                     format!(
-                        "rename operator call only supports assignment from source. Found {other:?}"
+                        "rename operator only supports assignment from source. Found {other:?}"
                     ),
                 ));
             }
@@ -112,7 +112,6 @@ pub(crate) fn parse_rename_operator_call(
         )),
         keys,
     ));
-
     pipeline_builder.push_data_expression(DataExpression::Transform(transform_expr));
 
     Ok(())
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/opl/src/parser/operator.rs
```diff
@@ -37,15 +40,59 @@ pub(crate) fn parse_operator_call(
...
+        }
+    }
+
+    let transform_expr = TransformExpression::RenameMapKeys(RenameMapKeysTransformExpression::new(
+        query_location.clone(),
+        MutableValueExpression::Source(SourceScalarExpression::new(
+            query_location,
+            ValueAccessor::new(),
+        )),
+        keys,
```
gouslu, line 113: what if "keys" is empty here, is it better to return Ok(()) early?

albertlockett: Thanks for the suggestion @gouslu! I'm not sure it's necessary to have the early return here. Currently, the grammar we've defined wouldn't actually allow for keys to be empty. There always needs to be at least one assignment expression (which gets mapped into this `keys` array:
https://github.com/open-telemetry/otel-arrow/blob/4a732793fb2f7d043749d3c125a96586dd72f1fb/rust/otap-dataflow/crates/opl/src/opl.pest#L159-L161

lalitb:
LGTM. nicely done. thanks for fixing the earlier PR comments.

PR #1780, opened 2026-01-13
PerfTest - test with varying input batch size

cijothomas:
Running the 100K LRPS tests, with varying batch_size 1000,5000, 10,000
The batch size here is referring to the batch size from load-generator. We haven't incorporated BatchProcessor yet to the tests

### DIFF
--- a/docs/tls-architecture.md
+++ b/docs/tls-architecture.md
@@ -1,0 +1,706 @@
+# TLS/mTLS Architecture Documentation
+
+This document describes the high-level architecture and design principles
+of TLS/mTLS support for OTLP/OTAP receivers and exporters.
+
+## Table of Contents
+
+- [Overview](#overview)
+- [Design Principles](#design-principles)
+- [Receiver Architecture](#receiver-architecture)
+- [Exporter Architecture](#exporter-architecture)
+- [Certificate Hot-Reload](#certificate-hot-reload)
+- [Performance Characteristics](#performance-characteristics)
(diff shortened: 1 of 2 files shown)


.github/workflows/pipeline-perf-test-nightly.yml
```diff
@@ -4,7 +4,8 @@ permissions:
 
 on:
     schedule:
-      - cron: '0 10 * * *'  # runs every day at 10 AM UTC
+      - cron: '0 16 * * *'  # runs every day at 8 AM PT (4 PM UTC)
```
cijothomas, line 7: moving to run twice a day, so in case things don't look as expected, I can fix it without waiting one full day.

Can change to once a day in future.

tools/pipeline_perf_test/test_suites/integration/continuous/100klrps-docker.yaml
```diff
@@ -81,24 +81,28 @@ tests:
         engine_config_template: test_suites/integration/templates/configs/engine/continuous/otlp-attr-otlp.yaml
         loadgen_exporter_type: otlp
         backend_receiver_type: otlp
+        max_batch_size: 1000
```
cijothomas, line 84: this is not changing anything - the default was 1000 anyway, this just makes it obvious.
Separately, I think the regular nightly tests should default to 512 as that is the default Log batch coming from OTel SDK's/OTLP exporters.

PR #1781, opened 2026-01-13
Update benchmarks doc to list all current tests

cijothomas:
Fixed one TODO!

https://github.com/open-telemetry/otel-arrow/pull/1528 - Still working on this separately, which will include actual numbers for key scenarios, so readers don't have to go through the graphs themselves!

### DIFF
--- a/.github/workflows/pipeline-perf-test-nightly.yml
+++ b/.github/workflows/pipeline-perf-test-nightly.yml
@@ -4,7 +4,8 @@ permissions:
 
 on:
     schedule:
-      - cron: '0 10 * * *'  # runs every day at 10 AM UTC
+      - cron: '0 16 * * *'  # runs every day at 8 AM PT (4 PM UTC)
+      - cron: '0 0 * * *'   # runs every day at 4 PM PT (midnight UTC)
     workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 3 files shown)


docs/benchmarks.md
```diff
@@ -17,9 +17,83 @@ We run two types of automated benchmark tests for Phase 2:
...
+### Metrics Collected
+
+All benchmark tests measure the following metrics:
+
+- **Logs/sec input** - Input throughput
+- **Logs/sec output** - Output throughput
+- **RAM** - Average and maximum memory usage
+- **Normalized CPU** - Average and maximum CPU usage normalized to a 0-100%
+  range relative to the available cores in the test configuration (e.g., 80% in
+  a 4-core test means 80% of 4 cores)
```
lalitb: This phrasing feels ambiguous, maybe it's just me, but could we spell out the math? Eg - 80% in a 4‑core test means roughly 3.2 cores worth of CPU time (0.8 × 4), so 100% equals all cores saturated.

lalitb (approved):
LGTM.

docs/benchmarks.md
cijothomas: rephrased. LMK if it is better now.

lalitb: Thanks.

PR #1785, opened 2026-01-14
Create entity and expose entity keys via thread locals and task locals

lquerel:
The engine now creates the following entities:

* **Pipeline** -> Stored in a thread local associated with the pipeline thread.
* **Node** -> Stored in the task local of the node.
* **Channel**
  * **Sender entity** stored in the task local of the sender node.
  * **Receiver entity** stored in the task local of the receiver node.

An entity cleanup mechanism is in place. A unit test has been added to validate this cleanup process.

The final goal is to be able to use these entities directly when reporting metric sets and events. This allows us to report the attributes of all our entities using a simple numerical ID.

Closes https://github.com/open-telemetry/otel-arrow/issues/1791

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/exporter.rs
+++ b/rust/otap-dataflow/crates/engine/src/exporter.rs
@@ -198,9 +198,6 @@ impl<PData> ExporterWrapper<PData> {
         channel_metrics: &mut ChannelMetricsRegistry,
         channel_metrics_enabled: bool,
     ) -> Self {
-        if !channel_metrics_enabled {
-            return self;
-        }
         match self {
             ExporterWrapper::Local {
                 node_id,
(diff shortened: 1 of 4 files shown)


utpilla (approved):

rust/otap-dataflow/crates/engine/src/receiver.rs
```diff
@@ -155,29 +220,52 @@ impl<PData> ReceiverWrapper<PData> {
...
-                        channel_id,
-                        CHANNEL_KIND_CONTROL,
-                        runtime_config.control_channel.capacity as u64,
-                    ),
-                    Err(receiver) => receiver,
+                let control_sender = control_sender.into_mpsc();
+                let control_receiver = control_receiver.into_mpsc();
+                let (control_sender, control_receiver) = match (control_sender, control_receiver) {
+                    (Ok(sender), Ok(receiver)) => {
+                        let channel_entity_key = pipeline_ctx.register_channel_entity(
```
lalitb, line 227: The control channel registration is only happening when then channel_metrics_enabled is enabled, because of short circuit at line 210. Is this intended? Just ruling out the case where we need control channel for other telemetry and it's not yet registered.

rust/otap-dataflow/crates/engine/src/exporter.rs
```diff
@@ -149,32 +210,73 @@ impl<PData> ExporterWrapper<PData> {
...
+                            LocalSender::mpsc_with_metrics(sender, channel_metrics, sender_metrics),
+                            Some(LocalReceiver::mpsc_with_metrics(
+                                receiver,
+                                channel_metrics,
+                                receiver_metrics,
+                                runtime_config.control_channel.capacity as u64,
+                            )),
+                        )
+                    }
+                    (Ok(sender), None) => {
```
jmacd: What does it mean that there is a None for control_receiver?

rust/otap-dataflow/crates/engine/src/entity_context.rs
```diff
@@ -0,0 +1,410 @@
...
+#[must_use]
+pub fn pipeline_entity_key() -> Option<EntityKey> {
+    PIPELINE_ENTITY_KEY.with(Cell::get)
+}
+
+tokio::task_local! {
+    /// Task-local context for nodes, including entity keys and telemetry handle.
+    /// This task local can be used by our internal telemetry instrumentation to associate metrics
+    /// and events with the correct node entity or the correct input/output channels.
+    static NODE_TASK_CONTEXT: NodeTaskContext;
```
jmacd, line 57: 👍

rust/otap-dataflow/crates/engine/src/exporter.rs
lquerel: I thought we had tests that did not specify the control receiver. It turns out we did not, so I took the opportunity to remove the option from the control receiver and was able to delete that branch as a result. Good catch.

rust/otap-dataflow/crates/engine/src/receiver.rs
lquerel: It is indeed debatable and a bit particular once events start using these entities. We will want to have them available even if metrics are disabled. I will make the necessary changes. Good catch.

Issue #1784, opened 2026-01-14
Columnar Query Engine `route to` operator


albertlockett:
We'd like to have a way to specify in the query engine / transform processor to send the output of some pipeline to some next destination.

For example, imagine an OTAP DF engine like:
```
OTLP receiver -> Transform Processor ---> Exporter1 (for important telemetry)
                                     |--> Exporter2 (for less important telemetry)
```

And a program like:
```kql
logs | if (severity_number > 10) {
  route_to "exporter1_out_port"
} else {
  route_to "exporter2_out_port"
}
```

Note that:
1. It would be nice if the columnar query engine remained decoupled from the OTAP DF Engine pipeline machinery. This would make it possible to use the columnar query-engine in a different system (like the SDK for example). This means whatever does the routing should be generic enough that it doesn't necessarily depend on something like `EffectHandler`.
2. In the program above, the batch gets split, which for correct ack/nack semantics means we need to create new outgoing contexts, and ack the incomining context in response to Ack/Nacks received for the outgoing contexts (similar to how Batch Processor works).
3. Transform Processor currently only accepts KQL program, but we'll only be supporting route_to in OPL initially, so we need to enhance the processor config to receive OPL programs and select the correct parser accordingly when deserializing the config.

To do for this issue:
- [ ] Add route_to operator to OPL Parser
- [ ] Add expressions for this operation to expressions AST
- [ ] Implement columnar query engine pipeline stage for this routing operation
- [ ] Add capability to transform_processor to be configured by OPL program
- [ ] Add tests to transform processor to ensure batches are correctly routed (via calls to effect handler)
- [ ] Handle proper Ack/Nack behaviour in transform processor.

PR #1786, opened 2026-01-14
feat: initial implementation of `route to` pipeline stage

albertlockett:
Related to #1784

Adds an operator the columnar query engine that can be used to route an OTAP batch to some destination. The main use case is to have the transform processor capable of sending telemetry batches to different out ports, where the behaviour is defined by the query it is executing.

- A new `PipelineStage` is implemented called `RouterPipelineStage`
- A new data expression type is added to our AST called the `OutputDataExpression`
- A new operator is added to OPL Parser that is parsed into the new data expression variant.

Example:
```kql
logs
| if (severity_text == "ERROR") {
    route_to "error_port"
} else if (severity_text == "INFO") {
    route_to "info_port"
} // else - route to default out_port
```

### Some additional notes on the design:

**Routing implementation is pluggable:**
Although the main use case is to direct the batches to some out port, I didn't want to couple the implementation of the columnar query engine to the DF pipeline. This means I didn't want code in the query-engine crate referencing things that handle pdata routing like `EffectHandler` or `message::Sender` from the engine crate. 

In general, I'm imagine use cases where pipelines driven by OPL could be executed in a variety of contexts, that may need to route data to a variety of destinations.

To make the router routing behaviour customizable, the `pipeline::router` module exposes a `Router` trait which users of columnar query-engine can implement.

**Extensions & Execution State:**
`RouterPipelineStage` will need to be able to find the implementation of `Router`. This PR adds the concept of `ExecutionState` and "extensions", which are a map of instances of types that pipeline stages may need during their execution.

The benefit of this "extension" pattern is that it helps improve future extensibility. For example, we could imagine users may eventually implement custom `PipelineStages`, which have external dependencies that need to be injected at runtime. Having these "extension"s available makes this possible.

The concept of "extensions" is similar to datafusion's [`SessionConfig` extensions](https://docs.rs/datafusion/latest/datafusion/prelude/struct.SessionConfig.html#method.with_extension), but having our own implementation provides us with some benefits: our pipeline stages execute in a single threaded runtime, so extension's types don't need to be `Send` + `Sync` and can be accessed mutably.

The `ExecutionState` as a concept also has some auxiliary benefits beyond simply being a repository of extensions. In the future, there may be other mutable state that needs to be updated by pipeline stages such as metrics or state related to stream processing. Introducing this type now is the foundation for these future features.

### Followups:
- Ack/Nack will be handled in a followup PR. Since this kind of conditional routing splits the batch, we need to juggle incoming/outgoing contexts (much like the batch processor).
- `RouteToPipelineStage` emits an empty batch after the incoming batch is sent elsewhere.  It's forced to do this by the trait signature of `PipelineStage`. This is OK for now, but in the future we probably want to introduce the concept of a "terminal pipeline stage" as a special type of pipeline stage consumes the batch.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/transform_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/transform_processor.rs
@@ -245,6 +245,10 @@ impl Processor<OtapPdata> for TransformProcessor {
                             otap_batch.into()
                         }
                         Err(e) => {
+                            // forward the routed messages in the event of a failure to avoid
+                            // caching any batches that were routed before the pipeline fails
+                            self.handle_routed_messages(effect_handler).await?;
+
                             self.metrics.msgs_transform_failed.inc();
                             return Err(EngineError::ProcessorError {
                                 processor: effect_handler.processor_id(),


rust/experimental/query_engine/expressions/src/data_expressions.rs
```diff
@@ -295,3 +302,89 @@ impl ConditionalDataExpressionBranch {
...
+}
+
+/// Expression representing an operation that emits data to a sink.
+#[derive(Debug, Clone, PartialEq)]
+pub enum OutputExpression {
+    /// Output data to a sink identified by name.
+    // Currently this contains a static string because it's the only way we handle identifying
+    // where to output the data. In the future we could support dynamic sink identified by a
+    // variable, result of a function call, or other some expression, at which point we can change
+    // this to contain the more general `StaticExpression`.
```
lquerel:
```suggestion
    /// Output data to a sink identified by name.
    /// Currently this contains a static string because it's the only way we handle identifying
    /// where to output the data. In the future we could support dynamic sink identified by a
    /// variable, result of a function call, or other some expression, at which point we can change
    /// this to contain the more general `StaticExpression`.
```

rust/otap-dataflow/crates/otap/src/transform_processor.rs
```diff
@@ -100,20 +107,27 @@ impl TransformProcessor {
...
+            error: format!("Could not parse TransformProcessor query: {e:?}"),
+        })?
+        .pipeline;
 
         // TODO: it would be nice if we could validate that the pipeline expr is supported by the
         // query engine here. Currently, validation happens lazily when the first batch is seen.
         // https://github.com/open-telemetry/otel-arrow/issues/1634
 
+        let mut execution_state = ExecutionState::new();
+        execution_state.set_extension::<RouterExtType>(Box::new(RouterImpl::new()));
```
lquerel, line 124: I like the concept.

rust/otap-dataflow/crates/otap/src/transform_processor.rs
```diff
@@ -124,6 +138,46 @@ impl TransformProcessor {
...
+                .connected_ports()
+                .iter()
+                .find(|p| p.as_ref() == route_name.as_str())
+                .ok_or_else(|| EngineError::ProcessorError {
+                    processor: effect_handler.processor_id(),
+                    kind: ProcessorErrorKind::Transport,
+                    error: "Routing error: ".into(),
+                    source_detail: format!("out_port name {} not configured", route_name),
+                })?
+                .clone();
```
lquerel, line 174: Why not using the method `send_message_to` and checking the error ?

rust/otap-dataflow/crates/query-engine/src/pipeline/state.rs
```diff
@@ -0,0 +1,180 @@
...
+/// This is similar to datafusion's `ExecutionState`, which it also uses for extensions, but
+/// without the need for `Send + Sync` bounds, as those are not required in this context due to these
+/// pipeline stages executing in a single threaded runtime. This also means that pipeline
+// stages can get mutable references to extensions if needed.
+//
+// In the future, this may be expanded to include other execution-related state like metrics,or
+// other state needed for stateful stream processing.
+//
+// When the pipeline is executed without an ExecutionState being provided, a default one will be
+// created. This means that anything added to this in the should be inexpensive to initialize in
```
lquerel: "...added to this in the <missing word> should be..."

lquerel (approved):
I made few non-blocking suggestions.

rust/otap-dataflow/crates/otap/src/transform_processor.rs
```diff
@@ -180,9 +234,14 @@ impl Processor<OtapPdata> for TransformProcessor {
...
                     otap_batch.decode_transport_optimized_ids()?;
-                    match self.pipeline.execute(otap_batch).await {
+                    match self
+                        .pipeline
+                        .execute_with_state(otap_batch, &mut self.execution_state)
+                        .await
+                    {
                         Ok(otap_batch) => {
                             self.metrics.msgs_transformed.inc();
+                            self.handle_routed_messages(effect_handler).await?;
```
lalitb, line 244: This will drain/send the routed batches only when the pipeline execution succeed. What if the pipeline fails - these routed batches will remain cached, and leak into next successful run ? Should we be clearing the router in Err branch below (line 247) ?

rust/otap-dataflow/crates/otap/src/transform_processor.rs
albertlockett: `send_message_to` takes as an arg `P: Into<PortName>` where `type PortName = Cow<'static, &str>`
https://github.com/open-telemetry/otel-arrow/blob/4e7f6a95ab06ee873b78bdba726d5f7b1d5ff7f5/rust/otap-dataflow/crates/engine/src/local/processor.rs#L175-L179

https://github.com/open-telemetry/otel-arrow/blob/4e7f6a95ab06ee873b78bdba726d5f7b1d5ff7f5/rust/otap-dataflow/crates/config/src/lib.rs#L64-L65

But the `route_name` here is an `Rc<String>` that might not have the `static lifetime, so I think to make this into a PortName I'd need to copy it, which I was trying to avoid

rust/otap-dataflow/crates/otap/src/transform_processor.rs
albertlockett: Thanks @lalitb. Yes you're right, I think it's reasonable to try to handle the routed batches in the error branch as well. Made this change in [16446e4](https://github.com/open-telemetry/otel-arrow/pull/1786/commits/16446e42bd6c26ed54f4d34be798a2f6fe3bb6b4)

rust/otap-dataflow/crates/query-engine/src/pipeline/state.rs
albertlockett: fixed, thanks for catching that

rust/experimental/query_engine/expressions/src/data_expressions.rs
```diff
@@ -295,3 +302,89 @@ impl ConditionalDataExpressionBranch {
...
+
+/// Expression representing an operation that emits data to a sink.
+#[derive(Debug, Clone, PartialEq)]
+pub enum OutputExpression {
+    /// Output data to a sink identified by name.
+    /// Currently this contains a static string because it's the only way we handle identifying
+    /// where to output the data. In the future we could support dynamic sink identified by a
+    /// variable, result of a function call, or other some expression, at which point we can change
+    /// this to contain the more general `StaticExpression`.
+    NamedSink(StringScalarExpression),
```
CodeBlanch, line 359: @albertlockett I just noticed this. Sorry for the late feedback but here it is...

1) "Sink" What is a sink in the context of AST? Maybe "NamedDestination" would be better. Remember it is an explicit goal this AST\code NOT be coupled to the collector.

2) Current design seems very limited. As far as routing is concerned, I can see different needs. Fan out\send\split, stuff like that.

   Consider something like this:

   ```rust
   enum RouteDataExpression {
     MapTo(RouteDestinationExpression), // Send the data and quit
     SplitTo(RouteDestinationExpression) // Send a copy of the data and continue on
   }

   enum RouteDestinationExpression {
      Named(StringScalarExpression)
   }
   ```

albertlockett: re point 1: imo "sink" is kind of a synonym for to "destination". I don't see how one is any more coupled to the collector than the other, but I'm fine to change it to destination if you think it's a better name. I'm not personally hung up on the naming one way or the other.

re point 2: the intention of the output expression was simply that the data would be emitted to some destination, and that the variants of the enum would control how the destination is defined. We could change the design to have the expression variants control how data is routed in the pipeline, but that could also be achieved through other types of data expressions. For example, we could imagine a fork with multiple branches as child expressions, which could achieve the "copy the data and continue on" scenario in your example.

CodeBlanch: Sounds like we need a design discussion 😄

albertlockett: Yeah that'd be great! We can discuss during wednesday's SIG meeting

PR #1794, opened 2026-01-15
Internal telemetry pipeline nodes config

jmacd:
Part of #1771.

Part of #1736.

Follows #1741.

This moves the HashMap<_, _> of nodes into a struct `PipelineNodes` and re-uses it to parse an identical graph of `internal` nodes. This internal graph will be use when an internal logging provider is configured to output to an internal pipeline.

### DIFF
--- a/rust/otap-dataflow/crates/config/src/pipeline.rs
+++ b/rust/otap-dataflow/crates/config/src/pipeline.rs
@@ -43,9 +43,6 @@ pub struct PipelineConfig {
     settings: PipelineSettings,
 
     /// All nodes in this pipeline, keyed by node ID.
-    ///
-    /// Note: We use `Arc<NodeUserConfig>` to allow sharing the same pipeline configuration
-    /// across multiple cores/threads without cloning the entire configuration.
     #[serde(default)]
     nodes: PipelineNodes,
 
@@ -77,6 +74,9 @@ pub enum PipelineType {
 }
 
 /// A collection of nodes forming a pipeline graph.
+///
+/// Note: We use `Arc<NodeUserConfig>` to allow sharing the same pipeline configuration
+/// across multiple cores/threads without cloning the entire configuration.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Default)]
 #[serde(transparent)]
 pub struct PipelineNodes(HashMap<NodeId, Arc<NodeUserConfig>>);
@@ -504,8 +504,14 @@ impl PipelineConfig {
 
         // Validate internal pipeline if present
         if !self.internal.is_empty() {
+            // TODO: the location of the internal telemetry pipeline
+            // nodes is subject to change. Temporarily, we append
+            // ("_internal") to the pipeline_id. We need a way to
+            // refer to the set of node defining the internal
+            // pipeline.
+            let internal_id: PipelineId = format!("{}_internal", &pipeline_id).into();
             self.internal
-                .validate(pipeline_group_id, pipeline_id, &mut errors);
+                .validate(pipeline_group_id, &internal_id, &mut errors);
         }
 
         if !errors.is_empty() {


rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -68,6 +75,183 @@ pub enum PipelineType {
     /// OpenTelemetry with Apache Arrow Protocol (OTAP) pipeline.
     Otap,
 }
+
+/// A collection of nodes forming a pipeline graph.
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, Default)]
+#[serde(transparent)]
+pub struct PipelineNodes(HashMap<NodeId, Arc<NodeUserConfig>>);
```
andborja, line 82: Same comment as in the other PR:
nit: We can do something more "structured" using named attributes instead of a tuple

rust/otap-dataflow/crates/config/tests/fixtures/test_pipeline.yaml
```diff
@@ -18,6 +18,31 @@ nodes:
     plugin_urn: "urn:test:exporter"
     config: null
     out_ports: {}
+
+# Note the internal and nodes graphs are separate.
+# They do not share a namespace.
+internal:
```
andborja, line 24: Also copied from the other PR:

Suggestion: We can make it multi-graph with the same effort and risk.
Something like:
```yaml
graphs: # array of graphs
   - internal: # name of the graph
     # other graph level properties, including the logger type
     nodes: # copied from the original graph structure
       telemetry:  # same block that is configured here
        ...
```
We can keep the existing configuration of single graph as is. This would be more an extension of the configuration.

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -46,7 +46,14 @@ pub struct PipelineConfig {
     ///
     /// Note: We use `Arc<NodeUserConfig>` to allow sharing the same pipeline configuration
     /// across multiple cores/threads without cloning the entire configuration.
-    nodes: HashMap<NodeId, Arc<NodeUserConfig>>,
+    #[serde(default)]
+    nodes: PipelineNodes,
```
utpilla, line 47: Let's update the doc comments on the lines above to not mention `Arc<NodeUserConfig>` now. We should mention that for `pub struct PipelineNodes(HashMap<NodeId, Arc<NodeUserConfig>>);`

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -46,7 +46,14 @@ pub struct PipelineConfig {
...
+
+    /// Internal telemetry pipeline nodes. These have the same structure
+    /// as `nodes` but are independent and isolated to a separate internal
+    /// telemetry runtime.
+    #[serde(default, skip_serializing_if = "PipelineNodes::is_empty")]
+    internal: PipelineNodes,
```
lquerel, line 53: I have an issue with this approach. Even though, as things stand, our engine only supports a single pipeline configuration, it is designed to support multiple pipelines (with different config) running in parallel within the same process. In the configuration crate, we have this logical grouping called `PipelineGroup`, whose long term goal is to group multiple pipelines that belong to the same group, for example a group could be mapped to a tenant.

All of this is to say that we will have multiple pipeline configurations running at the same time, and therefore a single internal telemetry system (ITS) receiving internal telemetry from different pipeline configurations. Because of that, it does not seem ideal to start tying the ITS configuration to a specific pipeline configuration. In the long run, I agree that we will need a way to define pipeline specific processing for our internal telemetry, but I propose that we address this in a second phase.

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -278,55 +468,44 @@ impl PipelineConfig {
...
-            for cycle in cycles {
-                errors.push(Error::CycleDetected {
-                    context: Context::new(pipeline_group_id.clone(), pipeline_id.clone()),
-                    nodes: cycle,
-                });
-            }
+        // Validate internal pipeline if present
+        if !self.internal.is_empty() {
+            self.internal
+                .validate(pipeline_group_id, pipeline_id, &mut errors);
```
utpilla: This could lead to some inconvenience in some cases of validation errors. For example, if the main pipeline and the internal pipeline have nodes with the same name and there is a validation error, it wouldn't be clear if the error belongs to the main pipeline section or the internal pipeline section.

Maybe we can add a `bool` field `is_internal` to `Context` or just append "internal" to the `pipeline_id` when validating internal pipeline?

rust/otap-dataflow/crates/config/src/pipeline.rs
jmacd: I tried to make this type transparent to serde, which works (IIUC) for tuples.

rust/otap-dataflow/crates/config/tests/fixtures/test_pipeline.yaml
jmacd: Agree. I will defer to Laurent who has in mind a use-case for pipeline groups.

rust/otap-dataflow/crates/config/src/pipeline.rs
jmacd: Done

rust/otap-dataflow/crates/config/src/pipeline.rs
jmacd: I have no strong opinion on how we organize the graph as far as where the internal telemetry definition belongs.

rust/otap-dataflow/crates/config/src/pipeline.rs
jmacd: I appended _internal.

PR #1795, opened 2026-01-15
Add service::telemetry::logs::providers settings for internal logging setup

jmacd:

As documented in https://github.com/open-telemetry/otel-arrow/pull/1741.

~Updates that document to match this change reflecting the prototype in #1771.~

Revised relative to #1771.

Adds LoggingProviders (choice of default logging provider for global, engine, and internal-telemetry threads).
Adds ProviderMode with names to select instrumentation behavior, with `its` referring to internal telemetry system.

Note: These settings are somehow not ideally placed. They belong also in the top-level settings, or with observed_state settings. However, since logging is configured with resource and level, which are part of the service::telemetry config area presently, we use that structure. After the bulk of #1736 is finished we can restructure.

### DIFF
--- a/rust/otap-dataflow/docs/self_tracing_architecture.md
+++ b/rust/otap-dataflow/docs/self_tracing_architecture.md
@@ -56,7 +56,7 @@ telemetry pitfalls, as follows:
 - Non-blocking interfaces. We prefer to drop and count dropped
   internal log events than to block the pipeline.
 - Option to configure internal telemetry multiple ways, including the
-  no-op implementation, direct console output, and global or regional 
+  no-op implementation, direct console output, and global or regional
   logs consumers.
 
 ## OTLP-bytes first
@@ -107,7 +107,7 @@ Provider mode values are:
 - Noop: Ignore logging.
 - ITS: Use the internal telemetry system.
 - OpenTelemetry: Use the OpenTelemetry SDK.
-- ConsoleDirect: Synchronously write to the console. 
+- ConsoleDirect: Synchronously write to the console.
 - ConsoleAsync: Asynchronously write to the console.
 
 Note that the ITS and ConsoleAsync modes share a the same provider
@@ -158,7 +158,7 @@ service:
 # Normal pipeline node
 nodes:
   ...
-  
+
 # Internal telemetry pipeline nodes
 internal:
   kind: receiver


rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -103,84 +99,28 @@ settings:
...
-- Noop: Not a real output mode, this setting will cause an error when
-  any provider uses Buffered or Unbuffered.
-- Raw: Use raw console logging from the global or regional logging
-  thread.
-- Internal: Use an Internal Telemetry Receiver as the destination.
+- Noop: Ignore logging.
+- ITS: Use the internal telemetry system.
+- OpenTelemetry: Use the OpenTelemetry SDK.
+- ConsoleDirect: Synchronously write to the console. 
+- ConsoleAsync: Synchronously write to the console.
```
utpilla:
```suggestion
- ConsoleAsync: Asynchronously write to the console.
```

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -5,74 +5,298 @@
...
     Warn,
     /// Error level logging.
     Error,
 }
 
+/// Logging providers for different execution contexts.
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
+pub struct LoggingProviders {
+    /// Provider mode for non-engine threads. This defines the global Tokio
+    /// `tracing` subscriber. Default is Unbuffered. Note that Buffered
```
utpilla: Update to newer terminology instead of mentioning "Buffered" and "Unbuffered".

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -5,74 +5,298 @@
...
+
+        let config = config_with(Noop, Noop, ConsoleAsync);
+        assert_invalid(&config, "internal provider is invalid");
+    }
+
+    #[test]
+    fn test_validate_engine_otel_requires_global_otel() {
+        use ProviderMode::*;
+        // Engine OpenTelemetry without global OpenTelemetry fails
+        for global in [Noop, ITS, ConsoleDirect] {
```
utpilla:
```suggestion
        for global in [Noop, ITS, ConsoleAsync, ConsoleDirect] {
```

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -5,74 +5,298 @@
...
+impl ProviderMode {
+    /// Returns true if this requires a LogsReporter channel for
+    /// asynchronous logging.
+    #[must_use]
+    pub fn needs_reporter(&self) -> bool {
+        matches!(self, Self::ITS | Self::ConsoleAsync)
+    }
+}
+
+fn default_global_provider() -> ProviderMode {
```
utpilla: nit: These methods can be marked `const`.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -5,74 +5,298 @@
...
+    ProviderMode::ITS
+}
+
+fn default_engine_provider() -> ProviderMode {
+    ProviderMode::ITS
+}
```
utpilla: These are not consistent with the default config yaml shared in the `self_tracing_architecture.md` which has `ConsoleAsync` as the default.

```yaml
service:
  telemetry:
    logs:
      level: info
      providers:
        global: console_async
        engine: console_async
        internal: noop
```

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -5,74 +5,298 @@
...
+    Noop,
+
+    /// Delivery using the internal telemetry system.
+    ITS,
+
+    /// Use OTel-Rust as the provider.
+    OpenTelemetry,
+
+    /// Asynchronous console logging.
+    /// The caller writes to a channel the same as ITS deliver, but bypasses
```
utpilla:
```suggestion
    /// The caller writes to a channel the same as ITS delivery, but bypasses
```

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -5,74 +5,298 @@
...
+    /// Synchronous console logging. Note! This can block the producing thread.
+    /// The caller writes directly to the console.
+    ConsoleDirect,
+}
+
+impl ProviderMode {
+    /// Returns true if this requires a LogsReporter channel for
+    /// asynchronous logging.
+    #[must_use]
+    pub fn needs_reporter(&self) -> bool {
```
utpilla:
```suggestion
    pub const fn needs_reporter(&self) -> bool {
```

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
jmacd: Done. In the long term I hope ITS becomes the default, but for now ConsoleAsync is safest.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
jmacd: Done. I think I don't fully understand why I should use const keywords.

utpilla (approved):

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
jmacd: Fixed.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
jmacd: :+1:

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
jmacd: Done.

rust/otap-dataflow/docs/self_tracing_architecture.md
jmacd: Done.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
utpilla: It's mainly to help the compiler apply optimizations. However, marking a method `const` doesn't guarantee a performance improvement in all cases. The most evident benefit of using `const` on the method is that it allows your method to be called from `const` expressions.

On thinking some more, I think we also have to keep in mind that for public methods removing `const` later would be a breaking change.

PR #1797, opened 2026-01-15
[otap-df-quiver] Migrate Quiver from blocking to async I/O using Tokio

AaronRM:
Migrates Quiver from blocking to async I/O using Tokio

Key Changes
- All hot-path methods are now async: `open`, `ingest`, `next_bundle`, `flush`, `maintain`, `shutdown`
- `next_bundle(id, timeout, cancel)` supports timeout and cancellation for graceful shutdown
- `poll_next_bundle(id)` provides sync non-blocking polling (renamed from old `next_bundle`)
- Re-exports `CancellationToken` from `tokio_util` for shutdown signaling
- WAL reader remains sync (intentional - only used during crash recovery)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -7,6 +7,9 @@ go:
 ci-repo:
 - changed-files:
   - any-glob-to-any-file: '.github/**/*'
+opl-parser:
+- changed-files:
+  - any-glob-to-any-file: 'rust/otap-dataflow/crates/opl/**/*'
 query-engine:
 - changed-files:
   - any-glob-to-any-file: 'rust/experimental/query_engine/**/*'
--- a/.github/workflows/pipeline-perf-test-continuous.yml
+++ b/.github/workflows/pipeline-perf-test-continuous.yml
@@ -88,10 +88,14 @@ jobs:
         - name: Run pipeline performance test log suite - Saturation
           run: |
             cd tools/pipeline_perf_test
-            for config in test_suites/integration/continuous/saturation-*.yaml; do
-              echo "Running saturation test: $config"
-              python orchestrator/run_orchestrator.py --config "$config"
-            done
+            ./scripts/run-saturation-tests.sh --output-json results/scaling-efficiency.json \
+              | tee saturation-scaling-report.txt
+            # Add scaling report to job summary
+            echo "### Saturation/Scaling Test Analysis" >> $GITHUB_STEP_SUMMARY
+            echo '```' >> $GITHUB_STEP_SUMMARY
+            # Extract just the analysis section from the report
+            sed -n '/^=.*SCALING ANALYSIS/,/^=\{10,\}$/p' saturation-scaling-report.txt >> $GITHUB_STEP_SUMMARY
+            echo '```' >> $GITHUB_STEP_SUMMARY
 
         - name: Run idle state performance tests
           run: |
@@ -110,6 +114,12 @@ jobs:
           with:
             name: benchmark-results-saturation
             path: tools/pipeline_perf_test/results/continuous_saturation*/gh-actions-benchmark/*.json
+
+        - name: Upload scaling efficiency metrics
+          uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+          with:
+            name: benchmark-results-scaling-efficiency
+            path: tools/pipeline_perf_test/results/scaling-efficiency.json
 
         - name: Upload benchmark results for processing (Idle State)
           uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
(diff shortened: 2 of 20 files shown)


rust/otap-dataflow/crates/quiver/ARCHITECTURE.md
```diff
@@ -1259,6 +1259,270 @@ returns a NACK, the embedding layer (e.g., persistence_processor):
...
+                    self.schedule_retry(bundle_ref);
+                }
+                Err(_) => handle.reject(),
+            }
+        }
+        Ok(())
+    }
+}
+```
+
```
jmacd, line 1525: Looks good, thanks for the updates.

jmacd (approved):
It looks like some of the logic required more refactoring work than other logic, mainly the wal_writer.rs file 🚀.

rust/otap-dataflow/crates/quiver/src/segment/writer.rs
```diff
@@ -153,26 +155,30 @@ impl SegmentWriter {
...
         segment: super::OpenSegment,
     ) -> Result<(u64, u32), SegmentError> {
         let path = path.as_ref();
         let (accumulators, manifest) = segment.into_parts()?;
 
+        // Create file and write data (sync - Arrow IPC doesn't support async)
         let file = File::create(path).map_err(|e| SegmentError::io(path.to_path_buf(), e))?;
         let mut writer = BufWriter::new(file);
 
         let result = self.write_streaming(&mut writer, accumulators, manifest, path)?;
```
lalitb: Seems the heavy work of IPC encoding and writing the entire segment is still synchronous. Do we plan to make it async eventually ?

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
```diff
@@ -1374,10 +1412,25 @@ fn system_time_to_nanos(ts: SystemTime) -> WalResult<i64> {
...
+/// This opens `/dev/null` (Unix) or `NUL` (Windows) to create a valid but
+/// inert file handle. The placeholder is immediately replaced with the
+/// real file after the sync operation completes.
+fn placeholder_file() -> File {
+    #[cfg(unix)]
+    let path = "/dev/null";
+    #[cfg(windows)]
+    let path = "NUL";
+    // Use std::fs::File::open which is sync, then convert to tokio
+    let std_file = std::fs::File::open(path).expect("failed to open null device");
```
lalitb: `expect` can panic if /dev/null fails (may sandboxed env, odd permissions, fd exhaust) - should there be more graceful handling. Maybe in later PR, avoid using this placeholder.

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
```diff
@@ -773,43 +780,71 @@ impl ActiveWalFile {
...
+        // Perform sync_data
+        #[cfg(test)]
+        test_support::record_sync_data();
+        let _ = std_file.sync_data();
+
+        // Convert back to tokio::fs::File
+        self.file = File::from_std(std_file);
+        self.last_flush = Instant::now();
+        self.unflushed_bytes = 0;
+    }
```
lalitb, line 877: nit - this function is probably too silent if some error occurs. At minimum, to emit a warn/debug, so operators know we skipped the drop flush. Not blocking for this PR though.

AaronRM: Added warnings for failure cases in `flush_now_sync`.

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
AaronRM: It's a valid concern. Made changes to avoid the `placeholder_file()` pattern & function entirely.

rust/otap-dataflow/crates/quiver/src/segment/writer.rs
AaronRM: Unfortunately [the arrow IPC writer doesn't support proper async](https://github.com/apache/arrow-rs/issues/7812) currently. I refactored `write_segment` to make it more async aware and moved the Arrow IPC writing to tokio's blocking thread pool.  (One other alternative is serializing to a Vec<u8> buffer and then using tokio to write the buffer async, but that would incur additional memory cost.)

AaronRM:
Fixes #1800

lquerel:
@AaronRM  I think for now this is fine, but we will probably need to revisit it in the future, because it seems to me that `tokio::fs` is really just a facade on top of synchronous syscalls. I know there were "tokio" discussions about moving toward a truly async version, similar to what exists for sockets or via `io_uring`, but I do not believe this is in place yet (to be verified).

All of this is to say, and it aligns with what @albertlockett  was saying, that later this year we will need to look at whether we should make the engine more independent from Tokio in general. In the meantime, Tokio may gain better support for asynchronous filesystem operations.

AaronRM:
@albertlockett @lquerel Ack on both comments. I went with `tokio` for now since, as noted, it's already being used elsewhere in otap-dataflow, and seems to be the current documented guidance per `async-best-practices.md`.

Abstracting out the async runtime (from all of otap-dataflow) will be challenging. It's a big topic worthy of a strategy discussion.

cijothomas:
> @albertlockett @lquerel Ack on both comments. I went with `tokio` for now since, as noted, it's already being used elsewhere in otap-dataflow, and seems to be the current documented guidance per `async-best-practices.md`.
> 
> Abstracting out the async runtime (from all of otap-dataflow) will be challenging. It's a big topic worthy of a strategy discussion.

💯 (We have similar challenge in OTel Rust too about abstracting async runtime)

PR #1798, opened 2026-01-15
Ack/Nack for routing in transform_processor

albertlockett:
Opening this as draft for now. There are some addition test cases and cleanup I want to do, but I wanted to first ensure my handling of the Pdata Contexts was correct.

---

Closes: #1784 

Now that we have `route_to` in OPL, in combination with `if/else`, this can create a scenario where we split the batch.

```kql
logs |
  route_to "out_port1"
}
// implicit collect everything that didn't go in "if" branch
```
A pipeline like this would emit two batches:
1. "ERROR" logs on the processor's "out_port1"
2. all other logs on the default out port in two 

If the batch had subscribers, when we process a pdata we must keep the context for the inbound batch, and create new contexts for the outbound batches. When all the outbound batches Ack/Nack'd, we must then Ack/Nack the inbound context.

This PR adds a `Contexts` type for juggling the inbound/outbound contexts and updates the transform processor to manage contexts + Ack/NAck correctly.


albertlockett:
@jmacd could I ask you for a review on this :)? I was using the batch_processor Ack/Nack implementation as inspiration.

One thing in particular I'd like to ensure is correct is the behaviour of juggling the contexts.

Basically when we find we've split the batch
1. For the inbound batch, put it in the `inbounds` slot map in `Contexts` which contains number of outbound batches + the original inbound context
2. For each outbound batch, put it in the `outbounds` slot map in `Contexts` with a value pointing back at the `inbounds` key.
3. Subscribe the outbound context using call data derived from the the `outbounds` slotmap key

Then when we receive an Ack/Nack:
1. Lookup the inbound key from `outbounds` slot map using the key from the calldata. Clear the `outbounds` slot map
2. decrement the count of outbounds in the `inbounds` slot map. If the count is zero, then we Ack/Nack the original inbound context.

I'm also curious if, when testing, is this the correct way to setup test `Ack` messages using the outbound contexts to simulate a downstream component having Ack/Nacked the batch:
https://github.com/open-telemetry/otel-arrow/blob/e5a0975737cc969a448930348a0e649b6ccfbdd4/rust/otap-dataflow/crates/otap/src/transform_processor.rs#L996-L1017

I realize I'm asking you to reverse engineer a lot of code here, so happy to walk through on teams if it's easier :)

jmacd:
@albertlockett I think you want to use a call to Context::next_ack, for the test section you quoted. This does what the effect handler would have done when the recipient responded with an Ack. See how batch_processor.rs tests use Context::next_ack, basically.

albertlockett:
> @albertlockett I think you want to use a call to Context::next_ack, for the test section you quoted. This does what the effect handler would have done when the recipient responded with an Ack. See how batch_processor.rs tests use Context::next_ack, basically.

thanks @jmacd, that worked! made this change in c031e629

albertlockett:
> Looks good. Looks like maybe the new code could be applied to the batch_processor in a future PR, maybe.

Yeah I think we could reuse the `Contexts` with a little bit of refactoring. The gap with the current implementation is that it doesn't expect an outbound batch to be associated with more than one inbound batch (because currently we only split, we don't combine batches). I imagine eventually this change will need to be made b/c we'd want OPL to support batching, so once that is in place we could reuse this in batch_processor.

rust/otap-dataflow/crates/otap/src/transform_processor/context.rs
```diff
@@ -0,0 +1,286 @@
...
+            self.outbound
+                .allocate(|| (outbound, ()))
+                .map(|(key, _)| key)
+        } else {
+            Some(Key::null())
+        }
+    }
+
+    pub fn set_failed(&mut self, outbound_key: Key, error_reason: String) {
+        if let Some(inbound) = self.inbound.get_mut(outbound_key) {
```
lalitb: Should we first lookup the outbound to get the inbound_key, as done in  `clear_outbound` (line 115) ?

rust/otap-dataflow/crates/otap/src/transform_processor/context.rs
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    /// Clears the outbound slot and returns the context and error reason if the inbound slot is now empty.
+    ///
+    /// Returns `Some((context, error_reason))` if the inbound slot is now empty. This would mean that
+    /// all outbound batches for this inbound slot have been processed and the inbound batch could be
+    /// Ack/NAck'd
+    pub fn clear_outbound(&mut self, outbound_key: Key) -> Option<(Context, Option<String>)> {
+        let inbound_key = {
+            let outbound = self.outbound.get(outbound_key)?;
```
lalitb: Maybe I am missing something, but seems we get the slot, but never remove it from `self.outbound` ?

rust/otap-dataflow/crates/otap/src/transform_processor/context.rs
```diff
@@ -0,0 +1,286 @@
...
+    /// Returns `None` if the outbound slot map is full.
+    pub fn insert_outbound(&mut self, inbound_key: Key) -> Option<Key> {
+        // incr inbound
+        if let Some(inbound) = self.inbound.get_mut(inbound_key) {
+            inbound.num_outbound += 1;
+
+            // insert outbound
+            let outbound = Outbound { inbound_key };
+            self.outbound
+                .allocate(|| (outbound, ()))
```
lalitb: If allocate fails, we return early but already incremented num_outbound at line 88 - this would leave the inbound context stuck.

rust/otap-dataflow/crates/otap/src/transform_processor/context.rs
albertlockett: Thanks @lalitb , that's a good catch. This should call `take`, not `get`. Fixed in https://github.com/open-telemetry/otel-arrow/pull/1798/commits/79788a0eadfb210235ee30445e9fc4600de20acc

rust/otap-dataflow/crates/otap/src/transform_processor/context.rs
albertlockett: Yes, you're right @lalitb 

Fixed in https://github.com/open-telemetry/otel-arrow/pull/1798/commits/4ab041d5c8d44ad8d360f88220431e45cf9256d8 and added some tests that exercise this code path

rust/otap-dataflow/crates/otap/src/transform_processor/context.rs
albertlockett: Thanks @lalitb . Fixed, and added tests that would catch this bug in https://github.com/open-telemetry/otel-arrow/pull/1798/commits/1d3e9a3f16046e82124b6730c8acbb3d5d178156

albertlockett:
Converting to draft as there's still some work to do to clean up `handle_exec_result`. I'd like to make the error handling more consistent and have better handling/coverage on the case where the pipeline routed some batches, but the batch destined for the default out port was not returned b/c the pipeline failed to execute to completion

Issue #1637, opened 2025-12-16
Perf Test - DF engine is dropping logs


cijothomas:
https://open-telemetry.github.io/otel-arrow/benchmarks/nightly/backpressure/

OTLP-Attr-OTAP is losing 100% logs.

cijothomas:
The behavior is same even for syslog too - just pump up load, and we are losing logs.

PR #1799, opened 2026-01-15
PerfTest -  Saturation tests to use 10 sec wait

cijothomas:
Fixes https://github.com/open-telemetry/otel-arrow/issues/1637

Matching the currently hardcoded 10 sec on pipeline shutdown path.
https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/admin/src/pipeline_group.rs#L68

Only modified the saturation tests to use 10 sec. Rest of tests seems fine with 3. In future, we can be smarter and block until engine_shutdown occurs up-to a deadline. If deadline is hit, then we can hint in test itself that we didn't have clean shutdown, so data loss likely occurred due to this.

Local run shows 0% dataloss with this change.

### DIFF
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -5,6 +5,9 @@ coverage:
         target: auto # Increase
         threshold: 0.1% # Allow small fluctuation
 
+ignore:
+  - "rust/otap-dataflow/crates/quiver-e2e/**" # Test tool, not production code
+
 comment:
   layout: "header, diff, flags, components"
 
@@ -30,3 +33,7 @@ component_management:
       name: "otel-arrow-go"
       paths:
         - "go/"
+    - component_id: "quiver"
+      name: "quiver"
+      paths:
+        - "rust/otap-dataflow/crates/quiver/"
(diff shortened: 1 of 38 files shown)


cijothomas:
https://github.com/open-telemetry/otel-arrow/issues/1801 is opened to track improving the engine shutdown.

lquerel (approved):
LGTM

tools/pipeline_perf_test/test_suites/integration/continuous/saturation-16cores.yaml
```diff
@@ -6,6 +6,6 @@ from_template:
     engine_core_range: "0-15"
     loadgen_cores: 48
     loadgen_core_range: "16-63"
-    backend_cores: 8
-    backend_core_range: "64-71"
+    backend_cores: 16
```
cijothomas, line 9: this was a typo! It should have 16 cores, to pump load to 16 engines!

cijothomas, line 9: this was a typo! It should have 16 cores, to pump load to 16 engines!

Issue #1801, opened 2026-01-15
Add optional wait and deadline parameter to /pipeline-groups/shutdown endpoint


cijothomas:
Currently, [POST /pipeline-groups/shutdown](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/admin/src/pipeline_group.rs#L81l) returns 202 Accepted immediately after queuing shutdown signals and uses hardcoded 10 sec deadline. Callers must guess how long to wait for the pipeline to fully drain.

Proposal: Add optional query parameters:
wait=true — block until all pipelines reach terminal state
deadline_secs=N — override the default 10-second shutdown deadline

POST /pipeline-groups/shutdown                              → 202 (immediate, 10s deadline)
POST /pipeline-groups/shutdown?wait=true                    → 200 (after drain, 10s deadline)
POST /pipeline-groups/shutdown?wait=true&deadline_secs=30   → 200 (after drain, 30s deadline)
POST /pipeline-groups/shutdown?deadline_secs=60             → 202 (immediate, 60s deadline)

Immediate use case: In performance/saturation testing, pipelines under heavy load may need longer drain times. Hardcoding wait times leads to either data loss (too short) or wasted time (too long).

cijothomas:
Alternatively, callers could poll [/pipeline-groups/status](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/admin/src/pipeline_group.rs#L53) until phase is Stopped, but a blocking ?wait=true would be simpler.

cijothomas:
Self-assign.

PR #1803, opened 2026-01-16
Shutdown improvements to pipeline

cijothomas:
1. Fixes https://github.com/open-telemetry/otel-arrow/issues/1801
2. Perf tests modified to leverage this, instead of optimistic sleep and wait. Most importantly, we now invoke shutdown on load-gen, engine, backend (in that order), to ensure clean shutdown all throughout. No more data loss!
3. For saturation/scale test - use batch-size of 512 (OTel SDK's default batch size). We previously used 200 due to data loss, but now that dataloss is root caused and fixes, switching to 512 as default batch size.
4. Added 24 core scenario too - this is the max we can do in the perf machine, as we ran out of cores in it. (24 cores for engine means 24 more cores for backend and 72 for loadgen!)

### DIFF
--- a/rust/otap-dataflow/crates/opl/src/opl.pest
+++ b/rust/otap-dataflow/crates/opl/src/opl.pest
@@ -160,6 +160,10 @@ rename_operator_call = {
     ("rename" | "project-rename") ~ assignment_expression ~ ("," ~ assignment_expression)*
 }
 
+remove_map_keys_operator_call = {
+    ("exclude" | "project-away") ~ (attribute_selection_expression | index_expression) ~ ("," ~ (attribute_selection_expression | index_expression))*
+}
+
 set_operator_call = {
     ("set" | "extend") ~ assignment_expression
 }
@@ -175,6 +179,7 @@ route_to_operator_call = {
 operator_call = {
     set_operator_call
     | rename_operator_call
+    | remove_map_keys_operator_call
     | if_else_operator_call
     | route_to_operator_call
     | where_operator_call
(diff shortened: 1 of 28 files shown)


rust/otap-dataflow/crates/state/src/pipeline_status.rs
```diff
@@ -54,6 +54,21 @@ impl PipelineStatus {
...
+    /// Returns true if all cores have reached a terminal state (Stopped, Deleted, or Failed).
+    /// Returns false if there are no cores tracked or if any core is still active.
+    pub fn is_terminated(&self) -> bool {
+        if self.cores.is_empty() {
+            return false;
+        }
+        self.cores.values().all(|c| {
+            matches!(
+                c.phase,
+                PipelinePhase::Stopped | PipelinePhase::Deleted | PipelinePhase::Failed(_)
```
albertlockett: Would it be better to expose this as a method on `PipelinePhase`?
```rs
impl PipelinePhase {
    pub fn is_terminated(&self) -> bool {
        match self {
            Self::Stopped | Self::Deleted | Self::Failed(_) => true,
            Self::Pending
            | Self::Starting
            | Self::Running
            | Self::Draining
            | Self::Updating
            | Self::RollingBack
            | Self::Rejected(_)
            | Self::Deleting(_) => false,
        }
    }
}
```

I feel like if we ever added another terminal phase, it might be safer b/c we'd get a compiler error if the new variant wasn't added to the is_terminated method.

Also, I wonder if `PipelinePhase::Rejected(_)` counts as a terminated state 🤔 ?

cijothomas: yes makes sense.
I am now treating Rejected as terminal (added code comment with a reasoning. LMK).

Anything with "..ing" in it, is not terminal 🤣

albertlockett: > Anything with "..ing" in it, is not terminal 🤣

lol, this makes sense but I am glad you didn't implement `is_terminated()` as 
`format!("{self:?}").has_suffix("ed")` 
😆

PR #1804, opened 2026-01-16
OPL parser support remove map keys operator call

### LINKED ISSUE #1667: Support conditionally applying pipeline to some data in columnar query engine

### LINKED ISSUE #1763: OPL Parser: delete map keys expression
We should support syntax for renaming map keys. rename and project-rename will be synonms
```
project_away attributes["x"] = attributes["y"]
exclude attributes["x"] = attributes["y"]
```
This should get parsed into DataExpression::Transform(TransformExpression::RemoveMapKeys(..)) from our expressions AST

albertlockett:
Closes #1763
Closes #1667 

Adds to OPL parser support for an operator to remove keys from maps (attributes). The name of this operator, like in KQL, is `project-away`, but there is an alias called `exclude`.

```kql
logs | project-away attributes["x"], attributes["y"]
logs | exclude resource.attributes["z"]
// etc.
```

This PR also uses the OPL parser in tests in the columnar query engine which use this operator. Finally, this cleans up the test code in `pipeline::conditional` to remove the `ConditionalTest` helper type that was needed to setup the tests until we had this parser support.

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
@@ -327,9 +327,8 @@ mod test {
         test_invalid_renames_are_errors::<OplParser>().await;
     }
 
-    #[tokio::test]
-    async fn test_delete_attributes() {
-        let result = exec_logs_pipeline::<KqlParser>(
+    async fn test_delete_attributes<P: Parser>() {
+        let result = exec_logs_pipeline::<P>(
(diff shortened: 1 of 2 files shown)


lquerel (approved):
LGTM

PR #1807, opened 2026-01-16
[syslog-cef-receiver] Add support for parsing tags for RFC 3164

### LINKED ISSUE #1729: `syslog_cef_receiver` deviation from Go Collector `syslog` parsing of TAG in RFC 3164
The [Collector-Contrib syslogreceiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/receiver/syslogreceiver) utilizes an underlying parser library that DOES produce `appname` and `proc_id` attributes on Syslog RFC 3164:

[Collector-Contrib code link](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/b077ec3f9096acbaa234be9f60dc0a419f495c1b/pkg/stanza/operator/parser/syslog/parser.go#L135C1-L144C3)
```go
// parseRFC3164 will parse an RFC3164 syslog message.
func (p *Parser) parseRFC3164(syslogMessage *rfc3164.SyslogMessage, skipPriHeaderValues bool) (map[string]any, error) {
    value := map[string]any{
	"timestamp": syslogMessage.Timestamp,
	"hostname":  syslogMessage.Hostname,
	"appname":   syslogMessage.Appname,
	"proc_id":   syslogMessage.ProcID,
	"msg_id":    syslogMessage.MsgID,
	"message":   syslogMessage.Message,
    }
    ...
```

[Internal parser dependency code link](https://github.com/leodido/go-syslog/blob/aae00dca59f5d850e8900663cb2405a2631a78fe/rfc3164/syslog_message.go#L47-L54)
```go
	if sm.tag != "-" && sm.tag != "" {
		out.Appname = &sm.tag
	}
	if sm.content != "-" && sm.content != "" {
		// Content is usually process ID
		// See https://tools.ietf.org/html/rfc3164#section-5.3
		out.ProcID = &sm.content
	}
```

The [RFC 3164 spec](https://datatracker.ietf.org/doc/html/rfc3164#section-5.3) has this to say:
> It has also been considered to be a good practice to include some
  information about the process on the device that generated the
  message - if that concept exists.  This is usually the process name
  and process id (often known as the "pid") for robust operating
  systems.  The process name is commonly displayed in the TAG field.
  Quite often, additional information is included at the beginning of
  the CONTENT field.  The format of "TAG[pid]:" - without the quote
  marks - is common.  The left square bracket is used to terminate the
  TAG field in this case and is then the first character in the CONTENT
  field.  If the process id is immaterial, it may be left off.

> In that case, a colon and a space character usually follow the TAG.
  This would be displayed as "TAG: " without the quotes.  In that case,
  the colon is the first character in the CONTENT field.

When the current Rust implementation gets a message like:
> <134>{timestamp} securityhost myapp[1234]: User admin logged in from 10.0.0.1 successfully

It produces
> LogRecord #0:
   -> ObservedTimestamp: 1767725764424071771
   -> Timestamp: 1767725764000000000
   -> SeverityText: INFO
   -> SeverityNumber: 9
   -> Body: <134>Jan 06 18:56:04 securityhost myapp[1234]: User admin logged in from 10.0.0.1 successfully
   -> Attributes:
      -> syslog.facility: 16
      -> syslog.severity: 6
      -> syslog.host_name: securityhost
      -> syslog.tag: myapp[1234]
      -> syslog.content: User admin logged in from 10.0.0.1 successfully
   -> Trace ID:
   -> Span ID:
   -> Flags: 0

Can/should we optimistically produce `app_name` (`myapp`) and `proc_id` (`1234`) attributes even on RFC 3164? One possible alternative is using later processors to transform the `tag` attribute.

utpilla:
Fixes #1729 

## Changes
- Parse `syslog.tag` field further into `syslog.app_name` and `syslog.process_id` when applicable for RFC 3164

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/parser/rfc3164.rs
+++ b/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/parser/rfc3164.rs
@@ -10,6 +10,8 @@ pub struct Rfc3164Message<'a> {
     pub(super) timestamp: Option<&'a [u8]>,
     pub(super) hostname: Option<&'a [u8]>,
     pub(super) tag: Option<&'a [u8]>,
+    pub(super) app_name: Option<&'a [u8]>,
+    pub(super) proc_id: Option<&'a [u8]>,
     pub(super) content: Option<&'a [u8]>,
     pub(super) input: &'a [u8],
 }
@@ -25,6 +27,12 @@ pub struct Rfc3164Message<'a> {
 /// - The message is parsed for any identifiable fields (timestamp, hostname, etc.)
 /// - No default priority is assigned
 /// - The entire message may be treated as content if no structure is found
+///
+/// # TAG Parsing
+/// The TAG field in RFC 3164 often contains the application name and optionally
+/// a process ID in the format `appname[pid]`. This parser extracts:
+/// - `app_name`: The part before `[` (or the entire TAG if no `[` is present)
+/// - `proc_id`: The numeric content between `[` and `]` (only if it's a valid number)
 pub fn parse_rfc3164(input: &[u8]) -> Result<Rfc3164Message<'_>, parser::ParseError> {
     if input.is_empty() {
         return Err(parser::ParseError::EmptyInput);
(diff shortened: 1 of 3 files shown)


rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/parser/rfc3164.rs
```diff
@@ -100,16 +108,74 @@ pub fn parse_rfc3164(input: &[u8]) -> Result<Rfc3164Message<'_>, parser::ParseEr
...
+/// This function extracts:
+/// - `app_name`: Everything before the `[` character (or the entire TAG if no `[`)
+/// - `proc_id`: The content between `[` and `]`, only if it's a valid numeric value
+///
+/// # Examples
+/// - `"su"` -> app_name=`"su"`, proc_id=None
+/// - `"sshd[5678]"` -> app_name=`"sshd"`, proc_id=`"5678"`
+/// - `"app[worker-1]"` -> app_name=`"app"`, proc_id=None (non-numeric)
+/// - `"app[]"` -> app_name=`"app"`, proc_id=None (empty)
+fn parse_tag_components(tag: Option<&[u8]>) -> (Option<&[u8]>, Option<&[u8]>) {
```
lquerel, line 139: What do we want to do with tag=`app[123]extra` ?

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/parser/rfc3164.rs
```diff
@@ -100,16 +108,74 @@ pub fn parse_rfc3164(input: &[u8]) -> Result<Rfc3164Message<'_>, parser::ParseEr
...
+    };
+
+    // Extract proc_id (content between '[' and ']')
+    if close_bracket_pos > bracket_pos + 1 {
+        let proc_id_bytes = &tag[bracket_pos + 1..close_bracket_pos];
+        // Only accept numeric proc_id values
+        if proc_id_bytes.iter().all(|&b| b.is_ascii_digit()) {
+            return (app_name, Some(proc_id_bytes));
+        }
+    }
```
lquerel, line 174: Not a big deal, but this algo looks like a multi-pass starting at the beginning of tag for the first bracket and then another time for the second bracket. This could probably be improved with a small state machine going over the tag bytes.

lquerel (approved):
Approved but made 2 suggestions.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/parser/rfc3164.rs
utpilla: For this message, we would have:
- `syslog.tag` set to `app[123]extra`
- `syslog.app_name` set to `app`
- `syslog.process_id` set to `123`

We would ignore "extra". It would be present in the tag though.

PR #1808, opened 2026-01-16
Split opentelemetry_client into otel_sdk, tracing_init, and ITS parts

jmacd:

This is a non-functional refactoring of `opentelemetry_client.rs` into other places. This will make it clearer what changes in #1771 and what is just moving around.

Moves runtime elements into the InternalTelemetrySystem, simplifies setup for the controller where logs/metrics were separated.

Moves OTel-SDK specific pieces into `otel_sdk` module, separates the Tokio `tracing` setup.

### DIFF
--- a/.github/workflows/pipeline-perf-test-continuous.yml
+++ b/.github/workflows/pipeline-perf-test-continuous.yml
@@ -103,6 +103,11 @@ jobs:
             python orchestrator/run_orchestrator.py --config test_suites/integration/continuous/idle-state-docker.yaml
             python orchestrator/run_orchestrator.py --config test_suites/integration/continuous/idle-state-allcores-docker.yaml
 
+        - name: Run passthrough performance test
+          run: |
+            cd tools/pipeline_perf_test
+            python orchestrator/run_orchestrator.py --config test_suites/integration/continuous/passthrough.yaml
+
         - name: Upload benchmark results for processing
           uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
           with:
@@ -126,6 +131,12 @@ jobs:
           with:
             name: benchmark-results-idle-state
             path: tools/pipeline_perf_test/results/idle_state*/gh-actions-benchmark/*.json
+
+        - name: Upload benchmark results for processing (Passthrough)
+          uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+          with:
+            name: benchmark-results-passthrough
+            path: tools/pipeline_perf_test/results/continuous_passthrough/gh-actions-benchmark/*.json
 
         - name: Add benchmark link to job summary
           run: |
@@ -170,6 +181,13 @@ jobs:
               merge-multiple: true
               path: results-idle-state
 
+          - name: Download benchmark artifacts (Passthrough)
+            uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5.0.0
+            with:
+              pattern: benchmark-results-passthrough*
+              merge-multiple: true
+              path: results-passthrough
+
           - name: Consolidate pipeline benchmark data
             run: |
               bash ./.github/workflows/scripts/consolidate-benchmarks.sh results-pipeline output-pipeline.json
@@ -181,6 +199,10 @@ jobs:
           - name: Consolidate idle state benchmark data
             run: |
               bash ./.github/workflows/scripts/consolidate-benchmarks.sh results-idle-state output-idle-state.json
+
+          - name: Consolidate passthrough benchmark data
+            run: |
+              bash ./.github/workflows/scripts/consolidate-benchmarks.sh results-passthrough output-passthrough.json
 
           - name: Update pipeline benchmark data and deploy to GitHub Pages
             uses: benchmark-action/github-action-benchmark@4bdcce38c94cec68da58d012ac24b7b1155efe8b # v1.20.7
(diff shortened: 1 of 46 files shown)


rust/otap-dataflow/crates/telemetry/src/otel_sdk/logger_provider.rs
```diff
@@ -29,39 +25,11 @@ pub struct LoggerProvider {
...
-    ///    flushes them periodically.
-    ///
-    /// The TODO here is to evaluate these options and implement one of them.
-    /// As of now, this causes contention, and we just need to accept temporarily.
-    ///
-    /// TODO: Evaluate also alternatives for the contention caused by the global
-    /// OpenTelemetry logger provider added as layer.
+    /// This creates the `SdkLoggerProvider` with the configured exporters.
+    /// Note: This does NOT initialize the global tracing subscriber.
+    /// Use `tracing_init::init_global_subscriber()` separately to set up tracing.
```
cijothomas: that alone is not sufficient.. somewhere we have to add a layer that is connected to OTel..

cijothomas: looks like that is done correctly, so just this comment need to mention that the global subscriber would be something that has OTel Layer, Fmt layer.

jmacd: The comment didn't make future sense. I removed it and just pointed to init_global_subscriber where the OTel bridge and fmt layer are configured. This is the area where #1771 adds new choices that I'm preparing for.

rust/otap-dataflow/crates/telemetry/src/lib.rs
```diff
@@ -81,25 +92,53 @@ pub struct InternalTelemetrySystem {
...
+
+    // === OTel SDK Subsystem ===
+    /// OTel SDK meter provider for metrics export.
+    meter_provider: SdkMeterProvider,
+
+    /// OTel SDK logger provider for logs export.
+    logger_provider: SdkLoggerProvider,
+
+    /// Tokio runtime for OTLP exporters (kept alive).
+    _otel_runtime: Option<tokio::runtime::Runtime>,
```
lalitb, line 103: We introduced owned OTel providers/runtime here, but `run`/`run_collection_loop` still just drop self without calling shutdown or returning a handle. Any caller that drives ITS via run will now exit without flushing the meter/logger providers or keeping the runtime alive. Can we either invoke shutdown() before returning or return the shutdown handle from run so callers can flush?

rust/otap-dataflow/crates/telemetry/src/lib.rs
```diff
@@ -146,10 +203,58 @@ impl InternalTelemetrySystem {
     pub async fn run_collection_loop(self) -> Result<(), Error> {
         self.collector.run_collection_loop().await
     }
+
+    /// Shuts down the OpenTelemetry SDK providers.
+    pub fn shutdown(&self) -> Result<(), Error> {
```
lquerel: nit: The name of this method gives the impression that it shuts down the entire ITS.

rust/otap-dataflow/crates/telemetry/src/collector.rs
```diff
@@ -61,6 +62,22 @@ impl InternalCollector {
...
+    /// Runs the collection loop until cancellation is requested.
+    ///
+    /// This method starts the internal signal collection loop and listens for a shutdown signal.
+    /// It returns when either the collection loop ends (Ok/Err) or the shutdown signal fires.
+    pub async fn run(self, cancel: CancellationToken) -> Result<(), Error> {
+        tokio::select! {
+            res = self.run_collection_loop() => {
+                res
+            }
+            _ = cancel.cancelled() => {
```
utpilla, line 72: Use `biased;` to prioritize shutdown handling?

rust/otap-dataflow/crates/telemetry/src/lib.rs
lalitb: Adding more details to be clear what I meant:
```rust
 async fn broken_example() {                                                                                                                                                                                   
      let its = InternalTelemetrySystem::new(config)?;                                                                                                                                                          
                                                                                                                                                                                                                
      // run() consumes `its`                                                                                                                                                                                   
      its.run(cancel_token).await?;                                                                                                                                                                             
                                                                                                                                                                                                                
      // its.shutdown()  <-- Can't do this, `its` is gone!                                                                                                                                                      
      // Providers dropped without flush, telemetry lost                                                                                                                                                        
  }   
```

rust/otap-dataflow/crates/telemetry/src/lib.rs
utpilla: This code is identical to `OtelShutdownHandle::shutdown` and also seems unnecessary. After calling `into_parts()`, the `InternalTelemetrySystem` is consumed (destroyed). So `InternalTelemetrySystem::shutdown()` can never be called in the `into_parts()` usage pattern.

utpilla: Same concern with `run()` and `run_collection_loop()` methods in `InternalTelemetrySystem`. With the `into_parts()` usage, we would only be needing the run methods of `InternalCollector`, right?

rust/otap-dataflow/crates/telemetry/src/lib.rs
jmacd: Fixed, I think! I removed the run() method. 

The caller is expected to use into_parts() and hold the shutdown before running. My mistake.

jmacd:
Thanks @utpilla for helping clean this up. Thanks @lalitb for reviewing!

rust/otap-dataflow/crates/telemetry/src/lib.rs
jmacd: @utpilla helped clean this up. Now I understand that the reason OTel SDKs require shutdown and the InternalCollector doesn't, this is from being spawned differently.

PR #1809, opened 2026-01-16
Update batch size variation perf tests

cijothomas:
Modified to use 10, 100, 512, 1024, 4096, 8192 as sizes.


cijothomas:
@lquerel Could you re-add to merge queue? This got kicked out.

PR #1810, opened 2026-01-16
PerfTest - add passthrough scenario

cijothomas:
All existing tests add a dummy processor in the middle to force conversion into internal format. But there are real scenarios where engine can act as "pass-through" without having to do this conversion. 
This adds a perf-test to continuously measure the throughput when operating as pass-through. 
Modelled after saturation tests - where we put as much load as required to hit 100% CPU in engine. Local run shows it can do twice (minimum) the throughput!


clhain:
nice, lgtm

PR #1811, opened 2026-01-16
remove transport optimized ID encoding in filter processor

albertlockett:
I'm pretty sure this is why we're seeing the OTAP -> Filter benchmark dropping the logs.

We were assuming the IDs were plain encoded in the filter processor, but if they have the transport optimized encoding (e.g., if they were received by OTAP receiver), they'd have the delta encoding which would really mess things up when we try to join the attributes to the logs.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/filter_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/filter_processor.rs
@@ -113,7 +113,8 @@ impl local::Processor<OtapPdata> for FilterProcessor {
                 // convert to arrow records
                 let (context, payload) = pdata.into_parts();
 
-                let arrow_records: OtapArrowRecords = payload.try_into()?;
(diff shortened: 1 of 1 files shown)


albertlockett:
cc @cijothomas

albertlockett:
Leaving this as draft until we have time to confirm this fixes the integration test

albertlockett:
Confirming that this fixes the dropped logs issue.

Before:
```
## Table: log_loss

**Description**: Lost logs (failed or sent but not received) total and as a percentage

|   logs_lost |   logs_lost_pct |
|------------:|----------------:|
|     6.3e+06 |               1 |
```

After:
```

## Table: log_loss

**Description**: Lost logs (failed or sent but not received) total and as a percentage

|   logs_lost |   logs_lost_pct |
|------------:|----------------:|
|  6.1555e+06 |           0.947 |
```

PR #1813, opened 2026-01-16
Snapshot: mechanism for internal Tokio tracing subscriber to OTLP byes or console output

jmacd:
For demonstration. 
This is the main mechanism of #1771


jmacd:
For discussion with @utpilla about what #1808 is supposed to help with.

PR #1818, opened 2026-01-17
Move ObservedEvent into crates/telemetry, consolidated with self_tracing::LogRecord

jmacd:

The ObservedEvent has associated flume channels and a connection with the existing metrics and admin component which make it an appealing way to transport log events in the engine.

Move PipelineKey, DeployedPipelineKey, CoreId types into crates/config.

Therefore, moving ObservedEvent into crates/telemetry lets us (optionally) use the same channel already use for lifecycle events for tokio log records. The existing event structure is extended with a `EventMessage` enum which supports None, String, or LogRecord messages. This way we can use a log record as the event message for all existing event types. The `event.rs` file moves, only ObservedEventRingBuffer from that file remains in crates/state.

The LogRecord has been storing a timestamp. Now, we leave that to the ObservedEvent struct. LogRecord passes through SystemTime everywhere it has been used. Callers generally compute this and pass it in. Minor cleanup in self_tracing/formatter.rs, do not pass SavedCallsite it can be calculated from record metadata as needed.

In internal_events, the raw_error! macro has been replaced with a helper to generate LogRecord values first, by level. This lets us pass `info_event!("string", key=value)` to any of the event constructors and construct an OTLP bytes message instead of a String message.

### DIFF
--- a/rust/otap-dataflow/benchmarks/benches/self_tracing/main.rs
+++ b/rust/otap-dataflow/benchmarks/benches/self_tracing/main.rs
@@ -11,15 +11,13 @@
 //! Example: `encode/3_attrs/1000_events` = 300 µs → 300 ns per event
 
 use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+use otap_df_pdata::otlp::ProtoBuffer;
+use otap_df_telemetry::self_tracing::{ConsoleWriter, DirectLogRecordEncoder, LogRecord};
+use std::time::SystemTime;
 use tracing::{Event, Subscriber};
 use tracing_subscriber::layer::Layer;
 use tracing_subscriber::prelude::*;
 use tracing_subscriber::registry::LookupSpan;
-
-use otap_df_pdata::otlp::ProtoBuffer;
-use otap_df_telemetry::self_tracing::{
-    ConsoleWriter, DirectLogRecordEncoder, LogRecord, SavedCallsite,
-};
 
 #[cfg(not(windows))]
 use tikv_jemallocator::Jemalloc;
@@ -28,16 +26,21 @@ use tikv_jemallocator::Jemalloc;
 #[global_allocator]
 static GLOBAL: Jemalloc = Jemalloc;
 
-/// The operation to perform on each event within the layer.
+/// The operation to perform on each event within the layer.  The cost
+/// of generating the timestamp is not included in the  measurement.
 #[derive(Clone, Copy)]
 enum BenchOp {
-    /// Encode the event into a LogRecord only.
-    Encode,
-    /// Encode once, then format N times.
+    /// Encode the event body into a new LogRecord.  This includes
+    /// encoding the body and attributes, not callsite details or
+    /// timestamp.
+    NewRecord,
+    /// Encode once, then format standard representation (with
+    /// timestamp) N times.
     Format,
-    /// Encode and format together N times.
+    /// Encode and format standard representation (with timestamp) N
+    /// times.
     EncodeAndFormat,
-    /// Encode to protobuf N times.
+    /// Encode to complete protobuf (with timestamp) N times.
     EncodeProto,
 }
 
@@ -58,8 +61,9 @@ where
     S: Subscriber + for<'a> LookupSpan<'a>,
 {
     fn on_event(&self, event: &Event<'_>, _ctx: tracing_subscriber::layer::Context<'_, S>) {
+        let now = SystemTime::now();
         match self.op {
-            BenchOp::Encode => {
+            BenchOp::NewRecord => {
                 for _ in 0..self.iterations {
                     let record = LogRecord::new(event);
                     let _ = std::hint::black_box(record);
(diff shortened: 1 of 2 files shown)


rust/otap-dataflow/crates/telemetry/src/event.rs
```diff
@@ -1,80 +1,94 @@
...
-        self.buf.is_empty()
-    }
-}
-
 /// An observed event emitted by the engine.
 #[derive(Debug, Clone, Serialize)]
 pub struct ObservedEvent {
     // ---- Source identification ----
-    /// Unique key identifying the pipeline instance.
+    /// Unique key identifying the pipeline instance (None for global/log events).
```
lquerel: What does None mean here, given that `DeployedPipelineKey` is not an Option?

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
```diff
@@ -142,33 +142,88 @@ macro_rules! otel_error {
 #[macro_export]
 macro_rules! raw_error {
     ($name:expr $(, $($fields:tt)*)?) => {{
-        use $crate::self_tracing::{ConsoleWriter, RawLoggingLayer};
+        use $crate::self_tracing::ConsoleWriter;
+        let now = std::time::SystemTime::now();
+        let record = $crate::error_event!($name $(, $($fields)*)?);
+        ConsoleWriter::no_color().print_log_record(now, &record);
+    }};
+}
```
lquerel, line 150: If I understand correctly, we should use `raw_error!` with great care, since there is no way to eliminate its cost, neither at compile time nor at runtime. Is that correct?
Also, for `println!`, we rely on Clippy to catch it in CI (and prevent them to end up in regular code). So what should we do for `raw_error!`?

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
```diff
@@ -142,33 +142,88 @@ macro_rules! otel_error {
...
+        raw_error!("simple error message");
+    }
+
+    #[test]
+    fn test_log_record_macros() {
+        // Test info_event! with attributes
+        let record = info_event!("test.event", count = 42i64, name = "test");
+        let callsite = record.callsite();
+        assert_eq!(*callsite.level(), Level::INFO);
+        assert_eq!(callsite.name(), "test.event");
```
lquerel, line 217: Is the line number returned by the callsite correct? 
I'm asking because I'm not sure I understand how the call site behaves with the nesting of the macros defined earlier.

rust/otap-dataflow/crates/telemetry/src/event.rs
jmacd: I backed out a part of this change, the comment's out of date. I imagined we'll prefer to use an internal pipeline key instead of a None here.
```suggestion
    /// Unique key identifying the pipeline instance.
```

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
jmacd: Yes. There will be very few of these, for cases where all/else fails. Is it difficult to configure clippy with a similar rule for raw_error? This statement is the same as println/eprintln but with the structured syntax of Tokio tracing and use of our code path.
I can see an argument that having raw_error become an eprintln!() statement makes sense; my preference is to use the same instrumentation interface use everywhere, so my goal is to replace log::error! (formatting) with tracing::error! or otel_error! (structured) and eprintln! (formatting) with raw_error! (structured).

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
jmacd: I added a test. These macros are designed for the kind of nesting here, IIUC.

jmacd:
> long term is that we must have a way to compile the engine in a mode where only vital events are compiled in, and ...

I can see your concern. I showed how to use special macros to encode LogRecord (i.e., structured body and key/values) with static callsite, but the current lifecycle events are dispatched unconditionally. I think you would prefer to see structured errors, but I also see formatting happening in the existing events.

My goals here are:

1. the admin thread can be assigned the role of collecting LogRecords from subscribers, this for the ConsoleAsync mode of internal logging (this is the fallback of not using ITS i.e., not using an internal pipeline). This means we can send logs on the same channel as the current lifecycle events.
2. To suggest that anywhere you see an Option<String> with format!(...) body you might instead use an info_event!(...) macro to encode OTLP bytes and use that instead of formatting, defer formatting to the consumer if at all

Answering your concern at the top, this leaves a slight misalignment. We're using Tokio macros to construct LogRecords which are the encoded parts and the Metadata i.e., callsite data. Your question is about whether can can compile-time disable these events. 

Yes, we can but not as written. Tokio macros have support for compile-time disablement base on Level, part of every callsite metadata. The fact that we have error_event! and info_event! is to suggest that we could choose to compile with only error events, but we'd have to modify ObservedEvent to treat those as the EventMessage::None value vs the EventMessage::Log event.

Issue #1812, opened 2026-01-16
Support for multiple pipeline groups and multiple pipelines


lquerel:
Until now, our OTAP DF Engine could only start a single pipeline configuration, even though internally our configuration model is already designed to support multiple pipeline configurations grouped into pipeline groups.

We need to add a new parameter `--config` or `-c`, mutually exclusive with `--pipeline` or `-p`, which will allow specifying an engine configuration declaring 0 to n pipeline groups, each composed of 0 to m pipeline configurations.

<img width="719" height="419" alt="Image" src="https://github.com/user-attachments/assets/a7a4efea-b0fd-434b-a1c5-d272203ef782" />

PR #1823, opened 2026-01-19
Support multiple groups and pipelines

lquerel:
Closes https://github.com/open-telemetry/otel-arrow/issues/1812

This PR adds support for multiple pipeline groups and multiple pipelines per group. It is still possible to start the engine with a single pipeline, but it is now also possible to start the engine with a configuration file describing one or more pipeline groups. An example configuration is included in this PR.

The controller documentation has been updated to provide more details. Optimal placement of the ITS and pipelines per NUMA node is not yet implemented. Named channels are not implemented yet.

I did not observe any performance regression with this PR.

### DIFF
--- a/rust/otap-dataflow/README.md
+++ b/rust/otap-dataflow/README.md
@@ -87,7 +87,7 @@ the many N-to-1 relationships expressed within an OTAP request.
 
 ### Controller and Engine
 
-See the controller and engine crate READMEs: 
+See the controller and engine crate READMEs:
 
 - [controller](./crates/controller/README.md).
 - [engine](./crates/engine/README.md),
--- a/rust/otap-dataflow/crates/controller/README.md
+++ b/rust/otap-dataflow/crates/controller/README.md
@@ -82,7 +82,7 @@ be explicitly declared in the pipeline topology.
 ### NUMA-Aware Controller
 
 Future versions of the controller are expected to become fully NUMA-aware.
-This includes:      
+This includes:
 
 - allocating pipelines on specific NUMA nodes,
 - pinning threads accordingly,
@@ -95,7 +95,7 @@ multi-socket systems.
 ### Morsel-Driven Parallelism
 
 In the OTAP Dataflow Engine, **batches** are the natural equivalent of
-“morsels”. Future designs may introduce batching at the ingress stage, before
+"morsels". Future designs may introduce batching at the ingress stage, before
 data traverses load-balancing boundaries such as per-core queues or MPMC named
 channels. This enables a form of **morsel-driven parallelism**, where:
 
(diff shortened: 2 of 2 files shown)


lalitb:
Thanks for the PR. Minor question on `core_count` with multiple pipelines: I noticed the sample config uses `core_set` with explicit non-overlapping ranges for each pipeline.

1. Is `core_set` the recommended approach when running multiple pipelines that need core isolation?
2. When using `core_count` across multiple pipelines, is it expected that they may end up sharing the same cores?

rust/otap-dataflow/crates/config/src/engine.rs
```diff
@@ -87,6 +76,83 @@ impl EngineConfig {
...
+        })?;
+        Self::from_yaml(&contents)
+    }
+
+    /// Load an [`EngineConfig`] from a file, automatically detecting the format based on file extension.
+    ///
+    /// Supports:
+    /// - JSON files: `.json`
+    /// - YAML files: `.yaml`, `.yml`
+    pub fn from_file<P: AsRef<Path>>(path: P) -> Result<Self, Error> {
```
albertlockett, line 114: In the future, maybe we'd want to consider the config crate for this? https://crates.io/crates/config . It might reduce some of the boilerplate loading config

lquerel: There is indeed this repetition issue, but at the same time there is also this error reporting problem described here: [https://github.com/open-telemetry/otel-arrow/issues/1832](https://github.com/open-telemetry/otel-arrow/issues/1832). I need to test this crate `config` and other approaches to see what will help resolve a number of issues around configuration parsing.

lquerel:
> @lalitb Thanks for the PR. Minor question on `core_count` with multiple pipelines: I noticed the sample config uses `core_set` with explicit non-overlapping ranges for each pipeline.
> 
>     1. Is `core_set` the recommended approach when running multiple pipelines that need core isolation?
> 
>     2. When using `core_count` across multiple pipelines, is it expected that they may end up sharing the same cores?

This part of the system still needs improvement. I would like to reach a situation where:
- Either the **number of cores** (without specific allocation) to use is defined in the pipeline configuration and the controller decides on the optimal placement.
* Or the exact core allocation is defined in the pipeline configuration and the controller follows the configuration strictly.

I created the following GH issue for tracking: https://github.com/open-telemetry/otel-arrow/issues/1837

rust/otap-dataflow/crates/config/src/engine.rs
albertlockett: Sounds good. One thing that's nice about this `config` crate is that it's possible to override the config file with environment variables:
https://docs.rs/config/0.15.19/config/#example

This can be helpful for users who have some dynamic config, but don't want to set up the machinery to deploy a templated config file.

jmacd: This sounds good. Whatever we do, let's not invent a crate for configuration 😂

jmacd (approved):
Question: It's not obvious from the PR how the distinction between single-pipeline configuration and pipeline-group configuration is done. I see the EngineConfig type already existed, so I suspect that's making it hard to see how it all connects.

lquerel:
> @jmacd Question: It's not obvious from the PR how the distinction between single-pipeline configuration and pipeline-group configuration is done. I see the EngineConfig type already existed, so I suspect that's making it hard to see how it all connects.

For now, I have only added an extra parameter to the command line (`--config`, `-c`) to allow specifying the global engine configuration, and I have kept the parameter (`--pipeline`, `-p`) for starting a single pipeline.

rust/otap-dataflow/crates/config/src/engine.rs
lquerel: I agree. I will check this crate and validate the quality of the error messages

PR #1824, opened 2026-01-19
fix: Always materialize the parent IDs when we transform attributes

### LINKED ISSUE #966: `transform_attributes` could create an invalid sequence of `parent_ids` using `rename` with no deletes
Currently `transform_attributes` will only remove the quasi-delta encoding from `parent_id` column if there were some deleted rows. We're not materializing the parent IDs if there were some only replacements.

https://github.com/open-telemetry/otel-arrow/blob/d4879f499c1819a268de6bb399c842db626b7a4d/rust/otel-arrow-rust/src/otap/transform.rs#L556-L569

The assumption was that deletes would be the only way for this method to create invalid sequence of quasi-delta encoding (see [here](https://github.com/open-telemetry/otel-arrow/blob/d4879f499c1819a268de6bb399c842db626b7a4d/rust/otel-arrow-rust/src/otap/transform.rs#L2870-L2885) for an explanation of how this could happen).

Thinking about this a bit more deeply, I'm not convinced this assumption is correct .. for example if there was a sequence of records like:
```
(type, key, value, parent_id)
str, "k1", "a", 0,
str, "k1", "a", 1,
str, "k2", "a", 0,
str, "k1", "a", 0,
str, "k1", "a", 1,
```
If the parent IDs were quasi-delta encoded, this would represent a sequence actual IDs like `0, 1, 0, 0, 1`.

If we rename the attribute "k2" to "k1", the sequence becomes:
```
(type, key, value, parent_id)
str, "k1", "a", 0,
str, "k1", "a", 1,
str, "k1", "a", 0,
str, "k1", "a", 0,
str, "k1", "a", 1,
```
And then if the parent IDs are considered to be quasi-delta encoded, this sequence now represents plain encoded IDs of `0, 1, 1, 1, 2`. This is not correct!

Note that the batch in this example is a bit unusual -- typically when the parent IDs are quasi-delta encoded, by default we always sort by the key column. But a user _could_ technically create a batch like this if the second sequence of "k1" was created by a previous replacement of a key that would get sorted after "k2". e.g. user had previously renamed "k3" -> "k1", where "k3" was following "k2" -- so even though it's unusual, we shouldn't assume it would _never_ happen.

**To fix:**

The safest & easiest way to avoid this is probably to always materialize the parent IDs when we transform attributes. Although, we could actually do this in a more optimized way by checking if the rows bordering the contiguous ranges we transform have the same type/key/value, and if this never happens, there's no reason to materialize the IDs in any case.

Apostlex0:
fixes #966

### DIFF
--- a/go/go.mod
+++ b/go/go.mod
@@ -2,7 +2,7 @@ module github.com/open-telemetry/otel-arrow/go
 
 go 1.24.0
 
-toolchain go1.25.5
+toolchain go1.25.6
 
 require (
 	github.com/HdrHistogram/hdrhistogram-go v1.2.0
--- a/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/arrow_records_encoder.rs
+++ b/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/arrow_records_encoder.rs
@@ -1012,12 +1012,16 @@ mod tests {
                         "'su root' failed for lonvick on /dev/pts/8".to_string()
                     ))
                 );
+                assert_eq!(
+                    log1_attrs.get("syslog.app_name"),
+                    Some(&AttributeValue::String("su".to_string()))
+                );
 
-                // Check that we have exactly the expected number of attributes (5 for RFC3164)
+                // Check that we have exactly the expected number of attributes (6 for RFC3164 with app_name)
                 assert_eq!(
                     log1_attrs.len(),
-                    5,
-                    "Log 1 should have exactly 5 attributes, got {}",
+                    6,
+                    "Log 1 should have exactly 6 attributes, got {}",
                     log1_attrs.len()
                 );
 
@@ -1049,12 +1053,20 @@ mod tests {
                         "Application started successfully".to_string()
                     ))
                 );
+                assert_eq!(
+                    log2_attrs.get("syslog.app_name"),
+                    Some(&AttributeValue::String("app".to_string()))
+                );
+                assert_eq!(
+                    log2_attrs.get("syslog.process_id"),
+                    Some(&AttributeValue::Integer(1234))
+                );
 
-                // Check that we have exactly the expected number of attributes (5 for RFC3164)
+                // Check that we have exactly the expected number of attributes (7 for RFC3164 with app_name and proc_id)
                 assert_eq!(
                     log2_attrs.len(),
-                    5,
-                    "Log 2 should have exactly 5 attributes, got {}",
+                    7,
+                    "Log 2 should have exactly 7 attributes, got {}",
                     log2_attrs.len()
                 );
 
@@ -1086,12 +1098,16 @@ mod tests {
                         "Kernel panic - not syncing: VFS".to_string()
                     ))
                 );
+                assert_eq!(
+                    log3_attrs.get("syslog.app_name"),
+                    Some(&AttributeValue::String("kernel".to_string()))
+                );
 
-                // Check that we have exactly the expected number of attributes (5 for RFC3164)
+                // Check that we have exactly the expected number of attributes (6 for RFC3164 with app_name)
                 assert_eq!(
                     log3_attrs.len(),
-                    5,
-                    "Log 3 should have exactly 5 attributes, got {}",
+                    6,
+                    "Log 3 should have exactly 6 attributes, got {}",
                     log3_attrs.len()
                 );
             }
@@ -1969,12 +1985,16 @@ mod tests {
                         "'su root' failed for lonvick on /dev/pts/8".to_string()
                     ))
                 );
+                assert_eq!(
+                    log2_attrs.get("syslog.app_name"),
+                    Some(&AttributeValue::String("su".to_string()))
+                );
 
-                // Check that we have exactly the expected number of attributes for RFC3164 (5)
+                // Check that we have exactly the expected number of attributes for RFC3164 (6 with app_name)
                 assert_eq!(
                     log2_attrs.len(),
-                    5,
-                    "Log 2 should have exactly 5 attributes, got {}",
+                    6,
+                    "Log 2 should have exactly 6 attributes, got {}",
                     log2_attrs.len()
                 );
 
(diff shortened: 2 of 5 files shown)


Apostlex0:
Hi @albertlockett ready for review!

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3486,6 +3484,155 @@ mod test {
...
+            Field::new(consts::PARENT_ID, DataType::UInt16, false).with_plain_encoding(),
+            Field::new(consts::ATTRIBUTE_TYPE, DataType::UInt8, false),
+            Field::new(consts::ATTRIBUTE_KEY, DataType::Utf8, false),
+            Field::new(consts::ATTRIBUTE_STR, DataType::Utf8, true),
+        ]));
+
+        let input = RecordBatch::try_new(
+            schema.clone(),
+            vec![
+                Arc::new(UInt16Array::from_iter_values(vec![0, 1, 0, 0, 1])),
```
albertlockett: I wonder if we should craft a sequence of IDs that happen to be different when we remove the transport optimized encoding. That might make the test case just a bit more clear
```suggestion
                Arc::new(UInt16Array::from_iter_values(vec![1, 1, 1, 1, 1]))
```

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3486,6 +3484,155 @@ mod test {
...
+                Arc::new(StringArray::from_iter_values(vec!["a", "a", "a", "a", "a"])),
+            ],
+        )
+        .unwrap();
+
+        let expected = RecordBatch::try_new(
+            expected_schema,
+            vec![
+                // Must remain the correct plain IDs after rename (i.e., materialized before rename):
+                Arc::new(UInt16Array::from_iter_values(vec![0, 1, 0, 0, 1])),
```
albertlockett:
```suggestion
                Arc::new(UInt16Array::from_iter_values(vec![1, 2, 1, 1, 2])),
```

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3486,6 +3484,155 @@ mod test {
...
+            // note: absence of encoding metadata means we assume it's quasi-delta encoded
+            Field::new(consts::PARENT_ID, DataType::UInt16, false),
+            Field::new(consts::ATTRIBUTE_TYPE, DataType::UInt8, false),
+            Field::new(consts::ATTRIBUTE_KEY, DataType::Utf8, false),
+            Field::new(consts::ATTRIBUTE_STR, DataType::Utf8, true),
+        ]));
+
+        // After rename k2 -> k1, keys become all k1, if we didn't materialize first, the
+        // quasi-delta decoding would incorrectly yield 0,1,1,1,2 but we want the correct plain
+        // IDs to remain 0,1,0,0,1 and the parent_id column to be marked as plain.
```
albertlockett:
```suggestion
        // quasi-delta decoding would incorrectly yield 1,2,3,4,5 but we want the correct plain
        // IDs to remain 1,2,1,1,2 and the parent_id column to be marked as plain.
```

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3486,6 +3484,155 @@ mod test {
...
+                DataType::Dictionary(Box::new(DataType::UInt8), Box::new(DataType::Utf8)),
+                false,
+            ),
+            Field::new(consts::ATTRIBUTE_STR, DataType::Utf8, true),
+        ]));
+
+        let input = RecordBatch::try_new(
+            schema.clone(),
+            vec![
+                Arc::new(UInt16Array::from_iter_values(vec![0, 1, 0, 0, 1])),
```
albertlockett:
```suggestion
                Arc::new(UInt16Array::from_iter_values(vec![1, 1, 1, 1, 1])),
```

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3486,6 +3484,155 @@ mod test {
...
+                )),
+                Arc::new(StringArray::from_iter_values(vec!["a", "a", "a", "a", "a"])),
+            ],
+        )
+        .unwrap();
+
+        let expected = RecordBatch::try_new(
+            expected_schema,
+            vec![
+                Arc::new(UInt16Array::from_iter_values(vec![0, 1, 0, 0, 1])),
```
albertlockett:
```suggestion
                Arc::new(UInt16Array::from_iter_values(vec![1, 2, 1, 1, 2])),
```

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3355,6 +3351,7 @@ mod test {
     }
 
     #[test]
+    #[ignore = "We can update this tests once we have a better method to materialize the parent_id than the current one"]
```
albertlockett: I think we can probably remove these test cases now, rather than just ignoring them

I might create a followup issue to add some more involved logic to check whether we'd end up merging adjacent ranges when renaming, and I can link this PR with instructions on the issue to restore the tests.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3419,6 +3416,7 @@ mod test {
...
```
albertlockett: Similar comment to https://github.com/open-telemetry/otel-arrow/pull/1824/files#r2705665025 . can probably be removed

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
Apostlex0: I think using [1,1,1,1,1] should be good enough since if we are going to improve the logic later on, the tests would need to be updated anyways

Apostlex0:
@albertlockett should be fine now!

albertlockett:
Thank you @Apostlex0!

PR #1825, opened 2026-01-19
chore(deps): update zip crate to 7.1.0

albertlockett:
supersedes https://github.com/open-telemetry/otel-arrow/pull/1822 which was failing on a cargo deny license check for an optional dependency (bzip2)


albertlockett:
ugh, looks like this is still failing the cargo deny check b/c we run this with --all-features which will include the bzip feature on the zip crate. I'll come back to this later.

PR #1836, opened 2026-01-20
feat: filter tests updated to use OPL parser

### LINKED ISSUE #1790: OPL Parser: null support
Add support for parsing null literal from OPL. This should get parsed into a `StaticScalarExpression::Null`.

Once support is added, update the tests in columnar query engine `pipeline::filter` related to null filter handling to use OPL Parser and this syntax.

Apostlex0:
closes #1790


Apostlex0:
@albertlockett the null literal was already being parsed as you wanted, so i just updated the filter tests similar to what we did before by passing a &str parameter("null" for opl and "string(null) for kql") in the shared test helper.

PR #1839, opened 2026-01-20
Add ACK/NACK support to debug_processor

AaronRM:
Ensuring proper forwarding of ACK and NACK control messages upstream from the debug_processor and add tests to verify this behavior.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/debug_processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/debug_processor.rs
@@ -246,22 +246,45 @@ impl local::Processor<OtapPdata> for DebugProcessor {
                             .await?;
                     }
                     NodeControlMsg::Ack(ackmsg) => {
-                        if let Ok(dd) = DebugCallData::try_from(ackmsg.calldata.clone()) {
-                            debug_output
-                                .output_message(&format!("ACK received after {:?}\n", dd.elapsed()))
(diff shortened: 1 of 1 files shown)


jmacd (approved):
Yes! thank you.

rust/otap-dataflow/crates/otap/src/debug_processor.rs
```diff
@@ -245,16 +246,25 @@ impl local::Processor<OtapPdata> for DebugProcessor {
                             .await?;
                     }
                     NodeControlMsg::Ack(ackmsg) => {
-                        let dd: DebugCallData = ackmsg.calldata.try_into()?;
-                        debug_output
-                            .output_message(&format!("ACK received after {:?}\n", dd.elapsed()))
-                            .await?;
+                        if let Ok(dd) = DebugCallData::try_from(ackmsg.calldata.clone()) {
```
lalitb: nit - this will swallow the malformed `calldata` - in this case `try_from` will fail but the error will get swallowed by `Ok(dd)`. Earlier this was propagated back using `try_into()?`. Maybe we can log the decode error so bad calldata isn't invisible? Same applied to NACK path.

AaronRM: Fixed

PR #1841, opened 2026-01-20
Internal logging provider setup; console_async support

jmacd:

Implements 4 of the 5 ProviderMode values.

Uses the ObservedStateStore's associated thread and channel to process console_async messages.

Replaces most of #1771.

Undoes portions of #1818:

- ObservedEvent is an enum for Engine, Log events
- Engine events return to `Option<String>` message, no structured message
- Removes info_event! and error_event! structured message constructor macros
- Moves LogRecord::Serialize support to where it's used

Adds new LoggingProviders selector `admin` to configure how the admin threads use internal logging. The new setting defaults to ConsoleDirect, i.e., the admin components will use synchronous console logging.

Configures the Tokio tracing subscriber globally, in engine threads, and in admin threads according to the ProviderMode.

The asynchronous tracing subscriber (which sends to console_async; will send to ITS in the future) uses the `internal` provider mode itself as a fallback. However, it does this directly, choosing the Noop or ConsoleDirect modes, OpenTelemetry mode is not supported here.

~Resolves a TODO about inconsistency in the otel_xxx! macros. These now support full Tokio syntax following raw_error!~
EDIT: portions of this PR were moved into #1843. This PR removes the top-level `log` dependency.

### DIFF
--- a/rust/otap-dataflow/crates/telemetry/src/event.rs
+++ b/rust/otap-dataflow/crates/telemetry/src/event.rs
@@ -34,10 +34,6 @@ impl ObservedEventReporter {
     }
 
     fn observe(&self, event: ObservedEvent) {
-        // Note: we expect fallback_mode is equal to providers.internal
-        // so it's in the set (Noop, ConsoleDirect, OpenTelemetry).
-        //
-        // OpenTelemetry is silently ignored in this location.
         match self.policy.blocking_timeout {
             None => match self.sender.try_send(event) {
                 Ok(_) => {}
@@ -47,10 +43,10 @@ impl ObservedEventReporter {
                     }
                     match err {
                         flume::TrySendError::Full(event) => {
-                            crate::raw_error!("Timeout sending observed event", event = ?event);
+                            crate::raw_error!("Channel full, dropping observed event", event = ?event);
                         }
                         flume::TrySendError::Disconnected(event) => {
-                            crate::raw_error!("Disconnect sending observed event", event = ?event);
+                            crate::raw_error!("Disconnect, dropping observed event", event = ?event);
                         }
                     }
                 }
@@ -63,10 +59,10 @@ impl ObservedEventReporter {
                     }
                     match err {
                         flume::SendTimeoutError::Timeout(event) => {
-                            crate::raw_error!("Timeout sending observed event", event = ?event);
+                            crate::raw_error!("Timeout, dropping observed event", event = ?event);
                         }
                         flume::SendTimeoutError::Disconnected(event) => {
-                            crate::raw_error!("Disconnect sending observed event", event = ?event);
+                            crate::raw_error!("Disconnect, dropping observed event", event = ?event);
                         }
                     }
                 }


rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -114,6 +115,167 @@ Note that the ITS and ConsoleAsync modes share a the same provider
...
+┌─────────────────────┐
+│      Console        │
+│      (stdout)       │
+└─────────────────────┘
+```
+
+### ConsoleAsync Provider
+
+Asynchronous console output via a bounded channel. Non-blocking for
+the producing thread; logs may be dropped if the channel is full.
```
cijothomas:
```suggestion
the producing thread; logs will be dropped if the channel is full.
```

rust/otap-dataflow/crates/telemetry/src/event.rs
```diff
@@ -4,17 +4,105 @@
...
+        self.observe(ObservedEvent::Engine(event))
+    }
+
+    /// Report a log event.
+    pub fn log(&self, event: LogEvent) {
+        self.observe(ObservedEvent::Log(event))
+    }
+
+    fn observe(&self, event: ObservedEvent) {
+        let sent = self.sender.send_timeout(event, self.timeout);
```
utpilla: If the channel is full, this would [block the thread](https://docs.rs/flume/latest/flume/struct.Sender.html#method.send_timeout:~:text=If%20the%20channel%20is%20bounded%20and%20is%20full%2C%20this%20method%20will%20block%20until%20space%20is%20available%2C%20the%20timeout%20has%20expired%2C%20or%20all%20receivers%20have%20been%20dropped.) until the channel has space or the timeout expires? Since this would be called by the engine threads, do we want a non-blocking behavior instead?

rust/otap-dataflow/crates/telemetry/src/lib.rs
```diff
@@ -98,50 +105,145 @@ pub struct InternalTelemetrySystem {
...
         let dispatcher = Arc::new(metrics::dispatcher::MetricsDispatcher::new(
             telemetry_registry.clone(),
             config.reporting_interval,
         ));
 
         // 2. Create OTel SDK providers
-        let otel_client = otel_sdk::OpentelemetryClient::new(config)?;
+        // Logger provider is only needed for OpenTelemetry mode
+        let needs_logger_provider = config.logs.providers.global == ProviderMode::OpenTelemetry
+            || config.logs.providers.engine == ProviderMode::OpenTelemetry;
```
lalitb: We should also be checking for `admin` mode here?

rust/otap-dataflow/crates/telemetry/src/event.rs
```diff
@@ -4,17 +4,105 @@
...
+        if self.fallback_mode != ProviderMode::ConsoleDirect {
```
utpilla: This feels odd. The controller provides engine's `ProviderMode` as the value for `fallback_mode` which would ideally never be `ConsoleDirect`. That means we would never log these errors for most configurations. I think we should be providing the `ProviderMode` of `admin` and not the `engine` as the `fallback_mode`.

From `run_forever` method in `controller.rs`:

```rust
// Create the observed state store first - the telemetry system needs its reporter
let obs_state_store =
    ObservedStateStore::new(&engine_settings.observed_state, telemetry_config.logs.providers.engine);
let obs_evt_reporter = obs_state_store.reporter();
```

rust/otap-dataflow/crates/telemetry/src/lib.rs
utpilla: I think instead of doing that, we should be validating that `admin` mode cannot have `OpenTelemetry` as their `ProviderMode`.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
```diff
@@ -144,6 +155,16 @@ impl LogsConfig {
                 ),
             });
         }
+        // Admin provider cannot use ConsoleAsync because the observed-state-store
+        // runs in an admin thread and implements console_async logging. Using
+        // ConsoleAsync for admin would create a feedback loop.
+        if self.providers.admin == ProviderMode::ConsoleAsync {
```
utpilla, line 188: I think we don't want to allow `ITS` and `OpenTelemetry` as well for `admin`.

Related: https://github.com/open-telemetry/otel-arrow/pull/1841#discussion_r2710550742

rust/otap-dataflow/crates/telemetry/src/event.rs
```diff
@@ -4,17 +4,105 @@
...
+            }
+            Err(flume::SendTimeoutError::Disconnected(event)) => {
+                crate::raw_error!("Disconnect sending observed event", event = ?event);
+            }
+            Ok(_) => {}
+        };
+    }
+}
+
+/// The obseved event
```
lalitb:
```suggestion
/// The observed event
```

rust/otap-dataflow/crates/telemetry/src/lib.rs
lalitb: I’m okay with blocking admin/internal from using opentelemetry for now. If we ever need it, we can later change this to build the logger provider for those modes too. Main concern was the crash if anyone tries `admin: opentelemetry` in config , as we never build a logger provider in that case.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -439,59 +459,63 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
...
+            // entity is registered here for proper context tracking and set into thread-local storage
+            // in order to be accessible by all components within this thread.
+            let pipeline_entity_key = pipeline_context.register_pipeline_entity();
+            let _pipeline_entity_guard =
+                set_pipeline_entity_key(pipeline_context.metrics_registry(), pipeline_entity_key);
+
+            // Pin thread to specific core
+            if !core_affinity::set_for_current(core_id) {
+                // Continue execution even if pinning fails.
+                // This is acceptable because the OS will still schedule the thread, but performance may be less predictable.
```
lquerel: In `run_pipeline_thread`, we now call `tracing_setup.with_subscriber` before `core_affinity::set_for_current`. `with_subscriber` builds a new `Dispatch` before running the closure, so any allocations for filters/layers/providers happen pre-pin. If the thread runs on
  a different core during that window, NUMA‑aware allocators (like jemalloc and mimalloc) can place those tracing objects on the wrong node. This regression (pre‑pin allocations) already existed (in the previous PR, I missed it) because we created the span/logs before pinning, but this change widens the pre‑pin
  allocation window and amplifies the NUMA locality risk.

I suggest the following: pin immediately after thread start (at the beginning of this method), then call `with_subscriber` (and create the span) inside the pinned region so tracing allocations happen on the intended core.

rust/otap-dataflow/crates/controller/src/thread_task.rs
```diff
@@ -9,6 +9,8 @@ use std::thread;
 use tokio::{runtime::Builder as RtBuilder, task::LocalSet};
 use tokio_util::sync::CancellationToken;
 
+use otap_df_telemetry::TracingSetup;
+
 /// Handle to a task running on a dedicated thread.
```
lquerel: I didn't pay close attention to the description of this method in the previous PR, but I think we should clarify right at the start of the comment describing this method that it is only intended to create an OS thread and a single-threaded async runtime, **without thread pinning, dedicated to things like internal metrics aggregation or internal event processing**.

rust/otap-dataflow/crates/telemetry/src/event.rs
lquerel: This `send_timeout` is systematically used for all event reports. Since this is on the hot path, I think this is something to watch, as it likely forces Tokio to add a timeout to a timer queue. Another approach would be to use `try_send`, which will work in most cases, and only fall back to `send_timeout` when `try_send` fails.

rust/otap-dataflow/crates/telemetry/src/event.rs
```diff
@@ -4,17 +4,105 @@
...
+}
 
 /// An observed event emitted by the engine.
 #[derive(Debug, Clone, Serialize)]
-pub struct ObservedEvent {
+pub struct EngineEvent {
     // ---- Source identification ----
-    /// Unique key identifying the pipeline instance.
-    #[serde(skip_serializing)]
+    /// Unique key identifying the pipeline instance (None for global/controller-level events).
```
lquerel: is the comment still valid? I mean the None part.

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -114,6 +115,165 @@ Note that the ITS and ConsoleAsync modes share a the same provider
...
+│   Bounded Channel   │
+│   (flume::Sender)   │
+└──────────┬──────────┘
+           │
+┌─────────────────────────────────────────────────┐
+│         Internal Telemetry Pipeline             │
+│  ┌───────────────┐  ┌───────────┐  ┌─────────┐  │
+│  │  ITR Receiver │─>│ Processor │─>│Exporter │  │
+│  └───────────────┘  └───────────┘  └─────────┘  │
+└─────────────────────────────────────────────────┘
```
lquerel: Nice :-)

rust/otap-dataflow/docs/self_tracing_architecture.md
```diff
@@ -114,6 +115,165 @@ Note that the ITS and ConsoleAsync modes share a the same provider
...
+│                                                                     │
+│  ┌───────────────────────────┐                                      │
+│  │  Admin                    │  (e.g., for ConsoleAsync mode)       │
+│  │  Observer Thread          │                                      │
+│  │  ┌─────────────────────┐  │                                      │
+│  │  │ Uses ConsoleDirect  │  │                                      │
+│  │  │ or Noop             │  │                                      │
+│  │  └─────────────────────┘  │                                      │
+│  └───────────────────────────┘                                      │
+└─────────────────────────────────────────────────────────────────────┘
```
lquerel: Cool to see that !

lquerel (approved):
I think there are a few adjustments to make, but overall I'm fine with it. If my comments aren't clear enough, contact me on Slack. 
Thanks for the update on the documentation.

rust/otap-dataflow/crates/telemetry/src/event.rs
lquerel: I had also paused on this point. In addition to what I mentioned in another comment, I think that 1) we should first use `try_send`, and if that fails, we should then consult a policy, for example an enum with variants like timeout, drop, or block, and take the appropriate action based on that policy.

jmacd: Added a `SendPolicy` 

```
/// How to act when an asynchronous event can't be sent.
#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
pub struct SendPolicy {
    /// If set, wait for a timeout.
    pub blocking_timeout: Option<Duration>,

    /// If failed, issue a raw error to the console.
    pub console_fallback: bool,
}
```

and let the existing engine events block / console log, while ordinary logging events will not block and drop.

rust/otap-dataflow/crates/telemetry/src/lib.rs
jmacd: I removed the fallback_mode in favor of the new SendPolicy.

There isn't anything dangerous (that I know of) from using the OpenTelemetry provider anywhere. The existing validation check was out of date. It can be configured because it does not use the engine or async channel.

There was a point about how fallback_mode would disrespect this setting, it's fixed.

rust/otap-dataflow/crates/telemetry/src/event.rs
jmacd: That was a mistake. The fallback was meant to the the `providers.internal` mode, not the engine mode.

With the new SendPolicy, there's less confusion, no ProviderMode here.

rust/otap-dataflow/crates/config/src/pipeline/service/telemetry/logs.rs
jmacd: I think we only need to prevent a cycle. If the admin component uses ITS or OpenTelemetry it will not be logging "to itself".

rust/otap-dataflow/crates/controller/src/lib.rs
jmacd: Fixed.

rust/otap-dataflow/crates/controller/src/thread_task.rs
jmacd: Done.

rust/otap-dataflow/crates/telemetry/src/event.rs
jmacd: I've added a `policy` as discussed, which gives a configuration to decide whether to block (how long) and whether to print to the console when it fails. There is no fallback provider, now, only these two options.

rust/otap-dataflow/crates/telemetry/src/event.rs
jmacd: Out of date. Fixed.

rust/otap-dataflow/crates/config/src/observed_state.rs
```diff
@@ -13,15 +13,45 @@ pub struct ObservedStateSettings {
...
+                blocking_timeout: Some(Duration::from_millis(1)),
+                console_fallback: true,
+            },
+            logging_events: SendPolicy {
+                blocking_timeout: None,
+                console_fallback: false,
+            },
+        }
+    }
+}
```
jmacd, line 47: FYI

rust/otap-dataflow/crates/telemetry/src/event.rs
```diff
@@ -4,17 +4,118 @@
...
+        match self.policy.blocking_timeout {
+            None => match self.sender.try_send(event) {
+                Ok(_) => {}
+                Err(err) => {
+                    if !self.policy.console_fallback {
+                        return;
+                    }
+                    match err {
+                        flume::TrySendError::Full(event) => {
+                            crate::raw_error!("Timeout sending observed event", event = ?event);
```
utpilla: Update the error message to say that channel is full instead of timeout. Something like:

`Channel full, dropping observed event`

rust/otap-dataflow/crates/telemetry/src/event.rs
```diff
@@ -4,17 +4,118 @@
...
+    /// Report a log event.
+    pub fn log(&self, event: LogEvent) {
+        self.observe(ObservedEvent::Log(event))
+    }
+
+    fn observe(&self, event: ObservedEvent) {
+        // Note: we expect fallback_mode is equal to providers.internal
+        // so it's in the set (Noop, ConsoleDirect, OpenTelemetry).
+        //
+        // OpenTelemetry is silently ignored in this location.
```
utpilla: This comment should be removed now that we don't have `fallback_mode`.

rust/otap-dataflow/crates/telemetry/src/event.rs
jmacd: Done. (Updated all messages to say "dropping" e.g.,  instead of "disconnected sending"

rust/otap-dataflow/crates/telemetry/src/event.rs
jmacd: Fixed

PR #1842, opened 2026-01-20
Add try_send_message/try_send_method_to to EffectHandler for non-awaiting message sending

AaronRM:
Add `try_send_message`/`try_send_method_to` to EffectHandler for non-awaiting message sending

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/local/processor.rs
+++ b/rust/otap-dataflow/crates/engine/src/local/processor.rs
@@ -34,7 +34,7 @@
 
 use crate::control::{AckMsg, NackMsg};
 use crate::effect_handler::{EffectHandlerCore, TelemetryTimerCancelHandle, TimerCancelHandle};
-use crate::error::{Error, ProcessorErrorKind, TypedError};
+use crate::error::{Error, TypedError};
 use crate::local::message::LocalSender;
 use crate::message::Message;
 use crate::node::NodeId;
@@ -155,7 +155,9 @@ impl<PData> EffectHandler<PData> {
                 .send(data)
                 .await
                 .map_err(TypedError::ChannelSendError),
-            None => Err(self.no_default_port_error()),
+            None => Err(TypedError::Error(Error::NoDefaultOutPort {
+                node: self.processor_id(),
+            })),
         }
     }
 
@@ -173,19 +175,10 @@ impl<PData> EffectHandler<PData> {
     pub fn try_send_message(&self, data: PData) -> Result<(), TypedError<PData>> {
         match &self.default_sender {
             Some(sender) => sender.try_send(data).map_err(TypedError::ChannelSendError),
-            None => Err(self.no_default_port_error()),
+            None => Err(TypedError::Error(Error::NoDefaultOutPort {
+                node: self.processor_id(),
+            })),
         }
-    }
-
-    /// Creates an error for when no default output port is configured.
-    fn no_default_port_error<T>(&self) -> TypedError<T> {
-        TypedError::Error(Error::ProcessorError {
-            processor: self.processor_id(),
-            kind: ProcessorErrorKind::Configuration,
-            error: "Ambiguous default out port: multiple ports connected and no default configured"
-                .to_string(),
-            source_detail: String::new(),
-        })
     }
 
     /// Sends a message to a specific named out port.
@@ -205,7 +198,10 @@ impl<PData> EffectHandler<PData> {
                 .send(data)
                 .await
                 .map_err(TypedError::ChannelSendError),
-            None => Err(self.unknown_port_error(&port_name)),
+            None => Err(TypedError::Error(Error::UnknownOutPort {
+                node: self.processor_id(),
+                port: port_name,
+            })),
         }
     }
 
@@ -227,21 +223,11 @@ impl<PData> EffectHandler<PData> {
         let port_name: PortName = port.into();
         match self.msg_senders.get(&port_name) {
             Some(sender) => sender.try_send(data).map_err(TypedError::ChannelSendError),
-            None => Err(self.unknown_port_error(&port_name)),
+            None => Err(TypedError::Error(Error::UnknownOutPort {
+                node: self.processor_id(),
+                port: port_name,
+            })),
         }
-    }
-
-    /// Creates an error for when an unknown output port is specified.
-    fn unknown_port_error<T>(&self, port_name: &PortName) -> TypedError<T> {
-        TypedError::Error(Error::ProcessorError {
-            processor: self.processor_id(),
-            kind: ProcessorErrorKind::Configuration,
-            error: format!(
-                "Unknown out port '{port_name}' for node {}",
-                self.processor_id()
-            ),
-            source_detail: String::new(),
-        })
     }
 
     /// Print an info message to stdout.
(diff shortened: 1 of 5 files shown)


rust/otap-dataflow/crates/engine/src/local/processor.rs
```diff
@@ -195,6 +188,35 @@ impl<PData> EffectHandler<PData> {
         }
     }
 
+    /// Attempts to send a message without blocking.
```
utpilla:
```suggestion
    /// Attempts to send a message without awaiting.
```

rust/otap-dataflow/crates/engine/src/local/processor.rs
```diff
@@ -195,6 +188,35 @@ impl<PData> EffectHandler<PData> {
...
+    ///
+    /// Unlike `send_message`, this method returns immediately if the downstream
+    /// channel is full, allowing the caller to handle backpressure without blocking.
```
utpilla: I think using the term "blocking" would suggest that the OS thread is blocked but that's not the case with `send_message`. It doesn't block the thread. It simply yields the control back to the async runtime when the channel is full.

```suggestion
    /// Unlike `send_message`, this method returns immediately if the downstream
    /// channel is full, allowing the caller to handle backpressure without awaiting.
```

AaronRM: Fair point. Fixed.

rust/otap-dataflow/crates/engine/src/local/processor.rs
```diff
@@ -376,4 +399,67 @@ mod tests {
...
+        let _ = senders.insert("b".into(), LocalSender::mpsc(b_tx));
+
+        let (_metrics_rx, metrics_reporter) = MetricsReporter::create_new_and_receiver(1);
+        // No default port specified with multiple ports = ambiguous
+        let eh = EffectHandler::new(test_node("proc"), senders, None, metrics_reporter);
+
+        // Should return configuration error when no default sender
+        let result = eh.try_send_message(99);
+        assert!(matches!(result, Err(TypedError::Error(_))));
+    }
```
utpilla: I'd suggest make the tests non-async especially since you're trying to test a non-async API.

```suggestion
    #[test]
    fn effect_handler_try_send_message_success() {
        let (tx, rx) = channel::<u64>(10);
        let mut senders = HashMap::new();
        let _ = senders.insert("out".into(), LocalSender::mpsc(tx));

        let (_metrics_rx, metrics_reporter) = MetricsReporter::create_new_and_receiver(1);
        let eh = EffectHandler::new(
            test_node("proc"),
            senders,
            Some("out".into()),
            metrics_reporter,
        );

        // Should succeed when channel has capacity
        assert!(eh.try_send_message(42).is_ok());
        assert_eq!(rx.try_recv().unwrap(), 42);
    }

    #[test]
    fn effect_handler_try_send_message_channel_full() {
        // Create a channel with capacity 1
        let (tx, _rx) = channel::<u64>(1);
        let mut senders = HashMap::new();
        let _ = senders.insert("out".into(), LocalSender::mpsc(tx));

        let (_metrics_rx, metrics_reporter) = MetricsReporter::create_new_and_receiver(1);
        let eh = EffectHandler::new(
            test_node("proc"),
            senders,
            Some("out".into()),
            metrics_reporter,
        );

        // First send should succeed
        assert!(eh.try_send_message(1).is_ok());

        // Second send should fail with Full since channel capacity is 1
        let result = eh.try_send_message(2);
        assert!(matches!(
            result,
            Err(TypedError::ChannelSendError(SendError::Full(2)))
        ));
    }

    #[test]
    fn effect_handler_try_send_message_no_default_sender() {
        let (a_tx, _a_rx) = channel::<u64>(10);
        let (b_tx, _b_rx) = channel::<u64>(10);

        let mut senders = HashMap::new();
        let _ = senders.insert("a".into(), LocalSender::mpsc(a_tx));
        let _ = senders.insert("b".into(), LocalSender::mpsc(b_tx));

        let (_metrics_rx, metrics_reporter) = MetricsReporter::create_new_and_receiver(1);
        // No default port specified with multiple ports = ambiguous
        let eh = EffectHandler::new(test_node("proc"), senders, None, metrics_reporter);

        // Should return configuration error when no default sender
        let result = eh.try_send_message(99);
        assert!(matches!(result, Err(TypedError::Error(_))));
    }
```

rust/otap-dataflow/crates/engine/src/local/processor.rs
```diff
@@ -195,6 +188,35 @@ impl<PData> EffectHandler<PData> {
...
+    /// Returns a [`TypedError::ChannelSendError`] containing [`SendError::Full`] if the
+    /// channel is full, or [`SendError::Closed`] if the channel is closed.
+    /// Returns a [`TypedError::Error`] if no default port is configured.
+    #[inline]
+    pub fn try_send_message(&self, data: PData) -> Result<(), TypedError<PData>> {
+        match &self.default_sender {
+            Some(sender) => sender.try_send(data).map_err(TypedError::ChannelSendError),
+            None => Err(self.no_default_port_error()),
+        }
+    }
```
lalitb: nit: Should we also add a `try_send_message_to` counterpart for named ports (mirroring the `send_message` / `send_message_to` pair)? Feel free to ignore if this was intentionally omitted - should be fine to defer until there's a concrete use case.

lalitb (approved):

rust/otap-dataflow/crates/engine/src/local/processor.rs
AaronRM: I originally created this PR because I needed a `try_send_message` as a prereq for another PR, but you raise a good point. I'll get clarity on [the correct scope along with Josh's question](https://github.com/open-telemetry/otel-arrow/pull/1842#pullrequestreview-3688670431).

AaronRM:
> Should the same interface be added to the `shared` component definitions? I would like guidance from @lquerel on whether we need (local/shared)x(receiver/processor) i.e., variants of this new interface.

I've updated this PR to include additions of `try_send_message` and `try_send_message_to` to (local/shared)x(receiver/processor) to keep the API's consistent all around. Awaiting feedback from @lquerel.

lquerel:
@jmacd and @AaronRM I think we need to offer the same interface between the local version and the shared version of the effect handler.

rust/otap-dataflow/crates/engine/src/shared/receiver.rs
```diff
@@ -157,17 +157,39 @@ impl<PData> EffectHandler<PData> {
...
+    #[inline]
+    pub fn try_send_message(&self, data: PData) -> Result<(), TypedError<PData>> {
+        match &self.default_sender {
+            Some(sender) => sender.try_send(data).map_err(TypedError::ChannelSendError),
+            None => Err(self.no_default_port_error()),
+        }
+    }
+
+    /// Creates an error for when no default output port is configured.
+    fn no_default_port_error<T>(&self) -> TypedError<T> {
```
jmacd: I think given that there are now two copies of no_default/unknown port error helpers, it would make sense to add new enum variants to crates/engine/src/error.rs so that this reads more like TypedError::Error(Error::NoDefaultPort{ node: self.receiver_id() })

or

TypedError::Error(Error::no_default_port(self.receiver_id())

with a helper in `impl Error` like `pub fn no_default_port(node: NodeId) -> Self {  ... }`.

(Additional commentary: this `source_detail` is begging for a macro to fill it in automatically.)

AaronRM: Good suggestion, fixed.

PR #1843, opened 2026-01-20
Replace uses of log:: with otel_ macros in crates/engine, crates/otap

jmacd:

Overlaps with #1841 by copying the file crates/telemetry/src/internal_events.rs to extend the otel_xxx macros to full Tokio syntax, to replace uses of log formatting as needed.

After this, #1841 can remove "log" from the workspace Cargo.toml b/c crates/state will have the remaining "log" references fixed there.

### DIFF
--- a/rust/otap-dataflow/crates/telemetry/src/internal_events.rs
+++ b/rust/otap-dataflow/crates/telemetry/src/internal_events.rs
@@ -36,7 +36,7 @@ pub mod _private {
 #[macro_export]
 macro_rules! otel_info {
     ($name:expr $(, $($fields:tt)*)?) => {
-        $crate::_private::info!(name: $name, target: env!("CARGO_PKG_NAME"), { $($($fields)*)? });
+        $crate::_private::info!(name: $name, target: env!("CARGO_PKG_NAME"), { $($($fields)*)? }, "");
     };
 }
 
@@ -55,7 +55,7 @@ macro_rules! otel_info {
 #[macro_export]
 macro_rules! otel_warn {
     ($name:expr $(, $($fields:tt)*)?) => {
-        $crate::_private::warn!(name: $name, target: env!("CARGO_PKG_NAME"), { $($($fields)*)? });
+        $crate::_private::warn!(name: $name, target: env!("CARGO_PKG_NAME"), { $($($fields)*)? }, "");
     };
 }
 
@@ -74,7 +74,7 @@ macro_rules! otel_warn {
 #[macro_export]
 macro_rules! otel_debug {
     ($name:expr $(, $($fields:tt)*)?) => {
-        $crate::_private::debug!(name: $name, target: env!("CARGO_PKG_NAME"), { $($($fields)*)? });
+        $crate::_private::debug!(name: $name, target: env!("CARGO_PKG_NAME"), { $($($fields)*)? }, "");
     };
 }
 
@@ -93,7 +93,7 @@ macro_rules! otel_debug {
 #[macro_export]
 macro_rules! otel_error {
     ($name:expr $(, $($fields:tt)*)?) => {
-        $crate::_private::error!(name: $name, target: env!("CARGO_PKG_NAME"), { $($($fields)*)? });
+        $crate::_private::error!(name: $name, target: env!("CARGO_PKG_NAME"), { $($($fields)*)? }, "");
     };
 }
 


rust/otap-dataflow/crates/telemetry/src/internal_events.rs
```diff
@@ -33,15 +33,10 @@ pub mod _private {
...
-///
-/// TODO: Update to use valueset! for full `tracing` syntax, see raw_error!
 #[macro_export]
 macro_rules! otel_info {
-    ($name:expr $(,)?) => {
-        $crate::_private::info!( name: $name, target: env!("CARGO_PKG_NAME"), name = $name, "");
-    };
-    ($name:expr, $($key:ident = $value:expr),+ $(,)?) => {
-        $crate::_private::info!(name: $name, target: env!("CARGO_PKG_NAME"), name = $name, $($key = $value),+, "");
+    ($name:expr $(, $($fields:tt)*)?) => {
```
lalitb, line 38: Just curious why is the two-arm macro removed ? I believe it was required to enforce key = value structure and keep compiler errors clear at the call site. This change will now allow calls like:
```
otel_info!("x", { not-valid-data });
```

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
```diff
@@ -33,15 +33,10 @@ pub mod _private {
...
+        $crate::_private::info!(name: $name, target: env!("CARGO_PKG_NAME"), { $($($fields)*)? }, "");
```
lalitb: nit - In later PRs, the  trailing `""` can be removed. This was also removed in the otel-rust repo now - https://github.com/open-telemetry/opentelemetry-rust/pull/3317

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
lalitb: Looking closer, I believe it will still fail for above example with compile time error, maybe with cryptic error. Given that, the current approach seems acceptable.

lalitb (approved):
LGTM. There are still some `log::` reference in the code, and CI is failing for it.

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
jmacd: I think I don't understand how this is avoided or what it looks like when you do this.
I didn't realize anything was lost, just wanted to be able to write e.g., `error = ?err`.

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
jmacd: Fixed (w/ ref to that PR) thanks.

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
jmacd: I tried this like:

```
        let error = Error::ShutdownError("panicked".into()).to_string();
        raw_error!("very important! logging failed", error);
```

and it prints IMO reasonably with a key named "error".

2026-01-20T22:45:22.055Z  ERROR  otap-df-telemetry::very important! logging failed (crates/telemetry/src/internal_events.rs:157):  [error=Shutdown error: panicked]

jmacd: Removing the "" has introduced new complications.

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
jmacd: That "fix" appears to have cause the other problem that you mentioned, so I put back the "". This appears to require all arguments to have simple key or key=value form. (I'm just trying to follow what `tracing` supports, not intentionally changing anything.)

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
lalitb: I mean with your change to make the format more flexible, the error for the above example would be caught deep inside the tracing macro internal, and the error would be bit cryptic. Previously, the stricter pattern caught these errors earlier with clearer messages. But should be fine to keep as it is. 

That said, this should be fine to keep as-is since it's still a compile-time error. One suggestion: consider adding a doc comment to each macro documenting the supported syntax: 

```rust
/// Logs an INFO-level event with structured fields.                                                                                                                                           
  ///                                                                                                                                                                                            
  /// # Supported field syntax                                                                                                                                                                   
  /// - `key = value` - explicit value                                                                                                                                                           
  /// - `key = ?value` - debug format (`{:?}`)                                                                                                                                                   
  /// - `key = %value` - display format (`{}`)                                                                                                                                                   
  /// - `?key` - shorthand for `key = ?key`                                                                                                                                                      
  /// - `%key` - shorthand for `key = %key`                                                                                                                                                      
  ///                                                                                                                                                                                            
  /// # Examples                                                                                                                                                                                 
  /// ```                                                                                                                                                                                        
  /// otel_info!("exporter.start", endpoint = cfg.endpoint);                                                                                                                                     
  /// otel_info!("request.failed", error = ?err, %status_code);                                                                                                                                  
  /// ```                                                                  
```

lalitb: ok ignore above, if the error is reasonable :)

Issue #1746, opened 2026-01-09
Improvements in OTLP encoding logic required for internal logging safety


jmacd:
Following #1735, here are a list of improvements we desire in the OTLP ProtoBuffer encoding object of crates/pdata/src/otlp/common.rs

1. Ability to encode into a limited-size buffer to avoid unbounded memory use
2. Ability to encode into a single pre-allocated &mut [u8]
3. Ability to safely truncate long strings appending a "..." or other notice
4. Ability to safely truncate a list of top-level fields (e.g., the attributes list of a log record) 
5. Ability to automatically generate the dropped_attributes_count field of LogRecord, Span, Resource etc by reserving 5 bytes
6. Ability to choose (somehow) the number of bytes dedicated to placeholders; the current 4-byte approach is wasteful for small buffers

Ultimately with all of this done we will be able to safely log into pre-allocated buffers and obtain usefully truncated and/or formatted messages without allocating memory inside a tokio tracing event.

Part of #1736

PR #1849, opened 2026-01-21
Console exporter for OTAP/OTLP logs

jmacd:

Uses the new internal logging support to format OTLP logs data. This prints RESOURCE and SCOPE lines with ASCII or Unicode pipe structures to identify the OTLP hierarchy:

```
2026-01-21T03:12:22.165Z  RESOURCE   v1.Resource: [fake_data_generator=v1]
2026-01-21T03:12:22.165Z  │ SCOPE    v1.InstrumentationScope:
2026-01-21T03:12:22.165Z  │ ├─ INFO  session.start:  [session.id=00112233-4455-6677-8899-aabbccddeeff, session.previous_id=00112233-4455-6677-8899-aabbccddeeff]
2026-01-21T03:12:22.165Z  │ ├─ INFO  session.end:  [session.id=00112233-4455-6677-8899-aabbccddeeff]
2026-01-21T03:12:22.165Z  │ ├─ INFO  device.app.lifecycle:  [ios.app.state=active, android.app.state=created]
2026-01-21T03:12:22.165Z  │ ├─ INFO  rpc.message:  [rpc.message.type=SENT, rpc.message.id=42, rpc.message.compressed_size=42, rpc.message.uncompressed_size=42]
```

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/console_exporter.rs
+++ b/rust/otap-dataflow/crates/otap/src/console_exporter.rs
@@ -118,9 +118,9 @@ impl Exporter<OtapPdata> for ConsoleExporter {
 impl ConsoleExporter {
     async fn export(&self, payload: &OtapPayload) {
         match payload.signal_type() {
-            SignalType::Logs => self.export_logs(&payload).await,
-            SignalType::Traces => self.export_traces(&payload).await,
-            SignalType::Metrics => self.export_metrics(&payload).await,
+            SignalType::Logs => self.export_logs(payload).await,
+            SignalType::Traces => self.export_traces(payload).await,
+            SignalType::Metrics => self.export_metrics(payload).await,
         }
     }
 


rust/otap-dataflow/crates/otap/src/console_exporter.rs
```diff
@@ -0,0 +1,436 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Console exporter that prints OTLP data with hierarchical formatting.
```
cijothomas, line 4: do we want to offer stability guarantees about output format or we want to explicitly mention that output format is not stable and not recommended to be relied for automation/parsing.?

rust/otap-dataflow/crates/otap/src/console_exporter.rs
```diff
@@ -0,0 +1,436 @@
...
+        let time = log_record
+            .time_unix_nano()
+            .or_else(|| log_record.observed_time_unix_nano())
+            .map(nanos_to_time)
+            .unwrap_or(SystemTime::UNIX_EPOCH);
+
+        let event_name = log_record
+            .event_name()
+            .map(|s| String::from_utf8_lossy(s).into_owned())
+            .unwrap_or_else(|| "event".to_string());
```
cijothomas: why "event" instead of leaving it empty?

rust/otap-dataflow/crates/otap/src/console_exporter.rs
```diff
@@ -0,0 +1,436 @@
...
+
+    fn export_logs(&self, payload: &OtapPayload) {
+        match payload {
+            OtapPayload::OtlpBytes(bytes) => {
+                self.formatter.format_logs_bytes(bytes);
+            }
+            OtapPayload::OtapArrowRecords(records) => match OtapLogsView::try_from(records) {
+                Ok(logs_view) => {
+                    self.formatter.format_logs_arrow(&logs_view);
+                }
```
utpilla: It'd be nice to implement `impl<'a> TryFrom<&'a OtlpProtoBytes> for OtapLogsView<'a>` as well similar to `OtapArrowRecords`. That would simplify the code here by having both paths go through the same set of steps.

OtlpBytes/OtapArrowRecords -> OtapLogsView -> Write to console

rust/otap-dataflow/crates/otap/src/console_exporter.rs
```diff
@@ -0,0 +1,436 @@
...
+            }
+        }
+
+        Ok(TerminalState::default())
+    }
+}
+
+impl ConsoleExporter {
+    fn export(&self, data: &OtapPdata) {
+        let (_, payload) = data.clone().into_parts();
```
utpilla: We could avoid the cloning of payload here for each export. We are only using references here in the `export_*` methods. It could be worth adding another method to `OtapPData` which returns the reference to payload without consuming the payload:

```rust
/// Returns a reference to the payload
#[must_use]
pub const fn payload_ref(&self) -> &OtapPayload {
    &self.payload
}
```

rust/otap-dataflow/crates/otap/src/console_exporter.rs
```diff
@@ -0,0 +1,436 @@
...
+        is_last_scope: bool,
+        output: &mut Vec<u8>,
+    ) {
+        let first_ts = scope_logs
+            .log_records()
+            .find_map(|lr| lr.time_unix_nano().or_else(|| lr.observed_time_unix_nano()))
+            .map(nanos_to_time)
+            .unwrap_or(SystemTime::UNIX_EPOCH);
+
+        let prefix = format!("{} ", self.tree.vertical);
```
utpilla: This would allocate a new string for each scope. Since `tree.vertical` is a `&static str`, we could pre-compute the prefix as a const.

rust/otap-dataflow/crates/otap/src/console_exporter.rs
```diff
@@ -0,0 +1,436 @@
...
+            );
+        });
+
+        // Format each scope
+        let scopes: Vec<_> = resource_logs.scopes().collect();
+        let scope_count = scopes.len();
+        for (i, scope_logs) in scopes.into_iter().enumerate() {
+            let is_last_scope = i == scope_count - 1;
+            self.format_scope_logs_to(&scope_logs, is_last_scope, output);
+        }
```
utpilla, line 254: It looks like we're allocating a vec here just to identify the last element. Could we use `peekable()` instead to avoid the allocation?

rust/otap-dataflow/crates/otap/src/console_exporter.rs
```diff
@@ -0,0 +1,436 @@
...
+            );
+        });
+
+        // Format each log record
+        let records: Vec<_> = scope_logs.log_records().collect();
+        let record_count = records.len();
+        for (i, log_record) in records.into_iter().enumerate() {
+            let is_last_record = i == record_count - 1;
+            self.format_log_record_to(&log_record, is_last_scope, is_last_record, output);
+        }
```
utpilla, line 326: Same as https://github.com/open-telemetry/otel-arrow/pull/1849#discussion_r2713380889

rust/otap-dataflow/crates/otap/src/console_exporter.rs
```diff
@@ -0,0 +1,436 @@
...
+
+        // Extract name/version for inline display
+        let name = scope
+            .as_ref()
+            .and_then(|s| s.name())
+            .map(|n| String::from_utf8_lossy(n).into_owned());
+        let version = scope
+            .as_ref()
+            .and_then(|s| s.version())
+            .map(|v| String::from_utf8_lossy(v).into_owned());
```
utpilla: We could rid of the `into_owned` and avoid the string allocation here by using a `Cow<str>` or raw bytes directly.

rust/otap-dataflow/crates/otap/src/console_exporter.rs
utpilla: Good question. We've seen users relying on SDK's ConsoleExporter in production with the expectation that the output format won't change.

rust/otap-dataflow/crates/otap/src/console_exporter.rs
```diff
@@ -0,0 +1,436 @@
...
+    pub unicode: bool,
+}
+
+fn default_color() -> bool {
+    true
+}
+
+fn default_unicode() -> bool {
+    true
+}
```
utpilla:
```suggestion
const fn default_color() -> bool {
    true
}

const fn default_unicode() -> bool {
    true
}
```

rust/otap-dataflow/crates/otap/src/console_exporter.rs
```diff
@@ -0,0 +1,436 @@
...
+                TreeChars::ASCII
+            },
+        }
+    }
+
+    /// Format logs from OTLP bytes to stdout.
+    pub fn format_logs_bytes(&self, bytes: &OtlpProtoBytes) {
+        let mut output = Vec::new();
+        self.format_logs_bytes_to(bytes, &mut output);
+        let _ = std::io::stdout().write_all(&output);
```
utpilla: Use the async version?

`let _ = tokio::io::stdout().write_all(&output).await;`

rust/otap-dataflow/crates/otap/src/console_exporter.rs
```diff
@@ -0,0 +1,436 @@
...
+    }
+
+    /// Format logs from Arrow format to stdout.
+    pub fn format_logs_arrow(&self, logs_view: &OtapLogsView<'_>) {
+        let mut output = Vec::new();
+        self.format_logs_data_to(logs_view, &mut output);
+        let _ = std::io::stdout().write_all(&output);
```
utpilla: Do we want to log errors here if `write_all` fails?

utpilla (approved):
Most of my comments are perf related which may not be as much of a concern for a ConsoleExporter.

There is one [comment](https://github.com/open-telemetry/otel-arrow/pull/1849#discussion_r2713445604) about using the std I/O blocking `write_all` vs tokio's async std I/O which I think we should address.

jmacd:
I appreciate all of this feedback! I see this component function as not a high priority, however all the points made about performance are likely to impact other export paths so I wouldn't dismiss any it.

rust/otap-dataflow/crates/otap/src/console_exporter.rs
lalitb: @utpilla - Wondering if this would need decoding the protobuf into rows and then building Arrow arrays (and the OTAP layout).  How about keeping `RawLogsData` for OTLP bytes and `OtapLogsView` for Arrow; a generic `format_logs<L: LogsDataView>` to keep call sites tidy?

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
```diff
@@ -317,6 +301,109 @@ impl ConsoleWriter {
...
+    /// Write a tracing Level with appropriate color and padding.
+    #[inline]
+    pub fn write_level(&self, w: &mut BufWriter<'_>, level: &Level) {
+        self.write_severity(w, Some(level_to_severity_number(level) as i32));
+    }
+
+    /// Write severity number with appropriate color and padding.
+    /// Severity numbers follow OTLP conventions (1-24, where INFO=9).
+    #[inline]
+    pub fn write_severity(&self, w: &mut BufWriter<'_>, severity: Option<i32>) {
```
lalitb: should we extend this method to also write severity-text along with the severity number?

jmacd: ah, yes! I had not considered severity text since I dropped it from the LogRecord encoding as being redundant for the 5 Tokio levels.

rust/otap-dataflow/crates/otap/src/console_exporter.rs
jmacd: Noted as unstable for now! We can open an explicit stabilization issue.

rust/otap-dataflow/crates/otap/src/console_exporter.rs
jmacd: That was a mistake, fixed.

rust/otap-dataflow/crates/otap/src/console_exporter.rs
jmacd: This should not require decoding! Trying...

rust/otap-dataflow/crates/otap/src/console_exporter.rs
```diff
@@ -0,0 +1,436 @@
...
+            );
+        });
+    }
+
+    /// Format a line to the output buffer.
+    fn format_line<F>(&self, output: &mut Vec<u8>, f: F)
+    where
+        F: FnOnce(&mut BufWriter<'_>),
+    {
+        let mut buf = [0u8; LOG_BUFFER_SIZE];
```
lalitb, line 381: More for later PR, should be good to add TODO for now - this is silently truncating if the writable size if over 4KB. Which means - A log record with a big body or lots of attributes will get cut off without warning. We can detect overflow, and append truncation marker, so user doesn't get misled with truncated content. Or else not to truncate by using growable buffer.

rust/otap-dataflow/crates/otap/src/console_exporter.rs
lalitb: Oh yes, but looking into the code, there may be lifetime issue.

jmacd: I cleaned this up somewhat, but yes, I'm not sure how to handle this kind of try-various-impls pattern here. I did add a try_from so that at least the two are symmetric:

    fn export_logs(&self, payload: &OtapPayload) {
        match payload {
            OtapPayload::OtlpBytes(bytes) => match RawLogsData::try_from(bytes) {
                Ok(logs_view) => {
                    self.formatter.print_logs_data(&logs_view);
                }
                Err(e) => {
                    otel_error!("Failed to create OTLP logs view", error = ?e);
                }
            },
            OtapPayload::OtapArrowRecords(records) => match OtapLogsView::try_from(records) {
                Ok(logs_view) => {
                    self.formatter.print_logs_data(&logs_view);
                }
                Err(e) => {
                    otel_error!("Failed to create OTAP logs view", error = ?e);
                }
            },
        }
    }

lalitb: This looks good actually. The trait-based approach is the right pattern here.

rust/otap-dataflow/crates/otap/src/console_exporter.rs
jmacd: Done! https://github.com/open-telemetry/otel-arrow/pull/1849/commits/398518356da64d22aa402f81f4824a4cfc85a3ac

rust/otap-dataflow/crates/otap/src/console_exporter.rs
jmacd: Fixed!

rust/otap-dataflow/crates/otap/src/console_exporter.rs
jmacd: https://github.com/open-telemetry/otel-arrow/pull/1849/commits/978d44b13aeccc3283f007c69520beef345c280b

rust/otap-dataflow/crates/otap/src/console_exporter.rs
jmacd: Fixed! Thanks. https://github.com/open-telemetry/otel-arrow/pull/1849/commits/978d44b13aeccc3283f007c69520beef345c280b

rust/otap-dataflow/crates/otap/src/console_exporter.rs
jmacd: Done (raw bytes works).

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: Added, tested.

rust/otap-dataflow/crates/otap/src/console_exporter.rs
jmacd: Now reads:
```
        if let Err(err) = tokio::io::stdout().write_all(&output).await {
            otel_error!("could not write to console", error = ?err);
        }
```

rust/otap-dataflow/crates/otap/src/console_exporter.rs
jmacd: I have this and some related topics recorded in https://github.com/open-telemetry/otel-arrow/issues/1746 🚀

PR #1852, opened 2026-01-21
feat(opl): add contains and regex match expressions to parser

### LINKED ISSUE #1789: OPL Parser: contains substring and regex matches
We need to design a syntax for OPL for substring contains and regex match.

KQL uses a syntax like:
```
x contains "y"
x matches regex ".*albert.*"
```
We could do something similar, or do something more like a function call `contains(x, "y")`. Will leave the actualy syntax unprescribed for this moment.

Once we've decided on the syntax:
- add support to the parser these should parse into `LogicalExpression::Matches` and `LogicalExpression::Contains`
- update the tests in columnar query engine `pipeline::filter` module to be driven by OPLParser as well as KQL parser.

Apostlex0:
closes #1789

### DIFF
--- a/rust/otap-dataflow/crates/opl/src/parser/expression.rs
+++ b/rust/otap-dataflow/crates/opl/src/parser/expression.rs
@@ -595,6 +595,12 @@ fn parse_primitive_expression(rule: Pair<'_, Rule>) -> Result<LogicalOrScalarExp
     }
 }
 
+/// Parses invocation of function
+///
+/// Currently we only support two known function `matches` and `contains`, both of which
+/// take two arguments and return a boolean value. In the future, our function library
+/// may be expanded to include more functions with various signatures and user defined
+/// functions, so the implementation of this function will be updated accordingly.
 fn parse_function_call(rule: Pair<'_, Rule>) -> Result<LogicalOrScalarExpr, ParserError> {
     let query_location = to_query_location(&rule);
     let mut inner_rules = rule.into_inner();
@@ -1553,12 +1559,26 @@ mod test {
     }
 
     #[test]
+    fn test_parse_function_call_unknown_function_name() {
+        let input = "unknown_fn(severity_text, \"ERR\")";
+        let mut rules = OplPestParser::parse(Rule::member_expression, input).unwrap();
+        assert_eq!(rules.len(), 1);
+
+        let err = parse_member_expression(rules.next().unwrap()).unwrap_err();
+        assert!(err.to_string().contains("Unsupported function"));
+    }
+
+    #[test]
     fn test_parse_function_call_wrong_arity_one_arg_errors() {
         let input = "contains(severity_text)";
         let mut rules = OplPestParser::parse(Rule::member_expression, input).unwrap();
         assert_eq!(rules.len(), 1);
 
-        assert!(parse_member_expression(rules.next().unwrap()).is_err());
+        let err = parse_member_expression(rules.next().unwrap()).unwrap_err();
+        assert!(
+            err.to_string()
+                .contains("Function 'contains' expects 2 arguments, got 1")
+        )
     }
 
     #[test]
@@ -1567,6 +1587,10 @@ mod test {
         let mut rules = OplPestParser::parse(Rule::member_expression, input).unwrap();
         assert_eq!(rules.len(), 1);
 
-        assert!(parse_member_expression(rules.next().unwrap()).is_err());
+        let err = parse_member_expression(rules.next().unwrap()).unwrap_err();
+        assert!(
+            err.to_string()
+                .contains("Function 'matches' expects 2 arguments, got 3")
+        )
     }
 }


Apostlex0:
@albertlockett should be ready for review!

rust/otap-dataflow/crates/opl/src/parser/expression.rs
```diff
@@ -1311,4 +1412,161 @@ mod test {
...
+    }
+
+    #[test]
+    fn test_parse_function_call_wrong_arity_three_args_errors() {
+        let input = "matches(severity_text, \"ERR.*\", \"extra\")";
+        let mut rules = OplPestParser::parse(Rule::member_expression, input).unwrap();
+        assert_eq!(rules.len(), 1);
+
+        assert!(parse_member_expression(rules.next().unwrap()).is_err());
+    }
```
albertlockett, line 1595: Should we add a test for invalid function name?

```suggestion
    }

    #[test]
    fn test_parse_function_call_unknown_function_name() {
        let input = "matches(severity_text, \"ERR.*\", \"extra\")";
        let mut rules = OplPestParser::parse(Rule::member_expression, input).unwrap();
        assert_eq!(rules.len(), 1);
        let err = parse_member_expression(rules.next().unwrap()).unwrap_err();
        assert!(err.to_string().contains("Unknown function"));
    }

File: rust/otap-dataflow/crates/opl/src/parser/expression.rs [hunk 2]
```diff
@@ -1311,4 +1412,161 @@ mod test {
...
+
+    #[test]
+    fn test_parse_function_call_wrong_arity_three_args_errors() {
+        let input = "matches(severity_text, \"ERR.*\", \"extra\")";
+        let mut rules = OplPestParser::parse(Rule::member_expression, input).unwrap();
+        assert_eq!(rules.len(), 1);
+
+        assert!(parse_member_expression(rules.next().unwrap()).is_err());
```
albertlockett: Wondering if we should also assert on the error message?
```suggestion
        let err = parse_member_expression(rules.next().unwrap_err()
        assert!(err.to_string().contains("Function 'matches' expects 2 arguments, got 3"));
```

rust/otap-dataflow/crates/opl/src/parser/expression.rs
```diff
@@ -1311,4 +1412,161 @@ mod test {
...
+        assert_eq!(result, expected);
+    }
+
+    #[test]
+    fn test_parse_function_call_wrong_arity_one_arg_errors() {
+        let input = "contains(severity_text)";
+        let mut rules = OplPestParser::parse(Rule::member_expression, input).unwrap();
+        assert_eq!(rules.len(), 1);
+
+        assert!(parse_member_expression(rules.next().unwrap()).is_err());
```
albertlockett:
```suggestion
        let err = parse_member_expression(rules.next().unwrap_err()
        assert!(err.to_string().contains("Function 'contains' expects 2 arguments, got 1"));
```

rust/otap-dataflow/crates/opl/src/parser/expression.rs
```diff
@@ -592,15 +594,114 @@ fn parse_primitive_expression(rule: Pair<'_, Rule>) -> Result<LogicalOrScalarExp
         )),
     }
 }
+
+fn parse_function_call(rule: Pair<'_, Rule>) -> Result<LogicalOrScalarExpr, ParserError> {
```
albertlockett, line 604: In the future, we'll probably adapt this implementation to involve the use of a known function library, rather than making implicit assumptions about the function signatures. I think we should just add a comment stating that for now, since there's only two functions, we have a simpler parsing approach and the implementation may be adapted down the line
```suggestion
/// Parses invocation of function
///
/// Currently we only support two known function `matches` and `contains`, both of which
/// take two arguments and return a boolean value. In the future, our function library
/// may be expanded to include more functions with various signatures and user defined
/// functions, so the implementation of this function will be updated accordingly.
fn parse_function_call(rule: Pair<'_, Rule>) -> Result<LogicalOrScalarExpr, ParserError> {
```

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
```diff
@@ -1904,8 +1904,7 @@ mod test {
         test_simple_attrs_filter::<OplParser>().await;
     }
 
-    #[tokio::test]
-    async fn test_filter_text_contains() {
+    async fn test_filter_text_contains<P: Parser>(q1: &str, q2: &str, q3: &str, q4: &str) {
```
albertlockett: nit: for these tests where we have `q1`, `q2`, etc. I wonder if there's a way we could make it more clear what the arguments are supposed to be. Maybe we could we use a more descriptive variable names?

```suggestion
    async fn test_filter_text_contains<P: Parser>(
        q_event_name_contains_error: &str, 
        q_1234_contains_event_name: &str,
        q_attrs_username_contains_y: &str, 
        q_albert_contains_attrs_username: &str
    ) {
```

rust/otap-dataflow/crates/opl/src/parser/expression.rs
Apostlex0: Got it👍

rust/otap-dataflow/crates/opl/src/parser/expression.rs
Apostlex0: Sure, we could do that given that we only support 2 for now👍

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
Apostlex0: this was a helper function so I thought it won't be necessary since the actual call sites for this do provide descriptive arguments, but sure I'll make the change👍

Apostlex0:
@albertlockett changes made🫡

Issue #1396, opened 2025-11-06
Add Microsoft Log Analytics (Logs Ingestion API) Exporter for OTel Arrow


gouslu:
In order to export logs to Logs Ingestion API, we need a new exporter in Rust. We plan to add it as an experimental feature initially with the goal of making it mainstream in the future.

[Logs Ingestion API](https://learn.microsoft.com/en-us/azure/azure-monitor/logs/logs-ingestion-api-overview)

Implementation will be done in multiple phases.

- [ ] Basic functionality
- [ ] Performance optimizations
- [ ] Private Link Support

PR #1854, opened 2026-01-22
feat(azure-monitor-exporter): Add heartbeat support and refactor auth handling

gouslu:
# Change Summary

Adds heartbeat functionality to the Azure Monitor Exporter and refactors authentication to use a dedicated `AuthHeader` module for reusability

- **Heartbeat support**: Sends periodic health heartbeats to Azure Monitor every 60 seconds via the `HEALTH_ASSESSMENT_BLOB` stream
- Heartbeat metadata sourced from environment variables (`IMAGE`, `POD_NAME`, `EXPORTER_ID`, `ARM_RESOURCE_ID`, `HOSTNAME`) with sensible fallbacks
- Move auth out of clients and update auth header of clients pro-actively using periodic tasks.

* Closes heartbeat item on issue #1396

Local manual tests and unit tests

### DIFF
--- a/rust/otap-dataflow/benchmarks/Cargo.toml
+++ b/rust/otap-dataflow/benchmarks/Cargo.toml
@@ -41,6 +41,7 @@ unsync.workspace = true
 portpicker.workspace = true
 tokio-stream.workspace = true
 weaver_common.workspace = true
+bytes.workspace = true
 
 [target.'cfg(not(windows))'.dev-dependencies]
 tikv-jemallocator.workspace = true
--- a/rust/otap-dataflow/benchmarks/benches/self_tracing/main.rs
+++ b/rust/otap-dataflow/benchmarks/benches/self_tracing/main.rs
@@ -12,7 +12,14 @@
 
 use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
 use otap_df_pdata::otlp::ProtoBuffer;
-use otap_df_telemetry::self_tracing::{ConsoleWriter, DirectLogRecordEncoder, LogRecord};
+use otap_df_telemetry::attributes::{AttributeSetHandler, AttributeValue};
+use otap_df_telemetry::descriptor::{AttributeField, AttributeValueType, AttributesDescriptor};
+use otap_df_telemetry::event::LogEvent;
+use otap_df_telemetry::registry::TelemetryRegistryHandle;
+use otap_df_telemetry::self_tracing::{
+    DirectLogRecordEncoder, LogContext, LogRecord, ScopeToBytesMap, encode_export_logs_request,
+    format_log_record_to_string,
+};
 use std::time::SystemTime;
 use tracing::{Event, Subscriber};
 use tracing_subscriber::layer::Layer;
@@ -25,6 +32,47 @@ use tikv_jemallocator::Jemalloc;
 #[cfg(not(windows))]
 #[global_allocator]
 static GLOBAL: Jemalloc = Jemalloc;
+
+// Test scope attributes for entity benchmarks
+static BENCH_SCOPE_ATTRIBUTES_DESCRIPTOR: AttributesDescriptor = AttributesDescriptor {
+    name: "BenchScope",
+    fields: &[
+        AttributeField {
+            key: "pipeline.name",
+            r#type: AttributeValueType::String,
+            brief: "Pipeline name",
+        },
+        AttributeField {
+            key: "cpu.id",
+            r#type: AttributeValueType::Int,
+            brief: "CPU ID",
+        },
+    ],
+};
+
+/// Mock attribute set for benchmarking scope attributes.
+#[derive(Debug)]
+struct BenchScopeAttributes {
+    values: Vec<AttributeValue>,
+}
+
+impl BenchScopeAttributes {
+    fn new(name: &str, id: i64) -> Self {
+        Self {
+            values: vec![AttributeValue::String(name.into()), AttributeValue::Int(id)],
+        }
+    }
+}
+
+impl AttributeSetHandler for BenchScopeAttributes {
+    fn descriptor(&self) -> &'static AttributesDescriptor {
+        &BENCH_SCOPE_ATTRIBUTES_DESCRIPTOR
+    }
+
+    fn attribute_values(&self) -> &[AttributeValue] {
+        &self.values
+    }
+}
 
 /// The operation to perform on each event within the layer.  The cost
 /// of generating the timestamp is not included in the  measurement.
@@ -42,17 +90,62 @@ enum BenchOp {
     EncodeAndFormat,
     /// Encode to complete protobuf (with timestamp) N times.
     EncodeProto,
+    /// Encode to complete protobuf with scope cache (with entity context) N times.
+    /// Measures overhead of scope attribute encoding/caching.
+    EncodeProtoWithScope,
+    /// Format with entity context N times.
+    /// Measures overhead of formatting entity keys.
+    FormatWithEntity,
+}
+
+impl BenchOp {
+    /// Returns true if this operation requires entity context.
+    fn needs_entity(self) -> bool {
+        matches!(
+            self,
+            BenchOp::EncodeProtoWithScope | BenchOp::FormatWithEntity
+        )
+    }
 }
 
 /// A layer that performs a configurable operation N times per event.
 struct BenchLayer {
     iterations: usize,
     op: BenchOp,
+    /// Optional entity context for scope benchmarks.
+    entity_context: Option<EntityContext>,
+}
+
+/// Entity context for scope benchmarks (registry, entity key, resource bytes).
+struct EntityContext {
+    registry: TelemetryRegistryHandle,
+    context: LogContext,
+    resource_bytes: bytes::Bytes,
 }
 
 impl BenchLayer {
     fn new(iterations: usize, op: BenchOp) -> Self {
-        Self { iterations, op }
+        Self {
+            iterations,
+            op,
+            entity_context: None,
+        }
+    }
+
+    fn with_entity(iterations: usize, op: BenchOp) -> Self {
+        let registry = TelemetryRegistryHandle::new();
+        let entity_key = registry.register_entity(BenchScopeAttributes::new("bench-pipeline", 42));
+        // Empty resource bytes - we're benchmarking scope encoding, not resource encoding
+        let resource_bytes = bytes::Bytes::new();
+        Self {
+            iterations,
+            op,
+            entity_context: Some(EntityContext {
+                registry,
+                context: LogContext::from_buf([entity_key]),
+                resource_bytes,
+            }),
+        }
     }
 }
 
@@ -65,26 +158,23 @@ where
         match self.op {
             BenchOp::NewRecord => {
                 for _ in 0..self.iterations {
-                    let record = LogRecord::new(event);
+                    let record = LogRecord::new(event, LogContext::new());
                     let _ = std::hint::black_box(record);
                 }
             }
             BenchOp::Format => {
                 // Encode once, format N times
-                let record = LogRecord::new(event);
-                let writer = ConsoleWriter::no_color();
+                let record = LogRecord::new(event, LogContext::new());
 
                 for _ in 0..self.iterations {
-                    let line = writer.format_log_record(Some(now), &record);
+                    let line = format_log_record_to_string(Some(now), &record);
                     let _ = std::hint::black_box(line);
                 }
             }
             BenchOp::EncodeAndFormat => {
-                let writer = ConsoleWriter::no_color();
-
                 for _ in 0..self.iterations {
-                    let record = LogRecord::new(event);
-                    let line = writer.format_log_record(Some(now), &record);
+                    let record = LogRecord::new(event, LogContext::new());
+                    let line = format_log_record_to_string(Some(now), &record);
                     let _ = std::hint::black_box(line);
                 }
             }
@@ -94,8 +184,42 @@ where
 
                 for _ in 0..self.iterations {
                     encoder.clear();
-                    let size = encoder.encode_log_record(now, &LogRecord::new(event));
+                    let size =
+                        encoder.encode_log_record(now, &LogRecord::new(event, LogContext::new()));
                     let _ = std::hint::black_box(size);
+                }
+            }
+            BenchOp::EncodeProtoWithScope => {
+                let ctx = self
+                    .entity_context
+                    .as_ref()
+                    .expect("entity context required");
+                let mut scope_cache = ScopeToBytesMap::new(ctx.registry.clone());
+                let mut buf = ProtoBuffer::with_capacity(512);
+
+                for _ in 0..self.iterations {
+                    buf.clear();
+                    let record = LogRecord::new(event, ctx.context.clone());
+                    let log_event = LogEvent { time: now, record };
+                    encode_export_logs_request(
+                        &mut buf,
+                        &log_event,
+                        &ctx.resource_bytes,
+                        &mut scope_cache,
+                    );
+                    let _ = std::hint::black_box(buf.len());
+                }
+            }
+            BenchOp::FormatWithEntity => {
+                let ctx = self
+                    .entity_context
+                    .as_ref()
+                    .expect("entity context required");
+
+                for _ in 0..self.iterations {
+                    let record = LogRecord::new(event, ctx.context.clone());
+                    let line = format_log_record_to_string(Some(now), &record);
+                    let _ = std::hint::black_box(line);
                 }
             }
         }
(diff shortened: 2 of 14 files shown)


rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/heartbeat.rs
```diff
@@ -0,0 +1,602 @@
...
+    })
+}
+
+impl Heartbeat {
+    /// Create a new Heartbeat instance.
+    pub fn new(config: &ApiConfig, auth: Auth) -> Result<Self, Error> {
+        let http_client = Client::builder()
+            .http1_only()
+            .timeout(Duration::from_secs(30))
+            .pool_max_idle_per_host(MAX_IDLE_CONNECTIONS_PER_HOST) // e.g., 32/4 = 8 connections per pool
```
utpilla: This comment does not match the value set for `MAX_IDLE_CONNECTIONS_PER_HOST`.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth_header.rs
```diff
@@ -0,0 +1,338 @@
...
+        // Pre-format the authorization header to avoid repeated allocation
+        self.value = HeaderValue::from_str(&format!("Bearer {}", token.token.secret()))
+            .map_err(Error::InvalidHeader)?;
+
+        // Calculate validity using Instant for faster comparisons
+        // Refresh 5 minutes before expiry
+        let valid_seconds = (token.expires_on - OffsetDateTime::now_utc()).whole_seconds();
+
+        self.token_valid_until = Instant::now() + Duration::from_secs(valid_seconds.max(0) as u64);
+        self.token_refresh_after = self.token_valid_until - Duration::from_secs(300);
```
utpilla: This is a good candidate for using named constant.

```rust
const TOKEN_REFRESH_BUFFER_SECS: u64 = 300;
...
self.token_refresh_after = self.token_valid_until - Duration::from_secs(TOKEN_REFRESH_BUFFER_SECS);
```

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/heartbeat.rs
```diff
@@ -0,0 +1,602 @@
...
+fn default_heartbeat_os_major_version() -> String {
+    std::env::var("POD_NAME").unwrap_or_else(|_| {
+        let (major, _) = parse_os_version();
+        major
+    })
+}
+
+#[inline]
+fn default_heartbeat_os_minor_version() -> String {
+    std::env::var("EXPORTER_ID").unwrap_or_else(|_| {
```
cijothomas: what is "EXPORTER_ID" env variable, and how is it OS minor version?

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/heartbeat.rs
```diff
@@ -0,0 +1,602 @@
...
+                os_major_version: "1".to_string(),
+                os_minor_version: "0".to_string(),
+            },
+            auth_header: AuthHeader::new(auth),
+        }
+    }
+
+    /// Send a heartbeat to the Azure Monitor Logs Ingestion endpoint.
+    pub async fn send(&mut self) -> Result<(), Error> {
+        self.heartbeat_row.time = Utc::now().to_rfc3339();
```
utpilla, line 166: Why not check token refresh proactively?

```suggestion
    pub async fn send(&mut self) -> Result<(), Error> {
        self.auth_header.ensure_valid_token().await?; 
```

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/heartbeat.rs
```diff
@@ -0,0 +1,602 @@
...
+    })
+}
+
+#[inline]
+fn default_heartbeat_os_minor_version() -> String {
+    std::env::var("EXPORTER_ID").unwrap_or_else(|_| {
+        let (_, minor) = parse_os_version();
+        minor
+    })
+}
```
utpilla, line 108: These seem odd. `POD_NAME` for OS major version and `EXPORTER_ID` for OS minor version.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
```diff
@@ -498,6 +511,14 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
...
+                    next_heartbeat_send = tokio::time::Instant::now() + tokio::time::Duration::from_secs(HEARTBEAT_INTERVAL_SECONDS);
+                    match heartbeat.send().await {
+                        Ok(_) => println!("[AzureMonitorExporter] Heartbeat sent"),
+                        Err(e) => println!("[AzureMonitorExporter] Heartbeat send failed: {:?}", e),
+                    }
+                }
+
                 completed = self.in_flight_exports.next_completion() => {
                     if let Some(completed_export) = completed {
                         self.finalize_export(&effect_handler, completed_export).await?;
```
utpilla, line 503: Not changed in this PR but there are some blocking File I/O calls being made in the exporter for stats collection:

```rust
let status = std::fs::read_to_string("/proc/self/status").unwrap_or_default();
let smaps = std::fs::read_to_string("/proc/self/smaps_rollup").unwrap_or_default();
```

- These should use async versions of the API offered by `tokio::fs`
- These would only work on Linux, therefore the whole branch should be placed under `#[cfg(target_os = "linux")]`

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
```diff
@@ -498,6 +511,14 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
                     next_token_refresh = Self::get_next_token_refresh(new_token);
                 }
 
+                _ = tokio::time::sleep_until(next_heartbeat_send) => {
+                    next_heartbeat_send = tokio::time::Instant::now() + tokio::time::Duration::from_secs(HEARTBEAT_INTERVAL_SECONDS);
+                    match heartbeat.send().await {
+                        Ok(_) => println!("[AzureMonitorExporter] Heartbeat sent"),
```
utpilla, line 496: There are few other places as well in this crate which use `println!` which could block the thread. Use `otel_info!` macro instead. You can specify `console_async` as the provider mode for `engine` in the config yaml to make sure that `otel_info!` statements are getting printed to console asynchronously. 

This could be done in a different PR.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/heartbeat.rs
gouslu: oh, thanks for catching that

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/heartbeat.rs
gouslu: token is refreshed in the background, the code below that handles 401 to cover edge cases.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/heartbeat.rs
gouslu: these are from old exporter

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
gouslu: I want to keep that overhaul of removing println statements to a separate PR. currently I'd like to keep the println.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth_header.rs
```diff
@@ -0,0 +1,338 @@
...
+        AuthHeader {
+            auth,
+            value: HeaderValue::from_static("Bearer "),
+            token_valid_until: Instant::now(),
+            token_refresh_after: Instant::now(),
+        }
+    }
+
+    /// Refresh the token and update the pre-formatted header
+    pub async fn refresh_token(&mut self) -> Result<(), Error> {
```
lalitb: this seems to be identical login as [LogsIngestionClient::refresh_token](https://github.com/open-telemetry/otel-arrow/blob/faeb878584aefc1cba85fea5a8ea11a9aa5350ae/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs#L156) - would be good to reuse?

gouslu: it is moved from there to auth header, so it is not in the client anymore. purpose of this auth_header module is that reusability between client and heartbeat.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/heartbeat.rs
lalitb: @gouslu  The background refresh in `exporter.rs` only updates the `auth` used by the ingestion clients; it never updates `heartbeat.auth_header`. That header is cloned at startup, refreshed once, and then only refreshed in the 401 branch. We should refresh before sending rather than relying on the 401 path, as @utpilla  suggested.

gouslu: thanks for pointing it out, I think this bug was there even before this PR for the client and I didn't realize it. will fix it!

gouslu:
per @lalitb and @utpilla 's comments, I realized that there is a bug with the background token refresh, and it seems like it has been there even before this PR. I'll apply those fixes and update the PR, until then this PR should be on hold.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
```diff
@@ -494,8 +457,46 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
...
+                                    // Retry every 10 seconds
+                                    next_token_refresh = tokio::time::Instant::now() + tokio::time::Duration::from_secs(10);
+                                }
+                            }
+
+                        }
+                        Err(e) => {
+                            println!("[AzureMonitorExporter] Failed to refresh access token: {:?}", e);
+                            // Retry every 10 seconds
+                            next_token_refresh = tokio::time::Instant::now() + tokio::time::Duration::from_secs(10);
```
ThomsonTan, line 488: If token refresh failed and delayed for another 10s, would the following heatbeat still be sent? If yes, the send will fail without a valid token?

gouslu: idk if there is a bug you can see, but this is the idea:

- we got a token, it is still valid, we refresh it 295 secs prior to expiry
- refresh failed but we still have a valid token header in the clients
- we let the work continue, while waiting for 10 seconds to try to get the token again

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
```diff
@@ -2,27 +2,28 @@
...
 use super::Error;
 use super::config::{AuthConfig, AuthMethod};
 
-#[derive(Clone)]
+/// Minimum delay between token refresh retry attempts in seconds.
+const MIN_RETRY_DELAY_SECS: f64 = 5.0;
+/// Maximum delay between token refresh retry attempts in seconds.
+const MAX_RETRY_DELAY_SECS: f64 = 30.0;
+/// Maximum jitter percentage (±10%) to add to retry delays.
+const MAX_RETRY_JITTER_PERC: f64 = 0.10;
```
jmacd: If 0.1 represents 10%, let's call this var `_RATIO`

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
```diff
@@ -494,8 +456,46 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
...
+                                Err(e) => {
+                                    println!("[AzureMonitorExporter] Failed to create auth header: {:?}", e);
+                                    // Retry every 10 seconds
+                                    next_token_refresh = tokio::time::Instant::now() + tokio::time::Duration::from_secs(10);
+                                }
+                            }
+
+                        }
+                        Err(e) => {
+                            println!("[AzureMonitorExporter] Failed to refresh access token: {:?}", e);
```
jmacd, line 486:
```suggestion
                                        otel_error!("Failed to refresh access token", error = ?e);
```

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
jmacd: I'd prefer to see anything general-purpose for monitoring machine health move into a stand-alone library, fwiw.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
gouslu: I want to create a PR with this type of change across the folder. Already have TODO for that.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/exporter.rs
gouslu: I agree, I'll remove these for GA, I'm keeping them for now for my own dev purposes.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
```diff
@@ -2,27 +2,28 @@
...
 use super::Error;
 use super::config::{AuthConfig, AuthMethod};
 
-#[derive(Clone)]
+/// Minimum delay between token refresh retry attempts in seconds.
+const MIN_RETRY_DELAY_SECS: f64 = 5.0;
+/// Maximum delay between token refresh retry attempts in seconds.
+const MAX_RETRY_DELAY_SECS: f64 = 30.0;
+/// Maximum jitter percentage (±10%) to add to retry delays.
+const MAX_RETRY_JITTER_RATIO: f64 = 0.10;
```
jmacd, line 19: FYI #1913.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
```diff
@@ -78,15 +53,49 @@ impl Auth {
...
-        let mut cached = self.cached_token.write().await;
-        *cached = None;
+    pub async fn get_token(&mut self) -> Result<AccessToken, Error> {
+        let mut attempt = 0_i32;
+        loop {
+            attempt += 1;
+
+            match self.get_token_internal().await {
+                Ok(token) => {
+                    println!(
```
jmacd, line 66: I'd like to see all println! replaced with otel_info!

jmacd: (non-blocking, sorry)

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
gouslu: I'll do it in a separate PR as well, thanks!

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/heartbeat.rs
```diff
@@ -0,0 +1,566 @@
...
+
+#[inline]
+fn default_heartbeat_os_name() -> String {
+    System::name().unwrap_or_else(|| std::env::consts::OS.to_string())
+}
+
+#[inline]
+fn default_heartbeat_computer() -> String {
+    std::env::var("ARM_RESOURCE_ID")
```
cijothomas, line 63: why we prefer ARM related env variables here?

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/heartbeat.rs
```diff
@@ -0,0 +1,566 @@
...
+
+/// Parse OS version into (major, minor) components
+/// e.g., "22.04.3" -> ("22", "04")
+/// e.g., "10 (22H2)" -> ("10", "22H2")
+/// e.g., "5.15.0-generic" -> ("5", "15")
+fn parse_os_version() -> (String, String) {
+    let version = System::os_version().unwrap_or_default();
+
+    if version.is_empty() {
+        return ("Unknown".to_string(), "Unknown".to_string());
```
cijothomas, line 76: nit: we don't need unwrap_or_default above, only to return Unknown here!

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/heartbeat.rs
```diff
@@ -0,0 +1,566 @@
...
+fn default_heartbeat_os_major_version() -> String {
+    std::env::var("POD_NAME").unwrap_or_else(|_| {
+        let (major, _) = parse_os_version();
+        major
+    })
+}
+
+#[inline]
+fn default_heartbeat_os_minor_version() -> String {
+    AZURE_MONITOR_EXPORTER_URN.to_string()
```
cijothomas, line 107: urn is not OS_minor_version

PR #1856, opened 2026-01-22
[otap-df-quiver] Track highest segment sequence during startup to avoid ID reuse

### LINKED ISSUE #1855: [otap-df-quiver] Segment sequence collision after restart causes write failures
On engine restart, Quiver does not scan existing segment files to determine the next segment sequence number. This causes the engine to attempt creating segments with sequence numbers that already exist, resulting in write failures because finalized segment files are marked read-only.

### Steps to Reproduce

1. Start Quiver engine and ingest data until segments are finalized (e.g., `00000000.qseg`, `00000001.qseg`)
2. Shut down the engine
3. Restart the engine
4. Ingest more data until a new segment is finalized
5. Observe error when attempting to create `00000000.qseg`

### Expected Behavior

After restart, new segments should use sequence numbers that don't conflict with existing files. If `00000001.qseg` exists, the next segment should be `00000002.qseg`.

### Actual Behavior

`next_segment_seq` is initialized to `0` on every startup, ignoring existing segment files. When the engine attempts to finalize a new segment, it tries to create a file that already exists. Since finalized segments are marked read-only, this results in a permission error.

### Root Cause

In `QuiverEngine::open()`, the segment store's `scan_existing()` is called for recovery but its return value (list of found segments) is ignored. The `next_segment_seq` atomic is always initialized to `0`.

### Proposed Fix

After calling `scan_existing()`, iterate over the discovered segments and set `next_segment_seq` to `max(segment_seq) + 1`.

AaronRM:
# Change Summary

Fixes #1855 by ensuring the Quiver engine scans existing segment files and starts with the next sequence number on recovery.

* Closes #1855 

Add `restart_recovers_segment_sequence_numbers` to validate that the issue is resolved & manual validation

No.

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/engine.rs
+++ b/rust/otap-dataflow/crates/quiver/src/engine.rs
@@ -255,9 +255,9 @@ impl QuiverEngine {
         let mut next_segment_seq = 0u64;
         match segment_store.scan_existing() {
             Ok(found_segments) => {
-                for (seq, _bundle_count) in &found_segments {
-                    // Next segment seq should be 1 + the highest found
-                    next_segment_seq = next_segment_seq.max(seq.raw() + 1);
+                // scan_existing returns segments sorted by sequence, so last is highest
+                if let Some((seq, _bundle_count)) = found_segments.last() {
+                    next_segment_seq = seq.raw() + 1;
                 }
                 if !found_segments.is_empty() {
                     tracing::info!(


rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -3142,4 +3159,127 @@ mod tests {
...
+
+        // === Second engine run: verify sequence recovery ===
+        let engine2 = QuiverEngine::open(config.clone(), test_budget())
+            .await
+            .expect("second engine open");
+
+        // The next segment sequence should be >= segments created in first run
+        // (it tracks total segments ever created, not just loaded)
+        assert!(
+            engine2.total_segments_written() >= segments_before_restart,
```
lalitb, line 3229: shouldn't the total segments after restart be exactly equal until new segments are written? In that case, we should have check with `assert_eq`?

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -251,11 +251,28 @@ impl QuiverEngine {
...
-                error = %e,
-                "failed to scan existing segments during startup; continuing with empty store"
-            );
+        // We need to track the highest segment sequence found to avoid reusing IDs
+        let mut next_segment_seq = 0u64;
+        match segment_store.scan_existing() {
+            Ok(found_segments) => {
+                for (seq, _bundle_count) in &found_segments {
+                    // Next segment seq should be 1 + the highest found
+                    next_segment_seq = next_segment_seq.max(seq.raw() + 1);
```
utpilla: It looks like you already sort the segments by sequence number before returning the vec in `scan_existing()`. So, it should be enough to just check the last segment instead of iterating through each one of them.

rust/otap-dataflow/crates/quiver/src/engine.rs
AaronRM: It could be possible (esp. if the test setup code changes) that `shutdown()` causes another segment to get written out. In that case, it would be `>` instead of `==`, so `>=` is the more robust check here.

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -251,11 +251,28 @@ impl QuiverEngine {
...
+                        next_segment_seq,
+                        "recovered existing segments from previous run"
+                    );
+                }
+            }
+            Err(e) => {
+                tracing::warn!(
+                    error = %e,
+                    "failed to scan existing segments during startup; continuing with empty store"
+                );
```
jmacd, line 274: This makes me wonder about the risk of a bug in scan_existing causing problems at runtime. Would a user ever want to have a hard failure at this spot?

Issue #1817, opened 2026-01-17
Improve load-generator


cijothomas:
For perf tests, we use df-engine with fake-receiver. Currently it generates traffic from otel semantic convention. We noticed that it takes a lot of cpu/memory to saturate a target 1-core engine.

This issue is opened to improve the fake-receiver to generate data more efficiently. (like reusing/replaying same data etc., similar to https://github.com/open-telemetry/otel-arrow/blob/main/tools/pipeline_perf_test/load_generator/loadgen.py#L218-L243)

This is important to continue scaling tests, as we run out of cores in lab machine🤣 There are 12+ load-gen cores, trying to saturation 1-core engine (https://github.com/open-telemetry/otel-arrow/pull/1810/changes)

PR #1857, opened 2026-01-22
Improve fake-signal-generator to better suit it for use as load generator.

cijothomas:
# Change Summary

Part of https://github.com/open-telemetry/otel-arrow/issues/1817
Fake-generator was generating new batches continuously, taking up CPU itself. For load-tests, we want to minimize the CPU taken by the fake-generator, so this PR adds additional options to it to re-use batches of telemetry.
It also adds option to generate data using hardcodes values, while maintaining existing ability to generate based on OTel semantic conventions.
When using hardcoded values, each log is designed to be approximately 300KB in size, similar to the ones from semantic convention. (We can move to 1 KB size in a future version)

Locally ran perf tests.

Fake-signal now supports more options.

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/error.rs
+++ b/rust/otap-dataflow/crates/engine/src/error.rs
@@ -394,6 +394,24 @@ pub enum Error {
         /// Prost error string
         reason: String,
     },
+
+    /// No default output port is configured when multiple ports are connected.
+    #[error(
+        "Ambiguous default out port for node {node}: multiple ports connected and no default configured"
+    )]
+    NoDefaultOutPort {
+        /// The node that has no default port configured.
+        node: NodeId,
+    },
+
+    /// An unknown output port was specified.
+    #[error("Unknown out port '{port}' for node {node}")]
+    UnknownOutPort {
+        /// The node where the unknown port was referenced.
+        node: NodeId,
+        /// The name of the unknown port.
+        port: PortName,
+    },
 }
 
 impl Error {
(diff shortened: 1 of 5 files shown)


rust/otap-dataflow/crates/otap/src/fake_data_generator.rs
```diff
@@ -39,10 +36,12 @@ pub mod attributes;
...
-pub mod fake_signal;
 /// fake signal metrics implementation
 pub mod metrics;
+/// generates signals based on OTel semantic conventions registry
+pub mod semconv_signal;
+/// Static hardcoded signal generators for lightweight load testing
+pub mod static_signal;
 
 /// The URN for the fake data generator receiver
 pub const OTAP_FAKE_DATA_GENERATOR_URN: &str = "urn:otel:otap:fake_data_generator:receiver";
```
lquerel, line 47: We could take advantage of this PR to align the URN with the pattern discussed during the last SIG, i.e. `urn:otel:<id>:receiver`. I left `id` as a placeholder because I was thinking we could also use this opportunity to give this receiver a better name. What do you think about <id>=`traffic_gen`?

This kind of change could of course be done in a follow-up PR as well, I mainly wanted to surface the idea while reviewing this PR.

cijothomas: Agree. https://github.com/open-telemetry/otel-arrow/issues/1867 opened for tracking in follow up.

rust/otap-dataflow/benchmarks/benches/otlp_bytes/main.rs
```diff
@@ -92,5 +95,27 @@ fn otlp_pdata_to_bytes_logs(c: &mut Criterion) {
     group.finish();
 }
 
-criterion_group!(otlp_bytes, otlp_pdata_to_bytes_logs);
+/// Benchmark cloning OtapPayload containing OtlpProtoBytes (logs).
+/// This confirms that clone is cheap due to bytes::Bytes being reference-counted.
```
cijothomas, line 99: I added this to confirm that the clone is cheap, because the fake-generator creates a batch at startup, and keep replaying it (by cloning). If the clone is expensive, then it'll eat up CPU cycles!
But this benchmark confirmed that the clone is cheap (few ns in my box)

Issue #1832, opened 2026-01-20
Proposal 5 - Better configuration parsing and error reporting


lquerel:
The current configuration parsing relies on `serde` and format-specific crates such as `serde_yaml`. This approach has two important limitations:

- `serde` does not provide precise error localization within configuration files, in particular line and column information for structural errors.
- The `serde_yaml` crate is deprecated, which raises maintainability and long-term support concerns.

This proposal suggests replacing `serde` with the `facet` ecosystem (`facet_yaml`, `facet_json`,...) for configuration parsing.

### Rationale

Adopting `facet` would address the limitations identified above:

- `facet` provides more precise and actionable error reporting, including accurate line and column information.
- The `facet` ecosystem offers actively maintained parsers for YAML and JSON.
- `facet` supports advanced schema features that align well with the goals of a stable and evolvable configuration model.

## Scope

- This change applies to configuration parsing only.
- The internal configuration model remains unchanged.
- Error messages and validation diagnostics become part of the user-facing contract.

This change is expected to significantly improve the developer and operator experience when authoring and validating configuration files.

lquerel:
@jmacd I updated this issue to take into account your env var requirement.

PR #1861, opened 2026-01-22
Internal Telemetry Receiver logging provider

### LINKED ISSUE #1736: Internal logging bridge, internral logging receiver

jmacd:

Final part of #1771.

Fixes #1736.

New tests. A new configs/internal-telemetry.yaml demonstrates the ITR configuration with the new console exporter.

Yes. See README.md update.

### DIFF
--- a/rust/otap-dataflow/configs/engine-conf/continuous_benchmark.yaml
+++ b/rust/otap-dataflow/configs/engine-conf/continuous_benchmark.yaml
@@ -17,15 +17,47 @@ pipeline_groups:
   continuous_benchmark:
     pipelines:
       # ======================================================================
-      # Pipeline generating traffic
+      # Pipelines generating traffic
       # ======================================================================
-      traffic_gen:
+      # First group of traffic generators based a static pre-generated dataset
+      traffic_gen1:
+        quota:
+          core_allocation:
+            type: core_count
+            count: 15
+
+        nodes:
+          receiver:
+            kind: receiver
+            plugin_urn: "urn:otel:otap:fake_data_generator:receiver"
+            out_ports:
+              out_port:
+                destinations:
+                  - exporter
+                dispatch_strategy: round_robin
+            config:
+              data_source: static
+              generation_strategy: pre_generated
+              traffic_config:
+                signals_per_second: 100000
+                max_signal_count: null
+                metric_weight: 0
+                trace_weight: 0
+                log_weight: 30
+          exporter:
+            kind: exporter
+            plugin_urn: "urn:otel:otlp:exporter"
+            config:
+              grpc_endpoint: "http://127.0.0.1:4327"
+      # Second group of traffic generators based on on-the-fly data generation
+      # derived from semantic conventions
+      traffic_gen2:
         quota:
           core_allocation:
             type: core_set
             set:
               - type: CoreRange
-                start: 10
+                start: 21
                 end: 35
 
         nodes:
@@ -49,7 +81,7 @@ pipeline_groups:
             kind: exporter
             plugin_urn: "urn:otel:otlp:exporter"
             config:
-              grpc_endpoint: "http://127.0.0.1:4327"
+              grpc_endpoint: "http://127.0.0.1:4337"
 
       # ======================================================================
       # System Under Test pipeline
@@ -61,9 +93,9 @@ pipeline_groups:
             set:
               - type: CoreRange
                 start: 0
-                end: 0
+                end: 1
         nodes:
-          otlp_recv:
+          otlp_recv1:
             kind: receiver
             plugin_urn: "urn:otel:otlp:receiver"
             out_ports:
@@ -74,18 +106,17 @@ pipeline_groups:
             config:
               listening_addr: "127.0.0.1:4327"
               wait_for_result: true
-
-          otap_recv:
+          otlp_recv2:
             kind: receiver
-            plugin_urn: "urn:otel:otap:receiver"
+            plugin_urn: "urn:otel:otlp:receiver"
             out_ports:
               out_port:
                 destinations:
-                  - otap_exporter
+                  - router
                 dispatch_strategy: round_robin
             config:
-              listening_addr: "127.0.0.1:4329"
-              response_stream_channel_size: 256
+              listening_addr: "127.0.0.1:4337"
+              wait_for_result: true
 
           router:
             kind: processor
@@ -133,10 +164,6 @@ pipeline_groups:
             plugin_urn: "urn:otel:noop:exporter"
             config:
 
-          otap_exporter:
-            kind: exporter
-            plugin_urn: "urn:otel:noop:exporter"
-            config:
 
       # ======================================================================
       # Performance No-Op pipeline to measure receiver performance
--- a/rust/otap-dataflow/crates/engine/src/attributes.rs
+++ b/rust/otap-dataflow/crates/engine/src/attributes.rs
@@ -82,6 +82,13 @@ pub struct ChannelAttributeSet {
     #[compose]
     pub node_attrs: NodeAttributeSet,
 
+    /// Port name for the channel endpoint.
+    ///
+    /// On the sender side, this is the port to which the channel is connected.
+    /// On the receiver side, this defaults to the node's input port.
+    #[attribute(key = "node.port")]
+    pub node_port: Cow<'static, str>,
+
     /// Unique channel identifier (in scope of the pipeline).
     #[attribute(key = "channel.id")]
     pub channel_id: Cow<'static, str>,
--- a/rust/otap-dataflow/crates/engine/src/channel_metrics.rs
+++ b/rust/otap-dataflow/crates/engine/src/channel_metrics.rs
@@ -249,6 +249,7 @@ mod tests {
         let (sender, receiver) = mpsc::Channel::new(1);
         let channel_entity_key = pipeline_ctx.register_channel_entity(
             "test:sender".into(),
+            "out".into(),
             CHANNEL_KIND_PDATA,
             CHANNEL_MODE_LOCAL,
             CHANNEL_TYPE_MPSC,
@@ -278,6 +279,7 @@ mod tests {
         let sender = LocalSender::mpsc(sender);
         let channel_entity_key = pipeline_ctx.register_channel_entity(
             "test:receiver".into(),
+            "input".into(),
             CHANNEL_KIND_PDATA,
             CHANNEL_MODE_LOCAL,
             CHANNEL_TYPE_MPSC,
--- a/rust/otap-dataflow/crates/engine/src/context.rs
+++ b/rust/otap-dataflow/crates/engine/src/context.rs
@@ -306,6 +306,7 @@ impl PipelineContext {
     pub fn channel_attribute_set(
         &self,
         channel_id: Cow<'static, str>,
+        node_port: Cow<'static, str>,
         channel_kind: &'static str,
         channel_mode: &'static str,
         channel_type: &'static str,
@@ -313,6 +314,7 @@ impl PipelineContext {
     ) -> ChannelAttributeSet {
         ChannelAttributeSet {
             node_attrs: self.node_attribute_set(),
+            node_port,
             channel_id,
             channel_kind: Cow::Borrowed(channel_kind),
             channel_mode: Cow::Borrowed(channel_mode),
@@ -326,6 +328,7 @@ impl PipelineContext {
     pub fn register_channel_entity(
         &self,
         channel_id: Cow<'static, str>,
+        node_port: Cow<'static, str>,
         channel_kind: &'static str,
         channel_mode: &'static str,
         channel_type: &'static str,
@@ -333,6 +336,7 @@ impl PipelineContext {
     ) -> EntityKey {
         let attrs = self.channel_attribute_set(
             channel_id,
+            node_port,
             channel_kind,
             channel_mode,
             channel_type,
--- a/rust/otap-dataflow/crates/engine/src/entity_context.rs
+++ b/rust/otap-dataflow/crates/engine/src/entity_context.rs
@@ -365,6 +365,7 @@ mod tests {
         let channel_id: Cow<'static, str> = "chan:pdata".into();
         let out_key = source_ctx.register_channel_entity(
             channel_id.clone(),
+            "out".into(),
             CHANNEL_KIND_PDATA,
             CHANNEL_MODE_LOCAL,
             CHANNEL_TYPE_MPSC,
@@ -373,6 +374,7 @@ mod tests {
         source_handle.add_output_channel_key("out".into(), out_key);
         let in_key = dest_ctx.register_channel_entity(
             channel_id,
+            "input".into(),
             CHANNEL_KIND_PDATA,
             CHANNEL_MODE_LOCAL,
             CHANNEL_TYPE_MPSC,
@@ -381,6 +383,7 @@ mod tests {
         dest_handle.set_input_channel_key(in_key);
         let ctrl_key = source_ctx.register_channel_entity(
             "chan:ctrl".into(),
+            "input".into(),
             CHANNEL_KIND_CONTROL,
             CHANNEL_MODE_LOCAL,
             CHANNEL_TYPE_MPSC,
--- a/rust/otap-dataflow/crates/engine/src/exporter.rs
+++ b/rust/otap-dataflow/crates/engine/src/exporter.rs
@@ -216,6 +216,7 @@ impl<PData> ExporterWrapper<PData> {
                     (Ok(sender), Ok(receiver)) => {
                         let channel_entity_key = pipeline_ctx.register_channel_entity(
                             control_channel_id(&node_id),
+                            "input".into(),
                             CHANNEL_KIND_CONTROL,
                             CHANNEL_MODE_LOCAL,
                             CHANNEL_TYPE_MPSC,
@@ -291,6 +292,7 @@ impl<PData> ExporterWrapper<PData> {
                     (Ok(sender), Ok(receiver)) => {
                         let channel_entity_key = pipeline_ctx.register_channel_entity(
                             control_channel_id(&node_id),
+                            "input".into(),
                             CHANNEL_KIND_CONTROL,
                             CHANNEL_MODE_SHARED,
                             CHANNEL_TYPE_MPSC,
--- a/rust/otap-dataflow/crates/engine/src/lib.rs
+++ b/rust/otap-dataflow/crates/engine/src/lib.rs
@@ -27,13 +27,13 @@ use async_trait::async_trait;
 use context::PipelineContext;
 pub use linkme::distributed_slice;
 use otap_df_config::{
-    PortName,
+    PipelineGroupId, PipelineId, PortName,
     node::{DispatchStrategy, NodeUserConfig},
     pipeline::PipelineConfig,
 };
 use otap_df_telemetry::INTERNAL_TELEMETRY_RECEIVER_URN;
 use otap_df_telemetry::InternalTelemetrySettings;
-use otap_df_telemetry::otel_debug;
+use otap_df_telemetry::{otel_debug, otel_debug_span};
 use std::borrow::Cow;
 use std::fmt::Debug;
 use std::num::NonZeroUsize;
@@ -294,12 +294,14 @@ impl<PData: 'static + Clone + Debug> PipelineFactory<PData> {
 
     /// Builds a runtime pipeline from the given pipeline configuration.
     ///
-    /// Steps:
-    /// - Create all runtime nodes based on the pipeline configuration.
-    /// - Analyze both the local vs shared nature of each pair of connected nodes and the nature of
-    ///   the hyper-edges between them to determine the best channel type.
-    /// - Assign channels to the source nodes and their destination nodes based on the previous
-    ///   analysis.
+    /// Main phases:
+    /// 1) Create runtime nodes and register telemetry.
+    /// 2) Plan hyper edge wiring: resolve destinations, pick channel type (shared/local,
+    ///    MPSC/MPMC), create channel endpoints, and register channel metrics.
+    /// 3) Apply wiring: attach senders to source ports and receivers to destination nodes,
+    ///    then publish collected channel metrics on the pipeline.
+    ///
+    /// [config] -> [nodes] -> [hyper-edges] -> [wiring plan] -> [pipeline]
     ///
     /// The `internal_telemetry` settings are injected into any receiver with the
     /// `INTERNAL_TELEMETRY_RECEIVER_URN` plugin URN, enabling it to consume logs
@@ -313,24 +315,19 @@ impl<PData: 'static + Clone + Debug> PipelineFactory<PData> {
         let mut receivers = Vec::new();
         let mut processors = Vec::new();
         let mut exporters = Vec::new();
-        let mut receiver_names = HashMap::new();
-        let mut processor_names = HashMap::new();
-        let mut exporter_names = HashMap::new();
-        let mut node_contexts = HashMap::new();
-        let mut node_telemetry_handles = HashMap::new();
-        let mut nodes = NodeDefs::default();
-        let mut channel_metrics = ChannelMetricsRegistry::default();
+        let mut build_state = BuildState::new();
 
         let pipeline_group_id = pipeline_ctx.pipeline_group_id();
         let pipeline_id = pipeline_ctx.pipeline_id();
         let core_id = pipeline_ctx.core_id();
 
-        otel_debug!(
-            "pipeline.build.start",
+        let span = otel_debug_span!(
+            "pipeline.build",
             pipeline_group_id = pipeline_group_id.as_ref(),
             pipeline_id = pipeline_id.as_ref(),
             core_id = core_id
         );
+        let _enter = span.enter();
 
         let channel_metrics_enabled = config.pipeline_settings().telemetry.channel_metrics;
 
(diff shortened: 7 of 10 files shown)


rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -1573,4 +1608,68 @@ mod tests {
...
+
+        let internal = internal_config.unwrap();
+
+        // Should have the internal pipeline settings
+        assert_eq!(internal.settings.default_node_ctrl_msg_channel_size, 50);
+        assert_eq!(internal.settings.default_pdata_channel_size, 50);
+
+        // Should have the internal nodes as its main nodes
+        assert_eq!(internal.nodes.len(), 2);
+
```
utpilla, line 1668: Add assertions to verify the node names against the provided config

```rust
assert_eq!(
    internal.nodes["itr"].plugin_urn.as_ref(),
    "urn:otel:otap:internal_telemetry:receiver"
);
assert_eq!(
    internal.nodes["console"].plugin_urn.as_ref(),
    "urn:otel:console:exporter"
);
```

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -114,11 +116,11 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
 
         // Create the telemetry system, pass the observed state store
         // as the admin reporter for console_async support.
-        let telemetry_system =
+        // The ITS logs receiver is returned if any provider uses ITS mode.
```
utpilla: We should update these comments. Maybe something like:

```rust
// Create the telemetry system. The logging_evt_reporter is used for
// ConsoleAsync provider mode (logs go through the ObservedStateStore).
// The ITS logs receiver is returned if any provider uses ITS mode.
```

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -521,6 +665,79 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
...
+        pipeline_context: PipelineContext,
+        obs_evt_reporter: ObservedEventReporter,
+        metrics_reporter: MetricsReporter,
+        pipeline_ctrl_msg_tx: PipelineCtrlMsgSender<PData>,
+        pipeline_ctrl_msg_rx: PipelineCtrlMsgReceiver<PData>,
+        tracing_setup: TracingSetup,
+        internal_telemetry: Option<InternalTelemetrySettings>,
+        startup_tx: std_mpsc::Sender<Result<(), Error>>,
+    ) -> Result<Vec<()>, Error> {
+        // No core pinning for internal pipeline - it's lightweight
```
lalitb: Curious about the cross-thread channel here in the context of the thread-per-core design. Since logs from pinned engine threads flow through `flume::bounded` to this unpinned thread, do you anticipate any cache/NUMA considerations at high log volumes?
Not a blocker - just wondering if batching per-core or optional affinity config or expose dedicated internal pipeline per NUMA domain might be worth considering in a future iteration.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -451,6 +470,129 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
...
+    ) -> Result<Option<(String, thread::JoinHandle<Result<Vec<()>, Error>>)>, Error> {
+        // Find the first pipeline with an internal config
+        let internal_config: Option<PipelineConfig> = pipeline_groups
+            .iter()
+            .flat_map(|(_, group)| group.pipelines.iter())
+            .find_map(|(_, pipeline)| pipeline.extract_internal_config());
+
+        let internal_config = match internal_config {
+            Some(config) => config,
+            None => return Ok(None), // No internal pipeline configured
```
utpilla: Maybe log a message here as well indicating that internal pipeline is being skipped.

rust/otap-dataflow/crates/controller/src/lib.rs
lalitb: To clarify my earlier point - the pipeline would be lightweight only if the log volume is low. The internal telemetry thread is `unpinned`, so the scheduler can run it on any core, including ones already hosting pinned pipeline threads, which can introduce contention and jitter. The actual impact depends on available idle cores and how busy the pinned threads are, but without affinity the ITS thread can end up sharing those cores.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -451,6 +470,129 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
...
+
+        // Create control message channel for internal pipeline
+        let (internal_ctrl_tx, internal_ctrl_rx) = pipeline_ctrl_msg_channel(
+            internal_config
+                .pipeline_settings()
+                .default_pipeline_ctrl_msg_channel_size,
+        );
+
+        // Create a channel to signal startup success/failure
+        let (startup_tx, startup_rx) = std_mpsc::channel::<Result<(), Error>>();
```
utpilla: Use a bounded channel to make the intent clear about this being a oneshot channel.

```suggestion
        let (startup_tx, startup_rx) = std_mpsc::channel::<Result<(), Error>>(1);
```

rust/otap-dataflow/crates/telemetry/src/lib.rs
```diff
@@ -129,16 +152,19 @@ impl InternalTelemetrySystem {
     /// This creates:
     /// 1. The internal metrics subsystem (registry, collector, reporter, dispatcher)
     /// 2. The OpenTelemetry SDK providers (meter, and optionally logger)
+    /// 3. If any provider uses ITS mode, a dedicated logs channel for the ITR
     ///
-    /// The `admin_reporter` is required for async logging modes (`ConsoleAsync`, future `ITS`).
+    /// The `admin_reporter` is required for async logging modes (`ConsoleAsync`).
```
utpilla: This name is a bit misleading as it suggests some relation with admin threads. Could we rename this to something more intuitive such as `console_async_reporter`?

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
```diff
@@ -0,0 +1,175 @@
...
+        effect_handler: &local::EffectHandler<OtapPdata>,
+        log_event: LogEvent,
+        buf: &mut ProtoBuffer,
+        resource_bytes: &Bytes,
+    ) -> Result<(), Error> {
+        encode_export_logs_request(buf, &log_event, resource_bytes);
+
+        let pdata = OtapPdata::new(
+            Context::default(),
+            OtlpProtoBytes::ExportLogsRequest(Bytes::copy_from_slice(buf.as_ref())).into(),
```
utpilla: It might be better to just allocate a new `ProtoBuffer` here and use `ProtoBuffer::into_bytes()` instead of using `Bytes::copy_from_sllice()` which does allocation as well as copying.

```rust
async fn send_log_event(...) -> Result<(), Error> {
    let mut buf = ProtoBuffer::with_capacity(512);
    encode_export_logs_request(&mut buf, &log_event, resource_bytes);
    
    let pdata = OtapPdata::new(
        Context::default(),
        OtlpProtoBytes::ExportLogsRequest(buf.into_bytes()).into(),  // No copy
    );
    effect_handler.send_message(pdata).await?;
    Ok(())
}
```

rust/otap-dataflow/configs/internal-telemetry.yaml
```diff
@@ -0,0 +1,55 @@
...
+          - console
+        dispatch_strategy: round_robin
+    config: {}
+  console:
+    kind: exporter
+    plugin_urn: "urn:otel:console:exporter"
+    config: {}
+
+service:
+  telemetry:
```
cijothomas, line 45: let add comment about how *this* telemetry section is different from the *internal* one.

jmacd: I am concerned we should consolidate one way or other. I don't have a proposal on how to proceed, presently we have three semi-related areas including service::telemetry::*  pipeline::settings::telemetry, EngineSettings::observed_state. I'm not sure what to do because of the overlap with OTel declarative configuration.

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
```diff
@@ -0,0 +1,175 @@
...
+                             // Ignore other control messages
+                        }
+                    }
+                }
+
+                // Receive logs from the channel
+                result = logs_receiver.recv_async() => {
+                    match result {
+                        Ok(ObservedEvent::Log(log_event)) => {
+                            self.send_log_event(&effect_handler, log_event, &mut buf, &resource_bytes).await?;
```
utpilla: Should we consider batching the events? Currently, each log becomes a separate `ExportLogsRequest`. Batching can help better amortize the per-message overhead:
- Fewer messages sent to the channel downstream
- Resource bytes would be copies only once per batch etc.

rust/otap-dataflow/crates/telemetry/README.md
```diff
@@ -55,9 +55,61 @@ Rust struct definition.
 
 See the [telemetry-macros crate](../telemetry-macros) for details.
 
+## Logging Macros
+
+There are internal macros defined in `otal_df_telemetry` with names
```
utpilla:
```suggestion
There are internal macros defined in `otap_df_telemetry` with names
```

rust/otap-dataflow/crates/controller/src/lib.rs
jmacd: Yes, future work! The current metrics collection path is also global / not pinned, but the design documentation calls for a future step where we factor into per-NUMA-region telemetry collectors. I would expect at that time to transition to per-NUMA-region ITRs: https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/telemetry/README.md

rust/otap-dataflow/crates/telemetry/README.md
```diff
@@ -55,9 +55,61 @@ Rust struct definition.
...
+crate](https://docs.rs/env_logger/latest/env_logger/), making the
+`RUST_LOG` environment variable available for controlling internal
+logging.
+
+There are four aspects that can be configured:
+
+- `engine`: configure logging for main engine threads
+- `global`: configure logging for the global `tracing` subscriber
+- `admin`: configure logging for the engine administrative threads
+- `internal`: configure logging for components involved in internal logging.
```
utpilla: I think we could provide clearer descriptions of these:

```suggestion
- `engine`: logging for pipeline threads that run dataflow processing (receivers, processors, exporters)
- `global`: fallback logging for code outside engine/admin contexts (e.g., libraries, startup code)
- `admin`: logging for administrative threads (metrics aggregation, observed state store, controller tasks)
- `internal`: logging for the internal telemetry pipeline itself; restricted to `console_direct` or `noop` to avoid feedback loops
```

rust/otap-dataflow/configs/internal-telemetry.yaml
lquerel: Just for tracking I created this issue https://github.com/open-telemetry/otel-arrow/issues/1872
No real proposal yet, but if you can add a more detailed description of the current mess feel free to edit the issue description. 
Based on that we will figure out a way to put some order there :-)

rust/otap-dataflow/crates/telemetry/README.md
```diff
@@ -55,9 +55,61 @@ Rust struct definition.
...
+
+These modes are configured through the `service::telemetry::logs::providers`
+field, with the following choices:
+
+- `its`: configure the Internal Telemetry System using nodes
+  defined in the pipeline's `internal` nodes section. These nodes
+  are configured in a dedicated engine thread.
```
utpilla: Do we want to use the word "engine" thread? It makes it sound like it's one of the main pipeline threads.

```suggestion
  are configured in a dedicated thread.
```

rust/otap-dataflow/crates/telemetry/README.md
```diff
@@ -55,9 +55,61 @@ Rust struct definition.
...
+- `console_async`: configure asynchronous console logging. In this
+  mode log records are printed to the console from the administrative
+  component.
+- `console_direct`: configure synchronous logging. In this mode, logging
+  interrupts the calling thread to print each log statement.
```
utpilla: I think we can be more specific here.

```suggestion
- `console_async`: configure asynchronous console logging. In this
  mode log records are printed to the console by the observed-state-store thread, avoiding 
  blocking the caller.
- `console_direct`: configure synchronous logging. This mode blocks the calling 
  thread to print each log statement immediately.
```

rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
```diff
@@ -291,3 +294,142 @@ pub fn level_to_severity_number(level: &Level) -> u8 {
...
+        buf.encode_string(RESOURCE_LOGS_SCHEMA_URL, url);
+    }
+}
+
+/// Encode an SDK Resource to bytes for later reuse.
+#[must_use]
+pub fn encode_resource_to_bytes(resource: &opentelemetry_sdk::Resource) -> bytes::Bytes {
+    let mut buf = ProtoBuffer::with_capacity(256);
+    encode_resource(&mut buf, resource.iter(), resource.schema_url());
+    bytes::Bytes::copy_from_slice(buf.as_ref())
```
utpilla: No need to use `Bytes::copy_from_slice()` which allocates and copies.

```suggestion
    buf.into_bytes()
```

rust/otap-dataflow/crates/telemetry/src/tracing_init.rs
```diff
@@ -117,6 +124,13 @@ impl ProviderSetup {
                 Dispatch::new(Registry::default().with(filter()).with(layer))
             }
 
+            ProviderSetup::ITS { reporter } => {
+                // ITS uses the same async layer as ConsoleAsync, but the
+                // receiver is consumed by the Internal Telemetry Receiver
```
utpilla: "receiver is consumed by the Internal Telemetry Receiver" sounds a bit confusing. We could either mention that we are referring to the flume receiver or just write something simple:

```rust
// ITS uses the same async layer as ConsoleAsync. The difference
// is that logs go to the Internal Telemetry Receiver (for OTLP export)
// instead of being printed to the console
```

rust/otap-dataflow/crates/telemetry/src/tracing_init.rs
```diff
@@ -432,12 +446,74 @@ mod tests {
...
+    #[test]
+    fn its_provider_all_levels() {
+        for level in [
+            LogLevel::Off,
+            LogLevel::Debug,
+            LogLevel::Info,
+            LogLevel::Warn,
+            LogLevel::Error,
+        ] {
+            let (reporter, _receiver) = test_reporter();
```
utpilla: To have a better test, we could use the `receiver` to assert that we receive the expected number of logs for each of the setups.

rust/otap-dataflow/crates/controller/src/lib.rs
jmacd:
```
        // TODO: This should be validated somewhere, that internal node are defined when
        // its is requested. Possibly we could fill in a default.
        match (
            internal_pipeline_handle.is_some(),
            telemetry_config.logs.providers.uses_its_mode(),
        ) {
            (false, true) => {
                otel_warn!("internal pipeline nodes not defined yet ITS mode requested")
            }
            (true, false) => {
                otel_warn!("internal pipeline nodes defined yet ITS mode not requested")
            }
            _ => {}
        };
```

rust/otap-dataflow/crates/controller/src/lib.rs
jmacd:
```
+        // Create the telemetry system. The console_async_reporter is passed when any
+        // providers use ConsoleAsync. The its_logs_receiver is passed when any
+        // providers use the ITS mode.
```

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -134,6 +136,23 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
             }
         }
 
+        // Spawn internal telemetry pipeline thread, if configured.
+        let _internal_pipeline_handle = Self::spawn_internal_pipeline_if_configured(
```
utpilla: We should use the `JoinHandle` for graceful shutdown. We could either add it to `threads` vec which contains the handles for the main pipeline threads or just shut it down separately.

rust/otap-dataflow/crates/telemetry/src/lib.rs
jmacd: Replaced with
```
    /// Creates a new [`InternalTelemetrySystem`] initialized with the given configuration.
    ///
    /// Depending on logging provider mode choices, multiple telemetry backends can be
    /// initialized:
    ///
    /// OpenTelemetry: the OTel logging provider is created if any
    /// service::telemetry::logs::providers uses this choice.  Note: the OTel meter
    /// provider is created unconditionally.
    ///
    /// ConsoleAsync: the ObservedEventReporer is passed in, having been created for
    /// use by the admin component unconditionally, to support the ConsoleAsync mode.
    ///
    /// ITS: if any logging provider is configured with for the internal telemetry system,
    /// an InternalReceiver will be returned.
    ///
    /// **Note:** The global tracing subscriber is NOT initialized here. Call
    /// `init_global_subscriber()` when ready to start logging.
```

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
jmacd: We should, but not in this PR. There are several considerations:
 
 1. I want to wait until after we add scope attributes, because we can sort the records by scope to reduce overhead
 2. Question whether batching after the channel is better than batching before the channel; in the case of the engine thread, we could use a thread-local batch and avoid allocating for each event, for example.
3. Why not direct to OTAP if we're being so elaborate?
4. How much better is it compared with letting `batch_processor` do the work?

rust/otap-dataflow/crates/telemetry/src/tracing_init.rs
jmacd: This led to a refactoring and cleanup. The InternalTelemetrySettings is a bit more proper now, see I've consolidated the two internal paths into a common provider setup, now.

https://github.com/open-telemetry/otel-arrow/pull/1861/commits/997ce6c0987c94a17a13c8165b84769bed6ad73c

rust/otap-dataflow/crates/telemetry/src/tracing_init.rs
jmacd: Fixed! I used a pattern like

```
drop(setup); // closes sender

let cnt = receiver.into_iter().count();
assert_eq!(cnt, ...);
```

rust/otap-dataflow/crates/controller/src/lib.rs
jmacd: Fixed. I changed threads to copy the DeployedPipelineKey (instead of a 3-tuple).

rust/otap-dataflow/configs/internal-telemetry.yaml
andborja: Once the internal telemetry pipeline is stable, we can remove the SDK integration. It replaces it in a better way.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -58,9 +58,11 @@ use otap_df_state::store::ObservedStateStore;
 use otap_df_telemetry::event::{EngineEvent, ErrorSummary, ObservedEventReporter};
 use otap_df_telemetry::reporter::MetricsReporter;
 use otap_df_telemetry::{
-    InternalTelemetrySystem, TracingSetup, otel_info, otel_info_span, otel_warn,
+    InternalTelemetrySettings, InternalTelemetrySystem, TracingSetup, otel_info, otel_info_span,
+    otel_warn,
 };
 use std::sync::Arc;
+use std::sync::mpsc as std_mpsc;
```
lquerel, line 65: I think tokio::sync::Notify will be a better fit for the signaling mechanism used later in the code.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -134,6 +139,37 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
...
+            telemetry_config.logs.providers.uses_its_provider(),
+        ) {
+            (false, true) => {
+                otel_warn!("ITS provider requested yet internal pipeline nodes not defined")
+            }
+            (true, false) => {
+                otel_warn!("internal pipeline nodes defined yet ITS provider not requested")
+            }
+            _ => {}
+        };
```
lquerel, line 182: Not for this PR, but would it be possible to imagine a solution where, when an ITS pipeline is defined, we automatically consider it as the one to use?

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -173,6 +209,17 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
         let mut threads = Vec::new();
         let mut ctrl_msg_senders = Vec::new();
 
+        // Add internal pipeline to threads list if present
+        if let Some((thread_name, handle)) = internal_pipeline_handle {
+            let ikey = Self::internal_pipeline_key();
+            threads.push((
+                thread_name,
+                0, // TODO: thread_id is not defined
```
lquerel: Can't we fix that ?

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -451,6 +486,123 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
...
+        );
+
+        // Create a channel to signal startup success/failure
+        let (startup_tx, startup_rx) = std_mpsc::sync_channel::<Result<(), Error>>(1);
+
+        let thread_name = "internal-pipeline".to_string();
+        let internal_evt_reporter = engine_evt_reporter.clone();
+        let internal_metrics_reporter = metrics_reporter.clone();
+
+        let handle = thread::Builder::new()
```
lquerel, line 573: To get the thread id, you could use something like.

```rust
let (tx, rx) = oneshot::channel();

let handle = thread::Builder::new()
    .name(thread_name.clone())
    .spawn(move || {
        let _ = tx.send(thread::current().id());
        // Self::run_internal_pipeline_thread...
    })
    .unwrap();

let tid = rx.await.expect("thread exited before sending id");
```

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -451,6 +486,123 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
...
+            core_id: 0, // TODO: Internal thread is unpinned, core_id is not accurate.
+        }
+    }
+
+    /// Spawns the internal telemetry pipeline if any pipeline has an internal config.
+    ///
+    /// Returns the thread handle if an internal pipeline was spawned
+    /// and waits for it to start, or None.
+    #[allow(clippy::too_many_arguments)]
+    fn spawn_internal_pipeline_if_configured(
```
lquerel, line 510: I’m wondering about thread pinning. I think we should pin this thread to core 0 (for now), since our pipeline engine relies on this concept, especially for CPU and memory metrics. It doesn’t cost much to do it in any case.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -521,6 +675,80 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
...
+        pipeline_config: PipelineConfig,
+        pipeline_factory: &'static PipelineFactory<PData>,
+        pipeline_context: PipelineContext,
+        obs_evt_reporter: ObservedEventReporter,
+        metrics_reporter: MetricsReporter,
+        pipeline_ctrl_msg_tx: PipelineCtrlMsgSender<PData>,
+        pipeline_ctrl_msg_rx: PipelineCtrlMsgReceiver<PData>,
+        tracing_setup: TracingSetup,
+        internal_telemetry: Option<InternalTelemetrySettings>,
+        startup_tx: std_mpsc::SyncSender<Result<(), Error>>,
```
lquerel: Like I suggested, I believe a tokio notify will be better here

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
```diff
@@ -0,0 +1,173 @@
...
+impl local::Receiver<OtapPdata> for InternalTelemetryReceiver {
+    async fn start(
+        self: Box<Self>,
+        mut ctrl_msg_recv: local::ControlChannel<OtapPdata>,
+        effect_handler: local::EffectHandler<OtapPdata>,
+    ) -> Result<TerminalState, Error> {
+        // Get the logs receiver channel from the effect handler
+        let logs_receiver = effect_handler
+            .logs_receiver()
+            .expect("InternalTelemetryReceiver requires a logs_receiver to be configured");
```
lquerel: We will change that once we have the named channel in place.

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
```diff
@@ -0,0 +1,173 @@
...
+
+        // Get the pre-encoded resource bytes
+        let resource_bytes = effect_handler
+            .resource_bytes()
```
lquerel: Not a big fan of polluting the effect handler of every receivers with something that seems to be unique for the ITS. We probably need to figure out a better approach in a future PR.

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
```diff
@@ -0,0 +1,173 @@
...
+                }
+
+                // Receive logs from the channel
+                result = logs_receiver.recv_async() => {
+                    match result {
+                        Ok(ObservedEvent::Log(log_event)) => {
+                            self.send_log_event(&effect_handler, log_event, &resource_bytes).await?;
+                        }
+                        Ok(ObservedEvent::Engine(_)) => {
+                            // Engine events are not yet processed
```
lquerel, line 152: Does that mean that the `/status` is no longer working? It's probably ok for a short period of time but we will need that in a future PR soon :-)

rust/otap-dataflow/crates/telemetry/src/lib.rs
```diff
@@ -74,14 +75,28 @@ pub use tracing::info_span as otel_info_span;
 pub use tracing::trace_span as otel_trace_span;
 pub use tracing::warn_span as otel_warn_span;
 
+/// The URN for the internal telemetry receiver.
+/// Defined here so it can be used by controller, engine, otap, and other crates.
+pub const INTERNAL_TELEMETRY_RECEIVER_URN: &str = "urn:otel:otap:internal_telemetry:receiver";
```
lquerel:
```suggestion
pub const INTERNAL_TELEMETRY_RECEIVER_URN: &str = "urn:otel:internal_telemetry:receiver";
```

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
```diff
@@ -0,0 +1,173 @@
...
+                            return Err(Error::ChannelRecvError(e));
+                        }
+                        _ => {
+                             // Ignore other control messages
+                        }
+                    }
+                }
+
+                // Receive logs from the channel
+                result = logs_receiver.recv_async() => {
```
lquerel, line 146: It's ok for this PR, but IMO medium term we will have a generic named_channel receiver, and a collection of ITS oriented processors/exporters to update internal states, or export signals outside.

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
jmacd: Done.

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
jmacd: Something terrible happened here. Let me clean this up!

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
jmacd: No. The choice made in service::telemetry::logs::providers controls where log events are delivered. The `EngineEvent` events continue going to the admin component, no change in `/status`.

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
jmacd: This sounds great.  Although, keep in mind that this receiver returns a half-encoded structured log event with (timestamp, entity_key(s), callsite/metadata, body/attributes). If we wanted to encode logs into a buffer before sending them on the channel, we could form whole OtlpProtoBytes messages, otherwise we would pay to encode Scope/Resource for every log event.

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
jmacd: In general, I don't understand how config validation works precisely or what the `#[derive(JsonSchema)]` does, so mostly you're seeing that I don't know how to prevent strange configurations from getting this far at startup.

rust/otap-dataflow/crates/controller/src/lib.rs
jmacd: Yes. I can also imagine inserting a simple pipeline it ITS is requested and none is configured, like just a `console_exporter`.

Part of https://github.com/open-telemetry/otel-arrow/pull/1861#discussion_r2723171279 and #1832.

rust/otap-dataflow/crates/controller/src/lib.rs
jmacd: I can't figure this! Will be glad to in a follow-up. The problem is to know the difference between success or failure. A tokio::sync::oneshot would be appropriate in async context.

rust/otap-dataflow/crates/controller/src/lib.rs
jmacd: https://github.com/open-telemetry/otel-arrow/pull/1861#discussion_r2723223229

rust/otap-dataflow/crates/controller/src/lib.rs
jmacd: This code snippet requires an asynchronous context, I think? I'm having trouble also because the existing runtime thread_ids are not assigned using thread::current().id(), they are integers representing the ordinal position of the requested_cores input. so I could use a value like requested_cores.len().

rust/otap-dataflow/crates/controller/src/lib.rs
jmacd: I removed `run_internal_pipeline_thread`, call `run_pipeline_thread` instead with an Option for telemetry settings and startup_tx. This makes less code and it means passing through the same CPU-pinning call.

I used the first CPU in the list of available CPUs.

utpilla: Not for this PR, but I was wondering should we assign a lower priority to internal pipeline's thread to prioritize the engine threads.

rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs
jmacd: I moved a the optional telemetry settings into PipelineContext, so there's no more impact on effect handler.

rust/otap-dataflow/crates/controller/src/lib.rs
jmacd: Since I understand that your requirement is to not re-use `thread_id` assignments, I've fixed this temporarily. Now, thread_id values will be unique across the engine but they are assigned using a sequence number.

Previously, they were the result of `requested_cores.enumerate()`.

Issue #1829, opened 2026-01-20
Proposal 2 - Outputs representation [config-model-v1]


lquerel:
The current representation of out-ports is verbose and exposes implementation details that are unnecessary for the majority of node configurations.

This proposal introduces:
- a simple output form for the common case of a single logical output
- an advanced output form for nodes exposing multiple named outputs

## Simple form (single logical output)

For nodes exposing a single logical output, two equivalent sub-forms are supported depending on the number of destinations.

### Single destination

```yaml
output: exporter
```

Semantics:
- `output` refers to the node’s default output.
- The value is a single destination node identifier.

### Multiple destinations

```yaml
outputs:
  - exporter
  - debug
```

Semantics:
- `outputs` refers to the node’s default output.
- Each entry is a destination node identifier.
- Dispatch behavior relies on well-defined defaults (i.e. balanced)

These two sub-forms are mutually exclusive and cover the most common configurations.

## Named outputs

For nodes exposing multiple named outputs, the named_outputs field is used.

```yaml
named_outputs:
  success:
    - exporter
  failure:
    - debug
```

Semantics:
- Each key under named_outputs represents a named output.
- The `default` key is reserved and MUST NOT be used with `named_outputs`.
- Each value follows the same rules as the simple form destinations.
- Named outputs MUST be explicitly declared by the node type.
- Nodes that do not declare multiple outputs MUST NOT use this form.

**Validation rules**:
- Only one of `output`, `outputs`, or `named_outputs` MAY be specified.
- `output` represents a single destination.
- `outputs` represents multiple destinations for a single logical output.
- `named_outputs` represents multiple named outputs.
- Destination identifiers MUST refer to valid node identifiers.
- Output normalization MUST occur before pipeline instantiation.

drewrelmas:
~~@lquerel~~

~~> Dispatch behavior relies on well-defined defaults (i.e. balanced)~~

~~Could you please clarify - are you proposing that `dispatch_strategy` will no longer be customizable? Or it will move to a parallel level in node config?~~

```yaml
type: otlp:receiver
outputs:
  - processor1
  - processor2
dispatch_strategy: broadcast
```

EDIT: This is discussed in #1830. Not relevant here.

lquerel:
@andborja I believe you had some concerns with the approach mentioned in this issue. Could you describe your alternative proposal so that we can review it? Thanks.

lquerel:
**Open question**

[...]

PR #1878, opened 2026-01-23
Fanout Processor

lalitb:
## Fan-out Processor Implementation

Implements all four discussed scenarios:

| Scenario | Config | Description |
|----------|--------|-------------|
| 1 | `mode: parallel, await_ack: primary` | Duplicate to all, wait for primary only |
| 2 | `mode: parallel, await_ack: all` | Duplicate to all, wait for all (with per-destination timeout) |
| 3 | `mode: sequential` | Send one-by-one, advance after ack |
| 4 | `fallback_for: <port>` | Failover to backup on nack/timeout |

### Why Stateful (not Stateless like Go collector)

The Go Collector's fanout is stateless because it uses **synchronous, blocking calls**:
```go
err := consumer.ConsumeLogs(ctx, ld)  // blocks until complete, error returns directly
```

Our OTAP engine uses async message passing with explicit ack/nack routing:

```rust
effect_handler.send_message_to(port, pdata).await?;  // returns immediately
// ack arrives later as separate NodeControlMsg::Ack
```
I explored making scenarios 1 and 3 stateless but hit three blockers:

1. **`subscribe_to()` mutates context** - Fanout must subscribe to receive acks, which pushes a frame onto the context stack. For correct upstream routing, we need the *original* pdata (pre-subscription). We cannot use `ack.accepted` from downstream.

2. **Downstream may mutate/drop payloads** - `into_parts()`, transformers, and filters mean we can't rely on getting intact pdata back in ack/nack messages.

3. **Sequential/fallback/timeout require coordination** - Need to know which destination is active, when to advance to the next, and when to trigger fallbacks or finish.

Even if downstream guaranteed returning intact payloads, we'd still need state for `await_all` completion tracking, fallback chains, and sequential advancement. The only gain would be a minor memory optimization (not storing `original_pdata`), not true statelessness.

Adopting Go's synchronous model would require fundamental engine architecture changes, not just fanout changes.

### Memory Optimizations

While full statelessness isn't possible, I have implemented fast paths to minimize allocations for common configurations:

| Configuration                                              | Fast Path        | State Per Request                              |
|-----------------------------------------------------------|------------------|------------------------------------------------|
| `await_ack: none`                                         | Fire-and-forget  | None (zero inflight tracking)                  |
| `parallel + primary + no fallback + no timeout`           | Slim primary     | Minimal (`request_id → original_pdata`)        |
| All other configs                                         | Full             | Complete endpoint tracking                     |

#### Fast Path Details

- **Fire-and-forget (`await_ack: none`)**  
  Bypasses all inflight state. Clone, send, and ACK upstream immediately.  
  Zero allocations per request.

- **Slim primary path**  
  Uses a tiny `HashMap<u64, OtapPdata>` instead of the full `Inflight` struct with `EndpointVec`.  
  Ignores non-primary ACKs and NACKs.

- **Full path**  
  Required for:
  - Sequential mode  
  - `await_all`  
  - Any fallback  
  - Any timeout  

  Tracks all endpoints and request state.

### Code Structure

`Inflight` holds per-request state:
- `original_pdata` - pre-subscription pdata, used for all upstream acks/nacks
- `endpoints[]` - per-destination status (`Acked`/`Nacked`/`InFlight`/`PendingSend`)
- `next_send_queue` - drives sequential mode advancement
- `completed_origins` - tracks completion for `await_ack: all`
- `timeout_at` - per-destination deadlines for timeout/fallback triggering

Not all fields are used for every scenario, but the overhead is minimal - empty HashSets don't allocate, SmallVec is inline for ≤4 items, and clone cost is O(1) for `bytes::Bytes`.

### Documentation

See [`crates/otap/src/fanout_processor/README.md`](crates/otap/src/fanout_processor/README.md) for configuration examples and behavior details.

### DIFF
--- a/rust/otap-dataflow/configs/engine-conf/continuous_benchmark.yaml
+++ b/rust/otap-dataflow/configs/engine-conf/continuous_benchmark.yaml
@@ -104,8 +104,10 @@ pipeline_groups:
                   - router
                 dispatch_strategy: round_robin
             config:
-              listening_addr: "127.0.0.1:4327"
-              wait_for_result: true
+              protocols:
+                grpc:
+                  listening_addr: "127.0.0.1:4327"
+                  wait_for_result: true
           otlp_recv2:
             kind: receiver
             plugin_urn: "urn:otel:otlp:receiver"
@@ -115,8 +117,10 @@ pipeline_groups:
                   - router
                 dispatch_strategy: round_robin
             config:
-              listening_addr: "127.0.0.1:4337"
-              wait_for_result: true
+              protocols:
+                grpc:
+                  listening_addr: "127.0.0.1:4337"
+                  wait_for_result: true
 
           router:
             kind: processor
@@ -187,7 +191,9 @@ pipeline_groups:
                   - perf_noop
                 dispatch_strategy: round_robin
             config:
-              listening_addr: 127.0.0.1:4328
+              protocols:
+                grpc:
+                  listening_addr: 127.0.0.1:4328
 
           perf_noop:
             kind: exporter
--- a/rust/otap-dataflow/configs/otlp-both-otlp.yaml
+++ b/rust/otap-dataflow/configs/otlp-both-otlp.yaml
@@ -1,0 +1,100 @@
+# OTLP receiver (both gRPC and HTTP) -> OTLP exporter pipeline
+#
+# This configuration demonstrates both protocols enabled with shared concurrency.
+# For gRPC-only, see otlp-otlp.yaml
+# For HTTP-only, see otlp-http-otlp.yaml
+
+settings:
+  default_pipeline_ctrl_msg_channel_size: 100
+  default_node_ctrl_msg_channel_size: 100
+  default_pdata_channel_size: 100
+
+nodes:
+  receiver:
+    kind: receiver
+    plugin_urn: "urn:otel:otlp:receiver"
+    out_ports:
+      out_port:
+        destinations:
+          - exporter
+        dispatch_strategy: round_robin
+    config:
+      protocols:
+        # OTLP/gRPC receiver (default port: 4317)
+        grpc:
+          listening_addr: "127.0.0.1:4317"
+
+          # --- Request/Response behavior ---
+          # wait_for_result: false        # Wait for downstream ACK before responding
+          # timeout: "30s"                # Timeout for gRPC requests
+
+          # --- Concurrency limits ---
+          # max_concurrent_requests: 0    # 0 = auto (matches downstream capacity)
+          # max_concurrent_streams: null  # HTTP/2 streams per connection (null = auto)
+          # transport_concurrency_limit: null  # Per-connection limit (null = auto)
+          # load_shed: true               # Fast reject when overloaded
+
+          # --- Compression ---
+          # request_compression: [zstd, gzip, deflate]  # or "none" to disable
+          # response_compression: []      # Empty = no response compression
+
+          # --- TCP settings ---
+          # tcp_nodelay: true             # Disable Nagle's algorithm
+          # tcp_keepalive: "45s"          # TCP keepalive timeout (null to disable)
+          # tcp_keepalive_interval: "15s" # Probe interval
+          # tcp_keepalive_retries: 5      # Probes before connection dead
+
+          # --- HTTP/2 tuning ---
+          # initial_stream_window_size: "8MiB"
+          # initial_connection_window_size: "24MiB"
+          # http2_adaptive_window: false  # Dynamic window sizing
+          # max_frame_size: "16KiB"
+          # max_decoding_message_size: "4MiB"
+          # http2_keepalive_interval: "30s"
+          # http2_keepalive_timeout: "10s"
+
+          # --- TLS configuration ---
+          # tls:
+          #   cert_file: "/path/to/server.crt"
+          #   key_file: "/path/to/server.key"
+          #   client_ca_file: "/path/to/client-ca.crt"  # For mTLS
+          #   include_system_ca_certs_pool: false
+          #   handshake_timeout: "10s"
+          #   reload_interval: "5m"
+
+        # OTLP/HTTP receiver (default port: 4318)
+        http:
+          listening_addr: "127.0.0.1:4318"
+
+          # --- Request/Response behavior ---
+          # wait_for_result: false        # Wait for downstream ACK before responding
+          # timeout: "30s"                # Request processing timeout
+          # max_request_body_size: "4MiB" # Max body size (applies to compressed & decompressed)
+          # accept_compressed_requests: true  # Accept gzip/deflate/zstd bodies
+
+          # --- Concurrency limits ---
+          # max_concurrent_requests: 0    # 0 = auto (matches downstream capacity)
+
+          # --- TCP settings ---
+          # tcp_nodelay: true
+          # tcp_keepalive: "45s"
+          # tcp_keepalive_interval: "15s"
+          # tcp_keepalive_retries: 5
+
+          # --- TLS configuration ---
+          # tls:
+          #   cert_file: "/path/to/server.crt"
+          #   key_file: "/path/to/server.key"
+          #   client_ca_file: "/path/to/client-ca.crt"  # For mTLS
+          #   include_system_ca_certs_pool: false
+          #   handshake_timeout: "10s"
+          #   reload_interval: "5m"
+
+  exporter:
+    kind: exporter
+    plugin_urn: "urn:otel:otlp:exporter"
+    config:
+      # Downstream OTLP/gRPC collector endpoint.
+      # Uses a different port than this receiver to avoid accidental loops.
+      grpc_endpoint: "http://127.0.0.1:4319"
+      # timeout: "15s"
--- a/rust/otap-dataflow/configs/otlp-http-otlp.yaml
+++ b/rust/otap-dataflow/configs/otlp-http-otlp.yaml
@@ -1,3 +1,9 @@
+# OTLP/HTTP receiver -> OTLP exporter pipeline (HTTP-only mode)
+#
+# This configuration demonstrates HTTP-only mode for the OTLP receiver.
+# For gRPC-only, see otlp-otlp.yaml
+# For both protocols enabled, see otlp-both-otlp.yaml
+
 settings:
   default_pipeline_ctrl_msg_channel_size: 100
   default_node_ctrl_msg_channel_size: 100
@@ -13,27 +19,28 @@ nodes:
           - exporter
         dispatch_strategy: round_robin
     config:
-      # OTLP/gRPC receiver (default port: 4317)
-      listening_addr: "127.0.0.1:4317"
+      protocols:
+        # OTLP/HTTP receiver only (default port: 4318)
+        http:
+          listening_addr: "127.0.0.1:4318"
 
-      # OTLP/HTTP receiver (default port: 4318)
-      http:
-        listening_addr: "127.0.0.1:4318"
-        # wait_for_result: false
-        # max_concurrent_requests: 0
-        # max_request_body_size: "4MiB"
-        # accept_compressed_requests: true
+          # --- Request/Response behavior ---
+          # wait_for_result: false        # Wait for downstream ACK before responding
+          # timeout: "30s"                # Request processing timeout
+          # max_request_body_size: "4MiB" # Max body size (applies to compressed & decompressed)
+          # accept_compressed_requests: true  # Accept gzip/deflate/zstd bodies
 
-      # timeout: "30s"  # Optional: timeout for gRPC requests
+          # --- Concurrency limits ---
+          # max_concurrent_requests: 0    # 0 = auto (matches downstream capacity)
 
-      # TLS / mTLS (receiver-side; gRPC only today)
-      # tls:
-      #   cert_file: "/path/to/server.crt"   # or `cert_pem: | ...`
-      #   key_file: "/path/to/server.key"    # or `key_pem: | ...`
-      #   client_ca_file: "/path/to/client-ca.crt"  # or `client_ca_pem: | ...`
-      #   include_system_ca_certs_pool: false
-      #   handshake_timeout: "10s"
-      #   reload_interval: "5m"
+          # --- TLS configuration ---
+          # tls:
+          #   cert_file: "/path/to/server.crt"
+          #   key_file: "/path/to/server.key"
+          #   client_ca_file: "/path/to/client-ca.crt"  # For mTLS
+          #   include_system_ca_certs_pool: false
+          #   handshake_timeout: "10s"
+          #   reload_interval: "5m"
 
   exporter:
     kind: exporter
--- a/rust/otap-dataflow/configs/otlp-otap.yaml
+++ b/rust/otap-dataflow/configs/otlp-otap.yaml
@@ -13,8 +13,10 @@ nodes:
           - exporter
         dispatch_strategy: round_robin
     config:
-      listening_addr: "127.0.0.1:4317"
-      # timeout: "30s"  # Optional: timeout for RPC requests
+      protocols:
+        grpc:
+          listening_addr: "127.0.0.1:4317"
+          # timeout: "30s"  # Optional: timeout for RPC requests
 
       # TLS / mTLS (receiver-side)
       # tls:
@@ -33,14 +35,16 @@ nodes:
     kind: exporter
     plugin_urn: "urn:otel:otap:exporter"
     config:
-      grpc_endpoint: "http://127.0.0.1:4318"
+      # Downstream OTAP/gRPC collector endpoint.
+      # Uses port 4319 to avoid conflict with standard OTLP ports (4317=gRPC, 4318=HTTP)
+      grpc_endpoint: "http://127.0.0.1:4319"
       # timeout: "15s"  # Optional: timeout for RPC requests
 
       # TLS / mTLS (exporter-side)
       # - TLS is enabled when `grpc_endpoint` uses https:// (scheme default).
       # - Setting `tls` enables TLS/mTLS and provides additional configuration options.
       # - If `tls` is omitted and `grpc_endpoint` uses http://, the connection is plaintext.
-      # grpc_endpoint: "https://collector.example.com:4318"
+      # grpc_endpoint: "https://collector.example.com:4319"
       # tls:
       #   ca_file: "/path/to/ca.crt"          # or `ca_pem: | ...`
       #   include_system_ca_certs_pool: false # default is true
--- a/rust/otap-dataflow/configs/otlp-otlp.yaml
+++ b/rust/otap-dataflow/configs/otlp-otlp.yaml
@@ -1,3 +1,9 @@
+# OTLP/gRPC receiver -> OTLP exporter pipeline (gRPC-only mode)
+#
+# This configuration demonstrates gRPC-only mode for the OTLP receiver.
+# For HTTP-only, see otlp-http-otlp.yaml
+# For both protocols enabled, see otlp-both-otlp.yaml
+
 settings:
   default_pipeline_ctrl_msg_channel_size: 100
   default_node_ctrl_msg_channel_size: 100
@@ -13,27 +19,38 @@ nodes:
           - exporter
         dispatch_strategy: round_robin
     config:
-      listening_addr: "127.0.0.1:4317"
-      # timeout: "30s"  # Optional: timeout for RPC requests
+      protocols:
+        # OTLP/gRPC receiver only (default port: 4317)
+        grpc:
+          listening_addr: "127.0.0.1:4317"
 
-      # TLS / mTLS (receiver-side)
-      # tls:
-      #   # Server identity (required for TLS)
-      #   cert_file: "/path/to/server.crt"   # or `cert_pem: | ...`
-      #   key_file: "/path/to/server.key"    # or `key_pem: | ...`
-      #
-      #   # Client certificate verification for mTLS (optional)
-      #   client_ca_file: "/path/to/client-ca.crt"  # or `client_ca_pem: | ...`
-      #   include_system_ca_certs_pool: false       # default is false
-      #
-      #   # Advanced settings
-      #   handshake_timeout: "10s"           # default is 10s
-      #   reload_interval: "5m"              # default is 5m
+          # --- Request/Response behavior ---
+          # wait_for_result: false        # Wait for downstream ACK before responding
+          # timeout: "30s"                # Timeout for gRPC requests
+
+          # --- Concurrency limits ---
+          # max_concurrent_requests: 0    # 0 = auto (matches downstream capacity)
+
+          # --- Compression ---
+          # request_compression: [zstd, gzip, deflate]  # or "none" to disable
+          # response_compression: []      # Empty = no response compression
+
+          # --- TLS configuration ---
+          # tls:
+          #   cert_file: "/path/to/server.crt"
+          #   key_file: "/path/to/server.key"
+          #   client_ca_file: "/path/to/client-ca.crt"  # For mTLS
+          #   include_system_ca_certs_pool: false
+          #   handshake_timeout: "10s"
+          #   reload_interval: "5m"
+
   exporter:
     kind: exporter
     plugin_urn: "urn:otel:otlp:exporter"
     config:
-      grpc_endpoint: "http://127.0.0.1:4318"
+      # Downstream OTLP/gRPC collector endpoint.
+      # Uses a different port than this receiver to avoid accidental loops.
+      grpc_endpoint: "http://127.0.0.1:4319"
       # timeout: "15s"  # Optional: timeout for RPC requests
 
       # TLS / mTLS (exporter-side)
--- a/rust/otap-dataflow/configs/otlp-perf.yaml
+++ b/rust/otap-dataflow/configs/otlp-perf.yaml
@@ -13,7 +13,9 @@ nodes:
           - perf
         dispatch_strategy: round_robin
     config:
-      listening_addr: "127.0.0.1:4318"
+      protocols:
+        grpc:
+          listening_addr: "127.0.0.1:4318"
   perf:
     kind: exporter
     plugin_urn: "urn:otel:otap:perf:exporter"
--- a/rust/otap-dataflow/configs/test-mtls.yaml
+++ b/rust/otap-dataflow/configs/test-mtls.yaml
@@ -21,24 +21,26 @@ nodes:
           - exporter
         dispatch_strategy: round_robin
     config:
-      listening_addr: "127.0.0.1:4317"
+      protocols:
+        grpc:
+          listening_addr: "127.0.0.1:4317"
+          # Receiver mTLS configuration (server-side)
+          tls:
+            # Server identity (required for TLS)
+            cert_file: "./test-certs/server.crt"
+            key_file: "./test-certs/server.key"
 
-      # Receiver mTLS configuration (server-side)
-      tls:
-        # Server identity (required for TLS)
-        cert_file: "./test-certs/server.crt"
-        key_file: "./test-certs/server.key"
-
-        # Client certificate verification for mTLS
-        client_ca_file: "./test-certs/ca.crt"
-        include_system_ca_certs_pool: false
+            # Client certificate verification for mTLS
+            client_ca_file: "./test-certs/ca.crt"
+            include_system_ca_certs_pool: false
 
   exporter:
     kind: exporter
     plugin_urn: "urn:otel:otlp:exporter"
     config:
       # Send to another OTLP receiver with mTLS enabled
-      grpc_endpoint: "https://127.0.0.1:4318"
+      # Uses port 4319 to avoid conflict with standard OTLP ports (4317=gRPC, 4318=HTTP)
+      grpc_endpoint: "https://127.0.0.1:4319"
 
       # Exporter mTLS configuration (client-side)
       tls:
--- a/rust/otap-dataflow/configs/test-tls-only.yaml
+++ b/rust/otap-dataflow/configs/test-tls-only.yaml
@@ -13,10 +13,12 @@
             - exporter
           dispatch_strategy: round_robin
       config:
-        listening_addr: "127.0.0.1:4317"
-        tls:
-          cert_file: "./test-certs/server.crt"
-          key_file: "./test-certs/server.key"
+        protocols:
+          grpc:
+            listening_addr: "127.0.0.1:4317"
+            tls:
+              cert_file: "./test-certs/server.crt"
+              key_file: "./test-certs/server.key"
 
     exporter:
       kind: exporter
--- a/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/otlp-ame.yaml
+++ b/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/otlp-ame.yaml
@@ -24,8 +24,10 @@ nodes:
           - azure-monitor-exporter
         dispatch_strategy: round_robin
     config:
-      # Listen on all interfaces in container
-      listening_addr: "0.0.0.0:4317"
+      protocols:
+        grpc:
+          # Listen on all interfaces in container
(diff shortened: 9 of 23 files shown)


lalitb:
Fanout now runs `stateless` in one case (`await_ack: none` fire-and-forget) and `slim state` in one case (parallel + primary-only with no fallback/timeout: just `request_id -> original_pdata`). All other configuration remains fully stateful to preserve ordering, failover, and upstream routing semantics. Have updated the PR desc and `Readme.md` accordingly. Ready for review with that explicit split.

lquerel:
Regarding the stateless vs stateful debate, I'm wondering if it's actually more a question of stack vs stackless, since we're based on message passing. I'm not a Go Collector specialist, but from what I understand the state is kept on the stack across the different sequential calls. So in the end, the presence and management of state don't seem that different to me.

In our case, having support for both a parallel mode and a sequential mode should allow users to choose between speed with limited control, or a precisely defined sequence that can be interrupted at any step. That would make this fanout processor something truly powerful and expressive.

lalitb:
> Regarding the stateless vs stateful debate, I'm wondering if it's actually more a question of stack vs stackless, since we're based on message passing. I'm not a Go Collector specialist, but from what I understand the state is kept on the stack across the different sequential calls. So in the end, the presence and management of state don't seem that different to me.

You're right - state always exists somewhere. In Go fanout it's implicit on the call stack during the synchronous call, then gone when it returns; in our async pipeline we make it explicit via the message's context stack or a map when coordinating multiple outcomes. The key difference is `blocking` vs `non-blocking` - Go's blocking model gives implicit correlation, while our async model requires explicit correlation since acks arrive later as separate messages. 

> In our case, having support for both a parallel mode and a sequential mode should allow users to choose between speed with limited control, or a precisely defined sequence that can be interrupted at any step. That would make this fanout processor something truly powerful and expressive.

Yes, parallel vs sequential gives users that flexibility - speed when order doesn't matter, precise control when it does. Currently Go collector fanout is only sequential.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
```diff
@@ -0,0 +1,256 @@
...
+    await_ack: primary      # or "all" or "none"
+    timeout_check_interval: 200ms
+    destinations:
+      - port: primary_export
+        primary: true
+        timeout: 5s
+      - port: backup_export
+        fallback_for: primary_export
+        timeout: 10s
+      - port: analytics_export
```
lquerel, line 26: I think that for a first PR this is fine, but to improve the user experience we should, in my opinion, find a way to combine the output mechanism (see [here](https://github.com/open-telemetry/otel-arrow/issues/1829)) with this destination system, because there is a lot of redundancy. That could be something like:


```yaml
processor:
  urn: urn:otel:fanout:processor
  config:
      mode: parallel
      await_ack: primary
      timeout_check_interval: 200ms
  outputs:
    - port: primary_export
      primary: true
      timeout: 5s
   - port: backup_export
      fallback_for: primary_export
      timeout: 10s
   - port: analytics_export
```

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Configuration
+
+```yaml
+processor:
+  urn: urn:otel:fanout:processor
+  config:
+    mode: parallel          # or "sequential"
+    await_ack: primary      # or "all" or "none"
+    timeout_check_interval: 200ms
```
lquerel, line 17: We probably need a `max_inflight` field to specify the maximum number of in-flight messages processed by the fanout processor, in order to keep its internal state properly bounded.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
```diff
@@ -0,0 +1,256 @@
...
+    destinations:
+      - port: primary_export
+        primary: true
+        timeout: 5s
```
lquerel, line 22: We might also need some kind of overflow policy such as block, drop, error (nack), ...

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
```diff
@@ -0,0 +1,256 @@
...
+- Timeouts are checked on periodic `TimerTick` (default interval: 200ms)
+- Timeout is treated as a failure (same as nack)
+
+## Metrics
+
+| Metric | Description |
+|--------|-------------|
+| `sent` | Requests dispatched (per incoming PData) |
+| `acked` | Requests successfully acked upstream |
+| `nacked` | Requests nacked upstream |
```
lquerel: There metrics will be redondant with the existing pdata channel metrics.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
```diff
@@ -0,0 +1,256 @@
...
+## Edge Cases
+
+- **Duplicate ports**: Rejected at config validation
+- **Multiple primaries**: Rejected; exactly one primary allowed
+- **Primary as fallback**: Rejected; primary cannot have `fallback_for`
+- **Unknown fallback target**: Rejected; `fallback_for` must reference port
+- **Fallback cycles**: Detected and rejected at config validation
+- **Fallback with `await_ack: none`**: Rejected; fire-and-forget ignores fallbacks
+- **Timeout with `await_ack: none`**: Rejected; fire-and-forget doesn't track responses
+- **Shutdown**: Inflight requests are dropped (not proactively nacked)
```
lquerel: Not sure to understand this one.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
```diff
@@ -0,0 +1,256 @@
...
+## Performance Optimizations
+
+The processor uses optimized fast paths for common configurations to minimize
+memory allocations per request:
+
+| Configuration | Fast Path | State Per Request |
+|---------------|-----------|-------------------|
+| `await_ack: none` | Fire-and-forget | **None** (zero tracking) |
+| `parallel` + `primary` (no fallback/timeout) | Slim primary | Minimal map |
+| All other configs | Full | Complete endpoint tracking |
```
lquerel, line 332: This need to be bounded, see my initial comment on `max_inflight` (or a better name).

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
```diff
@@ -0,0 +1,256 @@
+# Fan-out Processor
```
lquerel, line 1: We need a section describing how backpressure will work with this fanout node.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2268 @@
...
+    create: |pipeline_ctx: PipelineContext,
+             node: NodeId,
+             node_config: Arc<NodeUserConfig>,
+             proc_cfg: &ProcessorConfig| {
+        create_fanout_processor(pipeline_ctx, node, node_config, proc_cfg)
+    },
+};
+
+#[cfg(test)]
+mod tests {
```
lquerel, line 1173: A few focused tests here would really help lock down tricky behavior and prevent regressions:
- Late responses after failover
   Sequence: a destination times out -> fallback is dispatched -> a late Ack/Nack arrives from the original destination.
   Expected: the late response from the original destination is ignored and must not override the fallback outcome.
- Sequential mode should not dispatch fallbacks on success
   Sequence: sequential mode with a fallback configured for destination A -> A acks successfully.
   Expected: A's fallback is not dispatched; only the next normal destination runs.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2268 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum DestStatus {
+    PendingSend,
+    InFlight,
+    Acked,
+    Nacked,
+}
+
+#[derive(Debug)]
+struct EndpointState {
```
lquerel: For consistency with your `DestStatus` why not using `DestinationState` ?

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2268 @@
...
+    mode: DeliveryMode,
+    primary: usize,
+    /// Inline storage for up to 4 endpoints to avoid heap allocation in common cases.
+    endpoints: EndpointVec,
+    completed_origins: HashSet<usize>,
+    required_origins: usize,
+    /// Original pdata (before fanout's subscription was added) for upstream ack/nack.
+    /// This ensures upstream routing uses the correct context stack.
+    original_pdata: OtapPdata,
+    next_send_queue: SmallVec<[usize; 4]>,
```
lquerel: Will be nice to add comments for completed_origins, required_origins, and next_send_queue to ease the understanding.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2268 @@
...
+        effect_handler: &EffectHandler<OtapPdata>,
+    ) -> Result<u64, Error> {
+        self.ensure_timer(effect_handler).await?;
+
+        let request_id = self.next_id;
+        self.next_id = self.next_id.wrapping_add(1).max(1);
+
+        let mut endpoints = EndpointVec::new();
+        let mut queue = SmallVec::<[usize; 4]>::new();
+        let interests = Interests::ACKS_OR_NACKS | Interests::RETURN_DATA;
```
lquerel: Do you need `Interests::RETURN_DATA` for all combinations of mode and await ack?

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2268 @@
...
+            let _ = inflight.completed_origins.insert(origin);
+            if inflight.completed_origins.len() >= inflight.required_origins {
+                let entry = self.inflight.remove(&request_id)?;
+                return Some(entry.original_pdata);
+            }
+        }
+        None
+    }
+
+    fn handle_failure(
```
lquerel, line 605: `handle_failure` does a linear scan to find a fallback. You could probably precompute a dest_index -> immediate_fallback_index map (or reject ambiguity) to avoid repeated scans.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2268 @@
...
+type EndpointVec = SmallVec<[EndpointState; 4]>;
+
+#[derive(Debug)]
+struct Inflight {
+    await_ack: AwaitAck,
+    mode: DeliveryMode,
+    primary: usize,
+    /// Inline storage for up to 4 endpoints to avoid heap allocation in common cases.
+    endpoints: EndpointVec,
+    completed_origins: HashSet<usize>,
```
lquerel: May be a bitset could be good enough here and smaller :-)

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2268 @@
...
+
+        // No fallback, produce a nack using original pdata for correct upstream routing.
+        Some(NackMsg {
+            reason,
+            calldata: smallvec![],
+            refused: Box::new(inflight.original_pdata.clone()),
+        })
+    }
+
+    async fn handle_timeout(
```
lquerel, line 652: `handle_timeout` scans every inflight request on each tick... A min-heap of deadlines (or tracking the next deadline) could certainly drastically cut CPU when inflight is large which will matter in our case I believe :-)

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2268 @@
...
+        let mut endpoints = EndpointVec::new();
+        let mut queue = SmallVec::<[usize; 4]>::new();
+        let interests = Interests::ACKS_OR_NACKS | Interests::RETURN_DATA;
+        // Deadlines are initialized here and reset at actual dispatch time; a future tightening
+        // could defer setting timeout_at until send if needed.
+        let now = now();
+
+        for (idx, dest) in self.config.destinations.iter().enumerate() {
+            // Create a clone for this destination and subscribe with fanout calldata.
+            let mut dest_data = pdata.clone();
```
lquerel, line 501: You currently clone/retain payloads for all destinations up front. That might a bit difficult to achieve but an optimization could be to defer clone+subscribe until your call to dispatch_ready.

jmacd:
I think we can overcome the blockers and avoid memory allocation for the "slim primary" case, meaning really this case ought to work and we can modify OtapPdata and/or Context to help. Here's how I think this will work:

When the request arrives, use into_parts() to split into context/data; clone the data once for each output. For the primary request, use OtapPdata::new(original_context, data), and for the other requests (for now) we can use Context::new() creating an empty context. Send all the requests in parallel.

You don't need any calls to subscribe_to() in this arrangement, since the primary keeps the context it had any Ack/Nack it receives will go straight through the fanout processor directly to the original recipient. If we have other useful information in the Context (later), we will want a way to erase the subscription information and keep everything else (I think).

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
lalitb: Good point. I have added `max_inflight` config field (default 10,000, 0 = unlimited). New messages are nacked when the limit is exceeded. Also updated the document to reflect that.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
lalitb: Good point. For tracked modes (`primary/all`), we currently have a fixed policy: `nack` when `max_inflight` is reached. This signals backpressure upstream, allowing the pipeline to handle it appropriately (retry, buffer, or propagate further).

For fire-and-forget (`await_ack: none`), there's no inflight tracking, so it naturally relies on channel backpressure (effectively block).

I've documented this in the `README` under "Backpressure and Overflow". If users need configurable overflow policies in the future, we can add an overflow_policy field. For now, the fixed nack behavior keeps the config surface small and avoids footgun options like silent drops. I will create a tracking issue for this.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
lalitb: These metrics answer different questions:

- `Channel metrics`: Per-destination sends/receives/errors (one count per destination)
- `Fanout metrics`: Aggregated request outcomes after await_ack/fallback logic

Specifically:

`sent`/`acked`/`nacked` - request-level outcomes, not per-destination
`timed_out` - destination-level (how many destinations hit their timeout)
`rejected_max_inflight` - fanout-only, channels don't track this

For example, with `await_ack: primary` + `fallback`: if primary fails but fallback succeeds, fanout reports 1 ack. Channel metrics would show 1 nack + 1 ack across destinations.

Happy to remove these if we see metrics bloat - some could potentially be aggregated at channel level to get the equivalent.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
lalitb: Agreed the original text was terse. Updated to:

``
Shutdown: Inflight state is dropped; no nacks sent to upstream. Upstream will not receive notification for in-progress requests.
``

This clarifies that on shutdown, we don't proactively nack pending requests - upstream simply won't hear back. Downstream may still process them successfully; we just won't see the ack.

Alternative would be to nack all inflight requests on shutdown, so upstream knows they weren't confirmed. But:

- Downstream may still process them successfully (we just won't see the ack)
- Proactive nacking could cause duplicate processing if upstream retries
- Shutdown is typically a terminal event anyway

If you think we should `nack` inflight on shutdown (fail-safe), I can add that in next PR, and add tracking issue for now.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
lalitb: Done - `max_inflight` (default: 10,000) bounds the inflight maps for tracked modes (primary/all). When the limit is reached, new requests are nacked immediately.

This is documented in the "Backpressure and Overflow" section and enforced in both the slim and full inflight paths.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
lalitb: Good point. Added -  see the "Backpressure" section. This covers nack propagation, max_inflight limit, and bounded channels.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Good suggestions. Added both tests.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Good point. Have done accordingly.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Done. Added doc comments for all fields.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Good catch - we don't need `RETURN_DATA` at all. Subscriptions now use only `ACKS_OR_NACKS` and we rely on `calldata` plus stored `original_pdata`. Removed it.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Done. Added `fallback_for_dest: Vec<Option<usize>>` to `ValidatedConfig`, precomputed during validation. Now `handle_failure` does O(1) lookup:

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Good point -  Replaced `HashSet<usize>` with `u64` bitset. Also added validation to reject configs with >64 destinations (since u64 can only track 64 bits). Typical configs have 2-4 destinations, so this is a non-issue in practice.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Good point. Added a min-heap for O(log n) timeout checking - on timer tick, we pop expired entries instead of scanning all inflight. This is O(k log n) where k = expired deadlines, instead of O(n) for all inflight.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Good optimization idea. Currently we clone upfront for all destinations, but for sequential mode or when fallbacks aren't triggered, some clones are wasted. This adds some complexity - need to keep original pdata longer, handle clone failures at dispatch time. As of now, I have added a TODO comment in code, and will do a follow up PR for this - 
```
// TODO(optimization): Defer clone+subscribe to dispatch_ready to avoid
// upfront clones for destinations that may never be dispatched (sequential
// mode, untriggered fallbacks). Tracked for follow-up PR.
```

lalitb:
> I think we can overcome the blockers and avoid memory allocation for the "slim primary" case, meaning really this case ought to work and we can modify OtapPdata and/or Context to help. Here's how I think this will work:
> 
> When the request arrives, use into_parts() to split into context/data; clone the data once for each output. For the primary request, use OtapPdata::new(original_context, data), and for the other requests (for now) we can use Context::new() creating an empty context. Send all the requests in parallel.
> 
> You don't need any calls to subscribe_to() in this arrangement, since the primary keeps the context it had any Ack/Nack it receives will go straight through the fanout processor directly to the original recipient. If we have other useful information in the Context (later), we will want a way to erase the subscription information and keep everything else (I think).

Interesting idea - keeping the original context on primary so acks route directly upstream would eliminate the subscription overhead entirely.

Right now the slim path still needs the subscription so fanout can see the primary's ack/nack - it clears `slim_inflight`, bumps metrics, and enforces `max_inflight`. Letting acks bypass fanout would leave entries dangling and break backpressure. Non-primary acks would also be unrouted/dropped with empty contexts, which is fine only if the engine treats them as benign.

I'm open to a zero-alloc slim path, but it would require engine support for an "observer" mode where the fanout can see the primary outcome without adding a subscription frame, or a way to strip/restore subscription state in Context while still delivering the control message back to the fanout.

Given the tiny context clone cost today, I kept the existing slim path. Happy to revisit with a concrete engine change or a safe "passthrough observer" design.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
lalitb: created #1906 to track this.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: created #1905 to track this.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
lalitb: Agreed - there's redundancy between destinations and the standard output mechanism. As you've already captured in #1829, the redesign should allow nodes to extend output descriptions, which would address this.

For now, keeping destinations in processor config avoids engine changes while the config-model-v1 proposals are finalized. Added a note in the README acknowledging this as a future improvement.

lalitb:
> I think this is a very good first PR on this new processor, which is definitely complex (and will likely become even more so in the future). I have a few comments that I think should improve a couple of things here and there.

 Thanks for the thorough review! Addressed the comments in this PR and created tracking issues for follow-up improvements.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
```diff
@@ -0,0 +1,356 @@
...
+
+The fan-out processor clones incoming PData to multiple downstream destinations
+with configurable delivery modes, ack policies, and fallback routing.
+
```
jmacd, line 5:
```suggestion
The fan-out processor clones incoming PData to multiple downstream destinations
with configurable delivery modes, ack policies, and fallback routing.

Fallback settings on a per-port basis supports configuring alternative
destinations after failures, supporting a broad range of failover policies.

```

This is to help the reader below around lines 55 where "fallback" starts being used prior to the full explanation on line 110.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
```diff
@@ -0,0 +1,356 @@
...
+## Configuration
+
+```yaml
+processor:
+  urn: urn:otel:fanout:processor
+  config:
+    mode: parallel          # or "sequential"
+    await_ack: primary      # or "all" or "none"
```
jmacd:
```suggestion
    await_ack: all          # or "primary" or "none"
```

This is so that the `max_inflight` setting applies. If we have this as primary, delete line 15.

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
```diff
@@ -0,0 +1,356 @@
...
+provides the explicit bound on concurrent requests:
+
+| `await_ack` | Overflow Mechanism | Behavior |
+|-------------|-------------------|----------|
+| `primary`/`all` | `max_inflight` | Nack when limit reached |
+| `none` | Channel capacity | Block until downstream drains |
+
+> **Note**: `max_inflight` only applies to tracked modes.
+> Fire-and-forget (`await_ack: none`) has no inflight tracking and relies
+> solely on channel backpressure.
```
jmacd, line 219: :+1:

rust/otap-dataflow/crates/otap/src/fanout_processor/README.md
```diff
@@ -0,0 +1,356 @@
...
+Each destination receives an independent **clone** of the incoming PData.
+
+### Clone Cost
+
+Cloning is cheap for both data formats:
+
+| Format | Storage | Clone Cost |
+|--------|---------|------------|
+| OTLP (OtlpProtoBytes) | `bytes::Bytes` | O(1) refcount bump |
+| OTAP (OtapArrowRecords) | `Arc` per column | O(columns) refcount bumps |
```
jmacd:
```suggestion
| OTLP (OtlpProtoBytes) | `bytes::Bytes` | O(1) refcount increment |
| OTAP (OtapArrowRecords) | `Arc` per column | O(columns) refcount increments |
```

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2982 @@
...
+    next_send_queue: SmallVec<[usize; 4]>,
+}
+
+#[metric_set(name = "fanout.processor.metrics")]
+#[derive(Debug, Default, Clone)]
+struct FanoutMetrics {
+    /// Requests dispatched. Note: This is a convenience metric that overlaps with
+    /// channel-level send metrics. Consider removing if metric bloat is a concern.
+    #[metric(unit = "{item}")]
+    pub sent: Counter<u64>,
```
jmacd, line 378: This is OK for now!
I believe the metrics specified in https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md which may at first seem quite complex are complex because they completely cover what this component is doing. If channel-level metrics are properly tagged we'll have counters for sent by endpoint as well as outcomes by endpoint (producer side), and independently we'll have the acked/nacked-returned counts below (consumer side).

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2982 @@
...
+    request_id: u64,
+    dest_index: usize,
+}
+
+/// Fan-out processor implementation.
+pub struct FanoutProcessor {
+    config: ValidatedConfig,
+    metrics: MetricSet<FanoutMetrics>,
+    /// Full inflight tracking for complex scenarios (sequential, await_all, fallback, timeout).
+    inflight: HashMap<u64, Inflight>,
```
jmacd, line 406: I wonder if you considered a single hash map with an enum value like

```
enum InflightRecord {
  Slim(OtapPdata),
  Full(InflightStruct),
}
```

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2982 @@
...
+        pdata: OtapPdata,
+        effect_handler: &EffectHandler<OtapPdata>,
+    ) -> Result<u64, Error> {
+        self.ensure_timer(effect_handler).await?;
+
+        let request_id = self.next_id;
+        self.next_id = self.next_id.wrapping_add(1).max(1);
+
+        let mut destinations_state = DestinationVec::new();
+        let mut queue = SmallVec::<[usize; 4]>::new();
```
jmacd: I would like to see a `pub type DestinationQueue = SmallVec::<[usize; 4]>;` in the same location DestinationVec is allocated. This lets us write DestinationQueue::new() instead of smallvec![] for example.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2982 @@
...
+    }
+
+    /// Dispatch ready destinations and return any new deadlines for the heap.
+    async fn dispatch_ready(
+        request_id: u64,
+        inflight: &mut Inflight,
+        destinations: &[DestinationConfig],
+        effect_handler: &EffectHandler<OtapPdata>,
+    ) -> Result<SmallVec<[Deadline; 4]>, TypedError<OtapPdata>> {
+        let mut to_send = SmallVec::<[usize; 4]>::new();
```
jmacd: Repeating myself: I would like to see `pub type ... = SmallVec<...>;` used throughout, so that all the `4` value choices appear in once place.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2982 @@
...
+                    inflight.next_send_queue.push(fb_idx);
+                }
+                return None;
+            }
+        }
+
+        // No fallback, produce a nack using original pdata for correct upstream routing.
+        Some(NackMsg {
+            reason,
+            calldata: smallvec![],
```
jmacd: Here I also would prefer not to see `smallvec![]` but to see `CallData::new()`.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2982 @@
...
+        })
+    }
+
+    async fn handle_timeout(
+        &mut self,
+        effect_handler: &EffectHandler<OtapPdata>,
+    ) -> Result<Vec<NackMsg<OtapPdata>>, Error> {
+        let now = now();
+        let mut expired = Vec::new();
+        let mut dispatch_requests = SmallVec::<[u64; 8]>::new();
```
jmacd: (... again) let's have one block of type aliases where these constants are co-located.

jmacd (changes requested):
This is great. Requesting a bit of clarification in the README and one more thing.

nit: Asking for a small change to lift all SmallVec<[...; SIZE]> into a common area where all the SmallVec types used in this code are given an alias. Then instead of smallvec![] use AliasType::new().

lquerel:
@lalitb @jmacd Do you think we are far from being able to merge this PR?

lalitb:
> @lalitb @jmacd Do you think we are far from being able to merge this PR?

I don't think we're far at all. Let me go through Josh's comments - I should be able to address them by end of day.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -0,0 +1,2982 @@
...
+                }
+            }
+            (origin, inflight.await_ack, inflight.primary, inflight.mode)
+        };
+
+        if matches!(await_ack, AwaitAck::None) {
+            return Ok(());
+        }
+
+        // Await primary: if this ack corresponds to primary origin (or its fallback) we can finish.
```
ThomsonTan, line 792: This could cause data-loss for non-primary destinations, is this expected?

lalitb: This doesn’t cause data loss for non-primary destinations. The data is still sent; the difference is that we don’t wait for ACK/NACK from non-primary destinations. That behavior is intentional and matches the design of `await_ack` as documented in the README.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Thought about this - decided to keep them separate for clarity and to avoid enum matching overhead in the common case.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Thanks. done.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: done

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: done.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: done.

lalitb:
@lquerel, @jmacd - Thanks for the thorough review! All comments have been addressed, responded to, or tracked for
   follow-up

lquerel (approved):
Thanks. LGTM

jmacd (approved):

PR #1879, opened 2026-01-23
Validation process, updated pipeline support

c1ly:
Added pipeline validation to the validation tests which uses a pipeline group to validate various pipelines. 

- Pipeline group will have 3 pipelines the first one to generate messages, second would be the pipeline being validated, and the third pipeline to validate the messages.

Added a engine config template

- template will change (swap between otlp/otap) the exporters and receivers of the starting and ending pipelines in the pipeline group

### DIFF
--- a/rust/otap-dataflow/crates/validation/src/metrics_types.rs
+++ b/rust/otap-dataflow/crates/validation/src/metrics_types.rs
@@ -87,7 +87,7 @@ mod tests {
         );
 
         assert_eq!(format_metric_value(&MetricValue::U64(42)), "42");
-        assert_eq!(format_metric_value(&MetricValue::F64(3.14)), "3.14");
+        assert_eq!(format_metric_value(&MetricValue::F64(4.1)), "4.1");
     }
 
     #[test]
--- a/rust/otap-dataflow/crates/validation/src/pipeline.rs
+++ b/rust/otap-dataflow/crates/validation/src/pipeline.rs
@@ -25,6 +25,7 @@ impl Pipeline {
     }
 
     /// Load a pipeline from a YAML string slice.
+    #[must_use]
     pub fn from_yaml(yaml: &str) -> Self {
         let suv_yaml: Value = serde_yaml::from_str(yaml).expect("invalid pipeline yaml");
         Self {
@@ -269,36 +270,36 @@ nodes:
         let doc: Value = serde_yaml::from_str(&rendered).unwrap();
         let nodes = doc.get("nodes").and_then(Value::as_mapping).unwrap();
         let recv = nodes
-            .get(&Value::from("receiver"))
+            .get(Value::from("receiver"))
             .and_then(Value::as_mapping)
             .unwrap();
         let recv_cfg = recv
-            .get(&Value::from("config"))
+            .get(Value::from("config"))
             .and_then(Value::as_mapping)
             .unwrap();
         let protocols = recv_cfg
-            .get(&Value::from("protocols"))
+            .get(Value::from("protocols"))
             .and_then(Value::as_mapping)
             .unwrap();
         let grpc = protocols
-            .get(&Value::from("grpc"))
+            .get(Value::from("grpc"))
             .and_then(Value::as_mapping)
             .unwrap();
         assert_eq!(
-            grpc.get(&Value::from("listening_addr")),
+            grpc.get(Value::from("listening_addr")),
             Some(&Value::from("127.0.0.1:5555"))
         );
 
         let exp = nodes
-            .get(&Value::from("exporter"))
+            .get(Value::from("exporter"))
             .and_then(Value::as_mapping)
             .unwrap();
         let exp_cfg = exp
-            .get(&Value::from("config"))
+            .get(Value::from("config"))
             .and_then(Value::as_mapping)
             .unwrap();
         assert_eq!(
-            exp_cfg.get(&Value::from("grpc_endpoint")),
+            exp_cfg.get(Value::from("grpc_endpoint")),
             Some(&Value::from("http://127.0.0.1:7777"))
         );
     }
@@ -318,28 +319,28 @@ nodes:
         let doc: Value = serde_yaml::from_str(&rendered).unwrap();
         let nodes = doc.get("nodes").and_then(Value::as_mapping).unwrap();
         let recv = nodes
-            .get(&Value::from("otap_recv"))
+            .get(Value::from("otap_recv"))
             .and_then(Value::as_mapping)
             .unwrap();
         let recv_cfg = recv
-            .get(&Value::from("config"))
+            .get(Value::from("config"))
             .and_then(Value::as_mapping)
             .unwrap();
         assert_eq!(
-            recv_cfg.get(&Value::from("listening_addr")),
+            recv_cfg.get(Value::from("listening_addr")),
             Some(&Value::from("127.0.0.1:6000"))
         );
 
         let exp = nodes
-            .get(&Value::from("otap_exp"))
+            .get(Value::from("otap_exp"))
             .and_then(Value::as_mapping)
             .unwrap();
         let exp_cfg = exp
-            .get(&Value::from("config"))
+            .get(Value::from("config"))
             .and_then(Value::as_mapping)
             .unwrap();
         assert_eq!(
-            exp_cfg.get(&Value::from("grpc_endpoint")),
+            exp_cfg.get(Value::from("grpc_endpoint")),
             Some(&Value::from("http://127.0.0.1:7000"))
         );
     }
--- a/rust/otap-dataflow/crates/validation/src/scenario.rs
+++ b/rust/otap-dataflow/crates/validation/src/scenario.rs
@@ -35,6 +35,12 @@ pub struct Scenario {
     runtime: Duration,
 }
 
+impl Default for Scenario {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl Scenario {
     /// Start a new scenario builder with sensible defaults.
     #[must_use]


rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -343,6 +343,259 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
         Ok(())
     }
 
+    /// Starts the controller with the given engine configurations. Runs till pipeline is shutdown then closes everything
+    /// Used in Validation test where shutdown signal is used to end pipeline groups and admin endpoint
+    /// Starts the controller with the given engine configurations.
```
lquerel, line 416:
```suggestion
    /// Starts the controller with the given engine configurations.
    /// ToDo [LQ] We need to minimize duplication of code here
```

rust/otap-dataflow/crates/validation/src/validation_exporter.rs
```diff
@@ -0,0 +1,129 @@
...
+    expect_failure: bool,
+}
+
+#[metric_set(name = "validation.assert.exporter.metrics")]
+#[derive(Debug, Default, Clone)]
+struct AssertValidExporterMetrics {
+    /// Number of comparisons that did not match expectation.
+    /// 0 -> not valid
+    /// 1 -> valid
+    valid: otap_df_telemetry::instrument::Gauge<u64>,
```
lquerel, line 56: Why a gauge and not a counter?

rust/validation_test/src/error.rs
```diff
@@ -0,0 +1,50 @@
...
+use std::io;
+
+#[derive(Debug)]
+pub enum PipelineError {
+    Io(io::Error),
+    Template(tera::Error),
+    Config(String),
+    Http(reqwest::Error),
+    Status(String),
+    Ready(String),
```
lquerel: We are trying to avoid leaking errors from our internal dependencies at the error API level. Even if we do not expose the functions or structs of a dependency, simply including that dependency's errors in our own error types makes users of our crate implicitly dependent on our internal dependencies.

In this case, we either need to explicitly extract the information we care about from the internal error, or fall back to a textual representation of it.

rust/otap-dataflow/crates/validation/src/metrics_types.rs
```diff
@@ -0,0 +1,119 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Deserialization types for the admin telemetry metrics endpoint.
```
lquerel, line 4: Will not that be possible to reuse the structs already used by the admin endpoint? 
TBH, I didn't check if it's possible but that could be a good idea to double check.

c1ly: the structs used by the admin endpoint are not public, reused some of the telemetry enums and structs

rust/otap-dataflow/crates/validation/src/validation_exporter.rs
c1ly: Added a counter for tracking comparisons passes and fails, and a gauge to keep track of the most recent comparison made

rust/validation_test/src/pipeline.rs
```diff
@@ -0,0 +1,310 @@
...
+const PIPELINE_GROUP_ID: &str = "validation_test";
+const PIPELINE_ID_TRAFFIC: &str = "traffic_gen";
+const PIPELINE_ID_SUV: &str = "suv";
+const PIPELINE_ID_VALIDATION: &str = "validate";
+pub const PIPELINE_CONFIG_YAML: &str = "pipeline_validation_configs.yaml";
+
+/// Helps distinguish between the message types
+#[derive(Debug, Deserialize, Serialize)]
+#[serde(rename_all = "snake_case")]
+pub enum PlatformType {
```
lquerel: PlatformType -> MessageType

rust/validation_test/README.md
```diff
@@ -0,0 +1,45 @@
+# Validation Tests
+
+The validation test validates a OTLP message after
```
lquerel:
```suggestion
The validation test validates OTLP or OTAP messages after
```

rust/validation_test/README.md
```diff
@@ -0,0 +1,45 @@
...
+experiencing various processes such as encoding/decoding
+or going through a pipeline.
+
+## Encoding/Decoding Validation
+
+To validate whether encoding/decoding is working properly
+we comparing the input and output to check that they are equal
```
lquerel:
```suggestion
we comparing the input and output to check that they are equivalent.
```

rust/otap-dataflow/crates/validation/README.md
```diff
@@ -0,0 +1,45 @@
...
+file. There are already some pipelines defined in the
+`pipeline_validation_configs.yaml` file feel free to use
+these as a reference when making your additions.
+Below are what each required key is used for
+
+- name -> Validation test name for your pipeline
+- pipeline_config_path -> Path to your pipeline config
+- loadgen_exporter_type -> What receiver type are you using (`otlp` or `otap`)
+- backend_receiver_type -> What exporter type are you using (`otlp` or `otap`)
+- transformative -> Does your pipeline modify or alter the data
```
lquerel:
```suggestion
- transformative -> Does your pipeline modify or alter the data

## Future directions

- Automatically trigger the validation process when a PR becomes "Ready for review".
- Manually trigger the validation process when a comment containing `@validation` is added to a PR (strecht goal).
- Extend the validation exporter to support more complex validation procedure. 
```

rust/otap-dataflow/crates/validation/src/pipeline.rs
```diff
@@ -0,0 +1,310 @@
...
+            if Instant::now() >= loadgen_deadline {
+                return Err(PipelineError::Ready(format!(
+                    "load generator did not reach {max_signals_sent} signals before timeout"
+                )));
+            }
+            sleep(Duration::from_secs(metric_poll_cooldown)).await;
+        }
+
+        // wait sometime to allow msgs to propogate through the pipelines
+        sleep(Duration::from_secs(pipeline_delay_sec)).await;
```
ThomsonTan: Wait until validation_exporter has received all messages?

rust/validation_test/README.md
```diff
@@ -0,0 +1,52 @@
+# Validation Tests
```
cijothomas: nit: Wondering why we use the word "validation" tests? Aren't all tests validating something already ?

cijothomas:
From a quick look - we should be able to leverage what we do in perf tests by adding a validation phase. Perftests orchestrates a setup, and check for count of messages sent and count of messages received - it does not assert that the count is same today, but doing that (plus validating the content itself, not just count) would effectively achieve something like an "integration test".. Is that also the intend behind the validation test added in this PR?

lquerel:
@cijothomas To provide some context for this PR, I asked @c1ly to work on a validation mechanism that:
- can be easily run both in CI and locally:
  The goal is to make correctness checks easily accessible to all contributors, whether they are running tests locally or validating changes in CI. This keeps correctness verification close to where changes are made and avoids coupling basic validation to the benchmark infrastructure or specialized environments.

- is fast to execute:
  Validation tests are designed to run in seconds and be part of the inner development loop. Unlike perf tests, which are long-running and resource-intensive by design, these checks need to be lightweight so they can be executed frequently and across many configurations.

- is extremely flexible in terms of validation scenarios:
  The intent is to validate pipeline behavior, not just message counts. This includes checking equivalence between input and output when appropriate, asserting semantic invariants or expected outputs, and validating engine behavior when faced with invalid or malformed data, which is explicitly out of scope for continuous performance benchmarks.
  
For now, this framework is defined outside of the main project, but I think it should eventually become a set of integration tests built on top of it. All of this is to say that this is clearly a work in progress, and that it makes sense to keep this effort separate from the continuous benchmark, which in my view has different requirements.

cijothomas:
Got it. (I didn't mean to merge with perf tests.)

https://github.com/open-telemetry/otel-arrow/issues/1931 is looking to add integration tests, which would be leveraging this (or vice-versa)

rust/validation_test/src/main.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Entry point to run pipeline validations manually instead of via tests.
```
cijothomas: what is the reason why this is *not* modeled as test and run via cargo test?

rust/otap-dataflow/crates/validation/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Future directions
+
+- Automatically trigger the validation process when a PR becomes "Ready for review".
```
cijothomas: this PR itself adds validation test to CI, so this is run for every PR already.

rust/validation_test/src/main.rs
c1ly: this was originally ran via cargo test but I changed it to provide a simple way for people to run the validation process locally, I'm open to changing it back to be triggered via cargo test especially when we migrate this crate inside otap-dataflow

cijothomas: I see. I am not quite sure if cargo run is any simpler than cargo test. cargo test can also spin up suv/load/ and assert things. 
The current main executable has a single exit which exits with non-zero if any tests fails, but with cargo test, we can cleanly know which tests failed vs not.

rust/otap-dataflow/crates/validation/src/pipeline.rs
c1ly: we can do with for the control msgs, my main concern is the msgs coming from the system under validation as some pipelines could filter the messages or batch them all up

rust/otap-dataflow/crates/validation/README.md
c1ly: instead of running on every PR we only want to run the validation process if a PR is ready for review

cijothomas: "ready for review" - did you mean Draft PR vs normal PR? 
Unless these tests are complicated and eats up lot of resource, we don't need to do any such selective running - we can always run them like any normal tests. 
But if they take complex setup and time, then we can make it conditional to be run only when a selected Label is added (eg: "integration-test")

c1ly: normal PR. right now they don't take that long to run since there are only two pipeline validation jobs defined, but the run duration will increase as people add more pipelines to validate. A Label to trigger the job would be useful here.

rust/validation_test/src/main.rs
c1ly: moved to test function, test function is ignored so it is only ran in a separate job when the PR is ready for review

rust/otap-dataflow/crates/validation/README.md
c1ly: added job to run pipeline validation when pr is ready for review, will add label trigger if needed in the future

rust/validation_test/README.md
c1ly: changed crate name validation_test -> validation and moved into otap-dataflow. README.md has been updated to reflect the name change

lalitb:
> https://github.com/open-telemetry/otel-arrow/issues/1931 is looking to add integration tests, which would be leveraging this (or vice-versa

@cijothomas - They're orthogonal but complement each other nicely. I believe this PR focuses on making sure data comes out of a pipeline intact, while the TLS tests make sure secure connections actually work - handshakes, certs, mTLS, all of that. Down the line, we should plug TLS scenarios into this validation setup so we can confirm that data still flows correctly when TLS is in the picture, without worrying about the TLS plumbing itself.

rust/otap-dataflow/crates/validation/Cargo.toml
```diff
@@ -0,0 +1,39 @@
...
+license.workspace = true
+publish.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+
+[lints]
+workspace = true
+
+[dependencies]
+async-trait = "0.1"
```
cijothomas: could we use workspace dependencies everywhere?

jmacd: Agree, please follow the pattern elsewhere like `= { workspace = true }` so we always pick a consistent version of async-trait for example across all crates.

c1ly: updated

Issue #1886, opened 2026-01-25
OPL Parser / columnar engine: suppor multiple assignments in one `set`/`extend` operator call


albertlockett:
Currently our expression for setting an value is defined like this:
https://github.com/open-telemetry/otel-arrow/blob/58180138e1dfd8118f9b7d39cad784272aa50b74/rust/otap-dataflow/crates/opl/src/opl.pest#L186-L188

But ideally, we'd support expressions like:
```
set attributes["x"] = 1, attributes["y"] = 2, resource.attributes["z"] = 3
```

To do:
- update the grammar & parser that were added in https://github.com/open-telemetry/otel-arrow/pull/1768
- update the planner and tests that will be added as part of https://github.com/open-telemetry/otel-arrow/issues/1875

albertlockett:
Noticed this feature gap when reviewing https://github.com/open-telemetry/otel-arrow/pull/1885

lquerel:
I imagine that `scope.attributes["x"] = 1` will also be supported.

albertlockett:
> I imagine that `scope.attributes["x"] = 1` will also be supported.

yes absolutely

albertlockett:
Closed in https://github.com/open-telemetry/otel-arrow/pull/1885

PR #1891, opened 2026-01-26
feat: support multiple assignments in set/extend operator (#1886)

gyanranjanpanda:
Update the OPL grammar and parser to accept multiple comma-separated assignment expressions in a single `set` or `extend` operator call.

- Modified `opl.pest` to allow `("," ~ assignment_expression)*` in `set_operator_call`.
- Updated `operator.rs` to iterate over all inner assignment rules.
- Added unit tests to verify multiple assignments are parsed correctly.

Resolves: #1886

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
@@ -457,4 +457,80 @@ mod test {
     async fn test_delete_all_attributes_opl_parser() {
         test_delete_all_attributes::<OplParser>().await;
     }
+
+    async fn test_set_multiple_attributes_same_destination<P: Parser>() {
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs
```diff
@@ -457,4 +457,40 @@ mod test {
     async fn test_delete_all_attributes_opl_parser() {
         test_delete_all_attributes::<OplParser>().await;
     }
+
```
albertlockett: I think we should have tests covering multiple assignments here.

We should have a test case like:
- `attributes["x"] = "y", attributes["x2"] = "y2"` (set multiple attributes)
- `attributes["x"] = "y", resource.attributes["x2"] = "y2"` (set multiple attributes on different destinations

PR #1897, opened 2026-01-27
Fix cargo warning about 'profiles for the non root package will be ignored'

AaronRM:
# Change Summary

Fixes a cargo warning about 'profiles for the non root package will be ignored' from the `query-engine` crate.

n/a

Validated that build warning is resolved

No.


albertlockett:
Thanks @AaronRM !

PR #1895, opened 2026-01-27
Self-instrumentation scope attributes 

jmacd:
# Change Summary

Part of #1576. 

Fixes #1903.

Portions of this PR were merged in #1898. Lines that are crossed out below have been merged already.

- ~Telemetry crate defines LogContext = SmallVec<[EntityKey; 1]>, LogContextFn = fn() -> LogContext~
- InternalTelemetrySystem: accepts context function; registry now passed in
- Controller passes node_context() first or pipeline_context() second
- Entity attribute set definition structs are re-ordered; first field becomes identifying for console logs
- Console exporter: now prints entity definition in scope attributes
- ~Internal telemetry receiver: now encodes scope information on receipt of each record (as singletons, currently)~
- Observed state store: prints scope information in symbolic form (for console_async, examples below)
- Entity registry: logs definition of each entity set for correlation in console logs
- Console direct logging: prints unsymbolized information (examples below)
- ~Self tracing encoder.rs: now encodes scope attributes from cached information~
- ~Self tracing formatter.rs: refactored for clarity, now supports optional suffix for use in console_direct, console_async modes~

## How are these changes tested?

Internal logging example configurations revised.

## Are there any user-facing changes?

Yes. Example logs, e.g., console exporter:

```
2026-01-27T01:29:54.567Z  RESOURCE   v1.Resource [service.id=1234, service.name=test]
2026-01-27T01:29:54.567Z  │ SCOPE    v1.InstrumentationScope [node.id=generator, node.urn=urn:otel:otap:fake_data_generator:receiver, node.type=receiver, pipeline.id=default_pipeline, pipeline.group.id=default_pipeline_group, core.id=0, numa.node.id=0, process.instance.id=AGN72ERHGR5OFI24GZVBC7YCNU, host.id=JoshCorpSurfaceLaptop, container.id=]
2026-01-27T01:29:54.567Z  │ └─ DEBUG otap-df-otap::rate_limit.sleep (crates/otap/src/fake_data_generator.rs:35
```

E.g., defining a new pipeline entity:

```
[...]

PR #1898, opened 2026-01-27
Formatting and encoding for scope attributes

jmacd:
# Change Summary

Part of https://github.com/open-telemetry/otel-arrow/issues/1576, part of #1903.

Half of #1895, for a reasonable sized PR.

This PR:

- Refactors the self_tracing formatter to fix poor structure. A new type StyledBufWriter separates the behavior of formatting log messages (w/ color option) from the behavior of ConsoleWriter.
- Adds ScopeFormatter argument to the basic log format, which formats a suffix. Different callers use this differently, e.g., raw_error! ignores it, console_direct/async will append a suffix, and console_exporter bypasses b/c scopes print on a separate line
- Adds ScopeToBytesMap for caching pre-calculated OTLP scope attributes as Bytes (with benchmark) and add a use in ITR
- Extends LogRecord with LogContext, defines LogContextFn to be configured later in #1895 
- Adds TODOs for console_direct, console_async, and ITS provider mode, currently using empty context

New test for encoding and formatting a scope/entity key.

No. See #1895.

### DIFF
--- a/rust/otap-dataflow/crates/telemetry/src/registry.rs
+++ b/rust/otap-dataflow/crates/telemetry/src/registry.rs
@@ -94,11 +94,7 @@ impl TelemetryRegistryHandle {
         F: FnOnce(&dyn AttributeSetHandler) -> T,
     {
         let reg = self.registry.lock();
-        if let Some(attrs) = reg.entities.get(key) {
-            Some(f(attrs))
-        } else {
-            None
-        }
+        reg.entities.get(key).map(|attrs| f(attrs))
     }
 
     /// Registers a metric set type with the given static attributes and returns a `MetricSet`


rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
```diff
@@ -372,11 +374,111 @@ fn encode_resource_attribute(buf: &mut ProtoBuffer, key: &str, value: &opentelem
...
 
+/// Pre-encoded scope attribute bytes keyed by EntityKey.
+///
+/// The Internal Telemetry Receiver uses this to avoid re-encoding scope
+/// attributes for each log event. Entity attributes are looked up once
+/// from the registry and encoded as InstrumentationScope.attributes bytes.
+///
+/// TODO: ScopeToBytesMap grows without any attention to managing memory.
+/// We will require a way to de-register entities that are no longer use
+/// or to flush memory to maintain a fixed size.
```
utpilla, line 385: Or we could simply cap the size and stop caching newer entries once it's full. Since we should typically have a fixed small number of entities, we can avoid the complexities of evicting entries from the cache.

```rust
const MAX_CACHED_ENTITIES: usize = 256;  // Something more than enough for any realistic config

pub fn get_or_encode(&mut self, key: EntityKey) -> Bytes {
    if let Some(cached) = self.cache.get(&key) {
        return cached.clone();
    }

    // Encode the entity attributes
    let mut buf = ProtoBuffer::with_capacity(128);
    self.registry.visit_entity(key, |attrs| {
        for (attr_key, attr_value) in attrs.iter_attributes() {
            encode_scope_attribute(&mut buf, attr_key, attr_value);
        }
    });
    let bytes = buf.into_bytes();

    // Only cache if under the limit
    if self.cache.len() < MAX_CACHED_ENTITIES {
        let _ = self.cache.insert(key, bytes.clone());
    }
    
    bytes
}
```

rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
```diff
@@ -372,11 +374,111 @@ fn encode_resource_attribute(buf: &mut ProtoBuffer, key: &str, value: &opentelem
...
+    pub fn get_or_encode(&mut self, key: EntityKey) -> Bytes {
+        if let Some(cached) = self.cache.get(&key) {
+            return cached.clone();
+        }
+
+        // Encode the entity attributes as KeyValue messages
+        let mut buf = ProtoBuffer::with_capacity(128);
+        self.registry.visit_entity(key, |attrs| {
+            for (attr_key, attr_value) in attrs.iter_attributes() {
+                encode_scope_attribute(&mut buf, attr_key, attr_value);
```
utpilla: It looks like we are encoding the scope attributes while holding the lock on registry. Is that something to worry about? If yes, we could consider cloning the attributes with the lock so that encoding can happen outside of the lock.

rust/otap-dataflow/crates/telemetry/src/registry.rs
```diff
@@ -88,6 +88,17 @@ impl TelemetryRegistryHandle {
         self.registry.lock().entities.visit_entities(f);
     }
 
+    /// Visits a single entity by key.
+    pub fn visit_entity<F>(&self, key: EntityKey, mut f: F)
+    where
+        F: FnMut(&dyn AttributeSetHandler),
```
utpilla: nit: I think `FnOnce` would be semantically closer to the usage here.

```suggestion
    pub fn visit_entity<F>(&self, key: EntityKey, f: F)
    where
        F: FnOnce(&dyn AttributeSetHandler),
```

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
```diff
@@ -54,14 +54,58 @@ pub enum ColorMode {
...
     /// Write an ANSI escape sequence (no-op for NoColor).
     #[inline]
-    fn write_ansi(self, w: &mut BufWriter<'_>, code: AnsiCode) {
-        if let ColorMode::Color = self {
-            let _ = write!(w, "\x1b[{}m", code as u8);
+    fn write_ansi(&mut self, code: AnsiCode) {
+        if let ColorMode::Color = self.color_mode {
+            let _ = write!(self.buf, "\x1b[{}m", code as u8);
         }
     }
```
utpilla, line 84: nit: We could precompute the byte arrays for each `AnsiCode` instead of calling `write!` each time.

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
```diff
@@ -54,14 +54,58 @@ pub enum ColorMode {
...
+    pub fn position(&self) -> usize {
+        self.buf.position() as usize
+    }
+
+    /// Check if the buffer is full (position >= capacity).
+    #[inline]
+    #[must_use]
+    pub fn is_full(&self) -> bool {
+        self.buf.position() as usize >= self.buf.get_ref().len()
+    }
```
utpilla:
```suggestion
    pub const fn position(&self) -> usize {
        self.buf.position() as usize
    }

    /// Check if the buffer is full (position >= capacity).
    #[inline]
    #[must_use]
    pub const fn is_full(&self) -> bool {
        self.buf.position() as usize >= self.buf.get_ref().len()
    }
```

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
```diff
@@ -388,102 +438,120 @@ impl ConsoleWriter {
...
+                w.write_body_and_attrs(record, has_event_name);
             },
+            write_suffix,
         );
     }
 
     /// Format a header line (RESOURCE, SCOPE) with attributes and custom formatters.
     ///
-    /// Unlike `format_log_line`, this takes raw attributes instead of a LogRecordView,
+    /// Unlike \`format_log_line\`, this takes raw attributes instead of a LogRecordView,
```
utpilla: Why the backslashes?

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
```diff
@@ -216,153 +277,142 @@ impl ConsoleWriter {
...
     where
-        F: FnOnce(&mut BufWriter<'_>),
+        F: FnOnce(&mut Self),
     {
-        self.color_mode.write_ansi(w, code);
-        f(w);
-        self.color_mode.write_ansi(w, AnsiCode::Reset);
+        self.write_ansi(code);
+        f(self);
+        self.write_ansi(AnsiCode::Reset);
```
utpilla, line 406: This method may leave the terminal in a colored state if the buffer gets full before the reset sequence gets written. It might not be worth fixing since the buffer has 4KB space so it shouldn't run out of space. Maybe just add a comment about it.

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: https://doc.rust-lang.org/std/keyword.const.html#compile-time-evaluable-functions
I am struggling with this still, like how get_ref() can be a constant function--on the surface of it, it seems like I have to know that get_ref() and position() are both constant before I can think to apply this `const`. I wonder if there's a clippy warning to tell you when things could be const that aren't?

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: I refactored _from_ that arrangement of code into this one because I thought it was more readable. :)

I'll add a TODO.

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: Because an agent updated the comment for me. (I don't want to admit how much time I spend deleting and rewriting comments to look like agents didn't write them.) 😞

rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
jmacd: I filed #1904

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
jmacd: I added a TODO, see this as part of #1746.

rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
jmacd: Nice catch! I cloned the values however note that AttrbuteSetHandler itself can't be cloned for dyn-compatibility reasons. https://github.com/open-telemetry/otel-arrow/pull/1898/commits/af06c41afdf02f2d029d88179d618c7cff384e46

jmacd: (Note: clippy didn't like my approach, but same idea:

```
    /// Visits a single entity by key.
    pub fn visit_entity<F, T>(&self, key: EntityKey, f: F) -> Option<T>
    where
        F: FnOnce(&dyn AttributeSetHandler) -> T,
    {
        let reg = self.registry.lock();
        reg.entities.get(key).map(|attrs| f(attrs))
    }
```

rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
lquerel: Soon we will support live reconfiguration (add/delete/update of pipelines and to some extent nodes). So the number of entities can fluctuate a lot and can for some scenario be more important than you think.

rust/otap-dataflow/crates/telemetry/src/self_tracing/formatter.rs
utpilla: @jmacd There's a Clippy warning for this: [`missing_const_for_fn`](https://rust-lang.github.io/rust-clippy/master/index.html?search=missing_const_for_fn)

However, this is in the `nursery` [category](https://github.com/rust-lang/rust-clippy?tab=readme-ov-file#clippy) which means it's still under development and is not enforced by default.

PR #1899, opened 2026-01-27
[otap-dataflow] Save source node in Pdata msg

### LINKED ISSUE #1880: Distinguish message source when using a fan in configuration
extension to #1860 
When using a fan in connection for example we have
nodes X and Y -> node Z

I would like to distinguish message source via port or node id when reading in messages in node z

Currently we receive messages from a singular message channel, for multiple inputs we could have multiple message channels (a channel for each node) or tag each incoming message with a node id

c1ly:
Defined two new effect handler extension traits one for local, one for shared that allows us to update otap pdata with the source node

```rust
/// Effect handler extension for adding message source
#[async_trait(?Send)]
pub trait MessageSourceLocalEffectHandlerExtension<PData> {
    /// Send data after tagging with the source node.
    async fn send_message_with_source_node(&self, data: PData) -> Result<(), TypedError<PData>>;
    /// Try to send data after tagging with the source node.
    fn try_send_message_with_source_node(&self, data: PData) -> Result<(), TypedError<PData>>;
    /// Send data to a specific port after tagging with the source node.
    async fn send_message_with_source_node_to<P>(
        &self,
        port: P,
        data: PData,
    ) -> Result<(), TypedError<PData>>
    where
        P: Into<PortName> + Send + 'static;
    /// Try to send data to a specific port after tagging with the source node.
    fn try_send_message_with_source_node_to<P>(
        &self,
        port: P,
        data: PData,
    ) -> Result<(), TypedError<PData>>
    where
        P: Into<PortName> + Send + 'static;
}

/// Send-friendly variant for use in `Send` contexts (e.g., `tokio::spawn`).
#[async_trait]
pub trait MessageSourceSharedEffectHandlerExtension<PData: Send + 'static> {
    /// Send data after tagging with the source node.
    async fn send_message_with_source_node(&self, data: PData) -> Result<(), TypedError<PData>>;
    /// Try to send data after tagging with the source node.
    fn try_send_message_with_source_node(&self, data: PData) -> Result<(), TypedError<PData>>;
    /// Send data to a specific port after tagging with the source node.
    async fn send_message_with_source_node_to<P>(
        &self,
        port: P,
        data: PData,
    ) -> Result<(), TypedError<PData>>
    where
        P: Into<PortName> + Send + 'static;
    /// Try to send data to a specific port after tagging with the source node.
    fn try_send_message_with_source_node_to<P>(
        &self,
        port: P,
        data: PData,
    ) -> Result<(), TypedError<PData>>
    where
        P: Into<PortName> + Send + 'static;
}
```

Added a field to the Context struct that will store the node information and added new functions for OtapPdata and Context getting and setting the source node

```rust
pub struct Context {
    source_node: Option<NodeId>,
    stack: Vec<Frame>,
}

...


  /// update the source node
  pub fn add_source_node(mut self, node_id: Option<NodeId>) -> Self {
      self.source_node = node_id;
      self
  }

  /// return the source node field
  pub fn get_source_node(&self) -> Option<NodeId> {
      self.source_node.clone()
  }
```

Updated pipeline nodes to use send_message functions that will tag otappdata with source node name

Closes #1880

### DIFF
--- a/rust/otap-dataflow/benchmarks/Cargo.toml
+++ b/rust/otap-dataflow/benchmarks/Cargo.toml
@@ -26,6 +26,7 @@ otap-df-engine = { path = "../crates/engine"}
 otap-df-telemetry = { path="../crates/telemetry"}
 otap-df-pdata = { path="../crates/pdata", features = ["bench"]}
 
+smallvec.workspace = true
 tracing.workspace = true
 tracing-subscriber = { workspace = true, features = ["registry"] }
 
--- a/rust/otap-dataflow/configs/fake-batch-console.yaml
+++ b/rust/otap-dataflow/configs/fake-batch-console.yaml
@@ -5,7 +5,7 @@ settings:
   default_pdata_channel_size: 100
 
 nodes:
-  receiver:
+  generator:
     kind: receiver
     plugin_urn: "urn:otel:otap:fake_data_generator:receiver"
     out_ports:
@@ -16,8 +16,8 @@ nodes:
     config:
       traffic_config:
         max_signal_count: 10000
-        max_batch_size: 100
-        signals_per_second: 100
+        max_batch_size: 10
+        signals_per_second: 10
         log_weight: 100
       registry_path: https://github.com/open-telemetry/semantic-conventions.git[model]
   batch:
@@ -32,9 +32,19 @@ nodes:
       otap:
         min_size: 1000
         sizer: items
-      format: otap
+      format: preserve
       flush_timeout: 3s
   console:
     kind: exporter
     plugin_urn: "urn:otel:console:exporter"
     config: {}
+
+service:
+  telemetry:
+    logs:
+      level: debug
+      providers:
+        engine: console_async
+        global: console_async
+        admin:    console_direct
+        internal: console_direct
--- a/rust/otap-dataflow/configs/internal-telemetry.yaml
+++ b/rust/otap-dataflow/configs/internal-telemetry.yaml
@@ -4,7 +4,7 @@ settings:
   default_pdata_channel_size: 100
 
 nodes:
-  receiver:
+  generator:
     kind: receiver
     plugin_urn: "urn:otel:otap:fake_data_generator:receiver"
     out_ports:
--- a/rust/otap-dataflow/crates/controller/Cargo.toml
+++ b/rust/otap-dataflow/crates/controller/Cargo.toml
@@ -28,3 +28,4 @@ core_affinity = { workspace = true }
 tokio = { workspace = true }
 tokio-util = { workspace = true }
 flume = { workspace = true }
+smallvec = { workspace = true }
--- a/rust/otap-dataflow/crates/controller/src/lib.rs
+++ b/rust/otap-dataflow/crates/controller/src/lib.rs
@@ -52,15 +52,19 @@ use otap_df_engine::context::{ControllerContext, PipelineContext};
 use otap_df_engine::control::{
     PipelineCtrlMsgReceiver, PipelineCtrlMsgSender, pipeline_ctrl_msg_channel,
 };
-use otap_df_engine::entity_context::set_pipeline_entity_key;
+use otap_df_engine::entity_context::{
+    node_entity_key, pipeline_entity_key, set_pipeline_entity_key,
+};
 use otap_df_engine::error::{Error as EngineError, error_summary_from};
 use otap_df_state::store::ObservedStateStore;
 use otap_df_telemetry::event::{EngineEvent, ErrorSummary, ObservedEventReporter};
+use otap_df_telemetry::registry::TelemetryRegistryHandle;
 use otap_df_telemetry::reporter::MetricsReporter;
 use otap_df_telemetry::{
     InternalTelemetrySettings, InternalTelemetrySystem, TracingSetup, otel_info, otel_info_span,
-    otel_warn,
+    otel_warn, self_tracing::LogContext,
 };
+use smallvec::smallvec;
 use std::sync::Arc;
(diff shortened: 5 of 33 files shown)


rust/otap-dataflow/crates/engine/src/local/processor.rs
```diff
@@ -48,7 +49,10 @@ use std::time::{Duration, Instant};
 
 /// A trait for processors in the pipeline (!Send definition).
 #[async_trait(?Send)]
-pub trait Processor<PData> {
+pub trait Processor<PData> 
+where 
+    PData: MessageSource
+{
```
jmacd: As I read this PR (still draft), it reminds me of the work to introduce the otap_df_otap::pdata::Context type. I struggled with a similar change to this, which introduce a trait binding for all PData and made for a large change. The choice to use an extension trait residing in otap_df_pdata worked around having to introduce this trait.

I wonder if we should treat the addition of source-node in the way `subscribe_to()` works? (for discussion)

c1ly: I'm not too familiar with how subscribe_to() works can you elaborate? would ideally like to avoid making lots of changes

jmacd:
> would ideally like to avoid making lots of changes

(a) I'll be glad to discuss this in depth!
(b) take a look at the effect handler extensions

```
/// Effect handler extensions for producers specific to data type.
#[async_trait(?Send)]
pub trait ProducerEffectHandlerExtension<PData> {
    /// Subscribe to a set of interests.
    fn subscribe_to(&self, int: Interests, ctx: CallData, data: &mut PData);
}

/// Effect handler extensions for consumers specific to data type.
#[async_trait(?Send)]
pub trait ConsumerEffectHandlerExtension<PData> {
    /// Triggers the next step of work (if any) in Ack processing.
    async fn notify_ack(&self, ack: AckMsg<PData>) -> Result<(), Error>;

    /// Triggers the next step of work (if any) in Nack processing.
    async fn notify_nack(&self, nack: NackMsg<PData>) -> Result<(), Error>;
}
```

These are features that the `crates/otap` components have access to but the engine does not directly. The way I think about this, these extensions enable certain types of behavior on PData that depends on the concrete layout of the struct, you can call `subscribe_to()` because you are holding an `OtapPdata` object that implements the extensions, but the engine has no way to use these APIs.

c1ly:
> > would ideally like to avoid making lots of changes
> 
> (a) I'll be glad to discuss this in depth! (b) take a look at the effect handler extensions
> 
> ```
> /// Effect handler extensions for producers specific to data type.
> #[async_trait(?Send)]
> pub trait ProducerEffectHandlerExtension<PData> {
>     /// Subscribe to a set of interests.
>     fn subscribe_to(&self, int: Interests, ctx: CallData, data: &mut PData);
> }
> 
> /// Effect handler extensions for consumers specific to data type.
> #[async_trait(?Send)]
> pub trait ConsumerEffectHandlerExtension<PData> {
>     /// Triggers the next step of work (if any) in Ack processing.
>     async fn notify_ack(&self, ack: AckMsg<PData>) -> Result<(), Error>;
> 
>     /// Triggers the next step of work (if any) in Nack processing.
>     async fn notify_nack(&self, nack: NackMsg<PData>) -> Result<(), Error>;
> }
> ```
> 
> These are features that the `crates/otap` components have access to but the engine does not directly. The way I think about this, these extensions enable certain types of behavior on PData that depends on the concrete layout of the struct, you can call `subscribe_to()` because you are holding an `OtapPdata` object that implements the extensions, but the engine has no way to use these APIs.

Ah I see, I like this approach to allow the engine to be flexible with the generic PData, while giving us what we want specifically for OtapPData. 

I'm thinking of defining a trait like this where these send message functions will update the OtapPdata with the source_node information 
```rust
#[async_trait(?Send)]
pub trait MessageSourceEffectHandlerExtension<PData> {
    async fn send_message_with_source_node(&self, data: &mut PData) -> Result<(), TypedError<PData>>;
    fn try_send_message_with_source_node(&self, data: &mut PData) -> Result<(), TypedError<PData>>;
    async fn send_message_with_source_node_to<P>(&self, port: P, data: &mut PData) -> Result<(), TypedError<PData>>;
    fn try_send_message_with_source_nodeto<P>(&self, port: P, data: &mut PData) -> Result<(), TypedError<PData>>;
}
```

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -246,6 +256,17 @@ impl OtapPdata {
...
     }
+
+    /// update the source node
+    pub fn add_source_node(mut self, node_id: Option<NodeId>) -> Self {
+        self.source_node = node_id;
+        self
+    }
+
+    /// return the source node field
+    pub fn get_source_node(&self) -> Option<NodeId> {
```
ThomsonTan, line 274: Seems this method is only called by test function? If so, mark it as `#[cfg(test)]`?

c1ly: I plan to use this function outside of a test in my other MR #1879 to get node information from the messages

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -158,13 +163,18 @@ impl OtapPdata {
         Self {
             context: Context::default(),
             payload,
+            source_node: None,
```
jmacd: I would prefer to see this in `Context`, not `OtapPdata`, unless you see a reason not to. I would still have the default be None, so it won't impact current Context users much.

jmacd (changes requested):
This approach and the new extension traits look good. I would like to see the new `source_node` field be a part of Context.

PR #1911, opened 2026-01-29
fix: Field to batch index mapping in otap batch unify

JakeDern:
# Change Summary

Part of the otap batch `unify` logic tracks which otap batches have which fields. The implementation extracts the schemas for some payload type from each batch and assumes that the index in the schemas list is equivalent to the index in the `batches` slice. 

However, `select` filters out missing payload types from each batch, so if some batches are missing a payload then the index is not the same.

The fix is to maintain the 1:1 mapping of schema index to batch index by not filtering out missing batches.

Related to #1334, but there are still more issues listed there.

Uncommenting the complex metrics tests. The tests now make it farther and some scenarios see more success, but there are still at least two more known issues.

No.


JakeDern:
CC: @albertlockett @jmacd - More issues to solve as mentioned in the PR description, but some progress!

albertlockett (approved):

PR #1916, opened 2026-01-29
fix: premature pipeline control channel closure during shutdown

### LINKED ISSUE #1915: Pipeline control channel closed prematurely during shutdown, preventing nodes from sending cleanup messages
## Summary

The `PipelineCtrlMsgManager` breaks immediately after receiving a `Shutdown` message, which drops the pipeline control channel before nodes have finished their cleanup sequence. This causes nodes that try to send control messages (e.g., `CancelTimer`, `CancelTelemetryTimer`) during shutdown to fail with channel closed errors.

## Symptoms

Nodes that attempt to cancel timers or send other cleanup messages after receiving a shutdown signal encounter errors like:

`PipelineControlMsgError { error: "Channel is closed..." }`

This affects:
- **Receivers** (e.g., `otlp_receiver`, `otap_receiver`, `syslog_cef_receiver`) that cancel telemetry timers on shutdown
- **Processors** that cancel periodic telemetry collection timers when their message channel closes
- **Exporters** (e.g., `otlp_exporter`, `otap_exporter`, `parquet_exporter`) that cancel timers during shutdown

## Root Cause

In `PipelineCtrlMsgManager::run()`, when a `Shutdown` message is received:

1. The manager sends shutdown to all receivers via `shutdown_receivers()`
2. The manager immediately `break`s out of the event loop
3. This drops `self.pipeline_ctrl_msg_receiver`, closing the channel
4. Nodes that try to send cleanup messages find the channel already closed

The issue is a race condition: the manager exits before nodes have a chance to complete their cleanup.

## Impact

There are 13 call sites in the codebase that call `.cancel().await` during shutdown handling, all of which could potentially be affected:

| File | Occurrences |
|------|-------------|
| `otlp_receiver.rs` | 4 |
| `otap_receiver.rs` | 1 |
| `otlp_exporter.rs` | 1 |
| `otap_exporter.rs` | 1 |
| `parquet_exporter.rs` | 1 |
| `perf_exporter.rs` | 1 |
| `syslog_cef_receiver.rs` | 2 |
| `processor.rs` | 2 |

## Expected Behavior

After receiving a shutdown message, the manager should continue running in a "draining" mode to allow nodes to send cleanup messages. The manager should only exit when all senders have dropped their end of the channel, indicating all nodes have completed their cleanup.

## Proposed Fix

1. Add an `is_draining` flag to track shutdown state
2. After receiving `Shutdown`, enter draining mode instead of breaking immediately
3. Continue processing cleanup messages (`CancelTimer`, `CancelTelemetryTimer`, `DeliverAck`, `DeliverNack`) while draining
4. Ignore new work requests (timer starts, delayed data) during draining
5. Exit the loop when the channel is closed (all senders dropped)

AaronRM:
# Change Summary

Fixes an issue where the PipelineCtrlMsgManager breaks immediately after receiving a Shutdown message, which drops the pipeline control channel before nodes have finished their cleanup sequence. This causes nodes that try to send control messages (e.g., CancelTimer, CancelTelemetryTimer) during shutdown to fail with channel closed errors.

* Closes #1915 

Added 5 additional tests to validate the shutdown handling in `PipelineCtrlMsgManager`.

No

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
+++ b/rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
@@ -25,9 +25,9 @@ use std::collections::{BinaryHeap, HashMap};
 use std::time::{Duration, Instant};
 
 /// Maximum time buffer between the end of draining and the shutdown deadline.
-/// The actual buffer is the minimum of this value and 10% of the time until deadline,
-/// ensuring we use at least 90% of available time for draining while capping the
-/// post-draining buffer at 1 second.
+/// The actual buffer is the minimum of this value and 10% of the time until deadline.
+/// This is a best-effort attempt to ensure that `PipelineCtrlMsgManager` exits before
+/// the caller's timeout fires, avoiding a race where both expire simultaneously.
 const MAX_DRAINING_BUFFER_DURATION: Duration = Duration::from_secs(1);
 
 /// Represents delayed data with scheduling information.


rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -223,18 +229,68 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
+                    break;
+                }
+            }
+
+            // Get the next expirations, if any (skip during draining).
+            let next_expiry = if is_draining {
+                None
+            } else {
+                self.tick_timers.next_expiry()
+            };
```
jmacd:
```suggestion
            let next_expiry = is_draining.then(self.tick_timers.next_expiry);
```
I think (this compiles and) is nicer to read.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -223,18 +229,68 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
+                None
+            } else {
+                self.delayed_data.peek().map(|d| d.when)
+            };
+            // During draining, use the deadline as the next expiry to ensure we check it
+            let next_earliest = if is_draining {
+                draining_deadline
+            } else {
+                opt_min(opt_min(next_expiry, next_tel_expiry), next_delay_expiry)
+            };
```
jmacd: I'd do the same here too, https://doc.rust-lang.org/std/primitive.bool.html#method.then

jmacd (approved):
Looks good! I have a recommendation about boolean-to-option combinators then() (also excited for the future boolean::ok_or and ok_or_else).

lalitb (approved):
LGTM - optional nit: could add tests validating that `TimerTick` doesn't fire during draining and `StartTimer` is ignored during draining, but not necessary.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -243,36 +294,85 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
-                            break;
+                            // Enter draining mode instead of breaking immediately.
+                            // This allows nodes to send cleanup messages (e.g., CancelTimer,
+                            // CancelTelemetryTimer) during their shutdown sequence.
+                            // The loop will exit when all senders drop their channel ends
+                            // or when the draining deadline is reached.
+                            //
+                            // Calculate draining deadline to use at least 90% of available time,
+                            // with the gap capped at MAX_DRAINING_BUFFER. This maximizes draining
+                            // time while ensuring some buffer for post-draining work.
```
utpilla: Do we need this 90% calculation? What's the post draining work that requires this?

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -243,36 +294,85 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
-                                    self.event_reporter.report(EngineEvent::pipeline_runtime_error(self.pipeline_key, "Shutdown failed", err_summary))
+                                    self.event_reporter.report(EngineEvent::pipeline_runtime_error(self.pipeline_key.clone(), "Shutdown failed", err_summary))
                                 }
                             }
-                            break;
+                            // Enter draining mode instead of breaking immediately.
+                            // This allows nodes to send cleanup messages (e.g., CancelTimer,
+                            // CancelTelemetryTimer) during their shutdown sequence.
+                            // The loop will exit when all senders drop their channel ends
+                            // or when the draining deadline is reached.
```
utpilla: I think we shouldn't worry about processing `CancelTimer` and `CancelTelemetryTimer` messages once the pipeline has received shutdown message. The nodes that push the cancel timer messages already ignore the result of `cancel()` operation.

```rust
```
Instead, to keep things simple, we could cancel all the timers and telemetry timers proactively on receiving the shutdown message. This would also give more time to process `DeliverAck` and `DeliverNack` messages which are more important.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -243,36 +294,85 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
+                                otel_debug!(
+                                    "pipeline.draining.ignored_delay_data",
+                                    node_id = node_id,
+                                    "Ignoring DelayData during shutdown draining"
+                                );
+                            } else {
+                                let delayed = Delayed { node_id, when, data };
+                                self.delayed_data.push(delayed);
+                            }
                         }
```
utpilla, line 377: There is another `tokio::select` [branch](https://github.com/open-telemetry/otel-arrow/blob/6ad291b19e1b329ce9441810ea2b4a41cd1085eb/rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs#L285-L355) in this loop which also shouldn't be run during shutdown.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -243,36 +294,85 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
                         PipelineControlMsg::CancelTelemetryTimer { node_id, _temp } => {
                             self.telemetry_timers.cancel(node_id);
                         }
                         PipelineControlMsg::DelayData { node_id, when, data } => {
-                            let delayed = Delayed { node_id, when, data };
-                            self.delayed_data.push(delayed);
+                            if is_draining {
+                                otel_debug!(
```
utpilla: Maybe we should use `otel_warn!` instead since this would be a data loss issue?

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -223,18 +229,63 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
+            let next_delay_expiry = (!is_draining)
+                .then(|| self.delayed_data.peek().map(|d| d.when))
+                .flatten();
+            // During draining, use the deadline as the next expiry to ensure we check it.
+            // Otherwise, find the earliest of all timer expirations using opt_min (which
+            // returns the minimum of two Option<T> values, treating None as "no value").
+            let next_earliest = is_draining
+                .then_some(draining_deadline)
+                .flatten()
+                .or_else(|| opt_min(opt_min(next_expiry, next_tel_expiry), next_delay_expiry));
```
utpilla: This seems a bit complicated. I think we could simplify it by only checking `is_draining` once.

```rust
let next_earliest = if is_draining {
    draining_deadline
} else {
    opt_min(
        opt_min(self.tick_timers.next_expiry(), self.telemetry_timers.next_expiry()),
        self.delayed_data.peek().map(|d| d.when)
    )
};
```

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
AaronRM: There's nothing special about "10% up to 1s", but we need the draining to complete prior to the deadline received from the PipelineControlMsg::Shutdown message so we use this heuristic.

If we get a 60s shutdown deadline, then up to 59s can be used for draining. If we get a 1s shutdown deadline, then .9s is used for draining. If we get a .1s shutdown deadline then .09s is used for draining.

There is no real "post draining work" to do. It's just providing some additional buffer in case of delays in the timer callback due to async scheduling jitter, etc....

utpilla: Thanks for the explanation but I still don't see any additional guaranteed value in using this buffer approach. The deadline check only happens at the top of each loop iteration. If we're mid-iteration (e.g., awaiting a `DeliverAck` send), we could still overshoot both the buffer and the original deadline. We're essentially "best effort" either way.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
AaronRM: Sure, I had switched all of these from warn => debug per request from @lalitb, but will switch this one back to warn. FWIW, this shouldn't result in data loss with the incoming 'persistence' processor. 😄

utpilla: Oh. Sorry I didn't check the previous comments on the PR 😄

> FWIW, this shouldn't result in data loss with the incoming 'persistence' processor. 😄

But not every pipeline would have persistence processor though. Even if it's present, I believe it would depend on where it’s placed in the pipeline. If used appropriately, it would definitely help avoid this issue.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
AaronRM: Yes, it's best effort. However, timing out the draining 'internally' (before the external timeout happens) gives us a better chance for a graceful shutdown. Without it we are more likely to hit timeout issues on shutdown [like this one which occurred on an earlier commit in this branch](https://github.com/open-telemetry/otel-arrow/actions/runs/21493586111/job/61922362659) without the 'buffer'.

This additional 'buffer' was meant as a small change to address that issue. I think a separate PR might be better to focus on improving the full shutdown sequence to possibly get beyond best effort.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -18,12 +18,18 @@ use crate::pipeline_metrics::PipelineMetricsMonitor;
...
+use otap_df_telemetry::{otel_debug, otel_warn};
 use std::cmp::Reverse;
 use std::collections::{BinaryHeap, HashMap};
 use std::time::{Duration, Instant};
 
+/// Maximum time buffer between the end of draining and the shutdown deadline.
+/// The actual buffer is the minimum of this value and 10% of the time until deadline,
+/// ensuring we use at least 90% of available time for draining while capping the
+/// post-draining buffer at 1 second.
+const MAX_DRAINING_BUFFER: Duration = Duration::from_secs(1);
```
ThomsonTan: nit: name it like `MAX_DRAINING_BUFFER_IN_SEC`, or it looks a config of buffer size.

utpilla (approved):

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
AaronRM: Renamed to `MAX_DRAINING_BUFFER_DURATION`

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
utpilla: Thanks! I understand how that buffer helps now.

I think "post-draining work" could be misleading. We could rephrase it to something like this maybe?

```diff
/// The draining deadline uses at least 90% of the available time, with the gap between
/// draining end and shutdown deadline capped at [`MAX_DRAINING_BUFFER`]. This ensures
- /// we maximize draining time while reserving a reasonable buffer for post-draining work.
+ /// that `PipelineCtrlMsgManager` exits before the caller's timeout fires, avoiding a race where
+ /// both expire simultaneously.
```

AaronRM: Great suggestion, thanks! Updated the comment.

utpilla: We should do it here as well:

```rust
/// Maximum time buffer between the end of draining and the shutdown deadline.
/// The actual buffer is the minimum of this value and 10% of the time until deadline,
/// ensuring we use at least 90% of available time for draining while capping the
/// post-draining buffer at 1 second.
const MAX_DRAINING_BUFFER_DURATION: Duration = Duration::from_secs(1);
```

PR #1919, opened 2026-01-29
Allow mixed local/shared pdata senders

lquerel:
# Change Summary

  - Allow local receivers/processors to use the generic message::Sender so mixed local/shared edges can share channels safely.
  - Introduce ChannelMode to centralize control-channel wiring and metrics, reducing duplication across wrappers making the overall design less error-prone.
  - Add pipeline test for mixed local/shared receivers targeting the same exporter.
  
  NA
  
 See pipeline_tests.rs

  No


lalitb (approved):
LGTM - Nice cleanup using ChannelMode.

PR #1922, opened 2026-01-29
fix: Batches may differ by field order after unification

JakeDern:
# Change Summary

Note this is a band-aid to avoid larger changes, but it does solve a bunch of panics.

- Project batches to the merged schema before coalescing (reorder the fields to be the same)

Related to: https://github.com/open-telemetry/otel-arrow/issues/1334.

New unit tests for the coalescing.

No.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap/groups.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/groups.rs
@@ -1006,7 +1006,7 @@ fn generic_schemaless_concatenate<const N: usize>(
         if select(batches, i).next().is_some() {
             let schema = Arc::new(
                 // Note: The schemas of all the record batches should only differ
-                // by orer at this point. Schema::try_merge is just going to pick
+                // by order at this point. Schema::try_merge is just going to pick
                 // a canonical order for the fields, which we will project
                 // every record batch to.
                 //


JakeDern:
@jmacd - There is still at least one metrics panic after this, but it is much rarer on my machine (when using the below config). I know for a fact that the re-indexing logic is also bust, so the output is junk regardless.

I'm hoping that this stops the bleeding enough to start deeper refactoring, but please ping me if there's a frequent panic plaguing you. 

```
settings:
  default_pipeline_ctrl_msg_channel_size: 100
  default_node_ctrl_msg_channel_size: 100
  default_pdata_channel_size: 100

nodes:
  receiver:
    kind: receiver
    plugin_urn: "urn:otel:otap:fake_data_generator:receiver"
    out_ports:
      out_port:
        destinations:
          - noop
    config:
      traffic_config:
        max_signal_count: 10000
        max_batch_size: 10
        signals_per_second: 1
        log_weight: 100
  noop:
    kind: exporter
    plugin_urn: "urn:otel:noop:exporter"
    config: {}

# Internal telemetry pipeline - separate from main pipeline
# Uses hardcoded settings: single thread, no admin server
internal:
  telemetry:
    kind: receiver
    plugin_urn: "urn:otel:internal_telemetry:receiver"
    out_ports:
      out_port:
        destinations:
          - batch
    config: {}
  batch:
    kind: processor
    plugin_urn: "urn:otel:batch:processor"
    out_ports:
      out_port:
        destinations:
          - console
    config:
      otap:
        min_size: 1000
        sizer: items
      format: otap
      flush_timeout: 3s
  console:
    kind: exporter
    plugin_urn: "urn:otel:console:exporter"
    config: {}

service:
  telemetry:
    logs:
      level: "debug"
      providers:
        global: its
        engine: its
        internal: console_direct
        admin: its
    resource:
      service.id: 1234
      service.name: test

```

rust/otap-dataflow/crates/pdata/src/otap/groups.rs
```diff
@@ -1004,6 +1005,13 @@ fn generic_schemaless_concatenate<const N: usize>(
         // ignore all the rows where every item is None
         if select(batches, i).next().is_some() {
             let schema = Arc::new(
+                // Note: The schemas of all the record batches should only differ
+                // by orer at this point. Schema::try_merge is just going to pick
```
jmacd:
```suggestion
                // by order at this point. Schema::try_merge is just going to pick
```

jmacd (approved):

rust/otap-dataflow/crates/pdata/src/otap/groups.rs
```diff
@@ -1034,6 +1039,19 @@ fn generic_schemaless_concatenate<const N: usize>(
     Ok(result)
 }
 
+/// Given a target schema and a record batch, align the columns to the target schema.
+/// Fails if the record batch is not a subset of the target schema.
+fn project_to_schema(schema: SchemaRef, rb: &RecordBatch) -> Result<RecordBatch> {
+    let projection: Vec<usize> = schema
+        .fields()
+        .iter()
+        .filter_map(|f| rb.schema().index_of(f.name()).ok())
```
albertlockett, line 1048: The error created by indexof is somewhat expensive to compute as it tries to format the error string. We should probably use schema.fields().find().is_some() in place of indexof().ok()

https://docs.rs/arrow-schema/57.2.0/src/arrow_schema/schema.rs.html#398

JakeDern: Thanks for the tip! In the future I'll look for APIs with options instead of results when I'm not interested in propagating errors: https://github.com/open-telemetry/otel-arrow/pull/1924

albertlockett: Thanks! Yeah, I need to get into that habbit myself as well. This particular issue with `index_of` I noticed when I was profiling some of the transport encoding stuff (albeit in a different part of the codebase).

PR #1924, opened 2026-01-30
refactor: Use `.fields.find()` instead of `.index_of()` to look up field indices when batching

JakeDern:
# Change Summary

Swap out the `index_of` API which creates and expensive string on the failure/missing case for `.fields.find()` API which just returns an option.

Albert pointed this out to me here: https://github.com/open-telemetry/otel-arrow/pull/1922#discussion_r2744264230

No.


albertlockett (approved):
Thanks @JakeDern :D

PR #1925, opened 2026-01-30
OTLP receiver: enable HTTP-only mode

### LINKED ISSUE #1893: `otlpreceiver` support for `http`-only Configurations
I noticed one other disparity between the [Go collector `otlpreceiver`](https://github.com/open-telemetry/opentelemetry-collector/tree/main/receiver/otlpreceiver) configuration block and the implementation in this repository.

Go configuration:
```yaml
receivers:
  otlp:
    protocols:
      grpc:
      http:
```

[Rust configuration](https://github.com/open-telemetry/otel-arrow/blob/3f0c85c4d65a91562de3165088edececc378f0eb/rust/otap-dataflow/crates/otap/src/otlp_receiver.rs#L73):
```rust
pub struct Config {
    /// Shared gRPC server settings reused across gRPC-based receivers.
    #[serde(flatten)]
    pub grpc: GrpcServerSettings,

    /// Optional OTLP/HTTP server settings.
    ///
    /// When configured, the receiver will start an additional HTTP server implementing
    /// `POST /v1/logs`, `POST /v1/metrics`, and `POST /v1/traces`.
    #[serde(default)]
    pub http: Option<HttpServerSettings>,
}
```

In the Rust version, `grpc` configuration is required - there is no way for a customer to configure ONLY an `http` endpoint. This was loosely mentioned in the PR description of #1765:
> Adds HTTP/1.1 support for the OTLP receiver alongside the existing gRPC server. HTTP is disabled by default—existing gRPC-only deployments are unaffected.

Now that the initial support for HTTP endpoint is in, should we refactor to match the same flexibility of the Go component in supporting ALL of the following scenarios:
- Single receiver component listens for gRPC OTLP on port 4317 (no HTTP server)
- Single receiver component listens for HTTP OTLP on port 4318 (no gRPC server) <-- Not possible today
- Single receiver component listens for gRPC OTLP on port 4317 and HTTP OTLP on port 4318

lalitb:
# Change Summary

  This PR restructures the OTLP receiver configuration to support flexible protocol deployment modes, aligning with the Go collector's otlpreceiver model:                            
                                                                                                                                                                                      
  - gRPC only - Configure only protocols.grpc                                                                                                                                         
  - HTTP only - Configure only protocols.http (new!)                                                                                                                                  
  - Both protocols - Configure both with a global concurrency cap   

## Key Changes                                                                                                                                                                         
                                                                                                                                                                                      
  ### Configuration restructure:                                                                                                                                                          
  - Moved from flat config to protocols.grpc / protocols.http structure                                                                                                               
  - TLS configuration is now per-protocol (under each protocol's config)                                                                                                              
  - At least one protocol must be configured (validated at startup)                                                                                                                   
                                                                                                                                                                                      
  ### Concurrency model for dual-protocol mode:                                                                                                                                           
  - Each protocol enforces its own max_concurrent_requests limit                                                                                                                      
  - When both protocols are enabled, an additional global semaphore caps combined load to prevent exceeding downstream capacity                                                       
  - Permits acquired in consistent order (global -> local) to prevent deadlocks         

* Closes #1893

 Manual tested, along with unit tests.

 ⚠️ Breaking change: The OTLP receiver configuration format has changed.                                                                                                             
                                                                                                                                                                                      
  **_Before_**:

```yaml                                                                                              
  config:                                                                                                                                                                             
    listening_addr: "127.0.0.1:4317"                                                                                                                                                  
    tls:                                                                                                                                                                              
      cert_file: "/path/to/cert"                                                                                                                                                      
    http:                                                                                                                                                                             
      listening_addr: "127.0.0.1:4318"
```                                                                                                                                              
                                                                                                                                                                                      
  **_After_**:

```yaml                                                                                                                                                                   
  config:                                                                                                                                                                             
    protocols:                                                                                                                                                                        
      grpc:                                                                                                                                                                           
        listening_addr: "127.0.0.1:4317"                                                                                                                                              
        tls:                                                                                                                                                                          
          cert_file: "/path/to/cert"                                                                                                                                                  
      http:                                                                                                                                                                           
        listening_addr: "127.0.0.1:4318"                                                                                                                                              
        tls:                                                                                                                                                                          
          cert_file: "/path/to/cert"
```                                                                                                                                                
                                                                                                                                                                                                   
Refer to `otlp_receiver.md` (updated in this PR) for more details.


rust/otap-dataflow/docs/otlp-receiver.md
```diff
@@ -28,34 +28,37 @@ and HTTP/1.1 protocols with unified concurrency control.
...
+               |                                          |
+               | poll_ready                               | acquire_owned()
+               | (backpressure)                           | + timeout
+               +--------------------+---------------------+
+                                    |
+                          +---------v---------+
+                          | OtapPdata         |
+                          |   context: Context|
+                          |   payload: OtlpBytes  | <-- Raw serialized
+                          +-----------+-----------+
```
albertlockett, line 50: Cleaning up the ascii art :) 
```suggestion
                          +---------v-------------+
                          | OtapPdata             |
                          |   context: Context    |
                          |   payload: OtlpBytes  | <-- Raw serialized
                          +-----------+-----------+
```

rust/otap-dataflow/docs/otlp-receiver.md
```diff
@@ -28,34 +28,37 @@ and HTTP/1.1 protocols with unified concurrency control.
...
+       +------------------------------v-------------------------------+
+       |              Shared Infrastructure (gRPC + HTTP)             |
+       |  - AckRegistry (wait_for_result subscriptions)               |
+       |  - Unified Metrics (requests, acks, nacks, errors)           |
+       |  - Control Loop (ACK/NACK routing, telemetry, shutdown)      |
+       +------------------------------+-------------------------------+
+                                      |
+                          +-----------v-----------+
+                          | Pipeline (Bounded MPSC)|
+                          +-----------------------+
```
albertlockett, line 61:
```suggestion
                          +-----------v------------+
                          | Pipeline (Bounded MPSC)|
                          +------------------------+
```

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
```diff
@@ -67,24 +66,101 @@ pub const OTLP_RECEIVER_URN: &str = "urn:otel:otlp:receiver";
...
-    /// TLS configuration
-    #[cfg(feature = "experimental-tls")]
-    pub tls: Option<TlsServerConfig>,
+impl Protocols {
+    /// Returns `true` if at least one protocol is configured.
+    #[inline]
+    #[must_use]
+    pub fn is_valid(&self) -> bool {
+        self.grpc.is_some() || self.http.is_some()
+    }
```
albertlockett, line 156: What would happen if user specified both http and grpc using the same listening address? I assume one of the servers would fail to start? Should we also check for this case in this `is_valid` method?

jmacd: I would not see that as a problem, necessarily. In the Golang environment, I know that it is possible to share ports between gRPC and HTTP/2, however it is not convenient or well supported. The Collector does not support shared ports between the two.

jmacd (approved):

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
```diff
@@ -67,24 +66,101 @@ pub const OTLP_RECEIVER_URN: &str = "urn:otel:otlp:receiver";
...
+
+    /// Returns `true` if both protocols are configured.
+    #[inline]
+    #[must_use]
+    pub fn has_both(&self) -> bool {
+        self.grpc.is_some() && self.http.is_some()
+    }
```
utpilla, line 163: Do we need to support both HTTP and gRPC on the same receiver instance? Users can always create dedicated receivers for each protocol.

I'm asking mainly because supporting both adds complexity:

- Dual semaphore logic for unified backpressure (covered in this PR)
- Metrics don't distinguish traffic by protocol (not covered yet)
- ~~Port conflict validation (not covered yet?)~~ Covered in https://github.com/open-telemetry/otel-arrow/pull/1925/commits/ab1678a9dd17a19dd21c47e749266445f257c4f9

Having separate receivers avoid all of these.

lalitb: Good question! The Go collector's `otlpreceiver` supports both protocols in a single receiver, so the approach was to maintain the parity with Go in terms of user experience and config structure.  The approach was outlined in the original Github issue [here](https://github.com/open-telemetry/otel-arrow/issues/1144).            
                                                                                                                                                             
  The unified backpressure is actually an advantage of the single-instance approach, not just added complexity. When both protocols share one receiver, we can enforce a global concurrency cap tied to downstream pipeline capacity - ensuring combined HTTP + gRPC load doesn't overwhelm the pipeline. With separate receivers, each would independently push traffic, and coordinating backpressure across them would be harder (or require engine-level support).    
                                                                                                                                                             
  For the other points:                                                                                                                                      
  - Metrics: Currently unified. Could add per-protocol differentiation if there's demand, but keeping it simple for now.                                     
  - Port conflict validation: Added - configs with the same address for both protocols now fail early with a clear error message.

lalitb: Also, users who prefer isolation can still create separate receiver instances (e.g., otlp-grpc and otlp-http nodes each with one protocol) - the single-instance approach is optional, giving users flexibility based on their needs.

```yaml
 receivers:                                                                                                                                                 
    otlp-grpc:                                                                                                                                               
      config:                                                                                                                                                
        protocols:                                                                                                                                           
          grpc:                                                                                                                                              
    otlp-http:                                                                                                                                               
      config:                                                                                                                                                
        protocols:                                                                                                                                           
          http:                                                                                                                                              
            listening_addr: "0.0.0.0:4318"  
```

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
```diff
@@ -142,48 +225,82 @@ impl OTLPReceiver {
...
+        if !config.protocols.is_valid() {
+            return Err(otap_df_config::error::Error::InvalidUserConfig {
+                error: "At least one protocol (grpc or http) must be configured under 'protocols'"
+                    .to_string(),
+            });
+        }
+
+        // Validate that gRPC and HTTP do not use the same listening address.
+        if let (Some(grpc), Some(http)) = (&config.protocols.grpc, &config.protocols.http) {
+            if grpc.listening_addr == http.listening_addr {
```
ThomsonTan: This check might be a little loose, like `0.0.0.0:4317` and `127.0.0.1:4317` passes the check, but this should still not be allowed. Perhaps just check the port number.

lalitb: Fair point - however, there are also other edge cases like the IPv4/IPv6 dimension (0.0.0.0 vs [::]) and dual-stack socket behavior that varies by OS. Also, address conflicts can happen across components (e.g., a receiver and an exporter both binding to the same port). Given these complexities, engine-level validation (tracking all listening addresses across components) seems like the right long-term approach. For now, the local check catches the obvious same-address mistake, and the OS handles the rest.

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
lalitb: Thanks, Added validation that rejects configs where both protocols use the same listening address, with a clear error message at config parsing time.

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
lalitb: Also to add,

> Perhaps just check the port number.

This would be too restrictive, as the same port number can be configured on different interfaces, which is a valid use case: 

```yaml
 protocols:                                                                                                                                                 
    grpc:                                                                                                                                                    
      listening_addr: "192.168.1.100:4317"  # LAN                                                                                                            
    http:                                                                                                                                                    
      listening_addr: "10.0.0.1:4317"       # VPN
```

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
utpilla: > With separate receivers, each would independently push traffic, and coordinating backpressure across them would be harder (or require engine-level support).

There's no additional engine-level support required. Each receiver instance independently enforces its own `max_concurrent_requests` cap.

> The unified backpressure is actually an advantage of the single-instance approach, not just added complexity.

Dynamic capacity sharing is nice to have, but do we need it? The original design in #1144 had independent limits per protocol.

@jmacd Since you mentioned:

> In the Golang environment, I know that it is possible to share ports between gRPC and HTTP/2, **however it is not convenient or well supported**.

Do we even want parity for this particular feature?

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
ThomsonTan: Yes, there seems issue on just check the port number. Need some more check like below.

```rust
if grpc.listening_addr.port() == http.listening_addr.port() {
    let g_ip = grpc.listening_addr.ip();
    let h_ip = http.listening_addr.ip();
    if g_ip.is_unspecified() || h_ip.is_unspecified() || g_ip == h_ip {
        return Err(InvalidUserConfig { /* ... */ });
    }
}
```

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
lalitb: > There's no additional engine-level support required. Each receiver instance independently enforces its own max_concurrent_requests cap.

You're right - each receiver can independently enforces its own cap. However, the distinction is about total load on the downstream pipeline. With separate receivers  each capped at N, the pipeline could see up to 2N concurrent requests. The single-instance global semaphore is automatically derived from the downstream  `pdata_channel_size` - ensuring combined HTTP + gRPC traffic doesn't exceed what the pipeline can absorb, without needing to manually split limits between protocols. 

> Dynamic capacity sharing is nice to have, but do we need it? The original design in https://github.com/open-telemetry/otel-arrow/issues/1144 had independent limits per protocol.

Actually, the original design had a `single max_concurrent_requests` at the top level - not independent limits per protocol. This PR introduces per-protocol limits under protocols.grpc and protocols.http, plus the global semaphore for combined backpressure when both are enabled.

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
lalitb: Cool, have added this now.

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
lalitb: To clarify - shared backpressure isn't new in this PR. The original design already had it - when HTTP was enabled, both protocols shared a single semaphore pool capped at `downstream_capacity`. This PR preserves that existing behavior while restructuring the config format. If there are concerns about this  design choice, happy to discuss separately, but it shouldn't block this PR since we're maintaining existing semantics.

lalitb: > In the Golang environment, I know that it is possible to share ports between gRPC and HTTP/2, however it is not convenient or well supported.

  >> Do we even want parity for this particular feature?

Just to be clear -  Go collector doesn't support the sharing of ports. It's though possible in broader Go ecosystem (e.g., using `cmux`),

utpilla: > shared backpressure isn't new in this PR.

Ah, I see. I wasn't up to date with the OTLP receiver developments.

jmacd:
> Do we even want parity for this particular feature?

Mixed feelings. My old company ran a shared port configuration for gRPC and HTTP and it was always a bit of a source of confusion. It can work if we have a reason to make it work. The Collector does not do this.

rust/otap-dataflow/crates/otap/src/otlp_receiver.rs
```diff
@@ -67,24 +66,101 @@ pub const OTLP_RECEIVER_URN: &str = "urn:otel:otlp:receiver";
...
 #[serde(deny_unknown_fields)]
 pub struct Config {
-    /// Shared gRPC server settings reused across gRPC-based receivers.
-    #[serde(flatten)]
-    pub grpc: GrpcServerSettings,
+    /// Protocol configurations.
+    ///
+    /// At least one protocol (gRPC or HTTP) must be configured.
+    /// Each protocol can have its own TLS configuration.
+    pub protocols: Protocols,
```
ThomsonTan, line 126: Could we provide a backward compatible change to the config, like parse the previous config as `Protocols.grpc`?

lalitb: Since we're not stable yet, I think breaking changes are acceptable here. I'd prefer to keep things simple and support just one config format going forward.

Issue #1853, opened 2026-01-21
Perf test regression caused by `decode_transport_optimized_ids`


albertlockett:
We introduced a fix for the issue #966 which was to eagerly materialize the transport encoded parent IDs in the attribute procesor's `transform_attributes` util function. This caused a noticable performance regression in our benchmarks:

<img width="1209" height="682" alt="Image" src="https://github.com/user-attachments/assets/08ac4cca-62be-4f01-8bf4-60dc8db1f485" />

Three approaches need to be taken to tackle this:
- be more selective about when to call this function. Technically, we don't need to eagerly call it when renaming attributes. We only need it when renaming attributes would join a sequence of delta encoded attributes, and even when this happens, we might just be able to adjust the parent_id encoded values directly
- optimize this `decode_transport_optimized_ids` function, as there are some cases where we really do need to call this, and we want it to be as optimal as possible

PR #1927, opened 2026-01-30
Performance improvement for adding transport optimized encoding

albertlockett:
# Change Summary

- Optimizes the implementation of applying transport optimized encoding.
- Renames `materialize_parent_id` bench to `transport_optimize` as this now contains benchmarks that do both encoding & decoding

**Benchmark summary:**

| Benchmark | Size | Nulls | Before (µs) | After (µs) | Speedup | Improvement |
|-----------|------|-------|-------------|------------|---------|-------------|
| encode_transport_optimized_ids | 127 | No | 48.037 | 16.298 | 2.95x | 66.1% faster |
| encode_transport_optimized_ids | 127 | Yes | 47.768 | 18.446 | 2.59x | 61.4% faster |
| encode_transport_optimized_ids | 1536 | No | 518.36 | 98.955 | 5.24x | 80.9% faster |
| encode_transport_optimized_ids | 1536 | Yes | 520.94 | 107.01 | 4.87x | 79.5% faster |
| encode_transport_optimized_ids | 8096 | No | 3418.3 | 508.92 | 6.72x | 85.1% faster |
| encode_transport_optimized_ids | 8096 | Yes | 3359.5 | 545.16 | 6.16x | 83.8% faster |

Nulls* column above signifies there were null rows in the attribute values column. Ordinarily we wouldn't encode attributes like this in OTAP because it we'd use the AttributeValuesType::Empty value in the type column, but we handle it because it is valid arrow data since the columns are nullable.

**Context:** 
when fixing #966 we added code to eagerly remove the transport optimized encoding from when transforming attributes, and noticed a significant regression in the performance benchmarks, especially on OTAP-ATTR-OTAP scenario because we do a round trip decode/encode of the transport optimized encoding.

**Changes**

This PR specifically focuses on optimizing adding the transport optimized encoding for attributes, as this is where all the time was being spent. Adding this encoding involves sorting the attribute record batch by type, key, value, then parent_id, and adding delta encoding to the parent_id column for sequences where type, key and value are all equal to the previous row (unless value is null, or the type is Map or Slice).

Before this change, we were doing this sorting using arrow's `RowConverter`. We'd then do a second pass over the dataset to find sequences where type/key/value were equal, and apply the delta encoding to the parent_id column.

Although using the `RowConverter` is sometimes [an efficient way to sort multiple columns](https://arrow.apache.org/blog/2022/11/07/multi-column-sorts-in-arrow-rust-part-2/), it's notable that the `RowConverter` actually expands the dictionaries for all the columns before it sorts (see https://github.com/apache/arrow-rs/issues/4811). This is extremely expensive for us since most of our attribute columns are dictionary encoded.

This PR changes the implementation to sort the attributes record batch directly, starting by combining type & key together (using the sorted dictionary values from the keys column), then sorting this hybrid column. It then partitions the type column to identify the attributes value column for this segment of the sorted result, and partitions the key column to find segments of the value column to sort together. For each segment, it sorts it, appends it to a builder for the new values column. It then partitions the sorted segment of values and for each segment takes the parent_ids for the value segment, sorts them, adds delta encoding, and appends these to a buffer containing the encoded parent IDs. Then it combines everything together and produces the result.

The advantages of this approach are a) it's a lot faster and b) we build up enough state during the sorting that we don't need to do a second pass over the `RecordBatch` to add delta encoding.

There are quite a few transformations that happen, and I tried to do these as efficiently as possible. This means working with arrow's buffers directly in many places, instead of always using immutable `Array`s and compute kernels, which reduces quite a lot the amount of allocations.

**Future Work/Followups**
There are some code paths I didn't spent a lot of time optimizing:
- If the parent_id is a u32 which may be dictionary encoded, we simply cast it to a primitive array and then cast it back into a dict when we're done. I did some quick testing and figure this adds ~10% overhead. 
- If the value type is something that could be in a dictionary (string, int, bytes, & ser columns), but isn't dictionary encoded, or if the type is boolean, the way we build up the result column allocates many small arrays. This could be improved
- If the key column is not dictionary encoded. I didn't spend very much time optimizing this.

There's also probably some methods that we were using before to encode the ID column that I need to go back and delete

Related to #1853 

Existing unit tests plus new ones

No

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
@@ -1110,6 +1110,7 @@ impl AttrValuesSorter {
                 // if there were any nulls in the original values column, fill in any null segment
                 // flags in the ranges result
                 if let Some(nulls) = &nulls {
+                    let nulls = nulls.slice(range.start, range.len());
                     for nullable_range in result {
                         let (start_idx, _) = self.float64_sort[nullable_range.range.start];
                         nullable_range.is_null = nulls.is_null(start_idx)
@@ -1197,6 +1198,7 @@ impl AttrValuesSorter {
                 // if there were any nulls in the original values column, fill in any null segment/
                 // flags in the ranges result
                 if let Some(nulls) = &keys_and_ranks.nulls {
+                    let nulls = nulls.slice(range.start, range.len());
                     for nullable_range in result {
                         let (start_idx, _) = self.rank_sort[nullable_range.range.start];
                         nullable_range.is_null = nulls.is_null(start_idx)
@@ -2540,7 +2542,7 @@ mod test {
             ])),
             vec![
                 Arc::new(UInt16Array::from_iter_values([
-                    0, 1, 2, 3, 4, 5, 6, 4, 7, 8, 9, 10, 11,
+                    0, 1, 2, 3, 4, 5, 6, 4, 7, 8, 9, 10, 11, 12, 13,
                 ])),
                 Arc::new(UInt8Array::from_iter_values([
                     AttributeValueType::Str as u8,
@@ -2554,11 +2556,13 @@ mod test {
                     AttributeValueType::Double as u8,
                     AttributeValueType::Str as u8,
                     AttributeValueType::Str as u8,
+                    AttributeValueType::Str as u8,
+                    AttributeValueType::Double as u8,
                     AttributeValueType::Double as u8,
                     AttributeValueType::Double as u8,
                 ])),
                 Arc::new(DictionaryArray::new(
-                    UInt8Array::from_iter_values([0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1]),
+                    UInt8Array::from_iter_values([0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1]),
                     Arc::new(StringArray::from_iter_values(["ka", "kb"])),
                 )),
                 Arc::new(DictionaryArray::new(
@@ -2573,7 +2577,9 @@ mod test {
                         None, // null str attr (dict encoded)
                         None,
                         None, // null str attr (dict encoded) with different key
+                        None, // null str attr (dict encoded) with different key
                         Some(1),
+                        None,
                         None,
                         None,
                     ]),
@@ -2591,6 +2597,8 @@ mod test {
                     None, // null float attr (not dict encoded)
                     None,
                     None,
+                    None,
+                    None, // null float attr (not dict encoded) with different key
                     None, // null float attr (not dict encoded) with different key
                     Some(2.0),
                 ])),
@@ -2617,7 +2625,7 @@ mod test {
             ])),
             vec![
                 Arc::new(UInt16Array::from_iter_values([
-                    0, 1, 4, 6, 4, 9, 8, 5, 2, 3, 7, 11, 10,
+                    0, 1, 4, 6, 4, 10, 8, 9, 5, 2, 3, 7, 13, 11, 12,
                 ])),
                 Arc::new(UInt8Array::from_iter_values([
                     AttributeValueType::Str as u8,
@@ -2627,6 +2635,8 @@ mod test {
                     AttributeValueType::Str as u8,
                     AttributeValueType::Str as u8,
                     AttributeValueType::Str as u8,
+                    AttributeValueType::Str as u8,
+                    AttributeValueType::Double as u8,
                     AttributeValueType::Double as u8,
                     AttributeValueType::Double as u8,
                     AttributeValueType::Double as u8,
@@ -2635,7 +2645,7 @@ mod test {
                     AttributeValueType::Double as u8,
                 ])),
                 Arc::new(DictionaryArray::new(
-                    UInt8Array::from_iter_values([0, 0, 0, 0, 0, 1, 1, 0, 0, 0, 0, 1, 1]),
+                    UInt8Array::from_iter_values([0, 0, 0, 0, 0, 1, 1, 1, 0, 0, 0, 0, 1, 1, 1]),
                     Arc::new(StringArray::from_iter_values(["ka", "kb"])),
                 )),
                 Arc::new(DictionaryArray::new(
@@ -2647,6 +2657,8 @@ mod test {
                         None, // null str attr (dict encoded)
                         Some(1),
                         None, // null str attr (dict encoded) with different key
+                        None, // null str attr (dict encoded) with different key
+                        None,
                         None,
                         None,
                         None,
@@ -2664,11 +2676,13 @@ mod test {
                     None,
                     None,
                     None,
+                    None,
                     Some(1.5),
                     Some(2.0),
                     None, // null float attr (not dict encoded)
                     None, // null float attr (not dict encoded)
                     Some(2.0),
+                    None, // null float attr (not dict encoded) with different key
                     None, // null float attr (not dict encoded) with different key
                 ])),
             ],


albertlockett:
Benchmark results

before
```
encode_transport_optimized/encode_transport_optimized_ids_no_nulls/127
                        time:   [47.863 µs 48.037 µs 48.214 µs]

encode_transport_optimized/encode_transport_optimized_ids_with_nulls/127
                        time:   [47.625 µs 47.768 µs 47.948 µs]

File /tmp/input_1536.proto already exists, skipping generation
encode_transport_optimized/encode_transport_optimized_ids_no_nulls/1536
                        time:   [515.30 µs 518.36 µs 521.43 µs]

encode_transport_optimized/encode_transport_optimized_ids_with_nulls/1536
                        time:   [518.30 µs 520.94 µs 523.42 µs]

encode_transport_optimized/encode_transport_optimized_ids_no_nulls/8096
                        time:   [3.4128 ms 3.4183 ms 3.4250 ms]

encode_transport_optimized/encode_transport_optimized_ids_with_nulls/8096
                        time:   [3.3545 ms 3.3595 ms 3.3654 ms]
```

after
```
encode_transport_optimized/encode_transport_optimized_ids_no_nulls/127
                        time:   [16.246 µs 16.298 µs 16.352 µs]

encode_transport_optimized/encode_transport_optimized_ids_with_nulls/127
                        time:   [18.415 µs 18.446 µs 18.482 µs]

encode_transport_optimized/encode_transport_optimized_ids_no_nulls/1536
                        time:   [98.761 µs 98.955 µs 99.161 µs]

encode_transport_optimized/encode_transport_optimized_ids_with_nulls/1536
                        time:   [106.83 µs 107.01 µs 107.18 µs]

encode_transport_optimized/encode_transport_optimized_ids_no_nulls/8096
                        time:   [508.02 µs 508.92 µs 510.11 µs]


encode_transport_optimized/encode_transport_optimized_ids_with_nulls/8096
                        time:   [544.07 µs 545.16 µs 546.34 µs]
```

albertlockett:
CPU benchmarks show improvement for scenarios that use OTAP exporter

https://github.com/open-telemetry/otel-arrow/actions/runs/21532929307/job/62052656550?pr=1927

```
| name                          | unit      |            value | extra                                            |
|:------------------------------|:----------|-----------------:|:-------------------------------------------------|
| dropped_logs_percentage       | %         |     -1.4         | CI 100kLRPS/OTLP-ATTR-OTAP - Dropped Logs %      |
| cpu_percentage_normalized_avg | %         |     29.3419      | CI 100kLRPS/OTLP-ATTR-OTAP - CPU % (Normalized)  |
| cpu_percentage_normalized_max | %         |     30.2025      | CI 100kLRPS/OTLP-ATTR-OTAP - CPU % (Normalized)  |
| ram_mib_avg                   | MiB       |     40.6336      | CI 100kLRPS/OTLP-ATTR-OTAP - RAM (MiB)           |
| ram_mib_max                   | MiB       |     42.4766      | CI 100kLRPS/OTLP-ATTR-OTAP - RAM (MiB)           |
| logs_produced_rate            | logs/sec  | 124996           | CI 100kLRPS/OTLP-ATTR-OTAP - Log Throughput      |
| logs_received_rate            | logs/sec  | 126746           | CI 100kLRPS/OTLP-ATTR-OTAP - Log Throughput      |
| test_duration                 | seconds   |     20.0007      | CI 100kLRPS/OTLP-ATTR-OTAP - Test Duration       |
| network_tx_bytes_rate_avg     | bytes/sec | 566187           | CI 100kLRPS/OTLP-ATTR-OTAP - Network Utilization |
| network_rx_bytes_rate_avg     | bytes/sec |      2.23792e+06 | CI 100kLRPS/OTLP-ATTR-OTAP - Network Utilization |

  
| name                          | unit      |        value | extra                                            |
|:------------------------------|:----------|-------------:|:-------------------------------------------------|
| dropped_logs_percentage       | %         |     -4.16667 | CI 100kLRPS/OTAP-ATTR-OTAP - Dropped Logs %      |
| cpu_percentage_normalized_avg | %         |     19.2     | CI 100kLRPS/OTAP-ATTR-OTAP - CPU % (Normalized)  |
| cpu_percentage_normalized_max | %         |     21.865   | CI 100kLRPS/OTAP-ATTR-OTAP - CPU % (Normalized)  |
| ram_mib_avg                   | MiB       |     44.0008  | CI 100kLRPS/OTAP-ATTR-OTAP - RAM (MiB)           |
| ram_mib_max                   | MiB       |     46.6328  | CI 100kLRPS/OTAP-ATTR-OTAP - RAM (MiB)           |
| logs_produced_rate            | logs/sec  | 119996       | CI 100kLRPS/OTAP-ATTR-OTAP - Log Throughput      |
| logs_received_rate            | logs/sec  | 124996       | CI 100kLRPS/OTAP-ATTR-OTAP - Log Throughput      |
| test_duration                 | seconds   |     20.0007  | CI 100kLRPS/OTAP-ATTR-OTAP - Test Duration       |
| network_tx_bytes_rate_avg     | bytes/sec | 549549       | CI 100kLRPS/OTAP-ATTR-OTAP - Network Utilization |
| network_rx_bytes_rate_avg     | bytes/sec | 518852       | CI 100kLRPS/OTAP-ATTR-OTAP - Network Utilization |
```

Baseline:
image

rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
```diff
@@ -0,0 +1,2812 @@
...
+                .take()
+                .expect("serialized attr column is present");
+            columns.push(sorted_col.finish()?);
+            continue;
+        }
+
+        if field_name == consts::PARENT_ID {
+            columns.push(parent_id_col.clone());
+            continue;
+        }
```
jmacd: duplicate case?

rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
```diff
@@ -0,0 +1,2812 @@
...
+    }
+
+    // rebuild the record batch with all the sorted/encoded columns ...
+    let mut fields = vec![];
+    let mut columns = vec![];
+    for field in record_batch.schema().fields() {
+        let field_name = field.name();
+
+        if field.name() == consts::PARENT_ID {
```
jmacd:
```suggestion
        if field_name == consts::PARENT_ID {
```

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -435,12 +435,6 @@ where
                         batch_idx += 1;
                     }
                 }
-
-                // handle any remaining non-delta values after last delta range ...
-                // just read the last value if there are any remaining
-                if batch_idx <= eq_range_end {
-                    curr_parent_id = materialized_parent_ids[eq_range_end];
-                }
```
jmacd, line 443: Nice.

rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
albertlockett: nice catch, fixed in 45de30e0d. Thanks @jmacd!

rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
```diff
@@ -0,0 +1,2809 @@
...
+    // finalize the new parent_id column
+    let mut parent_id_col = Arc::new(PrimitiveArray::<T>::new(
+        ScalarBuffer::from(encoded_parent_id_column),
+        None,
+    )) as ArrayRef;
+
+    // If the original Parent ID type was a dictionary, cast it back to a dictionary of this type.
+    if let Some(dict_key_type) = parent_dict_key_type {
+        parent_id_col = match cast(
+            &parent_id_col.clone(),
```
utpilla:
```suggestion
            &parent_id_col,
```

rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
```diff
@@ -0,0 +1,2809 @@
...
+    ) -> Result<()> {
+        match &self.inner {
+            AttrsValueSorterInner::Float64((floats, nulls)) => {
+                let slice = floats.slice(range.start, range.len());
+                self.float64_sort.clear();
+                self.float64_sort.extend(slice.iter().copied().enumerate());
+
+                if let Some(nulls) = nulls {
+                    self.float64_sort.sort_unstable_by(|a, b| {
+                        match (nulls.is_null(a.0), nulls.is_null(b.0)) {
```
utpilla: Shouldn't we add `range.start` when indexing into `nulls` which is the shared buffer for the entire original range for the type?  Since you use enumerate the input range here `self.float64_sort.extend(slice.iter().copied().enumerate());`, `a.0` and `b.0` would always be `0` which would lead to checking the wrong item in `nulls`?

```suggestion
                        match (nulls.is_null(range.start + a.0), nulls.is_null(range.start + b.0)) {
```

albertlockett: Great catch @utpilla thank you! 

fixed in b887eb8dbb29b023f4e6c1d395b01f0ef1ae9fbf with updated tests to cover this

rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
```diff
@@ -0,0 +1,2843 @@
...
+                    range,
+                    is_null: false,
+                }));
+
+                // if there were any nulls in the original values column, fill in any null segment
+                // flags in the ranges result
+                if let Some(nulls) = &nulls {
+                    for nullable_range in result {
+                        let (start_idx, _) = self.float64_sort[nullable_range.range.start];
+                        nullable_range.is_null = nulls.is_null(start_idx)
```
utpilla, line 1116: This also seems to be problematic. `nulls` is the original unsliced buffer but `start_idx` is extracted from `self.float64_sort` which is slice relative.

rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
```diff
@@ -0,0 +1,2843 @@
...
+                    range,
+                    is_null: false,
+                }));
+
+                // if there were any nulls in the original values column, fill in any null segment/
+                // flags in the ranges result
+                if let Some(nulls) = &keys_and_ranks.nulls {
+                    for nullable_range in result {
+                        let (start_idx, _) = self.rank_sort[nullable_range.range.start];
+                        nullable_range.is_null = nulls.is_null(start_idx)
```
utpilla, line 1204: Same as https://github.com/open-telemetry/otel-arrow/pull/1927#discussion_r2756446448

rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
albertlockett: thanks, you're 100% right @utpilla . Fixed in e160e3cb7

rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
albertlockett: fixed, https://github.com/open-telemetry/otel-arrow/pull/1927#discussion_r2756489579

PR #1928, opened 2026-01-31
[otap-df-otap] Add TLS support for Syslog CEF Receiver

### LINKED ISSUE #1260: TLS support for Syslog
Stretch goal?

utpilla:
# Change Summary

- Add TLS support for Syslog/CEF over TCP

* Closes #1260 

- Only through some unit tests targeting TLS functionality for now
- Need to add integration tests

- Receiver config now allows for TLS settings

### DIFF
--- a/.github/workflows/dataflow-engine-binary-size.yml
+++ b/.github/workflows/dataflow-engine-binary-size.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ${{ matrix.platform.runner }}
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
+        uses: step-security/harden-runner@e3f713f2d8f53843e71c69a996d56f51aa9adfb9 # v2.14.1
         with:
           egress-policy: audit
 
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -29,9 +29,9 @@ jobs:
       - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
         with:
           submodules: true
-      - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c # v6.1.0
+      - uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
         with:
-          go-version: "1.25.5"
+          go-version: "1.25.6"
       - name: Run tests with coverage
         run: go test -cover -coverprofile=coverage.txt -covermode=atomic -coverpkg=./... ./...
         working-directory: ./go/${{ matrix.folder }}
@@ -58,11 +58,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
-    - uses: actions/setup-go@4dc6199c7b1a012772edbd06daecab0f50c9053c # v6.1.0
+    - uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
       with:
-        go-version: "1.25.5"
-    - uses: github/codeql-action/init@5d4e8d1aca955e8d8589aabd499c5cae939e33c7 # v4.31.9
+        go-version: "1.25.6"
+    - uses: github/codeql-action/init@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
       with:
         languages: go
-    - uses: github/codeql-action/analyze@5d4e8d1aca955e8d8589aabd499c5cae939e33c7 # v4.31.9
+    - uses: github/codeql-action/analyze@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
       timeout-minutes: 60
--- a/.github/workflows/ossf-scorecard.yml
+++ b/.github/workflows/ossf-scorecard.yml
@@ -42,6 +42,6 @@ jobs:
       # Upload the results to GitHub's code scanning dashboard (optional).
       # Commenting out will disable upload of results to your repo's Code Scanning dashboard
       - name: "Upload to code-scanning"
-        uses: github/codeql-action/upload-sarif@5d4e8d1aca955e8d8589aabd499c5cae939e33c7 # v4.31.9
+        uses: github/codeql-action/upload-sarif@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
         with:
           sarif_file: results.sarif
--- a/.github/workflows/pipeline-perf-on-label.yaml
+++ b/.github/workflows/pipeline-perf-on-label.yaml
@@ -44,7 +44,7 @@ jobs:
     runs-on: oracle-bare-metal-64cpu-512gb-x86-64
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
+        uses: step-security/harden-runner@e3f713f2d8f53843e71c69a996d56f51aa9adfb9 # v2.14.1
         with:
           egress-policy: audit
 
@@ -124,7 +124,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@20cf305ff2072d973412fa9b1e3a4f227bda3c76 # v2.14.0
+        uses: step-security/harden-runner@e3f713f2d8f53843e71c69a996d56f51aa9adfb9 # v2.14.1
         with:
           egress-policy: audit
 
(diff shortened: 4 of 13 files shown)


rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -138,13 +168,36 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
 
         // Start periodic telemetry collection (1s), similar to other nodes
         let timer_cancel_handle = effect_handler
-            .start_periodic_telemetry(std::time::Duration::from_secs(1))
+            .start_periodic_telemetry(Duration::from_secs(1))
```
ThomsonTan, line 158: Emit a warning or error for UDP protocol but still has TLS enabled? Because TLS is for TCP only.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -68,8 +83,23 @@ impl Config {
...
+    ///
+    /// The `TlsServerConfig.handshake_timeout` field has a serde default of 10 seconds,
+    /// so this will always return a value when TLS is configured.
+    #[cfg(feature = "experimental-tls")]
+    fn handshake_timeout(&self) -> Duration {
+        self.tls
+            .as_ref()
+            .expect("handshake_timeout called without TLS config")
+            .handshake_timeout
+            .expect("handshake_timeout should have serde default")
```
ThomsonTan: Avoid invoking `.expect()` which could panic on production?

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -1583,6 +1635,119 @@ mod tests {
...
+    /// 2. `ExtensionValueInvalid` error on macOS when adding non-CA certs to root stores
+    ///
+    /// # Arguments
+    /// * `dir` - Directory to write certificate files
+    /// * `ca_name` - Base name for CA files (creates `{ca_name}.crt` and `{ca_name}.key`)
+    /// * `cert_name` - Base name for end-entity cert files
+    /// * `cn` - Common Name for the end-entity cert (also used as SAN for hostname verification)
+    fn generate_ca_signed_cert(dir: &Path, ca_name: &str, cert_name: &str, cn: &str) {
+        // Step 1: Generate CA key and self-signed CA cert
+        let output = Command::new("openssl")
```
ThomsonTan: Probably use rcgen which is already in our dependence list, so no need to install openssl to run this test.

https://github.com/utpilla/otel-arrow/blob/utpilla/Add-TLS-support-for-Syslog-CEF-Receiver/rust/otap-dataflow/crates/otap/tests/otlp_exporter_proxy_tls.rs#L21

lalitb: We have similar `generate_cert` helper in `mtls_tests.rs` that uses `rcgen` instead of OpenSSL. We could potentially reuse that logic here (perhaps with updates for the CA-signed-leaf scenario) to avoid the extra dependency? If the OpenSSL approach is necessary for specific validation cases, feel free to keep it as is.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -59,6 +68,12 @@ struct Config {
     listening_addr: SocketAddr,
     /// The protocol to use for receiving messages
     protocol: Protocol,
+    /// TLS configuration for secure TCP connections (Syslog over TLS, RFC 5425).
+    ///
+    /// When configured, TCP connections will require TLS. This is only applicable
+    /// when `protocol` is `tcp`. UDP does not support TLS (use DTLS separately if needed).
```
lalitb: With "use DTLS separately if needed", do we mean deploying an external/sidecar DTLS-terminating proxy in front of the receiver? As written, it could read like DTLS is supported out-of-the-box here (without code changes), which I don’t think is the case.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -173,18 +226,60 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+                                    let tls_acceptor = maybe_tls_acceptor.clone();
+                                    #[cfg(feature = "experimental-tls")]
+                                    let tls_handshake_timeout = maybe_handshake_timeout;
+
                                     // Spawn a task to handle the connection.
                                     // ToDo should this be abstracted and exposed a method in the effect handler?
                                     _ = tokio::task::spawn_local(async move {
+                                        // Perform TLS handshake if configured, creating a unified reader type
+                                        #[cfg(feature = "experimental-tls")]
+                                        let mut reader: Box<dyn tokio::io::AsyncBufRead + Unpin> = if let Some(acceptor) = tls_acceptor {
```
lalitb, line 239: Nit (non-blocking): The Box<dyn AsyncBufRead> adds a heap allocation per connection. An enum wrapper could avoid this, but given the allocation is once per connection (not per message) and TLS handshake costs dominate, this is likely negligible. Worth revisiting if benchmarks show otherwise.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -1479,6 +1479,58 @@ pub async fn build_tls_acceptor(
...
+    acceptor: &tokio_rustls::TlsAcceptor,
+    handshake_timeout: Duration,
+) -> Result<tokio_rustls::server::TlsStream<T>, io::Error>
+where
+    T: tokio::io::AsyncRead + tokio::io::AsyncWrite + Unpin,
+{
+    match tokio::time::timeout(handshake_timeout, acceptor.accept(stream)).await {
+        Ok(Ok(tls_stream)) => Ok(tls_stream),
+        Ok(Err(e)) => {
+            otel_warn!("TLS handshake failed", error = ?e);
```
lalitb: Minor: TLS handshake failures log twice (WARN as here, and then DEBUG in receiver with peer context). This is fine, but could consolidate if the dual logging feels noisy.

lalitb (approved):
LGTM. Nice work! The implementation cleanly integrates with the existing TLS infrastructure. Have some nit comments, nothing blocking as such.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
utpilla: As you mentioned, the allocation is only once per connection and that cost would be negligible compared to TCP/TLS handshake costs. While an `enum` wrapper would eliminate the allocation, it would require manually implementing `AsyncRead` and `AsyncBufRead` for each variant, adding maintenance overhead for minimal gain.

Since network I/O costs dwarf a single allocation, I prefer `Box<dyn>` for simplicity.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
utpilla: The existing tests in `tls_utils.rs` were already using OpenSSL CLI, which is why I initially went with the same approach for consistency. I have changed the tests added in this PR to use `rcgen`.

To keep this PR scoped, I'd like to defer the following to a follow-up PR:

- Migrating the other existing tests in `tls_utils.rs` from OpenSSL to `rcgen`
- Consolidating test helpers into a shared module that both `tls_utils.rs` and `mtls_tests.rs` can use

How does that sound?

rust/otap-dataflow/crates/otap/src/tls_utils.rs
utpilla: Good catch! I have removed `otel_warn!` statements from `accept_tls_connection` method and letting the caller decide the log level that they want to use at the call site. I have updated the usage of this method in` syslog_cef_receiver.rs` to use `otel_warn!` instead of `otel_debug!`.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
utpilla: Good catch. I have removed that from the comment.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
utpilla: I have removed this method to simply extract the timeout value from the config.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
lalitb: Yes agree, this can be handled separate of this PR.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
utpilla: Thanks for raising this concern! I think there a few approaches that we could follow here:
1. Emit a warning and ignore the TLS config when configured with UDP
2. Validate at config parsing time and return an error in `from_config()` method
3. Enforce at type level by using enum like `Protocol::Tcp { tls: Option<TlsServerConfig> } | Protocol::Udp`. This would make the invalid states unrepresentable at config level.

I like Option 3, but it would bring some nesting into that YAML config (which I guess is fine?). It would look like this for

UDP
```yaml
listening_addr: "127.0.0.1:514"
protocol: udp
```

TCP without TLS:
```yaml
listening_addr: "127.0.0.1:514"
protocol:
  tcp: {}
```

TCP with TLS:
```yaml
listening_addr: "127.0.0.1:6514"
protocol:
  tcp:
    tls:
      cert_file: /path/to/cert.pem
      key_file: /path/to/key.pem
```

@jmacd @lquerel I would like to hear your thoughts as well on this one. The core problem here is of conditional configuration which is a general issue and could be faced by other nodes as well.  A set of configuration options are only valid for a specific value of another configuration. Which one of the three approaches do we want to use to solve this?

PR #1937, opened 2026-02-02
fix(tests): Switch from assert!(result.is_ok()) => result.unwrap() for CI diagnosability

AaronRM:
# Change Summary

Switch pattern from `assert!(result.is_ok())` to `result.unwrap()` in exporter tests. This is to improve diagnostics for flakey tests in CI. Currently, failures output the following which is not actionable:

```
    thread 'parquet_exporter::test::test_traces' (2500) panicked at crates\otap\src\parquet_exporter.rs:1299:21:
    assertion failed: exporter_result.is_ok()
```

With the change above, the error string from the result will be properly logged.

n/a

n/a

No

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,9 @@ We value all contributions, whether big or small, and encourage you to join us
 in improving this project. If you have questions, don't hesitate to reach out to
 the OpenTelemetry community - we're here to help!
 
+Looking for a place to start? See the repository's
+[Contribute page](https://github.com/open-telemetry/otel-arrow/contribute).
+
 ## Pre-requisites
 
 To work with this repository, you'll need:
--- a/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
+++ b/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
@@ -61,7 +61,7 @@ enum Protocol {
     Udp,
 }
 
-/// config for a syslog cef receiver
+/// Config for a syslog cef receiver
 #[derive(Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
 struct Config {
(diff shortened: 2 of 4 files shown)


albertlockett (approved):
seems reasonable to me. Thanks @AaronRM

rust/otap-dataflow/crates/otap/src/parquet_exporter.rs
```diff
@@ -577,7 +577,7 @@ mod test {
             })
             .run_validation(move |_ctx, exporter_result| {
                 Box::pin(async move {
-                    assert!(exporter_result.is_ok());
+                    exporter_result.expect("Exporter should succeed");
```
utpilla: Looking at all usages of `ValidationPhase::run_validation`, every caller asserts/expects the result to be `Ok`.

It would be better to simply move this check (expect the result) to `run_validation` method itself: https://github.com/open-telemetry/otel-arrow/blob/27476382aa19c9a7610aa1ab54e6a54bc774343b/rust/otap-dataflow/crates/engine/src/testing/exporter.rs#L346-L348

That would also simplify the `FnOnce` trait signature and remove some boilerplate code from the callsites.

The call sites would change from:
```rust
.run_validation(|ctx, result| async move {
    result.expect("success");
    // actual validation...
})
```

to 

```rust
.run_validation(|ctx| async move {
    // actual validation...
})
```

albertlockett: Yeah I think this sounds reasonable.

The original idea behind returning the result was that we may want to assert in our tests that the exporter returned an error. The parquet exporter used to have a test that asserted that the exporter returned an error if it wasn't able to flush its writers during the shutdown timeout. But that test has been refactored, and now it looks like we're no longer making assertions on the error anywhere in our code

utpilla: Oh okay. In that case, maybe we should let it be. It might help with some test in the future.

PR #1942, opened 2026-02-03
PerfTest - tweak idle state test to confirm memory growth pattern

cijothomas:
# Change Summary

Modified the Idle State Test to run on 1/2/4/8/16/32 cores and confirm if the memory growth (idle state) is predictable.

Part of the comment https://github.com/open-telemetry/otel-arrow/pull/1528/changes#r2710193083

Ran locally.

No

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,9 @@ We value all contributions, whether big or small, and encourage you to join us
 in improving this project. If you have questions, don't hesitate to reach out to
 the OpenTelemetry community - we're here to help!
 
+Looking for a place to start? See the repository's
+[Contribute page](https://github.com/open-telemetry/otel-arrow/contribute).
+
 ## Pre-requisites
 
 To work with this repository, you'll need:


cijothomas:
After validation for couple days, will move this to nightly. Not much value running this in every commit.

.github/workflows/scripts/analyze-idle-state-scaling.py
```diff
@@ -0,0 +1,314 @@
...
+    benchmark_data = generate_benchmark_json(memory_data, C, R, r_squared)
+    
+    if output_json_path:
+        with open(output_json_path, 'w') as f:
+            json.dump(benchmark_data, f, indent=2)
+        print(f"Benchmark JSON written to: {output_json_path}", file=sys.stderr)
+    
+    # Exit with warning if fit is poor
+    if r_squared < 0.85:
+        print("Warning: Poor linear fit detected! Memory scaling may not follow expected model.", file=sys.stderr)
```
ThomsonTan, line 310: Fail the CI with an non-zero exit code? The output to stderror might just be ignored.

cijothomas: Good point! 
So far, the perf tests don't cause CI to fail, even if perf regresses. I agree with your suggestion, that we need to do better here by failing CI itself. However, that is something I'll delay for sometime to make sure the CI perf tests are super stable, and all scenarios are valid. Happy to come back to this after few more days.

PR #1943, opened 2026-02-03
[otap-df-otap] Update Syslog CEF Receiver README

utpilla:
# Change Summary
- Update Syslog CEF Receiver README


rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
```diff
@@ -1,5 +1,372 @@
...
+  syslog_cef:
+    listening_addr: "0.0.0.0:514"
+    protocol: tcp  # or "udp"
+
+    # Optional: TLS configuration (TCP only, requires "experimental-tls" feature)
+    tls:
+      cert_file: "/path/to/server.crt"
+      key_file: "/path/to/server.key"
+      ca_file: "/path/to/ca.crt"           # Optional: client cert verification
+      handshake_timeout: "10s"              # Optional: default 10s
```
ThomsonTan: nit: align the comments with the comment in the line above?

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
```diff
@@ -1,5 +1,372 @@
...
+receivers:
+  syslog_cef:
+    listening_addr: "0.0.0.0:514"
+    protocol: tcp  # or "udp"
+
+    # Optional: TLS configuration (TCP only, requires "experimental-tls" feature)
+    tls:
+      cert_file: "/path/to/server.crt"
+      key_file: "/path/to/server.key"
+      ca_file: "/path/to/ca.crt"           # Optional: client cert verification
```
ThomsonTan: Should here be `client_ca_file` which is for TLS server?

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
```diff
@@ -1,5 +1,372 @@
...
+- **Simpler state management**: Metrics and buffers use `Rc<RefCell<...>>`
+  instead of `Arc<Mutex<...>>`
+- **Predictable performance**: No cross-thread coordination latency
+
+For TCP, each connection spawns a separate local task, allowing concurrent
+connection handling within the same thread.
+
+### Batching Strategy
+
+To optimize throughput and reduce per-message overhead, the receiver batches
```
cijothomas, line 306: wondering how does this compare with not batching in the receiver at all, and instead using batch_processor? batch_processor allows customization by user.

utpilla: The problem is that `batch_processor` already expects `OtapPdata`. It is not efficient to convert every single Syslog message received to an OTAP payload.

cijothomas: Thanks for clarifying. This will be another place where telemetry is in memory and hence not-durable, and hence lost in event of crash. If this becomes a problem, we can offer a opt-in switch to avoid this batching and accept the perf hit.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
utpilla: Thanks! I have fixed it.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
```diff
@@ -1,5 +1,372 @@
...
+| `cef.version` | `0` |
+| `cef.device_vendor` | `Security` |
+| `cef.device_product` | `IDS` |
+| `cef.device_version` | `1.0` |
+| `cef.device_event_class_id` | `100` |
+| `cef.name` | `Attack detected` |
+| `cef.severity` | `8` |
+| `src` | `10.0.0.1` |
+| `dst` | `192.168.1.1` |
+| `spt` | `12345` |
```
lalitb, line 258: From where is `spt` coming from - it's not there in example input.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
```diff
@@ -1,5 +1,372 @@
...
+
+> **Note:** RFC 3164 timestamps don't include the year, so the receiver
+> assumes the current year.
+
+### CEF Parsing
+
+CEF messages follow this structure:
+
+```text
+CEF:version|vendor|product|version|event_class_id|name|severity|extensions
```
lalitb: The `version` is twice here? 

EDIT - ok, the second one seems to device_version. Maybe it can be explicit.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
```diff
@@ -1,5 +1,372 @@
...
+    receivers: [syslog_cef]
+    processors: [batch]
+    exporters: [otlp]
+```
+
+## Feature Flags
+
+| Feature | Description |
+|---------|-------------|
+| `experimental-tls` | Enables TLS support for secure TCP connections |
```
lalitb, line 387: jfyi if we want to add that -  the feature flag is temporary, and would be eventually removed.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
```diff
@@ -1,5 +1,372 @@
...
+<priority>version timestamp hostname app-name procid msgid [structured-data] msg
+```
+
+**Example Input:**
+
+```text
+<34>1 2003-10-11T22:14:15.003Z mymachine.example.com su 12345 ID47
+[exampleSDID@32473 iut="3" eventSource="Application"]
+'su root' failed for lonvick on /dev/pts/8
+```
```
lalitb, line 109: nit - the example is spanning across multiple lines. We can either make it as single line, or add a comment that this is multi-line for readability..

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
```diff
@@ -1,5 +1,372 @@
 # Syslog and CEF Receiver
 
-Status: **Skeleton**
+A high-performance receiver for ingesting syslog messages (RFC 3164 and RFC 5424)
+and Common Event Format (CEF) security logs. The receiver automatically detects
+the message format and converts them into `OtapPdata` using the efficient
+Apache Arrow columnar format for downstream processing.
```
lalitb, line 8: nit - if it make sense, consider also mentioning the Plugin URN name `urn:otel:syslog_cef:receiver`.

lalitb:
Nice docs! Non-blocking suggestion for future: consider adding a brief "Operational Notes" or similar section for

   - Backpressure behavior (blocks on bounded downstream channel when slow)
   - In-memory batching durability caveat (as @cijothomas noted - up to 100 msgs / 100ms can be lost on crash)
   - Known TODOs from code: message size limit for TCP, or max connection limit

  Can be a follow-up PR. LGTM!

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
```diff
@@ -1,5 +1,372 @@
...
+```text
+CEF:0|Security|threatmanager|1.0|100|worm successfully stopped|10|
+src=10.0.0.1 dst=2.1.2.2 spt=1232
+```
+
+**Resulting LogRecord:**
+
+| Field | Value |
+|-------|-------|
+| `timestamp` | *(observed time)* |
```
lalitb: I believe in code `timestamp` is set to 0, and `observed timestamp` should be separate entry?

lalitb: Or maybe - Just make it explicit here:
```
   | `timestamp` | *(not available - falls back to observed_time)* |
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
```diff
@@ -1,5 +1,372 @@
...
+```
+
+**Resulting LogRecord:**
+
+| Field | Value |
+|-------|-------|
+| `timestamp` | `2003-10-11T22:14:15.003Z` |
+| `severity_number` | `18` (ERROR2 - maps from syslog severity 2/Critical) |
+| `severity_text` | `ERROR2` |
+| `body` | *(empty - fully parsed)* |
```
albertlockett: nit: should this be `null` instead of `empty`, just to avoid confusion where `empty` might be interpreted as meaning empty string?

```suggestion
| `body` | *(null - fully parsed)* |
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
```diff
@@ -1,5 +1,372 @@
...
+```
+
+**Resulting LogRecord:**
+
+| Field | Value |
+|-------|-------|
+| `timestamp` | `<current_year>-10-11T22:14:15` (local timezone) |
+| `severity_number` | `18` (ERROR2) |
+| `severity_text` | `ERROR2` |
+| `body` | *(empty - fully parsed)* |
```
albertlockett: nit: same question as https://github.com/open-telemetry/otel-arrow/pull/1943#discussion_r2765137319
```suggestion
| `body` | *(null - fully parsed)* |
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
jmacd: It would be good to document the potential for dataloss due to the buffer--I would call this "data in memory". Since Syslog doesn't have any Ack/Nack mechanism built in, there's not a great distinction between buffering in the component and buffering in the input queue of the next component. Users should put the durable processor next if they want to reduce this risk, otherwise-- I do not how to make Syslog more reliable.

The `internal_telemetry_receiver` component does not buffer, but if it did, then I would call it "telemetry in memory". There the use of batching should be considered dangerous for diagnostic capability.

utpilla: > Users should put the durable processor next if they want to reduce this risk,

I would make the batching related settings public so that the user can set the max batch size to `1`. Otherwise, there would still be some data batched before it makes it to durable buffer. Even then, we would still have scope for data loss as Syslog receiver doesn't respond back to clients.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/README.md
utpilla: I think I'll just wait for it to be stabilized and remove it from the README 😄

PR #1952, opened 2026-02-04
perf: avoid eagerly removing transport optimized encoding when transforming attributes

### LINKED ISSUE #966: `transform_attributes` could create an invalid sequence of `parent_ids` using `rename` with no deletes

### LINKED ISSUE #1853: Perf test regression caused by `decode_transport_optimized_ids`

albertlockett:
# Change Summary

To fix #966 we added a change to eargerly remove transport optimized encoding when transforming attributes, which led to the performance regression documented in #1853.

Removing this encoding is actually only necessary under somewhat rare conditions, at least for delete and renaming attributes. Basically, if the operation would join sequences where type/key/value or two adjacent rows are equal after the transformation, but were not equal before. For example if we had attribute columns like:
```
key | str val
--- | ------
 A  |  1    
 B  |  1        <-- if key "B" were renamed to "A"
 ...
 A  |  1    
 B  |  1        <--- if key "B" were deleted
 A  |  1    
```

This PR removes the eager decoding behaviour, and instead adds code to check if we've made a transform that produces such a sequence. As part of the transformation process, we already compute a sequence of ranges of the attribute keys column that will be renamed or deleted. We use these ranges to compute the neighbouring rows of each transformed range, and check that the post-transform type/key/value sequences are not equal. If we find any neighbours w/ equal values for these columns, we remove the transport optimized encoding.

Performing this check isn't free: For dictionary encoded keys, we compute the transformed ranges only for the dictionary values, so the ranges need to be mapped back to the equivalent ranges for the dictionary keys. Despite this it is faster to check than removing the transport optimized encoding. Moreover, if we remove the encoding and then have to re-apply it to use the OTAP exporter, performing this check becomes even more worthwhile. Also, once we've computed the input to perform this check, we're able to use the resulting ranges both to calculate statistics (for attribute transform metrics) and to speed up the computation of which rows were deleted.

Bench results:

num_rows | dict_keys | operation | mode | before_us | after_us | percent change
-- | -- | -- | -- | -- | -- | --
128 | FALSE | rename | decode=true | 1.8455 | 1.9344 | 4.60%
128 | FALSE | rename | decode=false | 1.8486 | 0.6966 | -62.32%
128 | FALSE | delete | decode=true | 2.8462 | 2.9733 | 4.27%
128 | FALSE | delete | decode=false | 2.8477 | 1.6827 | -40.91%
128 | FALSE | rename | no_encode | 0.84347 | 0.84786 | 0.52%
128 | FALSE | rename | no_encode+stat | 0.84413 | 0.84534 | 0.14%
128 | TRUE | rename | decode=true | 1.3191 | 1.3912 | 5.18%
128 | TRUE | rename | decode=false | 1.3286 | 0.4037 | -69.61%
128 | TRUE | delete | decode=true | 2.5701 | 2.5734 | 0.13%
128 | TRUE | delete | decode=false | 2.5649 | 1.5479 | -39.65%
128 | TRUE | rename | no_encode | 0.37642 | 0.29236 | -22.33%
128 | TRUE | rename | no_encode+stat | 0.37804 | 0.32997 | -12.72%
1536 | FALSE | rename | decode=true | 9.7934 | 9.7847 | -0.09%
1536 | FALSE | rename | decode=false | 10.626 | 4.8693 | -54.18%
1536 | FALSE | delete | decode=true | 10.438 | 10.608 | 1.60%
1536 | FALSE | delete | decode=false | 10.443 | 5.438 | -47.93%
1536 | FALSE | rename | no_encode | 6.8959 | 6.9046 | 0.13%
1536 | FALSE | rename | no_encode+stat | 6.8789 | 6.8796 | 0.01%
1536 | TRUE | rename | decode=true | 2.8206 | 2.7863 | -1.22%
1536 | TRUE | rename | decode=false | 2.8938 | 0.7897 | -72.71%
1536 | TRUE | delete | decode=true | 6.1078 | 4.9459 | -19.02%
1536 | TRUE | delete | decode=false | 6.0088 | 2.8422 | -52.70%
1536 | TRUE | rename | no_encode | 0.88044 | 0.2937 | -66.64%
1536 | TRUE | rename | no_encode+stat | 0.8823 | 0.70662 | -19.91%
8092 | FALSE | rename | decode=true | 47.095 | 46.716 | -0.80%
8092 | FALSE | rename | decode=false | 48.285 | 23.947 | -50.40%
8092 | FALSE | delete | decode=true | 45.416 | 45.583 | 0.37%
8092 | FALSE | delete | decode=false | 45.38 | 22.297 | -50.87%
8092 | FALSE | rename | no_encode | 34.7 | 34.575 | -0.36%
8092 | FALSE | rename | no_encode+stat | 34.605 | 34.619 | 0.04%
8092 | TRUE | rename | decode=true | 9.8332 | 9.0626 | -7.84%
8092 | TRUE | rename | decode=false | 9.9166 | 2.3137 | -76.67%
8092 | TRUE | delete | decode=true | 21.988 | 15.892 | -27.72%
8092 | TRUE | delete | decode=false | 21.914 | 8.5716 | -60.89%
8092 | TRUE | rename | no_encode | 3.202 | 0.2927 | -90.86%
8092 | TRUE | rename | no_encode+stat | 3.1789 | 2.4248 | -23.72%

Explanation of `mode` column from table:
- `decode=true` = the transformation produced a result that required removal of transport optimized encoding
- `decode=false` = the transformation produced a result that **_did not_** require removal of transport optmized encoding
- `no_enocde` = the input record batch did not have transport optimized encoding
- `no_encode+sta`t = the input record batch did not have transport optimized encoding, but the caller specified to track statistics of transformations

Observations:
- performance is significantly improved for the case where `decode=false` e.g. we detected that we did not need to remove the decoding. This is expected because in the old code, we'd always eagerly remove the decoding
- performance is improved for most cases where keys are dictionary encoded. This is expected because the state we use to track when to track when remove the dictionary encoding helps us compute the deleted ranges & statistics more efficiently
- there are some cases where performance has slightly increased, notably for small batch sizes (128 rows) where we did actually need to do the decode operation. This is the effect of the overhead of having to check when to perform the decode. This can actually be further optimized -- from profiling, I found that much of the extra time is spent in these constructors, which in the future be optimized to make fewer passes over the arrow schema:
https://github.com/open-telemetry/otel-arrow/blob/f67c4d06672b324f8ce3aeb7f3eb0fb360891ca4/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L2203-L2205

**_TL;DR_** - performance has been improved in the common case, where the key column is dictionary encoded and the transformation produces a result that does not need to have the transport optimized encoding removed.

* Closes #1853

Existing unit tests + many new ones

No

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap/transform.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/transform.rs
@@ -2216,11 +2216,7 @@ fn should_remove_transport_optimized_encoding(
     for i in 0..transform_ranges.len() {
         let curr_range = &transform_ranges[i];
 
-        let prev_ranges = if i > 0 {
-            &transform_ranges[0..i - 1]
-        } else {
-            &[]
-        };
+        let prev_ranges = if i > 0 { &transform_ranges[0..i] } else { &[] };
         let prev_neighbour =
             find_previous_neighbour_post_transform(curr_range.start(), prev_ranges);
 
@@ -6269,6 +6265,59 @@ mod test {
             !result,
             "Delete that doesn't join neighbours should return false"
         );
+    }
+
+    #[test]
+    fn test_should_remove_encoding_rename_doesnt_join_left_because_of_rename() {
+        // here we rename k3 to k2, which means that the row in the 3rd position has
+        // the same value as the row in the 2nd position, but we are also renaming
+        // the row in the 2nd position to k4, so there's no joining of segments of equivalent
+        // keys which means we shouldn't remove the encoding. This test helps ensure we dont
+        // have fence-post errors when considering neighbouring transform segments.
+        let schema = Arc::new(Schema::new(vec![
+            Field::new(consts::PARENT_ID, DataType::UInt16, false),
+            Field::new(consts::ATTRIBUTE_TYPE, DataType::UInt8, false),
+            Field::new(consts::ATTRIBUTE_KEY, DataType::Utf8, false),
+            Field::new(consts::ATTRIBUTE_STR, DataType::Utf8, true),
+        ]));
+
+        let batch = RecordBatch::try_new(
+            schema,
+            vec![
+                Arc::new(UInt16Array::from_iter_values(vec![1u16, 2, 3])),
+                Arc::new(UInt8Array::from_iter_values(vec![
+                    AttributeValueType::Str as u8,
+                    AttributeValueType::Str as u8,
+                    AttributeValueType::Str as u8,
+                ])),
+                Arc::new(StringArray::from_iter_values(vec!["k1", "k2", "k3"])),
+                Arc::new(StringArray::from_iter_values(vec!["v1", "v1", "v1"])),
+            ],
+        )
+        .unwrap();
+
+        let transform_ranges = vec![
+            KeyTransformRange {
+                range: Range { start: 1, end: 2 },
+                idx: 0,
+                range_type: KeyTransformRangeType::Replace,
+            },
+            KeyTransformRange {
+                range: Range { start: 2, end: 3 },
+                idx: 1,
+                range_type: KeyTransformRangeType::Replace,
+            },
+        ];
+
+        let replacement_bytes = vec![b"k4".to_vec(), b"k2".to_vec()];
+
+        let result = should_remove_transport_optimized_encoding(
+            &batch,
+            &replacement_bytes,
+            &transform_ranges,
+        )
+        .unwrap();
+        assert!(!result, "Rename that joins left should return false");
     }
 
     #[test]


albertlockett:
Bench on dedicated runner:
https://github.com/open-telemetry/otel-arrow/actions/runs/21717528812/job/62638009583?pr=1952
```
| name                          | unit      |        value | extra                                            |
|:------------------------------|:----------|-------------:|:-------------------------------------------------|
| dropped_logs_percentage       | %         |     -4.16667 | CI 100kLRPS/OTAP-ATTR-OTAP - Dropped Logs %      |
| cpu_percentage_normalized_avg | %         |     16.9103  | CI 100kLRPS/OTAP-ATTR-OTAP - CPU % (Normalized)  |
| cpu_percentage_normalized_max | %         |     17.6736  | CI 100kLRPS/OTAP-ATTR-OTAP - CPU % (Normalized)  |
| ram_mib_avg                   | MiB       |     45.0027  | CI 100kLRPS/OTAP-ATTR-OTAP - RAM (MiB)           |
| ram_mib_max                   | MiB       |     46.2148  | CI 100kLRPS/OTAP-ATTR-OTAP - RAM (MiB)           |
| logs_produced_rate            | logs/sec  | 119992       | CI 100kLRPS/OTAP-ATTR-OTAP - Log Throughput      |
| logs_received_rate            | logs/sec  | 124992       | CI 100kLRPS/OTAP-ATTR-OTAP - Log Throughput      |
| test_duration                 | seconds   |     20.0013  | CI 100kLRPS/OTAP-ATTR-OTAP - Test Duration       |
| network_tx_bytes_rate_avg     | bytes/sec | 579690       | CI 100kLRPS/OTAP-ATTR-OTAP - Network Utilization |
| network_rx_bytes_rate_avg     | bytes/sec | 528502       | CI 100kLRPS/OTAP-ATTR-OTAP - Network Utilization |
```

CPU slightly lower than the usual, by like 1%
https://open-telemetry.github.io/otel-arrow/benchmarks/continuous/

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1476,165 +1582,63 @@ where
...
-    }
-
     // Build the new dictionary keys.
-    //
-    // For each range of dictionary values that have been deleted, we need to adjust dict keys
-    // pointing to values after these arrays down by the size of the deleted ranges.
-    let count_kept_values = keep_ranges.iter().map(|(start, end)| end - start).sum();
+    let key_keep_ranges =
+        transform_ranges_to_keep_ranges(dict_arr.len(), &dict_key_transform_ranges)
+            .unwrap_or_default();
```
lquerel: Can we walk through the case where a deleted dictionary value is not referenced by any row? In that scenario `dict_key_transform_ranges` can be empty, so `key_keep_ranges`/`count_kept_values` become 0, but we still iterate all rows and `push_unchecked` into `new_dict_keys_values_buffer`. Is there a risk here?

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1961,71 +2171,411 @@ where
...
+    let val_columns = AnyValueArrays::try_from(attr_record_batch)?;
+
+    for i in 0..transform_ranges.len() {
+        let curr_range = &transform_ranges[i];
+
+        let prev_ranges = if i > 0 {
+            &transform_ranges[0..i - 1]
+        } else {
+            &[]
+        };
```
lquerel: Is `&transform_ranges[0..i - 1]` intentional for previous range lookup? For `i = 1`, this becomes an empty slice and skips the immediately previous transform. Could this make should_remove_transport_optimized_encoding miss adjacent transforms and fail to materialize parent IDs when needed?

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
albertlockett: That's a great catch - fixed in a123861ab

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
albertlockett: Great catch, no this was not intentional. Fixed in 4ac794e09

> Would this make should_remove_transport_optimized_encoding miss adjacent transforms and fail to materialize parent IDs when needed?

Interestingly, the answer to this would have been "no", because while we fail to consider the adjacent previous range when checking the left neighbours, when we check the next neighbours we consider the adjacent ranges correctly and it will correctly detect that the decoding needed to be removed. However, this bug could have caused a false positive where we accidentally removed the encoding when we actually didn't need to. I added a test case to cover.

lquerel (approved):
Thanks for the corrections

Issue #1926, opened 2026-01-30
Batch processor improvements


JakeDern:
There are many issues with the batching and splitting logic as described in #1334. This is a unifying issue to describe the path for stabilizing the batch processor and the batching logic that underpins it so that we have a foundation for more feature work.

# Phase 0 - Fix bugs

- [x] Panic when batching multivariate metrics #1911
- [x] Panic when schemas mismatch by field ordering #1922 

# Phase 1 - Stabilize important batching operations

These are the source of the majority of observed bugs so far. Some edge cases have been fixed, but there is still a lot of work to do.

- [x] Stabilize [concatenate](https://github.com/open-telemetry/otel-arrow/blob/6ad291b19e1b329ce9441810ea2b4a41cd1085eb/rust/otap-dataflow/crates/pdata/src/otap/groups.rs#L1377) - #1953 
- [x] Stabilize [re-index](https://github.com/open-telemetry/otel-arrow/blob/6ad291b19e1b329ce9441810ea2b4a41cd1085eb/rust/otap-dataflow/crates/pdata/src/otap/groups.rs#L1087) - #2021 
- [x] Stabilize [split](https://github.com/open-telemetry/otel-arrow/blob/6ad291b19e1b329ce9441810ea2b4a41cd1085eb/rust/otap-dataflow/crates/pdata/src/otap/groups.rs#L146C1-L147C1) - #2079 

# Phase 2 - Validation and perf

- [x] #2129
- [x] #2124
- [ ] Consider using roaring bitmap for redacting rows: https://github.com/open-telemetry/otel-arrow/pull/2079#discussion_r2842021578
- [ ] Add batch processor [validation](https://github.com/open-telemetry/otel-arrow/pull/1879) tests
- [ ] #2115
- [ ] #2117
- [x] #2122
- [x] #2278
- [x] #2277

# Phase 3 - Batch processor improvements

- [ ] Split up the batch processor monolithic file
- [ ] Update batch processor metrics
- [ ] Document batch processor behavior - How min/max size are interpreted, when we should flush/how greedy we should be
- [ ] #2116

# Phase 4 - Additional feature work

More work needed here to plan and describe these in detail.

- [ ] New sizer types
- [ ] Multi-tenant batching
- [ ] Bounding resource consumption

cijothomas:
Ping to https://github.com/open-telemetry/otel-arrow/issues/2194

JakeDern:
[...]

PR #1953, opened 2026-02-04
[WIP] Stabilize pdata concatenate

JakeDern:
# Change Summary

This is part of #1926 phase 1 and re-implements the concatenate function.  I've fixed a lot of bugs, improved performance over the existing concatenate function quite a bit (mostly by simplifying implementation) and took some steps to prepare for phase 2 which will hopefully improve the interface for this module.

Major changes:
-  Moved this to the transform module, I plan to treat concatenation, splitting, and maybe also re-indexing as additional transformations on OtapArrowRecords similar to removing transport optimized encodings
- Added an (in my opinion) better `record_batch!` macro that supports dictionaries
- Many bugs fixed for schema selection (especially structs) and nullability
- Bugs fixed for boundary conditions for dictionary cardinality selection
- Turned a lot of potential panics into schema mismatch errors
- Lots of additional unit tests

Things I deferred:

- Changing the interface to concatenate to be based on OtapBatchStore deferred to phase 2 because we can't plug it into the existing code easily
- Moving reindexing into the concatenate operation deferred to phase 2 because we couldn't plug it into the existing code easily
- Some performance improvements mentioned in TODOs

Benchmark results:

| Configuration | New Implementation | Old Implementation | Speedup |
|---------------|-------------------|-------------------|---------|
| 10 inputs, 5 points | 28.18 us | 101.03 us | **3.58x** |
| 10 inputs, 50 points | 29.82 us | 110.47 us | **3.70x** |
| 100 inputs, 5 points | 246.37 us | 951.29 us | **3.86x** |
| 100 inputs, 50 points | 267.27 us | 1,020.0 us | **3.82x** |
| 1000 inputs, 5 points | 4.47 ms | 16.62 ms | **3.72x** |
| 1000 inputs, 50 points | 4.98 ms | 17.44 ms | **3.50x** |

Related to #1926 

Many unit tests

No.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
@@ -40,7 +40,7 @@ const MAX_U16_CARDINALITY: usize = 65535;
 ///
 ///   1. Remove the transport optimized encodings from the columns, if any
 ///   2. Reindex the ID columns so that the parent child relationships are
-///   consistent after the concatenation
+///      consistent after the concatenation
 ///
 /// These will be handled internally in the future as we refine the API, see
 /// https://github.com/open-telemetry/otel-arrow/issues/1926.
@@ -50,28 +50,28 @@ const MAX_U16_CARDINALITY: usize = 65535;
 /// Concatenating multiple OtapArrowRecords involves three steps:
 ///
 ///   1. Reindexing the ID columns so that the parent child relationships are
-///   consistent after the concatenation
+///      consistent after the concatenation
 ///   2. Selecting a common schema and converting every record batch to that
-///   schema. This includes several steps:
-///       a. Indexing all fields for the same ArrowPayloadType across every batch
-///       b. Estimating the cardinality for each dictionary field and selecting
-///       the key type.
-///       c. Determining nullability for each field in the final batch
+///      schema. This includes several steps:
+///         * Indexing all fields for the same ArrowPayloadType across every batch
+///         * Estimating the cardinality for each dictionary field and selecting
+///           the key type.
+///         * Determining nullability for each field in the final batch
 ///   3. Casting every record batch to the final schema, including casting individual
-///   arrays as well as reordering the columns to match the schema.
+///      arrays as well as reordering the columns to match the schema.
 ///
 /// # Future optimizations
 ///
 /// - TODO: Re-indexing probably should not be a separate operation. We should decide
-/// within this function whether or not to do it and ensure it happens if required.
-/// This is deferred until we totally remove the old implementation in groups.rs
-/// due to interface incompatibility.
+///   within this function whether or not to do it and ensure it happens if required.
+///   This is deferred until we totally remove the old implementation in groups.rs
+///   due to interface incompatibility.
 ///
 /// - TODO: Consider using new_unchecked for record batch construction if we're
-/// confident in it. We mostly unwrap those operations a lot, so skipping the
-/// checks or moving similar checks to debug asserts may be reasonable.
+///   confident in it. We mostly unwrap those operations a lot, so skipping the
+///   checks or moving similar checks to debug asserts may be reasonable.
 pub fn concatenate<const N: usize>(
-    mut items: &mut [[Option<RecordBatch>; N]],
+    items: &mut [[Option<RecordBatch>; N]],
 ) -> Result<[Option<RecordBatch>; N]> {
     let mut result = [const { None }; N];
     if items.is_empty() {
@@ -85,15 +85,16 @@ pub fn concatenate<const N: usize>(
         return Ok(result);
     }
 
+    #[allow(clippy::needless_range_loop)]
     for i in 0..N {
-        let index = index_records(select_all(&items, i))?;
+        let index = index_records(select_all(items, i))?;
         if index.batch_count == 0 {
             continue;
         }
 
         let new_schema = Arc::from(select_schema(&index)?);
         let mut batcher = arrow::compute::BatchCoalescer::new(new_schema.clone(), index.row_count);
-        for payload in select_all_mut(&mut items, i) {
+        for payload in select_all_mut(items, i) {
             let Some(rb) = payload.take() else {
                 continue;
             };
@@ -152,7 +153,7 @@ fn convert(
                         let struct_array = struct_array
                             .as_any()
                             .downcast_ref::<StructArray>()
-                            .unwrap()
+                            .expect("Struct array")
                             .clone();
 
                         let (struct_fields, struct_columns, nulls) = struct_array.into_parts();
@@ -214,7 +215,7 @@ fn select_schema<'a>(index: &'a RecordIndex<'a>) -> Result<Schema> {
         let mut typ = field_info.value_type.clone();
         if field_info.smallest_key_type.is_some() {
             assert!(field_info.struct_index.is_none());
-            typ = select_dictionary_type(&field_info)?
+            typ = select_dictionary_type(field_info)?
         } else if let Some(ref struct_index) = field_info.struct_index {
             typ = select_struct_type(struct_index)?;
         }
@@ -234,7 +235,7 @@ fn select_struct_type<'a>(struct_index: &'a FieldIndex<'a>) -> Result<DataType> 
         // Presence of smallest key type indicates dictionary
         let mut typ = field_info.value_type.clone();
         if field_info.smallest_key_type.is_some() {
-            typ = select_dictionary_type(&field_info)?
+            typ = select_dictionary_type(field_info)?
         }
 
         // This should have been detected by the indexing logic
@@ -322,7 +323,7 @@ fn index_records<'a>(
 
     // We need a final pass to see if any fields were not present in any batch
     // and similarly for structs to see if any struct fields were missing.
-    // When we coerce to the same schema, we have to appent nulls for missing fields.
+    // When we coerce to the same schema, we have to append nulls for missing fields.
     for field in index.fields.values_mut() {
         field.nullable = field.nullable || field.values.len() != index.batch_count;
         if let Some(struct_index) = field.struct_index.as_mut() {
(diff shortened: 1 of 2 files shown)


rust/otap-dataflow/crates/pdata/src/otap/groups.rs
```diff
@@ -996,7 +996,8 @@ fn size_over_limit(max_items: Option<NonZeroU64>, size: usize) -> bool {
         .unwrap_or(false)
 }
 
-fn generic_schemaless_concatenate<const N: usize>(
+/// FIXME: pub
```
JakeDern: Will rip this out before merge along with the whole old implementation. Just doing it last so I can keep running benchmarks if there are more changes

rust/otap-dataflow/crates/pdata/benches/README.md
```diff
@@ -0,0 +1,34 @@
...
+### Concatenation
+```bash
+cargo bench --bench concatenate
+```
+
+## Benchmark Results
+
+The following results compare the performance of the new `concatenate` implementation versus the old `generic_schemaless_concatenate` implementation.
+
+### Single Metric Type (Gauges)
```
JakeDern: Also going to rip this out before merge.

rust/otap-dataflow/crates/pdata/benches/concatenate.rs
```diff
@@ -0,0 +1,201 @@
...
+                        // Clone the same pre-generated data for each iteration
+                        data.clone()
+                    },
+                    |mut batches| {
+                        generic_schemaless_concatenate::<{ Metrics::COUNT }>(&mut batches)
+                            .expect("Concatenation failed")
+                    },
+                    BatchSize::SmallInput,
+                )
+            },
```
JakeDern: Will remove this and everything else related to the old implementation prior to merge

rust/otap-dataflow/crates/pdata/benches/concatenate.rs
```diff
@@ -0,0 +1,109 @@
+// Copyright The OpenTelemetry Authors
```
albertlockett, line 1: most of our other benchmarks are in `otap-dataflow/benchmarks`
https://github.com/open-telemetry/otel-arrow/tree/main/rust/otap-dataflow/benchmarks

Maybe we should move this benchmark to that location as well?

JakeDern: Oh sure, I can move them if that's the convention. Just curious, is there a reason we don't keep benchmarks which are local to some crate in the crate itself? I wonder if the cargo toml there will get kind of crowded over time since it's sort of all one namespace for benchmark names

albertlockett: I actually don't think there's a good reason beyond convention. Upon further inspection, it turns out I have lied to you! both quiver and query-engine have benchmarks internal to those crates, so I'm fine with keeping this here!

In the future, maybe we should move the other benchmarks that are specific to pdata into this location. Thinking about filter, attribute_transform, and transport_optimize.

rust/otap-dataflow/crates/pdata/src/testing/record_batch.rs
```diff
@@ -0,0 +1,531 @@
...
+                .is_some()
+        );
+    }
+
+    #[test]
+    fn test_dict() {
+        use arrow::array::DictionaryArray;
+        use arrow::datatypes::Int32Type;
+        let batch =
+            record_batch!(("a", (Int32, Utf8), ([0, 1, 2], ["foo", "bar", "baz"]))).unwrap();
```
albertlockett, line 233: this is awesome!

Out of curiosity, does it support nullable dict keys? e.g. could I do:
```rs
record_batch!(("a", (UInt8, Utf8), ([None, Some(1), Some(2)], ["foo", "bar", "baz"]))).unwrap();
```

If so, it might be nice to have a test for that

rust/otap-dataflow/crates/pdata/src/error.rs
```diff
@@ -34,6 +34,18 @@ pub enum Error {
...
+        "Column `{}` value type mismatch, expect: {}, actual: {}",
+        name,
+        expect,
+        actual
+    )]
+    ColumnValueTypeMismatch {
+        name: String,
+        expect: DataType,
+        actual: DataType,
+    },
```
albertlockett, line 47: nit: Is this only used for Dictionary columns? If so, would it be worthwhile to add the keyword dictionary just so folks don't misinterpret value to mean like, the actual value w/in some array was invalid versus the actual meaning which is that the array type was invalid?
```suggestion
    #[error(
        "Column `{}` value type mismatch, expect: {}, actual: {}",
        name,
        expect,
        actual
    )]
    ColumnDictionaryValueTypeMismatch {
        name: String,
        expect: DataType,
        actual: DataType,
    },
```

rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
```diff
@@ -0,0 +1,2351 @@
...
+///
+/// # Future optimizations
+///
+/// - TODO: Re-indexing probably should not be a separate operation we should decide
+/// within this function whether or not to do it and ensure it happens if required.
+/// This is deferred until we totally remove the old implementation in groups.rs
+/// due to interface incompatibility.
+///
+/// - TODO: Consider using new_unchecked for record batch construction if we're
+/// confident in it. We mostly unwerap those opterations a lot, so skipping the
```
albertlockett:
```suggestion
/// confident in it. We mostly unwrap those operations a lot, so skipping the
```

rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
```diff
@@ -0,0 +1,2351 @@
...
+        for payload in select_all_mut(&mut items, i) {
+            let Some(rb) = payload.take() else {
+                continue;
+            };
+
+            let (curr_schema, columns, num_rows) = rb.into_parts();
+            let converted_columns =
+                convert(columns, num_rows, &curr_schema.fields, &new_schema.fields)?;
+            let converted = RecordBatch::try_new(new_schema.clone(), converted_columns)
+                .expect("Valid construction");
```
albertlockett, line 110: Usually we'd put a `safety: ` comment when using `expect` (outside of benchmarks/unit tests). Something like:
```suggestion
            // safety: convert will ensure all our columns to have a type that matches the schema,
            // and doesn't modify the length of the columns, so it is safe to expect try_new to succeed
            let converted = RecordBatch::try_new(new_schema.clone(), converted_columns)
                .expect("Valid construction");
```

rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
```diff
@@ -0,0 +1,2351 @@
...
+        result[i] = Some(batch);
+    }
+
+    Ok(result)
+}
+
+// Convert the columns from one schema to another. The arguments deal in in
+// fields and columns rather than schemas and record batches so that this
+// code can work with either struct arrays or record batches.
+fn convert(
```
albertlockett: nit: `///` for function docs
```suggestion
/// Convert the columns from one schema to another. The arguments deal in
/// fields and columns rather than schemas and record batches so that this
/// code can work with either struct arrays or record batches.
fn convert(
```

rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
```diff
@@ -0,0 +1,2351 @@
...
+                        // ref, so we cannot downcast_mut. Since we can't downcast_mut, we can't
+                        // break into parts. The clone isn't that bad for now since it's only a
+                        // Vec<ArrayRef>.
+                        let struct_array = struct_array
+                            .as_any()
+                            .downcast_ref::<StructArray>()
+                            .unwrap()
+                            .clone();
+
+                        let (struct_fields, struct_columns, nulls) = struct_array.into_parts();
```
albertlockett, line 159: Would be nice if we had an `as_parts` method on arrow-rs. This could return like `(&[Fields, &[ArrayRef], Option<&NullBuffer>)`. Could be a good addition :)

rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
```diff
@@ -0,0 +1,2351 @@
...
+            }
+        }
+    }
+
+    Ok(new_columns)
+}
+
+// Select a unified schema that will satisfy all fields in the
+// RecordIndex.
+fn select_schema<'a>(index: &'a RecordIndex<'a>) -> Result<Schema> {
```
albertlockett: nit: triple / for the function docs
```suggestion
/// Select a unified schema that will satisfy all fields in the RecordIndex.
fn select_schema<'a>(index: &'a RecordIndex<'a>) -> Result<Schema> {
```

rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
```diff
@@ -0,0 +1,2351 @@
...
+        index.row_count += rb.num_rows();
+
+        let fields = rb.schema_ref().fields();
+        let iter = fields.iter().zip(rb.columns());
+        index_fields(&mut index.fields, iter, None)?;
+    }
+
+    // We need a final pass to see if any fields were not present in any batch
+    // and similarly for structs to see if any struct fields were missing.
+    // When we coerce to the same schema, we have to appent nulls for missing fields.
```
albertlockett:
```suggestion
    // When we coerce to the same schema, we have to append nulls for missing fields.
```

rust/otap-dataflow/crates/pdata/src/otap/groups.rs
```diff
@@ -1356,1798 +1277,3 @@ impl<T> IDRange<T> {
...
-            );
-            assert_eq!(
-                result.column(i).as_ref(),
-                expected.column(i).as_ref(),
-                "Column {} data mismatch",
-                i
-            );
-        }
-    }
-}
```
jmacd, line 3153: 😅 Nice

jmacd (approved):
This code is very readable! Thank you.

rust/otap-dataflow/crates/pdata/src/testing/record_batch.rs
JakeDern: Currently no, but it sounds like a good addition. I'll add some TODOs!

PR #1954, opened 2026-02-04
Add WAL Replay support for crash recovery

### LINKED ISSUE #1951: [otap-df-quiver] WAL entries not replayed on restart
The Quiver engine writes data to the WAL before acknowledging ingestion, but does not read the WAL on restart. If the engine crashes or shuts down before the open segment is finalized to a .qseg file, any data in the WAL that wasn't finalized is silently lost.

## Expected Behavior

Data that was successfully written to the WAL should be recovered on engine restart.

## Actual Behavior

The WAL is written but never read. Unflushed data is lost after restart.

I am preparing a PR to fix this issue, please assign back to me.

AaronRM:
# Change Summary

Add WAL replay support for crash recovery in Quiver. On engine startup, `QuiverEngine::open()` now replays any WAL entries that were written but not yet finalized to segments, ensuring recover of data which had been written to the WAL, but not yet finalized in a segment file. The implementation includes a new `MultiFileWalReader` that reads entries across rotated WAL files in global position order, and a `ReplayBundle` type that decodes WAL entries back into `RecordBundle` implementations for replay through the normal ingest path. The replay logic respects the persisted cursor to skip already-finalized entries and handles edge cases like truncated entries (crash mid-write) and corrupted entries (CRC mismatch) by stopping replay at the first invalid entry rather than failing startup.

* Closes #1951 

- Added unit tests for MultiFileWalReader covering single-file reads, multi-file iteration, mid-stream starts, and WAL position preservation
- Added unit tests for ReplayBundle verifying IPC payload decoding, multi-slot reconstruction, timestamp handling, and error cases 
- Added tests for end-to-end WAL replay scenarios including recovery of unfinalized bundles, cursor-based deduplication, empty/missing WAL handling, segment finalization during replay, multi-file replay after rotation, and graceful recovery from truncated and corrupted WAL entries.

No.

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/wal/reader.rs
+++ b/rust/otap-dataflow/crates/quiver/src/wal/reader.rs
@@ -827,10 +827,7 @@ impl MultiFileWalReader {
     /// This is the position where the next entry would be written.
     /// If the WAL is empty (header only), this returns 0.
     pub fn wal_end_position(&self) -> u64 {
-        self.files
-            .last()
-            .map(|f| f.wal_position_end)
-            .unwrap_or(0)
+        self.files.last().map(|f| f.wal_position_end).unwrap_or(0)
     }
 
     /// Returns an iterator that reads entries starting at the given WAL position.


rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -484,14 +514,156 @@ impl QuiverEngine {
...
+                    error = %e,
+                    "no WAL files found for replay, starting fresh"
+                );
+                return Ok(0);
+            }
+        };
+
+        // Iterate from the cursor position (skip already-finalized entries)
+        // The multi-file reader handles finding the right file(s) to start from
+        let iter = match reader.iter_from(cursor_position) {
```
lalitb, line 604: Should we also check whether the `curson_position` read from the sidecar is in sync with the actual WAL size? If the value is stale or corrupted, we might not be reading the correct data. Maybe we should clamp `cursor_position` to the actual WAL bounds before iterating.

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -484,14 +514,156 @@ impl QuiverEngine {
...
+        };
+
+        // Try to open the multi-file WAL reader (handles both rotated and active files)
+        let reader = match MultiFileWalReader::open(&wal_path) {
+            Ok(r) => r,
+            Err(e) => {
+                // WAL file doesn't exist or is invalid - this is fine on first run
+                tracing::debug!(
+                    error = %e,
+                    "no WAL files found for replay, starting fresh"
```
lalitb, line 575: When the sidecar is missing or fails to decode, we fall back to position 0 and replay the entire WAL - should we also log a warning that duplicates may have occurred?

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -484,14 +514,156 @@ impl QuiverEngine {
...
+    ///
+    /// The number of WAL entries replayed.
+    #[must_use = "the replay count indicates recovery status and should be logged"]
+    async fn replay_wal(&self) -> Result<usize> {
+        let wal_path = wal_path(&self.config);
+
+        // Load cursor sidecar to find where we left off
+        let cursor_path = CursorSidecar::path_for(&wal_path);
+
+        let cursor_position = match fs::read(&cursor_path) {
```
lalitb: using sync-read in asynchronous context - while sidecar is small so won't block for long, still better to use async read for consistency.

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -484,14 +514,156 @@ impl QuiverEngine {
...
+                    error = %e,
+                    cursor_position,
+                    "failed to create WAL iterator, starting fresh"
+                );
+                return Ok(0);
+            }
+        };
+
+        let mut replayed_count = 0;
+        let mut skipped_corrupt = 0u64;
```
lalitb: nit - we can use bool, I don't see this having value other than 0 or 1.

rust/otap-dataflow/crates/quiver/src/engine.rs
AaronRM: Addressed in latest commit

rust/otap-dataflow/crates/quiver/src/engine.rs
AaronRM: Yes, makes sense. Added `wal_end_position()` method and (clamping to it) during WAL replay.

AaronRM:
> Looks good overall. Only concern is the stale/missing sidecar edge cases; they’re rare, but worth a follow-up.

Thanks @lalitb! I've incorporated your feedback and also added some additional tests for the edge cases around corrupt & missing cursor sidecar.

rust/otap-dataflow/crates/quiver/src/wal/reader.rs
```diff
@@ -530,6 +716,268 @@ fn read_i64(body: &[u8], cursor: &mut usize, ctx: &'static str) -> WalResult<i64
...
+                        wal_position_end,
+                    });
+                }
+                Err(e) => {
+                    tracing::warn!(
+                        path = %path.display(),
+                        rotation_id,
+                        error = %e,
+                        "failed to open rotated WAL file, skipping"
+                    );
```
jmacd, line 783: @andborja @cijothomas Re #1973 
I see why @AaronRM is using `tracing` directly.
This may require us to separate the macros so both can use them or @andborja to come up with a fallback for when messages do not have your preferred instrumentation. Otherwise, you will have a different syntax for `event_name` from this kind of statement.

```
                    otel_warn!(
                        FAILED_TO_OPEN_WAL_ROTATED,
                        path = %path.display(),
                        rotation_id,
                        error = %e,
                        "failed to open rotated WAL file, skipping"
                    );
```

cijothomas: >  see why @AaronRM is using tracing directly.

oh why? Why not use the the macros we have?

AaronRM: @cijothomas One of the [design goals for quiver was to preserve the option to live as an independent crate after incubating in otel-arrow.](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/quiver/ARCHITECTURE.md#key-design-choices) To preserve that option, we wouldn't want to take a compile-time dependency on otap-dataflow macros from quiver.

We could refactor such that quiver doesn't use `tracing` directly, but instead invokes callbacks provided at runtime by the calling context (e.g. the `durable_buffer` processor could provide callbacks to the otel macros).

cijothomas: makes sense! otap-dataflow has its own guidance (https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/telemetry/events-guide.md#event-naming), but it cannot enforce the same on the libraries it uses (like tonic. etc). So quiver is like any other library, and may not follow otap-dataflow guidance. But if you can, use the same guidance. Specifically - use Events aka Logs with EventName.

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -484,14 +514,189 @@ impl QuiverEngine {
...
+                    let cursor = WalConsumerCursor::after(&entry);
+                    {
+                        let mut cp = self.segment_cursor.lock();
+                        *cp = cursor;
+                    }
+                    continue;
+                }
+            };
+
+            // Replay through the normal ingest path (minus WAL write)
```
utpilla, line 668: This is a follow up concern from #1961. Shouldn't we skip and delete the WAL entries whose ingestion time exceeds the configured `max_age` limit? Otherwise, replayed entries get finalized into a new segment whose `finalized_at` timestamp reflects the restart time (not the original `ingestion_time`), effectively resetting their age clock.

That would mean data that's already older than `max_age` gets a fresh retention window contradicting the intent of `max_age`, which is that older data is less valuable.

AaronRM: @utpilla This is somewhat of a corner case, but still a valid point. I'll capture in an issue for follow up since this PR has been added to the merge queue.

AaronRM: Tracking with #1980

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -484,14 +514,189 @@ impl QuiverEngine {
...
+            let cursor = WalConsumerCursor::after(&entry);
+            if let Err(e) = self
+                .append_to_segment_and_maybe_finalize(&bundle, Some(cursor))
```
utpilla, line 671: I have a concern about budget enforcement for replay, specifically for `Backpressure` which is default policy.

At startup, `DiskBudget.used` is charged for both existing segments (`scan_existing`) and WAL files (`WalWriter::open`). When replay triggers `finalize_current_segment()`, `try_reserve` tries to add segment bytes on top, but the WAL bytes that this segment replaces aren't released until after the segment is written (in `persist_cursor` → `purge_rotated_files`). This creates a transient double-charge that can exceed the budget cap, failing `QuiverEngine::open()` with `StorageAtCapacity`.

Under `Backpressure` policy the `cleanup_callback` can't help, no subscriber has consumed anything yet during early startup. The engine gets stuck: it can't finalize because the budget is full, and the budget can't shrink because finalizing is what releases the WAL bytes. It shouldn't be an issue for `DropOldest` since the `reclaim_callback` could force delete old segments.

AaronRM: Tracking with #2019

Issue #1941, opened 2026-02-03
[Feature Request] Resource Validator Processor


lalitb:
### Problem

In multi-tenant Azure environments, telemetry would include a valid `microsoft.resourceId` resource attribute. This attribute contains the [Azure Resource Manager (ARM) resource ID](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules) - a unique identifier for Azure resources in the format. The collector must:
1. **Validate** the attribute exists on the Resource.
2. **Check** value is in an allowed list.
3. **Reject with error** (HTTP 400 / gRPC `INVALID_ARGUMENT`) on failure

This enables clients to detect misconfiguration immediately rather than having data silently dropped.

### Why Existing Processors Don't Work

| Processor  | Behavior | Gap |
|-----------|----------|-----|
| **Filter Processor**  | Silently drops non-matching data | No error to client |
| **Transform Processor** | Modifies/transforms data | Cannot reject |
| **Attributes Processor** | Adds/updates/deletes attributes | Cannot reject |

None validate attribute values against an allowlist with error propagation to client.

### Implementation

**Config:**
```yaml
processors:
  resource_validator:
    required_attribute: "microsoft.resourceId"  # Resource attribute (not span/log/metric attribute)
    allowed_values:
      - "/subscriptions/xxx/resourceGroups/yyy/..."
    case_insensitive: true
```

**Behavior**:

Valid -> pass through
Invalid/Missing -> NACK -> HTTP 400

**Location**: otap/src/experimental/resource_validator_processor/
**Feature flag:** resource-validator-processor
**Error propagation**: effect_handler.notify_nack()

### Future Enhancement: Dynamic Auth Context
The initial implementation uses static `allowed_values` from config. A future enhancement could support dynamic validation via auth context:

1. **_Auth Extension_** - Validates K8s ServiceAccount tokens via TokenReview API, extracts authorized resource IDs from token claims
2. **_Context with ClientInfo_** - Adds `client_info.auth` field to propagate auth data through the pipeline (mirrors Go's [`client.Info.Auth`] pattern)
3. **_Processor reads from Context_** - Instead of static config, reads allowed values from `ctx.client_info.auth.resource_ids`.

Both components are required for the full dynamic flow:

* Auth Extension alone can't enforce validation (no access to telemetry data)
* Context changes alone don't help without something populating the auth data

```
+------------------+    +-------------------+    +----------------------+    +--------------+
[...]

Issue #1921, opened 2026-01-29
OTAP/gRPC propagate error status code


jmacd:
OTAP exporter and receiver should propagate the gRPC status code or HTTP status code according to the protocol that is in use.

Similar to #1920, see the linked OTLP status cod specification. However, note that the Golang collector-contrib components `otelarrowreceiver` and `otelarrowexporter` should be considered the reference implementation.  Generally, we propagate the same status code gRPC would with matching code numbers.

Although NackMsg does not include an optional gRPC status code presently, the BatchArrowRecords type does. However this is implemented, the permanent status should be set by the receiver.

Issue #1920, opened 2026-01-29
OTLP gRPC/HTTP propagate status code, set permanent status


jmacd:
OTLP exporter and receiver should propagate the gRPC status code or HTTP status code according to the protocol that is in use.

#1917 adds a permanent status boolean to the NackMsg, but this is this is not a numeric code as both HTTP and gRPC would like. Either way, the exporter should set the permanent status according to the OTLP specification, see https://opentelemetry.io/docs/specs/otlp/.

albertlockett:
It looks like the OTLP gRPC exporter currently treats everything as retryable. It constructs the NackMsg using the `new` constructor which always sets `permanent` false
https://github.com/open-telemetry/otel-arrow/blob/4f522d2e9dff85834790fe8aba98e28340df21eb/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs#L439-L454

The OTLP gRPC spec has well defined retry behaviour for each status code that should be followed:
https://opentelemetry.io/docs/specs/otlp/#failures


gRPC Code | Retryable?
-- | --
CANCELLED | Yes
UNKNOWN | No
INVALID_ARGUMENT | No
DEADLINE_EXCEEDED | Yes
NOT_FOUND | No
ALREADY_EXISTS | No
PERMISSION_DENIED | No
UNAUTHENTICATED | No
RESOURCE_EXHAUSTED | Only if the server can recover (see below)
FAILED_PRECONDITION | No
ABORTED | Yes
OUT_OF_RANGE | Yes
UNIMPLEMENTED | No
INTERNAL | No
UNAVAILABLE | Yes
DATA_LOSS | Yes

The OTLP HTTP exporter already has the logic to handle this retry correctly based on status codes:
https://github.com/open-telemetry/otel-arrow/blob/4f522d2e9dff85834790fe8aba98e28340df21eb/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/mod.rs#L550-L592

PR #1956, opened 2026-02-04
Add experimental resource validator processor

lalitb:
fixes: #1941 

 ### Description:

   Adds an experimental `resource_validator_processor` that validates telemetry resources against an allowlist.

   **Behavior:**
   - Validates `microsoft.resourceId` (or configurable attribute) exists on resources
   - Checks value against allowed list (case-insensitive option)
   - NACKs with permanent error on validation failure (HTTP 400 / gRPC INVALID_ARGUMENT)

   **Config:**
   ```yaml
   processors:
     resource_validator:
       required_attribute: "microsoft.resourceId"
       allowed_values:
         - "/subscriptions/xxx/resourceGroups/yyy/..."
       case_insensitive: true
```

  **Feature flag**: resource-validator-processor

  **Status**: Static config only. Extensible for future dynamic auth context support.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
@@ -423,11 +423,7 @@ impl ResourceValidatorProcessor {
     }
 
     /// Updates metrics and logs warnings based on validation result
-    fn update_metrics(
-        &mut self,
-        result: &Result<(), (ValidationFailure, String)>,
-        num_items: u64,
-    ) {
+    fn update_metrics(&mut self, result: &Result<(), (ValidationFailure, String)>, num_items: u64) {
         match result {
             Ok(()) => {
                 self.metrics.batches_accepted.add(1);

### CODE REVIEW

File: rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs [hunk 1]
```diff
@@ -0,0 +1,777 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! Resource Validator Processor
+//!
+//! This processor validates that a required resource attribute exists and its value
+//! is in an allowed list. Requests that fail validation are permanently NACKed,
+//! enabling clients to detect misconfiguration immediately rather than having data
+//! silently dropped.
+//!
+//! # Use Case
```
cijothomas:
```suggestion
//! # Example Use Case
```

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,777 @@
...
+                let str_value = std::str::from_utf8(str_bytes)
+                    .map_err(|_| ValidationFailure::InvalidAttributeType)?;
+
+                // Skip allowed list check if no values are configured (presence-only validation)
+                if allowed_values.is_empty() {
+                    return Ok(());
+                }
+
+                // Check if value is in allowed list
+                let lookup_value = if self.case_insensitive {
```
cijothomas: nit: we don't need owned lookup_value for the contains check, so maybe possible to avoid the String allocation?

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/metrics.rs
```diff
@@ -0,0 +1,24 @@
...
+//! Metrics for the Resource Validator Processor
+
+use otap_df_telemetry::instrument::Counter;
+use otap_df_telemetry_macros::metric_set;
+
+/// Metrics collected by the Resource Validator Processor
+#[metric_set(name = "resource_validator.processor.metrics")]
+#[derive(Debug, Default, Clone)]
+pub struct ResourceValidatorMetrics {
+    /// Number of batches that passed validation
```
cijothomas, line 17: its intentional we count the number of batches, and not the count of telemetry items inside that batch right?

lalitb: Good question, yes this is intentional - have added the comment to clarify.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
lalitb: Good point. Done.

lalitb:
@AaronRM  - I’m seeing some intermittent CI failures related to durable_buffer_processor test. This PR doesn’t touch that code, and the failures went away on rerun. Flagging it in case you’ve seen this before.

```
**_for_attributes_no_double_encoding
        PASS [   0.012s] (1086/1780) otap-df-pdata otlp::attributes::test::test_default_anyvalue_encoded_when_column_missing
        PASS [   0.013s] (1087/1780) otap-df-pdata otlp::batching_tests::test_corrupted_protobuf_handling
        PASS [   0.106s] (1088/1780) otap-df-pdata otlp::batching_tests::test_simple_batch_logs
        PASS [   0.447s] (1089/1780) otap-df-pdata otlp::batching_tests::test_simple_batch_metrics
        FAIL [  11.518s] (1090/1780) otap-df-otap::durable_buffer_processor_tests test_durable_buffer_recovery_after_outage
  stdout ───

    running 1 test
    test test_durable_buffer_recovery_after_outage ... FAILED

    failures:

    failures:
        test_durable_buffer_recovery_after_outage

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 6 filtered out; finished in 11.49s
    
  stderr ───

    thread 'test_durable_buffer_recovery_after_outage' (6904) panicked at crates\otap\tests\durable_buffer_processor_tests.rs:570:5:
    Recovery should deliver at least 35 items (25+10), got 20
```

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
```diff
@@ -0,0 +1,168 @@
...
+///   resource_validator:
+///     required_attribute: "microsoft.resourceId"
+///     allowed_values:
+///       - "/subscriptions/xxx/resourceGroups/yyy/..."
+///     case_insensitive: true
+/// ```
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Config {
+    /// The resource attribute key that must be present on all resources.
+    /// Default: "microsoft.resourceId"
```
cijothomas: don't think a default of microsoft.resourceId should be used given the processor is designed to be quite generic.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    /// List of allowed values for the required attribute.
+    /// Empty list rejects all values.
+    #[serde(default)]
+    pub allowed_values: Vec<String>,
+
+    /// Whether to perform case-insensitive comparison of attribute values.
+    /// Default: false
+    #[serde(default)]
+    pub case_insensitive: bool,
```
cijothomas: my personal preference is `case_sensitive` with default(true) - avoids double negative.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
```diff
@@ -0,0 +1,168 @@
...
+/// ```
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Config {
+    /// The resource attribute key that must be present on all resources.
+    /// Default: "microsoft.resourceId"
+    #[serde(default = "default_required_attribute")]
+    pub required_attribute: String,
+
+    /// List of allowed values for the required attribute.
+    /// Empty list rejects all values.
```
cijothomas, line 32: Could this be used to merely check the presence of an attribute? I am not seeing how, as allowed_values is not `Option`...

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
lalitb: Yeah, I thought about that. There is no equivalent processor in Go collector main and contributor repo - which means there is no such use-case/need outside. So I thought this can be another of the MS specific component similar to Geneva and Azure Monitor exporter - currently in experimental and eventually in contrib repo. But no strong opinion, happy not to keep this default :)

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
lalitb: Kept it on parity with the Microsoft internal processor used for same purpose.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1131 @@
...
+    /// context is not available for a particular request.
+    #[allow(dead_code)]
+    Dynamic {
+        /// Fallback values from config (used when auth context is unavailable)
+        fallback: HashSet<String>,
+    },
+}
+
+impl AllowedValuesSource {
+    /// Gets the static/fallback allowed values set.
```
cijothomas: nit: Somewhat misleading - since the method name is static, lets just return the static one, and don't bother with Dynamic at all.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1131 @@
...
+    ConversionError,
+}
+
+impl std::fmt::Display for ValidationFailure {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ValidationFailure::MissingAttribute => write!(f, "missing"),
+            ValidationFailure::InvalidAttributeType => write!(f, "invalid_type"),
+            ValidationFailure::NotInAllowedList => write!(f, "not_allowed"),
+            ValidationFailure::ConversionError => write!(f, "conversion_error"),
```
cijothomas:
```suggestion
            ValidationFailure::ConversionError => write!(f, "conversion_error(internal)"),
```

nit: something to make it clear that user is not at fault here and Arrow engine is at fault!

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1131 @@
...
+        let result = validator.validate_logs(&data);
+        assert!(matches!(
+            result,
+            Err((ValidationFailure::NotInAllowedList, _))
+        ));
+    }
+
+    #[test]
+    fn test_validate_logs_invalid_attribute_type() {
+        let logs_bytes = create_logs_request_with_resource(vec![KeyValue::new(
```
cijothomas, line 866: lets add a test for the edge case of resource.attribute being empty in the input.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1131 @@
...
+    use bytes::Bytes;
+    use otap_df_pdata::proto::opentelemetry::{
+        collector::logs::v1::ExportLogsServiceRequest,
+        common::v1::{AnyValue, InstrumentationScope, KeyValue},
+        logs::v1::{LogRecord, ResourceLogs, ScopeLogs, SeverityNumber},
+        resource::v1::Resource,
+    };
+    use prost::Message as ProstMessage;
+
+    /// Test helper struct for validation testing without metrics
```
cijothomas, line 563: mm.. with this test, are we really validating the product code itself?

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
cijothomas: do we need to keep parity or use this opportunity to fix double negatives? (They are generally avoided in OTel from what I see... Not a big concern but if we can avoid it, that's great.)

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
cijothomas: then it maybe better to make this specific to Azure and rename to AzureResourceIdChecker/similar. Having a generic processor with Azure (or any vendor) specific default is incorrect.

lalitb: Make sense. I have removed the default, and made "required_attribute" a mandatory config parameter.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
lalitb: Yeah, thinking more "case_sensitive true" would be better. Have done the change now.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
lalitb: Good point! Currently this isn't supported - `allowed_values` is required and an empty list is rejected during validation.

We could add a "presence-only" mode in the future by making `allowed_values` optional - if omitted, only validate that the attribute exists without checking its value.

  This would be backward-compatible since existing configs with `allowed_values` would continue working unchanged. Only currently-invalid configs (those missing the field) would gain new meaning. Adding this later if there's demand.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
lalitb: Good point! You're right that `TestValidator` reimplements the validation logic rather than testing `ResourceValidatorProcessor` directly. That said, these tests still provide value as they validate the core validation algorithm, document expected behavior for edge cases (empty attributes, wrong types, case sensitivity), and catch regressions if the logic changes. The gap is that changes to `ResourceValidatorProcessor` without updating `TestValidator` wouldn't be caught. 

 The difficulty in using `ResourceValidatorProcessor` directly is that it requires a `PipelineContext` for metrics registration. Setting up a valid test `PipelineContext` needs additional test infrastructure that doesn't exist yet.

 I have added a `TODO` noting this limitation, and we can refactor to test the real processor once we have test helpers for PipelineContext set up.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
lalitb: Done. thanks.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
lalitb: Good point! Renamed the method from `get_static_values()` to `get_allowed_values()` to better reflect what it returns. Kept the Dynamic variant since we'll be adding SAT auth support soon - will avoid churn of removing and re-adding it.

AaronRM:
> @AaronRM - I’m seeing some intermittent CI failures related to durable_buffer_processor test. This PR doesn’t touch that code, and the failures went away on rerun. Flagging it in case you’ve seen this before.

@lalitb Thanks for the heads up. I think the flakiness is *probably* related to the lack of WAL replay (which is being addressed in #1954).

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
```diff
@@ -0,0 +1,150 @@
...
+    #[serde(default)]
+    pub allowed_values: Vec<String>,
+
+    /// Whether to perform case-sensitive comparison of attribute values.
+    /// Default: true
+    #[serde(default = "default_case_sensitive")]
+    pub case_sensitive: bool,
```
cijothomas, line 41: nit: Do an explicit callout what `required_attribute` is not affected by this and is always case_sensitive.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+                if let Err(failure) = self.validate_resource_with_allowed(&resource, allowed_values)
+                {
+                    return Err((failure, self.format_error_message(failure)));
+                }
+            } else {
+                let failure = ValidationFailure::MissingAttribute;
+                return Err((failure, self.format_error_message(failure)));
+            }
+        }
+        Ok(())
```
jmacd, line 395: are these the same as 325-337?
if yes, maybe refactor (3x); if not, feel free to ignore :)

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+                            // but pass through rather than fail
+                            Ok(())
+                        }
+                    },
+                    OtapPayload::OtapArrowRecords(arrow_records) => match signal_type {
+                        SignalType::Logs => {
+                            self.validate_arrow_logs(arrow_records, &allowed_values)
+                        }
+                        // Metrics/Traces Arrow views not yet available - convert to OTLP
+                        // TODO: Implement OtapMetricsView/OtapTracesView to avoid clone + conversion
```
jmacd, line 503: nit: if we have an issue number already, please link it, otherwise please create a new issue in the repo!

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+        arrow_records: &OtapArrowRecords,
+        allowed_values: &HashSet<String>,
+    ) -> Result<(), (ValidationFailure, String)> {
+        let logs_view = OtapLogsView::try_from(arrow_records).map_err(|_| {
+            let failure = ValidationFailure::ConversionError;
+            (failure, self.format_error_message(failure))
+        })?;
+
+        for resource_logs in logs_view.resources() {
+            if let Some(resource) = resource_logs.resource() {
```
jmacd, line 385: I can imagine that if we wanted to, that we could optimize the Arrow-based implementation of this logic in a way that would be faster, maybe a single pass over the ResourceAttrs table. @albertlockett could probably guide us. :-)

jmacd (approved):

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/README.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Behavior
+
+| Condition                                    | Result                    |
+| -------------------------------------------- | ------------------------- |
+| Attribute present with value in allowed list | Pass through              |
+| Attribute present, empty allowed_values list | Permanent NACK - HTTP 400 |
+| Attribute missing                            | Permanent NACK - HTTP 400 |
+| Attribute wrong type (not string)            | Permanent NACK - HTTP 400 |
+| Attribute value not in allowed list          | Permanent NACK - HTTP 400 |
```
utpilla: Something to keep in mind that this is not done currently. I believe the receiver needs to be updated to check the NACK details and return a 400. I think as of today, it only returns 503.

rust/otap-dataflow/crates/otap/Cargo.toml
```diff
@@ -129,6 +129,7 @@ recordset-kql-processor = [
     "dep:data_engine_recordset_otlp_bridge",
     "dep:data_engine_recordset"
 ]
+resource-validator-processor = ["experimental-processors"]
```
utpilla, line 132: Define this feature at `otap-dataflow` level as well similar to other experimental processors:

https://github.com/open-telemetry/otel-arrow/blob/e95eee989bfc60af479e1e780b82f76b0702a897/rust/otap-dataflow/Cargo.toml#L218-L221

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+    fn update_metrics(&mut self, result: &Result<(), (ValidationFailure, String)>) {
+        match result {
+            Ok(()) => {
+                self.metrics.batches_accepted.add(1);
+            }
+            Err((failure, msg)) => {
+                warn!(reason = %failure, "{}", msg);
+                match failure {
+                    ValidationFailure::MissingAttribute
+                    | ValidationFailure::InvalidAttributeType
```
ThomsonTan: Instead of tracking `InvalidAttributeType` and `ConversionError` in `batches_rejected_missing`, could create dedicated metrics to track them?

utpilla:
> LGTM. One minor point - this looks quite specific for Azure, so maybe rename it to be Azure specific processor. If there is interest to have a general purpose one, then we can add such a one in future too. Not a blocker, still in experimental feature only.

Good point. If the interest is to have a general-purpose processor, then we should probably consider checking non-string values as well and possibly look for multiple resource attributes.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/README.md
jmacd: See #1920 #1921

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
```diff
@@ -0,0 +1,150 @@
...
+///     required_attribute: "cloud.resource_id"
+///     allowed_values:
+///       - "/subscriptions/xxx/resourceGroups/yyy/..."
+///     case_sensitive: false  # optional, defaults to true
+/// ```
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Config {
+    /// The resource attribute key that must be present on all resources.
+    /// This is a required field with no default.
+    pub required_attribute: String,
```
utpilla: We could be more specific here.

```suggestion
    pub required_attribute_key: String,
```

jmacd:
I think qualifies as a general purpose component, but we can start with it in experimental.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
```diff
@@ -0,0 +1,150 @@
...
+    #[serde(default)]
+    pub allowed_values: Vec<String>,
+
+    /// Whether to perform case-sensitive comparison of attribute values.
+    /// Default: true
+    #[serde(default = "default_case_sensitive")]
+    pub case_sensitive: bool,
+}
+
+fn default_case_sensitive() -> bool {
```
utpilla:
```suggestion
const fn default_case_sensitive() -> bool {
```

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/config.rs
```diff
@@ -0,0 +1,150 @@
...
+    true
+}
+
+impl Config {
+    /// Validates the configuration.
+    pub fn validate(&self) -> Result<(), ConfigError> {
+        if self.required_attribute.is_empty() {
```
utpilla: Check for whitespace as well?

```suggestion
        if self.required_attribute.trim().is_empty() {
```

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+    /// indicate that allowed values should be read from the request context.
+    ///
+    /// The optional HashSet provides fallback values from config if auth
+    /// context is not available for a particular request.
+    #[allow(dead_code)]
+    Dynamic {
+        /// Fallback values from config (used when auth context is unavailable)
+        fallback: HashSet<String>,
+    },
+}
```
utpilla: This is a minor one. If both the variants would have the same data, that is, `HashSet<String>`, then it would be simpler to just move the `HashSet` to the processor struct and make the `enum` simpler:

```rust
pub enum AllowedValuesSource {
    /// Use only the static config values
    Static,
    /// Check auth context first, fall back to config values
    Dynamic,
}

pub struct ResourceValidatorProcessor {
    required_attribute: String,
    /// Pre-normalized allowed values (used as-is for Static, as fallback for Dynamic)
    allowed_values: HashSet<String>,
    /// Where to get allowed values from
    source_mode: AllowedValuesSource,
    case_sensitive: bool,
    metrics: MetricSet<ResourceValidatorMetrics>,
}
```

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+
+impl AllowedValuesSource {
+    /// Gets the allowed values set (static config or fallback for dynamic).
+    fn get_allowed_values(&self) -> &HashSet<String> {
```
utpilla:
```suggestion
    const fn get_allowed_values(&self) -> &HashSet<String> {
```

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+    ///                 }
+    ///             }
+    ///             // Fall back to config
+    ///             Cow::Borrowed(fallback)
+    ///         }
+    ///         AllowedValuesSource::Static(values) => Cow::Borrowed(values),
+    ///     }
+    /// }
+    /// ```
+    fn get_allowed_values(&self, _pdata: &OtapPdata) -> Cow<'_, HashSet<String>> {
```
utpilla, line 251:
```suggestion
    const fn get_allowed_values(&self, _pdata: &OtapPdata) -> Cow<'_, HashSet<String>> {
```

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+                // Empty allowed_values rejects all values.
+                // Case-sensitive: zero allocation with Cow::Borrowed.
+                // Case-insensitive: allocates via to_lowercase() for O(1) HashSet lookup.
+                let lookup_value: Cow<'_, str> = if self.case_sensitive {
+                    Cow::Borrowed(str_value)
+                } else {
+                    Cow::Owned(str_value.to_lowercase())
+                };
+
+                if allowed_values.contains(lookup_value.as_ref()) {
```
utpilla, line 295: Not for this PR, but I think the expected usage of this processor (tracking tenants) wouldn't have a lot of allowed values. It would be more performant to just rely on simple comparisons against a vec of allowed values instead of using HashSet which would come at a cost of always computing hash and doing the equality check.

Using a vec would also allow you to do case insensitive check without allocation by using [`to_lowercase`](https://doc.rust-lang.org/std/primitive.char.html#method.to_lowercase) iterator from `char`

```rust
// allowed_values is a vec in this example
// Case-sensitive: simple equality
allowed_values.iter().any(|v| v == str_value)

// Case-insensitive: iterator-based, no allocation
allowed_values.iter().any(|allowed| {
    allowed.chars().flat_map(char::to_lowercase)
        .eq(str_value.chars().flat_map(char::to_lowercase))
})
```

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/metrics.rs
```diff
@@ -0,0 +1,28 @@
...
+//! Metrics for the Resource Validator Processor
+
+use otap_df_telemetry::instrument::Counter;
+use otap_df_telemetry_macros::metric_set;
+
+/// Metrics collected by the Resource Validator Processor.
+///
+/// Note: Metrics are counted at the batch level, not per telemetry item.
+/// This is intentional because validation is pass/fail for the entire batch -
+/// if any resource in a batch fails validation, the whole batch is NACKed.
```
utpilla: It would be better to track it a telemetry item level. You should call `num_items()` on the pdata batch at least for the failed metrics so that we know how many logs etc. were rejects. Tracking the batch counts erases the magnitude of data loss.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
```diff
@@ -0,0 +1,1157 @@
...
+    /// These are pre-normalized (lowercased if case_sensitive is false).
+    Static(HashSet<String>),
+
+    /// Dynamic allowed values from auth context (future).
+    /// When SAT auth extension is ready, this variant will be used to
+    /// indicate that allowed values should be read from the request context.
+    ///
+    /// The optional HashSet provides fallback values from config if auth
+    /// context is not available for a particular request.
+    #[allow(dead_code)]
```
ThomsonTan, line 100: Comment out dynamic variant until it's actually needed? I think another change to remove the `#[allow(dead_code)]` is needed anyway.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
lalitb: Yeah, I noticed this duplication too. The three view traits (`ResourceLogsView`, `ResourceMetricsView`, `ResourceSpansView`) each have their own Resource GAT with no common super-trait, so a generic helper doesn't work cleanly without introducing one in `pdata`. Didn't feel worth touching that foundational layer for a single consumer - added a comment explaining why. Happy to revisit if we see more use cases for signal-generic resource iteration.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/README.md
lalitb: Good point. Updated the README table and mod.rs doc comment to reflect current behavior - permanent NACK without the HTTP 400 claim, plus a note explaining the receiver-side gap.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
lalitb: Good point. done. thanks

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
lalitb: `Cow::Borrowed()` isn't allowed in const fn contexts (it requires trait methods). This one can't be made const

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
lalitb: Good point about avoiding the allocation for case-insensitive comparison. Though but if plan to have it as a general-purpose validator (not just tenant tracking), the allowed list could potentially be large, so I'd want to keep efficient lookup. Maybe a hybrid approach - Vec for small lists, HashSet above a threshold - or at least the zero-allocation case-insensitive iterator trick. Will look into it as a follow-up.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/metrics.rs
lalitb: Good call - added `items_accepted` and `items_rejected` counters using `pdata.num_items()` to capture the magnitude of data loss on rejection.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
lalitb: Fair point. Removing `#[allow(dead_code)]` will be easy once SAT auth lands. I’m inclined to keep `Dynamic` for now since it makes the planned extension obvious in the code, keeps the struct and `get_allowed_values()` ready for SAT, and serves as a reminder that this path is intentional, not hypothetical. Happy to drop it if folks feel strongly.

lalitb:
> > LGTM. One minor point - this looks quite specific for Azure, so maybe rename it to be Azure specific processor. If there is interest to have a general purpose one, then we can add such a one in future too. Not a blocker, still in experimental feature only.
> 
> Good point. If the interest is to have a general-purpose processor, then we should probably consider checking non-string values as well and possibly look for multiple resource attributes.


Thanks for the feedback! The processor itself is fully generic - with no Azure-specific logic. The use case that motivated it happens to be an Azure scenario , but the same pattern applies to any multi-tenant environment.

I think keeping the generic name makes sense since the implementation doesn't have anything Azure-specific in it. The reason it's under experimental (and also under feature flag) is that there's no equivalent processor in the Go collector, so moving it to core would mean shipping it in the binary without any existing users outside our use case.

 Great point about non-string values and multiple attribute keys - I think those would be wonderful additions that could be contributed by folks who have those use cases. Keeping the current scope minimal and letting the community extend it as needed feels like a nice collaborative path forward.

rust/otap-dataflow/crates/otap/src/experimental/resource_validator_processor/mod.rs
albertlockett: Hey -- Sorry it took me a few days to reply to this, I was on vacation 🌴 !

This is something we might be able to optimize using Arrow compute kernels, but performing this check isn't trivial. If we did implement this, it would also be good to benchmark the against the existing approach for different batch sizes.

For each row, there's basically three things we need to check:
- key == required_attribute_key
- type == AttributeValueType::Str
- value in (allowed_values)

So we could do something like:
```rs
use otap_df_pdata::{
  otlp::attributes::AttributeValueType,
  schema::consts
};
use arrow::compute::kernels::{boolean::{and, or}, cmp::eq, filter::filter};

// TODO handle missing columns & other errors correctly instead of using unwrapping
let type_col = resource_attrs_record_batch.column_by_name(consts::ATTRIBUTE_TYPE).unwrap();
let key_col = reousrce_attrs_record_batch.column_by_name(conts::ATTRIBUTE_KEY).unwrap();

let has_type = eq(type_col, &UInt8Array::new_scalar(AttributeValuesType::Str as u8).unwrap();
let has_key = eq(key_col, &StringArray::new_scalar(required_attribute_key).unwrap();

// Note - I'm unwrapping this, but str column is actually optional - if it's not present, 
// it means there are no non-null/empty string attributes, which would effectively mean 
// we reject this batch, unless empty string is one of allowed_values the allowed values,
// then I guess we accept it?
let str_col = reousrce_attrs_record_batch.column_by_name(consts::ATTRIBUTE_STR).unwrap();

// here, we compare against every row in the values column.. Depending on the cardinality, 
// it might actually be faster to prefilter down str_col using and(has_type, has_key) and compare
// each value of the prefiltered str_col against each allowed value. Clearly that would change how
// we take the parent_id column below (we'd just do valid_rows = filter(&parent_ids, &has_val))
let mut has_val = BooleanArray::new_unset(str_col.len());
for val in allowed_values {
  let has_this_val = eq(str_col, &StringArray::new_scalar(val)).unwrap();
  has_val = or(&has_val, &has_this_val).unwrap();
}

// combine filter results together - this will be a boolean array where the rows represent the
// rows that have valid resource attribtues
let valid_rows = and(&has_type, and(&has_key, &has_val).unwrap()).unwrap();

//
let parent_ids = resource_attrs_record_batch.column_by_name(consts::PARENT_ID).unwrap();
let valid_resource_ids = filter(&parent_ids, &valid_rows).unwrap();
```

One you have the valid resource IDs, the next step would be to get the Resource ID column from the logs/traces/metrics record batch (it's inside a struct column), and ensure that every row contains a non-null value that is in the valid_resource_ids array.

Typically, we've done this by collecting `valid_resource_ids` into a Roaring Bitmap. 

```
let id_mask: RoaringBitmap = valid_resource_ids
    .as_any()
    .downcast_ref::<UInt16Array>()
    .unwrap()
    .iter().flatten().map(|i| i.into()).collect();
let resource_ids = root_rb
                .column_by_name(consts::RESOURCE)
                .and_then(|arr| arr.as_any().downcast_ref::<StructArray>())
                .and_then(|arr| arr.column_by_name(consts::ID));

// then do something like:
// for id in resource_ids {
//     if id.is_none() |!id_mask.contains(id) { return Err }
// }
```

Note: if the resource ID column contains a null value, it means the log/trace/metric has no resource attributes, which I guess would mean it fails the validity check.

albertlockett: Hey -- Sorry it took me a few days to reply to this, I was on vacation 🌴 !

This is something we might be able to optimize using Arrow compute kernels, but performing this check isn't trivial. If we did implement this, it would also be good to benchmark the against the existing approach for different batch sizes.

For each row, there's basically three things we need to check:
- key == required_attribute_key
- type == AttributeValueType::Str
- value in (allowed_values)

So we could do something like:
```rs
use otap_df_pdata::{
  otlp::attributes::AttributeValueType,
  schema::consts
};
use arrow::compute::kernels::{boolean::{and, or}, cmp::eq, filter::filter};

// TODO handle missing columns & other errors correctly instead of using unwrapping
let type_col = resource_attrs_record_batch.column_by_name(consts::ATTRIBUTE_TYPE).unwrap();
let key_col = reousrce_attrs_record_batch.column_by_name(conts::ATTRIBUTE_KEY).unwrap();

let has_type = eq(type_col, &UInt8Array::new_scalar(AttributeValuesType::Str as u8).unwrap();
let has_key = eq(key_col, &StringArray::new_scalar(required_attribute_key).unwrap();

// Note - I'm unwrapping this, but str column is actually optional - if it's not present, 
// it means there are no non-null/empty string attributes, which would effectively mean 
// we reject this batch, unless empty string is one of allowed_values the allowed values,
// then I guess we accept it?
let str_col = reousrce_attrs_record_batch.column_by_name(consts::ATTRIBUTE_STR).unwrap();

// here, we compare against every row in the values column.. Depending on the cardinality, 
// it might actually be faster to prefilter down str_col using and(has_type, has_key) and compare
// each value of the prefiltered str_col against each allowed value. Clearly that would change how
// we take the parent_id column below (we'd just do valid_rows = filter(&parent_ids, &has_val))
let mut has_val = BooleanArray::new_unset(str_col.len());
for val in allowed_values {
  let has_this_val = eq(str_col, &StringArray::new_scalar(val)).unwrap();
  has_val = or(&has_val, &has_this_val).unwrap();
}

// combine filter results together - this will be a boolean array where the rows represent the
// rows that have valid resource attribtues
let valid_rows = and(&has_type, and(&has_key, &has_val).unwrap()).unwrap();

//
let parent_ids = resource_attrs_record_batch.column_by_name(consts::PARENT_ID).unwrap();
let valid_resource_ids = filter(&parent_ids, &valid_rows).unwrap();
```

One you have the valid resource IDs, the next step would be to get the Resource ID column from the logs/traces/metrics record batch (it's inside a struct column), and ensure that every row contains a non-null value that is in the valid_resource_ids array.

Typically, we've done this by collecting `valid_resource_ids` into a Roaring Bitmap. 

```
let id_mask: RoaringBitmap = valid_resource_ids
    .as_any()
    .downcast_ref::<UInt16Array>()
    .unwrap()
    .iter().flatten().map(|i| i.into()).collect();
let resource_ids = root_rb
                .column_by_name(consts::RESOURCE)
                .and_then(|arr| arr.as_any().downcast_ref::<StructArray>())
                .and_then(|arr| arr.column_by_name(consts::ID));

// then do something like:
// for id in resource_ids {
//     if id.is_none() |!id_mask.contains(id) { return Err }
// }
```

Note: if the resource ID column contains a null value, it means the log/trace/metric has no resource attributes, which I guess would mean it fails the validity check.

albertlockett: BTW, the [filter](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs) module in columnar query-engine already has a bunch of code for doing this kind of attribute filter & joining the parent_ids back to the main record batch.

Unfortunately what we don't yet have would be a mechanism to apply this filter & then reject the entire batch, nor any way to expose that behaviour via OPL. 

I might spend some time this week to figure out how we'd do that. I think it would be a good opportunity to improve OPL/Columnar Query Engine w/ this real world use case.

albertlockett: @jmacd @lalitb (cc @lquerel): We might be able to use OPL/columnar query engine to check if the batch is valid in the case where the signal is OTAP. This would save a round-trip from OTAP to OTLP for metrics/traces (since we don't yet have OTAP backed views implemented signal types), and the filtering might be faster (although it would be good to benchmark to confirm).

Below is an example of a pipeline that we could invoke to check if the signal is valid. 

The pipeline will basically split the batch, and if there's any rows that have an invalid resource ID, it will try to "route" them to a route called "invalid". We then provide a custom "Router" that checks if it receives any rows, and sets a flag that some rows were invalid. We then make the Nack/Forward decision based on this flag.
```rust
  use data_engine_kql_parser::Parser;
  use otap_df_opl::parser::OplParser;
  use otap_df_query_engine::pipeline::{
        Pipeline,
        routing::{RouteName, Router, RouterExtType},
        state::ExecutionState,
    };
    
  let pipeline_expr = OplParser::parse(r#"
      signals |
      if (not(
          resource.attributes["microsoft.resourceId"] == "value1" or
          resource.attributes["microsoft.resourceId"] == "value2"
      )) {
          route_to "invalid"
      }
  "#).unwrap().pipeline;

  let mut pipeline = Pipeline::new(pipeline_expr);

  // "Router" impl that just sets the valid flag to false if anything is routed
  struct ValidityRouter {
      valid: bool
  }

  #[async_trait(?Send)]
  impl Router for ValidityRouter {
      fn as_any(&self) -> &dyn std::any::Any {
          self
      }

      fn as_any_mut(&mut self) -> &mut dyn std::any::Any {
          self
      }

      async fn send(&mut self,
          _route_name: RouteName,
          otap_batch: OtapArrowRecords
      ) -> otap_df_query_engine::error::Result<()> {
          let is_empty = otap_batch.root_record_batch().is_none();

          if !is_empty {
              self.valid = false;
          }
          Ok(())
      }
  }

  let mut exec_state = ExecutionState::new();
  exec_state.set_extension::<RouterExtType>(Box::new(ValidityRouter{ valid: true }));

  // batch 1 (valid)
  let logs_bytes = create_logs_request_with_resource(vec![KeyValue::new(
      "microsoft.resourceId",
      AnyValue::new_string("value1"),
  )]);
  let pdata: OtapPayload = OtlpProtoBytes::ExportLogsRequest(logs_bytes).into();
  let otap_batch1: OtapArrowRecords = pdata.try_into().unwrap();

  // batch 2 (not valid)
  let logs_bytes = create_logs_request_with_resource(vec![KeyValue::new(
      "microsoft.resourceId",
      AnyValue::new_string("value_not_valid"),
  )]);
  let pdata: OtapPayload = OtlpProtoBytes::ExportLogsRequest(logs_bytes).into();
  let otap_batch2: OtapArrowRecords = pdata.try_into().unwrap();

  // batch 3 (valid)
  let logs_bytes = create_logs_request_with_resource(vec![KeyValue::new(
      "microsoft.resourceId",
      AnyValue::new_string("value2"),
  )]);
  let pdata: OtapPayload = OtlpProtoBytes::ExportLogsRequest(logs_bytes).into();
  let otap_batch3: OtapArrowRecords = pdata.try_into().unwrap();

  for otap_batch in [otap_batch1, otap_batch2, otap_batch3] {
      let _ = pipeline.execute_with_state(otap_batch, &mut exec_state).await.unwrap();

      let router_impl = exec_state
          .get_extension_mut::<RouterExtType>()
          .and_then(|router| router.as_any_mut().downcast_mut::<ValidityRouter>())
          .unwrap();

      if !router_impl.valid {
          // reset validity flag
          router_impl.valid = true;
          println!("not valid (NACK)");

      } else {
          println!("valid (send to default out port)");
      }
  }
```

Prints:
```
valid (send to default out port)
not valid (NACK)
valid (send to default out port)
```

I think the major caveat here is that building the actual OPL query is a bit tricky, as we don't yet support any kind of parameterized pipelines (certainly its something we can/should add soon). Doing something like the following might open us up to injection attacks if we don't trust the processor config:
```rust
  let pipeline_expr = OplParser::parse(&format!(
      r#"
      signals |
      if (not(
          {}
      )) {{
          route_to "invalid"
      }}
  "#,
      allowed_values
          .iter()
          .map(|val| {
              format!(
                  "resource.attributes[\"microsoft.resourceId\"] == \"{}\"",
                  val
              )
          })
          .collect::<Vec<String>>()
          .join(" or ")
  ))
  .unwrap()
  .pipeline;
```

To get around this, we could technically build the pipeline expression manually using AST Expression types in the query_engine/expressions crate
https://github.com/open-telemetry/otel-arrow/tree/main/rust/experimental/query_engine/expressions

PR #1961, opened 2026-02-05
[otap-df-quiver] Implement time-based segment retention (max_age) for quiver & durable_buffer processor

### LINKED ISSUE #1960: [otap-df-quiver] Need to implement time-based segment retention (max_age) for quiver & durable_buffer processor
The Quiver storage engine currently supports size-based retention (`size_cap_bytes` with `DropOldest` policy), but the `max_retain_after_ingestion_hours` setting mentioned in the Quiver ARCHITECTURE.md document has not yet been implemented. This means segments can accumulate indefinitely if subscribers are slow or offline, even when the data is no longer useful. Users need a way to automatically expire old data after a configurable time period, regardless of whether subscribers have consumed it. This is important for compliance (data retention policies), resource management (preventing unbounded disk growth when subscribers lag), and operational simplicity (automatic cleanup of stale data).

Additionally, the `durable_buffer` processor already exposes a `max_age` configuration option, but it was marked as a placeholder and not wired through to the underlying quiver engine. The implementation should be opt-in to avoid unexpected data loss for existing deployments, and the `durable_buffer` processor should be updated to pass the configured `max_age` to quiver.

(I have a branch which I will be submitting for PR shortly which resolves these issues.)

AaronRM:
# Change Summary

This PR implements time-based segment retention (`max_age`) for the quiver storage engine, allowing segments to be automatically deleted after a configurable duration regardless of subscriber consumption status. The feature is *opt-in* (`max_age: None` by default) to avoid unexpected data loss. Segments are timestamped using file modification time when finalized, and expired segments are cleaned up both during startup (without loading them) and during periodic maintenance. The implementation coordinates with the subscriber registry to force-complete expired segments before deletion, ensuring subscribers don't attempt to read from deleted files.

Also updates the `durable_buffer` processor to pass its existing `max_age` config option through to quiver, replacing the previous placeholder implementation.

* Closes #1960 

Comprehensive unit tests cover the new functionality.

After this change, the user-facing `max_age` setting on the `durable_buffer` processor will work as expected. (A `max_age` setting is being added to the Quiver configuration.)

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/engine.rs
+++ b/rust/otap-dataflow/crates/quiver/src/engine.rs
@@ -37,7 +37,9 @@ use crate::subscriber::{
     SubscriberId, SubscriberRegistry,
 };
 use crate::telemetry::PersistenceMetrics;
-use crate::wal::{WalConsumerCursor, WalWriter, WalWriterOptions};
+use crate::wal::{
+    CursorSidecar, MultiFileWalReader, ReplayBundle, WalConsumerCursor, WalWriter, WalWriterOptions,
+};
 
 /// WAL statistics for observability.
 #[cfg(test)]
@@ -308,6 +310,8 @@ impl QuiverEngine {
             registry.force_complete_segments(&deleted_during_scan);
         }
 
+        // Start with empty open segment and default cursor
+        // WAL replay will populate these through the normal ingest path
         let engine = Arc::new(Self {
             config,
             metrics: PersistenceMetrics::new(),
@@ -361,6 +365,13 @@ impl QuiverEngine {
             }
             used_before.saturating_sub(budget.used())
         });
+
+        // Replay WAL entries that weren't finalized to segments before shutdown/crash
+        // This uses the same ingest path as live ingestion (minus WAL writes)
+        let replayed = engine.replay_wal().await?;
+        if replayed > 0 {
+            tracing::info!(replayed, "replayed WAL entries during startup");
+        }
 
         Ok(engine)
     }
@@ -482,7 +493,7 @@ impl QuiverEngine {
         // Step 1: Append to WAL for durability (if enabled)
         // Note: WAL bytes are NOT tracked in budget - they're temporary and purged after
         // segment finalization. Only segment bytes are budget-tracked.
-        if self.config.durability == DurabilityMode::Wal {
+        let cursor = if self.config.durability == DurabilityMode::Wal {
             let wal_offset = self.append_to_wal_with_capacity_handling(bundle).await?;
 
             // Track cumulative WAL bytes for throughput measurement
@@ -491,15 +502,34 @@ impl QuiverEngine {
                 .cumulative_wal_bytes
                 .fetch_add(wal_entry_bytes, Ordering::Relaxed);
 
-            // Step 2: Update cursor to include this entry
-            let cursor = WalConsumerCursor::from_offset(&wal_offset);
-            {
-                let mut cp = self.segment_cursor.lock();
-                *cp = cursor;
-            }
+            Some(WalConsumerCursor::from_offset(&wal_offset))
+        } else {
+            None
+        };
+
+        // Step 2: Append to open segment and finalize if threshold exceeded
+        self.append_to_segment_and_maybe_finalize(bundle, cursor)
+            .await
+    }
+
+    /// Core ingest logic: appends a bundle to the open segment and finalizes if thresholds are met.
+    ///
+    /// This is the shared path used by both live ingestion and WAL replay.
+    /// During live ingestion, `cursor` is populated from the WAL offset.
+    /// During WAL replay, `cursor` is populated from the WAL entry being replayed.
+    #[inline]
+    async fn append_to_segment_and_maybe_finalize<B: RecordBundle>(
+        &self,
+        bundle: &B,
+        cursor: Option<WalConsumerCursor>,
+    ) -> Result<()> {
+        // Update cursor if provided (tracks WAL position for segment finalization)
+        if let Some(c) = cursor {
+            let mut cp = self.segment_cursor.lock();
+            *cp = c;
         }
 
-        // Step 3: Append to open segment accumulator
+        // Append to open segment accumulator
         let should_finalize = {
             let mut segment = self.open_segment.lock();
             let _manifest_entry = segment.append(bundle)?;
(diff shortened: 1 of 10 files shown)


rust/otap-dataflow/crates/quiver/src/segment_store.rs
```diff
@@ -304,7 +324,21 @@ impl SegmentStore {
...
+                    // On Windows, the file may still be memory-mapped by outstanding
+                    // BundleHandles. Add to pending deletes for retry later.
+                    tracing::debug!(
+                        segment = seq.raw(),
+                        "Segment file in use, deferring deletion"
+                    );
+                    let _ = self.pending_deletes.lock().insert(seq);
+                    // Don't return error - we've handled it by deferring
+                }
+                Err(e) => return Err(e),
```
lalitb, line 350: nit - Consider adding a check for std::io::ErrorKind::NotFound in the match block as well.

This handles the edge case where a file is deleted by a user or external process in the window between the path.exists() check and the remove_file call. In that scenario, the deletion will fail with a NotFound error, but this is actually a success outcome (the file is gone) and shouldn't be treated as a failure or log a warning.

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -250,10 +257,11 @@ impl QuiverEngine {
...
         ));
 
-        // Scan for existing segments from previous runs (recovery)
-        // We need to track the highest segment sequence found to avoid reusing IDs
+        // Scan for existing segments from previous runs (recovery).
+        // Pass max_age to skip loading segments that are already expired -
+        // they'll be deleted during scan without the overhead of parsing them.
         let mut next_segment_seq = 0u64;
-        match segment_store.scan_existing() {
+        match segment_store.scan_existing_with_max_age(config.retention.max_age) {
```
lalitb, line 267: I think there’s a race here between the startup cleanup and the subscriber registry -  

`scan_existing_with_max_age` deletes expired segments on disk, but `SubscriberRegistry::open` runs afterward and reloads subscriber state from `progress.json`. If a subscriber was in the middle of a segment when the process last shut down, the registry can still think that segment exists even though we just deleted it. When the subscriber resumes, it may try to read a file that’s no longer there. 

Would it would be safer to return the deleted segment IDs from `scan_existing_with_max_age` and mark them complete in the registry right after it opens, so subscribers can move on cleanly ?

lalitb, line 267: Also, there is a related issue with sequence numbering: if `scan_existing_with_max_age` deletes all segments (returning an empty list), `next_segment_seq` will default to 0.
   
This will resets the sequence counter, potentially reusing IDs that subscribers have already processed and checkpointed. We need to track the highest sequence number observed during the scan (even among deleted files) to ensure `next_segment_seq` continues monotonically.

rust/otap-dataflow/crates/quiver/src/segment_store.rs
AaronRM: Updated to handle the edge case

rust/otap-dataflow/crates/quiver/src/engine.rs
AaronRM: Yes, this is a valid point. Added handling for this so that the scan returns a ScanResult with a list of the found & deleted segments. Any deleted segments are force completed in the subscriber.

AaronRM: Also addressed in the latest commit.

rust/otap-dataflow/crates/otap/src/durable_buffer_processor/config.rs
```diff
@@ -115,8 +115,11 @@ pub struct DurableBufferConfig {
     pub retention_size_cap: Byte,
 
     /// Maximum age of data to retain (e.g., "24h", "7d").
-    /// When set, data older than this will be eligible for removal.
-    /// TODO: This setting is currently a placeholder and is not yet enforced.
+    ///
+    /// When set, segments older than this duration are automatically deleted
+    /// during maintenance, regardless of whether they have been consumed.
+    /// When `None` (the default), segments are retained indefinitely until
```
utpilla, line 124: I think we should clearly specify what "older" means.

```suggestion
    /// Age is measured from when the segment was **finalized** (written to disk),
    /// not from the timestamps within the telemetry data itself.
    ///
    /// When `None` (the default), segments are retained indefinitely until
```

rust/otap-dataflow/crates/otap/src/durable_buffer_processor/config.rs
```diff
@@ -115,8 +115,11 @@ pub struct DurableBufferConfig {
...
+    /// consumed or evicted by size-based retention.
```
utpilla, line 125: Maybe also explicitly add this guidance.

```suggestion
    /// consumed or evicted by size-based retention.
    ///
    /// To survive extended downstream outages without data loss,
    /// set `max_age` longer than the maximum outage duration you need to tolerate.
```

rust/otap-dataflow/crates/quiver/src/segment_store.rs
```diff
@@ -364,8 +548,49 @@ impl SegmentStore {
...
+    ///
+    /// Used for cleaning up expired segments during scan without the overhead
+    /// of opening and parsing the segment.
+    fn delete_segment_file(path: &Path, budget: &Option<Arc<DiskBudget>>) -> std::io::Result<()> {
+        // Get file size for budget release before deleting
+        let file_size = std::fs::metadata(path).map(|m| m.len()).ok();
+
+        // Segment files are read-only after finalization, make writable first
+        if let Ok(metadata) = std::fs::metadata(path) {
+            let mut perms = metadata.permissions();
```
utpilla: You could get file size and permissions with a `std::fs::metadata` read operation.

rust/otap-dataflow/crates/quiver/src/segment_store.rs
```diff
@@ -386,6 +611,28 @@ impl SegmentStore {
...
+        let segments = self.segments.read();
+
+        segments
+            .iter()
+            .filter_map(|(seq, handle)| {
+                // Calculate age since finalization
+                let age = now.duration_since(handle.finalized_at).ok()?;
+                if age > max_age { Some(*seq) } else { None }
+            })
+            .collect()
```
utpilla, line 623: You could precompute the cut-off time instead of doing that computation for each segment. So, instead of asking "is this segment's age greater than `max_age`?" for each segment, we flip it to ask "was this segment finalized before the cutoff time?"

Something like this:

```rust
// Pre-compute cutoff time: segments finalized before this are expired.
// This is more efficient than computing duration_since() for each segment.
let Some(cutoff) = now.checked_sub(max_age) else {
    // max_age exceeds time since UNIX_EPOCH - nothing can be expired
    return Vec::new();
};

let segments = self.segments.read();

segments
    .iter()
    .filter_map(|(seq, handle)| (handle.finalized_at < cutoff).then_some(*seq))
    .collect()
```

lalitb (approved):

rust/otap-dataflow/crates/quiver/src/segment_store.rs
AaronRM: Yes, good call. Also applied a similar optimization to other functions in segment_store.

rust/otap-dataflow/crates/quiver/src/segment_store.rs
AaronRM: Started refactoring to address this and in the process discovered we don't actually need the metadata check for file length. The new simpler equivalent method in the refactored code is `remove_readonly_file`.

jmacd (approved):

Issue #1149, opened 2025-09-18
Syslog CEF Receiver improvements


utpilla:
Here's a list of things needed to be done for making Syslog CEF Receiver production ready.

**Correctness/Reliability/Usability:**
- [x] Support parsing of CEF messages with the Syslog header #1073 (Fixed in #1264)
- [ ] Cap max number of connections
- [x] Avoid populating LogRecord body when parsing is successful. Here's a link to the [discussion](https://cloud-native.slack.com/archives/C08RRSJR7FD/p1757358254837529?thread_ts=1757351715.549509&cid=C08RRSJR7FD) (Fixed in #1940)
- [ ] TLS support for TCP? (Added in #1928) Waiting on #1742 to stabilize TLS functionality
- [x] Enforce message size limits to protect the receiver from malicious inputs. (For example, the TCP receiver trying to parse a very long message with no newline character) (Fixed in #2452)
- [x] Instrument the receiver to emit metrics about the count of messages received, processed, dropped etc. (Added in #1309)
- [ ] Improve parsing related error handling
- [x] Implement graceful shutdown (either wait for the already spawned tasks to finish or cancel them) (Fixed in #1962)
- [ ] ~~Add support for a parsing configuration that allows selective inclusion of attributes and optional renaming. Attributes not explicitly listed may either be excluded from the output or aggregated into a single string under a designated fallback attribute.~~ Out of scope (This is now being done using `Condense` processor #1695)

**Performance Exploration** (not needed for being production ready)
- [ ] Optimize representation for certain types #1104
- [ ] Use `memchr` for CEF parsing
- [ ] `simdutf8` to validate/convert UTF8
- [ ] `aho-corasick` crate to detect well know CEF keys
- [ ] `atoi` to parse integers directly from `&[u8]` without UTF-8 conversion or allocation
- [ ] For UDP traffic there might be a way to use a combination of `socket2` (already being used) and `nix` to reduce the number of syscalls from `recv_from` method

Dylan-M:
This conflicts with the proposed, more generic, solution in [contrib #37442](https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/37442). That solution has a generic decoupled syslog receiver, and offloads all parsing to the transform processor OTTL functions.

Why that is better?
- There is often a need to do parsing differently from a single receiver to multiple exporters (each exporter has a different set of parsing for the same data inbound, normally to send unmolested data to cold storage for compliance reasons)
[...]

PR #1962, opened 2026-02-05
[otap-df-otap] Implement graceful shutdown for Syslog CEF Receiver

utpilla:
# Change Summary

1. Proper Shutdown Deadline Handling: Both TCP and UDP now capture the deadline from `NodeControlMsg::Shutdown` and return `TerminalState::new(deadline, [snapshot])` instead of `TerminalState::default()`
2. UDP Graceful Flush: On shutdown, flushes any pending records in `arrow_records_builder` using `try_send_message_with_source_node()` before returning. Uses `try_send` (non-blocking) since we're shutting down and can't wait indefinitely
3. TCP Task Shutdown Signaling:
   - Added `Rc<Cell<bool>>` shutdown flag to signal spawned connection tasks to flush and exit
   - Tasks check `shutdown_flag.get()` at the top of each loop iteration (cheap bool read, no locks)
   - When flag is set, tasks flush pending records via `try_send` and exit cleanly
5. TCP Task Tracking & Graceful Drain:
   - Added `Rc<Cell<usize>>` to track active spawned tasks
   - Tasks increment counter when starting, decrement at all exit points (shutdown, EOF, read error, TLS handshake failure)
   - On shutdown, waits for tasks to finish with timeout:
     - Uses 90% of time until deadline, capped at 1 second (`MAX_TASK_DRAIN_WAIT`)
     - Busy-spins with `yield_now()` should be rare (acceptable during shutdown)
   - Takes final metrics snapshot only after drain wait completes
   
# Key Design Decisions
- Used `Rc<Cell<T>>` instead of `CancellationToken` - simpler, no external dependency, cheaper (just pointer deref + bool read)
- Used `try_send` during shutdown flush - non-blocking, won't hang if downstream is full
- Rare case: All the tasks handling the active connections during shutdown are awaitnng I/O, then we could have a busy-spin during drain wait which would keep checking if the active task count is zero. I think this is acceptable behavior during shutdown.

Related to #1149

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
+++ b/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
@@ -20,7 +20,7 @@ use otap_df_engine::{
 };
 use otap_df_telemetry::instrument::{Counter, UpDownCounter};
 use otap_df_telemetry::metrics::MetricSet;
-use otap_df_telemetry::otel_info;
+use otap_df_telemetry::{otel_info, otel_warn};
 use otap_df_telemetry_macros::metric_set;
 use serde::Deserialize;
 use serde_json::Value;
@@ -36,7 +36,7 @@ use crate::tls_utils::{accept_tls_connection, build_tls_acceptor};
 #[cfg(feature = "experimental-tls")]
 use otap_df_config::tls::TlsServerConfig;
 #[cfg(feature = "experimental-tls")]
-use otap_df_telemetry::{otel_debug, otel_warn};
+use otap_df_telemetry::otel_debug;
 
 /// Arrow records encoder for syslog messages
 pub mod arrow_records_encoder;
@@ -213,11 +213,19 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
                                     // capped at MAX_TASK_DRAIN_WAIT.
                                     let time_until_deadline = deadline.saturating_duration_since(std::time::Instant::now());
                                     let drain_wait = std::cmp::min(time_until_deadline * 9 / 10, MAX_TASK_DRAIN_WAIT);
-                                    let _ = tokio::time::timeout(drain_wait, async {
+                                    let drain_result = tokio::time::timeout(drain_wait, async {
                                         while active_task_count.get() > 0 {
                                             tokio::task::yield_now().await;
                                         }
                                     }).await;
+
+                                    if drain_result.is_err() {
+                                        otel_warn!(
+                                            "receiver.shutdown.drain_timeout",
+                                            active_tasks = active_task_count.get(),
+                                            message = "Shutdown drain timeout expired with tasks still active"
+                                        );
+                                    }
 
                                     let snapshot = self.metrics.borrow().snapshot();
                                     return Ok(TerminalState::new(deadline, [snapshot]));


rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -189,17 +192,35 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+                                    // Wait for active tasks to finish flushing.
+                                    // Use 90% of remaining time (keeping 10% buffer for cleanup),
+                                    // capped at MAX_TASK_DRAIN_WAIT.
+                                    let time_until_deadline = deadline.saturating_duration_since(std::time::Instant::now());
+                                    let drain_wait = std::cmp::min(time_until_deadline * 9 / 10, MAX_TASK_DRAIN_WAIT);
+                                    let _ = tokio::time::timeout(drain_wait, async {
+                                        while active_task_count.get() > 0 {
+                                            tokio::task::yield_now().await;
+                                        }
+                                    }).await;
```
lquerel: Let's try this approach and if we see an issue we can always move to a tokio cancellation token or a notify/watch.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -189,17 +192,35 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
-                                    break;
+                                    shutdown_flag.set(true); // Signal all connection tasks to flush and exit
+
+                                    // Wait for active tasks to finish flushing.
+                                    // Use 90% of remaining time (keeping 10% buffer for cleanup),
+                                    // capped at MAX_TASK_DRAIN_WAIT.
+                                    let time_until_deadline = deadline.saturating_duration_since(std::time::Instant::now());
+                                    let drain_wait = std::cmp::min(time_until_deadline * 9 / 10, MAX_TASK_DRAIN_WAIT);
+                                    let _ = tokio::time::timeout(drain_wait, async {
+                                        while active_task_count.get() > 0 {
```
ThomsonTan, line 217: Busy-spin with yield_now() during drain wait may consume CPU unnecessarily, consider using a small sleep interval?

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -231,9 +252,17 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+                                    // Clone shutdown flag for the spawned task
+                                    let task_shutdown_flag = shutdown_flag.clone();
+                                    // Clone active task counter for the spawned task
+                                    let task_active_count = active_task_count.clone();
+
                                     // Spawn a task to handle the connection.
                                     // ToDo should this be abstracted and exposed a method in the effect handler?
                                     _ = tokio::task::spawn_local(async move {
+                                        // Track this task for graceful shutdown
+                                        task_active_count.set(task_active_count.get() + 1);
```
lalitb, line 275: Seems there is a race condition between the main thread and the spawned task - 

 `active_task_count` is incremented inside the spawned task, so during shutdown the main loop can observe 0 before the task actually runs and increments it. A task spawned just before shutdown can then run outside the drain window, meaning shutdown may complete before that task flushes or decrements the counter.

Incrementing the counter before spawn_local (or using an RAII guard) would make the task visible to the drain loop immediately and avoid this window.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
lalitb, line 220: also consider logging here if the drain timeout expires with tasks still active - this will help distinguish graceful shutdown (all tasks flushed) from forced timeout during debugging/operations.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
utpilla: I've addressed this differently than incrementing before `spawn_local` https://github.com/open-telemetry/otel-arrow/pull/1962/commits/e409de52a32659f4ebd47eec03786716f293e9f0

My concern with incrementing before spawn is that we'd then wait in the drain loop for tasks that haven't even started processing yet (scheduler hasn't run them). Instead, the task now checks the shutdown flag first:

```rust
_ = tokio::task::spawn_local(async move {
    // If already shutting down, exit immediately (nothing to process yet)
    if task_shutdown_flag.get() {
        return;
    }
    
    // Now we're committed to handling this connection
    metrics.borrow_mut().tcp_connections_active.inc();
    task_active_count.set(task_active_count.get() + 1);
    // ...
});
```

This way tasks that are spawned just before shutdown would exit immediately without incrementing the count and they don't have any data to flush since they just started.

Also, moved `tcp_connections_active.inc()` inside the task so both counters stay in sync.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
utpilla: The busy spin should be rare in practice. The spawned tasks are blocked in `tokio::select!` waiting on IO or a 100ms timer. When we `yield_now()`, we go to the back of the pending queue, but the tasks need their timer to fire to wake them. Even if all tasks are blocked waiting on IO, the 100ms batch timer ensures they'll wake up shortly, check the flag, and exit.

That said, this only happens during shutdown (not on the hot path) and is bounded by the 1 second timeout cap. Seemed like an acceptable one-time shutdown cost.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
utpilla: Thanks! Fixed in https://github.com/open-telemetry/otel-arrow/pull/1962/commits/ac49181c406341a6f17c609e6f259da42dada711.

PR #1968, opened 2026-02-05
[ci] add triage:deciding label to new issues

ThomsonTan:
# Change Summary

Based on the discussion in today's SIG, add CI task to apply label triage:diciding to new issues for later triage.


.github/workflows/issue_triage.yml
```diff
@@ -0,0 +1,31 @@
...
+          egress-policy: audit
+
+      - uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
+        with:
+          script: |
+            github.rest.issues.addLabels({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              labels: ["triage:deciding"]
```
cijothomas, line 30: could we use issue templates and add deciding label automatically there instead of a workflow?

https://github.com/open-telemetry/opentelemetry-rust/blob/main/.github/ISSUE_TEMPLATE/FEATURE-REQUEST.yml?plain=1#L5

ThomsonTan: Yes, that is another option. Probably we could do that once we have them.

There could still be some issue like if the blank issue template is chosen, the label thing will be bypassed. But I think we want to make the `triage:deciding` label mandatory.

cijothomas: ya true. Anyway, not harm keeping this workflow. Pretty cheap to run.

Issue #1972, opened 2026-02-05
Feature: Durable logging identifiers for internal log statements


jmacd:
In Microsoft the term "Durable Identifier" is used to describe a permanent identifier assigned to each log statement.

For example,
https://learn.microsoft.com/en-us/dotnet/core/extensions/logging/source-generation

The OTAP dataflow engine uses a set of internally developed macros like `otel_info!`, `otel_error!` to report diagnostics at runtime. These are simple wrappers around the Tokio tracing APIs that place a constant string message first in the statement.

There are two ways to interpret the constant message argument:

1. As an event identifier
2. As a message format or simple message

I believe we would be served by separating these concepts, and I would like us to do both. The .NET runtime (see link above) has a compiler feature to support a stable hash function for exactly this purpose, but we can do this with human-provided identifiers.

Here are two examples of log statements in these styles:

```
    otel_info!("Help, I've fallen down", key = value);
    otel_error!("event.stability.fallen_down", key1, key2, key3 = a+b, key4 = xyz);
```

In the current implementation, the OTLP LogRecord `event_name` field is initialized with a concatenation of crate name, constant argument (i.e., identifier or message), plus file and line. (I've noticed myself not wanting to see the file and line on the console!)

Here are ways we could fix this (without compiler support and/or without a compile-time stable hash function).

```
    otel_info!(YELLING_OUT_LOUD, "Help, I've fallen down", key = value);
    otel_error!(STABILITY_EVENT_FALLEN, "Rotation: Fallen down", key1, key2, key3 = a+b, key4 = xyz);
```

The use of all-caps makes these identifiers look like constants; they are in fact compiled into the event_name. 

Our basic requirement is that these identifiers are durable. They should also be distinct, so not duplicated within a crate for example.

cijothomas:
`otel_info!("Help, I've fallen down", key = value);` is not following our guidelines for event names:
https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/telemetry/events-guide.md#event-naming

`otel_info!("climber.help_requested", reason = "fallen down");` would be apt.

cijothomas:
> In Microsoft the term "Durable Identifier" is used to describe a permanent identifier assigned to each log statement.

This is same as what OTel calls "Event Name" :)

lquerel:
@jmacd @cijothomas 

Shouldn't we reserve the `body` parameter to handle this type of case?

```rust
[...]

PR #1973, opened 2026-02-05
feat: Add message parameter to otel_debug macro

andborja:
# Change Summary

Adding "message" attribute to the otel_debug macro.

Part of https://github.com/open-telemetry/otel-arrow/issues/1972

* Closes NA

Building the package

No

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
@@ -101,9 +101,9 @@ impl RecordsetKqlProcessor {
             OtlpProtoBytes::ExportLogsRequest(bytes) => {
                 otap_df_telemetry::otel_debug!(
                     "processor.processing_logs",
-                    "Processing KQL query",
                     processor = "recordset_kql",
                     input_items = input_items,
+                    "Processing KQL query"
                 );
                 self.process_logs(bytes, signal)
             }


rust/otap-dataflow/crates/telemetry/src/internal_events.rs
```diff
@@ -69,12 +69,12 @@ macro_rules! otel_warn {
...
 /// ```ignore
 /// use otap_df_telemetry::otel_debug;
-/// otel_debug!("processing.batch", batch_size = 100);
+/// otel_debug!("processing.batch", "Processing batch of items", batch_size = 100);
 /// ```
 #[macro_export]
 macro_rules! otel_debug {
-    ($name:expr $(, $($fields:tt)*)?) => {
-        $crate::_private::debug!(name: $name, target: env!("CARGO_PKG_NAME"), { $($($fields)*)? }, "");
+    ($name:expr, $message:expr $(, $($fields:tt)*)?) => {
```
cijothomas: this is *forcing* one to pass message, which is not required. Let's keep it optional. (that's how it is in `tracing`, and even in OTel world, "body" is optional, and not recommended at all except in some scenarios)

jmacd:
Part of https://github.com/open-telemetry/otel-arrow/issues/1972

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
jmacd: @cijothomas I take you mean we will have two branches in the macro, one as @andborja has done here, and one like the original form.

cijothomas: yes

rust/otap-dataflow/crates/engine/src/lib.rs
```diff
@@ -998,6 +998,7 @@ impl<PData: 'static + Clone + Debug> PipelineFactory<PData> {
 
         otel_debug!(
             "receiver.create.start",
+            "Creating receiver node...",
```
cijothomas: These don't look good. The message is not really adding any useful *additional* information. It makes sense to use message, when the short event name is not sufficient to convey the full thing.

rust/otap-dataflow/crates/telemetry/src/tracing_init.rs
```diff
@@ -206,7 +206,7 @@ mod tests {
         for level in ALL_LEVELS {
             let setup = test_setup(ProviderSetup::Noop, level);
             setup.with_subscriber(|| {
-                otel_debug!("debug");
+                otel_debug!("debug", "debug message");
```
cijothomas, line 209: lets not just add messages for the sake of adding messages.

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
andborja: better

andborja:
> I am not opposed to the idea of using Human readable message in logs. Requesting changes to address the following.
> 
> 1. Make it optional. Only few scenarios benefit from human readable message, so no need of making it mandatory.
> 2. Make it for all severity, not just debug.
> 3. Don't add message for the sake of adding it. If the event.name says "component.started", what additional value does a message "Component is started" providing?
> 
> [open-telemetry/semantic-conventions#3343](https://github.com/open-telemetry/semantic-conventions/pull/3343) We are totally aligned with this proposal in OTel conventions as well with the above. (this PR in semantic convention is, in turn, aligned with OTel specification too)

1. Agree.
2. This PR is for debug only. I'll send others for the rest.
3. Agree.

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
```diff
@@ -69,10 +69,17 @@ macro_rules! otel_warn {
 /// # Example:
 /// ```ignore
 /// use otap_df_telemetry::otel_debug;
-/// otel_debug!("processing.batch", batch_size = 100);
+/// otel_debug!("processing.batch", "Processing batch of items", batch_size = 100);
```
cijothomas: `tracing::event!(Level::INFO, field = value, field2 = value2, "message here");`

The popular way in Rust (tracing, log crates that are used pretty much everywhere) is to have message as the very last argument, after all the k=v things..

jmacd: The `log` crate uses message formatting strings, where the message comes first, as in `log::info!("message {arg}")` which is still my preferred way to log.

I (personally) find this `tracing` keys-and-values-before-message style to be out of importance order, so I like what we have more than what `tracing` gives us.

I would still prefer an unquoted PROCESSING_BATCH in the code, instead of "processing.batch".

cijothomas: `log` crate also puts message as the last argument, when using Structured logging:
https://docs.rs/log/latest/log/#structured-logging
```rust
info!(target: "yak_events", yak:serde; "Commencing yak shaving - message is last");
```

cijothomas: > I (personally) find this tracing keys-and-values-before-message style to be out of importance order,

If we agree that message is the most important one, then it makes sense that message is first!. IMHO, in structured logging, the message is least important thing. Message is only needed in very small scenarios, and can often be provided by backend/UI if a *human* is reading and prefers reading detailed explanation.

OTel semantic convention also thinks so, and they recommend avoiding *body* and instead recommends Attributes. (This may not be the final recommendation, but the direction is to shift towards attributes (KV pairs), and body/message is reserved for very small use cases...)

andborja: I also prefer it as the second parameter. It is more readable when reading the code. It is anyway optional...

jmacd: See my remarks in #1954

cijothomas: > See my remarks in #1954

I am not sure how is that related to the discussion around whether message should be the 2nd or last parameter?

(I was so used to the .NET style where the message was first, followed by arguments, But Rust's popular crates `log` and `tracing` put them at the very last, so it makes sense to adhere to Rust style here.)

cijothomas:
https://github.com/open-telemetry/otel-arrow/pull/1973#discussion_r2771609152 is unresolved, and my suggestion is to keep message at the very last, matching existing pattern in `tracing` and `log`. Or special case "message" attribute name like done by OTel Rust (https://github.com/open-telemetry/opentelemetry-rust/pull/2689/changes) with blessing from `tracing` maintainers.

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
andborja: Updated to have message as the last parameter.

jmacd (approved):

jmacd:
I think we can try to improve the macros later, or change this stuff again -- what's important is we have both event name and optional message. (No more complaints from me!)

cijothomas:
> I think we can try to improve the macros later, or change this stuff again -- what's important is we have both event name and optional message. (No more complaints from me!)

As suggested here, we already had the ability to send optional message with special casing "message" attribute.
https://github.com/open-telemetry/otel-arrow/pull/1973#issuecomment-3862068816

PR #1976, opened 2026-02-05
fix: improve reliability of test_durable_buffer_recovery_after_outage

### LINKED ISSUE #1975: Flaky test: test_durable_buffer_recovery_after_outage
Some PR checks ([example](https://github.com/open-telemetry/otel-arrow/actions/runs/21727002917/job/62671602009?pr=1961#step:11:2595)) have been failing due to flaky behavior in `test_durable_buffer_recovery_after_outage`:

```
    thread 'test_durable_buffer_recovery_after_outage' (208) panicked at crates\otap\tests\durable_buffer_processor_tests.rs:570:5:
    Recovery should deliver at least 35 items (25+10), got 20
```

This test is subject to timing issues and should be more deterministic.

AaronRM:
# Change Summary

Improve the reliability of `test_durable_buffer_recovery_after_outage` so that it is not subject to minor timing differences across runs that may lead to test failure. Make test more precise by validating the exact number of signals persisted and received by the exporter.

* Closes #1975

* Code inspection, manually running the test to attempt failure repro.

No, this is change only affects test code.

### DIFF
--- a/rust/otap-dataflow/crates/otap/tests/durable_buffer_processor_tests.rs
+++ b/rust/otap-dataflow/crates/otap/tests/durable_buffer_processor_tests.rs
@@ -399,6 +399,9 @@ fn count_signals_in_segments(
                 .filter_map(|e| e.ok())
                 .filter(|e| e.path().extension().is_some_and(|ext| ext == "qseg"))
                 .map(|e| {
+                    // Errors are intentionally ignored here: this function polls for
+                    // segments that may be actively being written, so open failures
+                    // (incomplete header, locked file, etc.) are expected and transient.
                     SegmentReader::open(e.path())
                         .map(|reader| {
                             reader


rust/otap-dataflow/crates/otap/tests/durable_buffer_processor_tests.rs
```diff
@@ -379,6 +381,56 @@ where
...
+                    SegmentReader::open(e.path())
+                        .map(|reader| {
+                            reader
+                                .streams()
+                                .iter()
+                                .filter(|s| s.slot_id.raw() == slot_id_raw)
+                                .map(|s| s.row_count)
+                                .sum::<u64>()
+                        })
+                        .unwrap_or(0)
```
ThomsonTan, line 414: Log the error from `SegmentReader::open(...)` before `.unwrap_or(0)` instead of returning 0 silently?

AaronRM: In general I would agree, but in this case we are in test code which is expected to hit errors (since it is reading from segment files that are also being written to). I added a comment explaining why we are ignoring error results.

jmacd:
The test failure is a flake (port binding).

PR #1977, opened 2026-02-06
feat: Add message parameter to otel_warn macro

andborja:
# Change Summary

Add event name to missing otel_warn! calls.

Part of https://github.com/open-telemetry/otel-arrow/issues/1972

* Closes #NA

local builds and tests

No

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/tls_utils.rs
+++ b/rust/otap-dataflow/crates/otap/src/tls_utils.rs
@@ -71,7 +71,7 @@ fn convert_native_certs_to_pem(cert_res: &rustls_native_certs::CertificateResult
     let mut pem_data = Vec::new();
 
     for error in &cert_res.errors {
-        otel_warn!("tls.native_cert.load_error", error = ?error, message = "Error loading native cert");
+        otel_warn!("tls.native_cert.load_error", error = ?error);
     }
 
     for cert in &cert_res.certs {


rust/otap-dataflow/crates/telemetry/src/internal_events.rs
albertlockett, line 66: In #1978 we added a pattern to the otel_error macro to handle the message:
https://github.com/andborja/otel-arrow/blob/5851e7c9f04697d97d226e5f7c0583b72b9bc451/rust/otap-dataflow/crates/telemetry/src/internal_events.rs#L97-L100

Do we not have to do the same thing to the `otel_warn` macro?

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -667,7 +673,7 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
             // This is acceptable because the OS will still schedule the thread, but performance may be less predictable.
             otel_warn!(
                 "core_affinity.set_failed",
-                message = "Failed to set core affinity for pipeline thread. Performance may be less predictable."
```
albertlockett, line 975: Does providing the field name `message = ` still work  correctly after this PR will be merged?

If not, @cijothomas we might want to update the documentation that will be added in this PR https://github.com/open-telemetry/otel-arrow/pull/1987 as it provides explicitly the field name like this.

afborja: No. I'll update this PR with the latest format.

Thanks,

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
andborja: Updated to follow the latest definition of message as attribute.

rust/otap-dataflow/crates/controller/src/lib.rs
andborja: Updated to use the latest definition of message as an attribute.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -71,7 +71,7 @@ fn convert_native_certs_to_pem(cert_res: &rustls_native_certs::CertificateResult
     let mut pem_data = Vec::new();
 
     for error in &cert_res.errors {
-        otel_warn!("Error loading native cert", error = ?error);
+        otel_warn!("tls.native_cert.load_error", error = ?error, message = "Error loading native cert");
```
cijothomas: message not required here. It is not adding totally new information that the EventName is not covering.

andborja: That message was there already. I'm just relocating it.

cijothomas: Previously, that was the only thing there. This PR added the proper EventName which made the message low/no-value, hence the suggestion to remove it, so we are following the agreed guidance:
https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/telemetry/events-guide.md#the-message-attribute

andborja: Removed

PR #1979, opened 2026-02-06
Add AuthError source in azure_monitor_exporter

drewrelmas:
# Change Summary

Troubleshooting some transient Auth errors using `azure_monitor_exporter` component. We should expose the error coming from `azure_core` crate.

n/a

Unit tests

No

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/error.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/error.rs
@@ -22,7 +22,7 @@ pub enum Error {
 
     // ==================== Authentication Errors ====================
     /// Authentication/authorization error.
-    #[error("Auth error ({kind}){}", .source.as_ref().map(|e| format!(": {}", e)).unwrap_or_default())]
+    #[error("Auth error ({kind}){}", source.as_ref().map(|e| format!(": {}", e)).unwrap_or_default())]
     Auth {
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/error.rs
```diff
@@ -22,7 +22,7 @@ pub enum Error {
 
     // ==================== Authentication Errors ====================
     /// Authentication/authorization error.
-    #[error("Auth error ({kind})")]
+    #[error("Auth error ({kind}){}", .source.as_ref().map(|e| format!(": {}", e)).unwrap_or_default())]
```
cijothomas: some of the hardest thing to debug is when internal logs are not helpful!
See our similar attempt in OTLP world : https://github.com/open-telemetry/opentelemetry-rust/pull/3343

cijothomas: @drewrelmas I missed to share my main point here. When doing such logs, we need to super-careful, as it is entirely possible that logs contain token or other sensitive things. https://github.com/open-telemetry/opentelemetry-rust/issues/3021 is a recent example in Rust, but this was a constant pain.

Not a blocker for this PR, but we need to make sure we do structured logging properly, avoiding sensitive things from appearing in logs.

PR #1981, opened 2026-02-06
Improved logging for OTLP receiver

cijothomas:
# Change Summary

Minor but important improvement to logging to include the endpoint where OTLP receiver is listening.
Also downgraded a registry log to debug - its polluting startup with a ton of logs, which I don't find useful, so downgrading to DEBUG from INFO. Its too much for INFO level, IMHO.

### DIFF
--- a/rust/otap-dataflow/crates/telemetry/src/registry.rs
+++ b/rust/otap-dataflow/crates/telemetry/src/registry.rs
@@ -76,11 +76,7 @@ impl TelemetryRegistryHandle {
             // TODO(#1907): This could benefit from logging a human-readable form
             // of the entity that we refer to later in the logs, instead of logging
             // every key/value in every line of console_async output.
-            otel_debug!(
-                "registry.define_entity",
-                schema,
-                definition
-            );
+            otel_debug!("registry.define_entity", schema, definition);
         }
         outcome.key()
     }


rust/otap-dataflow/crates/telemetry/src/registry.rs
```diff
@@ -76,7 +76,11 @@ impl TelemetryRegistryHandle {
             // TODO(#1907): This could benefit from logging a human-readable form
             // of the entity that we refer to later in the logs, instead of logging
             // every key/value in every line of console_async output.
-            otel_info!("registry.define_entity", schema, definition);
```
cijothomas, line 79: This don't compile after https://github.com/open-telemetry/otel-arrow/pull/1973
@andborja is that intentional?

andborja: yes. We need to add them as name = value

jmacd: I sort of like the feature where a simple identifier translates into `simple = simple`. 
I would've kept that feature.

rust/otap-dataflow/crates/telemetry/src/registry.rs
```diff
@@ -76,7 +76,11 @@ impl TelemetryRegistryHandle {
...
+            otel_debug!(
+                "registry.define_entity",
+                schema = schema,
```
cijothomas:
```suggestion
                schema,
```

rust/otap-dataflow/crates/telemetry/src/registry.rs
```diff
@@ -76,7 +76,11 @@ impl TelemetryRegistryHandle {
...
+                definition = definition
```
cijothomas:
```suggestion
                definition
```

PR #1984, opened 2026-02-06
feat(quiver): skip expired WAL entries during replay

### LINKED ISSUE #1980: Quiver WAL replay should skip items older than max_age
This is a follow up concern from #1961. Shouldn't we skip and delete the WAL entries whose ingestion time exceeds the configured `max_age` limit? Otherwise, replayed entries get finalized into a new segment whose `finalized_at` timestamp reflects the restart time (not the original `ingestion_time`), effectively resetting their age clock.

That would mean data that's already older than `max_age` gets a fresh retention window contradicting the intent of `max_age`, which is that older data is less valuable.

_Originally posted by @utpilla in https://github.com/open-telemetry/otel-arrow/pull/1954#discussion_r2775449539_

AaronRM:
# Change Summary

During WAL replay, entries older than the configured `max_age` retention are now skipped rather than replayed into new segments. Without this filtering, replaying expired WAL entries would effectively reset their age to zero, causing data to be retained longer than intended by the configured policy. The cutoff is computed once before the replay loop and compared against each entry's ingestion_time (with no assumption about WAL ordering). Skipped entries advance the cursor so they won't be retried, and the expired_bundles counter is incremented so operators have visibility into filtered data. When *all* replayed entries are expired (nothing is replayed), the cursor is explicitly persisted to the sidecar to avoid redundant re-scanning on subsequent restarts.

* Closes #1980

Two new tests cover the mixed old/fresh filtering case and the all-expired edge case, the latter including a third engine reopen to verify cursor persistence.

No, this is an optimization to the WAL recovery behavior. No config or user-facing changes.

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/engine.rs
+++ b/rust/otap-dataflow/crates/quiver/src/engine.rs
@@ -697,6 +697,33 @@ impl QuiverEngine {
                 }
             };
 
+            // Skip WAL entries that are older than max_age — replaying them
+            // would effectively reset their age to zero and cause them to be
+            // retained longer than intended.
+            //
+            // Note: WAL entries are NOT assumed to be sorted by ingestion_time,
+            // so we check every entry individually rather than short-circuiting.
+            //
+            // We check the raw timestamp BEFORE decoding to a ReplayBundle to
+            // avoid spending cycles on IPC deserialization for expired entries.
+            if let Some(cutoff) = max_age_cutoff {
+                let entry_time = if entry.ingestion_ts_nanos >= 0 {
+                    UNIX_EPOCH + Duration::from_nanos(entry.ingestion_ts_nanos as u64)
+                } else {
+                    UNIX_EPOCH
+                };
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -705,6 +717,24 @@ impl QuiverEngine {
                 }
             };
 
+            // Skip WAL entries that are older than max_age — replaying them
+            // would effectively reset their age to zero and cause them to be
+            // retained longer than intended.
+            // Note: WAL entries are NOT assumed to be sorted by ingestion_time,
+            // so we check every entry individually rather than short-circuiting.
+            if let Some(cutoff) = max_age_cutoff {
```
ThomsonTan: Move this check to above the WAL decoding? Then we could avoid decoding the expired WAL entries.

jmacd (approved):

utpilla (approved):

Issue #1613, opened 2025-12-12
Use weaver live-check in CI to ensure internal instrumentation follows OTel conventions


cijothomas:
https://github.com/open-telemetry/otel-arrow/pull/1605#pullrequestreview-3572974106

Let's follow what we advise everyone to follow!
For internal Events, there aren't any conventions defined yet. But for internal metrics, there are some defined for OTel SDKs, which could be followed? (https://github.com/open-telemetry/semantic-conventions/blob/main/docs/otel/sdk-metrics.md)

cijothomas:
https://github.com/open-telemetry/semantic-conventions/issues/3718 Opened an issue in semantic conventions to test the water. Most likely, we'll list our conventions in this repo itself, and then use a weaver CI guard to confirm we are producing what we claim to produce. 

https://github.com/open-telemetry/otel-arrow/pull/2986 was done to test viability of Weaver live-check, and that looks very promising already!

cijothomas:
From today's community call:

1. We should start with defining out own Semantic Convention registry for all the Events we produce. (and use weaver-live-check to enforce comformance)
2. And this gives us a realistic semantic convention pool of Events to be leveraged in our TrafficGenerator to give realistic loads.

So two problems are being solved by this!

PR #1987, opened 2026-02-06
Update events guide

cijothomas:
Guidance updated on proper usage of Events (AKA Structured Logs with Durable Name)
Enforce at CI any direct usage of tokio/tracing or log macros.

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/engine.rs
+++ b/rust/otap-dataflow/crates/quiver/src/engine.rs
@@ -22,7 +22,7 @@ use std::fs;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
-use std::time::Duration;
+use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
 use parking_lot::Mutex;
 use tokio::sync::Mutex as TokioMutex;
@@ -654,7 +654,19 @@ impl QuiverEngine {
         };
 
         let mut replayed_count = 0;
+        let mut skipped_expired = 0u64;
         let mut stopped_at_corruption = false;
+
+        // Compute max_age cutoff once before the loop so expired WAL entries
+        // are not replayed into new segments, which would reset their age and
+        // cause them to be retained longer than intended.
+        // If max_age is so large that it underflows past the epoch, clamp to
+        // UNIX_EPOCH (effectively disabling filtering, which is correct).
+        let max_age_cutoff = self
+            .config
+            .retention
+            .max_age
+            .map(|max_age| SystemTime::now().checked_sub(max_age).unwrap_or(UNIX_EPOCH));
 
         for entry_result in iter {
             let entry = match entry_result {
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/docs/telemetry/events-guide.md
```diff
@@ -36,13 +36,89 @@ Exception rule (traces):
...
+- `?value` -- Debug formatting (`fmt::Debug`)
+- `value` -- passed directly (integers, booleans, etc.)
+
+```rust
+otel_info!("node.connect",
+    endpoint = %addr,
+    config   = ?node_config,
+    count    = 42,
+);
+```
```
jmacd, line 121: Thanks!

rust/otap-dataflow/scripts/check-direct-telemetry-macros.sh
```diff
@@ -0,0 +1,111 @@
...
+    echo "  • tracing::warn!(...)  → otel_warn!(\"event.name\", ...)"
+    echo "  • tracing::error!(...) → otel_error!(\"event.name\", ...)"
+    echo "  • tracing::debug!(...) → otel_debug!(\"event.name\", ...)"
+    echo ""
+    echo "  Import from otap_df_telemetry:"
+    echo "    use otap_df_telemetry::otel_info;"
+    echo ""
+    echo "  See docs/telemetry/events-guide.md for full details."
+    exit 1
+fi
```
jmacd, line 111: Should we add a test that there are no space characters in the event name? (Is that a weaver thing?)
(Can it be checked in the macro?)

cijothomas: Yes we need to enforce as much as possible in compile time, and then validate against published schema in [CI time (via weaver).](https://github.com/open-telemetry/otel-arrow/issues/1613)

This PR is one baby step towards that Goal.

rust/otap-dataflow/scripts/check-direct-telemetry-macros.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
```
albertlockett, line 1: Will this script catch the case where someone uses the macro in a way where, for example, the case `tracing::warn!` doesn't actually appear in the file? e.g.:
```rs
use tracing::warn;
// ...

warn!("...");
```

Would it be easier or more effective to just use the clippy rule for banned macros? We already do this for lazy_static:

https://github.com/open-telemetry/otel-arrow/blob/466fa0268c689767413b158469cb93852522eeed/rust/otap-dataflow/.clippy.toml#L17-L19

jmacd:
I suggest we merge and improve clippy settings in a future PR. Thank you @cijothomas!

rust/otap-dataflow/scripts/check-direct-telemetry-macros.sh
cijothomas: I started with clippy for this, but clippy wasn't able to distinguish "the developer wrote tracing::info!" from "the developer wrote otel_info! which expanded to tracing::info!". The script here was my cheap way to do some enforcement, while I can find better ways!

PR #1988, opened 2026-02-06
feat: durable event names to quiver logging

AaronRM:
# Change Summary

All log/trace calls in the quiver crate now use crate-private `otel_info!`, `otel_warn!`, `otel_error!`, and `otel_debug!` macros that enforce a required event name as the first argument. This ensures every log event has a stable, machine-readable OpenTelemetry Event name following the `quiver.<component>.<action>` convention.

n/a

Minimal unit test for the macros, ensured existing tests pass.

No

### DIFF
--- a/rust/otap-dataflow/scripts/check-direct-telemetry-macros.sh
+++ b/rust/otap-dataflow/scripts/check-direct-telemetry-macros.sh
@@ -20,12 +20,14 @@ check_banned_pattern() {
     #   - .git/           git internals
     #   - crates/telemetry/  implements the otel_* wrappers (legitimate tracing usage)
     #   - crates/quiver/src/logging.rs     implements the otel_* wrappers internal to quiver (legitimate tracing usage)
+    #   - crates/quiver-e2e/  end-to-end tests may legitimately use tracing directly
     #   - benchmarks/        benchmarks may legitimately use tracing directly
     files=$(find . -name "*.rs" -type f \
         | grep -v target/ \
         | grep -v .git/ \
         | grep -v crates/telemetry/ \
         | grep -v crates/quiver/src/logging.rs \
+        | grep -v crates/quiver-e2e/ \
         | grep -v benchmarks/ \
         || true)
 


rust/otap-dataflow/crates/quiver/src/subscriber/progress.rs
```diff
@@ -654,10 +655,12 @@ pub fn scan_progress_files(dir: &Path) -> Result<Vec<SubscriberId>> {
             match SubscriberId::new(id_str) {
                 Ok(id) => subscriber_ids.push(id),
                 Err(e) => {
-                    tracing::warn!(
+                    otel_warn!(
```
cijothomas, line 658: `quiver.subscriber.progress_scan.file_ignored` would be more apt name.

rust/otap-dataflow/crates/quiver/src/wal/reader.rs
```diff
@@ -929,21 +935,27 @@ impl Iterator for MultiFileWalIter {
                         // Seek to appropriate position within this file
                         let seek_pos = self.start_position.max(file_info.wal_position_start);
                         if let Err(e) = reader.seek_to_position(seek_pos) {
-                            tracing::warn!(
+                            otel_error!(
+                                "quiver.wal.file.init",
```
cijothomas, line 939: using Error severity for an "init" operation sounds too aggressive? Or this was meant to mean something else?

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
```diff
@@ -447,10 +456,28 @@ impl WalWriter {
             file.flush().await?;
             (0, header.encoded_len()) // New file starts at WAL position 0
         } else if metadata.len() < WAL_HEADER_MIN_LEN as u64 {
+            otel_error!(
+                "quiver.wal.init",
```
cijothomas, line 460: we can use a more specific event name to indicate this is not just an "init" event, its more precisely "init.failed" - the attributes explain who and why failed

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
```diff
@@ -1251,6 +1303,13 @@ impl WalCoordinator {
 
     async fn rotate_active_file(&mut self, active_file: &mut ActiveWalFile) -> WalResult<()> {
         if self.rotated_files.len() >= self.options.max_rotated_files {
+            otel_warn!(
+                "quiver.wal.backpressure",
+                reason = "rotated_files_cap",
```
ThomsonTan, line 1309: Is reusing the same reason and messag with the above `preflight_append` (L1067) intentional?

AaronRM: Yes. Added a `phase` attribute to clarify.

rust/otap-dataflow/crates/quiver/src/wal/writer.rs
AaronRM: Awaiting feedback on https://github.com/open-telemetry/otel-arrow/pull/2010 to determine direction.

AaronRM: [This paragraph](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/telemetry/events-guide.md?plain=1#L175) suggests that failure events be differentiated only by level, but use the same event name:

> Regarding severity, some events may be logged at different levels depending on
> their severity or impact. For example, a `node.shutdown` event may be logged at
> INFO level during a graceful shutdown, but at ERROR level if the shutdown is due
> to a critical failure. When exporting events as logs, choose the log level that
> best reflects the significance of the event.

However, elsewhere the guide does suggest using 'fail' as a verb in the event name. We need to clear up the ambiguity relating to naming for error/failure events.

lquerel: Please see my comment on PR #2010 
I think it would be best to wait for feedback from Liudmila and/or Josh before finalizing our decision.

cijothomas (approved):
Agree on the direction to use `tracing` via wrapper to enforce EventName.
The actual event names - lets come back once we have clearer guidance.

Issue #1989, opened 2026-02-06
Error: cannot allocate memory in static TLS block


cbandy:
I'm trying some OTAP crates in a shared library. After a `cargo update`, I now encounter this error when dynamically loading:
```
cannot allocate memory in static TLS block
```

As I understand it, TLS is [thread local storage](https://doc.rust-lang.org/stable/unstable-book/compiler-flags/tls-model.html), and this error can be caused by custom allocators, e.g.

- https://www.github.com/jemalloc/jemalloc/issues/937
- https://www.github.com/purpleprotocol/mimalloc_rust/issues/138

This error went away when I moved jemalloc behind a disabled feature flag in otap-df-engine.

drewrelmas:
Appears closed based on linked PR description.

PR #1990, opened 2026-02-06
Make jemalloc optional in otap-df-engine on Linux

cbandy:
# Change Summary

This tucks the jemalloc pipeline metrics behind a `jemalloc` feature in the engine crate.

#1989

`cargo tree` with the feature on and off.

No, the default feature still includes jemalloc.

### DIFF
--- a/rust/otap-dataflow/crates/engine/Cargo.toml
+++ b/rust/otap-dataflow/crates/engine/Cargo.toml
@@ -14,7 +14,8 @@ workspace = true
 
 [features]
 testing = []
-jemalloc-testing = []
+jemalloc = ["dep:tikv-jemalloc-ctl"]
+jemalloc-testing = ["jemalloc"]
 test-utils = []
 
 [dependencies]
@@ -42,13 +43,12 @@ uuid = { workspace = true }
 once_cell = { workspace = true }
 data-encoding = { workspace = true }
 prost = { workspace = true }
-libmimalloc-sys = { workspace = true }
 byte-unit = { workspace = true }
 cpu-time = { workspace = true }
 nix = { workspace = true, features = ["resource"] }
 
 [target.'cfg(not(windows))'.dependencies]
-tikv-jemalloc-ctl = { workspace = true }
+tikv-jemalloc-ctl = { workspace = true, optional = true }
 
 [dev-dependencies]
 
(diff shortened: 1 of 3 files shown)


rust/otap-dataflow/crates/engine/Cargo.toml
```diff
@@ -14,6 +14,7 @@ workspace = true
 
 [features]
 testing = []
+jemalloc = ["dep:tikv-jemalloc-ctl"]
```
cbandy, line 17: 🤔 Hmm. How does Cargo feel about this on Windows?

cbandy: `cargo check` is happy when I flip all the `not(windows)` to `windows` on my Linux machine.

rust/otap-dataflow/crates/engine/Cargo.toml
```diff
@@ -14,6 +14,7 @@ workspace = true
...
 jemalloc-testing = []
```
ThomsonTan: The feature `jemalloc` should be added to the list of `jemalloc-testing`?

lquerel:
The description attached to this PR is a bit too brief and not well aligned with the title.

We already have the option to choose between mimalloc and jemalloc. What does this PR enable that we could not do before, and what is the underlying motivation?

Also, jemalloc is currently the default allocator not because it is the fastest (mimalloc is slightly faster), but because it exposes per-thread memory usage metrics, and therefore per pipeline instance, which are important for this project.

I would like to better understand what we are trying to achieve here before merging this PR.

@jmacd @albertlockett @drewrelmas for visibility.

cbandy:
> We already have the option to choose between mimalloc and jemalloc. What does this PR enable that we could not do before, and what is the underlying motivation?

`tikv-jemallocator` is optional in the workspace, and this makes `tikv-jemalloc-ctl` optional in otap-df-engine.

As I described briefly in #1989, I'm trying to use OTAP crates in a dynamically loaded library. Including any of the `jemalloc` crates then causes a Thread Local Storage error. I'd like to disable `jemalloc` for now.

lquerel:
@cbandy Ok, understood. As long as jemalloc remains enabled by default for the main process, that works for me.

rust/otap-dataflow/crates/engine/Cargo.toml
albertlockett: Windows checks are passing in CI :+1:

Issue #1611, opened 2025-12-12
OTAP Exporter remove sleeps from unit tests


albertlockett:
The OTAP exporter uses some random sleeps while the server is starting/stopping to test the behaviour of sending gRPC requests while the server is not ready tor receive connections.
https://github.com/open-telemetry/otel-arrow/blob/785f92edfcbf9f2f9797933c9236a8a35a7d5369/rust/otap-dataflow/crates/otap/src/otap_exporter.rs#L846-L854

However, this is fragile and actually seems to not work in windows:
https://github.com/open-telemetry/otel-arrow/actions/runs/20154182152/job/57898503464?pr=1594

We could update the test to wait to receive ACK or NACK before proceeding instead of relying on random sleeps. This would probably make the test more reliable.

pritishnahar95:
Since https://github.com/open-telemetry/otel-arrow/pull/1994 has been merged, this issue should be closed.

Issue #1325, opened 2025-10-21
OTAP exporter Ack/Nack handling and concurrency


jmacd:
Similar to #1323 and #1324:

The OTAP exporter will require an upgrade to support Ack/Nack. We will need to decide how to limit concurrency at the same time.

PR #1994, opened 2026-02-07
feat: handle Ack and Nack messages in OTAP Exporter.

alochaus:
# Change Summary

Part of #1325 (see also #1324).

## Problem: flaky test due to sleep-based synchronization.

The OTAP Exporter had two related issues:

### 1. Dropped Context

When the exporter received pipeline data (OtapPdata), it split the message into context and payload, then threw away the context:

https://github.com/open-telemetry/otel-arrow/blob/82f71508f8e598e78853335cce82195e894894cd/rust/otap-dataflow/crates/otap/src/otap_exporter.rs#L233-L235

The context carries routing information that tells the pipeline who to notify when a message succeeds or fails. Without it, the exporter was a black hole — data went in, but no confirmation ever came back.

### 2. Flaky Test

The test `test_receiver_not_ready_on_start` had no way to know when the exporter finished processing, so it used arbitrary sleeps:

https://github.com/open-telemetry/otel-arrow/blob/82f71508f8e598e78853335cce82195e894894cd/rust/otap-dataflow/crates/otap/src/otap_exporter.rs#L866-L868

https://github.com/open-telemetry/otel-arrow/blob/82f71508f8e598e78853335cce82195e894894cd/rust/otap-dataflow/crates/otap/src/otap_exporter.rs#L870-L872

https://github.com/open-telemetry/otel-arrow/blob/82f71508f8e598e78853335cce82195e894894cd/rust/otap-dataflow/crates/otap/src/otap_exporter.rs#L883-L885

On a slow machine or under load, these might not be long enough. On a fast machine, they waste time.

## Solution: thread context through for ACK/NACK.

The core idea is to preserve the `OtapPdata` context by using `take_payload()` instead of `into_parts()`.

This extracts the payload for gRPC transmission while keeping the context alive in the original OtapPdata. Then pass that context through the entire streaming pipeline so it can be returned with ACK (success) or NACK (failure) notifications.

## Key changes

1. Preserve context at the entry point

```rust
  // Before: context discarded
  let (_context, payload) = pdata.into_parts();

  // After: payload extracted, context preserved
  let payload = pdata.take_payload();
```

2. Pair each batch with its context through the pipeline

The internal channels changed from carrying just data to carrying (context, data) tuples:

```rust
  // Before
  channel::<OtapArrowRecords>(64)

  // After
  channel::<(OtapPdata, OtapArrowRecords)>(64)
```

3. Correlate gRPC responses with their original requests

The exporter uses bidirectional gRPC streaming — requests go out on one stream, responses come back on another. A FIFO correlation channel pairs them:

```
  create_req_stream ──sends pdata──→ [correlation channel] ──recv pdata──→ handle_res_stream
    (yielded batch)                                                         (got response)
```

Since both streams are ordered, the first response always corresponds to the first request.

4. Send ACK/NACK in the main loop

```rust
  Some(PDataMetricsUpdate::Exported(signal_type, pdata)) => {
      self.pdata_metrics.inc_exported(signal_type);
      effect_handler.notify_ack(AckMsg::new(pdata)).await?;
  },
  Some(PDataMetricsUpdate::Failed(signal_type, pdata)) => {
      self.pdata_metrics.inc_failed(signal_type);
      effect_handler.notify_nack(NackMsg::new("export failed", pdata)).await?;
  },
```

The effect_handler uses the context inside pdata to route ACK/NACK back through the pipeline.

5. Replace sleeps with deterministic waits in the test

```Rust
  // Before: sleep and hope
  tokio::time::sleep(Duration::from_millis(5)).await;

  // After: wait for the actual event
  timeout(Duration::from_secs(5), async {
      loop {
          match pipeline_ctrl_msg_rx.recv().await {
              Ok(PipelineControlMsg::DeliverNack { .. }) => break,
              Ok(_) => continue,
              Err(_) => panic!("pipeline ctrl channel closed"),
          }
      }
  }).await.expect("Timed out waiting for NACK");
```

The test is now event-driven: it proceeds as soon as the NACK/ACK arrives, with a 5-second timeout as a safety net.

* https://github.com/open-telemetry/otel-arrow/issues/1611


lalitb:
Just to confirm, I believe this is not an issue with OTLP exporter -  the exporter captures the context from `into_parts()`, threads it through the export pipeline, and sends proper ACK/NACK messages via `route_export_result()`.

alochaus:
You're right. This problem doesn't happen in OTLP Exporter, only in OTAP exporter.

jmacd:
@alochaus FYI see #1324 and #1325

jmacd:
Thank you @alochaus!

FYI

```
  // After: payload extracted, context preserved
  let payload = pdata.take_payload();
```

In the future we may want to extend this with an option to save the payload in case the retry_processor is subscribed with a RETURN_DATA interest. I can see how we might like a helper on OtapPdata and Context, a sort of clone_or_take() operation that clones otherwise takes that handles the RETURN_DATA case.

PR #2003, opened 2026-02-09
chore: Replace duplicated OpenSSL/rcgen test cert helpers with shared dev crate

lalitb:
### Summary

  Pulls the TLS/mTLS test certificate generation code into a small, unpublished workspace crate (`otap-test-tls-certs`) so all the exporter, receiver, and tls_utils unit tests share one `rcgen`-based implementation. Removes all `OpenSSL` CLI calls from test paths.

### Motivation

  We had the same cert-generation logic duplicated across multiple test files, some using `OpenSSL` CLI calls and some using inline `rcgen`. This made tests flaky on systems without `OpenSSL` installed and meant bug fixes had to be applied in several places. This consolidates everything into one shared rcgen-only helper.

### What changed

   - New internal crate `crates/otap-test-tls-certs` (publish = false) with CA, leaf, and self-signed cert helpers using rcgen.
   - All TLS integration tests and `tls_utils` unit tests now import from `otap_test_tls_certs` instead of per-file OpenSSL/rcgen copies.
   - Removed `skip_if_no_openssl()` guards - tests no longer depend on the `OpenSSL` CLI being installed.
   - Added `#[must_use]` where clippy asked for it.

### Alternatives considered

   1. **_Shared module in tests/common/_** - works for integration tests, but unit tests in src/ can't import from tests/common/ without include!(), which is fragile and poorly supported by tooling.
   2. **_Feature-gated module in src/_** - avoids a new crate, but #[cfg(test)] doesn't apply when the crate is built as a dependency for integration tests, so you end up 
  needing an extra Cargo feature and awkward wiring. Mixing test-only code into the main crate felt wrong.
   3. **_Dedicated dev helper crate (this PR)_** - standard pattern in Rust workspaces (publish = false, listed in [dev-dependencies]). Clean imports everywhere, no special tricks, no impact on production builds.

  **_Went with option 3 because it's the most straightforward to maintain and extend._**

### DIFF
--- a/rust/otap-dataflow/Cargo.toml
+++ b/rust/otap-dataflow/Cargo.toml
@@ -49,7 +49,7 @@ otap-df-pdata-otlp-model = { path = "./crates/pdata/src/otlp/model"}
 otap-df-config = { path = "crates/config" }
 otap-df-controller = { path = "crates/controller" }
 otap-df-otap = { path = "crates/otap" }
-otap-test-tls-certs = { path = "crates/otap-test-tls-certs" }
+otap-test-tls-certs = { package = "otap-df-test-tls-certs", path = "crates/otap-test-tls-certs" }
 otap-df-pdata = { path = "crates/pdata" }
 otap-df-telemetry = { path = "crates/telemetry" }
 quiver = { package = "otap-df-quiver", path = "crates/quiver" }
--- a/rust/otap-dataflow/crates/otap-test-tls-certs/Cargo.toml
+++ b/rust/otap-dataflow/crates/otap-test-tls-certs/Cargo.toml
@@ -1,10 +1,10 @@
 [package]
-name = "otap-test-tls-certs"
+name = "otap-df-test-tls-certs"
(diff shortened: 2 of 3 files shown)


albertlockett (approved):
This is great, thanks @lalitb

rust/otap-dataflow/crates/otap-test-tls-certs/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+    let cert = params.self_signed(&key_pair).expect("self-signed cert");
+
+    GeneratedCert {
+        cert_pem: cert.pem(),
+        key_pem: key_pair.serialize_pem(),
+    }
+}
+
+/// Generate CA + signed leaf cert and write them to disk.
+#[must_use]
```
ThomsonTan, line 122: Seems the return value is ignored in all callsites (`let _ =`), probably just remove `#[must_use]` because the the function is mostly for its side-effect.

PR #2004, opened 2026-02-09
Mark methods const

utpilla:
# Change Summary
- Mark methods `const` when applicable

No

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/segment_store.rs
+++ b/rust/otap-dataflow/crates/quiver/src/segment_store.rs
@@ -365,7 +365,7 @@ impl SegmentStore {
     /// On Windows, this occurs when trying to delete a file that's still open
     /// (e.g., memory-mapped by outstanding `BundleHandle`s).
     #[cfg(windows)]
-    const fn is_sharing_violation(error: &std::io::Error) -> bool {
+    fn is_sharing_violation(error: &std::io::Error) -> bool {
         // Windows error code 32: ERROR_SHARING_VIOLATION
         // "The process cannot access the file because it is being used by another process."
         error.raw_os_error() == Some(32)


cijothomas:
My browser is struggling to render all the files! Could this be split to few files at a time? 
(Or we can give a blind approval trusting the author)

AaronRM:
@utpilla Would it make sense to [enable the relevant clippy rule](https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn)?

utpilla:
> @utpilla Would it make sense to [enable the relevant clippy rule](https://rust-lang.github.io/rust-clippy/master/index.html#missing_const_for_fn)?

@AaronRM We could do that. However, this clippy rule is in the ["nursery"](https://github.com/rust-lang/rust-clippy?tab=readme-ov-file#clippy) category which means it's still under development.

I think we could enable this clippy rule. The only concern would be if later they make changes so that methods can't be marked `const`, we would have to update them. Also, something to note: if we mark a public function with const, removing it again will break API compatibility.

AaronRM:
> if we mark a public function with const, removing it again will break API compatibility.

@utpilla Another consideration...the `otel_*!` macros can't be called from within a `const fn`. If we mark a method without telemetry `const fn`, it may require *a breaking change* (switching to non-const) to add telemetry statements to it later.

utpilla:
> > if we mark a public function with const, removing it again will break API compatibility.
> 
> @utpilla Another consideration...the `otel_*!` macros can't be called from within a `const fn`. If we mark a method without telemetry `const fn`, it may require _a breaking change_ (switching to non-const) to add telemetry statements to it later.

@AaronRM Yes, that's true. However, I believe this is unlikely to be an issue in practice. The methods we're marking `const` are pure operations (constructors, property accessors, simple transformations) that by nature don't have side effects worth observing. Any logging we'd care about would generally be at the call boundary for these `const` methods.

PR #2010, opened 2026-02-10
Clarify event naming for different outcomes

### LINKED ISSUE #1972: Feature: Durable logging identifiers for internal log statements

cijothomas:
The "Severity and placement" section previously suggested using the same event name (node.shutdown) at different severity levels to distinguish a graceful shutdown (INFO) from a critical failure (ERROR).

This conflicts with the guidance in Event Naming and Verbs, which recommends distinct event names for different outcomes. Updated the example to use node.shutdown.complete (INFO) and node.shutdown.fail (ERROR) so the guide is internally consistent.

This is an intentional semantic change, not just a wording tweak. The guide now consistently says: different outcomes → different event names. Severity reflects significance, but should not be the sole way to distinguish success from failure.

Fixes #1972

### DIFF
--- a/rust/otap-dataflow/crates/config/src/pipeline.rs
+++ b/rust/otap-dataflow/crates/config/src/pipeline.rs
@@ -336,6 +336,98 @@ impl PipelineNodes {
 
         cycles
     }
+
+    /// Remove nodes that are not properly connected in the pipeline graph.
+    ///
+    /// This iteratively removes:
+    /// - Receivers with no outgoing connections (empty `out_ports` or all destinations empty)
+    /// - Processors and exporters that are not referenced as a destination by any other node
+    /// - Processors whose only outgoing destinations have all been removed
+    ///
+    /// Removal is iterative because removing one node may cause downstream or upstream
+    /// nodes to become disconnected.
+    ///
+    /// Returns `(NodeId, NodeKind)` for every removed node (for warning purposes).
+    pub fn remove_unconnected_nodes(&mut self) -> Vec<(NodeId, NodeKind)> {
+        let mut removed = Vec::new();
+
+        loop {
+            let mut to_remove = Vec::new();
+
+            // Build the set of all node IDs referenced as a destination by any remaining node.
+            let mut referenced_as_destination: HashSet<NodeId> = HashSet::new();
+            for node_config in self.0.values() {
+                for edge in node_config.out_ports.values() {
+                    for dest in &edge.destinations {
+                        _ = referenced_as_destination.insert(dest.clone());
+                    }
+                }
+            }
+
+            for (node_id, node_config) in &self.0 {
+                match node_config.kind {
+                    NodeKind::Receiver => {
+                        // A receiver is unconnected if it has no out_ports or all its
+                        // out_ports have empty destinations.
+                        let has_destinations = node_config
+                            .out_ports
+                            .values()
+                            .any(|edge| !edge.destinations.is_empty());
+                        if !has_destinations {
+                            to_remove.push((node_id.clone(), node_config.kind));
+                        }
+                    }
+                    NodeKind::Processor | NodeKind::ProcessorChain => {
+                        if !referenced_as_destination.contains(node_id) {
+                            to_remove.push((node_id.clone(), node_config.kind));
+                        } else {
+                            // Also check if processor has no outgoing connections
+                            let has_destinations = node_config
+                                .out_ports
+                                .values()
+                                .any(|edge| !edge.destinations.is_empty());
+                            if !has_destinations {
+                                to_remove.push((node_id.clone(), node_config.kind));
+                            }
+                        }
+                    }
+                    NodeKind::Exporter => {
+                        if !referenced_as_destination.contains(node_id) {
+                            to_remove.push((node_id.clone(), node_config.kind));
+                        }
+                    }
+                }
+            }
+
+            if to_remove.is_empty() {
+                break;
+            }
+
+            // Remove the unconnected nodes
+            for (node_id, kind) in &to_remove {
+                _ = self.0.remove(node_id);
+                removed.push((node_id.clone(), *kind));
+            }
+
+            // Clean up references to removed nodes from remaining nodes' destinations.
+            let removed_ids: HashSet<&NodeId> = to_remove.iter().map(|(id, _)| id).collect();
+            for node_config in self.0.values_mut() {
+                let references_removed = node_config
+                    .out_ports
+                    .values()
+                    .any(|edge| edge.destinations.iter().any(|d| removed_ids.contains(d)));
+                if references_removed {
+                    let mut updated = (**node_config).clone();
+                    for edge in updated.out_ports.values_mut() {
+                        edge.destinations.retain(|dest| !removed_ids.contains(dest));
+                    }
+                    *node_config = Arc::new(updated);
+                }
+            }
+        }
+
+        removed
+    }
 }
 
 impl std::ops::Index<&str> for PipelineNodes {
@@ -541,6 +633,13 @@ impl PipelineConfig {
     /// Creates a consuming iterator over the nodes in the pipeline.
     pub fn node_into_iter(self) -> impl Iterator<Item = (NodeId, Arc<NodeUserConfig>)> {
         self.nodes.into_iter()
+    }
+
+    /// Remove unconnected nodes from the pipeline and return descriptions of what was removed.
+    ///
+    /// This delegates to [`PipelineNodes::remove_unconnected_nodes`] on the main pipeline nodes.
+    pub fn remove_unconnected_nodes(&mut self) -> Vec<(NodeId, NodeKind)> {
+        self.nodes.remove_unconnected_nodes()
     }
 
     /// Returns the service-level telemetry configuration.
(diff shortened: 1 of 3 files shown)


cijothomas:
@lquerel @albertlockett Requesting more eyes for this.

rust/otap-dataflow/docs/telemetry/events-guide.md
```diff
@@ -172,11 +175,10 @@ Optionally, add occurrence-specific attributes (dynamic context):
...
 
-Regarding severity, some events may be logged at different levels depending on
-their severity or impact. For example, a `node.shutdown` event may be logged at
-INFO level during a graceful shutdown, but at ERROR level if the shutdown is due
-to a critical failure. When exporting events as logs, choose the log level that
-best reflects the significance of the event.
+Regarding severity, choose the log level that best reflects the significance of
+the event. For example, `node.shutdown.complete` at INFO for a graceful
+shutdown and `node.shutdown.fail` at ERROR for a critical failure -- these are
+distinct events, not the same event at different severity levels.
```
AaronRM, line 181: FWIW, I have a slight preference for the previous guidance (i.e. same event name, different level based on outcome) only because it can be easier to review in a downstream dashboard, etc...

When filtering by event name (e.g. `node.shutdown`) the user sees both the info and warning/error messages interleaved without needing to get into the internals of the event names.

Of course, the counter argument is that a fuzzy search on `node.shutdown` would yield the same results either way.

cijothomas: https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-eventname 
I will ask if OTel semantic conventions has thoughts on this as well. From what I can remember from old discussions, each occurrence of an Event should have same structure - but it does not explicitly say if Severity is part of the structure! (it does say attributes and body are)

jmacd:
I'm so, so very tempted to share my opinion, but then I realized it's wasteful of all of our time. I agree with Cijo we should follow OTel and there is a nuanced question about whether Level is considered independent of event identity to be addressed. I approve if @AaronRM and @cijothomas and others agree.

AaronRM:
> I'm so, so very tempted to share my opinion, but then I realized it's wasteful of all of our time. I agree with Cijo we should follow OTel and there is a nuanced question about whether Level is considered independent of event identity to be addressed. I approve if @AaronRM and @cijothomas and others agree.

I agree. 👍 My main concern is that the guidelines are consistent to know how to address comments properly in #1988.

cijothomas:
> node.shutdown.complete is of the form <entity>.<verb>.<verb> isn't it? Or is shutdown now the "thing"?

I'll follow-up separately for this one.

Only goal of this PR is to clarify if we should use same EventName for different outcomes of an operation. 


One event:
`process.shutdown` with attribute status=Sucess/Fail , severity depends on status

vs

`process.shutdown.ok` , INFO
`process.shutdown.failed` WARN/ERROR attributes reason, error_code, ....

We currently have former, and this PR is changing to the latter.

Both approaches are valid, but distinct event names better reflect that a successful shutdown and a failed shutdown are fundamentally different event types with different operational significance and mental models—success is a routine lifecycle transition while failure is an anomalous condition requiring investigation.

This could be a time consuming discussion without much returns! Lets give it another day to see if anyone has strong preference either way.

lquerel:
There are indeed a few ambiguities in our guidelines around events that need clarification.

I believe the semantic conventions lean toward Approach 1, but I'm adding @lmolkova  and @jsuereth  (semconv and weaver maintainers) to get their perspective on this topic.

Concretely, my thinking is that the event name would be:

```
<ns>.node.shutdown
```

where `node` represents the thing, and the attributes would include:

```
<ns>.shutdown.outcome = "success" | "failure"
error.type            # when severity is ERROR or WARN
error.message.   # when severity is ERROR or WARN
```

I've intentionally left the `<ns>` placeholder undefined for now, since there is still some discussion about which namespace we should use for this project (`otelcol` or something else).

cijothomas:
https://github.com/open-telemetry/opentelemetry-dotnet/blob/main/src/OpenTelemetry.Exporter.OpenTelemetryProtocol/Implementation/OpenTelemetryProtocolExporterEventSource.cs#L98-L114
In OTel .NET's internal logging, we follow Approach 2 - normal export vs failed export are distinct events. Not saying that is the best, but it has certainly influenced my proposal. Here's why:
Separately EventName would help when we eventually use weaver like tools to auto-generate code for reporting each event. With distinct event name, it'll be easier to enforce things - node.shutdown.fail event's generated code can enforce passing an error.type.

lquerel:
@cijothomas 

> Here's why: Separately EventName would help when we eventually use weaver like tools to auto-generate code for reporting each event. With distinct event name, it'll be easier to enforce things - node.shutdown.fail event's generated code can enforce passing an error.type.

An interesting argument to take into consideration.

jmacd:
I think we should merge this. I do not think we should wait for the philosophical question or OTel to change its mind. 
Mostly, it's because I want #1988 to merge!

rust/otap-dataflow/docs/telemetry/events-guide.md
```diff
@@ -172,11 +175,10 @@ Optionally, add occurrence-specific attributes (dynamic context):
...
 When events are exported as logs, set an appropriate severity.
 
-Regarding severity, some events may be logged at different levels depending on
-their severity or impact. For example, a `node.shutdown` event may be logged at
-INFO level during a graceful shutdown, but at ERROR level if the shutdown is due
-to a critical failure. When exporting events as logs, choose the log level that
-best reflects the significance of the event.
+Regarding severity, choose the log level that best reflects the significance of
+the event. For example, `node.shutdown.complete` at INFO for a graceful
+shutdown and `node.shutdown.fail` at ERROR for a critical failure -- these are
```
lmolkova, line 180: this is not a formal guidance (at least not yet), but event naming follows the same principles as metric naming - one name for specific kind of event, outcome recorded as an attribute.

E.g. `node.shutdown` with attribute `error.type = something_specific`. 

Caveat: assuming there are events along the way, during the process of shutdown, then there could be 
`node.shutdown.error` event(s) and the process of shutdown can also be recorded as a span (or an event if spans don't work). 
**TL;DR: `node.shutdown` event represents the shutdown process with any outcome (success, failure, something in betwwen). `node.shutdown.error` represents a specific occurrence of an error during shutdown process. Naming comes from what you want to record.**

Severity considerations (in review): https://github.com/open-telemetry/semantic-conventions/pull/3311
**TL;DR: if it's the end of the world situation - FATAL, if it clearly affects user experience - ERROR, if it's retriable and effect on user experience is minimal/not known - WARN or below**

cijothomas: Thanks for sharing this @lmolkova. When looking at the OTel Collector's internal metrics, I see they use separate metric names for different outcomes:
otelcol_receiver_accepted_spans and otelcol_receiver_refused_spans
Rather than:
otelcol_receiver_spans{status="accepted|refused"}

Of course it is Collector, and while it was being built, semantic conventions may not have existed. Anyway, I can open an issue with sem.conv repo to continue this discussion in a better forum for this.

cijothomas: > TL;DR: node.shutdown event represents the shutdown process with any outcome (success, failure, something in betwwen). node.shutdown.error represents a specific occurrence of an error during shutdown process. Naming comes from what you want to record.

This is quite reasonable.

lmolkova: I believe there is an issue on collector - https://github.com/open-telemetry/opentelemetry-collector/issues/14350. Having arrow interested in a similar set of metrics is a good justification to move the issue to semconv (or create a new one).

We've gone through a round of internal OTel SDK metric renames recently - https://github.com/open-telemetry/semantic-conventions/blob/main/docs/otel/sdk-metrics.md and I imagine collector would do something similar at some point

jmacd: SDK and Collectors/pipelines can use the same metrics, but I've begun to see the SDK case as special compared with the pipeline case.

@lmolkova please see https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md which I have been working to emulate.

lquerel: Thank you, @lmolkova, for the clarification. We will incorporate this into our guidelines.

Issue #1950, opened 2026-02-04
Internal telemetry receiver and pipeline for metrics


jmacd:
The current configuration for self-diagnostic metrics is confusingly split between `settings::telemetry` and `service::telemetry::metrics`.

In #1936 we have cleaned this up for logging by removing the OpenTelemetry SDK for logging (only). 

This issue is to do the same configuration cleanup for the metrics SDK,

1. to move non-OTel metrics configuration into settings::telemetry::metrics
2. to move towards removing the OTel SDK in favor of writing directly to ITR

If we will keep the OTel SDK, as a separate issue, we may consider renaming `service::telemetry` to `opentelemetry` or `otel_sdk` indicating it is an area exclusive to the OTel SDK.

PR #2011, opened 2026-02-10
otap_df_otap::pdata::Context remove source_node, use last frame's node_id

jmacd:
# Change Summary

Removes the `Option<Cow<str>>` field in `OtapPdata`.

Uses the `.stack.last().map(|frame| frame.node_id)` i.e., the last node's frame.

This makes a simpler/smaller `OtapPdata` at the expense of always adding a Frame.

I think this makes sense because for us to implement the Collector's specification for component-level telemetry (with producer/consumer counts, with outcome attributes) requires maintaining a small amount of state for every node that opts in. The current PR does not support an "opt-in" concept, so all nodes produce a frame and the current user of source_node information (validation logic) continues to work. As a TODO for the future, we can add this opt-in mechanism, for example to let users disable this frame behavior when it is not useful (thus disabling certain metrics).

See https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md

Renames `current_calldata()` which was confusingly named and used in two real cases (retry behavior w/ delayed data) and some test cases. Now that we have `source_node()`, the semantics are more clear if we name this `source_calldata()`. Added comments to indicate that this changes after a call to subscribe_to() otherwise is automatically maintained.

Part of #2018 
Part of #1950 
Historical connection with #487 
Follows #1899 

New tests.

No.

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/processor.rs
+++ b/rust/otap-dataflow/crates/engine/src/processor.rs
@@ -326,6 +326,7 @@ impl<PData> ProcessorWrapper<PData> {
                 pdata_senders,
                 pdata_receiver,
                 user_config,
+                source_tag,
                 ..
             } => {
                 let message_channel = MessageChannel::new(
@@ -338,12 +339,13 @@ impl<PData> ProcessorWrapper<PData> {
                     })?,
                 );
                 let default_port = user_config.default_output.clone();
-                let effect_handler = local::EffectHandler::new(
+                let mut effect_handler = local::EffectHandler::new(
                     node_id,
                     pdata_senders,
                     default_port,
                     metrics_reporter,
                 );
+                effect_handler.set_source_tagging(source_tag);
                 Ok(ProcessorWrapperRuntime::Local {
                     processor,
                     effect_handler,
@@ -357,6 +359,7 @@ impl<PData> ProcessorWrapper<PData> {
                 pdata_senders,
                 pdata_receiver,
                 user_config,
+                source_tag,
                 ..
             } => {
                 let message_channel = MessageChannel::new(
@@ -369,12 +372,13 @@ impl<PData> ProcessorWrapper<PData> {
                     })?),
                 );
                 let default_port = user_config.default_output.clone();
-                let effect_handler = shared::EffectHandler::new(
+                let mut effect_handler = shared::EffectHandler::new(
                     node_id,
                     pdata_senders,
                     default_port,
                     metrics_reporter,
                 );
+                effect_handler.set_source_tagging(source_tag);
                 Ok(ProcessorWrapperRuntime::Shared {
                     processor,
                     effect_handler,
--- a/rust/otap-dataflow/crates/engine/src/receiver.rs
+++ b/rust/otap-dataflow/crates/engine/src/receiver.rs
@@ -310,6 +310,7 @@ impl<PData> ReceiverWrapper<PData> {
                     control_receiver,
                     pdata_senders,
                     user_config,
+                    source_tag,
                     ..
                 },
                 metrics_reporter,
@@ -326,13 +327,14 @@ impl<PData> ReceiverWrapper<PData> {
                 };
                 let default_port = user_config.default_output.clone();
                 let ctrl_msg_chan = local::ControlChannel::new(Receiver::Local(control_receiver));
-                let effect_handler = local::EffectHandler::new(
+                let mut effect_handler = local::EffectHandler::new(
                     node_id,
                     msg_senders,
                     default_port,
                     pipeline_ctrl_msg_tx,
                     metrics_reporter,
                 );
+                effect_handler.set_source_tagging(source_tag);
                 receiver.start(ctrl_msg_chan, effect_handler).await
             }
             (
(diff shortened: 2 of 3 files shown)


jmacd:
@c1ly please take a look

c1ly:
LGTM. Since we are keeping track of all nodes visited in the context stack this could allow us to add some visualization on the front end to show the signal path, I know that this is something @lquerel is interested in

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -50,9 +47,14 @@ impl Context {
...
     ) {
-        if let Some(last) = self.stack.last() {
+        if let Some(last) = self.stack.last_mut() {
             // Inherit the preceding frame's RETURN_DATA bit
             interests |= last.interests & Interests::RETURN_DATA;
+
+            if node_id == last.node_id {
+                last.interests = interests;
+                return;
+            }
```
lquerel: Is it to protect the pdata API from calling multiple times the `subscribe_to` method from the same node?

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -117,21 +119,30 @@ impl Context {
         self.stack.last().map(|f| f.calldata.clone())
     }
```
lquerel, line 123: Now that frames without declared interest and return_data are stacked on `self.stack` should we change the behavior of these 2 methods?

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -50,9 +47,14 @@ impl Context {
...
         node_id: usize,
     ) {
-        if let Some(last) = self.stack.last() {
+        if let Some(last) = self.stack.last_mut() {
             // Inherit the preceding frame's RETURN_DATA bit
             interests |= last.interests & Interests::RETURN_DATA;
+
+            if node_id == last.node_id {
+                last.interests = interests;
+                return;
```
lquerel: What happens to the `calldata` parameter when `node_id` == `last.node_id` ?

rust/otap-dataflow/crates/engine/src/lib.rs
```diff
@@ -381,6 +381,18 @@ impl<PData: 'static + Clone + Debug> PipelineFactory<PData> {
...
+                .node_iter()
+                .enumerate()
+                .map(|(idx, (name, _))| (name.clone(), idx))
+                .collect(),
+        );
+        pipeline_ctx.set_node_names(node_names);
+
         // Create runtime nodes based on the pipeline configuration.
         // ToDo(LQ): Collect all errors instead of failing fast to provide better feedback.
         for (name, node_config) in config.node_iter() {
```
lquerel, line 476: Here we rely on the implicit ordering behavior of `node_iter()` on a HashMap which for the same hashmap instance without changes is probably the same between to call on `node_iter()` but I don't there is any guarantee at the API level of the HashMap. Can we build the `node_names` inside the existing for loop to avoid this implicit behavior?

rust/otap-dataflow/crates/otap/src/pdata.rs
jmacd: That's what it was -- and thanks, I realized this is not appropriate. I added a debug_assert_ne!() to make sure a node doesn't accidentally subscribe on top of itself, since I expect that would break something like automatic metrics.

rust/otap-dataflow/crates/otap/src/pdata.rs
jmacd: Thanks for pointing at this -- the first is still accurate, and the second gave me confusion. To try to improve the confusion: (a) I added comments, (b) renamed this function to `source_calldata` so that this method matches naming for source_node.

rust/otap-dataflow/crates/otap/src/pdata.rs
jmacd: Changed to avoid this scenario:
```
            // We should never subscribe twice.
            debug_assert_ne!(node_id, last.node_id);
```

rust/otap-dataflow/crates/engine/src/lib.rs
jmacd: Fixed! 🤦🤦
The  map has to be computed in two passes.

rust/otap-dataflow/crates/engine/src/local/receiver.rs
```diff
@@ -123,7 +123,7 @@ impl<PData> ControlChannel<PData> {
 /// A `!Send` implementation of the EffectHandler.
 #[derive(Clone)]
 pub struct EffectHandler<PData> {
-    core: EffectHandlerCore<PData>,
+    pub(crate) core: EffectHandlerCore<PData>,
```
lquerel: Why not keeping this field private when you already expose the set_needs_source_tag?

lquerel: Same for the other receiver impls.

lquerel (approved):
LGTM. 
Just one nit comment.

rust/otap-dataflow/crates/engine/src/local/receiver.rs
jmacd: Fixed.

PR #2013, opened 2026-02-10
fix: error applying transport optimized encoding to plain encoded String/Binary types

### LINKED ISSUE #1974: Adding transport encoding fails with plain encoded string values
```rust
let schema = Arc::new(Schema::new(vec![
            Field::new(consts::PARENT_ID, DataType::UInt16, false)
                .with_encoding(metadata::encodings::PLAIN),
            Field::new(consts::ATTRIBUTE_TYPE, DataType::UInt8, false),
            Field::new(consts::ATTRIBUTE_KEY, DataType::Utf8, false),
            Field::new(consts::ATTRIBUTE_STR, DataType::Utf8, true),
        ]));

        let batch = RecordBatch::try_new(
            schema,
            vec![
                Arc::new(UInt16Array::from_iter_values([1, 1, 1, 1,1,  1, 1])),
                Arc::new(UInt8Array::from_iter_values([
                    AttributeValueType::Str as u8,
                    AttributeValueType::Str as u8,
                    AttributeValueType::Str as u8,
                    AttributeValueType::Str as u8,
                    AttributeValueType::Str as u8,
                    AttributeValueType::Str as u8,
                ])),
                Arc::new(StringArray::from_iter_values([
                    "b", "b", "b", "c", "a", "a", "a",
                ])),
                Arc::new(StringArray::from_iter_values([
                    "d", "f", "e", "g", "a", "c", "b",
                ])),
            ],
        )
        .expect("record batch OK");

        let (result, _) = apply_transport_optimized_encodings(&ArrowPayloadType::LogAttrs, &batch)
            .expect("transport optimize encoding apply");
    }
```

```
thread 'otap::transform::insert_tests::test' (689003) panicked at crates/pdata/src/otap/transform.rs:7714:10:
record batch OK: InvalidArgumentError("all columns in a record batch must have the same length")
test otap::transform::insert_tests::test ... FAILED
```

Fortunately, most of the time these will be dict encoded, but we should fix this ASAP

albertlockett:
# Change Summary

Properly track the current offsets when appending segments of the sorted values column when the type is a plain encoded String/Binary array.

Prior to this fix, we'd have an error when we built the final value column if more than one non-null segment of the array was appended. This would happen because we appended offsets generated from slices of the values array, which were offset from the start of the slice, not from the start of the array. This would cause a non-monotonically increasing offsets array for the resulting `StringArray`/`BinaryArray`, which is invalid.

* Closes #1974 

New unit tests

No.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
@@ -2946,6 +2946,84 @@ mod test {
     }
 
     #[test]
+    fn test_sort_and_apply_transport_delta_encoding_for_attr_plain_encoded_string_values_with_nulls()
+     {
+        let input = RecordBatch::try_new(
+            Arc::new(Schema::new(vec![
+                Field::new(consts::PARENT_ID, DataType::UInt16, false)
+                    .with_encoding(consts::metadata::encodings::PLAIN),
+                Field::new(consts::ATTRIBUTE_TYPE, DataType::UInt8, false),
+                Field::new(consts::ATTRIBUTE_KEY, DataType::Utf8, false),
+                Field::new(consts::ATTRIBUTE_STR, DataType::Utf8, true),
+            ])),
+            vec![
+                Arc::new(UInt16Array::from_iter_values([1, 1, 1, 1, 1, 1, 1])),
+                Arc::new(UInt8Array::from_iter_values([
+                    AttributeValueType::Str as u8,
+                    AttributeValueType::Str as u8,
+                    AttributeValueType::Str as u8,
+                    AttributeValueType::Str as u8,
+                    AttributeValueType::Str as u8,
+                    AttributeValueType::Str as u8,
+                    AttributeValueType::Str as u8,
+                ])),
+                Arc::new(StringArray::from_iter_values([
+                    "b", "b", "b", "c", "a", "a", "a",
+                ])),
+                Arc::new(StringArray::from_iter([
+                    None,
+                    Some("f"),
+                    Some("e"),
+                    Some("g"),
+                    Some("a"),
+                    Some("c"),
+                    None,
+                ])),
+            ],
+        )
+        .expect("record batch OK");
+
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
```diff
@@ -2854,4 +2879,135 @@ mod test {
...
+        let result = transport_optimize_encode_attrs::<UInt16Type>(&input).unwrap();
+
+        let expected = RecordBatch::try_new(
+            Arc::new(Schema::new(vec![
+                Field::new(consts::PARENT_ID, DataType::UInt16, false)
+                    .with_encoding(consts::metadata::encodings::QUASI_DELTA),
+                Field::new(consts::ATTRIBUTE_TYPE, DataType::UInt8, false),
+                Field::new(consts::ATTRIBUTE_KEY, DataType::Utf8, false),
+                Field::new(consts::ATTRIBUTE_BYTES, DataType::Binary, true),
+                // TODO add type types as well
```
lalitb: nit - are there TODOs remaining, or it's just left-over.

rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
```diff
@@ -2854,4 +2879,135 @@ mod test {
...
+                ])),
+                Arc::new(BinaryArray::from_iter_values([
+                    b"a", b"b", b"c", b"d", b"e", b"f", b"g",
+                ])),
+            ],
+        )
+        .expect("record batch OK");
+
+        pretty_assertions::assert_eq!(result, expected);
+    }
```
lalitb, line 3089: nit - Both new tests use all non-null values. It would be valuable to add a test case where some attribute values are null.

rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs
albertlockett: oops - leftover. removed

PR #2021, opened 2026-02-11
Reindex implementation

JakeDern:
# Change Summary

This is a reimplementation of reindex as a part of the ongoing work of #1926. This now reindexes all the required columns, has support for metrics, and has support for dictionary encoded columns.

I also was able to uncomment most of the batching tests after this change 🥳. One more to go which requires split.

Other minor changes:

- Made it so that `allowed_payload_types` returns payload types in the exact same order that they are stored. This is occasionally handy to have.

Things deferred:

- Benchmarks. I had nothing to compare it to since the original didn't reindex a bunch of the necessary columns anyway like scope or resource id nor did it support dictionaries. I'll add these in when I get to the next point..
- Some optimization opportunities like using naive offsets instead of sorting and reindexing everything starting at 0. We need this path because it's possible to get into situations where we absolutely need to compact things down to fit into u16, but we can likely skip it a decent portion of the time.

Part of #1926.

I added a big unit test suite.

No.


rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
```diff
@@ -0,0 +1,2159 @@
...
+    let mut remaining_slice = &mut sorted_ids[..];
+    for mapping in mappings.iter() {
+        if remaining_slice.is_empty() {
+            break;
+        }
+
+        // If there are elements left that come before the current mapping then these
+        // were never re-mapped. In this case we can either return an error, drop
+        // them, or try to remove them entirely from the record batch by nulling
+        // them out.
```
JakeDern: @albertlockett @lquerel I'm very interested to hear your thoughts on what to do here.

This is a referential integrity violation as-in we have primary ids in one table which define what actually exists and then a child table which references that column made up IDs that don't exist in the parent.

If we ignore them, then we risk messing with things that they are batched with as we could remap those ids into a range that is valid and would join them to some other record. 

That would be bad, so I think we should avoid that scenario if we can. A few options:

1. We remove these rows from the parent record batch which is probably expensive
2. We try to come up with a value to assign to them so that they remain invalid and not tied to anything. We could always reserve the highest ID for this case, in exchange for -1 max batch size. But then we have a problem if someone legitimately uses that ID. 
3. We could try to null these IDs, but I don't know that we will let required id columns be nullable nor is there a guarantee the column was marked nullable in the first place so we might have to edit all of that.

Will capture an issue to follow up on it whatever we decide.

rust/otap-dataflow/crates/pdata/src/otlp/metrics/data_points/exp_histogram.rs
```diff
@@ -30,7 +30,7 @@ use arrow::datatypes::{DataType, Field, FieldRef, Fields, UInt64Type};
 
 #[allow(missing_docs)]
 pub struct ExpHistogramDpArrays<'a> {
-    pub id: Option<&'a UInt32Array>,
+    pub id: Option<UInt32ArrayAccessor<'a>>,
```
JakeDern: @jmacd For awareness on these changes to support dictionary encodings for these IDs. Everything seemed to still just work after I did this and the encode/decode stuff seemed to only depend on APIs that the Accessors also supported.

rust/otap-dataflow/crates/pdata/src/otap.rs
```diff
@@ -466,6 +492,8 @@ impl OtapBatchStore for Metrics {
 
     const COUNT: usize = 19;
 
+    const ROOT_PAYLOAD_TYPE: ArrowPayloadType = ArrowPayloadType::UnivariateMetrics;
```
albertlockett: We might need to revisit this if/when we support `MultivariateMetrics`. If the OTAP batch is multivariate metrics, that would be the root payload type, and I think the implementation here would look in the wrong position for the root batch.

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
```diff
@@ -0,0 +1,2159 @@
...
+            max: u16::MAX as usize,
+            message: "Too many items to reindex".to_string(),
+        });
+    }
+
+    // Iterate over all allowed payload types for this signal
+    for &payload_type in S::allowed_payload_types() {
+        for rb in store.select_mut(payload_type) {
+            *rb = remove_transport_optimized_encodings(payload_type, rb)?;
+        }
```
albertlockett: Would it be a helpful addition in the future to have MultiBatchStore contain `&mut [S]` (e.g. contain `OtapBatchStore`)?

Then we could call helper methods for this decoding:
https://github.com/open-telemetry/otel-arrow/blob/70c62ad23a1d932f7e95bf93f57d4c86c82927c3/rust/otap-dataflow/crates/pdata/src/otap.rs#L190-L192

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
```diff
@@ -0,0 +1,2159 @@
...
+            Err(_) => {
+                // No ID column, put the batch back and continue
+                store.get_mut(i)[parent_idx] = Some(parent_batch);
+                continue;
+            }
+        };
+
+        // Create mappings for the parent IDs
+        let ids = materialize_id_column::<T>(id_col.as_ref())?;
+        let ids = ids.values().to_vec();
```
albertlockett: We can avoid cloning these by calling `to_vec` I think. I'm that  ScalarBuffer would deref to &[T::Native], which I'm pretty sure we can pass into sort_vec_to_indices

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
```diff
@@ -0,0 +1,2159 @@
...
+    mappings: &[IdMapping<T::Native>],
+) -> Result<RecordBatch>
+where
+    T: ArrowPrimitiveType,
+    T::Native: Ord + Copy + Add<Output = T::Native> + AddAssign + SubAssign + ArrowNativeType,
+{
+    // Extract ID column
+    let id_col = extract_id_column(&batch, column_path)?;
+    let ids = materialize_id_column::<T>(id_col.as_ref())?;
+    let mut ids = ids.values().to_vec();
```
albertlockett, line 372: These will need to be the plain encoded IDs for this to work, right?

Here we've removed any encoding for the parent, but we haven't yet removed it for the child batch.
https://github.com/open-telemetry/otel-arrow/blob/ef203f2acfcae72b6b054734d8cb1ba94118386a/rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs#L124-L126

And It looks like we're calling this function with the child batch here:
https://github.com/open-telemetry/otel-arrow/blob/ef203f2acfcae72b6b054734d8cb1ba94118386a/rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs#L282-L285

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
albertlockett: > remove these rows from the parent record batch

Wouldn't we remove the rows from the _child_ record batch? IIUC, it's the child record batch that has an ID in it's parent_id column pointing back to an ID that doesn't exist in the parent?

My gut tells me the "right" thing to do is Option #1 - remove the rows. I believe it would be very unusual for us to receive batches like this and it indicates something has gone wrong upstream. IMO, it's OK to have a slower path to correct bad data (which we shouldn't have received in the first place).

Option #2 could also work. It promulgates the bad data issue, but it doesn't make it any worse and it would be more performant than removing the rows. Instread of making `T::Native::MAX - 1` the sentinel value for invalid ID, we could just choose some value that's not used in `IdMapping` couldn't we? Then we wouldn't pay the penalty of -1 max batch size. Also if there's no unused value in `IdMapping`, it means we're using the full ID range so no ID column can be invalid in the child batch.

WRT to option #3, I don't think we should use nulls for this. We expect the parent_id column not to be nullable.

rust/otap-dataflow/crates/pdata/src/otlp/metrics/data_points/exp_histogram.rs
albertlockett: @JakeDern was this change necessary?

With U32 IDs, I thought we were only use dictionary encoding for parent_id columns, not "id" columns.

For example, ExpHistogram Datapoint attributes would have a u32 "parent_id" column pointing back at this "id" column. The "parent_id" column may be dictionary encoded, but I thought this "id" column was always a straight UInt32Array.

These "id" columns are intended to be unique, so dictionary encoding them just would just add extra overhead. By contrast, the "parent_id" column may have the same ID multiple times (for example, if the ExpHistogram Datapoint has many attributes), so dict encoding can make sense.

If we're somehow producing a dictionary encoded u32 "id" column, this seems like something we should correct.

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: Ah yes, for (1) removing the junk rows from the current/child record batch is what I meant. 

For (2) I don't think we can always find an ID to remap it to that isn't used across any BAR we're trying to reindex. We have to avoid colliding with IDs from any other BAR or risk associating this junk data with a record batch from that other BAR. 

The current mappings will be for some relation in the current BAR. We can assume that all IDs prior to the `start_id` of the first mapping are used, since we are mapping to contiguous ranges starting at 0. We can also assume there will be no holes left through the end of the mapped range (defined as `(first_mapping.start_id +/- first_mapping.offset) + sorted_ids.len()` for the same reason. That only leaves IDs after the current mapped range, but we might end up using them all as we process the next BAR(s). 

We could retool the code to calculate all mappings up front along so we know the highest ID we will need and apply the optimization if there are any left. Two notes on that though:

- The optimization for doing a naive remapping by finding the lowest/highest value in each record batch, computing the offset, and applying it using a vectorized add/sub operation without sorting will have to do a similar calculation
- That optimization will also be unable to detect any junk rows so we will have to accept junk rows existing as a possibility

What do you think about the following:

1. Implement removing the bad rows now as the fallback case because I think we always need it
2. Implement the junk ID optimization and the naive remapping optimization together later since they need a similar up-front calculation

JakeDern: And I think it goes without saying, but I also don't like 3 🙂

rust/otap-dataflow/crates/pdata/src/otlp/metrics/data_points/exp_histogram.rs
JakeDern: Ah yeah, I agree with you about it being pointless for someone to do this. I put it in because I wanted to cover all the possible cases in the batching code and there's nothing today preventing a dictionary encoded id column from getting here (that I'm aware of). To do that I needed support in this code for it since my tests leverage conversions that use these models.

I guess the question is - When we define the OTAP spec, is dictionary encoding these IDs forbidden or discouraged and when would we enforce that?

Anyways I'm not opinionated on keeping this in, I can scrap this and the associated tests and file it under "cases that we will soon prevent from happening upstream of this code".

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: Yeah could be, I'm still playing around with how to deal with multiple OtapArrowRecords at a time when we want one specific `OtapBatchStore` variant. I think it will come down to what's most convenient for the callers once we get to refining the interface to the batching code. 

Another possibility is instead of having to unwrap OtapArrowRecords and re-collect into some structure of Logs/Metrics/Traces, we could accept any `impl FixedSizeIterator<Item = &mut T> where T: OtapBatchStore` and then if we had an existing `Vec<OtapArrowRecords>` where we know they're all the same variant it would be easy to process that in place.

That said as I'm writing this, I'm remembering that I might need to sort these by min/max ids in some instances and jump around a bit.

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: I'll give that a try!

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: Yeah good call, let me add some test cases for this. I think our data generation code generally doesn't apply transport optimized encodings, but that's an easy thing to add.

rust/otap-dataflow/crates/pdata/src/otap.rs
JakeDern: Ah yeah good point - I'll try to eliminate this in favor of just using a more general construct because all I'm trying to do is bounds checking on the IDs, but we're not handling the u32 id columns with this anyway.

JakeDern: This got me thinking - We won't really be able to determine how many IDs we have for resource and scope ahead of time by just checking the length since they're many to many and reference a sort of invisible resource and scope entity that don't have their own table.

Since we're checking the id column in Logs/Metrics/Traces, we don't have to worry about resource.id and scope.id since those are bounded by the length of the record batch/length of the id column. But it's possible for someone to add a bunch of rows with junk `parent_id` to the ResourceAttrs, ScopeAttrs, or Log/Metrics/TracesAttrs tables and cause us problems if we're not careful.

In those cases, if we're doing the slow version and compacting everything to a contiguous range then we will detect and remove any bad rows and be fine. If we're doing the naive offset version then we probably need to check not just the primary id columns, but also the parent references and see if they disagree about the min/max which is what could push us over.

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: I refactored this code a bit and saved some allocations. We still call `to_vec` here, but now it saves us calling it again later so same savings. One clone of the original ids is unavoidable because we have to get ownership and mutate them all, but we just do one now.

rust/otap-dataflow/crates/pdata/src/otap.rs
albertlockett: ohhh yeah that's a good point, you're 100% right about that. 

FWIW the parquet_exporter has the exact problem you describe:
https://github.com/open-telemetry/otel-arrow/blob/70c62ad23a1d932f7e95bf93f57d4c86c82927c3/rust/otap-dataflow/crates/otap/src/parquet_exporter/idgen.rs#L71-L111

I don't think we have any code that would purposely leave attributes record batches w/ parent IDs pointing at nothing. I'd also like to just believe that all users are benevolent saints who wouldn't craft evil attributes record batches 😇. We should probably fix it anyway though

JakeDern: If everyone was just trustworthy programming would be so much easier 😅

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
albertlockett: > What do you think about the following:
>
> 1.  Implement removing the bad rows now as the fallback case because I think we always need it
> 2. Implement the junk ID optimization and the naive remapping optimization together later since they need a similar up-front calculation

Yup, this sounds like a good plan!

rust/otap-dataflow/crates/pdata/src/otlp/metrics/data_points/exp_histogram.rs
albertlockett: I'm of the opinion that we should forbid using dict encoded IDs. It just seems like extra complexity to support it, with no benefit.

There are other places in the code we assume the ID column is not dict encoded. One other places is when we remove delta encoding.
https://github.com/open-telemetry/otel-arrow/blob/70c62ad23a1d932f7e95bf93f57d4c86c82927c3/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L66-L73

In terms of where this _should_ get validated, TBD :). I'd love to see our record batches get validated all in one place, but we don't do that yet, and the structure just kind of gets validated piece by piece as it goes along.

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
albertlockett: Sounds good. I'm not opposed to the type we're currently using btw, just figured I'd call out that `OtapBatchStore` is available if it makes life easier.

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
albertlockett: Sounds good - yeah it can be added by calling `encode_transport_optimized` on `&mut OtapArrowRecords`. 

Currently the only place this is called, outside of tests, is right before we serialize the `RecordBatch` to Arrow IPC before we export it
https://github.com/open-telemetry/otel-arrow/blob/ef203f2acfcae72b6b054734d8cb1ba94118386a/rust/otap-dataflow/crates/pdata/src/encode/producer.rs#L250

rust/otap-dataflow/crates/pdata/src/otlp/metrics/data_points/exp_histogram.rs
JakeDern: Sounds good, unless anyone thinks otherwise I'll revert the changes there and remove the tests for these specific cases, though they'll still work for now. 

At some point soon we'll hopefully validate these properties upstream of this code between the work you're doing to provide an API for working with otap payloads and the work I want to do to add more validations on OtapBatchStore implementations :)

JakeDern: And wouldn't you know it, once I apply the transport optimized encodings to my test data I start seeing things like this. Will definitely remove since other code is already making this assumption.

image

JakeDern: ~Ok have some dedicated transport optimized tests now and I just remove the encoding on everything up front in the implementation. Resolving this.~

Got my comments mixed up, but this is still resolved with tests :).

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: Have some dedicated transport optimized tests now and I just remove the encoding on everything up front in the implementation. Resolving this

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: Added this to #1926 and I have the fallback implemented for removing junk IDs. Resolving.

rust/otap-dataflow/crates/pdata/src/otap/batching_tests.rs
```diff
@@ -102,19 +102,18 @@ fn generate_metrics_batching_test_cases() -> Vec<MetricsBatchingTestCase> {
...
-    //         &format!("all_types_small_limit_{}", limit),
-    //         MetricsConfig::new()
-    //             .with_gauges(vec![2, 3])
-    //             .with_sums(vec![2])
-    //             .with_histograms(vec![1, 2])
-    //             .with_summaries(vec![1]),
-    //         limit as u64,
-    //         3,
-    //     );
-    // }
```
jmacd, line 117: Nice!

rust/otap-dataflow/crates/pdata/src/otap.rs
jmacd: I'm glad I read to the end of this PR 💯

jmacd (approved):

rust/otap-dataflow/crates/pdata/src/otap/batching_tests.rs
JakeDern: Almost got them all, hopefully soon!

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
```diff
@@ -0,0 +1,2502 @@
...
+fn extract_id_column(rb: &RecordBatch, column_path: &str) -> Result<ArrayRef> {
+    access_column(column_path, &rb.schema(), rb.columns()).ok_or_else(|| Error::ColumnNotFound {
+        name: column_path.to_string(),
+    })
+}
+
+/// Sorts a vector of values and returns the resulting sort indices
+fn sort_vec_to_indices<T: Ord>(values: &[T]) -> Vec<u32> {
+    let mut indices: Vec<u32> = (0u32..values.len() as u32).collect();
+    indices.sort_by_key(|&i| &values[i as usize]);
```
utpilla: Is it safe to use unstable sorting here?

```suggestion
    indices.indices.sort_unstable_by_key(|&i| &values[i as usize]);(|&i| &values[i as usize]);
```

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
```diff
@@ -0,0 +1,2502 @@
...
+                child_types: &[ArrowPayloadType::SpanEventAttrs],
+            }],
+        },
+        ArrowPayloadType::SpanLinkAttrs => PayloadRelationInfo {
+            primary_id: None,
+            relations: &[Relation {
+                key_col: ID,
+                child_types: &[ArrowPayloadType::SpanEventAttrs],
+            }],
+        },
```
utpilla: This should be `SpanLinks` with ID having `SpanLinkAttrs` as child, right?

```suggestion
        ArrowPayloadType::SpanLinks => PayloadRelationInfo {
            primary_id: Some(PrimaryIdInfo {
                name: ID,
                size: IdColumnType::U32,
            }),
            relations: &[Relation {
                key_col: ID,
                child_types: &[ArrowPayloadType::SpanLinkAttrs],
            }],
        },
```

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
```diff
@@ -0,0 +1,2502 @@
...
+        // test along with `-- --no-capture`
+        // FIXME: Comment out
+        for (idx, b) in batches.iter().enumerate() {
+            use arrow::util::pretty;
+            eprintln!("-----Batch #{}------", idx);
+            for rb in b.iter().flatten().cloned() {
+                eprintln!("{}", pretty::pretty_format_batches(&[rb]).unwrap());
+            }
+            eprintln!("-----End Batch #{}------", idx);
+        }
```
utpilla: nit: It looks like we want this to be commented out and only use it for debugging.

JakeDern: Yep, still finalizing a couple changes from Alberts feedback - Will take this out at the end.

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: Yep should be ok, we only sort one time and then use that sort throughout. Will make the commit locally!

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: Wow good eye! I must have taken the wrong completion on that one. Fixed

rust/otap-dataflow/crates/pdata/src/otap.rs
JakeDern: Ok we have the bounds checking in for all the primary ID columns, just missing the edge cases with the resource/scope ids. I have that filed as a follow up along with the optimizations. I took that other root payload stuff out

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: Removed

PR #2024, opened 2026-02-11
feat: Add upsert action to Attribute Processor

### LINKED ISSUE #2015: Attribute Processor - add ability to upsert attributes
In https://github.com/open-telemetry/otel-arrow/pull/1737 we added the ability to insert new attributes. This implementation follows the convention of the OpenTelemetry Collector's Attribute Processor implementation where it will not replace existing attributes if they exist.

We should add the capability to upsert attributes, wherein we'll either insert a new attribute, or update the value of an existing attribute if it exists.

**Performance Considerations:**

The easiest way to do this is probably to treat it like a delete then insert, but this might be less optimal than updating the values column in place when we can. 

For example, if the values column is dictionary encoded (as it may be for string, binary, and int column types), we can just copy the dict key's data buffer, update the values in place, and then possibly add create new dictionary values with the appended key (assuming no dictionary overflow). For non dict encoded, fixed with values columns types like float and boolean, we can do something similar - just copy the data buffer and update the value in place. We'll also need to update the null buffers.

We'll likely want to have benchmarks for the implementation.

ThomsonTan:
# Change Summary

Adds the ability to `upsert` attributes in the Attribute Processor. An `upsert` either inserts a new attribute or updates the value of an existing attribute if one already exists — unlike `insert`, which skips keys that are already present.

Here are the benchmark results on my devbox (optimized build, fresh baseline):

| Scenario | 128 plain | 1536 plain | 8192 plain | 128 dict | 1536 dict | 8192 dict |
|---|---|---|---|---|---|---|
| `upsert_new_key` | 36.3 µs | 324 µs | 1.93 ms | 37.8 µs | 321 µs | 1.80 ms |
| `upsert_existing_key` | 35.4 µs | 320 µs | 1.90 ms | 42.1 µs | 309 µs | 1.72 ms |
| `upsert_with_delete` | 40.3 µs | 286 µs | 1.61 ms | 37.6 µs | 248 µs | 1.28 ms |
| `upsert_multiple_keys` | 38.4 µs | 371 µs | 2.18 ms | 42.9 µs | 295 µs | 1.58 ms |

Scaling seems linear (128→8192 is 64× rows, ~50× time). Insert and update paths are essentially the same cost now that `Upsert` is a first-class `KeyTransformRangeType` variant - no extra pre-scan or set merging. Dictionary-encoded keys are competitive or faster at scale (e.g. `upsert_with_delete` at 8192: 1.28 ms dict vs 1.61 ms plain).

* Closes #2015

### DIFF
--- a/.github/workflows/project_management_comment.yml
+++ b/.github/workflows/project_management_comment.yml
@@ -1,0 +1,27 @@
+name: Add comment
+on:
+  issues:
+    types:
+      - labeled
+permissions:
+  contents: read
+
+jobs:
+  add-comment:
+    if: github.event.label.name == 'help wanted'
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@58077d3c7e43986b6b15fba718e8ea69e387dfcc # v2.15.1
+        with:
+          egress-policy: audit
+
+      - name: Add comment
+        uses: peter-evans/create-or-update-comment@e8674b075228eee787fea43ef493e45ece1004c9
+        with:
+          issue-number: ${{ github.event.issue.number }}
+          body: |
+            This issue is available for anyone to work on. Feel free to leave a comment if you'd like to be assigned. **Make sure to reference this issue in your pull request.**
+            :sparkles: Thank you for your contribution! :sparkles:
--- a/rust/otap-dataflow/crates/admin/src/lib.rs
+++ b/rust/otap-dataflow/crates/admin/src/lib.rs
@@ -14,6 +14,7 @@ use axum::Router;
 use std::net::SocketAddr;
 use std::sync::Arc;
 use tokio::net::TcpListener;
+use tokio::sync::Mutex;
 use tokio_util::sync::CancellationToken;
 use tower::ServiceBuilder;
 
@@ -34,7 +35,7 @@ struct AppState {
     metrics_registry: TelemetryRegistryHandle,
 
     /// The control message senders for controlling pipelines.
-    ctrl_msg_senders: Vec<Arc<dyn PipelineAdminSender>>,
+    ctrl_msg_senders: Arc<Mutex<Vec<Arc<dyn PipelineAdminSender>>>>,
 }
 
 /// Run the admin HTTP server until shutdown is requested.
@@ -48,7 +49,7 @@ pub async fn run(
     let app_state = AppState {
         observed_state_store: observed_store,
         metrics_registry,
-        ctrl_msg_senders,
+        ctrl_msg_senders: Arc::new(Mutex::new(ctrl_msg_senders)),
     };
 
     let app = Router::new()
--- a/rust/otap-dataflow/crates/admin/src/pipeline_group.rs
+++ b/rust/otap-dataflow/crates/admin/src/pipeline_group.rs
@@ -105,9 +105,8 @@ async fn shutdown_all_pipelines(
     );
 
     // Send shutdown message to all pipelines
-    let errors: Vec<_> = state
-        .ctrl_msg_senders
-        .iter()
+    let errors: Vec<_> = (*state.ctrl_msg_senders.lock().await)
+        .drain(..)
         .filter_map(|sender| {
             // Use the timeout from params for the shutdown deadline
             let deadline = Instant::now() + Duration::from_secs(params.timeout_secs);
--- a/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/auth.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/auth.rs
@@ -82,7 +82,10 @@ impl Auth {
                     return Ok(token);
                 }
                 Err(e) => {
-                    otel_warn!("azure_monitor_exporter.auth.get_token_failed", attempt = attempt, error = %e);
+                    let error_msg = e.to_string();
+                    let first_line = error_msg.lines().next().unwrap_or(&error_msg);
+                    otel_warn!("azure_monitor_exporter.auth.get_token_failed", attempt = attempt, error = %first_line);
+                    otel_debug!("azure_monitor_exporter.auth.get_token_failed.details", attempt = attempt, error = %e);
                     self.metrics.borrow_mut().add_auth_failure();
                 }
             }
--- a/rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
+++ b/rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
@@ -340,6 +340,11 @@ impl<PData: 'static + Debug + Clone + ReceivedAtNode + Unwindable> RuntimePipeli
             node_metric_handles[id] = Some(handles);
         }
 
+        // Drop the original sender so the channel closes when all node tasks complete.
+        // Each node holds its own clone and without this drop, the PipelinCtrlMsgManager
+        // can only exit via a timeout.
+        drop(pipeline_ctrl_msg_tx);
+
         // Spawn the control-plane task that routes node control messages to the pipeline engine.
         futures.push(local_tasks.spawn_local(async move {
             let manager = PipelineCtrlMsgManager::new(
--- a/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
+++ b/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
@@ -1646,9 +1646,20 @@ mod telemetry_tests {
             let listening_port = portpicker::pick_unused_port().expect("No free ports");
             let listening_addr: SocketAddr = format!("127.0.0.1:{listening_port}").parse().unwrap();
 
-            // Receiver with metrics enabled via pipeline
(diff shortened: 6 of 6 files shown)


rust/otap-dataflow/benchmarks/benches/attribute_transform/main.rs
```diff
@@ -528,13 +534,125 @@ fn bench_transport_optimized_transform_attributes(c: &mut Criterion) {
     group.finish();
 }
 
+/// Benchmarks for upsert transforms (insert-or-update semantics).
+/// Measures both the case where keys already exist (update path) and where they don't (insert
+/// path), as well as combined upsert+delete scenarios.
+fn bench_upsert_attributes(c: &mut Criterion) {
```
albertlockett, line 565: It would be interesting to see the result of these benchmarks. Maybe we could post them in a comment on the PR?

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -858,6 +891,8 @@ pub struct TransformStats {
     pub renamed_entries: u64,
     /// Exact number of attribute entries added by insert rules
     pub inserted_entries: u64,
+    /// Exact number of attribute entries updated by upsert rules
+    pub upserted_entries: u64,
```
albertlockett, line 905: The main purpose of these stats is so we can update metrics in the attributes processor about how many rows were transformed. Could you use these to update the metrics, similar to what we do with `renamed_entries` and `deleted_entries`?

https://github.com/open-telemetry/otel-arrow/blob/75a2f71ba0765bd22358a4a2c772bf4eabc66c35/rust/otap-dataflow/crates/otap/src/attributes_processor.rs#L274-L302

https://github.com/open-telemetry/otel-arrow/blob/75a2f71ba0765bd22358a4a2c772bf4eabc66c35/rust/otap-dataflow/crates/otap/src/attributes_processor.rs#L351-L361

And also add to the test suite?
https://github.com/open-telemetry/otel-arrow/blob/75a2f71ba0765bd22358a4a2c772bf4eabc66c35/rust/otap-dataflow/crates/otap/src/attributes_processor.rs#L1516-L1530

albertlockett: Doesn't have to be part of this PR btw -- I'm fine if we do this as as a followup if you prefer

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1122,71 +1189,145 @@ pub fn transform_attributes_impl(
...
+                            upsert.entries.len() as u64
+                                * original_parent_ids.len() as u64,
+                        );
+                        // We can't precisely know how many were deleted due to upsert vs
+                        // explicit delete without more tracking. Use a conservative approach:
+                        // subtract the actual upsert insert count from deleted (since each
+                        // upsert that updated an existing key would have caused one delete).
+                        // The upserted_entries count itself tells us how many were inserted,
+                        // and we can figure out updates from context.
+                        let _ = upsert_deleted; // ignore for now
```
albertlockett: Are you sure this math works? 

For example if I had a batch like:
```
key | val | parent_id
----|----| ----------
"a" | "1" | 1
"b" | "1" | 1
"a" | "1" | 2
"c" | "1" | 3
```

```
transform = {
  delete: "b",
  upsert: { "a": "2" }
}
```
I'd delete 3 rows, then insert (via upsert) 3 to get:
```
key | val | parent_id
----|-----|-----------
"a" | "2" | 1
"a" | "2" | 2
"c" | "1" | 3
"a" | "2" | 3
```

But the final statistics I'd expect are:
```
{
  delete: 1,
  upsert: 3
}
```

I don't see how we can compute the stats accurately from the values we have in this location.

I think that probably what we need to do to the additional tracking you mention here.

albertlockett: I think the one way to do do this is probably as follows ...

Instead of creating an effective transform [here](https://github.com/open-telemetry/otel-arrow/blob/7c07d6a28f09775edd11bf0c9238f4e997968630/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L933-L956) with the upserts rolled up into the deletes, we can pass the unmodified transform directly into `transform_keys`/`transform_dictionary_keys` [here](https://github.com/open-telemetry/otel-arrow/blob/7c07d6a28f09775edd11bf0c9238f4e997968630/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L1009) and [here](https://github.com/open-telemetry/otel-arrow/blob/7c07d6a28f09775edd11bf0c9238f4e997968630/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L1069).

We'll probably need to add a new variant to the KeyTransformRange type:
https://github.com/open-telemetry/otel-arrow/blob/7c07d6a28f09775edd11bf0c9238f4e997968630/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L1945-L1949

The nice thing about doing this is, it gives us the chance to calculate the statistics based on the number of rows we transform, like we do here for replace/delete
https://github.com/open-telemetry/otel-arrow/blob/7c07d6a28f09775edd11bf0c9238f4e997968630/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L1407-L1412
and here
https://github.com/open-telemetry/otel-arrow/blob/7c07d6a28f09775edd11bf0c9238f4e997968630/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L1901-L1913



Internally to to these methods, we handle upserts like deletes for now. For example, 
- in this location where we're just choosing to delete ranges:
https://github.com/open-telemetry/otel-arrow/blob/7c07d6a28f09775edd11bf0c9238f4e997968630/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L1535-L1544
- in this location, wher we're calculating which ranges to keep from the original batch:
https://github.com/open-telemetry/otel-arrow/blob/7c07d6a28f09775edd11bf0c9238f4e997968630/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L2031-L2074


The nice thing about doing it this way is that I feel it moves us in a direction where we be able to treat upserts as deletes/inserts, which might unlock some improved performance. For example - if we detect that all our attributes already have some attribute (no need to insert new rows), we only need to adjust the value column for this attribute, instead of deleting and re-adding rows to every column.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -6845,6 +7018,167 @@ where
...
 
+/// Create a batch of upserted attributes.
+/// Unlike `create_inserted_batch`, this function does NOT check for existing keys - it
+/// unconditionally creates rows for all (parent, key, value) combinations. This is used
+/// after the upsert keys have already been deleted from the batch, so we know there are
+/// no existing entries to conflict with.
+///
+/// This function is generic over `T: ParentId` to handle different parent ID types (u16, u32)
+/// as well as dictionary-encoded parent IDs.
+fn create_upserted_batch<T>(
```
albertlockett: This is almost the same as `create_inserted_batch` except for this small block:
https://github.com/open-telemetry/otel-arrow/blob/7c07d6a28f09775edd11bf0c9238f4e997968630/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L6895-L6901

I wonder if we could cut down on the code duplication by just passing an additional parameter for whether to check for existing attributes?

Although if, in the context in which we're using this existing attributes that we're upserting have already been deleted, then do we actually need a separate function?

ThomsonTan: Removed `create_upserted_batch` entirely and switched the call site to `create_inserted_batch` with the post-deletion batch.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
ThomsonTan: Added `inserted_entries` and `upserted_entries` counters to `AttributesProcessorMetrics`. The stats now flow end-to-end just like renamed_entries and deleted_entries.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
ThomsonTan: Yes, it is an issue. Fixed by adding Upsert as a first-class variant of KeyTransformRangeType as suggested.

rust/otap-dataflow/benchmarks/benches/attribute_transform/main.rs
ThomsonTan: Pasted the benchmark result into the change summary.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
albertlockett: great! yes I see those are now returned and that we also have tests covering the returned counts. nice addition, thank you!

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
albertlockett: Thanks for making this change

albertlockett:
Thanks @ThomsonTan .

Issue #1949, opened 2026-02-04
Feature: Optional node-level and pipeline-level attributes


jmacd:
I am aware of a need for the user to configure additional attributes on nodes and pipelines.

Current internal instrumentation (logs and metrics) uses a set of key-values computed by the controller and engine during startup. The user has control over names of the node, the pipeline, and so on, and this information is included in the InstrumentationScope attributes.

I am aware of a use for additional scoping attributes. This could look something like:

```
nodes:
  receiver:
    kind: receiver
    plugin_urn: "urn:otel:otap:fake_data_generator:receiver"
    attributes:
      my-team-label: my-team-value
```

for the node, or for pipeline and pipeline-group:

```
  continuous_benchmark:
    attributes:
      group-key: group-value
    pipelines:
      traffic_gen1:
        quota: ...
        attributes:
          pipe-key: pipe-value
```

PR #2026, opened 2026-02-11
feat: Node level telemetry attributes support for logs

andborja:
# Change Summary

Node level telemetry attributes support for logs

Part of https://github.com/open-telemetry/otel-arrow/issues/1949

Unit tests + Running a local collector

### Example of a log with attributes:
2026-02-11T23:36:21.549Z  INFO  otap-df-otap::receiver.start: Fake data generator receiver started [signals_per_second=1000 signals/sec, max_batch_size=1000, metrics_per_iteration=0, traces_per_iteration=0, logs_per_iteration=1000, data_source=SemanticConventions, generation_strategy=Fresh, node_sequence=1, **pipeline_name=MyCustomPipelineName**] entity/node.attrs: node.id=generator node.urn=urn:otel:traffic_generator:receiver node.type=receiver pipeline.id=default_pipeline pipeline.group.id=default_pipeline_group core.id=0 numa.node.id=0 process.instance.id=AGOE6EAHVBYUBPD3NXBBQ7TT3Y host.id=andborja container.id=

Users can add node level telemetry attributes for logs now, so they are part of the logs

### DIFF
--- a/rust/otap-dataflow/crates/config/src/node.rs
+++ b/rust/otap-dataflow/crates/config/src/node.rs
@@ -53,7 +53,11 @@ pub struct NodeUserConfig {
     pub default_out_port: Option<PortName>,
 
     /// Node specific attributes to be added to internal telemetry.
-    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    #[serde(
+        default,
+        skip_serializing_if = "HashMap::is_empty",
+        deserialize_with = "deserialize_telemetry_attributes"
+    )]
     pub telemetry_attributes: HashMap<String, AttributeValue>,
 
     /// Node-specific configuration.
@@ -197,6 +201,25 @@ impl NodeUserConfig {
     }
 }
 
+/// Deserializes `telemetry_attributes` and rejects any attribute with an `Array` value,
+/// which is not supported for log record attributes.
+fn deserialize_telemetry_attributes<'de, D>(
+    deserializer: D,
+) -> Result<HashMap<String, AttributeValue>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let attrs: HashMap<String, AttributeValue> = HashMap::deserialize(deserializer)?;
+    for (key, value) in &attrs {
+        if matches!(value, AttributeValue::Array(_)) {
+            return Err(serde::de::Error::custom(format!(
+                "unsupported telemetry attribute type for `{key}`: array attributes are not supported"
+            )));
+        }
+    }
+    Ok(attrs)
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
@@ -236,5 +259,37 @@ config: {}
         let cfg: NodeUserConfig = serde_yaml::from_str(yaml).unwrap();
         assert!(matches!(cfg.kind, NodeKind::Processor));
         assert_eq!(cfg.out_ports.len(), 3);
+    }
+
+    #[test]
+    fn node_user_config_with_tellemetry_attributes_valid() {
+        let json = r#"{
+            "kind": "receiver",
+            "plugin_urn": "urn:example:demo:receiver",
+            "telemetry_attributes": {
+                "attr1": "value1",
+                "attr2": 123,
+                "attr3": true
+            },
+            "out_ports": {}
+        }"#;
+        let cfg: NodeUserConfig = serde_json::from_str(json).unwrap();
+        assert!(matches!(cfg.kind, NodeKind::Receiver));
+        assert!(cfg.out_ports.is_empty());
+    }
+
+    #[test]
+    fn node_user_config_with_tellemetry_attribute_array_expects_error() {
+        let json = r#"{
+            "kind": "receiver",
+            "plugin_urn": "urn:example:demo:receiver",
+            "telemetry_attributes": {
+                "attr1": "value1",
+                "attr2": [1, 2, 3]
+            },
+            "out_ports": {}
+        }"#;
+        let cfg: Result<NodeUserConfig, _> = serde_json::from_str(json);
+        assert!(cfg.is_err());
     }
 }
--- a/rust/otap-dataflow/crates/engine/src/entity_context.rs
+++ b/rust/otap-dataflow/crates/engine/src/entity_context.rs
@@ -9,6 +9,7 @@ use otap_df_config::PortName;
 use otap_df_config::pipeline::service::telemetry::AttributeValue;
 use otap_df_telemetry::metrics::{MetricSet, MetricSetHandler};
 use otap_df_telemetry::registry::{EntityKey, MetricSetKey, TelemetryRegistryHandle};
+use otap_df_telemetry::self_tracing::encode_custom_attrs_to_bytes;
 use std::cell::{Cell, RefCell};
 use std::collections::HashMap;
 use std::fmt::Debug;
@@ -97,13 +98,13 @@ pub fn node_entity_key() -> Option<EntityKey> {
         .flatten()
 }
 
-/// Returns the custom node attributes for the current node.
+/// Returns the pre-encoded custom node attribute bytes for the current node.
 ///
 /// At runtime, reads from the task-local `NODE_TASK_CONTEXT`.
 /// During build time, falls back to the thread-local `BUILD_NODE_TELEMETRY_HANDLE`.
 #[inline]
 #[must_use]
-pub fn node_custom_attrs() -> HashMap<String, AttributeValue> {
+pub fn node_custom_attrs() -> bytes::Bytes {
     // Runtime: check task-local first.
     if let Ok(attrs) = NODE_TASK_CONTEXT.try_with(|ctx| {
         ctx.telemetry_handle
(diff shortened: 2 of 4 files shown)


rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
```diff
@@ -116,6 +171,21 @@ impl LogRecord {
         let mut visitor = DirectFieldVisitor::new(&mut buf);
         event.record(&mut visitor);
 
+        // Encode custom node attributes from context, if any.
+        if !context.custom_attrs.is_empty() {
+            let mut sorted_attrs: Vec<_> = context.custom_attrs.iter().collect();
```
cijothomas: This is hot path of logging, and we should avoid allocations and expensive operations like sorting here.

Haven't looked at full details, but Nodelevel telemetry means it should be available at Node startup time itself, so we should prepare and store it (including encoding) at startup time itself, instead of doing it at hot path.

rust/otap-dataflow/crates/config/src/node.rs
```diff
@@ -51,6 +52,10 @@ pub struct NodeUserConfig {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub default_out_port: Option<PortName>,
 
+    /// Node specific attributes to be added to internal telemetry.
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub attributes: HashMap<String, AttributeValue>,
```
cijothomas: if this is only used for telemetry purposes, then could we use "telemetry_attrributes" or something to clearly indicate this has only telemetry purpose and cannot be used for anything else.
(K8s labels are used to select nodes - we don't do such capability in df engine, hence the suggestion for a more obvious name)

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
andborja: Sure. Updated.

rust/otap-dataflow/crates/config/src/node.rs
andborja: Updated.

rust/otap-dataflow/crates/engine/src/entity_context.rs
```diff
@@ -208,6 +241,11 @@ impl NodeTelemetryHandle {
         self.state.borrow().entity_key
     }
 
+    /// Return the custom node attributes for log records.
+    pub(crate) fn custom_attrs(&self) -> HashMap<String, AttributeValue> {
+        self.state.borrow().custom_attrs.clone()
+    }
+
```
lalitb: This gets called on every log event via `engine_context()` -> `node_custom_attrs()` -> here, cloning the full `HashMap` each time. The clone then gets re-encoded to protobuf in `encode_custom_attrs_to_bytes()` - So we currently pay clone + sort + protobuf encode per event, despite comments implying pre-encoding.
                                                                                                             

One option could be to store the pre-encoded `bytes::Bytes` in `NodeTelemetryHandle` at construction time (inside new()), so `node_custom_attrs()` returns a cheap `Bytes` clone instead of cloning and re-encoding the HashMap on every log event.

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
```diff
@@ -41,15 +67,72 @@ pub struct LogRecord {
...
+}
+
+/// Allow iterating entity keys via `Deref` to the inner SmallVec.
+impl Deref for LogContext {
+    type Target = SmallVec<[EntityKey; 1]>;
+
+    fn deref(&self) -> &Self::Target {
+        &self.entity_keys
+    }
+}
```
lalitb: Nit (non-blocking): `LogContext` now has two dimensions (`entity_keys` + `custom_attrs_bytes`), but `Deref<Target = SmallVec<...>>` still makes it look like a plain vector. Could we drop the `Deref` impl and make access explicit (`context.entity_keys` / `context.entity_keys.is_empty()`) everywhere? It keeps the API honest about what `LogContext` actually represents.

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
```diff
@@ -11,16 +11,42 @@ pub mod encoder;
...
+    let mut buf = ProtoBuffer::with_capacity(128);
+    let mut visitor = DirectFieldVisitor::new(&mut buf);
+    let mut sorted_attrs: Vec<_> = attrs.iter().collect();
+    sorted_attrs.sort_by_key(|(k, _)| k.as_str());
+    for (key, value) in sorted_attrs {
+        match value {
+            AttributeValue::String(s) => visitor.encode_string_attribute(key, s),
+            AttributeValue::Bool(b) => visitor.encode_bool_attribute(key, *b),
+            AttributeValue::I64(i) => visitor.encode_int_attribute(key, *i),
+            AttributeValue::F64(f) => visitor.encode_double_attribute(key, *f),
```
lalitb, line 42: Nit (non-blocking): If someone configures an array attribute in YAML, it gets silently dropped here. Could we either reject unsupported types during config validation, or at least log a warning? Silent data loss can be confusing to debug.

lquerel:
@andborja I'd like to better understand what you are trying to achieve with this PR. The example provided seems somewhat redundant with the existing concept of entity attributes, where each node should already have the pipeline ID defined. I may be misinterpreting your intent, and I apologize if that's the case.

Are these node-level attributes known at node startup, including their values?

Do these node-level attributes participate in the identification of the node?

Based on the answers to these questions, I think we can determine what the most appropriate approach would be.

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
utpilla: This is an anti-pattern in Rust. `Deref` is intended for smart pointer types (Box, Arc, etc.), not for exposing inner fields on types.

andborja:
> @andborja I'd like to better understand what you are trying to achieve with this PR. The example provided seems somewhat redundant with the existing concept of entity attributes, where each node should already have the pipeline ID defined. I may be misinterpreting your intent, and I apologize if that's the case.
> 
> Are these node-level attributes known at node startup, including their values?
> 
> Do these node-level attributes participate in the identification of the node?
> 
> Based on the answers to these questions, I think we can determine what the most appropriate approach would be.

My example is adding 2 arbitrary attributes at the node level. They are in addition to the ones automatically introduced by the engine (like pipeline ID).
They are known at node startup, including their values. Similar to the resource level attributes.
Those are telemetry attributes, so they are not used anywhere else (including node identification).

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -85,14 +85,18 @@ pub struct Controller<PData: 'static + Clone + Send + Sync + std::fmt::Debug> {
...
+    let custom_attrs = node_custom_attrs();
     if let Some(node) = node_entity_key() {
-        smallvec![node]
+        LogContext::with_custom_attrs(smallvec![node], custom_attrs)
     } else if let Some(pipeline) = pipeline_entity_key() {
-        smallvec![pipeline]
+        LogContext::with_custom_attrs(smallvec![pipeline], custom_attrs)
     } else {
-        smallvec![]
+        LogContext::new()
```
jmacd, line 99: What I am hoping to see would not result in a larger vector of entity keys at this point, but instead that the new attributes would be inserted in the node attributes. This function would be unchanged.

rust/otap-dataflow/crates/engine/src/entity_context.rs
```diff
@@ -95,6 +97,28 @@ pub fn node_entity_key() -> Option<EntityKey> {
...
+            .unwrap_or_default()
+    }) {
+        return attrs;
+    }
+    // Build-time fallback: check thread-local handle.
+    BUILD_NODE_TELEMETRY_HANDLE
+        .with(|cell| cell.borrow().as_ref().map(|h| h.custom_attrs()))
+        .unwrap_or_default()
+}
+
```
jmacd: Then, we would re-use existing thread local state, this would be unnecessary.

rust/otap-dataflow/crates/engine/src/entity_context.rs
jmacd: Thanks @lalitb. 

If we simply add new attributes to the existing Node entity, we get what you describe for free. The ConsoleAsync and ITR consumers will see the Node entity key and expand it into a pre-calculated OTLP bytes.,

andborja: Good idea. Updated.

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
andborja: Agree. Removed.

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
andborja: Added a validation at the yaml parsing.

rust/otap-dataflow/crates/telemetry/src/self_tracing.rs
```diff
@@ -116,6 +188,11 @@ impl LogRecord {
         let mut visitor = DirectFieldVisitor::new(&mut buf);
         event.record(&mut visitor);
 
+        // Append pre-encoded custom node attributes bytes, if any.
+        if !context.custom_attrs_bytes.is_empty() {
+            buf.extend_from_slice(&context.custom_attrs_bytes);
```
ThomsonTan, line 193: Custom node attributes are appended raw to body_attrs_bytes. If a tracing event field has the same name as a custom attribute, the LogRecord will contain duplicate attribute keys with potentially different values. OTLP spec does not define behavior for duplicate keys.

Issue #2008, opened 2026-02-09
validation framework updates


c1ly:
Extend the validation framework to:
- [x]  support multiple inputs and outputs (can configure and map more than one traffic::Generator and traffic::Capture)
- [x] support various validation checks right now we only support equivalence checks

     1. Filter check
     2. attribute check
     3. Batch check (check we are receiving appropriate batch sizes)

- [ ] run validation checks as a processor that is injected between receivers and processors
- [x] #2258
- [ ] fault injection
- [ ] custom receiver exporters for generator/capture
- [ ] extend framework to support multi pipelines
- [ ] add http support for otlp receiver/exporter in framework
- [ ] extend error to emit which validation tests failed

lquerel:
@c1ly what are the remaining sub-tasks here?

c1ly:
> [@c1ly](https://github.com/c1ly) what are the remaining sub-tasks here?

here are the list of features we want to support for the validation framework

- [ ] run validation checks as a processor that is injected between receivers and processors
- [ ]  support test containers
- [ ] fault injection
- [ ] custom receiver exporters for generator/capture
- [ ] extend framework to support multi pipelines

PR #2027, opened 2026-02-12
[otap-dataflow ] support additional validation methods in the validation framework

c1ly:
# Change Summary

Updated the validation exporter to perform different validation checks (equivalence, signal drop, attribute, batch size)

- users can configure the validation framework to perform a list of validation checks for their testing
- Defined ValidationKind enum to describe various validations to make

1. ValidationKind::SignalDrop -> checks for any change in signal count
2. ValidationKind::Attribute -> checks for existence of key or key values and can check for nonexistent keys
3. ValidationKind::Batch -> check for correct batch sizes
4. ValidationKind::Equivalence -> checks that message hasn't been transformed

related to #2008 

Added tests for each of the new validation check methods.
Added example validation framework tests for filter processor and attribute processor pipelines

No

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/testing/equiv/logs.rs
+++ b/rust/otap-dataflow/crates/pdata/src/testing/equiv/logs.rs
@@ -265,8 +265,8 @@ mod tests {
         };
 
         assert!(validate_logs_equivalent(
-            &[request1.clone()],
-            &[request2.clone()]
+            std::slice::from_ref(&request1),
+            std::slice::from_ref(&request2)
         ));
         assert_logs_equivalent(&[request1], &[request2]);
     }
--- a/rust/otap-dataflow/crates/pdata/src/testing/equiv/metrics.rs
+++ b/rust/otap-dataflow/crates/pdata/src/testing/equiv/metrics.rs
@@ -388,8 +388,8 @@ mod tests {
         };
 
         assert!(validate_metrics_equivalent(
-            &[request1.clone()],
-            &[request2.clone()]
+            std::slice::from_ref(&request1),
+            std::slice::from_ref(&request2)
         ));
         assert_metrics_equivalent(&[request1], &[request2]);
     }
--- a/rust/otap-dataflow/crates/pdata/src/testing/equiv/traces.rs
+++ b/rust/otap-dataflow/crates/pdata/src/testing/equiv/traces.rs
@@ -289,8 +289,8 @@ mod tests {
         };
 
         assert!(validate_traces_equivalent(
-            &[request1.clone()],
-            &[request2.clone()]
+            std::slice::from_ref(&request1),
+            std::slice::from_ref(&request2)
         ));
         assert_traces_equivalent(&[request1], &[request2]);
     }


rust/otap-dataflow/crates/validation/src/validation_types/batch.rs
```diff
@@ -0,0 +1,84 @@
...
+
+/// Ensure every message size is within `[min_items, max_items]` (if provided).
+/// Returns false if the slice is empty or any message violates the bounds.
+pub fn check_batch_size(
+    messages: &[OtlpProtoMessage],
+    min_items: Option<usize>,
+    max_items: Option<usize>,
+) -> bool {
+    if messages.is_empty() {
+        return false;
```
JakeDern: Is there a reason we would not return true in this case? My intuition would be that any no items implies all items satisfy the conditions

rust/otap-dataflow/crates/validation/src/validation_types/batch.rs
```diff
@@ -0,0 +1,84 @@
...
+
+//! Batch validation utilities.
+//!
+//! Validates that each message in a collection meets configured batch size bounds.
+
+use otap_df_pdata::proto::OtlpProtoMessage;
+
+/// Ensure every message size is within `[min_items, max_items]` (if provided).
+/// Returns false if the slice is empty or any message violates the bounds.
+pub fn check_batch_size(
```
JakeDern: Can we replace the implementation of some of these validation functions with only `all`/`any` if we change the empty handling? 

```rust
    !v.iter().any(|x| {
        let min = min.unwrap_or(0);
        let max = max.unwrap_or(usize::MAX);
        *x < min || *x > max
    })
```

JakeDern: Just something to be aware of - We eventually plan to support different [Sizers](https://github.com/open-telemetry/otel-arrow/blob/cbc03d838832e2dedba932c899b95cdf95b07594/rust/otap-dataflow/crates/otap/src/batch_processor.rs#L88-L96) than items. We're not there yet, but just mentioning in case it changes how you want to structure things in here!

rust/otap-dataflow/crates/validation/src/validation_types/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    },
+}
+
+impl ValidationKind {
+    /// Evaluate this validation against control and system-under-validation messages.
+    #[must_use]
+    pub fn evaluate(&self, control: &[OtlpProtoMessage], suv: &[OtlpProtoMessage]) -> bool {
+        match self {
+            ValidationKind::Equivalence => {
+                std::panic::catch_unwind(AssertUnwindSafe(|| assert_equivalent(control, suv)))
```
JakeDern: Maybe a nice improvement to the assert_equivelent code would be to add some non-panicing variants so that we don't have to catch and unwind here.

 I don't think this would be too hard to do, it looks like there is just a single assert at the end that we could extract out: https://github.com/open-telemetry/otel-arrow/blob/cbc03d838832e2dedba932c899b95cdf95b07594/rust/otap-dataflow/crates/pdata/src/testing/equiv/canonical.rs#L154C5-L163C1

rust/otap-dataflow/crates/validation/src/validation_types/attributes.rs
```diff
@@ -0,0 +1,395 @@
...
+        let mut out = Vec::new();
+
+        let include_resource = self.domains.contains(&AttributeDomain::Resource);
+        let include_scope = self.domains.contains(&AttributeDomain::Scope);
+        let include_signal = self.domains.contains(&AttributeDomain::Signal);
+
+        for resource_logs in &logs.resource_logs {
+            if include_resource {
+                if let Some(resource) = resource_logs.resource.as_ref() {
+                    out.push(resource.attributes.clone());
```
JakeDern: I don't think the validation code is particularly performance sensitive, but it could be nice to avoid collecting all the attributes by cloning the vectors and return references to them instead.

rust/otap-dataflow/crates/validation/src/validation_types/attributes.rs
```diff
@@ -0,0 +1,395 @@
...
+                && self.require.is_empty()
+                && self.forbid_keys.is_empty();
+        }
+
+        attr_lists
+            .iter()
+            .all(|attrs| self.validate_attr_list(attrs))
+    }
+
+    fn validate_attr_list(&self, attrs: &[ProtoKeyValue]) -> bool {
```
JakeDern: Another generic thing that might be nice to check is that we don't have duplicate attributes defined for any of the messages. It's probably an error somewhere if we do

rust/otap-dataflow/crates/validation/src/validation_exporter.rs
```diff
@@ -35,20 +34,26 @@ pub const VALIDATION_EXPORTER_URN: &str = "urn:otel:validation:exporter";
...
     control_input: NodeName,
-    /// When true, a failing equivalence check is considered expected.
-    #[serde(default)]
-    expect_failure: bool,
+    /// Validation rules to run. Defaults to a single equivalence check.
+    #[serde(default = "ValidationExporterConfig::default_validations")]
+    validations: Vec<ValidationKind>,
+}
+
+impl ValidationExporterConfig {
```
JakeDern, line 46: Do you think it would be a good idea to implement more component types e.g. Processor for this Validations component? I'm imagining, for example, that we want to uphold some invariants across an entire pipeline and that we could inject validation components after every receiver and processor runs so that if there is a problem detected it's isolated to a particular component. For more complicated pipelines with many stages this could be a really nice thing to have

rust/otap-dataflow/crates/validation/src/validation_types/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+use signal_dropped::check_signal_drop;
+
+/// Supported validation kinds executed by the validation exporter.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum ValidationKind {
+    /// Check semantic equivalence between control and suv outputs.
+    Equivalence,
+    /// Check that after contains fewer signals than before.
+    SignalDrop,
```
lquerel: May be we should enable more expressiveness here. I'm thinking about things like:
- signals drop by at least a %
- signals drop by at most a %
- signals drop by any amount (but count are equals)
- ...

rust/otap-dataflow/crates/validation/src/validation_types/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+    /// Check that each message meets a minimum and/or maximum batch size.
+    Batch {
```
lquerel: We probably need to extend batch validation criteria with temporality related conditions.

rust/otap-dataflow/crates/validation/src/validation_types/mod.rs
```diff
@@ -0,0 +1,144 @@
...
+use attributes::AttributeCheck;
+use batch::check_batch_size;
+use otap_df_pdata::proto::OtlpProtoMessage;
+use otap_df_pdata::testing::equiv::assert_equivalent;
+use signal_dropped::check_signal_drop;
+
+/// Supported validation kinds executed by the validation exporter.
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case")]
+pub enum ValidationKind {
```
lquerel: I think it's more than a validation kind. I suggest to rename this enum with something like:
- ValidationInstructions
- ValidationConditions
- ...

rust/otap-dataflow/crates/validation/src/validation_types/attributes.rs
```diff
@@ -0,0 +1,395 @@
...
+
+/// Declarative attribute validation.
+///
+/// - `require_keys`: every attribute list in the selected domains must contain
+///   these keys (value is ignored).
+/// - `require`: required key/value pairs that must appear in each attribute list.
+/// - `forbid_keys`: keys that must **not** appear in each attribute list.
+/// - `domains`: which domains to inspect; defaults to [`AttributeDomain::Signal`].
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Default)]
+pub struct AttributeCheck {
```
lquerel: I suggest to split this AttributeCheck into additional variants of the ValidationConditions/Instructions so users can create very explicit/clear scenarii.

rust/otap-dataflow/crates/validation/src/validation_exporter.rs
c1ly: this makes sense, the framework can be updated to so that if a test fails we then inject the component after each receiver and processor node to hone in on where the validation failed. Updated #2008 with this feature request

rust/otap-dataflow/crates/validation/src/validation_types/batch.rs
c1ly: we can define specific Validation enums variants for each batch sizer to keep things simple

rust/otap-dataflow/crates/validation/src/validation_types/batch.rs
c1ly: updated the params to take in a &OtlpProtoMessage instead of a slice of them so we aren't checking the same message again and again

Issue #2029, opened 2026-02-12
feat: Content router processor for resource-attribute-based routing


lalitb:
A `content_router` processor is needed to route telemetry signals (logs, metrics, traces) to named output ports based on a resource attribute value (e.g., `microsoft.resourceId`), enabling per-tenant isolation in multi-tenant pipeline deployments.                                                                                                                   
                                                                                                                                                                                        
  Conceptually equivalent to the [routing connector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/connector/routingconnector/README.md) in `opentelemetry-collector-contrib`, but adapted to the OTAP dataflow engine's local processor model — using zero-copy protobuf views for route resolution.

  **When to use this vs. the transform processor:**
  - **Pure OTLP pipelines** (e.g., Geneva/Azure Monitor with OTLP receivers) — prefer the content router. It operates directly on OTLP protobuf bytes via zero-copy views, avoiding the
  intermediate OTLP-to-OTAP conversion that the transform processor would require solely for routing.
  - **Mixed OTLP/OTAP pipelines** — prefer the transform processor, which handles Arrow data natively without conversion overhead.

Related discussion: #1834

drewrelmas:
Closed - component added

PR #2030, opened 2026-02-12
feat: add content_router processor

lalitb:
Description:

   Implements the content router processor described in #2029

   - Registered as `urn:otel:content_router:processor`
   - Zero-copy routing via `RawLogsData`/`RawMetricsData`/`RawTraceData` protobuf views
   - Native `OtapLogsView` for Arrow logs (metrics/traces Arrow views pending — falls back to OTLP conversion)
   - Destination-aware mixed-batch detection with single-pass fold

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
@@ -6,6 +6,7 @@ use azure_identity::{
     DeveloperToolsCredential, DeveloperToolsCredentialOptions, ManagedIdentityCredential,
     ManagedIdentityCredentialOptions, UserAssignedId,
 };
+use otap_df_telemetry::{otel_info, otel_warn};
 use std::sync::Arc;
 
 use super::Error;
@@ -20,14 +21,11 @@ const MAX_RETRY_JITTER_RATIO: f64 = 0.10;
 
 #[derive(Clone, Debug)]
 // TODO - Consolidate with crates/otap/src/{cloud_auth,object_store)/azure.rs
-#[allow(clippy::print_stdout)]
 pub struct Auth {
     credential: Arc<dyn TokenCredential>,
     scope: String,
 }
 
-// TODO: Remove print_stdout after logging is set up
-#[allow(clippy::print_stdout)]
 impl Auth {
     pub fn new(auth_config: &AuthConfig) -> Result<Self, Error> {
         let credential = Self::create_credential(auth_config)?;
@@ -63,17 +61,11 @@ impl Auth {
 
             match self.get_token_internal().await {
                 Ok(token) => {
-                    println!(
-                        "[AzureMonitorExporter] Obtained access token, expires on {}",
-                        token.expires_on
-                    );
+                    otel_info!("azure_monitor_exporter.auth.get_token_succeeded", expires_on = %token.expires_on);
                     return Ok(token);
                 }
                 Err(e) => {
-                    println!(
-                        "[AzureMonitorExporter] Failed to obtain access token (attempt {}): {e}",
-                        attempt
-                    );
+                    otel_warn!("azure_monitor_exporter.auth.get_token_failed", attempt = attempt, error = %e);
                 }
             }
 
@@ -93,7 +85,10 @@ impl Auth {
             let delay_secs = (capped_delay_secs + jitter).max(1.0);
             let delay = tokio::time::Duration::from_secs_f64(delay_secs);
 
-            println!("[AzureMonitorExporter] Retrying in {:.1}s...", delay_secs);
+            otel_warn!(
+                "azure_monitor_exporter.auth.retry_scheduled",
+                delay_secs = %delay_secs
+            );
             tokio::time::sleep(delay).await;
         }
     }
@@ -104,17 +99,23 @@ impl Auth {
                 let mut options = ManagedIdentityCredentialOptions::default();
 
                 if let Some(client_id) = &auth_config.client_id {
-                    println!("Using user-assigned managed identity with client_id: {client_id}");
+                    otel_info!("azure_monitor_exporter.auth.credential_type", method = "user_assigned_managed_identity", client_id = %client_id);
                     options.user_assigned_id = Some(UserAssignedId::ClientId(client_id.clone()));
                 } else {
-                    println!("Using system-assigned managed identity");
+                    otel_info!(
+                        "azure_monitor_exporter.auth.credential_type",
+                        method = "system_assigned_managed_identity"
+                    );
                 }
 
                 Ok(ManagedIdentityCredential::new(Some(options))
                     .map_err(|e| Error::create_credential(AuthMethod::ManagedIdentity, e))?)
             }
             AuthMethod::Development => {
-                println!("Using developer tools credential (Azure CLI / Azure Developer CLI)");
+                otel_info!(
+                    "azure_monitor_exporter.auth.credential_type",
+                    method = "developer_tools"
+                );
                 Ok(
                     DeveloperToolsCredential::new(Some(DeveloperToolsCredentialOptions::default()))
                         .map_err(|e| Error::create_credential(AuthMethod::Development, e))?,
--- a/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
@@ -3,6 +3,7 @@
 
 use bytes::Bytes;
 
+use otap_df_telemetry::otel_warn;
 use rand::{RngExt, SeedableRng, rngs::SmallRng};
 use reqwest::{
     Client,
@@ -23,8 +24,6 @@ const MAX_IDLE_CONNECTIONS_PER_HOST: usize = 2;
 /// Handles authentication, compression, and HTTP communication with the
 /// Azure Monitor Logs Ingestion API.
 #[derive(Clone)]
-// TODO: Remove print_stdout after logging is set up
-#[allow(clippy::print_stdout)]
 pub struct LogsIngestionClient {
     http_client: Client,
     endpoint: String,
@@ -91,8 +90,6 @@ impl LogsIngestionClientPool {
     }
 }
 
-// TODO: Remove print_stdout after logging is set up
-#[allow(clippy::print_stdout)]
 impl LogsIngestionClient {
     /// Creates a new Azure Monitor logs ingestion client instance from provided components.
     ///
@@ -144,8 +141,6 @@ impl LogsIngestionClient {
         self.auth_header = header;
     }
 
-    // TODO: Remove print_stdout after logging is set up
-    #[allow(clippy::print_stdout)]
     /// Export compressed data to Log Analytics ingestion API with automatic retry.
     ///
     /// Retries on:
@@ -199,11 +194,7 @@ impl LogsIngestionClient {
                         base_delay.mul_f64(jitter_factor)
                     };
 
-                    println!(
-                        "[AzureMonitorExporter] Retry after {}ms: {:?}",
-                        delay.as_millis(),
-                        e
-                    );
+                    otel_warn!("azure_monitor_exporter.export.retry_delay", delay_ms = delay.as_millis() as u64, error = ?e);
 
                     tokio::time::sleep(delay).await;
                 }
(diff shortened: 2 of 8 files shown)


rust/otap-dataflow/crates/otap/src/content_router.rs
```diff
@@ -0,0 +1,1518 @@
...
+//!     routing_key: "service.namespace"
+//!     case_sensitive: false
+//!     routes:
+//!       "frontend": "frontend_pipeline"
+//!       "backend": "backend_pipeline"
+//!     default_output: "fallback"
+//! ```
+//!
+//! Each route value corresponds to a named output port that must be wired
+//! in the pipeline configuration.
```
albertlockett, line 28: nit: Should we add a sentence or two to the module docs here clarifying the behaviour for unmatched or mix-matched batches?

albertlockett (approved):
LGTM, thanks @lalitb

rust/otap-dataflow/crates/otap/src/content_router.rs
```diff
@@ -0,0 +1,1518 @@
...
+                // Defensive: signal_type/payload mismatch cannot occur for OtlpBytes
+                // since signal_type() is derived from the OtlpProtoBytes variant itself.
+                _ => RouteResolution::ConversionError,
+            },
+            OtapPayload::OtapArrowRecords(arrow_records) => {
+                match signal_type {
+                    // Use native OTAP Arrow view for logs (avoids clone + OTLP round-trip)
+                    SignalType::Logs => self.resolve_arrow_logs_route(arrow_records),
+                    // Metrics/Traces Arrow views not yet available — convert to OTLP.
+                    // TODO: Use OtapMetricsView/OtapTracesView when available.
```
jmacd, line 484: 👍

rust/otap-dataflow/crates/otap/src/content_router.rs
```diff
@@ -0,0 +1,1518 @@
...
+    /// Output port for messages that don't match any route.
+    /// If not set, unmatched messages are permanently NACKed.
+    #[serde(default)]
+    pub default_output: Option<String>,
+    /// Whether to perform case-sensitive matching on attribute values.
+    #[serde(default = "default_case_sensitive")]
+    pub case_sensitive: bool,
+}
+
+fn default_case_sensitive() -> bool {
```
utpilla:
```suggestion
const fn default_case_sensitive() -> bool {
```

rust/otap-dataflow/crates/otap/src/content_router.rs
```diff
@@ -0,0 +1,1518 @@
...
+use otap_df_telemetry::metrics::MetricSet;
+use otap_df_telemetry::otel_warn;
+use otap_df_telemetry_macros::metric_set;
+use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+/// URN for the ContentRouter processor
+pub const CONTENT_ROUTER_URN: &str = "urn:otel:content_router:processor";
```
utpilla, line 80: The name `content_router` feels too generic. It doesn't communicate what about the content drives routing. Something like `resource_attribute_router` would be more descriptive for what this processor currently does.

However, before we settle on a name, I think it's worth discussing the broader routing story. Ideally, for multi-tenant routing, we'd want to extract the tenant identifier from HTTP headers or gRPC metadata. That way we can route without parsing the payload at all, which would be significantly cheaper.

That raises a design question: Should we have one routing component that can route based on either headers or resource attributes (configured via the routing key source), or separate processors for each (e.g., `header_router`, `resource_attribute_router`)? The naming of this component woul depend on which direction we take.

lalitb: Great questions - let me share my thinking on both:

  **On the name:** I'd prefer to keep **_content_router._** I agree it sounds generic if you only look at what it does today (resource attributes), but that's intentional. We can extend it in the future to support other content-level keys - scope attributes, or even context metadata stamped by the receiver. The name leaves room for that without a rename. `resource_attribute_router` would lock us in.

**On header-based routing**: Architecturally, a separate `header_router` processor can't exist in our design -`headers` and `gRPC` metadata are consumed at the transport layer by the receiver. By the time a processor sees the message, it only has `OtapPdata`. If we ever needed header values for routing, the receiver would have to stamp them into the `Context` struct that travels with each message, and then `content_router` could read them. So it would naturally be a config option within this same component, not a separate processor.

rust/otap-dataflow/crates/otap/src/content_router.rs
lalitb: Good call - added a section to the module docs covering both cases. 

 On **_mixed batches_** specifically: we intentionally NACK the entire batch rather than splitting it per-resource (which is what the Go collector's `routingconnector` does). Splitting requires deserializing, partitioning the resource arrays, and re-serializing separate payloads per destination - which would destroy the zero-copy advantage we get from routing on `protobuf` views. Batch-level routing keeps it to a single-pass fold over resource attributes with O(1) forwarding of the untouched payload.

 In practice, mixed batches should be rare - SDKs typically produce batches from a single resource, and upstream collectors should already be grouping by.tenant. The permanent NACK surfaces it as a configuration issue to fix upstream rather than silently absorbing the splitting cost. If splitting were ever needed, it would belong in a separate upstream processor, not in the router.

rust/otap-dataflow/crates/otap/src/content_router.rs
utpilla: > If we ever needed header values for routing, the receiver would have to stamp them into the Context struct that travels with each message, and then content_router could read them. So it would naturally be a config option within this same component, not a separate processor.

Yes, I think we should consider doing that. it would be more efficient to look up the Context instead of iterating over resource attributes on the hot path. So, it sounds like we're aligned on having this same processor extend its capabilities over time. I prefer that too.

In that case, I'd suggest we drop the word "content" and simply call it the `router` processor. Just like how we use `retry` and `batch` processors and not `content_retry` or `content_batch` processors.

lalitb: >  In that case, I'd suggest we drop the word "content" and simply call it the router processor. Just like how we use retry and batch processors and not content_retry or content_batch processors.

We already have `signal_type_router` in the codebase - calling this one `router` would be ambiguous (`router` based on what?). The prefix here disambiguates: `signal_type_router` routes by signal type, `content_router` routes by content values.

 Also, regarding naming - content-based routing - [CBR](https://www.enterpriseintegrationpatterns.com/ContentBasedRouter.html) is a well-established messaging pattern - a router that inspects message content to determine the destination. Naming for `retry` and `batch` processors make sense - but here we have two different types of routers.

jmacd: I think this is pretty good reasoning, even if the link to "CBR" looks to be from the 1990s😀. 

@lquerel please take a look.

On this thread, I think we should rename `signal_type_router` to `signal_router`.

lquerel: The URN of the signal type router is already `urn:otel:type_router:processor`, so I would be in favor of using the URN `urn:otel:content_router:processor` for this new processor. We should also rename the file `signal_type_router.rs` to `type_router.rs` for consistency.

I'm still reviewing this PR.

rust/otap-dataflow/crates/otap/src/content_router.rs
```diff
@@ -0,0 +1,1529 @@
...
+    pub case_sensitive: bool,
+}
+
+const fn default_case_sensitive() -> bool {
+    true
+}
+
+impl ContentRouterConfig {
+    /// Validates the configuration.
+    fn validate(&self) -> Result<(), ConfigError> {
```
lquerel: This method should validate that the route destination is connected to a declared node output. `outputs` are accessible from the node config object IIRC.

rust/otap-dataflow/crates/otap/src/content_router.rs
```diff
@@ -0,0 +1,1529 @@
...
+                                format!(
+                                    "no matching route for routing key '{}' and no default output configured",
+                                    self.routing_key
+                                )
+                            };
+                            otel_warn!(
+                                "content_router.nack",
+                                routing_key = self.routing_key.as_str(),
+                                message = reason.as_str()
+                            );
```
lquerel: Emitting a warn event for every unroutable or conversion error is not a good idea, since an external user could deliberately send unroutable messages and degrade the performance of the entire pipeline. I'll start with a metric instead, and potentially (in the future) introduce some form of sampling for this event later on.

rust/otap-dataflow/crates/otap/src/content_router.rs
```diff
@@ -0,0 +1,1529 @@
...
+///     case_sensitive: false
+///     routes:
+///       "frontend": "frontend_pipeline"
+///       "backend": "backend_pipeline"
+///     default_output: "fallback"
+/// ```
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ContentRouterConfig {
+    /// The resource attribute key used for routing decisions.
+    pub routing_key: String,
```
lquerel: Since we may extend the scope of this router, I think we should find a way to make this routing key more explicit. It is currently a routing key that targets resource attributes. The question we should ask is how we would evolve this configuration to support routing based on scope attributes or signal attributes.

rust/otap-dataflow/crates/otap/src/content_router.rs
```diff
@@ -0,0 +1,1529 @@
...
+    ConversionError,
+}
+
+/// The ContentRouter processor routes messages to output ports based on
+/// a resource attribute value.
+pub struct ContentRouter {
+    /// The attribute key to inspect for routing.
+    routing_key: String,
+    /// Normalized routes: attribute value -> output port name.
+    routes: HashMap<String, String>,
```
lquerel:
```suggestion
    routes: HashMap<String, PortName>,
```

rust/otap-dataflow/crates/otap/src/content_router.rs
lquerel: Do we have a GH issue for that?

rust/otap-dataflow/crates/otap/src/content_router.rs
jmacd: (or metric name)!

rust/otap-dataflow/crates/otap/src/content_router.rs
lalitb: We didn't had, created now - #2053

rust/otap-dataflow/crates/otap/src/content_router.rs
lalitb: Done - `validate` now takes `&[PortName]` (the declared node outputs) and checks that every route destination and `default_output` refers to a declared output port.

rust/otap-dataflow/crates/otap/src/content_router.rs
lalitb: Agreed — removed all three `otel_warn!` calls (unroutable, mixed batch, conversion error). The existing metrics (`signals_nacked`, `signals_no_routing_key`, `signals_conversion_error`) are the primary           
  observability mechanism for these cases.

rust/otap-dataflow/crates/otap/src/content_router.rs
lalitb: Good point - introduced a `RoutingKeyExpr` enum to make the source explicit and extensible:
                
  ```rust                                                                                                                                                                                                            
  pub enum RoutingKeyExpr {
      ResourceAttribute(String),                                                                                                                                                                                     
      // Future: ScopeAttribute(String), MetricName, ...
  }
```
The config format is now routing_key: { resource_attribute: "service.namespace" }. Adding support for scope attributes, signal attributes, or metric names (👍 @jmacd) only requires a new variant and a corresponding match arm in resolve_route.

rust/otap-dataflow/crates/otap/src/content_router.rs
lalitb: done.

lalitb:
> I approve the PR, but I think there are a few minor improvements to make. I trust you to address them before merging.

Thanks, comments are relevant. Have addressed them.

rust/otap-dataflow/crates/otap/src/content_router.rs
lalitb: Agreed on the URN - `urn:otel:content_router:processor` is already in place. On the renaming of `signal_type_router.rs` -> `type_router.rs`: happy to do this, but would prefer to keep it out of this PR to keep the diff focused. Can follow up in a separate cleanup PR.

lalitb:
@lquerel  - do you think this can be merged now ?

PR #2032, opened 2026-02-12
[query-engine] RecordSet engine diagnostic level adjustments

CodeBlanch:
# Changes

* Lowers some spammy diagnostics from "Warn" to "Info" in RecordSet engine

# Details

@drewrelmas has been doing some integration testing and noticed these "warnings" showing up for cases that didn't really need any attention:

* When using `coalesce(thing1, thing2)` we don't really need to "warn" if "thing1" couldn't be found.
* When using `project-away thing1, thing2` we don't really need to "warn" if "thing1" wasn't found (just a no-op in that case).


jmacd:
LGTM

PR #2038, opened 2026-02-13
perf: improve take performance when deleting rows in transform attributes

### LINKED ISSUE #2037: Slow performance taking non-deleted rows when transforming attributes when converted from OTLP
In attribute transformation code that is used by the attribute processor/columnar query engine, when it does deletes, it figures out a list of ranges to keep (the ranges of non-deleted rows), then takes the rows for these ranges.

This is pretty low overhead when we have received an OTAP batch from the OTAP receiver, because the batch is already sorted by key, so usually we're only splitting/taking a handful of ranges

However, when we convert attributes from OTLP, the batch ends up sorted by parent ID, which means the ranges of non-deleted rows are all over the place. This means it behooves us not to do something slow when we take these ranges.

Currently however, we're doing this:
https://github.com/open-telemetry/otel-arrow/blob/cbc03d838832e2dedba932c899b95cdf95b07594/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L2174-L2187

Which creates a new `ArrayRef` for every range (with all the associated `Arc`s, etc) and then concats them. As expected, when doing this kind of transformation, this is where we spend the vast majority of time:

Image

albertlockett:
# Change Summary

When we delete attributes, we compute the ranges of rows in the attributes record batch that did not have deletions and then take them. Before, we were using `slice` and `concat` for each range, which has a bit of overhead.

When there are many ranges, this overhead adds up and makes the performance quite bad. This can happen when we convert OTLP to OTAP, because the attributes end up sorted by parent_id so the ranges of deleted/kept attribtue keys are all over the place.

This change uses the MutableArrayData to take the ranges instead, inspired by what the arrow [`filter`](https://docs.rs/arrow/latest/arrow/compute/kernels/filter/fn.filter.html) kernel does internally:
https://github.com/apache/arrow-rs/blob/d8946ca0775ab7fe0eef2fdea4b8bb3d55ec6664/arrow-select/src/filter.rs#L475-L498

The PR also adds a benchmark for transforming attributes that are in the problematic sort order, to measure the performance imrovement.

* Closes #2037 

Existing unit tests cover this function

No


albertlockett:
Before:
```
transform_attributes_transport_optimized/num_rows=128,dict_keys=true,delete,no_encode
                        time:   [10.736 µs 10.751 µs 10.767 µs]

transform_attributes_transport_optimized/num_rows=1536,dict_keys=true,delete,no_encode
                        time:   [126.55 µs 126.74 µs 126.94 µs]

transform_attributes_transport_optimized/num_rows=8092,dict_keys=true,delete,no_encode
                        time:   [670.02 µs 670.92 µs 671.81 µs]

  ```

After:

```
transform_attributes_transport_optimized/num_rows=128,dict_keys=true,delete,no_encode
                        time:   [1.9626 µs 1.9671 µs 1.9723 µs]
                        change: [−81.723% −81.661% −81.601%] (p = 0.00 < 0.05)
                        Performance has improved.


transform_attributes_transport_optimized/num_rows=1536,dict_keys=true,delete,no_encode
                        time:   [8.7713 µs 8.7856 µs 8.8015 µs]
                        change: [−93.072% −93.050% −93.026%] (p = 0.00 < 0.05)
                        Performance has improved.


transform_attributes_transport_optimized/num_rows=8092,dict_keys=true,delete,no_encode
                        time:   [40.037 µs 40.174 µs 40.340 µs]
                        change: [−93.839% −93.793% −93.753%] (p = 0.00 < 0.05)
                        Performance has improved.
```

JakeDern:
Wow, nice! I will be stealing this for the reindexing code. Tracking it on #1926

Issue #1957, opened 2026-02-04
Documentation: Create a formal OTAP spec


JakeDern:
## Background

Information on OpenTelemetry Arrow Protocol (OTAP) is currently scattered. There is a good amount of information between the partial [data model](https://github.com/open-telemetry/otel-arrow/blob/main/docs/data_model.md), [protbuf definition](https://github.com/open-telemetry/otel-arrow/tree/main/proto/opentelemetry/proto/experimental/arrow/v1),  [informal introduction](https://github.com/open-telemetry/otel-arrow/blob/main/docs/otap_basics.md), and mostly complete [Go](https://github.com/open-telemetry/otel-arrow/tree/main/go/pkg/otel)/[Rust](https://github.com/open-telemetry/otel-arrow/tree/main/rust/otap-dataflow/crates/pdata) implementations, but no agreed upon formal spec written down. This makes it difficult to gain a mental model of the protocol if you are new to the project.

But beyond that ambiguities also cause trouble during implementation and review. Without an agreed upon written spec, I check what the Go implementation does, ask someone with more experience, or program defensively against cases I'm not sure can happen. But the Go implementation is just _one_ implementation in a range of acceptable behaviors, experts can't remember everything or be guaranteed to agree, and code often becomes more complicated or less performant the less constraints there are.

## Proposed process

This is the parent issue for getting the spec to an accepted state. After merging an initial draft, issues can be opened and closed with a PR like any other improvement. Once we're satisfied with the state of the spec, this parent issue can be closed with the PR declaring accepted.

## Sub Issues

Opening issues is encouraged even if the goal is not to push for a particular change. Issues are a great way to publicly document tradeoffs that are not obvious and can serve as decision records.

- [ ] #2134
- [ ] #2135
- [ ] #2136
- [x] #2149
- [ ] #2168
- [ ] #2369
- [ ] #2459

## TODOs

- [ ] Error handling documentation is lacking i.e. what do the various error handling codes mean? what's retryable and what's not?
- [ ] Are there any interoperability requirements worth discussing. [See here](https://github.com/open-telemetry/otel-arrow/pull/2040#discussion_r2854448631)
- [ ] Leveraging the format in some way to generate code and/or documentation
- [ ] How does OTAP become "official" e.g. [Pull requests · open-telemetry/opentelemetry-specification](https://github.com/open-telemetry/opentelemetry-specification/pull/4791)

PR #2040, opened 2026-02-14
docs: Initial OTAP Spec Draft

JakeDern:
# Change Summary

This is a working draft of an OTAP Spec. The goal is now to get a solid foundation merged so that we can continue to iterate on spec changes as sub issues listed in #1957.

Merging this PR will open up contributions to anyone rather than keeping everything on just on private branch and make spec changes explicit and subject to the same review process as any other change going forward.

* Part of #1957


OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+See [Appendix F: Glossary](#appendix-f-glossary) for terminology and [Appendix G: References](#appendix-g-references) for reference material.
+
+---
+
+## 2. Data Model
+
+### 2.1 Normalized Representation
+
+OTAP represents telemetry Signals as a set of normalized tables connected by foreign key relationships; 
+effectively a database. Each Signal type has a different number and set of tables reflecting the data
```
albertlockett: nit: should we drop this phrase "effectively a database"? I feel it may sew confusion for those who philosophically disagree about what technically constitutes a  "database"

> effectively a database.

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+
+Each table within a Signal has a designated Payload Type that identifies it. The foreign key
+relationships between these tables form a Rooted Directed Acyclic Graph (DAG) with the 
+Root Payload Type being the root of that graph. 
+
+For example, the Logs signal type consists of four Payload Types: LOGS, LOG_ATTRS, RESOURCE_ATTRS, 
+and SCOPE_ATTRS. The LOGS table is the Root Payload Type for Logs and fills a similar role as an OTLP Log. 
+Each Log has a unique `id` which identifies it, and links it to the LOG_ATTRS table which defines the log's
+attributes. LOGS similarly contains `resource` and `scope` fields, each having an `id`  which links them 
+to the RESOURCE_ATTRS and SCOPE_ATTRS tables.
```
albertlockett: > Each Log has a unique `id` which identifies it

It might be helpful to call out here that the `id` isn't globally unique - it's only unique within one "batch".

If we add this, maybe we should also have previously introduced the concept of what means a "batch" by adding a section or paragraph before here specifying something like how OTAP data is a stream of batches/frames, and within each batch there are multiple  normalized tables.

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+| LOG_ATTRS | Log-level attributes | LOGS |
+| RESOURCE_ATTRS | Resource attributes | LOGS |
+| SCOPE_ATTRS | Scope attributes | LOGS |
+
+#### 2.1.2 Metrics Signal Tables
+
+| Payload Type | Description | Parent Payload Type |
+|---|---|---|
+| UNIVARIATE_METRICS | Core metric metadata (Root) | — |
+| MULTIVARIATE_METRICS | Core metric metadata (Root) | — |
```
albertlockett: I feel the description should not use the word "metadata". 

The reason I feel this way is because what we call METRIC_ATTRS in OTAP data model is actually called metadata in the OTLP data model:
https://github.com/open-telemetry/opentelemetry-proto/blob/d3fb76d70deb0874692bd0ebe03148580d85f3bb/opentelemetry/proto/metrics/v1/metrics.proto#L212-L220

For these rows, maybe say something like "Definition of metric (Root)". FWIW, the OTLP data model describes the equivalent structure as "Defines a Metric which has one or more timeseries."
https://github.com/open-telemetry/opentelemetry-proto/blob/d3fb76d70deb0874692bd0ebe03148580d85f3bb/opentelemetry/proto/metrics/v1/metrics.proto#L104

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+| SUMMARY_DP_ATTRS | Attributes for summary data points | SUMMARY_DATA_POINTS |
+| HISTOGRAM_DP_ATTRS | Attributes for histogram data points | HISTOGRAM_DATA_POINTS |
+| EXP_HISTOGRAM_DP_ATTRS | Attributes for exponential histogram data points | EXP_HISTOGRAM_DATA_POINTS |
+| NUMBER_DP_EXEMPLARS | Exemplars for number data points | NUMBER_DATA_POINTS |
+| HISTOGRAM_DP_EXEMPLARS | Exemplars for histogram data points | HISTOGRAM_DATA_POINTS |
+| EXP_HISTOGRAM_DP_EXEMPLARS | Exemplars for exponential histogram data points | EXP_HISTOGRAM_DATA_POINTS |
+| NUMBER_DP_EXEMPLAR_ATTRS | Exemplar attributes for number data points | NUMBER_DP_EXEMPLARS |
+| HISTOGRAM_DP_EXEMPLAR_ATTRS | Exemplar attributes for histogram data points | HISTOGRAM_DP_EXEMPLARS |
+| EXP_HISTOGRAM_DP_EXEMPLAR_ATTRS | Exemplar attributes for exponential histogram data points | EXP_HISTOGRAM_DP_EXEMPLARS |
+| METRIC_ATTRS | Metric-level attributes | METRICS |
```
albertlockett: nit: maybe add "metadata" in brackets? This is what the equivalent field is called in OTLP
https://github.com/open-telemetry/opentelemetry-proto/blob/d3fb76d70deb0874692bd0ebe03148580d85f3bb/opentelemetry/proto/metrics/v1/metrics.proto#L212-L220

```suggestion
| METRIC_ATTRS | Metric-level attributes (metadata) | METRICS |
```

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+
+### 3.3 Arrow IPC Layer
+
+The Arrow IPC (Interprocess Communication) layer is the innermost layer containing the actual telemetry data 
+in Apache Arrow's columnar format. Arrow IPC defines how schemas and data are serialized into byte streams using 
+a standardized format that can be read by any Arrow-compatible library. This layer enables:
+
+- **Dictionary/Delta Dictionary encoding**: Efficient representation of repeated string values
+- **Zero-copy deserialization**: Data can be read directly from wire format without copying
+- **Columnar layout**: Data organized by column rather than by row, enabling better compression and SIMD processing
```
albertlockett: nit: Do you feel this third point may be redundant? The structure/benefits of columnar layout are something inherent to arrow, not necessarily arrow IPC

> - **Columnar layout**: Data organized by column rather than by row, enabling better compression and SIMD processing

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+| str | Utf8 | Dict(u8), Dict(u16) | Yes | No | — | — | String value (when type=1) |
+| int | Int64 | — | Yes | No | — | — | Integer value (when type=3) |
+| double | Float64 | — | Yes | No | — | — | Double value (when type=4) |
+| bool | Boolean | — | Yes | No | — | — | Boolean value (when type=2) |
+| bytes | Binary | — | Yes | No | — | — | Bytes value (when type=5) |
+| ser | Binary | — | Yes | No | — | — | CBOR-encoded Array or Map (when type=6 or 7) |
+
+### 4.2 Logs Payloads
+
+#### LOGS
```
albertlockett: Some comments for logs table:
- `body` is a struct, so we should probably add `.` instead of `_` in the field names. E.g. change `body_type` to `body.type`
- Some additional columns in the table below do support dictionary encoding:
`schema_url`, `trace_id`, `span_id`, `severity_text`, `severity_number`, `body.str`, `body.int`, `body.bytes`, `body.ser`, `resource.schema_url`, `scope.name`, `scope.version`
- `event_name` column looks to be missing

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+| body_str | Utf8 | — | No | Yes | — | — | String body (may be empty) |
+| body_int | Int64 | — | Yes | No | — | — | Integer body (when body_type=3) |
+| body_double | Float64 | — | Yes | No | — | — | Double body (when body_type=4) |
+| body_bool | Boolean | — | Yes | No | — | — | Boolean body (when body_type=2) |
+| body_bytes | Binary | — | Yes | No | — | — | Bytes body (when body_type=5) |
+| body_ser | Binary | — | Yes | No | — | — | CBOR-encoded complex body (when body_type=6 or 7) |
+| dropped_attributes_count | UInt32 | — | Yes | No | — | — | Number of dropped log attributes |
+| flags | UInt32 | — | Yes | No | — | — | Trace flags |
+
+#### LOG_ATTRS
```
albertlockett: LOG_ATTRS is the same as Scope, Resource, Span & Metrics attributes. Would it save some repetition to define this structure once?

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+
+| Name | Type | Alt Representations | Nullable | Required | Id Encoding | Metadata | Description |
+|------|------|---------------------|----------|----------|-------------|----------|-------------|
+| id | UInt32 | — | Yes | Yes | [DELTA](#652-delta-encoding) | encoding | Event identifier (primary key) |
+| parent_id | UInt16 | — | No | Yes | [COLUMNAR QUASI-DELTA](#653-quasi-delta-encoding) (name) | encoding | Foreign key to [SPANS](#spans).id |
+| time_unix_nano | Timestamp(Nanosecond) | — | Yes | No | — | — | Event timestamp in Unix nanoseconds |
+| name | Utf8 | — | No | Yes | — | — | Event name |
+| dropped_attributes_count | UInt32 | — | Yes | No | — | — | Number of dropped event attributes |
+
+#### SPAN_EVENT_ATTRS
```
albertlockett: `SPAN_EVENT_ATTRS` is the same structure as DP_ATTRs, _DP_EXEMPLAR_ATTRS, and SPAN_LINK_ATTRS. Would it cut down on some repetition to define the structure of the attributes table one time?

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+---
+
+## 4. Payload Specifications
+
+This section defines the complete Arrow schema for all OTAP payload types, organized by signal category.
+
+**Table Column Descriptions:**
+- **Name**: Field name in the Arrow schema
+- **Type**: Base Arrow type for this field
+- **Alt Representations**: Alternative encodings allowed for this field (e.g., `Dict(u8)` for Dictionary(UInt8, Type), `List(T)` for list types). See also: [Dictionary Encoding](https://arrow.apache.org/docs/format/Columnar.html#dictionary-encoded-layout)
```
albertlockett: I wonder if there's a way to restructure how we define these type columns to make it clearer when each type is used. I'm worried the "Alt Representation" might be interpreted as types that are accepted albeit non-standard, or are chosen in some laissez-faire fashion when in reality there is some logic to which type is used.

For the case of dictionaries, the rules are basically like - OTAP will try to use the smallest dictionary key it can, and eventually overflow to the native (non dict-encoded) array.

Maybe instead of Alt Representation, we have a column like "Dict key min size"?

Also, do we actually have columns that are either Native/List? For example, I see this `List` being used for Summary DP Quantile values, but I thought that field was actually a List of structs, e.g.:
```rs
 DataType::List(Arc::new(Field::new(
                "item",
                DataType::Struct(Fields::from(vec![
                    Field::new(consts::SUMMARY_QUANTILE, DataType::Float64, false),
                    Field::new(consts::SUMMARY_VALUE, DataType::Float64, false),
                ])),
                true,
            ))),
```

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+| Name | Type | Alt Representations | Nullable | Required | Id Encoding | Metadata | Description |
+|------|------|---------------------|----------|----------|-------------|----------|-------------|
+| id | UInt32 | — | Yes | Yes | [DELTA](#652-delta-encoding) | encoding | Data point identifier (primary key) |
+| parent_id | UInt16 | — | No | Yes | [DELTA](#652-delta-encoding) | encoding | Foreign key to [UNIVARIATE_METRICS / MULTIVARIATE_METRICS](#univariate_metrics--multivariate_metrics).id |
+| start_time_unix_nano | Timestamp(Nanosecond) | — | Yes | No | — | — | Start time in Unix nanoseconds |
+| time_unix_nano | Timestamp(Nanosecond) | — | Yes | No | — | — | Timestamp in Unix nanoseconds |
+| count | UInt64 | — | Yes | No | — | — | Count of observations |
+| sum | Float64 | — | Yes | No | — | — | Sum of observations |
+| quantile | Float64 | List(Float64) | Yes | No | — | — | Quantile values |
+| value | Float64 | List(Float64) | Yes | No | — | — | Quantile observation values |
```
albertlockett: Not sure this is correct - Pretty sure this is a list of structs (see previous comment)

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+- **Name**: Field name in the Arrow schema
+- **Type**: Base Arrow type for this field
+- **Alt Representations**: Alternative encodings allowed for this field (e.g., `Dict(u8)` for Dictionary(UInt8, Type), `List(T)` for list types). See also: [Dictionary Encoding](https://arrow.apache.org/docs/format/Columnar.html#dictionary-encoded-layout)
+- **Nullable**: Whether the field can contain null values
+- **Required**: Whether this field must be present in every record
+- **Id Encoding**: The encoding method used for id columns (see [Section 5.5](#65-transport-optimized-encodings))
+- **Metadata**: Arrow field-level metadata keys that MAY be present (see [Section 5.5.4](#654-field-metadata))
+- **Description**: Human-readable description of the field's purpose
+
+NOTE: Column names that contain a `.` character indicate the presence of a struct typed field. For
```
albertlockett: Do you think we should add a row to each table for the struct column itself, specifying that the type is a struct? This would also give us a place to:
- describe the struct (in description field)
- specify whether the struct itself is nullable/required

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+
+| type value | Active field |
+|------------|--------------|
+| 1 | str |
+| 2 | bool |
+| 3 | int |
+| 4 | double |
+| 5 | bytes |
+| 6 | ser (Array encoded as CBOR) |
+| 7 | ser (Map encoded as CBOR) |
```
albertlockett: Seems like the numbers in the table might be off:
https://github.com/open-telemetry/otel-arrow/blob/a6e0e74750abf6f949784dfada1df293d2c7250b/rust/otap-dataflow/crates/pdata/src/otlp/attributes.rs#L33-L42
5 = Map
6 = Slice
7 = Bytes
There's also 0 which is represents the 'Empty' value type

It might be helpful to flesh out a few details here:

First, we can clarify what is meant by "Active Field". E.g. this is the column in a RecordBatch/Struct (in the case of log body) to read the value from for a given row depending on the value of the type field.

Also, we can specify how null values should/may be encoded. For example, we'd prefer if an attribute value is null to set type = 0 (empty), but if the type is non-zero and the value column is null in some position (or if value column is missing), the attribute value will be interpreted as null. Annoyingly for Logs (which uses this same scheme) there is yet another way to determine that the body is null which is that the body struct itself could be null for some row.

docs/otap-spec.md
```diff
@@ -0,0 +1,1268 @@
...
+**Algorithm for Columnar Quasi-Delta**:
+Similar, but matching based on specified column values (e.g., span event `name` field, exemplar `int_value`/`double_value`).
+
+**Applicability**: `parent_id` columns in attribute and related tables
+
+**Typical use**:
+- Attribute table `parent_id` columns
+- Span event/link `parent_id` columns
+- Exemplar `parent_id` columns
+
```
albertlockett, line 806: In the future, we can probably provide some extra details about when the parent_id is really delta encoded.

There are additional rules, such as:
- different fields are used for different payload types when determining the contiguous runs. for example, we call out that it's `int_value`/`double_value` for exmplars, but for attributes it's `type`/`key`/ value (where value is any of `str`, `int` etc. depending on type)
- we don't delta encode when the attributes are null
- we don't delta encode for some attr types (Map & Slice)

https://github.com/open-telemetry/otel-arrow/blob/a6e0e74750abf6f949784dfada1df293d2c7250b/rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs#L10-L64

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+characteristics:
+
+1. There is a many-to-many relationship relationship between RESOURCE_ATTRS/SCOPE_ATTRS tables and their
+parent payload types
+2. The corresponding column in the LOGS/METRICS/SPANS tables for RESOURCE_ATTRS.parent_id and SCOPE_ATTRS.parent_id
+are `resource.id` and `scope.id` respectively rather than just `id`.
+3. Unlike other identifiers, `resource.id` and `scope.id` have no single table that "owns" them and defines the valid
+set of Ids.
+
+### 5.5 Transport Optimized Encodings
```
albertlockett: I wonder if we should also add a section here specifying that we also sort the record batches prior to applying these encodings?

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1268 @@
...
+## 5. Id Columns
+
+This section defines more details related to identifier columns used to establish relationships 
+between payload types in the OTAP data model.
+
+### 5.1 Primary Keys and Foreign Keys
+
+All parent-child relationships in the OTAP data model follow a uniform convention:
+
+- **Parent tables** define an `id` column as their primary key
```
albertlockett: This might also be a good place to say a few additional things about the ID column:

1. It is nullable - null meaning the row had no children (e.g. a Log w/ no attributes)
2. It is unique within a batch, but is not globally unique (the same ID will be reused across subsequent batches)

docs/otap-spec.md
```diff
@@ -0,0 +1,1268 @@
...
+## 4. Payload Specifications
+
+This section defines the complete Arrow schema for all OTAP payload types, organized by signal category.
+
+**Table Column Descriptions:**
+- **Name**: Field name in the Arrow schema
+- **Type**: Base Arrow type for this field
+- **Alt Representations**: Alternative encodings allowed for this field (e.g., `Dict(u8)` for Dictionary(UInt8, Type), `List(T)` for list types). See also: [Dictionary Encoding](https://arrow.apache.org/docs/format/Columnar.html#dictionary-encoded-layout)
+- **Nullable**: Whether the field can contain null values
+- **Required**: Whether this field must be present in every record
```
albertlockett, line 382: Regarding whether things are required, there is a bit more we can say on this.

1. Non-required (optional) columns may (or maybe _should_) be excluded if their values are either entirely null, or default value (meaning "" for str, 0 for int, 0.0 for floats, etc.)
  - The exception to this rule about an all default-value column being excluded which is ID columns. E.g. If there was a single Log row with attributes so it had a single 0 in the ID column. This would actually be more common in the resource.id column, if for example all logs/metrics/spans came form a single resource.
  
2. RecordBatchs for certain Payload Types are optional as well (this might have already been called out elsewhere). E.g. if there are no log attributes, we don't send a en empty record batch - the record batch for this payload type would just not be present.

OTAP_FORMAL_SPEC.md
albertlockett: Also, wrt to uniqueness, I guess resource.id and span.id are special cases because they are not unique for each row.

docs/otap-spec.md
albertlockett: With respect to null vs. default value, something I unfortunately don't yet have my head around is under what circumstances these are semantically equivalent in OTel's data model (and whether we're actually handling them correctly).

OTAP_FORMAL_SPEC.md
JakeDern: Sure, will drop for now. I trust your instinct on it

OTAP_FORMAL_SPEC.md
JakeDern: Agreed, I've dropped the word "unique" and taken a note to find somewhere to talk about the scope of uniqueness of the ids. Let me know if you have a good idea on it, I'll update here based on where I slot it back in.

OTAP_FORMAL_SPEC.md
JakeDern: Agreed, updated to `Definition of univariate/multivariate metrics (Root)` as appropriate.

OTAP_FORMAL_SPEC.md
JakeDern: Ah good call on the body, I keep forgetting that it's nested (that anyvalue thing we discussed a long time ago). I fixed the dictionary encoding stuff as well.

Really good catch on the `event_name` too - Neither our current data model diagram ~nor the go schema have it either, though there is a builder in the go implementation for it. I'll open an issue about it, not sure why it hasn't been causing problems.~ Just kidding I'm just blind. It's only missing from the data diagram

OTAP_FORMAL_SPEC.md
JakeDern: Yeah I originally tried to structure by signal type and then group within signals, but I think that ended up being a silly idea. I'll reorganize this.

JakeDern: Reorganized and consolidate these.  I did end up organizing by signal, but only for "entities". Then At the end I have a section for attributes and split into u16 and u32.

OTAP_FORMAL_SPEC.md
JakeDern: Fixed the quantiles

OTAP_FORMAL_SPEC.md
JakeDern: Reorganized all of the definitions for attributes

OTAP_FORMAL_SPEC.md
JakeDern: >I wonder if there's a way to restructure how we define these type columns to make it clearer when each type is used. I'm worried the "Alt Representation" might be interpreted as types that are accepted albeit non-standard, or are chosen in some laissez-faire fashion when in reality there is some logic to which type is used.

Yeah "Alt Representation" is not great, I changed the type columns like 3 times already and still don't like it 😅.

The one thing I do like about the current version is that the Native Type is stated plainly in the `Type` field and not mixed in with other stuff. I think presenting the column as it's Native type and then letting the exact representation be an additional implementation detail maps well to how we think about OtapArrowRecords in code.

The first name I tried for "Alt Representations" was "Encodings" which soon could include REE, but I was afraid of overloading the word "Encoding" especially in the case of IDs. I also didn't want to have to state "Native" every time under the Representations/Encodings column so that's why I tried labeling as "Alt".

> For the case of dictionaries, the rules are basically like - OTAP will try to use the smallest dictionary key it can, and eventually overflow to the native (non dict-encoded) array.

I think discussion about which dictionary key types are selected and how is best left to an appendix or documentation about individual components/implementations i.e. "Producer implementations SHOULD prefer dictionary encodings where possible and SHOULD select the smallest possible key size for the type and only increase size when detecting overflow". 

I wasn't sure if "Producers" was actually a concept I should even include in the spec directly since there's technically nothing requiring a valid client have the same Producer abstraction that we have.

>Maybe instead of Alt Representation, we have a column like "Dict key min size"?

I'm not sure about "Dict key min size". I think if there were a lot of dictionary sizes it might make sense, but with only two possible I kind of like stating it plainly. If the column is not allowed to be dictionary encoded then leaving this as `-` could also be interpreted as "no minimum, therefore any dictionary size is allowed".  I'd also like to leave the column open for adding REE in the future which could be another valid "encoding".

Since Arrow does canonically [refer to Dictionary/REE as "Encodings"](https://arrow.apache.org/docs/format/Columnar.html#dictionary-encoded-layout), maybe we could call it "Arrow Encodings" to make it clear that these are these are in the realm of Arrow and then we could refer to our Delta Encodings as something else?

OTAP_FORMAL_SPEC.md
JakeDern: >This might also be a good place to say a few additional things about the ID column:
>1. It is nullable - null meaning the row had no children (e.g. a Log w/ no attributes)
>2. It is unique within a batch, but is not globally unique (the same ID will be reused across subsequent batches)

Added!

>Also, wrt to uniqueness, I guess resource.id and span.id are special cases because they are not unique for each row.

Yes, I do have a section in there on the special case of resource and span ids! Currently section 5.3.

JakeDern: Random thought on cooperative optimization opportunities between batching and producers. In the reindexing code, I should check if the offset I compute is `0` and skip reassigning Ids.

 If producers don't start at 0 for Ids and instead try to continue until they overflow, it _might_ make that optimization get hit more often especially because I plan to sort the RecordBatches we're trying to reindex by their min/max Id in order to avoid reassignment at all when possible.

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1199 @@
...
+| LOG_ATTRS | Log-level attributes | LOGS |
+| RESOURCE_ATTRS | Resource attributes | LOGS |
+| SCOPE_ATTRS | Scope attributes | LOGS |
+
+#### 2.1.2 Metrics Signal Tables
+
+| Payload Type | Description | Parent Payload Type |
+|---|---|---|
+| UNIVARIATE_METRICS | Definition of univariate metrics (Root) | — |
+| MULTIVARIATE_METRICS | Definition of multivariate metrics (Root) | — |
```
lquerel: I'm wondering whether we should omit multivariate metrics from this formal specification for now, since they are not implemented yet and this would give us flexibility to revisit the design in the future.

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1199 @@
...
+The OTAP message layer defines additional restrictions and requirements for the contents of a BAR,
+the ArrowPayload, and BatchStatus messages. It defines which Payload Types are valid for which Services/Signals;
+rules around Schema Evolution, Schema Resets, and Error Handling; and when it is allowable to omit payloads
+entirely.
+
+#### 3.2.1 BatchArrowRecords Message
+
+The BatchArrowRecords (BAR) message is the fundamental unit of data transmission in OTAP. It represents 
+a complete set of related telemetry tables for a single signal type, containing all the tables needed 
+to reconstruct that signal.
```
lquerel: We reconstruct a **batch** of signals of same type.

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1199 @@
...
+in Apache Arrow's columnar format. Arrow IPC defines how schemas and data are serialized into byte streams using 
+a standardized format that can be read by any Arrow-compatible library. This layer enables:
+
+- **Dictionary/Delta Dictionary encoding**: Efficient representation of repeated string values
+- **Zero-copy deserialization**: Data can be read directly from wire format without copying
+
+Clients and servers communicating over a Stream are managing multiple independent Arrow IPC streams,
+one for each Payload Type that the client has sent. On the server side, each `record` is routed to a separate 
+IPC stream consumer based on its `type` and `schema_id`. This means that a single BAR may feed multiple 
+independent Arrow IPC stream readers simultaneously.
```
lquerel: I'm not sure this paragraph is clear enough. A BAR message is sent over a **gRPC stream** between the client and the server. This BAR contains a collection of ArrowPayloads (of different types).

So within a single gRPC stream, multiple BAR messages are exchanged, each containing multiple ArrowPayloads, and for each (payload type, schema id) tuple, there is an **Arrow IPC stream**.

OTAP_FORMAL_SPEC.md
lquerel: Perhaps the term adaptive representation (or data-optimized encoding, performance-adaptive encoding) better captures the intent here. We aim to use the most appropriate and optimal type/encoding based on the data being transported.

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1199 @@
...
+
+---
+
+## Appendix C: Implementation Notes
+
+### C.1 Performance Considerations
+
+1. **BAR sizing**: Larger BARs improve compression but increase memory usage
+2. **Dictionary encoding**: Most effective for low-to-medium cardinality (10-10,000 unique values)
+3. **Transport encoding**: Most effective when data has strong sequential patterns
```
lquerel: Do you mean delta encoding?

lquerel (commented):
There are still a few gaps here and there, **but wow, this is truly an outstanding spec**! I'm really happy to see this. It's a key milestone toward formalizing the protocol and enabling other implementations more easily.

OTAP_FORMAL_SPEC.md
JakeDern: Redid this whole section in the second iteration of the draft.

OTAP_FORMAL_SPEC.md
JakeDern: I rewrote much of the document, this will be covered in section: `4. IPC Stream Management and Schema Resets` when I push the new version shortly.

docs/otap-spec.md
JakeDern: Some good points to discuss! Non-required optional columns being excluded if their values are entirely null is outlined in 4.2:

```md
If the column is marked as optional in the OTAP spec, it SHOULD be omitted entirely by the Client if there 
is no data for that Column.
```

Regarding interpreting a missing column as null or a default value, this seems like a confusing point. Like you mention, different columns like `id` would want different interpretations when the column in missing. I'm not actually sure which columns could regularly take advantage of an "all default" value. Maybe this is a situation where we should be recommending REE instead which would be more explicit? I know it's not in the spec yet as allowed, but maybe we can explore that more.

In 5.5.1 I did try to define rules for AnyValues specifically:

```md
A `type` of `0` indicates that the value for the attribute is empty.

In the case of the LOGS payload, note that the AnyValue is contained in the 
`body` column which is also nullable. In this case a null entry is semantically
equivalent to a `type` of `0`.

If the column for the Active Field is not present, or the `type` falls outside 
of the allowed range, then the value for the key is also interpreted as empty.
```

The [OTel specification](https://opentelemetry.io/docs/specs/otel/common/#anyvalue) uses this term "empty" instead of null.

JakeDern: I also do mention the omission of payloads with 0 rows in 3.1:

```
`arrow_payloads` additionally MUST include the primary/root table (LOGS, SPANS,
or UNIVARIATE_METRICS) and SHOULD omit payloads with 0 rows.
```

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1174 @@
...
+["AnyValue"](https://opentelemetry.io/docs/specs/otel/common/#anyvalue). 
+This is a value representing a union of one or more types: 
+`string`, `int`, `double`, `bool`, `map`, `slice`, or `bytes`.
+
+In OTAP this is implemented via multiple "value" columns either `str`, `int`, 
+`double`, `bool`, `bytes`, or `ser`.  `type` field that serves as the 
+discriminant and note that multiple AnyValue types can map to the `bytes` 
+column. See the below table for the exact mappings.
+
+// TODO: I see some mention of Arrow Sparse and Dense Unions in the Go code, is that out of date
```
JakeDern: @lquerel  - Just checking if I missed anything regarding how we represent AnyValues in OTAP.

I _think_ in the Go code that the only usage of Arrow Dense/Sparse union is this AttributesDT which has a comment indicating that it maybe should be removed in the future. AttributesDT also has no references outside of test code that I can find.

https://github.com/open-telemetry/otel-arrow/blob/9737434e98d32d80ae6d020117bde5b7909f0840/go/pkg/otel/common/arrow/attributes.go#L22

This code looks somewhat ancient (2022), maybe it's time to remove it?

docs/otap-spec.md
JakeDern: There are definitely some knowledge gaps on my end for the encodings, I'll read up on these more. This is one section that was not updated much in the second iteration.

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1174 @@
...
+
+**Dictionary used before definition**:
+- **Cause**: RecordBatch references dictionary not yet sent
+- **Status**: INVALID_ARGUMENT
+- **Action**: Client MUST send DictionaryBatch before referencing in RecordBatch
+
+---
+
+### 7.5 Interoperability
+
```
JakeDern: I would love some feedback on what to talk about (if anything) for this topic. There are many capabilities in OTAP, which puts a lot of burden on an implementation. Some of which you could consider just an optimization opportunity rather than strictly essential for speaking OTAP. Some are also dependent on arrow features which may or may not exist for the language's implementation.

 Some things that could be considered "optional" that come to mind:

- Delta dictionary support for IPC
- REE support in the future
- Support for delta encoded columns or subsets of the delta encodings

Additionally, we have the ability to downgrade to an OTLP connection in the go otel arrow exporter. That may be just an application specific capability, but if there is an explicit protocol negotiation mechanism we should prescribe then that's another thing we could talk about here.

Let me know if you have any thoughts!

CC: @jmacd @lquerel

OTAP_FORMAL_SPEC.md
albertlockett: I see we've changed to optimized encoding, and I think this is a good heading. 

It's accurate, both in the sense that it's space optimized for transport and some of our processing is actually more optimized this these dict encoded columns as well (think of attribute renaming).

It's nice phrasing because calling it "optimized" instead of "alternate" will gently encourage use of the dictionary encoding

docs/otap-spec.md
albertlockett: > If the column for the Active Field is not present, or the `type` falls outside 
of the allowed range, then the value for the key is also interpreted as empty.

Interesting - I have been interpreting these as invalid batches. 

For example, [here](https://github.com/open-telemetry/otel-arrow/blob/67bf2696f227001736b8c58b2aed4a6de0b648ef/rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs#L251-L252) and [here](https://github.com/open-telemetry/otel-arrow/blob/4aa67d74db11d16a333c588e0e51cda8854b5a10/rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs#L605-L607) we return errors while processing the batch if the attibrute type is outside the range 0-7.

I think we should probably handle these cases as invalid batches, rather than being permissive and treating the type as empty. I feel that in these cases, the type isn't actually indicating empty, it's indicating something ambiguous

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1171 @@
...
+---
+
+## 1. Introduction
+
+### 1.1 Purpose
+
+The OpenTelemetry Arrow Protocol (OTAP) defines a wire protocol for transmitting
+OpenTelemetry telemetry signals (logs, metrics, and traces) from a Client to a 
+Server. OTAP optimizes on multiple axis for compression efficiency, memory 
+usage, and processing speed while being semanticallyequivalent to OpenTelemetry
```
albertlockett:
```suggestion
usage, and processing speed while being semantically equivalent to OpenTelemetry
```

docs/otap-spec.md
JakeDern: I'm good with treating this as an error instead - I wasn't sure of the precedent here.

When writing this I was thinking of a pipeline where this wasn't detected (I think that's currently possible depending on components configured) and when we then output or forward data in this form. 

Defining it as invalid batches makes me think we need to make sure we always screen this in a pipeline otherwise we have the possibility of propagating junk. You could always write that off as "garbage in, garbage out" but then I worry about what happens when valid data is mixed with invalid like via the batch processor

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1171 @@
...
+
+service ArrowMetricsService {
+  rpc ArrowMetrics(stream BatchArrowRecords) returns (stream BatchStatus) {}
+}
+```
+
+Clients create an OTAP connection by intiating a gRPC connection and starting a
+gRPC stream for one or more services. `BatchArrowRecords` (BAR) messages are
+streamed from the client with corresponding `BatchStatus` acknowledgments
+streamed back from the server. The bi-directional streaming pattern allows the
```
albertlockett: nit: it might be helpful to work the word "stateful" into this section somehow. 
```suggestion
streamed back from the server. The bi-directional stateful streaming pattern allows the
```

The fact that it's stateful is implied by the fact that w/in the ArrowPayload contained w/in the BAR, we each record batch is a member of an Arrow IPC stream, which is stateful. However a user might not grok that straight a way, so just thinking it'll be helpful to give a little hint here about the statefulness.

OTAP_FORMAL_SPEC.md
albertlockett: > I wasn't sure if "Producers" was actually a concept

hmmm, I'm not sure if the word "Producer" would engender confusion.

"Encoders" instead of "Producers" would align with what we call the code that basically takes telemetry from some source and formulates OTAP from it. The module containing utilities for this in our pdata crate is called encode
https://github.com/open-telemetry/otel-arrow/tree/8e05fa92fc91eb708ed8f47eca8e15434bceab2f/rust/otap-dataflow/crates/pdata/src/encode

"Producer" we've used to name something that takes OTAP batches and serializes them as BARs
https://github.com/open-telemetry/otel-arrow/blob/8e05fa92fc91eb708ed8f47eca8e15434bceab2f/rust/otap-dataflow/crates/pdata/src/encode/producer.rs

docs/otap-spec.md
```diff
@@ -0,0 +1,1171 @@
...
+Schema Message prior to any Dictionary, Delta Dictionary, or Record Batch 
+Messages in the Stream.
+
+Servers MUST keep track of Arrow IPC Stream state per gRPC Stream and per
+Payload Type and `schema_id` within the Stream.
+
+When a Client changes the `schema_id` sent for some Payload Type, this is known
+as a Schema Reset. The Server, upon receiving such an `ArrowPayload`, MUST 
+begin processing the Messages in the `record` field of that message as if they 
+begin a new IPC Stream.
```
albertlockett, line 354: Should we callout that the server only keeps track of one schema_id per payload time at a time?

e.g. a client couldn't send
```
{ type: Logs, schema_id: 1, record: ... }, // start of IPC stream
{ type: Logs, schema_id: 2, record: ... }, // start of new IPC stream
{ type: Logs, schema_id: 1, record: ... }, // continuation of stream from first message (no Schema in IPC payload)
```
The third message wouldn't work

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1171 @@
...
+| span_id | FixedSizeBinary(8) | — | No | Yes | — | — | Span id |
+| trace_state | Utf8 | — | Yes | No | — | — | W3C trace state |
+| parent_span_id | FixedSizeBinary(8) | — | Yes | No | — | — | Parent span id |
+| name | Utf8 | — | No | Yes | — | — | Span name |
+| kind | Int32 | — | Yes | No | — | — | Span kind enum |
+| dropped_attributes_count | UInt32 | — | Yes | No | — | — | Number of dropped span attributes |
+| dropped_events_count | UInt32 | — | Yes | No | — | — | Number of dropped events |
+| dropped_links_count | UInt32 | — | Yes | No | — | — | Number of dropped links |
+| status_code | Int32 | — | Yes | No | — | — | Span status code |
+| status_status_message | Utf8 | — | Yes | No | — | — | Status message |
```
albertlockett: Status is actually a struct column with two fields, code and message:
https://github.com/open-telemetry/otel-arrow/blob/ca375646ebd9f257c341d03230a745532342ce8f/go/pkg/otel/traces/arrow/traces.go#L23-L46

https://github.com/open-telemetry/otel-arrow/blob/ca375646ebd9f257c341d03230a745532342ce8f/go/pkg/otel/traces/arrow/status.go#L22-L25

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1171 @@
...
+This is a value representing a union of one or more types: 
+`string`, `int`, `double`, `bool`, `map`, `slice`, or `bytes`.
+
+In OTAP this is implemented via multiple "value" columns either `str`, `int`, 
+`double`, `bool`, `bytes`, or `ser`.  `type` field that serves as the 
+discriminant and note that multiple AnyValue types can map to the `bytes` 
+column. See the below table for the exact mappings.
+
+// TODO: I see some mention of Arrow Sparse and Dense Unions in the Go code,
+is that out of date or am I missing some details there?
```
albertlockett: As far as I know these types aren't used in OTAP

docs/otap-spec.md
```diff
@@ -0,0 +1,1171 @@
...
+column. See the below table for the exact mappings.
+
+// TODO: I see some mention of Arrow Sparse and Dense Unions in the Go code,
+is that out of date or am I missing some details there?
+
+This technique is used in all Payload Types that utilize Attribute16 and 
+Attribute32 schemas as well as the LOGS `body` field.
+
+The "Active Field" is defined as the field that contains valid data for the 
+row, all other value fields MUST be ignored for that row.
```
albertlockett: I like that we have a term for this now, the "Active Field"!

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1171 @@
...
+
+| type value | AnyValue type | Active Field |
+|------------|--------------|---------|
+| 0 | empty | N/A |
+| 1 | str | str |
+| 2 | int | int |
+| 3 | double | double |
+| 4 | bool | bool |
+| 5 | map | bytes |
+| 6 | slice | bytes |
```
albertlockett:
```suggestion
| 5 | map | ser |
| 6 | slice | ser |

File: docs/otap-spec.md (hunk 25 above)
Comment (JakeDern) [reply docs/otap-spec.md:354]: Yeah I was also wondering about the server tracking one schema id per payload - Is this a requirement or an implementation detail? 

I have this TODO written that kind of hints at an edge case related to this:

```
TODO: Should we require batch_ids be strictly increasing? Is there a reason we'd 
like to know which messages came first/second from a client? What if a schema 
reset occurs and then we get a nack and need to re-transmit?
```

I'm imagining that a client sends

```
{ type: Logs, schema_id: 1, batch_id: 1 }
{ type: Logs, schema_id: 1, batch_id: 2 }
{ type: Logs, schema_id: 2, batch_id: 3 }
```
And the server streams back:

```
Ack { batch_id: 1 }
Ack { batch_id: 3 }
Nack { batch_id: 2}
```

So we've already moved on from schema_id 1. Now the client gets a _retryable_ nack back, but we're juggling more than one in-flight request. How do we re-transmit? Based on whether there was a schema reset since that batch id went out, we could make better choices, but we have to be able to detect that.

Assuming there wasn't a schema reset, we could just re-transmit it.  If we can detect that there's been a reset since this batch_id went out, then we could maybe run it back through the producer again and reconstruct it with the current schema + a new batch_id rather than resetting. 

So maybe baking in a requirement that ids are increasing so both sides have information about what order things happened in is useful. We could store something like: `{id_of_last_reset: 3, payload_type: Logs}`, so that when we get `Nack { batch_id: 2 }` from the server, we know it came before.

File: OTAP_FORMAL_SPEC.md (hunk 27 above)
Comment (JakeDern) [reply OTAP_FORMAL_SPEC.md]: Yeah I think the code is just stale, but checking with Laurent here: https://github.com/open-telemetry/otel-arrow/pull/2040#discussion_r2854373594

File: OTAP_FORMAL_SPEC.md [hunk 30]
```diff
@@ -0,0 +1,1171 @@
...
+Resource and scope entities are **not** represented as separate payload types.
+Instead, they are embedded as struct fields within root tables (LOGS, SPANS, 
+or METRICS).
+
+Each root table contains:
+
+- `resource.id` (UInt16): Identifier for the resource
+- `scope.id` (UInt16): Identifier for the instrumentation scope
+
+Note that there are no RESOURCE or SCOPE payload types. Resources and scopes
```
albertlockett: > Note that there are no RESOURCE or SCOPE payload types.

nit: this phrase redundant? repeated from previous paragraph.

OTAP_FORMAL_SPEC.md
```diff
@@ -0,0 +1,1171 @@
...
+### 7.3 Error Handling Rules
+
+#### 7.3.1 Schema Errors
+
+// TODO: We probably need to define behaviors for all of these
+
+**Invalid schema**:
+- **Cause**: Schema message is malformed or uses unsupported types
+- **Status**: INVALID_ARGUMENT
+- **Action**: Client MUST fix schema definition
```
albertlockett: The client only has to fix the schema if it wishes to retry the request, right? If it's OK with dropping the data, it's free to not do any correction. Maybe i'm interpretting MUST too literally here 😆

OTAP_FORMAL_SPEC.md
JakeDern: Fixed! What's your opinion on `status.status_message`? It kinda looks like from the proto spec that it should be `status.code` and `status.message`.

https://github.com/open-telemetry/opentelemetry-proto/blob/56bc2ce8643a824c36eee4acb1bf0eaf24e87458/opentelemetry/proto/trace/v1/trace.proto#L308C1-L329C1

docs/otap-spec.md
JakeDern: Proposal related to this here for discussion! https://github.com/open-telemetry/otel-arrow/issues/2168

Going to resolve this comment in favor of the issue

OTAP_FORMAL_SPEC.md
JakeDern: https://github.com/open-telemetry/otel-arrow/issues/2136

OTAP_FORMAL_SPEC.md
JakeDern: deleted this

docs/otap-spec.md
JakeDern: These notes were added!

OTAP_FORMAL_SPEC.md
JakeDern: Spoke with Laurent, this should be good to remove

OTAP_FORMAL_SPEC.md
JakeDern: Added this to the parent issue for future discussion

JakeDern: #1957

JakeDern:
> LGTM! I think we can merge this in its current form eh @JakeDern ? We can certainly evolve the spec in future PRs as well

Yeah, definitely far from completely done, but I think in a good spot to be merged in draft status!

PR #2041, opened 2026-02-14
fix: quiver persistence engine fails with EPERM on filesystems that don't support chmod/set_readonly

### LINKED ISSUE #2039: Quiver persistence engine fails with EPERM on filesystems that don't support chmod/set_readonly
When Quiver runs on filesystems that reject permission changes, such as certain Kubernetes volumeMounts, the engine's attempts to set finalized segment files and rotated WAL files to read-only via set_permissions fail with EPERM (Operation not permitted). This is because Quiver unconditionally calls `chmod`/`set_readonly` after writing segment and WAL files to enforce immutability, but these filesystems do not support POSIX permission semantics.

AaronRM:
# Change Summary

Fix an issue where Quiver may fail (`EPERM`) on filesystems that don't support `chmod`/`set_readonly`. The fix introduces a lightweight startup probe (`probe_set_permissions_support`) that creates a temporary file, attempts a `chmod`/`set_readonly`, and records whether the operation succeeds. When the probe detects an unsupported filesystem, the `enforce_file_readonly` flag is propagated to both the `SegmentWriter` and `WalWriter`, causing them to skip read-only enforcement gracefully while logging a warning so operators are aware that file-level immutability protection is degraded.

* Closes #2039

Added relevant unit tests, tested with supported filesystems (ext4) and unsupported (FAT32 from Linux)

No.

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/engine.rs
+++ b/rust/otap-dataflow/crates/quiver/src/engine.rs
@@ -1472,11 +1472,28 @@ fn probe_set_permissions_support(dir: &Path) -> bool {
     );
     let probe_path = dir.join(&probe_name);
 
-    let result = (|| -> std::io::Result<()> {
-        // Create a temporary probe file
-        let _file = fs::File::create(&probe_path)?;
+    // Create a temporary probe file. If this fails (disk full, quota exceeded,
+    // directory missing, etc.) we cannot probe permissions at all. Treat this as
+    // a setup error rather than evidence that permissions are unsupported —
+    // default to `true` so we don't silently disable readonly enforcement.
+    let file_created = match fs::File::create(&probe_path) {
+        Ok(_file) => true,
+        Err(e) => {
+            otel_warn!(
+                "quiver.engine.init",
+                error = %e,
+                path = %dir.display(),
+                reason = "permissions_probe_create_failed",
+                message = "could not create probe file to test permission support; \
+                    assuming permissions are supported",
+            );
+            return true;
+        }
+    };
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -1420,14 +1439,114 @@ async fn initialize_wal_writer(
...
+        SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .unwrap_or(Duration::ZERO)
+            .as_nanos(),
+    );
+    let probe_path = dir.join(&probe_name);
+
+    let result = (|| -> std::io::Result<()> {
+        // Create a temporary probe file
+        let _file = fs::File::create(&probe_path)?;
```
lalitb: nit - If file creation fails (for disk full, quota exceeded etc), we still return the error as "filesystem does not support permission changes" - this could be misleading.

lalitb:
Should [purge_rotated_files](https://github.com/open-telemetry/otel-arrow/blob/a6e0e74750abf6f949784dfada1df293d2c7250b/rust/otap-dataflow/crates/quiver/src/wal/writer.rs#L1489) also respect `enforce_file_readonly`  before calling `set_permissions` on non-Unix? Otherwise we may still hit permission errors even when readonly enforcement is disabled.

rust/otap-dataflow/crates/quiver/src/engine.rs
AaronRM: Separated creation of the file from probing read-only support.

PR #2042, opened 2026-02-15
Add user-configurable custom node attributes (for telemetry)

jmacd:
# Change Summary

This is an alternative to #2026.

This adds a `#[compose]` for a single-key attribute ("custom") with map-value attribute supporting user-configurable key:value pairs. The reason this is a single attribute and map value is because the AttributeSet descriptor() returns a static, and this was the simplest approach (to me). On the other hand, this required adding a new map-valued attribute type, which was a little bit more code.

#1949 

✅ 

```yaml
    telemetry_attributes:
      pipeline_name: "MyCustomPipelineName"
      node_sequence: 1
```

yields

> 2026-02-15T21:39:25.317Z  INFO  otap-df-otap::receiver.start: Fake data generator receiver started [signals_per_second=1000 signals/sec, max_batch_size=1000, metrics_per_iteration=0, traces_per_iteration=0, logs_per_iteration=1000, data_source=SemanticConventions, generation_strategy=Fresh] entity/node.attrs: node.id=generator node.urn=urn:otel:traffic_generator:receiver node.type=receiver pipeline.id=default_pipeline pipeline.group.id=default_pipeline_group core.id=0 numa.node.id=0 process.instance.id=AGOGGPTN6F2SFIS46XKA4GKQGM host.id=JoshCorpSurfaceLaptop container.id= **custom={node_sequence=1, pipeline_name=MyCustomPipelineName}**

Yes.

### DIFF
--- a/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
+++ b/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
@@ -63,7 +63,7 @@ pub fn get_log_record_schema() -> &'static ParserMapSchema {
 pub struct BridgePipeline {
     attributes_schema: Option<ParserMapSchema>,
     pipeline: PipelineExpression,
-    include_dropped_records: bool,
+    options: BridgeOptions,
 }
 
 impl BridgePipeline {
@@ -117,10 +117,9 @@ pub fn parse_kql_query_into_pipeline(
     query: &str,
     options: Option<BridgeOptions>,
 ) -> Result<BridgePipeline, Vec<ParserError>> {
-    let include_dropped_records = options
-        .as_ref()
-        .is_none_or(|v| v.get_include_dropped_records());
-    let parser_options = build_parser_options(options).map_err(|e| vec![e])?;
+    let mut options = options.unwrap_or_default();
+
+    let parser_options = build_parser_options(&mut options).map_err(|e| vec![e])?;
     let attributes_schema = match parser_options
         .get_source_map_schema()
         .and_then(|s| s.get_schema_for_key("Attributes"))
@@ -132,7 +131,7 @@ pub fn parse_kql_query_into_pipeline(
     Ok(BridgePipeline {
         attributes_schema,
         pipeline: result.pipeline,
-        include_dropped_records,
+        options,
     })
 }
 
@@ -241,7 +240,7 @@ pub fn process_export_logs_service_request_using_pipeline(
     let mut final_results = batch.flush();
 
     let mut dropped_records = None;
-    let dropped_record_count = if pipeline.include_dropped_records {
+    let dropped_record_count = if pipeline.options.get_include_dropped_records() {
         for record in initial_dropped_records.into_iter() {
             final_results.dropped_records.push(record);
         }
@@ -270,7 +269,11 @@ pub fn process_export_logs_service_request_using_pipeline(
 
             let count = final_results.dropped_records.len();
 
-            process_log_record_results(&mut dropped_records_request, final_results.dropped_records);
+            process_log_record_results(
+                pipeline.options.get_diagnostic_options(),
+                &mut dropped_records_request,
+                final_results.dropped_records,
+            );
 
             dropped_records = Some(dropped_records_request);
 
@@ -291,6 +294,7 @@ pub fn process_export_logs_service_request_using_pipeline(
     if has_summaries || has_included_records {
         if has_included_records {
             process_log_record_results(
+                pipeline.options.get_diagnostic_options(),
                 &mut export_logs_service_request,
                 final_results.included_records,
             );
@@ -300,9 +304,9 @@ pub fn process_export_logs_service_request_using_pipeline(
             let mut log_records = Vec::new();
 
             for summary in final_results.summaries.included_summaries {
-                let diagnostics = summary.to_string();
+                let diagnostics = summary.diagnostics;
 
-                if let Some(map) = summary.map {
+                let mut log_record = if let Some(map) = summary.map {
                     let mut attributes: Vec<(Box<str>, AnyValue)> =
                         Vec::with_capacity(map.len() + 1);
 
@@ -312,17 +316,7 @@ pub fn process_export_logs_service_request_using_pipeline(
                             .map(|(key, value)| (key, value.into())),
                     );
 
-                    if !diagnostics.is_empty() {
-                        attributes.push((
-                            "query_engine.output".into(),
-                            AnyValue::Native(OtlpAnyValue::StringValue(StringValueStorage::new(
-                                diagnostics,
-                            ))),
-                        ));
-                    }
-
-                    log_records.push(LogRecord::new().with_attributes(attributes));
-                    included_record_count += 1;
+                    LogRecord::new().with_attributes(attributes)
                 } else {
                     let mut attributes: Vec<(Box<str>, AnyValue)> = Vec::with_capacity(
                         summary.aggregation_values.len() + summary.group_by_values.len() + 1,
@@ -369,18 +363,20 @@ pub fn process_export_logs_service_request_using_pipeline(
                         }
                     }
 
-                    if !diagnostics.is_empty() {
-                        attributes.push((
-                            "query_engine.output".into(),
-                            AnyValue::Native(OtlpAnyValue::StringValue(StringValueStorage::new(
-                                diagnostics,
-                            ))),
-                        ));
-                    }
+                    LogRecord::new().with_attributes(attributes)
+                };
 
-                    log_records.push(LogRecord::new().with_attributes(attributes));
-                    included_record_count += 1;
-                }
+                handle_diagnostics(
+                    pipeline.options.get_diagnostic_options(),
+                    &mut (&mut log_record, &diagnostics),
+                    |s| BridgeLogRecordDiagnostics::new(&pipeline.pipeline, s.0, s.1),
+                    |s, n, v| {
+                        s.0.attributes.get_values_mut().insert(n.into(), v);
+                    },
+                );
+
+                log_records.push(log_record);
+                included_record_count += 1;
             }
 
             let summary_otlp = ResourceLogs::new().with_scope_logs(
@@ -407,7 +403,7 @@ pub fn process_export_logs_service_request_using_pipeline(
     })
 }
 
-fn build_parser_options(options: Option<BridgeOptions>) -> Result<ParserOptions, ParserError> {
+fn build_parser_options(options: &mut BridgeOptions) -> Result<ParserOptions, ParserError> {
     let mut parser_options = ParserOptions::new().with_attached_data_names(&[
         "resource",
         "instrumentation_scope",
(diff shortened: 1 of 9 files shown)


rust/otap-dataflow/crates/config/src/node.rs
```diff
@@ -206,4 +240,41 @@ config: {}
             .collect();
         assert_eq!(cfg.outputs, expected);
     }
+
+    #[test]
+    fn node_user_config_with_tellemetry_attributes_valid() {
```
utpilla:
```suggestion
    fn node_user_config_with_telemetry_attributes_valid() {
```

rust/otap-dataflow/crates/config/src/node.rs
```diff
@@ -206,4 +240,41 @@ config: {}
...
+            BTreeSet::from([
+                "attr1".to_string(),
+                "attr2".to_string(),
+                "attr3".to_string(),
+            ])
+        );
+    }
+
+    #[test]
+    fn node_user_config_with_tellemetry_attribute_array_expects_error() {
```
utpilla:
```suggestion
    fn node_user_config_with_telemetry_attribute_array_expects_error() {
```

rust/otap-dataflow/crates/config/src/node.rs
```diff
@@ -206,4 +240,41 @@ config: {}
...
+    }
+
+    #[test]
+    fn node_user_config_with_telemetry_attribute_array_expects_error() {
+        let json = r#"{
+            "type": "urn:example:demo:receiver",
+            "telemetry_attributes": {
+                "attr1": "value1",
+                "attr2": [1, 2, 3]
+            },
```
utpilla: This is not valid JSON because of the trailing comma. The test works because it just checks for the deserialization to fail and not the reason.

```suggestion
            }
```

rust/otap-dataflow/crates/engine/src/attributes.rs
```diff
@@ -75,6 +110,64 @@ pub struct NodeAttributeSet {
...
+/// with user-defined attributes without requiring static descriptors.
+#[derive(Debug, Clone)]
+pub struct CustomAttributeSet {
+    values: Vec<AttributeValue>,
+}
+
+impl Default for CustomAttributeSet {
+    fn default() -> Self {
+        Self {
+            values: vec![AttributeValue::Map(BTreeMap::new())],
```
utpilla, line 144: This default always produces a single-element vec with an empty map, which means every node, even those without any `telemetry_attributes` configured — will emit `custom={}` in self-tracing log scope attributes and `custom="{}"` as a metrics attribute.

I think we should use an empty `vec` here instead (and also update `new()` to also skip inserting when the map is empty). That way the `#[compose]` flattening via `attribute_values()` would contribute zero entries when there are no custom attributes, and the custom key would only appear in telemetry output for nodes that actually configure it.

Alternatively, the encoder/dispatcher could skip map-valued attributes when the map is empty, but filtering at the source seems cleaner.

utpilla (approved):

rust/otap-dataflow/crates/config/src/node.rs
```diff
@@ -43,6 +45,14 @@ pub struct NodeUserConfig {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub default_output: Option<PortName>,
 
+    /// Node specific attributes to be added to internal telemetry.
+    #[serde(
+        default,
+        skip_serializing_if = "HashMap::is_empty",
+        deserialize_with = "deserialize_telemetry_attributes"
+    )]
+    pub telemetry_attributes: HashMap<String, AttributeValue>,
```
lalitb: The schema for this field (generated via `JsonSchema` on `AttributeValue`) includes the `Array` variant, but `deserialize_telemetry_attributes` doesn't actually accept arrays at runtime. So if someone passes an array in their config, it'll pass right through schema validation and then break when the config is loaded — and there's nothing to warn them it was ever going to fail.

Either annotate this field with `#[schemars(schema_with = "...")]` to emit a schema that excludes `Array`, or introduce a narrower `ScalarAttributeValue` type    
 (`String | Bool | I64 | F64`, no `Array`) and use that here instead of `AttributeValue` - which also lets you delete `deserialize_telemetry_attributes` entirely.

rust/otap-dataflow/crates/config/src/node.rs
```diff
@@ -165,9 +179,29 @@ impl NodeUserConfig {
...
+/// which is not supported for log record attributes.
+fn deserialize_telemetry_attributes<'de, D>(
+    deserializer: D,
+) -> Result<HashMap<String, AttributeValue>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let attrs: HashMap<String, AttributeValue> = HashMap::deserialize(deserializer)?;
+    for (key, value) in &attrs {
+        if matches!(value, AttributeValue::Array(_)) {
```
lalitb: As mentioned [here](https://github.com/open-telemetry/otel-arrow/pull/2042/changes#r2820877752) - This rejects arrays at runtime, but the JSON Schema generated for the `telemetry_attributes` field still advertises `Array` as valid.

rust/otap-dataflow/crates/telemetry/src/attributes.rs
```diff
@@ -176,6 +180,13 @@ impl AttributeValue {
             AttributeValue::UInt(u) => u.to_string(),
             AttributeValue::Double(f) => f.to_string(),
             AttributeValue::Boolean(b) => b.to_string(),
+            AttributeValue::Map(m) => {
```
lalitb, line 183: The `Map` serialization here could cause a subtle metrics issue worth addressing.

`to_string_value()` collapses the whole map into an unescaped `key=value, key=value` string, but `=` and `, ` are also valid characters inside keys and values. So two completely different maps can produce the exact same string:

```
{"a": "b", "c": "d"}  ->  {a=b, c=d}
{"a": "b, c=d"}       ->  {a=b, c=d}  <- collision here
```

Since this string ends up as an OTel metric label, two nodes with different  telemetry_attributes` configs could silently share a label and get aggregated together silently.

In practice this probably won't happen often since most configs use simple identifiers, but it's the kind of bug that's really hard to track down when it does.

Two options to fix this:
- Quote values in `to_string_value()` for `Map` (e.g. `key="value"` with `"` escaped) so the format is unambiguous.
- Or expand each map entry into its own OTel attribute (`custom <key> = value`) in the dispatcher instead of collapsing everything into one string.

rust/otap-dataflow/crates/engine/src/context.rs
```diff
@@ -110,9 +112,12 @@ pub struct PipelineContext {
     thread_id: usize,
     pipeline_group_id: PipelineGroupId,
     pipeline_id: PipelineId,
+    pipeline_telemetry_attrs: HashMap<String, AttributeValue>,
```
lalitb: nit: `pipeline_telemetry_attrs` - this seems to be declared and cloned, but never used.

rust/otap-dataflow/crates/engine/src/attributes.rs
```diff
@@ -5,8 +5,43 @@
...
+) -> AttributeValue {
+    use otap_df_config::pipeline::service::telemetry::AttributeValue as ConfigValue;
+    match value {
+        ConfigValue::String(s) => AttributeValue::String(s.clone()),
+        ConfigValue::Bool(b) => AttributeValue::Boolean(*b),
+        ConfigValue::I64(i) => AttributeValue::Int(*i),
+        ConfigValue::F64(f) => AttributeValue::Double(*f),
+        ConfigValue::Array(arr) => {
+            // Encode arrays as a string representation
+            AttributeValue::String(format!("{:?}", arr))
```
lalitb, line 28: - I believe this arm is unreachable as node.rs hard-rejects any `Array` value before the config map is built.
-And also, `{:?}` on AttributeValue Array produces Rust internal Debug output, which will expose rust internal data structures.

lquerel:
EDIT: Based on a conversation with @utpilla, it seems I misinterpreted the requirement. The idea appears to be to allow attributes to be added via configuration. That seems obvious after rereading the description. I'm not sure why it didn't immediately stand out to me.

--
I'm working under the assumptions that 1) the attributes are known at initialization time, both keys and values, and 2) the attribute values do not change after registration. If these assumptions hold, then why not extend `crates/engine/src/context.rs` with a method like the one below?

Note: I haven't tested it, so I may be missing something, but in principle it seems to me that it should work.

```rust
pub fn register_metrics_with_attributes<T: MetricSetHandler + Default + Debug + Send + Sync, A: AttributeSetHandler + Send + Sync + 'static>(
        &self,
        attrs: A,
    ) -> MetricSet<T> {
        if let Some(telemetry) = current_node_telemetry_handle() {
            telemetry.register_metric_set_with_attributes::<T>(attrs)
        } else {
            self.controller_context
                .telemetry_registry_handle
                .register_metric_set::<T>(attrs)
        }
}
```

Then we would extend `entity_context.rs` with a method like:

```rust
pub(crate) fn register_metric_set_with_attributes<T: MetricSetHandler + Default + Debug + Send + Sync>(
    &self,
    attrs: impl AttributeSetHandler + Send + Sync + 'static,
) -> MetricSet<T> { ... }
```

From there, in a node implementation, we could define a custom `AttributeSet`:

```rust
#[attribute_set(name = "<xyz>.node.attrs")]
#[derive(Debug, Clone, Default, Hash)]
struct XyzAttributeSet {
    #[compose]
    node_attrs: NodeAttributeSet,
    #[attribute(key = "receiver.protocol")]
    protocol: Cow<'static, str>,
    #[attribute(key = "receiver.wait_for_result")]
    wait_for_result: bool,
}
```

And then, instead of performing the usual metric registration, we could call the previously defined method:

```rust
let metrics = pipeline_ctx.register_metrics_with_attributes::<XyzMetrics, _>(
    XyzAttributeSet {
        node_attrs: pipeline_ctx.node_attribute_set(),
        protocol: Cow::Borrowed("blabla"),
        wait_for_result: config.wait_for_result,
    },
);
```

rust/otap-dataflow/crates/config/src/node.rs
lquerel: It's a detail, but I'd really like us to have the same level of description for these attributes as we already have internally. At a minimum, this would mean:
- The ability to add a brief description, optionally for each attribute
- A clear split between `entity_attributes` and `descriptive_attributes` instead of a single `telemetry_attributes` section

rust/otap-dataflow/crates/engine/src/attributes.rs
lquerel: I think we could simply remove `node_telemetry_attrs` entirely from the `NodeAttributeSet` structure and instead create a separate `NodeWithCustomAttributeSet` structure that would contain these custom attributes. We could then introduce a metric registration method that supports these custom attributes, as described in my comment at the end of the thread.

jmacd:
Thank you @utpilla and @lquerel 

I will follow up with https://github.com/open-telemetry/otel-arrow/issues/2078

PR #2048, opened 2026-02-16
Support datetime in OPL Parser

### LINKED ISSUE #2035: OPL Parser support datetime literals
In KQL, we can have datetime literals supported like: `datetime(2025-02-04 00:00:01.1)`. We should support the same thing in OPL.

Once this is supported, update the tests that filter in datetime in columnar query engine to have cases that use the OPL Parser.

albertlockett:
# Change Summary

Adds support for datetime literals in OPL Parser. This uses the same syntax as KQL, where various formats are supported either as literal text, or embedded w/in a string.
```kql
logs | where time_unix_nano > datetime(2026-02-04)
logs | where time_unix_nano > datetime("2026-02-04")
```

Supported formats include those currently also supported in KQL parser:
- US/mid-endian (MM/DD/YYYY)
- RFC 822 (e.g. "Wed, Feb 4, 2026, 14:10:00 GMT")
- ISO 8601 (e.g. 2026-02-04T05:30:00-05:00).

* Closes #2035

New unit tests. 

Note: the tests for the actual String to DateTime parsing of many formats is already covered by the test cases in [`data_engine_expressions::primitives::date_utils`](https://github.com/open-telemetry/otel-arrow/blob/947ff506f68fad78f3e97294f38795274cc1fc3f/rust/experimental/query_engine/expressions/src/primitives/date_utils.rs#L329-L532)

No

### DIFF
--- a/rust/otap-dataflow/crates/opl/src/parser/expression.rs
+++ b/rust/otap-dataflow/crates/opl/src/parser/expression.rs
@@ -451,6 +451,10 @@ pub fn parse_member_expression(rule: Pair<'_, Rule>) -> Result<LogicalOrScalarEx
         Rule::index_expression => parse_index_expression(rule),
         Rule::primitive_expression => parse_primitive_expression(rule),
         Rule::attribute_selection_expression => parse_attribute_selection_expression(rule),
+        Rule::datetime_expression => {
+            let date_time_expr = parse_datetime_expression(rule)?;
+            Ok(ScalarExpression::Static(date_time_expr).into())
+        }
         Rule::function_call => parse_function_call(rule),
         invalid_rule => Err(invalid_child_rule_error(
             query_location,
@@ -581,15 +585,10 @@ fn parse_primitive_expression(rule: Pair<'_, Rule>) -> Result<LogicalOrScalarExp
             BooleanScalarExpression::new(query_location, false),
         ))
         .into()),
-
         Rule::null_token => Ok(ScalarExpression::Static(StaticScalarExpression::Null(
             NullScalarExpression::new(query_location),
         ))
         .into()),
-        Rule::datetime_expression => {
-            let date_time_expr = parse_datetime_expression(rule)?;
-            Ok(ScalarExpression::Static(date_time_expr).into())
-        }
         Rule::expression => parse_expression(rule),
         invalid_rule => Err(invalid_child_rule_error(
             query_location,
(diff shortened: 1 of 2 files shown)


rust/otap-dataflow/crates/opl/src/parser/temporal.rs
```diff
@@ -0,0 +1,199 @@
...
+                        )
+                    })?
+                }
+                _ => {
+                    todo!("missed the buss")
+                }
+            }
+        }
+        _ => {
+            todo!("AH HELL MA")
```
lquerel: Shouldn't we return an error here instead of panicking, especially since this `todo` can be triggered by an OPL expression that is just configuration and therefore easy to forge?

rust/otap-dataflow/Cargo.toml
```diff
@@ -291,15 +292,15 @@ debug = "line-tables-only"  # minimum required for profiling
 [profile.bench]
 inherits = "release"
 opt-level = 3
-debug = false
+# debug = false
```
lquerel: Is it intentional?

rust/otap-dataflow/crates/opl/src/parser/expression.rs
```diff
@@ -778,6 +783,13 @@ mod test {
                 "null",
                 StaticScalarExpression::Null(NullScalarExpression::new(QueryLocation::new_fake())),
             ),
+            (
+                "datetime(2026-02-04)",
```
lquerel, line 786: Should we also add another test with `datetime("2026-02-04")` and datetime(\`2026-02-04\`) (if this last one is supported) ?

rust/otap-dataflow/crates/opl/src/parser/temporal.rs
albertlockett: oops, looks like I have missing commits

albertlockett: fixed in 1a4784354

rust/otap-dataflow/Cargo.toml
albertlockett: oops, no sorry. Those were leftover from when I was running some profiling for this PR https://github.com/open-telemetry/otel-arrow/pull/2038

Fixed in 80ee7e34d

rust/otap-dataflow/crates/opl/src/parser/expression.rs
albertlockett: Sure, that can't hurt. Added in abd445223.

Backtick string isn't yet supported - only double quote and single quote.

PR #2050, opened 2026-02-17
Add support for MMSC instruments by introducing Histogram bridge

utpilla:
# Change Summary

### Motivation
We need to record latency-style metrics (e.g., request duration) that capture min, max, sum, and count — capabilities provided by histograms. However, the OpenTelemetry specification does not support Async/Observable Histograms, and our internal telemetry subsystem collects metrics via periodic snapshots of pre-aggregated state rather than individual observations.

This PR introduces an internal MMSC instrument (`Min/Max/Sum/Count`) and exports it as an **OpenTelemetry SDK Histogram with no bucket boundaries.** This preserves the exact MMSC semantics through the standard SDK pipeline without requiring spec-level changes.

### Why This Works? Correctness:
An OpenTelemetry SDK histogram built with `.with_boundaries(vec![])` disables bucket counting and only exports four values: **min, max, sum, count**. To reconstruct these four values from a pre-aggregated `MmscSnapshot { min, max, sum, count }`, we issue synthetic `histogram.record()` calls:

| Count of measurements | Strategy |
|------:|----------|
| 0 | No-op |
| 1 | `record(sum)` — when count is 1, `min = max = sum` |
| 2 | `record(min)`, `record(max)` |
| ≥ 3 | `record(min)`, `record(max)`, then `record(fill)` × (`count − 2`), where `fill = (sum − min − max) / (count − 2)` |

The SDK histogram tracks `min = min(observations)`, `max = max(observations`), `sum = Σ observations`, `count = len(observations)`. Since we always record the exact min and max, those are preserved. The remaining `count − 2` observations sum to `sum − min − max`, and dividing evenly preserves the total sum. Therefore, the exported `HistogramDataPoint` carries the exact original min, max, sum, and count.

### Code Changes
**New instrument** — `instrument.rs`
- `Mmsc` struct with `record(f64)`, `get() -> MmscSnapshot`, `reset()` methods
- `MmscSnapshot`- immutable snapshot holding min/max/sum/count

**New descriptor variant** — `descriptor.rs`
- Added `Instrument::Mmsc` to the instrument enum

**Snapshot pipeline** — `metrics.rs`
- Introduced `SnapshotValue` enum (`Scalar(MetricValue)` | `Mmsc(MmscSnapshot)`) replacing bare `MetricValue` throughout the snapshot pipeline
- MMSC-aware accumulation in `MetricSetRegistry` - merges via min-of-mins, max-of-maxes, sum-of-sums, count-of-counts
- Updated `MetricSetHandler::snapshot_values()` return type, `MetricsEntry`, and `MetricsIterator` to use `SnapshotValue`

**OTel export** — `dispatcher.rs`
- `record_synthetic_histogram()` implements the formula above
- `add_opentelemetry_metric()` routes `Instrument::Mmsc` to the synthetic histogram path
- Full end-to-end test using `InMemoryMetricExporter` validates min/max/sum/count and empty bucket boundaries

**Derive macro** — `lib.rs`
- `#[derive(MetricSetHandler)]` now handles `Mmsc` fields (no generic type parameter, always F64/Delta)

**Admin / observability endpoints** — `telemetry.rs`
- MMSC values expand into four sub-metrics in both Prometheus (`_min`, `_max`, `_sum`, `_count` with appropriate types) and Line Protocol formats
- Tests for both output formats

**Downstream updates** — `registry.rs`, `collector.rs`, `parquet_exporter.rs`, `metrics_types.rs`
- Migrated from `MetricValue` to `SnapshotValue` throughout

Partially addresses #2051 by adding support for MMSC instrument.

Unit tests

Yes, component authors would now be able to use MMSC instrument.

### DIFF
--- a/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
+++ b/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge.rs
@@ -63,7 +63,7 @@ pub fn get_log_record_schema() -> &'static ParserMapSchema {
 pub struct BridgePipeline {
     attributes_schema: Option<ParserMapSchema>,
     pipeline: PipelineExpression,
-    include_dropped_records: bool,
+    options: BridgeOptions,
 }
 
 impl BridgePipeline {
@@ -117,10 +117,9 @@ pub fn parse_kql_query_into_pipeline(
     query: &str,
     options: Option<BridgeOptions>,
 ) -> Result<BridgePipeline, Vec<ParserError>> {
-    let include_dropped_records = options
-        .as_ref()
-        .is_none_or(|v| v.get_include_dropped_records());
-    let parser_options = build_parser_options(options).map_err(|e| vec![e])?;
+    let mut options = options.unwrap_or_default();
+
+    let parser_options = build_parser_options(&mut options).map_err(|e| vec![e])?;
     let attributes_schema = match parser_options
         .get_source_map_schema()
         .and_then(|s| s.get_schema_for_key("Attributes"))
@@ -132,7 +131,7 @@ pub fn parse_kql_query_into_pipeline(
     Ok(BridgePipeline {
         attributes_schema,
         pipeline: result.pipeline,
-        include_dropped_records,
+        options,
     })
 }
 
@@ -241,7 +240,7 @@ pub fn process_export_logs_service_request_using_pipeline(
     let mut final_results = batch.flush();
 
     let mut dropped_records = None;
-    let dropped_record_count = if pipeline.include_dropped_records {
+    let dropped_record_count = if pipeline.options.get_include_dropped_records() {
         for record in initial_dropped_records.into_iter() {
             final_results.dropped_records.push(record);
         }
@@ -270,7 +269,11 @@ pub fn process_export_logs_service_request_using_pipeline(
 
             let count = final_results.dropped_records.len();
 
-            process_log_record_results(&mut dropped_records_request, final_results.dropped_records);
+            process_log_record_results(
+                pipeline.options.get_diagnostic_options(),
+                &mut dropped_records_request,
+                final_results.dropped_records,
+            );
 
             dropped_records = Some(dropped_records_request);
 
@@ -291,6 +294,7 @@ pub fn process_export_logs_service_request_using_pipeline(
     if has_summaries || has_included_records {
         if has_included_records {
             process_log_record_results(
+                pipeline.options.get_diagnostic_options(),
                 &mut export_logs_service_request,
                 final_results.included_records,
             );
@@ -300,9 +304,9 @@ pub fn process_export_logs_service_request_using_pipeline(
             let mut log_records = Vec::new();
 
             for summary in final_results.summaries.included_summaries {
-                let diagnostics = summary.to_string();
+                let diagnostics = summary.diagnostics;
 
-                if let Some(map) = summary.map {
+                let mut log_record = if let Some(map) = summary.map {
                     let mut attributes: Vec<(Box<str>, AnyValue)> =
                         Vec::with_capacity(map.len() + 1);
 
@@ -312,17 +316,7 @@ pub fn process_export_logs_service_request_using_pipeline(
                             .map(|(key, value)| (key, value.into())),
                     );
 
-                    if !diagnostics.is_empty() {
-                        attributes.push((
-                            "query_engine.output".into(),
-                            AnyValue::Native(OtlpAnyValue::StringValue(StringValueStorage::new(
-                                diagnostics,
-                            ))),
-                        ));
-                    }
-
-                    log_records.push(LogRecord::new().with_attributes(attributes));
-                    included_record_count += 1;
+                    LogRecord::new().with_attributes(attributes)
                 } else {
                     let mut attributes: Vec<(Box<str>, AnyValue)> = Vec::with_capacity(
                         summary.aggregation_values.len() + summary.group_by_values.len() + 1,
@@ -369,18 +363,20 @@ pub fn process_export_logs_service_request_using_pipeline(
                         }
                     }
 
-                    if !diagnostics.is_empty() {
-                        attributes.push((
-                            "query_engine.output".into(),
-                            AnyValue::Native(OtlpAnyValue::StringValue(StringValueStorage::new(
-                                diagnostics,
-                            ))),
-                        ));
-                    }
+                    LogRecord::new().with_attributes(attributes)
+                };
 
-                    log_records.push(LogRecord::new().with_attributes(attributes));
-                    included_record_count += 1;
-                }
+                handle_diagnostics(
+                    pipeline.options.get_diagnostic_options(),
+                    &mut (&mut log_record, &diagnostics),
+                    |s| BridgeLogRecordDiagnostics::new(&pipeline.pipeline, s.0, s.1),
+                    |s, n, v| {
+                        s.0.attributes.get_values_mut().insert(n.into(), v);
+                    },
+                );
+
+                log_records.push(log_record);
+                included_record_count += 1;
             }
 
             let summary_otlp = ResourceLogs::new().with_scope_logs(
@@ -407,7 +403,7 @@ pub fn process_export_logs_service_request_using_pipeline(
     })
 }
 
-fn build_parser_options(options: Option<BridgeOptions>) -> Result<ParserOptions, ParserError> {
+fn build_parser_options(options: &mut BridgeOptions) -> Result<ParserOptions, ParserError> {
     let mut parser_options = ParserOptions::new().with_attached_data_names(&[
         "resource",
         "instrumentation_scope",
@@ -415,7 +411,7 @@ fn build_parser_options(options: Option<BridgeOptions>) -> Result<ParserOptions,
     ]);
 
     let (log_record_schema, summary_schema) =
-        build_log_record_schema(options.and_then(|mut v| v.take_attributes_schema()))?;
+        build_log_record_schema(options.take_attributes_schema())?;
 
     if let Some(summary_schema) = summary_schema {
         parser_options = parser_options.with_summary_map_schema(summary_schema);
@@ -491,24 +487,24 @@ fn build_log_record_schema(
 }
 
 fn process_log_record_results(
+    options: &BridgeDiagnosticOptions,
     response: &mut ExportLogsServiceRequest,
     records: Vec<RecordSetEngineRecord<LogRecord>>,
 ) {
-    for record_result in records {
-        let mut diagnostic_output = None;
+    for mut record_result in records {
+        handle_diagnostics(
+            options,
+            &mut record_result,
+            |s| s.into(),
+            |s, n, v| {
+                s.get_record_mut()
+                    .attributes
+                    .get_values_mut()
+                    .insert(n.into(), v);
+            },
+        );
 
-        let diagnostics = record_result.get_diagnostics();
-        if !diagnostics.is_empty() {
-            diagnostic_output = Some(record_result.to_string());
-        }
-
-        let mut log_record = record_result.take_record();
-        if let Some(d) = diagnostic_output {
-            log_record.attributes.get_values_mut().insert(
-                "query_engine.output".into(),
-                AnyValue::Native(OtlpAnyValue::StringValue(StringValueStorage::new(d))),
-            );
-        }
+        let log_record = record_result.take_record();
 
         let resource_id = log_record
             .resource_id
@@ -529,6 +525,34 @@ fn process_log_record_results(
             .expect("scope_logs were not found");
 
         scope_logs.log_records.push(log_record);
+    }
+}
+
+fn handle_diagnostics<S, FBuild, FAddAddtribute>(
+    options: &BridgeDiagnosticOptions,
+    state: &mut S,
+    build: FBuild,
+    add_attribute: FAddAddtribute,
+) where
+    FBuild: Fn(&S) -> BridgeLogRecordDiagnostics,
+    FAddAddtribute: Fn(&mut S, &str, AnyValue),
+{
+    let diagnostics = build(state);
+
+    if !diagnostics.is_empty() {
+        match options {
+            BridgeDiagnosticOptions::AddAttribute { attribute_name } => {
+                let diagnostics = diagnostics.to_string();
+                add_attribute(
+                    state,
+                    attribute_name.as_ref(),
+                    AnyValue::Native(OtlpAnyValue::StringValue(StringValueStorage::new(
+                        diagnostics,
+                    ))),
+                );
+            }
+            BridgeDiagnosticOptions::Callback(c) => c(diagnostics),
+        }
     }
 }
 
--- a/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge_options.rs
+++ b/rust/experimental/query_engine/engine-recordset-otlp-bridge/src/bridge_options.rs
@@ -1,12 +1,19 @@
 // Copyright The OpenTelemetry Authors
 // SPDX-License-Identifier: Apache-2.0
 
-use data_engine_kql_parser::*;
+use std::fmt::Display;
 
-#[derive(Debug, Clone, PartialEq)]
+use data_engine_expressions::PipelineExpression;
+use data_engine_kql_parser::*;
+use data_engine_recordset::{RecordSetEngineDiagnostic, RecordSetEngineRecord, format_diagnostics};
+
+use crate::LogRecord;
+
+#[derive(Debug, Clone)]
 pub struct BridgeOptions {
     attributes_schema: Option<ParserMapSchema>,
     include_dropped_records: bool,
+    diagnostic_options: BridgeDiagnosticOptions,
 }
 
 impl Default for BridgeOptions {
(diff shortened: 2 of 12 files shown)


rust/otap-dataflow/crates/telemetry/src/instrument.rs
```diff
@@ -454,6 +454,96 @@ impl From<f64> for Gauge<f64> {
...
+/// are exported. See [`record_synthetic_histogram`] for details.
+///
+/// [`record_synthetic_histogram`]: crate::metrics::dispatcher::MetricsDispatcher
+#[derive(Clone, Copy)]
+pub struct Mmsc {
+    min: f64,
+    max: f64,
+    sum: f64,
+    count: u64,
+}
```
lquerel, line 478: Unless I'm mistaken, since `Mmsc` appears to be a subset of the Summary metric as defined in OTLP, shouldn't we call this metric `Summary`?

rust/otap-dataflow/crates/telemetry/src/metrics.rs
```diff
@@ -120,6 +121,162 @@ impl std::ops::AddAssign for MetricValue {
...
+/// instruments produce `Mmsc(MmscSnapshot)`.
+#[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
+#[serde(untagged)]
+#[allow(variant_size_differences)] // Mmsc is 32 bytes vs 8 for Scalar; acceptable for internal telemetry.
+pub enum SnapshotValue {
+    /// A single scalar metric value (counter, gauge, histogram, up-down counter).
+    Scalar(MetricValue),
+    /// A pre-aggregated min/max/sum/count summary from an [`crate::instrument::Mmsc`] instrument.
+    Mmsc(MmscSnapshot),
+}
```
lquerel: Why not extend the `MetricValue` enum as follows:

```rust
pub enum MetricValue {
    /// Unsigned 64-bit integer value.
    U64(u64),
    /// 64-bit floating point value.
    F64(f64),
    Summary {
        min: f64,
        max: f64,
        count: u64,
        sum: f64,
    }
}
```

I find the current approach ambiguous because, up to now, snapshots are what is emitted by the pipelines on the hot path, and depending on the nature of the metric, these snapshots are either accumulated or replace the current metric value in the registry.

rust/otap-dataflow/crates/telemetry/src/metrics/dispatcher.rs
```diff
@@ -219,6 +252,52 @@ impl MetricsDispatcher {
...
+    /// This produces `count` calls to `histogram.record()` that yield
+    /// the same min/max/sum/count on the exported `HistogramDataPoint`:
+    /// - count == 0: no records
+    /// - count == 1: record(sum)  (since min == max == sum)
+    /// - count == 2: record(min), record(max)
+    /// - count >= 3: record(min), record(max), then record a fill value
+    ///   `(sum - min - max) / (count - 2)` for the remaining `count - 2` times.
+    ///
+    /// **Note:** The histogram is built with `.with_boundaries(vec![])` to
+    /// disable bucket counting, so only min/max/sum/count are exported.
```
lquerel, line 245: Isn't this closer to a summary than to a histogram?

rust/otap-dataflow/crates/telemetry/src/metrics/dispatcher.rs
```diff
@@ -219,6 +252,52 @@ impl MetricsDispatcher {
...
+                histogram.record(s.max, attributes);
+            }
+            n => {
+                histogram.record(s.min, attributes);
+                histogram.record(s.max, attributes);
+                let fill = (s.sum - s.min - s.max) / (n - 2) as f64;
+                for _ in 0..(n - 2) {
+                    histogram.record(fill, attributes);
+                }
+            }
```
lquerel, line 276: That potentially results in a lot of iterations for a relatively low-quality histogram :-)

lquerel (commented):
Made few suggestions.

rust/otap-dataflow/crates/telemetry/src/instrument.rs
utpilla: It is a subset of Summary metric, but the official docs seem to discourage from using it: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#summary-legacy

> This point type is not recommended for new applications and exists for compatibility with other formats.

Moreover, it would be a very poor summary metric as it would only have two quantile values: `0.0` and `1.0` (`min` and `max`). We wouldn't have median or 99th percentile value etc. which would be doing injustice to the summary metric.

However, OpenTelemetry Histogram [natively supports](https://github.com/open-telemetry/opentelemetry-proto/blob/d3fb76d70deb0874692bd0ebe03148580d85f3bb/opentelemetry/proto/metrics/v1/metrics.proto#L472-L473) the "MMSC" scenario by disabling bucket distribution computation.

rust/otap-dataflow/crates/telemetry/src/metrics/dispatcher.rs
utpilla: Agreed. However, this runs once per reporting interval on already-aggregated snapshots, not on the hot path, so the iteration cost should be acceptable.

rust/otap-dataflow/crates/telemetry/src/metrics.rs
utpilla: Fixed in https://github.com/open-telemetry/otel-arrow/pull/2050/changes/5f3a862c21e22f3abe2d78543e268f1ff08be96b

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -814,37 +934,76 @@ fn format_prometheus_text(
...
+                                "# HELP {} {}",
+                                metric_name,
+                                escape_prom_help(field.brief)
+                            );
+                        }
+                        let prom_type = match field.instrument {
+                            Instrument::Counter => "counter",
+                            Instrument::UpDownCounter => "gauge",
+                            Instrument::Gauge => "gauge",
+                            Instrument::Histogram => "histogram",
```
ThomsonTan: Why was this mapping changed to "histogram"? Seems like it is emit as a single scalar which sounds like a gauge.

utpilla: There was an existing inconsistency between `agg_prometheus_text` and `format_prometheus_text` in terms of what `Instrument::Histogram` should map. I have updated both to use `gauge`.

rust/otap-dataflow/crates/telemetry/src/metrics/dispatcher.rs
```diff
@@ -216,6 +227,53 @@ impl MetricsDispatcher {
...
+    fn record_synthetic_histogram(
+        field: &MetricsField,
+        s: &MmscSnapshot,
+        attributes: &[opentelemetry::KeyValue],
+        meter: &Meter,
+    ) {
+        if s.count == 0 {
+            return;
+        }
+        let histogram = meter
```
cijothomas, line 255: not introduced in this PR, but this is creating Histogram instrument without re-using. It's not in hot hot path, but something we could explore and avoid. (For Future PR consideration).

utpilla: Yeah, I found it odd when I saw this setup for other instruments. I followed the existing pattern for simplicity of review.

rust/otap-dataflow/crates/telemetry/src/instrument.rs
```diff
@@ -454,6 +454,97 @@ impl From<f64> for Gauge<f64> {
...
+impl Mmsc {
+    /// Records a single observation, updating min/max/sum/count.
+    #[inline]
+    pub fn record(&mut self, value: f64) {
+        if value < self.min {
+            self.min = value;
+        }
+        if value > self.max {
+            self.max = value;
+        }
```
lalitb, line 511: nit - good to also guard against  non-finite inputs?

rust/otap-dataflow/crates/telemetry/src/metrics/dispatcher.rs
```diff
@@ -216,6 +227,53 @@ impl MetricsDispatcher {
...
+                histogram.record(s.sum, attributes);
+            }
+            2 => {
+                histogram.record(s.min, attributes);
+                histogram.record(s.max, attributes);
+            }
+            n => {
+                histogram.record(s.min, attributes);
+                histogram.record(s.max, attributes);
+                let fill = (s.sum - s.min - s.max) / (n - 2) as f64;
```
lalitb, line 272: nit - `fill` can become negative if sum < min + max (eg bad snapshot). Should we guard against that - eg, clamp tiny negatives to 0.0, and log/drop when clearly negative

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -749,17 +841,45 @@ fn format_line_protocol(
...
+                    first = false;
+                    let _ = write!(
+                        &mut fields,
+                        "{}={}",
+                        escape_lp_field_key(field.name),
+                        format_lp_value(value, Some(field.value_type))
+                    );
+                }
+                MetricValue::Mmsc(s) => {
+                    for (suffix, fval) in [("_min", s.min), ("_max", s.max), ("_sum", s.sum)] {
```
lalitb, line 867: When `s.count == 0`, `s.min`/`s.max` are sentinel values (`f64::MAX`/`f64::MIN`). In mixed metric sets, this branch still emits `_min`/`_max`, which leaks invalid values into output.

   Please guard `count == 0` here (skip MMSC emission, or emit only `_count=0` and `_sum=0`).

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -546,31 +592,77 @@ fn agg_prometheus_text(groups: &[AggregateGroup], timestamp_millis: Option<i64>)
...
+                                "{metric_name}{{{base_labels}}} {value_str}{ts_suffix}"
+                            );
+                        }
+                    }
+                    MetricValue::Mmsc(s) => {
+                        let brief = escape_prom_help(field.brief);
+                        for (suffix, prom_type, val) in [
+                            ("_min", "gauge", s.min),
+                            ("_max", "gauge", s.max),
+                            ("_sum", "counter", s.sum),
```
lalitb, line 636: `_sum` is emitted as Prometheus counter, but `Mmsc::record()` allows negative observations. Counters must be non-negative and monotonic - Can we either (1) emit `_sum` as gauge, or (2) enforce non-negative MMSC inputs so `_sum` is truly counter-safe?

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -814,37 +934,76 @@ fn format_prometheus_text(
...
+                            "{metric_name}{{{base_labels}}} {value_str}{ts_suffix}"
+                        );
+                    }
+                }
+                MetricValue::Mmsc(s) => {
+                    let brief = escape_prom_help(field.brief);
+                    for (suffix, prom_type, val) in [
+                        ("_min", "gauge", s.min),
+                        ("_max", "gauge", s.max),
+                        ("_sum", "counter", s.sum),
```
lalitb, line 985: Same issue as [here](https://github.com/open-telemetry/otel-arrow/pull/2050/changes#r2823443579)

rust/otap-dataflow/crates/telemetry/src/metrics.rs
```diff
@@ -66,6 +86,9 @@ impl MetricValue {
                 MetricValue::F64(lhs) => {
                     *lhs += rhs as f64;
                 }
+                MetricValue::Mmsc(_) => {
+                    debug_assert!(false, "add_in_place: cannot add U64 to Mmsc");
```
lalitb, line 90: nit - `debug_assert!` is stripped in release, so this type mismatch becomes a silent no-op in prod. Is this intended?

rust/otap-dataflow/crates/telemetry/src/metrics.rs
```diff
@@ -74,6 +97,20 @@ impl MetricValue {
                 MetricValue::F64(lhs) => {
                     *lhs += rhs;
                 }
+                MetricValue::Mmsc(_) => {
+                    debug_assert!(false, "add_in_place: cannot add F64 to Mmsc");
```
lalitb, line 101: Same as [here](https://github.com/open-telemetry/otel-arrow/pull/2050/changes#r2823480360), also line 112.

rust/otap-dataflow/crates/telemetry/src/instrument.rs
lalitb: Actually thinking more, there is another issue - The default uses `f64::MAX/f64::MIN` as sentinels, but these break with `+-Inf` observations: first record `+Inf` leaves `min` stuck at `f64::MAX` as `+Inf < f64::MAX`; first record `-Inf` leaves `max` stuck at `f64::MIN` as `-Inf > f64::MIN`. 

 Use `f64::INFINITY`/`f64::NEG_INFINITY` as sentinels instead - then any finite-or-infinite value updates them correctly. Combined with an `is_finite()` guard in `record()`, both problems are solved.

lalitb (commented):
Really elegant solution - the synthetic histogram bridge is clever and well-executed. I've left a few inline comments worth addressing.  None are blocking for this PR, can be a follow-up.

rust/otap-dataflow/crates/telemetry/src/metrics.rs
utpilla: Yes, it would be no-op. That's an incorrect usage of `add_in_place` method so ideally nobody should call it like that.

rust/otap-dataflow/crates/telemetry/src/instrument.rs
utpilla: Deferring this to a follow-up. Today this matches what the OpenTelemetry SDK Histogram does. It accepts whatever `f64` you hand it with no `is_finite()` guard or rejection of `NaN`/`Inf`. The same gap exists in our other internal instruments of the telemetry crate too (e.g., Counters could receive negative or `NaN`/`Inf` values). This would be worth a broader discussion on whether we want runtime enforcement across all instruments or just a doc-level contract that callers pass valid values.

rust/otap-dataflow/crates/telemetry/src/metrics/dispatcher.rs
utpilla: Related comment: https://github.com/open-telemetry/otel-arrow/pull/2050#discussion_r2823991551

If the recorded values are negative, then the correct fill value could be negative as well. Similar to how if you use OpenTelemetry SDK Counter to record negative numbers, you will end up with a negative sum.

rust/otap-dataflow/crates/admin/src/telemetry.rs
utpilla: Related comment: https://github.com/open-telemetry/otel-arrow/pull/2050#discussion_r2823991551

This is following the same trust model as Counter today. `Counter<f64>::add()` also accepts negative values at the API level without enforcement, yet we emit it as a Prometheus counter. Both rely on callers passing valid inputs. Changing `_sum` to gauge only for MMSC while keeping Counter as counter would be inconsistent. Happy to revisit this holistically across all instruments if we decide to add runtime enforcement, but that's a broader conversation.

rust/otap-dataflow/crates/admin/src/telemetry.rs
utpilla: https://github.com/open-telemetry/otel-arrow/pull/2050#discussion_r2824046766

rust/otap-dataflow/crates/telemetry/src/metrics.rs
utpilla: https://github.com/open-telemetry/otel-arrow/pull/2050#discussion_r2823935123

rust/otap-dataflow/crates/admin/src/telemetry.rs
utpilla: Good point. Fixed in https://github.com/open-telemetry/otel-arrow/pull/2050/commits/da0cf4dc77b61997270df0e1039f3ee3bba03ab8

lquerel (approved):
LGTM

rust/otap-dataflow/crates/admin/src/telemetry.rs
jmacd: https://github.com/open-telemetry/otel-arrow/issues/2100

rust/otap-dataflow/crates/telemetry/src/instrument.rs
jmacd: Yes +1 to representing MMSC as Histogram w/o buckets. 💯

rust/otap-dataflow/crates/telemetry/src/instrument.rs
jmacd: See also #2100. We have to protect against Inf and NaN measurements, I would say.

rust/otap-dataflow/crates/telemetry/src/metrics.rs
jmacd: Looks good!

PR #2054, opened 2026-02-17
start using otel_ macros for logs

gouslu:
# Change Summary

replaced println statements to otel_ macros for logging. left out the stats, will be converted to metrics in a later PR.

Internal logging item in [this issue](https://github.com/open-telemetry/otel-arrow/issues/1396).

Manual testing.

No

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
@@ -87,7 +87,7 @@ impl Auth {
 
             otel_warn!(
                 "azure_monitor_exporter.auth.retry_scheduled",
-                delay_secs = format!("{:.1}", delay_secs)
+                delay_secs = %delay_secs
             );
             tokio::time::sleep(delay).await;
         }


rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
```diff
@@ -199,11 +194,7 @@ impl LogsIngestionClient {
                         base_delay.mul_f64(jitter_factor)
                     };
 
-                    println!(
-                        "[AzureMonitorExporter] Retry after {}ms: {:?}",
-                        delay.as_millis(),
-                        e
-                    );
+                    otel_warn!("export.retry_delay", delay_ms = delay.as_millis() as u64, error = ?e);
```
cijothomas: Warn needed?

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
```diff
@@ -93,7 +85,10 @@ impl Auth {
             let delay_secs = (capped_delay_secs + jitter).max(1.0);
             let delay = tokio::time::Duration::from_secs_f64(delay_secs);
 
-            println!("[AzureMonitorExporter] Retrying in {:.1}s...", delay_secs);
+            otel_info!(
+                "auth.retry_scheduled",
```
cijothomas: lets prefix azure_monitor.exporter  or something. the default target used crate name, which would be "otap-df-otap" only and no sufficient to tell this is coming from azure monitor exporter.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/transformer.rs
```diff
@@ -180,7 +180,7 @@ impl Transformer {
                         let writer = (&mut buf).writer();
                         let mut ser = serde_json::Serializer::new(writer);
                         if let Err(e) = record_map.serialize(&mut ser) {
-                            println!("Failed to serialize log entry: {e}");
+                            otel_warn!("transform.serialize_failed", error = %e);
```
cijothomas: if we are emitting a warning, could we include the impact? i.e are we going to drop this log or logs forever? is user at fault or this is just internal error, and users just need to report the issue to the owners of this crate?

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
gouslu: I thought you mentioned that the component itself is inferred. (?)

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/transformer.rs
gouslu: makes sense, I'll look at it a little more.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
cijothomas: crate name only. For this crate, need to do more specific as it has many

gouslu: ah ok, well, I'll prefix all of them then.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
gouslu: I am in general leaning towards more telemetry rather than less telemetry. this can be useful information that both indicates failure of export internally to the client and captures retry information. I can make it even more useful but adding more information though.

.vscode/settings.json
```diff
@@ -22,6 +22,13 @@
         "rust/experimental/query_abstraction/Cargo.toml",
         "rust/experimental/query_engine/Cargo.toml"
     ],
+    
+    // Enable specific features for rust-analyzer
+    "rust-analyzer.cargo.features": [
+        "experimental-exporters",
```
ThomsonTan: Is this intended? Like this will enable running rust-analyzer on these components for everyone.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
cijothomas: there is cost for each telemetry (not just the compute cost, but costs paid by users to ship the logs someplace, like a paid backend etc.)

"export.retry_delay" tells me the exporter is retrying after a delay, which is normal operation, not worth WARN levels. If this is something more serious, then I can see why WARN would makes sense.

(Agree we haven't written down rules in this repo (or anywhere) about guidance of picking severity.)

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
```diff
@@ -242,10 +233,26 @@ impl LogsIngestionClient {
...
+                Err(Error::forbidden(body))
+            }
             413 => Err(Error::PayloadTooLarge),
-            429 => Err(Error::RateLimited { body, retry_after }),
+            429 => {
+                let retry_after_secs = retry_after.map(|d| d.as_secs());
+                otel_warn!("azure_monitor_exporter.export.rate_limited", retry_after_secs = ?retry_after_secs);
+                Err(Error::RateLimited { body, retry_after })
+            }
             500..=599 => Err(Error::ServerError {
```
ThomsonTan, line 246: `otel_warn!()` for 500..=599 too?

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
```diff
@@ -93,7 +85,10 @@ impl Auth {
             let delay_secs = (capped_delay_secs + jitter).max(1.0);
             let delay = tokio::time::Duration::from_secs_f64(delay_secs);
 
-            println!("[AzureMonitorExporter] Retrying in {:.1}s...", delay_secs);
+            otel_info!(
+                "azure_monitor_exporter.auth.retry_scheduled",
+                delay_secs = format!("{:.1}", delay_secs)
```
ThomsonTan: This allocates a new `String` even the `INFO` level is disabled. Probalby use `%` display format or use the raw value to avoid allocating a new `String` every time.

.vscode/settings.json
gouslu: I think yes, but let me know why not?

drewrelmas: Not everyone may want to compile the experimental/'contrib' components - I don't think it should be the default. I've kept that in my own user `settings.json`.

In addition, for simplicity I use
```json
    "rust-analyzer.cargo.features": "all",
```

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
```diff
@@ -93,7 +85,10 @@ impl Auth {
             let delay_secs = (capped_delay_secs + jitter).max(1.0);
             let delay = tokio::time::Duration::from_secs_f64(delay_secs);
 
-            println!("[AzureMonitorExporter] Retrying in {:.1}s...", delay_secs);
+            otel_info!(
```
drewrelmas: Should the logs about retry be `debug` level? If this is standard operation of the exporter, I'm thinking don't need to clutter up `info` logs with it.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
gouslu: I'll park this for now. if it proves to have a heavy impact as you anticipate, then maybe we can revise?

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
gouslu: no but I think I'll convert to warn, I expect this to be rare.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
gouslu: this is how I see it btw

terminal failure or failures that are rather important -> error
failures, issues, potentially transient and shouldn't happen a lot -> warn
helpful with prod operation and diagnostics and health of the app and not too frequent -> info
frequent and used for detailed debugging and diagnostics -> debug

gouslu:
@cijothomas and @drewrelmas I am open to changing retry attempts from warn to debug if you insist, let me know. I am still on the edge when it comes to using them as warn, the main reason I chose warn is basically that I don't expect them to be often, especially auth I think deserves to be warn, maybe LA client can be relaxed a bit more, but I can also re-adjust and re-align after adding metrics based on how the metrics turn out to be.

drewrelmas (approved):
Your reasoning makes sense to me, but I'd also like to get @cijothomas's signoff too. Thanks for swapping away from `println` @gouslu!

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
cijothomas: This is unresolved @gouslu

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
```diff
@@ -104,17 +99,23 @@ impl Auth {
                 let mut options = ManagedIdentityCredentialOptions::default();
 
                 if let Some(client_id) = &auth_config.client_id {
-                    println!("Using user-assigned managed identity with client_id: {client_id}");
+                    otel_info!("azure_monitor_exporter.auth.credential_type", method = "user_assigned_managed_identity", client_id = %client_id);
```
cijothomas, line 102: we probably can leverage a startup event to include credential_type attribute to it, instead of dedicated event just for that. There are costs with each log statements (each callsite adds to static memory)

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/client.rs
```diff
@@ -242,15 +233,36 @@ impl LogsIngestionClient {
         let body = response.text().await.unwrap_or_default();
 
         match status.as_u16() {
-            401 => Err(Error::unauthorized(body)),
-            403 => Err(Error::forbidden(body)),
+            401 => {
```
cijothomas: Use single Event with status_code as attribute.

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
gouslu: can you give an example, I am not sure if I get it.

cijothomas: wait for https://github.com/open-telemetry/otel-arrow/pull/2061

rust/otap-dataflow/crates/otap/src/experimental/azure_monitor_exporter/auth.rs
gouslu: mb, I thought I got that one already.

PR #2060, opened 2026-02-19
feat: simplify quiver DiskBudget implementation to eliminate deadlocks

### LINKED ISSUE #2019: Bug: WAL replay may deadlock under Backpressure policy due to transient budget double-charge
When a Quiver engine restarts with the default Backpressure retention policy, `DiskBudget.used` is charged for both existing segments (via `scan_existing`) and WAL files (via `WalWriter::open`). If un-finalized WAL entries need to be replayed during startup, segment finalization calls `try_reserve` to pre-allocate space for the new segment file. However, the WAL bytes that the segment replaces are not released until after the segment is written and the WAL is purged (`persist_cursor` -> `purge_rotated_files`). This creates a transient double-charge window where both the WAL data and its corresponding segment data count against the budget simultaneously, pushing used above the cap and causing `try_reserve` to fail with `StorageAtCapacity`. Under Backpressure policy the cleanup callback cannot help because no subscriber has consumed anything during early startup, so the engine deadlocks: it cannot finalize because the budget is full, and the budget cannot shrink because finalization is what releases the WAL bytes.

The root cause is that the reservation-based budget model (`try_reserve` / `PendingWrite` / callbacks) is fundamentally at odds with the WAL-then-segment write ordering. There will always be a window where both representations coexist on disk.

## Proposed Fix: Simpler DiskBudget Implementation

We propose replacing the existing implementation with a significantly simpler watermark-based `DiskBudget` implementation that has two thresholds: a `soft_cap` that gates ingest, and a `hard_cap` that represents the true ceiling. In this model, `DiskBudget` serves as an accounting mechanism, not a reservation system. Plain `add(bytes)` / `remove(bytes)` methods record what has actually been written and deleted — no reservations, no callbacks, no CAS loops. The `soft_cap` is computed as `hard_cap` - `segment_headroom` (typically one segment target size), and the engine validates at startup that `hard_cap` ≥ `wal_max` + 2 × `segment_target_size`. This structurally guarantees room for exactly one segment finalization even when the budget is at the soft cap, so finalization always proceeds unconditionally and the deadlock is eliminated. Cleanup and drop-oldest reclaim move into the engine's `ingest()` path where they can be retried naturally, rather than hiding behind opaque budget callbacks.

Issue [was identified](https://github.com/open-telemetry/otel-arrow/pull/1954#discussion_r2775607673) by @utpilla (Thanks Utkarsh!)

### LINKED ISSUE #2020: Bug: quiver-e2e terminal left in raw mode after errors
When the quiver-e2e tool encounters an error or panics (e.g. from running `./target/release/quiver-e2e --disk-budget-mb 1` which causes a disk budget error), the terminal is left in raw mode with the alternate screen still active. This renders the user's shell unusable until they manually reset it. This happens because `Dashboard::cleanup()` is only called on the happy path. If the process exits via an early error return or a panic, the `crossterm` raw mode and alternate screen are never restored.

AaronRM:
# Change Summary

- Add a significantly simpler watermark-based `DiskBudget` implementation that has two thresholds: a `soft_cap` that gates ingest, and a `hard_cap` that represents the true ceiling. In this model, DiskBudget serves as an accounting mechanism, not a reservation system. Plain `add(bytes)` / `remove(bytes)` methods record what has actually been written and deleted without reservations, callbacks, or CAS loops. The `soft_cap` is computed as `hard_cap` - `segment_headroom` (typically one segment target size), and the engine validates at startup that `hard_cap >= wal_max + 2 * segment_target_size`. This structurally guarantees room for exactly one segment finalization even when the budget is at the soft cap, so finalization always proceeds unconditionally. Cleanup and drop-oldest reclaim move into the engine's `ingest()` path where they can be retried naturally, rather than hiding behind opaque budget callbacks.
- Fix an issue in `quiver-e2e` which could lead to bad terminal state after an error occurred.

* Closes #2019 
* Closes #2020

- Added / refactored tests for budget scenarios
- Local chaos/stress testing for validating budget boundary cases

No, this change is primarily about improving quiver's internal behavior.

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/budget.rs
+++ b/rust/otap-dataflow/crates/quiver/src/budget.rs
@@ -284,11 +284,7 @@ impl DiskBudget {
     /// (that's expected — finalization overshoot is bounded by
     /// `segment_headroom`).
     pub fn add(&self, bytes: u64) {
-        let _ = self
-            .used
-            .fetch_update(Ordering::Release, Ordering::Acquire, |current| {
-                Some(current.saturating_add(bytes))
-            });
+        let _ = self.used.fetch_add(bytes, Ordering::Release);
     }
 
     /// Records bytes deleted from disk.


rust/otap-dataflow/crates/quiver/src/budget.rs
```diff
@@ -324,563 +277,197 @@ impl DiskBudget {
...
-                reserved: 0,
-                committed: AtomicBool::new(false),
+    /// Called after WAL appends, segment writes, or when discovering
+    /// existing files at startup. May push `used` above `soft_cap`
+    /// (that's expected — finalization overshoot is bounded by
+    /// `segment_headroom`).
+    pub fn add(&self, bytes: u64) {
+        let _ = self
+            .used
+            .fetch_update(Ordering::Release, Ordering::Acquire, |current| {
```
utpilla: Could we simply use `fetch_add` instead of `fetch_update` which would be CAS loop ?

AaronRM: Yes, the tradeoff is we could *technically* overflow the u64 without using `fetch_update`/`saturating_add`, but that's very, very unlikely in practice. Will simplify to use `fetch_add`.

PR #2064, opened 2026-02-19
[otap-df-otap] Restructure Syslog CEF Receiver config to scope settings per protocol

utpilla:
# Change Summary

Restructures the Syslog CEF Receiver configuration so that endpoint and TLS settings are scoped under their respective protocol, rather than being flat top-level fields. This aligns with the OTLP Receiver's config pattern and makes the config model extensible for future protocol additions (e.g., Unix Domain Sockets).

### Motivation

The previous config had listening_addr, protocol, and tls as sibling fields:
- TLS is inherently TCP-specific, but nothing in the config structure enforced that
- A flat `listening_addr` (`SocketAddr`) wouldn't generalize to protocols which might be added in future such as Unix Domain Sockets that use a path instead
- Adding new protocols would require awkward conditional validation

### Config change

**Before**
```
Config
├── listening_addr: SocketAddr
├── protocol: "tcp" | "udp"
└── tls: Option<TlsServerConfig>     ← TCP-only, but not enforced by structure
```

**After**
```
Config
└── protocol: Protocol                ← tagged enum, exactly one variant
    ├── tcp: TcpConfig
    │   ├── listening_addr: SocketAddr
    │   └── tls: Option<TlsServerConfig>
    └── udp: UdpConfig
        └── listening_addr: SocketAddr
```

### YAML example

**Before**
```yaml
config:
  listening_addr: "0.0.0.0:5140"
  protocol: udp
```

**After**
```yaml
config:
  protocol:
    udp:
      listening_addr: "0.0.0.0:5140"
```

### Key design decisions
- **Tagged enum instead of struct with optional fields**: Unlike the OTLP Receiver (which supports running gRPC + HTTP simultaneously), the Syslog CEF Receiver supports exactly one protocol per instance. A serde externally-tagged enum enforces this at deserialization time with no runtime validation needed.
- `protocol` (singular): Reflects the one-protocol-per-instance constraint.
- TLS scoped to `TcpConfig`: UDP's `deny_unknown_fields` rejects any `tls` field, enforcing the constraint structurally.

Added unit tests

Yes

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
+++ b/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
@@ -1292,6 +1292,27 @@ mod config_tests {
         }
     }
 
+    #[test]
+    fn both_protocols_rejected() {
+        // The Protocol enum is externally tagged, so specifying both tcp and udp
+        // in the same config must be rejected — only one protocol per instance.
+        let json = serde_json::json!({
+            "protocol": {
+                "tcp": {
+                    "listening_addr": "127.0.0.1:5140"
+                },
+                "udp": {
+                    "listening_addr": "127.0.0.1:5145"
+                }
+            }
+        });
+        let config: Result<Config, _> = serde_json::from_value(json);
+        assert!(
+            config.is_err(),
+            "Config with both tcp and udp should be rejected"
+        );
+    }
+
     #[cfg(feature = "experimental-tls")]
     #[test]
     fn valid_tcp_with_tls() {


rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -53,41 +53,68 @@ const MAX_BATCH_SIZE: u16 = 100; // Maximum number of messages to build an Arrow
...
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct TcpConfig {
+    /// The address to listen on for TCP connections.
+    listening_addr: SocketAddr,
+    /// TLS configuration for secure TCP connections (Syslog over TLS, RFC 5425).
+    ///
+    /// When configured, TCP connections will require TLS.
+    #[cfg(feature = "experimental-tls")]
+    pub tls: Option<TlsServerConfig>,
```
lalitb: nit - tls is pub while listening_addr is private, and the struct itself is private.

utpilla:
```suggestion
    tls: Option<TlsServerConfig>,
```

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -53,41 +53,68 @@ const MAX_BATCH_SIZE: u16 = 100; // Maximum number of messages to build an Arrow
...
+#[derive(Debug, Deserialize)]
+#[serde(deny_unknown_fields)]
+struct UdpConfig {
+    /// The address to listen on for UDP datagrams.
+    listening_addr: SocketAddr,
+}
+
+/// Protocol configuration for the syslog CEF receiver.
+///
+/// Exactly one protocol variant must be specified.
```
lalitb, line 79: Nit: With serde's externally-tagged enum, input like this won't produce an error:

```yaml
    protocol:
      tcp:
        listening_addr: "0.0.0.0:514"
      udp:
        listening_addr: "0.0.0.0:513"
   ```
  Serde will silently deserialize one variant and ignore the other. The in-memory value is always valid (single variant), but the user won't be told their config has a redundant/ignored protocol block.   Not a blocker - just worth being aware of,  or probably adding a TODO that this would be fixed later with custom serde deserializer.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
lquerel: That's funny, a suggestion on your own PR 🙂

cijothomas: I do it all the time. It is easy to commit the change from UI itself for minor things, than going back to desktop and switch to right branch and do this!

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
lquerel: `serde` is great, but it does indeed have several issues, including this one. Another problem is that it doesn't preserve line and column numbers in error messages, which aren't particularly helpful to begin with.

I see two possible ways to address this. Either we continue using `serde` but combine it with a JSON Schema and a logic similar to what Weaver uses to generate more meaningful error messages, or we take a completely different approach based on a facet-style introspection model.

Nothing urgent here, but I think it's worth thinking about.

utpilla: @lalitb I investigated and it turns out we're safe due to our config deserialization path, but it's subtle.

The YAML config is first deserialized into `NodeUserConfig`, where the config field is serde_json::Value. This means a YAML block with both `tcp:` and `udp:` keys gets materialized as a JSON object with both keys preserved. Then the receiver calls `serde_json::from_value` to deserialize into `Config` (with the `Protocol` enum).

`serde_json::from_value` for externally-tagged enums rejects objects with more than one key So, for our production path (YAML → serde_json::Value → `serde_json::from_value` → `Config`), both protocols are rejected. The `both_protocols_rejected` test added in https://github.com/open-telemetry/otel-arrow/pull/2064/commits/2c8aaa7a7d3b5dfddc4102da4230ba2d9fad6f16 validates this.

You're right that `serde_yaml` would silently pick one variant if we were deserializing YAML directly into `Config`, but our intermediate `serde_json::Value` step prevents that issue.

PR #2066, opened 2026-02-19
[query-engine] Support routing RecordSet engine diagnostics into log streams

CodeBlanch:
# Changes

* Adds options so that hosts can manually handle diagnostics emitted when using OTLP bridge.
* Use the new options in KQL processor to route diagnostics into standard log streams.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
+++ b/rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
@@ -103,28 +103,28 @@ impl RecordsetKqlProcessor {
                         otap_df_telemetry::otel_debug!(
                             "processor.query_output",
                             processor = "recordset_kql",
-                            formatted_diagnostics = display(d)
+                            formatted_diagnostics = %d
                         );
                     }
                     RecordSetEngineDiagnosticLevel::Info => {
                         otap_df_telemetry::otel_info!(
                             "processor.query_output",
                             processor = "recordset_kql",
-                            formatted_diagnostics = display(d)
+                            formatted_diagnostics = %d
                         );
                     }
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
```diff
@@ -69,20 +71,64 @@ impl RecordsetKqlProcessor {
...
+                    .get_diagnostics()
+                    .iter()
+                    .map(|v| v.get_diagnostic_level())
+                    .max()
+                    .unwrap_or_default();
+
+                match max {
+                    RecordSetEngineDiagnosticLevel::Verbose => {
+                        otap_df_telemetry::otel_debug!(
+                            "processor.query_output",
```
drewrelmas, line 104: Can you take a look at guidance from Cijo at https://github.com/open-telemetry/otel-arrow/pull/2054#discussion_r2819519210

I think we're trending towards having `recordset_kql_processor.` prefix for logs from this processor.

cijothomas: yes we need to add this prefix. The automatic thing only gets the crate name, so for situations like this where a single crate host multiple component, lets explicitly prefix with the component name.

CodeBlanch: There are a bunch of log calls in here I'll need to switch. Let me do those all together in small follow-up to this PR.

drewrelmas: Approving for now - talking with @CodeBlanch offline about better representation. Follow-up PR is coming.

drewrelmas (approved):

PR #2056, opened 2026-02-17
[config] Move to `otel_dataflow/v1` config file structure, centralize ITS, and normalize policy resolution

lquerel:
# Change Summary

  This PR completes the config model consolidation and policy refactor for the dataflow engine.

  - Simplified configuration loading to a single root format: OtelDataflowSpec (version: otel_dataflow/v1).
  - Removed support for loading standalone PipelineConfig files as engine runtime input.
  - Renamed top-level pipeline_groups to groups.
  - Moved internal telemetry pipeline (ITS) declaration from pipeline-level fields (internal, internal_connections) to:
      - engine.observability.pipeline.nodes
      - engine.observability.pipeline.connections
  - Consolidated engine settings into EngineConfig.
  - Removed pipeline ServiceConfig / duplicated telemetry config path; telemetry is now configured centrally and via policy resolution.
  - Introduced/expanded hierarchical policy resolution (top-level -> group -> pipeline, with observability-specific override where applicable):
      - policies.channel_capacity.control.node
      - policies.channel_capacity.control.pipeline
      - policies.channel_capacity.pdata (default set to 128)
      - policies.health
      - policies.telemetry
      - policies.resources.core_allocation (default top-level behavior: all cores)
  - Explicitly rejected resources policy on observability pipeline for now.
  - Unified resolved config handling:
      - OtelDataflowSpec::resolve() produces a deterministic resolved snapshot.
      - Observability pipeline is represented in resolved pipelines with role tagging.
  - Controller cleanup:
      - deduplicated run paths (run_forever / run_till_shutdown) through a shared execution path
      - consumes resolved config once and uses config-owned observability internal IDs.
  - Updated docs/tests/config fixtures accordingly (including README.md and files under configs/).

## What issue does this PR close?

  - Closes #1833 
  - Closes #1871 
  - Partially #1830 
  
  ## How are these changes tested?

Checked with `cargo xtask check` 
  
## Are there any user-facing changes?

Yes (breaking config changes):

  - Runtime config must be OtelDataflowSpec with version: otel_dataflow/v1.
  - pipeline_groups is now groups.
  - ITS config moved to engine.observability.pipeline.
  - Pipeline-level service / telemetry path removed.
  - Policy fields moved/standardized under hierarchical policies sections.
  
## Example configuration file

```yaml
version: otel_dataflow/v1

# This configuration file reproduces the continuous benchmarking setup used
# in our CI pipelines. The traffic generators, system under test, and backend
[...]

PR #2069, opened 2026-02-19
Add aws s3 storage support to otap-dataflow

antonmry:
# Change Summary

Add AWS S3 storage support to the otap-dataflow parquet exporter, mirroring the existing Azure storage integration. This includes configurable authentication methods (default credential chain, static credentials, web identity, and assume role) and S3-specific options like custom endpoints for S3-compatible stores (e.g., LocalStack).

* Related to https://github.com/open-telemetry/otel-arrow/issues/501

  - Unit tests for all AWS auth method deserialization variants (`default`, `static_credentials`, `web_identity`, `assume_role`)
  - Unit test for S3 `ObjectStore` construction via `from_storage_type`
  - Unit tests for S3 storage config deserialization with each auth method
  - All tests gated behind the `aws` feature flag
  - Example config file (`fake-parquet-s3.yaml`) for local tested with LocalStack

Yes. Users can now configure the parquet exporter to write to AWS S3 (or S3-compatible stores) by specifying an `s3` storage block in the pipeline config. A new `aws` feature flag must be enabled at build time. An example configuration is provided in `configs/fake-parquet-s3.yaml`.

### DIFF
--- a/rust/otap-dataflow/configs/fake-parquet-s3.yaml
+++ b/rust/otap-dataflow/configs/fake-parquet-s3.yaml
@@ -1,39 +1,39 @@
-settings:
-  default_pipeline_ctrl_msg_channel_size: 100
-  default_node_ctrl_msg_channel_size: 100
-  default_pdata_channel_size: 100
+version: otel_dataflow/v1
+engine: { }
+groups:
+  default:
+    pipelines:
+      main:
+        policies:
+          channel_capacity:
+            control:
+              node: 100
+              pipeline: 100
+            pdata: 128
 
-nodes:
-  receiver:
-    type: traffic_generator:receiver
-    config:
-      traffic_config:
-        max_batch_size: 100
-        signals_per_second: 10
-        log_weight: 100
-      registry_path: https://github.com/open-telemetry/semantic-conventions.git[model]
-  exporter:
-    type: parquet:exporter
-    config:
-      storage:
-        s3:
-          base_uri: s3://otel-test-bucket/telemetry
-          region: us-east-1
+        nodes:
+          receiver:
+            type: traffic_generator:receiver
+            config:
+              traffic_config:
+                max_batch_size: 100
+                signals_per_second: 10
+                log_weight: 100
+              registry_path: https://github.com/open-telemetry/semantic-conventions.git[model]
+          exporter:
+            type: parquet:exporter
+            config:
+              storage:
+                s3:
+                  base_uri: s3://my-bucket/telemetry
+                  region: us-east-1
+                  auth:
+                    type: default
+              partitioning_strategies:
+                - schema_metadata: ["_part_id"]
+              writer_options:
+                flush_when_older_than: 10s
 
-          # Optional config options available for development against S3
-          # compatible store like LocalStack:
-          endpoint: http://localhost:4566
-          allow_http: true
-          virtual_hosted_style_request: false
-          auth:
-            type: static_credentials
-            access_key_id: test
-            secret_access_key: test
-      partitioning_strategies:
-      - schema_metadata: ["_part_id"]
-      writer_options:
-        flush_when_older_than: 10s
-
-connections:
-  - from: receiver
-    to: exporter
+        connections:
+          - from: receiver
+            to: exporter
--- a/rust/otap-dataflow/configs/fake-parquet-azure.yaml
+++ b/rust/otap-dataflow/configs/fake-parquet-azure.yaml
@@ -16,22 +16,23 @@ groups:
             type: traffic_generator:receiver
             config:
               traffic_config:
-                max_batch_size: 1000
-                signals_per_second: 30
+                max_batch_size: 100
+                signals_per_second: 10
                 log_weight: 100
               registry_path: https://github.com/open-telemetry/semantic-conventions.git[model]
           exporter:
             type: parquet:exporter
             config:
               storage:
-                file:
-                  base_uri: /tmp
+                azure:
+                  base_uri: https://mystorageaccount.blob.core.windows.net/mycontainer/telemetry
+                  auth:
+                    type: azure_cli
               partitioning_strategies:
-                - schema_metadata: [ "_part_id" ]
+                - schema_metadata: ["_part_id"]
               writer_options:
                 flush_when_older_than: 10s
 
         connections:
           - from: receiver
             to: exporter
-
--- a/rust/otap-dataflow/configs/fake-parquet-local-s3.yaml
+++ b/rust/otap-dataflow/configs/fake-parquet-local-s3.yaml
@@ -16,22 +16,29 @@ groups:
             type: traffic_generator:receiver
             config:
               traffic_config:
-                max_batch_size: 1000
-                signals_per_second: 30
+                max_batch_size: 100
+                signals_per_second: 10
                 log_weight: 100
               registry_path: https://github.com/open-telemetry/semantic-conventions.git[model]
           exporter:
             type: parquet:exporter
             config:
               storage:
-                file:
-                  base_uri: /tmp
+                s3:
+                  base_uri: s3://otel-test-bucket/telemetry
+                  region: us-east-1
+                  endpoint: http://localhost:4566
+                  allow_http: true
+                  virtual_hosted_style_request: false
+                  auth:
+                    type: static_credentials
+                    access_key_id: test
+                    secret_access_key: test
               partitioning_strategies:
-                - schema_metadata: [ "_part_id" ]
+                - schema_metadata: ["_part_id"]
               writer_options:
                 flush_when_older_than: 10s
 
         connections:
           - from: receiver
             to: exporter
-


rust/otap-dataflow/crates/otap/src/cloud_auth/aws.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// (env vars, shared config files, instance metadata, etc.).
+    Default,
+
+    /// Static credentials.
+    StaticCredentials {
+        /// AWS access key id.
+        access_key_id: String,
+
+        /// AWS secret access key.
+        secret_access_key: String,
```
JakeDern: One thing that this raises to me is the potential need for an `OpaqueString` or `SecretString` type that has a different `Debug` and `Display` implementations so that we don't potentially leak credentials via logs.

Not sure if we currently have that anywhere yet CC: @lquerel @albertlockett

antonmry:
Thank you for the review. It's still a draft, I would like to test it more but I appreciate the early feedback. I'm looking at options for your comment, it's a good point.

rust/otap-dataflow/crates/otap/src/object_store.rs
```diff
@@ -42,6 +42,31 @@ pub enum StorageType {
         /// The auth settings, see [cloud_auth::azure::AuthMethod]
         auth: cloud_auth::azure::AuthMethod,
     },
+
+    /// AWS S3 storage
+    #[cfg(feature = "aws")]
+    S3 {
+        /// The S3 bucket URI, e.g. `s3://my-bucket/prefix`
+        base_uri: String,
```
albertlockett, line 55: When we construct the AWS object store like this, I think it will actually lose the path from the URI
https://github.com/open-telemetry/otel-arrow/blob/226129d3d43fa8618ea48147fb45a22642a4d83d/rust/otap-dataflow/crates/otap/src/object_store.rs#L128

Indeed, if I run the code in this PR with config:
```yaml
          exporter:
            type: parquet:exporter
            config:
              storage:
                s3:
                  base_uri: s3://lockett-share-ti836471537/telemetry
                  region: ca-central-1
                  auth:
                    # ....
```

I see the telemetry isn't nested under /telemetry
```
$ AWS_PROFILE=albertlockett aws s3 ls s3://lockett-share-ti836471537
                           PRE log_attrs/
                           PRE logs/
                           PRE resource_attrs/
```

It's probably a bigger refactor to fix that - I think we'd need to have our `from_storage_type` function also return an optional path prefix, and we'd have to pass that along to the parquet writer. 

That's probably something we can document and fix later on, but for now I'd just advise that maybe we remove the `/prefix` from this comment. Does that sound OK?

```suggestion
        /// The S3 bucket URI, e.g. `s3://my-bucket`
        base_uri: String,
```

rust/otap-dataflow/configs/fake-parquet-s3.yaml
```diff
@@ -0,0 +1,44 @@
...
+          exporter:
+            type: parquet:exporter
+            config:
+              storage:
+                s3:
+                  base_uri: s3://otel-test-bucket/telemetry
+                  region: us-east-1
+                  endpoint: http://localhost:4566
+                  allow_http: true
+                  virtual_hosted_style_request: false
```
albertlockett: nit: maybe we can document that these options are for local development only:

```suggestion

                  # Optional config options available for development against S3 compatible store like LocalStack:
                  endpoint: http://localhost:4566
                  allow_http: true
                  virtual_hosted_style_request: false

```

rust/otap-dataflow/crates/otap/src/object_store.rs
albertlockett: cc @JakeDern . I'm less familiar w/ azure Object Store implementation, but it might have the same problem

albertlockett (commented):
This is a fantastic contribution @antonmry ! Thanks so much for taking this on.

I had a few small doc-related comments.

When testing this, I imagine you were working testing LocalStack? Now that we have support for this, I'll probably create a future issue for us to run parquet exporter against an AWS compatible store in CI, just so we have better test coverage on this feature

rust/otap-dataflow/crates/otap/src/object_store.rs
JakeDern: I don't have the environment to test it anymore, but I'm pretty sure it worked for the Azure implementation and based on the docs seems like it should (assuming the storage type for your scheme supports a path) : 

https://github.com/apache/arrow-rs-object-store/blob/72cd13608ba9c2c566c609c377bd627281f83501/src/azure/builder.rs#L531C1-L562

rust/otap-dataflow/crates/otap/src/cloud_auth/aws.rs
antonmry: I didn't find anything similar, Azure works differently. I've just committed this one as a potential implementation:  https://github.com/open-telemetry/otel-arrow/pull/2069/changes/038ec8fc65324ba1ce54aa1311ab838cc38fcd8c

antonmry:
@albertlockett thank you for the feedback. You are right, I only tested it with LocalStack, I'm in the process of testing it with a real AWS S3 bucket, I will mark it as Ready for review once it's done.

I didn't find a GH action for Azure but I agree it would be useful to have it. I've some experience with Garage and Rustfs for S3, but AFAICT they are only S3-compatible, not Azure. If you open an issue, let me know and I will try to open a PR.

rust/otap-dataflow/configs/fake-parquet-s3.yaml
antonmry: Good point, solved in https://github.com/open-telemetry/otel-arrow/pull/2069/commits/6c1fa0a079489811dcaf39e62ed2313773e828a4

rust/otap-dataflow/crates/otap/src/object_store.rs
antonmry: I didn't test it with Azure, but based on https://github.com/apache/arrow-rs-object-store/blob/main/src/azure/builder.rs#L686 , I think it has the same behaviour.

I've commited a fix both for Azure and AWS, https://github.com/open-telemetry/otel-arrow/pull/2069/commits/83f7ab62154f1381dd4a8ab6f56e2643ffb4a617 , but it might be better to skip the prefix support in this PR and, if you think it's valuable, create a new one only for that

rust/otap-dataflow/crates/otap/src/cloud_auth/aws.rs
antonmry: I'm closing this conversation, feel free to open it again if there's something I should change on this

antonmry:
Tested with AWS and it worked well, with and without prefix. The PR is Ready for review.  


image

rust/otap-dataflow/crates/otap/Cargo.toml
```diff
@@ -76,6 +76,7 @@ socket2 = { workspace = true }
 ipnet = { workspace = true }
 flate2 = { workspace = true }
 zstd = { workspace = true }
+url = { workspace = true }
```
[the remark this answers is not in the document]
antonmry: url is also used in #[cfg(test)] (the delayed_test_object_store helper), so making it optional would break cargo test without feature flags. It's a tiny crate and already a transitive dep of object_store itself

rust/otap-dataflow/crates/otap/src/cloud_auth/aws.rs
```diff
@@ -0,0 +1,189 @@
...
+        AuthMethod::AssumeRole {
+            role_arn,
+            external_id: _,
+            session_name,
+        } => {
+            let mut configured = builder.with_config(AmazonS3ConfigKey::RoleArn, role_arn);
+            if let Some(value) = session_name {
+                configured = configured.with_config(AmazonS3ConfigKey::RoleSessionName, value);
+            }
+            configured
```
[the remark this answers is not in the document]
antonmry: Already documented in cloud_auth/aws.rs:43-45. There's no ExternalId variant in AmazonS3ConfigKey upstream, we can't wire it through. The field exists for forward-compat. Rejecting Some values with an error would be a breaking config change for no benefit

rust/otap-dataflow/crates/otap/src/object_store.rs
```diff
@@ -42,6 +47,102 @@ pub enum StorageType {
...
+        store: "cloud",
+        source: Box::new(e),
+    })?;
+
+    let path = url.path();
+
+    // For scheme-based URIs (s3://, az://, abfs://), the path starts with /
+    // and the first segment after the leading slash is the prefix.
+    // For HTTPS Azure URIs, the first path segment is the container name,
+    // so the prefix starts from the second segment.
```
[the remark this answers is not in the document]
antonmry: https://github.com/open-telemetry/otel-arrow/pull/2069/commits/ad5f6e2cbca53472735f12f15b14f52cbcd0bcb8

rust/otap-dataflow/crates/otap/src/object_store.rs
```diff
@@ -81,12 +182,42 @@ pub fn from_storage_type(
...
+            let mut builder = AmazonS3Builder::from_env().with_url(base_uri);
+
+            if let Some(region) = region {
+                builder = builder.with_region(region);
+            }
+            if let Some(endpoint) = endpoint {
+                builder = builder.with_endpoint(endpoint);
+            }
+            if let Some(true) = allow_http {
+                builder = builder.with_allow_http(true);
```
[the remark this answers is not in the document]
antonmry: https://github.com/open-telemetry/otel-arrow/pull/2069/commits/ad5f6e2cbca53472735f12f15b14f52cbcd0bcb8

rust/otap-dataflow/crates/otap/src/object_store.rs
JakeDern: Yeah I think you're right, looking at the code it looks like it just parses the container out of the path if applicable. If we're not able to test it with an azure account right now it might be better to file an issue for follow up like you mentioned though

antonmry:
Tested with Azure and it works well, including support for the prefix:

image

Confirmed that the current version doesn't support prefix with the same config:

image

The test setup I used is [here](https://github.com/antonmry/otel-playground/tree/main/otel-arrow).

It's ready for review and merge if there are no more suggestions.

rust/otap-dataflow/crates/otap/src/object_store.rs
antonmry: I tested it with Azure and it worked well, it's documented [here](https://github.com/open-telemetry/otel-arrow/pull/2069/#issuecomment-3933702061).

I'm going to resolve this conversation, but feel free to reopen it if you think we should go with a different approach.

rust/otap-dataflow/crates/otap/src/cloud_auth/aws.rs
```diff
@@ -0,0 +1,189 @@
...
+                configured = configured.with_config(AmazonS3ConfigKey::RoleArn, value);
+            }
+            if let Some(value) = token_file_path {
+                configured = configured.with_config(AmazonS3ConfigKey::WebIdentityTokenFile, value);
+            }
+            configured
+        }
+        AuthMethod::AssumeRole {
+            role_arn,
+            external_id: _,
```
lquerel, line 90: Why external_id is ignored here? Is there a reason to have it in the config but not using it to configure the builder?

rust/otap-dataflow/crates/otap/src/object_store.rs
albertlockett: Awesome - yeah the `PrefixStore` is a great solution for this. thanks @antonmry

rust/otap-dataflow/crates/otap/src/cloud_auth/aws.rs
antonmry: external_id is a standard parameter of the AWS STS ([link](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html)) but object_store's AmazonS3ConfigKey doesn't expose a way to pass it through to the underlying STS call. I added it with a comment to properly model AssumeRole in case it's added in the future but it isn't strictly required and it could be removed

antonmry:
I've rebased the branch to solve a conflict with main and I tested it again with Azure and AWS.

rust/otap-dataflow/crates/otap/src/cloud_auth/aws.rs
antonmry: I'm resolving this conversation. Please, feel free to reopen it if you prefer external_id to be removed. We can always add it later

rust/otap-dataflow/configs/fake-parquet-s3.yaml
```diff
@@ -0,0 +1,39 @@
...
+            access_key_id: test
+            secret_access_key: test
+      partitioning_strategies:
+      - schema_metadata: ["_part_id"]
+      writer_options:
+        flush_when_older_than: 10s
+
+connections:
+  - from: receiver
+    to: exporter
```
albertlockett: @antonmry we made a change to the configuration structure in https://github.com/open-telemetry/otel-arrow/pull/2056 . This config file will need to be updated
```suggestion
version: otel_dataflow/v1
engine: { }
groups:
  default:
    pipelines:
      main:
        policies:
          channel_capacity:
            control:
              node: 100
              pipeline: 100
            pdata: 128
        nodes:
        receiver:
            type: traffic_generator:receiver
            config:
            traffic_config:
                max_batch_size: 100
                signals_per_second: 10
                log_weight: 100
            registry_path: https://github.com/open-telemetry/semantic-conventions.git[model]
        exporter:
            type: parquet:exporter
            config:
            storage:
                s3:
                base_uri: s3://otel-test-bucket/telemetry
                region: us-east-1
        
                # Optional config options available for development against S3
                # compatible store like LocalStack:
                endpoint: http://localhost:4566
                allow_http: true
                virtual_hosted_style_request: false
                auth:
                    type: static_credentials
                    access_key_id: test
                    secret_access_key: test
            partitioning_strategies:
            - schema_metadata: ["_part_id"]
            writer_options:
                flush_when_older_than: 10s

        connections:
        - from: receiver
            to: exporter
```

This should fix the test failure that we're seeing in CI
https://github.com/open-telemetry/otel-arrow/actions/runs/22300219228/job/64527141863?pr=2069

```
        FAIL [   0.025s] (  49/2172) otap-df-config engine::tests::bundled_configs_parse_as_engine_configs
  stdout ───

    running 1 test
    test engine::tests::bundled_configs_parse_as_engine_configs ... FAILED

    failures:

    failures:
        engine::tests::bundled_configs_parse_as_engine_configs

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 110 filtered out; finished in 0.01s
    
  stderr ───

    thread 'engine::tests::bundled_configs_parse_as_engine_configs' (107825) panicked at crates/config/src/engine.rs:989:17:
    failed to parse engine config /Users/runner/work/otel-arrow/otel-arrow/rust/otap-dataflow/crates/config/../../configs/fake-parquet-s3.yaml: Err(DeserializationError { context: Context { pipeline_group_id: None, pipeline_id: None }, format: "YAML", details: "unknown field `settings`, expected one of `version`, `policies`, `topics`, `engine`, `groups`" })
```

antonmry: Oh, it seems an issue during the rebase. I've fixed it with https://github.com/open-telemetry/otel-arrow/pull/2069/changes/88ec4f29af15bf285eb811ccfd1777b37c94928d 

I have also removed the comment in fake-parquet.yaml and added files for S3 (local and AWS) and Azure. It should be more clear in this way.

Issue #1145, opened 2025-09-17
OTLP/HTTP Exporter


lquerel:
See Go implementation for reference.

albertlockett:
closing, this work is completed

PR #2070, opened 2026-02-19
Initial implementation of OTLP HTTP Exporter

albertlockett:
# Change Summary

This PR adds a new exporter for exporting telemetry via OTLP over HTTP.

Many of the same patterns from the OTLP gRPC exporter are followed, including the optimizations that were added in https://github.com/open-telemetry/otel-arrow/pull/1474.

This PR adds a new `otlp_http::client_settings` to the otap crate, which contains `HttpClientSettings`, a reusable configuration for an HTTP client, which can produce a configured `reqwest::ClientBuilder`. This is analogous to the `otlp_grpc::client_settings::GrpcClientSettings`.

The exporter uses the `otlp_exporter:: InFlightExports` type for managing a limited number of concurrent in flight requests and, and handling request completions.

The exporter uses a pool of `request::Clients`, implemented as the `HttpClientPool` type. The intention is to force requests to be distributed over multiple connections to the OTLP server. In our cast, there may be many servers listening on the same port using SOREUSEPORT (for example, one per instance of a pipeline, each on different cores). The intention is to hopefully have better load balancing. When using HTTP version 1, this is likely unnecessary as with a high enough concurrent request volume, multiple connections will be opened anyway by the internal reqwest Client connection pool. However, this will likely be needed when we add TLS because HTTP version 2 may be negotiated, in which case multiple requests can be multiplexed over the same connection, leading to the same issue we had in the OTLP gRPC exporter (which also uses a client pool).

There are a variety of additional tasks that I will handle in followup PRs:
- [ ] TLS/mTLS support
- [ ] Proxy support
- [ ] Compression - compressed request bodies / accepting compressed responses
- [ ] Allow endpoint overrides for each signal type (similar to Go collector implementation)
- [ ] Unit test metrics produced by component

* Relates to #1145 

A suite of unit tests covering successful requests both OTLP & OTAP pdatas, plus tests covering various errors including connection refused, non 200 response, OTLP service responses w/ partial success and malformed OTAP batches.

Users now have a new component available for their telemetry pipelines.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
@@ -1027,7 +1027,6 @@ mod test {
                         }
                     }
 
-                    // validate we received three Acks
                     let mut ack_count = 0;
                     let mut pipeline_ctrl_rx = ctx.take_pipeline_ctrl_receiver().unwrap();
                     loop {
@@ -1198,7 +1197,6 @@ mod test {
                     // ensure exit success
                     result.unwrap();
 
-                    // validate we received three Nacks
                     let mut ack_count = 0;
                     let num_expected_nacks = 1;
                     let mut pipeline_ctrl_rx = ctx.take_pipeline_ctrl_receiver().unwrap();
@@ -1301,7 +1299,6 @@ mod test {
                     // ensure exit success
                     result.unwrap();
 
-                    // validate we received three Nacks
                     let mut ack_count = 0;
                     let num_expected_nacks = 3;
                     let mut pipeline_ctrl_rx = ctx.take_pipeline_ctrl_receiver().unwrap();
@@ -1395,7 +1392,6 @@ mod test {
                     // ensure exit success
                     result.unwrap();
 
-                    // validate we received three Nacks
                     let mut ack_count = 0;
                     let num_expected_nacks = 1;
                     let mut pipeline_ctrl_rx = ctx.take_pipeline_ctrl_receiver().unwrap();
@@ -1494,7 +1490,6 @@ mod test {
                     // ensure exit success
                     result.unwrap();
 
-                    // validate we received three Nacks
                     let mut ack_count = 0;
                     let num_expected_nacks = 1;
                     let mut pipeline_ctrl_rx = ctx.take_pipeline_ctrl_receiver().unwrap();
@@ -1583,7 +1578,6 @@ mod test {
 
                     server_cancellation_token.cancel();
 
-                    // validate we received three Nacks
                     let mut ack_count = 0;
                     let num_expected_nacks = 1;
                     let mut pipeline_ctrl_rx = ctx.take_pipeline_ctrl_receiver().unwrap();
@@ -1680,7 +1674,6 @@ mod test {
 
                     server_cancellation_token.cancel();
 
-                    // validate we received three Nacks
                     let mut ack_count = 0;
                     let num_expected_nacks = 1;
                     let mut pipeline_ctrl_rx = ctx.take_pipeline_ctrl_receiver().unwrap();
@@ -1821,7 +1814,6 @@ mod test {
                         }
                     }
 
-                    // validate we received three Acks
                     let mut ack_count = 0;
                     let mut pipeline_ctrl_rx = ctx.take_pipeline_ctrl_receiver().unwrap();
                     loop {


albertlockett:
Keeping this as Draft for now, since it seems to be breaking the pipeline perf tests

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
```diff
@@ -0,0 +1,1590 @@
...
+use crate::otlp_exporter::InFlightExports;
+use crate::otlp_http::client_settings::HttpClientSettings;
+use crate::otlp_http::{LOGS_PATH, METRICS_PATH, PROTOBUF_CONTENT_TYPE, TRACES_PATH};
+use crate::otlp_http_exporter::config::Config;
+use crate::pdata::{Context, OtapPdata};
+
+mod config;
+
+/// The URN for the OTLP HTTP exporter
+pub const OTLP_HTTP_EXPORTER_URN: &str = "urn:otel:otlphttp:exporter";
```
lquerel: Should we use `urn:otel:otlp_http:exporter` for clarity?

I also suggest we add to the agenda of the next SIG meeting the question of potentially renaming the OTLP/gRPC exporter.

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
```diff
@@ -0,0 +1,1590 @@
...
+
+    let pdata = OtapPdata::new(context, saved_payload);
+
+    let err_msg = match result {
+        Ok(service_resp) => service_resp.partial_success.map(|partial_success| {
+            format!(
+                "{} ({} rejected)",
+                partial_success.error_message, partial_success.rejected
+            )
+        }),
```
lquerel: This might be the right approach but I'm not sure. 
I'm wondering if a partial success should alway result into a Nack (as in this impl). Do you know what the Go collector do in this case?

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
```diff
@@ -0,0 +1,1590 @@
...
+        Some(src) => format!(": {src}"),
+        None => String::new(),
+    }
+}
+
+async fn query_result_to_service_response(
+    signal_type: &SignalType,
+    result: Result<Response, reqwest::Error>,
+) -> Result<ServiceResponse, ServiceRequestError> {
+    let mut bytes = result?.error_for_status()?.bytes().await?;
```
lquerel: For security reason, we probably should add a strict maximum response size in order to prevent malicious or misbehaving server to return a BIG answer and so trigger memory exhaustion.

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
```diff
@@ -0,0 +1,1590 @@
...
+                                        .encode(&mut otap_batch, &mut proto_buffer),
+                                };
+                            let body = if let Err(e) = encode_result {
+                                // encoding error, we must have received an invalid structured batch
+                                _ = effect_handler
+                                    .notify_nack(NackMsg::new(
+                                        e.to_string(),
+                                        OtapPdata::new(context, otap_batch.into()),
+                                    ))
+                                    .await;
```
lquerel: We should check the context.may_return_payload() method to determine if the full otap_batch payload must be included into the Nack message.

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
```diff
@@ -0,0 +1,1590 @@
...
+                    let (context, payload) = pdata.into_parts();
+
+                    // proto encode the payload into the request body, while keeping a copy of the
+                    // original payload if the context allows it to be returned
+                    let (body, saved_payload) = match payload {
+                        OtapPayload::OtlpBytes(mut otlp_bytes) => {
+                            if context.may_return_payload() {
+                                // copy the bytes to the request body
+                                let body = Bytes::copy_from_slice(otlp_bytes.as_bytes());
+                                (body, otlp_bytes.into())
```
lquerel: We could avoid this copy by adding the following method in OtlpProtoBytes

```rust
/// Get a cloned reference-counted view of the serialized proto bytes.
#[must_use]
pub fn clone_bytes(&self) -> Bytes {
      match self {
          OtlpProtoBytes::ExportLogsRequest(bytes)
          | OtlpProtoBytes::ExportMetricsRequest(bytes)
          | OtlpProtoBytes::ExportTracesRequest(bytes) => bytes.clone(),
      }
}
```

and then replace `let body = Bytes::copy_from_slice(otlp_bytes.as_bytes());` with `let body = otlp_bytes.clone_bytes();`

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
```diff
@@ -0,0 +1,1590 @@
...
+                                _ = effect_handler
+                                    .notify_nack(NackMsg::new(
+                                        e.to_string(),
+                                        OtapPdata::new(context, otap_batch.into()),
+                                    ))
+                                    .await;
+                                self.pdata_metrics.add_failed(signal_type, 1);
+                                continue;
+                            } else {
+                                Bytes::copy_from_slice(proto_buffer.as_ref())
```
lquerel, line 287: The same apply here I think

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
albertlockett: I agree - that keeps things consistent with go colllector as well.

Fixed in [7620115](https://github.com/open-telemetry/otel-arrow/pull/2070/commits/76201158c98e597eb906773b347c142836e3885d)

albertlockett: Added to next week's SIG agenda https://docs.google.com/document/d/1z8_Ra-ALDaYNa88mMj1gOZtOpLZLRk0-dZEmDjPmcUs/edit?tab=t.0

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
albertlockett: Good idea, fixed in [c36df31](https://github.com/open-telemetry/otel-arrow/pull/2070/commits/c36df31924f7d8b3ca860c6c2d96966201ec06be)

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
albertlockett: Thanks, fixed in 1a7db4b65

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
lalitb: As per the [specs](https://opentelemetry.io/docs/specs/otlp/#partial-success) -

>If the request is only partially accepted (i.e. when the server accepts only parts of the data and rejects the rest), the server response MUST be the same [Export<signal>ServiceResponse](https://github.com/open-telemetry/opentelemetry-proto/tree/v1.9.0/opentelemetry/proto/collector) message as in the [Full Success](https://opentelemetry.io/docs/specs/otlp/#full-success) case.

Sending a `Nack` here will make the upstream retry the full batch.

rust/otap-dataflow/crates/otap/Cargo.toml
```diff
@@ -67,6 +67,7 @@ weaver_semconv.workspace = true
 weaver_resolved_schema.workspace = true
 weaver_resolver.workspace = true
 rand.workspace = true
+reqwest = { workspace = true, features = ["rustls", "stream"] }
```
lalitb, line 70: This was optional earlier. But I believe the OTLP/HTTP exporter is core component, so this has to be mandatory. Not an issue, just reiterating that this would be added dependency now.

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
```diff
@@ -0,0 +1,1731 @@
...
+                Message::Control(NodeControlMsg::Shutdown { deadline, reason }) => {
+                    otel_info!("otlp.exporter.http.shutdown", reason = reason);
+                    while !inflight_exports.is_empty() {
+                        if let Some(completed) = inflight_exports.next_completion().await {
+                            finalize_completed_export(
+                                completed,
+                                &effect_handler,
+                                &mut self.pdata_metrics,
+                            )
+                            .await;
```
lalitb, line 230: The drain loop ignores the `deadline` already in scope. With `http.timeout = None` (default), a slow/hung server will block shutdown indefinitely past the deadline                                                                                               Worth racing `next_completion()` against `sleep_until(deadline)` here?

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
```diff
@@ -0,0 +1,1731 @@
...
+    fn try_new(
+        client_settings: &HttpClientSettings,
+        pool_size: NonZeroUsize,
+    ) -> Result<Self, reqwest::Error> {
+        let mut default_headers = HeaderMap::new();
+
+        // TODO eventually this header value should be dynamic once we support JSON OTLP payloads
+        _ = default_headers.insert(
+            http::header::CONTENT_TYPE,
+            HeaderValue::from_static(PROTOBUF_CONTENT_TYPE),
```
lalitb, line 587: should we also add `Accept: application/x-protobuf` here?

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/config.rs
```diff
@@ -0,0 +1,42 @@
...
+pub struct Config {
+    /// Configuration for the HTTP client that will be used by this exporter
+    pub http: HttpClientSettings,
+
+    /// The endpoint to which the exporter will send OTLP HTTP requests. This should include the
+    /// scheme, host and port, but not the paths (/v1/logs) as these will be appended to requests
+    /// automatically for each batch of signals depending on the signal type.
+    ///
+    /// Example: "http://localhost:4318" or "https://otel-collector:4318"
+    pub endpoint: String,
```
lalitb, line 22: nit - should we have a valuation for `endpoint`, as part of `OtlpHttpExporter::from_config`, something like:

```rust
  reqwest::Url::parse(&config.endpoint).map_err(|e| {
      ConfigError::InvalidUserConfig { error: format!("invalid endpoint URL: {e}") }
  })?;
```

rust/otap-dataflow/crates/pdata/src/otlp/mod.rs
```diff
@@ -79,6 +79,29 @@ impl OtlpProtoBytes {
         }
     }
 
+    /// Replaces the internal bytes with new bytes and returns the old bytes.
+    pub fn replace_bytes<B>(&mut self, b: B) -> Bytes
+    where
+        B: Into<Vec<u8>>,
```
lalitb: nit - This abstraction can be removed? Since both sides are `Bytes`, taking `Bytes` directly with `std::mem::replace` would be cleaner and avoids an unnecessary copy for any future non-empty caller.

lalitb:
Nice work. High level looks good! Left a few inline comments. I believe worth addressing before merge would be the partial success handling for OTEL specs compliance.

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
albertlockett: Oops yes you're right - we shouldn't send the Nack here because we don't want to retry. the spec also says:
https://opentelemetry.io/docs/specs/otlp/#partial-success-1

> The client MUST NOT retry the request when it receives a partial success response where the partial_success 

Looks like there are also semantics for when to retry:

> The server SHOULD use HTTP response status codes to indicate retryable and not-retryable errors for a particular erroneous situation. The client SHOULD honour HTTP response status codes as retryable or not-retryable.
> 
> **Retryable Response Codes**
> The requests that receive a response status code listed in following table SHOULD be retried. All other 4xx or 5xx response status codes MUST NOT be retried.
> 
> **HTTP response status code**
> 429 Too Many Requests
> 502 Bad Gateway
> 503 Service Unavailable
> 504 Gateway Timeout

I'll fix this

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
albertlockett: :+1: fixed in 03166d933

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
albertlockett: I don't think we can do something similar to what is suggest [here](https://github.com/open-telemetry/otel-arrow/pull/2070#discussion_r2831002974).

Internally this proto_buffer has a Vec<u8>, not `Bytes`.
https://github.com/open-telemetry/otel-arrow/blob/3d35559b3b13c8f3082c1baf5bb0c90ab9dc418c/rust/otap-dataflow/crates/pdata/src/otlp/common.rs#L330-L333

We also clear and reuse this instance for each OTAP batch that we encode to proto.

rust/otap-dataflow/crates/otap/Cargo.toml
albertlockett: Indeed - no longer optional

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
albertlockett: Sure, it can't hurt. Added in 185a2e0b8

rust/otap-dataflow/crates/pdata/src/otlp/mod.rs
albertlockett: Sounds good. I was originally trying to be consistent with the constructor, but you're right it's better to avoid the copy.

Fixed in 348f9f399

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/config.rs
albertlockett: Seems reasonable - fixed in bd8459dd3

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
albertlockett: @lalitb & @lquerel I've updated the way we handle errors/produce Nacks in 3bb9a33cc

The behaviour is we receive a 200 response code w/ a partial request we emit a Nack, _unless_ rejected signals == 0. This Nack will have `permanent = true`, meaning the payload should not be retried. 

When partial success rejects 0 signals, we treat this as a success. According to the spec, the server send a partial_success even if it fully accepted the request:
> servers MAY also use the partial_success field to convey warnings/suggestions to clients even when it fully accepts the request. In such cases, the rejected_<signal> field MUST have a value of 0, and the error_message field MUST be non-empty.

In the case where we receive a 200 response, but couldn't decode the body, we also send a Nack with `permanent=true`. The assumption here is that either there was a success, or a partial success, but either way we're not supposed to retry the request.

When we receive a non-200 response, we now follow the suggestions in the spec vis-à-vis retries: we send a Nack, and IF the response status wasn't 429, 502, 503 or 504, we set `permanent=true` meaning don't retry it.

For non-http errors, the spec only gives some limited clarity about what to do:
https://opentelemetry.io/docs/specs/otlp/#all-other-responses
>All other HTTP responses that are not explicitly listed in this document should be treated according to HTTP specifications.

> If the server disconnects without returning a response, the client SHOULD retry and send the same request. The client SHOULD implement an exponential backoff strategy between retries to avoid overwhelming the server.

The behaviour I have for these is to retry (e.g. send Nack w/ `permanent=false`) if the error was a connection error or a TCP timeout. My thinking was that it's reasonable to retry in these cases which may be intermittent networking errors.

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
```diff
@@ -0,0 +1,1862 @@
...
+                    let mut pipeline_ctrl_rx = ctx.take_pipeline_ctrl_receiver().unwrap();
+                    loop {
+                        let msg = match pipeline_ctrl_rx.recv().await {
+                            Ok(msg) => msg,
+                            Err(_) => break, // channel closed, no more messages will be received
+                        };
+
+                        match msg {
+                            PipelineControlMsg::DeliverAck { node_id, .. } => {
+                                assert_eq!(node_id, node_id);
```
utpilla: Remove this assertion. There are a few more places where we have this.

albertlockett: ok. [4ad24a4](https://github.com/open-telemetry/otel-arrow/pull/2070/commits/4ad24a42ed400bb502e2bb39ccd654693dd346ac)

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
```diff
@@ -0,0 +1,1854 @@
...
+async fn collect_body(response: Response, max_len: usize) -> Result<Bytes, ServiceRequestError> {
+    let mut stream = response.bytes_stream();
+
+    let size_hint = stream.size_hint();
+    if let Some(upper) = size_hint.1 {
+        return Err(ServiceRequestError::BodyTooLarge {
+            body_size: upper,
+            max_size: max_len,
+        });
+    }
```
utpilla: The `size_hint()` pre-check seems a bit odd:

- It returns an error whenever an upper bound exists, without comparing it to `max_len`. Even `Some(1)` would be rejected.
- Also, `Stream::size_hint()` counts remaining chunks, not bytes, so comparing it against a byte limit (`max_len`) would also not be meaningful.

The streaming loop below already does correct byte-level enforcement, so I think this pre-check can simply be removed. Alternatively, if we want to keep an early-out for Content-Length, but that will need to come from the response headers rather than the stream's chunk hint.

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
```diff
@@ -0,0 +1,90 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Shared configuration for gRPC-based clients.
```
utpilla:
```suggestion
//! Shared configuration for HTTP-based clients.
```

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
```diff
@@ -0,0 +1,1854 @@
...
+                })
+            })
+            .run_validation(|mut ctx, result| {
+                Box::pin(async move {
+                    // ensure exit success
+                    result.unwrap();
+
+                    server_cancellation_token.cancel();
+
+                    // validate we received three Nacks
```
utpilla: This looks like a copy-paste error. There is only one test `test_handles_partial_success` which expects three NACKs. Rest of the tests which have this comment only expect 1 NACK.

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
albertlockett: great catch, thanks @utpilla

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
```diff
@@ -0,0 +1,1852 @@
...
+                        .await
+                        .unwrap();
+                })
+            })
+            .run_validation(|mut ctx, result| {
+                Box::pin(async move {
+                    // ensure exit success
+                    result.unwrap();
+
+                    // validate we received three Nacks
```
utpilla: This comment can be removed/updated for these tests as well:
- `test_handles_connection_refused_errors`
- `test_handles_response_body_too_large`
- `test_handles_invalid_otap_payloads`
- `test_nacks_for_otap_payloads_when_context_indicates_no_payload_return`
- `test_nacks_for_otlp_payloads_when_context_indicates_no_payload_return`

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
albertlockett: I created https://github.com/open-telemetry/otel-arrow/issues/2099 to track this. OTLP gRPC exporter has the same behaviour

PR #2074, opened 2026-02-19
[recordset_kql_processor] Logging tweaks

CodeBlanch:
Relates to #2066

# Changes

* Switch to logging each diagnostic returned from engine individually (instead of logging a single formatted version).
* Tweaks to the other logs in the processor for consistency.


rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs
```diff
@@ -162,8 +170,8 @@ impl RecordsetKqlProcessor {
 
                 otap_df_telemetry::otel_debug!(
                     "recordset_kql_processor.success",
-                    input_items = input_items,
-                    output_items = output_items,
+                    input_items,
```
cijothomas, line 173: nice simplification!

PR #2079, opened 2026-02-20
feat: Split implementation

### LINKED ISSUE #1334: otel-arrow-rust batching logic panic

JakeDern:
# Change Summary

This is the initial implementation of the third major item batching operation, `split`, and is the last part of Phase 1 of the batch processor work. This also closes #1334 by enabling all of the remaining batching tests.

Major pieces of work:

- Split implementation
- Pulled out shared utilities from the `transport_optimize` and `transform` modules into a `utils` module as they're shared across the batching related transforms as well.
- Pulled out shared testing utilities from the `reindex` module into a dedicated `testing` module as the `split` module also leverages these heavily 
- Fixed a bug in reindexing when redacting rows with dictionary arrays
- Remove the old groups.rs implementation

* Part of #1926 
* Closes #1334 
 
A large unit testing suite was added, and the last disabled tests were enabled in the comprehensive batching tests.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
@@ -20,7 +20,7 @@ use crate::otap::transform::util::{
 };
 use crate::otap::{Logs, Metrics, OtapBatchStore, Traces};
 use crate::proto::opentelemetry::arrow::v1::ArrowPayloadType;
-use crate::schema::consts::PARENT_ID;
+use crate::schema::consts::{ID, PARENT_ID};
 
 use super::util::{PrimaryIdInfo, payload_relations};
 
@@ -629,7 +629,7 @@ where
         }
         _ => {
             return Err(Error::ColumnDataTypeMismatch {
-                name: "id".to_string(),
+                name: ID.to_string(),
                 expect: T::DATA_TYPE,
                 actual: array.data_type().clone(),
             });


rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
```diff
@@ -0,0 +1,1394 @@
...
+        let key_col = relation.key_col;
+
+        for &child_type in relation.child_types {
+            let child_idx = POSITION_LOOKUP[child_type as usize];
+
+            let Some(child_rb) = &source_batch[child_idx] else {
+                continue;
+            };
+
+            let key_ranges = get_contiguous_id_ranges(parent_slice, key_col)?;
```
utpilla: It looks like this is only dependent on `parent_slice` and `key_col` and not child data. We could move this outside of the `for &child_type in relation.child_types` to only compute it once.

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
```diff
@@ -0,0 +1,1394 @@
...
+    let n = dict.len();
+
+    // Dereference each row's dictionary key to its logical value.
+    let logical_val = |i: usize| -> u32 {
+        let key_idx = dict.key(i).expect("non-null parent_id");
+        vals.value(key_idx).into()
+    };
+
+    let start = row_partition_point(n, |i| logical_val(i) < *key_range.start());
+    let end = row_partition_point(n, |i| logical_val(i) <= *key_range.end());
```
utpilla, line 717: nit: Consider a merge scan in `plan_metrics_split_impl`

`find_rows_by_parent_id_range` does two binary searches per call. In `plan_metrics_split_impl`, it's called once per metric row against sorted data-point tables. Both sides are ascending (root sorted by `sort_otap_batch_by_parent_then_id`, child by `parent_id`). A single merge-scan cursor would be `O(N + M)` instead of `O(N log M)`, and should also be more cache-friendly since it accesses data sequentially, unlike binary search, which jumps around.

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
```diff
@@ -0,0 +1,1394 @@
...
+///   +-----------+----+
+///   |         0 |  1 |
+///   |         0 |  2 |
+///   |         1 |  0 |  <-- id restarts here
+///   |         1 |  3 |
+///   +-----------+----+
+/// ```
+///
+/// In this case we'll initially produce two ranges, one for `[1, 2]` and one
+/// for `[0, 3]`, which are overlapping. We need to reduce these to a single range
```
utpilla: We should be producing three ranges, right? `[1..=2, 0..=0, 3..=3]` which would then be reduced and merge to `[0,3]`?

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
```diff
@@ -0,0 +1,1394 @@
...
+///
+/// 1. Start by creating windows for chunks of contiguous values.
+/// 2. Verify that the chunks are contiguous and non-overlapping, if not, proceed
+///    to the next step.
+/// 3. Reduce and merge the ranges via [`reduce_ranges`]
+///
+/// Note: id columns are never dictionary-encoded; this dispatches on plain
+/// UInt16/UInt32 only.
+///
+/// See [`test_split_traces_non_unique_child_ids`] for the exact scenario
```
utpilla: This test isn't there,

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
```diff
@@ -458,6 +456,123 @@ fn remove_vec_ranges<T>(vec: &mut Vec<T>, ranges: &[Range<usize>]) {
...
+    debug_assert!(
+        value_redactions.windows(2).all(|w| w[0].end <= w[1].start),
+        "value_redactions must be sorted and non-overlapping"
+    );
+
+    if value_redactions.is_empty() {
+        return Vec::new();
+    }
+
+    let dict = id_col.as_dictionary::<K>();
```
albertlockett, line 520: nit: this will panic if the passed id_col cannot be converted into a dict array. Maybe we could do one of:
- change the signature to accept &DictionaryArray<K> and do the conversion before calling
- add a comment about why this is safe

wdyt?

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
```diff
@@ -458,6 +456,123 @@ fn remove_vec_ranges<T>(vec: &mut Vec<T>, ranges: &[Range<usize>]) {
...
+
+    // Keys are indices into the values array — directly comparable to the
+    // redaction ranges which are also value-array indices. We keep the keys
+    // in their native type (u8 or u16) and cast the range bounds to match.
+    let mut sorted_keys: Vec<K::Native> = dict.keys().values().to_vec();
+    sorted_keys.sort_unstable();
```
albertlockett, line 526: I wonder if performance of this could be improved if we kept Roaring bitmap of keys found/not found in a values_redaction range, and used this to compute the ranges? That way we wouldn't have to sort the keys

We'd probably want to benchmark this first to make sure this is actually worth doing, e.g. measure whether sorting is actually a bottleneck.

```rs
let redacted = RoaringBitmap::new();
let not_redacted = RoaringBitmap::new();
for (key_idx, key) in dict.keys().values() {
    if not_redacted.contains(key as u32) {
            if let Some(start) = current_start.take() {
                key_redactions.push(start..key_idx);
            }
            continue
    }
    
    if redacted.contains(key as u32) {
            if current_start.is_none() {
                current_start = Some(key_idx);
            }
            continue
    }
    
    let is_found = { // binary search the values redaction ranges };
    if found {
        redacted.insert(key);
    } else {
        not_redacted.insert(key);
    }
}
```

albertlockett: By doing this, we'd get the removed key ranges in the order of the original record batch, and we could also avoid re-sorting the original record batch in the case we detect violations here:
https://github.com/open-telemetry/otel-arrow/blob/674e6aa355e44b36b3ea608eccb042407ff1c782/rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs#L424-L426

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
```diff
@@ -0,0 +1,1394 @@
...
+/// matching rows.
+///
+/// Dispatches on the id column's native type (UInt16/UInt32), following the
+/// same pattern as [`get_contiguous_id_ranges`].
+fn plan_metrics_split(
+    root_rb: &RecordBatch,
+    batch: &[Option<RecordBatch>],
+    max_items: NonZeroU32,
+) -> Result<Vec<Range<usize>>> {
+    let id_col = extract_id_column(root_rb, "id")?;
```
albertlockett: nit: we have consts module with the column names `otap_df_pdata::schema::consts::ID` that can be used here

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
```diff
@@ -0,0 +1,1394 @@
...
+        .values();
+
+    let metric_types =
+        root_rb
+            .column_by_name(METRIC_TYPE)
+            .ok_or_else(|| Error::ColumnNotFound {
+                name: METRIC_TYPE.to_string(),
+            })?;
+
+    // safety: Assuming this property will be uphelp in the future by OtapBatchStore
```
albertlockett:
```suggestion
    // safety: Assuming this property will be upheld in the future by OtapBatchStore
```

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
```diff
@@ -0,0 +1,1394 @@
...
+    // with the downstream concatenation step. Without this, trailing chunks
+    // from one batch may be too small to meet the batch processor's
+    // lower_limit when concatenated.
+    let mut total_items: usize = 0;
+    #[allow(clippy::needless_range_loop)]
+    for i in 0..original_len {
+        // TODO: Can we make this impossible by OtapBatchStore guarantees?
+        if batches[i][root_idx].is_none() {
+            continue;
+        };
```
albertlockett, line 84: That's an interesting idea to try to guarantee this. Currently if the root batch is None, it basically means an empty batch. Some components do emit this - for example a FilterProcessor or TransformProcessor that filters out all the rows.

The reason this happens currently is, something upstream might have been waiting for an Ack on the batch and something downstream might be what produces that Ack. So these components that don't want to handle Ack/Nack logic just emit the empty batch and defer the Ack/Nack handling to something downstream.

rust/otap-dataflow/crates/pdata/src/otap/transform/util.rs
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    Ok(())
+}
+
+// TODO [JD]: Optimize this because doing a column sort with dictionary encoded
+// causes complete expansion of the dictionary. We can probably do this faster
+// for a two column sort. transport_optimize has some optimizations for this
+// and we can steal some techniques.
+pub(crate) fn sort_by_parent_then_id(rb: RecordBatch) -> Result<RecordBatch> {
```
albertlockett, line 128: :+1: Yeah `super::sort_to_indices` will use the row converter.

Something like this might be faster:
https://github.com/open-telemetry/otel-arrow/blob/674e6aa355e44b36b3ea608eccb042407ff1c782/rust/otap-dataflow/crates/pdata/src/otap/transform/transport_optimize/attributes.rs#L578-L617

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
JakeDern: Yeah this is a good point, my brain was in binary search mode :D. Added a better implementation with a merge scan

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
JakeDern: Ah yeah I renamed it to `test_split_overlapping_parent_ranges` and since it's test code it looks like it doesn't automatically update doc refs. I just removed the comment since it doesn't seem to link nicely to test code.

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
JakeDern: Yes, you are right 🙂 - I updated it and for fun wrote a test that is this exact case

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: Yeah good point, I forgot this is a panicking variant. I added a safety comment explaining the caller is meant to check.

Normally I'd rather make the function take &DictionaryArray, but arrow casting code ends up really verbose when you do it a bunch of times on the caller side 😅. 

 I keep ending up with this pattern of checking types in the caller and dispatching to some generic function that just does it. Maybe this is a case where some helper macros could help improve the readability. I'd also be happy to change it the other way though!

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
JakeDern: removed that one already, but got one more in the reindex implementation!

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
JakeDern: Having components possibly emit empty record batches after Filtering or Transform makes sense. 

I don't quite follow the Ack/Nack stuff, but I'm not very familiar with that system yet!

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: Yeah it's a good idea! I'll track it on #1926 and when I have the benchmarks running come back around to it

rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs
JakeDern: Feel free to reopen if you'd rather see the parameter type changed

rust/otap-dataflow/crates/pdata/src/otap/transform/util.rs
JakeDern: I'm also tracking this on #1926! Will be very satisfying to take these items once decent benchmarks are in place 🙂

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
jmacd: I agree, we should feel free to drop payloads that have accessory tables and no primary table in any signal https://github.com/open-telemetry/opentelemetry-proto/pull/622

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
```diff
@@ -0,0 +1,1507 @@
...
+    offset: usize,
+) -> Result<Vec<Range<usize>>> {
+    let n_rows = root_rb.num_rows();
+    let max = max_items.get() as usize;
+
+    // Size the first chunk to fill the remainder of the current
+    // concatenation bin. This ensures that when `generic_concatenate`
+    // later processes the split output, the trailing chunk from the
+    // previous batch plus this first chunk fit exactly into `max_items`.
+    // This may not be a behavior that's important in the future.
```
jmacd, line 125:
```suggestion
    // This behavior is required for `batch_processor` to work correctly.
```

jmacd (approved):

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
JakeDern: I think there is some discussion to be had about the tradeoff between producing batches of the exact minimum/maximum allowable size and the amount of work needed to process the records. I plan to file an issue about defining batch processor behavior, so stay tuned!

jmacd: I see. The batch_processor just needs a way to ensure that pending items held back for being under-size are not held-back again and again. This is implemented by requiring that the first input request is always emitted completely by the first output request.

albertlockett (approved):

rust/otap-dataflow/crates/pdata/src/otap/transform/split.rs
JakeDern: Couple of related issues - #2115, #2116, #2117

Issue #2081, opened 2026-02-20
Terminal state metrics are not collected successfully by `exporter::TestRuntime`


albertlockett:
Here in the exporter `TestRuntime`, we create a task that runs the exporter to completion, then tries to collect the final metrics shapshot when it exits:
https://github.com/open-telemetry/otel-arrow/blob/fd92a9e19d4b95d0ad022c0132f4fb47dda71241/rust/otap-dataflow/crates/engine/src/testing/exporter.rs#L263-L267

This gets awaited on in the `ValidationPhase` here:
https://github.com/open-telemetry/otel-arrow/blob/fd92a9e19d4b95d0ad022c0132f4fb47dda71241/rust/otap-dataflow/crates/engine/src/testing/exporter.rs#L346-L349

When this gets called, the metrics reporter will try to send the telemetry to the channel:
https://github.com/open-telemetry/otel-arrow/blob/fd92a9e19d4b95d0ad022c0132f4fb47dda71241/rust/otap-dataflow/crates/telemetry/src/reporter.rs#L64-L70

One issue with this is that receiver side of the channel is already dropped. This would have been owned by the `InternelCollector` which is was an `Arc` in the `InternalTelemetrySystem`, which was owned by the `TestRuntime`, which gets dropped in the call to `set_exporter`.

The second issue here is that, even if if the collector side of the telemetry channel hadn't been dropped, the collection loop is never started by the `TestRuntime`.

This becomes a problem for test writers who want to test their telemetry component and see what telemetry it returns when it's finished.
```rs
#[test]
fn test_my_exporter() {
    let test_runtime = TestRuntime::<OtapPdata>::new();
    let telemetry_registry_handle = test_runtime.metrics_registry();
    let exporter = setup_exporter_wrapper();
    test_runtime
        .set_exporter(exporter)
        .run_test(scenario())
        .run_validation(|ctx, result| {
            // currently not working ...
            telemetry_registry_handle.visit_current_metrics(|desc, attrs, iter| {
                println!("desc = {:?}", desc);
                for v in iter {
                    println!("v = {:?}", v)
                }
            });
        })
}

PR #2082, opened 2026-02-20
collect terminal metrics snapshot in exporter TestRuntime

albertlockett:
# Change Summary

Collects the telemetry returned by the component in the exporter `TestRuntime`.

A new function is added to manually invoke the telemetry loop one time on `InternalCollector`. This is called if the exporter returned a terminal state with some metrics snapshots.

* Closes https://github.com/open-telemetry/otel-arrow/issues/2081

I tested these manually in a unit test I had on a feature branch:
https://github.com/open-telemetry/otel-arrow/commit/2722031d02d56477fd09ad05d78a543417172a21

More test coverage will be added later on: I plan to use this in a followup to PR https://github.com/open-telemetry/otel-arrow/pull/2070 to add a comprehensive suite of unit tests to the metrics reported by the OTLP HTTP exporter

 No

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/testing/exporter.rs
+++ b/rust/otap-dataflow/crates/engine/src/testing/exporter.rs
@@ -264,8 +264,8 @@ impl<PData: Clone + Debug + 'static> TestRuntime<PData> {
                 .map(|terminal_state| {
                     for snapshot in terminal_state.into_metrics() {
                         let _ = metrics_reporter_terminal.try_report_snapshot(snapshot);
-                        metrics_collector.clone().collect_once();
                     }
+                    metrics_collector.collect_pending(); // Collect after sending all the
                 })
         });
         TestPhase {
--- a/rust/otap-dataflow/crates/telemetry/src/collector.rs
+++ b/rust/otap-dataflow/crates/telemetry/src/collector.rs
@@ -43,10 +43,10 @@ impl InternalCollector {
         )
     }
 
-    /// Performs a single collection of metrics from the reporting channel and,
-    /// if there are any buffered messages, aggregates them into the `registry`.
-    pub fn collect_once(self: Arc<Self>) {
-        if let Ok(metrics) = self.metrics_receiver.try_recv() {
+    /// Drains all pending metrics from the reporting channel and aggregates
+    /// them into the `registry`.
+    pub fn collect_pending(&self) {
+        while let Ok(metrics) = self.metrics_receiver.try_recv() {
             self.registry
                 .accumulate_metric_set_snapshot(metrics.key, &metrics.metrics);
         }


rust/otap-dataflow/crates/telemetry/src/collector.rs
```diff
@@ -43,6 +43,15 @@ impl InternalCollector {
         )
     }
 
+    /// Performs a single collection of metrics from the reporting channel and,
+    /// if there are any buffered messages, aggregates them into the `registry`.
+    pub fn collect_once(self: Arc<Self>) {
```
utpilla: I don't think we need an `Arc` here. You could simply use `&self` and also simplify the call site to remove the `clone()`.

```suggestion
    pub fn collect_once(&self) {
```

rust/otap-dataflow/crates/engine/src/testing/exporter.rs
```diff
@@ -256,13 +256,15 @@ impl<PData: Clone + Debug + 'static> TestRuntime<PData> {
...
+        let metrics_collector = self.metrics_system.collector();
         let run_exporter_handle = self.local_tasks.spawn_local(async move {
             exporter
                 .start(pipeline_ctrl_msg_tx, metrics_reporter_start)
                 .await
                 .map(|terminal_state| {
                     for snapshot in terminal_state.into_metrics() {
                         let _ = metrics_reporter_terminal.try_report_snapshot(snapshot);
+                        metrics_collector.clone().collect_once();
                     }
```
utpilla, line 267: It would be cleaner to send all the snapshots first and then collect everything in one sweep instead of doing per iteration channel roundtrips of `try_send` and `try_recv`.

```rust
.map(|terminal_state| {
    for snapshot in terminal_state.into_metrics() {
        let _ = metrics_reporter_terminal.try_report_snapshot(snapshot);
    }
    metrics_collector.collect_pending(); // Collect after sending all the
})
```

In `collector.rs`:

```rust
/// Drains all pending metrics from the reporting channel and aggregates
/// them into the `registry`.
pub fn collect_pending(&self) {
    while let Ok(metrics) = self.metrics_receiver.try_recv() {
        self.registry
            .accumulate_metric_set_snapshot(metrics.key, &metrics.metrics);
    }
}
```

utpilla:
Is this issue applicable to receiver `TestRuntime` as well?

https://github.com/open-telemetry/otel-arrow/blob/fd92a9e19d4b95d0ad022c0132f4fb47dda71241/rust/otap-dataflow/crates/engine/src/testing/receiver.rs#L288-L290

rust/otap-dataflow/crates/engine/src/testing/exporter.rs
albertlockett: @utpilla made this change in [ce25fd8](https://github.com/open-telemetry/otel-arrow/pull/2082/commits/ce25fd82ea5d738850b139042f700a7ca14a5b68)

I made you a coauthor since I basically copied your comment verbatim :)

albertlockett:
> Is this issue applicable to receiver `TestRuntime` as well?
> 
> https://github.com/open-telemetry/otel-arrow/blob/fd92a9e19d4b95d0ad022c0132f4fb47dda71241/rust/otap-dataflow/crates/engine/src/testing/receiver.rs#L288-L290

@utpilla yes, but it has a somewhat different issue as well where the receiver of the metrics is created in such a way where the test would have no way to access it.
https://github.com/open-telemetry/otel-arrow/blob/fd92a9e19d4b95d0ad022c0132f4fb47dda71241/rust/otap-dataflow/crates/engine/src/testing/receiver.rs#L278-L279

The solution here would probably be to implement something like what we've done for the exporter TestRuntime, where we have an `InternalTelemetrySystem` owned by the `receiver::TestRuntime`, but given that this is a bigger change, I'd prefer to maybe do it as a followup and just keep this PR focused on the exporter.

rust/otap-dataflow/crates/engine/src/testing/exporter.rs
utpilla: Haha, happy to help!

Issue #1847, opened 2026-01-20
Rearrange node folder structure


drewrelmas:
As discussed in 2025-01-13 SIG meeting, the current [otap subfolder](https://github.com/open-telemetry/otel-arrow/tree/main/rust/otap-dataflow/crates/otap) should be rearranged better identify component types (`receiver`, `processor`, and `exporter`) as well as core/contrib status.

This would bring the code organization closer to the known style of the Collector/Collector-Contrib repositories.

## Proposal

~~Group components by `kind`. In addition, rename `/experimental` [subfolder](https://github.com/open-telemetry/otel-arrow/tree/main/rust/otap-dataflow/crates/otap/src/experimental) to `/contrib` to better indicate status. If there is ever a need for a `otel-arrow-contrib` repository, this setup enables an easy lift and shift.~~

* ~~`otel-arrow/rust/otap-dataflow/crates`~~
  * ~~`/nodes`~~
    * ~~`/receiver`~~
      * ~~`/otlp_receiver`~~
      * ~~`/syslog_cef_receiver`~~
      * ~~...~~
    * ~~`processor`~~
      * ~~`/attributes_processor`~~
      * ~~`/debug_processor`~~
      * ~~...~~
    * ~~`exporter`~~
      * ~~`/otap_exporter`~~
      * ~~`/otlp_exporter`~~
      * ~~...~~
    * ~~`/contrib`~~
      * ~~`/receiver`~~
      * ~~`/processor`~~
        * ~~`/recordset_kql_processor`~~
        * ~~`/condese_attributes_processor`~~
      * ~~`/exporter`~~
        * ~~`/geneva_exporter`~~
        * ~~`/azure_monitor_exporter`~~

Miscellaneous code (non-node-related) in the [otap subfolder](https://github.com/open-telemetry/otel-arrow/tree/main/rust/otap-dataflow/crates/otap) may be moved around as needed after every node component has been moved.

EDIT: See comment thread below, the group has decided on an alternate approach.

vihaan-that:
@drewrelmas  I have made an attempt at solving the issue. I would be grateful if a review could be performed on the PR.

lquerel:
I'd like to propose an alternative organization that I think scales better, especially given that the project is already a multi-crate, workspace-based setup.

### Alternative proposal: split core and contrib into separate crates

Instead of keeping core and contrib nodes in a single `nodes` crate with an internal `contrib/` hierarchy, we could split them at the crate level:

[...]

PR #2084, opened 2026-02-20
Refactor otap/experimental to new contrib-nodes crate

### LINKED ISSUE #2085: Reorganize experimental nodes to contrib-nodes crate

drewrelmas:
# Change Summary

First step towards the goal state mentioned in #1847. I chose to start with the experimental subfolder as it has a limited number of nodes to confirm methodology.

I tried to also clean up the `Cargo.toml` dependencies in both `otap-df/otap` and `otap-df/contrib-nodes` to separate concerns, but it is always possible there are more unused that should be removed.

* Progress towards #1847
* Closes #2085 

Unit tests and ran `main.rs` with nodes from `contrib-nodes` enabled.

No, only developer-facing.

### DIFF
--- a/rust/otap-dataflow/crates/otap/README.md
+++ b/rust/otap-dataflow/crates/otap/README.md
@@ -34,7 +34,8 @@ optional contrib processors) live in `crates/contrib-nodes`.
 - Error Exporter (`src/error_exporter.rs`)
 - Noop Exporter (`src/noop_exporter.rs`)
 - OTAP Exporter (`src/otap_exporter/`, `src/otap_exporter.rs`)
-- OTLP Exporter (`src/otlp_exporter.rs`)
+- OTLP gRPC Exporter (`src/otlp_exporter.rs`)
+- OTLP HTTP Exporter (`src/otlp_http_exporter/`)
 - Parquet Exporter (`src/parquet_exporter/`, `src/parquet_exporter.rs`)
 - Perf Exporter (`src/perf_exporter/`)
 


rust/otap-dataflow/crates/contrib-nodes/README.md
```diff
@@ -0,0 +1,29 @@
...
+- `geneva-exporter`
+- `azure-monitor-exporter`
+
+### Processors
+
+- `condense-attributes-processor`
+- `recordset-kql-processor`
+- `resource-validator-processor`
+
+When these features are enabled in the top-level binary, their factories are
```
lquerel, line 36: I suggest to list these features enabling those contrib nodes

lquerel (approved):
It's really great to see this reorganization and split of the crate folders around the nodes.

There's still that unfinished discussion about the common prefix used for all our crates. Right now we use `otap-df-`, and I believe we previously mentioned `otel-dataflow-`, which would align nicely with the `version` field format we recently introduced in our configuration files.

This doesn’t necessarily need to be resolved in this PR, I just wanted to use this opportunity to reopen the discussion.

rust/otap-dataflow/crates/contrib-nodes/README.md
drewrelmas: In https://github.com/open-telemetry/otel-arrow/pull/2084/commits/e7525c4a84d3797828e685fdb16e3321ed3078f8 I've improved the documentation both in this file and the root `otap-dataflow` `README.md`. In addition, I changed how the `contrib-exporters` and `contrib-processors` features work - I assume they were meant to enable all contrib nodes of that category. (Can still opt-in to specific nodes as well, they each retain their own flag).

rust/otap-dataflow/crates/contrib-nodes/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+
+/// Exporter implementations for contrib nodes.
+pub mod exporters;
+
+/// Processor implementations for contrib nodes.
+pub mod processors;
+
+/// Factory for Contrib-based pipeline components
+#[pipeline_factory(CONTRIB, OtapPdata)]
+pub static CONTRIB_PIPELINE_FACTORY: PipelineFactory<OtapPdata> = build_factory();
```
utpilla: Do we need this? Is the plan to have the contrib nodes register against this factory?

drewrelmas: I took another look, was initially mimicking the lib.rs at `crates/otap` but I think you're correct, it isn't necessary. The important bit is the 
```rust
// Keep this side-effect import so the crate is linked and its `linkme`
// distributed-slice registrations (contrib processors/exporters) are visible
// in `OTAP_PIPELINE_FACTORY` at runtime.
use otap_df_contrib_nodes as _;
```
in `src/main.rs`.

Without it, even if the build specifies the proper features, the engine process doesn't recognize the `contrib` node types.

rust/otap-dataflow/crates/otap/README.md
```diff
@@ -1,46 +1,46 @@
...
-- Noop Exporter: An exporter that drops all data.
-- Parquet Exporter: An exporter that writes data to Parquet files.
-- Geneva Exporter (Experimental): An exporter for Microsoft Geneva monitoring
-  backend. Supports logs and traces. See [experimental/geneva_exporter](src/experimental/geneva_exporter/)
-  for details. Enable with `--features geneva-exporter`.
+- Console Exporter (`src/console_exporter.rs`)
+- Error Exporter (`src/error_exporter.rs`)
+- Noop Exporter (`src/noop_exporter.rs`)
+- OTAP Exporter (`src/otap_exporter/`, `src/otap_exporter.rs`)
+- OTLP Exporter (`src/otlp_exporter.rs`)
```
utpilla: Now that #2070 is merged, we could be more specific: OTLP gRPC Exporter and OTLP HTTP Exporter.

rust/otap-dataflow/crates/contrib-nodes/src/lib.rs
drewrelmas: Removed in https://github.com/open-telemetry/otel-arrow/pull/2084/changes/79e07b555633ce3b094deed498d3443986b84543

rust/otap-dataflow/crates/otap/README.md
drewrelmas: Agreed, updated.

utpilla (approved):

PR #2089, opened 2026-02-21
Add endpoint override config options for OTLP HTTP exporter

albertlockett:
# Change Summary

small followup from https://github.com/open-telemetry/otel-arrow/pull/2070.

Adds new config options for each signal type to override the endpoint to which the OTLP HTTP exporter sends data.

This is to aid with parity between this implementation and the analogous Go collector component, which also has these options: https://github.com/open-telemetry/opentelemetry-collector/tree/main/exporter/otlphttpexporter#otlp-http-exporter

* Part of #1145 

A new unit test is added.

 Users can configure the component with these new options.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs
@@ -117,6 +117,23 @@ impl OtlpHttpExporter {
             error: format!("invalid endpoint URL: {e}"),
         })?;
 
+        // validate the endpoint overrides if supplied
+        if let Some(endpoint) = config.logs_endpoint.as_ref() {
+            _ = reqwest::Url::parse(endpoint).map_err(|e| ConfigError::InvalidUserConfig {
+                error: format!("invalid logs endpoint URL: {e}"),
+            })?;
+        }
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/otap/src/otlp_http_exporter/config.rs
```diff
@@ -21,6 +21,24 @@ pub struct Config {
     /// Example: "http://localhost:4318" or "https://otel-collector:4318"
     pub endpoint: String,
 
+    /// The target URL to send trace data to, including the path. If this setting is present the
+    /// endpoint setting is ignored for traces.
+    ///
+    /// Example: https://example.com:4318/v1/traces
+    pub traces_endpoint: Option<String>,
```
lalitb, line 28: Good to also add validation for these signal-override endpoints in `from_config`, similar to how we do for 'endpoint`.

lalitb (approved):

rust/otap-dataflow/crates/otap/src/otlp_http_exporter/config.rs
albertlockett: Thanks @lalitb , added in 77b800dd2

PR #2096, opened 2026-02-23
   [Geneva exporter] Add UserManagedIdentityByResourceId auth variant

lalitb:
### Description:

   - Adds `UserManagedIdentityByResourceId` variant to the Geneva exporter's `AuthConfig` enum, enabling authentication via Azure Resource Manager resource ID. This maps to the existing `AuthMethod::UserManagedIdentityByResourceId` in `geneva-uploader` - no new auth logic added.

### Changes:
   - New `AuthConfig` variant with `resource_id` and `msi_resource` fields
   - Mapping to `AuthMethod::UserManagedIdentityByResourceId` in `from_config`
   - `msi_resource` extraction for the new variant
   - Extended `test_auth_config_variants` to cover all five auth variants


rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -134,6 +134,14 @@ pub enum AuthConfig {
         /// MSI resource identifier
         msi_resource: String,
     },
+    /// User-assigned managed identity (by ARM resource ID)
+    /// Identifies the managed identity by its Azure Resource Manager resource ID.
+    UserManagedIdentityByResourceId {
```
utpilla: Would it be better to include Arm in the name?

```suggestion
    UserManagedIdentityByArmResourceId {
```

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -872,13 +886,50 @@ mod tests {
...
             "msi_resource": "https://resource"
         });
-        let _system_mi: AuthConfig = serde_json::from_value(system_mi_json).unwrap();
+        let system_mi: AuthConfig = serde_json::from_value(system_mi_json).unwrap();
+        assert!(matches!(
+            system_mi,
+            AuthConfig::SystemManagedIdentity { .. }
+        ));
+
+        let user_mi_json = serde_json::json!({
```
ThomsonTan, line 903: Add a test to call `create_exporter_from_factory` with a config using the `usermanagedidentitybyresourceid` auth type, verifying the exporter initialises without error. This would exercise the from_config mapping path for the new variant.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
lalitb: Good point. Done.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
lalitb: Thanks. Added.

gouslu:
GRAIN OF SALT: I'm still fairly new to rust, so I might be missing something here, but I was wondering if it could be done in a way without creating multiple variants of `UserManagedIdentity`.

Maybe a late comment, but I wonder if we could consolidate the three `UserManagedIdentity*` variants under a single `UserManagedIdentity` variant with an implicit selector. The identity type would be inferred from which field the user provides — no extra discriminator needed.

**Config examples:**

```yaml
# by client_id (most common)
auth:
  type: usermanagedidentity
  client_id: "abc-123"
  msi_resource: "https://..."

# by object_id
auth:
  type: usermanagedidentity
  object_id: "def-456"
  msi_resource: "https://..."

# by resource_id
auth:
  type: usermanagedidentity
  resource_id: "/subscriptions/.../identities/my-id"
  msi_resource: "https://..."
```

**`AuthConfig` variant using `#[serde(flatten)]`** — the selector fields appear as siblings of `msi_resource` in the config:

```rust
#[derive(Debug, Deserialize, Clone)]
#[serde(tag = "type", rename_all = "lowercase")]
pub enum AuthConfig {
    // ... other variants ...
    UserManagedIdentity {
        #[serde(flatten)]
        selector: ManagedIdentitySelector,
        msi_resource: String,
    },
}
```

**Inner enum with `#[serde(untagged)]`** — serde picks the variant based on which field name is present:

```rust
#[derive(Debug, Deserialize, Clone)]
#[serde(untagged)]
pub enum ManagedIdentitySelector {
    ClientId { client_id: String },
    ObjectId { object_id: String },
    ResourceId { resource_id: String },
}
```

**Mapping to upstream `AuthMethod`:**

```rust
AuthConfig::UserManagedIdentity { selector, msi_resource } => {
    let auth_method = match selector {
        ManagedIdentitySelector::ClientId { client_id } => {
            AuthMethod::UserManagedIdentity { client_id }
        }
        ManagedIdentitySelector::ObjectId { object_id } => {
            AuthMethod::UserManagedIdentityByObjectId { object_id }
        }
        ManagedIdentitySelector::ResourceId { resource_id } => {
            AuthMethod::UserManagedIdentityByResourceId { resource_id }
        }
    };
}
```

This keeps the config surface clean (one `type` value instead of three) while still supporting all upstream identity selectors. The `msi_resource` field — shared across all sub-variants — lives at the `UserManagedIdentity` level where it belongs.

lalitb:
> Maybe a late comment, but I wonder if we could consolidate the three `UserManagedIdentity*` variants under a single `UserManagedIdentity` variant with an implicit selector. The identity type would be inferred from which field the user provides — no extra discriminator needed.
> 
>
> 
> This keeps the config surface clean (one `type` value instead of three) while still supporting all upstream identity selectors. The `msi_resource` field — shared across all sub-variants — lives at the `UserManagedIdentity` level where it belongs.


 @gouslu  - Thanks for the suggestion. I like the idea from a config UX perspective. Having a single `usermanagedidentity` type and inferring the selector from whichever field is present definitely makes the YAML cleaner.                                                                
                  
  My only concern is the serde shape. This would combine a tagged outer enum with `flatten` and an `inner` untagged enum. It works right now because each selector is a single unique field, but `untagged` relies on trial deserialization - if we ever extend one of those variants with additional optional fields or defaults, we could accidentally introduce ambiguity or order-dependent behavior. Error messages from `untagged` failures can also be quite opaque.

Since this is config-facing and likely to evolve, I lean toward keeping the variants explicit for now. That said, happy to explore it further in a follow-up if you think the UX benefit is worth it.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -134,6 +134,14 @@ pub enum AuthConfig {
         /// MSI resource identifier
         msi_resource: String,
     },
+    /// User-assigned managed identity (by ARM resource ID)
+    /// Identifies the managed identity by its Azure Resource Manager resource ID.
+    UserManagedIdentityByArmResourceId {
```
ThomsonTan, line 139: serde tag `rename_all = "lowercase"` makes this into "usermanagedidentitybyarmresourceid" which is hard to recognize, perhaps set an alias like `#[serde(alias = "user_managed_identity_by_arm_resource_id")]` to make it more readable.

jmacd: OK to do this in another PR!

PR #2101, opened 2026-02-24
Node telemetry_attributes into entity::extend::identity_attributes

### LINKED ISSUE #2078: Custom node attributes - Follow-up
We recently introduced the concept of `telemetry_attributes` to allow adding custom attributes to nodes, in addition to the entity attributes automatically defined by the engine.

It is important to note that in the semantic conventions, an entity is characterized by a set of attributes that uniquely identify it, along with an optional set of descriptive attributes that do not participate in its identification.

As a result, the current `telemetry_attributes` concept appears incomplete. This issue proposes introducing the following structure instead:

```yaml
type: <urn>
config:
  ... custom node config
entity:    # a generic way to extend the entity attributes
  extend:
    identity_attributes: ...
    # and later we could add descriptive_attributes
```

This would make the intent much clearer and align more naturally with the semantic conventions model.

jmacd:
# Change Summary

Moves telemetry_attributes as requested.

Fixes #2078.

✅ 

Yes, documented.

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/lib.rs
+++ b/rust/otap-dataflow/crates/engine/src/lib.rs
@@ -483,7 +483,7 @@ impl<PData: 'static + Clone + Debug> PipelineFactory<PData> {
                 name.clone(),
                 node_config.r#type.clone(),
                 node_kind,
-                node_config.identity_attributes().clone(),
+                node_config.identity_attributes(),
             );
 
             match node_kind {


rust/otap-dataflow/crates/engine/src/lib.rs
```diff
@@ -504,7 +483,7 @@ impl<PData: 'static + Clone + Debug> PipelineFactory<PData> {
                 name.clone(),
                 node_config.r#type.clone(),
                 node_kind,
-                node_config.telemetry_attributes.clone(),
+                node_config.identity_attributes().clone(),
```
utpilla:
```suggestion
                node_config.identity_attributes(),
```

PR #2103, opened 2026-02-24
feat(metrics): enhance DurableBufferMetrics with per-item, expired, (re)queued tracking

### LINKED ISSUE #2104: Durable buffer processor lacks per-item observability
The `durable_buffer_processor` currently only tracks bundle-level metrics (ACK/NACK counts, in-flight bundles, dropped/expired segments and bundles). There is no visibility into the number of individual telemetry items (log records, metric data points, spans) flowing through the processor, i.e. how many were ingested, how many are queued awaiting ACK, how many were lost to retention policies, or how many were requeued after a NACK. These item-level counts are more relevant for pipeline operators vs. record bundles / pdata batches.

Additionally, the existing `consumed_arrow_*`/`produced_arrow_*` counters only work for Arrow-format data; OTLP pass-through bundles are completely uncounted. This makes it difficult for operators to assess data loss severity (e.g., "we dropped 2 segments" says nothing about how many log records that was), monitor queue depth in item-meaningful units, or understand retry impact per signal type. Item counts should also survive restarts so queued gauges reflect accurate state after recovery.

AaronRM:
# Change Summary

This commit adds per-item, per-signal-type metrics to the durable buffer processor. Previously only Arrow-format bundles were counted; now OTLP pass-through bundles are also counted by scanning the protobuf wire format. Item counts are persisted in the Quiver segment manifest and restored on restart, enabling accurate queued-item gauges across restarts.

New metrics under `otelcol.node.durable_buffer`:

- Consumed/produced counters per signal type (log records, metric points, spans): renamed from the previous *_arrow_* variants
- Queued gauges per signal type: items ingested but not yet ACKed, seeded from WAL on restart
- Requeued counters per signal type: items in NACKed bundles scheduled for retry
- Dropped/expired item gauges: item-granularity equivalents of the existing bundle-level gauges

- Closes #2104

- Added relevant unit tests to validate counters from `durable_buffer_processor`
- Manually validated counters behave as expected from prometheus endpoint

Yes. There are new metrics exposed under the durable_buffer metric set. The previous `consumed.arrow.*` and `produced.arrow.*` metrics are renamed to the signal-agnostic names above (e.g., `consumed.arrow.logs` => `consumed.log.records`). This is a breaking change for any dashboards or alerts referencing the old metric names.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/durable_buffer_processor/mod.rs
+++ b/rust/otap-dataflow/crates/otap/src/durable_buffer_processor/mod.rs
@@ -111,7 +111,7 @@ use otap_df_engine::{
     ConsumerEffectHandlerExtension, Interests, ProcessorFactory, ProducerEffectHandlerExtension,
 };
 use otap_df_pdata::{OtapArrowRecords, OtapPayload};
-use otap_df_telemetry::instrument::{Counter, Gauge};
+use otap_df_telemetry::instrument::{Counter, Gauge, ObserveCounter};
 use otap_df_telemetry::metrics::MetricSet;
 use otap_df_telemetry_macros::metric_set;
 
@@ -207,28 +207,28 @@ pub struct DurableBufferMetrics {
     /// Total segments force-dropped due to DropOldest retention policy.
     /// Non-zero values indicate data loss.
     #[metric(unit = "{segment}")]
-    pub dropped_segments: Gauge<u64>,
+    pub dropped_segments: ObserveCounter<u64>,
 
     /// Total bundles lost due to force-dropped segments (DropOldest policy).
     /// Non-zero values indicate data loss.
     #[metric(unit = "{bundle}")]
-    pub dropped_bundles: Gauge<u64>,
+    pub dropped_bundles: ObserveCounter<u64>,
 
     /// Total individual items (log records, data points, spans) lost due to
     /// force-dropped segments (DropOldest policy). Non-zero values indicate data loss.
     #[metric(unit = "{item}")]
-    pub dropped_items: Gauge<u64>,
+    pub dropped_items: ObserveCounter<u64>,
 
     /// Total bundles lost due to expired segments (max_age retention).
     /// Non-zero values indicate data aged out before delivery.
     #[metric(unit = "{bundle}")]
-    pub expired_bundles: Gauge<u64>,
+    pub expired_bundles: ObserveCounter<u64>,
 
     /// Total individual items (log records, data points, spans) lost due to
     /// expired segments (max_age retention). Non-zero values indicate data
     /// aged out before delivery.
     #[metric(unit = "{item}")]
-    pub expired_items: Gauge<u64>,
+    pub expired_items: ObserveCounter<u64>,
 
     // ─── Retry metrics ──────────────────────────────────────────────────────
     /// Number of retry attempts scheduled.
@@ -1695,11 +1695,11 @@ impl otap_df_engine::local::processor::Processor<OtapPdata> for DurableBuffer {
                     {
                         self.metrics.storage_bytes_used.set(used);
                         self.metrics.storage_bytes_cap.set(cap);
-                        self.metrics.dropped_segments.set(dropped_segs);
-                        self.metrics.dropped_bundles.set(dropped_buns);
-                        self.metrics.dropped_items.set(dropped_items);
-                        self.metrics.expired_bundles.set(expired_buns);
-                        self.metrics.expired_items.set(expired_items);
+                        self.metrics.dropped_segments.observe(dropped_segs);
+                        self.metrics.dropped_bundles.observe(dropped_buns);
+                        self.metrics.dropped_items.observe(dropped_items);
+                        self.metrics.expired_bundles.observe(expired_buns);
+                        self.metrics.expired_items.observe(expired_items);
                     }
 
                     metrics_reporter


rust/otap-dataflow/crates/otap/src/durable_buffer_processor/mod.rs
```diff
@@ -559,6 +628,65 @@ impl DurableBuffer {
...
+    /// should be reflected in the queued gauges. This iterates all available
+    /// segments, reads per-bundle metadata (item count + first slot ID), and
+    /// classifies each bundle's signal type so the per-signal gauges start at
+    /// the correct value.
+    ///
+    /// This is a one-time cost at engine initialization — O(segments × bundles)
+    /// with no I/O beyond what the segment store already has in memory.
+    fn seed_queued_counters_from_wal(&mut self, engine: &QuiverEngine) {
+        let store = engine.segment_store();
+        for seg_seq in store.available_segments() {
```
ThomsonTan, line 643: If the subscriber had already ACKed some bundles before the last shutdown, those segments still exist in the store (`cleanup_completed_segments` hasn't run yet). This could cause the queued gauges to over-count at startup. So probably before iterating segments, call `engine.cleanup_completed_segments()`.

rust/otap-dataflow/crates/otap/src/durable_buffer_processor/mod.rs
```diff
@@ -210,6 +214,22 @@ pub struct DurableBufferMetrics {
...
+    /// Total bundles lost due to expired segments (max_age retention).
+    /// Non-zero values indicate data aged out before delivery.
+    #[metric(unit = "{bundle}")]
+    pub expired_bundles: Gauge<u64>,
+
+    /// Total individual items (log records, data points, spans) lost due to
+    /// expired segments (max_age retention). Non-zero values indicate data
+    /// aged out before delivery.
+    #[metric(unit = "{item}")]
+    pub expired_items: Gauge<u64>,
```
utpilla: `dropped_segments`, `dropped_bundles`, `dropped_items`, `expired_bundles`, `expired_items` should be `ObserveCounter` instruments instead of `Gauge`.

All of these are monotonically increasing cumulative totals, sampled from Quiver's atomic counters. They would only ever go up.

AaronRM: Fixed

rust/otap-dataflow/crates/quiver/src/segment/reader.rs
```diff
@@ -866,6 +867,10 @@ impl SegmentReader {
         let bundle_indices =
             Self::get_primitive_column::<arrow_array::types::UInt32Type>(&batch, "bundle_index")?;
 
+        // item_count is optional for backward compatibility with legacy segments.
+        let item_counts: Option<Vec<u64>> =
+            Self::get_primitive_column::<arrow_array::types::UInt64Type>(&batch, "item_count").ok();
```
utpilla, line 872: I'm assuming we don't have to worry about backward compatibility since it's not used in production yet. It would be better to make `item_count` required rather than silently defaulting to `0`.

rust/otap-dataflow/crates/otap/src/durable_buffer_processor/mod.rs
AaronRM: @ThomsonTan Yes, there are some over-counting problems. It will require more than calling `cleanup_completed_segments()` to address. Quiver's `SubscriberRegistry` will need to expose the subscriber's state so that this module can accurately count in the case of partially (or fully) ACK'ed segments on startup. I will be pushing a commit shortly with this change.

AaronRM: This PR has been added to the merge queue. I will create a follow up PR to address.

rust/otap-dataflow/crates/otap/src/durable_buffer_processor/mod.rs
```diff
@@ -146,37 +150,37 @@ pub struct DurableBufferMetrics {
...
-    pub produced_arrow_metrics: Counter<u64>,
-
-    /// Number of Arrow trace spans produced (sent downstream).
-    /// OTLP pass-through mode skips counting to avoid parsing overhead.
-    #[metric(unit = "{span}")]
-    pub produced_arrow_traces: Counter<u64>,
+    // ─── Consumed item metrics (per signal type) ────────────────────────
+    /// Number of log records consumed (ingested to WAL).
+    /// For OTLP bytes, counted by scanning the protobuf wire format without full deserialization.
+    #[metric(unit = "{records}")]
```
lalitb, line 156: Follow up: for a future PR, consider using canonical semconv unit annotations (`{log_record}`, `{data_point}`, `{span}`) so units stay singular and aligned with OTel conventions.

Reference: [Instrument units](https://opentelemetry.io/docs/specs/semconv/general/metrics/#instrument-units)

> "All non-units that use curly braces to annotate a quantity need to match the grammatical number of the quantity it represent. For example if measuring the number of individual requests to a process the unit would be `{request}`, not `{requests}`."

Examples from [OTel SDK metrics semconv](https://opentelemetry.io/docs/specs/semconv/otel/sdk-metrics/):
- `otel.sdk.log.created` -> `{log_record}`
- `otel.sdk.processor.log.queue.size` -> `{log_record}`
- `otel.sdk.exporter.metric_data_point.inflight` -> `{data_point}`
- `otel.sdk.exporter.metric_data_point.exported` -> `{data_point}`
- `otel.sdk.span.started` -> `{span}`

AaronRM: Addressing in #2112.

rust/otap-dataflow/crates/otap/src/durable_buffer_processor/mod.rs
AaronRM: Tracking with #2193

PR #2105, opened 2026-02-24
chore: use debug build for validation as its slightly faster

cijothomas:
Nit improvement. For config validation, debug build is fine and is faster.


cijothomas:
5m 6s vs 13m :) (Not really significant for CI times, as this was anyway in parallel. But why waste CPU cycles!)

Issue #2018, opened 2026-02-10
Component-level pipeline metrics implementation


jmacd:
I considered this to be our "north star" document for internal metrics about the pipeline:

https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md

This document describes how to record complete information about the traversal of requests (or items, or bytes) through the pipeline when complete observability is required, meaning to count every incoming and separately count every outgoing request. 

In addition to counts, this document prescribes semantic conventions for naming attributes to describe the pipeline name, signal type, and component name(type).  Lastly, it prescribes the use of an `outcome` attribute which carries the accounting backwards as success, failure, and refusal propagates backwards in the pipeline.

We have implemented the mechanism for counting requests in the forward direction through integration with our channel types. We do not have any automatic accounting for Ack/Nack or to distinguish internally-generated failures from externally-propagated refusals (i.e., downstream failures). We are not using the same metric names as the document above, but it's is just a naming issue.

This issue is to cover completing this story, and not necessarily by following the guideline above exactly. Because of our message-passing style, there is no state automatically available on the backward path for easy metric accounting. For example, by the time an Ack or Nack is received we do not automatically know anything about the request, it's gone. We can auto-instrument on the forward path where the outcome is still unknown.

I've investigated how to address this: one proposal in #2011. We can use the Context vector of Frames to store any information needed to instrument the return path, as that data automatically propagates with Ack/Nack. For example, we could store a timestamp during the forward path (e.g., an argument to the receiver's subscribe_to()) in the call data and auto-instrument duration on the return (with a status indicating Ack or Nack).

My proposal is to support opt-in pipeline instrumentation for the return path. The way to opt-in to counting outcomes this way will be to subscribe to Ack/Nack. In other words, you will have to use the Ack/Nack mechanism if you want a metric with outcomes counted.

As a minor complication, we may want to use different metric instrument names when outcome-accounting is enabled or not. If you are not subscribed to Ack/Nack, then you would count requests on the forward path. If you are subscribed to Ack/Nack, you would up-down count active requests, increment on the forward path and decrement on the return path while emitting counts with outcomes. This logic applies only when both Ack and Nack are requested.

[...]

PR #2106, opened 2026-02-24
Azure monitor exporter metrics

gouslu:
**### PR Description is generated with the help of AI ###**

# Change Summary

Migrate the Azure Monitor Exporter from the custom `Stats` reporting system (periodic `print_stdout`) to the `otap-df-telemetry` metrics framework, and audit all log statements for correct severity levels.

**Metrics migration:**
- Remove `stats.rs` (325 lines) and replace with `metrics.rs` (448 lines) using the `#[metric_set]` proc macro and `AzureMonitorExporterMetricsTracker` wrapper
- 21 metric instruments: counters for success/failure (rows, batches, messages), HTTP status codes (2xx/401/403/413/429/5xx), auth success/failure, `log_entry_too_large`, `heartbeats`, `transform_failures`; `Mmsc` histograms for HTTP and auth latency; `Gauge` for in-flight exports and internal state map sizes
- Wire shared metrics (`Rc<RefCell<…>>`) into exporter, client, auth, and transformer

**Log severity audit:**
- Correct `auth.get_token_succeeded` from `otel_info!` to `otel_debug!` (redundant with exporter's token refresh info log)
- Add 5 new log statements for previously silent failure paths: `message.unsupported_signal` (warn ×2), `message.no_valid_entries` (debug), `message.log_entry_too_large` (warn with `size_bytes`), `message.batch_push_failed` (error)

**Cleanup:**
- Remove `#[allow(clippy::print_stdout)]` and stale TODO from `gzip_batcher.rs`

Checks off the "Internal metrics" item from #1396.

- 152 existing unit tests pass, covering metrics initialization, counter increments, histogram recording, report generation, transformer field mappings, gzip batching, auth flows, and client HTTP handling
- `cargo clippy` clean (zero warnings)
- All tests: `cargo test -p otap-df-contrib-nodes --features azure-monitor-exporter`
- Manual testing and debugging to validate whether the telemetry is being collected and is recorded accurately

Yes:
- **Removed:** periodic `Stats` printout to stdout is replaced by structured OTel metrics emitted through the telemetry framework
- **New log statements:** 5 additional structured logs for failure paths that were previously silent

### DIFF
--- a/rust/otap-dataflow/crates/otap/Cargo.toml
+++ b/rust/otap-dataflow/crates/otap/Cargo.toml
@@ -67,7 +67,7 @@ weaver_semconv.workspace = true
 weaver_resolved_schema.workspace = true
 weaver_resolver.workspace = true
 rand.workspace = true
-reqwest = { workspace = true, features = ["rustls", "stream"] }
+reqwest = { workspace = true, features = ["http2", "rustls", "stream"] }
 zip.workspace = true
 tower = { workspace = true }
 hyper-util = { workspace = true }
--- a/rust/otap-dataflow/crates/otap/src/otlp_http.rs
+++ b/rust/otap-dataflow/crates/otap/src/otlp_http.rs
@@ -848,8 +848,9 @@ pub async fn serve(
                                 Err(_) => return,
                             };
                             let io = TokioIo::new(tls_stream);
-                            let conn = hyper::server::conn::http1::Builder::new()
-                                .serve_connection(io, service_fn(move |req| handler.clone().handle(req)));
+                            let executor = hyper_util::rt::TokioExecutor::new();
+                            let conn = hyper_util::server::conn::auto::Builder::new(executor);
+                            let conn = conn.serve_connection(io, service_fn(move |req| handler.clone().handle(req)));
 
                             let mut conn = std::pin::pin!(conn);
 
--- a/rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
+++ b/rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
@@ -5,8 +5,15 @@
 
 use reqwest::ClientBuilder;
 use serde::Deserialize;
+use std::io;
 use std::time::Duration;
 use tower::limit::ConcurrencyLimitLayer;
+
+#[cfg(feature = "experimental-tls")]
+use {
+    crate::tls_utils::read_file_with_limit_async, otap_df_config::tls::TlsClientConfig,
+    otap_df_telemetry::otel_error, reqwest::Certificate, reqwest::Identity,
+};
 
 use crate::otap_grpc::client_settings::{
     default_concurrency_limit, default_connect_timeout, default_tcp_keepalive, default_tcp_nodelay,
@@ -42,6 +49,12 @@ pub struct HttpClientSettings {
     /// Timeout for HTTP requests. If not specified, no timeout is applied.
     #[serde(default, with = "humantime_serde")]
     pub timeout: Option<Duration>,
+
+    /// Client-side TLS/mTLS configuration.
+    /// Requires the `experimental-tls` feature to be enabled.
+    #[cfg(feature = "experimental-tls")]
+    #[serde(default)]
+    pub tls: Option<TlsClientConfig>,
 }
 
 impl HttpClientSettings {
@@ -51,8 +64,8 @@ impl HttpClientSettings {
         self.concurrency_limit.max(1)
     }
 
-    /// Returns a configured client-bulder
-    pub fn client_builder(&self) -> ClientBuilder {
+    /// Returns a configured client-builder
+    pub async fn client_builder(&self) -> Result<ClientBuilder, HttpClientError> {
         let mut client_builder = ClientBuilder::new()
             .connect_timeout(self.connect_timeout)
             .tcp_nodelay(self.tcp_nodelay)
@@ -72,8 +85,124 @@ impl HttpClientSettings {
             client_builder = client_builder.timeout(timeout)
         }
 
-        client_builder
+        #[cfg(feature = "experimental-tls")]
+        if let Some(tls) = &self.tls {
+            let mut certs = vec![];
+
+            if let Some(ca_pem) = &tls.ca_pem {
+                let cert = Certificate::from_pem(ca_pem.as_bytes())?;
+                certs.push(cert);
+            }
+
+            if let Some(ca_file) = &tls.ca_file {
+                let ca_pem = read_file_with_limit_async(ca_file).await.map_err(|e| {
+                    otel_error!(
+                        "tls.ca_file.read_error",
+                        ca_file = ?ca_file,
+                        error = ?e,
+                        message = "Failed to read CA file"
+                    );
+                    e
+                })?;
+                let cert = Certificate::from_pem(&ca_pem)?;
+                certs.push(cert);
+            }
+
+            if tls.include_system_ca_certs_pool.unwrap_or(true) {
+                client_builder = client_builder.tls_certs_merge(certs);
+            } else {
+                client_builder = client_builder.tls_certs_only(certs);
+            }
+
+            if let Some(true) = &tls.insecure_skip_verify {
+                client_builder = client_builder.danger_accept_invalid_certs(true);
+            }
+
+            // mTLS client certificate configuration
+            let client_cert_configured = tls.config.cert_file.is_some()
+                || tls
+                    .config
+                    .cert_pem
+                    .as_ref()
+                    .is_some_and(|pem| !pem.trim().is_empty());
+            let client_key_configured = tls.config.key_file.is_some()
+                || tls
+                    .config
+                    .key_pem
+                    .as_ref()
+                    .is_some_and(|pem| !pem.trim().is_empty());
+
+            if client_cert_configured || client_key_configured {
+                if !(client_cert_configured && client_key_configured) {
+                    return Err(io::Error::new(
+                        io::ErrorKind::InvalidInput,
+                        "TLS configuration error: both cert and key must be provided for mTLS. \
+                             Provide both cert_file/cert_pem and key_file/key_pem, or neither.",
+                    )
+                    .into());
+                }
+
+                // Read cert and key
+                let cert_pem = if let Some(cert_file) = &tls.config.cert_file {
+                    read_file_with_limit_async(cert_file).await.map_err(|e| {
+                        otel_error!(
+                            "tls.cert_file.read_error",
+                            cert_file = ?cert_file,
+                            error = ?e, message = "Failed to read client cert file"
+                        );
+                        e
+                    })?
+                } else if let Some(cert_pem) = &tls.config.cert_pem {
+                    cert_pem.as_bytes().to_vec()
+                } else {
+                    unreachable!()
+                };
+
+                let key_pem = if let Some(key_file) = &tls.config.key_file {
+                    read_file_with_limit_async(key_file).await.map_err(|e| {
+                        otel_error!(
+                            "tls.key_file.read_error",
+                            key_file = ?key_file,
+                            error = ?e,
+                            message = "Failed to read client key file"
+                        );
+                        e
+                    })?
+                } else if let Some(key_pem) = &tls.config.key_pem {
+                    key_pem.as_bytes().to_vec()
+                } else {
+                    unreachable!()
+                };
+
+                // Combine cert and key into PEM format for Identity
+                let mut identity_pem = cert_pem;
+                identity_pem.extend_from_slice(&key_pem);
+
+                let identity = Identity::from_pem(&identity_pem).map_err(|e| {
+                    io::Error::new(
+                        io::ErrorKind::InvalidInput,
+                        format!("Failed to create identity from cert and key: {}", e),
+                    )
+                })?;
+
+                client_builder = client_builder.identity(identity);
+            }
+        }
+
+        Ok(client_builder)
     }
+}
+
+/// Errors that occur configuring Http ClientBuilder
+#[derive(thiserror::Error, Debug)]
+pub enum HttpClientError {
+    /// Error occurred configuring reqwest client
+    #[error("http client build error: {0}")]
+    Reqwest(#[from] reqwest::Error),
+
+    /// IO Error occurred reading tls cert from file
+    #[error("http client build io error: {0}")]
+    Io(#[from] io::Error),
 }
 
 impl Default for HttpClientSettings {
@@ -85,6 +214,8 @@ impl Default for HttpClientSettings {
             tcp_keepalive: default_tcp_keepalive(),
             tcp_keepalive_interval: None,
             timeout: None,
+            #[cfg(feature = "experimental-tls")]
+            tls: None,
         }
     }
 }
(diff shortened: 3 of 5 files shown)


rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
```diff
@@ -16,8 +17,35 @@ pub struct Config {
...
     pub auth: AuthConfig,
+
+    /// Self-telemetry configuration for the exporter's own operational metrics
+    #[serde(default)]
+    pub telemetry: TelemetryConfig,
+}
+
+/// Configuration for the exporter's self-telemetry
+#[derive(Debug, Deserialize, Clone)]
+pub struct TelemetryConfig {
```
andborja: Why a custom telemetry config here? Where is it used?

cijothomas: +1. Should not have component specific duration etc.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
```diff
@@ -0,0 +1,448 @@
...
+#[metric_set(name = "azure_monitor_exporter.metrics")]
+#[derive(Debug, Default, Clone)]
+pub struct AzureMonitorExporterMetrics {
+    /// Number of rows successfully exported.
+    #[metric(unit = "{row}")]
+    pub successful_rows: Counter<u64>,
+    /// Number of batches successfully exported.
+    #[metric(unit = "{batch}")]
+    pub successful_batches: Counter<u64>,
+    /// Number of messages successfully exported.
```
cijothomas, line 31: what's the difference between row and message?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
```diff
@@ -0,0 +1,448 @@
...
+    /// Current number of message-to-data mappings (leak detector).
+    #[metric(unit = "{entry}")]
+    pub msg_to_data_count: Gauge<u64>,
+    /// Number of log entries rejected for exceeding the batch size limit.
+    #[metric(unit = "{entry}")]
+    pub log_entries_too_large: Counter<u64>,
+    /// Number of successful heartbeat sends.
+    #[metric(unit = "{heartbeat}")]
+    pub heartbeats: Counter<u64>,
+    /// Number of log records that failed to serialize during transformation.
```
cijothomas, line 90: what is transform inside an exporter?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
```diff
@@ -0,0 +1,448 @@
...
+    /// Current number of batch-to-message mappings (leak detector).
+    #[metric(unit = "{entry}")]
+    pub batch_to_msg_count: Gauge<u64>,
+    /// Current number of message-to-batch mappings (leak detector).
+    #[metric(unit = "{entry}")]
+    pub msg_to_batch_count: Gauge<u64>,
+    /// Current number of message-to-data mappings (leak detector).
+    #[metric(unit = "{entry}")]
+    pub msg_to_data_count: Gauge<u64>,
+    /// Number of log entries rejected for exceeding the batch size limit.
```
cijothomas, line 84: interesting.. when would this occur?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
```diff
@@ -452,9 +489,15 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
                 }
             })?;
 
+        // Start periodic telemetry collection (every 1 second)
+        _ = effect_handler
+            .start_periodic_telemetry(self.config.telemetry.metrics_report_interval)
```
andborja: The metrics internal sdk has its own configuration for periodic telemetry.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
drewrelmas: This isn't the true 'transform', it is the flattening from OTAP batch to key:value pair format that Azure Monitor expects. It was historically called the 'transformer' part of the exporter. I would almost prefer 'flattening' for differentiation.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
gouslu: message is each message we receive, which can contain multiple log records. Usually each message has multiple rows, each batch has ~1mb of rows, which usually has multiple messages.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
gouslu: that's a separate conversation, I don't think it belongs to this PR if it should be changed.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
gouslu: These msg_to_data etc are metrics for internal state of the exporter. I wanted to track them to detect any potential leaks based on bugs. Normally the state should be clean when it is idle.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
gouslu: ok I can get rid of it, thanks! I wasn't getting any collection event, that's why I thought I needed to add this, maybe I was doing something wrong, let me revise.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
gouslu: Can you show me how to use that and configure that? Do I not need start_periodic_telemetry here? I want to be able to configure colllection internval and start the collection as well.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
gouslu: > Why a custom telemetry config here? Where is it used?

it is used to setup metric collection interval. is it configured in any other way? to me it seemed like each component starts its own collection, and specifying interval here is useful for me.

image

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/auth.rs
```diff
@@ -56,16 +70,21 @@ impl Auth {
...
-                    otel_info!("azure_monitor_exporter.auth.get_token_succeeded", expires_on = %token.expires_on);
+                    otel_debug!("azure_monitor_exporter.auth.get_token_succeeded", expires_on = %token.expires_on);
+                    let mut m = self.metrics.borrow_mut();
+                    m.add_auth_latency(start.elapsed().as_millis() as f64);
+                    m.add_auth_success();
                     return Ok(token);
                 }
                 Err(e) => {
                     otel_warn!("azure_monitor_exporter.auth.get_token_failed", attempt = attempt, error = %e);
+                    self.metrics.borrow_mut().add_auth_failure();
```
cijothomas, line 86: the auth latency is only calculated in success path? That'd skew results. Is that intended?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/client.rs
```diff
@@ -217,8 +234,19 @@ impl LogsIngestionClient {
             .await
             .map_err(Error::network)?;
 
+        let status_code = response.status().as_u16();
+        let elapsed = start.elapsed();
+
+        // Record per-attempt latency, status code, and batch size
+        {
+            let mut m = self.metrics.borrow_mut();
+            m.add_client_latency(elapsed.as_millis() as f64);
```
cijothomas: incase of network error, we exist early and won't hit this code path, missing reporting latency for such failures..

gouslu: I think network error is a different concept than http round-trip latency.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/auth.rs
gouslu: yeah, this in fact made me realize that I should also only record success latency on laclient and rename the metrics as "success latency" because it helps analyze performance and failures are not something we want to analyze performance of strictly, rather just record it and fix it.

gouslu: done

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/client.rs
```diff
@@ -217,8 +234,21 @@ impl LogsIngestionClient {
...
             .map_err(Error::network)?;
 
+        let status_code = response.status().as_u16();
+        let elapsed = start.elapsed();
+
+        // Record per-attempt status code and batch size
+        {
+            let mut m = self.metrics.borrow_mut();
+            m.record_laclient_status_code(status_code);
+            m.add_batch_size(body_len as f64);
```
cijothomas, line 244: batch_size won't be reported for Error::network

gouslu: that's OK, batch size especially matters if we can correlate it with 413, which should be impossible in this code base if there's no bug.

gouslu: it also matters for successful requests. if we wanted to track every batch size we created, the code would be in gzip batcher.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
jmacd: (There is a crates/config/src/pipeline/telemetry.rs with a TelemetryConfig and reporting_interval. What I want to avoid is for every component to reinvent standard features. It's true that we want to control reporting interval at the node level, but we shouldn't define a new TelemetryConfig, we'll find a way to override telemetry config at the node level.)

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
```diff
@@ -0,0 +1,494 @@
...
+    #[metric(unit = "{response}")]
+    pub laclient_http_413: Counter<u64>,
+    /// Number of HTTP 429 (rate limited) responses.
+    #[metric(unit = "{response}")]
+    pub laclient_http_429: Counter<u64>,
+    /// Number of HTTP 5xx (server error) responses.
+    #[metric(unit = "{response}")]
+    pub laclient_http_5xx: Counter<u64>,
+    /// Number of successful authentication attempts.
+    pub auth_successes: Counter<u64>,
```
cijothomas: there is auth_success_latency defined as well, which would give the count already. No need of duplicating it with this metric.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
cijothomas: Sorry, I meant "log_entries_too_large" when would this occur?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
```diff
@@ -0,0 +1,494 @@
...
+    pub msg_to_data_count: Gauge<u64>,
+    /// Number of log entries rejected for exceeding the batch size limit.
+    #[metric(unit = "{entry}")]
+    pub log_entries_too_large: Counter<u64>,
+    /// Number of successful heartbeat sends.
+    #[metric(unit = "{heartbeat}")]
+    pub heartbeats: Counter<u64>,
+    /// Number of log records that failed to serialize during transformation.
+    #[metric(unit = "{record}")]
+    pub transform_failures: Counter<u64>,
```
cijothomas, line 92: if transform failures do occur, would we still increment the "failed_rows" metrics? If I were to check at total failed log record count, which metric would that be? will I be required to add up transform_failures + failed_rows ?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/client.rs
cijothomas: if it is a conscious choice to track batch_size only for successful case, please modify the metric description to indicate so.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
gouslu: good catch!

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
gouslu: failed rows are for the rows we failed to send to LA, transform failures is before that, they get rejected outright.

gouslu: what I mean is that if something cannot even be transformed into a "row", then they can't be failed to be sent.

cijothomas: Ya I assumed so. Now I am not sure which metric would I use to find how many logs are lost.. 
If the transform failed, do we NACK back?
if the export failed, do we NACK back? 
Depending on these, the retry/durable buffer before exporter does retries. 
(Sorry just writing my mind, not clear what is the solution at the moment)

jmacd: #2018 I don't think we should worry about this level of detail for every component.

cijothomas:
A lot of changes to review, I think it is better to merge it in - the replacement of stdout metrics with proper one itself is a huge improvement! Let us follow up on addressing issues in follow up, short PRs.

cijothomas (approved):
Not reviewed each metric for accuracy- there are some concerns, Better to address them in follow up. The switch from stdout printing of metrics to proper reporting is good direction, and approving to unblock progress.

PR #2113, opened 2026-02-25
Add extension support to the pipeline engine

utpilla:
# Change Summary

Introduces first-class extension support into the dataflow pipeline engine. Extensions are non-pipeline components that provide cross-cutting capabilities (e.g., authentication, health checks, service discovery) to receivers and exporters without participating in the `pdata` flow.

### Motivation
Pipeline components like receivers and exporters often need shared services — credential management, token refresh, header validation — that don't fit the receiver → processor → exporter data-flow model. Extensions provide a clean separation: an independent task produces service handles, and pipeline components consume them at startup via a type-safe registry.

### What's included
Engine core
- `ExtensionWrapper` — Unified wrapper supporting both `Send` and `!Send` extension implementations, analogous to ReceiverWrapper/ExporterWrapper.
- `local::Extension` / `shared::Extension` traits — Lifecycle trait with a `start()` method that receives a control channel and effect handler.
- `ExtensionConfig` — Runtime configuration (control channel capacity) for extensions. Extensions only receive control messages, no pdata channels.
- `ExtensionControlMsg` — PData-free control message enum (`Shutdown`, `TimerTick`, `CollectTelemetry`).
- `ExtensionFactory` — Factory struct (not generic over PData) registered via `#[distributed_slice]`.
- `ExtensionHandles` / `ExtensionRegistryBuilder` / `ExtensionRegistry` — Type-safe, `Clone + Send` registry. Extension factories register typed handles; pipeline components retrieve them by `(extension_name, TypeId)` at startup.
- `ServerAuthenticator` / `ClientAuthenticator` traits — Pluggable auth contract for receivers (validate incoming requests) and exporters (attach outgoing credentials), with cloneable handle wrappers (`ServerAuthenticatorHandle`, `ClientAuthenticatorHandle`).
- **Pipeline lifecycle integration** — Extensions are created before other nodes and started before the pipeline, ensuring handles are available when components call start(). They shut down after pipeline components.
- **Error variants** — ExtensionHandleAlreadyRegistered, ExtensionHandleNotFound, UnknownExtension.

Engine macros (engine-macros)
- `#[pipeline_factory]` macro now generates an `EXTENSION_FACTORIES` distributed slice and a `get_<prefix>_extension_factory_map()` helper.

Config (config)
- `NodeKind::Extension` recognized in URN parsing (`:extension` suffix).
- Extensions excluded from connectivity pruning (they have no data-flow edges).

Pipeline components (`otap`, `contrib-nodes`, `validation`, `benchmarks`)
- All receiver/exporter `start()` signatures now accept `ExtensionRegistry` as a third parameter.
- Existing components pass `_extension_registry` (unused) — no behavioral changes.

### What's NOT included

No concrete extension implementations are shipped yet (no entries in OTAP_EXTENSION_FACTORIES).

* Closes #NNN

Yes


rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,312 @@
...
+    /// The extension task runs independently, processing only control messages.
+    /// Unlike pipeline node wrappers, extensions do not receive a `PipelineCtrlMsgSender`
+    /// — they manage their own timers via `tokio::time` if needed.
+    pub async fn start(self, metrics_reporter: MetricsReporter) -> Result<(), Error> {
+        match self {
+            ExtensionWrapper::Local {
+                node_id,
+                extension,
+                control_receiver,
+                ..
```
lalitb, line 309: We seems to be dropping the `control_sender` here - which means after this method executes, there is no way to send shutdown or any other control message to running extension task.

rust/otap-dataflow/crates/engine/src/lib.rs
```diff
@@ -494,11 +556,61 @@ impl<PData: 'static + Clone + Debug> PipelineFactory<PData> {
...
+                        control_channel_capacity,
+                    )
+                },
+            )?;
+            extensions.push(extension_wrapper);
+        }
+        // Extract handles from each extension and merge into the registry builder.
+        for ext in &mut extensions {
+            if let Some(handles) = ext.take_handles() {
+                extension_registry_builder.merge(ext.name(), handles)?;
```
lalitb: This is merging the plugin URN, not the config node name. which means, we can't have the two extension of same type configured:
Eg:
```yaml
    nodes:                                                                    
      auth_for_team_a:                                                        
        type: "extension:bearer_auth"                                         
      auth_for_team_b:                                                        
        type: "extension:bearer_auth"   
```

lalitb:
> Pipeline components like receivers and exporters often need shared services 

Receivers and exporters both receive `ExtensionRegistry`, but processors do not. Is this intended, and to be added later ? As there could be the real-world use-cases where processors would need extension access (I believe Go collector also support this).

rust/otap-dataflow/crates/engine/src/local/extension.rs
```diff
@@ -0,0 +1,126 @@
...
+        self.ctrl_rx.recv().await
+    }
+}
+
+/// A `!Send` effect handler for extensions.
+///
+/// Provides a minimal set of capabilities — primarily node identity and logging.
+/// Extensions that need periodic timers should use `tokio::time::interval` directly.
+#[derive(Clone)]
+pub struct EffectHandler {
```
lalitb, line 96: In local mode, extensions and pipeline nodes share a single `LocalSet` thread, so anything that blocks between `.await` points - sync I/O, heavy crypto, `thread::sleep` - will stall the whole pipeline silently. Probably worth documenting the non-blocking requirement on the `Extension` trait so implementors know upfront.                                                                                      
                                                                                                                                                                                                                
  Also noticed `EffectHandler` doesn't have a `spawn_blocking` helper. Authors who need to run blocking work will either reach for `tokio::task::spawn_blocking` directly (works, but not discoverable) or block the thread without realising. Something like:                                                                                                                                
                  
  ```rust
  pub async fn spawn_blocking<F, R>(&self, f: F) -> R
  where
      F: FnOnce() -> R + Send + 'static,
      R: Send + 'static,
  {
      tokio::task::spawn_blocking(f)
          .await
          .expect("blocking task panicked")
  }
```

  would make the safe path obvious. One thing to note - !Send fields can't cross into the closure, so callers need to extract/clone before passing in, might be worth a doc note on the method.

File: rust/otap-dataflow/crates/engine/src/local/extension.rs [hunk 4]
```diff
@@ -0,0 +1,126 @@
+// Copyright The OpenTelemetry Authors
```
gouslu, line 1: do we need local/shared versions of extensions? both here and in the extensionwrapper? it seems like we already use arc<mutex> for cloning and sync support anyway, and extensions are send only. so maybe we can just make it so that extensions don't have this separation?

rust/otap-dataflow/crates/engine/src/extensions/auth.rs
```diff
@@ -0,0 +1,503 @@
...
+    /// Returns the headers to attach to an outgoing request.
+    ///
+    /// Each entry is a `(header_name, header_value)` pair. The exporter
+    /// inserts them into the request's header map before sending.
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`AuthError`] if credentials are unavailable
+    /// (e.g., token not yet refreshed, provider unreachable).
+    fn get_request_metadata(&self)
```
gouslu, line 142: does client here mean clients that use http headers? I think something that is more agnostic and focuses more on atomic functionality could be more widely useful. something like I did in my pr -> BearerTokenProvider or sth like that, that returns bearer token. How the consumer uses it is none of our concern. This is also very beneficial if consumer wants to have access to stuff like expiration date  of bearer token etc easily.

lquerel:
> > Pipeline components like receivers and exporters often need shared services
> 
> Receivers and exporters both receive `ExtensionRegistry`, but processors do not. Is this intended, and to be added later ? As there could be the real-world use-cases where processors would need extension access (I believe Go collector also support this).

@lalitb @utpilla, I second this. In my view, all node types should be able to access extensions. However, before we can get there, we first need to introduce an `init` method (or extend the constructor function used by the factories) in our `Receiver`, `Processor`, and `Exporter` traits. That will also solve quite a few issues along the way and will allow us to pass the `ExtensionRegistry`, including to processors. I think this can be introduced in a separate PR.

lquerel:
@utpilla First feedback, given that I haven't read the entire PR. I like the idea of reusing the `#[distributed_slice]` concept and factories for extensions. It's clean and extensible. However, I didn't see any integration with our configuration model. In the configuration for a specific pipeline, how can I specify that I want to instantiate a specific implementation and configure the extension that is compatible with the `ServerAuthenticator` or `ClientAuthenticator` trait?

Before continuing the review, I'd really like to see a concrete example of an extension configuration (in our YAML files) and how it hooks up to, for example, a receiver.

rust/otap-dataflow/crates/engine/src/extensions/auth.rs
```diff
@@ -0,0 +1,503 @@
...
+}
+
+// ─── Client side (exporters attach outgoing credentials) ───────────────────
+
+/// Trait for producing credentials to attach to outgoing requests.
+///
+/// Implement this trait in an auth extension to provide client-side
+/// authentication. The extension decides what headers to attach
+/// (e.g., `Authorization: Bearer <token>`, custom API key headers).
+pub trait ClientAuthenticator: Send {
```
gouslu, line 132: can async traits (funcs) be supported in this pattern?

rust/otap-dataflow/crates/engine/src/lib.rs
utpilla: Thanks for catching that! Fixed in https://github.com/open-telemetry/otel-arrow/pull/2113/changes/bb447141ca3cc27f0dcdc4b1f0324738facc458a.

rust/otap-dataflow/crates/engine/src/extension.rs
utpilla: Thanks! Fixed in https://github.com/open-telemetry/otel-arrow/pull/2113/changes/9dda9c479813ef85c570134654ecd27f0e5855dd

rust/otap-dataflow/crates/engine/src/local/extension.rs
utpilla: This issue applies equally to all node types (receivers, processors, exporters). They all share the same single-threaded runtime. None of them currently provide a `spawn_blocking` helper. I think documenting the non-blocking contract and potentially adding a `spawn_blocking` helper would be better as a follow-up that covers all node types uniformly, not just extensions.

rust/otap-dataflow/crates/engine/src/local/extension.rs
utpilla: The extension's service handles are `Arc`-based and `Send + Sync`, but the extension implementation itself can hold `!Send` internal state. This mirrors the pattern used by receivers, processors, and exporters. They all have local/shared variants. Since the engine runs on `current_thread` + `LocalSet`, `!Send` is the natural default. Removing the local variant would force extension authors to add unnecessary `Send` boilerplate for state that never leaves the thread. I'd prefer to keep it for consistency and flexibility.

gouslu:
Please don't merge this PR without my approval. I have been working on an extension system as well and I have a different opinion on how I think extensions should be implemented. I think the core idea is very similar in many cases, but I would like us to work together on this feature @utpilla.

gouslu:
@utpilla this is what I have put together as an alternative -> https://github.com/open-telemetry/otel-arrow/pull/2141

utpilla:
> However, I didn't see any integration with our configuration model. In the configuration for a specific pipeline, how can I specify that I want to instantiate a specific implementation and configure the extension that is compatible with the `ServerAuthenticator` or `ClientAuthenticator` trait?
> 
> Before continuing the review, I'd really like to see a concrete example of an extension configuration (in our YAML files) and how it hooks up to, for example, a receiver.

@lquerel You could check this diff to get an idea of how a sample config could look like on both receiver and exporter end: https://github.com/utpilla/otel-arrow/pull/3

jmacd:
Closing in favor of #2293 design, we will incorporate elements of this design as we go ahead. Thanks @utpilla.

PR #2118, opened 2026-02-26
Fix eventname for OTAP nodes to include identifying prefix

cijothomas:
# Change Summary

Prefix node name to help identify events properly, as all nodes have same crate name.

Local run.

yes better event names.


rust/otap-dataflow/crates/otap/src/otap_receiver.rs
```diff
@@ -237,7 +237,7 @@ impl shared::Receiver<OtapPdata> for OTAPReceiver {
         effect_handler: shared::EffectHandler<OtapPdata>,
     ) -> Result<TerminalState, Error> {
         otap_df_telemetry::otel_info!(
-            "receiver.start",
+            "otap.receiver.start",
```
utpilla: Let's be consistent in the naming convention. Either use `_` or `.`.

We already have the `_` pattern for `syslog_cef_receiver`, `fake_data_generator`.

rust/otap-dataflow/crates/otap/src/tls_utils.rs
```diff
@@ -366,7 +366,7 @@ async fn add_system_trust_anchors_if_enabled(
     // Best-effort: accept that some system certs might not parse.
     let (added, ignored) = store.add_parsable_certificates(roots);
     otel_debug!(
-        "loaded.system.ca.certificates",
+        "tls.native_cert.loaded",
```
utpilla, line 369: Was the change in the name intentional? `system.ca.certificates` vs `native_cert`?

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -251,7 +251,7 @@ impl GrpcClientSettings {
         if proxy.has_proxy() && !self.grpc_endpoint.trim_start().starts_with("https://") {
             let proxy_str = proxy.to_string();
             otap_df_telemetry::otel_info!(
-                "proxy.configured",
+                "otap.grpc.proxy.configured",
```
utpilla: Would it better to use `otap_grpc_exporter` instead to group all the OTAP gRPC exporter logs together?

jmacd (approved):

rust/otap-dataflow/crates/otap/src/tls_utils.rs
cijothomas: Yes. The other events in this module used tis.native_cert*. I can revert if this is not correct.

jmacd (approved):

PR #2121, opened 2026-02-26
perf: Optimizing sorting by id/parent id

### LINKED ISSUE #2122: Optimize id column sorts
[sort_by_parent_then_id](https://github.com/open-telemetry/otel-arrow/blob/45c3ea7b3beee63354c2c362f5e1f45cd091e017/rust/otap-dataflow/crates/pdata/src/otap/transform/util.rs#L128)  sorts record batches by parent_id then id before the split transform. It currently
delegates all sorting to Arrow's RowConverter via sort_to_indices. This has two inefficiencies:

  1. Dictionary expansion: RowConverter fully expands dictionary-encoded columns into their row representation.
  Since parent_id is frequently dictionary-encoded (Dict<UInt8, UInt16/UInt32>), this is wasteful — the underlying
  values are small unsigned integers that can be resolved with a simple index lookup.
  2. No early return for sorted data: OTAP data is often already sorted by parent_id/id. The current implementation
  always runs the full sort pipeline even when the input is already in order.

We can optimize this by specializing the sort implementation based on the number and types of columns:

- Two-column sort: Pack resolved parent_id (high 32 bits) and id (low 32 bits) into a Vec<u64> and sort that
  directly. This avoids RowConverter entirely - dictionary values are resolved inline via values[keys[i]], and the
  packed u64 comparison is a single integer operation per step.
- Single-column sort: Add an is_sorted check before delegating to arrow::compute::sort_to_indices. For dictionary
  columns, resolve through keys→values to check sorted order.
- Already-sorted early return: Check is_sorted on the packed keys (two-column) or raw values (single-column) and
  return the batch unchanged when already in order.

JakeDern:
# Change Summary

This PR optimizes some id sorting logic that we have for split using techniques for two column sorts that we also use elsewhere in the transport optimize code.

| Scenario                               | Baseline (RowConverter) | Optimized | Speedup |
  |----------------------------------------|-------------------------|-----------|---------|
  | u16 id only (random)                   | 7.23 us                 | 7.27 us   | ~same   |
  | u16 id only (sorted)                   | 1.20 us                 | 71 ns     | 17x     |
  | u16 pid + u16 id (native)              | 25.4 us                 | 7.33 us   | 3.5x    |
  | u32 id + Dict<u8,u32> pid             | 28.7 us                 | 7.89 us   | 3.6x    |
  | u32 id + Dict<u16,u32> pid            | 28.3 us                 | 7.93 us   | 3.6x    |
  | u32 id (25% null) + Dict<u8,u32> pid  | 30.4 us                 | 10.6 us   | 2.9x    |

* Closes #2122

New unit testing suite comparing results with the old and also indirectly covered by all the split tests.

No


rust/otap-dataflow/crates/pdata/src/otap/transform/util.rs
```diff
@@ -527,3 +676,166 @@ pub(crate) fn payload_to_idx(payload_type: ArrowPayloadType) -> usize {
...
+            ("id", UInt16, [4, 3, 2, 1])
+        )
+        .unwrap();
+        assert_matches_oracle(rb);
+    }
+
+    #[test]
+    fn test_dict_16_u32() {
+        let rb = record_batch!(
+            ("parent_id", (UInt8, UInt32), ([1, 0, 1, 0], [10, 20])),
```
ThomsonTan:
```suggestion
            ("parent_id", (UInt16, UInt32), ([1, 0, 1, 0], [10, 20])),
```

JakeDern: Ah good catch, thanks

rust/otap-dataflow/crates/pdata/src/otap/transform/util.rs
```diff
@@ -121,42 +124,188 @@ pub(crate) fn sort_otap_batch_by_parent_then_id<const N: usize>(
...
+/// sort each independently, then merge them with nulls_first semantics for id.
+fn sort_two_columns_packed<T: ArrowNativeType>(
+    rb: RecordBatch,
+    resolve_pid: impl Fn(u32) -> u64,
+    id_col: &dyn Array,
+    id_vals: &[T],
+) -> Result<RecordBatch> {
+    let len = id_vals.len();
+
+    let pack = |i: u32| -> u64 { (resolve_pid(i) << 32) | id_vals[i as usize].as_usize() as u64 };
```
albertlockett, line 268: nit: I'm not sure if it'll make a huge difference, but if both the ID/parent_id were u16, we could pack them into a u32, then I guess we'd maybe be able to fit more rows of the packed arrays in the CPU cache, which might make sorting faster.

I realize this significantly complicates things, and might not be worth it, so feel free to ignore if you don't feel like trying this :)

JakeDern: Nothing slips past you 🙂 - I thought about this, but then I thought that a next optimization would be to re-use the packed vector across every call since we're usually sorting a whole otap arrow records in a loop when calling from `sort_by_parent_then_id`. So I thought maybe it would be faster to just use the same Vec<u64> for everything rather than use one of each. Although maybe we can just cast the `Vec<u64>` into a `Vec<u32>` somehow?

I didn't implement that yet, I kind of wanted to profile it first since that would require the caller to pass in the scratch space or otherwise have us create some kind of `RecordBatchIdSorter` or something.

albertlockett: Sounds good. Btw, for profiler I highly recommend [`samply`](https://github.com/mstange/samply), which can be used when running benches like:

```
samply record cargo bench -p benchmarks --bench "sort_by_parent_then_id" -- "<benchmark group,id, etc>"
```

JakeDern: Thanks for the tip, I was going to ask you at some point what you were using for a profiler! I've had mixed luck with perf + cargo flamegraph, the generation of the visual takes forever for code that pulls in a lot of deps which criterion does

albertlockett: Also need to adjust these when using samply:
https://github.com/open-telemetry/otel-arrow/blob/05a881850552252d5d19119ab26ae61c0362082f/rust/otap-dataflow/Cargo.toml#L293-L304

want something like:
```yaml
[profile.bench]
inherits = "release"
opt-level = 3
# debug = false <--- want debug symbols
incremental = false
# lto = "fat" <---- makes build slow when iterating
codegen-units = 1
# Potentially override lto for faster builds:
lto = "thin"
# Potentially override debug/strip for easier debugging:
debug = true
strip = "none"
```

rust/otap-dataflow/crates/pdata/src/otap/transform/util.rs
```diff
@@ -121,42 +124,188 @@ pub(crate) fn sort_otap_batch_by_parent_then_id<const N: usize>(
...
+}
+
+/// Two-column sort. Packs parent_id and id into u64 sort keys for a single-pass sort.
+fn sort_two_columns(
+    rb: RecordBatch,
+    parent_id_col: &ArrayRef,
+    id_col: &ArrayRef,
+) -> Result<RecordBatch> {
+    if parent_id_col.null_count() != 0 {
+        return Err(Error::NullParentId);
```
lquerel, line 203: Question: What do we do with malformed batches that, for example, have null parent IDs, and more importantly, what impact do they have on the batch processor and on the pipeline as a whole?

For now, I would say that at a minimum we should increment a metric (maybe it's already done, I didn't check), and in the future we should be able to define an optional policy at the configuration level to route invalid pdata to a DLQ or something similar.

rust/otap-dataflow/crates/pdata/src/otap/transform/util.rs
```diff
@@ -121,42 +124,188 @@ pub(crate) fn sort_otap_batch_by_parent_then_id<const N: usize>(
...
+/// Sorts a record batch by `parent_id` (primary) then `id` (secondary) columns.
+///
+/// Dispatches to specialized sort paths based on the column types:
+/// - No sort columns: returns batch unchanged.
+/// - Single column (native): sort via the normal arrow sort_to_indices with an
+///   early return check if we're already sorted.
+/// - Two columns (both native): Sort via a specialized two column sort that packs
+///   parent_id into the high bits and id into the low bits of a u64 and sorts that
+///   instead.
+pub fn sort_by_parent_then_id(rb: RecordBatch) -> Result<RecordBatch> {
```
lquerel, line 136: Why make this function public?

rust/otap-dataflow/crates/pdata/src/otap/transform/util.rs
JakeDern: >Question: What do we do with malformed batches that, for example, have null parent IDs, and more importantly, what impact do they have on the batch processor and on the pipeline as a whole?

I think it depends on which component the data runs through and whether or not it interferes with whatever it was trying to do.

In general for the batching code, I'm trying to take a pretty defensive stance against malformed data and either deal with it if I can do so reasonably or sometimes reject it like in this case. If we return errors from batching code, the batch processor does currently bump a counter: 

https://github.com/open-telemetry/otel-arrow/blob/05a881850552252d5d19119ab26ae61c0362082f/rust/otap-dataflow/crates/otap/src/batch_processor.rs#L856C1-L872C1

But I'm really not sure how the whole pipeline holds up against edge cases/spec violations like:

- Missing Id columns (just means there's nothing to join to)
- Primary Id columns with duplicate values 
- Data with referential integrity violations
- Data that's missing required columns

It's on my near-term list to investigate that more thoroughly.

rust/otap-dataflow/crates/pdata/src/otap/transform/util.rs
```diff
@@ -121,42 +124,188 @@ pub(crate) fn sort_otap_batch_by_parent_then_id<const N: usize>(
...
+
+    // Precompute packed keys for sorting. Only valid row slots are populated
+    let mut packed = vec![0u64; len];
+    for &i in &valid_indices {
+        packed[i as usize] = pack(i);
     }
+    valid_indices.sort_unstable_by_key(|&i| packed[i as usize]);
+
+    // Merge: null-id rows first at equal parent_id (nulls_first for id column).
+    let mut indices: Vec<u32> = Vec::with_capacity(len);
```
lquerel, line 285: I'm wondering if we could avoid allocating new `Vec` on every call to this function for `packed` and `indices`, by taking advantage of the fact that we're in a single threaded environment.
We could pass a SortBuffers to this function allowing us to reuse those allocations.

JakeDern: Yes - I had the same idea 🙂: https://github.com/open-telemetry/otel-arrow/pull/2121#discussion_r2865316622

Going to profile and see if we want to change the interface for that

rust/otap-dataflow/crates/pdata/src/otap/transform/util.rs
JakeDern: I had it this way to benchmark it - Maybe I can define some kind of feature that selectively re-exports it through the parent module

Issue #2058, opened 2026-02-18
Columnar query engine expression evaluation


albertlockett:
Currently the only kind of expressions the columnar query engine really evaluates are those which do comparisons with literals and produce boolean values. We use these for filtering and if/else conditions.

There are various other types of expressions we need to support, such as Math, comparing with non literals, function calls, etc.

For example, imagine an expression like the following (assume that + concatenates strings).
```js
"[" + resource.attributes["service.name"]  + " " + format(time_unix_nano) + "] ERROR: "   
  +  coalesce(attributes["exception.type"], "Unknown") + attributes["exception.message"] 
```

This might produce a string value like:
```
[df_engine 2026-02-18T20:30:32Z] ERROR: EngineError proto decode error
```

We should be able to evaluate expressions such as this.

There are a few challenges to overcome. Below are some rough implementation notes that should eventually be formalized by a spec:

**Null handling semantics**

Most fields in OTAP are nullable, so we need to have a plan for handling nulls. I feel that nulls should propagate. For example:
```
attributes["x"] + attributes["y"]
```
If `attribtues["x"]` or `attributes["y"]` are null, the result of this operation should be null.

We may consider making an exception for comparison operators like `==` and `!=`, where we consider `null == null` true and `null == <not null>` to be false, etc.

**Type mismatches**

Again consider:
```
attributes["x"] + attributes["y"]
```
This should fail to evaluate and produce an error if these fields are both not null but do not have the same type. In the future, we may consider adding capability to either cast or explicitly type check for users who wish to avoid this default behaviour.

**Transparent Joins**

It would be _nice_ if we could use the capabilities of datafusion's expression evaluation for this feature, but it presents a challenge. 

Again consider:
```
attributes["x"] + attributes["y"]
```

This isn't just a simple string concat -- we actually need to do something like:
```js
// pesudocode
let attrs_x = `select * from attributes where key = 'x'`;
let attrs_y = `select * from attributes where key = 'y'`;

let result = `
  select concat(attrs_x.value, attrs_y.value) 
  from attrs_x join attrs_y on attrs_x.parent_id == attrs_y.parent_id`;
```
(In reality it's actually more complicated than this, because we also need to check the values types and do null handling).

I think what we probably need to do is break our expression into into different "data domains" - e.g. segments of record batches that we can evaluate using datafusion expressions, and then manage "joining" the results of the expressions for each domain ourselves.

[...]

PR #2126, opened 2026-02-27
Columnar query engine expression evaluation: simple arithmetic

albertlockett:
# Change Summary

Adds a module to the columnar query engine with the ability evaluate simple arithmetic expressions on OTAP record batches. For example, it could evaluate expressions such as `severity_number + attributes["x"] * 2`. 

Note that the expression evaluation isn't yet integrated into any `PipelineStage` implementation, but the intention is that this can soon be used to implement more advanced filtering, attribute insertion, column updates and so on.

While on the surface, this isolated simple arithmetic not appear terribly useful, the main/important contributions in this PR are to lay down some foundations for future expression evaluation:

**Transparent joins in the expression tree**
This PR adds is the ability to evaluate a set of DataFusion expressions while transparently joining data from different record batches as the expression evaluates.

For example, consider we had `severity_number * 2 + attributes["x"] + attributes["y"]`, we'd need to first evaluate:
- 1. `severity_number * 2`
- 2. `attributes where key = "x"`, then select the value column based on the type
- 3. `attributes where key = "y"`, then select the value column based on the type,

Then we need to join these three expressions on the ID/parent ID relationship, then perform the additions.

This PR builds the expression tree in such a way that it can manage where these joins need to happen on either side of a binary expression, and it performs the joins automatically during expression evaluation while keeping track of the ID scope/row order of the current data at each stage.

**Type evaluation and coercion**

While planning the expression, the planner attempts to keep track of what are the possible types that the expression could produce if it were to successfully evaluate. When it detects invalid types, it is able to produce an error indicating an invalid expression.

For example, we'd be able to detect at planning time that `severity_text + 2` is invalid, because text can't be added to a number.

For expressions where we can't determine that the types are invalid at planning time, it will be determined and runtime and an error will be produced when the expression evaluates on some batch. For example `attributes["x"] + 2`, it's unknown whether `attributes["x"]` is an int64, so it's assumed that the expression will produce an int64, and if `attributes["x"]` is found not to be this type, an ExecutionError will be produced.

The planner automatically coerces integer types when necessary. Currently when adding two integers, they will be coerced into the largest type that could contain the value, while keeping the signed-ness of one side. For example, uint8 + int32 will produce an int32. I realize this type of automatic integer coercion is probably controversial, so in the future I'm happy to get rid of this in favour of forcing explicit casting if that is preferred. 

**Missing data / null propagation**

When one side of an expression is null, for the purposes of arithmetic the expression will evaluate to null. This includes the case of null values, missing attributes, missing columns, and missing optional record batches.

For example: `attributes["x"] + 2` would evaluate as null if the attribtues record batch was not present, there were no attributes with `key == "x"`, or the attributes where `key=="x"` had type empty, and so on.

**Relocated the projection code**

Adds a new module called `pipeline::project` which has the projection code that was previously inside the filter module. We need to project the input record batches into a known schema to evaluate the expressions, and also consider the expression evaluation to result in a `null` if the projection could evaluate due to missing data.

* Relates to https://github.com/open-telemetry/otel-arrow/issues/2058

There are 64 new unit tests covering these changes

No

## Future work/followups:
There are many, but most pressing are:
- Other types of expression evaluation, including string expressions, unary math expressions, function invocation and bridging this with the filtering code (for expressions that produce boolean arrays).
- Integrating expression evaluation with various pipeline stages including those which set attributes, set values, and filtering
- OPL Parser support for the type of expressions we're able to evaluate


rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
```diff
@@ -0,0 +1,2534 @@
...
+                        Ok(ScopedLogicalExpr {
+                            logical_expr: col(column_name),
+                            requires_dict_downcast: false,
+                            source: LogicalExprDataSource::DataSource(DataScope::Root),
+                            expr_type: field_type,
+                        })
+                    }
+                    ColumnAccessor::StructCol(column_name, struct_field_name) => {
+                        let field_type =
+                            root_field_type(&struct_field_name).ok_or_else(|| Error::InvalidPipelineError {
```
lquerel: Should you use `root_field_type` or `nested_struct_field_type` here?

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
```diff
@@ -0,0 +1,2534 @@
...
+            LogRecord::build()
+                .severity_text("debug")
+                .severity_number(40)
+                .attributes(vec![KeyValue::new("k1", AnyValue::new_int(7))])
+                .finish(),
+        ]);
+
+        let otap_batch = otlp_to_otap(&OtlpProtoMessage::Logs(logs));
+        let logs_input_2 = otap_batch.get(ArrowPayloadType::Logs).unwrap();
+        // ensure the column isn't in the same location
```
lquerel, line 1903: Isn't or is in the same location?
The comment indicates "isn't" but the code looks like a "is"

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
```diff
@@ -0,0 +1,2534 @@
...
+impl ExprPhysicalPlanner {
+    /// Converts a ScopedLogicalExpr into an executable ScopedPhysicalExpr.
+    fn plan(&self, logical_expr: ScopedLogicalExpr) -> Result<ScopedPhysicalExpr> {
+        logical_expr.into_physical()
+    }
+}
+
+/// A node in the expression tree used for expression evaluation.
+///
+/// This encapsulates a datafusion PhysicalExpr that evaluates some section of of the overall
```
lquerel:
```suggestion
/// This encapsulates a datafusion PhysicalExpr that evaluates some section of the overall
```

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
```diff
@@ -0,0 +1,2534 @@
...
+                result.scope_ids = Some(Arc::clone(scope_ids))
+            }
+        }
+
+        result
+    }
+}
+
+#[cfg(test)]
+mod test {
```
lquerel, line 732: I didn't see it, but maybe I missed that test. Do you have a test that simulates two batches with attributes with the same name but different types across the two batches (attributes used in an expression)?

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
```diff
@@ -0,0 +1,2534 @@
...
+                        };
+
+                        otap_batch
+                            .get(attrs_payload_type)
+                            .map(|rb| {
+                                Self::try_project_attrs(
+                                    rb,
+                                    key.as_str(),
+                                    self.projection_opts.downcast_dicts,
+                                )
```
lquerel, line 466: How does that work when in the same expression we have multiple times the same attribute/field referenced? May be there is here a possible optimization consisting to project only 1 time and reuse the projection each time the same attribute/field appears in the expression.

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
albertlockett: lol, good catch! Fixed in 51cc3ef65. Thanks @lquerel

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
albertlockett: yep, good catch, thanks @lquerel

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
albertlockett: Good idea! added in 0118369a6 . Thanks @lquerel

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
albertlockett: It depends on the expression.

If you were to do something like this `attributes["x"] + attributes["x"]`, it would only get projected once and the. However, if you did something like `severity_type * attributes["x"] - attributes["x"]`, this would require that we evaluate  the multiplication first, then we'd try to join it to join `attributes["x"]` before subtracting, and by virtue of how the expressions are  currently executed, we would unfortunately project `attributes["x"]` a second time.

You're 100% right though that there's an optimization we can/should make to cache the projected attributes in this case.

Issue #2120, opened 2026-02-26
Proposal: extract view traits into standalone zero-dependency leaf crate (otap-pdata-views)


lalitb:
## Problem                                                                                                                                                                                                      
                                                                                                                                                                                                                  
  The view traits in `otap-df-pdata` (`LogsDataView`, `LogRecordView`, `TracesView`, `SpanView`, etc.) are pure trait definitions with zero external dependencies. However they currently live inside a crate that transitively pulls in Arrow, tokio, tonic, roaring and ~30 other dependencies.                                                                                                                                 
                                                                                                                                                                                                                  
  Any downstream consumer that wants to implement or consume these traits - without needing the full Arrow/OTAP processing stack - is forced to take on the entire `otap-df-pdata` dependency graph. This is a significant and unnecessary overhead for lightweight consumers.

  ## Proposed Solution

  Extract the zero-dependency view traits into a new standalone leaf crate `otap-pdata-views` with no external dependencies. `otap-df-pdata` re-exports from it, so existing consumers see no breaking change.

  **Moves to `otap-pdata-views`:**
  - `views/logs.rs` - log view traits
  - `views/trace.rs` - trace/span view traits
  - `views/common.rs` - shared value/attribute view traits
  - `views/resource.rs` - resource view trait
  - `TraceId` / `SpanId` type aliases

  **Stays in `otap-df-pdata` unchanged:**
  - `views/metrics.rs` - the trait definitions are zero-dep but the file also contains `From<proto::...>` conversion impls that couple it to proto types. Moving metrics is deferred to a follow-up once those conversions are relocated to the proto adapter layer.  
  - `views/otap/` - Arrow backend implementations
  - `views/otlp/` - proto/bytes backend implementations

  ## Motivation

The immediate use case is `geneva-uploader` in `opentelemetry-rust-contrib`, which is being extended to support pluggable encoder formats in addition to the existing OTLP path. The existing OTLP ingestion support remains unchanged. The extension adds the ability for custom sources - such as MsgPack - to encode directly to proprietary `Bond` format without going through an OTLP intermediate representation. To enable this, `geneva-uploader` needs to optionally depend on `LogsDataView`/`TracesView` as part of its encoder trait contract. Adding full `otap-df-pdata` as a dependency to a lightweight transport library is not viable. A zero-dependency trait crate is.

[...]

PR #2130, opened 2026-02-27
feat: extract otap-df-pdata-views as a standalone zero-dependency leaf crate

lalitb:
Fixes: #2120 

## Summary                                                                                                                                                                                                         
                                                                                                                                                                                                                     
  Extracts the backend-agnostic view traits for logs, traces, resource, and common types from                                                                                                                        
  `otap-df-pdata` into a new standalone crate `otap-df-pdata-views`.                                                                                                                                                    
                                                                                                                                                                                                                     
  The motivation is to make these traits consumable outside the otel-arrow ecosystem - for example,
  by exporters like `geneva-uploader` in `opentelemetry-rust-contrib` - without pulling in the full                                                                                                                  
  `otap-df-pdata` dependency stack. See discussion in #2120.                                                                                                                                                         
                                                                                                                                                                                                                     
  ## Changes                                                                       

  - **New crate** `crates/pdata-views/` - contains `views::common`, `views::logs`, `views::trace`,
    `views::resource` with zero external dependencies
  - `TraceId` and `SpanId` type aliases are defined at the crate root
  - All consumers within `otap-df-pdata` (otlp/proto, otlp/bytes, otap, encode, payload) updated
    to import directly from `otap_pdata_views` - the re-export shim layer has been removed
  - External crates (`otap`, `telemetry`, `contrib-nodes`, `benchmarks`) updated to import view
    traits directly from `otap_pdata_views`
  - `#[non_exhaustive]` added to `ValueType` for forward compatibility

  ## Out of scope

  **Metrics view traits are intentionally excluded** for now - the metrics view hierarchy is
  significantly more complex and can be extracted in a follow-up once the pattern is validated
  for logs and traces.

  ## Verification

  - `cargo tree -p otap-pdata-views` confirms zero external dependencies
  - All existing tests in `otap-df-pdata` pass unchanged

### DIFF
--- a/rust/otap-dataflow/crates/pdata-views/README.md
+++ b/rust/otap-dataflow/crates/pdata-views/README.md
@@ -4,22 +4,26 @@ Zero-dependency, backend-agnostic view traits for OTLP/OTAP telemetry data.
 
 ## Overview
 
-This crate provides read-only view traits for traversing hierarchical telemetry data structures
-(logs, traces, resources) without any external dependencies. It is designed to be consumed both
-within the `otap-dataflow` workspace and by external crates that need a lightweight integration
-point without pulling in the full `otap-df-pdata` stack.
+This crate provides read-only view traits for traversing hierarchical
+telemetry data structures (logs, traces, resources) without any external
+dependencies. It is designed to be consumed both within the
+`otap-dataflow` workspace and by external crates that need a lightweight
+integration point without pulling in the full `otap-df-pdata` stack.
 
 ## Traits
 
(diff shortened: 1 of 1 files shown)


lquerel:
@lalitb we can't merge because there is an issue with one of the required check.

lalitb:
> @lalitb we can't merge because there is an issue with one of the required check.

@lquerel  - should be good now. Had to change the crate name from `otap-pdata-views` to `otap-df-pdata-views` to have the structure CI check pass.

PR #2141, opened 2026-03-01
[Prototype] Pipeline Components - Extension Support 

gouslu:
## Pipeline Components - Extension Support Proposal

Extensions are pipeline components alongside receivers, processors, and exporters -- but they do not participate in data-path connections (no PData channels, control messages only).

Extensions can be implemented as local (`!Send` futures, single-threaded `LocalSet`) or shared (`Send` futures, multi-threaded), giving extension authors flexibility with some caveats based on the `Send` only extension traits in this design.

Extensions can optionally implement extension traits (e.g., `BearerTokenProvider`) defined in the engine crate to expose capabilities to other components, or opt out and serve purely as background tasks.

Extension traits require `Send + Clone + 'static`. If an extension publishes any trait, the concrete type implementing that trait must satisfy these bounds -- meaning shared state must be managed via `Arc` (or similar) by the extension author.

The extension's lifecycle method (`start`) takes `Box<Self>` by move -- the extension instance is consumed, not cloned. What is cloned is the extension *struct itself* during `extension_traits()`: the macro clones `self` into each `TraitRegistration`, and those clones are inserted into the `ExtensionRegistry`. After trait collection, the original extension is consumed by `start()`. Each consumer (receiver/exporter) receives a clone of the registry, and calling `registry.get::<dyn Trait>(name)` returns a fresh clone of the stored trait object.

Extensions are configured as a sibling to `nodes` in the pipeline YAML (dedicated `extensions:` section), not inside `nodes`.

Extension trait types are sealed -- new trait types can only be added inside the engine crate; external crates can implement existing traits but cannot define new ones.

Extensions are started first, before exporters, processors, and receivers, so their capabilities are available at component initialization.

Extension registry is passed via start methods of exporters and receivers. Can be added to "process" as well if needed. Considered putting it into the EffectHandler, but it didn't feel like the right place for that.

Created a separate control message channel for Extensions based on the changes that were done by @utpilla 's PR https://github.com/open-telemetry/otel-arrow/pull/2141.

### Alternatives Considered

- **`Arc`-based registry** -- true single instance with `Arc` clones, but enforces a `Sync` boundary on all extensions.
- **`Rc`-based registry** -- only works for local components. It might be acceptable to say that shared components don't have extension support, but this limits flexibility.
- All other alternatives seem to require at least one deep clone or multiple instances and never achieve a true single instance. For this reason I opted for a design where deep clone is accepted and shared extension state is the responsibility of the extension author via an explicit `Clone` requirement.

### Why Local and Shared Variants

Local and shared extension variants exist for two reasons:

1. **Consistency** -- receivers and exporters already have local/shared variants; extensions follow the same pattern.
2. **Event-loop optimizations** -- despite extension *traits* requiring `Send` (because the registry is `Send`), the `start()` async body of a local extension can use `!Send` types (`Rc`, `RefCell`, `LocalSet` spawning, etc.). This means a local extension that publishes `Send` traits can still use non-thread-safe optimizations inside its own event loop, even though its struct fields must be `Send + Clone` for trait registration.

### DIFF
--- a/docs/extension-system-architecture.md
+++ b/docs/extension-system-architecture.md
@@ -119,6 +119,76 @@ no wasted resources.
    pure background tasks simply use the default (empty)
    implementation and never appear in the registry.
 
+### State Management: Clone + Arc
+
+Extensions use the **Clone + Arc** pattern for shared state
+management. This is the same convention used by widely-adopted
+Rust libraries including:
+
+- **[axum](https://docs.rs/axum/latest/axum/extract/struct.State.html#shared-mutable-state)**
+  (25k stars, by the tokio team) — documents this pattern
+  explicitly:
+
+  > *"As state is global within a Router you can't directly
+  > get a mutable reference to the state. The most basic
+  > solution is to use an `Arc<Mutex<_>>`. Which kind of
+  > mutex you need depends on your use case. See the tokio
+  > docs for more details."*
+
+- **tonic** (gRPC framework, by the tokio team) — services
+  must be `Clone`; shared state is wrapped in `Arc`.
+- **tower** (`Service` trait) — clone-per-request model;
+  documentation recommends `Arc` for shared state.
+- **reqwest::Client**, **kube::Client**, **AWS SDK clients**
+  — all use `Clone` wrapping an `Arc<Inner>`.
+
+In our extension system, the extension struct is `Clone`.
+When cloned into the capability registry and distributed to
+consumers, all clones share the same underlying state via
+`Arc`. No per-clone copies are made of shared resources
+like credentials or broadcast channels.
+
+**Example — shared state via `Arc`:**
+
+```rust
+#[derive(Clone)]
+pub struct MyExtension {
+    // Shared across all clones — Arc is the sharing primitive
+    credential: Arc<dyn TokenCredential>,
+    token_sender: Arc<watch::Sender<Option<BearerToken>>>,
+
+    // Cheap to clone — small owned values
+    scope: String,
+    method: AuthMethod,
+}
+```
+
+**Interior mutability without `Mutex`:**
+
+For hot-path operations, `Arc<Mutex<_>>` is not the only
+option. The following `Send`-compatible primitives provide
+lock-free interior mutability:
+
+| Primitive | Cost | Use case |
+|---|---|---|
+| `AtomicU64` | ~1ns | Counters, flags |
+| `ArcSwap<T>` | ~2ns read | Swapping configs, token caches |
+| `watch::Sender` | ~3ns read | Push-based state updates |
+| `DashMap<K,V>` | ~10-15ns | Concurrent hash maps |
+
+These are preferred over `Mutex` for extension state that
+is read frequently on the hot path. `Mutex` (~15ns) is
+acceptable for state accessed infrequently (e.g., config
+updates).
+
+**Why `Send` only, not `Rc`/`RefCell`:**
+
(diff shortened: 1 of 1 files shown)


docs/extension-system.md
```diff
@@ -0,0 +1,422 @@
...
+3. **Local/Shared split.** Like receivers and exporters, extensions have both
+   local (`!Send` futures) and shared (`Send` futures) variants. Local
+   extensions run on the single-threaded `LocalSet`; shared extensions can be
+   spawned on multi-threaded runtimes. The `ExtensionWrapper` enum abstracts
+   over both variants.
+
+4. **Registry-based lookup.** Receivers and exporters receive an
+   `ExtensionRegistry` at `start()` and look up extensions by name and trait.
+   Processors do not receive the registry (they don't need cross-cutting
+   capabilities directly).
```
jmacd: nit: Eventually, we will find ways to use processor extensions. In the Collector, we find cross cutting concerns like memory limiters and persistent key/value stores.

docs/extension-system.md
```diff
@@ -0,0 +1,422 @@
...
+```
+
+### Key Design Decisions
+
+1. **Extensions start first.** The engine spawns extensions before receivers,
+   processors, and exporters so their capabilities are available when
+   data-path components initialize.
+
+2. **PData-free.** Extensions are completely decoupled from the pipeline data
+   type (`PData`). They receive their own `ExtensionControlMsg` messages
```
jmacd: Note: the admin component is in a similar position, needing to be decoupled from the PData type of the engine, yet interoperating with the engine.

docs/extension-system.md
```diff
@@ -0,0 +1,422 @@
...
+| Requirement            | Reason                                           |
+|------------------------|--------------------------------------------------|
+| `Clone`                | Registry stores clones; `get()` returns clones   |
+| `Send`                 | Registry is `Send`; extensions may cross threads |
+| `'static`              | Required for `Any`-based type erasure            |
+| Shared state via `Arc` | Clones must observe the same state               |
+
+Pure background-task extensions (no published traits) have no special
+requirements beyond implementing the `Extension` trait. Local extensions
+can even use `!Send` futures.
```
jmacd: I think we should develop this section a bit more. I know that @lquerel is thinking about extensions for topics, and it raises essentially a new dimension here. We understand that extensions can be local or they can be shared across threads; separately, we understand that an extension could be instantiated per-core or per-engine. In some cases, we may want a single extension instance shared across CPU cores, and in other cases we may want per-engine-core extension instances that are shared only across pipeline threads in a single core.

Since the admin component has already come up in my review, there's another parallel. In many ways the current admin component is like a shared extension across engine threads. After we have topic receivers, we can think of the admin component as a singleton pipeline connected through topic used for collecting internal telemetry.

gouslu: Sounds good, I think this design can meaninfully expand into percore perengine. But every time we start creating variations, we are adding complexity. Based on the feedback I can work on how to incorporate it to this design, if the overal design gets a thumps up review and can update both the code base and the arch document accordingly.

I believe the design supports adding topic-listening and per-engine scope incrementally. Currently extension registries are per core/pipeline and we could look into how we can support per engine extensions as well. Are we interested in every extension to be availabe both in per engine and per core variations? Or do we want to have certain extension types to be per engine, meanwhile certain extension types to be per pipeline?

gouslu: btw, let's continue discussion on this PR https://github.com/open-telemetry/otel-arrow/pull/2230

PR #2146, opened 2026-03-02
feat: Support EnvFilter directives in log level config

cijothomas:
# Change Summary

Change LogLevel from a fixed enum (Off/Debug/Info/Warn/Error) to a string that accepts full tracing EnvFilter directive strings. This allows per-crate log filtering directly from the YAML config without relying on RUST_LOG.

Example:

The change is backward compatible — simple keywords like "info" or "warn" continue to work as before. The default value is "info,h2=off,hyper=off" to maintain the existing behavior of silencing noisy dependencies.

Locally.

Yes, but back-compatible. Advanced filtering can be now provided in config itself.

### DIFF
--- a/rust/otap-dataflow/crates/config/src/settings/telemetry/logs.rs
+++ b/rust/otap-dataflow/crates/config/src/settings/telemetry/logs.rs
@@ -11,6 +11,15 @@ use serde::{Deserialize, Serialize};
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
 pub struct LogsConfig {
     /// The log level for internal engine logs.
+    ///
+    /// Accepts either a simple level keyword (`off`, `debug`, `info`, `warn`, `error`)
+    /// or a full [`tracing_subscriber::EnvFilter`] directive string for fine-grained
+    /// control (e.g., `"info,typespec_client_core=warn,azure_core=off"`).
+    ///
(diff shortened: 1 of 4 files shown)


rust/otap-dataflow/crates/config/src/settings/telemetry/logs.rs
```diff
@@ -20,20 +29,28 @@ pub struct LogsConfig {
...
-    Error,
+///
+/// Accepts either a simple level keyword (`off`, `debug`, `info`, `warn`, `error`)
+/// or a full [`tracing_subscriber::EnvFilter`] directive string for fine-grained
+/// control (e.g., `"info,typespec_client_core=warn,azure_core=off"`).
+///
+/// Defaults to `"info,h2=off,hyper=off"`.
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq)]
+#[serde(transparent)]
+pub struct LogLevel(String);
```
cijothomas, line 40: kept the name as LogLevel, as changing it to LogFilter might give the impression that users now need to know how to write filter expressions... Open to other ideas to make this better.

jmacd: I think this is reasonable. We could say we support RUST_LOG syntax for the level and point to its documentation.

PR #2147, opened 2026-03-02
Topic implementation

### LINKED ISSUE #1834: Proposal 7 - Support named channels [config-model-v1]
## Goals
- Support complex multi-pipeline deployments in a single process with strict isolation (per-pipeline runtime, metrics, shutdown, quota).
- Provide an efficient in-process communication fabric between pipelines via named channels.
- Keep the configuration model aligned with existing EngineConfig, PipelineGroupConfig, and PipelineConfig.

More details soon

lquerel:
# Change Summary

- Backend-agnostic API for topic publish/subscribe.
- Decouple publisher and subscriber nodes through named topics.
- Support both balanced and broadcast subscriptions behind one contract.
- Keep engine-facing code stable while allowing multiple backends (in-memory now, persistent/distributed later).

image

**YAML Topic Configuration**

Topics are declared either globally under `topics.<name>` or per pipeline group under `groups.<group>.topics.<name>`. Each topic has a description plus separate policy blocks for balanced and broadcast delivery:

```yaml
topics:
  raw_signals:
    description: "raw ingest stream"
    policies:
      balanced:
        queue_capacity: 1024
        on_full: block          # or drop_newest
      broadcast:
        queue_capacity: 8192
        on_lag: drop_oldest     # or disconnect
      ack_propagation: auto     # optional
```

`exporter:topic` nodes publish into a declared topic, and `receiver:topic` nodes subscribe to it in either balanced or broadcast mode. queue_on_full can still be overridden locally on a topic exporter, but capacities remain topic-level configuration.

**General capabilities**

- Decouple pipelines with in-memory topics instead of direct pipeline-to-pipeline wiring.
- Support worker-pool style processing through balanced consumer groups.
- Support fan-out/tap scenarios through broadcast subscribers.
- Support mixed topologies where the same topic serves both balanced and broadcast consumers.
- Tune balanced and broadcast behavior independently:
    - balanced queue capacity and full-queue policy
    - broadcast ring capacity and lag policy
- Optionally bridge Ack/Nack semantics across topic hops with ack_propagation.
- Allow group-local topic declarations to override top-level topics with the same local name.
  
* Closes #1834 

### Topic (balanced only mode) vs MPMC Flume channel - benchmarks

Below is a single comparison table across **message size** and **#consumers**, using the **middle throughput value** reported by Criterion (`thrpt: [low mid high]`), in **Melem/s**.

Msg size | Consumers | Topic (balanced) thrpt | Flume MPMC thrpt | Topic vs Flume
-- | -- | -- | -- | --
32 B | 1 | 4.3437 | 4.0805 | +6.45%
32 B | 2 | 3.9108 | 3.8037 | +2.82%
32 B | 4 | 3.2708 | 3.0329 | +7.84%
32 B | 8 | 2.7416 | 2.1232 | +29.13%
256 B | 1 | 4.9314 | 4.0027 | +23.20%
256 B | 2 | 3.7602 | 3.7473 | +0.34%
256 B | 4 | 3.1690 | 3.0948 | +2.40%
256 B | 8 | 2.8371 | 2.0671 | +37.25%
4096 B | 1 | 5.4115 | 4.0623 | +33.21%
4096 B | 2 | 3.6807 | 3.7129 | −0.87%
4096 B | 4 | 3.1721 | 3.1296 | +1.36%
4096 B | 8 | 2.6819 | 2.0787 | +29.02%

### Topic (broadcast only mode) vs Tokio tokio::sync::broadcast

Msg size | Consumers | Topic (broker) thrpt | Tokio broadcast thrpt | Topic vs Tokio
-- | -- | -- | -- | --
32 B | 1 | 4.3664 Melem/s | 3.0919 Melem/s | +41.22%
32 B | 2 | 2.7762 Melem/s | 2.1046 Melem/s | +31.92%
32 B | 4 | 1.8264 Melem/s | 1.5887 Melem/s | +14.96%
32 B | 8 | 1.0579 Melem/s | 1.0304 Melem/s | +2.67%
256 B | 1 | 4.8100 Melem/s | 3.1258 Melem/s | +53.88%
256 B | 2 | 2.8709 Melem/s | 2.1110 Melem/s | +35.99%
256 B | 4 | 1.8009 Melem/s | 1.5707 Melem/s | +14.66%
256 B | 8 | 1.0632 Melem/s | 1.0280 Melem/s | +3.42%
4096 B | 1 | 5.1536 Melem/s | 3.2001 Melem/s | +61.05%
4096 B | 2 | 2.9282 Melem/s | 2.1086 Melem/s | +38.86%
4096 B | 4 | 1.8497 Melem/s | 1.5447 Melem/s | +19.75%
4096 B | 8 | 1.0812 Melem/s | 1.0274 Melem/s | +5.23%

### Topic (mixed mode broadcast) vs Tokio tokio::sync::broadcast

Msg size | Consumers | Topic (mixed) thrpt | Tokio broadcast thrpt | Topic vs Tokio
-- | -- | -- | -- | --
32 B | 1 | 2.9991 Melem/s | 3.0042 Melem/s | −0.17%
32 B | 2 | 2.0465 Melem/s | 2.1419 Melem/s | −4.45%
32 B | 4 | 1.2015 Melem/s | 1.5551 Melem/s | −22.74%
32 B | 8 | 0.74308 Melem/s | 0.98701 Melem/s | −24.71%
256 B | 1 | 3.0205 Melem/s | 3.0079 Melem/s | +0.42%
256 B | 2 | 2.0323 Melem/s | 2.1317 Melem/s | −4.66%
256 B | 4 | 1.2382 Melem/s | 1.5964 Melem/s | −22.44%
256 B | 8 | 0.75830 Melem/s | 1.0248 Melem/s | −26.01%
4096 B | 1 | 3.1358 Melem/s | 3.1391 Melem/s | −0.11%
4096 B | 2 | 2.2230 Melem/s | 2.0835 Melem/s | +6.69%
4096 B | 4 | 1.2814 Melem/s | 1.5674 Melem/s | −18.25%
4096 B | 8 | 0.75007 Melem/s | 1.0143 Melem/s | −26.05%

**Note**: 

For this third scenario we are slightly slower, but our topic implementation can support **both balanced and broadcast modes simultaneously** in this configuration. This explains why it is somewhat less performant than the pure `tokio::sync::broadcast` implementation.

It is therefore important for the controller to analyze the **pipeline and topic topology** in order to select the most appropriate topic implementation and configuration depending on the topology and the subscription modes.

**Important Note**: I ran some load tests and I think there may still be a risk of deadlock or starvation. I will investigate this in more detail in a separate PR because, according to my tests, the issue only appears at load levels above 10M signals/s

image


rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -248,6 +248,19 @@ impl OtapPdata {
         &self.payload
     }
 
+    /// Clone this pdata while resetting transport context to default.
+    ///
+    /// This is used at transport boundaries (for example topic hops between
+    /// pipelines) where in-process Ack/Nack routing state must not leak across
+    /// boundaries.
+    #[must_use]
+    pub fn clone_without_context(&self) -> Self {
```
jmacd, line 401: Looks good.

rust/otap-dataflow/crates/engine/src/topic/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+The topic module is organized in three layers:
+
+1. Public API layer
+- `TopicBroker<T>`: create, lookup, list, remove, and close topics.
+- `TopicHandle<T>`: publish and subscribe entry point.
+- `Subscription<T>`: receive + ack/nack interface.
+- `TopicSet<T>`: per-pipeline resolved map of local topic alias -> `TopicHandle<T>`.
+
+2. Backend abstraction layer
```
jmacd:
```suggestion
2. Backend abstraction layer

In the future, we will implement alternative topic implementations by supplying an implementation of these traits. The default in-memory topic backend is provided, and we anticipate adding direct topic-export integration to Quiver.

```

PTAL @AaronRM

rust/otap-dataflow/crates/otap/src/topic_exporter.rs
```diff
@@ -85,25 +145,221 @@ impl Exporter<OtapPdata> for TopicExporter {
...
+                    biased;
+
+                    maybe_event = async {
+                        match &mut ack_events_rx {
+                            Some(rx) => rx.recv().await,
+                            None => None,
+                        }
+                    }, if ack_events_rx.is_some() => {
+                        if let Some(event) = maybe_event {
+                            if let Some(data) = pending_messages.remove(&event.message_id) {
```
lalitb: This seems to be removing from the pending messages on first ack/nack. How about the broadcast scenario when there are multiple subscribers - all subsequent acks/nacks would be orphaned. Is this intended?

rust/otap-dataflow/crates/engine/src/topic/topic.rs
```diff
@@ -0,0 +1,1078 @@
...
+                Self::try_send(sender, event)
+            }
+            None => Err(AckNotEnabled),
+        }
+    }
+
+    fn try_send(sender: &mpsc::Sender<AckEvent>, event: AckEvent) -> Result<(), Error> {
+        match sender.try_send(event) {
+            Ok(()) => Ok(()),
+            Err(mpsc::error::TrySendError::Full(_)) => Err(AckChannelFull),
```
lalitb: What happens when this returns `AckChannelFull` - It looks like the receiver treats it as a warning and moves on, but the exporter still has the message sitting in `pending_messages` waiting for an ack that will never arrive. Under sustained backpressure this could grow unboundedly with no recovery path until shutdown. Am I reading this right?

rust/otap-dataflow/crates/engine/src/topic/topic.rs
```diff
@@ -0,0 +1,1078 @@
...
+            return Err(TopicClosed);
+        }
+
+        let id = self.next_message_id(publisher_id);
+
+        if let Some(sg) = self.group.get() {
+            let envelope = Envelope { id, payload: msg };
+            sg.tx.send(envelope).await.map_err(|_| TopicClosed)?;
+        }
+        // No subscribers yet → message silently dropped (consistent with Mixed mode).
```
lalitb: Quick question on the contract here - when no subscriber has connected yet, the message is dropped but `PublishOutcome::Published` is still returned to the caller. Is that intentional for in-memory topics? Asking because during startup or a rolling restart of the subscriber pipeline, the publisher side would see a success signal while the data is silently lost.

rust/otap-dataflow/configs/engine-conf/topic_multitenant_isolation.yaml
```diff
@@ -0,0 +1,358 @@
...
+            type: processor:content_router
+            outputs: ["tenant_a", "tenant_b", "tenant_c", "unmatched"]
+            config:
+              routing_key:
+                resource_attribute: "service.namespace"
+              case_sensitive: false
+              routes:
+                tenant-a: tenant_a
+                tenant-b: tenant_b
+                tenant-c: tenant_c
```
lalitb, line 230: We plan to use this pattern for multi-tenant Geneva exporter isolation - one content_router at ingress dispatching to per -tenant topics, with each tenant group running its own retry -> batch -> Geneva pipeline.  Good to see this is already captured in the yaml here.

lalitb (approved):
LGTM. Thanks, awesome work! Left a few non-blocking comments. Looking forward to using this for tenant-level isolation in the Geneva exporter.

rust/otap-dataflow/crates/otap/src/topic_exporter.rs
lquerel: This is a limitation of the current system. The rationale is that, to my knowledge, the priority use cases for Microsoft and F5 are topics with balanced subscriptions, so I implemented only the strict minimum for Ack/Nack support in broadcast mode. I will add a limitations section in the documentation.

lquerel: Created this GH issue for tracking -> https://github.com/open-telemetry/otel-arrow/issues/2252

rust/otap-dataflow/crates/engine/src/topic/topic.rs
lquerel: Working on a better solution.

rust/otap-dataflow/crates/otap/src/topic_exporter.rs
lalitb: Yes, this is not the use-case for us, so not a blocker. Thanks.

drewrelmas: @lquerel Want to note - this is not relevant for @lalitb's use case but it IS for mine (ingress pipeline reliably broadcasting copied data to multiple processing pipelines)

Not a blocker for this PR but it is something pending for me to be able to adopt the topic implementation.

rust/otap-dataflow/crates/engine/src/topic/topic.rs
lquerel: This is now addressed. `exporter:topic` uses tracked publishes when ack_propagation: auto is enabled, and it only Ack/Nacks upstream once the tracked outcome resolves. If no receiver is connected, no downstream Ack arrives, so the publish times out and is Nacked upstream instead of being reported as an immediate success.

PR #2153, opened 2026-03-02
Add process-wide RSS metric to engine metrics

cijothomas:
Adds a memory_rss metric that reports the process-wide Resident Set Size (physical RAM) using the memory-stats crate. Unlike the existing jemalloc-based memory_usage (per-thread heap only), this captures the full process memory footprint — matching what external tools like kubectl top pod or htop report. Works on Linux, macOS, and Windows with no feature flags required. This is reported under engine level. If the direction is okay, will introduce more engine level metrics (total CPU utilization etc.)

(OTel Collector has this metric already. Slightly different name, just like every other metric)

Locally ran engine, and then query `http://127.0.0.1:8080/metrics?reset=false&format=prometheus` and look for the new metric - they match what I see using external tools for RSS memory tracking.

```txt
# HELP memory_rss Process-wide Resident Set Size — physical RAM currently used by the process. Matches what external tools report (e.g. `kubectl top pod`, `htop`, `ps rss`).
# TYPE memory_rss gauge
memory_rss{set="engine.metrics",process_instance_id="AGOLC2UIVF4SFAEKCW6BLZ5XMM",host_id="",container_id=""} 354533376 1772503088570
```

New metric, no change to existing.

Note: Decided to add a new dependency which brings libc which is already a dependency. If concerns about external crate, we can hand roll this ourselves.

### DIFF
--- a/rust/otap-dataflow/crates/controller/src/lib.rs
+++ b/rust/otap-dataflow/crates/controller/src/lib.rs
@@ -284,10 +284,13 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
                 use std::time::Duration;
                 use tokio::time::{MissedTickBehavior, interval};
 
+                // TODO: Make this interval configurable via engine config.
+                const ENGINE_METRICS_INTERVAL: Duration = Duration::from_secs(5);
+
                 let mut monitor =
                     EngineMetricsMonitor::new(engine_registry, engine_entity_key, engine_reporter);
 
-                let mut ticker = interval(Duration::from_secs(5));
+                let mut ticker = interval(ENGINE_METRICS_INTERVAL);
                 ticker.set_missed_tick_behavior(MissedTickBehavior::Skip);
 
                 loop {


rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -270,6 +270,48 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug> Controller<PData> {
...
+        let engine_metrics_handle = spawn_thread_local_task(
+            "engine-metrics",
+            admin_tracing_setup.clone(),
+            move |cancellation_token| async move {
+                use otap_df_engine::engine_metrics::EngineMetricsMonitor;
+                use std::time::Duration;
+                use tokio::time::{MissedTickBehavior, interval};
+
+                // TODO: Make this interval configurable via engine config.
+                const ENGINE_METRICS_INTERVAL: Duration = Duration::from_secs(5);
```
lquerel, line 288: Ideally, it should be possible not only to configure the interval but also to define a flag specifying whether the metric should be collected or not.

cijothomas: https://github.com/open-telemetry/otel-arrow/pull/2169 is introducing metric severity, so I think we can rely on that once ready? Or you were thinking of per-metric on/off flag? Or engine level on/off flag?

(I'll open an issue for tracking to not lose track)

lquerel:
@cijothomas Fully aligned with this direction. Additional process-level metrics can be added later to complement the pipeline-level ones. In the long term, everything should be configurable in a unified way.

cijothomas:
> @cijothomas Fully aligned with this direction. Additional process-level metrics can be added later to complement the pipeline-level ones. In the long term, everything should be configurable in a unified way.

https://github.com/open-telemetry/otel-arrow/pull/2160 Added one more engine/process level metric.

rust/otap-dataflow/crates/controller/src/lib.rs
lquerel: Metric severity is probable a good start.

PR #2156, opened 2026-03-03
Azure Monitor Exporter: surface auth misconfigurations quickly at startup

cijothomas:
Adds a 5-second token acquisition probe (with max 3 retry) before the main loop so auth misconfigurations (e.g., MSI on a non-Azure machine) surface in logs within ~5s instead of ~93s currently. On failure, the main loop retries indefinitely as before — no change to resilience

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
+++ b/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
@@ -1646,9 +1646,20 @@ mod telemetry_tests {
             let listening_port = portpicker::pick_unused_port().expect("No free ports");
             let listening_addr: SocketAddr = format!("127.0.0.1:{listening_port}").parse().unwrap();
 
-            // Receiver with metrics enabled via pipeline
-            let receiver =
-                SyslogCefReceiver::with_pipeline(pipeline, Config::new_udp(listening_addr));
+            // Receiver with metrics enabled via pipeline.
+            // Use max_batch_size=1 so that the single record is flushed
+            // immediately in the recv_from handler instead of waiting for
+            // the interval tick, which avoids timing-dependent flakiness.
+            let receiver = SyslogCefReceiver::with_pipeline(
+                pipeline,
+                Config {
+                    protocol: Protocol::Udp(UdpConfig { listening_addr }),
+                    batch: Some(BatchConfig {
+                        flush_timeout_ms: None,
+                        max_size: NonZeroU16::new(1),
+                    }),
+                },
+            );
 
             // Keep downstream open to avoid refused
             let (out_tx, mut _out_rx) = otap_df_channel::mpsc::Channel::new(8);
@@ -1693,7 +1704,7 @@ mod telemetry_tests {
             // To exercise the "invalid" path, send an empty datagram which is rejected by the parser.
             let _ = sock.send_to(b"", listening_addr).await.unwrap();
 
-            // Allow interval to tick
+            // Allow the receiver task to process the incoming messages.
             tokio::time::sleep(Duration::from_millis(150)).await;
 
             // Trigger telemetry collection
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
```diff
@@ -497,7 +497,50 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
                 message: format!("Failed to start telemetry timer: {e}"),
             })?;
 
-        let mut next_token_refresh = tokio::time::Instant::now();
+        // Quick startup probe: try to acquire a token with a short timeout.
+        // On success, auth headers are set immediately and token refresh is scheduled normally.
+        // On failure (e.g., MSI on a non-Azure machine), a warning surfaces within ~1s
+        // and the main loop retries via the regular get_token() infinite retry path.
+        let mut next_token_refresh = match auth
```
gouslu, line 504: this is too short, initial token acquisition takes 3secs.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/auth.rs
```diff
@@ -68,6 +68,19 @@ impl Auth {
         Ok(token_response)
     }
 
+    /// Attempt a single token acquisition with a timeout.
+    /// Used at startup to surface auth misconfigurations quickly.
+    /// Returns Ok(token) on success, or Err if it fails or times out.
+    pub async fn try_get_token_once(
```
gouslu: if token acquisition fails for any reason at all, should we assume misconfiguration right away? there has to be better ways to detected that, an I think we should at least give a few tries.

maybe I would change the code in such a way that I'd update get_token method so that it either fails at max x tries, or it keeps trying forever.

or maybe something like a retry policy being passed into get_token, sth like:

pub enum RetryPolicy {
    MaxRetries(u32),
    Infinite,
}

and initial get_token we pass 3, the latter one we pass infinite?

gouslu: it seems like network related issues are being handled inside azure_identity. if timeout is a big concern, then we can also make it part of "retrypolicy" in such a way that it is either max retries or timeout, whichever comes first. but, I don't think i'd be so worried about actual underlying request timeouts, as it seems to be capped and misconfiguration wouldn't fall under a retry attempt in azure_identity. so I think if there's no strong reason, I'd delete timeouts and make it a retry attempt based approach as a retry policy for get_token

cijothomas: This is single attempt at startup of the exporter itself, not in the other paths. Failure here is not assuming misconfiguration. It just logs that the startup attempt failed. May be the log should make it clear that it'll retry in background so this is not FATAL?

(btw, without this, it takes 90+ seconds to first see a WARN log about this, which is what prompted me to make this change)

gouslu: ok, so SDK's retry config also needs to be configured. but most importantly I think I need to change the auth.rs and the event loop such that it doesn't block. similar to in_flight_exports we should use futures for auth token acquisitions and we should only allow other events in the loop if auth has succeeded. does it make sense? I can take care of that happily btw.

gouslu: and initial token acquisition can and should also be moved into the event loop I think.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
jmacd: I'd accept 5s :)

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/auth.rs
jmacd: The best of both worlds, I guess, would be to have an async task start warning the user if the token can't be acquired after a short period? So, try to get a token with a timeout asynchronously, and log if you do not succeed after the short timeout. In addition, we should log warnings about retries, etc.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
```diff
@@ -497,7 +497,50 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
...
+            .try_get_token_once(tokio::time::Duration::from_secs(1))
```
jmacd:
```suggestion
            .try_get_token_once(tokio::time::Duration::from_secs(5))
```

cijothomas:
@jmacd @gouslu I modified this to do 5 sec (with upto 3 attempts) in the startup path to surface failures earlier. The log message from failure indicate that the main loop will continue token acquisition in background forever.

See if this is okay. Gokhan's concern about blocking event loop due to token acquisition - this can be addressed separately - Tracking https://github.com/open-telemetry/otel-arrow/issues/2283

jmacd (approved):

PR #2152, opened 2026-03-02
Investigate Engine stall

cijothomas:
sharing some investigation for engine "stall".

PR #2158, opened 2026-03-03
Fix PipelineCtrlMsgManager deadlock when node control channel is full

### LINKED ISSUE #2157: PipelineCtrlMsgManager deadlocks when a node's control channel is full
## Problem

The `PipelineCtrlMsgManager` can stall the entire pipeline on the single-threaded `LocalSet` runtime when any node's control channel is full.

@cijothomas reported an issue with pipeline stalling in his performance testing of the pipeline. He also found that the changes in his PR #2152 address the concern.

## Root Cause

`PipelineCtrlMsgManager::send()` is an `async` method that calls `sender.send(msg).await` when a node's control channel is full. https://github.com/open-telemetry/otel-arrow/blob/63a23cf282c43a3dceb453f6fd17c794a4e9bb70/rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs#L467-L486

Because all nodes and the manager run as cooperative tasks on a single-threaded Tokio `LocalSet`, this `.await` parks the manager's task and prevents it from processing *any* other messages — including acks that could free up space in the congested channel.

This creates a circular-wait stall:

1. The manager reads a message (e.g., `DeliverAck`) from the pipeline ctrl channel and tries to forward it to the target node's control channel.
2. That node's control channel is full → the manager calls `.send(msg).await` → **manager blocks**.
3. The node whose channel is full (e.g., an exporter that just finished a batch export) is itself trying to send more messages to the pipeline ctrl channel.
4. The pipeline ctrl channel fills up because the manager is no longer draining it → **the node blocks**.
5. **Deadlock**: The manager waits for space in the node's control channel, while the node waits for space in the pipeline ctrl channel. Neither can make progress. All other nodes are starved — their messages sit in the pipeline ctrl queue, never processed.

This affects all pipeline message types that flow through the manager (acks, nacks, timer ticks, telemetry, delayed data), though acks are the most likely trigger. Exporters that batch messages and then ack each one in a tight loop (e.g., `azure_monitor_exporter`) are especially susceptible:

```rust
// Batch-ack pattern in azure_monitor_exporter
for (_, context, payload) in completed_messages {
    effect_handler.notify_ack(AckMsg::new(payload)).await?;
    // ^^^^^^^^^ sends DeliverAck to pipeline ctrl channel
}
```

utpilla:
# Change Summary

This is almost the same as @cijothomas's #2152 but with changes around draining logic and reusing the `VecDeque` instance.

Change `send()` from blocking-async to non-blocking with buffered retry:

- **`try_send()` + `VecDeque` buffer**: When a node's control channel is full, the message is pushed into a `pending_sends: VecDeque<(usize, NodeControlMsg<PData>)>` instead of blocking.
- **`drain_pending_sends()`**: At the top of each event loop iteration, buffered messages are retried. Messages that still can't be delivered are re-queued at the back of the deque.
- **Retry wakeup via `select!`**: A 5ms sleep branch in `tokio::select!` wakes the loop when there are pending sends, placed *last* (after the message recv and timer branches) so incoming messages — especially acks that could unblock congested nodes — are always prioritized.
- **Observability**: A warning is logged when the pending sends buffer reaches a threshold (100), helping operators diagnose sustained backpressure.

The manager is now fully non-blocking on the outbound path (manager -> nodes), eliminating the circular-wait possibility.

### Note on the unbounded `VecDeque`

The `pending_sends` buffer is unbounded, but this is acceptable in practice. The buffer only grows when a node's control channel is persistently full — each event loop iteration can add at most one message to the buffer. Under normal operation the buffer stays empty; it only accumulates messages during transient backpressure spikes and drains as soon as the congested node catches up. If a node is permanently stuck (never draining its control channel), the buffer will grow indefinitely, but this indicates a fundamental problem with that node — the warning logged at 100 entries surfaces this to operators so they can investigate.

* Closes #2157 

A test (`test_circular_wait_between_node_and_manager`) reproduces the exact production scenario:
- Pipeline ctrl channel capacity 3, Node A control channel capacity 1, Node B capacity 10
- An active Node A task sends `DeliverAck` in a loop (simulating the exporter batch-ack pattern)
- Asserts that Node B receives its ack within 500ms — without the fix, this times out due to the circular wait

No

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
+++ b/rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
@@ -275,9 +275,27 @@ impl<PData> PipelineCtrlMsgManager<PData> {
         let mut is_draining = false;
         let mut draining_deadline: Option<Instant> = None;
 
+        // Single reusable timer for retrying buffered sends. Created once,
+        // reset only when `pending_sends` transitions from empty to non-empty.
+        // This avoids allocating a new Sleep future on every loop iteration
+        // (the standard tokio::select! pinned-sleep pattern).
+        let retry_delay = tokio::time::sleep(Duration::from_millis(5));
+        tokio::pin!(retry_delay);
+        let mut retry_armed = false;
+
         loop {
             // Drain any buffered sends before processing new messages.
             self.drain_pending_sends();
+
+            // Arm the retry timer when pending sends appear; disarm when drained.
+            if !self.pending_sends.is_empty() && !retry_armed {
+                retry_delay
+                    .as_mut()
+                    .reset(tokio::time::Instant::now() + Duration::from_millis(5));
+                retry_armed = true;
+            } else if self.pending_sends.is_empty() {
+                retry_armed = false;
+            }
 
             // Check if we've exceeded the draining deadline
             if let Some(deadline) = draining_deadline {
@@ -482,8 +500,10 @@ impl<PData> PipelineCtrlMsgManager<PData> {
                 // Placed last so incoming messages and timers are always
                 // prioritized — incoming acks may be exactly what unblocks
                 // the congested node.
-                _ = tokio::time::sleep(Duration::from_millis(5)),
-                    if !self.pending_sends.is_empty() => {
+                // Uses a pinned/reused sleep to avoid allocating a new timer
+                // future on every loop iteration.
+                _ = &mut retry_delay, if retry_armed => {
+                    retry_armed = false;
                     // drain_pending_sends() runs at loop top; just wake up.
                     continue;
                 }


utpilla:
## Considered options to fix the issue

### 1. Buffer-and-retry **(chosen)**

Use `try_send()` instead of `send().await`. On `Full`, buffer the message in a `VecDeque` and retry on subsequent event loop iterations.

**Pros:** Minimal code change, preserves the single event loop architecture, FIFO ordering maintained per node, manager never blocks.
**Cons:** Unbounded buffer — if a node's control channel stays full indefinitely, the buffer grows without limit (see note above). Retry uses a 5ms timer wakeup, adding slight latency to buffered messages.

### 2. Unbounded node control channels

Replace bounded `mpsc` channels with unbounded ones for the manager→node direction.

**Pros:** Simplest change — `send()` never blocks on an unbounded channel.
**Cons:** Removes all backpressure on the outbound path. A slow node could accumulate unbounded memory. Masks the real problem rather than handling it.

### 3. Increase channel capacities

Raise the default node control channel capacity (currently 256) or make it proportional to node count.

**Pros:** No code change, just a config tweak.
**Cons:** Only delays the deadlock — doesn't eliminate it. With enough ack volume the channel still fills up. False sense of safety.

### 4. Drop messages on full channel

On `Full`, log a warning and drop the message instead of blocking or buffering.

**Pros:** Simple, manager never blocks, bounded memory.
**Cons:** Drops acks/nacks, which breaks end-to-end delivery guarantees. Upstream nodes would never learn their data was delivered, potentially causing retries or data loss.

### 5. Per-node send tasks

Spawn a dedicated `tokio::task::spawn_local` for each outbound send, letting the manager fire-and-forget.

**Pros:** Manager never blocks, each send is independent.
**Cons:** Significant complexity — ordering guarantees within a node are lost (two tasks for the same node can complete out of order), shutdown coordination becomes harder (spawned tasks may outlive the manager), and spawning a task per message adds allocation and scheduling overhead.

### 6. Restructure the exporter ack pattern

Change exporters to use `try_send()` or batch their acks differently so they don't flood the pipeline ctrl channel.

**Pros:** Fixes the trigger at the source.
**Cons:** Pushes the fix to every exporter/processor individually. Doesn't fix the fundamental issue — the manager's `send()` can still block on any message type, not just acks. Fragile and easy to regress.

### 7. Select-based drain in `send()`

Inside `send()`, when the channel is full, use `tokio::select!` to simultaneously await the send *and* drain the pipeline ctrl channel.

**Pros:** Keeps the manager responsive while waiting.
**Cons:** Any messages received inside `send()`'s inner `select!` need full dispatch — the same message-handling logic from the outer event loop must be duplicated or refactored into a shared function. Processing a received message may itself call `send()`, creating recursive nesting. This effectively turns `send()` into a mini event loop, adding substantial complexity for a problem that the simpler buffer-and-retry approach solves.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -456,33 +474,75 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
+                break;
+            };
+            if let Some(sender) = self.control_senders.get(node_id) {
+                match sender.try_send(msg) {
+                    Ok(()) => {}
+                    Err(otap_df_channel::error::SendError::Full(msg)) => {
+                        self.pending_sends.push_back((node_id, msg));
+                    }
+                    Err(otap_df_channel::error::SendError::Closed(_)) => {
+                        // Drop message for closed channel
```
cijothomas, line 560: when would this occur? internal log required?

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -1824,4 +1884,188 @@ mod tests {
...
+    ///      The first Ack succeeds (fills A's cap-1 channel).
+    ///      The second Ack finds A's channel full → manager blocks on `.await`.
+    ///   4. Now both are stuck:
+    ///      - Manager is blocked sending to A's control channel (full)
+    ///      - Node A is blocked sending to pipeline ctrl channel (full, refilled
+    ///        after manager freed the initial two slots)
+    ///      - Neither can make progress.
+    ///   5. DeliverAck{B} sits in the pipeline ctrl queue — never processed.
+    ///
+    /// The test asserts Node B receives its ack within 500 ms.  On the current
```
cijothomas, line 1943: what is special about 500 ms? Is it to account for the fact that the pending sends are drained in best effort fashion only?
Such tests are flaky in CI/windows machines typically.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -456,33 +474,75 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
             match sender.try_send(msg) {
                 Ok(()) => {}
                 Err(otap_df_channel::error::SendError::Full(msg)) => {
-                    // Channel backpressured: await until space is available
-                    let _ = sender.send(msg).await;
+                    self.pending_sends.push_back((node_id, msg));
+                    if self.pending_sends.len() == PENDING_SENDS_WARN_THRESHOLD {
+                        otel_warn!(
+                            "pipeline.ctrl.pending_sends.high",
+                            count = self.pending_sends.len(),
```
cijothomas, line 532: 2026-03-03T16:34:57.453Z  WARN  otap-df-engine::pipeline.ctrl.pending_sends.high: Pending sends buffer reached threshold; a node's control channel may be persistently full [count=100] entity/pipeline.attrs: pipeline.id=main pipeline.group.id=perftest1 core.id=3 numa.node.id=0 process.instance.id=AGOLJB4V4R7SBL2PTBBYDQ7T2Y host.id=ubuntuperf container.id=
2026-03-03T16:35:00.460Z  WARN  otap-df-engine::pipeline.ctrl.pending_sends.high: Pending sends buffer reached threshold; a node's control channel may be persistently full [count=100] entity/pipeline.attrs: pipeline.id=main pipeline.group.id=perftest1 core.id=3 numa.node.id=0 process.instance.id=AGOLJB4V4R7SBL2PTBBYDQ7T2Y host.id=ubuntuperf container.id=
2026-03-03T16:35:09.688Z  WARN  otap-df-engine::pipeline.ctrl.pending_sends.high: Pending sends buffer reached threshold; a node's control channel may be persistently full [count=100] entity/pipeline.attrs: pipeline.id=main pipeline.group.id=perftest1 core.id=3 numa.node.id=0 process.instance.id=AGOLJB4V4R7SBL2PTBBYDQ7T2Y host.id=ubuntuperf container.id=
2026-03-03T16:35:15.762Z  WARN  otap-df-engine::pipeline.ctrl.pending_sends.high: Pending sends buffer reached threshold; a node's control channel may be persistently full [count=100] entity/pipeline.attrs: pipeline.id=main pipeline.group.id=perftest1 core.id=3 numa.node.id=0 process.instance.id=AGOLJB4V4R7SBL2PTBBYDQ7T2Y host.id=ubuntuperf container.id=
2026-03-03T16:35:18.756Z  WARN  otap-df-engine::pipeline.ctrl.pending_sends.high: Pending sends buffer reached threshold; a node's control channel may be persistently full [count=100] entity/pipeline.attrs: pipeline.id=main pipeline.group.id=perftest1 core.id=3 numa.node.id=0 process.instance.id=AGOLJB4V4R7SBL2PTBBYDQ7T2Y host.id=ubuntuperf container.id=
2026-03-03T16:35:21.858Z  WARN  otap-df-engine::pipeline.ctrl.pending_sends.high: Pending sends buffer reached threshold; a node's control channel may be persistently full [count=100] entity/pipeline.attrs: pipeline.id=main pipeline.group.id=perftest1 core.id=3 numa.node.id=0 process.instance.id=AGOLJB4V4R7SBL2PTBBYDQ7T2Y host.id=ubuntuperf container.id=
2026-03-03T16:35:27.957Z  WARN  otap-df-engine::pipeline.ctrl.pending_sends.high: Pending sends buffer reached threshold; a node's control channel may be persistently full [count=100] entity/pipeline.attrs: pipeline.id=main pipeline.group.id=perftest1 core.id=3 numa.node.id=0 process.instance.id=AGOLJB4V4R7SBL2PTBBYDQ7T2Y host.id=ubuntuperf container.id=
2026-03-03T16:35:30.960Z  WARN  otap-df-engine::pipeline.ctrl.pending_sends.high: Pending sends buffer reached threshold; a node's control channel may be persistently full [count=100] entity/pipeline.attrs: pipeline.id=main pipeline.group.id=perftest1 core.id=3 numa.node.id=0 process.instance.id=AGOLJB4V4R7SBL2PTBBYDQ7T2Y host.id=ubuntuperf container.id=

This does seem to fix the issue, but the above is from my load testing.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -456,33 +474,75 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
-                        self.send(node_id, msg).await;
+                        self.send(node_id, msg);
                     }
                 }
+
+                // Retry buffered sends after a short delay.
+                // Placed last so incoming messages and timers are always
+                // prioritized — incoming acks may be exactly what unblocks
+                // the congested node.
+                _ = tokio::time::sleep(Duration::from_millis(5)),
```
lalitb: nit: The retry `sleep(5ms)` is recreated on every loop iteration, so under steady traffic it almost never gets a chance to fire because the loop keeps selecting `recv` and dropping the timer. That means this branch is effectively inert when busy, and when  idle-with-pending it devolves into periodic polling. Functionally it still works because `drain_pending_sends()` runs at the top of the loop, but the timer lifecycle is inefficient and hard to reason about. Consider switching to a single pinned/reused sleep  (arm/reset on pending transition) so retry wakeups are explicit and stable.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -30,6 +30,10 @@ use std::time::{Duration, Instant};
 /// the caller's timeout fires, avoiding a race where both expire simultaneously.
 const MAX_DRAINING_BUFFER_DURATION: Duration = Duration::from_secs(1);
 
+/// Threshold for the pending sends buffer. When the buffer exceeds this size,
+/// a warning is logged to help operators diagnose sustained backpressure.
+const PENDING_SENDS_WARN_THRESHOLD: usize = 100;
```
lalitb, line 35: suggestion (non-blocking): Makes sense that the buffer is unbounded for now. One small addition to consider for a follow-up: a configurable hard cap (e.g. 10k) that drops the message and increments a counter/logs an error when exceeded - just as a safety net against OOM in pathological cases

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
utpilla: This is the channel-closed case. We ignored it even in the original code before this PR: https://github.com/open-telemetry/otel-arrow/blob/1548992db682cce1e88414cee20ff678a5c253e0/rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs#L480-L482

It happens during shutdown when a node's receiver is dropped before the manager finishes draining.  It's an expected race during shutdown. That's probably why it was always ignored.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
utpilla: Nothing special about 500ms. We're using the timeout to detect a deadlock. Without the fix it blocks forever, with the fix it completes in microseconds. The best-effort drain isn't a factor here: Node B's channel has plenty of capacity, so `drain_pending_sends` delivers on the first iteration. The timeout just needs to be finite, and 500ms is already very generous. There's no race condition, so it shouldn't be flaky.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
lalitb: Actually, on second thought - dropping acks silently isn't safe here. A dropped ack means the upstream node (fanout/receiver) waits indefinitely with no notification. If a cap is ever added, it would need a recovery path (e.g. nack the dropped message so upstream can retry, or a staleness timeout on inflight tracking). So the unbounded buffer is probably the right call for now.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -456,33 +474,75 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
+                // prioritized — incoming acks may be exactly what unblocks
+                // the congested node.
+                _ = tokio::time::sleep(Duration::from_millis(5)),
+                    if !self.pending_sends.is_empty() => {
+                    // drain_pending_sends() runs at loop top; just wake up.
+                    continue;
+                }
             }
         }
         Ok(())
```
ThomsonTan, line 512: Log a message if `!self.pending_sends.is_empty()` before dropping it if there are still any pending sends?

lalitb (approved):

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
cijothomas: we can create a tracking issue and follow up; having an unbounded buffer in the core engine is not good.

lalitb: +1 Agree on tracking this.

utpilla: This is already covered in the PR description under "Note on the unbounded VecDeque." The buffer stays empty during normal operation. It only grows when a node is persistently not draining its control channel, which signals a problem with that node. The warning at 100 entries surfaces this to operators. (We can discuss if that's too noisy)

As @lalitb pointed out, we can't simply cap and drop. A dropped ack leaves the upstream node (e.g., fanout) waiting indefinitely since there's no staleness timeout on inflight tracking today. We'd need a recovery path (nack the dropped message or add inflight staleness timeouts) before a cap is safe.

I'll create a tracking issue for bounded-with-recovery as a follow-up. That issue should probably also cover the other unbounded structures in the manager that I noticed while working on this (e.g., `BinaryHeap<Delayed<PData>>` for delayed data, which holds full `Box<PData>` payloads with no cap).

cijothomas: > The warning at 100 entries surfaces this to operators. (We can discuss if that's too noisy)

I'll share more feedback after this is merged. I can see it is somewhat noisy already, so I see the need of making it a metric. But not a blocker for this PR.

lquerel:
Great fix, and great analysis of the stall mechanism. The root cause breakdown in this PR (and the other original PR) made the failure mode very clear. I also want to acknowledge that I missed this issue earlier on my side, sorry about that. Thanks @cijothomas and @utpilla for this analysis and combined effort!

I agree with this PR as a strong first step: removing the blocking `await` from manager -> node sends addresses the circular-wait deadlock with minimal churn, and the reproducer test is strong.

For a long term solution, I'd suggest a follow up that keeps this non-blocking model but adds bounded backpressure semantics:
- Keep `pending_sends`, but add explicit bounds:
  - `max_pending_per_node`: prevents one stuck node from monopolizing the buffer.
  - `max_pending_total`: gives a hard global memory bound for the manager.
- Split policy by message class:
  - `critical`: `Ack/Nack/Shutdown` (and likely `DelayedData`) should not be silently dropped.
  - `best_effort`: `TimerTick/CollectTelemetry/Config` can be coalesced or latest-only under pressure.
- Add a retry/drain budget per loop (instead of draining all pending every iteration) so large pending queues don't impact deadline handling or control loop responsiveness.
- Expose metrics for pending depth, enqueue/dequeue rate, coalesced/dropped counts, and oldest pending age.

One note on the PR description: "at most one message added per loop" is not always true (timer branch can enqueue multiple messages in one iteration), so unbounded growth can happen faster than expected under sustained congestion.

Also worth calling out: this issue is runtime-agnostic. A multi-threaded work-stealing runtime may reduce scheduler starvation, but it does not break this logical wait cycle (`node A waiting on full pipeline-ctrl` <-> `manager waiting on full node-A control channel`). So bounded guardrails are still important even if runtime strategy changes - which I don't expect :-).

For initial guardrail values, an easy starting point is to derive from existing channel policy:
- `max_pending_per_node ~= control.node capacity`
- `max_pending_total ~= node_count * max_pending_per_node` (optionally capped)

So +1 as immediate mitigation, and I'd strongly support a bounded policy follow up for memory safety and predictable behavior under persistent backpressure.

lquerel (approved):
LGTM. See my general comment above in the thread for a follow up.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -456,33 +474,75 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
                     }
                 }
+
+                // Retry buffered sends after a short delay.
+                // Placed last so incoming messages and timers are always
+                // prioritized — incoming acks may be exactly what unblocks
+                // the congested node.
+                _ = tokio::time::sleep(Duration::from_millis(5)),
+                    if !self.pending_sends.is_empty() => {
+                    // drain_pending_sends() runs at loop top; just wake up.
```
ThomsonTan, line 507: nit: seems these 3 lines are not indented with the `if` statement.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
utpilla: Thanks for catching that! Fixed in https://github.com/open-telemetry/otel-arrow/pull/2158/changes/0f021a570f63fc53a11428a009d19bc8fff8f5d6.

Issue #2036, opened 2026-02-13
Columnar query engine: support additional assignment expressions


albertlockett:
Currently the only type of `set` operation supported is setting an attribute to a literal value. E.g. `set attributes["x"] = "y"`.

There are additional types of assignment expressions that need to be supported. The list of Lefts/Rights below form a matrix of what should be supported in expressions like `<left> = <right>`

**Left side:**
-  simple field assignment `set severity_text = <right>`
-  assignment to non-attribute `AnyValue` (e.g. log body). `logs | set body = <right>`
-  nested struct field. `resource.schema_url = <right>.`
-  attribute `set attribute["x"] = "y"` (already supported where "right" is a literal).

**Right side:**
- literal: e.g.`<left> = "x"`. 
  - Already somewhat supported for attributes.
  - Literal can be string, int, float, bool, or nested types such lists.
  - May require type aware context when parsing - should now allow assignment to string fields of non-string literals for example
  - If the literal is a default value, such as 0 or "", for some optional columns this can be implemented as dropping the column
- result of function call: e.g. `<left> = func()`
- assignment value from another field: `set body = event_name`, `attributes["x"] = attributes["y"]`.
  - In some cases, this will not be possible. For example, if the right-side contains more values than could be assigned to the left. E.g. `logs | set resource.schema_url = scope.schema_url` (because each resource contains multiple scopes). Similarly `resource.attributes["x"] = attributes["x"]` should not be allowed.
- null: e.g. `<left> = null`.
  - Note that in OTAP, in some cases this can be implemented by simply dropping optional columns

albertlockett:
Marking this not ready for work - I think some analysis needs to happen about what is the best way to do this ...

For example, we can imagine quite complicated assignment expressions like:
```js 
set body = "[" + resource.attributes["service.name"]  + " " + format(time_unix_nano) + "] ERROR: " 
  + attributes["exception.message"] + coalesce(attributes["exception.type"], "Unknown")
```
(assume + does a concat operation here)

This is a bit sophisticated. Consider that there are many four different data domains:
- `"[" + resource.attributes["service.name"]  + " "` (resource attrs)
- ` format(time_unix_nano) + "] ERROR: " ` (root batch)
- `attributes["exception.message"]` (attributes where key = exception.message)
- `coalesce(attributes["exception.type"], "Unknown")` (attributes where key = exception.message)

[...]

PR #2159, opened 2026-03-03
Columnar query engine support assigning to root batch fields

albertlockett:
# Change Summary

Adds support to the columnar query engine to assign values to columns on the root OTAP batch.

For example:
```kql
logs | set severity_text = "INFO"
```

Through this work, I wanted to validate that it's possible to bridge the expression evaluation code that was added in https://github.com/open-telemetry/otel-arrow/pull/2126 with a PipelineStage implementation. In doing so, many things that were private/dead-code in the `expr` module are now marked public.

This also means that we support assigning values from various types of expressions, not simply literals. For example:
```kql
logs | set event_name = attributes["event_name"] // assign from attribute
logs | set severity_number = severity_number * 5 + 1 // assign from arithmetic
```

This code handles:
- type checking the result when doing the assignment, at planning time if possible and otherwise at runtime
- converting the expression eval result back into a dictionary if the column supports it
- removing the column if the expression evaluated to null (and returning an error if the column is not nullable)
- not assigning nulls to non-null columns (returning error if this would happen)

To make it easier to take the results of the evaluated expression, I also modified the JoinExec trait that is used to expression evaluation expose a method to return the rows to take from the values being joined. This helps to align the rows produced by the expression evaluation to the order of the destination record batch.

* Related to #2036 

Existing unit tests ensure we didn't break the planner.

There are over 30 new unit tests for the assignment pipeline stage.

I suppose the ability for the transform processor to evaluate these new types of expressions could be considered user facing. 

## Near-term followups:

In followup PRs, I'll add support to assign values attributes. We already support assigning literals to values (added here https://github.com/open-telemetry/otel-arrow/pull/1885), but this only supports literals.

We'll also now need to fix this TODO (in fact, we should have done after #1885, but the code in this PR could also cause the concatenation we're doing here to fail).
https://github.com/open-telemetry/otel-arrow/blob/63a23cf282c43a3dceb453f6fd17c794a4e9bb70/rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs#L190-L207

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
@@ -1208,25 +1208,4 @@ mod test {
             }
         }
     }
-
-    #[tokio::test]
-    async fn test_duration_update() {
-        let traces_data = to_traces_data(vec![
-            Span::build()
-                .start_time_unix_nano(1u64)
-                .end_time_unix_nano(5u64)
-                .finish(),
-        ]);
-        let otap_batch = otlp_to_otap(&OtlpProtoMessage::Traces(traces_data));
-
-        // check the duration
-
-        let pipeline_expr =
-            OplParser::parse("traces | set duration_time_unix_nano = duration_time_unix_nano * 2")
-                .unwrap()
-                .pipeline;
-        let mut pipeline = Pipeline::new(pipeline_expr);
-
-        let result = pipeline.execute(otap_batch).await.unwrap();
-    }
 }


rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -0,0 +1,1211 @@
...
+    }
+
+    #[tokio::test]
+    async fn test_set_root_column_rejects_invalid_type_during_planning_opl_parser() {
+        test_set_root_column_rejects_invalid_type_during_planning::<OplParser>().await
+    }
+
+    #[tokio::test]
+    async fn test_set_root_column_rejects_invalid_type_during_planning_kql_parser() {
+        test_set_root_column_rejects_invalid_type_during_planning::<OplParser>().await
```
lalitb:
```suggestion
        test_set_root_column_rejects_invalid_type_during_planning::<KqlParser>().await
```

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -0,0 +1,1211 @@
...
+                            // already smallest dict size
+                            Ok(Arc::clone(array_vals))
+                        }
+                        DataType::UInt16 => {
+                            // check if we can use a smaller dictionary key
+                            let values_as_dict = array_vals
+                                .as_any()
+                                .downcast_ref::<DictionaryArray<UInt16Type>>()
+                                .expect("can downcast to dict");
+                            if values_as_dict.values().len() <= 255 {
```
lalitb: Seems to be off-by-one for UInt8 key downsizing :)

rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
```diff
@@ -574,7 +590,9 @@ impl Cardinality {
     }
 }
 
-fn estimate_cardinality<'a>(info: &FieldInfo<'a>) -> Cardinality {
+/// Estimate the cardinality of a set of arrays
+#[must_use]
+pub fn estimate_cardinality<'a>(info: &FieldInfo<'a>) -> Cardinality {
```
lalitb, line 595: Maybe I am missing something, but this is now called from `eval_result_to_array` - and new caller can pass types this function doesn't handle. Specifically, `Duration(Nanosecond)` will hit the `unreachable!()` because `duration_time_unix_nano` is listed in `root_field_supports_dict_encoding` but there's no match arm for `Duration` here. A query like 

```
 traces | extend duration_time_unix_nano = duration_time_unix_nano * 2
```
would trigger a panic.

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
albertlockett: ah thanks, fixed in c3e6dbd8c

Originally I was thinking this needed to be one less due to an open bug in arrow-rs, but turns out that bug exists is for a different kernel (concat, not cast)

rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
albertlockett: Ah yeah good catch. I didn't realize that function didn't handle duration type.

We actually don't have for date/time arithmetic in this engine yet. The query you give would fail during planning, because the engine would detect the type Duration and would consider the type invalid for arithmetic.

I plan to add support for better arithmetic on datetimes/durations in future PR, but for now I just added some TODO comments and also removed this duration field from `root_field_supports_dict_encoding` for good measure.

PR #2165, opened 2026-03-03
add recv_when to backpressure the message channel pdata meanwhile still receiving control messages

gouslu:
# Change Summary

add recv_when to backpressure the message channel pdata meanwhile still receiving control messages. this fixes an issue where state would keep growing with a very high boundary under heavy load

Part of #2188 

manual testing - further load testing is required.

no

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/message.rs
+++ b/rust/otap-dataflow/crates/engine/src/message.rs
@@ -285,7 +285,7 @@ impl<PData> MessageChannel<PData> {
                 if self
                     .pdata_rx
                     .as_ref()
-                    .expect("pdata_rs must exist")
+                    .expect("pdata_rx must exist")
                     .is_empty()
                 {
                     let shutdown = self


rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
```diff
@@ -190,6 +191,12 @@ impl Heartbeat {
         let status = response.status();
         let body = response.text().await.unwrap_or_default();
 
+        otel_warn!(
```
lalitb, line 194: this looks unconditional, logging warning even for success?

cijothomas, line 194: is this new log necessary? if yes, and is not duplicating existing one, please do that in separate PR.

rust/otap-dataflow/crates/engine/src/message.rs
```diff
@@ -310,8 +328,8 @@ impl<PData> MessageChannel<PData> {
                     Err(e)  => return Err(e),
                 },
 
-                // B) Then pdata
-                pdata = self.pdata_rx.as_mut().expect("pdata_rx must exist").recv() => {
+                // B) Then pdata (guarded by accept_pdata)
+                pdata = self.pdata_rx.as_mut().expect("pdata_rx must exist").recv(), if accept_pdata => {
```
lalitb, line 357: At minimum, there should be a test to verify that `recv_when(false)` returns control messages while leaving `pdata` in the channel, and that the subsequent `recv_when(true)` can retrieve the `pdata`.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
gouslu: line 179 says

        if response.status().is_success() {
            return Ok(());
        }


so by 194 it is not success anymore.

gouslu: it was necessary, yes. I want to keep that, it is a minor addition, harmless, and doesn't need a separate PR. it is following the same convention as log analytics client.

rust/otap-dataflow/crates/engine/src/message.rs
gouslu: yeah, I haven't seen any tests in the file, so I didn't add any. will look into adding some.

gouslu: I figured out that it is in exporter.rs, added tests there.

ThomsonTan, line 357: If accept_pdata is set to false, seems the handling of closed `RecvError::Closed` is also bypassed as it is guarded by `if accept_pdata`.  Probably should check this (channel has been showdown) at the beginning of the current loop.

rust/otap-dataflow/crates/engine/src/message.rs
```diff
@@ -228,6 +228,24 @@ impl<PData> MessageChannel<PData> {
...
+    ///
+    /// During shutdown draining the guard is ignored — pdata is
+    /// always drained until the deadline, regardless of
+    /// `accept_pdata`.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [`RecvError`] if both channels are closed, or if the
+    /// shutdown deadline has passed.
+    pub async fn recv_when(&mut self, accept_pdata: bool) -> Result<Message<PData>, RecvError> {
```
jmacd, line 248: I feel `recv_when` doesn't quite capture what this does. A better name, maybe, recv_conditional(control: bool, data: bool) that would support any combination of one or the other channel?

You would call recv_conditional(true, !at_capacity)

rust/otap-dataflow/crates/engine/src/message.rs
gouslu: great catch, working on a fix

rust/otap-dataflow/crates/engine/src/message.rs
gouslu: I feel that if we had 2 split channels that'd be even better, but it seems like it is making it tricky to do the shutdown and draining. Do you want me to give it a try? Also, when do you think we'd never want to receive from control channel? the whole purpose of this is to apply backpressure, if I want to ignore control channel messages, I can do that, but applying backpressure to control doesn't seem to be necessary?

cijothomas:
based on my load testing, this is fixing the issue. After overnight run for 10+ hours, memory hold steady, despite accepting 100s of millions of logs and exporting them. IOW, I approve this PR :)

jmacd:
I will merge this and create an issue to follow up on any unresolved questions above.

jmacd (approved):

PR #2167, opened 2026-03-03
perf: Use naive offsets when reindexing whenever possible

### LINKED ISSUE #2124: Add a fast path for reindexing based on a naive offset

JakeDern:
# Change Summary

This PR adds a bunch of fast paths for reindexing to avoid sorting and compacting ids whenever it's determined that it's possible to do so without overflowing the type and while also not causing junk data violations.

The intuition is that as long as we have enough ids to spare, we don't care if a column has ids like `[1, 1000, 2000]`. We would rather burn 3000 available ids on this batch and just reindex them by applying an offset to get them out of the range of the previous a batch than have to sort the whole column, find all the contiguous ranges, and then compact them down. We fall back to that slow path if we need the room. 

This PR also adds more data generation and benchmarks.

The speedup is highly variable based on scenario but is overall pretty good compared to the existing one. Some notes:

- The `gapped` cases should be considered somewhat pathological for all implementations. There are many gaps with basically 0 contiguous runs so in real workloads these are not going to show as good of gains.
- The `contiguous` categories generally show less gain because the old implementation didn't have to redact rows from those (the most expensive operation) but still had to sort.
- The `gapped` improvements are must smaller for logs at the larger batch size because they're forced to compact. Metrics is "lucky" in that the datapoints can have u32 ids.
- The `unsorted_contiguous` case show a gigantic improvement, because the profile shows way more time spent in sorting for that the old implementation because it falls back to a general quicksort. I think rust tries to speed up sorts by noticing if the data is mostly sorted already and takes a different path. 

Here's the perf stats that show how much time was spent sorting for some of those cases:

Unsorted contiguous -- quicksort:
```
 64.16%  core::slice::sort::unstable::quicksort::quicksort
   3.24%  core::slice::sort::shared::smallsort::insertion_sort_shift_left
   2.91%  core::slice::sort::shared::pivot::median3_rec
   1.03%  sort_vec_to_indices
```
  Contiguous -- ipnsort
```
  15.36%  core::slice::sort::unstable::ipnsort
   3.17%  sort_vec_to_indices
```
  Gapped -- ipnsort again
```
  11.48%  core::slice::sort::unstable::ipnsort
   2.43%  sort_vec_to_indices
  21.68%  apply_mappings
  18.85%  create_mappings
```

  | Scenario | Signal | New (us) | Old (us) | Speedup |
  |---|---|---|---|---|
  | 1r1s/contiguous | metrics | 47.4 | 75.6 | 1.59x |
  | 1r1s/contiguous | logs | 43.8 | 118.1 | 2.70x |
  | 1r1s/contiguous | traces | 50.1 | 128.3 | 2.56x |
  | 1r1s/unsorted_contiguous | metrics | 48.3 | 123.4 | 2.56x |
  | 1r1s/unsorted_contiguous | logs | 43.0 | 327.6 | 7.62x |
  | 1r1s/unsorted_contiguous | traces | 49.8 | 340.9 | 6.85x |
  | 1r1s/gapped | metrics | 46.5 | 109.7 | 2.36x |
  | 1r1s/gapped | logs | 43.2 | 166.1 | 3.84x |
  | 1r1s/gapped | traces | 49.6 | 177.3 | 3.58x |
  | 3r2s/contiguous | metrics | 76.0 | 226.8 | 2.98x |
  | 3r2s/contiguous | logs | 81.0 | 504.1 | 6.22x |
  | 3r2s/contiguous | traces | 88.5 | 535.4 | 6.05x |
  | 3r2s/unsorted_contiguous | metrics | 76.1 | 689.0 | 9.06x |
  | 3r2s/unsorted_contiguous | logs | 78.1 | 2200.9 | 28.2x |
  | 3r2s/unsorted_contiguous | traces | 86.1 | 2219.9 | 25.8x |
  | 3r2s/gapped | metrics | 75.8 | 412.8 | 5.45x |
  | 3r2s/gapped | logs | 575.5 | 770.4 | 1.34x |
  | 3r2s/gapped | traces | 596.8 | 825.7 | 1.38x |

* Closes #2124 

- Added new unit tests and benchmarks

No


albertlockett (approved):

albertlockett:
LGTM! Great work @JakeDern

lquerel:
@JakeDern @albertlockett @jmacd @cijothomas 
We should keep an eye on the compression rate before and after this change. I want to make sure there’s no significant impact on delta encoding, and therefore on the compression rate. The less regularity we have in the IDs, the less we can benefit from delta encoding.

I’m thinking we should consider extending the traffic generator so we can place the batcher and the OTAP exporter in different extreme scenarios.

PR #2169, opened 2026-03-03
Node-level pipeline metrics instrumentation

### LINKED ISSUE #2018: Component-level pipeline metrics implementation

jmacd:
# Change Summary

Implements semantic conventions from [Collector node telemetry RFC](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md).

Adds `MetricLevel`: with values None, Basic, Normal, Detailed

Change `policies::telemetry::channel_metrics` from bool to MetricLevel, default Basic, former behavior Basic

Adds new types for forward- and reverse-directed engine context.

- RouteData automatic fields on the forward path (entry_time_ns, output_port_index, calldata) replaces direct CallData
- UnwindData automatic fields on the return path (return_time_ns)

New `Interests`:

- PRODUCER_METRICS: capture output port index, receivers capture entry time
- CONSUMER_METRICS: processors and exporters capture entry time
- ENTRY_TIMESTAMP: capture the timestamp (at detailed level)
- SOURCE_TAGGING: creates a frame with source tagging

A new node-level field named `node_interests` makes all the default behaviors associated with these interests simple, as each node's interests are derived from the effective MetricLevel policy plus source tagging (when sending to multi-source destinations).

Adds new producer and consumer metric instruments following the RFC linked above:

Metric set `node.consumer` with three request counts:

- `consumed.success`
- `consumed.failure`
- `consumed.refused`

At detailed level:

- `consumed.duration`: Mmsc, future Histogram.

A similar group `node.producer` with metrics named `produced.*`, and an enum for categorizing responses `RequestOutcome` into Success, Failure, Refused.

New use of macros to simplify `pdata.rs` (see how you like it), much shorter now.

New traits and implemented on OtapPdata:

- `Unwindable`  for unwinding frames of the context.
- `ReceivedAtNode` for stamping the entry time
- `StampOutputPort` for stamping the exit path
- `OutputSend` for handling output ports

New structs

- `OutputRouter` which encapsulates send/send_to/try_send/try_send_to for use with the `OutputSend` trait
- `NodeMetricHandles` which encapsulates the consumer and multiple producer metrics handles

New pipeline_ctrl.rs logic:

- Instrument producer and consumer metrics on the Ack/Nack return path while routing
- Former use of next_ack and next_nack replaced by `Unwindable` trait; frames are popped inside engine's pipeline_ctrl.rs not otap's pdata.rs

In `pdata.rs`, new API `update_send_context` for updating route data without touching user CallData, also APIs for output port index and entry frame updates.

Testing support: next_ack and next_nack move into testing module, adds a harness that simulates the pipeline controller.

Fixes #2018 .

✅ 

✅

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/control.rs
+++ b/rust/otap-dataflow/crates/engine/src/control.rs
@@ -13,10 +13,10 @@ use bytemuck::Pod;
 use otap_df_channel::error::SendError;
 use otap_df_telemetry::reporter::MetricsReporter;
 use smallvec::SmallVec;
+use std::cell::Cell;
 use std::collections::HashMap;
 use std::marker::PhantomData;
 use std::time::{Duration, Instant};
-use std::cell::Cell;
 
 thread_local! {
     /// Temporary; see nanos_since_birth


rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -374,11 +413,15 @@ impl<PData> PipelineCtrlMsgManager<PData> {
                                 self.delayed_data.push(delayed);
                             }
                         }
-                        PipelineControlMsg::DeliverAck { node_id, ack } => {
-                            self.send(node_id, NodeControlMsg::Ack(ack)).await;
+                        PipelineControlMsg::DeliverAck {
+                            ack,
+                        } => {
+                            self.unwind_ack(ack).await;
```
utpilla: Could you update this PR branch to get the latest changes from #2158?

We want to remove blocking await tasks for `manager -> node` communications.

rust/otap-dataflow/crates/engine/src/control.rs
```diff
@@ -12,11 +12,22 @@ use crate::shared::message::{SharedReceiver, SharedSender};
...
 use std::marker::PhantomData;
+use std::sync::OnceLock;
 use std::time::{Duration, Instant};
 
+/// Returns a monotonic timestamp in nanoseconds since an arbitrary process epoch.
+/// Used for duration calculations in pipeline component metrics.
+///
+/// TODO: This could be perhaps more platform-specific.
+pub fn nanos_since_birth() -> u64 {
+    static EPOCH: OnceLock<Instant> = OnceLock::new();
```
utpilla: (non-blocking) It might be better to use `thread_local!` instead of `OnceLock`. Thread local should work because in our usage we only care about the difference between timestamps.

jmacd: Interesting. I was trying to take the simplest path, here.

I think we ought to use arch-specific built-in functions to read a u64 monotonic counter the same way `Instant` does, like the x86 `rdtsc`, in which case we would not require any state.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
jmacd: Done!

utpilla: Could you update `unwind_ack` and `unwind_nack` methods to not be `async`? Then you can get rid of the `.await` here.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -312,351 +483,166 @@ impl OtapPdata {
...
+
+/// Implements `ConsumerEffectHandlerExtension<OtapPdata>` for an EffectHandler type.
+macro_rules! impl_consumer_ext {
+    ($handler:ty) => {
+        #[async_trait(?Send)]
+        impl ConsumerEffectHandlerExtension<OtapPdata> for $handler {
+            async fn notify_ack(&self, mut ack: AckMsg<OtapPdata>) -> Result<(), Error> {
+                if ack.accepted.has_pending_metrics(Interests::ACKS) {
+                    ack.unwind.return_time_ns = nanos_since_birth();
+                }
```
utpilla: At `Normal` level, `has_pending_metrics()` sees `PIPELINE_METRICS` on the frames and returns `true`, so we end up calling `nanos_since_birth()` here. But the thing is, `entry_time_ns` is always 0 at `Normal`. It only gets set when `ENTRY_TIMESTAMP` is present (i.e., `Detailed` level). So, `record_frame_metrics` skips the duration recording anyway because of the `entry_time_ns > 0` guard. The counters still work correctly. We're just doing an unnecessary clock read per ACK/NACK.

Would it make sense to check for `ENTRY_TIMESTAMP` instead of `PIPELINE_METRICS` in `has_pending_metrics`, so the timestamp is only captured when it'll actually be used?

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
jmacd: Fixed.

rust/otap-dataflow/crates/otap/src/pdata.rs
jmacd: Yes!

jmacd: Fixed. Also I renamed the method `has_timing` and made it private.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -142,7 +168,93 @@ impl Context {
...
+    ///
+    /// When `ENTRY_TIMESTAMP` is present in `interests`, the frame's
+    /// entry timestamp is captured automatically.
+    pub(crate) fn update_send_context(&mut self, node_id: usize, interests: Interests) {
+        if let Some(top) = self.stack.last_mut() {
+            if top.node_id == node_id {
+                top.interests |= interests;
+                if interests.contains(Interests::ENTRY_TIMESTAMP) {
+                    top.route.entry_time_ns = nanos_since_birth();
+                }
```
utpilla: This would overwrite the receive-time timestamp for processors at `Detailed` level, leading to incorrect `consumed_duration_ns` metrics.

For a processor, `push_entry_frame` captures `entry_time_ns = T1` (receive time), then `update_send_context` overwrites it with `T2` (send time) when merging into the same frame. So `consumed_duration_ns` ends up as `ack_time - send_time` instead of `ack_time - receive_time`, losing the node's processing time.

It looks like we handle this correctly in `subscribe_to`.

I think this should fix it:

```suggestion
                if interests.contains(Interests::ENTRY_TIMESTAMP) && top.route.entry_time_ns == 0 {
                    top.route.entry_time_ns = nanos_since_birth();
                }
```

rust/otap-dataflow/crates/engine/src/output_router.rs
```diff
@@ -0,0 +1,261 @@
...
+        Self {
+            node_id,
+            ports,
+            default,
+        }
+    }
+
+    /// Returns the list of connected output port names.
+    #[must_use]
+    pub fn connected_ports(&self) -> Vec<PortName> {
```
utpilla, line 93: `connected_ports()` collects from a HashMap, so the order is nondeterministic. Since the constructor sorts entries to assign stable indices, should we sort the result here too for consistency? If this is intended for future use, having the order match the index assignment would prevent subtle bugs.

rust/otap-dataflow/crates/engine/src/output_router.rs
```diff
@@ -0,0 +1,261 @@
...
+        self.ports.keys().cloned().collect()
+    }
+
+    /// Returns the stable output port index for the default port.
+    #[must_use]
+    pub fn default_output_port_index(&self) -> u16 {
+        self.default.as_ref().map_or(0, |(_, idx)| *idx)
```
utpilla: `default_output_port_index()` returns `0` when there's no default port, but `0` is a valid port index. If this is intended for future use, should it return `Option<u16>` to distinguish "no default" from "default is port 0"?

rust/otap-dataflow/crates/engine/src/local/processor.rs
```diff
@@ -92,12 +94,8 @@ pub trait Processor<PData> {
...
 pub struct EffectHandler<PData> {
     pub(crate) core: EffectHandlerCore<PData>,
-
-    /// A sender used to forward messages from the processor.
-    /// Supports multiple named output ports.
-    msg_senders: HashMap<PortName, Sender<PData>>,
-    /// Cached default sender for fast access in the hot path
-    default_sender: Option<Sender<PData>>,
+    /// Output-port router.
+    pub router: OutputRouter<Sender<PData>>,
```
utpilla, line 98: These could be `pub(crate)` in all the other effect handlers as well.

rust/otap-dataflow/crates/engine/src/control.rs
utpilla: > I think we ought to use arch-specific built-in functions to read a u64 monotonic counter the same way Instant does, like the x86 rdtsc, in which case we would not require any state.

That's a great idea! However, one concern is that [_rdtsc()](https://doc.rust-lang.org/beta/core/arch/x86/fn._rdtsc.html) is marked `unsafe` in Rust (it's a raw intrinsic), and it returns cycles rather than nanoseconds, so we'd need TSC frequency calibration plus per-arch implementations.

My suggestion of switching from `OnceLock` to `thread_local!` was a much simpler and not so significant improvement. It also makes the intent clear. `OnceLock` is a synchronization primitive which we don't need because of the thread-per-core, shared nothing architecture. Also, since we only care about timestamp differences, each thread having its own epoch is fine.

jmacd: I didn't mean actual _rdtsc()--

Instant::now() is implemented by vDSO system calls underneath, https://doc.rust-lang.org/std/time/struct.Instant.html#underlying-system-calls which on Linux will be clock_gettime(), that's what I actually want here.

Using a thread_local, anyway.

rust/otap-dataflow/crates/otap/src/pdata.rs
jmacd: Yes. Thank you and very good eye.

It's more subtle than I realized at first. This logic is only needed for receivers which may capture a timestamp in subscribe_to(), but for receivers that do not call subscribe_to() we set it here. So I added the != 0 guard and require PRODUCER_METRICS.

```
+                if interests.contains(Interests::ENTRY_TIMESTAMP | Interests::PRODUCER_METRICS)
+                    && top.route.entry_time_ns == 0
```

rust/otap-dataflow/crates/engine/src/output_router.rs
jmacd: Fixed.

rust/otap-dataflow/crates/engine/src/output_router.rs
jmacd: This was dead code. Removed.

rust/otap-dataflow/crates/engine/src/local/processor.rs
jmacd: These are called from the `pdata.rs` macro in crates/otap/src/pdata.rs. 
I feel the organization could improve, but I can't see it right now!

jmacd:
I think this would benefit from a round of more-thorough testing, but I think better to merge and address later. The more-complex components like fanout worry me a little bit.

PR #2174, opened 2026-03-04
[otap-df-otap] Update Syslog CEF receiver to add input format attribute

utpilla:
# Change Summary

Adds an `input.format` log attribute to every log record emitted by the Syslog CEF Receiver. This attribute indicates which message format was detected by the receiver's auto-detection logic, enabling downstream processors to filter, route, or transform records based on the originating format.

### Motivation

The receiver supports multiple input formats (RFC 5424, RFC 3164, CEF, and CEF over syslog) and auto-detects the format of each incoming message. Until now, there was no way for downstream pipeline components to know which format a given log record originated from. With `input.format`, users can build processor-based filtering rules (e.g., drop all RFC 3164 messages, or route CEF messages to a separate exporter) without needing receiver-level configuration options.

### Changes

**Core implementation** (`parsed_message.rs`):
- Added `INPUT_FORMAT` constant (`"input.format"`)
- Added `format()` method on `ParsedSyslogMessage` that returns the detected format as a static string
- Updated `add_attributes_to_arrow()` to emit `input.format` as an attribute on every log record

Unit tests

Yes, users would now find an additional attribute in the syslog log record messages.


cijothomas (approved):
Nice addition!

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver/parser/parsed_message.rs
```diff
@@ -34,6 +34,9 @@ const CEF_DEVICE_EVENT_CLASS_ID: &str = "cef.device_event_class_id";
 const CEF_NAME: &str = "cef.name";
 const CEF_SEVERITY: &str = "cef.severity";
 
+// Attribute key constant for detected input format
+const INPUT_FORMAT: &str = "input.format";
```
ThomsonTan, line 38: Could name attribute conflict with the the attribute coming OTel sementic convention? Probably add namespace to prevent potential collision like `cef.input.format`?

lquerel: @utpilla  Why not making this attribute part of the syslog configuration with a default value set to `input.format` ?

utpilla: If the concern is a potential naming conflict with Semantic Conventions, configurability doesn’t actually solve the problem. Even if the attribute is configurable and defaults to `input.format`, we would still need to change that default if the Semantic Conventions group later reserves the name, which would be a breaking change.

cijothomas: All the attribute names used by the receiver today - they are not really part of any stable OTel conventions but rather self defined.

PR #2176, opened 2026-03-04
feat: Add environment variable substitution support (${env:VAR})

### LINKED ISSUE #2175: Add environment variable substitution support (${env:VAR})
Add support for environment variable substitution in configuration files, matching the behavior of the [OpenTelemetry Go Collector](https://opentelemetry.io/docs/collector/configuration/#environment-variables).

Motivation: Hard-coding values such as endpoints, ports, and credentials in configuration files is impractical in containerized and CI/CD environments. Environment variable substitution allows operators to inject runtime values without modifying config files.

Example:
```yaml
receivers:
  otlp:
    endpoint: ${env:OTLP_ENDPOINT:-localhost:4317}

exporters:
  otlp:
    endpoint: ${env:OTLP_ENDPOINT}
```

andborja:
# Change Summary

Add support for environment variable substitution in configuration files,
matching the behavior of the OpenTelemetry Go Collector. Ref: https://opentelemetry.io/docs/collector/configuration/#environment-variables

Substitution is applied to the raw config string before deserialization,
making it format-agnostic (YAML and JSON). A zero-dependency left-to-right
parser handles the substitution in `crates/config/src/env_substitution.rs`.

Supported syntax:
- ${env:VAR}         — substituted; error if unset
- ${env:VAR:-default} — substituted; falls back to `default` if unset
- ${env:VAR:-}       — substituted; falls back to empty string if unset
- $$                 — produces a literal `$`
- ${file:...} etc.   — passed through unchanged (future providers)

* Closes #2175
* Part (3) of #1832 

Unit tests and running the dataflow engine locally using the updated config file.

Users are able to use environment variables in the configuration.

### DIFF
--- a/rust/otap-dataflow/crates/config/src/env_substitution.rs
+++ b/rust/otap-dataflow/crates/config/src/env_substitution.rs
@@ -68,12 +68,23 @@ pub fn substitute_env_vars(input: &str) -> Result<String, Error> {
 
                     let value = match std::env::var(var_name) {
                         Ok(v) => v,
-                        Err(_) => match default {
+                        Err(env_var_error) => match default {
                             Some(d) => d.to_string(),
                             None => {
-                                return Err(Error::EnvVarNotFound {
-                                    var: var_name.to_string(),
-                                });
+                                match env_var_error {
+                                    std::env::VarError::NotPresent => {
+                                        // Variable is simply not set.
+                                        return Err(Error::EnvVarNotFound {
+                                            var: var_name.to_string(),
+                                        });
+                                    }
+                                    std::env::VarError::NotUnicode(_) => {
+                                        // Variable is set but contains invalid Unicode.
+                                        return Err(Error::EnvVarCannotBeParsed {
+                                            var: var_name.to_string(),
+                                        });
+                                    }
+                                }
                             }
                         },
                     };
--- a/rust/otap-dataflow/crates/config/src/error.rs
+++ b/rust/otap-dataflow/crates/config/src/error.rs
@@ -200,6 +200,14 @@ pub enum Error {
         var: String,
     },
 
+    /// An environment variable referenced in the configuration cannot be parsed and no default was given.
+    #[error("Environment variable `{var}` cannot be parsed and no default was provided")]
+    #[diagnostic(code(data_plane::env_var_cannot_be_parsed), url(docsrs))]
+    EnvVarCannotBeParsed {
+        /// The name of the environment variable that cannot be parsed.
+        var: String,
+    },
+
     /// An invalid user configuration occurred.
     #[error("An invalid user configuration occurred: {error}")]
     InvalidUserConfig {


rust/otap-dataflow/crates/config/src/engine/io.rs
```diff
@@ -27,8 +28,9 @@ impl OtelDataflowSpec {
 
     /// Creates a new [`OtelDataflowSpec`] with the given YAML string.
     pub fn from_yaml(yaml: &str) -> Result<Self, Error> {
+        let yaml = crate::env_substitution::substitute_env_vars(yaml)?;
```
albertlockett, line 31: Will this make it so we can substitute environment variables anywhere in the YAML, not just in yaml string values? For example:
```yaml
    resource:
      service.name: "fake-debug-noop-service"
      ${env:SOME_RESOURCE}: "${env:STATEFULSET_ORDINAL:-unknown}"
```
If so is this something we want to support? Will it cause any problems? I think it's probably OK.  

FWIW it seems the Otel Collector does not allow this, but that doesn't mean we're necessarily obligated to enforce the same restriction.
```
$ cat ~/Desktop/otel-col-exampl.yaml | head -5
receivers:
  ${env:ALBERT_TEST}:
    protocols:
      grpc:
        endpoint: localhost:4317

$ ALBERT_TEST=otlp ./bin/otelarrowcol --config ~/Desktop/otel-col-exampl.yaml
Error: failed to get config: cannot unmarshal the configuration: decoding failed due to the following error(s):

'receivers' '' in "${env:ALBERT_TEST}" id: invalid character(s) in type "${env:ALBERT_TEST}"
2026/03/04 08:37:42 collector server run finished with error: failed to get config: cannot unmarshal the configuration: decoding failed due to the following error(s):

'receivers' '' in "${env:ALBERT_TEST}" id: invalid character(s) in type "${env:ALBERT_TEST}"
```

andborja: I think it is a good feature to be able to configure blocks and not only values. This is specially useful under k8s solutions where the configuration is more dynamic.

cijothomas: if this is intentional design, lets capture that by adding test cases where we replace not just values.

andborja: Sure. Added.

rust/otap-dataflow/crates/config/src/env_substitution.rs
```diff
@@ -0,0 +1,241 @@
...
+                } else {
+                    // Not an `env:` provider — pass through verbatim.
+                    output.push_str(&input[i..end + 1]);
+                    i = end + 1;
+                }
+                continue;
+            }
+            // No matching `}` found — pass the `$` through as-is and advance
+            // by one so we do not loop forever.
+            output.push(bytes[i] as char);
```
gouslu: `bytes[i] as char` could corrupt non-ASCII characters

jmacd:
@andborja Thank you. I verified that this solution is in-line with others we could adopt--it's normal that environment variable expansion applied at the textual level before YAML or JSON parsing, making it possible to insert fragments of syntax, and we know that environment variables are relatively unsafe (hence `unsafe` keywords). I would rather not import the `envsubst` crate for a piece of code as small as this, and besides that crate doesn't do exactly what we want.

I have a style concern, having to do with the loop structure and the number of levels of indentation, it shows up as the twice-repeating stanza:

```
        output.push(bytes[i] as char);
        i += 1;
```

We can fix this to address the concern by @gouslu, meaning to use only utf8-safe find-and-copy operations. It's also an efficiency concern: if we structure the loop to find the next `$` character then most of the work is done by system libraries, it should lead to fewer conditionals and faster parsing. Something like this:

```
pub fn substitute_env_vars(input: &str) -> Result<String, Error> {
    let mut output = String::with_capacity(input.len());
    let mut rest = input;

    while let Some(pos) = rest.find('$') {
        output.push_str(&rest[..pos]);
        rest = &rest[pos..];

        if rest.starts_with("$$") {
            output.push('$');
            rest = &rest[2..];
            continue;
        }

        if rest.starts_with("${") {
            if let Some(close) = rest[2..].find('}') {
                /// (...) expand variable here
                continue;
            }
        }

        output.push('$');
        rest = &rest[1..];
    }

    output.push_str(rest);
    Ok(output)

```

rust/otap-dataflow/crates/config/src/env_substitution.rs
jmacd: Yes. Let's make sure we're testing what happens (1) when environment variables yield non-ascii characters, to be comprehensive, and (2) ~when configuration files contain non-ascii variables inside and outside of the ${...} syntactic markers.~

rust/otap-dataflow/crates/config/src/env_substitution.rs
```diff
@@ -0,0 +1,241 @@
...
+                let inner = &input[start..end]; // content between `${` and `}`
+
+                if let Some(rest) = inner.strip_prefix("env:") {
+                    // Split on the first `:-` to allow an optional default.
+                    let (var_name, default) = match rest.find(":-") {
+                        Some(pos) => (&rest[..pos], Some(&rest[pos + 2..])),
+                        None => (rest, None),
+                    };
+
+                    let value = match std::env::var(var_name) {
```
lquerel, line 69: The VarError returned by `std::env::var` is composed of 2 variants: NotPresent and NotUnicode.
We probably should keep this distinction in the error we are reporting.

See: https://doc.rust-lang.org/std/env/fn.var.html

rust/otap-dataflow/crates/config/src/env_substitution.rs
```diff
@@ -0,0 +1,241 @@
...
+                    output.push_str(&input[i..end + 1]);
+                    i = end + 1;
+                }
+                continue;
+            }
+            // No matching `}` found — pass the `$` through as-is and advance
+            // by one so we do not loop forever.
+            output.push(bytes[i] as char);
+            i += 1;
+            continue;
```
lquerel: The no-closing-brace path currently advances one byte and retries. For malformed input like `${${${...` this repeatedly rescans the tail and can become quadratic. We can preserve the same pass-through behavior while keeping linear time by appending the remaining slice once and exiting.

rust/otap-dataflow/crates/config/src/env_substitution.rs
andborja: Good catch. Updated.

rust/otap-dataflow/crates/config/src/env_substitution.rs
andborja: Updated based on @jmacd suggestion

rust/otap-dataflow/crates/config/src/env_substitution.rs
andborja: Updated

Issue #1710, opened 2026-01-02
Add HTTPS proxy support (TLS to proxy) for OTLP/OTAP exporters


lalitb:
The OTLP and OTAP gRPC exporters currently support proxying via HTTP CONNECT (PR #1679). However, many enterprise environments require the client-to-proxy connection itself to be encrypted, typically configured using an `https://` proxy URL.

Supporting HTTPS proxies involves more than accepting a different scheme. It requires establishing a TLS session with the proxy before issuing the CONNECT host:port request. This, in turn, requires explicit configuration for TLS behavior, including trusted root CAs (system or custom), hostname verification and SNI, and optionally client certificates for mTLS.

PR #2177, opened 2026-03-04
feat(otap): add HTTPS proxy support for gRPC exporters

lalitb:
Fixes: #1710 

  ### Description:

   Adds support for `https://` proxy URLs in the OTLP/gRPC and OTAP/gRPC exporters, enabling TLS (and optionally mTLS) connections to the proxy server itself.
  
 Previously, proxy URLs were restricted to `http://` - the connection to the proxy was always plaintext, even when the tunneled traffic to the backend was encrypted. This doesn't work in enterprise environments where proxy servers require TLS connections.

   ## Changes
   
   - **`proxy.rs`**: Accept `https://` proxy URLs (gated behind `experimental-tls`). Perform a TLS handshake with the proxy before sending the HTTP CONNECT request. Refactored `http_connect_tunnel_on_stream` to 
  be generic over `AsyncRead + AsyncWrite` instead of `TcpStream`-specific.
   - **`client_settings.rs`**: Updated connector return type to `ProxyTcpStream` (trait object supporting both plain and TLS streams).
   - **`proxy-support.md`**: Removed "No HTTPS to proxy" limitation, added HTTPS proxy configuration examples including the new `proxy.tls` block.
   
   ## Configuration
   
 
   ### Simple case — TLS to proxy, system CAs
   
 ```yaml
   nodes:
     exporter:
       type: exporter:otlp
       config:
         grpc_endpoint: "https://telemetry.backend.com:4317"
         proxy:
           https_proxy: "https://secure-proxy.corp.com:8443"
```

  Or via environment variable:
```bash
   export HTTPS_PROXY=https://secure-proxy.corp.com:8443
```

### Full case — separate TLS/mTLS for proxy and backend

```yaml
   nodes:
     exporter:
       type: exporter:otlp
       config:
         grpc_endpoint: "https://telemetry.backend.com:4317"
   
         # TLS/mTLS to the backend server (inside the tunnel)
         tls:
           ca_file: "/etc/ssl/certs/backend-ca.pem"
           cert_file: "/etc/ssl/certs/backend-client.pem"
           key_file: "/etc/ssl/private/backend-client-key.pem"
           server_name_override: "telemetry.backend.com"
   
         # HTTPS proxy with its own TLS/mTLS
         proxy:
           https_proxy: "https://secure-proxy.corp.com:8443"
           no_proxy: "localhost,*.internal,10.0.0.0/8"
           tls:
             ca_file: "/etc/ssl/certs/proxy-ca.pem"
             cert_file: "/etc/ssl/certs/proxy-client.pem"
             key_file: "/etc/ssl/private/proxy-client-key.pem"
             server_name_override: "secure-proxy.corp.com"
```

  This supports two independent trust chains — the proxy and backend can use different CAs, different client certificates, and different SNI names.

### DIFF
--- a/rust/otap-dataflow/crates/engine/Telemetry.md
+++ b/rust/otap-dataflow/crates/engine/Telemetry.md
@@ -1,0 +1,101 @@
+# Engine Crate Telemetry
+
+This document lists telemetry emitted directly by the `crates/engine` crate.
+It includes metric instruments registered by the crate and log events
+emitted via `otel_*` log macros.
+
+## Metrics
+
+| Metric name | Description | Produced in file |
+| --- | --- | --- |
+| `engine.metrics.memory_rss` | Process resident memory (RSS) in bytes. | `crates/engine/src/engine_metrics.rs` |
+| `channel.sender.send.count` | Number of messages successfully sent to a channel. | `crates/engine/src/channel_metrics.rs` |
+| `channel.sender.send.error_full` | Number of send attempts that failed because the channel was full. | `crates/engine/src/channel_metrics.rs` |
+| `channel.sender.send.error_closed` | Number of send attempts that failed because the channel was closed. | `crates/engine/src/channel_metrics.rs` |
+| `channel.receiver.recv.count` | Number of messages successfully received from a channel. | `crates/engine/src/channel_metrics.rs` |
+| `channel.receiver.recv.error_empty` | Number of receive attempts when the channel was empty. | `crates/engine/src/channel_metrics.rs` |
+| `channel.receiver.recv.error_closed` | Number of receive attempts after the channel was closed. | `crates/engine/src/channel_metrics.rs` |
+| `channel.receiver.capacity` | Configured channel buffer capacity. | `crates/engine/src/channel_metrics.rs` |
+| `pipeline.metrics.uptime` | Time since pipeline instance start. | `crates/engine/src/pipeline_metrics.rs` |
+| `pipeline.metrics.memory_usage` | Current heap memory in use by the pipeline thread (jemalloc only). | `crates/engine/src/pipeline_metrics.rs` |
+| `pipeline.metrics.memory_allocated` | Cumulative bytes allocated by the pipeline thread (jemalloc only). | `crates/engine/src/pipeline_metrics.rs` |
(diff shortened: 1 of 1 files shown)


gouslu:
Proxy support should be possible to configure at container/OS level. Why do we need at the component level for this case?

edit: I see that this is in addition to env based config, so you may ignore this.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
```diff
@@ -434,7 +517,228 @@ traffic to the final destination for https:// targets.",
...
+            "failed to read proxy.tls.{field} from {}: {e}",
+            path.display()
+        ))
+    })
+}
+
+#[cfg(feature = "experimental-tls")]
+async fn load_proxy_system_root_certs() -> Result<Vec<CertificateDer<'static>>, ProxyError> {
+    // Cache system roots for the process lifetime to avoid repeated blocking OS/keychain reads.
+    static SYSTEM_ROOTS: OnceCell<Vec<CertificateDer<'static>>> = OnceCell::const_new();
```
ThomsonTan, line 561: Seems these caches the roots for the very first connection via proxy, could it have problem if it has cert issue, but later connection will not do retry on it?

lalitb:
> Proxy support should be possible to configure at container/OS level. Why do we need at the component level for this case?

Can you clarify your question, as this seems to be missing some basic transport details?

For this specific flow:

```
Exporter -> HTTPS proxy -> HTTPS backend
```

container/OS env vars can select a proxy, but they do not, by themselves, express separate proxy-side TLS identity/trust settings versus backend TLS settings in the exporter config. How would you expect container-level proxy config alone to handle proxy-specific CA/client cert/SNI for this case?

Also, this is specific to the gRPC exporter path in this PR. For Geneva exporter and Azure Monitor (HTTP path via reqwest), OS/container-level proxy handling is already provided by the underlying client stack.

rust/otap-dataflow/crates/otap/src/otap_grpc/proxy.rs
lalitb: `SYSTEM_ROOTS` caches only successful loads (get_or_try_init). If loading system roots fails, the error is returned and nothing is cached, so later connections will retry initialization.

PR #2180, opened 2026-03-04
[otap-df-otap] Add error handling for arrow records build in Syslog CEF Receiver

### LINKED ISSUE #2182: Syslog CEF receiver can panic if there is an error constructing OTAP record batches
For example, from invalid UTF-8 sequences:
```
thread 'pipeline-perftest1-main-core-3' (238483) panicked at crates/otap/src/syslog_cef_receiver.rs:454:115:
Failed to build Arrow records: ArrowError(InvalidArgumentError("Encountered non UTF-8 data: invalid utf-8 sequence of 1 bytes from index 4"))
```

utpilla:
# Change Summary

@cijothomas reported intermittent errors logs saying that arrow records build within Syslog CEF receiver failed with the following error:
```
thread 'pipeline-perftest1-main-core-3' (238483) panicked at crates/otap/src/syslog_cef_receiver.rs:454:115:
Failed to build Arrow records: ArrowError(InvalidArgumentError("Encountered non UTF-8 data: invalid utf-8 sequence of 1 bytes from index 4"))
```

The PR adds error handling for building Arrow records.

Closes #2182

### DIFF
--- a/rust/otap-dataflow/crates/pdata/benches/concatenate.rs
+++ b/rust/otap-dataflow/crates/pdata/benches/concatenate.rs
@@ -3,8 +3,8 @@
 
 //! Benchmarks for concatenation and reindexing.
 //!
-//! Reindex benchmarks include contiguous and gapped (doubled IDs) variants
-//! to exercise different code paths.
+//! Reindex benchmarks include contiguous, gapped (doubled IDs), and unsorted
+//! variants to exercise different code paths.
 
 use std::ops::Mul;
 use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
utpilla, line 1: Note to reviewers: Please use the "Hide Whitespace" option for easier review.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -400,14 +400,21 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+                                                                OtapPdata::new_todo_context(arrow_records.into())
+                                                            );
+                                                            let mut m = metrics.borrow_mut();
+                                                            match &res {
+                                                                Ok(_) => m.received_logs_forwarded.add(items),
+                                                                Err(_) => m.received_logs_forward_failed.add(items),
+                                                            }
+                                                        }
+                                                        Err(e) => {
+                                                            otel_warn!("syslog_cef_receiver.arrow_records.build_failed", error = %e, message = "Failed to build Arrow records, dropping batch");
```
cijothomas, line 415: could the raw payload be debug logged too?

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
```diff
@@ -400,14 +400,21 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
+                                                            metrics.borrow_mut().received_logs_forward_failed.add(items);
+                                                        }
```
cijothomas, line 417: metrics looks good to me! The WARN might be too noisy since it'll print on each failed log.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
utpilla: The problem is that error type doesn't have any information about which entry in the batch caused the failure. All we know is batch couldn't be built into `OtapArrowRecords` and the reason it failed but not the problematic entry.

rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
utpilla: I also created #2181 to fix the handling of invalid UTF-8 str values. With the fix in #2181, it should be uncommon for arrow records build to fail.

So, using `otel_warn` should be okay now. In fact, it would help point out any weird issue if at all it comes up.

PR #2181, opened 2026-03-04
Gracefully handle invalid UTF-8 in binary_to_utf8_array

### LINKED ISSUE #1232: Gracefully handle invalid UTF-8 when constructing OTAP records
related to discussion https://github.com/open-telemetry/otel-arrow/pull/1203#discussion_r2395331926

Since https://github.com/open-telemetry/otel-arrow/pull/1203 if we encounter a single invalid UTF-8 sequence, we error when constructing the OTAP record.

We should handle this more gracefully to avoid dropping the entire batch when a single invalid UTF-8 sequence is produced.

To do this, we should look into doing a `from_utf8_lossy` in `otel_arrow_rust::encode::record::array::binary_to_utf8_array` instead of just using the default arrow array conversion, the idea being we keep the valid sequences and use the placeholder ? character for invalid sequences.

We should benchmark while working on this to ensure we get the optimal performance of the solution.

Since we may need to manually reconstruct the array depending on how the values buffer is converted, we should look for opportunities to optimize the array construction. For example, can we skip offset/dictionary key validation since we know those will be valid when converting from the BinaryArray

utpilla:
# Change Summary

### Problem

`binary_to_utf8_array` uses `StringArray::try_from_binary()` to convert binary columns to UTF-8 string columns. If **any** value in the array contains invalid UTF-8, this call fails and propagates an error that causes the **entire batch** to be dropped — even if only a single value is malformed. In an observability pipeline, losing an entire batch of valid records because of one bad message is unacceptable.

### Solution

Introduce a `binary_to_utf8_lossy` helper that uses a two-tier strategy:

1. **Fast path** — attempts `StringArray::try_from_binary()` first. When all values are valid UTF-8 (the common case), this is a zero-copy conversion with no additional allocation.
2. **Slow path** — if the fast path fails, falls back to per-value `String::from_utf8_lossy()`, which replaces invalid byte sequences with the Unicode replacement character (U+FFFD `�`) instead of returning an error.

Both `binary_to_utf8_array` (native `BinaryArray`) and `binary_dict_to_utf8_dict_array` (dictionary-encoded columns) now use this helper, so all signal types that flow through the OTAP encoding path are protected.

### Changes

- **`binary_to_utf8_array`** — calls `binary_to_utf8_lossy` instead of `StringArray::try_from_binary` directly.
- **`binary_dict_to_utf8_dict_array`** — same change on the dictionary values array.
- **New `binary_to_utf8_lossy`** — fast-path + lossy-fallback helper. Preserves nulls. Pre-allocates the `StringBuilder` using the source array's row count and byte length

* Closes #1232

Updated the existing invalid-UTF-8 test to expect lossy conversion instead of an error. Added new test cases for dictionary arrays with invalid values and null preservation during lossy conversion.

Yes, users would now see their invalid UTF-8 messages sanitized and getting exported successfully

### DIFF
--- a/rust/otap-dataflow/crates/pdata/benches/concatenate.rs
+++ b/rust/otap-dataflow/crates/pdata/benches/concatenate.rs
@@ -3,8 +3,8 @@
 
 //! Benchmarks for concatenation and reindexing.
 //!
-//! Reindex benchmarks include contiguous and gapped (doubled IDs) variants
-//! to exercise different code paths.
+//! Reindex benchmarks include contiguous, gapped (doubled IDs), and unsorted
+//! variants to exercise different code paths.
 
 use std::ops::Mul;
 use std::sync::Arc;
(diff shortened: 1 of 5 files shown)


albertlockett (approved):
LGTM! Thanks for fixing this @utpilla

lquerel (commented):
@utpilla @jmacd @albertlockett 
As a default behavior, that's probably the best thing to do. Longer term, I wouldn't be surprised if some users would prefer to have batches with invalid UTF-8 encoding routed to a DLQ or something similar. That's the kind of policy we could introduce in the future. Maybe we should track this in an issue.

Issue #2034, opened 2026-02-13
OPL/Columnar Query Engine: Support pipelines for nested types


albertlockett:
Some users may wish to use OPL to programatically modify nested fields within OTAP data. This includes things like Attributes or Metrics datapoints.

For example, imagine I wanted to redact attributes with the value hash if the values matched some regex. We don't currently have a way to do this. 

This could be built using a bespoke redaction operator, but I think it would be better implemented with something flexible.

I think we should design an operator that can take some non-parent `RecordBatch` batch and apply a transformation pipeline to it, and assign the result back to the OTAP batch.

For example, here is how it could look for the attribute redaction example discussed above:
```rust
logs 
| apply attributes {
  if (matches(value, '4[0-9]{12}(?:[0-9]{3})?') or matches(value, '(5[1-5][0-9]{14})'))) {
    set value = hash(value)
  }
}
```

TODO: I'll add more details to how I think this should work when I have more time to work out the details. For now, documenting the idea for posterity.

PR #2190, opened 2026-03-04
Columnar Query Engine support treating attributes as elements of a nested pipeline

albertlockett:
Current Status: Ready for Review ~WIP/Draft - after the 4th March SIG meeting, deciding we can rework the AST/Expression tree used to represent these nested pipelines as function invocations. This change is still in-progress~

# Change Summary

Adds support for an operation in OPL/Columnar query engine to treat attributes as elements in a nested stream and execute a pipeline on this stream. The result of the stream will be assigned back to the OTAP batch attributes.

Currently the only type of pipeline stage supported in the nested pipeline is filtering, but already this can be used to build some sophisticated attribute redaction capability.

Some simple examples
```kql
// keep only attributes where the value isn't a mastercard credit card number:
logs | apply attributes {
  where not(contains(value, "5198"))
}

// keep only k8s resource attributes:
logs | apply attributes {
  where matches(key, ".*k8s.*")
}
```

`value` in these expressions case is actually treated as a virtual column, and we determine which actual attribute value column to use in the expression based on the type on the other side of the binary expression.

~To support this functionality, this PR adds a new type of `DataExpression` to our AST expression crate called `NestedDataExpression`, which contains a source and a list of inner child `DataExpression`s.~ (<-- _this is no longer correct. On SIG meeting 4 March, decided to change how the AST is modelled_)

A new operator call is added to OPL called the `apply` operator call which contains a nested pipeline of operations to apply to some attributes record batch. These operations are parsed into the body of a `PipelineFunction`, and we use a `DataExpression::Transform(TransformExpression::Set(..))` where the source is an `InvokeFunctionExpression` referencing this function and the destination is the attributes record batch.

To support the types of operations we want to apply to attribute, which for this particular PR is simply filtering, this PR adds a `Discard` variant to the `PipelineFunctionExpression` enum. It implements handling of this type of expression in the record set engine, by simply returning `Null` from the function if such a statement is encountered and if the input to the statement does not pass the predicate. Because this also constitutes a statement from which we an return from the function, this statement is taken into account when calling `InvokeFunctionExpression::try_fold`.

Our `PipelinePlanner` is responsible for taking this expression/function definition and planning the pipeline. It converts the `PipelineFunctionExpression`s in the function body back into a list of `DataExpression`s. The planner now has a new mode (indicated by a flag) for whether it's planning for an attributes pipeline, or an OTAP batches pipeline. Internally, when it encounters the `NestedDataExpression`, it creates a new instance of itself and plans the inner expression in the attributes planning mode. 

The columnar query engine's planning stage also expects that `optimize` / `try_fold` has been called on the `DataExpression`s it receives. For this reason, we also implement calling `try_fold` on the expressions within the pipeline function body.

The `PipelineStage` trait now exposes two new methods: `execute_on_attributes` and `supports_exec_on_attributes`. The latter is used by the planner in attribute planning mode - if it plans a pipeline stage where this method returns false (which is the default in the trait definition), it returns a planning error. Only `PipelineStage` implementations that support executing on attributes will implement these methods (and return `true` from `supports_exec_on_attributes`).

The Filter pipeline stage is currently the only `PipelineStage` implementation that supports operating on attributes record batches. In this case, the `FilterPlan` is planned as normal, and then some new filter optimizers are used to combine the various `Composite<FilterPlan>` variants into a single datafusion logical `Expr`, and to change references to the virutal "value" column into the actual typed attribute value column.

A new `PipelineStage` impl called the `ApplyToAttributesPipelineStage` is what drives the inner pipeline. For each record batch it receives, it pulls out the attributes record batch, calls the chain of inner pipeline stages, and assigns the end result back to the OTAP batch.

* Related to #2034

There are new unit tests for the parser, and the new pipeline stage.

 This new OPL syntax would be available for users using the transform processor.

### DIFF
--- a/rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
+++ b/rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
@@ -1576,7 +1576,7 @@ impl InvokeFunctionScalarExpression {
             }
 
             if return_count == 1
-                // safety: we can "expect" on return_statement here because if return_count is 
+                // safety: we can "expect" one return_statement here because if return_count is 
                 // non-zero, then we've set this variable to `Some` in the loop above
                 && let ScalarExpression::Static(s) = return_statement.expect("return_statement not None")
             {


rust/experimental/query_engine/expressions/src/data_expressions.rs
```diff
@@ -359,6 +366,97 @@ pub enum OutputExpression {
...
+/// has list of nested objects, this expression can be used to express a transformation to the list
+#[derive(Clone, Debug, PartialEq)]
+pub struct NestedDataExpression {
+    query_location: QueryLocation,
+
+    /// steps to process the nested data
+    children: Vec<DataExpression>,
+
+    // selector of the input to the nested data pipeline
+    target: Option<SourceScalarExpression>,
```
albertlockett: TODO after discussion w/ @CodeBlanch change this to mutable value expression

albertlockett: After more discussion with @CodeBlanch - thinking we can probably actually get rid of this bespoke `NestedDataExpression`.

The set of child pipeline steps can be set up using a PipelineFunctionExpression/Implementation expression
https://github.com/open-telemetry/otel-arrow/blob/19ce05d29655224e526a2e1f241ae409c857ad75/rust/experimental/query_engine/expressions/src/pipeline_expression.rs#L321-L324
https://github.com/open-telemetry/otel-arrow/blob/19ce05d29655224e526a2e1f241ae409c857ad75/rust/experimental/query_engine/expressions/src/pipeline_expression.rs#L353-L357
alongside adding variants for Conditional/Discard in the `PipelineFunctionExpression` enum.

Then we'll just need to figure out how parse it into something that can be invoked by the columnar engine.

One option could be to parse this into parse it into a Set transform expression where the destination is the target MutableValueExpression and the source is a scalar InvokeFunction expression:
https://github.com/open-telemetry/otel-arrow/blob/19ce05d29655224e526a2e1f241ae409c857ad75/rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs#L46-L47
If that doesn't work, the other option could be to just have an `InvokeFunction` data expression

rust/experimental/query_engine/expressions/src/pipeline_expression.rs
```diff
@@ -89,15 +89,33 @@ impl PipelineExpression {
...
-                errors.push(e);
+        for func in &mut self.functions {
+            if let PipelineFunctionImplementation::Expressions(func_exprs) =
+                &mut func.implementation
+            {
+                for e in func_exprs {
+                    if let Err(e) = e.try_fold(&scope) {
+                        errors.push(e);
+                    }
+                }
```
albertlockett: @CodeBlanch working through the refactoring we discussed yesterday. Wondering if it's OK to call try_fold on the expressions inside the `PipelineFunctionExpressions`?

rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
```diff
@@ -1569,14 +1569,29 @@ impl InvokeFunctionScalarExpression {
...
+                    PipelineFunctionExpression::Discard(_d) => {
+                        discard_count += 1;
+                    }
+                    _ => {}
                 }
             }
 
             if return_count == 1
+                && discard_count == 0
                 && let ScalarExpression::Static(s) = return_statement.unwrap()
```
lquerel: Should we replace this unwrap with an error?

lquerel (approved):
There is this unwrap that I think should be converted into an error, otherwise it LGTM

rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
albertlockett: I agree that we should probably avoid using unwrap generally. However, in this case it's safe because if `return_count == 1`, we'll have initialized initialized `return_statement` to `Some`.

In 27ffc5a7b I switched this to `expect` and added a "safety" comment. I also added a test case that will have us hit this branch in a case where `return_statement` is `None` just for good measure.

rust/experimental/query_engine/expressions/src/pipeline_expression.rs
albertlockett: confirmed on SIG meeting that this is OK :+1:

rust/experimental/query_engine/expressions/src/pipeline_expression.rs
```diff
@@ -89,15 +89,68 @@ impl PipelineExpression {
...
+        }
+
+        for i in 0..self.functions.len() {
+            // There's a challenge here where `PipelineResolutionScope` needs to shared-borrow the
+            // functions, but we need to mutably borrow the expressions inside the function body so
+            // that they can be folded. The trick here is that we take the expressions from the
+            // body temporarily.
+            //
+            // The main limitation of this approach is that we will not to optimize an expression
+            // inside a recursive function by inspecting the body the function's body. However,
```
jmacd:
```suggestion
            // inside a recursive function by inspecting the function's body. However,
```

rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
```diff
@@ -1577,7 +1576,9 @@ impl InvokeFunctionScalarExpression {
             }
 
             if return_count == 1
-                && let ScalarExpression::Static(s) = return_statement.unwrap()
+                // safety: we can "expect" on return_statement here because if return_count is 
+                // non-zero, then we've set this variable to `Some` in the loop above
```
CodeBlanch:
```suggestion
                // safety: we can "expect" one return_statement here because if return_count is 
                // non-zero, then we've set this variable to `Some` in the loop above
```

PR #2195, opened 2026-03-05
Rename flush_timeout to max_batch_duration in batch processor and syslog receiver

cijothomas:
The name flush_timeout is misleading — "timeout" typically means "the maximum time allowed for an operation before it fails." However, this field bounds how long data can sit in a batch before being flushed — it's not a failure condition or a fixed periodic cadence.

An operator seeing flush_timeout: 5s may reasonably assume that if a flush takes longer than 5s, it will be aborted or logged as an error. That's not what happens — the field controls the maximum time the oldest item in a batch can wait before the batch is sealed and sent.

For context, the three OTel ecosystems each use different names for this concept:

OTel Collector batch processor: timeout — misleading, as in this codebase today.
OTel SDK spec (Batch Span Processor): schedule_delay / scheduledDelayMillis — accurate but SDK-specific.
This codebase (before this PR): flush_timeout — combines the right verb ("flush") with the wrong noun ("timeout").
max_batch_duration clearly communicates that this is the maximum time a batch will accumulate data before being flushed. The timer starts when the first item arrives into an empty batch, and the batch is sealed when either the duration elapses or the size threshold is reached — whichever comes first.

### DIFF
--- a/rust/experimental/query_engine/engine-recordset/src/scalars/scalar_expressions.rs
+++ b/rust/experimental/query_engine/engine-recordset/src/scalars/scalar_expressions.rs
@@ -411,6 +411,12 @@ where
 
                     for e in expressions {
                         match e {
+                            PipelineFunctionExpression::Discard(d) => {
+                                return Err(ExpressionError::NotSupported(
+                                    d.get_query_location().clone(),
+                                    format!("{} not supported in pipeline function", d.get_name()),
+                                ));
+                            }
                             PipelineFunctionExpression::Transform(t) => {
                                 execute_transform_expression(&func_execution_context, t)?
                             }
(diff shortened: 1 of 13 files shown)


JakeDern:
I'm not super strongly opinionated on it, but I don't think `flush_interval` is semantically correct either. To me, `flush_interval: 100ms` makes it sound like we flush regularly every 100ms, but that's not true.

The operation subject to "timeout" being referred to in this case is "accumulating enough data to flush". If we don't manage to accumulate enough data to flush before the timeout, then we seal the batch and send it anyway. 

Something that might be an improvement over both of these is `max_flush_interval` or `max_time_to_flush` indicating this is how long we'll wait before sealing a batch or the maximum time between flushes.

utpilla:
> The name `flush_timeout` is misleading — "timeout" typically means "the maximum time allowed for an operation before it fails."

I agree that `flush_timeout` is misleading.

> However, this field controls the interval after which pending data is flushed regardless of batch size — it's a periodic trigger, not a failure condition.

That's not the right interpretation. The actual semantic for the config in both the components is bounding how long data can sit in a batch. It's not fixed periodic cadence.

I would say `max_batch_duration` is a good name for this. It captures this more precisely: "a batch will be flushed if it has been accumulating for this long." It also avoids a potential misread where someone thinks the system flushes on a fixed schedule regardless of whether there's data.

lalitb:
+1 that `flush_timeout` is misleading.

Looking at the implementation, the timer is only scheduled when the buffer transitions from empty to non-empty. So the behavior is really bounding how long the oldest data can sit in a batch before it gets flushed, not a periodic interval.

`max_batch_duration` seems to capture that more precisely than `flush_interval`.

cijothomas:
Thanks for the feedback. Picking `max_batch_duration` in the PR

jmacd:
For the record, the term `flush_timeout` is verbatim from the Collector config in the exporterhelper batcher. (Batch processor is the legacy batcher, exporterhelper is new.)

https://pkg.go.dev/go.opentelemetry.io/collector/exporter/exporterhelper@v0.147.0/internal/queuebatch#BatchConfig

cijothomas:
> For the record, the term `flush_timeout` is verbatim from the Collector config in the exporterhelper batcher. (Batch processor is the legacy batcher, exporterhelper is new.)
> 
> https://pkg.go.dev/go.opentelemetry.io/collector/exporter/exporterhelper@v0.147.0/internal/queuebatch#BatchConfig

I had (have) the same concern with that name but never had the time to pursue it in Collector repo.

cijothomas:
@JakeDern can you re-review and see if the new name makes sense.

JakeDern:
@cijothomas I'm good with `max_batch_duration` too, thanks!

PR #2198, opened 2026-03-05
Batch processor: nack on inbound slot exhaustion instead of fatal error

### LINKED ISSUE #2194: Batch processor returns fatal error on inbound slot exhaustion and crashes engine
When the batch processor's inbound slots are fully occupied (default limit: 1024), the next subscribed message causes a fatal `ProcessorError("inbound slots not available")` that crashes the entire pipeline.

I think BatchProcessor should just `Nack` the request instead of returning Err.
Also, another question - if a processor returns Err, is it expected that engine crashes?

Repro: Run engine with this config:

```yaml
# Repro: batch processor crashes when inbound slots are exhausted.
# The retry processor subscribes (ACKS|NACKS) on every message, so
# has_subscribers()=true at the batch processor. With a long flush
# timeout the default 1024 inbound slots exhaust in ~2s, hitting a
# fatal ProcessorError("inbound slots not available") that kills the
# pipeline.

version: otel_dataflow/v1
engine: {}
groups:
  default:
    pipelines:
      main:
        policies:
          channel_capacity:
            control:
              node: 100
              pipeline: 100
            pdata: 128
          resources:
            core_allocation:
              type: core_count
              count: 1

        nodes:
          receiver:
            config:
              traffic_config:
                max_batch_size: 1
                signals_per_second: 5000
                log_weight: 100
          retry:
            type: processor:retry
            config: {}

          batch:
            type: processor:batch
            config:
              otap:
                min_size: 100000
                sizer: items
              flush_timeout: 10s

          noop:
            type: exporter:noop
            config:

        connections:
          - from: receiver
            to: retry
          - from: retry
            to: batch
          - from: batch
            to: noop
```

cijothomas:
# Change Summary

When the batch processor's inbound slot limit is reached, it returns a fatal error that crashes the entire pipeline. This changes it to nack the request instead, applying backpressure to upstream nodes.

* Closes #2194 

Unit test. The issue has a repro step, that is validated to be fixed.

Yes. Backpressure instead of crash.


rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -764,27 +767,29 @@ where
...
-            .transpose()?
-            .map(|(bc, _)| bc);
+        // If there are subscribers, calculate an inbound slot key. When inbound
+        // slots are exhausted, nack the request to apply backpressure.
+        let inkey = if ctx.has_subscribers() {
+            if self.buffer.inbound.is_full() {
+                self.metrics.nacked_inbound_slots.inc();
+                let refused = OtapPdata::new(ctx, payload.into());
+                effect
+                    .notify_nack(NackMsg::new("inbound slots not available", refused))
```
lalitb, line 782: Nit: this was the existing error message, but since this path represents capacity/backpressure rather than a generic failure, something like `inbound slots exhausted` or `inbound slot capacity exhausted` might be a bit clearer.

jmacd (approved):

drewrelmas (approved):

drewrelmas:
Looks like test is actually failing still, not quite ready for merge.

cijothomas:
Tagging @lalitb who is investigating fan_out for similar issues.

lalitb:
> Tagging @lalitb who is investigating fan_out for similar issues.


Should we be letting the back pressure propagate upstream as in #2225 instead of Nacking ?

JakeDern:
> > Tagging @lalitb who is investigating fan_out for similar issues.
> 
> Should we be letting the back pressure propagate upstream as in #2225 instead of Nacking ?

I think I'm up to date on #2225 now. I like the solution there better because it gives us the chance to clear space rather than immediately applying backpressure with a nack, but I don't think it will work for the current batch processor.

The batch processor doesn't know whether it can accept pdata or not until it knows what is in the pdata, this includes:

- Whether the pdata has subscribers (if it doesn't then we don't need to allocate a slot because we don't care about ack/nack): https://github.com/open-telemetry/otel-arrow/blob/6571763319b2cc89d899d81c5a9bf5e0e841e81d/rust/otap-dataflow/crates/otap/src/batch_processor.rs#L768-L788
- Which form the pdata is in, because we track OtapArrowRecords separate from ProtoBytes: https://github.com/open-telemetry/otel-arrow/blob/6571763319b2cc89d899d81c5a9bf5e0e841e81d/rust/otap-dataflow/crates/otap/src/batch_processor.rs#L421-L422
- Finally, which signal type the pdata is - We track inbound request context separately by signal as well: https://github.com/open-telemetry/otel-arrow/blob/6571763319b2cc89d899d81c5a9bf5e0e841e81d/rust/otap-dataflow/crates/otap/src/batch_processor.rs#L362-L366

I don't think we can implement `accept_pdata` in the batch processor given the current signature, we would need a ref to the pdata to decide. The pdata which would probably have to be buffered in ProcessorWrapper which would need us to take a couple branches on every message, something like:

- Check if we have a pdata buffered. 
    - If yes, check the processor's `accept_pdata()`.
        -  If `true`, call `process` on that
        -  If `false`, buffer the pdata and read from the message channel with `accept_pdata: false` and call `process` on that
    -  If no, read from the message channel and call `process` on that

Curious what folks think about this overall.

lquerel:
@JakeDern I don't have a complete solution to propose, but a simplification element that could perhaps be considered. If we could have just one table tracking the pending Acks/Nacks within the batch processor, then we wouldn't need to know about pdata to address points 2 and 3 (of your list of examples).

JakeDern:
> @JakeDern I don't have a complete solution to propose, but a simplification element that could perhaps be considered. If we could have just one table tracking the pending Acks/Nacks within the batch processor, then we wouldn't need to know about pdata to address points 2 and 3 (of your list of examples).

For sure consolidating to one table could be worth looking at to address those points, but I'm not sure how much batch processor code we'll have to rework to make sure ack/nack work with that as it's dependent on buffering signals separately and in a particular order. Maybe we can split the tables tracking pending acks/nacks from the Signal Buffers, but we still have to solve the context checking problem as well.

IMO this PR is still worth merging as the ack/nack based solution is simple and improves the situation significantly.

lalitb:
> IMO this PR is still worth merging as the ack/nack based solution is simple and improves the situation significantly.

I agree, the actual solution is non-trivial than I initially thought. This PR fix is the right pragmatic solution for now.

JakeDern:
Issue open to track implementing a better solution here: https://github.com/open-telemetry/otel-arrow/issues/2305

cijothomas:
> Issue open to track implementing a better solution here: #2305

I'll close this PR, in favor of 2305's exploration. Thanks @JakeDern

JakeDern:
@cijothomas We got our wires crossed 🙂 - The consensus was to merge this PR and then see if we can come up with a better solution later as it will be non-trivial. #2305 is tracking that future work

cijothomas:
@jmacd will pick this one up. Either same PR or a new PR.

PR #2199, opened 2026-03-05
Add dashboard to admin endpoint

cijothomas:
https://github.com/open-telemetry/otel-arrow/pull/2178 Following up to get a built-in dashboard endpoint (/dashboard) in admin panel. This auto refresh and show metrics.

Just a basic version to show core metric. If this is desired, we can improve and show all metrics, grouped/separate by core etc.

An alternate is modify the python script to serve this webpage, if there are concerns with embedding one inside engine itself.

image

### DIFF
--- a/rust/otap-dataflow/benchmarks/benches/exporter/main.rs
+++ b/rust/otap-dataflow/benchmarks/benches/exporter/main.rs
@@ -9,6 +9,7 @@ use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
 use fluke_hpack::Encoder;
 use otap_df_channel::mpsc;
 use otap_df_engine::{
+    Interests,
     config::ExporterConfig,
     exporter::ExporterWrapper,
     message::{Receiver, Sender},
@@ -454,7 +455,7 @@ fn bench_exporter(c: &mut Criterion) {
                     let local = LocalSet::new();
                     let _run_exporter_handle = local.spawn_local(async move {
                         exporter
-                            .start(node_req_tx, metrics_reporter)
+                            .start(node_req_tx, metrics_reporter, Interests::empty())
                             .await
                             .expect("Exporter event loop failed")
                     });
@@ -520,7 +521,7 @@ fn bench_exporter(c: &mut Criterion) {
                     let local = LocalSet::new();
                     let _run_exporter_handle = local.spawn_local(async move {
                         exporter
-                            .start(node_req_tx, metrics_reporter)
+                            .start(node_req_tx, metrics_reporter, Interests::empty())
                             .await
                             .expect("Exporter event loop failed")
                     });
@@ -591,7 +592,7 @@ fn bench_exporter(c: &mut Criterion) {
                     let local = LocalSet::new();
                     let _run_exporter_handle = local.spawn_local(async move {
                         exporter
-                            .start(node_req_tx, metrics_reporter)
+                            .start(node_req_tx, metrics_reporter, Interests::empty())
                             .await
                             .expect("Exporter event loop failed")
                     });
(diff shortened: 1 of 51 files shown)


lalitb:
Nice work on this @cijothomas ! Having a quick visual way to see engine metrics is really useful.
   
   One thought - since the engine already exposes `/metrics?format=prometheus`, we could also leverage an external dashboard tool like [Perses](https://perses.dev/) (CNCF sandbox project). The advantage is we get historical time-series, alerting support, and the dashboards are just JSON files in the repo - no Rust rebuild needed when the UI changes.
   
   I did a rough quick setup (with help from Copilot) to explore this idea:  [diff](https://github.com/open-telemetry/otel-arrow/compare/main...lalitb:otel-arrow:perses-monitoring-dashboard?expand=1). It's a Docker Compose stack (Prometheus + Perses) with a pre-built dashboard - `docker compose up -d && ./setup.sh` and you get 10 panels covering engine health, data flow, and runtime metrics.

image

cijothomas:
> Nice work on this @cijothomas ! Having a quick visual way to see engine metrics is really useful.
> 
> One thought - since the engine already exposes `/metrics?format=prometheus`, we could also leverage an external dashboard tool like [Perses](https://perses.dev/) (CNCF sandbox project). The advantage is we get historical time-series, alerting support, and the dashboards are just JSON files in the repo - no Rust rebuild needed when the UI changes.
> 
> I did a rough quick setup (with help from Copilot) to explore this idea: [diff](https://github.com/open-telemetry/otel-arrow/compare/main...lalitb:otel-arrow:perses-monitoring-dashboard?expand=1). It's a Docker Compose stack (Prometheus + Perses) with a pre-built dashboard - `docker compose up -d && ./setup.sh` and you get 10 panels covering engine health, data flow, and runtime metrics.
> 
> image

This is specifically to do things **without** additional dependency. Yes, any tool can be used to scrape metrics (given its in standard prom format).

lalitb:
> This is specifically to do things without additional dependency. Yes, any tool can be used to scrape metrics (given its in standard prom format).

That's a fair point about zero dependencies, and I see the appeal of a fully embedded developer experience - `cargo run` and open a browser, nothing else needed.
   
The Perses setup I shared is more aimed at longer debugging/perf sessions where you want historical trends and richer queries - and the only extra dependency is a docker container, so it's still pretty lightweight.
   
 My main suggestion for this PR would be to make the dashboard more dynamic - auto-discover metrics from `/metrics?format=json` by category (like `engine-metrics.py` does) rather than hardcoding  specific metric names in the HTML. That way it won't need updating every time we add new metrics.  And the pipeline group/pipeline tables already auto-discover dynamically, which is great.

cijothomas:
> > This is specifically to do things without additional dependency. Yes, any tool can be used to scrape metrics (given its in standard prom format).
> 
> That's a fair point about zero dependencies, and I see the appeal of a fully embedded developer experience - `cargo run` and open a browser, nothing else needed.
> 
> The Perses setup I shared is more aimed at longer debugging/perf sessions where you want historical trends and richer queries - and the only extra dependency is a docker container, so it's still pretty lightweight.
> 
> My main suggestion for this PR would be to make the dashboard more dynamic - auto-discover metrics from `/metrics?format=json` by category (like `engine-metrics.py` does) rather than hardcoding specific metric names in the HTML. That way it won't need updating every time we add new metrics. And the pipeline group/pipeline tables already auto-discover dynamically, which is great.

Yes. That (auto-discover) is the goal. I first want to know if we are even okay with an embedded html/js inside the engine before adding those.
This is just a basic version to show **core metric**. If this direction is agreed, we can improve and show **all** metrics (auto-discovered), grouped/separate by core etc.

cijothomas:
> The Perses setup I shared is more aimed at longer debugging/perf sessions where you want historical trends and richer queries - and the only extra dependency is a docker container, so it's still pretty lightweight.

Yes. Perses or any solution is still doable without this PR.

lalitb:
> Yes. That (auto-discover) is the goal. I first want to know if we are even okay with an embedded html/js inside the engine before adding those.
> This is just a basic version to show **core metric**. If this direction is agreed, we can improve and show **all** metrics (auto-discovered), grouped/separate by core etc.

Yes, the direction looks good to me. Good to see others feedback too.

PR #2200, opened 2026-03-05
PerfTest - modify saturation test script to detect and run selective tests

cijothomas:
Saturation test script now detects total cores in the machine, and filters out tests it cannot run due to lack of cores. This has no impact of lab runs, as our machine has sufficient cores. This is mostly to help local runs where we don't have such powerful machines. And I want to re-use the script as-is, to investigate the lack of scalability with number of cores.

### DIFF
--- a/tools/pipeline_perf_test/scripts/run-saturation-tests.sh
+++ b/tools/pipeline_perf_test/scripts/run-saturation-tests.sh
@@ -41,19 +41,49 @@ done
 
 cd "$PERF_TEST_DIR"
 
+# Detect available CPU cores on this machine
+AVAILABLE_CORES=$(nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || echo 0)
 echo "=============================================="
 echo "Running Saturation/Scaling Tests"
+echo "Detected $AVAILABLE_CORES CPU cores on this machine"
 echo "=============================================="
 echo ""
 
-# Find and run all saturation test configs
+# Compute total cores needed for a saturation test config.
+# Each config defines: num_cores (engine) + loadgen_cores + backend_cores
+# The pattern is: engine=N, loadgen=3*N, backend=N => total=5*N
+# We parse the actual values from each YAML to be safe.
+compute_required_cores() {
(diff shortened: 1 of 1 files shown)


tools/pipeline_perf_test/scripts/run-saturation-tests.sh
```diff
@@ -41,20 +41,50 @@ done
...
+# Each config defines: num_cores (engine) + loadgen_cores + backend_cores
+# The pattern is: engine=N, loadgen=3*N, backend=N => total=5*N
+# We parse the actual values from each YAML to be safe.
+compute_required_cores() {
+    local config="$1"
+    local engine loadgen backend
+
+    engine=$(grep -oP 'num_cores:\s*\K[0-9]+' "$config" 2>/dev/null || echo 0)
+    loadgen=$(grep -oP 'loadgen_cores:\s*\K[0-9]+' "$config" 2>/dev/null || echo 0)
+    backend=$(grep -oP 'backend_cores:\s*\K[0-9]+' "$config" 2>/dev/null || echo 0)
```
lalitb, line 62: `grep -P` won't work on MacOS, and and the `|| echo 0` fallback will silently mask the error:
```
lalitb@Lalits-MacBook-Pro-2 otap-dataflow % grep -P
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
lalitb@Lalits-MacBook-Pro-2 otap-dataflow % 
```

Good to use more portable alternative like `awk`.

cijothomas: Good point.
I'll tackle it in a future PR. Our perf tests don't work on Mac anyway today. I'll create a separate tracking issue for that. (something to do with df_engine inside docker+macOS! df_engine itself runs on Mac.)

PR #2209, opened 2026-03-05
Add instructions for AI Agents

cijothomas:
When asking Github's Copilot bot to fix trivial issues, it's missing some obvious things like fmt and running tests. This should make those better. Also added Claude.md as well, but to avoid duplication, all contents are in AGENTS.md.


rust/otap-dataflow/AGENTS.md
```diff
@@ -0,0 +1,16 @@
+# AGENTS.md — OTAP-Dataflow
```
AaronRM: We should also add instructions to:

- Use only ASCII characters in Markdown files. No smart quotes, em-dashes, or Unicode symbols.
- Run `markdownlint` to validate any changes to Markdown files.

rust/otap-dataflow/AGENTS.md
```diff
@@ -0,0 +1,16 @@
...
+
+Read and follow [CONTRIBUTING.md](CONTRIBUTING.md) for development guidelines,
```
AaronRM, line 3: Also suggest it consult rust/otap-dataflow/docs (and esp. `async-best-practices.md` and `events-guide.md`)?

rust/otap-dataflow/AGENTS.md
```diff
@@ -0,0 +1,16 @@
...
+coding conventions, and telemetry/logging rules.
+
+## After every code change
+
+After modifying any Rust file, **always** run this command before considering
+the task complete:
+
+```bash
+cargo xtask check
```
AaronRM: I know this AGENTS.md file is already in `rust/otap-dataflow`, but perhaps be explicit for agent instructions?

```
cd rust/otap-dataflow && cargo xtask check
```

rust/otap-dataflow/AGENTS.md
cijothomas: I wanted to start small to see if this actually improves things. I want to add more especially internal observability etc, after those guidelines are fully ironed out.

rust/otap-dataflow/AGENTS.md
cijothomas: thanks. yes I recall copilot putting non-ascii and then I had to ask again. Good point

tools/sanitycheck.py
```diff
@@ -2,18 +2,39 @@
...
 import glob
 import os
+import subprocess
 import sys
 
 CR = b'\r'
 CRLF = b'\r\n'
 LF = b'\n'
 
+# Build a set of git-ignored files to skip (e.g., build artifacts in target/).
```
cijothomas: I added this, so agents won't hit this issue. A lot of local files (like perf test results may have non-ascii etc.) It's not an issue in CI, but now we ask AGENTS to run this locally, want to avoid unnecessary noise about local only files.

.github/copilot-instructions.md
```diff
@@ -0,0 +1,3 @@
+# Copilot Instructions
+
+Read and follow all instructions in [rust/otap-dataflow/AGENTS.md](../rust/otap-dataflow/AGENTS.md).
```
lalitb: I think we should have project-specific instructions as documented here - https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions#creating-path-specific-custom-instructions as there are both Go and Rust projects here. 

Copilot will read .github/instructions/*.instructions.md files, when there are separate instructions for Go and Rust.

CLAUDE.md
```diff
@@ -0,0 +1,3 @@
+# CLAUDE.md
+
+Read and follow all instructions in [rust/otap-dataflow/AGENTS.md](rust/otap-dataflow/AGENTS.md).
```
lalitb: Same as [here](https://github.com/open-telemetry/otel-arrow/pull/2209/changes#r2898360806)

.github/copilot-instructions.md
cijothomas: see if the latest commit addresses this concern (I basically said - if working on Rust, then only...)

cijothomas:
https://github.com/open-telemetry/otel-arrow/actions/runs/22877245668/job/66371764529?pr=2244 Most common failure of introducing ASCII and failing to catch it via sanitycheck.py  - This PR should fix that!

rust/otap-dataflow/AGENTS.md
```diff
@@ -0,0 +1,41 @@
...
+
+After modifying any Markdown, YAML, or Python file, run:
+
+```bash
+python3 tools/sanitycheck.py
+```
+
+This checks for non-ASCII characters, trailing whitespace, consistent line
+endings (LF only), and correct indentation. Fix any errors before presenting
+the result.
```
reyang: What does "presenting the result" mean?

tools/sanitycheck.py
```diff
@@ -2,18 +2,39 @@
...
+_git_ignored_cache = None
+
+def is_git_ignored(filepath):
+    global _git_ignored_cache
+    if _git_ignored_cache is None:
+        _git_ignored_cache = set()
+        try:
+            result = subprocess.run(
+                ['git', 'ls-files', '--others', '--ignored', '--exclude-standard'],
```
reyang: This means we're spawning a process for every single file, I think perf is going to suffer a lot.

reyang: Oh, I see you're caching the files, interesting.

tools/sanitycheck.py
reyang: Can agents be intelligent enough to tell which file needs to be scanned?

cijothomas: let me think about this and make it efficient. If given enough iterations, agent do eventually fix these, but that is much more expensive than fixing it without having to do multiple iterations.

cijothomas:
> I have a general question - do we want to modify the CI/CD logic to handle coding agent specific scenarios? My gut feeling is no [#2209 (comment)](https://github.com/open-telemetry/otel-arrow/pull/2209#discussion_r2908461898) and would like to see more discussions.

This is needed irrespective of coding agents. When someone runs the sanitycheck.yml locally, its checks from repo root and includes all files, including non-git tracked. I'll come back with a better fix If I can.

reyang:
> > I have a general question - do we want to modify the CI/CD logic to handle coding agent specific scenarios? My gut feeling is no [#2209 (comment)](https://github.com/open-telemetry/otel-arrow/pull/2209#discussion_r2908461898) and would like to see more discussions.
> 
> This is needed irrespective of coding agents. When someone runs the sanitycheck.yml locally, its checks from repo root and includes all files, including non-git tracked. I'll come back with a better fix If I can.

When I do local work, here is my typical workflow:

1. make code changes without worrying about lint/sanity too much
2. build/test, if it doesn't meet my expectation, go back to step 1, otherwise move to step 3
3. git clean, run lint/sanity tools, prepare the PR

I assume AI would use similar flow?

rust/otap-dataflow/AGENTS.md
```diff
@@ -0,0 +1,41 @@
...
+directories).
+
+## Markdown rules
+
+- Use only ASCII characters in Markdown files. No smart quotes, em-dashes, or
+  Unicode symbols.
+- After editing Markdown files, run markdownlint on the changed files:
+
+```bash
+npx markdownlint-cli <file1.md> <file2.md>
```
ThomsonTan, line 16: This requires `node.js/npx` to be installed, but this prerequisite is not mentioned. An agent without Node.js would fail.

cijothomas:
@reyang @ThomsonTan Please re-review. I've pushed a new set of changes.

The sanitycheck.py is modified to ignore files not tracked by git. It does this by doing a single `git` call and caching. This is a change required irrespective of this PR.


> git clean, run lint/sanity tools, prepare the PR

git clean is not always possible. Especially, when we have python venv.

tools/sanitycheck.py
```diff
@@ -2,18 +2,35 @@
...
 CR = b'\r'
 CRLF = b'\r\n'
 LF = b'\n'
 
+# Only check git-tracked files so local build artifacts, venvs, perf results,
+# etc. are skipped automatically.  Single subprocess call, cached for all checks.
+_tracked_files = None
+
+def _is_tracked(filepath):
+    global _tracked_files
```
reyang: Curious, why taking this approach "updating a global variable inside a function, with potential race condition" rather than just putting it at the beginning of the script?

```python
_tracked_files = set(os.path.normpath(f) for f in subprocess.run(
                            ['git', 'ls-files'],
                            capture_output=True, text=True, check=True
                         ).stdout.splitlines())
```

cijothomas: agree. Simplified in latest commit.

rust/otap-dataflow/AGENTS.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+## Sanity checks
+
+After modifying any Markdown, YAML, or Python file, run:
+
+```bash
+python3 tools/sanitycheck.py
```
reyang, line 24: Is this covered by the CONTRIBUTING.md doc?

I think the goal is to have a self-contained CONTRIBUTING.md doc which serves as the single source of truth, and all contributors (whether human or AI bot) should follow it.
And if the coding bot isn't smart enough to find the CONTRIBUTING.md doc, the AGENTS.md will simply say "make sure you read and follow the CONTRIBUTING.md doc"? If different parts of the project have different guidelines, it makes sense to have specific CONTRIBUTING.md docs for subfolders.

lquerel:
@cijothomas Can we merge this PR?

cijothomas:
> @cijothomas Can we merge this PR?

Yes please.

PR #2211, opened 2026-03-05
[otap-df-telemetry] add duration metric helper for processors

### LINKED ISSUE #2210: [otap-df-telemetry] Processor duration measurement helper
The OTAP dataflow engine's trait for processors has a `process()`, called for each message received. We could measure the elapsed walltime of this operation, but it would not be a useful metric because it includes processing time and blocking time on the output channel(s).

Moreover, some processors do not define a meaningful time independent of the pipeline duration, which is measured at the node-level. 

For some processors, though, a simple duration measurement of the time spent processing will be useful. This should be under the control of the processor itself, since it is application-specific.

In the Collector, this sort of interface is called `processorhelper`, similar to `exporterhelper` and `receiverhelper`.

For our processors to measure duration, therefore, we should provide a telemetry helper API for timing. It makes sense to place this API on or near the current Mmsc (min-max-sum-count) instrument, and it makes sense to use a scoped call with a closure, the instrument can measure elapsed time and report a standard metric.

This would be added on a case-by-case basis to the processors.

jmacd:
Add a `node.processor` metric set with a `process.success.duration` and `process.failed.duration` Mmsc instrument for measuring the wall-clock duration of the work done in a process() call. A closure is used to prevent inclusion of async-await points in the measurement.

The metric is registered via the node telemetry context.

This is intended to be gated by MetricLevel >= Normal

Fixes #2210.

### DIFF
--- a/rust/otap-dataflow/crates/contrib-nodes/src/processors/recordset_kql_processor/processor.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/src/processors/recordset_kql_processor/processor.rs
@@ -187,7 +187,7 @@ impl RecordsetKqlProcessor {
     }
 
     fn process_logs(
-        &mut self,
+        &self,
         bytes: bytes::Bytes,
         signal: SignalType,
     ) -> Result<OtlpProtoBytes, Error> {


rust/otap-dataflow/crates/contrib-nodes/src/processors/condense_attributes_processor/mod.rs
```diff
@@ -208,20 +210,38 @@ pub static CONDENSE_ATTRIBUTES_PROCESSOR_FACTORY: otap_df_engine::ProcessorFacto
...
     }
 
     /// Condenses attributes in the given record batch according to the processor configuration.
     /// Returns the number of individual attributes that were condensed.
-    fn condense(&mut self, records: &mut OtapArrowRecords) -> Result<u64, Error> {
+    fn condense(&self, records: &mut OtapArrowRecords) -> Result<u64, Error> {
+        Self::do_condense(&self.config, records)
+    }
+
+    fn do_condense(config: &Config, records: &mut OtapArrowRecords) -> Result<u64, Error> {
```
utpilla: I think the simpler thing would be to keep the original `condense` method and remove `&mut self`:

```rust
fn condense(&self, records: &mut OtapArrowRecords) -> Result<u64, Error> {
```

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -0,0 +1,103 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Opt-in process-duration timing for processors.
+//!
+//! Processors that perform meaningful compute can add a
+//! [`ProcessDuration`] field and use either:
+//!
+//! - [`ProcessDuration::guard`] – a drop-guard that records the
```
utpilla: `ProcessDuration::guard` does not exist. The actual API is `start()` + `TimingGuard::stop()`, and the guard does not auto-record on drop, it silently discards.

rust/otap-dataflow/crates/contrib-nodes/src/processors/condense_attributes_processor/mod.rs
```diff
@@ -624,6 +652,8 @@ impl local::Processor<OtapPdata> for CondenseAttributesProcessor {
                     }),
                 };
 
+                timing.stop(&mut self.process_duration);
```
utpilla: There is an inconsistency in `TimingGuard::stop` calling pattern. Attributes processor does not call it in error path but condense processor and KQL processors call stop even in the error path.

I think we want to avoid calling it in the error path?

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -0,0 +1,103 @@
...
+//!   elapsed time when it goes out of scope.  This keeps the
+//!   original code structure intact (no closure / re-indentation).
+//! - [`ProcessDuration::timed`] – a scoped closure that records
+//!   duration on return.
+//!
+//! Both are gated on [`Interests::CONSUMER_METRICS`] (MetricLevel ≥ Normal).
```
utpilla: This works but the semantic connection isn't obvious. Process duration isn't really a "consumer metric". Would adding a dedicated interest bit be better? For example, `Interests::PROCESS_DURATION`?

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -0,0 +1,103 @@
...
+            Some(Timer::start())
+        } else {
+            None
+        };
+        TimingGuard { timer }
+    }
+
+    /// Time `f` if `interests` includes [`Interests::CONSUMER_METRICS`],
+    /// otherwise just call `f` without overhead.
+    pub fn timed<F, R>(&mut self, interests: Interests, f: F) -> R
```
utpilla: Do we need this method?

rust/otap-dataflow/crates/otap/src/filter_processor.rs
```diff
@@ -172,6 +179,9 @@ impl local::Processor<OtapPdata> for FilterProcessor {
                         filtered_arrow_records
                     }
                 };
+
+                timing.stop(&mut self.process_duration);
```
lalitb: This isn't reached if the `?` operators above bail early, so error durations aren't recorded. `CondenseAttributesProcessor` and `RecordsetKqlProcessor` always call stop() before result handling - should `FilterProcessor` follow the same pattern?

rust/otap-dataflow/crates/contrib-nodes/src/processors/condense_attributes_processor/mod.rs
jmacd: Fixed.

rust/otap-dataflow/crates/engine/src/process_duration.rs
jmacd: Lame! I think I fixed this.

rust/otap-dataflow/crates/contrib-nodes/src/processors/condense_attributes_processor/mod.rs
jmacd: Lame! Fixed.

rust/otap-dataflow/crates/engine/src/process_duration.rs
jmacd: Done. Added PROCESS_DURATION

rust/otap-dataflow/crates/engine/src/process_duration.rs
jmacd: Lame! removed.

rust/otap-dataflow/crates/otap/src/filter_processor.rs
jmacd: Redesigned.

jmacd:
e.g., 
```
# HELP process_duration_min Wall-clock duration of the processor compute section, in nanoseconds.
# TYPE process_duration_min gauge
process_duration_min{set="processor.process.duration",node_id="debug",node_urn="urn:otel:processor:debug",node_type="processor",pipeline_id="main",pipeline_group_id="default",core_id="19",numa_node_id="0",process_instance_id="AGOOAMAILN73HA4PQSJ3WPDKSA",host_id="JoshCorpSurfaceLaptop",container_id="",custom="{node_sequence=2, pipeline_name=MyCustomPipelineName}"} 4139070 1773287780962
# HELP process_duration_max Wall-clock duration of the processor compute section, in nanoseconds.
# TYPE process_duration_max gauge
process_duration_max{set="processor.process.duration",node_id="debug",node_urn="urn:otel:processor:debug",node_type="processor",pipeline_id="main",pipeline_group_id="default",core_id="19",numa_node_id="0",process_instance_id="AGOOAMAILN73HA4PQSJ3WPDKSA",host_id="JoshCorpSurfaceLaptop",container_id="",custom="{node_sequence=2, pipeline_name=MyCustomPipelineName}"} 4409694 1773287780962
# HELP process_duration_sum Wall-clock duration of the processor compute section, in nanoseconds.
# TYPE process_duration_sum counter
process_duration_sum{set="processor.process.duration",node_id="debug",node_urn="urn:otel:processor:debug",node_type="processor",pipeline_id="main",pipeline_group_id="default",core_id="19",numa_node_id="0",process_instance_id="AGOOAMAILN73HA4PQSJ3WPDKSA",host_id="JoshCorpSurfaceLaptop",container_id="",custom="{node_sequence=2, pipeline_name=MyCustomPipelineName}"} 8548764 1773287780962
# HELP process_duration_count Wall-clock duration of the processor compute section, in nanoseconds.
# TYPE process_duration_count counter
process_duration_count{set="processor.process.duration",node_id="debug",node_urn="urn:otel:processor:debug",node_type="processor",pipeline_id="main",pipeline_group_id="default",core_id="19",numa_node_id="0",process_instance_id="AGOOAMAILN73HA4PQSJ3WPDKSA",host_id="JoshCorpSurfaceLaptop",container_id="",custom="{node_sequence=2, pipeline_name=MyCustomPipelineName}"} 2 1773287780962
```

rust/otap-dataflow/crates/otap/src/filter_processor.rs
```diff
@@ -119,59 +132,73 @@ impl local::Processor<OtapPdata> for FilterProcessor {
                 let mut arrow_records: OtapArrowRecords = payload.try_into()?;
                 arrow_records.decode_transport_optimized_ids()?;
 
-                let filtered_arrow_records: OtapArrowRecords = match signal {
+                let _timing = self.process_duration.start(effect_handler.node_interests());
```
lquerel: I think that with this system we should measure only the **non-async sections**, meaning the blocks that do not contain `await`, in order to truly measure the processor’s compute section.

In this case (and in the other processors), the guard is kept alive across downstream `.await`. Under load, that folds queueing, wakeups, and backpressure delays into `process.duration`, so the metric stops being local to the processor and becomes partly a downstream congestion metric. This makes backpressure diagnosis materially harder and can mislead tuning decisions on hot processors such as `filter` and `attributes`.

We should scope the guard to the synchronous compute block only, for example:

```
let result = {
    let _timing = ...;
    compute(...)
};
```

and drop it before the first `.await`.

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -0,0 +1,100 @@
...
+        let timer = if interests.contains(Interests::PROCESS_DURATION) {
+            Some(Timer::start())
+        } else {
+            None
+        };
+        TimingGuard {
+            timer,
+            accumulator: Rc::clone(&self.accumulator),
+        }
+    }
```
lquerel: The feature is described as opt-in, but the disabled path still pays per-message overhead: every instrumented processor unconditionally calls start(...), and start always clones the Rc<Cell<Mmsc>> and constructs a guard even when PROCESS_DURATION is not enabled. It is still permanent hot-path work in MetricLevel::None/Basic, which is exactly where telemetry overhead should be near-zero.
We should make the disabled path allocation/refcount-free, for example with a TimingGuard::Disabled variant that carries no Rc, or gate before constructing the guard at the call site.

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -0,0 +1,100 @@
...
+use otap_df_telemetry::metrics::MetricSet;
+use otap_df_telemetry::reporter::MetricsReporter;
+use otap_df_telemetry_macros::metric_set;
+
+use crate::context::PipelineContext;
+
+/// Metric set containing a single process-duration histogram.
+#[metric_set(name = "processor.process.duration")]
+#[derive(Debug, Default, Clone)]
+pub struct ProcessDurationMetrics {
```
lquerel: I suggest to rename this struct `ProcessorComputeDuration` or `ProcessorWorkDuration` to avoid misinterpretation :-)

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -0,0 +1,100 @@
...
+    /// Wall-clock duration of the processor compute section, in nanoseconds.
+    #[metric(name = "process.duration", unit = "ns")]
```
lquerel: Then we should align the metric name accordingly.

rust/otap-dataflow/crates/otap/src/attributes_processor.rs
```diff
@@ -234,6 +237,7 @@ impl AttributesProcessor {
             has_scope_domain,
             has_signal_domain,
             metrics: None,
+            process_duration: None,
```
lquerel: And also rename those attributes with something like `compute_duration`.

lquerel (approved):
Few additional comments.

rust/otap-dataflow/crates/engine/src/process_duration.rs
jmacd: `ComputeDurationMetrics`

rust/otap-dataflow/crates/engine/src/process_duration.rs
jmacd: Now `compute.duration` in the MetricSet named `processor.compute`.

rust/otap-dataflow/crates/engine/src/process_duration.rs
jmacd: Fixed.

rust/otap-dataflow/crates/otap/src/filter_processor.rs
jmacd: Fixed.

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -0,0 +1,129 @@
...
+
+use crate::context::PipelineContext;
+
+/// Metric set containing a single compute-duration instrument.
+#[metric_set(name = "processor.compute")]
+#[derive(Debug, Default, Clone)]
+pub struct ComputeDurationMetrics {
+    /// Wall-clock duration of the processor compute section, in nanoseconds.
+    #[metric(name = "compute.duration", unit = "ns")]
+    pub compute_duration: Mmsc,
```
utpilla: I think we should split compute duration by outcome.

Currently `compute.duration` aggregates both successful and failed operations into a single Mmsc. This makes it hard to distinguish whether high durations come from actual compute work or error paths.

We could split this into two instruments: `compute.duration.ok` and `compute.duration.err`, and constrain `timed()` to `FnOnce() -> Result<T, E>` so it can route the elapsed time based on the outcome. All current callers already return `Result`, so no call-site changes would be needed.

```rust
#[metric_set(name = "processor.compute")]
pub struct ComputeDurationMetrics {
    #[metric(name = "compute.duration.ok", unit = "ns")]
    pub compute_duration_ok: Mmsc,

    #[metric(name = "compute.duration.err", unit = "ns")]
    pub compute_duration_err: Mmsc,
}
```

```rust
pub fn timed<T, E>(
    &self,
    interests: Interests,
    f: impl FnOnce() -> Result<T, E>,
) -> Result<T, E> {
    if interests.contains(Interests::PROCESS_DURATION) {
        let timer = Timer::start();
        let result = f();
        let elapsed = timer.elapsed_nanos();
        match &result {
            Ok(_) => { /* record into acc_ok */ }
            Err(_) => { /* record into acc_err */ }
        }
        result
    } else {
        f()
    }
}
```

This keeps the closure API intact, adds no overhead on the disabled path, and gives operators a clear signal when errors are contributing to duration spikes.

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -0,0 +1,206 @@
...
+use otap_df_telemetry_macros::metric_set;
+
+use crate::context::PipelineContext;
+
+/// Metric set containing compute-duration instruments split by outcome.
+#[metric_set(name = "processor.compute")]
+#[derive(Debug, Default, Clone)]
+pub struct ComputeDurationMetrics {
+    /// Wall-clock duration of successful compute sections, in nanoseconds.
+    #[metric(name = "compute.duration.success", unit = "ns")]
```
utpilla: `compute.duration.success` could be read as "computing duration was successful" rather than "duration of successful compute." Maybe reorder to `compute.success.duration` / `compute.error.duration`? That hopefully reads unambiguously as "the duration of successful/failed compute operations," with the qualifier scoping which compute and duration describing what's measured.

```rust
#[metric(name = "compute.success.duration", unit = "ns")]
pub compute_success_duration: Mmsc,

#[metric(name = "compute.error.duration", unit = "ns")]
pub compute_error_duration: Mmsc,
```

rust/otap-dataflow/crates/core-nodes/src/processors/filter_processor/mod.rs
```diff
@@ -119,59 +132,68 @@ impl local::Processor<OtapPdata> for FilterProcessor {
...
+                                        Error::ProcessorError {
+                                            processor: effect_handler.processor_id(),
+                                            kind: ProcessorErrorKind::Other,
+                                            error: format!("Filter error: {e}"),
+                                            source_detail,
+                                        }
+                                    })?;
+
+                                self.metrics.log_signals_consumed.add(log_signals_consumed);
+                                self.metrics.log_signals_filtered.add(log_signals_filtered);
```
utpilla: nit: The counter updates (`self.metrics.log_signals_consumed.add(...)`, `self.metrics.span_signals_consumed.add(...)`, etc.) are inside the `timed()` closure, so they're included in the compute duration measurement. These are cheap `add()` calls so the impact is minimal, but for purity the counters could be updated after the closure returns (similar to how the attributes processor updates its counters outside `timed()`).

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -0,0 +1,189 @@
...
+}
+
+/// Wrapper providing interests-gated duration recording and reporting.
+pub struct ComputeDuration {
+    metrics: MetricSet<ComputeDurationMetrics>,
+    /// Accumulator for successful durations.
+    /// Uses `RefCell` for interior mutability so `timed` can take `&self`,
+    /// allowing callers to hold shared borrows of sibling fields in the
+    /// closure.  `RefCell` is used instead of `Cell` to support future
+    /// instrument types that are not `Copy` (e.g. histogram buckets).
```
utpilla: nit: Compared to `Cell` which has zero runtime borrow-check cost, `RefCell` does come with a runtime borrow-check cost for every `borrow_mut` call. I understand that `Cell` might not help in future when we move to Histograms, but I guess that switch could have been made when we actually needed it.

rust/otap-dataflow/configs/fake-kql-debug-noop.yaml
```diff
@@ -1,4 +1,10 @@
 version: otel_dataflow/v1
+
+policies:
+  telemetry:
+    # Normal level enables processor compute-duration metrics (e.g. KQL).
```
utpilla:
```suggestion
    # 'normal' metric level enables processor compute-duration in addition to channel metrics.
```

rust/otap-dataflow/crates/engine/src/process_duration.rs
jmacd: I moved to compute.success.duration and compute.failed.duration.

rust/otap-dataflow/crates/engine/src/process_duration.rs
jmacd: Done

rust/otap-dataflow/crates/core-nodes/src/processors/debug_processor/mod.rs
```diff
@@ -325,36 +332,46 @@ impl local::Processor<OtapPdata> for DebugProcessor {
...
                 match otlp_bytes {
                     OtlpProtoBytes::ExportLogsRequest(bytes) => {
                         if active_signals.contains(&SignalActive::Logs) {
-                            let req = LogsData::decode(bytes.as_ref()).map_err(|e| {
-                                Error::PdataConversionError {
-                                    error: format!("error decoding proto bytes: {e}"),
-                                }
-                            })?;
+                            let req =
+                                effect_handler.timed(&self.compute_duration, || {
```
lalitb: This timing only covers protobuf decode. Most of the debug processor’s actual work happens afterward in process_log / process_metric / process_trace, so this metric will underreport the real processing cost.

rust/otap-dataflow/crates/engine/src/process_duration.rs
jmacd: Replaced with Cell<Mmsc> for now!

rust/otap-dataflow/crates/core-nodes/src/processors/filter_processor/mod.rs
jmacd: I agree, fixed. And also-- I sort of don't care whether metric updates are included in compute time (a) because the distortions will be small, (b) because compute is compute. 😀

rust/otap-dataflow/crates/core-nodes/src/processors/debug_processor/mod.rs
jmacd: I've gone back and forth on this! The blocking calls to the console will appear as compute if we're not careful, but I guess the work in formatting can be measured. Not sure if it's worth it. I mainly added this in the debug procsesor to test the code path because it's used in many test configs.

lalitb: Yeah it's just an observation, not blocking specifically for debug processor.

PR #2213, opened 2026-03-05
feat(metrics): recompute durable_buffer queued item gauges on telemetry tick

### LINKED ISSUE #2193: Durable_buffer queued gauges can be inflated after restart due to incomplete subscriber progress restoration
As noted in [an earlier PR comment](https://github.com/open-telemetry/otel-arrow/pull/2103#discussion_r2844033034) from @ThomsonTan, the `durable_buffer` processor’s queued gauges are seeded on startup by scanning persisted segments and summing bundle item counts, but that scan does not account for subscriber progress within those segments. If the previous run ACK'ed some bundles (or partially processed a segment) before shutdown, the underlying segment files can still exist and will be counted in full, inflating queued items. This produces a temporary over-count until normal processing catches up and maintenance cleans segments. The issue is metric accuracy on restart due to missing subscriber state in the seed path, not data loss or delivery semantics.

AaronRM:
# Change Summary

Replace incremental queued-item counters (which drifted on force-drops and segment expiry) with a periodic full-recompute model that derives gauges from subscriber progress bitmaps and the open segment snapshot.

Key changes:
- Add recompute_queued_counters() called at engine init and every CollectTelemetry tick, replacing all incremental gauge updates from ingest/ACK/NACK paths
- Introduce segment-level metadata caching (`SegmentMetricsSummary`, `CachedSegmentMetrics`) with LRU eviction bounded at 4096 entries
- Add `OpenSegmentBundleSummary` and `snapshot_bundles()` for visibility into the open (accumulating) segment
- Add `pending_segment_progress() to subscriber registry for obtaining a point-in-time snapshot of per-segment ACK/NACK state
- Add once-per-segment warning suppression for metadata load failures

The new model is self-correcting: even if a segment is force-dropped or expires, the next recompute cycle produces accurate counts from the current ground truth.

* Closes #2193 

- Updated unit tests covering gauge accuracy
- Local integration testing with the full engine showed no gauge drift under sustained backpressure and negligible recompute overhead

`durable_buffer` queued* gauges will show improved accuracy.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/durable_buffer_processor/mod.rs
+++ b/rust/otap-dataflow/crates/otap/src/durable_buffer_processor/mod.rs
@@ -720,6 +720,11 @@ impl DurableBuffer {
     ///
     /// The gauges reflect all ingested but not-yet-ACKed items across both sources.
     ///
+    /// *Snapshotting is intentionally non-atomic across these two sources.* During
+    /// segment finalization there is a brief window where a just-finalized segment
+    /// may appear in neither snapshot, causing a temporary under-count. Gauges
+    /// will self-correct on the next telemetry tick.
+    ///
     /// Uses a segment-level cache (`segment_cache`) for finalized segments to avoid
     /// per-tick allocations. Segments are immutable after finalization, so the
     /// per-bundle `(item_count, signal_type)` classification is computed once and


rust/otap-dataflow/crates/otap/src/durable_buffer_processor/mod.rs
```diff
@@ -648,63 +710,183 @@ impl DurableBuffer {
...
-        if total > 0 {
-            otel_info!(
-                "durable_buffer.queued.seeded",
-                logs = self.queued_log_records,
-                metrics = self.queued_metric_points,
-                traces = self.queued_spans,
-                total = total
-            );
+        // Step 3: add items from the open (accumulating) segment.
+        let open_bundles = engine.open_segment_bundle_summaries();
```
lalitb, line 865: @AaronRM quick question: is the brief under-count during segment finalization an intentional tradeoff? My read is that gauges self-correct on the next telemetry tick, and making the finalized/open-segment snapshots atomic would likely add contention to ingest. If that’s the intent, could we add a short note in `recompute_queued_counters` to document it?

AaronRM: @lalitb Yes, thanks for noting this. Yes, it is intentional because maximizing throughput and minimizing potential lock contention are more important concerns than perfect metric precision. I've added a comment to `recompute_queued_counters` about the tradeoff.

PR #2214, opened 2026-03-05
fix: repair arm64 Docker build by switching to musl.cc toolchain  

lalitb:
## Problem                                                                                                                                                                                                                                                                                                         
                                                    
  The `aarch64-unknown-linux-musl` build was silently compiling C code against                                                                                                                                                                                                                                       
  glibc headers while linking against musl libc.                       
                                                                                                                                                                                                                                                                                                                     
  `rust:1.94` ships Debian 13 / glibc 2.40+, where `_GNU_SOURCE` implies
  `_ISOC23_SOURCE`, redirecting `strtol` and `sscanf` to C23 variants.
  `aws-lc-sys/bcm.c` defines `_GNU_SOURCE`, so its object files reference
  `__isoc23_strtol` and `__isoc23_sscanf` — symbols that only exist in glibc,
  not musl. Linker explodes:

  undefined reference to __isoc23_strtol' undefined reference to __isoc23_sscanf'

  The culprits:
  - **Native arm64 builds** (macOS M1, Linux arm64 runners): no C compiler was
    explicitly set, Cargo fell back to the system `gcc` which uses glibc headers
  - **Cross amd64 -> arm64 builds**: `gcc-aarch64-linux-gnu` (Debian's glibc-based
    cross-compiler) brings glibc system headers into what is supposed to be a
    musl build

  ## Fix

  Split the arm64 path on `BUILDPLATFORM`:

  - **Native arm64** (`BUILDPLATFORM=linux/arm64`): use the musl.cc
    `aarch64-linux-musl-native.tgz` toolchain — an arm64-hosted compiler with
    its own musl headers. `musl-tools` from apt was considered but rejected
    because its unprefixed `musl-gcc` confuses jemalloc's configure into
    misdetecting atomics support. Also adds a symlink
    `aarch64-linux-musl-ar -> aarch64-linux-musl-gcc-ar` since the native tarball
    only ships the latter but cmake-based crates expect the former.

  - **Cross amd64 -> arm64** (`BUILDPLATFORM=linux/amd64`): use the musl.cc
    `aarch64-linux-musl-cross.tgz` toolchain - an x86_64-hosted cross-compiler
    with its own musl headers, replacing `gcc-aarch64-linux-gnu`. Tarball
    integrity verified with `sha256sum`.

  Both toolchains avoid glibc headers entirely, eliminating the C23 redirect.

## Testing                                                                                                                                                                                                                                                                                                         
                                                                       
  | Scenario | BUILDPLATFORM | TARGETPLATFORM | Tested on |                                                                                                                                                                                                                                                          
  |---|---|---|---|                                                    
  | Native arm64 | `linux/arm64` | `linux/arm64` | macOS Apple Silicon (Docker Desktop) - `docker buildx build --platform linux/arm64` |
  | Cross amd64→arm64 | `linux/amd64` | `linux/arm64` | macOS Apple Silicon (Docker Desktop) - `docker buildx build --platform linux/arm64 --build-arg BUILDPLATFORM=linux/amd64` |
  | Native amd64 | `linux/amd64` | `linux/amd64` | Not tested locally - unchanged code path, covered by CI `ubuntu-24.04` runner |

  Note: Scenario 3 was simulated on macOS by overriding BUILDPLATFORM via --build-arg. The x86_64 cross-compiler binaries ran under Docker Desktop's binfmt support. The true cross-compile path will be exercised by CI on an amd64 runner

### DIFF
--- a/rust/experimental/query_engine/engine-recordset/src/scalars/scalar_expressions.rs
+++ b/rust/experimental/query_engine/engine-recordset/src/scalars/scalar_expressions.rs
@@ -411,6 +411,12 @@ where
 
                     for e in expressions {
                         match e {
+                            PipelineFunctionExpression::Discard(d) => {
+                                return Err(ExpressionError::NotSupported(
+                                    d.get_query_location().clone(),
+                                    format!("{} not supported in pipeline function", d.get_name()),
+                                ));
+                            }
                             PipelineFunctionExpression::Transform(t) => {
                                 execute_transform_expression(&func_execution_context, t)?
                             }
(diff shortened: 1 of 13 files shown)


lalitb:
converting to draft - I believe there may be much easier solution by switching to Alphie based image. Will try out and update here.

lalitb:
> I believe there may be much easier solution by switching to Alphie based image. Will try out and update here.

Ok tested this. Switching to Alpine for the build image doesn't actually eliminate the muse.cc dependency for cross-compilation here. The current approach (Debian + must.cc for arm64) is the right call for the project given the cross-arch build requirement. Alpine would only be a clean win if you dropped cross-compilation support entirely. 

Making this PR ready for review.

cijothomas:
https://github.com/open-telemetry/otel-arrow/actions/runs/22930021955/job/66549239123 Broken nightly builds (to calculate binary size), which would be fixed with this. Thanks @lalitb

rust/otap-dataflow/cross-arch-build.sh
```diff
@@ -4,15 +4,30 @@ if [ "${TARGETPLATFORM}" = "linux/amd64" ]; then
...
-  fi
+  # Always use the musl.cc cross-toolchain for arm64, regardless of whether this
+  # is a native or cross build. The Debian gcc-aarch64-linux-gnu package (and
+  # musl-gcc wrappers) use glibc system headers; glibc >= 2.38 maps _GNU_SOURCE
+  # to _ISOC23_SOURCE, which redirects strtol/sscanf to __isoc23_strtol/
+  # __isoc23_sscanf. Those C23 symbols don't exist in musl libc, causing link
+  # failures in C dependencies that define _GNU_SOURCE (e.g. aws-lc-sys/bcm.c).
+  # The musl.cc toolchain ships its own musl headers, so this redirection never
+  # occurs, fixing the issue cleanly for both native and cross builds.
+  MUSL_CROSS_TARBALL=aarch64-linux-musl-cross.tgz
```
ThomsonTan: Is there a versioned tarball? With the current URL, our CI will break once there is a new version of toolchain released, and we need udpate the SHA256 here.

rust/otap-dataflow/cross-arch-build.sh
ThomsonTan, line 1: Set pipefail too?

rust/otap-dataflow/cross-arch-build.sh
jmacd: Curious: isn't there a more appropriate way to get this dependency?

lquerel:
There’s no reason for you to change this configuration, but just in case: you should never use a statically compiled binary with musl without using an allocator like jemalloc or mimalloc. The default musl allocator is incredibly slow.

lalitb:
> There’s no reason for you to change this configuration, but just in case: you should never use a statically compiled binary with musl without using an allocator like jemalloc or mimalloc. The default musl allocator is incredibly slow.

Agreed, thanks. We already use `jemalloc` as the global allocator, so this PR does not change allocator behavior for the musl build.

Also fixed a separate follow-up issue identified now by splitting the arm64 path: native arm64 builds now use `musl-tools`, and only amd64 -> arm64 cross-builds use the `musl.cc` toolchain.

rust/otap-dataflow/cross-arch-build.sh
lalitb: I don’t see a versioned artifact URL on musl.cc. The site appears to publish fixed tarball names plus checksums/signatures, so with the current layout the best we can do here is pin the exact SHA256 and fail closed if the artifact changes.

For the amd64 -> arm64 musl cross-build path, I didn’t find a better distro-provided option that avoids the glibc-header/C23 issue, which is why this PR uses the musl.cc toolchain. So this is meant as the minimal fix rather than the ideal long-term dependency model.

If we want to harden it further, the next step would be to mirror the tarball in a location we control, or switch to a pinned toolchain/container artifact that we own. I can file a follow-up for that if that seems worthwhile. The good thing is that updates do not seem frequent; the currently published main toolchain artifact is several years old.

rust/otap-dataflow/cross-arch-build.sh
lalitb: We can, but pipefail is a bash-ism, so adding it cleanly would also mean making the script explicitly bash. Since this script is used in the Debian Rust build image that should be fine, but if we want to keep it shell-portable I’d leave it as-is.

Issue #2207, opened 2026-03-05
Missing error handling in otap exporter


JakeDern:
There are several missing error handling paths in the otap exporter that can lead to silent failure without any metrics or logs reporting.

If the other half of the channel is closed, then this silently does nothing and drops data. Probably shouldn't happen, but I don't think we gain much by suppressing it:

https://github.com/open-telemetry/otel-arrow/blob/4185b93a1656879195dab482af90287a8130984b/rust/otap-dataflow/crates/otap/src/otap_exporter.rs#L248-L252

Producing the BARs here can also fail silently as can sending the metrics back up:

https://github.com/open-telemetry/otel-arrow/blob/4185b93a1656879195dab482af90287a8130984b/rust/otap-dataflow/crates/otap/src/otap_exporter.rs#L434-L455

JakeDern:
For what it's worth I did actually hit these and it was kind of painful 😁 - We should probably do a general pass on the error handling in the exporter.

gyanranjanpanda:
i would like to do this

JakeDern:
~Looks like the relevant code is still there, I can try to take a peek at this~

Actually, I was looking on mobile before which took me to an old commit. I think this has all been mostly reworked.

JakeDern:
Closing as stale, see above comment.

Issue #2022, opened 2026-02-11
Proposal: Improve OTAP PData Record Batch API


albertlockett:
### Problem Statement

Imagine I'm a contributor who wants to write a new processor for the df-engine. I want the processing to be as efficient as possible, so I choose to do my processing directly on the Arrow `Array`s. As I unpack the pdata to access the telemetry data, eventually I will call `OtapArrowRecords::get` and be given an arrow `RecordBatch`.

When considering the `RecordBatch` from this perspective, there are some implicit assumptions and unknowns which make it hard to work with this data. The `RecordBatch` contains only what could be called the "physical schema". E.g. which columns are actually present, and what are their types. 

However, It has no indication to the user of:
- Which columns are guaranteed to be present, and which are optional
- Which optional columns could be present, but are not
- Which columns are guaranteed to be a certain type
- Which columns, presenting as a given type, but may take on different types in certain situations
- Whether the ID columns are encoded

As a concrete example, imagine I inspect my telemetry data and see `RecordBatch` representing Resource Attributes. This RecordBatch may have the following schema:
```
"parent_id" - UInt16
"type"      - UInt8
"key"       - Dictionary<UInt8, Utf8>
"str"       - Dictionary<UInt16, Utf8>
```

Unless I was very familiar with the OTAP spec, it's not clear at all how I need to handle each column. For example, I may ask questions such as:
- Are there additional values columns?
- Is the type column always guaranteed to be present?
- Is the key column always dictionary encoded?
- Can I access the parent_id column directly, or are the values encoded?

A contributor may not even know they need to ask these questions, and in doing so may make invalid assumptions about the structure of the `RecordBatch`, leading to bugs (this has happened).

Moreover, from a long-term maintenance perspective there is an additional difficulty with the current level of abstraction: if we adapt the protocol to add new fields, or new arrow types to represent existing fields, it's difficult to grep the code base and determine all the places we're working with a `RecordBatch` representing some payload type that may break. For example, there are some places in the code where we assume attributes str/int column will never be a dict keyed by u8. It's not clear how to to find these these locations if we decide to change that rule.

### Proposed Solution

I feel there's probably room for an additional layer of indirection between OtapArrowRecords, which represents a single OTAP batch containing OTAP data in arrow format, and the RecordBatch for an individual payload type.

The objectives of such an abstraction would be:
- Provide some additional clarity over the structure of Arrow data (the "logical schema")
[...]

PR #2215, opened 2026-03-06
Enforce OTAP dictionary key sizes in concatenate (fixes #2204)

### LINKED ISSUE #2204: Concatenate does not take into account the otap spec when picking dictionary sizes
The concatenate code in concatenate.rs has to pick a unifying schema. This entire process is currently completely generic and does not take the spec into account at all. This can cause problems by picking dictionary key sizes that aren't allowed.

For example, if we're concatenating two attributes `str` columns where at least one is dictionary encoded, we could pick Dictionary<u8, UTF8> as the encoding even though we only technically allow Dictionary<u16, UTF8>.

This can cause downstream errors.

gyanranjanpanda:
## Summary

The generic cardinality logic in `select_dictionary_type()` picks the smallest dictionary key type that fits the data, but the OTAP spec (OTEP 0156) mandates **`UInt16` keys for string and binary dictionary columns**. This can produce `Dictionary<UInt8, Utf8>` when it should be `Dictionary<UInt16, Utf8>`, causing downstream IPC errors.

## Changes

Single file: `crates/pdata/src/otap/transform/concatenate.rs`

- **`needs_u16_key()`** — identifies value types subject to the OTAP UInt16 key constraint (Utf8, LargeUtf8, Utf8View, Binary, LargeBinary, BinaryView)
- **`fix_dict_key()` / `fix_dict_key_in_field()`** — recursively corrects dictionary key types in data types and fields, descending into composite types (List, Struct, Map)
- **`enforce_otap_dict_keys()`** — post-processes the unified schema to apply the fix
- **One-line wiring** at the `concatenate()` call site: `enforce_otap_dict_keys(&select_schema(&index)?)`

## Design Decision

This is a **post-processing** fix rather than modifying `select_dictionary_type()` directly. The generic cardinality estimation remains unchanged — OTAP spec enforcement is applied as a separate layer. This keeps the two concerns cleanly separated.

## Tests

- All 39 existing concatenate tests pass (updated to expect UInt16 for string dicts in the concat path)
- 2 new regression tests:
  - `regression_2204_str_dict_gets_u16_key` — the exact scenario from the issue
  - `int32_dict_keeps_u8_key` — non-string dicts are unaffected

Fixes #2204

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
@@ -20,7 +20,7 @@ use crate::error::Error;
 use crate::otap::Result;
 use crate::schema::consts::metadata::COLUMN_ENCODING;
 use crate::schema::consts::metadata::encodings::PLAIN;
-use crate::schema::consts::{ID, PARENT_ID};
+use crate::schema::consts::{self, ID, PARENT_ID};
 
 /// These are one less than the maximum cardinality of the key type. We should be
 /// able to go up to 256/65536 without overflow, but there is a bug in arrow-rs
@@ -92,7 +92,7 @@ pub fn concatenate<const N: usize>(
             continue;
         }
 
-        let new_schema = Arc::from(select_schema(&index)?);
+        let new_schema = Arc::from(enforce_otap_dict_keys(&select_schema(&index)?));
         let mut batcher = arrow::compute::BatchCoalescer::new(new_schema.clone(), index.row_count);
         for payload in select_all_mut(items, i) {
             let Some(rb) = payload.take() else {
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
```diff
@@ -573,6 +573,67 @@ fn select_dictionary_type<'a>(info: &FieldInfo<'a>) -> Result<DataType> {
...
+    matches!(
+        value_type,
+        DataType::Utf8
+            | DataType::LargeUtf8
+            | DataType::Utf8View
+            | DataType::Binary
+            | DataType::LargeBinary
+            | DataType::BinaryView
+    )
+}
```
albertlockett: The logic here is not correct. 

The issue in #2204 is not that all strings/binary columns should be should be dict encoded, it's that some columns, including, for example, one that is called `str`, should have the minimum key size as u16.

These are spelled out in the nascent spec https://github.com/open-telemetry/otel-arrow/pull/2040 . As far as I know, and feel free to correct me on this, the dict key size to use is not spelled out in the OTEP mentioned in the PR description?

gyanranjanpanda:
@albertlockett  could u review this now

JakeDern:
@gyanranjanpanda Thank you for your interest, but I think the approach here is a band-aid solution because:

1. We should be taking the OTAP spec into account when selecting the schema initially rather than fixing it afterwards
2. To properly account for the spec, we need to be applying rules based on the payload type and not just the names of the fields

Otap spec awareness is a more general problem for the repo that a good fix to this issue will make progress in addressing. The solution I was thinking about pursuing is to create `const` payload definitions that contain spec information and can be queried by any component. Something like:

```rust

pub const fn get_definition(typ: ArrowPayloadType) -> &PayloadDefinition {
    match typ {
        ArrowPayloadType::LOG_ATTRS => Attributes16Definition,
        ...
    }
}

const Attributes16Definition: &PayloadDefinition = &PayloadDefinition { ... };
const Attributes32Definition: &PayloaDefinition = &PayloadDefinition { ... };
...

struct PayloadDefinition {
    columns: Vec<ColumnDefinition>
    ...
}

impl PayloadDefinition {
    pub fn get(column_name: &str) -> Option<ColumnDefinition> { ... }
}

struct ColumnDefinition {
  min_dict_key_size: Option<DictionaryKeySize>
  ...
}
```

This would be complementary to and can eventually be integrated with the proposal in #2022 and provides a generic way to query or iterate over columns for a payload type. Note: There are many things not accounted for in the code snippet above such as describing relationships between payload types.

I should have put more context in the issue; that's my mistake so sorry for that. It was late at night when I filed it 🙂.

That all being said, I don't know that this is a good issue for you to take. IMO it's a somewhat urgent thing to fix and requires more experience with OTAP and the existing code base.

gyanranjanpanda:
Thanks for the clarification  that makes sense. I agree that fixing this at the schema/spec level with payload-type awareness is the better approach than a post-process workaround.
I'm happy to close this PR if you'd prefer to handle it, or I can try implementing the PayloadDefinition-based approach if you're open to it. Let me know what you'd prefer.

gyanranjanpanda:
https://github.com/open-telemetry/otel-arrow/issues/2207 could i do this issue that u open

JakeDern:
@gyanranjanpanda Yeah let's maybe close this one for now. I don't know if anyone is looking at #2207, but feel free to attempt something. You'll want to understand how the otap exporter component and our internal telemetry system works at a high level first so that you understand how to surface the appropriate logs and metrics.

JakeDern:
Opened #2240 with the approach described above

albertlockett:
@gyanranjanpanda I'm going to close this as the issue was resolved in #2240

PR #2219, opened 2026-03-06
fix(quiver): mmap segment files as shared

### LINKED ISSUE #2218: Quiver segment reader uses MAP_PRIVATE mmap, causing RSS to grow proportionally to disk usage during outages
When a downstream exporter is offline, the durable buffer accumulates data in Quiver segment files on disk as designed. However, `SegmentReader::open_mmap()` uses `map_copy_read_only()` (MAP_PRIVATE) for memory mapping. MAP_PRIVATE pages, once faulted in, become pinned in the process's RSS and cannot be reclaimed by the kernel without swapping—unlike MAP_SHARED pages which are backed by the file and freely evictable. The full-file CRC validation during `from_buffer()` faults in every page of every segment at open time (including during `scan_existing()` at restart). Combined with the retry loop in the durable_buffer processor continuously reading bundles from accumulated segments, this causes RSS to track disk usage ~1:1.

In testing, 2GB of buffered segment data on disk results in ~2.1GB RSS. At the default 10GB disk budget, this would consume ~10GB of RAM, defeating the purpose of buffering to disk.

I am publishing a PR to switch to `MmapOptions::map()` (MAP_SHARED) and calling `madvise(MADV_DONTNEED)` to immediately release pages from RSS. With this fix applied, RSS stays flat regardless of disk usage.

Originally introduced in #1649.

AaronRM:
# Change Summary

Switch Quiver `SegmentReader::open_mmap()` from MAP_PRIVATE (`map_copy_read_only`) to MAP_SHARED (`map`), and add `madvise(MADV_DONTNEED)` after CRC validation to release pages from RSS.

MAP_PRIVATE pages, once faulted in by the full-file CRC check, become pinned in RSS and cannot be reclaimed by the kernel. During downstream outages, as segments accumulate on disk, RSS grows ~1:1 with disk usage (e.g., 10GB disk budget -> ~10GB RSS), defeating the purpose of buffering to disk.

With MAP_SHARED, pages are backed by the file and freely reclaimable by the kernel. The `madvise(MADV_DONTNEED)` call after CRC proactively drops all pages from RSS immediately after validation. Subscribers re-fault only the specific pages they need on demand.

**Before:** 2GB on disk -> 2.1GB RSS, 3GB on disk -> 3.2GB RSS (linear growth)
**After:** 2GB on disk -> 130MB RSS, 3GB on disk -> 133MB RSS (flat)

* Closes #2218

- Existing unit tests pass (covering mmap correctness, zero-copy verification, CRC mismatch detection, bundle-outlives-reader, and multi-stream alignment)
- Validated by running `df_engine` configured with a durable_buffer processor against an error exporter (simulated outage).

No configuration changes. Users running `df_engine` or any Quiver-based pipeline with the durable buffer processor will see significantly reduced memory usage during downstream outages and recovery periods.


AaronRM:
FYI @lalitb and @albertlockett [you were previous reviewers on #1649].

lalitb (approved):
LGTM. Thanks,

Issue #919, opened 2025-08-11
Rate & Memory limiter framework


jmacd:
Part of https://github.com/open-telemetry/opentelemetry-collector/issues/9591
Part of https://github.com/open-telemetry/opentelemetry-collector/issues/12603

Covering a series of draft Collector changes, presently the most recent: 
https://github.com/open-telemetry/opentelemetry-collector/pull/13265

In #otel-arrow-dev a conversation on this topic specific to Rust/Arrow, selected fragments:

@jmacd wrote:

> Being familiar with Go and gRPC-Go, I've thought about how to manage the contextual information that goes with requests in the Go collector. This includes state [client.Info](http://client.info/) covering key:values from the arriving  gRPC or HTTP headers, from the Auth extension, timeout and cancellation, and e.g., tracing state (SpanContext).

> Looking to @lquerel especially for design thoughts on. How does the rate limiter processor briefly pause a request, if the token bucket says to wait for an interval? If a rate limiter forwards a request (with or without a delay) does it need to track active requests to match downstream Nack controls to upstream Nack controls? Will every component that maps incoming message IDs to outgoing message IDs ...

> After studying the code, my first guess would be to extend the Message::PData enum like

```
    /// A pipeline data message traversing the pipeline with context
    PData{ data: PData, ctx: Context },
```

The question is how to manage metadata used for limiter requests as part of the Context. See

PR #2224, opened 2026-03-06
Add delay processor for simulating pipeline latency

cijothomas:
Adds a new delay processor (urn:otel:processor:delay) that sleeps for a configurable duration before forwarding each message. This is useful for simulating slow pipeline stages when testing backpressure, inflight limits, and timeout behavior.

This is a standalone processor and hence it can be placed anywhere in a pipeline and/or paired with any exporter or other processors.

Note: This is an intentionally minimal component to help flush out engine issues now. A future rate limiter (#919) may provide more sophisticated flow control. Once that lands, this processor may be removed or kept as a simpler alternative for testing.

Open question: Should this be a processor (as this PR has), a dedicated sleeping exporter (e.g., [urn:otel:exporter:sleep], or a config option on the existing noop exporter? A processor is more composable — it can be placed anywhere in the pipeline and combined with any exporter — but a sleeping exporter would be simpler to configure for the common case of simulating a slow backend. Open to suggestions.


drewrelmas:
> Open question: Should this be a processor (as this PR has), a dedicated sleeping exporter (e.g., [urn:otel:exporter:sleep], or a config option on the existing noop exporter? A processor is more composable — it can be placed anywhere in the pipeline and combined with any exporter — but a sleeping exporter would be simpler to configure for the common case of simulating a slow backend. Open to suggestions.

It seems like there's an argument to be made for having both a slow processor (simulating an expensive transformation) and a slow exporter - they are slightly different scenarios.

Worth noting we have a test-only [flaky_exporter](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/otap/tests/common/flaky_exporter.rs) which does something similar.

lquerel:
Adding @c1ly to this PR for visibility as this processor could be used by its validation framework.

cijothomas:
> It seems like there's an argument to be made for having both a slow processor (simulating an expensive transformation) and a slow exporter

+1.
I'll follow up with a "slow exporter" too eventually. I am trying to test engine's stability/behavior in unfavorable conditions like a slow processor/exporter etc.

PR #2223, opened 2026-03-06
NOT for merge: Silent data loss

cijothomas:

PR #2225, opened 2026-03-06
fix(fanout): replace nack-on-full with accept_pdata() backpressure to prevent silent data loss

lalitb:
Issue reported in PR: #2223 

 ## Problem                                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                     
  When fanout hits its `max_inflight` limit it nacks incoming messages back upstream instead of applying backpressure. This causes silent data loss.                                                                                                                                 
                                                                                                                                                                                                                                                                                     
  The correct behavior is for fanout to stop consuming from its input channel when `max_inflight` is full. This causes the pdata channel to fill up, which blocks the upstream `send_message().await`, naturally slowing the receiver -- no data lost, no retry needed.

  Instead, fanout today does this:

  ```rust
  if self.slim_inflight.len() >= self.config.max_inflight {
      // nacks the message -- data is lost as receive is not supposed to retry.
      effect_handler.notify_nack(...).await?;
  }
  ```

  Receivers are ingress nodes and are not expected to handle nacks from downstream processors. There is no retry contract in the engine. The nack carries the original pdata in `nack.refused` but it goes out of scope and is dropped silently -- no log, no metric, no recovery.

  The workaround is `max_inflight: 0` (unlimited inflight), which prevents nacking and lets channel capacity provide natural backpressure, but removes the memory bound on in-flight state.

  ## Background: Stateless vs. Stateful Processors

  The engine's processor model was originally designed around stateless transforms: each call to `process()` is independent, and the engine feeds messages as fast as the channel delivers them.                                                                                     
   
  Fanout and batch are **stateful processors** -- they maintain in-flight tracking across multiple `process()` calls and must wait for acks before they can accept more work. This creates a throttling requirement that the engine's push model does not support:                   
                                                            
  - The processor knows when it is at capacity (inflight map full)
  - But it cannot signal the engine to pause -- the engine keeps calling `process()` regardless
  - The only escape valve was to nack the message, which causes data loss when receivers do not retry

  The fix closes this gap by letting processors signal their readiness back to the engine via `accept_pdata()`.

## Batch Processor

  Batch has similar stateful limits (`inbound_request_limit`, `outbound_request_limit`) but does not have the same silent data loss problem in practice. When batch nacks due to capacity, it expects a **retry processor** to be present upstream to catch and re-deliver the nacked message. This is the standard recommended pipeline topology:
                                                                                                                                                                                                                                                                                     
  receiver -> retry -> batch -> exporter                    

  Without a retry processor upstream, batch nacks would also cause silent data loss -- the same root cause. The `accept_pdata()` mechanism introduced here is the correct long-term fix for batch as well, but that is left as a follow-up since the immediate data loss issue only manifests in fanout (receivers do not retry; retry processors do).

## Fix  

  Introduces `accept_pdata()` on the `Processor` trait and wires it into the engine run loop.                                                                                                                                                                                        
   
  **`Processor` trait (`local` and `shared`)** -- new default method:                                                                                                                                                                                                                
                                                            
 ```rust
  fn accept_pdata(&self) -> bool { true }
```
  All existing stateless processors inherit true with no change. Only stateful processors that need throttling override it.

  Engine run loop -- changed from unconditional recv() to:

```rust
  while let Ok(msg) = message_channel.recv_when(processor.accept_pdata()).await {
      processor.process(msg, &mut effect_handler).await?;
  }
```

  `recv_when(false)` only reads from the control channel (acks/nacks), leaving `pdata` untouched in its channel. This lets acks drain the inflight map without deadlocking, and causes natural backpressure upstream once the pdata channel fills.

  **`FanoutProcessor`** -- implements accept_pdata():

```rust
  fn accept_pdata(&self) -> bool {
      if self.config.max_inflight == 0 { return true; }
      if self.config.use_fire_and_forget { return true; }
      if self.config.use_slim_primary {
          self.slim_inflight.len() < self.config.max_inflight
      } else {
          self.inflight.len() < self.config.max_inflight
      }
  }
```

  The nack-on-full blocks are removed from `process_slim_primary()` and the full path. No `pdata` is ever dropped due to `max_inflight` pressure.


## Coverage across all fanout modes                                                                                                                                                                                                                                                                           
                                         
  `accept_pdata()` handles all three paths correctly:                                                                                                                                                                                                                                                           
                                                            
  | Mode | `accept_pdata()` behavior | Was nacking before? |                                                                                                                                                                                                                                                    
  |------|--------------------------|---------------------| 
  | Fire-and-forget | Always `true` -- no inflight tracking, no throttling needed | No |
  | Slim primary | `slim_inflight.len() < max_inflight` | Yes -- fixed |
  | Full path (sequential / await_all / fallback / timeout) | `inflight.len() < max_inflight` | Yes -- fixed |

  Fire-and-forget is unaffected by design: it sends and acks upstream immediately, so inflight never accumulates.

  ## Performance

  **Hot path (`accept_pdata()` = `true`, the common case):**
  - `accept_pdata()` is a handful of integer comparisons -- effectively free
  - `recv_when(true)` is identical to `recv()` -- polls both channels, no added overhead
  - Zero latency impact on the steady-state hot path

  **When throttling kicks in (`accept_pdata()` = `false`):**
  - Engine loop only reads from `control_rx` -- proper async await, no busy loop, no CPU spin
  - Acks drain `inflight`, then `accept_pdata()` returns `true` again on the next iteration
  - Control messages (acks, nacks, shutdown, timeouts) are still delivered even while pdata is paused -- shutdown and timeout handling are unaffected

  **Heap:**
  - No new allocations. `inflight` and `slim_inflight` are still bounded by `max_inflight` exactly as before
  - The nack path removed actually slightly reduces code path length

  **One behavioral change worth noting:**
  Previously when `max_inflight` was hit, fanout nacked immediately and continued processing new messages (at the cost of data loss). Now, if the downstream exporter is permanently stuck and never acks, the pipeline stalls end-to-end rather than silently dropping. This is the correct behavior -- a stuck exporter should surface as backpressure, not silent loss -- but it is a visible behavioral change if `max_inflight` is configured and the downstream is unhealthy.

### DIFF
--- a/.github/workflows/project_management_comment.yml
+++ b/.github/workflows/project_management_comment.yml
@@ -1,0 +1,27 @@
+name: Add comment
+on:
+  issues:
+    types:
+      - labeled
+permissions:
+  contents: read
+
+jobs:
+  add-comment:
+    if: github.event.label.name == 'help wanted'
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@58077d3c7e43986b6b15fba718e8ea69e387dfcc # v2.15.1
+        with:
+          egress-policy: audit
+
+      - name: Add comment
+        uses: peter-evans/create-or-update-comment@e8674b075228eee787fea43ef493e45ece1004c9
+        with:
+          issue-number: ${{ github.event.issue.number }}
+          body: |
+            This issue is available for anyone to work on. Feel free to leave a comment if you'd like to be assigned. **Make sure to reference this issue in your pull request.**
+            :sparkles: Thank you for your contribution! :sparkles:
--- a/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
+++ b/rust/otap-dataflow/crates/otap/src/syslog_cef_receiver.rs
@@ -1646,9 +1646,20 @@ mod telemetry_tests {
             let listening_port = portpicker::pick_unused_port().expect("No free ports");
             let listening_addr: SocketAddr = format!("127.0.0.1:{listening_port}").parse().unwrap();
 
-            // Receiver with metrics enabled via pipeline
-            let receiver =
-                SyslogCefReceiver::with_pipeline(pipeline, Config::new_udp(listening_addr));
+            // Receiver with metrics enabled via pipeline.
+            // Use max_batch_size=1 so that the single record is flushed
+            // immediately in the recv_from handler instead of waiting for
+            // the interval tick, which avoids timing-dependent flakiness.
+            let receiver = SyslogCefReceiver::with_pipeline(
+                pipeline,
+                Config {
+                    protocol: Protocol::Udp(UdpConfig { listening_addr }),
+                    batch: Some(BatchConfig {
+                        flush_timeout_ms: None,
+                        max_size: NonZeroU16::new(1),
+                    }),
+                },
+            );
 
             // Keep downstream open to avoid refused
             let (out_tx, mut _out_rx) = otap_df_channel::mpsc::Channel::new(8);
@@ -1693,7 +1704,7 @@ mod telemetry_tests {
             // To exercise the "invalid" path, send an empty datagram which is rejected by the parser.
             let _ = sock.send_to(b"", listening_addr).await.unwrap();
 
-            // Allow interval to tick
+            // Allow the receiver task to process the incoming messages.
             tokio::time::sleep(Duration::from_millis(150)).await;
 
             // Trigger telemetry collection
@@ -1738,8 +1749,22 @@ mod telemetry_tests {
             let port = portpicker::pick_unused_port().expect("No free ports");
             let addr: SocketAddr = format!("127.0.0.1:{port}").parse().unwrap();
 
-            // Receiver with pipeline metrics
-            let receiver = SyslogCefReceiver::with_pipeline(pipeline, Config::new_udp(addr));
+            // Receiver with pipeline metrics.
+            // Use max_batch_size=1 so that the single record is flushed
+            // immediately in the recv_from handler instead of waiting for
+            // the interval tick, which avoids timing-dependent flakiness.
+            let receiver = SyslogCefReceiver::with_pipeline(
+                pipeline,
+                Config {
+                    protocol: Protocol::Udp(UdpConfig {
+                        listening_addr: addr,
+                    }),
+                    batch: Some(BatchConfig {
+                        flush_timeout_ms: None,
+                        max_size: NonZeroU16::new(1),
+                    }),
+                },
+            );
 
             // Wire a closed downstream to force refused
             let (tx, rx) = otap_df_channel::mpsc::Channel::new(1);


lalitb:
This is ready for review now.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -1091,23 +1094,8 @@ impl Processor<OtapPdata> for FanoutProcessor {
 
                 // === FULL PATH: Sequential, await_all, fallback, or timeout ===
                 // Full inflight tracking with DestinationVec.
-
-                // Check max_inflight limit before accepting new message.
-                if self.config.max_inflight > 0 && self.inflight.len() >= self.config.max_inflight {
-                    self.metrics.rejected_max_inflight.add(1);
```
utpilla, line 1097: This metric `rejected_max_inflight` can now be removed.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -1030,6 +1017,22 @@ impl FanoutProcessor {
 
 #[async_trait(?Send)]
 impl Processor<OtapPdata> for FanoutProcessor {
+    /// Returns `false` when the inflight map is at capacity, pausing pdata delivery until acks
+    /// drain the inflight map below the limit. This provides backpressure without data loss.
+    fn accept_pdata(&self) -> bool {
+        if self.config.max_inflight == 0 {
```
utpilla, line 1054: Not introduced in this PR but using `0` to represent unlimited looks odd. Using an `Option<usize>` would be much better. `None` could represent unlimited and `Some` would represent the finite values. I think we should fix this (in a different PR).

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -2823,4 +2792,83 @@ mod tests {
         // Request should be complete
         assert!(h.fanout.inflight.is_empty(), "request should be complete");
     }
+
+    /// Reproduces silent data loss when max_inflight is exceeded.
+    ///
+    /// # Bug
+    ///
+    /// When fanout hits its `max_inflight` limit it nacks the incoming message
+    /// upstream.  The expectation is that the upstream node retries the data,
```
utpilla: Why should that be the expectation? The upstream node such as a receiver could also just apply backpressure and let the client decide what they want to do.

utpilla:
> You mention that `batch` processor isn't affected in same way because there is an expectation for a `retry` processor ahead of it.
> 
> Could the same pattern not be followed between a receiver and `fanout` processor rather than changing the core engine loop?
> 
> > receiver -> retry -> fanout

A retry processor upstream doesn't fix this. It masks it differently:
- **Without retry:** NACK → receiver ignores → data lost.
- **With retry:** NACK → retry catches and re-queues, but fanout keeps accepting and rejecting at full speed, so retry accumulates unbounded delayed data → OOM under sustained pressure.

Both fail because **NACKing doesn't create backpressure**, `process()` returns immediately and the engine delivers the next message right away. `accept_pdata(false)` stops delivery at the source.

lalitb:
> > You mention that `batch` processor isn't affected in same way because there is an expectation for a `retry` processor ahead of it.
> > Could the same pattern not be followed between a receiver and `fanout` processor rather than changing the core engine loop?
> > > receiver -> retry -> fanout
> 
> A retry processor upstream doesn't fix this. It masks it differently:
> 
> * **Without retry:** NACK → receiver ignores → data lost.
> * **With retry:** NACK → retry catches and re-queues, but fanout keeps accepting and rejecting at full speed, so retry accumulates unbounded delayed data → OOM under sustained pressure.
> 
> Both fail because **NACKing doesn't create backpressure**, `process()` returns immediately and the engine delivers the next message right away. `accept_pdata(false)` stops delivery at the source.

Agreed with the above. Just want to add a couple more points:

Even if retry doesn't OOM, it eventually gives up - once `max_elapsed_time` (5 min default) is reached, it nacks upstream and the receiver drops it. So retry just delays the loss under sustained pressure rather than preventing it.

The other motivation for fixing this at the engine level: batch has the same class of problem. Today it crashes with `ProcessorError` when slots run out, and there's already a PR to replace that with a nack. If we go the retry-in-front route, we'd need retry before every stateful processor. `accept_pdata()` handles it once - and the `recv_when` guard already exists in the engine, we're just exposing it to processors via a trait method with a safe default.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -2823,4 +2792,83 @@ mod tests {
...
+        // All sent messages must reach the exporter -- zero silent drops.
+        assert_eq!(
+            nacked, 0,
+            "BUG: {nacked} message(s) were nacked and silently lost -- \
+             fanout should apply backpressure instead of nacking"
+        );
+        assert_eq!(
+            delivered, TOTAL_SENT,
+            "all {TOTAL_SENT} messages must reach the exporter, got {delivered}"
+        );
```
utpilla: The name of this test is misleading. The test only checks that no NACKs were produced by the fanout processor. Also, `delivered` matching `TOTAL_SENT` is an odd assertion to have. The whole idea is to apply backpressure and not allow the sender to send more data. Here we are sending `4` items, when the backpressure should have restricted that to `2`.

I understand that this testing weirdness arises mainly because we are not testing this using `recv_when` but maybe we could rename the test to no NACKs produced or something like that.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -2823,4 +2792,83 @@ mod tests {
...
+    /// Reproduces silent data loss when max_inflight is exceeded.
+    ///
+    /// # Bug
+    ///
+    /// When fanout hits its `max_inflight` limit it nacks the incoming message
+    /// upstream.  The expectation is that the upstream node retries the data,
+    /// but no receiver implements retry for nacks.  The traffic generator
+    /// (and all real receivers) handle `NodeControlMsg::Nack` with a catch-all
+    /// `_ => {}` arm -- the data is simply discarded with no log, no metric
+    /// increment, and no retry.
```
utpilla: The doc comments on this still describes the old behavior ("fanout nacks them upstream", "`assert!(nacked > 0)` currently passes, confirming the bug"). Now that the fix is in, this reads like a pre-fix snapshot that was left behind. Let's update that to describe what the test actually verifies today, that `process()` no longer nacks when inflight is full.

Also worth noting in the comment: this test calls `process()` directly, so it doesn't exercise `accept_pdata()` or the engine's `recv_when` gate. It confirms the nack-on-full code was removed, but not that backpressure works end-to-end.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
utpilla, line 1119: It'd be good to add a `debug_assert` here:

```rust
debug_assert!(
    self.accept_pdata(),
    "process() called with PData while at max_inflight ({}) — \
        engine must check accept_pdata() before delivering",
    self.config.max_inflight
);
```

utpilla (approved):
The fix looks good. Left some suggestions.

rust/otap-dataflow/crates/engine/src/processor.rs
```diff
@@ -423,7 +423,7 @@ impl<PData> ProcessorWrapper<PData> {
                     .start_periodic_telemetry(Duration::from_secs(1))
                     .await?;
 
-                while let Ok(msg) = message_channel.recv().await {
+                while let Ok(msg) = message_channel.recv_when(processor.accept_pdata()).await {
```
ThomsonTan, line 426: There could be a liveness issue here. `recv_when(false)` only detects pdata channel closure when the pdata queue is already empty. If upstream closes while buffered pdata is still queued, and `accept_pdata()` is false, the loop in `recv_when()"' won't read pdata or synthesize shutdown. It will wait only on control, which may never arrive.

lalitb: The liveness concern doesn't apply here. When `accept_pdata()` returns false, it's because `inflight.len() >= max_inflight` - meaning there are in-flight messages downstream that will generate ack/nack control messages. Those control messages unblock `recv_when(false)`, reduce inflight, and the next iteration calls `recv_when(true)` which drains the buffered pdata. The closure is detected once the buffer empties. There's also timeout handling for in-flight entries, so a hung downstream can't stall this indefinitely. The test `test_recv_when_false_closed_with_buffered_data` explicitly validates this drain-then-detect sequence.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Good point, have rephrased the text better.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Good point. Will take it in separate PR.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
JakeDern: I have the same question. I'm not super familiar with the engine yet but just trying to wrap my head around this so I can look at the batch processor behavior. 

My understanding is that we subscribe to nacks in the otap_grpc implementation:

https://github.com/open-telemetry/otel-arrow/blob/449692896ee264d3ba00bde14a5f6fd93a94b77e/rust/otap-dataflow/crates/otap/src/otap_grpc.rs#L341-L346

It looks like then we try to wait for an ack/nack after calling `send_message_with_source_node`, but I don't really understand how this works and how the nack makes it into the `rx` channel there:

https://github.com/open-telemetry/otel-arrow/blob/449692896ee264d3ba00bde14a5f6fd93a94b77e/rust/otap-dataflow/crates/otap/src/otap_grpc.rs#L365-L398

But at the layer above in the otap receiver we just increment some counter and discard like is mentioned in the PR:

https://github.com/open-telemetry/otel-arrow/blob/449692896ee264d3ba00bde14a5f6fd93a94b77e/rust/otap-dataflow/crates/otap/src/otap_receiver.rs#L195-L202

So is the problem that the ack/nack handling in the otap_grpc.rs implementation isn't receiving the nack and streaming back the right response today?

JakeDern: I'm speaking about the otap receiver here, but I'm trying to look at your statement about receivers not being expected to handle nacks:

> Receivers are ingress nodes and are not expected to handle nacks from downstream processors

I guess my main questions are:

- Why aren't receivers and ingress nodes expected to handle nacks?
- Does the nack handling in the otap receiver not work? Or where was the silent dropping observed?

lquerel:
@lalitb @jmacd @albertlockett @drewrelmas @utpilla 

Please do not merge this PR before I have had a chance to review it. Sorry for the delay, I have too many things going on in parallel right now.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -1075,6 +1075,13 @@ impl Processor<OtapPdata> for FanoutProcessor {
             // mirroring other stateless processors. Follow-up could proactively nack inflight on shutdown.
             Message::Control(_) => Ok(()),
             Message::PData(pdata) => {
+                debug_assert!(
+                    self.accept_pdata(),
+                    "process() called with PData while at max_inflight ({}) — \
+                     engine must check accept_pdata() before delivering",
+                    self.config.max_inflight
+                );
```
lquerel, line 1117: The comment for the `recv_when` method says:

```
    /// During shutdown draining the guard is ignored — pdata is
    /// always drained until the deadline, regardless of
    /// `accept_pdata`.
```

So in debug mode, this `debug_assert!` can panic during shutdown, especially when fanout is already in a saturated state.

I’m wondering why, once shutdown draining starts, we couldn’t continue honoring `accept_pdata` for `pdata`, while still delivering cleanup control messages (Ack/Nack) so that stateful processors can reduce their in-flight state and reopen capacity.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
```diff
@@ -376,9 +376,6 @@ struct FanoutMetrics {
     pub nacked: Counter<u64>,
     #[metric(unit = "{item}")]
     pub timed_out: Counter<u64>,
-    /// Messages rejected due to max_inflight limit (backpressure).
-    #[metric(unit = "{item}")]
-    pub rejected_max_inflight: Counter<u64>,
```
lquerel, line 381: For operability, I think we should add few saturation gauge metrics:
- `in_flight`: current tracked request count.
- `max_inflight_config`: configured max_inflight value, with 0 meaning unlimited.
- `throttled`: 1 when fanout is currently refusing new pdata via `accept_pdata()`, else 0.
- `throttle_episodes` (a counter): increments only on transition from not-throttled to throttled.

Those metrics could be updated after state mutation (not inside accept_pdata to avoid hot path impact).

lquerel:
Just for future reviewers who might be confused by the PR description (as I was), I'd like to provide an alternative description of the problem:

* by the time `process()` is called, the `pdata` item has already been dequeued from `MessageChannel`
* acks/nacks are delivered later as control messages through the engine
* if fanout tries to “wait inside `process()`” for an ack/nack that would free capacity, the engine stops polling the message channel while that call is blocked
* that means the ack/nack it is waiting for cannot be delivered to fanout, because it remains queued on the control channel

So the issue is not that the engine may call `process()` concurrently. The issue is that the current engine API has no way to say “stop consuming `pdata`, but keep draining control messages” before the next `pdata` is removed from the channel.

In any case, the solution developed in this PR seems appropriate to me.

JakeDern:
Some thoughts on this related to batch processor - https://github.com/open-telemetry/otel-arrow/pull/2198#issuecomment-4040575796

@cijothomas @lalitb @lquerel Any thoughts on merging the ack/nack based approach for now vs adding a ref to the actual pdata in accept_pdata?

lquerel:
@JakeDern Just to make sure we're on the same page. What you're proposing is to merge the Nack approach for the batch processor and keep the `accept_pdata` approach for the current PR. If so, I agree.
In the medium term, however, I think we will need to revisit the batch processor on this aspect.

JakeDern:
> @JakeDern Just to make sure we're on the same page. What you're proposing is to merge the Nack approach for the batch processor and keep the `accept_pdata` approach for the current PR. If so, I agree. In the medium term, however, I think we will need to revisit the batch processor on this aspect.

Yes exactly! We're playing comment tag, I just replied on the other PR 😁

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Good catch. This is fixed now.

rust/otap-dataflow/crates/otap/src/fanout_processor.rs
lalitb: Done. The `rejected_max_inflight` counter has been replaced with the four metrics you suggested - All four are updated via `update_saturation_metrics()` which is called at the end of `process()` after any state mutation, not inside `accept_pdata()` itself to keep the hot path clean. The transition detection uses a `was_throttled` flag to ensure `throttle_episodes` only increments on the edge, not while staying saturated.

lalitb:
Thanks for the review. all the comments are resolved now.,

drewrelmas:
Not merging yet to give @jmacd opportunity to comment.

Issue #1423, opened 2025-11-12
Add otap-dataflow Changelog generation


jmacd:
In order to release on crates.io, we should incorporate a Changelog practice a we begin versioning.

See #737.

drewrelmas:
@jmacd given the rarity of investment in the `go` subdirectory outside of dependency upgrades, I'm wondering if you think it makes sense to stick the [current Changelog](https://github.com/open-telemetry/otel-arrow/blob/main/CHANGELOG.md) that represents only Golang changes under `go/CHANGELOG.md` and start a new `rust/CHANGELOG.md`?

The [release automation](https://github.com/open-telemetry/otel-arrow/blob/main/.github/workflows/prepare-release.yml) would need some small updates too as well if that were the case.

We should also reach a conclusion on the topic mentioned in #737 about versioning. Given we already have git tags on the repository up to `v0.45.0` - is the simplest course of action to start versioning the Rust crates at the same value?

drewrelmas:
Given the interest in publishing at least an OTAP dataflow engine binary from the repo in #3096, I would like to make a push on the Changelog topic. I propose the following solution:

- [x] Adopt the Collector/Collector-Contrib `chloggen` contributor-facing process almost exactly ([source link](https://github.com/open-telemetry/opentelemetry-go-build-tools/blob/main/chloggen/README.md))
  - Contributors accumulate individual changelog entries in `.chloggen/` subdirectories
  - Avoids merge conflicts caused by many contributors all trying to touch one Changelog file
  - Maintain 2 separate Changelogs in the repo, one at `go/CHANGELOG.md` and one at `rust/otap-dataflow/CHANGELOG.md`
  - Bypassed by dependency upgrades, `chore` in PR title, or specific PR label
- [x] Incorporate the `chloggen` process into the `Prepare Release` and `Push Release` workflows
- [ ] (optional) Enforce PR title standardization
  - Potentially incorporate https://github.com/amannn/action-semantic-pull-request or a more lightweight custom version

lquerel:
@drewrelmas Could we add a CI check to ensure that a changelog entry is created for every new PR?

cijothomas:
https://github.com/open-telemetry/opentelemetry-rust/pull/2698 sharing something we did in OTel Rust, might be handy.
Agree with @drewrelmas proposal. (Thanks for already calling out chore title skipping the validation)

PR #2226, opened 2026-03-06
Add GitHub issue templates

ThomsonTan:
# Change Summary

Based on the discussion in the last community meeting, add GitHub issue templates to improve issue intake and triage for this repository.


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,117 @@
+name: Bug Report
+description: Report a bug in OTel-Arrow
+labels: ["Bug"]
```
drewrelmas, line 4: Is there a way for us to set issue Type `Bug` here instead of just a label? I thought that was the main point from the SIG call - only maintainers and triagers may modify labels and issue type.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+  - type: textarea
+    id: actual
+    attributes:
+      label: Actual Behavior
+      description: What actually happened. Include error messages and logs if applicable.
+    validations:
+      required: true
+
+  - type: input
```
drewrelmas, line 69: Not sure we need version at this stage - we technically aren't versioning `otap-df` crates yet. See #1423.

Release version in this repo currently only refers to Go modules.

.github/ISSUE_TEMPLATE/feature_request.yaml
```diff
@@ -0,0 +1,68 @@
+name: Feature Request
+description: Suggest a new feature or enhancement for OTel-Arrow
+labels: ["Feature"]
```
drewrelmas, line 4: Similar here, would like to use Type, not a label.

.github/ISSUE_TEMPLATE/bug_report.yaml
drewrelmas: Looking at the documentation, I see there is:

```yaml
type: bug
```

which would probably do what we want.

drewrelmas:
In addition, do you think we should have a template for issue Type `Task`?

.github/ISSUE_TEMPLATE/bug_report.yaml
ThomsonTan: Added type field to all templates and aligned them with the exact type name currently in use: Bug.

.github/ISSUE_TEMPLATE/bug_report.yaml
ThomsonTan: Remove version and uses commit hash instead, this is also optional.

ThomsonTan:
> In addition, do you think we should have a template for issue Type `Task`?

I initially thought this overlapped with the feature request type, but since the repo is still evolving rapidly and we already have a substantial number of task-oriented items, I added a dedicated task template as well.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,118 @@
...
+
+  - type: dropdown
+    id: component
+    attributes:
+      label: Component(s)
+      description: Which component(s) does this bug affect?
+      multiple: true
+      options:
+        - Rust OTAP dataflow (rust/otap-dataflow/)
+        - Rust Beaubourg (rust/beaubourg/)
```
lquerel: Instead of Beaubourg, we could specify here the Go OTel Arrow Receiver/Expoter

.github/ISSUE_TEMPLATE/feature_request.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+  - type: dropdown
+    id: component
+    attributes:
+      label: Component(s)
+      description: Which component(s) does this feature relate to?
+      multiple: true
+      options:
+        - Rust OTAP dataflow (rust/otap-dataflow/)
+        - Rust Beaubourg (rust/beaubourg/)
```
lquerel: See my previous comment

.github/ISSUE_TEMPLATE/other.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+  - type: dropdown
+    id: component
+    attributes:
+      label: Component(s)
+      description: Which component(s) does this relate to?
+      multiple: true
+      options:
+        - Rust OTAP dataflow (rust/otap-dataflow/)
+        - Rust Beaubourg (rust/beaubourg/)
```
lquerel: Same

.github/ISSUE_TEMPLATE/bug_report.yaml
drewrelmas: Yes - good point. @ThomsonTan the Beaubourg project is just a reference implementation that is not receiving active development.

Better replacement is the Go code in the repo which we still maintain.

ThomsonTan: Thanks for the clarification. Removed the Beaubourg from the list because it is superseded by otap-dataflow and added "Go OTel Arrow Receiver/Exporter" as suggested.

PR #2230, opened 2026-03-07
docs: add extension system architecture document

gouslu:
# Change Summary

PR to discuss extension system architecture in relation to implementation in PR https://github.com/open-telemetry/otel-arrow/pull/2141

Yes, adds extension section to configuration in the proposes design.

### DIFF
--- a/docs/extension-system-architecture.md
+++ b/docs/extension-system-architecture.md
@@ -22,7 +22,7 @@ Before extensions, cross-cutting concerns like authentication
 were embedded directly inside individual exporters. This led to:
 
 - **Duplication** -- every exporter needing auth carried its own
-  credential management, token refresh loop, and retry logic.
+  credential management and token refresh loop.
 - **Tight coupling** -- credential-specific dependencies (e.g.,
   `azure_identity`) leaked into exporter crates even when unused.
 - **No sharing** -- multiple exporters targeting the same tenant
@@ -32,8 +32,9 @@ Extensions solve this by extracting shared capabilities into
 named, independently-running components. An extension can
 optionally expose well-defined traits through a type-safe
 registry for data-path nodes to look up by name, or it can
-simply run as a pure background task (e.g., periodic cleanup,
-health reporting) without publishing any traits at all. Either
+simply run as a pure background task (e.g., certificate
+rotation, service discovery refresh) without publishing
+any capabilities at all. Either
 way: no direct dependencies between nodes, no duplicated logic,
 no wasted resources.
 
@@ -48,12 +49,12 @@ no wasted resources.
 |  | (auth)      |  | (background)|             |
 |  | Arc<State>  |  +-------------+             |
 |  +------+------+                              |
-|         | extension_traits!() macro           |
+|         | extension_capabilities!() macro     |
 |         | clones self per trait, producing    |
-|         | Vec<TraitRegistration>              |
+|         | Vec<CapabilityRegistration>         |
 |         v                                     |
 |  +----------------------+                     |
-|  |  ExtensionRegistry   |  (built once)       |
+|  |  CapabilityRegistry  |   (built once)      |
 |  |  stores cloned trait |                     |
 |  |  objects by name     |                     |
 |  +--+---------------+---+                     |
@@ -68,7 +69,8 @@ no wasted resources.
 |                                               |
 |  get() returns a cloned Box<dyn Trait>;       |
 |  all clones share state via Arc inside the    |
-|  extension -- the registry itself is stateless|
+|  extension -- the registry itself holds       |
+|  type-erased cloneable trait objects          |
 +-----------------------------------------------+
 ```
 
@@ -79,12 +81,14 @@ no wasted resources.
    capabilities are available at initialization. At shutdown,
    extensions terminate only after all data-path nodes have
    drained -- ensuring capabilities like auth tokens remain
-   available during final flushes.
+   available during final flushes. Extension instances are
+   scoped to a single pipeline -- they are not shared across
+   pipelines.
 
 2. **PData-free.** Extensions are completely decoupled from
    the pipeline data type (`PData`). They receive their own
-   `ExtensionControlMsg` messages (shutdown, timer ticks,
-   config updates, telemetry collection) through a dedicated
+   `ExtensionControlMsg` messages (shutdown, config
+   updates, telemetry collection) through a dedicated
    control channel and never process pipeline data directly.
 
 3. **Separate control channel.** Extensions use
@@ -101,22 +105,23 @@ no wasted resources.
    spawned on multi-threaded runtimes. `ExtensionWrapper`
    abstracts over both variants.
 
-5. **Registry-based lookup.** Receivers and exporters receive
-   an `ExtensionRegistry` at `start()` and look up extensions
-   by name and trait. Processors don't currently receive the
-   registry, but adding it is trivial -- the same
-   `extension_registry.clone()` pattern applies directly.
-   If you have a use case for this, please comment.
+5. **Registry-based lookup.** The `CapabilityRegistry` is
+   passed to receiver, processor, and exporter factories
+   at construction time -- not at `start()`. This means
+   capabilities are resolved during pipeline build, catching
+   missing extensions early. The API and naming are
+   node-agnostic -- all node types receive the registry
+   through the same factory parameter.
 
-6. **Optional trait publishing.** Extensions that expose
-   capabilities override `extension_traits()` to register
-   trait implementations in the registry. Extensions that are
+6. **Optional capability publishing.** Extensions that expose
+   capabilities override `extension_capabilities()` to register
+   capability implementations in the registry. Extensions that are
    pure background tasks simply use the default (empty)
    implementation and never appear in the registry.
 
 ## Core Types
 
-### Extension Trait
+### Extension Lifecycle Trait
 
 The lifecycle contract every extension implements. Two
 variants exist -- local and shared -- mirroring the pattern
@@ -133,8 +138,8 @@ pub trait Extension {
         effect_handler: EffectHandler,
     ) -> Result<TerminalState, Error>;
 
-    fn extension_traits(&self)
-        -> Vec<TraitRegistration>
+    fn extension_capabilities(&self)
+        -> Vec<CapabilityRegistration>
     {
         Vec::new()
     }
@@ -152,8 +157,8 @@ pub trait Extension: Send {
         effect_handler: EffectHandler,
     ) -> Result<TerminalState, Error>;
 
-    fn extension_traits(&self)
-        -> Vec<TraitRegistration>
+    fn extension_capabilities(&self)
+        -> Vec<CapabilityRegistration>
     {
         Vec::new()
     }
@@ -169,19 +174,19 @@ Key points:
   extension into its own task. After this, the engine can
   only reach it through the control channel.
 - **`ControlChannel`** wraps a receiver for
-  `ExtensionControlMsg` (shutdown, timer ticks, config
+  `ExtensionControlMsg` (shutdown, config
   updates, telemetry collection). No pipeline data ever
   flows through it.
 - **`EffectHandler`** provides node identity and metrics
   reporting. Extensions manage their own timers directly
   (e.g., `tokio::time`) rather than through the engine's
   timer infrastructure.
-- **`extension_traits()`** defaults to empty. Extensions
-  that publish traits override it (typically via the
-  `extension_traits!` macro) to return a
-  `Vec<TraitRegistration>`. During pipeline build, the
+- **`extension_capabilities()`** defaults to empty. Extensions
+  that publish capabilities override it (typically via the
+  `extension_capabilities!` macro) to return a
+  `Vec<CapabilityRegistration>`. During pipeline build, the
   engine calls this method on each extension and inserts
-  the returned registrations into the `ExtensionRegistry`
+  the returned registrations into the `CapabilityRegistry`
   under the extension's configured name. Pure background
   tasks leave the default and never appear in the
   registry.
@@ -218,9 +223,9 @@ Responsibilities:
 - **Construction** -- `ExtensionWrapper::local()` and
   `::shared()` create the control channel and box the
   extension.
-- **Trait registration** -- `register_traits()` calls
-  the extension's `extension_traits()` and inserts the
-  results into the `ExtensionRegistry` under the
+- **Trait registration** -- `register_capabilities()` calls
+  the extension's `extension_capabilities()` and inserts the
+  results into the `CapabilityRegistry` under the
   extension's name.
 - **Control sender** -- `extension_control_sender()`
   produces an `ExtensionControlSender` that the engine
@@ -246,7 +251,6 @@ so they have no `Ack`, `Nack`, or `DelayedData` variants.
 #[derive(Debug, Clone)]
 pub enum ExtensionControlMsg {
     Config { config: serde_json::Value },
-    TimerTick {},
     CollectTelemetry {
         metrics_reporter: MetricsReporter,
     },
@@ -258,11 +262,15 @@ Each variant:
 
 - **`Config`** -- notifies the extension of a configuration
   change (hot reload).
-- **`TimerTick`** -- periodic tick from the engine.
 - **`CollectTelemetry`** -- asks the extension to flush its
   local metrics into the provided `MetricsReporter`.
 - **`Shutdown`** -- requests graceful shutdown with a
   deadline and human-readable reason.
+
+Extensions manage their own periodic timers directly via
+`tokio::time` rather than receiving timer ticks from
+the engine. This keeps the control channel lean and
+avoids unnecessary wakeups.
 
 These messages flow through a dedicated channel per
 extension (created by `ExtensionWrapper`), kept separate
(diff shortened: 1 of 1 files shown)


docs/extension-system-architecture.md
```diff
@@ -0,0 +1,794 @@
...
+`nodes`, not as nodes themselves, and they never touch pipeline
+data directly.
+
+### Why Do We Need Them?
+
+Before extensions, cross-cutting concerns like authentication
+were embedded directly inside individual exporters. This led to:
+
+- **Duplication** -- every exporter needing auth carried its own
+  credential management, token refresh loop, and retry logic.
```
lquerel: Retry logic is not part of the exporters, we already have a retry processor.

docs/extension-system-architecture.md
```diff
@@ -0,0 +1,794 @@
...
+  `azure_identity`) leaked into exporter crates even when unused.
+- **No sharing** -- multiple exporters targeting the same tenant
+  each acquired and refreshed tokens independently.
+
+Extensions solve this by extracting shared capabilities into
+named, independently-running components. An extension can
+optionally expose well-defined traits through a type-safe
+registry for data-path nodes to look up by name, or it can
+simply run as a pure background task (e.g., periodic cleanup,
+health reporting) without publishing any traits at all. Either
```
lquerel: Not sure I understand what we are referring to here. Health reporting is already in place and is a core function, not really an extension-based system. Similarly, I'm not sure what is meant by periodic cleanup.

docs/extension-system-architecture.md
```diff
@@ -0,0 +1,794 @@
...
+system in the OTAP dataflow engine -- what extensions are,
+how they integrate into the pipeline lifecycle, and how to
+implement new ones.
+
+## What Are Extensions?
+
+Extensions are standalone pipeline components that provide
+**shared, cross-cutting capabilities** -- such as authentication,
+service discovery, or health checking -- to data-path nodes
+(receivers, exporters). They are configured as siblings to
```
lquerel, line 15: Why limiting extensions to receivers and exporters?

docs/extension-system-architecture.md
```diff
@@ -0,0 +1,794 @@
...
+|  +----------+  +----------+                   |
+|  | Receiver |  | Exporter |                   |
+|  | (own     |  | (own     |                   |
+|  |  registry|  |  registry|                   |
+|  |  clone)  |  |  clone)  |                   |
+|  +----------+  +----------+                   |
+|                                               |
+|  get() returns a cloned Box<dyn Trait>;       |
+|  all clones share state via Arc inside the    |
+|  extension -- the registry itself is stateless|
```
lquerel: Saying that the registry is stateless while it contains trait objects seems strange to me.

docs/extension-system-architecture.md
```diff
@@ -0,0 +1,794 @@
...
+
+4. **Local/Shared split.** Like receivers and exporters,
+   extensions have both local (`!Send` futures) and shared
+   (`Send` futures) variants. Local extensions run on the
+   single-threaded `LocalSet`; shared extensions can be
+   spawned on multi-threaded runtimes. `ExtensionWrapper`
+   abstracts over both variants.
+
+5. **Registry-based lookup.** Receivers and exporters receive
+   an `ExtensionRegistry` at `start()` and look up extensions
```
lquerel: Is the `start` method the best place to pass the `ExtensionRegistry`?

docs/extension-system-architecture.md
```diff
@@ -0,0 +1,794 @@
...
+|  |  clone)  |  |  clone)  |                   |
+|  +----------+  +----------+                   |
+|                                               |
+|  get() returns a cloned Box<dyn Trait>;       |
+|  all clones share state via Arc inside the    |
+|  extension -- the registry itself is stateless|
++-----------------------------------------------+
+```
+
+### Key Design Decisions
```
lquerel, line 77: In these design decisions, it is not explicitly stated that the lifecycle of extensions is directly tied to the lifecycle of pipeline instances. It is represented in the diagram, but I would like it to be clear to the reader that an extension instance is not shared between pipelines.

docs/extension-system-architecture.md
```diff
@@ -0,0 +1,794 @@
...
+  extension into its own task. After this, the engine can
+  only reach it through the control channel.
+- **`ControlChannel`** wraps a receiver for
+  `ExtensionControlMsg` (shutdown, timer ticks, config
+  updates, telemetry collection). No pipeline data ever
+  flows through it.
+- **`EffectHandler`** provides node identity and metrics
+  reporting. Extensions manage their own timers directly
+  (e.g., `tokio::time`) rather than through the engine's
+  timer infrastructure.
```
lquerel, line 183: Why do we have a `TimerTick` variant in the `ExtensionControlMsg` enum if extensions manage their own timers?

docs/extension-system-architecture.md
```diff
@@ -0,0 +1,794 @@
...
+correct `TypeId` and coerce functions -- a process that
+is both tedious and easy to get wrong.
+
+The macro's `Clone` requirement is intentional -- it
+signals to extension developers that their type will be
+cloned during registration (and again on each registry
+`get()` call). This encourages holding internal state
+behind `Arc` so that clones are cheap (just a reference
+count bump) and all clones observe the same underlying
+state.
```
utpilla, line 418: This seems problematic. While both approaches #2113 and #2141 agree on the fundamentals: extensions are PData-free, start first, shut down last, and expose capabilities via a registry, they differ in how these capabilities flow from extension to consumer.

In this PR (#2141), the extension struct is the capability. The `extension_traits!` macro clones self into the registry. Consumers get a clone of the extension. Correctness requires all mutable state to be Arc-wrapped so clones stay in sync.

In #2113, the factory builds two separate objects: the extension (lifecycle) and a handle (capability).
The extension doesn't need Clone. No macro needed.

> This encourages holding internal state behind `Arc`

The "**remember to Arc**" problem:

In the clone-self mode (this PR), the engine creates three independent copies of the extension struct during pipeline build:

1. The original created by the factory, lives inside `ExtensionWrapper`
2. The macro clone `extension_traits!()` calls `self.clone()` and stores the copy in the `ExtensionRegistry`
3. The consumer clone, `registry.get()` clones from #2 and hands it to the receiver/exporter

After registration, `start()` consumes #1 (the original). The exporter/receiver works with #3 (a clone of a clone). These are different objects and they only share state for fields wrapped in `Arc`. Any non-`Arc` mutable field silently diverges.

This compiles without warning:

```rust
#[derive(Clone)]
struct MyExtension {
    token_tx: Arc<watch::Sender<Token>>,  // ✅ shared across all clones
    retry_count: u32,                      // ❌ each clone gets its own copy
}
```

`start()` runs on copy #1 and increments `retry_count`. The exporter holds copy #3 and sees `retry_count` stuck at 0. The framework relies on a convention: "wrap everything mutable in Arc" which the compiler cannot enforce.

In #2113, this bug is structurally impossible. The handle and extension are different types constructed by the factory. There's nothing to diverge because they were never copies of the same object. No "remember to Arc everything" convention is needed. I think that's a better contract to offer the extension authors.

lquerel: I admit that detail escaped me, and it does seem error-prone. Separating extension and handle does seem like a good idea. 

@gouslu @utpilla  How does that integrate with the other idea of allowing multiple capabilities to be exposed by the same extension?

utpilla: > How does that integrate with the other idea of allowing multiple capabilities to be exposed by the same extension?

In #2113, `ExtensionHandles` supports registering multiple handles of different types from the same extension factory:

```rust
let mut handles = ExtensionHandles::new();
handles.register(ClientAuthenticatorHandle::new(auth));
handles.register(HealthCheckHandle::new(health));
// register as many distinct handle types as needed
```

Each [`register()` call adds a `(TypeId, ErasedHandle)` entry](https://github.com/open-telemetry/otel-arrow/pull/2113/changes#diff-030bd32770b02f6f72d220fed49b389940624ebecab25f5ddaef795a76ab3bffR90-R98). Consumers look them up by extension name + concrete type:

```rust
let auth = registry.get::<ClientAuthenticatorHandle>("azure-auth")?;
let health = registry.get::<HealthCheckHandle>("azure-auth")?;
```

The difference from this PR (or #2141) is what gets registered: concrete handle types vs trait objects. In this design, `extension_traits!(BearerTokenProvider, HealthCheck)` registers `dyn Trait` objects (clones of the extension itself). In my approach, the factory creates purpose-built handle structs and registers them independently. The 1-to-N relationship is the same; the unit of registration differs.

gouslu: Fair criticism @utpilla, but this is by design. Extensions will require clone to be implemented, that would be the signal that extension developer needs for shared state. However, it is not explicit. I thought, together with documentation, it could be fine. This design trade-off comes from multiple decisions that lead to this:

- simplify extension traits, easy to add and maintain.
- simplify extension development.
- avoid complexity of mutex or any other more complex ways of dealing with memory barriers within the framework.
- wrap all mutable fields with `Arc<>` is a very basic concept to grasp, lowers barrier of entry and complexity of implementation.
- reduce boilerplate and hide behind a simple macro that is owned by the engine.

In PR #2113, it still seems that if your start method had to regularly update the token, it would have to find a way to share it with the extension instance, which would require the author to have an `Arc<>` struct field, in your case it is `Arc<Mutex<>>` since you made the handles `Sync`, which doesn't have to be required if the job is to provide a tonic interceptor. In #2113 registry still requires `Clone + Send + 'static` which can (in theory) lead to same error proneness at some place, if not by the extension implementer. In addition to all the plumbing required to connect handle to trait.

**Edit:** I understand the handle based design prevents the mistake, and I think that's the main benefit of it. I don't like the boilerplate of handle based design, as well as the mutex requirement to satisfy `Sync`. If we change this design to be similar to your handle based design, but drop the `Sync` requirement and maybe find ways to get rid of the boilerplate, I could be happier. I'll look into it a little bit and maybe some other alternatives.

docs/extension-system-architecture.md
gouslu: maybe, but an exporter might have an http client with a configurable retry, that uses sth like 'HttpRetryExtension', that can be configured? I don't know, maybe not a great example.

docs/extension-system-architecture.md
gouslu: Right, not a good example.

docs/extension-system-architecture.md
gouslu: No reason, we can add it to processors, I just didn't see a start method for processors and I needed to think a little longer how to hook processors up with it.. I was expecting this feedback, thanks.

docs/extension-system-architecture.md
gouslu: Right, not that kind of state, but I get your point :D

docs/extension-system-architecture.md
gouslu: I am not sure of that. Probably factory function(?). I will think of alternatives. I had other ideas too, but let me see what I can come up with as alternatives.

docs/extension-system-architecture.md
utpilla: **On simplicity of usage and plumbing:** In #2113, the "plumbing" is three lines in the factory: `ExtensionHandles::new()`, `.register(handle)`, and passing handles to `ExtensionWrapper`. The extension author implements a plain trait — no `#[derive(Clone)]`, no macro, no `Arc` wrapping. That's not adding any significant complexity, it's just more explicit wiring.

**On `Arc<Mutex<>>` and `Sync`:** The `Arc<Mutex<>>` is inside the handles. It's framework code, not extension-author code. The extension author just implements a simple trait for example `ClientAuthenticator: Send` (a plain trait, no `Sync`). **Neither the extension author nor the consumer ever writes `Arc` or `Mutex`**. As for whether `Sync ` is needed for tonic interceptors, the handle is what enters tonic services, and the framework makes it `Sync` automatically. The extension author doesn't think about it.

**On "same error proneness":** The `Clone + Send + 'static` requirement on handles is fundamentally different from `Clone` on the extension struct. A handle is a read endpoint — a small, purpose-built struct whose only job is to let consumers read state that the extension produces. There's no lifecycle logic, no credentials, no retry state. It's just a thin bridge between the extensions and its consumers. The extension struct is the whole machine. Requiring `Clone` on that means every field the author adds — config, credentials, senders, counters — must be individually assessed for `Arc`-worthiness. A handle doesn't have that problem because it only contains what consumers need to read, not what the extension needs to run.

**On sharing tokens between `start()` and the handle:** The extension author doesn't use `Arc` for this. The factory creates two separate objects — the extension (which produces state) and the handle (which lets consumers read it). They're connected internally, but neither the extension author nor the consumer has to think about shared ownership. In #2141, the extension struct is the handle, so every field that `start()` mutates must be manually wrapped in `Arc` to stay in sync with the cloned copies consumers receive.

gouslu: @utpilla, @lquerel I have gone full circle, and I stand by this design, because every other pathway puts us in a situation where the trade-off seems worse to me. Handle based design seem to require `Arc<Mutex<>>` no matter what. Other alternatives introduce `Sync` boundary where we don't want. I still think that simple `Arc<>` wrapping of shared state can be very easy to review for and enforce. If the sentiment is very strong that leaving shared state management to the extension author is a bad idea, please let's talk about it in this week's OTel meeting.

gouslu: @utpilla I created a sample PR https://github.com/open-telemetry/otel-arrow/pull/2267 to demonstrate how an extension would look like with a `get_token` method that may or may not make an HTTP call. Async methods are essential for extensions in my opinion, and not having them is a deal breaker.

A few observations on the handle-based design:

1. **State sharing is not enforced** — when the provider and extension are separate structs, the author must manually wire shared state via `Arc::clone`. The compiler won't catch it if they create independent instances instead. This is the same error-proneness that you criticized this desgin for.

2. **Mutex lock overhead** — the `tokio::sync::Mutex` on the handle adds ~6ns per call compared to a direct vtable call - best case scenario when there is no CPU cache pressure. Minor, but worth noting for hot paths.

3. **Serialization of concurrent callers** — because the mutex is held across `.await` points inside `get_token()`, concurrent callers are forced sequential. If `get_token()` does I/O, their requests can't overlap — caller B waits for caller A to fully complete before it can even start. Without the mutex, both would proceed concurrently.

4. **Boilerplate** — the handle-based design requires two structs with overlapping fields, manual `Arc` plumbing between them, and a separate handle wrapper. In a single-struct design, the extension *is* the provider — no duplication, no wiring.

gouslu: This is the example extension based on what I propose in PR https://github.com/open-telemetry/otel-arrow/pull/2141 -> https://github.com/gouslu/otel-arrow/blob/97a702c1090c9df1a701bc5af7cf590547534603/rust/otap-dataflow/crates/contrib-nodes/src/extensions/azure_identity_auth_extension/extension.rs

gouslu: I have found out that there are multiple popular crates out there that takes the same approach to shared state as mine. Axum, built by the tokio team (tokio-rs/axum), uses the same `Clone + Arc` convention for shared state — documented, not enforced at compile time. You can see the documentation here for [shared mutable state](https://docs.rs/axum/latest/axum/extract/struct.State.html#shared-mutable-state) in this section:

image

This approach that I have also ended up with seems to be adopted by multiple widely used libraries, including `axum` with 1.4k forks and 25.3k stars.

image

What I have come up with the extension system I have developed is basically a service locator pattern while avoiding `Sync` boundaries entirely, and one of the main differences with `axum` seems to be that axum uses generics for state, such as `Extension<T>` where T is the state. Our design wants to avoid a defined state as a generic type parameter because each extension implementer can have their own shared state.

All these being said, inspired by `axum`, I am trying to figure out if I can actually group all shared state into a struct and make it available in trait implementations, somehow. I'll keep working on it, but I am not sure if it will get anywhere without a significant trade-off in the design.

Edit: The same state-sharing pattern (author is responsible for `Arc<>` wrapping due to `Clone` requirement) can be found in `tonic`, `tower`, and `warp` — all by the tokio team all of them rely on documentation and tutorials for this.

cc: @lquerel, @jmacd, @utpilla

gouslu: What I think I would document for the extension authors to make shared state management easy for them is to follow a convention where they group up their internal shared state into a struct and wrap the whole struct in `Arc<MySharedState>` and only access it via `self.state` or something like that. Which is still not enforced by the compiler but it is a convention to make shared state management easier for extension authors.

Like I mentioned, I will try to find a way to enforce this pattern, but I am not very hopeful that it'll come as an "all-win" situation.

utpilla: I prefer **correctness by construction** — where the framework's types and API make incorrect usage impossible — over **correctness by convention**, where the author has to remember to do the right thing (e.g., `Arc`-wrap shared fields).

That said, I'll admit this does make the clone-self approach in #2141 more acceptable than I initially thought. If Axum has normalized this pattern, we could too. But why settle for "acceptable" when we can make the problem impossible in the first place?

gouslu: The claim about the handle based system not having the same issue is incorrect and I have demonstrated it [here](https://github.com/open-telemetry/otel-arrow/pull/2267/changes#r2919488333).

Maybe you can demonstrate how it is made impossible with an example.

Imagine a scenario where both lifecycle and the extension instances need to write to the same data. It could, for example, be a ratelimiter, where lifecycle resets the quota based on time or information accessible to the extension from the engine via start method, and the extension has a "try_acquire" method to get quota.

PR #2234, opened 2026-03-07
Embedded o11y UI

lquerel:
# Change Summary

This PR replaces the previous admin dashboard with a richer observability UI. This UI is integrated with the following endpoints: `/metrics`, `/readyz`, `/livez`, and `/status`.

Key changes:
- Integrates the new UI into otap-df-admin and embeds assets into the binary.
- Serves UI from:
   - GET /
   - GET /dashboard (alias)
   - GET /static/* for embedded assets.
- Vendors frontend runtime dependencies (i.e. tailwind and chartjs) under crates/admin/ui/vendor (no CDN dependency at runtime).
- Removes the old dashboard.html flow.

- cargo check -p otap-df-admin
- scripts/run-ui-js-tests.sh

Yes, the UI has undergone many changes.

### DIFF
--- a/rust/otap-dataflow/docs/admin/architecture.md
+++ b/rust/otap-dataflow/docs/admin/architecture.md
@@ -27,9 +27,10 @@ Request flow:
 2. Browser loads static assets from `/static/*`.
 3. UI polls `/telemetry/metrics` (or `/metrics` alias) with:
    - `format=json`
-   - `reset=true`
+   - `reset=false`
    - optional `keep_all_zeroes=true|false`
-4. Client logic derives rates/summaries and updates cards, charts, and topology.
+4. Client logic derives deltas/rates/summaries and updates cards, charts, and
+   topology.
 
 When calling endpoints directly outside browser-relative paths, use:
 
--- a/rust/otap-dataflow/crates/admin/ui/js/main.js
+++ b/rust/otap-dataflow/crates/admin/ui/js/main.js
@@ -73,12 +73,13 @@
   // Query params tune metrics query behavior.
   const urlParams = new URLSearchParams(window.location.search);
 
-  // Metrics endpoint strategy: reset snapshots each poll, and keep-all-zeroes is configurable.
+  // Metrics endpoint strategy: read cumulative snapshots (no server-side reset),
+  // compute deltas client-side, and keep-all-zeroes is configurable.
   const keepAllZeroesParam = urlParams.get("keep_all_zeroes");
   const keepAllZeroes =
     keepAllZeroesParam == null ? true : keepAllZeroesParam === "true";
   const METRICS_URL_CANDIDATES = buildMetricsCandidates({
-    query: `format=json&reset=true&keep_all_zeroes=${keepAllZeroes ? "true" : "false"}`,
+    query: `format=json&reset=false&keep_all_zeroes=${keepAllZeroes ? "true" : "false"}`,
   });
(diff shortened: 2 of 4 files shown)


albertlockett (approved):

rust/otap-dataflow/scripts/run-ui-js-tests.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+# Run admin UI JavaScript module tests.
```
drewrelmas, line 2: Not a blocker for this PR, but do we want to include these tests in the Rust-CI workflow?

rust/otap-dataflow/docs/admin/architecture.md
```diff
@@ -0,0 +1,261 @@
...
+The UI and telemetry endpoints are served from one origin by the same admin
+HTTP server.
+
+Request flow:
+
+1. `GET /` or `GET /dashboard` serves the embedded UI shell.
+2. Browser loads static assets from `/static/*`.
+3. UI polls `/telemetry/metrics` (or `/metrics` alias) with:
+   - `format=json`
+   - `reset=true`
```
cijothomas: is it intentional we reset? if a user has a prometheus scraper, then this would likely make its job more difficult.

lquerel: At first, it seemed simpler to start with an approach based on resets and deltas, but I think I can do without it. I'm checking whether I can include this in this PR or implement it in a separate PR.

rust/otap-dataflow/scripts/run-ui-js-tests.sh
lquerel: I have this section in one of the Markdown docs. It describes what I had in mind for the CI integration. Let me know what you think.

```
## Future CI integration

UI module tests should be integrated as a **separate GitHub Actions job**
focused only on JavaScript UI validation.

Recommended UI-only CI job:

1. Trigger only when UI JS paths change:
   - `rust/otap-dataflow/crates/admin/ui/js/**`
   - `rust/otap-dataflow/scripts/run-ui-js-tests.sh`
2. Run only:
   - `./rust/otap-dataflow/scripts/run-ui-js-tests.sh`
3. Mark this job as its own required PR status check
   (independent from Rust checks).

Practical CI notes:

- Pin a Node.js version in CI to avoid drift.
- Keep the JS tests browserless and deterministic
  (no network, no timers by default).
- Keep this job independent from `rust-required-status-check` so UI failures
  are isolated.
```

drewrelmas: Looks great @lquerel - I missed this content, thanks

drewrelmas: > 3. Mark this job as its own required PR status check
   (independent from Rust checks).
   
One minor concern about making this a required check if it doesn't always run (dependent on path changes) but let's figure that out when we get there.

lquerel: I'm definitely not an expert when it comes to the workflow side of GitHub. I'd appreciate some help with this task.

rust/otap-dataflow/docs/admin/architecture.md
lquerel: The client is no longer using reset=true.

PR #2240, opened 2026-03-09
fix: Take otap spec into account when selecting schemas

### LINKED ISSUE #2204: Concatenate does not take into account the otap spec when picking dictionary sizes

JakeDern:
# Change Summary

This PR fixes a bug with concatenation not taking the spec into account by:

1. Adding some basic spec definitions that contain all of the allowed payload types per column
2. Updating the concatenate code to take the spec into account along with signal types

I think the spec definitions are the most interesting part and I'd like to expand on these in future PRs for a variety of purposes including generating more robust test cases.

* Closes #2204 

Unit

No


rust/otap-dataflow/crates/pdata/src/schema/consts.rs
```diff
@@ -73,6 +73,40 @@ pub const ATTRIBUTE_BOOL: &str = "bool";
...
+pub const BODY_INT: &str = "body.int";
+pub const BODY_DOUBLE: &str = "body.double";
+pub const BODY_BOOL: &str = "body.bool";
+pub const BODY_BYTES: &str = "body.bytes";
+pub const BODY_SER: &str = "body.ser";
+
+// resource sub-fields
+pub const RESOURCE_ID: &str = "resource.id";
+pub const RESOURCE_DROPPED_ATTRIBUTES_COUNT: &str = "resource.dropped_attributes_count";
+pub const RESOURCE_SCHEMA_URL: &str = "resource.schema_url";
```
albertlockett, line 88: Are these dotted paths used anywhere?

It looks like for accessing struct columns, we call get_nested which takes the struct name and field name iiuc

JakeDern: Ah yeah, I implemented the ability to do both since sometimes we refer to them as `resource.id` in the transform module - Maybe wasn't necessary for now though as I didn't end up referencing them. I can take them out or maybe leave it for now in case

albertlockett: ah OK - we can leave it in. I think it'll be useful, plus there are test cases covering the access pattern :) I was moreso just asking for my own edification!

PR #2242, opened 2026-03-09
feat: Add otel_event! generalized log macro with runtime severity support

### LINKED ISSUE #2071: feat: Generalized log macro
We have some logging macros: `otel_debug`, `otel_info`, `otel_warn`, and `otel_error`.

These require that you know the severity up front.

This code has to select the correct macro based on a `match`: https://github.com/open-telemetry/otel-arrow/blob/d38292fda6a22b9f1bcfcee6baa8825c5b6ab7da/rust/otap-dataflow/crates/otap/src/experimental/recordset_kql_processor/processor.rs#L101-L131

According to @cijothomas, tracing lib has an `event!` macro which is a lower-level thing where you can pass in the severity.

This issue is to capture the potential need for an "otel" macro which allows the severity to be passed in.

Having the specialized macro would save on some callsite static memory being wasted in cases like the code above.

CodeBlanch:
Fixes #2071
Relates to #2072

# Changes

* Uses the code `copilot` tried to introduce on #2072 to implement `otel_event!`
* Updates `RecordSetKqlProcessor` to use the new event


albertlockett:
LGTM! Thanks @CodeBlanch

PR #2244, opened 2026-03-09
Fix nightly batch-size perf tests

cijothomas:
1. Continue executing remaining test steps after a failure so container logs are always captured for diagnostics.
2. Disable all OTAP tests in 100klrps-batch-sizes-docker.yaml — they hit drain deadline exceeded at 100k logs/sec. I put a TODO in the OTAP section to get back to this once we know more. This is done to unblock the remaining tests. (OTLP OTLP tests are running fine)
3. Add if: !cancelled() to all nightly test and upload steps so independent suites run to completion even when a prior suite fails (as each suites are independent)


cijothomas:
Nightly Perf runs got broken due to this, this should get us up and running! (the infra issue also cost a week or so of runs)

PR #2246, opened 2026-03-10
Add batchprocessor to perf tests

cijothomas:
Blocked on https://github.com/open-telemetry/otel-arrow/issues/2194

Trying to introduce batch processor to Perf tests, so as to catch ^ issues earlier. And also to actually measure the perf impact of batching!

### DIFF
--- a/rust/experimental/query_engine/expressions/src/data_expressions.rs
+++ b/rust/experimental/query_engine/expressions/src/data_expressions.rs
@@ -199,24 +199,50 @@ impl ConditionalDataExpression {
         self.default_branch.as_deref()
     }
 
+    /// The try_fold implementation for this type of expression will recursively call the relevant
+    /// optimization methods on each branch's conditions and the expressions within each branch. It
+    /// also will remove branches if it determines that they would not evaluate on any rows.
     pub(crate) fn try_fold(
         &mut self,
-        _scope: &PipelineResolutionScope,
+        scope: &PipelineResolutionScope,
     ) -> Result<(), ExpressionError> {
-        // TODO support folding. What this should do is:
-        //
-        // 1) for each branch, check if's condition can be folded into a boolean literal using
-        // `LogicalExpression::try_resolve_static`. If so:
-        // - if the result is false, discard the branch because it would evaluate on zero rows
-        // - if the result is true, discard all the subsequent branches and the default branch
-        //   because no rows would be evaluated by them
-        //
-        // 2) recursively call try_fold on all the expressions in every remaining branch and the
-        // default branch if still present.
-        //
-        // Before doing this, we should support filtering by static literals so we can write unit
-        // tests that will evaluate the folded expr. Without this, we won't be able to evaluate the
-        // resolved static as a filter. https://github.com/open-telemetry/otel-arrow/issues/1508
+        let mut i = 0;
+        while i < self.branches.len() {
+            let static_logical_expr = {
+                let branch = &mut self.branches[i];
+                branch.condition.try_resolve_static(scope)?
+            };
+
+            if let Some(static_logical_expr) = static_logical_expr {
+                if static_logical_expr {
+                    // here everything will pass the filter. That means we can drop the test of the
+                    // branches because all remaining rows will be evaluated by this branch
+                    _ = self.branches.split_off(i + 1);
+
+                    // drop the default branch - no need to keep it as there will be now rows for
+                    // for it to evaluate
+                    self.default_branch = None;
+                } else {
+                    // here nothing will pass the filter, so we can basically discard this branch
+                    self.branches.remove(i);
+                    continue;
+                }
+            }
+
+            // optimize the expressions inside the branch
+            for expression in &mut self.branches[i].expressions {
+                expression.try_fold(scope)?;
+            }
+
+            i += 1;
+        }
+
+        if let Some(default_branch) = self.default_branch.as_mut() {
+            // optimize the expressions inside the branch
+            for expression in default_branch {
+                expression.try_fold(scope)?;
+            }
+        }
 
         Ok(())
     }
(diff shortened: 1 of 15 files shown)


JakeDern:
Are you hitting the issue described in #2194 when running these? I've actually been working on adding batch processor benchmarks too and whenever I've run them, they don't hit the limits with the current load generation setup

JakeDern:
This seems like as good a place as any to discuss benchmarks so I'm also curious about the scenarios that we want to test and whether to do continuous vs nightly. Just like with the current continuous benchmarks with the attribute processor, batching is sensitive to permutations across:

- otap -> batch -> otap
- otlp -> batch -> otlp
- otap -> batch -> otlp
- otlp -> batch -> otap

And across all three signal types as we take different code paths. My proposals would be:

1. We should add batch processor benches to continuous as it's a critical core component and very sensitive to changes in terms of perf
2. We should add otap -> otap and otlp -> otlp as a baseline to the continuous charts
3. We should add the signal type as a dimension to the above matrix of scenarios because perf characteristics can and likely will be different across them

We may want to add the ability to filter by dimensions in the aggregated charts to keep them sane - Something like checkboxes for signal (logs/metrics/traces) and scenario (batch/baseline/otap).

CC: @lquerel @clhain for any thoughts!

clhain:
Broadly agree with all that Jake.

I think generally it would be a good idea to re-visit the line between continuous and nightly in light of the many new engine capabilities and see how we can pack more tests in while still aiming to not monopolize the shared runner instance (e.g. we can probably avoid blowing away the backend/loadgenerator after each test, parallelizing some of them, running some with loadgen in-process, etc, etc).

I also agree that the chart layouts aren't really scaling well or providing an intuitive view of the world. Joe from F5 is looking at an alternative interface for comparing across different agents, which might be useful here when done. 

The orchestration framework has the ability to output all of the internal test data as parquet files (high level summary as well as second by second metrics)... I have a vision where we replace the whole 'store a big blob of json and use some static html charts on top of it' with 'store a few parquet files and use the duckdb wasm plugin on top of those to provide a highly interactive view of test results over much longer time periods'. One of these days...

Obviously this is all just brain dump, no need to address as part of this PR =P

JakeDern:
> Broadly agree with all that Jake.
> 
> I think generally it would be a good idea to re-visit the line between continuous and nightly in light of the many new engine capabilities and see how we can pack more tests in while still aiming to not monopolize the shared runner instance (e.g. we can probably avoid blowing away the backend/loadgenerator after each test, parallelizing some of them, running some with loadgen in-process, etc, etc).
> 
> I also agree that the chart layouts aren't really scaling well or providing an intuitive view of the world. Joe from F5 is looking at an alternative interface for comparing across different agents, which might be useful here when done.
> 
> The orchestration framework has the ability to output all of the internal test data as parquet files (high level summary as well as second by second metrics)... I have a vision where we replace the whole 'store a big blob of json and use some static html charts on top of it' with 'store a few parquet files and use the duckdb wasm plugin on top of those to provide a highly interactive view of test results over much longer time periods'. One of these days...
> 
> Obviously this is all just brain dump, no need to address as part of this PR =P

Totally makes sense! I think it probably wouldn't be too hard to get a couple of checkboxes for signal/scenario dimensions into the continuous benchmark html charts (famous last words) in the meantime. I'm motivated to get some good batch processor benches up and running in the near term so I'm happy to employ some robot helpers and see if I can hack that together if that's the direction we want to go.

If there's concerns about monopolizing the runner because the matrix is too big though also totally understand that. The proposal would be (batch(4) + baseline(2) + attr(4)) * signals (3) = 30 runs vs the 4 on that continuous chart today.

We can alternatively start smaller and just add the batch + baseline scenarios into the chart (6 more scenarios) and not multiply across signals until we make other changes to use the shared runner more efficiently.

JakeDern:
I could also experiment with the larger matrix as a nightly and we can see how we like it and then add a smaller additional set to the continuous

clhain:
Glancing at it now, the current setup is taking anywhere from 30min to 1hr+ (!), and there's like 5 runs queued up... Maybe it's mostly time spent building, but I'd say start with nightly and we need to get those numbers way down before we start multiplying anything =P

clhain:
> Glancing at it now, the current setup is taking anywhere from 30min to 1hr+ (!), and there's like 5 runs queued up... Maybe it's mostly time spent building, but I'd say start with nightly and we need to get those numbers way down before we start multiplying anything =P

ah nvm that 1hr+ must include queue time... looks like they mostly finish in ~30 min (breakdown: 4m build, 16+3+4m for logs, idle, passthrough - this seems high).

JakeDern:
That's a decent amount of time with the rate of PRs we have! How do we feel about adding just otap-batch-otap and otlp-batch-otlp to the continuous to give us something and have everything else be nightly? Maybe it's still too much given the queues we have...

cijothomas:
> That's a decent amount of time with the rate of PRs we have! How do we feel about adding just otap-batch-otap and otlp-batch-otlp to the continuous to give us something and have everything else be nightly? Maybe it's still too much given the queues we have...

Majority of the tests can be pushed to Nightly. I started with continuous to have it running frequently and adjust settings based on the runs. The plan was to always move them to nightly.

 I am also working on https://github.com/open-telemetry/otel-arrow/pull/1528 to make sure all the things we test/publish about perf is captured.

clhain:
I filed this, looks like something is broken with shutdown calls: https://github.com/open-telemetry/otel-arrow/issues/2257

That should drop the 16 minute one down to like 4 when fixed if someone has time to investigate.

I definitely support having some batch testing in continuous (seems more useful than ATTR in any case), starting with the basic 2 and the rest in nightly sounds like a solid plan for now.

rust/otap-dataflow/crates/otap/src/batch_processor.rs
```diff
@@ -195,7 +195,7 @@ const fn default_otap_max_size_items() -> Option<NonZeroUsize> {
 }
 
 const fn default_otap_sizer_items() -> Sizer {
-    Sizer::Bytes
+    Sizer::Items
```
jmacd, line 199: Yes! (How did ...)

PR #2250, opened 2026-03-10
AzureMonitorExporter - logging fixes

cijothomas:
Per-retry attempts were logged at WARN, creating noise during transient outages, while the actual data loss event (retries exhausted) had no dedicated log. The attempt counter was also incorrect for server-driven retries.

Changes

Downgrade per-retry logs (export.retrying, client.error) to debug
Add warn-level export.retries_exhausted when MAX_RETRIES is reached — the actionable data loss signal
Fix attempt counter to increment on all failures, not just non-server-driven ones
Add TODO for missing upper bound on server-driven retries (429 with Retry-After)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -1,7 +1,7 @@
 name: Bug Report
 description: Report a bug in OTel-Arrow
 type: Bug
-labels: ["Bug"]
+labels: ["bug"]
 body:
   - type: markdown
     attributes:
--- a/.github/ISSUE_TEMPLATE/feature_request.yaml
+++ b/.github/ISSUE_TEMPLATE/feature_request.yaml
@@ -1,7 +1,7 @@
 name: Feature Request
 description: Suggest a new feature or enhancement for OTel-Arrow
 type: Feature
-labels: ["Feature"]
+labels: ["enhancement"]
 body:
   - type: markdown
     attributes:
--- a/.github/workflows/pipeline-perf-on-label.yaml
+++ b/.github/workflows/pipeline-perf-on-label.yaml
@@ -108,7 +108,6 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --user --upgrade pip
           python -m pip install --user --require-hashes -r tools/pipeline_perf_test/orchestrator/requirements.lock.txt
           python -m pip install --user --require-hashes -r tools/pipeline_perf_test/load_generator/requirements.lock.txt
 
@@ -145,7 +144,6 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --user --upgrade pip
           python -m pip install --user --require-hashes -r tools/pipeline_perf_test/orchestrator/requirements.lock.txt
           python -m pip install --user --require-hashes -r tools/pipeline_perf_test/load_generator/requirements.lock.txt
 
--- a/.github/workflows/pipeline-perf-test-continuous.yml
+++ b/.github/workflows/pipeline-perf-test-continuous.yml
@@ -76,7 +76,6 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --user --upgrade pip
           python -m pip install --user --require-hashes -r tools/pipeline_perf_test/orchestrator/requirements.lock.txt
           python -m pip install --user --require-hashes -r tools/pipeline_perf_test/load_generator/requirements.lock.txt
 
--- a/.github/workflows/pipeline-perf-test-manual-pr.yaml
+++ b/.github/workflows/pipeline-perf-test-manual-pr.yaml
@@ -99,7 +99,6 @@ jobs:
 
       - name: Install dependencies
         run: |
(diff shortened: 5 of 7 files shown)


gouslu:
you should be able to use any of these metrics or logs for actual data loss:

image

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/client.rs
```diff
@@ -191,14 +191,20 @@ impl LogsIngestionClient {
                     });
                 }
                 Err(e) => {
+                    attempt += 1;
+
+                    // ToDo: Add an upper bound for server-driven retries (429/5xx with
+                    // Retry-After). Currently only the non-server-driven path enforces
+                    // MAX_RETRIES; a server that perpetually returns 429 with Retry-After
+                    // will cause this loop to retry indefinitely.
                     let delay = if let Some(server_delay) = e.retry_after() {
```
gouslu, line 200: when would the server perpetually return Retry-After? I chose to trust Retry-After header here especially and not count it towards attempts. I understand the concern though, but is your concern valid? Meaning, why would you not trust the Retry-After header if server explicitly returns it?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/client.rs
```diff
@@ -191,14 +191,20 @@ impl LogsIngestionClient {
...
                         let base_delay = server_delay.max(Duration::from_secs(5));
                         let jitter = Duration::from_secs(3)
                             + Duration::from_secs_f64(rng.random::<f64>() * 7.0);
                         base_delay + jitter
                     } else {
-                        attempt += 1;
                         if attempt >= MAX_RETRIES {
+                            otel_warn!("azure_monitor_exporter.export.retries_exhausted", attempts = attempt, error = ?e);
```
gouslu: this is useful, but may not be necessary as we always log failed export attempts after all retries are exhausted, and have 3 metrics for it.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/client.rs
```diff
@@ -210,7 +216,7 @@ impl LogsIngestionClient {
                         base_delay.mul_f64(jitter_factor)
                     };
 
-                    otel_warn!("azure_monitor_exporter.export.retry_delay", delay_ms = delay.as_millis() as u64, error = ?e);
+                    otel_debug!("azure_monitor_exporter.export.retrying", attempt = attempt, delay_ms = delay.as_millis() as u64, error = ?e);
```
gouslu, line 219: Why would you consider that to be a noise? I would say in a system where we expect everything to work almost all the time, transient errors should be warnings, and if we start seeing these warnings so often that it creates such a noise, maybe then we can do something about it. But, we need at least one way of tracking these transient issues. We can't just be like, oh, if it is transient, it is OK.

Try export method doesn't track network errors, so I would say either we should track all errors in try_export by tracking all errors there, or keep this as warn.

I also think that keeping this as warn can help me optimize retry logic and retry delay logic.

I am not sure of this change, I am OK with it as long as my concerns above are addressed.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/client.rs
gouslu: Also, even if you are going to add exhaustion to Retry-After, I would say make it higher and a separate limit.

cijothomas: Agreed — the TODO doesn't change any behavior, just flags the edge case for future consideration. If a server ever misbehaves, we can add a defensive upper bound then. Resolution may well be "do nothing."

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/client.rs
cijothomas: Good point. Yes the %Error also shows the last_error, so this is not really adding new value. removing.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/client.rs
cijothomas: Transient failures during retry are normal, business-as-usual situationr — that's what the retry loop is for. Logging each attempt at WARN creates noise (up to 80 WARN lines for a brief blip with 5 retries × 16 in-flight batches), making it harder to spot actual problems. I've moved these to DEBUG; open to INFO as a compromise, but WARN should be reserved for situations that truly need operator attention.

Good catch on -  that network errors aren't currently tracked via metrics (only HTTP status codes are). I'd prefer to address that separately — adding a laclient_network_errors counter would give visibility into transient connection issues without relying on log levels. I can file an issue to track metric gaps.

PR #2254, opened 2026-03-10
fix heartbeat table mappings

gouslu:
# Change Summary

Fixes the values pushed to the heartbeat table.

Local, manual testing and unit tests.

No

### DIFF
--- a/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
@@ -49,9 +49,6 @@ struct HeartbeatRow {
 
     #[serde(rename = "Version")]
     version: String,
-
-    #[serde(rename = "SCAgentChannel")]
-    sc_agent_channel: String,
 }
 
 #[inline]
@@ -147,7 +144,6 @@ impl Heartbeat {
                 os_major_version: default_heartbeat_os_major_version(),
                 os_minor_version: default_heartbeat_os_minor_version(),
                 version: default_heartbeat_version(),
-                sc_agent_channel: "Direct".to_string(),
             },
             auth_header: HeaderValue::from_static("Bearer "),
         })
@@ -168,7 +164,6 @@ impl Heartbeat {
                 os_major_version: "1".to_string(),
                 os_minor_version: "0".to_string(),
                 version: "test-version".to_string(),
-                sc_agent_channel: "Direct".to_string(),
             },
             auth_header: HeaderValue::from_static("Bearer "),
         }
@@ -266,7 +261,6 @@ mod tests {
             os_major_version: "22".to_string(),
             os_minor_version: "04".to_string(),
             version: "1.0.0".to_string(),
-            sc_agent_channel: "Direct".to_string(),
         };
 
         let json = serde_json::to_value(&row).unwrap();
@@ -278,7 +272,6 @@ mod tests {
         assert_eq!(json["OSMajorVersion"], "22");
         assert_eq!(json["OSMinorVersion"], "04");
         assert_eq!(json["Version"], "1.0.0");
-        assert_eq!(json["SCAgentChannel"], "Direct");
     }
 
     #[test]
@@ -291,7 +284,6 @@ mod tests {
             os_major_version: "".to_string(),
             os_minor_version: "".to_string(),
             version: "".to_string(),
-            sc_agent_channel: "".to_string(),
         };
 
         let json = serde_json::to_string(&row).unwrap();
@@ -304,14 +296,12 @@ mod tests {
         assert!(json.contains("\"OSMajorVersion\""));
         assert!(json.contains("\"OSMinorVersion\""));
         assert!(json.contains("\"Version\""));
-        assert!(json.contains("\"SCAgentChannel\""));
 
         // Verify no snake_case field names
         assert!(!json.contains("\"time\""));
         assert!(!json.contains("\"version\""));
         assert!(!json.contains("\"os_name\""));
         assert!(!json.contains("\"os_type\""));
-        assert!(!json.contains("\"sc_agent_channel\""));
     }
 
     #[test]
@@ -324,7 +314,6 @@ mod tests {
             os_major_version: "22".to_string(),
             os_minor_version: "04".to_string(),
             version: "1.0.0".to_string(),
-            sc_agent_channel: "Direct".to_string(),
         };
 
         let payload = serde_json::json!([row]);


rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
```diff
@@ -97,15 +97,45 @@ fn parse_os_version() -> (String, String) {
...
+    match std::env::consts::OS {
+        "linux" => "Linux".to_string(),
+        "windows" => "Windows".to_string(),
+        "macos" => "MacOS".to_string(),
+        other => other.to_string(),
+    }
+}
+
+/// Returns private IPv4 addresses as a JSON array string, e.g. `["10.0.0.1","192.168.1.5"]`.
+/// Matches the C++ ODSUploader behavior of `MdsdUtil::GetPrivateIps()`.
```
cijothomas: ?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
```diff
@@ -30,28 +28,31 @@ pub struct Heartbeat {
 
 #[derive(Serialize)]
 struct HeartbeatRow {
-    #[serde(rename = "Time")]
-    time: String,
+    #[serde(rename = "Computer")]
+    computer: String,
```
cijothomas, line 36: is time not required?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
```diff
@@ -30,28 +28,31 @@ pub struct Heartbeat {
...
     #[serde(rename = "OSMajorVersion")]
     os_major_version: String,
 
     #[serde(rename = "OSMinorVersion")]
     os_minor_version: String,
+
+    #[serde(rename = "Version")]
+    version: String,
+
+    #[serde(rename = "ComputerPrivateIps")]
```
cijothomas: don't think it is okay to collect IP address without explicit opt-in to such feature.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
```diff
@@ -97,15 +97,45 @@ fn parse_os_version() -> (String, String) {
...
+    minor
+}
+
+#[inline]
+fn default_heartbeat_os_type() -> String {
+    match std::env::consts::OS {
+        "linux" => "Linux".to_string(),
+        "windows" => "Windows".to_string(),
+        "macos" => "MacOS".to_string(),
+        other => other.to_string(),
```
cijothomas, line 117: https://learn.microsoft.com/en-us/azure/azure-monitor/reference/tables/heartbeat
Type of OS. Possible values are Windows or Linux.

Doc says only Windows or Linux allowed.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
```diff
@@ -97,15 +97,45 @@ fn parse_os_version() -> (String, String) {
...
+/// Returns private IPv4 addresses as a JSON array string, e.g. `["10.0.0.1","192.168.1.5"]`.
+/// Matches the C++ ODSUploader behavior of `MdsdUtil::GetPrivateIps()`.
+fn default_heartbeat_computer_private_ips() -> String {
+    let networks = Networks::new_with_refreshed_list();
+    let mut ips: Vec<String> = Vec::new();
+
+    for (_name, data) in &networks {
+        for ip_network in data.ip_networks() {
+            if let std::net::IpAddr::V4(v4) = ip_network.addr {
+                if v4.is_private() && !v4.is_loopback() {
```
ThomsonTan: Why is the loopback IP excluded which is also a private IP?

gouslu: based on AMA logic, it seems to exclude localhost.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
gouslu: MacOS works fine, I'll keep it.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
gouslu: yeah, I'll add it back based on new information I got from Drew.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
gouslu: yeah, you have a point. removing it.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
gouslu: either way, removing it per Cijo's comment.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
gouslu: image

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
```diff
@@ -33,25 +32,31 @@ struct HeartbeatRow {
...
     #[serde(rename = "OSMajorVersion")]
     os_major_version: String,
 
     #[serde(rename = "OSMinorVersion")]
     os_minor_version: String,
+
+    #[serde(rename = "Version")]
+    version: String,
+
+    #[serde(rename = "SCAgentChannel")]
```
cijothomas: not sure what is this particular field? Why we are adding it vs not adding others.

gouslu: Why do you want to remove it? Is that problematic for some reason?

gouslu: I have just added whatever seemed that would be provided by the client based on some info Drew shared with me. I can remove it if it is unnecessary or problematic (?).

cijothomas: heartbeat table has a lot of fields - why add some specially? What is the criterion used?

gouslu: image

I wanted to make sure that it is set to Direct, because we are sending it directly from this exporter.

cijothomas: Got it.
This rust library can be embedded anywhere. Unless it is a mandatory field, we don't need to send.

Alternatively, I think the best option is to make HeartBeat configurable (what to send, what values to send, and even a full turn-off flag). The thing that embeds this library can make best determination about what to send, instead of the librarymaking assumptions.

gouslu: this PR is not intended for configurability, we need to set an acceptable baseline, we can add configurability later based on expressed need. can you please tell me which fields you think should be removed from here?

cijothomas: if configurability is not the goal for this PR, that is okay. Lets not add new fields either in this PR - we can use this PR for fixing the values of what we emit. (Then we can follow up with additional fields and configurability separately).

gouslu: i'll remove sc_agent_channel, the others look good to me.

PR #2241, opened 2026-03-09
chore(deps): Configure Renovate to work with python requirements.lock.txt files

drewrelmas:
# Change Summary

Attempt to add `requirements.lock.txt` for the two `requirements.txt` files present in this repo. In addition, configure Renovate for working with those lock files.

Relevant docs:
* https://docs.renovatebot.com/modules/manager/pip-compile/
* https://docs.renovatebot.com/configuration-options/#lockfilemaintenance

## What issue does this PR close?

Security warnings with limited visibility.

We have multiple long-standing warnings about:
> pipCommand not pinned by hash

These impact the repo OpenSSF Scorecard.

## How are these changes tested?

CI tests Renovate config validity - will need to wait for next update to truly see behavior.

## Are there any user-facing changes?

N/A

PR #2256, opened 2026-03-10
chore(deps): Adjust remaining instances of pipCommand not being pinned by hash

drewrelmas:
# Change Summary

Follow-up to #2241 

Security warnings from GitHub

n/a

n/a

### DIFF
--- a/.github/workflows/pipeline-perf-on-label.yaml
+++ b/.github/workflows/pipeline-perf-on-label.yaml
@@ -108,7 +108,6 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --user --upgrade pip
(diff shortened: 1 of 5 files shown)


.github/workflows/pipeline-perf-on-label.yaml
```diff
@@ -108,7 +108,6 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --user --upgrade pip
```
drewrelmas, line 111: I could be wrong about this, but there isn't a specific value to ensuring latest `pip` every run? We will just use whatever version of `pip` is available on the runner?

Image

Even though `--require-hashes` is used below, GitHub scanning seems to consider the upgrade itself another non-compliant usage.

PR #2261, opened 2026-03-10
fix: Drop senders so that shutdown can complete

### LINKED ISSUE #2257: Perf Test - Unexpectedly long shutdown time
### Pre-filing checklist

### Component(s)

CI / Build

### Description

Took a quick look at perf test results and noticed it's taking a full minute to stop the engine (so 3 minutes for each test because loadgen/backend/sut)... seems like this sort of call takes the full 60s timeout (presumably unexpected):  https://github.com/open-telemetry/otel-arrow/blob/main/tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker.yaml#L132

[2026-03-10 18:10:28] INFO | ctx.type=StepContext ctx.name=Stop Load Generator | Executing main step action...
[2026-03-10 18:11:27] INFO | ctx.type=StepContext ctx.name=Stop Engine | Executing main step action...
[2026-03-10 18:12:26] INFO | ctx.type=StepContext ctx.name=Stop Backend | Executing main step action...

### Additional Context

_No response_

JakeDern:
# Change Summary

The pipeline shutdown sequence depends on all the nodes dropping their node control message senders so that the channel closes and the manager exits its loop, but the admin endpoint prevents that by holding references to the senders.

This PR wraps the admin senders in a mutex so that we can take ownership of them and drop them all and therefore let the sequence complete.

The original pipeline ctrl msg tx was also dangling.

* Closes #2257

With a curl command: `curl -X POST  "http://localhost:8082/pipeline-groups/shutdown?wait=true&timeout_secs=60" -H "Content-Type: application/json"`.

 I have no idea how to write a reasonable automated test for this :)

No

### DIFF
--- a/rust/otap-dataflow/configs/README.md
+++ b/rust/otap-dataflow/configs/README.md
@@ -128,20 +128,21 @@ Syslog/CEF receiver with performance metrics:
 - Measures and exports performance metrics
 - View metrics at: `http://127.0.0.1:8080/telemetry/metrics?format=prometheus&reset=false`
 
-To send test syslog messages:
+To send a quick test message (UDP):
 
 ```bash
-# Send a single syslog message
 echo "<134>$(date '+%b %d %H:%M:%S') testhost testtag: Test message" | nc -u -w1 127.0.0.1 5140
-
-# Send multiple messages
-for i in {1..100}; do
-  echo "<134>$(date '+%b %d %H:%M:%S') testhost testtag: Test message #$i" | nc -u -w1 127.0.0.1 5140
-done
-
-# Send CEF format message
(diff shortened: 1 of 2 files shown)


JakeDern:
I hit the wrong button and deleted the branch on my fork which closed the pr 🤦‍♂️. This is ready for review

lquerel (approved):
Thanks for fixing that! LGTM.

We will probably have to revisit this when we work on the live reconfiguration capability, since we will need to be able to shut down a specific pipeline instance. For now, though, we don’t need to be more granular.

PR #2262, opened 2026-03-10
[Geneva exporter] Add metrics, and enable telemetry for Geneva exporter.

lalitb:
# Change Summary

   The Geneva exporter was missing the `start_periodic_telemetry()` call, which meant `CollectTelemetry` messages were never delivered - all exporter metrics showed zero despite successful uploads.

   This PR fixes that and adds metrics to match what Azure Monitor exporter already tracks:
   
   - `batches_uploaded` / `batches_failed` — batch-level success/failure
   - `records_uploaded` / `records_failed` — individual record counts
   - `bytes_uploaded` — compressed payload throughput
   - `upload_duration` — upload latency (Mmsc)
   - `encode_duration` — encode + compress latency (Mmsc)
   - `conversion_errors`, `empty_payloads_skipped`, `unsupported_signals` — error path counters
   
   Tests updated to handle the new telemetry timer message.

* Closes #NNN

through unit tests.

N/A


rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -256,25 +293,43 @@ impl GenevaExporter {
...
+            Ok(()) => {
+                self.metrics.batches_uploaded.add(batches_encoded as u64);
+                self.metrics.records_uploaded.add(record_count);
+                let total_bytes: u64 = batches.iter().map(|b| b.data.len() as u64).sum();
+                self.metrics.bytes_uploaded.add(total_bytes);
+                Ok(batches_encoded)
+            }
+            Err(e) => {
+                self.metrics.batches_failed.add(batches_encoded as u64);
+                self.metrics.records_failed.add(record_count);
```
utpilla, line 329: When `try_for_each_concurrent` fails, it short-circuits on the first error. There are two issues here:
- This inflates `batches_failed`/`records_failed`
- It misses incrementing `batches_uploaded`/`records_uploaded` for the ones that did succeed

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -256,25 +293,43 @@ impl GenevaExporter {
...
                     .map_err(|e| format!("Failed to upload {:?} batch: {e}", signal_type))
             })
-            .await?;
-
-        self.metrics.batches_uploaded.add(batches_encoded as u64);
-        Ok(batches_encoded)
+            .await;
+
+        let upload_ms = upload_start.elapsed().as_secs_f64() * 1000.0;
+        self.metrics.upload_duration.record(upload_ms);
```
utpilla, line 317: `upload_duration` wraps the entire `try_for_each_concurrent` block, so it records one sample per call (the wall-clock time until the last concurrent upload finishes). That value reflects the aggregate of all concurrent uploads, not individual upload latency. For example, if a call uploads `5` batches concurrently and each takes `50ms`, you'd record a single `~50ms` sample with `count=1`, instead of `5` samples of `~50ms` each with `count=5`. The value reported doesn't represent any real upload that happened.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -158,9 +159,45 @@ struct ExporterMetrics {
...
+    #[metric(unit = "{msg}")]
+    pub empty_payloads_skipped: Counter<u64>,
+
+    /// Number of OTAP-to-OTLP conversion errors.
+    #[metric(unit = "{error}")]
+    pub conversion_errors: Counter<u64>,
+
+    /// Number of unsupported signal types rejected (e.g. metrics).
+    #[metric(unit = "{msg}")]
+    pub unsupported_signals: Counter<u64>,
```
utpilla, line 200: We could use a better name here. Metrics is the only unsupported signal. Maybe something like `metrics_payloads_dropped`?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -158,9 +159,45 @@ struct ExporterMetrics {
     #[metric(unit = "{batch}")]
     pub batches_encoded: Counter<u64>,
 
-    /// Total number of batches considered successfully uploaded.
+    /// Total number of batches successfully uploaded to Geneva.
     #[metric(unit = "{batch}")]
     pub batches_uploaded: Counter<u64>,
```
utpilla, line 164: All the upload/failure counters (`batches_uploaded`, `batches_failed`, `records_uploaded`, `records_failed`, `bytes_uploaded`) and the latency metrics (`upload_duration`, `encode_duration`) are shared across logs and traces. `signal_type` is passed into `upload_batches_concurrent` but only used in the error message string. The metrics don't distinguish which signal type they're counting.

If the exporter handles both logs and traces, there's no way to tell from the metrics which signal is failing or how throughput/latency differs between them.

One option: split the counters per signal, e.g. `log_batches_uploaded` / `trace_batches_uploaded`, or have **a small per-signal metrics struct** and select based on `signal_type`.

PR #2265, opened 2026-03-11
Add resource_attributes support to traffic generator

lalitb:
### Summary
  
The traffic generator's static data source produces hardcoded resource attributes with no way to customize them. This makes it impossible to load-test content-based routing pipelines (e.g. `content_router` routing by `tenant.id`) using the built-in generator.

This adds an optional `resource_attributes` field to the traffic generator config. It accepts three forms:

  **Single map** (all batches carry the same attributes):
  ```yaml
  resource_attributes:
    tenant.id: prod
    service.namespace: frontend
```

  **List of maps** (equal round-robin rotation per batch):
```yaml
  resource_attributes:
    - {tenant.id: prod, service.namespace: frontend}
    - {tenant.id: ppe, service.namespace: backend}
```

 **Weighted list** (proportional batch split):
``` yaml
resource_attributes:
    - attrs: {tenant.id: prod, service.namespace: frontend}
      weight: 3
    - attrs: {tenant.id: ppe, service.namespace: backend}
      weight: 1
```

  The weighted form produces a 75%/25% batch split, simulating realistic skewed  multi-tenant traffic on a single connection. All three forms are backward-compatible - existing configs are unaffected (defaults to empty).

 ### Implementation

  Rotation uses a precomputed index table built once at startup - [0, 0, 0, 1] for the 3:1 example above. The hot path is a single modulo lookup:

  slot  = rotation[batch_rotation_index % rotation.len()]
  attrs = &entries[slot].attrs

  `batch_rotation_index` is a dedicated counter incremented once per emitted batch, fully independent of `signal_count`. Rotation advances exactly once per OTLP message regardless of batch size.

 
 ### Limitations

  - `resource_attributes` only applies to `data_source: static`
  - With `generation_strategy: pre_generated`, only the first attribute set is used - rotation requires fresh (or templates)
  - Rotation order is naive ([0, 0, 0, 1] for 3:1), not smooth interleaved; smooth weighted round-robin is left as a follow-up


lalitb:
> I wonder if it might be necessary to add a concept of weight when you have a list of resource attributes. That would allow configuring a large tenant versus several smaller ones, for example.


+1 I had the same in mind, let me add that too in this PR.

lalitb:
> I wonder if it might be necessary to add a concept of weight when you have a list of resource attributes

 Done - weighted entries are already included in this PR. The weight field on each `resource_attributes` entry controls the batch share, so a large tenant vs. several smaller ones is directly expressible:                                                                                                                                                                                  
 
```yaml                                                                                                                                                                                                                                                                                                                                                                                         
  resource_attributes:                                                                                                                                                                                                                                                                                                                                                                               
    - attrs: {tenant.id: large-tenant}
      weight: 8                                                                                                                                                                                                                                                                                                                                                                                      
    - attrs: {tenant.id: small-tenant-a}                               
      weight: 1
    - attrs: {tenant.id: small-tenant-b}
      weight: 1
```

Updated docs and PR desc accordingly.

rust/otap-dataflow/configs/README.md
```diff
@@ -68,6 +68,26 @@ Generates fake data with performance metrics:
...
+
+- Uses `data_source: static` with two resource attribute sets (`tenant.id: prod` and
+  `tenant.id: ppe`) weighted 3:1, producing a 75% / 25% batch split per pipeline.
+- Generates fake data -> performance exporter
+- View metrics at: `http://127.0.0.1:8080/telemetry/metrics?format=prometheus&reset=false`
+
+The `resource_attributes` field accepts three forms:
+
+| Form | Description |
+|------|-------------|
```
albertlockett: markdown lint fix:
```suggestion
| ---- | ----------- |
```

rust/otap-dataflow/configs/README.md
```diff
@@ -68,6 +68,26 @@ Generates fake data with performance metrics:
 - Generates fake data -> performance exporter
 - View metrics at: `http://127.0.0.1:8080/telemetry/metrics?format=prometheus&reset=false`
 
+### `fake-multi-tenant-perf.yaml`
+
+Generates mixed-tenant traffic using weighted resource attribute rotation:
+
+- Uses `data_source: static` with two resource attribute sets (`tenant.id: prod` and
+  `tenant.id: ppe`) weighted 3:1, producing a 75% / 25% batch split per pipeline.
```
albertlockett: markdown lint fix:
```suggestion
- Uses `data_source: static` with two resource attribute sets (`tenant.id: 
  prod` and `tenant.id: ppe`) weighted 3:1, producing a 75% / 25% batch split
  per  pipeline.
```

lquerel:
@lalitb There is a merge issue. I can't merge this PR.

lalitb:
> @lalitb There is a merge issue. I can't merge this PR.

@lquerel  - This should be good to merge now.

PR #2267, opened 2026-03-11
example extension for comparison based on utpilla's design

gouslu:
# Change Summary

* Closes #NNN

### DIFF
--- a/rust/otap-dataflow/crates/contrib-nodes/src/extensions/azure_identity_auth_extension/extension.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/src/extensions/azure_identity_auth_extension/extension.rs
@@ -1,0 +1,329 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Azure Identity Auth Extension implementation.
+//!
+//! This extension provides Azure authentication services to the pipeline.
+//! It manages Azure credentials and provides token acquisition capabilities
+//! to consumers (e.g., exporters) via the [`BearerTokenProviderHandle`].
+//!
+//! # Architecture
+//!
+//! The extension creates a [`BearerTokenProviderHandle`] at construction time,
+//! which wraps an `AzureTokenProvider` implementing the [`BearerTokenProvider`]
+//! trait. The extension background task refreshes tokens periodically and
+//! broadcasts updates through a `watch` channel that subscribers receive via
+//! the handle.
+//!
+//! Consumers retrieve the handle from the extension registry:
+//!
+//! ```rust,ignore
+//! let handle = extension_registry
+//!     .get::<BearerTokenProviderHandle>("azure_auth")?;
+//! let token = handle.get_token().await?;
+//! ```
+
+use async_trait::async_trait;
+use azure_core::credentials::{AccessToken, TokenCredential};
+use azure_identity::{
+    DeveloperToolsCredential, DeveloperToolsCredentialOptions, ManagedIdentityCredential,
+    ManagedIdentityCredentialOptions, UserAssignedId,
+};
+use otap_df_engine::extensions::{
+    BearerToken, BearerTokenError, BearerTokenProvider,
+};
+use std::sync::{Arc, Mutex};
+use tokio::sync::watch;
+
+use otap_df_engine::error::Error as EngineError;
+use otap_df_engine::local::extension as local;
+
+use super::config::{AuthMethod, Config};
+use super::error::Error;
+
+/// Buffer time before token expiry to trigger refresh (in seconds).
+const TOKEN_EXPIRY_BUFFER_SECS: u64 = 299;
+/// Minimum interval between token refresh attempts (in seconds).
+const MIN_TOKEN_REFRESH_INTERVAL_SECS: u64 = 10;
+/// Minimum delay between token refresh retry attempts in seconds.
+const MIN_RETRY_DELAY_SECS: f64 = 5.0;
+/// Maximum delay between token refresh retry attempts in seconds.
+const MAX_RETRY_DELAY_SECS: f64 = 30.0;
+/// Maximum jitter percentage (±10%) to add to retry delays.
+const MAX_RETRY_JITTER_RATIO: f64 = 0.10;
+/// Retry interval when token refresh fails (in seconds).
+const TOKEN_REFRESH_RETRY_SECS: u64 = 10;
+
+/// The token provider implementation that backs the [`BearerTokenProviderHandle`].
+///
+/// Reads the latest token from a shared cache (updated by the extension's
+/// background refresh loop) and exposes a watch channel for subscribers.
+/// If the cache is empty or expired, falls back to fetching from Azure directly.
+pub(crate) struct AzureTokenProvider {
+    /// The Azure credential provider.
+    credential: Arc<dyn TokenCredential>,
+    /// The OAuth scope for token acquisition.
+    scope: String,
+    /// Sender for broadcasting token refresh events (used by `subscribe_token_refresh`).
+    token_sender: Arc<watch::Sender<Option<BearerToken>>>,
+    /// Cached token updated by the extension's refresh loop (used by `get_token`).
+    token_cache_for_demonstration: Arc<Mutex<Option<BearerToken>>>,
+}
+
+#[async_trait]
+impl BearerTokenProvider for AzureTokenProvider {
+    async fn get_token(&self) -> Result<BearerToken, BearerTokenError> {
+        // Fast path: return cached token if it hasn't expired.
+        {
+            let cache = self.token_cache_for_demonstration.lock().expect("token_cache lock poisoned");
+            if let Some(cached) = cache.as_ref() {
+                let now_secs = std::time::SystemTime::now()
+                    .duration_since(std::time::UNIX_EPOCH)
+                    .map(|d| d.as_secs() as i64)
+                    .unwrap_or(0);
+                if cached.expires_on > now_secs {
+                    return Ok(cached.clone());
+                }
+            }
+        }
+
+        // Slow path: cache is empty or expired — fetch from Azure.
+        let access_token = self
+            .credential
+            .get_token(
+                &[&self.scope],
+                Some(azure_core::credentials::TokenRequestOptions::default()),
+            )
+            .await
+            .map_err(|e| BearerTokenError {
+                message: e.to_string(),
+            })?;
+
+        let token = BearerToken::new(
+            access_token.token.secret().to_string(),
+            access_token.expires_on.unix_timestamp(),
+        );
+
+        // Update cache so next call hits the fast path.
+        *self.token_cache_for_demonstration.lock().expect("token_cache lock poisoned") = Some(token.clone());
+
+        Ok(token)
+    }
+
+    fn subscribe_token_refresh(&self) -> watch::Receiver<Option<BearerToken>> {
+        self.token_sender.subscribe()
+    }
+}
+
+/// Azure Identity Auth Extension.
+///
+/// Runs as a background task that periodically refreshes the Azure bearer
+/// token and broadcasts updates through the token provider handle.
+pub struct AzureIdentityAuthExtension {
+    /// The Azure credential provider (shared with the token provider).
+    credential: Arc<dyn TokenCredential>,
+    /// Human-readable description of the credential type.
+    credential_type: &'static str,
+    /// The OAuth scope for token acquisition.
+    scope: String,
+    /// Sender for broadcasting token refresh events.
+    token_sender: Arc<watch::Sender<Option<BearerToken>>>,
+    /// Shared cache updated by the refresh loop, read by the provider handle.
+    token_cache_for_demonstration: Arc<Mutex<Option<BearerToken>>>,
+}
+
+impl AzureIdentityAuthExtension {
+    /// Creates a new Azure Identity Auth Extension and its associated token provider.
+    ///
+    /// Returns both the extension (for the background task) and the provider
+    /// (to be wrapped in a `BearerTokenProviderHandle` and registered).
+    pub(crate) fn new(config: &Config) -> Result<(Self, AzureTokenProvider), Error> {
+        let (credential, credential_type) = Self::create_credential(config)?;
+        let (token_sender, _) = watch::channel(None);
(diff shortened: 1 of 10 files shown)


rust/otap-dataflow/crates/engine/src/extensions/bearer_token.rs
```diff
@@ -0,0 +1,316 @@
...
+            inner: Arc::new(tokio::sync::Mutex::new(Box::new(provider))),
+        }
+    }
+
+    /// Returns an authentication token.
+    ///
+    /// Acquires the internal lock, then delegates to the underlying
+    /// [`BearerTokenProvider`] implementation.
+    pub async fn get_token(&self) -> Result<BearerToken, BearerTokenError> {
+        self.inner.lock().await.get_token().await
```
gouslu, line 210: We need async methods, but that forces the handle's mutex to also be async `tokio::Mutex::lock().await`. This serializes all callers: if component A is mid-way through an async call (e.g., an HTTP request) while holding the lock, component B must wait for A to fully complete before it can even start. Without the mutex, both calls would be concurrent — their I/O would overlap. In this extension concurrent calls may not matter, but any extension whose handle wraps an async I/O method will hit this limitation.

gouslu: Also, I think not having async methods is a deal-breaker.

rust/otap-dataflow/crates/contrib-nodes/src/extensions/azure_identity_auth_extension/extension.rs
```diff
@@ -0,0 +1,329 @@
...
+const MAX_RETRY_JITTER_RATIO: f64 = 0.10;
+/// Retry interval when token refresh fails (in seconds).
+const TOKEN_REFRESH_RETRY_SECS: u64 = 10;
+
+/// The token provider implementation that backs the [`BearerTokenProviderHandle`].
+///
+/// Reads the latest token from a shared cache (updated by the extension's
+/// background refresh loop) and exposes a watch channel for subscribers.
+/// If the cache is empty or expired, falls back to fetching from Azure directly.
+pub(crate) struct AzureTokenProvider {
```
gouslu, line 62: This struct is in addition to extensions internal state on line 122, needed for state sharing, because lifecycle and extension are 2 different instances/structs. This causes redundancy.

rust/otap-dataflow/crates/contrib-nodes/src/extensions/azure_identity_auth_extension/extension.rs
```diff
@@ -0,0 +1,329 @@
...
+    ///
+    /// Returns both the extension (for the background task) and the provider
+    /// (to be wrapped in a `BearerTokenProviderHandle` and registered).
+    pub(crate) fn new(config: &Config) -> Result<(Self, AzureTokenProvider), Error> {
+        let (credential, credential_type) = Self::create_credential(config)?;
+        let (token_sender, _) = watch::channel(None);
+        let token_sender = Arc::new(token_sender);
+        let token_cache_for_demonstration = Arc::new(Mutex::new(None));
+
+        let provider = AzureTokenProvider {
```
gouslu, line 146: The compiler will error if you move the same value into both structs, but it won't stop you from creating independent instances with separate state. The author must understand that `Arc::clone` is needed to share state — the type system doesn't enforce it. So, the comment about handle based design not being error-prone in that regard is not entirely correct. It is more explicit, I will accept it, but it is still error-prone.

rust/otap-dataflow/crates/engine/src/extensions/auth.rs
```diff
@@ -0,0 +1,505 @@
...
+#[derive(Clone)]
+pub struct ServerAuthenticatorHandle {
+    inner: Arc<Mutex<Box<dyn ServerAuthenticator>>>,
+}
+
+impl ServerAuthenticatorHandle {
+    /// Creates a new handle wrapping the given authenticator implementation.
+    pub fn new(auth: impl ServerAuthenticator + 'static) -> Self {
+        Self {
+            inner: Arc::new(Mutex::new(Box::new(auth))),
```
gouslu, line 104: The `tokio::sync::Mutex` on the handle creates an additional layer of indirection. Benchmarks show ~6ns per call (lock + unlock) compared to <1ns for a direct vtable call — roughly 6x overhead. This is small in absolute terms, but worth noting if an extension method is called in a hot path (e.g., per-request auth header injection).

utpilla:
I think all the concerns in this PR trace back to one design question: **should handle methods be async?**

My position is no. The way I'm modeling it, handle is a read endpoint. Its only job is to return cached state that the extension's background loop already produced. All async work (token acquisition, retries, refresh scheduling) belongs in `start()`. Consumers never do I/O through the handle; they read what's already there.

This isn't a novel constraint — the Go Collector's auth extension interfaces follow the same principle. [`Server.Authenticate`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/extension/extensionauth/server.go), [`HTTPClient.RoundTripper`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/extension/extensionauth/client.go), and [`GRPCClient.PerRPCCredentials`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/extension/extensionauth/client.go) are all sync. The gRPC receiver calls [`Authenticate` synchronously inside its interceptor](https://github.com/open-telemetry/opentelemetry-collector/blob/85c4dac6438eecfff4e4447c6be2c2e5cf6bdf7a/config/configgrpc/configgrpc.go#L655), and the exporter calls [`PerRPCCredentials` synchronously during setup](https://github.com/open-telemetry/opentelemetry-collector/blob/85c4dac6438eecfff4e4447c6be2c2e5cf6bdf7a/config/configgrpc/configgrpc.go#L391). Background goroutines handle all the async I/O; the consumer-facing interfaces return cached state.

With that context, here's how the specific concerns in this PR play out:

**On tokio Mutex serializing the callers**

The `BearerTokenProvider` trait introduced here has `async fn get_token()`, which forces the handle to use `tokio::sync::Mutex` (`std::sync::Mutex` can't be held across `.await`). Yes, that serializes callers, if one is mid-way through an async operation, all others wait. However, **keeping handles sync avoids this entirely**.

**On `AzureTokenProvider` being "redundant":**

The `AzureTokenProvider` here has 4 fields (`credential`, `scope`, `token_sender`, `token_cache_for_demonstration`) and includes a fallback path that fetches tokens directly from Azure on a cache miss. That complexity comes from the async `get_token()` design choice — the handle needs enough state to perform I/O on demand.

In #2113, the equivalent struct is:

```rust
struct AzureIdentityClientAuth {
    token_rx: watch::Receiver<Option<HeaderValue>>,
}
```

One field. No credentials, no scope, no cache, no fallback fetching. It reads from the watch channel and returns the cached header. The "redundancy" was introduced by **making the handle async, not by the handle-based pattern itself**.

**On "same error-proneness" / Arc::clone:**

The `Arc::clone` usage here is needed because the extension and the provider both need shared access to `token_sender` and `token_cache_for_demonstration` — again, because the handle does its own I/O and needs the full credential state.

In #2113, there's no `Arc::clone` for state sharing. The factory calls `watch::channel()`, which produces two endpoints by construction — the sender moves into the extension, the receiver moves into the handle:

```rust
let (token_tx, token_rx) = watch::channel(None);
let auth = AzureIdentityClientAuth { token_rx };
let extension = AzureIdentityAuthExtension { token_tx, .. };
```

**Nothing to clone, nothing to get wrong.**

**On `tokio::Mutex` overhead (6ns vs <1ns):**

6ns per call is not a meaningful factor in a system where a single request involves socket I/O, deserialization, pipeline processing, and serialization — orders of magnitude more expensive. This shouldn't be a deciding factor for the architecture.

rust/otap-dataflow/crates/engine/src/extensions/bearer_token.rs
utpilla: > Also, I think not having async methods is a deal-breaker.

I agree that, with my approach in #2113, we don't have good support for async methods in the handles, which is covered well in your approach #2141. However, I don't know if we actually need async methods in the handle.

I like the clear separation of concerns, where the `Extensions::start` method is responsible for keeping the state updated (with async operations if required) but the handles consumed by the receivers, exporters etc. are only reading the updated state. I left a comment here with more details: https://github.com/open-telemetry/otel-arrow/pull/2267#issuecomment-4042002705

gouslu:
> I think all the concerns in this PR trace back to one design question: **should handle methods be async?**
> 
> My position is no. The way I'm modeling it, handle is a read endpoint. Its only job is to return cached state that the extension's background loop already produced. All async work (token acquisition, retries, refresh scheduling) belongs in `start()`. Consumers never do I/O through the handle; they read what's already there.
> 
> This isn't a novel constraint — the Go Collector's auth extension interfaces follow the same principle. [`Server.Authenticate`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/extension/extensionauth/server.go), [`HTTPClient.RoundTripper`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/extension/extensionauth/client.go), and [`GRPCClient.PerRPCCredentials`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/extension/extensionauth/client.go) are all sync. The gRPC receiver calls [`Authenticate` synchronously inside its interceptor](https://github.com/open-telemetry/opentelemetry-collector/blob/85c4dac6438eecfff4e4447c6be2c2e5cf6bdf7a/config/configgrpc/configgrpc.go#L655), and the exporter calls [`PerRPCCredentials` synchronously during setup](https://github.com/open-telemetry/opentelemetry-collector/blob/85c4dac6438eecfff4e4447c6be2c2e5cf6bdf7a/config/configgrpc/configgrpc.go#L391). Background goroutines handle all the async I/O; the consumer-facing interfaces return cached state.
> 
> With that context, here's how the specific concerns in this PR play out:
> 
> **On tokio Mutex serializing the callers**
> 
> The `BearerTokenProvider` trait introduced here has `async fn get_token()`, which forces the handle to use `tokio::sync::Mutex` (`std::sync::Mutex` can't be held across `.await`). Yes, that serializes callers, if one is mid-way through an async operation, all others wait. However, **keeping handles sync avoids this entirely**.
> 
> **On `AzureTokenProvider` being "redundant":**
> 
> The `AzureTokenProvider` here has 4 fields (`credential`, `scope`, `token_sender`, `token_cache_for_demonstration`) and includes a fallback path that fetches tokens directly from Azure on a cache miss. That complexity comes from the async `get_token()` design choice — the handle needs enough state to perform I/O on demand.
> 
> In #2113, the equivalent struct is:
> 
> ```rust
> struct AzureIdentityClientAuth {
>     token_rx: watch::Receiver<Option<HeaderValue>>,
> }
> ```
> 
> One field. No credentials, no scope, no cache, no fallback fetching. It reads from the watch channel and returns the cached header. The "redundancy" was introduced by **making the handle async, not by the handle-based pattern itself**.
> 
> **On "same error-proneness" / Arc::clone:**
> 
> The `Arc::clone` usage here is needed because the extension and the provider both need shared access to `token_sender` and `token_cache_for_demonstration` — again, because the handle does its own I/O and needs the full credential state.
> 
> In #2113, there's no `Arc::clone` for state sharing. The factory calls `watch::channel()`, which produces two endpoints by construction — the sender moves into the extension, the receiver moves into the handle:
> 
> ```rust
> let (token_tx, token_rx) = watch::channel(None);
> let auth = AzureIdentityClientAuth { token_rx };
> let extension = AzureIdentityAuthExtension { token_tx, .. };
> ```
> 
> **Nothing to clone, nothing to get wrong.**
> 
> **On `tokio::Mutex` overhead (6ns vs <1ns):**
> 
> 6ns per call is not a meaningful factor in a system where a single request involves socket I/O, deserialization, pipeline processing, and serialization — orders of magnitude more expensive. This shouldn't be a deciding factor for the architecture.

I was able to implement this and nothing stopped me from implementing my extension like I did in this PR. If you are going to put all these constraints and principles to authors to "not make mistakes", and make "good design decisions" then why not just put one simple constraint like I did in my design: "if you are going to share the state, put `Arc<>` around it. and be aware that your extensions will be cloned for every consumer". With this knowledge, in the design I propose, you have so much more freedom and simplicity.

On mutex cost, 6ns is best case. Under production load, this will likely be much higher, due to CPU cache pressure. 

Also, there is no real mechanism that protects anyone from doing anything multi-threaded, anyone could still run `tokio::task::spawn_blocking` and use your handle in it. A lot of the "mistakes" in the code has to be guarded by PR review; this is not something new and PRs are there for a reason. So, the one design flaw that you pointed out in the design I proposed can easily be managed by documentation and PR reviews. This is the same level of discipline we already apply to multi-threading, memory safety patterns, and API usage across the codebase.

If the argument is to say that, never share state between lifecycle and extension, isn't that a constraint you are putting on the developer? Same principal can be applied to the design proposal I made; however I wouldn't go that far. I would only say, `Arc<>` wrap your shared fields, and the rest is business as usual.

rust/otap-dataflow/crates/engine/src/extensions/bearer_token.rs
gouslu: I don't think that constraint is acceptable. What if you wanted to have a cache processor that uses different databases. Something like "KeyValueCacheProcessor" that uses "KeyValueCacheProvider", and you can implement it for "RedisKeyValueCacheProvider" and Amazon/Azure.. etc., you name it. Those providers can be extension capabilities and the cache processor doesn't have to know which tech it is using underneath. This would require async. I am just imagining scenarios, but things like that would be useful.

rust/otap-dataflow/crates/contrib-nodes/src/extensions/azure_identity_auth_extension/extension.rs
utpilla: Here's a way you can implement a functionally equivalent implementation of BearerToken extension with the handle approach without requiring the extension author to add any Mutex or redundant code:

https://github.com/utpilla/otel-arrow/pull/4

Check these files:
- `‎rust/otap-dataflow/crates/engine/src/extensions/bearer_token.rs` (Capability trait and handle definition)
- `‎rust/otap-dataflow/crates/contrib-nodes/src/extensions/azure_identity_auth/mod.rs` (Concrete implementation)

Issue #1742, opened 2026-01-08
Remove experimental-tls feature flag and make TLS support built-in by default


lalitb:
I propose removing the `experimental-tls` feature flag and making TLS/mTLS support a built-in, always-available capability of the collector. This would help establish TLS as a first-class, production-ready feature rather than an opt-in experimental one. The only concern is that this will increase the binary size as TLS dependencies (rustls, tokio-rustls, etc) will always be included, but security should be an inherent property of the collector.

utpilla:
This was completed in #2624.

PR #2269, opened 2026-03-11
Support pluggable rustls crypto providers (ring, aws-lc-rs, OpenSSL) …

### LINKED ISSUE #2251: Support pluggable rustls crypto providers (ring, aws-lc-rs, OpenSSL) via feature flags
### Summary
   
   TLS support currently hardcodes `ring` as the rustls crypto provider. Some deployment environments require the use of platform-approved or FIPS-validated cryptographic libraries (e.g., OpenSSL). We should introduces feature flags that allow users to select their preferred rustls `CryptoProvider` at compile time.
   
   ### Motivation
   
   - **Compliance:** Certain regulated environments do not permit the use of `ring` and require OpenSSL or other approved crypto libraries.
   - **Flexibility:** Different platforms and deployment targets have different crypto backend preferences. Users should be able to choose without forking the project.
   - **FIPS support:** The `rustls-openssl` crate offers a `fips` feature for environments that require FIPS 140 validated cryptography.
   
   ### Proposed Changes
   
   Introduce three mutually exclusive feature flags for selecting the rustls crypto backend:
   
   | Feature | Backend | Use Case |
   |---|---|---|
   | `crypto-ring` | `ring` | Default, backward-compatible |
   | `crypto-aws-lc` | `aws-lc-rs` | AWS environments, broader algorithm support |
   | `crypto-openssl` | `rustls-openssl` | Regulated environments, FIPS compliance |
   
   ### Scope
   
   - Add `rustls-openssl` as an optional workspace dependency
   - Remove hardcoded `ring` feature from `rustls` and `tokio-rustls` workspace deps
   - Replace the one direct `ring::sign` API call in `tls_utils.rs` with the provider-agnostic `CryptoProvider::get_default()` API
   - Add `install_crypto_provider()` helper to centralize provider initialization
   - Deduplicate crypto provider init in main.rs by delegating to install_crypto_provider() instead of inlining the same #[cfg] logic.
   - Enable tonic/tls-native-roots for crypto-openssl to provide client-side gRPC TLS without pulling in ring
   - Update test infrastructure to use the selected provider
   - Switch `reqwest` from `rustls` feature to `rustls-no-provider` in `otap` and `contrib-nodes` crates (see below)
   
   ### Direct `reqwest` usage
   
   The `otap` and `contrib-nodes` crates depend directly on `reqwest 0.13` with the `rustls` feature, which hardcodes `aws-lc-rs` as the crypto backend. This needs to change:
   
   - Switch from `reqwest` feature `rustls` -> `rustls-no-provider` so that reqwest uses whichever `CryptoProvider` has been globally installed (ring, aws-lc-rs, or OpenSSL depending on the selected `crypto-*`    feature flag).
   - No changes are needed in `reqwest` itself - it already supports the `rustls-no-provider` feature for this use case.
   
   ### What does NOT change
   
   - All existing TLS functionality (certificate loading, hot-reload, mTLS, etc.)
   - The public API surface
   
   ### Transitive `ring` dependencies
   
   After this change, `ring` is fully eliminated from the direct dependency chain when using `crypto-openssl`. However, `ring` still appears transitively through upstream crates:
   - **`opentelemetry-otlp` (0.31.0):** Its `tls` feature hardcodes `tonic/tls-ring`, and it depends on `reqwest ^0.12` which activates `rustls-tls-webpki-roots` -> `ring`. The `opentelemetry-rust` main branch 
  has partial fixes (`tls-ring`/`tls-aws-lc` feature split, `reqwest` updated to 0.13). However, to fully support OpenSSL as a crypto backend, additional work is needed: `opentelemetry-otlp` should offer a provider-agnostic TLS option (similar to reqwest's `rustls-no-provider`) so that consumers using `rustls-openssl` can install their provider globally without the exporter forcing a specific backend.
Note: tonic 0.14 already supports provider-agnostic TLS via its `tls-native-roots` feature, which enables TLS plumbing and OS certificate store without pulling in `ring` or `aws-lc-rs`. The `crypto-openssl` feature flag uses this to provide full client-side gRPC TLS with OpenSSL today.
   - **`weaver` (pinned at v0.17.0):** Pulls in `ring` through three transitive paths: `gix` (via `gix-transport` -> `reqwest 0.12`), `ureq` (which depends on `rustls` with `ring`), and `jsonschema` (via `reqwest 0.12`). To fully eliminate `ring`, weaver would need to update these dependencies to versions that either use `reqwest 0.13` with `rustls-no-provider` or `native-tls`, rather than hardcoding `ring` as the crypto backend. The fix on this repo's side is to bump the weaver version once a newer release is available with these updates
   Note: weaver is only used to power the fake data generator - a dev/testing/benchmarking tool for generating synthetic telemetry data. It is not a production pipeline component and does not process customer or internal data.
   
   These will be tracked and resolved as the upstream crates release updates.
   
   ### References
   
   - [`rustls-openssl`](https://github.com/tofay/rustls-openssl) - OpenSSL-backed `CryptoProvider` for rustls
   - [`rustls` CryptoProvider docs](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html)
   - [open-telemetry/opentelemetry-rust#2704](https://github.com/open-telemetry/opentelemetry-rust/issues/2704) — Related upstream issue for pluggable TLS in opentelemetry-otlp

pritishnahar95:
# Change Summary

Support pluggable rustls crypto providers (ring, aws-lc-rs, OpenSSL) via
compile-time feature flags. TLS support previously hardcoded `ring` as the
rustls crypto backend. This PR introduces three mutually exclusive feature
flags (`crypto-ring`, `crypto-aws-lc`, `crypto-openssl`) so users can select
their preferred `CryptoProvider` at build time, enabling compliance with
environments that require OpenSSL or FIPS-validated cryptography.

* closes #2251

- `cargo check` passes with default features (`crypto-ring`).
- `cargo check --no-default-features --features jemalloc,crypto-openssl,experimental-tls` passes.
- All existing TLS tests (`tls_utils`, `mtls_tests`, `tls_stream`, `tls_reload`,
  `otlp_exporter_tls`, `otlp_exporter_proxy_tls`) now use the centralized
  `install_crypto_provider()` helper and will exercise whichever backend is
  selected by feature flags.
- `compile_error!` guards prevent enabling multiple crypto features simultaneously.

Yes:

- **New feature flags** on the root crate and `otap-df-otap`:
  - `crypto-ring` (default) — uses `ring`, backward-compatible.
  - `crypto-aws-lc` — uses `aws-lc-rs` for AWS environments.
  - `crypto-openssl` — uses `rustls-openssl` for regulated/FIPS environments.
- **Default behavior is unchanged** — `crypto-ring` is included in the default
  feature set, so existing builds are unaffected.
- **To build with OpenSSL**: `cargo build --no-default-features --features jemalloc,crypto-openssl`
- **Transitive `ring` dependencies** from `opentelemetry-otlp` (via `reqwest 0.12`)
  and `weaver` (via `ureq`) remain and are tracked for resolution as upstream
  crates release updates. Weaver is dev/test tooling only, not a production
  pipeline component.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/crypto.rs
+++ b/rust/otap-dataflow/crates/otap/src/crypto.rs
@@ -51,6 +51,19 @@ pub fn install_crypto_provider() -> Result<(), String> {
             .map_err(|_| "crypto provider already installed (openssl)".to_string())?;
     }
 
+    #[cfg(not(any(
+        feature = "crypto-ring",
+        feature = "crypto-aws-lc",
+        feature = "crypto-openssl"
+    )))]
+    {
+        otap_df_telemetry::otel_warn!(
+            "crypto.no_provider",
+            message = "no crypto-* feature enabled: TLS operations will fail at runtime. \
+                       Enable exactly one of: crypto-ring, crypto-aws-lc, crypto-openssl"
+        );
+    }
+
     Ok(())
 }
 


rust/otap-dataflow/crates/contrib-nodes/src/lib.rs
```diff
@@ -8,3 +8,15 @@ pub mod exporters;
...
+/// Test support: installs the rustls crypto provider once for the entire test binary.
+/// This is needed because `reqwest` with `rustls-no-provider` requires a globally
+/// installed `CryptoProvider` before creating any HTTP client.
+#[cfg(test)]
+pub(crate) fn ensure_crypto_provider() {
+    use std::sync::Once;
+    static INIT: Once = Once::new();
+    INIT.call_once(|| {
+        let _ = otap_df_otap::crypto::install_crypto_provider();
+    });
```
drewrelmas: Not sure this is the right place for this helper function - I understand you want it to be available to tests in `contrib-nodes` but I think it should be possible to export this as part of the `otap` crate `test-utils` feature.

See https://github.com/open-telemetry/otel-arrow/blob/ee7dcbab5c0dec5d5cb234062849e4c747b940c9/rust/otap-dataflow/crates/contrib-nodes/Cargo.toml#L88-L89

rust/otap-dataflow/crates/otap/Cargo.toml
```diff
@@ -97,7 +98,10 @@ byte-unit.workspace = true
 
 [features]
 test-utils = []
-experimental-tls = ["dep:rustls", "dep:rustls-pki-types", "dep:tokio-rustls", "dep:rustls-native-certs", "tonic/tls-ring", "dep:arc-swap", "dep:notify"]
+experimental-tls = ["dep:rustls", "dep:rustls-pki-types", "dep:tokio-rustls", "dep:rustls-native-certs", "dep:arc-swap", "dep:notify", "tonic/tls-native-roots"]
```
drewrelmas, line 101: Question for the group - should TLS even be behind a feature flag? Should it be core part of the engine? At very least, maybe I would remove `experimental` from the feature name.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
```diff
@@ -222,6 +222,7 @@ mod tests {
     // ==================== Test Helpers ====================
 
     fn create_test_client() -> Client {
+        crate::ensure_crypto_provider();
```
drewrelmas: It feels a little weird to me to need test authors to always call this func - what is the behavior if they forget? Is the test failure obvious?

rust/otap-dataflow/Cargo.toml
```diff
@@ -206,7 +206,7 @@ wiremock = "0.6.5"
 
 [features]
 # ToDo When jemalloc is enabled, we could use jemalloc_pprof to profile where memory is allocated. See https://crates.io/crates/jemalloc_pprof
-default = ["jemalloc"]
+default = ["jemalloc", "crypto-ring"]
```
drewrelmas, line 210: Since `crypto-ring` is the default feature for crypto and the features are mutually exclusive, we're not actually exercising code paths for other features in CI, correct? I wonder if we need isolated test runs using matrix of TLS feature...

I imagine the Rust SDK folks among us have run into this problem and have recommended pattern.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
pritishnahar95: Yes, the failure is a panic with 'No provider set' from reqwest internals — not immediately obvious. I've placed the call in shared test helpers (create_test_http_client, setup_exporter, etc.) so most tests get it automatically. Only tests that create HTTP clients through unusual paths need an explicit call. New to Rust, so could be wrong, but Rust has no built-in "run before all tests" hook. Open to suggestions if there's a better pattern :)

rust/otap-dataflow/crates/otap/Cargo.toml
pritishnahar95: I like the idea of removing `experimental` from feature name(although I would recommend doing that in a separate PR). @lalitb what do you think?

lalitb: The TLS should be enabled by default - without any feature flag. There is a tracking issue for this - https://github.com/open-telemetry/otel-arrow/issues/1742.  I want still be sure of some scenario before this be removed.

rust/otap-dataflow/Cargo.toml
pritishnahar95: Correct — CI currently only exercises crypto-ring (default) and all-features (ring wins). The code divergence between providers is a single function call in install_crypto_provider(), so the risk of untested paths is low. The TLS plumbing, certificate loading, mTLS, hot-reload etc. are all identical regardless of provider. What changes is which cryptographic primitives run under the hood (inside rustls), which is tested by the upstream crates themselves.
That being said, adding a crypto feature matrix is a good idea for completeness but could be a follow-up?

rust/otap-dataflow/crates/contrib-nodes/src/lib.rs
pritishnahar95: Good point, I have pushed a new commit to address this.

rust/otap-dataflow/crates/otap/Cargo.toml
pritishnahar95: In that case, let's talk more about this in the linked issue, since that is outside the scope of this PR.

drewrelmas: Agreed

rust/otap-dataflow/Cargo.toml
drewrelmas: I think given how close each implementation is it is acceptable to only test one path at the moment.

.github/workflows/rust-ci.yml
```diff
@@ -464,6 +471,13 @@ jobs:
         with:
           toolchain: stable
           components: clippy
+      - name: Install OpenSSL (Windows)
```
drewrelmas, line 474: Minor note - if the openssl feature isn't being actually executed in our test suite, do we need to include this in our workflows?

pritishnahar95: The Install OpenSSL (Windows) step is needed because --all-features enables crypto-openssl, which pulls in openssl-sys at compile time. Without OpenSSL headers present, the build fails before any tests run.

We could remove these steps and instead exclude crypto-openssl from the Windows --all-features runs, but we need to use an explicit feature list. The tradeoff is that the explicit list needs manual updates when new features are added, but it avoids vcpkg build time and the OpenSSL install dependency on Windows.

lalitb: The CI time for the windows clippy and test build has now increased by 6mins, which is substantial - We may want to cache the vcpkg install step, or avoid using `--all-features` for clippy/test on Windows. Not blocking for this PR, but we should definitely revisit this.

rust/otap-dataflow/crates/otap/src/crypto.rs
```diff
@@ -0,0 +1,68 @@
...
+        not(feature = "crypto-ring"),
+        not(feature = "crypto-aws-lc")
+    ))]
+    {
+        rustls_openssl::default_provider()
+            .install_default()
+            .map_err(|_| "crypto provider already installed (openssl)".to_string())?;
+    }
+
+    Ok(())
```
lalitb, line 67: Nit: When none of the crypto-* features are enabled, this silently returns Ok(()) without installing any provider. Any later TLS usage (reqwest, tonic, etc.) will fail with a potentially confusing error. Consider logging a `tracing::warn!` here when no crypto feature is active, or returning an Err so callers get an early, clear signal.

.github/workflows/rust-ci.yml
pritishnahar95: I have noticed this too, we should definitely revisit this.

rust/otap-dataflow/crates/otap/src/crypto.rs
pritishnahar95: Good point, addressed.

rust/otap-dataflow/crates/otap/src/crypto.rs
```diff
@@ -0,0 +1,80 @@
...
+            .map_err(|_| "crypto provider already installed (openssl)".to_string())?;
+    }
+
+    #[cfg(not(any(
+        feature = "crypto-ring",
+        feature = "crypto-aws-lc",
+        feature = "crypto-openssl"
+    )))]
+    {
+        tracing::warn!(
```
drewrelmas: @pritishnahar95 please use the Telemetry macros found in other components instead of `tracing` crate.

pritishnahar95: done, amended commit.

lquerel:
@pritishnahar95 @jmacd 

> Transitive ring dependencies from opentelemetry-otlp (via reqwest 0.12)
and weaver (via ureq) remain and are tracked for resolution as upstream
crates release updates. Weaver is dev/test tooling only, not a production
pipeline component.

I'm wondering whether we actually have a real problem here. In the medium term, I expected that we would no longer have `opentelemetry-otlp` as a dependency once the refactoring of the internal telemetry pipeline is completed. OTLP and OTAP export would be handled by our exporters, which would rely on the encryption layer defined by these new feature flags.

Regarding Weaver, we could imagine introducing a feature that enables dev-only capabilities (disabled by default). If we do that, it seems to me that nothing would remain that is not configured through the feature flags introduced in this PR.

rust/otap-dataflow/crates/otap/src/crypto.rs
```diff
@@ -0,0 +1,81 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Cryptographic provider initialization for rustls.
```
lquerel, line 4: Shouldn’t we move this code into a dedicated crate that could be used by both `contrib-nodes` and `core-nodes` (currently named `otap`)?

lquerel (approved):
LGTM

pritishnahar95:
> @pritishnahar95 @jmacd
> 
> > Transitive ring dependencies from opentelemetry-otlp (via reqwest 0.12)
> > and weaver (via ureq) remain and are tracked for resolution as upstream
> > crates release updates. Weaver is dev/test tooling only, not a production
> > pipeline component.
> 
> I'm wondering whether we actually have a real problem here. In the medium term, I expected that we would no longer have `opentelemetry-otlp` as a dependency once the refactoring of the internal telemetry pipeline is completed. OTLP and OTAP export would be handled by our exporters, which would rely on the encryption layer defined by these new feature flags.
> 
> Regarding Weaver, we could imagine introducing a feature that enables dev-only capabilities (disabled by default). If we do that, it seems to me that nothing would remain that is not configured through the feature flags introduced in this PR.

That's a good point @lquerel! Let's tackle the weaver feature-gating in a separate issue/PR since it seems to touch the fake_data_generator module, its [linkme](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/072586267e/resources/app/out/vs/code/electron-browser/workbench/workbench.html) registrations, and test/validation configs — better as a focused follow-up.

rust/otap-dataflow/crates/otap/src/crypto.rs
drewrelmas: @lquerel yes, I think we should find proper place for it, I plan to incorporate this later with all the 'shared' content left in `otap` at the end of #2086.

PR #2282, opened 2026-03-12
fix: Ensure root signal payload is first in BatchArrowRecords

cijothomas:
Perf test involving collector has been failing. This is to address that.


cijothomas:
@JakeDern Based on slack discussion, this is not what we want right?

JakeDern:
> @JakeDern Based on slack discussion, this is not what we want right?

Correct, I think we should remove the assumption from the go code instead

cijothomas:
Closing https://github.com/open-telemetry/otel-arrow/pull/2282#issuecomment-4043274713

PR #2290, opened 2026-03-12
Columnar query engine condition and assign operator support in nested attributes pipelines

albertlockett:
# Change Summary

In #2190 we added the ability to execute an OPL "pipeline" on a stream of attributes. This PR extends the capability to implement the `set` and `if/else` syntax to operate on attributes.

It allows us to write expressions that can optionally modify attribute, for example to redact based on sensitive attribtue keys or values:
```js
logs | apply attributes {
  if (key == "api-key" or key == "secret" or value = "4519 0123 4567 8901") {
    set value = "<redacted>"
  }
}
```

(After https://github.com/open-telemetry/otel-arrow/pull/2273 merges, we'll also be able to use regex match in the "if" condition).

The expression on the left side of the `set` does not have to be a static value. This uses the expression evaluation code that was added in https://github.com/open-telemetry/otel-arrow/pull/2126. This paves the way for more interesting types of attribute value updates as our expression evaluation becomes more mature.

The rules for setting attribute values are currently a bit restrictive:
1. If the expression used to compute the value includes the `value` column (a virtual column representing attribute value), then all the attributes must be the same type. In the future, I'll add some capability to ensure that we can check this in the if statement
2. The attribute "key" column, as well as the other attribute columns (type, parent_id, int, float, etc) cannot be used in the `set` expressions at this time. This helps ensure we don't create invalid batches.

* Closes https://github.com/open-telemetry/otel-arrow/issues/2034

Unit tests

These types of expressions would now be available for users in the transform processor.

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs
@@ -219,6 +219,11 @@ impl PipelineStage for ConditionalPipelineStage {
         task_context: Arc<TaskContext>,
         exec_options: &mut ExecutionState,
     ) -> Result<RecordBatch> {
+        if attrs_record_batch.num_rows() == 0 {
+            // no branches would handle any rows, so nothing to do
+            return Ok(attrs_record_batch);
+        }
+
         // keep track of the rows that were selected by previous branches
         let mut already_selected_vec = BooleanArray::new(
             BooleanBuffer::new_unset(attrs_record_batch.num_rows()),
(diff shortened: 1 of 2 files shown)


rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs
```diff
@@ -210,6 +210,111 @@ impl PipelineStage for ConditionalPipelineStage {
...
+                        )
+                        .await?;
+                }
+            }
+            branch_results.push(default_branch_batch);
+        }
+
+        // reconstruct the result by concatenating the record batches from all branches
+        let batch_refs = branch_results.iter().collect::<Vec<_>>();
+        let final_result = concat_batches(branch_results[0].schema_ref(), batch_refs)?;
```
lquerel, line 315: Is it possible in a scenario where everything is filtered out that we end up with an empty branch_results vec? If yes then this code could panic.

albertlockett: oh yeah you're right -- if a previous batch filtered out all the rows, we could hit this. Fixed in 5da0dc3e5

Issue #2283, opened 2026-03-12
Azure Monitor Exporter: auth token acquisition blocks the tokio::select! event loop


cijothomas:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Bug Description

auth.get_token() in the exporter's main tokio::select! loop is an infinite-retry async call that blocks the entire loop until it succeeds. This means:

While token acquisition is retrying (e.g., IMDS unreachable, expired credentials), no other select branches can run — message processing, in-flight export completions, heartbeats, periodic flushes, and shutdown handling are all blocked.

The fix would be to make token acquisition non-blocking — run it as a future (similar to in_flight_exports) so the select loop continues processing other events while auth retries in the background. Exports should be paused (or queued) until a valid token is available.

### Steps to Reproduce

See desc

### Expected Behavior

See desc

### Actual Behavior

See desc

### OTel-Arrow Version

aa

### Environment

aa

### Configuration

```yaml

```

### Log Output

```shell

```

gouslu:
Assign it to me, there may be some regression on this once we change to extension model, I want to track it and address it together with it.

PR #2312, opened 2026-03-13
azure monitor exporter: move auth to event loop only - non-blocking auth.

gouslu:
# Change Summary

* Closes https://github.com/open-telemetry/otel-arrow/issues/2283

Local maual tests and unit tests.

No.


rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
```diff
@@ -611,8 +602,8 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
                     }
                 }
 
-                // Control always flows; pdata guarded by !at_capacity
-                msg = msg_chan.recv_when(!at_capacity) => {
+                // Control always flows; pdata guarded by has_token && !at_capacity
+                msg = msg_chan.recv_when(accepting_pdata) => {
```
lalitb, line 618: There's a shutdown issue when auth is failing. The engine only sends Shutdown on the control channel to receivers; processors and exporters detect shutdown by noticing their pdata channel closed. But `recv_when(false)` only checks for pdata channel closure when the buffer is empty. If the upstream processor sends a few messages before shutting down, those messages sit in the buffer, `is_empty()` returns false, and the closure check is skipped. Since no Shutdown arrives on control either, the exporter can hang forever retrying auth with buffered messages stuck in the channel.
     
  Those buffered messages are not lost immediately, but if a later forced shutdown tears the channel down, they are silently lost because the exporter never read them into its internal tracking state.
     
  Before this PR, `accepting_pdata` was only false briefly when the exporter was at capacity. Now it can stay false for a long time during auth outages, which makes this much easier to hit.
     
  **Normal scenario:**
   Engine -> Receiver Shutdown -> receiver exits -> pdata closes -> processor exits -> pdata closes -> exporter detects closure -> exits.
  
**With auth failing:**
  
   processor sends [A, B], then exits -> pdata is closed but buffer is not empty -> recv_when(false) sees is_empty() == false and skips the closure check -> exporter waits on control -> no Shutdown arrives on control for exporters -> exporter hangs.
  
   A possible fix is to detect pdata-channel closure even while `accepting_pdata` is false, instead of only doing that when the buffer is empty. Note that the data can still not be drained by the exporter, but at least it can exit cleanly without the collector in a hung state. This can be done in the exporter's select loop, but the same gap exists for other callers of `recv_when(false)`, so an engine-level fix in `recv_when` would be more complete.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/auth.rs
```diff
@@ -65,101 +107,22 @@ impl Auth {
...
-                    let mut m = self.metrics.borrow_mut();
-                    m.add_auth_success_latency(start.elapsed().as_millis() as f64);
-                    return Ok(token);
-                }
-                Err(e) => {
-                    let error_msg = e.to_string();
-                    let first_line = error_msg.lines().next().unwrap_or(&error_msg);
-                    otel_warn!("azure_monitor_exporter.auth.get_token_failed", attempt = attempt, error = %first_line);
-                    otel_debug!("azure_monitor_exporter.auth.get_token_failed.details", attempt = attempt, error = %e);
-                    self.metrics.borrow_mut().add_auth_failure();
```
lalitb, line 142: we removed this metrics from the new flow - should we bring it back ?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
lalitb: Raised #2315 to fix it at engine level. However, do we see if we can also control the issue through the exporter's select loop.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/auth.rs
gouslu: ‼️

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
gouslu: fixing with -> https://github.com/open-telemetry/otel-arrow/pull/2317

gouslu: As you are aware the fixes are on hold, but this PR can still be merged without those fixes. I understand the issue, but it is an edge case and we are already tracking the bug. If you are also OK, I'd like to merge this @lalitb.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
```diff
@@ -613,6 +613,7 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
                     }
                 }
 
+                // TODO: Ensure that when rejecting pdata, data loss doesn't occur. (pending on lquerel's msg channel rework)
```
lalitb, line 616: nit- this TODO to be removed once the #2370 is merged.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
lalitb: Agree, the PR is orthogonal to #2370  - this PR is about the auth lifetime, and should be safe to go. And #2317  should be closed once #2370 is merged.

lalitb (approved):
LGTM/

Issue #2315, opened 2026-03-13
recv_when(false) does not detect pdata channel closure when buffer has messages


lalitb:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Bug Description

 `recv_when(false)` only checks if the pdata channel is closed when the buffer is empty (`is_empty() == true`). If the upstream sends messages and then shuts down, the channel is closed but has buffered messages. `is_empty()`  returns false, the closure check is skipped, and `recv_when` blocks on the control channel waiting for a Shutdown that never arrives - because the engine only sends Shutdown via control to receivers, not to processors or exporters.
   
   This means any node calling `recv_when(false)` can hang if the pdata channel closes while there are still buffered messages.
   
   Known callers affected:
   - **Azure Monitor exporter** - `accepting_pdata` is false during auth outages (see comment in #2312)
   - **Fanout processor** - `accept_pdata()` returns false when inflight count hits max_inflight

### Steps to Reproduce

  1. Have a node that calls `recv_when(false)` for an extended period (e.g., Azure Monitor exporter during auth failure)
   2. Upstream sends a few pdata messages
   3. Upstream shuts down (drops its pdata sender)
   4. The node's pdata channel is now closed with buffered messages
   5. `recv_when(false)` never detects the closure - node hangs

### Expected Behavior

   `recv_when(false)` should detect pdata channel closure regardless of whether the buffer is empty, so the node can shut down cleanly.

### Actual Behavior

   `recv_when(false)` skips the closure check when the buffer is not empty, and the node hangs forever.

### OTel-Arrow Version

latest

### Environment

All env

### Configuration

```yaml

```

### Log Output

```shell

```

jmacd:
@lalitb do you think this is still an open issue?

PR #2316, opened 2026-03-13
fix: avoid consuming pdata close probe in recv_when()

ThomsonTan:
# Change Summary

- **Fix a data-loss race** in the pdata-closed probe inside `recv_when(false)`.
  The probe used `is_empty()` then `try_recv()` to detect channel closure on
  shared pdata channels. If pdata arrived between those two calls (possible from
  another OS thread), `try_recv()` would dequeue and silently drop the message.
- **Replace with `is_closed() && is_empty()`** - two non-consuming checks that
  eliminate the TOCTOU window entirely. No data is ever dequeued by the probe.
- **Add `Receiver::is_closed()`** through the full receiver stack (`otap-df-channel`
  mpsc/mpmc, `LocalReceiver`, `SharedReceiver`, `Receiver` enum).

* Closes #NNN

- [x] New `test_mpsc_is_closed` / `test_mpmc_is_closed` - unit tests for
  `SharedReceiver::is_closed()` on both backends
- [x] `cargo xtask check` passes (fmt, clippy, all tests)

### DIFF
--- a/rust/otap-dataflow/crates/channel/src/mpmc.rs
+++ b/rust/otap-dataflow/crates/channel/src/mpmc.rs
@@ -1052,7 +1052,7 @@ mod tests {
         let handle = local.spawn_local(async {
             let (tx, rx) = Channel::<i32>::new(NonZeroUsize::new(10).unwrap());
 
-            // Sender alive — not closed
+            // Sender alive - not closed
             assert!(!rx.is_closed());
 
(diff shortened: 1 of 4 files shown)


ThomsonTan:
Close this one and prefer https://github.com/open-telemetry/otel-arrow/pull/2317 which also fixes issue #2315

PR #2317, opened 2026-03-13
engine: fix recv_when shutdown hang when pdata channel closed with buffered data

gouslu:
# Change Summary

Add is_closed() to all channel receiver types (mpsc, mpmc, local, shared) and use it in recv_when() instead of the previous is_empty()+try_recv() approach.

Previously, recv_when(false) only detected pdata channel closure when the buffer was empty. If the upstream processor sent messages before shutting down, those messages sat in the buffer, is_empty() returned false, and the closure check was skipped. Since no Shutdown arrives on the control channel for exporters (engine only sends Shutdown to receivers), the node would hang forever.

This is particularly easy to trigger during auth outages where accepting_pdata stays false for extended periods.

The fix checks is_closed() regardless of buffer state. The buffered data cannot be drained while accept_pdata is false anyway, so detecting closure and initiating shutdown is the correct behavior.

* Closes https://github.com/open-telemetry/otel-arrow/issues/2315

- Added `test_auth_outage_shutdown_does_not_hang` test.
- When fix is removed `test_auth_outage_shutdown_does_not_hang` test fails.

no


jmacd:
@lalitb please take a look

rust/otap-dataflow/crates/engine/src/message.rs
```diff
@@ -270,27 +280,21 @@ impl<PData: ReceivedAtNode> MessageChannel<PData> {
...
+            // We check is_closed() regardless of buffer state — the sender side
+            // may have closed while items remain in the buffer. Without this,
+            // buffered items prevent closure detection and the node hangs.
             if !accept_pdata
                 && self
                     .pdata_rx
                     .as_ref()
                     .expect("pdata_rx must exist")
-                    .is_empty()
+                    .is_closed()
```
lalitb: The check here should be `is_closed() && is_empty()`, which is what #2316 does. That fixes the TOCTOU data-loss race by removing the consuming `try_recv()` probe, while preserving the existing behavior that buffered pdata is not discarded on immediate shutdown.
   
   Replacing `is_empty()` with `is_closed()` alone changes `recv_when(false)` semantics for every caller that uses `accept_pdata=false` as temporary backpressure. For example, the fanout processor at `max_inflight` can recover capacity via Ack/Nack and then drain the buffered data. With this change, if upstream closes during that backpressure window, the data is lost:
   
   ```
   Before: upstream sends [A, B], closes -> recv_when(false) waits on control -> Ack arrives -> accept_pdata=true -> A, B drained [OK]
   
   After: upstream sends [A, B], closes -> recv_when(false) sees is_closed()=true -> immediate Shutdown -> A, B discarded [DATA LOSS]
   ```
   
  Thinking more, the auth-outage hang (#2315) fixing likely requires either an exporter-side change that avoids long-lived `accepting_pdata=false` during auth failure, or an engine-level shutdown change (e.g., propagating explicit Shutdown control messages to all nodes, not just receivers). It should not be solved by changing `recv_when(false)` to treat closed-with-buffered pdata as immediate shutdown.
   
   Suggested path forward:
   - Use `is_closed() && is_empty()` here (as in #2316) to fix the TOCTOU race
   - Address the auth-outage hang separately with a targeted fix in azure monitor exporter.

Sorry , if my comment on the Azure monitor exporter PR created the confusion :(

cc @ThomsonTan

rust/otap-dataflow/crates/engine/src/message.rs
```diff
@@ -270,27 +280,21 @@ impl<PData: ReceivedAtNode> MessageChannel<PData> {
...
-                    .expect("pdata_rx must exist")
-                    .try_recv()
-                {
-                    self.shutdown();
-                    return Ok(Message::Control(NodeControlMsg::Shutdown {
-                        deadline: Instant::now().add(Duration::from_secs(1)),
-                        reason: "pdata channel closed".to_owned(),
-                    }));
-                }
+                self.shutdown();
```
lalitb: If a real Shutdown is already pending (draining mode with `shutting_down_deadline` set), this path runs before the draining-mode check and replaces the original deadline/reason with a synthetic 1-second shutdown. The caller-provided deadline is lost and the grace period is shortened.
But honestly, this is a pre-existing issue - the old `is_empty() + try_recv()` code had the same ordering. Not a blocker for this PR, I can create as separate issue for that.

gouslu: thanks for these insights, I'm trying to address all.

gouslu: ok, I also added `shutting_down_deadline.is_none()` to the check, I hope that does the trick.

rust/otap-dataflow/crates/engine/src/message.rs
```diff
@@ -268,29 +278,26 @@ impl<PData: ReceivedAtNode> MessageChannel<PData> {
...
+            // there is still buffered data waiting to be drained.
+            // We skip this when shutting_down_deadline is set because the
+            // draining-mode block below already handles channel closure and
+            // preserves the caller-provided deadline and reason.
+            if !accept_pdata && self.shutting_down_deadline.is_none() && {
+                let pdata = self.pdata_rx.as_ref().expect("pdata_rx must exist");
+                pdata.is_closed() && pdata.is_empty()
+            } {
+                self.shutdown();
+                return Ok(Message::Control(NodeControlMsg::Shutdown {
```
ThomsonTan: Preserving deadline on `self.pending_shutdown` instead of always creating a synthetic 1-second one.

gouslu: I need an explanation of this please :)

rust/otap-dataflow/crates/engine/src/message.rs
gouslu: Moving this logic to exporter feels wrong. It seems like multiplexing of the channels is there for a good reason, but it also makes shutdown logic far too complex. hmm 🤔 ...

gouslu: So, if we get shutdown, would it make sense to auto-nack every message that we failed to process instead of dropping? What I mean is that, we are at outage, and we receive shutdown, exporter drains the current request, but it is still at outage and it doesn't make sense to process any more data just to eventually nack them. So, basically, if a component is still not processing any data after shutdown, we take the bufferred data and nack all of them. This seem to require a relatively wide impact change for one of the implementation possiblities I am looking for, but it is mostly because of method signature change.

Either way, without getting into the implementation detail, does such a functionality make sense?

gouslu: @lalitb can you take a look at this -> https://github.com/open-telemetry/otel-arrow/pull/2326

I created this PR as a conversation starter, and if the idea is not right, I am fine with abandoning it.

rust/otap-dataflow/crates/engine/src/message.rs
ThomsonTan: I meant returning `self.pending_shutdown.take()` if available, like below:

```rust
    let shutdown_msg = self.pending_shutdown.take().unwrap_or_else(|| {
        NodeControlMsg::Shutdown {
            deadline: Instant::now().add(Duration::from_secs(1)),
            reason: "pdata channel closed".to_owned(),
        }
    });
    self.shutdown();
    return Ok(Message::Control(shutdown_msg));
```

jmacd: I believe @ThomsonTan is referring to the value in the line below. I believe we know the true deadline, the value below is made up.

lalitb: I may be missing something, but I think this concern applied to the earlier version before the `shutting_down_deadline.is_none()` guard. In the current code, this branch only runs when no real Shutdown is already pending, so there is no existing deadline to preserve here.

ThomsonTan: You're right. The `shutting_down_deadline.is_none()` guard ensures this branch only runs outside draining mode, so `pending_shutdown` is always None here. The synthetic 1-second deadline is correct for this path. Thanks for the clarification.

gouslu: I'll still do a small refactoring here, addressing both concerns.

gouslu: Paranoid programming!

gouslu: this -> https://github.com/open-telemetry/otel-arrow/pull/2317/changes/cae108cfaf804ba2e166380da1c6d74b3e4b80ba

ThomsonTan: The consolidated `showdown()` looks good to me, thanks @gouslu .

rust/otap-dataflow/crates/engine/src/message.rs
```diff
@@ -404,7 +390,30 @@ impl<PData: ReceivedAtNode> MessageChannel<PData> {
...
+        let msg = self.pending_shutdown.take().unwrap_or_else(|| {
+            otel_warn!(
+                "message_channel.synthetic_shutdown",
+                reason = "pdata channel closed before Shutdown was received"
+            );
+            NodeControlMsg::Shutdown {
+                deadline: Instant::now().add(Duration::from_secs(1)),
+                reason: "pdata channel closed".to_owned(),
+            }
+        });
```
jmacd, line 411: I get it. Thanks!

jmacd:
@gouslu Thank you, and thank you also for the proposal in https://github.com/gouslu/otel-arrow/blob/gouslu/channel-redesign/rust/otap-dataflow/docs/channel-redesign.md which I have read and agree with. I would close this and move our pending discussion to #2431. Thank you for your work on this.

PR #2318, opened 2026-03-13
azure monitor exporter: optimize transformer with direct JSON serialization

gouslu:
# Change Summary

- Pre-serialize resource+scope fields once per ScopeLogs as a JSON byte prefix
- Write per-record fields directly to byte buffer using itoa/ryu, bypassing serde_json::Map entirely
- Add write_json_string, write_json_hex, write_field_value_json for zero-allocation JSON output
- Make config and metrics modules public for benchmark access
- Add criterion benchmark under contrib-nodes/benches/exporters/azure_monitor_exporter/
- Added contrib bench to rust-bench workflow.

Benchmark results (1000 records):
  Original:       1.60ms (~625K records/s)
  Hoisted:        1.36ms (~735K records/s)  +17%
  Hoisted + Direct Serialization:  425us  (~2.35M records/s) +275%

- Existing unit tests cover already mapping uniqueness, also added tests to make sure that overlapping fields are rejected by the config valiation across resource -> scope -> log hiearchy.
- Added tests for validating against encoding issues for non ASCII characters.

None.

### DIFF
--- a/rust/otap-dataflow/crates/contrib-nodes/benches/exporters/azure_monitor_exporter/transformer.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/benches/exporters/azure_monitor_exporter/transformer.rs
@@ -2,6 +2,11 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! Benchmark for Azure Monitor Transformer: OTLP logs -> JSON conversion.
+//!
+//! MacOS M4 Pro baseline results (1000 records):
+//!   Original (per-record resource/scope mapping):  1.60ms (~625K records/s)
+//!   Hoisted (resource/scope computed once):        1.36ms (~735K records/s)  +17%
+//!   Hoisted + Direct Serialization (current):      425µs  (~2.35M records/s) +275%
 
 #![allow(unused_results)]
 
--- a/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
@@ -304,38 +304,33 @@ impl Transformer {
     /// Write an AnyValueView directly as JSON bytes.
     fn write_any_value_json<'a, V: AnyValueView<'a>>(value: &V, out: &mut Vec<u8>) {
         match value.value_type() {
-            ValueType::String => {
-                Self::write_json_string(value.as_string().unwrap_or(b""), out);
-            }
-            ValueType::Int64 => {
-                let n = value.as_int64().unwrap_or(0);
-                Self::write_i64(n, out);
-            }
-            ValueType::Double => {
-                let d = value.as_double().unwrap_or(0.0);
-                if d.is_finite() {
-                    // Use ryu for fast float formatting
+            ValueType::String => match value.as_string() {
+                Some(s) => Self::write_json_string(s, out),
+                None => out.extend_from_slice(b"null"),
+            },
+            ValueType::Int64 => match value.as_int64() {
+                Some(n) => Self::write_i64(n, out),
+                None => out.extend_from_slice(b"null"),
+            },
+            ValueType::Double => match value.as_double() {
+                Some(d) if d.is_finite() => {
                     let mut ryu_buf = ryu::Buffer::new();
-                    out.extend_from_slice(ryu_buf.format(d).as_bytes());
-                } else {
-                    out.extend_from_slice(b"null");
+                    out.extend_from_slice(ryu_buf.format_finite(d).as_bytes());
                 }
-            }
-            ValueType::Bool => {
-                if value.as_bool().unwrap_or(false) {
-                    out.extend_from_slice(b"true");
-                } else {
-                    out.extend_from_slice(b"false");
-                }
-            }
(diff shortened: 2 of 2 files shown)


rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
```diff
@@ -37,6 +37,16 @@ use otap_df_telemetry::{otel_debug, otel_error, otel_info, otel_warn};
...
+/// Max concurrent HTTP requests in flight to the Logs Ingestion API.
+///
+/// Approximation using Little's Law (L = λ × W):
+///   - λ (throughput): LA API allows ~500 req/min per DCR (~8 req/s, estimated)
+///   - W (latency): p99 response time ~0.5-1s (estimated, varies by region/load)
+///   - L ≈ 8 × 0.5 = 4 slots minimum, doubled for burst headroom ≈ 8
+///
+/// Set to 16 to absorb transient latency spikes without throttling.
+/// Worst-case memory for pending requests is roughly: 16 × 1MB = 16MB.
+/// These are rough estimates — tune based on observed metrics in production.
```
jmacd, line 49: Nice comment 🥇

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
```diff
@@ -202,6 +214,181 @@ impl Transformer {
...
+            ValueType::Int64 => {
+                let n = value.as_int64().unwrap_or(0);
+                Self::write_i64(n, out);
+            }
+            ValueType::Double => {
+                let d = value.as_double().unwrap_or(0.0);
+                if d.is_finite() {
+                    // Use ryu for fast float formatting
+                    let mut ryu_buf = ryu::Buffer::new();
+                    out.extend_from_slice(ryu_buf.format(d).as_bytes());
```
jmacd: should you use format_finite() since you've checked?
https://docs.rs/ryu/latest/ryu/struct.Buffer.html#method.format_finite

jmacd: I can't resist thinking about how to avoid the Buffer::new() and the extend_from_slice, for example but it requires an [unsafe method](https://docs.rs/ryu/latest/ryu/raw/fn.format64.html)!

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
```diff
@@ -136,61 +132,77 @@ impl Transformer {
...
+                let mut scope_map = base_map.clone();
+                if let Some(s) = scope_logs.scope() {
+                    self.apply_scope_mapping(&s, &mut scope_map);
+                }
 
-                    // apply resource mapping first to allow scope to override if needed
-                    if let Some(r) = resource_logs.resource() {
-                        self.apply_resource_mapping(&r, &mut record_map);
-                    }
+                // Pre-serialize resource+scope as JSON bytes (once per ScopeLogs)
```
cijothomas, line 176: 💯 That is awesome!

rust/otap-dataflow/crates/contrib-nodes/benches/exporters/azure_monitor_exporter/transformer.rs
```diff
@@ -0,0 +1,237 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Benchmark for Azure Monitor Transformer: OTLP logs -> JSON conversion.
```
cijothomas, line 4: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/benches/metrics_counter.rs#L2-L15

Can you share the numbers from your machine and paste it as comment? It is very convenient for a quick reference without having to run them all the time.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
gouslu: > should you use format_finite() since you've checked? https://docs.rs/ryu/latest/ryu/struct.Buffer.html#method.format_finite

applying this

rust/otap-dataflow/crates/contrib-nodes/benches/exporters/azure_monitor_exporter/transformer.rs
gouslu: on it

rust/otap-dataflow/crates/contrib-nodes/benches/exporters/azure_monitor_exporter/transformer.rs
```diff
@@ -0,0 +1,242 @@
...
+//!
+//! MacOS M4 Pro baseline results (1000 records):
+//!   Original (per-record resource/scope mapping):  1.60ms (~625K records/s)
```
cijothomas, line 7: these "original measurements"  new makes sense in PR desc. For this file, only the _current_ one is required.

Issue #2308, opened 2026-03-13
Proposal: Query recent internal engine logs via admin API


lalitb:
## Summary

  Add support for querying recent internal engine logs through the admin API.                                                                                                                                                                                                                                           
  Proposed endpoints:                                                                                                                                                                                                                                                                                                   
                  
  - `GET /telemetry/logs`
  - `GET /telemetry/logs?after=<seq>&limit=<n>`

  ## Current behavior

  Internal engine logs already flow through the telemetry/logging system and are emitted through the configured provider (`ConsoleDirect`, `ConsoleAsync`, `ITS`). That covers log emission, but there is no admin/API surface for retrieving a recent window of internal logs from the running engine.

  ## Why this would help

  - Admin and diagnostic tooling
  - Orchestration and readiness workflows that need to inspect startup logs without shell access
  - MCP/agent clients that need structured engine diagnostics
  - Tests that want to assert on engine log output through a stable API

  ## Design sketch

  An in-memory ring buffer is maintained inside the engine, fed from the existing internal logging layer via a non-blocking side-channel (does not replace or affect the configured log provider). Recent log entries are retained up to configurable limits and exposed through the admin HTTP server.

  Key properties:

  - **Non-blocking on the hot path** - log events are enqueued via `try_send`; drops are counted and reported, never block the caller
  - **Bounded retention** - dual limits: `max_entries` and `max_bytes`; oldest entries are evicted when either is exceeded
  - **Cursor-based polling** - `after=<seq>` allows incremental fetch so clients only receive new records since their last read
  - **Opt-in, zero cost when disabled** - no channel, no task, no lock is allocated unless `tap.enabled: true`
  - **Orthogonal to existing sinks** - tap is fed alongside the configured provider, not instead of it

 ## Prior art

  The Go collector has two relevant building blocks:

  - **`logger_tee`** - intercepts logs at the shared internal logging layer, which validates that layer as the right interception point. This proposal extends that approach by adding a queryable retention surface on top.
  - **`zpages`** - retains recent internal signals in-memory and exposes them via HTTP endpoints for live diagnostics (`/debug/tracez`, `/debug/pipelinez` etc.). This proposal applies the same pattern to internal log events - a gap that zpages does not currently cover.

  ## Proposed API

  ```http
  GET /telemetry/logs
  GET /telemetry/logs?after=<seq>&limit=<n>
  ```

  Example response:

  ```json
  {
    "oldest_seq": 101,
[...]

PR #2319, opened 2026-03-13
 feat: Query recent internal engine logs via admin API (log tap)

lalitb:
## Summary

This adds an in-memory internal log tap exposed through the admin API at `/telemetry/logs`. It retains recent internal logs with bounded memory and reports tap health stats such as `dropped_on_ingest`, `dropped_on_retention`, and `retained_bytes`.

**Current scope:**

- enable retained internal log access through `telemetry.logs.tap.enabled`
- query recent logs through the admin endpoint
- expose basic tap health counters
   

## Performance

- When `logs.tap.enabled` is off, the tap path stays inactive. No retained-log buffer is created, no extra log clone or lock is taken, and there should be no user-visible latency or meaningful per-log memory overhead beyond trivial `Option` checks.
- When `logs.tap.enabled` is on, the added work happens on the async consumer side, not on the emitting thread. This adds a small amount of retention bookkeeping and a shallow log clone, with memory bounded by the configured retention limits.

   ## Usage
   
   ```yaml
   engine:
     telemetry:
       logs:
         tap:
           enabled: true
   ```

   
   ## Tested with
   
   traffic generator -> retry -> batch -> geneva exporter.
   
### Latest 10 logs:
```
$ curl -s 'http://localhost:8080/telemetry/logs?limit=10'  | jq | grep rendered

      "rendered": "2026-03-13T20:32:11.619Z  DEBUG geneva-uploader::uploader.upload.post: Posting to ingestion gateway [event_name=Log, moniker=pademo2diagbl] entity=EntityKey(473v1)\n",
      "rendered": "2026-03-13T20:32:11.619Z  DEBUG otap-df-otap::fake_data_generator.rate_limit.sleep: Sleeping to maintain configured signal rate [sleep_duration_ms=999] entity=EntityKey(495v1)\n",
      "rendered": "2026-03-13T20:32:11.619Z  INFO  otap-df-contrib-nodes::geneva_exporter.upload: Uploading logs to Geneva using OTLP path entity=EntityKey(423v1)\n",
      "rendered": "2026-03-13T20:32:11.619Z  DEBUG geneva-uploader::client.encode_and_compress_logs: Encoding and compressing resource logs [resource_logs_count=5] entity=EntityKey(423v1)\n",
      "rendered": "2026-03-13T20:32:11.619Z  DEBUG geneva-uploader::encoder.encode_log_batch: Encoded log batch [event_name=Log, schemas=1, events=500, uncompressed_size=117752, compressed_size=3742] entity=EntityKey(423v1)\n",
      "rendered": "2026-03-13T20:32:11.619Z  DEBUG geneva-uploader::client.upload_batch: Uploading batch [event_name=Log, size=3742] entity=EntityKey(423v1)\n",
      "rendered": "2026-03-13T20:32:11.619Z  DEBUG geneva-uploader::uploader.upload: Starting upload [event_name=Log, size=3742] entity=EntityKey(423v1)\n",
      "rendered": "2026-03-13T20:32:11.619Z  DEBUG geneva-uploader::config_client.get_ingestion_info: Getting ingestion info (checking cache first) entity=EntityKey(423v1)\n",
      "rendered": "2026-03-13T20:32:11.619Z  DEBUG geneva-uploader::config_client.get_ingestion_info.cache_hit: Using cached ingestion info [expiry=2026-03-16 20:29:21.438837400 UTC] entity=EntityKey(423v1)\n",
      "rendered": "2026-03-13T20:32:11.619Z  DEBUG geneva-uploader::uploader.upload.post: Posting to ingestion gateway [event_name=Log, moniker=pademo2diagbl] entity=EntityKey(423v1)\n",
```

### Tap health stats

```bash
$ curl -s 'http://localhost:8080/telemetry/logs?limit=1' | jq '{dropped_on_ingest, dropped_on_retention, retained_bytes}'
{
  "dropped_on_ingest": 0,
  "dropped_on_retention": 123558,
  "retained_bytes": 1656680
}
```

  ### Filter by level
   
   ```bash
$ curl -s 'http://localhost:8080/telemetry/logs?limit=5' | jq '[.logs[] | select(.level == "DEBUG")] | .[].rendered'
"2026-03-13T20:41:37.897Z  DEBUG otap-df-otap::fake_data_generator.rate_limit.sleep: Sleeping to maintain configured signal rate [sleep_duration_ms=999] entity=EntityKey(23v1)\n"
"2026-03-13T20:41:37.909Z  DEBUG geneva-uploader::uploader.upload.success: Upload successful [event_name=Log, ticket=v1|westus-0-sfmc|6f537847340b460ba3255f7c2e7c5923|0000000000000000|https://test1.diagnostics.monitoring.core.windows.net/-pademo2diagbl-GenevaCommon-RedirectionEvent3876-Ver2v0-134179080978676520-0000000000000000-eastus-Default] entity=EntityKey(81v1)\n"
"2026-03-13T20:41:37.909Z  DEBUG geneva-uploader::client.upload_batch.success: Successfully uploaded batch [event_name=Log] entity=EntityKey(81v1)\n"
"2026-03-13T20:41:37.911Z  DEBUG otap-df-otap::fake_data_generator.rate_limit.sleep: Sleeping to maintain configured signal rate [sleep_duration_ms=999] entity=EntityKey(125v1)\n"
$ 
   ```

## Future possibilities
   
With a proper JSON client (e.g., MCP tool, CLI, or dashboard), this API could support richer workflows - server-side level filtering, streaming/tailing, per-pipeline log scoping, and structured queries over the entity context attributes already present in each log entry.

### DIFF
--- a/rust/otap-dataflow/crates/telemetry/src/tracing_init.rs
+++ b/rust/otap-dataflow/crates/telemetry/src/tracing_init.rs
@@ -103,20 +103,13 @@ impl ProviderSetup {
             ProviderSetup::Noop => Dispatch::new(tracing::subscriber::NoSubscriber::new()),
 
             ProviderSetup::ConsoleDirect => {
-                let layer = StructuredLoggingLayer::new(
-                    Some(ConsoleWriter::color()),
-                    None,
-                    context_fn,
-                );
+                let layer =
+                    StructuredLoggingLayer::new(Some(ConsoleWriter::color()), None, context_fn);
                 Dispatch::new(Registry::default().with(filter).with(layer))
             }
 
             ProviderSetup::InternalAsync { reporter } => {
-                let layer = StructuredLoggingLayer::new(
-                    None,
-                    Some(reporter.clone()),
-                    context_fn,
-                );
+                let layer = StructuredLoggingLayer::new(None, Some(reporter.clone()), context_fn);
                 Dispatch::new(Registry::default().with(filter).with(layer))
             }
         }
--- a/rust/otap-dataflow/crates/config/src/settings/telemetry/logs.rs
+++ b/rust/otap-dataflow/crates/config/src/settings/telemetry/logs.rs
@@ -438,7 +438,10 @@ mod tests {
             ..Default::default()
         };
         config.tap.enabled = true;
-        assert_invalid(&config, "logs.tap.enabled requires at least one async log provider");
+        assert_invalid(
+            &config,
+            "logs.tap.enabled requires at least one async log provider",
+        );
 
         for providers in [
             providers(ConsoleAsync, Noop, Noop, ConsoleDirect),
@@ -453,7 +456,9 @@ mod tests {
                 },
                 ..Default::default()
             };
-            config.validate().expect("async provider should support tap");
+            config
+                .validate()
+                .expect("async provider should support tap");
         }
     }
 
--- a/rust/otap-dataflow/crates/state/src/store.rs
+++ b/rust/otap-dataflow/crates/state/src/store.rs
@@ -431,13 +431,13 @@ mod tests {
     use otap_df_telemetry::registry::TelemetryRegistryHandle;
     use otap_df_telemetry::self_tracing::{LogContext, LogRecord};
     use std::borrow::Cow;
-    use std::time::SystemTime;
     use std::time::Duration;
-    use tracing::{Event, Subscriber};
-    use tracing_subscriber::layer::{Context, Layer};
-    use tracing_subscriber::layer::SubscriberExt;
-    use tracing_subscriber::registry::LookupSpan;
+    use std::time::SystemTime;
     use tokio_util::sync::CancellationToken;
+    use tracing::{Event, Subscriber};
+    use tracing_subscriber::layer::SubscriberExt;
+    use tracing_subscriber::layer::{Context, Layer};
+    use tracing_subscriber::registry::LookupSpan;
 
     fn make_key(core_id: usize) -> otap_df_config::DeployedPipelineKey {
         otap_df_config::DeployedPipelineKey {
@@ -704,11 +704,13 @@ mod tests {
                 console_fallback: false,
             },
         };
-        let log_tap = log_tap::build(&otap_df_config::settings::telemetry::logs::InternalLogTapConfig {
-            enabled: true,
-            max_entries: 8,
-            max_bytes: 1024 * 1024,
-        });
+        let log_tap = log_tap::build(
+            &otap_df_config::settings::telemetry::logs::InternalLogTapConfig {
+                enabled: true,
+                max_entries: 8,
+                max_bytes: 1024 * 1024,
+            },
+        );
         let store = ObservedStateStore::new_with_log_tap(
             &config,
             TelemetryRegistryHandle::new(),
@@ -753,11 +755,13 @@ mod tests {
                 console_fallback: false,
             },
         };
-        let log_tap = log_tap::build(&otap_df_config::settings::telemetry::logs::InternalLogTapConfig {
-            enabled: true,
-            max_entries: 8,
-            max_bytes: 1024 * 1024,
-        });
+        let log_tap = log_tap::build(
+            &otap_df_config::settings::telemetry::logs::InternalLogTapConfig {
+                enabled: true,
+                max_entries: 8,
+                max_bytes: 1024 * 1024,
+            },
+        );
         let store = ObservedStateStore::new_with_log_tap(
             &config,
             TelemetryRegistryHandle::new(),


cijothomas:
>  curl -s 'http://localhost:8080/telemetry/logs?limit=10'  | jq | grep rendered
> 
>       "rendered": "2026-03-13T20:32:11.619Z  DEBUG

How come you are getting DEBUG logs? Engine does not enable debug by default. Does tap somehow changes that?

lalitb:
> How come you are getting DEBUG logs? Engine does not enable debug by default. Does tap somehow changes that?

I enabled debug level before running the engine. We are still not there to enable level at runtime :)

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -129,11 +145,106 @@ struct AggregateGroup {
...
+    seq: u64,
+    timestamp: String,
+    level: String,
+    target: String,
+    event_name: String,
+    file: Option<String>,
+    line: Option<u32>,
+    rendered: String,
+    contexts: Vec<ResolvedLogContext>,
+}
```
jmacd, line 171: I'm curious if we can use OTLP as the representation here instead of a bespoke protocol? I'm thinking about how the internal telemetry receiver and instrumentation exchange OTLP bytes already, so we might be able to store those bytes directly, pass them to the UI directly, work with them using other OTel-friendly libraries, etc.

lalitb: Great suggestion. I think we can keep the current in-memory representation rather than store full OTLP bytes in the ring buffer, and still provide the API for users to request otlp_json or otlp_bytes?

One important clarification: the log tap does not receive full OTLP bytes today. Internally it receives LogEvent / ObservedEvent, where the log body and attributes are already captured in OTLP-shaped bytes, but the full OTLP payload would only be assembled later when we actually need to emit/export it.

The PR tries to set the right boudaries: log payload is captured in OTLP-friendly bytes, while callsite metadata and entity context stay as cheap local references to keep the hot path light. The is intentional as it keeps extra work off the logging hot path.

If we stored full OTLP for every retained log in ring buffer, we’d duplicate the same resource/scope wrapper bytes over and over for logs that share the same context. That adds memory cost without much benefit, and it would also push more encoding and context-resolution work into capture, which is exactly the path we’ve been trying to keep light. And there is also tap-specific state that OTLP doesn’t model, like sequence numbers, truncation, and drop counters, which the polling API depends on.

So my take is: keep the internal representation as-is, and if we want the interoperability benefits you called out, add an OTLP output mode on /telemetry/logs. That gives external tools standard OTLP without making the engine pay that cost on every log event.

jmacd: The engine uses https://github.com/open-telemetry/otel-arrow/blob/6933f8e2d5629a7b76983b60ff43526d67e7f33f/rust/otap-dataflow/crates/telemetry/src/self_tracing.rs#L35 LogRecord, an intermediate representation of log records that addresses the concern you have. 

This representation is used between call sites and raw logging events, and it is used between call sites and the console_async consumer as well. Only when ITR submits the data to a pipeline does it get fully encoded as OTLP bytes with the extra information and ITR is able to group by scope at that point (however it does not currently group by scope). So, it's probably possible to use LogRecord: there's an efficient translation into formatted record or OTLP bytes basically for the same reasons. I wouldn't let this idea block you, on the other hand, we could file it for later.

lalitb: Thanks @jmacd , that’s a useful clarification. You’re right that `LogRecord` is the closer internal representation here than full OTLP bytes, which makes an OTLP-oriented output more feasible than I’d framed it. In this PR, the tap still retains `LogEvent` plus tap-specific polling state (seq, truncation, drop counters), and the admin endpoint projects that into JSON for the UI. To keep this draft scoped, I can land the current API and file a follow-up for an `OTLP`/`LogRecord` output mode on `/telemetry/logs`.

lalitb:
@jmacd @lquerel - I’ve been using a version of this PR locally for my tests, and it’s been pretty handy for inspecting recent internal engine logs through the admin API, especially when running the engine at DEBUG level. No pressure at all to take this feature as-is - I know the original issue #2308 is still being triaged. Just sharing in case it feels useful more broadly. If this seems like the right direction, I can get it ready for formal review; otherwise I’m happy to drop it or rework it in whatever direction makes more sense :)

rust/otap-dataflow/crates/telemetry/src/tracing_init.rs
```diff
@@ -68,47 +70,83 @@ impl TracingSetup {
...
+        tap: Option<InternalLogTapReporter>,
+    },
 
-    /// Synchronous console logging via `RawLoggingLayer`.
-    ConsoleDirect,
+    /// Synchronous console logging via `StructuredLoggingLayer`.
+    ConsoleDirect {
+        /// Optional internal log tap.
+        tap: Option<InternalLogTapReporter>,
+    },
```
jmacd: I think if a user selects Noop or ConsoleDirect, they are making a choice other than logtap. Only the InternalAsync choice would support logtap, the way I imagine it.

rust/otap-dataflow/crates/telemetry/src/lib.rs
```diff
@@ -234,6 +253,9 @@ impl InternalTelemetrySystem {
             context_fn,
             console_async_reporter,
             its_reporter,
+            log_tap_reporter,
+            log_tap_handle,
+            log_tap_runtime: Mutex::new(log_tap_runtime),
```
jmacd: The way I was imagining this, the console_async_reporter and the log_tap_reporter would be mutually exclusive or else one and the same. I was thinking we would change ProviderMode,

```
    /// Delivery using the internal telemetry system.
    ITS,
    /// Asynchronous console logging. The caller writes to a channel
    /// the same as ITS delivery, but bypasses the internal pipeline
    /// with console logging.
    #[serde(rename = "console_async")]
    ConsoleAsync,
```

I was thinking
(a) Rename "console_async" to "admin_async"
(b) Let the admin component have two booleans: LogTap and Console.

It would be the same code path, same runtime task: "admin_async" covering the non-ITR cases with two builtin sinks (i.e., console and logtap). By my proposal, the logtap feature would not require its own runtime task, it would be handled either by the ITR in its receiver configuration or by the admin component in its receiver configuration.

jmacd (changes requested):
Asking for a minor change, to be reflected in the configuration for logs. The summary is: LogTap disabled for Noop and ConsoleDirect; LogTap optional for AdminAsync and ITR cases, with no need for a LogTap-dedicated runtime task.

rust/otap-dataflow/crates/telemetry/src/log_tap.rs
```diff
@@ -0,0 +1,573 @@
...
+    /// Current retained memory estimate.
+    pub retained_bytes: usize,
+    /// Retained entries matching the query.
+    pub logs: Vec<RetainedLogEvent>,
+}
+
+/// A non-blocking reporter used on the tracing hot path.
+#[derive(Clone, Debug)]
+pub struct InternalLogTapReporter {
+    next_seq: Arc<AtomicU64>,
```
lquerel: IMO, you shouldn't maintain the seq on the producer side. 

`seq` is "allocated" before `try_send`, but retention/query order is whatever the shared queue receives. If core A gets 10 and core B gets 11, B can enqueue first and the deque becomes out of sequence. From there after, `next_seq`, and `truncated_before_seq` can duplicate or skip entries for polling clients. The clean fix is to assign sequence numbers on the single consumer thread when draining the queue, not on the producers.

That will also remove an Arc + Atomic on the hot path.

lquerel: QueuedLogEvent could be redefined:

```rust
#[derive(Debug, Clone)]
struct QueuedLogEvent {
    event: LogEvent,
    estimated_bytes: usize,
}
```

rust/otap-dataflow/crates/telemetry/src/log_tap.rs
```diff
@@ -0,0 +1,573 @@
...
+    dropped_on_ingest: Arc<AtomicU64>,
+    tx: flume::Sender<QueuedLogEvent>,
+}
+
+impl InternalLogTapReporter {
+    /// Queue a structured internal log event without blocking the caller.
+    pub fn log(&self, event: LogEvent) {
+        let seq = self.next_seq.fetch_add(1, Ordering::Relaxed);
```
lquerel: Could be removed.

rust/otap-dataflow/crates/telemetry/src/tracing_init.rs
lalitb: I modeled log tap as an orthogonal secondary sink, not as a provider mode, which is why it can be enabled with `Noop` and `ConsoleDirect` too. If the intended ITS/logging semantics are that only `InternalAsync` should support log tap, I’m happy to change that.

rust/otap-dataflow/crates/telemetry/src/log_tap.rs
lalitb: Good catch. Fixed as suggested - sequence assignment now happens on the consumer side and producer-side `seq` tracking has been removed.

rust/otap-dataflow/crates/telemetry/src/log_tap.rs
lalitb: done.

lalitb:
> Asking for a minor change, to be reflected in the configuration for logs. The summary is: LogTap disabled for Noop and ConsoleDirect; LogTap optional for AdminAsync and ITR cases, with no need for a LogTap-dedicated runtime task.

@jmacd - thanks for reviewing this.

I think there are two possible scopes here, and I want to confirm which one you want:

**1. Semantic-only change**
- Remove LogTap support from Noop
- Remove LogTap support from ConsoleDirect
- Keep LogTap only on InternalAsync

This is the smaller change.

**2. Semantic change plus architectural unification**
- Remove LogTap support from Noop and ConsoleDirect
- Remove the dedicated log_tap_runtime
- Reuse the existing async path for LogTap
- Make the current async log path fan out to console and/or retained-log sinks

This is the larger change.

I’m happy to do either, but I want to confirm whether you want just (1) in this PR, or the broader refactor in (2).

lalitb:
> > Asking for a minor change, to be reflected in the configuration for logs. The summary is: LogTap disabled for Noop and ConsoleDirect; LogTap optional for AdminAsync and ITR cases, with no need for a LogTap-dedicated runtime task.
> 
> @jmacd - thanks for reviewing this.
> 
> I think there are two possible scopes here, and I want to confirm which one you want:
> 
> **1. Semantic-only change**
> 
> * Remove LogTap support from Noop
> * Remove LogTap support from ConsoleDirect
> * Keep LogTap only on InternalAsync
> 
> This is the smaller change.
> 
> **2. Semantic change plus architectural unification**
> 
> * Remove LogTap support from Noop and ConsoleDirect
> * Remove the dedicated log_tap_runtime
> * Reuse the existing async path for LogTap
> * Make the current async log path fan out to console and/or retained-log sinks
> 
> This is the larger change.
> 
> I’m happy to do either, but I want to confirm whether you want just (1) in this PR, or the broader refactor in (2).


I went with option 2 above - remove LogTap support from `Noop` / `ConsoleDirect`, remove the dedicated `log_tap_runtime`, and reuse the existing `ConsoleAsync` / `ITS` async paths for retained-log fanout

rust/otap-dataflow/crates/telemetry/src/event.rs
```diff
@@ -82,10 +93,16 @@ impl ObservedEventReporter {
...
+        let is_log = matches!(event, ObservedEvent::Log(_));
         match self.policy.blocking_timeout {
             None => match self.sender.try_send(event) {
                 Ok(_) => {}
                 Err(err) => {
+                    if is_log {
+                        if let Some(drop_counter) = &self.drop_counter {
+                            drop_counter.increment();
+                        }
+                    }
```
jmacd, line 105: @lquerel are you concerned about the relaxed-memory increment that happens here? This can happen in any thread when the logtap feature is enabled. I think we should ask for another policy boolean along the lines of `self.policy.console_fallback` e.g. maybe `dropped_logs_counting` with a false default, which would avoid constructing the drop_counter.

Then again, for threads where this is a concern, we should aim to use macros that we control. Then for engine threads where we have thread-local access to engine state, we may be able to implement drop counters without memory barriers. (To me this is non-blocking feedback!)

lalitb: >  relaxed-memory increment that happens here?

Just to add, This only runs on the channel-send failure path, i.e. when the async log channel is already full or disconnected and the log is being dropped - not a steady-state hot-path cost.

So:
- if log tap is disabled, there is no `drop_counter` and no atomic here
- if log tap is enabled but the send succeeds, this branch is not taken
- the `Relaxed` atomic only happens when a log is already being dropped

jmacd:
> I went with option 2 above

By the way, I think this turned out great. Thank you!

PR #2325, opened 2026-03-14
Handle OS signals (SIGTERM/SIGINT) for graceful pipeline shutdown

cijothomas:
The main executable did not handle OS termination signals, so Kubernetes SIGTERM or local Ctrl+C terminated the process without draining in-flight data.

This PR adds controller-managed OS signal handling with a double-signal convention:

- First SIGINT/SIGTERM requests graceful shutdown for all pipelines with a 60-second drain deadline.
- A second signal forces immediate exit via `process::exit(1)`.

The repository's standalone executable enables this behavior explicitly. Library embedders retain process-wide signal ownership by default (`ControllerRunOptions::handle_os_signals` defaults to `false`) and can opt in when the controller owns the process lifetime.

The second signal is an in-process escape hatch for a drain that stalls. Unlike an external SIGKILL, `process::exit(1)` emits a direct stderr log recording that the drain was deliberately abandoned and yields an exit code distinguishable from 137, which Kubernetes also reports for OOMKilled.

### Follow-ups

- #3829 captures termination signals earlier during engine startup.
- #3833 keeps the second-signal force-exit path responsive while graceful shutdown dispatch is blocked.


rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -1343,6 +1345,118 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug + ReceivedAtNode + U
...
+                            "shutdown.signal_received",
+                            signal = signal_name,
+                            message = "OS termination signal received, initiating graceful \
+                                   shutdown. Send the signal again to force immediate exit."
+                        );
+
+                        // Give pipelines a generous deadline to drain (60 s by default —
+                        // matches the default Kubernetes terminationGracePeriodSeconds).
+                        let deadline =
+                            std::time::Instant::now() + std::time::Duration::from_secs(60);
```
gouslu: nit: named constant for 60 secs?

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -1343,6 +1345,118 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug + ReceivedAtNode + U
...
+                            otel_error!(
+                                "shutdown.signal_dispatch_failed",
+                                error_count = errors.len(),
+                                message = "Some pipelines could not be notified of shutdown"
+                            );
+                        }
+
+                        // ── Second signal: force exit ───────────────────────
+                        let signal_name = Self::recv_termination_signal().await;
+                        otel_error!(
```
gouslu: Currently it seems like it is sync, but if this ever uses a bufferred writer (which I doubt it will) this message may not get printed before exit. Maybe an eprintln! right under this for good measure?

cijothomas: agree. Will swap to eprintln.

jmacd: This is the purpose of raw_error!
I would prefer to use it for consistency.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -1343,6 +1345,118 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug + ReceivedAtNode + U
...
+
+                        // Give pipelines a generous deadline to drain (60 s by default —
+                        // matches the default Kubernetes terminationGracePeriodSeconds).
+                        let deadline =
+                            std::time::Instant::now() + std::time::Duration::from_secs(60);
+
+                        let mut errors = Vec::new();
+                        for sender in &senders {
+                            if let Err(e) = sender.try_send_shutdown(
+                                deadline,
```
lalitb: One correctness issue here: `try_send_shutdown()` can drop the shutdown request if the pipeline control channel is full at signal time. That makes graceful shutdown best-effort under backpressure.
 
For this PR, I think the smallest fix could be a bounded retry inside the existing `rt.block_on(async { ... })` block, e.g. retry `try_send_shutdown()` a few times with a short `tokio::time::sleep(...).await` between attempts before giving up and logging the error. That avoids trait changes and closes the immediate gap.

 As a follow-up, we can either add a proper `async` shutdown send on the trait or move shutdown onto a dedicated out-of-band signal such as a watch channel.

Longer term, a dedicated out-of-band shutdown signal (e.g. watch channel per pipeline) also gives us a clean reusable `ShutdownHandle` that supervisor and `OpAMP` can call directly - same shutdown path regardless of trigger source.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -1343,6 +1345,118 @@ impl<PData: 'static + Clone + Send + Sync + std::fmt::Debug + ReceivedAtNode + U
...
+                signal(SignalKind::interrupt()).expect("failed to register SIGINT handler");
+
+            tokio::select! {
+                _ = sigterm.recv() => "SIGTERM",
+                _ = sigint.recv() => "SIGINT",
+            }
+        }
+
+        #[cfg(not(unix))]
+        {
```
sapatrjv: Here is the windows equivalent that handles both ctrl_c, ctrl_break. 
#[cfg(windows)]
{
    use tokio::signal::windows::{ctrl_c, ctrl_break};

    let mut sigint = ctrl_c()
        .expect("failed to register Ctrl-C handler");
    let mut sigterm = ctrl_break()
        .expect("failed to register Ctrl-Break handler");

    tokio::select! {
        _ = sigterm.recv() => "CTRL_BREAK (SIGTERM-equivalent)",
        _ = sigint.recv()  => "CTRL_C (SIGINT-equivalent)",
    }
}

sapatrjv: FYI: On windows platform Ctrl+C can't be reliably sent to a process without console handle and it will be ignored. Safest option is to use CTRL_BREAK.

rust/otap-dataflow/crates/controller/src/lib.rs
kamilon: +1, separate channel for control signals makes sense.

rust/otap-dataflow/crates/controller/src/lib.rs
kamilon: additional nit: should be configurable

cijothomas: Done — introduced `SHUTDOWN_TIMEOUT_SECS` named constant. Making it configurable via engine config tracked as follow-up.

rust/otap-dataflow/crates/controller/src/lib.rs
cijothomas: Using `raw_error!` as Josh suggested.

rust/otap-dataflow/crates/controller/src/lib.rs
cijothomas: Added bounded retry (3 attempts with 500ms sleep) inside the async block. Separate out-of-band shutdown channel tracked as future work.

rust/otap-dataflow/crates/controller/src/lib.rs
cijothomas: Done — added `#[cfg(windows)]` block handling both `ctrl_c` and `ctrl_break`.

jmacd (approved):
Thanks for pushing this through, and for the register-once fix. Reusing the signal streams across both waits is the right call to avoid dropping a second signal during the drain.

On the outstanding copilot-reviewer note about the dropped `JoinHandle`, I don't think it's a problem. In Rust the process terminates when `main` returns, and the detached signal-handler thread is reaped at that point rather than keeping the process alive. It would only block exit if something joined it, and nothing does. Your code comments already capture this, so I'm comfortable leaving the thread detached.

The double-signal convention matches the Go Collector, and a 60s drain deadline aligned with the default Kubernetes `terminationGracePeriodSeconds` is a sensible default. LGTM.

One optional and non-blocking follow-up: patch coverage on the new signal paths is on the lighter side, so a small test exercising the second-signal force-exit path would be welcome if it is easy to add.

drewrelmas (approved):

lquerel:
@cijothomas Sorry for removing this PR from the merge queue. I think there are a few issues that need to be addressed first. I'll finish my review soon. Once again, sorry for the delay.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -1789,9 +1798,14 @@ impl<
...
-        // In standard engine mode we keep the main thread parked after startup.
+        // In standard engine mode, keep the main thread blocked until an explicit
+        // engine-wide shutdown drains every pipeline. An empty engine, or one whose
+        // last pipeline stops independently, must remain available for live control.
         if run_mode == RunMode::ParkMainThread {
-            thread::park();
+            // Listen for SIGINT/SIGTERM and trigger graceful shutdown via the
+            // control plane.
+            Self::spawn_shutdown_signal_listener(signal_control_plane);
+            runtime.wait_until_global_shutdown_drains_or_released();
```
lquerel, line 1821: This condition-variable wait is not released by fatal runtime-recovery failures. fail_runtime_recovery still calls controller_thread.unpark(), but the main thread no longer calls thread::park(). If any instance fails to drain, the controller hangs instead of completing teardown and returning the fatal error. You should call release_instance_wait() from that path and add a regression test that verifies the controller returns with an active, non-draining instance.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -2028,6 +2045,114 @@ impl<
...
+            .build()
+            .expect("failed to create signal-handler runtime");
+        let mut signals = {
+            let _runtime_guard = rt.enter();
+            TerminationSignals::register()
+        };
+
+        // The JoinHandle is intentionally dropped -- the thread terminates when
+        // the process exits (after pipelines drain) or on force-exit.
+        drop(
```
lquerel: Dropping this JoinHandle detaches the listener, it does not stop the thread. `run_forever_with_options` can return while the host process continues, leaving a stale control plane and process wide signal listener alive. After a signal driven shutdown, a later signal can therefore be interpreted as the stale listener's second signal and call process::exit(1).

Please make the listener cancellable and cancel/join it during controller teardown.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -1789,9 +1798,14 @@ impl<
...
 
-        // In standard engine mode we keep the main thread parked after startup.
+        // In standard engine mode, keep the main thread blocked until an explicit
+        // engine-wide shutdown drains every pipeline. An empty engine, or one whose
+        // last pipeline stops independently, must remain available for live control.
         if run_mode == RunMode::ParkMainThread {
-            thread::park();
+            // Listen for SIGINT/SIGTERM and trigger graceful shutdown via the
+            // control plane.
+            Self::spawn_shutdown_signal_listener(signal_control_plane);
```
lquerel: Signal handlers are registered only here, after the observability and regular pipeline threads have started. A SIGTERM received during startup still performs the platform's default immediate termination, even though receivers may already be processing data.
Probably better to register signal capture before launching pipelines and retain any early signal until the control plane can initiate coordinated shutdown.

lquerel (changes requested):
The overall direction makes sense, but I found three lifecycle issues that should be addressed before merging:
- Fatal runtime-recovery still calls controller_thread.unpark(), although the main thread now waits on a condition variable. A non-draining instance can therefore prevent teardown indefinitely.
- The signal-listener thread is detached and has no cancellation path. If the controller returns while the host process continues, it leaves stale process-wide signal handling and controller state behind.
- Signal handlers are registered only after pipeline threads have started, so SIGTERM during startup still causes immediate termination and may drop data already being processed.

rust/otap-dataflow/crates/controller/src/lib.rs
cijothomas: Good catch.
That also made me think about the PR's unconditional registration to handle OS signals and process exit. If a host application is embedding us, then I think the host may prefer if host themselves handle the OS signals, instead of the library. I think its best if I also modify the PR to make this configurable, so host apps chose to disable this as well. (and make the listener cancellable ofcourse)

lquerel: Indeed good idea

cijothomas: Fixed in commit [d64c21d16](https://github.com/open-telemetry/otel-arrow/commit/d64c21d16bb93541257fbdd384916d5b256b5bd7).

rust/otap-dataflow/crates/controller/src/lib.rs
cijothomas: Agree. Let me tackle this as a follow up, as this is not simply moving the signal handler launch to earlier, but requires holding on to os-signal until pipelines is in reasonable shape.

cijothomas: https://github.com/open-telemetry/otel-arrow/issues/3829 Opened to track this in follow up.

rust/otap-dataflow/crates/controller/src/lib.rs
cijothomas: Addressed in [commit](https://github.com/open-telemetry/otel-arrow/commit/64bd7dbb6abb937c0daf9bf0f5266051cc3fd1c7)

cijothomas:
> The overall direction makes sense, but I found three lifecycle issues that should be addressed before merging:
> 
> * Fatal runtime-recovery still calls controller_thread.unpark(), although the main thread now waits on a condition variable. A non-draining instance can therefore prevent teardown indefinitely.
> * The signal-listener thread is detached and has no cancellation path. If the controller returns while the host process continues, it leaves stale process-wide signal handling and controller state behind.
> * Signal handlers are registered only after pipeline threads have started, so SIGTERM during startup still causes immediate termination and may drop data already being processed.

Thanks! All of them are addressed now. (The signal handler being registered too late - deferred for [follow up](https://github.com/open-telemetry/otel-arrow/issues/3829) to keep scope in control.)

cijothomas:
> > The overall direction makes sense, but I found three lifecycle issues that should be addressed before merging:
> > 
> > * Fatal runtime-recovery still calls controller_thread.unpark(), although the main thread now waits on a condition variable. A non-draining instance can therefore prevent teardown indefinitely.
> > * The signal-listener thread is detached and has no cancellation path. If the controller returns while the host process continues, it leaves stale process-wide signal handling and controller state behind.
> > * Signal handlers are registered only after pipeline threads have started, so SIGTERM during startup still causes immediate termination and may drop data already being processed.
> 
> Thanks! All of them are addressed now. (The signal handler being registered too late - deferred for [follow up](https://github.com/open-telemetry/otel-arrow/issues/3829) to keep scope in control.)

In latest commit, I made the feature opt-in so hosting applications can chose to opt-in if they want (our example binary enables this) - This was done as the tokio's sigterm handler registration is permanent with no "de-register" option, so I think its best if we default to false.

lquerel (approved):
I approve, assuming that #3833 and #3829 will be addressed in follow-up work.

PR #2326, opened 2026-03-14
feat(engine): auto-nack orphaned pdata on exporter shutdown

### LINKED ISSUE #2315: recv_when(false) does not detect pdata channel closure when buffer has messages

gouslu:
NOTE: This is a draft PR to start a discussion based on a concrete implementation.

## Problem

When an exporter shuts down with unread pdata in its channel buffer (e.g., due to auth outage holding `accept_pdata=false`), those messages were silently dropped. Upstream nodes were never notified the data wasn't exported, violating the ack/nack contract.

If an exporter is not accepting pData, because it would be meaningless to accept it, such as during outage, and a shutdown message is received, the pData that is not processed by the exporter should be automatically nacked to avoid data loss.

Additionally, `recv_when(false)` had a TOCTOU race in its pdata channel closure probe — the old `try_recv()` approach could consume (and lose) a message that arrived between the `is_empty()` guard and the `try_recv()` call.

## Changes

**TOCTOU fix (from #2317):**
- Replace `is_empty() + try_recv()` probe with `is_closed() && is_empty()` — non-consuming check that avoids data loss
- Add `is_closed()` to all channel receiver types (mpsc, mpmc, local, shared)
- Skip the early-exit probe during draining mode (`shutting_down_deadline.is_none()`) to preserve caller-provided deadline/reason

**Auto-nack orphaned pdata:**
- `MessageChannel::shutdown()` now drains any remaining pdata into an internal buffer before dropping the receiver
- `MessageChannel::take_orphaned_pdata()` returns the captured messages
- `Exporter` trait signature changed: `start()` returns `(TerminalState, MessageChannel)` so the engine gets the channel back
- `ExporterWrapper::start()` auto-nacks orphaned pdata via `PipelineControlMsg::DeliverNack`, ensuring upstream nodes are always notified
- In the normal healthy shutdown path (exporter drained everything), the buffer is empty and no nacks are sent — zero cost on the happy path

- Closes #2315

- `test_auth_outage_shutdown_does_not_hang` — verifies `recv_when(false)` with closed+buffered pdata channel returns shutdown
- `test_take_orphaned_pdata_after_shutdown` — fully drained channel → empty orphaned data
- `test_take_orphaned_pdata_with_unread_data` — auth-outage scenario: 2 unread messages captured
- `test_take_orphaned_pdata_empty_after_normal_drain` — normal shutdown → no orphans
- `test_shared_take_orphaned_pdata` — shared MessageChannel variant

`Exporter::start()` trait signature now returns `Result<(TerminalState, MessageChannel), Error>` instead of `Result<TerminalState, Error>`. All 16 exporter implementations updated.

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/exporter.rs
+++ b/rust/otap-dataflow/crates/engine/src/exporter.rs
@@ -442,7 +442,7 @@ impl<PData> NodeWithPDataReceiver<PData> for ExporterWrapper<PData> {
 #[cfg(test)]
 mod tests {
     use crate::Interests;
-    use crate::control::{AckMsg, NodeControlMsg};
+    use crate::control::{AckMsg, Controllable, NodeControlMsg};
     use crate::error::ExporterErrorKind;
     use crate::exporter::{Error, ExporterWrapper};
     use crate::local::exporter as local;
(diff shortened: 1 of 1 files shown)


lquerel:
It seems we keep running into issues around the pdata/control channel complex dynamics and shut down. I think this is a sign of a design problem, or at least an indication that we need a way to validate this more systematically. I’ll think about it.

In the meantime, I think you could add a few additional tests:
- Pre-fill the pipeline control channel to capacity, then shut down an exporter with unread buffered pdata. The expectation should be that the unread pdata is still eventually recovered/NACKed.
- Use a test exporter that receives Shutdown and then returns Err(...) while leaving unread pdata in its input buffer. The expectation should be that those buffered items are still turned into NACKs.

gouslu:
> It seems we keep running into issues around the pdata/control channel complex dynamics and shut down. I think this is a sign of a design problem, or at least an indication that we need a way to validate this more systematically. I’ll think about it.

OK, I am curious to see what you come up with. I have thought about it with the current level of knowledge I have about the engine, and I asked myself what if the channels weren't multiplexed. Multiplexing seems to solve and encapsulate some logic, especially the deadline and the draining; but it also creates other problems.

If we kept those channels separate, let nodes decide what they receive and when they receive it make it their responsibility to implement draining and all that. If exporters don't implement shutdown properly, they will lose data anyway without something like durable buffer. We can provide some shutdown handle to make draining easier and straightforward, and document some best practices.

Currently draining is transparent, so exporter processes pdata until the deadline. But alternative could be that when shutdown is received, they can explicitly drain the pdata channel in the shutdown code (or with some kind of helper/shutdown handle we provide to encapsulate some deadline or other logic).

Basically, I think we should consider a solution that exposes the channels separately and directly to the exporter (or other nodes for that regard), it also gives complex exporters more direct control over their event loop.

lalitb:
> I think this is a sign of a design problem, or at least an indication that we need a way to validate this more systematically.

 Agreed, the repeated issues around pdata/control-channel dynamics do point to a deeper design problem worth revisiting more systematically.

On @gouslu suggestion: exposing channels more directly is worth exploring. The key thing I’d want to preserve is that an engine-provided shutdown handle carries enough of the correctness burden, so individual exporters do not each have to get draining right on their own.

The ownership model feels like the crux. The fact that the engine has to get the channel back after `start()` returns just to recover leftovers is a sign that ownership was never clearly defined to begin with.

gouslu:
> It seems we keep running into issues around the pdata/control channel complex dynamics and shut down. I think this is a sign of a design problem, or at least an indication that we need a way to validate this more systematically. I’ll think about it.
> 
> In the meantime, I think you could add a few additional tests:
> 
> * Pre-fill the pipeline control channel to capacity, then shut down an exporter with unread buffered pdata. The expectation should be that the unread pdata is still eventually recovered/NACKed.
> * Use a test exporter that receives Shutdown and then returns Err(...) while leaving unread pdata in its input buffer. The expectation should be that those buffered items are still turned into NACKs.

Added the tests, they fail as expected. Maybe we shouldn't fix this issue at all and rework this area. The need is clear:

- Components like exporters should be able to pause pdata intake indefinitely (e.g., backpressure while at in-flight capacity, outage etc.) without losing data or blocking shutdown.
- Shutdown should avoid data loss as much as possible. I don't think it is entirely possible, but at least something that does a draining of the channels for us for the orphaned messages would be a good idea on all execution paths.
- Direct access to channels would be preferred to allow for more complex event loops.

PR #2329, opened 2026-03-15
Optimize bitmap usage in `FilterPipelineStage`

albertlockett:
# Change Summary

Last Wednesday's SIG meeting discussion alluded to me that our OTAP query engine probably had some had some performance to be gained in its filtering implementation. I ran the profiler on the benchmarks and, lo and behold, we were spending a lot of time creating `RoaringBitmap`s and checking if IDs were contained therein.

Note - when filtering we were using these `RoaringBitmap`s in two places:
a) when the filtering be an attribute value in a query like `where attributes["x"] = "y"`, we filter the attribute record batch where `key == "x" and str == "y"`, then create a bitmap of the parent_id column, then we scan the parent record batch, determining which values from the `id` column are in this bitmap, and keep only those rows. In these situations, we may also and/or/not the bitmaps together if the query was something like `attributes["x1"] == "y1" and attributes["x2"] == "y2"`.
b) when we're finished with the entire filter execution, we will have determined which rows to keep for the parent batch, and then we need to recursively filter the child record batches. We would do this by creating a bitmap of the ID column on the parent batch, then scanning the child `parent_id` column, checking which rows were in this bitmap and keeping only those rows.

This PR replaces the usage of `RoaringBitmap` with a simpler implementation based on an implementation called `IdBitmap` which contains pages, each containing a flat bitmap of based on an array of u64s.  This is much faster because, unlike `RoaringBitmap`, we never have to do search for the container during insert/contains. While the `RoaringBitmap` is designed for potentially sparse data, our data is dense where in the vast majority of cases the ID columns will be 0...max with few or no gaps, so a more traditional bitmap makes sense here.

This PR also reuses the heap allocations of the bitmap between batches. Note that because we potentially create many of these bitmaps during a filter execution (see note above about logically combining attribute filters), we implement reuse via a pool.

Benchmark results:
| Benchmark | Batch Size | Before | After | Change |
|---|---|---|---|---|
| `simple_field_filter` | 32 | 7.58 µs | 7.61 µs | ~0% |
| `simple_field_filter` | 1024 | 22.32 µs | 14.39 µs | **-35.5%** |
| `simple_field_filter` | 8192 | 140.86 µs | 65.67 µs | **-53.4%** |
| `simple_attr_filter` | 32 | 10.17 µs | 9.93 µs | **-2.4%** |
| `simple_attr_filter` | 1024 | 41.43 µs | 22.75 µs | **-45.1%** |
| `simple_attr_filter` | 8192 | 308.44 µs | 120.45 µs | **-60.9%** |
| `attr_or_attr_filter` | 32 | 12.92 µs | 12.57 µs | **-2.7%** |
| `attr_or_attr_filter` | 1024 | 56.77 µs | 31.04 µs | **-45.3%** |
| `attr_or_attr_filter` | 8192 | 427.42 µs | 167.81 µs | **-60.7%** |
| `attr_and_prop_filter` | 32 | 10.84 µs | 10.92 µs | ~0% |
| `attr_and_prop_filter` | 1024 | 35.98 µs | 21.54 µs | **-40.1%** |
| `attr_and_prop_filter` | 8192 | 263.80 µs | 100.78 µs | **-61.8%** |
| `attr_and_attr_filter` | 32 | 4.32 µs | 4.38 µs | ~0% |
| `attr_and_attr_filter` | 1024 | 13.40 µs | 9.41 µs | **-29.8%** |
| `attr_and_attr_filter` | 8192 | 92.65 µs | 49.49 µs | **-46.6%** |
| `attr_and_or_together_filter` | 32 | 16.31 µs | 16.37 µs | ~0% |
| `attr_and_or_together_filter` | 1024 | 52.78 µs | 34.01 µs | **-35.6%** |
| `attr_and_or_together_filter` | 8192 | 387.77 µs | 164.41 µs | **-57.6%** |
| `or_short_circuit` | 32 | 3.38 µs | 3.25 µs | **-3.8%** |
| `or_short_circuit` | 1024 | 24.19 µs | 8.49 µs | **-64.9%** |
| `or_short_circuit` | 8192 | 127.50 µs | 48.67 µs | **-61.8%** |

(Note that filter benchmarks `attr_and_or_together_filter` & `and_attrs_short_circuit` are omitted from the table because they filter out all the rows before any bitmap operations, so there is no performance change).

* Closes https://github.com/open-telemetry/otel-arrow/issues/2330

Unit tests

No

## Future work/followups

We should probably look at other places we use `RoaringBitmap` for this same purpose to see if we can get some performance benefit from using this new IdBitmap type. For example, I think this same type of procedure is also done in filter processor.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap/filter.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/filter.rs
@@ -214,7 +214,6 @@ impl IdBitmap {
                     }
                     _ => {
                         // No corresponding page in other — zero this page
-                        sp.last_used_generation = self.generation;
                         sp.words.fill(0);
                     }
                 }


lquerel:
So cool. That makes perfect sense!

lquerel (commented):
This seems like a very good optimization for "normal" use cases. However, I think we should add protection against malicious usage where someone could artificially create small batches with just a few entries but IDs close to 2^32, which would cause the system to allocate 512 MiB vectors (leading to potential DoS).

In my opinion, we should consider detecting such situations and then switching to a RoaringBitmap instead, in order to protect the system from these extreme cases.

albertlockett:
> This seems like a very good optimization for "normal" use cases. However, I think we should add protection against malicious usage where someone could artificially create small batches with just a few entries but IDs close to 2^32, which would cause the system to allocate 512 MiB vectors (leading to potential DoS).
> 
> In my opinion, we should consider detecting such situations and then switching to a RoaringBitmap instead, in order to protect the system from these extreme cases.

@lquerel good call - right be to careful!

I've pushed some new changes to try to protect agains this adversarial type of ID column, but it works a bit differently than what you've suggested (I hope that's OK).

I thought two different internal representations of the bitmap would present some challenges. First being that, the logic for choosing representation should be used is somewhat arbitrary and determining that an ID column is indeed sparse adds extra overhead which I didn't want. Also, when we combine bitmaps using and/or/not we need to handle the different combinations of one side being one representation and the other side being the other, which adds complexity that I didn't want.

What I've done is modified the `IdBitmap` implementation be constructed a Vec of optional pages, each containing a heap allocated segment. This is somewhat similar to how `RoaringBitmap` works internally, but it's still faster because we can jump right to the page we're looking to access, and we are never trying to change the representation during insert. This extra layer of indirection does add a small bit of overhead, but it's <2% which is probably a worthwhile trade off given the overall performance improvement this PR introduces (I updated the table of benchmark results in the description).

The pages are allocated on first use and the allocations can be reused across uses of the bitmap. To prevent a really adversarial actor from sending multiple different sparse IDs in different pages and eventually allocating all the memory, the bitmap keeps track of a generation ID of the last time a page was written plus it's own current generation. If a page's generation ID gets too far behind the current generation, the page gets dropped when we clear the bitmap. This helps keep the overall memory usage in check without thrashing the allocated pages.

rust/otap-dataflow/crates/pdata/src/otap/filter.rs
```diff
@@ -29,6 +29,310 @@ const ID_SET_MAX_LENGTH_THRESHOLD: usize = 20;
...
+///
+/// Each page tracks the generation (batch cycle) in which it was last written. On
+/// [`clear`](IdBitmap::clear), the generation counter is incremented and pages are evaluated:
+///
+/// - Pages used within the last [`PAGE_EVICTION_THRESHOLD`] generations are zeroed and retained.
+/// - Pages that haven't been used in more than [`PAGE_EVICTION_THRESHOLD`] generations are
+///   deallocated, preventing unbounded memory growth from adversarial or unusual input patterns.
+///
+/// This means pages that are used regularly (even intermittently) stay allocated, while pages
+/// from one-off anomalous batches are eventually freed.
```
jmacd, line 83: Looks good.

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
```diff
@@ -1359,83 +1431,172 @@ pub(crate) fn filter_otap_batch(
...
+    let result = match otap_batch.root_payload_type() {
         ArrowPayloadType::Logs => {
-            filter_child_batch::<UInt16Type>(&mut otap_batch, ArrowPayloadType::LogAttrs)?;
-            filter_child_batch::<UInt16Type>(&mut otap_batch, ArrowPayloadType::ScopeAttrs)?;
-            filter_child_batch::<UInt16Type>(&mut otap_batch, ArrowPayloadType::ResourceAttrs)?;
+            filter_child_batch::<UInt16Type>(
+                &mut otap_batch,
+                ArrowPayloadType::LogAttrs,
+                &mut id_bitmap,
+            )?;
```
utpilla: nit: The acquired bitmap skips `pool.release()` on error. The `?` operators inside the match arms cause an early return, bypassing `pool.release(id_bitmap)`. The bitmap gets dropped instead of recycled.

`filter_child_batch` only errors on malformed/corrupt batches, which shouldn't occur under normal conditions. But the code's intent is clearly for the release to always run. Maybe worth a minor fix for correctness, e.g. replacing `?` with `.and_then()` chains so errors flow into result without early returning.

Something like:

```rust
ArrowPayloadType::Logs => {
    filter_child_batch::<UInt16Type>(&mut otap_batch, ArrowPayloadType::LogAttrs, &mut id_bitmap)
        .and_then(|()| filter_child_batch::<UInt16Type>(&mut otap_batch, ArrowPayloadType::ScopeAttrs, &mut id_bitmap))
        .and_then(|()| filter_child_batch::<UInt16Type>(&mut otap_batch, ArrowPayloadType::ResourceAttrs, &mut id_bitmap))
        .map(|()| otap_batch)
}
```

albertlockett: Thanks @utpilla , fixed in [2921ee7](https://github.com/open-telemetry/otel-arrow/pull/2329/commits/2921ee70954b03b1a55c3d7c1510d0d37d1b43b9)

This will save us having dropped the `id_bitmap` in the case we get a malformed record batch, which means we won't need to allocate a new one when we filter the next OTAP batch, saving us a few heap allocations!

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
```diff
@@ -1068,22 +1104,22 @@ impl Composite<AttributeFilterExec> {
...
                 } else {
-                    left_result & right_result
+                    left_result.combine_and(right_result, pool)
                 })
             }
             Self::Or(left, right) => {
-                let left_result = left.execute(otap_batch, session_ctx, inverted)?;
-                let right_result = right.execute(otap_batch, session_ctx, inverted)?;
+                let left_result = left.execute(otap_batch, session_ctx, inverted, pool)?;
+                let right_result = right.execute(otap_batch, session_ctx, inverted, pool)?;
```
utpilla, line 1117: Not introduced in this PR but maybe we could have short-circuiting for `Self::Or` as well to avoid evaluating the right side when possible?

rust/otap-dataflow/crates/pdata/src/otap/filter.rs
```diff
@@ -29,6 +29,310 @@ const ID_SET_MAX_LENGTH_THRESHOLD: usize = 20;
...
+                        sp.last_used_generation = self.generation;
+                        for (sw, ow) in sp.words.iter_mut().zip(op.words.iter()) {
+                            *sw &= ow;
+                        }
+                    }
+                    _ => {
+                        // No corresponding page in other — zero this page
+                        sp.last_used_generation = self.generation;
+                        sp.words.fill(0);
+                    }
```
utpilla, line 218: When `other` has no corresponding page, the page in self is zeroed but also stamped with `last_used_generation = self.generation`. This marks the all-zero page as "just used," keeping it alive for another 16 eviction cycles.

Would it make sense to skip the `last_used_generation` update in this arm, so eviction can reclaim the page sooner based on when it was last meaningfully written?

albertlockett: Yeah that seems reasonable @utpilla . made this change in ba26a024c

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
albertlockett: Yeah that's a good idea. We should do that, but I probably won't do as part of this PR. I will create an issue to document

PR #2335, opened 2026-03-16
fix: Use dedicated channel for engine lifecycle events to prevent startup stalls

### LINKED ISSUE #2328: Pipeline cores can get stuck in Pending at startup on high-core machines
### Pre-filing checklist

### Component(s)

## Summary
On 64-core machines, some pipeline cores can get permanently stuck in `Pending` phase at startup.

The immediate failure mode is that the `Admitted` lifecycle event is dropped by the observed-state channel under startup burst, so the state machine never transitions `Pending` -> `Starting`. When `Ready` arrives later for the same core, it is rejected as an invalid transition.

## Cause
At startup, all pipeline cores emit `Admitted` events into a shared `flume::bounded(100)` channel. Engine events use `send_timeout(1ms)`. Under startup burst on a 64-core machine, the single consumer does not always drain fast enough, and some Admitted events are dropped.

The corresponding `Ready` event for the same core arrives later, while that core is still in `Pending`, producing:

```
InvalidTransition { phase: Pending, event: Success(Ready) }
```

## Repro

64-core VM, single pipeline, `core_allocation: all_cores`. Repeated startup/shutdown cycles showed intermittent failure:

| Run  | Cores | Dropped `Admitted` |
|------|-------|--------------------|
| 1    | 59/64 | 5                  |
| 5    | 60/64 | 4                  |
| 6    | 62/64 | 2                  |
| rest | 64/64 | 0                  |

3/10 runs affected.

## Logs

Dropped lifecycle event:

```
ERROR otap-df-telemetry::Timeout, dropping observed event:
  [event=Engine(EngineEvent { core_id: 49, type: Success(Admitted) })]
```

Followed by state-machine failure for the same startup:

```
ERROR otap-df-state::state.report_failed:
  [error=InvalidTransition { phase: Pending, event: Success(Ready) }]
```

In the repro runs, dropped `Admitted` count matched `InvalidTransition` count and matched the number of stuck cores.

## Potential fixes:

**Primary fix:** Make engine lifecycle events (Admitted, Ready, etc.) non-lossy. These events should not use a drop-on-timeout path. Possible approaches:

- blocking send for engine lifecycle events
- or a separate reliable channel for engine lifecycle events

**Mitigation:** Increase `reporting_channel_size` from 100 to 256+ to reduce repro frequency. But this does not eliminate the underlying bug because engine lifecycle events are still sent through a lossy timeout-based path.

### Steps to Reproduce

-

### Expected Behavior

All cores should be used.

### Actual Behavior

Some cores are in stuck state.

### OTel-Arrow Version

latest

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

lalitb:
# Change Summary

Fixes a startup race where pipeline cores could get stuck in `Pending` on high-core machines.                                                                     
 
Engine lifecycle events (`Admitted`, `Ready`) shared the same fixed-size bounded observed-state channel as lossy async log events. Under startup burst, `Admitted` could be dropped when `send_timeout(1ms)` expired. When `Ready` arrived later, the state machine rejected the `Pending -> Ready` transition as invalid, leaving the core stuck.

This change separates reliability classes: engine lifecycle events now go through a dedicated unbounded channel, while async log events stay on the existing bounded lossy path.

The unbounded channel is intentional: engine events are low-volume, correctness-critical, and naturally bounded by pipeline/core lifecycle activity.

**Alternate approaches considered:** 

- Increasing the bounded channel size would only reduce the probability of failure under burst; it would not guarantee delivery of lifecycle events.
- Making the state machine accept `Pending` -> `Ready` would mask dropped lifecycle events instead of fixing delivery.

* Closes #2328 

```bash
$ cargo test -p otap-df-state -- --nocapture
$ cargo check -p otap-df-controller -p otap-df-state -p otap-df-telemetry -p otap-df-config
```

No


rust/otap-dataflow/crates/telemetry/src/event.rs
```diff
@@ -11,22 +11,69 @@ use std::fmt;
...
-        self.observe(ObservedEvent::Engine(event))
+        if let Some(engine_sender) = &self.engine_sender {
+            if let Err(e) = engine_sender.send(event) {
+                crate::raw_error!(
+                    "Engine channel disconnected, dropping engine event",
+                    event = ?e.into_inner()
+                );
+            }
+        } else {
+            // Fallback: lossy path (used in tests and backward-compat scenarios).
```
jmacd, line 74: Should we warn the user or debug_assert when the fallback happens?

lalitb:
The intent behind the unbounded engine-events channel is to make lifecycle delivery reliable. These are low-volume control-plane events, not a high-throughput stream like logs, so the expected queue size is small in normal operation.

I agree there is still a tradeoff: if the consumer stalls, an unbounded queue can grow. For this fix, I think that is acceptable, because the alternative is reintroducing a bounded path where correctness-critical lifecycle events can be delayed or dropped under burst.

I'd be happy to add some visibility here, for example sender-side backlog tracking or watchdog-style monitoring, so a stuck consumer becomes visible without making lifecycle delivery lossy again.

jmacd:
I wonder if we should have a control-channel abstraction that is bounded independently by event kind. If we limit to at most K (>=1) events per kind of event in the channel, then there is a bound of K*N for at most N kinds of event.

utpilla:
> The intent behind the unbounded engine-events channel is to make lifecycle delivery reliable. These are low-volume control-plane events, not a high-throughput stream like logs, so the expected queue size is small in normal operation.
> 
> I agree there is still a tradeoff: if the consumer stalls, an unbounded queue can grow. For this fix, I think that is acceptable, because the alternative is reintroducing a bounded path where correctness-critical lifecycle events can be delayed or dropped under burst.
> 
> I'd be happy to add some visibility here, for example sender-side backlog tracking or watchdog-style monitoring, so a stuck consumer becomes visible without making lifecycle delivery lossy again.

@lalitb I don't think the tradeoff is between unbounded (reliable) vs. bounded (lossy). The PR already introduced a blocking `send()` on the engine channel which means switching to a bounded channel gives you reliable delivery and a memory ceiling with no send-side code change.

- `flume::Sender::send()` **on a bounded channel** blocks until space is available. It never drops.
- The `biased` select already prioritizes draining engine events, so any producer blocking would be extremely brief.
- Even if a producer briefly blocks, there's no data loss concerns. Lifecycle events like `Admitted` and `Ready` are sent before the pipeline starts processing data.

lalitb:
> I don't think the tradeoff is between unbounded (reliable) vs. bounded (lossy).

@utpilla  You’re right that a bounded channel would still be reliable with the current `send()` path, so the tradeoff is not loss vs. no loss.

The reason I still chose unbounded here is that send() is synchronous. With a bounded channel, if the store's event handler is slow or the store task falls behind, that backpressure propagates directly into the threads calling lifecycle paths. With unbounded, the failure mode is backlog growth instead of blocking those control-plane paths. That may hold for the current lifecycle events, but it is still an assumption about where and when those events are emitted, and that is easy to violate as new events are added. With unbounded, the producer stays decoupled from consumer speed regardless of when the event is emitted.

> I wonder if we should have a control-channel abstraction that is bounded independently by event kind. If we limit to at most K (>=1) events per kind of event in the channel, then there is a bound of K*N for at most N kinds of event.

@jmacd  I think this is the interesting bounded design to consider. We can have a broader follow-up on that direction; this PR is trying to make the lifecycle path reliable with the smallest behavioral change.  If we do move to a bounded design in a follow-up, I think the bound and its blocking semantics need explicit discussion rather than picking an arbitrary size here.

PR #2323, opened 2026-03-13
chore: Migrate topics, internal_telemetry_receiver, and perf_exporter…

drewrelmas:
… to core-nodes crate

# Change Summary

Next part of #1847 and #2086

Moves:

* perf_exporter
* topic_exporter
* internal_telemetry_receiver
* topic_receiver

## How are these changes tested?

* Unit tests / CI
* Compiled and ran `df_engine` and confirmed all nodes are still available

## Are there any user-facing changes?

No

PR #932, opened 2025-08-13
[otap-dataflow] move otap datagen test files to fixtures.rs and perf exporter test refactor

clhain:
Fixes #927.

This PR moves the parquet_exporter/test/datagen.rs code into the main otap library for shared test fixtures.

- Added header generation via options to all create_simple_*_arrow_record_batches functions
- Added config option for timestamp header generation (default true)
- Refactor of the perf_exporter tests to leverage the new fixtures.

PR #2339, opened 2026-03-16
chore: Migrate syslog_cef_receiver and parquet_exporter to core-nodes crate

drewrelmas:
# Change Summary

Next part of https://github.com/open-telemetry/otel-arrow/issues/1847 and https://github.com/open-telemetry/otel-arrow/issues/2086

* Unit tests / CLI
* Compiled and ran `df_engine` and confirmed all nodes are still
available

No


rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
```diff
@@ -322,7 +322,7 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
                         // Process incoming TCP connections.
                         accept_result = listener.accept() => {
                             match accept_result {
-                                Ok((socket, peer_addr)) => {
+                                Ok((socket, _peer_addr)) => {
```
utpilla: Did `clippy` or `cargo fmt` force you to add `_` as a prefix?

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/config.rs
```diff
@@ -380,7 +380,7 @@ fn default_resource_weight() -> NonZeroU32 {
 ///
 /// Each entry at position `i` contributes `entry.weight` copies of `i` to the
 /// table.  The hot path is then:
-/// ```
+/// ```text
```
drewrelmas, line 383: Random doc_test failure in local CI, just fixing in this PR too.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
drewrelmas: Reverted this - it was actually an artifact from asking Copilot to do local migration of this node following pattern of previous moves. It seems to prefer the `_` prefix compared to the other method of suppression.

Since the purpose of this PR is just migration, amended to focus change as much as possible on only required using statements.

rust/otap-dataflow/crates/core-nodes/src/exporters/parquet_exporter/fixtures.rs
```diff
@@ -1,7 +1,6 @@
 // Copyright The OpenTelemetry Authors
```
drewrelmas, line 1: @albertlockett I'm not very happy with how this iteration has moved `fixtures.rs` to `parquet_exporter`. `otap/src/object_store.rs` is in a similar situation - it has generic code that only `parquet_exporter` depends on at the moment. Any good ideas? 

I was trying to move all the nodes first and only then worry about 'shared' code organization, but there are some cross-crate test dependencies that make it a little harder.

albertlockett: Yeah I actually think it's fine to have move fixtures under parquet_exporter. We originally moved them out so the perf_exporter could rely on them as part of https://github.com/open-telemetry/otel-arrow/pull/932/ but since https://github.com/open-telemetry/otel-arrow/pull/2323 the tests in perf_exporter no longer rely on these fixtures.

For `object_store` I think I'd like to keep this as shared code as opposed to something only used by parquet. I can imagine other exporters might want to rely on this code, for example if we ever had an exporter that wrote to a different rust data/table format (like LanceDB of Vortex)

PR #2343, opened 2026-03-16
feat: Feature-gate weaver behind dev-tools to eliminate transitive ring dependency

### LINKED ISSUE #2322: Feature-gate weaver behind optional dev-tools flag to eliminate transitive ring dependency
### Pre-filing checklist

### Component(s)

## Summary

The `weaver` crates (`weaver_forge`, `weaver_semconv`, `weaver_resolved_schema`,
`weaver_resolver`, `weaver_common`) are regular dependencies of `otap-df-core-nodes`
and `otap-df-otap`, used to power the `fake_data_generator` receiver — a
dev/testing/benchmarking tool for generating synthetic telemetry data. They should
be gated behind an optional feature flag (e.g., `dev-tools`, included in default
features for backward compatibility) so that production builds can opt out and
avoid weaver's transitive `ring` dependency.

See https://github.com/open-telemetry/otel-arrow/pull/2269#issuecomment-4050756671

## Motivation

After #2269 (pluggable crypto providers), the only remaining transitive `ring`
path within the project's control is through weaver:

Feature-gating weaver would fully eliminate `ring` from `crypto-openssl`
production builds (built with `--no-default-features`).

(`opentelemetry-otlp` is the other transitive path, but will be removed once the
internal telemetry pipeline refactor is complete — tracked separately.)

### Proposed Solution

- **`crates/core-nodes`**: Make `weaver_*` dependencies optional, add `dev-tools`
  feature flag, gate `fake_data_generator` module (including its `linkme`
  `distributed_slice` registration) behind `#[cfg(feature = "dev-tools")]`
- **`crates/otap`**: Move `weaver_*` from regular dependencies to dev-dependencies
  (only used in `tests/pipeline_tests.rs`, not in source)
- **`crates/validation`**: Move `weaver_*` from regular dependencies to
  dev-dependencies (only used in `#[cfg(test)]` in `encode_decode.rs`); add
  `core-nodes` dev-dependency with `dev-tools` feature for test access to
  `fake_data_generator`
- **Root `Cargo.toml`**: Add `dev-tools` feature forwarding to
  `otap-df-core-nodes/dev-tools`, include in `default` features
- No Dockerfile changes needed — it already uses explicit `--features` (not
  defaults), so `dev-tools` is excluded from production images unless passed

### Alternatives Considered

_No response_

### Additional Context

_No response_

### LINKED ISSUE #2358: Include a --no-default-features test in build workflow
### Pre-filing checklist

### Component(s)

### Bug Description

As noted in #2343, we are not testing for broken features in the continuous build.

https://github.com/open-telemetry/otel-arrow/pull/2343#issuecomment-4077077305

### Steps to Reproduce

Follow the steps in #2343 for example.

### Expected Behavior

> The build should in general work with/without any combination of feature flags. This should have been caught by the CI - we should be add at least one --no-default-features check in CI for these targets so the optional-feature shape is validated too (in future PRs).

### Actual Behavior

> Both fail because fake_data_generator is now gated behind core-nodes/dev-tools, but validation and benchmarks still import it from normal code without enabling that feature. So if fake_data_generator is intended to be optional, these crates need to opt into the new feature as well, or gate their own usage accordingly.

### OTel-Arrow Version

head

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

pritishnahar95:
# Change Summary

Feature-gate the `weaver` crates behind an optional `dev-tools` feature flag to
eliminate the transitive `ring` dependency from production builds. The `weaver`
crates are only used by the `fake_data_generator` (traffic generator) receiver —
a dev/testing/benchmarking tool. By making them optional, builds with
`--no-default-features --features jemalloc,crypto-openssl` no longer pull in
`ring` through weaver.

`dev-tools` is included in default features for backward compatibility, so
existing builds, benchmarks, and configs are unaffected.

* Closes #2322
* Closes #2358

- `cargo check` with default features (includes `dev-tools`) — passes, all
  existing functionality preserved.
- `cargo check --no-default-features --features jemalloc,crypto-openssl` — passes,
  verifies the binary builds without weaver/fake_data_generator.
- Added a `no_default_features_check` CI job that runs `cargo check --no-default-features`
  against the workspace, validation crate, and benchmarks on every PR to prevent
  regressions (closes #2358).

Yes:

- **New feature flag**: `dev-tools` (enabled by default) controls whether the
  `fake_data_generator` / `traffic_generator` receiver is included in the build.
- **Default behavior is unchanged** — `cargo build` still includes the traffic
  generator and all dev tooling.
- **Builds** that want to eliminate `ring` dependency can now build with:
  ```
  cargo build --no-default-features --features <feature1>,<feature2>
  ```
  This excludes weaver and its transitive `ring` dependency entirely.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -607,6 +607,7 @@ jobs:
       - docs_required
       - structure_check
       - pest-fmt
+      - no_default_features_check
     steps:
       - name: Check if all required jobs succeeded
         run: |
@@ -636,6 +637,10 @@ jobs:
           fi
           if [[ "${{ needs.pest-fmt.result }}" != "success" ]]; then
             echo "pest-fmt failed or was cancelled"
+            exit 1
+          fi
+          if [[ "${{ needs.no_default_features_check.result }}" != "success" ]]; then
+            echo "no_default_features_check failed or was cancelled"
             exit 1
           fi
           echo "All required checks passed!"


rust/otap-dataflow/crates/core-nodes/src/receivers/mod.rs
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /// Fake data generator receiver.
+#[cfg(feature = "dev-tools")]
```
cijothomas, line 5: fake_data_generator has 2 modes internally - one of them uses weaver, but the other is self-contained. We only need to feature gate the weaver one with this feature flag.

pritishnahar95: Good point — the static mode is self-contained and doesn't need weaver. Two options:

Whole-module approach (current): Gate the entire `fake_data_generator` module behind #[cfg(feature = "dev-tools")]. One line, zero maintenance overhead, but loses the static traffic generator in `--no-default-features` builds. Note: dev-tools is included in default features, so this only affects explicit --no-default-features builds.

Sub-module approach: Move weaver source files into a `semconv` sub-module. The static mode, core dispatch, and linkme registration stay always-available. However, `config.rs` and `mod.rs` still need `#[cfg]` annotations in multiple places. I've prototyped this and it compiles, but the `#[cfg]` spread is non-trivial and future changes to the `semconv or static` dispatch will need to be aware of the feature gating.

I'd lean toward option 1 for simplicity unless keeping the static generator available in `--no-default-features` builds is important. Happy to go either way — what do you think?

jmacd: I agree with option 1. The fake data generator is I think a developer tool and I expect we will build with the necessary features for benchmarking.

cijothomas: "dev-tools" - the name-- are there other dev tools planned? if yes, then this is fine. If not, we can just guard with "fake-generator" feature flags. Not a blocker.

rust/otap-dataflow/Cargo.toml
```diff
@@ -218,6 +218,10 @@ unsafe-optimizations = ["unchecked-index", "unchecked-arithmetic"]
 unchecked-index = []
 unchecked-arithmetic = []
 
+# Dev/test tooling (fake data generator). Disabled in production builds to
```
cijothomas: What does "Disabled in production builds" mean here?

rust/otap-dataflow/crates/core-nodes/src/receivers/mod.rs
cijothomas: resolving this conversation. Agree with feature-flagging entire fake-generator. The name is not a blocker.

jmacd: I have mixed feelings--I like the name "dev-tools" and I agree we should be more specific. The problem with "fake-generator" is I'll never remember it, I'll type "fake-data-generator" or "fake_data_generator".

I was thinking "weaver-tools" might be better, but I don't think this is a big deal. The ring dependency is a big deal!

rust/otap-dataflow/Cargo.toml
pritishnahar95: Removed that line, it's confusing.

lalitb:
I did a quick check and these feature-off builds fail:

```bash
$ cargo check -p otap-df-validation --lib --no-default-features
error[E0433]: failed to resolve: could not find `fake_data_generator` in `receivers`
  --> crates/validation/src/traffic.rs:12:36
   |
12 | use otap_df_core_nodes::receivers::fake_data_generator::config::DataSource;
   |                                    ^^^^^^^^^^^^^^^^^^^ could not find `fake_data_generator` in `receivers`
   |
note: found an item that was configured out
  --> /Users/lalitb/work/obs/otel/rust/otel-arrow/rust/otap-dataflow/crates/core-nodes/src/receivers/mod.rs:6:9
   |
 5 | #[cfg(feature = "dev-tools")]
   |       --------------------- the item is gated behind the `dev-tools` feature
 6 | pub mod fake_data_generator;
   |         ^^^^^^^^^^^^^^^^^^^


$ cargo check -p benchmarks --benches --no-default-features
    Checking benchmarks v0.1.0 (/Users/lalitb/work/obs/otel/rust/otel-arrow/rust/otap-dataflow/benchmarks)
error[E0433]: failed to resolve: could not find `fake_data_generator` in `receivers`
  --> benchmarks/benches/transport_optimize/main.rs:16:36
   |
16 | use otap_df_core_nodes::receivers::fake_data_generator::config::{Config, TrafficConfig};
   |                                    ^^^^^^^^^^^^^^^^^^^ could not find `fake_data_generator` in `receivers`
   |
note: found an item that was configured out
  --> /Users/lalitb/work/obs/otel/rust/otel-arrow/rust/otap-dataflow/crates/core-nodes/src/receivers/mod.rs:6:9
   |
 5 | #[cfg(feature = "dev-tools")]
   |       --------------------- the item is gated behind the `dev-tools` feature
 6 | pub mod fake_data_generator;
   |         ^^^^^^^^^^^^^^^^^^^

error[E0433]: failed to resolve: could not find `fake_data_generator` in `receivers`
  --> benchmarks/benches/transport_optimize/main.rs:17:36
   |
17 | use otap_df_core_nodes::receivers::fake_data_generator::semconv_signal::semconv_otlp_logs;
   |                                    ^^^^^^^^^^^^^^^^^^^ could not find `fake_data_generator` in `receivers`
   |
note: found an item that was configured out
  --> /Users/lalitb/work/obs/otel/rust/otel-arrow/rust/otap-dataflow/crates/core-nodes/src/receivers/mod.rs:6:9
   |
 5 | #[cfg(feature = "dev-tools")]
   |       --------------------- the item is gated behind the `dev-tools` feature
 6 | pub mod fake_data_generator;
   |         ^^^^^^^^^^^^^^^^^^^


```

Both fail because `fake_data_generator` is now gated behind `core-nodes/dev-tools`, but `validation` and `benchmarks` still import it from normal code without enabling that feature. So if `fake_data_generator` is intended to be optional, these crates need to opt into the new feature as well, or gate their own usage accordingly.

The build should in general work with/without any combination of feature flags. This should have been caught by the CI - we should be add at least one `--no-default-features` check in CI for these targets so the optional-feature shape is validated too (in future PRs).

jmacd:
https://github.com/open-telemetry/otel-arrow/issues/2358

@pritishnahar95 have we fixed the broken feature combination problem @lalitb described?

pritishnahar95:
> #2358
> 
> @pritishnahar95 have we fixed the broken feature combination problem @lalitb described?

@jmacd The broken feature combinations are fixed and benchmarks now correctly opt into core-nodes/dev-tools. I've validated locally that the following all pass:

```
cargo check --no-default-features --workspace
cargo check -p otap-df-validation --lib --no-default-features
cargo check -p benchmarks --benches --no-default-features
```

For the CI side (adding a --no-default-features check to the workflow to catch this automatically), need to add a CI test job to validate this.

Edit/Update: Added the CI job, succeeds: https://github.com/open-telemetry/otel-arrow/actions/runs/23222305570/job/67497432374?pr=2343

rust/otap-dataflow/crates/core-nodes/src/receivers/mod.rs
lquerel: Side comment.

The urn of this receiver is `urn:otel:receiver:traffic_generator` but we still use internally the name `FakeDataGenerator` which doesn't help :-)

At some point we should rename the struct `FakeDataGenerator` into `TrafficGen`.

Issue #2289, opened 2026-03-12
Enforce that payloads in OtapArrowRecords conform to the spec


JakeDern:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Description

# Problem

OtapArrowRecords and the underlying OtapBatchStore implementations are the fundamental storage unit for otap record batches.

As noted in #2022 and outlined in the otap-spec, there are many constraints determine what a valid OTAP Record Batch looks like, but there are no guarantees provided by OtapArrowRecords that the contents obey these constraints.

This means that downstream components consuming OtapArrowRecords must be aware of the spec and program defensively. This introduces two major problems:

1. Reimplementing similar validations repeatedly is error prone and requires downstream consumers have knowledge of very lengthy spec
2. It introduces fallibility into many operations which otherwise should not be able to fail unless data is malformed

There are many examples of this, but here are two of them:

1. A simple operation to produce the values array for an id column is fallible because we don't have a guarantee that the columns are not malformed: https://github.com/open-telemetry/otel-arrow/blob/4fde1cc67a8724f2948cb80637e90546db7849f5/rust/otap-dataflow/crates/pdata/src/otap/transform/reindex.rs#L810-L839
2. A simple transform operation looking for monotonically increasing sequences is similarly fallible just in case there is an id data type mismatch: https://github.com/open-telemetry/otel-arrow/blob/4fde1cc67a8724f2948cb80637e90546db7849f5/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L649-L684

## Proposal

OtapArrowRecords/OtapBatchStore should be the place to centrally validate basic spec conformance and downstream consumers of these constructs should be able to take that conformance for granted. 

With the introduction of payload definitions in #2240, we have the foundations for enforcing the spec. This does, however, mean that we need to tighten up direct access to the underlying storage and treat bypassing validations as an _unsafe_ operation.

At least two major operations will likely be affected:

1. [set](https://github.com/open-telemetry/otel-arrow/blob/4fde1cc67a8724f2948cb80637e90546db7849f5/rust/otap-dataflow/crates/pdata/src/otap.rs#L48-L54) operations on both OtapArrowRecords and OtapBatchStore will become fallible. These have many callsites.
[...]

PR #2346, opened 2026-03-16
feat: OTAP Schema construct and complete definitions

JakeDern:
# Change Summary

I previously added payload definitions in #2240 which were mostly focused on solving the dictionary key size problem. This PR builds on that work, but with a more refined implementation that also includes things like required vs optional columns as a pre-req for #2289.

The major changes:

- Added an otap `Schema` type and redefined the payloads according to that. This is a better construct that is somewhat symmetrical to arrow Schemas, allows for defining recursive Struct and List types, and subsequently removes the requirement to have nested lookup tables.
- Added deep equality checks between record batches and `Schema` types
- Updated some small bits of the `transform` module for the changes 

* Part of #2289

Unit

No


JakeDern:
I think an exporter test flaked

PR #2349, opened 2026-03-17
feat: Improve pdata::otap::testing module and expose it to other crates

JakeDern:
# Change Summary

This is another PR on the path to #2289. During my initial implementation, I realized that a lot of tests were generating invalid otap batches. In order to make that easier, I've updated the logs/metrics/traces macros in the pdata crate to automatically fill in anything required to be spec compliant based on what's been specified. I had a half-baked version of this before, but now that we have the spec we can do this properly.

Additional changes:

- Added an `Into<OtapArrowRecords>` trait bound to OtapBatchStore. From<Logs/Metrics/Traces> was already implemented, it's just useful to have in the bound. 
- added a `testing` feature which lets these be consumed across other crates. 

* Part of #2289

Unit.

No


JakeDern:
This is a stacked diff on #2346 and most commits belong to that PR. Will publish this after that goes in

PR #2351, opened 2026-03-17
Disable Renovate updates of indirect python lock dependencies

drewrelmas:
# Change Summary

As noted in [this workflow run](https://github.com/open-telemetry/otel-arrow/actions/runs/23181837916/job/67356323365?pr=2306) - recent Renovate updates to some python dependencies for perf testing didn't work properly.

#2336 updated `pydantic-core` from `2.41.5` to `2.42.0`. However, this is an indirect dependency of `pydantic`, which is a direct dependency in the requirements file:

https://github.com/open-telemetry/otel-arrow/blob/59ef72fdbe2003a1425bb5c700d3de0579ffb050/tools/pipeline_perf_test/orchestrator/requirements.txt#L6

`pydantic` `2.12.5` requires EXACTLY `pydantic-core` `2.41.5`, so it was a bad Renovate update.

Based on Renovate docs, we should be able to disable indirect dependency updates like this by matching on `matchDepTypes: ["indirect"]` and disabling.

In addition, it seems we have some mismatch between python versions used in the repo: `3.11` and `3.14`. This can also lead to bad side effects if lock files are generated with a version different from the one being used at runtime during workflow runs.


drewrelmas:
Also seeing if @cijothomas or @clhain would approve

PR #2355, opened 2026-03-17
add resource id header to ame exporter requests

gouslu:
# Change Summary

If configured, sets resource id header for log analytics API.

New config field `azure_monitor_source_resourceid` under `api` config section of azure monitor exporter.

### DIFF
--- a/.github/workflows/pipeline-perf-test-continuous.yml
+++ b/.github/workflows/pipeline-perf-test-continuous.yml
@@ -18,6 +18,15 @@ jobs:
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
+      - name: System topology (NUMA, CPU, memory)
+        run: |
+          echo "### System Topology" >> $GITHUB_STEP_SUMMARY
+          echo '```' >> $GITHUB_STEP_SUMMARY
+          lscpu | grep -E "Model name|Socket|Core|Thread|NUMA|CPU\(s\)" | tee -a $GITHUB_STEP_SUMMARY
+          echo "" | tee -a $GITHUB_STEP_SUMMARY
+          numactl --hardware 2>/dev/null | tee -a $GITHUB_STEP_SUMMARY || echo "numactl not available"
+          echo '```' >> $GITHUB_STEP_SUMMARY
+
       - id: detect_os
         name: Detect OS (self-hosted)
         shell: bash
@@ -71,7 +80,7 @@ jobs:
           git submodule init
           git submodule update
           cd rust/otap-dataflow
-          docker build --build-context otel-arrow=../../ -f Dockerfile -t df_engine .
+          docker buildx build --load --build-context otel-arrow=../../ -f Dockerfile -t df_engine .
           cd ../..
 
       - name: Install dependencies
--- a/.github/workflows/pipeline-perf-test-nightly.yml
+++ b/.github/workflows/pipeline-perf-test-nightly.yml
@@ -72,7 +72,7 @@ jobs:
           git submodule init
           git submodule update
           cd rust/otap-dataflow
-          docker build --build-context otel-arrow=../../ -f Dockerfile -t df_engine .
+          docker buildx build --load --build-context otel-arrow=../../ -f Dockerfile -t df_engine .
           cd ../..
 
       - name: Build otelarrowcol
--- a/rust/otap-dataflow/Dockerfile
+++ b/rust/otap-dataflow/Dockerfile
@@ -6,7 +6,7 @@
 # In order to include these in the build context, and to avoid conflicting
 # with the arrow enabled go collector build files in the repo root, the image
 # must be built with additional --build-context flags as follows:
-# docker build --build-context otel-arrow=../../ -f Dockerfile -t df_engine .
+# docker buildx build --load --build-context otel-arrow=../../ -f Dockerfile -t df_engine .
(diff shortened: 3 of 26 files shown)


rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame.yaml
```diff
@@ -41,6 +41,7 @@ groups:
                 dcr_endpoint: "https://my-workspace.eastus-1.ingest.monitor.azure.com"
                 stream_name: "Custom-MyTable_CL"
                 dcr: "dcr-abc123"
+                azure_monitor_source_resourceid: "/subscriptions/<subscription-id>/resourceGroups/<resource-group>/providers/Microsoft.Kubernetes/connectedClusters/<cluster-name>/providers/microsoft.kubernetesconfiguration/extensions/<extension-name>" # Azure Specific - Optional
```
cijothomas, line 44: this is not required, if the env variable `AZURE_MONITOR_SOURCE_RESOURCEID` is set as it'll automatically gets applied?

drewrelmas: @gouslu Per Cijo's comment - I think perhaps config only is sufficient instead of also having the env variable method. It provides a more formal contract for consumers.

I don't believe we have any other nodes that rely on environment variables in place of configuration, this would be a new pattern.

cijothomas: yes `AZURE_MONITOR_SOURCE_RESOURCEID` feels like a magic env variable, not really a standard one.

gouslu: I added env variable as a backup just in case it'd be harder to implement for the config in our use case, but I also agree that it should be config based.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
```diff
@@ -588,4 +607,25 @@ mod tests {
             HeaderValue::from_static("Bearer new_token")
         );
     }
+
+    // ==================== Resource ID Header Tests ====================
+
+    #[test]
```
drewrelmas, line 613: nit: are these duplicate tests needed in `heartbeat`? They seem to be almost the same as in `client` tests?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
```diff
@@ -95,6 +95,9 @@ pub struct ApiConfig {
     /// Schema mapping configuration
     #[serde(default)]
     pub schema: SchemaConfig,
+
+    /// Arm Resource ID header for the logs exported to Azure Monitor (optional)
+    pub azure_monitor_source_resourceid: Option<String>,
```
cijothomas, line 100: can we default to None, so all the other places can avoid explicitly passing None.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame.yaml
cijothomas: https://github.com/open-telemetry/otel-arrow/pull/2176 added support for pulling config values from env variable, it could be leveraged, if needed env variable support.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
gouslu: yeah, they have both their own client setups, so need to be tested.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
gouslu: I am not sure I understand the ask here, can you explain?

drewrelmas: @gouslu see https://github.com/open-telemetry/otel-arrow/blob/4ec2c6406d0fac4d7877ce8ab0cf3b3806eb9772/rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/config.rs#L114-L115

You can define default value as `None` when not provided. Will this work?

gouslu: it doesn't seem to work that way, serde default sets a default value for serialization, but for fields with Option<> it seems to be not necessary.

gouslu: can this be merged?

drewrelmas: Will merge it as-is, we can revisit style-wise but I don't see an obvious improvement at the moment.

PR #2356, opened 2026-03-17
feat: Enforce basic OTAP Spec compliance at the OtapBatchStore level

### LINKED ISSUE #2289: Enforce that payloads in OtapArrowRecords conform to the spec

JakeDern:
# Change Summary

This PR implements some basic spec enforcement as described in #2289. It doesn't catch harder to detect things like duplicate primary ids, but it does account for making sure all fields match the spec and that there are no extraneous fields.

Majority of this PR is various tweaks to different modules that depended on setting invalid record batches being possible in tests or having the operation be infallible, but there are some major updates to be aware of:

- I introduced the concept of RawBatchStore which now underpins OtapBatchStore. Validations are applied at the OtapBatchStore level and raw operations are generally unchecked and/or can panic. I think this is actually a good direction for slimming down the public contract for OtapBatchStore and there's probably more to explore here.
- Updated the parquet exporter which was using OtapBatchStore as storage to use a new construct OtapParquetRecords which wraps RawBatchStore types becuase it needs to be able to do stuff like widen out the id columns
- Split schema errors out into their own type
- The `record_bundle` tests needed a little more reworking than was typical elsewhere, I also consolidated a function into a generic there.

* Closes #2289 

Unit.

No.


rust/otap-dataflow/crates/pdata/src/otap.rs
```diff
@@ -180,26 +180,39 @@ impl From<Traces> for OtapArrowRecords {
     }
 }
 
+pub(crate) mod sealed {
+    use arrow::array::RecordBatch;
+
+    /// Sealed trait providing access to less safe operations within this crate only.
+    pub trait OtapBatchStore {
```
lquerel, line 208: Cool

JakeDern: You beat me to the punch on starting a discussion on this point 🙂.

I also played around with the idea of marking this as unsafe, but didn't like that approach because this is not actually a soundness/safety issue (at least not today).

lquerel:
@JakeDern  Should we turn this PR into "Ready for review" mode?

JakeDern:
> @JakeDern Should we turn this PR into "Ready for review" mode?

Nearly there, just working through a last couple of tests that need updating.

rust/otap-dataflow/crates/pdata/src/otap/raw_batch_store.rs
```diff
@@ -0,0 +1,340 @@
...
+///
+/// The `TYPE_MASK` const generic is a bitmask indicating which
+/// [`ArrowPayloadType`] values are valid for this store. The `COUNT` const
+/// generic is the number of slots in the backing array.
+///
+/// This type provides **no** OTAP schema validation. Callers that need
+/// validation should use the [`OtapBatchStore`](super::OtapBatchStore) trait
+/// implementations which wrap this type.
+#[derive(Clone, Debug, PartialEq)]
+pub struct RawBatchStore<const TYPE_MASK: u64, const COUNT: usize> {
```
JakeDern, line 157: CC: @lquerel @albertlockett - This is likely of interest to you and was added late

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/bundle_adapter.rs
```diff
@@ -524,91 +524,58 @@ pub fn convert_bundle_to_pdata(
...
-        }
-    }
-
-    OtapArrowRecords::Traces(traces)
-}
-
-/// Create Metrics records from payloads.
-fn create_metrics_records(payloads: &HashMap<SlotId, RecordBatch>) -> OtapArrowRecords {
-    let mut metrics = Metrics::default();
+fn create_records<T>(
```
JakeDern, line 537: FYI: @AaronRM - I consolidated a couple of functions here and also had to update the tests to only use valid otap data. Happy to change anything here according to your preference!

rust/otap-dataflow/crates/core-nodes/src/exporters/parquet_exporter/records.rs
```diff
@@ -0,0 +1,76 @@
...
+    },
+    proto::opentelemetry::arrow::v1::ArrowPayloadType,
+};
+
+/// Parquet-exporter batch storage backed by unvalidated [`RawBatchStore`]s.
+///
+/// Converted from [`OtapArrowRecords`] at the exporter boundary via
+/// `From<OtapArrowRecords>`.
+#[allow(clippy::large_enum_variant)]
+pub enum OtapParquetRecords {
```
JakeDern, line 26: Just flagging this as the reason the RawBatchStore is immediately useful IMO

rust/otap-dataflow/crates/pdata/src/otap/raw_batch_store.rs
JakeDern: See also: https://github.com/open-telemetry/otel-arrow/pull/2356/changes#r2956705158

rust/otap-dataflow/crates/pdata/src/error.rs
```diff
@@ -218,35 +219,18 @@ pub enum Error {
         message: String,
     },
 
-    #[error("Encoding error: {}", error)]
-    Encoding {
```
JakeDern, line 222: There was circular problem here with encoding/decoding potentially producing one of these internal errors and then things that produce this error type potentially producing an Encoding error. I did was I thougt was reasonable and removed encoding from here for now and updated the places that returned this error variant to just return the inner `encode::Error` type instead.

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/bundle_adapter.rs
```diff
@@ -524,91 +524,58 @@ pub fn convert_bundle_to_pdata(
...
+{
+    let mut store = T::default();
 
     for (slot_id, batch) in payloads {
-        // Include signal-specific slots for Metrics
-        if let Some((SignalType::Metrics, payload_type)) = from_slot_id(*slot_id) {
-            metrics.set(payload_type, batch.clone());
+        // Include signal-specific slots (any signal type)
+        if let Some((signal_type, payload_type)) = from_slot_id(*slot_id) {
+            assert_eq!(signal_type, T::SIGNAL_TYPE);
```
AaronRM: This assert is in product code, so it could cause a panic and process crash. We should return a `BundleConversionError` for more graceful error handling.

rust/otap-dataflow/crates/pdata/src/otap/filter.rs
```diff
@@ -987,7 +987,10 @@ fn apply_filter(
     let filtered_record_batch = arrow::compute::filter_record_batch(record_batch, filter)
         .map_err(|e| Error::ColumnLengthMismatch { source: e })?;
     let num_rows_removed = num_rows_before - (filtered_record_batch.num_rows() as u64);
-    payload.set(payload_type, filtered_record_batch);
+    // safety: Removing rows from a valid payload should yield a valid payload
+    payload
+        .set(payload_type, filtered_record_batch)
+        .expect("filter did not change validity");
```
AaronRM, line 993: The added `expect` check is adds a path which could cause a panic in production. Consider replacing `expect` with `?`.

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
```diff
@@ -1429,7 +1429,11 @@ pub(crate) fn filter_otap_batch(
     let new_root_batch = filter_record_batch(root_batch, selection_vec)?;
 
     // replace the root batch
-    otap_batch.set(otap_batch.root_payload_type(), new_root_batch);
+    // safety: Filtering a valid payload should produce a valid payload of the
+    // same type.
+    otap_batch
+        .set(otap_batch.root_payload_type(), new_root_batch)
+        .expect("valid otap batch");
```
AaronRM, line 1436: Same comment here. Should be `?` instead of `expect`.

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/bundle_adapter.rs
JakeDern: Sure thing - I think it should be impossible unless our implementation has a fairly critical bug in it, but I'll downgrade it to a debug assert so it'll fire in non-prod builds only. Let me know if you prefer removal though.

BundleConversionError is returned right below this if the set fails.

rust/otap-dataflow/crates/pdata/src/otap/filter.rs
JakeDern: I'm open to adjusting this, but I don't think this is inconsistent with safety conventions in the pdata module. We often use expect + a safety comment when we believe failure preconditions are impossible and it would be a programming error otherwise. 

There are many examples in the crate, but here's one I grabbed just now: https://github.com/open-telemetry/otel-arrow/blob/f17759953287bee8d120867b02f74d148b1555a7/rust/otap-dataflow/crates/pdata/src/schema.rs#L39-L44

If you disagree with the safety comment or the practice in general, I'm totally open to discussing it!

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
JakeDern: Same discussion point as here: https://github.com/open-telemetry/otel-arrow/pull/2356#discussion_r2961411109

rust/otap-dataflow/crates/pdata/src/otap.rs
albertlockett: Yeah I agree with this being sealed. 

I think it would be better to make it unsafe if we had external users of this crate who that implemented this (similar to what happened w/ arrow Array trait). Since we probably don't have that, we can keep sealed until there's a compelling use case to allow custom implementations

rust/otap-dataflow/crates/pdata/src/schema/error.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use crate::schema::schema::DataType as OtapDataType;
+
+/// Result type for schema validation.
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// Errors that can occur when validating a [`RecordBatch`](arrow::array::RecordBatch)
+/// against an OTAP [`Schema`](crate::schema::schema::Schema) definition.
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
```
albertlockett, line 16: I wonder if eventually we should consolidate Error types? There's a bunch of errors in `crate::error::Error` related to like having the wrong type for some column, wrong dictionary keys, etc. Doesn't have to happen as part of this PR. can be ongoing cleanup

rust/otap-dataflow/crates/query-engine/src/error.rs
```diff
@@ -44,3 +44,11 @@ impl From<DataFusionError> for Error {
...
     }
 }
+
+impl From<otap_df_pdata::error::Error> for Error {
+    fn from(error: otap_df_pdata::error::Error) -> Self {
+        Self::ExecutionError {
+            cause: format!("PdataError: {error:?}"),
+        }
+    }
+}
```
albertlockett, line 54: nice -- this will be useful. There's a bunch of places where this conversion is done manually and this will clean things up.

rust/otap-dataflow/crates/pdata/src/schema/error.rs
JakeDern: Yeah I had the same thought - I was wondering about breaking things down into more focused error types as well that could be aggregated into a smaller number of categories as `#[source]`s in more crate-wide error type. Just an idea

JakeDern:
> LGTM!
> 
> Thank you for adding some sorely needed discipline, and apologies for creating tests with so many non-spec columns 😅

Thanks for the review! Don't worry we were all guilty as charged on the non-spec columns 😄

jmacd:
LGTM

PR #2359, opened 2026-03-18
Fix CI for Ubuntu 24 runner: Docker Buildx + Python setup

cijothomas:
The self-hosted perf test runner changed from Oracle Linux 8 to Ubuntu 24. This broke the CI in two ways:

1. **Docker:** The runner lacks the `buildx` component, which is needed for `--build-context` and `FROM --platform` in the Dockerfile. Fixed by adding `docker/setup-buildx-action` and using `docker buildx build --load`.

2. **Python:** The OL8-specific Python install step was skipped (condition `id == 'ol'` fails on Ubuntu), and Ubuntu 24 doesn't provide a `python` command by default. Fixed by replacing the OL8 block with `actions/setup-python`.

This also cleans up ~184 lines of now-unnecessary OL8-specific code (detect_os step, dnf installs, python shim, disk-cleanup).

### DIFF
--- a/.github/workflows/pipeline-perf-test-continuous.yml
+++ b/.github/workflows/pipeline-perf-test-continuous.yml
@@ -71,7 +71,7 @@ jobs:
           git submodule init
           git submodule update
           cd rust/otap-dataflow
-          docker build --build-context otel-arrow=../../ -f Dockerfile -t df_engine .
+          docker buildx build --load --build-context otel-arrow=../../ -f Dockerfile -t df_engine .
           cd ../..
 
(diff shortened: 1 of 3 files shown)


cijothomas:
> @cijothomas for my own knowledge, can you link to sample failure that helped you discover this?

https://github.com/open-telemetry/otel-arrow/actions/runs/23217009635/job/67480362748

(The machine was replaced yesterday with a new one. We might hit other issues as well, will monitor and update. Also the machine went offline sometime ago, have followed up to get it rebooted)

PR #2370, opened 2026-03-18
Ack nack redesign

lquerel:
**Not ready for review yet!**

# Change Summary

This PR redesigns the OTel dataflow engine control plane and graceful shutdown flow.

At a high level, it:
- splits shared runtime orchestration from Ack/Nack result unwinding, so control traffic and completion traffic progress independently
- introduces a receiver-first two-phase drain (`DrainIngress` / `ReceiverDrained`) so graceful shutdown stops new ingress first, then drains admitted work before downstream shutdown
- replaces absolute control priority with bounded-fair scheduling so `pdata`, timers, delayed data, and control all continue to make progress under load
- aligns engine/config terminology around `node`, `runtime`, and `results` (I'm still working on this naming...)
- expands validation with deterministic simulation testing (DST), targeted regression tests, and updated engine/config documentation

* Closes #2326 

- Expanded engine regression tests for:
  - control/result path liveness
  - `MessageChannel` fairness
  - timer / delayed-data progress under control pressure
  - shutdown-time handling of delayed data and timers
  - receiver `wait_for_result` behavior during graceful shutdown
- Added deterministic simulation testing (DST) for:
  - Ack/Nack unwind correctness
  - mixed `pdata` / runtime-control / result interleavings
  - heavy-ingress and backpressure scenarios
  - shutdown ordering and drain completion
  - known closed-admission behavior at shutdown deadline
- Validated with crate-level test runs including `otap-df-engine` and `otap-df-otap`

Yes.

- Channel-capacity policy names now use:
  - `policies.channel_capacity.control.node`
  - `policies.channel_capacity.control.runtime`
  - `policies.channel_capacity.control.results`
  - `policies.channel_capacity.pdata`
- Engine and config documentation were updated to reflect the new runtime/results terminology and the current graceful shutdown behavior.

### DIFF
--- a/rust/otap-dataflow/benchmarks/benches/otlp_bytes/main.rs
+++ b/rust/otap-dataflow/benchmarks/benches/otlp_bytes/main.rs
@@ -6,6 +6,11 @@
 //! This crate benchmarks OTLP.
 
 use criterion::{Criterion, criterion_group, criterion_main};
+use otap_df_pdata::otlp::ProtoBuffer;
+use otap_df_pdata::otlp::ProtoBytesEncoder;
+use otap_df_pdata::otlp::logs::LogsProtoBytesEncoder;
+use otap_df_pdata::proto::OtlpProtoMessage;
+use otap_df_pdata::testing::round_trip::otlp_to_otap;
 use prost::Message;
 use prost::bytes::BytesMut;
 
@@ -117,5 +122,32 @@ fn otlp_payload_clone(c: &mut Criterion) {
     group.finish();
 }
 
-criterion_group!(otlp_bytes, otlp_pdata_to_bytes_logs, otlp_payload_clone);
+/// Benchmark converting OTAP to OTLP proto bytes
+fn otap_to_bytes_logs(c: &mut Criterion) {
+    let mut group = c.benchmark_group("OTAP to OTLP Logs");
+
+    let logs = create_logs_data();
+    let otap_batch = otlp_to_otap(&OtlpProtoMessage::Logs(logs));
+    let mut proto_buffer = ProtoBuffer::new();
+    let mut encoder = LogsProtoBytesEncoder::new();
+
+    _ = group.bench_function("proto encode", |b| {
+        b.iter(|| {
+            let mut batch = otap_batch.clone();
+            proto_buffer.clear();
+            encoder
+                .encode(&mut batch, &mut proto_buffer)
+                .expect("can encode proto");
+        })
+    });
+
+    group.finish();
+}
+
+criterion_group!(
+    otlp_bytes,
+    otlp_pdata_to_bytes_logs,
+    otlp_payload_clone,
+    otap_to_bytes_logs
+);
 criterion_main!(otlp_bytes);
--- a/rust/otap-dataflow/crates/core-nodes/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/README.md
@@ -32,6 +32,7 @@ Each component lives in its own subfolder within a category:
         filter_processor/
         log_sampling_processor/
         retry_processor/
+        temporal_reaggregation_processor/
         signal_type_router/
         transform_processor/
       receivers/
@@ -127,6 +128,7 @@ Each component lives in its own subfolder within a category:
 | log_sampling_processor | `urn:otel:processor:log_sampling` | `src/processors/log_sampling_processor/` |
 | retry_processor | `urn:otel:processor:retry` | `src/processors/retry_processor/` |
 | signal_type_router | `urn:otel:processor:type_router` | `src/processors/signal_type_router/` |
+| temporal_reaggregation_processor | `urn:otel:processor:temporal_reaggregation` | `src/processors/temporal_reaggregation_processor/` |
 | transform_processor | `urn:otel:processor:transform` | `src/processors/transform_processor/` |
 
 #### attributes_processor
@@ -195,6 +197,10 @@ Each component lives in its own subfolder within a category:
 
 - Applies KQL or OPL transformations to OTAP batches via the query engine
 - Supports routed outputs while preserving upstream Ack/Nack semantics
+
+#### temporal_reaggregation_processor
+
+- Reaggregates metrics at a lower frequency to reduce telemetry volume
 
 ### Receivers
 
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/mod.rs
@@ -36,3 +36,6 @@ pub mod signal_type_router;
 
 /// Log sampling processor.
 pub mod log_sampling_processor;
+
+/// Temporal reaggregation processor.
+pub mod temporal_reaggregation_processor;
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/README.md
@@ -1,0 +1,28 @@
+# Temporal Reaggregation Processor
+
+Status: **WIP**
+
+URN: `urn:otel:processor:temporal_reaggregation`
+
+## Overview
+
+The temporal reaggregation processor decreases telemetry volume by
+reaggregating metrics collected at a higher frequency into a lower one.
+This is one of the three semantics-preserving transformations outlined in the
+[Metrics Data Model specification][metrics-data-model].
+
+This processor is partially modeled after the [Go interval processor][go-interval].
+
+[metrics-data-model]: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#events--data-stream--timeseries
+[go-interval]: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.147.0/processor/intervalprocessor
+
+## Configuration
+
+```yaml
+temporal-reaggregation:
+  type: urn:otel:processor:temporal_reaggregation
+  config:
+    # The interval at which aggregated metrics are emitted.
+    # Default: 60s
+    period: 60s
+```
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/config.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/config.rs
@@ -1,0 +1,94 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Configuration for the temporal reaggregation processor.
+
+use std::time::Duration;
+
+use otap_df_config::error::Error as ConfigError;
+use serde::{Deserialize, Serialize};
+
+const DEFAULT_PERIOD_SECONDS: u64 = 60;
+
+const MIN_PERIOD_MILLIS: u64 = 100;
+
+fn default_period() -> Duration {
+    Duration::from_secs(DEFAULT_PERIOD_SECONDS)
+}
+
+/// Configuration for the temporal reaggregation processor.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Config {
+    /// The interval at which the processor aggregates and emits metrics.
+    /// Must be at least 100ms. Default: 60s.
+    #[serde(with = "humantime_serde", default = "default_period")]
+    pub period: Duration,
+}
+
+impl Config {
+    /// Validates the configuration.
+    pub fn validate(&self) -> Result<(), ConfigError> {
+        if self.period < Duration::from_millis(MIN_PERIOD_MILLIS) {
+            return Err(ConfigError::InvalidUserConfig {
+                error: format!(
+                    "period must be at least {}ms, got {:?}",
+                    MIN_PERIOD_MILLIS, self.period
+                ),
+            });
+        }
+        Ok(())
+    }
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            period: default_period(),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use serde_json::json;
+
+    #[test]
+    fn test_default_config() {
+        let config = Config::default();
+        assert_eq!(config.period, Duration::from_secs(60));
+    }
+
+    #[test]
+    fn test_deserialize_empty_config() {
+        let config: Config = serde_json::from_value(json!({})).unwrap();
+        assert_eq!(config.period, Duration::from_secs(60));
+    }
+
+    #[test]
+    fn test_deserialize_with_period() {
+        let config: Config = serde_json::from_value(json!({
+            "period": "30s"
+        }))
+        .unwrap();
+        assert_eq!(config.period, Duration::from_secs(30));
+    }
+
+    #[test]
+    fn test_validate_minimum_period() {
+        let reject_zero = Config {
+            period: Duration::ZERO,
+        };
+        assert!(reject_zero.validate().is_err());
+
+        let reject_sub_100ms = Config {
+            period: Duration::from_millis(50),
+        };
+        assert!(reject_sub_100ms.validate().is_err());
+
+        let accept_100ms = Config {
+            period: Duration::from_millis(MIN_PERIOD_MILLIS),
+        };
+        assert!(accept_100ms.validate().is_ok());
+    }
+}
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/metrics.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/metrics.rs
@@ -1,0 +1,20 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Metrics for the temporal reaggregation processor.
+
+use otap_df_telemetry::instrument::Counter;
+use otap_df_telemetry_macros::metric_set;
+
+/// Metrics for the temporal reaggregation processor.
+#[metric_set(name = "temporal_reaggregation.processor.pdata.metrics")]
+#[derive(Debug, Default, Clone)]
+pub struct TemporalReaggregationMetrics {
+    /// Number of flushes triggered by the regular timer.
+    #[metric(unit = "{flush}")]
+    pub flushes_timer: Counter<u64>,
+
+    /// Number of flushes triggered by exceeding the maximum stream count.
+    #[metric(unit = "{flush}")]
+    pub flushes_overflow: Counter<u64>,
+}
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
@@ -1,0 +1,256 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Temporal reaggregation processor for OTAP metrics.
+//!
+//! This processor decreases telemetry volume by reaggregating metrics collected
+//! at a higher frequency into a lower one.
+
+pub mod config;
+mod metrics;
+
+use std::sync::Arc;
+use std::time::Duration;
+
+use async_trait::async_trait;
+use linkme::distributed_slice;
+use otap_df_config::error::Error as ConfigError;
+use otap_df_config::node::NodeUserConfig;
+use otap_df_engine::MessageSourceLocalEffectHandlerExtension;
+use otap_df_engine::config::ProcessorConfig;
+use otap_df_engine::context::PipelineContext;
+use otap_df_engine::control::NodeControlMsg;
+use otap_df_engine::error::Error;
+use otap_df_engine::local::processor as local;
+use otap_df_engine::message::Message;
+use otap_df_engine::node::NodeId;
+use otap_df_engine::processor::ProcessorWrapper;
+use otap_df_otap::OTAP_PROCESSOR_FACTORIES;
+use otap_df_otap::pdata::OtapPdata;
+use otap_df_telemetry::metrics::MetricSet;
+
+use self::config::Config;
+use self::metrics::TemporalReaggregationMetrics;
+
+/// URN
+pub const TEMPORAL_REAGGREGATION_PROCESSOR_URN: &str = "urn:otel:processor:temporal_reaggregation";
+
+/// The temporal reaggregation processor.
+pub struct TemporalReaggregationProcessor {
+    metrics: MetricSet<TemporalReaggregationMetrics>,
+    collection_period: Duration,
+    /// Whether the periodic flush timer has been started.
+    timer_started: bool,
+}
+
+/// Factory function to create a [`TemporalReaggregationProcessor`].
+pub fn create_temporal_reaggregation_processor(
+    pipeline_ctx: PipelineContext,
+    node: NodeId,
+    node_config: Arc<NodeUserConfig>,
+    processor_config: &ProcessorConfig,
+) -> Result<ProcessorWrapper<OtapPdata>, ConfigError> {
+    Ok(ProcessorWrapper::local(
+        TemporalReaggregationProcessor::from_config(pipeline_ctx, &node_config.config)?,
+        node,
+        node_config,
+        processor_config,
+    ))
+}
+
+/// Register the temporal reaggregation processor as an OTAP processor factory.
+#[allow(unsafe_code)]
+#[distributed_slice(OTAP_PROCESSOR_FACTORIES)]
+pub static TEMPORAL_REAGGREGATION_PROCESSOR_FACTORY: otap_df_engine::ProcessorFactory<OtapPdata> =
+    otap_df_engine::ProcessorFactory {
+        name: TEMPORAL_REAGGREGATION_PROCESSOR_URN,
+        create: |pipeline_ctx: PipelineContext,
+                 node: NodeId,
+                 node_config: Arc<NodeUserConfig>,
+                 proc_cfg: &ProcessorConfig| {
+            create_temporal_reaggregation_processor(pipeline_ctx, node, node_config, proc_cfg)
+        },
+        wiring_contract: otap_df_engine::wiring_contract::WiringContract::UNRESTRICTED,
+        validate_config: otap_df_config::validation::validate_typed_config::<Config>,
+    };
+
+impl TemporalReaggregationProcessor {
+    /// Creates a new processor from a configuration JSON value.
+    pub fn from_config(
+        pipeline_ctx: PipelineContext,
+        config: &serde_json::Value,
+    ) -> Result<Self, ConfigError> {
+        let metrics = pipeline_ctx.register_metrics::<TemporalReaggregationMetrics>();
+        let config: Config =
+            serde_json::from_value(config.clone()).map_err(|e| ConfigError::InvalidUserConfig {
+                error: e.to_string(),
+            })?;
+        config.validate()?;
+        Ok(Self {
+            metrics,
+            collection_period: config.period,
+            timer_started: false,
+        })
+    }
+
+    /// Starts the periodic flush timer if it has not already been started.
+    async fn ensure_timer_started(
+        &mut self,
+        effect_handler: &local::EffectHandler<OtapPdata>,
+    ) -> Result<(), Error> {
+        if !self.timer_started {
+            let _handle = effect_handler
+                .start_periodic_timer(self.collection_period)
+                .await?;
+            self.timer_started = true;
+        }
+        Ok(())
+    }
+}
+
+#[async_trait(?Send)]
+impl local::Processor<OtapPdata> for TemporalReaggregationProcessor {
+    async fn process(
+        &mut self,
+        msg: Message<OtapPdata>,
+        effect_handler: &mut local::EffectHandler<OtapPdata>,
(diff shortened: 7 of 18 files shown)


rust/otap-dataflow/crates/core-nodes/src/processors/retry_processor/mod.rs
```diff
@@ -836,6 +839,153 @@ mod test {
         )
     }
 
+    // Downstream Nacks are only retryable when RETURN_DATA preserves the original
+    // payload. This test locks in that losing the payload produces a terminal Nack
+    // immediately instead of wedging the request in a retry loop.
```
jmacd, line 844: :+1:

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
```diff
@@ -1966,6 +1982,251 @@ mod tests {
         test_timer_flush(datagen.generate_logs().into(), true);
     }
 
+    // The processor schedules one-shot DelayedData wakeups without cancelling older
```
jmacd, line 2013: :+1:

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
```diff
@@ -124,6 +124,15 @@ impl local::Receiver<OtapPdata> for InternalTelemetryReceiver {
                 // Handle control messages with priority
                 ctrl_msg = ctrl_msg_recv.recv() => {
                     match ctrl_msg {
+                        Ok(NodeControlMsg::DrainIngress { deadline, .. }) => {
+                            while let Ok(event) = logs_receiver.try_recv() {
+                                if let ObservedEvent::Log(log_event) = event {
+                                    Self::send_log_event(&effect_handler, log_event, &resource_bytes, &mut scope_cache).await?;
+                                }
+                            }
```
jmacd, line 132: This makes me wonder how we handle internal logging shutdown. (Not actionable.)

rust/otap-dataflow/crates/core-nodes/src/receivers/otap_receiver/mod.rs
```diff
@@ -1036,4 +1161,110 @@ mod tests {
             .run_test(nack_scenario(grpc_endpoint)) // Use NACK-specific scenario
             .run_validation_concurrent(nack_validation_procedure()); // Use NACK-specific validation
     }
+
+    // When wait_for_result is enabled, shutdown must resolve every in-flight batch
+    // response stream to an explicit unavailable status instead of leaving the OTAP
+    // client hanging on the server-side wait path.
```
jmacd, line 1178: 👍

rust/otap-dataflow/crates/core-nodes/src/receivers/otlp_receiver/mod.rs
```diff
@@ -2181,6 +2509,174 @@ mod tests {
             .run_validation_concurrent(validation);
     }
 
+    // Hold a wait_for_result gRPC request in flight, then trigger shutdown
+    // before any Ack or Nack is produced. The caller should receive a terminal
+    // Unavailable result that carries the shutdown reason instead of hanging.
```
jmacd, line 2527: :+1:

rust/otap-dataflow/crates/engine/src/clock.rs
```diff
@@ -0,0 +1,262 @@
...
+//! waiter bookkeeping and locking are only used when tests explicitly install a
+//! `SimClock`, so the expected production impact is negligible.
+
+use std::cell::{Cell, RefCell};
+use std::future::Future;
+use std::pin::Pin;
+use std::time::{Duration, Instant};
+
+/// Boxed sleep future used by the engine control plane.
+pub type Sleep = Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
```
jmacd, line 29: Cool

rust/otap-dataflow/crates/engine/src/testing/dst/heavy_ingress.rs
```diff
@@ -0,0 +1,501 @@
...
+    processor_paused_seen: bool,
+    processor_resumed_seen: bool,
+    receiver_drain_seen: bool,
+    receiver_drained_sent: bool,
+    processor_shutdown_seen: bool,
+    exporter_shutdown_seen: bool,
+    runtime_noise_sent: usize,
+}
+
+async fn run_backpressure_interblock_seed(seed: u64) {
```
jmacd, line 55: Please comment! I was wondering if this fn and the similar ones in control_plane.rs, closed-admission.rs are possibly `#[cfg(test)]`?

rust/otap-dataflow/crates/engine/src/completion_emission_metrics.rs
```diff
@@ -0,0 +1,79 @@
...
+/// Node-scoped completion counters emitted from effect handlers.
+#[metric_set(name = "node.completion_emission")]
+#[derive(Debug, Default, Clone)]
+pub(crate) struct CompletionEmissionMetrics {
+    /// Count of successful `notify_ack` routings into the shared completion lane.
+    #[metric(name = "notify_ack.routed", unit = "{message}")]
+    pub notify_ack_routed: Counter<u64>,
+    /// Count of successful `notify_nack` routings into the shared completion lane.
+    #[metric(name = "notify_nack.routed", unit = "{message}")]
+    pub notify_nack_routed: Counter<u64>,
```
jmacd, line 28: 👍

rust/otap-dataflow/crates/engine/src/control.rs
```diff
@@ -13,25 +14,19 @@ use bytemuck::Pod;
...
-}
-
 /// Returns a monotonic timestamp in nanoseconds since an arbitrary process epoch.
 /// Used for duration calculations in pipeline component metrics.
 ///
-/// TODO: This should not use a thread_local; it should not store any
-/// state at all. Use clock_gettime() or windows::QueryPerformanceCounter.
+/// `0` is reserved as the sentinel for "timestamp not set", so this function
+/// always returns a strictly positive value.
+///
```
jmacd, line 26: 👍

rust/otap-dataflow/crates/engine/src/control_plane_metrics.rs
```diff
@@ -0,0 +1,897 @@
...
+/// Pipeline-scoped metrics for the shared runtime-control actor.
+///
+/// This set describes shutdown/drain progress, pending control backlogs, and
+/// the state and throughput of the runtime-managed timer and delayed-data
+/// machinery. It is the primary metric set for understanding whether the
+/// pipeline runtime is still making progress under control-plane load and where
+/// graceful shutdown is spending time.
+#[metric_set(name = "pipeline.runtime_control")]
+#[derive(Debug, Default, Clone)]
+pub(crate) struct RuntimeControlMetrics {
```
jmacd, line 50: Looks good.

rust/otap-dataflow/crates/engine/src/control_plane_metrics.rs
```diff
@@ -0,0 +1,897 @@
...
+            self.collect_telemetry_sent += collect_telemetry as u64;
+            self.delayed_data_sent += delayed_data_sent as u64;
+            if timer_ticks > 0 || collect_telemetry > 0 || delayed_data_sent > 0 {
+                self.dirty = true;
+            }
+        }
+    }
+
+    pub(crate) fn report_if_needed(&mut self) -> Result<(), TelemetryError> {
+        if !self.dirty {
```
jmacd, line 616: How much do we save with this .dirty field? It appears error-prone, maybe.

rust/otap-dataflow/crates/engine/src/message.rs
```diff
@@ -199,68 +265,77 @@ pub struct MessageChannel<PData> {
...
+    /// - exporters switch to forced draining of already buffered channel data
     ///
-    /// Returns a [`RecvError`] if both channels are closed, or if the
-    /// shutdown deadline has passed.
-    pub async fn recv_when(&mut self, accept_pdata: bool) -> Result<Message<PData>, RecvError> {
-        let mut sleep_until_deadline: Option<Pin<Box<Sleep>>> = None;
+    /// This method is the single point where that shutdown-time distinction is
+    /// decided for the shared receive loop.
+    fn shutdown_drain_accepts_pdata(accept_pdata: bool, policy: DrainPolicy) -> bool {
+        accept_pdata || matches!(policy, DrainPolicy::ForceDrainDuringShutdown)
```
jmacd, line 330: I see!

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
```diff
@@ -280,6 +280,37 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
                         // Process incoming control messages.
                         ctrl_msg = ctrl_chan.recv() => {
                             match ctrl_msg {
+                                Ok(NodeControlMsg::DrainIngress { deadline, .. }) => {
```
jmacd, line 283: @utpilla Please take a look.

rust/otap-dataflow/crates/core-nodes/src/receivers/topic_receiver/mod.rs
```diff
@@ -340,6 +354,34 @@ impl local::Receiver<OtapPdata> for TopicReceiver {
...
+                                                    "topic_receiver.drain_ingress_reject_failed",
+                                                    node = receiver_id.name.as_ref(),
+                                                    topic = config.topic.as_ref(),
+                                                    error = e.to_string(),
+                                                    message = "Failed to reject topic message while receiver was draining ingress"
+                                                );
+                                            }
+                                        }
+                                    }
+                                    tokio::task::consume_budget().await;
```
jmacd, line 381: What does this do?

rust/otap-dataflow/crates/engine/src/testing/dst/closed_admission.rs
```diff
@@ -0,0 +1,100 @@
...
+
+use super::{DstRng, SimClock, dst_seeds};
+use crate::Interests;
+use crate::control::NodeControlMsg;
+use crate::message::{Message, ProcessorMessageChannel, Receiver};
+use crate::testing::dst::common::{setup_dst_runtime, yield_cycles};
+use otap_df_channel::mpsc;
+use std::time::Duration;
+
+async fn run_closed_admission_deadline_seed(seed: u64) {
```
jmacd, line 15: Comment.

rust/otap-dataflow/crates/engine/src/testing/dst/control_plane.rs
```diff
@@ -0,0 +1,310 @@
...
+    recv_controls, recv_until, setup_dst_runtime, yield_cycles,
+};
+use crate::testing::test_nodes;
+use otap_df_config::policy::TelemetryPolicy;
+use otap_df_telemetry::reporter::MetricsReporter;
+use std::collections::HashMap;
+use std::time::Duration;
+use tokio::time::timeout;
+
+async fn run_control_plane_seed(seed: u64) {
```
jmacd, line 31: Comment.

rust/otap-dataflow/crates/engine/src/testing/dst/message_channel.rs
```diff
@@ -0,0 +1,171 @@
...
+
+use super::{DstRng, SimClock, dst_seeds};
+use crate::Interests;
+use crate::control::NodeControlMsg;
+use crate::message::{ExporterMessageChannel, Message, ProcessorMessageChannel, Receiver};
+use crate::testing::dst::common::setup_dst_runtime;
+use otap_df_channel::mpsc;
+use std::time::Duration;
+
+async fn run_message_channel_seed(seed: u64) {
```
jmacd, line 15: Comment.

rust/otap-dataflow/crates/engine/src/message.rs
```diff
@@ -3,16 +3,20 @@
...
 use otap_df_channel::mpsc;
+use std::future::Future;
 use std::ops::Add;
-use std::pin::Pin;
 use std::time::{Duration, Instant};
-use tokio::time::{Sleep, sleep_until};
+
+/// Maximum number of consecutive control messages delivered before the channel
+/// forces one pdata attempt when pdata delivery is allowed.
+const CONTROL_BURST_LIMIT: usize = 32;
```
jmacd, line 19: Should this be configurable?

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -1,49 +1,53 @@
...
-/// the caller's timeout fires, avoiding a race where both expire simultaneously.
-const MAX_DRAINING_BUFFER_DURATION: Duration = Duration::from_secs(1);
-
 /// Threshold for the pending sends buffer. When the buffer exceeds this size,
 /// a warning is logged to help operators diagnose sustained backpressure.
 const PENDING_SENDS_WARN_THRESHOLD: usize = 100;
 
+/// Maximum number of consecutive runtime-control messages handled before the
+/// manager forces one due expiry pass.
+const RUNTIME_CTRL_BURST: usize = 64;
```
jmacd, line 49: Should this (eventually?) be configurable?

rust/otap-dataflow/crates/engine/README.md
```diff
@@ -1,402 +1,652 @@
...
+   with `pdata` through role-specific message channels.
+3. **Pipeline runtime channels**
+   Each pipeline runtime owns two bounded shared MPSC channels:
+   - a **runtime-control channel** for timers, delayed-data requests,
+     receiver-drain notifications, and shutdown requests, consumed by
+     `RuntimeCtrlMsgManager`
+   - a **pipeline-completion channel** for `DeliverAck` / `DeliverNack`, consumed by
+     `PipelineCompletionMsgDispatcher`
+
+### Why Runtime-Control and Completion Are Split
```
jmacd, line 156: Thanks for the documentation update.

jmacd (approved):

jmacd:
Thank you @lquerel.

With this PR I believe we can close
https://github.com/open-telemetry/otel-arrow/pull/2317
and
https://github.com/open-telemetry/otel-arrow/pull/2326

Please take a look @gouslu @utpilla @lalitb.

lalitb:
This looks headed in the right direction overall. Splitting timer/shutdown control traffic from Ack/Nack completion traffic makes sense, receiver-first drain feels like the right shutdown model, and the DST coverage is a nice addition.                                              
                                                                                                    
Want to go deeper on a few edge cases - in particular the forced-deadline path under backpressure, where there may be a liveness/deadlock issue around channel ownership and shutdown ordering.                                                                                         
 
Also want to run the benchmarks before signing off, just to sanity check the Ack/Nack path under load given the added dispatcher overhead.

rust/otap-dataflow/crates/engine/src/message.rs
```diff
@@ -312,101 +389,282 @@ impl<PData: ReceivedAtNode> MessageChannel<PData> {
...
+    pub(crate) async fn recv_internal(&mut self) -> Result<Message<PData>, RecvError> {
+        self.recv_when_internal(true).await
+    }
+
+    pub(crate) async fn recv_when_internal(
+        &mut self,
+        accept_pdata: bool,
+    ) -> Result<Message<PData>, RecvError> {
+        self.core
+            .recv_with_policy(accept_pdata, DrainPolicy::ForceDrainDuringShutdown)
```
gouslu, line 637: Should this be exposed to consumers? Explicit behavior would be appreciated here. I think that when accept_pdata is set to false, we should default to honor admission, otherwise it implies implicit behavior and can be confusing for people who are not aware of this. But if consumers can set this explicitly, then they will be aware.

I also think that, even if people set "honoradmission", there should be an internal mechanism that drains the pdata that has never made it to the next component. I think it is fair to assume that an exporter that doesn't accept pdata is either already full or facing an outage or something. Maybe shutdown gives a grace period for pdata, but after that grace period, and after exporter exists, it should just auto-nack whatever is remaining in the channel.

Those being said, auto-nack could be a separate work.

gouslu: Maybe auto-nack could also be a policy.

rust/otap-dataflow/crates/engine/src/message.rs
```diff
@@ -285,17 +360,19 @@ impl<PData: ReceivedAtNode> MessageChannel<PData> {
...
             // Draining mode: Shutdown pending
             if let Some(dl) = self.shutting_down_deadline {
-                // If shutdown pending and no pdata left, return Shutdown immediately
+                let drain_accepts_pdata =
+                    Self::shutdown_drain_accepts_pdata(accept_pdata, drain_policy);
+
+                // Once shutdown has been latched, the stored Shutdown is released
+                // only after the bounded pdata backlog is empty. This keeps the
+                // channel-level drain contract explicit: upstream work that was
+                // already accepted into the channel gets a chance to run first.
```
lalitb, line 375: The new drain-path logic appears to release `Shutdown` as soon as buffered `pdata` is empty. At a design level, I think this is worth another look, because it may allow shutdown to overtake queued cleanup/control work that the new graceful-drain model is otherwise trying to preserve.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -275,356 +312,461 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
-            // Check if we've exceeded the draining deadline
-            if let Some(deadline) = draining_deadline {
-                if Instant::now() >= deadline {
-                    otel_warn!(
-                        "pipeline.draining.deadline_exceeded",
-                        "Draining deadline exceeded, forcing shutdown"
-                    );
+            let now = clock::now();
+
+            if let Some(deadline) = shutdown_deadline {
```
lalitb, line 366: The receiver-first drain model is a good direction, but the deadline-forced shutdown path looks worth another look under backpressure. In particular, I think it is important to confirm that the forced-shutdown guarantee still holds even when node control channels are already saturated.

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -677,78 +821,297 @@ impl<PData> PipelineCtrlMsgManager<PData> {
...
-                            outcome,
-                            now_ns,
+/// Dedicated return-path dispatcher for Ack/Nack unwinding.
+pub struct PipelineCompletionMsgDispatcher<PData> {
+    pipeline_completion_msg_receiver: PipelineCompletionMsgReceiver<PData>,
+    control_senders: ControlSenders<PData>,
+    node_metric_handles: Rc<RefCell<Vec<Option<NodeMetricHandles>>>>,
+    pending_sends: VecDeque<(usize, NodeControlMsg<PData>)>,
+    completion_metrics: PipelineCompletionMetricsState,
+}
```
lalitb, line 878: The new completion-path buffering is directionally good for avoiding stalls, but it appears to shift the failure mode into heap growth and retry churn under sustained saturation. I think the backpressure model here deserves another pass.

gouslu:
For exporters specifically, I think control and pdata channels should be directly exposed to the consumer. The current approach bakes drain policy into the channel (ForceDrainDuringShutdown), but exporters already have their own shutdown/drain logic and need to coordinate pdata drain with their in-flight state (e.g., flush current batch, wait for HTTP completions, nack orphans).

Two options that would give exporters full control:

- Shutdown message includes a drain handle — the exporter calls something like drain_handle.drain(pdata_rx, DrainPolicy::NackAll, &effect_handler).await. A handle would be great for convenience.

- Alternatively, since exporters would have direct access to pdata_rx, it can handles remaining items inside its own shutdown code path and doesn't even use a handle.

Either approach avoids the channel needing to know about exporter-specific drain semantics, and eliminates the edge case where the early closed-channel probe can shadow a pending shutdown with a synthetic one. It would also simplify message.rs and make it easier to maintain and keep bug free.

edit: this approach also gives more options for the exporters that needs to carry more complex logic.

gouslu:
> Thank you @lquerel.
> 
> With this PR I believe we can close #2317 and #2326
> 
> Please take a look @gouslu @utpilla @lalitb.

The auto-nack might not be needed since this PR introduces a force drain policy that recv_when implicitly enforces during shutdown, but I don't think I like that personally. If an exporter rejects data, it shouldn't be forced to accept it. If someone writes code assuming the rejection is honored, they can get caught in a bug they didn't expect.

I think #2317 is still useful defensively. Even though ForceDrainDuringShutdown prevents the hang during normal shutdown (Shutdown arrives on the control channel and force drain kicks in), #2317 covers the case where the pdata channel closes without Shutdown arriving — e.g., an upstream node crashes before the controller can dispatch Shutdown.

lalitb:
> For exporters specifically, I think control and pdata channels should be directly exposed to the consumer. The current approach bakes drain policy into the channel (ForceDrainDuringShutdown), but exporters already have their own shutdown/drain logic and need to coordinate pdata drain with their in-flight state (e.g., flush current batch, wait for HTTP completions, nack orphans).

I agree with the goal of making exporter shutdown behavior more explicit and controllable.
 
 That said, I'd be cautious about exposing raw `control_rx`/`pdata_rx` directly. Once pdata has been accepted into an engine-managed channel, its shutdown handling becomes at least partly an engine-level concern, and pushing raw drain logic into each exporter would likely duplicate subtle shutdown/race-condition handling across implementations - exactly the kind of thing the `MessageChannelCore` centralization was designed to avoid.
 
 What seems more attractive to me is keeping the drain machinery centralized, but making the policy **explicit and exporter-chosen** rather than implicit. The `DrainPolicy` enum already exists internally (`HonorAdmission` / `ForceDrainDuringShutdown`); it could be extended with a `NackBuffered` variant and surfaced as part of the exporter API, either via a trait method like `fn drain_policy(&self) -> DrainPolicy` or through the shutdown/drain-session handle you mentioned.
 
 That would give exporters the flexibility to choose their shutdown semantics (flush everything, nack orphans, honor admission) without every exporter needing to reimplement the channel-draining state machine. The drain-handle idea feels especially promising as a middle ground: centralized machinery, exporter-chosen policy.

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
lquerel: Good point. For tracking -> https://github.com/open-telemetry/otel-arrow/issues/2417

rust/otap-dataflow/crates/engine/src/testing/dst/heavy_ingress.rs
lquerel: Yes the entire `dst` module is behind `#[cfg(any(test, feature = "test-utils"))]`

I will a comment to clarify that at the top of the files.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
utpilla, line 319: (non-blocking): Now that `Shutdown` is only sent to receivers after the deadline has expired (forced-shutdown path), the drain-wait/`yield_now` loop here is effectively a no-op since `saturating_duration_since` returns zero. Both the TCP and UDP `Shutdown` handling logic could be simplified to:

```rust
Ok(NodeControlMsg::Shutdown { deadline, .. }) => {
    let _ = timer_cancel_handle.cancel().await;
    shutdown_flag.set(true);
    let snapshot = self.metrics.borrow().snapshot();
    return Ok(TerminalState::new(deadline, [snapshot]));
}
```

We could simplify this in a follow-up.

rust/otap-dataflow/crates/engine/src/message.rs
utpilla: There is also a cross-node data loss risk: since `Shutdown` is sent to processors and exporters simultaneously, the exporter can see `is_empty() == true` and exit before the processor has pushed its drain output. The old design avoided this by cascading shutdown via channel closure (a permanent signal); `is_empty()` is a transient snapshot that can't distinguish "upstream is done" from "upstream hasn't run yet."

rust/otap-dataflow/crates/core-nodes/src/receivers/topic_receiver/mod.rs
lquerel: Hehe, it’s an interesting function!

`tokio::task::consume_budget` is a kind of conditional async yield. Basically, it yields back to the Tokio scheduler if the current task has exhausted its budget. Otherwise, execution continues in the current task.

In the present case, we are in draining mode and consuming the messages currently in the topic in order to send nacks back through the upstream pipeline (if ack/nack is enabled). In that mode, and in this loop, there was no other await point, which is why this method is used. Compared with `tokio::task::yield_now`, it has the advantage of not necessarily yielding to the async scheduler.

utpilla: Ah more `tokio` specific coupling 😄.

I think if we were to switch to a different runtime in the future, we could explore `yield_now()` every `n` iterations to achieve the same anti-starvation goal without the runtime lock-in.

rust/otap-dataflow/crates/engine/src/testing/dst/control_plane.rs
lquerel: done

rust/otap-dataflow/crates/engine/src/testing/dst/message_channel.rs
lquerel: done

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
lquerel: Thank you. I applied your suggestion in https://github.com/open-telemetry/otel-arrow/pull/2370/changes/bd992b413327d00b94bf89e4bb880fa03657cf1a

utpilla: Thanks, we could simplify it for UDP as well.

lquerel:
@lalitb and @gouslu 

For tracking purposes, I created this GitHub issue: [Make exporter drain/shutdown policy explicit and exporter-controlled](#2424 ).

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
lquerel: I agree with the concern. The current pending_sends queue is a correctness/liveness valve, not a complete backpressure solution. It prevents a full node-control inbox from stalling unrelated completion unwinding on the same LocalSet, but it is currently unbounded, so sustained saturation can indeed turn into heap growth and retry churn. I think this PR still improves the failure mode relative to dispatcher stalls, but the backlog policy deserves a follow-up pass.

For tracking -> https://github.com/open-telemetry/otel-arrow/issues/2427

lalitb:
> Also want to run the benchmarks before signing off, just to sanity check the Ack/Nack path under load given the added dispatcher overhead.

I did notice a small end-to-end throughput regression here, around ~4%, when benchmarking with the noop exporter. My guess is that this is mostly exposing additional engine-side state/branching/bookkeeping in the new control/message path, which shows up more clearly when the exporter itself is effectively free.

That said, I did not observe the same regression with Azure Monitor or Geneva exporter workloads, which is closer to our actual use cases. In those cases the exporter/network cost seems to dominate enough that the extra engine overhead is not materially visible.

jmacd (approved):

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
lquerel: I agree with this feedback. The deadline-forced branch currently reuses the non-blocking send() path, which can defer Shutdown into pending_sends when a node-control inbox is full. Because the runtime-control manager then exits immediately, that means the forced-shutdown guarantee is weaker than intended under saturation. We should fix that so deadline-forced shutdown is delivered through a path that remains guaranteed even when normal node-control buffering is already full.

For tracking -> https://github.com/open-telemetry/otel-arrow/issues/2431

rust/otap-dataflow/crates/engine/src/message.rs
lquerel: I created the 2 following issues for follow up PRs:
- https://github.com/open-telemetry/otel-arrow/issues/2432 
- https://github.com/open-telemetry/otel-arrow/issues/2433

lquerel:
I merged `main` into my PR, and I’m seeing a performance regression. Something is going on that I need to figure out before merging #2370.

lquerel:
@jmacd @lalitb @utpilla @gouslu 

Okay, I found it. It was a configuration issue and an interaction with the fix Joshua made in the policies. There’s no problem with Joshua's change, but the bug he fixed was preventing the configuration defined at the group level from being applied, because it was being implicitly masked by the default pipeline-level configuration.

In the group, I had configured a pdata channel capacity of 4096, while the default was 128. With your fix, the group-level configuration was now actually being applied, resulting in 4096 slots per channel.

So going from 128 to 4096 is a 32x increase in work in progress per edge. In this benchmark, batches can contain up to 1000 logs, and Ack/Nack propagation is automatic, so larger internal buffers allow the topic exporters, receivers, retry processor, and completion path to retain much more admitted work before backpressure kicks in. That usually means significantly more in-flight state, worse cache locality, more memory traffic, and longer completion windows.

PR #2371, opened 2026-03-18
chore(deps): remove unused zip dependency from otap crate and workspace

pritishnahar95:
# Change Summary
- Remove the direct zip dependency from crates/otap/Cargo.toml (never imported or called anywhere in the source) and the now-dead zip = "=4.2.0" pin from the workspace Cargo.toml.
- Add [bzip2-1.0.6](https://spdx.org/licenses/bzip2-1.0.6.html) to license allowlist in `deny.toml`

* Closes #NNN

CI tests

No

### DIFF
--- a/rust/otap-dataflow/deny.toml
+++ b/rust/otap-dataflow/deny.toml
@@ -110,7 +110,8 @@ allow = [
     "CDLA-Permissive-2.0", # See https://github.com/cncf/foundation/issues/1007
     "CC0-1.0",
     "ISC",  # Used by the `is_ci`, similar to MIT license
-    "OpenSSL"
+    "OpenSSL",
+    "bzip2-1.0.6"
 ]
 # The confidence threshold for detecting a license from license text.
 # The higher the value, the more closely the license text must be to the


rust/otap-dataflow/deny.toml
```diff
@@ -110,7 +110,8 @@ allow = [
     "CDLA-Permissive-2.0", # See https://github.com/cncf/foundation/issues/1007
     "CC0-1.0",
     "ISC",  # Used by the `is_ci`, similar to MIT license
-    "OpenSSL"
+    "OpenSSL",
+    "bzip2-1.0.6"
```
pritishnahar95, line 114: @lquerel @jmacd is this ok to add?
See https://spdx.org/licenses/bzip2-1.0.6.html

Seems like a license modeled on BSD.

jmacd: I think yes.

utpilla: Is this addition actually needed? Running `cargo deny check licenses` on this branch shows `bzip2-1.0.6` as an unmatched license allowance. Since `cargo deny` doesn't encounter the `bzip2-1.0.6` license from any crate metadata, this entry appears unnecessary.

pritishnahar95: This does fail in the CI without adding the license: https://github.com/open-telemetry/otel-arrow/actions/runs/23269569695/job/67658878770?pr=2371

utpilla: Ah you're right. I was testing with a stale `Cargo.lock` that still had zip pinned at `4.2.0`. With a fresh resolve (as CI does), removing the `zip = "=4.2.0"` workspace pin lets `zip` upgrade to `4.6.1`, which pulls in `libbz2-rs-sys v0.2.2` and that crate declares `bzip2-1.0.6` as its license.

PR #2372, opened 2026-03-18
Fix CI Docker build by enabling BuildKit

cijothomas:
https://github.com/open-telemetry/otel-arrow/pull/2359 didn't alone fix, so reverting and trying a diff fix.

The new CI machine defaults to the legacy Docker builder which doesn't support --build-context or FROM --platform. Add DOCKER_BUILDKIT=1 prefix to enable the built-in BuildKit engine (available since Docker 18.09) without requiring the buildx plugin.

### DIFF
--- a/.github/workflows/pipeline-perf-on-label.yaml
+++ b/.github/workflows/pipeline-perf-on-label.yaml
@@ -50,53 +50,13 @@ jobs:
 
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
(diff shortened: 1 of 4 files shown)


cijothomas:
Hold on from merging. I added the perf label, so we can validate if this indeed fixes. (its waiting to get a runner...)

cijothomas:
Docker issue seems fixed. Next is python issue :)

cijothomas:
> Docker issue seems fixed. Next is python issue :)

Looks good. Please merge!

PR #2376, opened 2026-03-19
perf: increase syslog TCP load generator target rate

cijothomas:
## Summary

Remove the 5K logs/sec rate cap on the syslog load generator tests (both TCP and UDP) so they can actually stress-test the df-engine.

## Changes

1. **`loadgen.py`**: Changed `target_rate` Pydantic field from `gt=0` to `ge=0` so that `0` is accepted. When `target_rate` is 0 (falsy), the existing worker code skips all timer/sleep logic entirely, running at max throughput.

2. **`syslog-tcp-docker.yaml`**: Added `syslog_rate: 0` to all 4 test template variables. (The template's `update_component_strategy` step was overriding the component-level config with a hardcoded default of 5000.)

3. **`syslog-docker.yaml`** (UDP): Added `syslog_rate: 0` and `syslog_threads: 8` to all 4 test template variables. UDP needs multiple threads because each message is a separate `sendto()` syscall, limiting a single Python thread to ~9K logs/sec.

### TCP (1 thread, no rate limit)
| Test | Before | After | Speedup |
|---|---|---|---|
| ATTR-OTLP | 5,418 | **197,321** | 36x |
| CEF-ATTR-OTLP | 5,420 | **93,609** | 17x |
| ATTR-OTAP | 5,420 | **90,130** | 17x |
| CEF-ATTR-OTAP | 5,420 | **80,434** | 15x |

### UDP (8 threads, no rate limit)
| Test | Before | After | Speedup |
|---|---|---|---|
| ATTR-OTLP | 5,418 | **57,127** | 11x |
| CEF-ATTR-OTLP | 5,420 | **56,819** | 10x |
| ATTR-OTAP | 5,420 | **55,397** | 10x |
| CEF-ATTR-OTAP | 5,420 | **55,153** | 10x |

All tests now saturate the 1-core df-engine with 0% log drops (except CEF-ATTR-OTAP UDP at 0.4%).

### DIFF
--- a/tools/pipeline_perf_test/test_suites/integration/nightly/syslog-tcp-docker.yaml
+++ b/tools/pipeline_perf_test/test_suites/integration/nightly/syslog-tcp-docker.yaml
@@ -80,7 +80,7 @@ tests:
         syslog_transport: tcp
         # Non CEF message of approximately the same size as CEF message
         syslog_message: 'NOT-CEF:0|Security|threatmanager|1.0|100|worm stopped|10|src=192.168.1.10 dst=10.0.0.25 spt=443 dpt=8080 proto=TCP act=blocked cs1Label=MalwareType cs1=Worm cs2Label=DetectionMethod cs2=Heuristic cs3Label=Severity cs3=High requestMethod=POST'
-        syslog_rate: 999999999
+        syslog_rate: 0
         observation_interval: 60
   - name: SYSLOG-TCP-3164-ATTR-OTLP
     from_template:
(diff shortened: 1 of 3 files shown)


tools/pipeline_perf_test/test_suites/integration/nightly/syslog-tcp-docker.yaml
```diff
@@ -80,6 +80,7 @@ tests:
         syslog_transport: tcp
         # Non CEF message of approximately the same size as CEF message
         syslog_message: 'NOT-CEF:0|Security|threatmanager|1.0|100|worm stopped|10|src=192.168.1.10 dst=10.0.0.25 spt=443 dpt=8080 proto=TCP act=blocked cs1Label=MalwareType cs1=Worm cs2Label=DetectionMethod cs2=Heuristic cs3Label=Severity cs3=High requestMethod=POST'
+        syslog_rate: 999999999
```
utpilla: Would it be better to use `syslog_rate: 0` instead of `999999999`? The load generator already has a built-in "no rate limit" code path in loadgen.py. When `target_rate` is falsy (`0` or `None`), it sets `batch_interval = None` and skips all the timer/sleep logic entirely:

https://github.com/open-telemetry/otel-arrow/blob/f17759953287bee8d120867b02f74d148b1555a7/tools/pipeline_perf_test/load_generator/loadgen.py#L284-L293

With `999999999`, the rate-limiting code still runs every iteration. It calls `perf_counter()`, computes a negative sleep time, skips the sleep, and increments `late_batches` on every single batch (making that metric meaningless). With `0`, the entire block is skipped (tighter send loop, no wasted syscalls, and clean metrics).

jmacd: Sounds good.

cijothomas: good call! done.

PR #2379, opened 2026-03-19
Columnar query engine `extend`/`set` operator support attiributes as destination

### LINKED ISSUE #2016: Change the implementation of columnar query-engine's set attribute handling to use upsert
Currently when we write an OPL expression like `set attributes["x"] = "y"`, the columnar query engine treats this as an attribute insert (implementation was added in https://github.com/open-telemetry/otel-arrow/pull/1885).

However, the insert behaviour will not update an existing attribute, which is not the desired behaviour for this expression. After #2015  is implemented, we should change the attribute transformation for this type of expression to be an attribute upsert.

### LINKED ISSUE #2350: AttributeTransform support signals with no attributes
### Pre-filing checklist

### Component(s)

Currently the AttributeTransform requires a record batch to update the attributes for, this assumes signals already have attributes but we should be able to accommodate signals that don't have any attributes where the record batch doesn't exist but we want to be able to insert attributes.

### Proposed Solution

AttributeTransform should be able to create the attribute record batch for OTAP signals and insert (delete and upsert don't really apply here since there are no attributes before transformation) defined attributes for all signals in the batch (will need parent id from primary record batch)

### Alternatives Considered

_No response_

### Additional Context

_No response_

albertlockett:
Apologies to reviewers: I'm sorry this is long. I'm happy to break it up if desired.

# Change Summary

Adds the ability to set attribute values using the columnar query engine.

Note that we already had this capability when the source was a static literal (e.g. `logs | extend attributes["x"] = "hello"`. This PR improves the capability so the attribute can be assigned from the results of the expression evaluation that was added in https://github.com/open-telemetry/otel-arrow/pull/2126

It means that now we can do things like:
```kql
logs | set attributes["event_name"] = event_name // use root field as source
logs | set attributes["x"] = resource.attributes["y"] // use some other attribute as the sourc
logs | set attributes["x"] = attributes["y"] * 2 // use arithmetic as the source
// fyi ^^^ "set" is an alias of "extend" in OPL
// etc.

```

Note: this does create Empty attributes if the expression evalutes to `null`. E.g. for something like this `logs | set attributes["x"] = attributes["y"]`, if `attributes["y"]` did not exist for some row, then an empty attribute would be created for `attributes["x"]`.

This also fixes a few bugs in the current set attributes implementation:
- corrects the semantics of `set`/`extend` to be an "upsert" - e.g. replace the attribute value or create a new attribute if one did not exist. Before this PR, we did not replace existing values.
- fixes issue where attributes would not be inserted if the attribute record batch did not previously exist

The core of changes introduced is a optimized kernel for upserting attributes. Currently this lives is located at`otap_df_query_engine::pipeline::assign:attributes::upsert_attributes`.  This expects the caller to pass the attribute key, the new values, the parent_id associated with the attribute and a mask of which rows should be updated. It then uses this to quickly merge the new values/attribute types, and append any inserts onto each column, inserting nulls where appropriate and maintaining correct dictionary encoding semantics. For best performance, this can upsert multiple attribute keys at once.

Because the `upsert_attributes` kernel can assign multiple attributes at once, the query-engine's planner code and the `AssignPipelineStage` have both been modified to accomodate this. The planner attempts to coalesce multiple "set" transformations into a single pipeline stage if possible, and the `AssignPipelineStage` handles evaluating the expression for each source, and driving the invocation of the `upsert_attributes` kernel to do all the attribute upserts in bulk.

| Benchmark | 128 rows | 1536 rows | 8192 rows |
|---|---|---|---|
| `upsert_new_str_key` | 4.54 µs | 15.37 µs | 67.04 µs |
| `upsert_existing_str_key` | 7.10 µs | 49.16 µs | 239.45 µs |
| `upsert_two_new_str_keys` | 6.06 µs | 20.81 µs | 90.04 µs |
| `upsert_two_existing_str_keys` | 9.50 µs | 58.26 µs | 271.99 µs |
| `upsert_two_existing_one_new` | 11.41 µs | 65.20 µs | 300.78 µs |

In all cases, we see this is a lot faster than the current attribute upsert. (see here https://github.com/open-telemetry/otel-arrow/pull/2024). Although it's not an apples-to-apples comparison, both these acheive the same result on the same order-of-magnitude of log data, but this code is much faster.

* Related to #2036 
* Closes #2016

Unit tests - many new tests added

Yes - in transform processors users will now have the capability to assign the value of attributes using a variety of new expressions for sources (not just static literals).

## Future work

In a future PR I'll integrate the new `upsert_attributes` kernel into the attribute processor to improve performance and also fix #2350

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/planner.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/planner.rs
@@ -692,7 +692,8 @@ impl PipelinePlanner {
                 if !assignments.is_empty() {
                     let pipeline_stage = AssignPipelineStage::try_new(&mut assignments)?;
                     results.push(Box::new(pipeline_stage));
-                    assignments.clear()
+                    assignments.clear();
+                    cols_or_keys_referenced.clear();
                 }
 
                 let function_id = func.get_function_id();
@@ -778,6 +779,7 @@ impl PipelinePlanner {
                 let pipeline_stage = AssignPipelineStage::try_new(&mut assignments)?;
                 results.push(Box::new(pipeline_stage));
                 assignments.clear();
+                cols_or_keys_referenced.clear();
             }
 
             // assignment will be combined with previous assignments
@@ -1214,6 +1216,24 @@ mod test {
     {
         let pipeline_expr = OplParser::parse(
             "logs | set attributes[\"x\"] = 5 | set attributes[\"y\"] = attributes[\"z\"] * attributes[\"x\"]",
+        )
+        .unwrap()
+        .pipeline;
+        let planner = PipelinePlanner::new();
+        let stages = planner
+            .plan_stages(
+                &pipeline_expr,
+                &Pipeline::create_session_context(),
+                &OtapArrowRecords::Logs(Logs::default()), // empty placeholder
+            )
+            .unwrap();
+        assert_eq!(stages.len(), 2)
+    }
+
+    #[test]
+    fn test_combine_key_state_reset_between_combinations() {
+        let pipeline_expr = OplParser::parse(
+            "logs | set attributes[\"x\"] = 5 | set attributes[\"y\"] = 5 | set attributes[\"x\"] = 6 | set attributes[\"y\"] = 7",
         )
         .unwrap()
         .pipeline;


lquerel:
> Because the upsert_attributes kernel can assign multiple attributes at once, the query-engine's planner code and the AssignPipelineStage have both been modified to accomodate this. The planner attempts to coalesce multiple "set" transformations into a single pipeline stage if possible, and the AssignPipelineStage handles evaluating the expression for each source, and driving the invocation of the upsert_attributes kernel to do all the attribute upserts in bulk.

So cool!

rust/otap-dataflow/crates/query-engine/src/pipeline/assign/attributes.rs
```diff
@@ -0,0 +1,3257 @@
...
+            let dict = col
+                .as_any()
+                .downcast_ref::<DictionaryArray<UInt8Type>>()
+                .expect("dict<u8>");
+            let keys: Vec<u16> = dict
+                .keys()
+                .values()
+                .iter()
+                .enumerate()
+                .map(|(i, &k)| if dict.is_null(i) { 0 } else { k as u16 })
```
lquerel: Why are we converting null into 0 ?

rust/otap-dataflow/crates/query-engine/src/pipeline/assign/attributes.rs
```diff
@@ -0,0 +1,3257 @@
...
+                _ => (Arc::clone(col), None),
+            };
+            Some((values_arr, dict_keys))
+        }
+        None => None,
+    };
+
+    // Build a bitset of "live" dict value indices — values still referenced by at least one
+    // passthrough (non-updated) row. This prevents dead values from being included in the
+    // unified dictionary, which is important for data redaction scenarios.
```
lquerel, line 1127: Cool!

rust/otap-dataflow/crates/query-engine/src/pipeline/planner.rs
```diff
@@ -507,139 +545,252 @@ impl PipelinePlanner {
...
+        // no guarantees about the logical order in which the assignments are executed. This can
+        // lead to ambiguity about the final result if the same column or attribute key is used
+        // in multiple expressions. For example, if we combine two expressions like:
+        // `attributes["x"] = "y"` and `attributes["x"] = "z"` the end result would be different
+        // depending on which executed "first". Similar situation exist for cases like:
+        // `attributes["x"] = "A"` and `attributes["y"] = attributes["x"]`.
+        //
+        // To avoid the ambiguity, we keep track of which keys or columns used in the combined
+        // expressions in this set..
+        let mut cols_or_keys_referenced = HashSet::new();
```
lquerel, line 586: What is the lifecycle of this set? Is it reset on new assignment group?

rust/otap-dataflow/crates/query-engine/src/pipeline/assign/attributes.rs
albertlockett: oops - yeah that's not needed :) Fixed in 75ace9461

rust/otap-dataflow/crates/query-engine/src/pipeline/planner.rs
albertlockett: Oops - yeah that should be getting reset on each assignment group. Fixed in b22261c2a and added test

PR #2381, opened 2026-03-19
Add mock LA server for local Azure Monitor Exporter testing

cijothomas:
Adds a lightweight mock Azure Monitor Logs Ingestion API server and a local test config for performance testing the Azure Monitor Exporter without incurring Log Analytics costs.

## New files

- `crates/mock-la-server/` - Standalone Axum-based mock server that accepts gzip-compressed JSON POSTs, decompresses/counts log entries, prints periodic throughput stats, and supports error simulation flags (--fail-rate, --retry-after, --latency, --unauthorized-rate, --payload-too-large, --fail-after)
- `fakegen-ame-local.yaml` - Pipeline config pointing the exporter at http://localhost:9999 with auth.method: dev

## Usage

```
# Terminal 1: start mock server
cargo run -p mock-la-server -- --port 9999

# Terminal 2: run pipeline
cargo run --features azure-monitor-exporter -- --config crates/contrib-nodes/src/exporters/azure_monitor_exporter/fakegen-ame-local.yaml --num-cores 1
```

### DIFF
--- a/rust/otap-dataflow/crates/contrib-nodes/examples/mock_la_server.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/examples/mock_la_server.rs
@@ -40,12 +40,12 @@ use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::Duration;
 
+use axum::Router;
 use axum::body::Bytes;
 use axum::extract::{Path, State};
 use axum::http::{HeaderMap, StatusCode, header};
 use axum::response::IntoResponse;
 use axum::routing::post;
-use axum::Router;
 use clap::Parser;
 use flate2::read::GzDecoder;
 use rand::RngExt;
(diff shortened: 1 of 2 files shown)


lquerel (approved):
LGTM

lalitb:
Nice, I would like to have this extended as subsequent PR for the Geneva Exporter.

cijothomas:
> Nice, I would like to have this extended as subsequent PR for the Geneva Exporter.

its easy to add a similar one inside Geneva Exporter as an example I think. (this is very useful to simulate and test behavior when server is slow/returns 500 etc.)

gouslu:
Thanks for this change! This looks good, I think we should consider adding all applicable API limits, but this is great in itself and checks the most important thing, the payload size.

https://learn.microsoft.com/en-us/azure/azure-monitor/fundamentals/service-limits#logs-ingestion-api

PR #2383, opened 2026-03-20
Batch processor: Nack when inbound or outbound slots are full

### LINKED ISSUE #2194: Batch processor returns fatal error on inbound slot exhaustion and crashes engine

jmacd:
# Change Summary

Fixes engine-level error when the batch_processor inbound/outbound slots are full.

For inbound failures: Nack the message and do not admit.
For outbound failures: Nack the message but let the request fly.

Adds two new metrics for these cases.

Adds a simpler SlotMap allocator: allocate_with_data() makes the two callsites in batch_processor easier to read.

Adds several notes and TODOs:

- It is easy for callers to mistakenly invoke route_ack() and route_nack(), but it breaks instrumentation b/c the timing measurement is not captured.
- Failure to Ack/Nack are engine-level errors; we should have something consistent for dealing with lost Ack/Nack messages.

Fixes #2194

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
@@ -645,7 +645,7 @@ impl BatchProcessor {
         if items == 0 {
             self.metrics.dropped_empty_records.inc();
             // Note: Failure to Ack/Nack is an engine-level error.
-            let _ = effect.notify_ack(AckMsg::new(request)).await?;
+            effect.notify_ack(AckMsg::new(request)).await?;
             return Ok(());
         }
 
@@ -788,7 +788,7 @@ where
                     self.metrics.nacked_inbound_slots.inc();
                     let refused = OtapPdata::new(bctx.ctx, payload.into());
                     // Note: Failure to Ack/Nack is an engine-level error.
-                    let _ = effect
+                    effect
                         .notify_nack(NackMsg::new("inbound routes exhausted", refused))
                         .await?;
                     // Note: Inbound slot exhaustion leads to dropping
@@ -960,7 +960,7 @@ where
                             {
                                 self.metrics.nacked_outbound_slots.inc();
                                 // Note: Failure to Ack/Nack is an engine-level error.
-                                let _ = effect
+                                effect
                                     .notify_nack(NackMsg::new(
                                         "outbound routes exhausted",
                                         OtapPdata::new(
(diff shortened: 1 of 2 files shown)


rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
```diff
@@ -514,6 +514,12 @@ pub struct BatchProcessorMetrics {
     /// Number of empty records dropped
     #[metric(unit = "{msg}")]
     dropped_empty_records: Counter<u64>,
+    /// Number of requests nacked due to inbound slot exhaustion
+    #[metric(unit = "{msg}")]
+    nacked_inbound_slots: Counter<u64>,
+    /// Number of requests nacked due to inbound slot exhaustion
+    #[metric(unit = "{msg}")]
+    nacked_outbound_slots: Counter<u64>,
```
lquerel, line 522: Probably for a future PR, but we could track these two metrics in the effect handlers so that nodes don't have to implement this instrumentation manually.

I'm adding that in #2370 .

lquerel (approved):
LGTM

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
jmacd: I covered this in #2386

PR #2390, opened 2026-03-20
feat: Log sampling processor

### LINKED ISSUE #2382: Simple logs subsampling processor
### Pre-filing checklist

### Component(s)

Sampling is a valuable tool to use to reduce telemetry volume. There are many ways to implement sampling, and each signal has its own individual challenges.

This is a proposal for a very simple processor that focuses only on logs as a first step.

#### Log Subsampling Processor

URN: `urn:otel:processor:log_subsampling`

##### Overview

The Log Subsampling processor reduces log volume by discarding a portion of incoming log records according to a configurable subsampling policy. The processor treats all incoming log records as equal and makes no attempt to classify them further. Pipeline administrators are expected to configure their pipeline such that all logs reaching a processor instance can be considered equivalent.

In a typical deployment, some routing node (perhaps using an OPL program to do classification) sits upstream of this processor.  The router classifies incoming telemetry and directs logs to a the appropriate pipeline branch where the log subsampling instance removes some the traffic.

##### Signal Handling

| Signal  | Behavior               |
|---------|------------------------|
| Logs    | Apply subsampling      |
| Metrics | Pass through unchanged |
| Traces  | Pass through unchanged |

Non-log signals are forwarded downstream without modification. This makes the processor resilient to pipeline misconfiguration where metrics or traces are inadvertently routed through it.

##### Configuration

The processor configuration uses an externally tagged enum on the `policy` field. Exactly one policy must be specified. We will support two initial sampling policies.

###### Zip Sampling

Emit at most N log records per time window.

```yaml
config:
  policy:
    zip:
      interval: 60s
      max_items: 100
```

| Field       | Type     | Required | Description                    |
| ----------- | -------- | -------- | ------------------------------ |
| `interval`  | duration | yes      | Length of the sampling window  |
| `max_items` | integer  | yes      | Max records to emit per window |

Constraints: `interval > 0`, `max_items > 0`.

###### Ratio Sampling

Emit the first N (`emit`) out of every M (`out_of`) logs.

```yaml
config:
  policy:
    ratio:
      emit: 1
      out_of: 10
```

| Field    | Type    | Required | Description                          |
|----------|---------|----------|--------------------------------------|
| `emit`   | integer | yes      | Numerator of the sampling fraction   |
| `out_of` | integer | yes      | Denominator of the sampling fraction |

Constraints: `emit > 0`, `out_of > 0`, `emit <= out_of`.

###### Record Selection

Policies are evaluated against each incoming OTAP batch. Some number of records are selected to be kept based on the policy and the rest are discarded. 

###### Zip Sampling Algorithm

The zip sampler tracks a `count` of records emitted in the current window and resets it to zero when the window elapses. We'll implement the timer via the engine's TimerTick message.

At startup (or on first message), the processor calls `effect_handler.start_periodic_timer(interval)` to register a
per-node periodic timer with the engine. The engine delivers `NodeControlMsg::TimerTick` at a fixed cadence.

On `TimerTick`: reset `count = 0`. 

On each incoming log batch of size B:

1. Compute `budget = max_items - count`.
2. Compute `to_keep = min(budget, B)`.
3. Update `count += to_keep`.
4. If `to_keep > 0`, slice and forward the first `to_keep` rows.
5. If `to_keep == 0`, ack the batch immediately (see Ack/Nack section).

Ticks will continue to fire during idle periods, but the reset on an already-zero count is a no-op.

###### Ratio Sampling Algorithm

The ratio sampler keeps M (`emit`) out of every N (`out_of`) records.

The exact way logs are selected is an implementation detail. If the incoming Logs batch is larger than N, we can try to consolidate into a single split.

For example, if we're keeping 7 out of every 10 logs and batch of 100 comes in, we can do a single slice to keep the first 70 logs rather than try to pick out 7 logs and skip 3 repeatedly.

The calculation is something roughly like the following where `seen` tracks how many records we've seen in the current cycle and `emitted` is how many we've already emitted in the current cycle.

```text
remaining_in_cycle = out_of - seen
from_current       = min(B, remaining_in_cycle)
keep_from_current  = min(max(emit - emitted, 0), from_current)

after_current = B - from_current
full_cycles   = after_current / outdo
leftover      = after_current % out_of

to_keep = keep_from_current + (full_cycles * emit) + min(emit, leftover)
```

After processing, update state:

```text
new_seen    = (seen + B) % out_of
new_emitted = min(emit, new_seen)
```

If `to_keep > 0`, slice and forward the first `to_keep` rows.
If `to_keep == 0`, ack the batch immediately (see Ack/Nack section).

Example with `emit: 2, out_of: 5`:

| Batch | Size | State before      | to_keep | State after       |
| ----- | ---- | ----------------- | ------- | ----------------- |
| 1     | 12   | emitted=0, seen=0 | 6       | emitted=2, seen=2 |
| 2     | 4    | emitted=2, seen=2 | 1       | emitted=1, seen=1 |
| 3     | 5    | emitted=1, seen=1 | 2       | emitted=1, seen=1 |
| 4     | 4    | emitted=1, seen=1 | 1       | emitted=2, seen=4 |

Total records in: 12 + 4 + 5 + 4 = 25
Total records emitted = 6 + 1 + 2 + 1 = 10 (exactly 2 out of 5)

##### Ack/Nack

By deploying a subsampling processor, the pipeline operator has explicitly opted into data loss as a volume reduction strategy. Records are expected to be regularly discarded and so we will limit participation in ack/nack.

For each incoming otap batch, if all records are dropped (to_keep == 0) then the processor immediately acks the inbound request by calling `notify_ack(AckMsg::new(pdata)) `. No data is sent downstream. 

If the there's an empty incoming batch (num_items == 0):** then we also Ack immediately. Otherwise, skip participating in the ack/nack chain.

The main thing we're exposing ourselves to here is the double-subsampling problem. If a downstream node nacks the subsampled data, the nack propagates upstream past the subsampling processor (since it did not subscribe). If a retry processor upstream re-sends the data, it re-sends its own saved copy of the original records. Those records arrive at the subsampling processor as new input and are subsampled again. For example, with a 1:10 ratio, 100 original records become 10 on the first attempt. On retry, those 10 are subsampled again to 1. The effective ratio on retry becomes 1:100 instead of 1:10.

To mitigate this, Pipeline operators can place a retry processor after the subsampling processor so that retried data is not re-subsampled. However, this does constrain pipeline topology.

Full Ack/Nack participation is discussed in the alternatives section.

##### Telemetry

```rust
#[metric_set(name = "log_subsampling.processor.pdata.metrics")]
pub struct LogSubsamplingMetrics {
    /// Total log records received
    #[metric(unit = "{log}")]
    pub log_signals_consumed: Counter<u64>,

    /// Log records dropped by subsampling
    #[metric(unit = "{log}")]
    pub log_signals_dropped: Counter<u64>,

    /// Batches where all records were dropped (acked immediately)
    #[metric(unit = "{batch}")]
    pub batches_fully_dropped: Counter<u64>,
}
```

##### Format handling 

Initially we will only operate on OtapArrowRecords and if the incoming payload
 is OTLP bytes, we convert to Arrow format first. Adding OTLP support later will
be fairly straightforward.

#### Alternative Zip Sampling timer implementation

I think this is worse than letting the engine handle as it's much harder to test. Abstracting out time will make it very easy to control the internal counter resets.

To implement we'd track two pieces of state: `window_expiry` (Instant), `count` (usize).

The window is checked and reset on demand when a batch arrives. On each incoming log batch of size B:

1. If `now >= window_expiry`, reset `count = 0` and set
   `window_expiry = now + interval`.
2. Compute `budget = max_items - count`.
3. Compute `to_keep = min(budget, B)`.
4. Update `count += to_keep`.
5. If `to_keep > 0`, slice and forward the first `to_keep` rows.
6. If `to_keep == 0`, ack the batch immediately (see Ack/Nack section).

Windows are anchored to traffic arrival, not the wall clock. If no data arrives for several intervals, only a single reset occurs on the next arrival. Unused budget from past windows does not carry forward. No timer lifecycle or control message handling is required.

#### Full Ack/Nack Participation

If we wanted to fully participate in ack/nack and not affect the sampling outcome then the cost is complexity.

The processor would subscribe to downstream ack/nack interests and maintain a slot-based correlation between inbound and outbound messages. This prevents double-subsampling on retry.

When `to_keep > 0`:

1. If the incoming pdata has subscribers (`context.has_subscribers()`),
   allocate a slot storing the original `Context`.
2. Call `effect.subscribe_to(Interests::ACKS | Interests::NACKS,
   slot_key.into(), &mut pdata)` to push a frame onto the outgoing
   context.
3. Send the subsampled data downstream.

The processor does NOT subscribe with `Interests::RETURN_DATA`. Only the original `Context` is stored -- not the Arrow record payload. This keeps memory overhead minimal.

On `Ack`: look up the slot by calldata, retrieve the original `Context`, and call `notify_ack` with an empty-payload `OtapPdata` constructed from the original context.

On `Nack`: same as ack, but call `notify_nack` instead. The upstream retry processor holds its own copy of the original data (via its own `RETURN_DATA` subscription) and will re-send the full original records. The subsampling processor then re-subsamples from the full original rather than from the already-subsampled result.

**Counter state on retry:** Because the subsampling counters (zip window count, ratio emitted/seen) continue to evolve between the original send and the retry, the retry may produce a different `to_keep` count than the original attempt. 

#### Additional Sampling policies

Sampling policy can be defined in many ways with many levels of sophistication. The sampling policy enum can be extended as needed in the future to support more options such as probabilistic sampling or adjustments to how records are selected from a batch.

### Additional Context

_No response_

JakeDern:
# Change Summary

This PR adds a very simple log sampling processor that's intended to be extensible and leave room for more sophisticated samplers in the future.

* Closes #2382

Unit tests added - I also ran it locally and observed the metrics in the dashboard. This was with an `emit: 1, out_of: 10` and it dropped precisely 90% according to the reported metrics.

image

And the debug processor reported getting batches of exactly 10 which is expected given traffic generator was spitting out batches of 100:

image

Yes, new processor!

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/mod.rs
@@ -6,8 +6,6 @@
 //! Reduces log volume by discarding a portion of incoming log records
 //! according to a configurable sampling strategy. Non-log signals
 //! (metrics and traces) pass through unchanged.
-//!
-//! See `README.md` in this module for full design documentation.
 
 mod config;
 mod metrics;
@@ -124,7 +122,7 @@ impl LogSamplingProcessor {
         // resized appropriately the next time around.
         //
         // This _should_ succeed if the sampler called finish as the builder
-        // will drop its buffer.
+        // will drop its reference.
         let (buf, _) = selection.into_parts();
         match buf.into_inner().into_mutable() {
             Ok(mutable) => self.filter_buffer = mutable,
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/samplers/ratio.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/samplers/ratio.rs
@@ -3,13 +3,6 @@
 
 //! Ratio sampler -- emit a fixed fraction of log records.
 //!
-//! # Algorithm
-//!
-//! The ratio sampler keeps `emit` out of every `out_of` records.
-//! Conceptually, for each record: increment `seen`, and if
-//! `emitted < emit` then keep the record and increment `emitted`. When
-//! `seen` reaches `out_of`, reset both counters to zero.
-//!
 //! At the batch level, the number of records to keep is computed in O(1)
 //! using a closed-form formula rather than iterating per record. See
 //! [RatioSampler::compute_to_keep].
@@ -90,6 +83,9 @@ impl RatioSampler {
     /// Uses an O(1) closed-form formula that accounts for the current
     /// position within the emit/out_of cycle, any number of full cycles
     /// within the batch, and the leftover partial cycle at the end.
+    ///
+    /// Note: This means that we sample every record out of the front of the
+    /// incoming batch if `out_of` is greater than the incoming batch size.
     ///
     /// ## Example
     ///
--- a/rust/otap-dataflow/crates/core-nodes/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/README.md
@@ -30,6 +30,7 @@ Each component lives in its own subfolder within a category:
         durable_buffer_processor/
         fanout_processor/
         filter_processor/
+        log_sampling_processor/
         retry_processor/
         signal_type_router/
         transform_processor/
@@ -123,6 +124,7 @@ Each component lives in its own subfolder within a category:
 | durable_buffer_processor | `urn:otel:processor:durable_buffer` | `src/processors/durable_buffer_processor/` |
 | fanout_processor | `urn:otel:processor:fanout` | `src/processors/fanout_processor/` |
 | filter_processor | `urn:otel:processor:filter` | `src/processors/filter_processor/` |
+| log_sampling_processor | `urn:otel:processor:log_sampling` | `src/processors/log_sampling_processor/` |
 | retry_processor | `urn:otel:processor:retry` | `src/processors/retry_processor/` |
 | signal_type_router | `urn:otel:processor:type_router` | `src/processors/signal_type_router/` |
 | transform_processor | `urn:otel:processor:transform` | `src/processors/transform_processor/` |
@@ -172,6 +174,11 @@ Each component lives in its own subfolder within a category:
 - Filters OTAP signals according to configured trace/log filter rules
 - Tracks consumed and filtered signal metrics for telemetry
 - Useful for reducing data volume before downstream processing
+
+#### log_sampling_processor
+
+- Discards a portion of incoming logs according to a configurable sampling policy
+- Useful for reducing data volume in a telemetry backend
 
 #### retry_processor
 


rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/samplers/ratio.rs
```diff
@@ -0,0 +1,311 @@
...
+        to_keep
+    }
+}
+
+#[async_trait(?Send)]
+impl Sampler for RatioSampler {
+    fn sample_arrow_records(&mut self, records: &OtapArrowRecords) -> BooleanArray {
+        let total = records.root_record_batch().map_or(0, |rb| rb.num_rows());
+        let to_keep = self.compute_to_keep(total);
+        BooleanArray::new(BooleanBuffer::collect_bool(total, |i| i < to_keep), None)
```
lalitb: This seems to be allocating a new buffer on every batch. Since the selection is always a contiguous prefix of true values, we could possibly reuse the buffer.

lalitb: nit: The doc above describes keeping records based on their position in the emit/out_of cycle, but the code always just keeps the first N. That's fine  for log sampling since the order doesn't really matter, but the doc should probably say "keeps a prefix" to match what actually happens.

rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/samplers/ratio.rs
```diff
@@ -0,0 +1,311 @@
...
+    /// ## Example
+    ///
+    /// With `emit: 2, out_of: 5`:
+    ///
+    /// | Batch | Size | State before      | to_keep | State after       |
+    /// |-------|------|-------------------|---------|-------------------|
+    /// | 1     | 12   | emitted=0, seen=0 | 6       | emitted=2, seen=2 |
+    /// | 2     | 4    | emitted=2, seen=2 | 1       | emitted=1, seen=1 |
+    /// | 3     | 5    | emitted=1, seen=1 | 2       | emitted=1, seen=1 |
+    /// | 4     | 4    | emitted=1, seen=1 | 1       | emitted=2, seen=4 |
```
lalitb: Can you revisit this example, specifically batch 4. The state after seems to be bit off.

rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/mod.rs
```diff
@@ -0,0 +1,448 @@
...
+                    let source_detail = format_error_sources(&e);
+                    EngineError::ProcessorError {
+                        processor: effect_handler.processor_id(),
+                        kind: ProcessorErrorKind::Other,
+                        error: format!("failed to convert payload to arrow records: {e}"),
+                        source_detail,
+                    }
+                })?;
+
+        let selection = self.sampler.sample_arrow_records(&records);
```
lalitb: I think we should call `decode_transport_optimized_ids()` on the records after `try_into()` - the filter processor does this too - https://github.com/open-telemetry/otel-arrow/blob/7c80ef943e417c5671b6dcb028cccc8d89935525/rust/otap-dataflow/crates/core-nodes/src/processors/filter_processor/mod.rs#L133

Without it, the delta-encoded IDs from transport-optimized Arrow input won't match correctly when filtering child batch. Also worth adding a test that exercises sampling against transport-optimized Arrow logs with child attribute batches.

rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/samplers/ratio.rs
JakeDern: Ah yes, I thought I had a note on this somewhere maybe just in the attached issue.

There's a discussion about naming and also about whether keeping the first N is part of the contract or not. Will leave this open and do a pass on docs. Either way will make sure it's mentioned somewhere in the implementation.

rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/samplers/ratio.rs
albertlockett: Are we sure the docs/logic here is right?

For the last line:
```rs
        // Update state
        let new_seen = (self.seen + batch_size) % self.out_of; // (1 + 4) % 5 = 5 % 5 = 0
        self.seen = new_seen;  // 0
        self.emitted = self.emit.min(new_seen); // 2.min(0) = 0
```

JakeDern: Yeah lalit made the same comment. I changed up the example to make it a round number and then did bad math on the last one. Just pushed the updated docs

JakeDern: Fixed

lalitb (approved):
LGTM, once this is fixed- https://github.com/open-telemetry/otel-arrow/pull/2390#discussion_r2976708864

rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/samplers/ratio.rs
JakeDern: Ah yeah @albertlockett and I also briefly discussed this idea. It would be nice to have the processor manage the storage and pass it into the samplers, but it's not super straightforward to do because Arrow arrays are immutable.

We can only try to reclaim the underlying MutableBuffer from the BooleanBuffer with `into_inner().into_mutable()` if there is only a single reference, but then we have to have the samplers work with either MutableBuffer directly or with BooleanBufferBuilder which might be ok.

I'll poke at it a little bit and see how it looks.

rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/mod.rs
JakeDern: Good call, thank you - Fixed.

rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/samplers/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+use arrow::array::BooleanArray;
+use async_trait::async_trait;
+use otap_df_engine::error::Error as EngineError;
+use otap_df_engine::local::processor as local;
+use otap_df_otap::pdata::OtapPdata;
+use otap_df_pdata::otap::OtapArrowRecords;
+
+/// Trait for log sampling strategies.
+#[async_trait(?Send)]
+pub trait Sampler: std::fmt::Debug {
```
lquerel, line 25: Nice

rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/README.md
```diff
@@ -0,0 +1,80 @@
+# Log Sampling Processor
+
+URN: `urn:otel:processor:log_sampling`
+
+## Overview
```
lquerel, line 5: [in addition to this doc] This new processor should be added to one of the higher-level `README.md` files that lists the nodes supported by the engine.

drewrelmas:
@JakeDern perhaps a silly question, but is there a specific motivation for intentionally only supporting sampling of logs as opposed to adding a `sampling_processor` which only supports Log signals for the moment?

EDIT: On a moment's reflection, I can understand rationale for not sampling metrics because of their nature, but I am less knowledgeable about Traces. If anyone more versed would like to educate me, would be happy to listen!

JakeDern:
> @JakeDern perhaps a silly question, but is there a specific motivation for intentionally only supporting sampling of logs as opposed to adding a `sampling_processor` which only supports Log signals for the moment?
> 
> EDIT: On a moment's reflection, I can understand rationale for not sampling metrics because of their nature, but I am less knowledgeable about Traces. If anyone more versed would like to educate me, would be happy to listen!

This is not 100% my wheelhouse either 😄 - But like you mentioned for metrics we're planning to do temporal re-aggregation instead of sampling which would have different mechanics.

I think for spans sampling is tricky too. IIRC one challenge there is making sure you consistently sample all spans for the same Trace ID so that when your frontend calls microservice X which calls microservice Y (and the spans are all exported at a different time) you end up with a full end to end trace of the system in your telemetry backend and not a bunch of fragments like you'd get if you randomly sampled spans.

JakeDern:
@jmacd Any final feedback on naming the policies? I left them as-is just to see if anyone else wanted to chime in. I agree with your points on the issue especially about ratio perhaps being too broad. We could go with something much more literal for both.

Some zip alternatives:

- `top_n_per_period`
- `systematic` (your suggestion)

Some ratio alternatives:

- `take_n_of_m`
- `n_of_m`

lquerel:
@JakeDern @jmacd 

If we’re going for the most explicit option, I’d vote for `top_n_per_period`.
And for the second policy, any alternatives work for me.

rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/samplers/ratio.rs
JakeDern: This worked out better than I thought it would actually - The builder had a pretty reasonable interface. I added a metric to track failures to reclaim the underlying buffer and verified it was happening, see https://github.com/open-telemetry/otel-arrow/pull/2390/changes/f16fb40c62e5f932e879d59acb2e8287d5d22e7f

image

JakeDern: Done.

rust/otap-dataflow/crates/core-nodes/src/processors/log_sampling_processor/README.md
JakeDern: Done

PR #2392, opened 2026-03-20
Pipeline/group/engine policy precedence: prevent against misuse of unresolved policies

### LINKED ISSUE #2389: Pipeline-level policy overrides engine-level policy
### Pre-filing checklist

### Component(s)

### Bug Description

Like #2154 but impacting telemetry. We need an `Option<>` for policy fields so they merge correctly, otherwise an omitted value in the pipeline-level will override the engine-level because `#[serde(default)]`.

This likely also impacts channel capacity and health settings. Note: there are correct `resolve()` methods in place, but because the fields were `pub` it is possible to access them incorrectly.

### Steps to Reproduce

See #2154, for telemetry policies.

### Expected Behavior

These values take effect when not set in the pipeline or group level.
```
policies:
  channel_capacity:
    ...
  telemetry:
    channel_metrics: detailed
```

### Actual Behavior

The detailed metrics setting does not take.

### OTel-Arrow Version

main

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

jmacd:
# Change Summary

Like #2154 but for the other three policy fields. Make all fields Option types. Adds a ResolvedPolicies type which strips the Options after resolving. There was existing resolve code, but it was not used consistently: this was observed for the `telemetry` policy.

Fixes #2389.

One new test. The `configs/internal-telemetry.yaml` configuration is modified to show the problem. Before the fix, no duration metrics. After the fix, duration metrics, as set by the top-level policy.

### DIFF
--- a/rust/otap-dataflow/crates/config/src/engine.rs
+++ b/rust/otap-dataflow/crates/config/src/engine.rs
@@ -666,6 +666,154 @@ groups:
     }
 
     #[test]
+    fn resolve_policies_mixed_engine_and_group_inheritance() {
+        // The reviewer asked for a test where some policies come from the
+        // group level and others from the engine (top) level, exercising
+        // the full three-level precedence chain.
+        let yaml = r#"
+version: otel_dataflow/v1
+policies:
+  telemetry:
+    channel_metrics: detailed
+  channel_capacity:
+      control:
+        node: 500
+        pipeline: 501
+      pdata: 502
+  health:
+    ready_if: [Running, Updating]
+engine: {}
+groups:
+  default:
+    policies:
+      # Group sets telemetry and health but NOT channel_capacity.
+      telemetry:
+        channel_metrics: normal
+        pipeline_metrics: false
+      health:
+        ready_if: [Failed]
+    pipelines:
+      # Pipeline sets only channel_capacity → gets telemetry from group,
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/config/src/engine.rs
```diff
@@ -533,11 +534,137 @@ groups:
...
         );
         assert_eq!(
-            p3_resolved.policies.effective_resources().core_allocation,
+            p3_resolved.policies.resources.core_allocation,
             crate::policy::CoreAllocation::CoreCount { count: 9 }
         );
     }
 
+    #[test]
+    fn resolve_policies_inherit_from_parent_scope() {
```
gouslu, line 543: are there any tests for when inheritance comes partially from engine level and partially from group level? if not, it is worth adding.

gouslu:
If applicable, there should be a test where partial configration from different tiers are taken in a way that it validates the fact that they are merged at the lowest level. And there should also be a test that validates both they are merged and the inner value wins and overrides an outer value.

rust/otap-dataflow/crates/config/src/engine.rs
jmacd: Done

PR #2393, opened 2026-03-20
Use OTAP spec aware `concatenate` when producing the results of `if`/`else` statements

### LINKED ISSUE #2216: Columnar Query Engine conditional if/else expression needs to account for schema modifications during concat
In the if/else expression evaluation by the columnar query engine, it splits the batch, and each branch of the statement evaluation processes the data, then the results of each branch are concatenated together. Currently we're using a naive concatenation strategy

otel-arrow/rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs](https://github.com/open-telemetry/otel-arrow/blob/63a23cf282c43a3dceb453f6fd17c794a4e9bb70/rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs#L190-L207

However, we'e now merged some PRs that can modify the arrow schema (see https://github.com/open-telemetry/otel-arrow/pull/2159), we should use the more sophisticated OTAP batch concat that is also used by the batch processor.

albertlockett:
# Change Summary

In the columnar query engine, when we write `if`/`else` statements in OPL, the results of each branch are concatenated together. Before this change, we were simply using arrow's `concat_batches` helper function with expects all the `RecordBatch`s to have the same schema. However, this would cause a problem if some branch of the statement changes the schema.

This PR corrects the issue by using OTAP's [`concatenate`](https://github.com/open-telemetry/otel-arrow/blob/eaa4103326057ef68125244171801bc010cb3571/rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs#L75) function instead which correctly expands each `RecordBatch` into a common schema.

There's one pipeline stage that also writes new IDs to the rows with null IDs (this happens when we assign attributes). In order for `concatenate` to produce a valid batch, we need to ensure the IDs are globally unique. This PR adds a mechanism to initialize shared state for different implementations of the same pipeline stage if they're being used in a nested branch within conditional pipeline stage, and uses it for the purpose of ensuring unique IDs when filling in these null rows.

* Closes #2216 

Unit tests

No

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
@@ -1028,6 +1028,17 @@ impl PipelineStage for AssignPipelineStage {
 
         Ok(())
     }
+
+    fn clear_state_for_conditional_branch(
+        &mut self,
+        exec_state: &mut ExecutionState,
+    ) -> Result<()> {
+        // if we've added the NextIdTracker, we'll need to remove it. Otherwise, if the
+        // ExecutionState is reused between batches, it will not be reinitialized for the next
+        // incoming OTAP batch
+        _ = exec_state.remove_extension::<NextIdTracker>();
+        Ok(())
+    }
 }
 
 /// Extension implementation used to keep track of the next max ID when the ID column
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs
@@ -269,6 +269,19 @@ impl PipelineStage for ConditionalPipelineStage {
             branch_results.push(default_branch_batch);
         }
 
+        // give the pipeline stages within each branch the opportunity to clear any
+        // state that was initialized for this batch.
+        for branch in &mut self.branches {
+            for stage in &mut branch.pipeline_stages {
+                stage.clear_state_for_conditional_branch(exec_state)?;
+            }
+        }
+        if let Some(branch) = self.default_branch.as_mut() {
+            for stage in branch {
+                stage.clear_state_for_conditional_branch(exec_state)?;
+            }
+        }
+
         // reconstruct the result with the results of each branch
         match otap_batch {
             OtapArrowRecords::Logs(_) => concatenate_logs(&mut branch_results),
@@ -391,24 +404,31 @@ impl PipelineStage for ConditionalPipelineStage {
 mod test {
     use crate::pipeline::{
         Pipeline,
-        test::{exec_logs_pipeline, exec_metrics_pipeline, exec_traces_pipeline},
+        test::{
+            exec_logs_pipeline, exec_metrics_pipeline, exec_traces_pipeline, otap_to_logs_data,
+        },
     };
+    use arrow::array::UInt16Array;
     use data_engine_parser_abstractions::Parser;
     use otap_df_opl::parser::OplParser;
-    use otap_df_pdata::{
-        proto::opentelemetry::{
-            common::v1::{AnyValue, KeyValue},
-            logs::v1::LogRecord,
-            trace::v1::Span,
-        },
-        testing::round_trip::{to_metrics_data, to_traces_data},
-    };
     use otap_df_pdata::{
         proto::opentelemetry::{
             metrics::v1::Metric,
             trace::v1::{Status, span::SpanKind},
         },
+        schema::consts,
         testing::round_trip::to_logs_data,
+    };
+    use otap_df_pdata::{
+        proto::{
+            OtlpProtoMessage,
+            opentelemetry::{
+                common::v1::{AnyValue, KeyValue},
+                logs::v1::LogRecord,
+                trace::v1::Span,
+            },
+        },
+        testing::round_trip::{otlp_to_otap, to_metrics_data, to_traces_data},
     };
 
     use super::*;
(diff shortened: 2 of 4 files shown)


rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs
```diff
@@ -65,6 +71,54 @@ impl ConditionalPipelineStage {
...
+fn concatenate_metrics(branch_results: &mut Vec<OtapArrowRecords>) -> Result<OtapArrowRecords> {
+    concat_generic::<Metrics, { METRICS_TYPE_MASK }, { Metrics::COUNT }>(branch_results)
+}
+
+fn concatenate_traces(branch_results: &mut Vec<OtapArrowRecords>) -> Result<OtapArrowRecords> {
+    concat_generic::<Traces, { TRACES_TYPE_MASK }, { Traces::COUNT }>(branch_results)
+}
+
+fn concatenate_attrs_record_batches(branch_results: &mut Vec<RecordBatch>) -> Result<RecordBatch> {
+    let mut batches = branch_results
```
JakeDern: Beware, today you still have to call reindex before concatenating! I have it on my todo to fix this when I finally eliminate groups.rs, but it'd be very quick to make that small change here as well if you wanted to 

https://github.com/open-telemetry/otel-arrow/blob/eaa4103326057ef68125244171801bc010cb3571/rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs#L41-L45

rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs
```diff
@@ -65,6 +71,54 @@ impl ConditionalPipelineStage {
...
+        .drain(..)
+        .map(|batch| [Some(batch)])
+        .collect::<Vec<_>>();
+
+    let mut result =
+        concatenate_with_def(&mut batches, |_| payloads::get(ArrowPayloadType::LogAttrs))?;
```
JakeDern: Is it possible to shove the attributes record batches into a `Logs` and work with the top level concatenate function and then pull it out of the LogAttrs slot? This `concatenate_with_def` function is admittedly a little bit ugly and maybe better to keep internal 😅

rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs
albertlockett: So it turns out in this case we actually do want to concat without reindexing. We've split the batch first, so when we concatenate the splits back together we want to keep the same IDs.

If we call `reindex`, we'd create new scopes/resources because the IDs of these columns get incremented when in reality the batches we're concatenating may contain signals belonging to the same scope/resource.

One thing I did fix however is that when we assign attributes to something that had a null `id`, we fill in the ID column. I've added code in this PR to keep a global view of the ID column across branches, so as we're filling in the next available ID there are no duplicates created.

rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs
albertlockett: Made this change! I'm not sure it made the code cleaner at the call site, but at least it avoids us making `concatenate_with_def` public when we maybe don't want it to be so.

JakeDern: Yeah, definitely puts some more burden on the caller especially in this case. Definitely a goal to have a cleaner API here and also design for concatenating a single payload type rather than an entire otap batch.

rust/otap-dataflow/crates/query-engine/src/pipeline/conditional.rs
JakeDern: Oh interesting, so we have a need to concatenate individual payloads in addition to entire OtapBatches and with/without reindexing first. I'll keep this in mind when I take a look at the interface to this module.

JakeDern (approved):

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -999,6 +1006,60 @@ impl PipelineStage for AssignPipelineStage {
...
+        // this pipeline stage may be seeing a different subset of the overall batch, but we need
+        // to ensure the IDs that are assigned are not duplicated across branches. That is why we
+        // add this extension.
+        if let ColumnAccessor::Attributes(attrs_id, _) = &self.dest_columns[0] {
+            if *attrs_id == AttributesIdentifier::Root
+                && exec_state.get_extension::<NextIdTracker>().is_none()
+            {
+                let next_id_tracker = NextIdTracker::try_new(otap_batch)?;
+                exec_state.set_extension(next_id_tracker);
+            }
```
lquerel, line 1026: I'm probably missing some context, but while reading this code I was wondering whether there might be a lifecycle mismatch here between the stage and the `NextIdTracker`.  

I thought the stage's lifecycle was tied to the processor's lifecycle, whereas in this code the `NextIdTracker` seems to be initialized with the current batch.  

Should the tracker be reinitialized for each new batch?

albertlockett: Thanks @lquerel . Yeah this is a good catch. 

Technically the lifecycle of these extensions are scoped to the `ExecutionState`, whose lifecycle can be managed externally from the query-engine pipeline. In the transform processor we reuse this across batches, so definitely the `NextIdTracker` needs to be reinitialized. Pushed a fix in 7da7a0e that resolves the issue

PR #2394, opened 2026-03-20
[engine] Split runtime control and return paths for receiver-first drain

lquerel:
# Change Summary

This PR lands the behavioral core of the ack/nack and graceful shutdown redesign.

It splits the old shared pipeline control path into two distinct paths:
- a runtime-control path used for shutdown orchestration, timers, delayed data, and receiver drain coordination
- a pipeline return path used for Ack/Nack delivery and unwind

It also introduces a receiver-first shutdown sequence:
- `ShutdownRequested`
- `DrainIngress` to receivers
- `ReceiverDrained` from receivers
- downstream `Shutdown` to the rest of the pipeline

The goal of this PR is to establish the new control-plane behavior with the minimum runtime, controller, and receiver/exporter adaptations needed to make that behavior work. Naming cleanup, docs churn, observability, DST, and broader liveness coverage are intentionally left to follow-up PRs.

* Closes #NNN

- `cargo xtask check`
- focused engine/control-plane correctness tests covering shutdown, drain sequencing, and Ack/Nack return-path behavior
- receiver and exporter regression coverage exercised through the workspace test suite

There are no intended new public docs or observability surfaces in this PR.

Behaviorally, this changes pipeline runtime semantics during graceful shutdown and Ack/Nack propagation:
- shutdown is now receiver-first
- Ack/Nack delivery is isolated from runtime-control traffic
- admitted work is less likely to be affected by mixed control-path interference during drain

### DIFF
--- a/rust/otap-dataflow/crates/validation/src/simulate.rs
+++ b/rust/otap-dataflow/crates/validation/src/simulate.rs
@@ -48,7 +48,7 @@ pub(crate) async fn run_pipelines_with_timeout(
         .await?;
         sleep(propagation_delay).await;
         let result = ensure_validation_passed(&admin_client, &admin_base).await;
-        shutdown_pipeline(&admin_client, &admin_base).await?;
+        shutdown_pipeline(&admin_client, &admin_base, timeout).await?;
         result
     })
     .await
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/config/src/policy.rs
```diff
@@ -222,13 +227,17 @@ pub struct ControlChannelCapacityPolicy {
     /// Capacity used for pipeline control channels.
     #[serde(default = "default_pipeline_control_channel_capacity")]
     pub pipeline: usize,
+    /// Capacity used for return-path Ack/Nack control channels.
+    #[serde(default = "default_return_control_channel_capacity", rename = "return")]
+    pub r#return: usize,
```
lquerel, line 232: This field will be renamed completion in a follow up PR.

PR #2395, opened 2026-03-20
fix: Move batch processor to 100klrps scenario and add otap-batch-otap

### LINKED ISSUE #2277: Batch processor continuous benchmarks
Tracking need for continuous benchmarks for both:

OTAP->Batch->OTAP
OTLP->Batch->OTLP

See here: https://github.com/open-telemetry/otel-arrow/pull/2246#issue-4048885041

JakeDern:
# Change Summary

This PR moves the continuous batch processor benchmarks to the 1ooklrps scenario and adds an otap-batch-otap configuration.

I think the batch processor benchmarks were mistakenly added to the "passthrough" scenario which states it's for scenarios with no processor in the middle. The dashboard also does not seem to be set up properly for these and we want to add otap-batch-otap as mentioned here: https://github.com/open-telemetry/otel-arrow/pull/2246#issuecomment-4033755985

- Closes #2277


JakeDern:
CC: @cijothomas

PR #2396, opened 2026-03-20
fix: Temporarily disable the nightly otap-filter-otap Go collector scenario

JakeDern:
# Change Summary

This scenario has been blocking all the nightly benchmarks for a few weeks now and we can't fix it until this is released and we take a version bump: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/46879

It looks like it will be another couple of weeks for the next otel collector contrib release as the last one was just a few days ago. I'm proposing to disable the scenario for now to unblock everything else.


JakeDern:
CC: @cijothomas

Issue #2270, opened 2026-03-11
Missing attribute after decoding OTAP to OTLP


albertlockett:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Bug Description

I noticed a case where when converting OTAP to OTLP we encounter a missing attribute. Documenting for posterity

Consider two batches like this:
```
Logs Record Batch:
+----+----------+---------+---------------------+-------------------------+---------------+
| id | resource | scope   | time_unix_nano      | observed_time_unix_nano | severity_text |
+----+----------+---------+---------------------+-------------------------+---------------+
| 0  | {id: 0}  | {id: 0} | 1970-01-01T00:00:00 | 1970-01-01T00:00:00     | ERROR         |
| 2  | {id: 0}  | {id: 0} | 1970-01-01T00:00:00 | 1970-01-01T00:00:00     | DEBUG         |
| 1  | {id: 0}  | {id: 0} | 1970-01-01T00:00:00 | 1970-01-01T00:00:00     | WARN          |
+----+----------+---------+---------------------+-------------------------+---------------+

Log Attrs Record Batch:
+-----------+-----+------+------+
| parent_id | key | type | str  |
+-----------+-----+------+------+
| 0         | y   | 1    | test |
| 2         | y   | 1    | test |
| 1         | x   | 1    | test |
+-----------+-----+------+------+
```

If I convert this to OTLP using something like:
```
    pub fn otap_to_logs_data(otap_batch: OtapArrowRecords) -> LogsData {
        let otap_payload: OtapPayload = otap_batch.into();
        let otlp_bytes: OtlpProtoBytes = otap_payload.try_into().unwrap();
        LogsData::decode(otlp_bytes.as_bytes()).unwrap()
    }
```

I end up with OTLP like this. Note how the 3rd `LogRecord` has no attributes, when it should have a single text attribute `{ key: "x", value: "test" }`
```
LogsData {
    resource_logs: [
        ResourceLogs {
            resource: Some(
                Resource {
                    attributes: [],
                    dropped_attributes_count: 0,
                    entity_refs: [],
                },
            ),
            scope_logs: [
                ScopeLogs {
                    scope: Some(
                        InstrumentationScope {
                            name: "",
                            version: "",
                            attributes: [],
                            dropped_attributes_count: 0,
                        },
                    ),
                    log_records: [
                        LogRecord {
[...]

PR #2421, opened 2026-03-24
fix attributes dropped when decoding OTAP -> OTLP proto bytes when IDs are out of order

albertlockett:
# Change Summary

Fixes issue decoding OTAP to OTLP proto bytes when the ID column of the root record batch is not in sorted order.

When we do this decoding, we initialize cursors for the order in which to visit the rows of each record batch. When we visit some row in the root record batch, we then try to find its attributes by advancing the attributes cursor until the parent_id column matches the ID of the current row in the root record batch.

The whole scheme is predicated on the assumption that we'll visit the `id`/`parent_id` column in the same order. However, we initialize the cursor for the attribtues record batch in sorted parent_id order, but for the root record batch we were only initializing it in sorted order of resource/scope ID.

The fix is to add the ID column to the multi-column sort which is used to initialize the cursor for the root record batch.

This PR also makes the change to avoid using RowConverter for the multi-column sort for the root record batch cursor init, instead opting to simply pack the IDs into a Vec of some unsized int (width depending on how many ID columns we're packing), and then sort this. This slightly improves performance (even after adding the additional column to the sort) and I added a benchmark to measure. Bench: main 337µs , after: 305µs 

Note: one side-effect of this change (which in my opinion is OK), is that all the rows with a null ID (e.g. rows w/ no attributes) will appear first in the decoded OTLP message. This is b/c arrow typically uses 0 as a placeholder in int arrays in null rows, and the sort we are doing does not bother looking at the null buffer for best performance.

* Relates to #2270

Unit tests

 No

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otlp/common.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otlp/common.rs
@@ -608,7 +608,7 @@ impl BatchSorter {
 
     /// Initializes the cursor to visit the root [`RecordBatch`] in the OTAP batch in order of
     /// ascending Resource ID & Scope ID. This allows the caller to easily iterate the root
-    /// [`RecordBatch`] in a depth-first trave rsal for each Resource -> Scope -> Log/Span/etc.
+    /// [`RecordBatch`] in a depth-first traversal for each Resource -> Scope -> Log/Span/etc.
     pub fn init_cursor_for_root_batch(
         &mut self,
         record_batch: &RecordBatch,


rust/otap-dataflow/crates/pdata/src/otlp/common.rs
```diff
@@ -632,61 +615,85 @@ impl BatchSorter {
...
-
-                // save the rows vec allocation for next batch needs sorting
-                self.rows = Self::reuse_rows_vec(sort);
-            }
-
-            //there's only one ID column, so we'll visit in the order of this column.
-            [Some(ids), None] => {
-                let ids = ids.as_any().downcast_ref::<UInt16Array>().ok_or_else(|| {
+        if let Some(ids) = record_batch.column_by_name(consts::ID) {
+            sort_columns[sort_columns_idx] =
```
ThomsonTan, line 644: nit: `sort_columns_idx += 1;` in case it is used in future?

albertlockett: That would be a clippy error because the value we assign never gets read. If a new ID column gets added to the list of sort columns in the future, we can update the sort_column_index then. However, I doubt we'll add more columns to this multi-column sort.

rust/otap-dataflow/crates/pdata/src/otlp/common.rs
```diff
@@ -632,61 +615,85 @@ impl BatchSorter {
...
                     Error::ColumnDataTypeMismatch {
                         name: consts::ID.into(),
                         expect: DataType::UInt16,
                         actual: ids.data_type().clone(),
                     }
-                })?;
+                })?);
+        }
+
+        match sort_columns {
```
JakeDern, line 654: Worth mentioning I think that we do something pretty similar here, but for the two column cases: https://github.com/open-telemetry/otel-arrow/blob/0dde9c12802beebd157468f013b01a33cd40b3ac/rust/otap-dataflow/crates/pdata/src/otap/transform/util.rs#L136-L137

I don't think you have to do it now, but I wonder if we can create one "sort by ids" function that always does the best thing based on the inputs

albertlockett:
After some discussion w/ @JakeDern , we've determined this only fixes the issue where there is an invariant such that all the values in the ID column in rows having some scope.id "X" will be less than all the values in the ID column in rows having some scope.id "Y" where X < Y. Or stated more simply, as the scope/resource ID increase, so to do we expect this of the ID column.

We currently maintain this invariant by pure coincidence by virtue of how we encode OTAP data, and by the fact that as we're transforming the data, we have no good reason to start mixing up the values in the ID columns. That said, there's nothing in the OTAP spec that says we _must_ have our ID columns ordered in such a way.

Indeed, it's possible to manually craft by hand a truely pathological  record batch that still doesn't decode properly:
```rust
        // Build an otap batch where:
        //
        //   Row 0: resource_id=0, scope_id=5   (high scope_id in first resource)
        //   Row 1: resource_id=1, scope_id=2   (low scope_id in second resource)
        //   Row 2: resource_id=1, scope_id=3   (another low scope_id in second resource)
        //
        // After sorting by (resource_id, scope_id), visit order is rows [0, 1, 2]
        // with scope_ids visited: [5, 2, 3] - non-monotonic at the 5 -> 2 boundary.
        let str_type = 1u8; // AttributeValueType::Str
        #[rustfmt::skip]
        let mut otap_batch: OtapArrowRecords = logs!(
            (Logs,
                ("id", UInt16, vec![0u16, 1, 2]),
                ("resource.id", UInt16, vec![0u16, 1, 1]),
                ("scope.id", UInt16, vec![5u16, 2, 3]),
                ("scope.name", Utf8, vec!["scope-five", "scope-two", "scope-three"])),
            (ScopeAttrs,
                ("parent_id", UInt16, vec![2u16, 3, 5]),
                ("type", UInt8, vec![str_type, str_type, str_type]),
                ("key", Utf8, vec!["sk", "sk", "sk"]),
                ("str", Utf8, vec!["scope2-attr", "scope3-attr", "scope5-attr"])),
        ).into();
```

I'm tempted to handle this additional case as a followup (I don't believe the solution will conflict), as it currently solves an issue that can be easily encountered in OPL/Transform processor. However I throw myself upon the mercy of my reviewers and if anyone feels this all must be fixed at once I will kindly do so.

JakeDern:
> After some discussion w/ @JakeDern , we've determined this only fixes the issue where there is an invariant such that all the values in the ID column in rows having some scope.id "X" will be less than all the values in the ID column in rows having some scope.id "Y" where X < Y. Or stated more simply, as the scope/resource ID increase, so to do we expect this of the ID column.
> 
> We currently maintain this invariant by pure coincidence by virtue of how we encode OTAP data, and by the fact that as we're transforming the data, we have no good reason to start mixing up the values in the ID columns. That said, there's nothing in the OTAP spec that says we _must_ have our ID columns ordered in such a way.
> 
> Indeed, it's possible to manually craft by hand a truely pathological record batch that still doesn't decode properly:
> 
> ```rust
>         // Build an otap batch where:
>         //
>         //   Row 0: resource_id=0, scope_id=5   (high scope_id in first resource)
>         //   Row 1: resource_id=1, scope_id=2   (low scope_id in second resource)
>         //   Row 2: resource_id=1, scope_id=3   (another low scope_id in second resource)
>         //
>         // After sorting by (resource_id, scope_id), visit order is rows [0, 1, 2]
>         // with scope_ids visited: [5, 2, 3] - non-monotonic at the 5 -> 2 boundary.
>         let str_type = 1u8; // AttributeValueType::Str
>         #[rustfmt::skip]
>         let mut otap_batch: OtapArrowRecords = logs!(
>             (Logs,
>                 ("id", UInt16, vec![0u16, 1, 2]),
>                 ("resource.id", UInt16, vec![0u16, 1, 1]),
>                 ("scope.id", UInt16, vec![5u16, 2, 3]),
>                 ("scope.name", Utf8, vec!["scope-five", "scope-two", "scope-three"])),
>             (ScopeAttrs,
>                 ("parent_id", UInt16, vec![2u16, 3, 5]),
>                 ("type", UInt8, vec![str_type, str_type, str_type]),
>                 ("key", Utf8, vec!["sk", "sk", "sk"]),
>                 ("str", Utf8, vec!["scope2-attr", "scope3-attr", "scope5-attr"])),
>         ).into();
> ```
> 
> I'm tempted to handle this additional case as a followup (I don't believe the solution will conflict), as it currently solves an issue that can be easily encountered in OPL/Transform processor. However I throw myself upon the mercy of my reviewers and if anyone feels this all must be fixed at once I will kindly do so.

Truly pathological record batches are something of an interest of mine 😉 - I think we should file another bug with this test case for follow-up and merge this PR as it's strictly an improvement IMO.

JakeDern (approved):

Issue #1650, opened 2025-12-18
`transform_attributes` can create duplicate keys when renaming


albertlockett:
`otap_df_pdata::otap::transform::transform_attributes` can create duplicate keys when doing a renaming operation if the destination key already exists for some record:
https://github.com/open-telemetry/otel-arrow/blob/ac99143900169732ae2a7c90ed7a61cc4bd37471/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L778

The OTel spec indicates that this is to be avoided, but can be allowed via an option:
https://opentelemetry.io/docs/specs/otel/common/#mapstring-anyvalue

> The implementation MUST by default enforce that the exported maps contain only unique keys. The enforcement of uniqueness may be performed in a variety of ways as it best fits the limitations of the particular implementation (e.g. by removing duplicates).
>
> The implementation MAY have an option to allow exporting maps with duplicate keys (e.g. for better performance). If such option is provided, it MUST be documented that for many receivers, handling of maps with duplicate keys is unpredictable and it is the users' responsibility to ensure keys are not duplicate.

We should fix this, then add probably add test cases for [`AttributesProcessor`](https://github.com/open-telemetry/otel-arrow/blob/ac99143900169732ae2a7c90ed7a61cc4bd37471/rust/otap-dataflow/crates/otap/src/attributes_processor.rs#L114) and [`AttributesTransformPipelineStage`](https://github.com/open-telemetry/otel-arrow/blob/5f5c2ed78405c96b701209ec004b39fdaa6ea5e6/rust/otap-dataflow/crates/query-engine/src/pipeline/attributes.rs#L21) which both use this method internally.

gyanranjanpanda:
is this issue still avilable

ThomsonTan:
@gyanranjanpanda I think so. Contributions are welcome.

albertlockett:
Yes @gyanranjanpanda  you can work on this.

I feel it probably makes sense to delete the rows that would become duplicates via renaming. e.g. if we are renaming attribute `x` to `y`, we'd delete the rows for any attribute key `y` having the same parent_id as some other attribute with key `x`.

e.g.
```
parent_id | key 
----------|-----
  0       | "x" 
  0       | "y"   <-- deleted because attribute key "x" will be renamed to "y"
  1       | "x" 
  1       | "y"   <-- deleted because attribute key "x" will be renamed to "y"
[...]

Issue #2014, opened 2026-02-10
Inserting attribute does not function correctly when there are no existing attributes


albertlockett:
In https://github.com/open-telemetry/otel-arrow/pull/1737 we added the ability to insert attributes using the attribute processor.

However, if the record batch to which we're inserting does not exist on the OTAP batch (e.g., if we're inserting a log attribute, but the batch has no log attributes at all to begin with), then the transformation is not properly applied:

For example, if I add a test like this to the attributes processor:

```rs
    #[test]
    fn test_insert_no_existing_attributes() {
        // Test inserting a boolean value via JSON configuration - there are no existing attributes:
        let input = build_logs_with_attrs(
            vec![], // no resource attrs
            vec![], // no scope attrs
            vec![], // no log attrs
        );

        let cfg = json!({
            "actions": [
                {"action": "insert", "key": "enabled", "value": true},
            ],
            "apply_to": ["signal"]
        });

        let metrics_registry_handle = TelemetryRegistryHandle::new();
        let controller_ctx = ControllerContext::new(metrics_registry_handle);
        let pipeline_ctx =
            controller_ctx.pipeline_context_with("grp".into(), "pipeline".into(), 0, 1, 0);

        let node = test_node("attributes-processor-insert-no-existing");
        let rt: TestRuntime<OtapPdata> = TestRuntime::new();
        let mut node_config = NodeUserConfig::new_processor_config(ATTRIBUTES_PROCESSOR_URN);
        node_config.config = cfg;
        let proc =
            create_attributes_processor(pipeline_ctx, node, Arc::new(node_config), rt.config())
                .expect("create processor");
        let phase = rt.set_processor(proc);

        phase
            .run_test(|mut ctx| async move {
                let mut bytes = BytesMut::new();
                input.encode(&mut bytes).expect("encode");
                let bytes = bytes.freeze();
                let pdata_in =
                    OtapPdata::new_default(OtlpProtoBytes::ExportLogsRequest(bytes).into());
                ctx.process(Message::PData(pdata_in))
                    .await
                    .expect("process");

                let out = ctx.drain_pdata().await;
                let first = out.into_iter().next().expect("one output").payload();

                let otlp_bytes: OtlpProtoBytes = first.try_into().expect("convert to otlp");
                let bytes = match otlp_bytes {
                    OtlpProtoBytes::ExportLogsRequest(b) => b,
                    _ => panic!("unexpected otlp variant"),
                };
[...]

PR #2423, opened 2026-03-24
Fix duplicate attribute keys in transform_attributes

gyanranjanpanda:
# Fix Duplicate Attribute Keys in `transform_attributes`

## Changes Made
This PR resolves issue #1650 by ensuring that dictionary keys are deduplicated when transformations such as `rename` are applied, as required by the OpenTelemetry specification ("Exported maps MUST contain only unique keys by default").

To accomplish this while maintaining strict performance requirements, we replaced the previous `RowConverter` deduplication strategy with a new high-performance, proactive pre-filter:
- We injected `filter_rename_collisions` into `transform_attributes_impl` inside `otap-dataflow/crates/pdata/src/otap/transform.rs`.
- Before a rename is processed, this function reads the `parent_id`s and target keys. It uses the `IdBitmap` type to find any existing target keys whose `parent_id` maps back to an old key that will be renamed.
- It proactively strips those collision rows from the batch via `arrow::compute::filter_record_batch` *before* the actual transform happens.

## Testing
- Extended the `AttributesProcessor` unit tests (`test_rename_removes_duplicate_keys`) to explicitly verify that renaming an attribute resulting in a collision automatically discards duplicate keys.
- Extended the `AttributesTransformPipelineStage` in `query-engine` tests with a parallel case ensuring OPL/KQL query pipelines (`project-rename`) properly drop duplicates when resolving duplicates.
- Refactored `otap_df_pdata` `transform.rs` tests to properly expect deduplicated keys using this plan-based method.
- Validated logic with `cargo test --workspace --all-features`.

## Validation Results
All tests pass. OTel semantic rules surrounding unique mapped keys map cleanly through down/upstream processors. The `IdBitmap` intersection approach completely resolves the multi-thousand percent `RowConverter` performance regressions, dropping collision resolution overhead to essentially zero through efficient bitmap operations.


gyanranjanpanda:
@albertlockett  and @ThomsonTan  waiting for your feed backl

albertlockett (changes requested):
Hey @gyanranjanpanda . I appreciate you taking the time to look at this, but I don't think we can accept this PR as is. 

Unfortunately, the benchmarks we have for this code on main are currently broken. But when I apply the fix from https://github.com/open-telemetry/otel-arrow/pull/2426 and run the benchmark we see that this change introduces significant performance regression:

```
transform_attributes_dict_keys/single_replace_no_deletes/keys=32,rows=128,rows_per_key=4
                        time:   [5.1300 µs 5.1348 µs 5.1394 µs]
                        change: [+1027.4% +1031.5% +1035.2%] (p = 0.00 < 0.05)
                        Performance has regressed.


transform_attributes_dict_keys/single_replace_single_delete/keys=32,rows=128,rows_per_key=4
                        time:   [5.5027 µs 5.5091 µs 5.5155 µs]
                        change: [+495.01% +497.37% +499.48%] (p = 0.00 < 0.05)
                        Performance has regressed.

transform_attributes_dict_keys/no_replace_single_delete/keys=32,rows=128,rows_per_key=4
                        time:   [5.3440 µs 5.3584 µs 5.3746 µs]
                        change: [+577.41% +580.27% +583.40%] (p = 0.00 < 0.05)
                        Performance has regressed.

transform_attributes_dict_keys/single_replace_no_deletes/keys=32,rows=1536,rows_per_key=48
                        time:   [34.015 µs 34.050 µs 34.086 µs]
                        change: [+4000.2% +4016.4% +4031.3%] (p = 0.00 < 0.05)
                        Performance has regressed.

transform_attributes_dict_keys/single_replace_single_delete/keys=32,rows=1536,rows_per_key=48
                        time:   [34.390 µs 34.472 µs 34.562 µs]
                        change: [+1421.9% +1433.5% +1443.9%] (p = 0.00 < 0.05)
                        Performance has regressed.

transform_attributes_dict_keys/no_replace_single_delete/keys=32,rows=1536,rows_per_key=48
                        time:   [34.302 µs 34.340 µs 34.379 µs]
                        change: [+1562.1% +1568.0% +1573.6%] (p = 0.00 < 0.05)
                        Performance has regressed.

transform_attributes_dict_keys/single_replace_no_deletes/keys=32,rows=8192,rows_per_key=256
                        time:   [171.62 µs 171.78 µs 171.96 µs]
                        change: [+6262.2% +6290.6% +6316.2%] (p = 0.00 < 0.05)
                        Performance has regressed.

transform_attributes_dict_keys/single_replace_single_delete/keys=32,rows=8192,rows_per_key=256
                        time:   [171.79 µs 171.92 µs 172.06 µs]
                        change: [+1771.2% +1835.7% +1893.0%] (p = 0.00 < 0.05)
                        Performance has regressed.

transform_attributes_dict_keys/no_replace_single_delete/keys=32,rows=8192,rows_per_key=256
                        time:   [171.20 µs 171.35 µs 171.49 µs]
                        change: [+1962.8% +1981.5% +1998.1%] (p = 0.00 < 0.05)
                        Performance has regressed.

transform_attributes_dict_keys/single_replace_no_deletes/keys=128,rows=128,rows_per_key=1
                        time:   [4.9566 µs 4.9693 µs 4.9819 µs]
                        change: [+587.52% +592.02% +597.47%] (p = 0.00 < 0.05)
                        Performance has regressed.

transform_attributes_dict_keys/single_replace_single_delete/keys=128,rows=128,rows_per_key=1
                        time:   [5.6185 µs 5.6284 µs 5.6377 µs]
                        change: [+292.54% +294.19% +296.01%] (p = 0.00 < 0.05)
                        Performance has regressed.

transform_attributes_dict_keys/no_replace_single_delete/keys=128,rows=128,rows_per_key=1
                        time:   [5.2733 µs 5.2831 µs 5.2938 µs]
                        change: [+385.50% +387.73% +389.92%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

While I expect to see some performance regression because we're doing extra work, I feel that such a serious regression in performance warrants some additional investigation into if/how we can do this in a more efficient way. 

Please see my comment here which prescribes an approach that I believe will be more performant than what is currently in this PR: https://github.com/open-telemetry/otel-arrow/issues/1650#issuecomment-4113152455

gyanranjanpanda:
thanks for your wonderful guidance i will make sure i could match your expectation

albertlockett:
Hey @gyanranjanpanda I wanted to give you a heads up that I am going to be working on #2014 and there may be some significant changes to the transform_attributes code. I will be touching code in transform_keys as well as transform_attributes_impl. I wanted to give you a heads up in case you want to hold off advancing your work until you can better understand the conflicts

gyanranjanpanda:
Thanks for the heads up! I’ll keep an eye on your changes to #2014 and try to align my work accordingly. If possible, could you share which parts might be most affected so I can avoid overlap? or should i wait  after u finished your work i should continue this work

albertlockett:
> Thanks for the heads up! I’ll keep an eye on your changes to #2014 and try to align my work accordingly. If possible, could you share which parts might be most affected so I can avoid overlap? or should i wait after u finished your work i should continue this work

It's probably easiest to hold off until I finish to avoid conflicts, but I'll leave it up to you. I _think_ I should have the changes I need to make for #2014 done by early next week, if not sooner. 

For now, I'll show you the in-progress changes: 
https://github.com/open-telemetry/otel-arrow/compare/main...albertlockett:otel-arrow:albert/2014?expand=1 

I was imagining that for #1650 you'd need to make changes to `plan_key_replacements` or `plan_key_deletes` (which actually haven't been modified) to produce ranges to be deleted in `transform_keys`.

albertlockett:
@gyanranjanpanda the changes I mentioned that could cause conflicts have now been merged (see #2442)

gyanranjanpanda:
i will fix this code as soon as possible while looking your merged pr

gyanranjanpanda:
@albertlockett Thanks for the detailed benchmark feedback! I have completely reworked the approach based on your guidance.

**What changed:**
- Replaced the old `RowConverter` + `filter_record_batch` approach with a **plan-based collision detection** strategy
- Uses `IdBitmap` (as you suggested) to efficiently detect rename collisions in O(N)
- Generates `KeyTransformRange::Delete` entries that are merged into the existing `transform_keys` / `transform_dictionary_keys` pipeline
- No physical batch filtering — collision rows are skipped naturally during materialization
- Only runs collision detection when parent_ids are plain-encoded (skips transport-optimized batches)

**Benchmark results (no regression):**

| Benchmark | Old PR (your review) | Current PR |
|-----------|---------------------|------------|
| `single_replace_no_deletes/keys=32,rows=128` | 5.13 µs (+1031%) | **695 ns** |
| `single_replace_single_delete/keys=32,rows=128` | 5.51 µs (+497%) | **1.36 µs** |
| `no_replace_single_delete/keys=32,rows=128` | 5.34 µs (+580%) | **1.14 µs** |
| `single_replace_no_deletes/keys=32,rows=1536` | 34.0 µs (+4016%) | **1.16 µs** |
| `single_replace_single_delete/keys=32,rows=1536` | 34.5 µs (+1434%) | **3.37 µs** |
| `no_replace_single_delete/keys=32,rows=1536` | 34.3 µs (+1568%) | **3.14 µs** |

The plan-based approach avoids the expensive `RowConverter` sorting and physical batch copy entirely. Would love your re-review when you get a chance!

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1727,11 +1771,21 @@ fn transform_keys(
                 KeyTransformRangeType::Delete => {
                     // for deleted ranges we don't need to append any offsets to the buffer, so we
                     // just decrement by how many total bytes were deleted from this range.
-                    let deleted_val_len = delete_plan
-                        .as_ref()
-                        .expect("delete plan should be initialized")
-                        .target_keys[transform_range.idx]
-                        .len();
+                    let deleted_val_len = if let Some(ref dp) = delete_plan {
+                        if transform_range.idx < dp.target_keys.len() {
```
albertlockett: We're setting `idx` to 0 here:
https://github.com/open-telemetry/otel-arrow/blob/71f2ee68e5571307af3789ca09bcb3e7af90da95/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L1931-L1944

So won't transform_range.idx always be less than `dp.target_keys.len()` if there are any deletes? 

We should probably have a test that does delete and rename at the same time to cover this kind of issue.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1306,14 +1307,46 @@ pub fn transform_attributes_impl(
...
+    let collision_delete_ranges = if has_renames && !is_transport_optimized {
+        let rename = transform
+            .rename
+            .as_ref()
+            .expect("has_renames guard ensures this is Some");
+        let parent_ids_vec = attrs_record_batch
+            .column_by_name(consts::PARENT_ID)
+            .map(read_parent_ids_as_u32)
+            .transpose()?;
+        if let Some(parent_ids) = parent_ids_vec.as_ref() {
```
albertlockett: `parent_id` column should always be present for attributes. If it's not present that would be an error (similar to how you've returned an error below for the attributes key). See the spec has this column as required: https://github.com/open-telemetry/otel-arrow/blob/main/docs/otap-spec.md#542-u16-attributes

Also, we are copying the ID column into a `Vec<u32>` here, only to pass it into `find_rename_collisions_to_delete` ranges simply so the IDs can be iterated. This kind of copying is wasteful

IMO we could probably just make `find_rename_collisions_to_delete` generic over `<K: ArrowPrimitiveType>` and have it accept parent_ids as a `&PrimitiveArray<K>`. If the parent_id column is a PrimitiveArray, we pass that directly. If it's a dictionary, we can actually just pass the keys and the logic we're using to look for which rows would be duplicated after rename should still work.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1874,127 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+
+    for (old_key, new_key) in &rename.map {
+        // Build IdBitmap of parent_ids that have the source key (old_key).
+        let mut source_parents = IdBitmap::new();
+        for (i, &pid) in parent_ids.iter().enumerate() {
+            if let Some(k) = key_accessor.str_at(i) {
+                if k == old_key.as_str() {
+                    source_parents.insert(pid);
+                }
+            }
```
albertlockett: The way we're checking the strings here could be optimized. Since attribute keys will most often be dictionary encoded, what will happen here is that this will effectively pull the string from the dict and compare it for every parent ID. It would actually be more optimal to do the string comparison one time on the dictionary values.

All this to say that we should probably be using the arrow `eq` kernel on the string up front to create a mask of which rows contain this key https://docs.rs/arrow/latest/arrow/compute/kernels/cmp/fn.eq.html and check if the mask is true at this index.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1874,127 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+        }
+
+        if source_parents.is_empty() {
+            continue;
+        }
+
+        // Mark target-key rows whose parent_id overlaps with source_parents.
+        for (i, &pid) in parent_ids.iter().enumerate() {
+            if let Some(k) = key_accessor.str_at(i) {
+                if k == new_key.as_str() && source_parents.contains(pid) {
```
albertlockett: Similar comment to the one above related to using `eq` compute kernel to compare the key value instead of calling `str_at` and == for every row.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1874,127 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+    }
+
+    if !has_collisions {
+        return Vec::new();
+    }
+
+    // Convert per-row booleans into contiguous KeyTransformRange::Delete entries
+    let mut ranges = Vec::new();
+    let mut start = None;
+    for (i, &should_delete) in collision_delete_rows.iter().enumerate() {
```
albertlockett: We're materializing a vector with a flag at each position indicating that the row is deleted, only to throw it away when we iterate it to convert to ranges here. Wouldn't it be more optimal to just compute the vec of ranges directly in the code above?

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
albertlockett: Maybe the solution here would be to make `idx` an Option<usize>

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
albertlockett: For example if you passed keys directly as an `ArrayRef`, you could so something similar to this:
```rs
// compare using optimized kernel
let keys = eq(keys, &StringArray::new_scalar(old_key))?;

// for each range that has equal keys
for (start, end) in BitSliceIterator::new(&keys.values().inner(), 0, keys.len()) {
    // push IDs into source_parents
    for pid in parent_ids[start..end] {
        source_parents.insert(pid)
    }
  }
```
I say "something similar" because I also think we may want to pass something other than a Vec<u32> for parent IDs. (see my other comment)

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1874,127 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+    key_accessor: &StringArrayAccessor<'_>,
+    parent_ids: &[u32],
+    rename: &RenameTransform,
+) -> Vec<KeyTransformRange> {
+    let mut collision_delete_rows = vec![false; num_rows];
+    let mut has_collisions = false;
+
+    for (old_key, new_key) in &rename.map {
+        // Build IdBitmap of parent_ids that have the source key (old_key).
+        let mut source_parents = IdBitmap::new();
```
albertlockett: The heap allocation of ID bitmap can be reused for each rename:
```suggestion
    let mut source_parents = IdBitmap::new();
    for (old_key, new_key) in &rename.map {
        source_parents.clear();
```

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1306,14 +1307,46 @@ pub fn transform_attributes_impl(
     let attrs_record_batch = attrs_record_batch_cow.as_ref();
     let schema = attrs_record_batch.schema();
 
+    // Compute collision delete ranges if renames exist.
+    // When renaming key "a" -> "b", any existing row with key "b" sharing a parent_id
+    // with a row having key "a" would become a duplicate. We proactively identify these
+    // collisions and generate Delete ranges that are merged into the transform pipeline.
+    // Only when parent_ids are plain-encoded (not transport-optimized) — quasi-delta
+    // encoded values don't represent actual parent IDs.
```
albertlockett: This is not true. The delta encoded IDs do represent actual parent_ids, and we can't just ignore this logic if we find delta encoded parent IDs.

Probably the most optimal thing to do is actually check if any rows have an attribute with the new key name. If this happens, we'll need to remove the delta delta encoding so we can figure out if a row with the same parent_id also has the old key name, which would indicate a collision after rename.

I realize this complicates your algorithm, because clearly both those key checks are happening inside the same helper `find_rename_collisions_to_delete_ranges`, so in order to get this to work as explained you may need to refactor a bit.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1617,7 +1653,15 @@ fn transform_keys(
...
-    let transform_ranges = merge_transform_ranges(replacement_plan.as_ref(), delete_plan.as_ref());
+    let mut transform_ranges =
+        merge_transform_ranges(replacement_plan.as_ref(), delete_plan.as_ref()).into_owned();
+
+    // Merge collision delete ranges into the transform plan
+    if !collision_delete_ranges.is_empty() {
+        transform_ranges.extend_from_slice(collision_delete_ranges);
+        transform_ranges.sort_by_key(|r| r.start());
+    }
+    let transform_ranges: Cow<'_, [KeyTransformRange]> = Cow::Owned(transform_ranges);
```
albertlockett: I don't feel that we should convert this to a Vec, then re-sort it, then convert it back to a Cow. The whole point of it being a Cow in the first place was to avoid a heap allocation.

Maybe we could find a way to have the collision_delete_ranges already appended into replacement_plan.ranges, maybe by passing it into plan_key_replacements?

albertlockett (changes requested):
Looks like some good progress, but still some things happening that are not as well optimized as they could be

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1897,159 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+    let mut source_parents = IdBitmap::new();
+
+    for (old_key, new_key) in &rename.map {
+        source_parents.clear();
+
+        let old_key_mask = eq(key_col, &StringArray::new_scalar(old_key)).map_err(|e| {
+            Error::UnexpectedRecordBatchState {
+                reason: format!("eq kernel failed for old_key: {e}"),
+            }
+        })?;
```
albertlockett: There is still quite a performance regression from what is on main. For example:
```
transform_attributes_native_keys/block_replace_no_delete/rows=1536
                        time:   [5.9484 µs 5.9712 µs 5.9930 µs]
                        change: [+241.53% +256.29% +269.79%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

When I profile this, I see we're spending a lot of time in the `eq` compute kernel:
image

I think we need to optimize how we check for the presence of the existing keys.

We actually have a highly optimized kernel for checking if the keys match some given value, which I think is what we should use here:
https://github.com/open-telemetry/otel-arrow/blob/9c54c8e469806b54ec6fa8b9721790c951ca0a97/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L2377-L2459

One caveat about this is that, it only works on the offset/values buffer from the arrow string arrays. That means if the keys column happens to be dictionary encoded, we can only run this on the dictionary values.

albertlockett: Also, because we're checking for the existence of the old keys both in this method, and in `plan_key_replacements`:
https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L2282

It'd be nice if we can avoid checking that twice, but we'd need to dramatically refactor how this function is called in order to do that (which we may want to do).

If that refactoring is not possible, we could consider the fact that it may be somewhat rare that someone may have existing keys that would become duplicates via renaming. Given this fact, imo it might be better logic here to first check for the existing of the new key, and if no rows are found, we exist early (instead of how we're currently doing it, where we look for old keys first).

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1897,159 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+        for (start, end) in BitSliceIterator::new(old_key_mask.values().inner(), 0, num_rows) {
+            for i in start..end {
+                let pid: u64 = parent_ids.value(i).into();
+                source_parents.insert(pid as u32);
+            }
+        }
+
+        if source_parents.is_empty() {
+            continue;
+        }
```
albertlockett: Careful about the unnecessary work here - we don't actually need to load all the IDs into the ID bitmap before returning early. We may be able to use the result of having checked for the existence of the key to determine if we can `continue` early on this iteration of the loop.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1897,159 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+/// This is a cheap pre-check used to decide if we need to materialize delta-encoded parent_ids.
+fn rename_has_target_key_in_column(key_col: &ArrayRef, rename: &RenameTransform) -> Result<bool> {
+    for new_key in rename.map.values() {
+        let scalar = StringArray::new_scalar(new_key);
+        let mask = eq(key_col, &scalar).map_err(|e| Error::UnexpectedRecordBatchState {
+            reason: format!("eq kernel failed on attribute keys: {e}"),
+        })?;
+        if mask.true_count() > 0 {
+            return Ok(true);
+        }
```
albertlockett: Similar to the comment I've made on the code below, the `eq` compute kernel is kind of expensive, and we might be able to use the more optimized 
https://github.com/open-telemetry/otel-arrow/blob/9c54c8e469806b54ec6fa8b9721790c951ca0a97/rust/otap-dataflow/crates/pdata/src/otap/transform.rs#L2377-L2384

That said, I also feel that the first step of `find_rename_collisions_to_delete_ranges` maybe should be be to call this for the new keys (again, see my comments on code below), so given that fact - we might want to be careful about doing duplicate work.

albertlockett (changes requested):
hey @gyanranjanpanda  - sorry it took me some time to review the last round of changes. The code looks like it's in much better shape, thanks for all your work! 

The performance is still not quite where we need it to be. I left some suggestions about how things could maybe be improved - specifically around how we're using the `eq` kernel.

I also noticed that these changes break the existing benchmarks, which would make this perf regression hard for you to measure locally. I've pushed a fix to my branch here: https://github.com/open-telemetry/otel-arrow/commit/32a55e5e971c5c5f46148963d5ba09deff82be50 (which you may actually want to cherry-pick).

FWIW, instructions for profiling have also beend added here https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/PROFILING.md . It's possible to use these same commands while running the benchmarks.

gyanranjanpanda:
@albertlockett  could  u rview this now

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1897,362 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+                        .downcast_ref::<DictionaryArray<UInt16Type>>()
+                        .expect("checked type")
+                        .keys()
+                        .values()
+                        .iter()
+                        .map(|v| *v as usize)
+                        .collect(),
+                    _ => unreachable!("unsupported dict key type"),
+                },
+                _ => unreachable!("checked dictionary type"),
```
albertlockett: We're eagerly collecting the dictionary keys into a `Vec<usize>`, and later on we're actually just iterating over the vec. Doing this collection seems wasteful.

I noticed we still have a performance regression in one of the existing benchmarks:

```
transform_attributes_dict_keys/single_replace_no_deletes/keys=128,rows=8192,rows_per_key=64
                        time:   [1.4321 µs 1.4362 µs 1.4401 µs]
                        change: [+172.60% +174.04% +175.64%] (p = 0.00 < 0.05)
                        Performance has regressed.
```

And indeed, we're spending a lot of time in collect:
image

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1897,362 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+            // Map back through dict_keys to find actual rows.
+            for range in &old_key_result.ranges {
+                for dict_val_idx in range.start()..range.end() {
+                    for (row, dk) in dict_keys.iter().enumerate() {
+                        if *dk == dict_val_idx {
+                            let pid: u64 = parent_ids.value(row).into();
+                            source_parents.insert(pid as u32);
+                        }
+                    }
+                }
```
albertlockett: For each value in the range, we iterate the entire dictionary keys array and check if the index from the range is equal to the key? If this range has a size of greater than one, this is not a very efficient way to do this check.

I actually think if you just took the dictionary keys as an arrow array (e.g. avoid materializing the Vec<usize>, as mentioned above). Then here it would be faster to do something like:
```rs
let row_mask = if range.len() == 1 {
     eq(dict_keys, UInt16Array::new_scalar(range.start() as u16))?
} else {
    let geq_start = gt_eq(dict_keys, UInt16Array::new_scalar(range.start() as u16))?;
    let lt_end = lt(dict_keys, UInt16Array::new_scalar(range.end() as u16))?;
    and(&geq_start, &lt_end)
}

let row_mask_buffer = row_mask.values();
for (start, end) in BitSliceIter::new(row_mask_buffer.inner(), row_mask_buffer.offset(), row_mask.len()) {
    for i in start..end {
         let pid: u64 = parent_ids.value(i).into();
         source_parents.insert(pid as u32);
    }
}
```
see: https://docs.rs/arrow/latest/arrow/compute/kernels/cmp/index.html
see: https://docs.rs/arrow-buffer/latest/arrow_buffer/bit_iterator/struct.BitSliceIterator.html

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
albertlockett: I actually think we can avoid materializing the vec, and just take the array. See my comment on the code below.

albertlockett: Also, we have variants of crate::error::Error that can be used for invalid dictionary types instead of using unreachable! here. I wonder if we should either use those, or comment on why the code is actually unreachable

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -9155,4 +9665,172 @@ mod upsert_tests {
...
+        .unwrap();
+
+        assert_eq!(result, expected);
+    }
+
+    /// Same scenario as [`test_rename_collision_with_real_delete_utf8`] but with
+    /// dictionary-encoded attribute keys. Verifies that collision removal and real
+    /// deletes interact correctly through the dictionary key transform path.
+    #[test]
+    fn test_rename_collision_with_real_delete_dict() {
```
albertlockett, line 9861: could you add an additional test where the dict key is  for the key column is u16 as well?

albertlockett (changes requested):
Thanks for the latest round of changes @gyanranjanpanda !

Still some performance issues with this code that I feel we should address

gyanranjanpanda:
@albertlockett  i did some changes that u mention

gyanranjanpanda:
@albertlockett  if any thing u want to change i am willing to change

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1897,576 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+                let lt_e =
+                    lt(dict_keys, &UInt8Array::new_scalar(range.end() as u8)).map_err(|e| {
+                        Error::UnexpectedRecordBatchState {
+                            reason: format!("lt kernel failed: {e}"),
+                        }
+                    })?;
+                and(&geq, &lt_e).map_err(|e| Error::UnexpectedRecordBatchState {
+                    reason: format!("and kernel failed: {e}"),
+                })?
+            };
```
albertlockett: I wonder if it would be possible to make dict_value_range_to_row_mask generic over `<K: ArrowPrimitiveType>` and change the arg type to `PrimitiveArray<K>`? That get rid of this duplicated code?

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1897,576 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+fn find_rename_collisions_dict_u8<K: ArrowPrimitiveType>(
+    num_rows: usize,
+    key_col: &ArrayRef,
+    parent_ids: &PrimitiveArray<K>,
+    rename: &RenameTransform,
+    dict_values: &StringArray,
+    offsets: OffsetBuffer<i32>,
+    values_buf: Buffer,
+) -> Result<Vec<KeyTransformRange>>
+where
```
albertlockett: Similar to my other comment about making `dict_value_range_to_row_mask`, I wonder if it would be possible to just make this generic over the dict key? That would make it so we don't have all this duplicated code in this method and in `find_rename_collisions_to_delete_ranges` for each dict key?

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1897,576 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+        and(&geq_start, &lt_end).map_err(|e| Error::UnexpectedRecordBatchState {
+            reason: format!("and kernel failed on dict keys: {e}"),
+        })?
+    };
+    Ok(mask)
+}
+
+/// UInt8 variant of the dictionary collision detection. Uses the same algorithm
+/// as the UInt16 path but operates on `UInt8Array` dict keys to avoid a costly
+/// `Vec<usize>` materialization.
```
albertlockett: > Uses the same algorithm as the UInt16 path but operates on `UInt8Array` dict keys to avoid a costly `Vec<usize>` materialization.

This kind of comment isn't necessary - generally it's not useful to have comments stating what the code doesn't do. It makes sense in the context of responding to some particular feedback, but it doesn't make sense in the overall context of file.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1897,576 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+    Ok(ranges)
+}
+
+/// Build a boolean row-mask that is `true` for every row in `dict_keys` whose
+/// value falls inside the half-open range `[range.start(), range.end())`.
+///
+/// Uses the Arrow `eq` kernel for single-element ranges and `gt_eq`/`lt`/`and`
+/// for multi-element ranges, which is much faster than iterating the entire
+/// dictionary keys array for each value index.
+fn dict_value_range_to_row_mask(
```
albertlockett:
```
/// Uses the Arrow `eq` kernel for single-element ranges and `gt_eq`/`lt`/`and`
/// for multi-element ranges, which is much faster than iterating the entire
/// dictionary keys array for each value index.
```
This comment is unnecessary. The docstring for the function doesn't need to explain the implementation, and similar to my other comment about the vec materialization, it's not helpful to say that the way it's implemented is faster than the way it's not implemented.

Again, this kind of comment makes sense when responding to PR feedback, and it's the kind of thing an LLM might generate when responding to a prompt, but in the overall context of this code base it's not a sensible thing to write.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1897,576 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+        )?;
+        if old_key_result.total_matches == 0 {
+            continue;
+        }
+
+        source_parents.clear();
+
+        if let Some(dict_keys) = dict_keys_ref {
+            // Dictionary-encoded (UInt16): use eq/gt_eq/lt kernels on dict keys
+            // to build a row mask, then iterate with BitSliceIterator.
```
albertlockett: > use eq/gt_eq/lt kernels on dict keys

Comment left over from code that was refactored into `dict_value_range_to_row_mask`?

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1897,576 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+        DataType::Dictionary(k, _) => {
+            let dict_values = extract_dict_string_values(key_col)?;
+            let offsets = dict_values.offsets().clone();
+            let values_buf = dict_values.values().clone();
+            // Take the dictionary keys array directly — no Vec<usize> materialization.
+            // We only support UInt8 and UInt16 dict key types; for UInt8 we cast to
+            // UInt16 so we can use a single code path with eq/gt_eq/lt kernels.
+            let dict_keys_arr: &UInt16Array = match k.as_ref() {
+                DataType::UInt8 => {
+                    return find_rename_collisions_dict_u8(
```
albertlockett: > no Vec<usize> materialization.

Again, not helpful generally to comment about what the code is not doing.

> eq/gt_eq/lt kernels.

It doesn't make sense to mention the kernels here that are used inside `find_rename_collisions_dict_u8`. From the perspective of this position in the code, how `find_rename_collisions_dict_u8` is implemented seems irrelevant. This reads like a comment on code that was maybe refactored and the comment didn't get updated.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -1820,6 +1897,576 @@ struct DictionaryKeysTransformResult<K: ArrowDictionaryKeyType> {
...
+    K::Native: Into<u64>,
+{
+    if num_rows == 0 || rename.map.is_empty() {
+        return Ok(Vec::new());
+    }
+
+    // Extract the raw offset/values buffers from the key column.
+    // For dictionary-encoded keys, we run find_matching_key_ranges on dictionary values
+    // then use arrow compute kernels (eq/gt_eq/lt) on the integer dict keys array
+    // to efficiently map dict value indices back to row indices.
```
albertlockett: Again, mentioning the compute kernels in a location where they're not used is not needed.

albertlockett (changes requested):
@gyanranjanpanda the performance is finally in a good place imo. Thanks for your diligence getting us here.

Still some issues related to cleanliness, comments and code duplication.

gyanranjanpanda:
i did  changes that u mention @albertlockett

albertlockett (approved):
Looks good to me @gyanranjanpanda  - thanks for your diligence with this work

PR #2430, opened 2026-03-24
Downgrade recordset convert_scalar_expression warnings on missing identifiers

drewrelmas:
# Change Summary

Same motivation as:
- #2032
- #2127
- #2132 

Queries with `toint()` calls on non-existent identifiers would get spammed with WARN logs.

N/A

Manual test of a variety of queries

Users of `recordset_kql_processor` will see fewer spammy logs on certain transforms.

### DIFF
--- a/rust/experimental/query_engine/engine-recordset/src/scalars/convert_scalar_expressions.rs
+++ b/rust/experimental/query_engine/engine-recordset/src/scalars/convert_scalar_expressions.rs
@@ -3,7 +3,35 @@
 
 use data_engine_expressions::*;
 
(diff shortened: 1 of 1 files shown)


drewrelmas:
Discussed offline with @CodeBlanch, realized the warnings are valid for raw `to*()` calls, but it would be better to wrap optional identifiers in a `coalesce` statement instead.

i.e.
> myattribute = coalesce(toint(myattribute), null)

And silence the warning if in a coalesce similar to the linked PRs. He will have an alternate solution coming shortly.

PR #2434, opened 2026-03-25
Add data sanitization step for transform processor results

### LINKED ISSUE #2313: Old values left in dictionary values arrays after attempted redaction using OPL/transform processor
### Pre-filing checklist

### Component(s)

### Bug Description

If I'm using transform processor and I write an OPL program to try to redact my sensitive value:
```
logs | apply attributes {
    if (key == "maybe_sensitive") {
        where value != "my_credit_card_number"
    }
}
```
(using the new capability added in https://github.com/open-telemetry/otel-arrow/pull/2290).

This will remove the attribute value "my_credit_card_number" semantically from the attribute value, but under certain conditions the actual text may still be in the dictionary in the string column. To avoid this footgun, we should probably have the capability in the query-engine to actually remove the dictionary values that don't have keys pointing to them.

Note - this is potentially expensive, so we probably want a way to opt out of the behaviour if it's not needed. This would be a somewhat advanced configuration for the engine.

### Steps to Reproduce

```rs
    #[tokio::test]
    async fn test_pipeline_where_conditional_example() {
        let input = to_logs_data(vec![
            LogRecord::build()
                .attributes(vec![
                    KeyValue::new(
                        "maybe_sensitive",
                        AnyValue::new_string("my_credit_card_number"),
                    ),
                    KeyValue::new("k1", AnyValue::new_string("albert")),
                ])
                .finish(),
            LogRecord::build()
                .attributes(vec![
                    KeyValue::new("maybe_sensitive", AnyValue::new_string("my_phone_number")),
                    KeyValue::new("k1", AnyValue::new_string("albert")),
                ])
                .finish(),
        ]);
        let query = r#"
            logs | apply attributes {
                if (key == "maybe_sensitive") {
                    where value != "my_credit_card_number"
                }
            }"#;

        let pipeline_expr = OplParser::parse(query).unwrap().pipeline;
        let mut pipeline = Pipeline::new(pipeline_expr);

        let input = otlp_to_otap(&OtlpProtoMessage::Logs(input));
        let result = pipeline.execute(input.clone()).await.unwrap();

        // also just assert there are no attrs remaining
        let attrs = result.get(ArrowPayloadType::LogAttrs).unwrap();
        let str_col = attrs.column_by_name(consts::ATTRIBUTE_STR);
        println!("{str_col:?}")
    }
```

Observe:
```
Some(DictionaryArray {keys: PrimitiveArray<UInt16>
[
  2,
  1,
  1,
] values: StringArray
[
  "my_credit_card_number",
  "albert",
  "my_phone_number",
]}
)
```

### Expected Behavior

Either
```
Some(DictionaryArray {keys: PrimitiveArray<UInt16>
[
  1,
  0,
] values: StringArray
[
  "albert",
  "my_phone_number",
]}
)
```

or possibly:
```
Some(DictionaryArray {keys: PrimitiveArray<UInt16>
[
  2,
  1,
  1,
] values: StringArray
[
  "",
  "albert",
  "my_phone_number",
]}
)
```

(whatever is faster to execute).

### Actual Behavior

See in steps to reproduce

### OTel-Arrow Version

83a824726

### Environment

albert's macbook

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

albertlockett:
# Change Summary

Adds a "sanitization" step which can be performed on the result produced by the transform processor. This passes over all the the columns in all the RecordBatch's and removes any values from dictionary columns that have no keys pointing to them.

The procedure has some performance overhead, so there is an option to skip if if, for example the transformation isn't removing sensitive data, or if something further along in the pipeline would remove the hidden arrow data (for example, serializing to OTLP in the OTLP exporter).

**Why is this necessary?**
Some of the arrow compute kernels will perform transformations ignoring certain buffers for best performance if the result would still be a semantically correct arrow array. For example when filtering, the arrow compute kernels only filter dictionary key columns without touching the dictionaries. 

I'm imagining that someday someone will try to use transform processor to try to redact sensitive data, but the values from the rows they deleted will still be present in the arrow buffers. If they then transmitted the data using OTAP exporter (which does a simple arrow IPC serialization), the "redacted" data has escaped.

**Why blindly do this for all transforms on all columns? Can't the query-engine be smarter about this?**
Maybe - but it's not as simple as it appears.

For example, consider when we're filtering. If we did something like: `logs | where event_name != "sensitive_event_name"`, it might be easy to think that `event_name` is the sensitive column so it's the only one that needs sanitizing. But _maybe_ the user actually knows a-priori that any log w/ this event name actually has sensitive data in some other column.

When it comes to the security of, I feel that it's better to be err on the side of caution. 

In the future we could maybe consider a better system where we let the user provide hints about what fields they consider sensitive.

**What's the performance impact?**

This sanitizing step adds significant overhead. When adding this sanitization step into the query engine's benchmarks, I saw anywhere between 5%-45% performance overhead on top of just executing the transform pipeline w/ no sanitization. The actual overhead depends on the complexity of the transform and size of output. 

This poor performance is another reason why there's an option to skip this step.

**Should sanitize be on by default?**

My feeling is yes. If someone forgets to configure this or misconfigures it, I feel that it's best to fail on the side of worse performance/better security.

**Does this have to happen in the transform processor?**

I feel that this is a reasonable place to do this, but open to suggestions if anyone feels differently.

* Closes #2313

Unit tests

There's a new user facing config field called `skip_sanitize_result`


rust/otap-dataflow/crates/pdata/src/arrays/sanitize.rs
```diff
@@ -0,0 +1,365 @@
...
+            // so it is safe to expect that try_new will not error here
+            let sanitized_batch = RecordBatch::try_new(record_batch.schema(), new_columns)
+                .expect("can create sanitized batch");
+            Some(sanitized_batch)
+        }
+    }
+}
+
+/// Sanitizes a struct array by removing orphaned dictionary values from any dictionary fields.
+fn sanitize_struct(struct_arr: &StructArray) -> Option<StructArray> {
```
lquerel, line 82: Is it possible for the struct itself to be nullified?

albertlockett: Sorry, I'm not sure I follow.. Are you thinking that after filtering or some such operation that we might still have a struct column but the null buffer is entirely null, in which case it could be removed entirely?

lquerel: I was wondering whether it's possible for a structure to be entirely nullified while the field dictionaries within that structure are not. It's probably a case that can't happen, but I wanted to make sure.

albertlockett: ohhh, sorry I get what you mean now. Yes it is possible to have a the validity bitmap indicate that some row in a struct column is null, but the columns contained in the struct array are not null in at this index.

We currently don't have a way to do a transform to set a struct to null, but maybe once we do this will be something we can add to this sanitization code? I think it would be as simple as going through each struct column and `or`ing, each column's null buffer with the struct null buffer, then zeroing out the values buffer (which would be the expensive part).

PR #2435, opened 2026-03-25
[query-engine] Adjust logging for conversion operations in RecordSet engine

CodeBlanch:
Relates to #2430

# Changes

* Pass along `SelectionOptions` to `execute_convert_scalar_expression`
* Tiddy:
  * Reduce code duplication in conversion code
  * Implement `Copy` for `RecordSetEngineDiagnosticLevel`

# Details

We pass along `SelectionOptions` so that `toint(Something)` will emit a warn if "Something" isn't found but `coalese(tonint(Something), null)` will emit an info. This follows the pattern we've established so far when some outer function is used to handle the "not found" case. The assumption being "not found" is expected\anticipated in these cases so a warning is not necessary\overkill.


drewrelmas:
Much better solution than #2430! Thanks!

Issue #2422, opened 2026-03-24
Metrics temporal reaggregation processor


JakeDern:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

A processor for doing temporal metrics reaggregations based on a similar one [implemented in go](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.147.0/processor/intervalprocessor).

### Proposed Solution

urn: `urn:otel:processor:temporal_reaggregation`

#### Overview

The temporal reaggregation processor decreases telemetry volume by reaggregating metrics collected at a higher frequency into a lower one. This temporal reaggregation is one of the three semantics preserving transformations outlined in the [Metrics Data Model specification](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#events--data-stream--timeseries) and is also [implemented in go](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/v0.147.0/processor/intervalprocessor).

Reaggregation is done by collecting metrics for some fixed interval and emitting a single data point per stream when the interval expires. If the number of streams or inbound requests becomes too large, then we may emit metrics at a higher frequency than requested.

#### Signal Handling

Anything other than metrics are passed through as-is.
#### Format Handling

We can support both OTLP and OTAP inputs using pdata views to iterate over the constructs, but we will initially only output OTAP batches with the aggregated results.
#### Config

```yaml
period: 5s -- The metrics aggregation period
inbound_request_limit: 8192 -- The maximum number of inbound requests
outbound_request_limit: 1024 -- The maximum number of outbound requests
-- TBD: possible limits on the number of metric streams we can aggregate at one time
```

In the future we can also support the interval processor's ability to configure passthrough since there is some actual data loss for some metric types like gauges and not just resolution loss by doing this re-aggregation.
#### Supported Metric Types 

Initially we can support the same metrics types as the interval processor. We can exclude anything with delta temporality and non-monotonic sums. 

In the future I don't believe there is anything blocking us from adding coarser delta-to-delta reaggregation for some metric types as well.

Any metrics types which are not supported by the processor will be passed through as-is.
#### Telemetry

[...]

PR #2436, opened 2026-03-25
feat: Metrics temporal reaggregation processor scaffolding

JakeDern:
# Change Summary

This is part 1 of implementing the Metrics Temporal Reaggregation processor just starting with basic scaffolding and documentation.

* Part of #2422
 
Unit

Yes, new processor!

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/config.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/config.rs
@@ -10,7 +10,6 @@ use serde::{Deserialize, Serialize};
 
 const DEFAULT_PERIOD_SECONDS: u64 = 60;
 
-/// Minimum allowed period in milliseconds.
 const MIN_PERIOD_MILLIS: u64 = 100;
 
 fn default_period() -> Duration {


rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
```diff
@@ -0,0 +1,255 @@
...
+        let metrics = pipeline_ctx.register_metrics::<TemporalReaggregationMetrics>();
+        let config: Config =
+            serde_json::from_value(config.clone()).map_err(|e| ConfigError::InvalidUserConfig {
+                error: e.to_string(),
+            })?;
+        Ok(Self {
+            metrics,
+            collection_period: config.period,
+            timer_started: false,
+        })
```
lquerel, line 93: We should probably set a minimum value greater than 0 for the period to avoid triggering the timer every 0 seconds.

lquerel (approved):
Minor suggestion. Otherwise LGTM

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
JakeDern: Done - I think 100ms seems like a generous minimum for a component that is supposed to be reaggregating to a higher interval, but happy to update it again

PR #2443, opened 2026-03-26
Improve cargo xtask developer checks and add diagnostics output

lquerel:
# Change Summary

Improve the OTAP-Dataflow developer check workflow by adding two targeted xtask
commands and a diagnostics mode for the full check.

This PR:
- adds `cargo xtask quick-check` for faster local iteration
- adds `cargo xtask check-benches` for bench-focused validation
- adds `cargo xtask check --diagnostics` to summarize where check time is spent
- documents the new commands and how to interpret diagnostics output

The goal is to make local validation more practical without weakening the
existing final validation path. `cargo xtask check` remains the required full
check, while the new commands help contributors iterate faster and troubleshoot
slow check runs more easily.

* Closes #NNN

- `cargo xtask check`

The new diagnostics parser/reporting logic is also covered by xtask unit tests,
including timing report parsing, test-binary parsing, and timing-report link
generation.

image

No runtime or product-facing changes.

There are contributor-facing workflow changes:
- new `cargo xtask quick-check`
- new `cargo xtask check-benches`
- new `cargo xtask check --diagnostics`
- new documentation in `docs/xtask-diagnostics.md`

### DIFF
--- a/rust/otap-dataflow/Cargo.toml
+++ b/rust/otap-dataflow/Cargo.toml
@@ -66,11 +66,11 @@ data_engine_parser_abstractions = { path = "../experimental/query_engine/parser-
 ahash = "0.8.11"
 arc-swap = "1.7"
 arrayvec = "0.7.6"
-arrow = { version = "57.0", features=["prettyprint"] }
-arrow-buffer = { version = "57.0" }
-arrow-ipc = { version = "57.0", features=["zstd"] }
-arrow-schema = { version = "57.0" }
-arrow-array = { version = "57.0" }
+arrow = { version = "58.1", features=["prettyprint"] }
+arrow-buffer = { version = "58.1" }
+arrow-ipc = { version = "58.1", features=["zstd"] }
+arrow-schema = { version = "58.1" }
+arrow-array = { version = "58.1" }
 async-stream = "0.3.6"
 async-trait = "0.1.88"
 async-unsync = "0.3.0"
@@ -87,7 +87,7 @@ clap = { version = "4.5.42", features = ["derive"] }
 core_affinity = "0.8.3"
 criterion = "0.8.0"
 data-encoding = "2.9.0"
-datafusion = { version = "51.0.0", default-features = false }
+datafusion = { version = "53.0.0", default-features = false }
 fluke-hpack = "0.3.1"
 flume = { version = "0.12.0", default-features = false, features = ["async"] }
 futures = "0.3.32"
@@ -112,7 +112,7 @@ memory-stats = "1"
 nix = { version = "0.31.0", features = ["process", "signal", "fs", "mman"] }
 notify = "8.0"  # Uses platform-native backend: inotify (Linux), kqueue (macOS), ReadDirectoryChanges (Windows)
 num_enum = "0.7"
-object_store = {version = "0.12.3", default-features = false}
+object_store = {version = "0.13.2", default-features = false}
 once_cell = "1.20.2"
 
 # TODO: Note this dependency is duplicate with crates/pdata/src/proto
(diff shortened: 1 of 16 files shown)


rust/otap-dataflow/xtask/src/main.rs
```diff
@@ -54,55 +69,381 @@ pub fn print_help() -> anyhow::Result<()> {
...
+            "--no-run",
+        ],
+    )?;
+    println!("✅ Fast workspace test targets compiled successfully.\n");
+    Ok(())
+}
+
+fn compile_benches() -> anyhow::Result<()> {
+    println!("🚀 Compiling workspace bench targets with cargo check...");
+    run("cargo", &["check", "--workspace", "--benches"])?;
```
lalitb, line 323: Should this be `cargo bench --workspace --no-run` as we need to compile here?

rust/otap-dataflow/xtask/src/main.rs
```diff
@@ -54,55 +69,381 @@ pub fn print_help() -> anyhow::Result<()> {
...
+        .ok_or_else(|| anyhow::anyhow!("failed to capture stdout from `{command}`"))?;
+    let stderr = child
+        .stderr
+        .take()
+        .ok_or_else(|| anyhow::anyhow!("failed to capture stderr from `{command}`"))?;
+
+    let (tx, rx) = mpsc::channel::<diagnostics::CapturedLine>();
+    let stdout_tx = tx.clone();
+    let stdout_handle = thread::spawn(move || stream_lines(stdout, false, stdout_tx));
+    let stderr_handle = thread::spawn(move || stream_lines(stderr, true, tx));
```
lalitb, line 362: `stdout` and `stderr` are read separately and then mixed into one queue here. Is the original output order for the command guaranteed to be preserved?

rust/otap-dataflow/xtask/src/main.rs
lquerel: This is a `cargo check` targeting the benches, so it compiles them but does not run the benchmarks. I don't think the `--no-run` option exists in this context.

rust/otap-dataflow/xtask/src/main.rs
lquerel: It's possible that I didn't understand this comment correctly. 
There isn't really any guaranteed ordering between `stderr` and `stdout` (since one is often buffered and the other isn't). What matters to me, though, is making sure we preserve the local ordering within each stream, which I believe is guaranteed with this approach. Do you see it differently?

rust/otap-dataflow/xtask/src/main.rs
lalitb: I agree `--no-run` does not exist for `cargo check`. My point is that `cargo check --benches` only checks bench targets for errors; it does not do the full build/codegen/link step. If the intent here is to actually compile the bench targets without executing them, `cargo bench --workspace --no-run` seems like the better match.

rust/otap-dataflow/xtask/src/main.rs
lalitb: I may be missing something, but `parse_test_binaries` looks like it can still misattribute timings if `Running ..`. and `test result ...` arrive from different streams and cross-stream ordering is not preserved.

Please ignore if this is more hypothetical scenario :)

rust/otap-dataflow/xtask/src/main.rs
lquerel: My intent was "compile enough to verify correctness". The following steps are performed by cargo check:
- parses the code
- expands macros
- resolves dependencies
- runs type checking and borrow checking
- builds enough metadata for the crate graph

Which I think is good enough for this use case (quick validation) but I'm open to move to a full codegen/link if we have a reason for it.

lalitb: Yes agree, we don't need full codegen here. Thanks for explanation.

Issue #313, opened 2025-03-31
[Rust-CI Follow-up] cargo deny license failures


drewrelmas:
Newly donated Rust projects in this repository have some licensing issues. There is a specific [allowlist maintained by the CNCF foundation](https://github.com/cncf/foundation/blob/main/allowed-third-party-license-policy.md) that suggests a specific list that its projects can use.

Need to confirm if OpenTelemetry has its own list, it appears not per conversation in [https://github.com/open-telemetry/community/issues/2383](https://github.com/open-telemetry/community/issues/2383).

Evaluate flagged licenses in CI cargo_deny step and resolve on a case by case basis.

utpilla:
@jmacd If you're following up with CNCF on this, could you please ask about "MPL-2.0" license as well? We did a [similar exercise](https://github.com/open-telemetry/opentelemetry-rust/pull/2936) for Rust SDK repo and found that some of crates transitively depend on [webpki-roots](https://github.com/rustls/webpki-roots) which uses "MPL-2.0" license.

PR #2444, opened 2026-03-27
feat(temporal_reaggregation_processor): Basic aggregations for OTAP payloads

JakeDern:
# Change Summary

This is Part 2 of the temporal reaggregation processor which adds some basic aggregation ability and a whole lot of plumbing. There are still many things to fix and implement, but this seemed like a good checkpoint.

This PR lets us:

- Compute an identity of a metric stream from a view without copying
- Incrementally build an output record batch while giving random write access to the data point tables so that we can update them with newer data as it comes in.

* Part of #2422 

Unit.

No.

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
@@ -324,6 +324,8 @@ mod tests {
             let batch1 = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0]),
+                    ("resource.id", UInt16, [0]),
+                    ("scope.id", UInt16, [0]),
                     ("metric_type", UInt8, [MetricType::Gauge as u8]),
                     ("name", Utf8, ["cpu"])),
                 (NumberDataPoints,
@@ -336,6 +338,8 @@ mod tests {
             let batch2 = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0]),
+                    ("resource.id", UInt16, [0]),
+                    ("scope.id", UInt16, [0]),
                     ("metric_type", UInt8, [MetricType::Gauge as u8]),
                     ("name", Utf8, ["cpu"])),
                 (NumberDataPoints,
@@ -369,6 +373,8 @@ mod tests {
             let batch1 = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0]),
+                    ("resource.id", UInt16, [0]),
+                    ("scope.id", UInt16, [0]),
                     ("metric_type", UInt8, [MetricType::Sum as u8]),
                     ("name", Utf8, ["requests"]),
                     ("aggregation_temporality", Int32, [2]),
@@ -383,6 +389,8 @@ mod tests {
             let batch2 = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0]),
+                    ("resource.id", UInt16, [0]),
+                    ("scope.id", UInt16, [0]),
                     ("metric_type", UInt8, [MetricType::Sum as u8]),
                     ("name", Utf8, ["requests"]),
                     ("aggregation_temporality", Int32, [2]),
@@ -605,6 +613,8 @@ mod tests {
             let batch1 = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0]),
+                    ("resource.id", UInt16, [0]),
+                    ("scope.id", UInt16, [0]),
                     ("metric_type", UInt8, [MetricType::Gauge as u8]),
                     ("name", Utf8, ["cpu"])),
                 (ResourceAttrs,
@@ -621,6 +631,8 @@ mod tests {
             let batch2 = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0]),
+                    ("resource.id", UInt16, [0]),
+                    ("scope.id", UInt16, [0]),
                     ("metric_type", UInt8, [MetricType::Gauge as u8]),
                     ("name", Utf8, ["cpu"])),
                 (ResourceAttrs,
@@ -659,9 +671,10 @@ mod tests {
             let input = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0, 1]),
+                    ("resource.id", UInt16, [0, 0]),
+                    ("scope.id", UInt16, [0, 1]),
                     ("metric_type", UInt8, [MetricType::Gauge as u8, MetricType::Gauge as u8]),
-                    ("name", Utf8, ["cpu", "cpu"]),
-                    ("scope.id", UInt16, [0, 1])),
+                    ("name", Utf8, ["cpu", "cpu"])),
                 (ScopeAttrs,
                     ("parent_id", UInt16, [0, 1]),
                     ("key", Utf8, ["lib", "lib"]),
@@ -694,9 +707,10 @@ mod tests {
             let batch1 = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0]),
+                    ("resource.id", UInt16, [0]),
+                    ("scope.id", UInt16, [0]),
                     ("metric_type", UInt8, [MetricType::Gauge as u8]),
                     ("name", Utf8, ["cpu"]),
-                    ("scope.id", UInt16, [0]),
                     ("scope.name", Utf8, ["scope-a"])),
                 (ResourceAttrs,
                     ("parent_id", UInt16, [0]),
@@ -712,9 +726,10 @@ mod tests {
             let batch2 = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0]),
+                    ("resource.id", UInt16, [0]),
+                    ("scope.id", UInt16, [0]),
                     ("metric_type", UInt8, [MetricType::Gauge as u8]),
                     ("name", Utf8, ["cpu"]),
-                    ("scope.id", UInt16, [0]),
                     ("scope.name", Utf8, ["scope-b"])),
                 (ResourceAttrs,
                     ("parent_id", UInt16, [0]),
@@ -752,6 +767,8 @@ mod tests {
             let input = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0, 1]),
+                    ("resource.id", UInt16, [0, 0]),
+                    ("scope.id", UInt16, [0, 0]),
                     ("metric_type", UInt8, [MetricType::Gauge as u8, MetricType::Gauge as u8]),
                     ("name", Utf8, ["cpu", "memory"])),
                 (NumberDataPoints,
@@ -781,6 +798,8 @@ mod tests {
             let input = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0, 1]),
+                    ("resource.id", UInt16, [0, 0]),
+                    ("scope.id", UInt16, [0, 0]),
                     ("metric_type", UInt8, [MetricType::Gauge as u8, MetricType::Sum as u8]),
                     ("name", Utf8, ["cpu", "cpu"]),
                     ("aggregation_temporality", Int32, [0, 2]),
@@ -812,6 +831,8 @@ mod tests {
             let input = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0]),
+                    ("resource.id", UInt16, [0]),
+                    ("scope.id", UInt16, [0]),
                     ("metric_type", UInt8, [MetricType::Gauge as u8]),
                     ("name", Utf8, ["cpu"])),
                 (NumberDataPoints,
@@ -846,6 +867,8 @@ mod tests {
             let input = OtapArrowRecords::Metrics(metrics!(
                 (UnivariateMetrics,
                     ("id", UInt16, [0, 1]),
+                    ("resource.id", UInt16, [0, 0]),
+                    ("scope.id", UInt16, [0, 0]),
                     ("metric_type", UInt8, [MetricType::Gauge as u8, MetricType::Sum as u8]),
                     ("name", Utf8, ["temperature", "requests"]),
                     ("aggregation_temporality", Int32, [0, 2]),


rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/aggregate.rs
```diff
@@ -0,0 +1,658 @@
...
+        dp: &V,
+        stream_id: StreamId<'_>,
+        otap_metric_id: u16,
+    ) {
+        let time = dp.time_unix_nano();
+        if let Some(state) = self.stream_map.get_mut(&StreamIdRef(&stream_id)) {
+            if time > state.time_unix_nano {
+                self.number_dps.replace(state.dp_row_index, dp);
+                state.time_unix_nano = time;
+            }
```
albertlockett, line 430: Am I reading this right that we're always just taking the latest datapoint?

albertlockett: I guess we can do this because we're checking the metric type/aggregation temporality in `is_aggregatable` eh?

JakeDern: Yeah I'm starting with the metric types that the go interval processor supports which can be aggregated by taking the latest - Basically monotonic cumulative counters and cumulative histograms, exp histograms, gauges, and summaries.

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/data_point_builder.rs
```diff
@@ -0,0 +1,836 @@
...
+                Arc::new(UInt16Array::from(std::mem::take(&mut self.parent_id))),
+                Arc::new(TimestampNanosecondArray::from(std::mem::take(
+                    &mut self.start_time_unix_nano,
+                ))),
+                Arc::new(TimestampNanosecondArray::from(std::mem::take(
+                    &mut self.time_unix_nano,
+                ))),
+                Arc::new(self.int_value.finish()),
+                Arc::new(self.double_value.finish()),
+                Arc::new(UInt32Array::from(std::mem::take(&mut self.flags))),
```
albertlockett, line 200: I thought the flag column is optional. Maybe in future we could keep use a builder that keeps track of wether all the flags are null/zeros, and if so do not include this array? (saving 32 bits per datapoint)

JakeDern: Yeah, certainly plenty to optimize here. I don't use dictionary encodings either because I didn't want to reimplement all the logic from the adaptive array builders.

I was thinking about trying to extend the adaptive array builders to support the random writes in a future PR.

rust/otap-dataflow/Cargo.toml
```diff
@@ -196,6 +197,7 @@ weaver_forge = { git = "https://github.com/open-telemetry/weaver.git", tag = "v0
 weaver_resolved_schema = { git = "https://github.com/open-telemetry/weaver.git", tag = "v0.21.2"}
 weaver_resolver = { git = "https://github.com/open-telemetry/weaver.git", tag = "v0.21.2"}
 weaver_semconv = { git = "https://github.com/open-telemetry/weaver.git", tag = "v0.21.2"}
+xxhash-rust = { version = "0.8", features = ["xxh3"] }
```
JakeDern, line 200: This has a BSL 1.0 license. It seems to be pretty permissive, and I think it would be safe to allow: https://github.com/doumanash/xxhash-rust?tab=BSL-1.0-1-ov-file

If not we can definitely try to find a different library for a 128-bit hash or we can compute two independent 64-bit hashes and combine them which is what the go implementation does.

JakeDern: To clarify - BSL 1.0 in this case is "Boost 1.0" not "Business Source License"

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
```diff
@@ -58,21 +73,64 @@ pub fn create_temporal_reaggregation_processor(
...
+                self.ensure_timer_started(effect_handler).await?;
+
+                match pdata.signal_type() {
+                    // TODO: Support for both view types
+                    SignalType::Metrics => {
+                        if let OtapPayload::OtapArrowRecords(ref records) = *pdata.payload_ref() {
+                            if let Ok(view) = OtapMetricsView::try_from(records) {
+                                self.accumulator.ingest(view);
+                            }
+                        }
```
lquerel, line 104: I assume a follow-up PR will make unsupported metrics or non-OTAP batches pass through this processor unchanged. As I understand it right now, non-OTAP payload are ignored and lost forever.

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/config.rs
lquerel, line 24: At some point, we need to introduce another config parameter to limit the size of the aggregator state.

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/aggregate.rs
```diff
@@ -0,0 +1,658 @@
...
+        &mut self,
+        resource_id: ResourceId,
+        view: &R,
+        schema_url: &[u8],
+    ) -> u16 {
+        if let Some(&id) = self.resource_ids.get(&resource_id) {
+            return id;
+        }
+        let id = self.next_resource_id;
+        debug_assert!(id < u16::MAX, "resource ID overflow");
```
lquerel, line 313: I assume this is a temporary piece of code.

In the long term, we should force a flush of the ongoing aggregations when such an overflow occurs, so that we can reset this ID back to zero.

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/identity.rs
```diff
@@ -0,0 +1,789 @@
...
+            return Self::EMPTY;
+        }
+
+        buf.entries.sort_by(|a, b| a.key().cmp(b.key()));
+
+        for attr in &buf.entries {
+            buf.buf.push(HashTag::Key as u8);
+            buf.buf.extend_from_slice(attr.key());
+            write_attr_value(&mut buf.buf, attr);
+        }
```
lquerel, line 329: Here we have a collision issue. I'm not talking about a probabilistic collision issue but it's a deterministic alias caused by an ambiguous encoding. With this algo:

`{ a = bytes([F4 62 F5]) }` and `{ a = bytes([]), b = empty }`

will have the same encoding byte for byte. Adding the len as a prefix for each encoded value will probably fix this issue.

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/identity.rs
```diff
@@ -0,0 +1,789 @@
...
+                buf.extend_from_slice(b);
+            }
+        }
+        ValueType::Empty => {
+            buf.push(HashTag::Empty as u8);
+        }
+        // FIXME: Views don't support these types yet either.
+        ValueType::Array | ValueType::KeyValueList => {
+            buf.push(HashTag::Empty as u8);
+        }
```
lquerel, line 378: Until this is fixed we will have potential collisions.

lquerel (commented):
I think we should revisit how the hashed value is serialized to avoid collisions.

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
JakeDern: Correct, next PR will focus on extending to OTLP including supporting views

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/config.rs
JakeDern: Yep, it's on my TODO list. One of the next couple PRs will address this

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/aggregate.rs
JakeDern: Agreed, it's on my todo list to sort out all of these rough edges prior to ack/nack support

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/identity.rs
JakeDern: Yep, it's on the todo list to get these added to the views as well

JakeDern:
@lquerel Here is a list of things not included in this PR I currently have marked for follow-up in a very approximate order:

- [ ] Support both payload types
- [ ] Support passing through unsupported metrics immediately (we're currently dropping them)
- [ ] Re-use the view objects between batches 
- [ ] ID overflow handling
- [ ] Add support for Array and Map attribute types (views and hashing)
- [ ] Add exemplar support
- [ ] Add benchmarks
- [ ] Add settings for bounding aggregator memory
- [ ] Add ack/nack support 
- [ ] Support for additional types of metrics if required
- [ ] Support for passthrough settings to exclude certain types from aggregations (gauges, others)

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/identity.rs
JakeDern: Yikes, yeah let me think about how to avoid this. We should probably also patch in go because I adapted it from there and it's the same scheme: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/pdatautil/hash.go

JakeDern: They actually write a suffix as well, which I originally missed, but I don't think it matters. Slightly tweaked collision for the FF suffix:

```text
{ a = bytes([FF F4 62 F6 BB]) } =>        F4 61 F6 FF F4 62 F6 BB FF
{ a = bytes([]),  b = bytes([BB]) } =>    F4 61 F6 FF F4 62 F6 BB FF
```

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/1ff44f3cc3104d0cc3660945a90f24dcf52efbdc/pkg/pdatautil/hash.go#L173-L178

(updated with an easier example)
(updated again, because I forgot a byte, but it doesn't matter)

JakeDern: I _think_ the length prefix for the byte arrays will work, but I'm also wondering if we should solve this for keys or not. I think if keys are guaranteed to be UTF-8 and we shift the tags into the F5-FF range then we'd be ok there. Otherwise it's trivial to create the same collision:

```
{ "a\xF6" = bytes([]) } =>   F4 61 F6 F6 FF
{ "a" = bytes([F6]) } =>     F4 61 F6 F6 FF
```

But there's a few layers to consider on the "guaranteed to be UTF-8" part...

1. Our views currently define Strings as `pub type Str<'src> = &'src [u8];`. It's reasonable to argue that we therefore can't assume anything about encoding because anyone could implement the trait as narrowly as it's defined.

2. Ignoring that for a second let's look at the existing implementation and ask, "can non-utf8 characters make it to this point in the program?"

For arrow we have some guarantees as the source data was parsed already as a string:

```rust
#[inline(always)]
pub(crate) fn get_attribute_key<'a>(
    attr_key: &'a StringArrayAccessor<'a>,
    row_idx: usize,
) -> Option<&'a [u8]> {
    attr_key.str_at(row_idx).map(|s| s.as_bytes())
}
```

For OTLP, I'm not sure we have such a guarantee because we're just slicing some byte buffer. You could argue that protobuf defines this field as a string and therefore these bytes must be utf-8, but someone could create a buggy or malicious implementation that does not do this and therefore we need to check before assuming:

```rust
    fn key(&self) -> otap_df_pdata_views::views::common::Str<'_> {
        loop {
            if let Some((start, end)) = from_option_nonzero_range_to_primitive(self.key_range.get())
            {
                return &self.buf[start..end];
            } else if self.pos.get() >= self.buf.len() {
                break;
            } else {
                self.advance();
            }
        }

        // return empty string when cannot read key
        &[]
    }
```

I think my opinion here is that someone could manufacture collisions which may or may not hurt depending on scenario so we should either:

1. Redefine the view type as `pub type Str<'src> = &'src str;` OR
2. Put a length field in front of the keys as well

_Note: The OpenTelemetry specification only says that attribute names must be valid Unicode but explicitly calls out there's no canonical encoding. This is not really relevant here, but just pointing it out: https://opentelemetry.io/docs/specs/semconv/general/naming/#general-naming-considerations_

JakeDern: Pushing a 4-byte length for both string/byte array keys and values for now

rust/otap-dataflow/Cargo.toml
JakeDern: @jmacd Any restrictions you know of on Boost 1.0 license? Our internal guidance allows it explicitly; I can add it to the list if there are no other restrictions.

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/identity.rs
JakeDern: Go fix here: https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/47242

jmacd: https://github.com/lightstep/otel-launcher-go/blob/main/lightstep/sdk/metric/internal/fprint/fprint.go for another example

rust/otap-dataflow/Cargo.toml
jmacd: Yes, I think this is OK, however please also note:

- The issue number #313 is gaining weight. I think we should go ahead and if someone thinks we should change this, address later. think BSL-1.0 is OK, however we have to make a couple of updates.
- The current `NOTICE` file is a little unclear, we might call this `THIRD_PARTY_NOTICES.txt`. We should add the BSL-1.0 requirement for xxhash-rs into this file
- The `df_engine` I think has to comply with this as well, we need to add a `--license` mode that prints the the inlined notice.

JakeDern: Took care of the first two points in this PR, maybe we can follow up with another PR for adding the `--license` mode for df_engine if that's ok

JakeDern:
Test failures look unrelated to the changes here - I see several PRs stuck on the pipeline perf test and the ubuntu 24.04 hanging after 55+ minutes

PR #2450, opened 2026-03-27
feat: Standardize otap nullable and optional columns

### LINKED ISSUE #2149: Proposal: Standardize the relationship between nullable and optional columns
# Context

Apache arrow has the concept of "nullability" - Columns where some values can be marked as null. This is a canonical property of any field in an arrow schema.

 OTAP has a related concept of "optionality" i.e. columns which are optional and can be omitted entirely. This is an OTAP concept and is indicated in the Go implementation with metadata e.g. the Logs table's id column:

 https://github.com/open-telemetry/otel-arrow/blob/9384bddc78fbd69e42092f5addf09142f90dab5e/go/pkg/otel/logs/arrow/logs.go#L28C1-L29C1

We also have some columns in the Go implementation which are marked as optional but not nullable:

https://github.com/open-telemetry/otel-arrow/blob/9384bddc78fbd69e42092f5addf09142f90dab5e/go/pkg/otel/metrics/arrow/histogram_dp.go#L47C1-L50C1

This is problematic because there is no way to combine two record batches where one provides one of these columns and the other does not unless they are nullable.

# Proposal

Whether or not a column is nullable in arrow terms determines whether it is optional in OTAP terms. Nullable columns are always optional because you could set the entire column to null anyway and therefore could just omit it. We should eliminate the optional metadata property from the Go code.

JakeDern:
# Change Summary

This is a spec compliance PR overhauling the relationship between nullable and optional columns as described in #2149 across both go and rust implementations.

Major changes:

- [spec] Removed the "nullable" column from payload specification tables since "required" is sufficient
- [spec] Changed logs "time_unix_nano" field to optional because I believe this is consistent with the [otlp spec](https://github.com/open-telemetry/opentelemetry-proto/blob/ca839c51f706f5d53bfb46f06c3e90c3af3a52c6/opentelemetry/proto/logs/v1/logs.proto#L139-L142)
- [spec] clarified the relationship between nullable and optional columns in a few spots
- [spec] added "flags" to the spans table which was missing 
- [spec] added "zero_threshold" to the exp histogram dp table which was missing
- [spec] made span links "trace_id" and "span_id" columns required
- [rust] Updated the "required" property of some fields in payloads.rs to match the spec
- [go] Updated the optional and required metadata across many fields to match the spec

* Closes #2149

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/README.md
@@ -356,18 +356,17 @@ To optimize throughput and reduce per-message overhead, the receiver batches
 messages into Apache Arrow record batches before sending downstream:
 
 ```text
-+-------------------------------------------------------------+
-|                    Batching Logic                           |
-+-------------------------------------------------------------+
-|                                                             |
-|   Messages arrive  -->  ArrowRecordsBuilder  -->  Batch     |
-|                              |                     sent     |
-|                              |                              |
-|   Flush conditions:          |                              |
-|   +- Size: max_size messages +-------------------------->   |
-|   +- Time: max_batch_duration_ms  |                              |
-|                                                             |
-+-------------------------------------------------------------+
++-------------------------------------------------------------------+
+|                         Batching Logic                            |
++-------------------------------------------------------------------+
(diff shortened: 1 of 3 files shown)


JakeDern:
I think the test failures here are also unrelated, but it still makes me a bit uncomfortable as this is a riskier change.

PR #2452, opened 2026-03-28
Bound Syslog/CEF message size to prevent unbounded memory growth

utpilla:
# Change Summary

Introduces a 16 KiB per-message size limit (`MAX_MESSAGE_SIZE`) for the syslog CEF receiver to prevent unbounded memory growth from malicious or misbehaving clients.

### Before vs After

| Aspect | Before | After |
|---|---|---|
| TCP message size | Unbounded (memory exhaustion risk) | Capped at 16 KiB per message |
| UDP receive buffer | 1 KiB stack array (silent truncation) | 16 KiB heap `Vec` |
| Truncation visibility | None | `received_logs_truncated` metric |
| TCP line buffer allocation | `Vec::new()` (multiple early reallocs) | `Vec::with_capacity(4096)` |
| Oversized message handling | N/A | Parsed and forwarded as partial data |

### Problem

- **TCP**: `read_until(b'\n')` would keep appending bytes to a `Vec` until a newline appeared. A client sending data without newlines could exhaust memory.
- **UDP**: The receive buffer was a 1 KiB stack array, too small for most real-world syslog messages and silently truncating them with no visibility.

### Changes

- Add `read_line_bounded()` helper that wraps `AsyncReadExt::take()` + `AsyncBufReadExt::read_until()` to cap TCP reads at `MAX_MESSAGE_SIZE` (16 KiB) per message. Returns a `BoundedReadResult` enum (`Complete`, `Truncated`, or `Eof`) so callers can distinguish outcomes.
- Upgrade UDP receive buffer from 1 KiB stack array to 16 KiB heap `Vec`, matching the TCP limit.
- Add `received_logs_truncated` metric to track messages that exceeded the size limit. For TCP, truncation is detected precisely. For UDP, it uses a best-effort heuristic (datagram fills buffer).
- Pre-allocate TCP line buffer with `Vec::with_capacity(4096)` to avoid early reallocations for typical messages.
- Add doc comments explaining cancellation safety: `buf` may contain partial data from `select!` cancellation and must not be cleared between calls unless discarding the current message.

### Design decisions

- **Truncated messages are still forwarded**: Preserves partial data (syslog header, timestamp, severity) rather than dropping it. The tail of an oversized message is also processed as its own message.
- **Metric semantics**: `received_logs_total` counts each bounded read as a separate message. An oversized message that splits into head + tail counts as 2.

Towards #1149 

- **TCP truncation integration test**: Sends an oversized message followed by a normal message, verifies all 3 records (truncated head, tail, normal) are received.
- **6 unit tests for `read_line_bounded`**: Empty reader, complete line, truncation, exact-limit edge case, EOF with partial data, and multi-call behavior after truncation.
- **Telemetry test updates**: Metric index adjustments to account for the new `received_logs_truncated` field.

No

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
@@ -60,6 +60,12 @@ const DEFAULT_MAX_BATCH_SIZE: u16 = 100;
 /// 2048 bytes).
 const MAX_MESSAGE_SIZE: usize = 16 * 1024;
 
+/// Initial capacity for the per-connection TCP message buffer.
+///
+/// Most syslog messages are well under 4 KiB, so this avoids early
+/// reallocations. The buffer can grow up to [`MAX_MESSAGE_SIZE`] if needed.
+const INITIAL_MSG_BUFFER_CAPACITY: usize = 4096;
+
 /// Maximum time to wait for spawned TCP tasks to drain during shutdown.
 const MAX_TASK_DRAIN_WAIT: Duration = Duration::from_secs(1);
 
@@ -462,7 +468,7 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
                                         // Suppress unused variable warning when TLS is disabled
                                         let _ = peer_addr;
 
-                                        let mut line_bytes = Vec::with_capacity(4096);
+                                        let mut line_bytes = Vec::with_capacity(INITIAL_MSG_BUFFER_CAPACITY);
 
                                         let mut arrow_records_builder = ArrowRecordsBuilder::new();
 


rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
```diff
@@ -399,7 +456,7 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
                                         // Suppress unused variable warning when TLS is disabled
                                         let _ = peer_addr;
 
-                                        let mut line_bytes = Vec::new();
+                                        let mut line_bytes = Vec::with_capacity(4096);
```
cijothomas: lets move to a const?

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
```diff
@@ -642,7 +695,7 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
                 );
 
                 let socket = effect_handler.udp_socket(udp_config.listening_addr)?;
-                let mut buf = [0u8; 1024]; // ToDo: Find out the maximum allowed size for syslog messages
+                let mut buf = vec![0u8; MAX_MESSAGE_SIZE];
```
cijothomas, line 710: wondering why would we allocate for worst possible scenario? Could we be more conservative, and accept that vec! will re-grow if we actually has such a big message? Maybe 1/2 or 1/4 of the MAX size for initial size?

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
```diff
@@ -164,6 +172,55 @@ impl Config {
...
+/// a read mid-stream, `buf` may contain partial data from a cancelled read that
+/// the next call must continue from. Do **not** clear `buf` between calls unless
+/// you are discarding the current message.
+async fn read_line_bounded<R: AsyncBufRead + Unpin>(
+    reader: &mut R,
+    buf: &mut Vec<u8>,
+    max_size: usize,
+) -> std::io::Result<BoundedReadResult> {
+    let n = (&mut *reader)
+        .take(max_size as u64)
```
lalitb: I think this doesn’t fully enforce the size cap.

Because the read is inside `tokio::select!`, it can be interrupted if the batch timer fires before a newline arrives. When that happens, any bytes already appended to `line_bytes` stay in the buffer, and the next loop iteration starts reading the same message again.

I think `read_line_bounded()` should consider how much is already in `buf` and only allow reading the remaining space. If `buf` is already at the limit, it should return `Truncated` immediately.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
```diff
@@ -493,16 +548,14 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
...
                                                         }
-                                                        Ok(_) => {
-                                                            let is_complete_line = line_bytes.last() == Some(&b'\n');
-                                                            if !is_complete_line {
-                                                                // ToDo: Handle incomplete lines
-                                                                // Do we process the incomplete line with partial data or discard it?
-                                                                // Handle incomplete line (log, emit metrics, etc.)
+                                                        Ok(bounded_result) => {
+                                                            if matches!(bounded_result, BoundedReadResult::Truncated) {
+                                                                metrics.borrow_mut().received_logs_truncated.inc();
```
lalitb, line 565: I see from the PR description that splitting an oversized TCP message into `head` + `tail` is intentional, but I’m not sure this is the right semantic for a single syslog message.

In this branch we parse and emit the truncated prefix, clear `line_bytes`, and then continue reading, which means the remaining bytes from the same original message get emitted as a second record. That makes one input message turn into multiple downstream records, and `received_logs_total` no longer maps cleanly to original messages.

Why do we want to preserve and emit the `tail` as a separate logical message instead of either keeping a single truncated record or discarding until the next newline after truncation?

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
utpilla: The UDP receive buffer is fundamentally different from TCP's `Vec`. `recv_from` writes into a fixed `&mut [u8]` slice. It doesn't append or grow. If the buffer is smaller than the incoming datagram, the OS kernel silently discards the excess bytes with no way to recover them. So, we must allocate the full `MAX_MESSAGE_SIZE` upfront to avoid permanent data loss. This is a one-time 16 KiB allocation for the lifetime of the receiver (one per core).

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
utpilla: Great catch @lalitb! Thank you! Fixed in https://github.com/open-telemetry/otel-arrow/pull/2452/commits/35c5f1dbe37ba678234f8e4e6ec82d7a7379acf2.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
utpilla: Fixed in https://github.com/open-telemetry/otel-arrow/pull/2452/changes/2da408347c9f6d4a6121323b2529ba710873bdde.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
utpilla: > Why do we want to preserve and emit the tail as a separate logical message instead of either keeping a single truncated record or discarding until the next newline after truncation?

The truncated head preserves the syslog header (timestamp, hostname, severity, facility) which is the most valuable structured metadata. The tail preserves part of the message body. Discarding either means losing data that could be useful for debugging.

The receiver's abstraction is "newline-delimited messages up to 16 KiB each." From its perspective, the head and tail are two separate protocol-level messages. The `received_logs_truncated metric` (incremented only once per oversized message, not per fragment) alerts operators that truncation is occurring, so they can investigate if needed.

Messages exceeding 16 KiB are extremely rare in practice. When one does occur, it's likely an unusual event worth investigating which makes preserving partial data better than silently dropping it.

lalitb: The motivation to preserve as much data as possible makes sense, and this direction is okay for now. The downstream semantics are still a bit surprising though - the `tail` gets emitted as an independent log record with `severity = UNSPECIFIED` and no original `timestamp`, so exporters like Azure Monitor or OTLP consumers have no way to tell it's a fragment of the preceding truncated message.                                                                                                                                                                                           
                                                                                                                                                                                                     
Could we add a `TODO` here and open a follow-up GitHub issue to revisit this? Specifically worth tracking: whether to add fragment correlation metadata (e.g. a shared ID on head and tail), or whether discarding after the newline boundary is actually safer for downstream consumers.

jmacd: I think I agree with @lalitb that preserving the tail and emitting it without context is not a great experience for users. Somewhere I have a wishlist item, that we could invent an inter-log-record reference scheme to allow you to link together multi-part log statements among other things.

It makes me think about a third option, maybe (not blocking), which would preserve the Syslog header, then insert a "(continued from #reference)", in the body with some kind of "overflow" attribute, so that users can find and diagnose the cause of large messages.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs
```diff
@@ -732,8 +797,14 @@ impl local::Receiver<OtapPdata> for SyslogCefReceiver {
                         result = socket.recv_from(&mut buf) => {
                             match result {
                                 Ok((n, _peer_addr)) => {
-                                    // ToDo: Validate the received data before processing
-                                    // ToDo: Consider logging or using peer_addr for security/auditing
+                                    // If the datagram filled the entire buffer, it was likely
+                                    // truncated by the OS (the actual message may have been larger).
+                                    // A message exactly MAX_MESSAGE_SIZE bytes would also trigger
+                                    // this, but there is no way to distinguish the two cases with
+                                    // UDP — this heuristic is the best we can do.
```
jmacd, line 804: This looks good to me. Taking @cijothomas's comment we could imagine growing the underlying buffer _after_ receiving this message, but it would be confusing to explain this to users.

PR #2453, opened 2026-03-29
Move dependency imports to workspace 

### LINKED ISSUE #525: [Rust CI] Avoid multiple imports of the same dependency
We should detect when a crate is imported multiple times in different crates within the same workspace. In such cases, the best practice is to move the dependency declaration to the main Cargo.toml so that there is only one version. 

There might already be some cargo commands that handle this, otherwise, another option would be to extend what we have in our xtask check/structure.

Upon cursory search, [cargo-autoinherit](https://github.com/mainmatter/cargo-autoinherit) might help.

### LINKED ISSUE #1218: All dependencies and dev-dependencies should be workspace-level
I believe that when we mix dependencies between workspace-level and crate-level, there are slow build times because of how `cargo` processes variations. As we switch batch and forth between workspace-level and crate-level cargo commands, cargo has to re-build things when I wish it wouldn't.

I believe it is best if we move all dependency definitions to the workspace level. All crate-level dependencies and dev-dependencies should use { workspace = true } for the best hygiene.

sumanthreddy542:
# Change Summary

Centralize duplicated dependency declarations

Dependencies that were specified with explicit paths or versions in multiple crates have been moved to the root Cargo.toml [workspace.dependencies] section. Each crate now inherits them via [workspace = true]. This was done using cargo autoinherit and prevents version drift across the workspace.

* Closes #525 
* Closes #1218 

cargo build and cargo test

 No

### DIFF
--- a/rust/otap-dataflow/Cargo.toml
+++ b/rust/otap-dataflow/Cargo.toml
@@ -203,7 +203,7 @@ cpu-time = "1.0.0"
 azure_core = {version = "0.33.0", default-features = false }
 azure_identity = {version = "0.33.0", default-features = false }
 flate2 = "1.1.5"
-reqwest = { version = "0.13.1", default-features = false }
+reqwest = { version = "0.13.1", features = ["rustls-no-provider", "json", "query", "http2", "stream"] }
 time = "0.3.44"
 wiremock = "0.6.5"
 anyhow = "1.0.98"
--- a/rust/otap-dataflow/crates/validation/Cargo.toml
+++ b/rust/otap-dataflow/crates/validation/Cargo.toml
@@ -25,7 +25,7 @@ humantime-serde = { workspace = true }
 
 testcontainers = { workspace = true }
 minijinja = { workspace = true }
-reqwest = { version = "0.12", features = ["json"] }
+reqwest = { workspace = true, features = ["json", "query"] }
 
 otap-df-engine = { workspace = true }
 otap-df-channel = { workspace = true }


rust/otap-dataflow/Cargo.toml
```diff
@@ -206,6 +206,27 @@ flate2 = "1.1.5"
...
+mime_guess = "2.0.5"
+minijinja = "1"
+otap-df-admin = { path = 'crates\admin' }
+otap-df-channel = { path = 'crates\channel' }
+otap-df-engine-macros = { path = 'crates\engine-macros' }
+otap-df-opl = { path = 'crates\opl' }
+otap-df-query-engine = { path = 'crates\query-engine' }
+otap-df-quiver = { path = 'crates\quiver' }
+otap-df-state = { path = 'crates\state' }
+otap-df-telemetry-macros = { path = 'crates\telemetry-macros' }
```
lalitb: Better to use the unix-style path separator, as per the convention used in rust. https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#no-local-path-traversal

sumanthreddy542: Sure, i will update

rust/otap-dataflow/crates/validation/Cargo.toml
```diff
@@ -23,19 +23,19 @@ tokio = { workspace = true }
 portpicker = { workspace = true }
 humantime-serde = { workspace = true }
 
-testcontainers = "0.27"
-minijinja = "1"
+testcontainers = { workspace = true }
+minijinja = { workspace = true }
 reqwest = { version = "0.12", features = ["json"] }
```
drewrelmas: Is there a reason this one can't be a workspace reference? I assume maybe it uses a different version than other crates?

If so, can you please open a follow-up issue to "Align reqwest version usage across otap-dataflow"?

sumanthreddy542: Yes, this uses a different version and it seems in the new version, "query" is added as a new feature. Updating the PR to move this as well.

drewrelmas: Thanks! That is ideal solution compared to opening a new issue.

rust/otap-dataflow/Cargo.toml
```diff
@@ -203,9 +203,31 @@ cpu-time = "1.0.0"
 azure_core = {version = "0.33.0", default-features = false }
 azure_identity = {version = "0.33.0", default-features = false }
 flate2 = "1.1.5"
-reqwest = { version = "0.13.1", default-features = false }
+reqwest = { version = "0.13.1", features = ["rustls-no-provider", "json", "query", "http2", "stream"] }
```
drewrelmas, line 206: Did you determine that these features are the minimal set required for all crates in the workspace?

sumanthreddy542: Yes, these are the features that were imported in other crates where reqwest is used.

PR #2460, opened 2026-03-30
azure-monitor-exporter performance improvements

kamilon:
# Change Summary

Avoid expensive clone call on arrow batch. This change alone almost doubles the performance while halving the memory consumption.

Directly serialize the timestamps from byte buffers without intermediary parse.

Pre-serialize known schema column name fields at configuration time. Avoids n allocations per batch where n is column count.

Couple other very minor refactors to avoid a couple negligible allocations.

No issue files for these changes.

Using the checked in unit tests.

Purely in lower memory and CPU consumption and higher max EPS.

### DIFF
--- a/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
@@ -144,14 +144,8 @@ impl Heartbeat {
                     .os_name
                     .clone()
                     .unwrap_or_else(default_heartbeat_os_name),
-                os_major_version: overrides
-                    .os_major_version
-                    .clone()
-                    .unwrap_or(os_major),
-                os_minor_version: overrides
-                    .os_minor_version
-                    .clone()
-                    .unwrap_or(os_minor),
+                os_major_version: overrides.os_major_version.clone().unwrap_or(os_major),
+                os_minor_version: overrides.os_minor_version.clone().unwrap_or(os_minor),
                 version: overrides
                     .version
                     .clone()


cijothomas:
can you run https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/fakegen-ame-local.yaml and share perf numbers before/after.

https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/scripts/engine-metrics.py is a handy script I use to print engine's internal metrics, so we can quickly see the memory, cpu utilization etc.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
```diff
@@ -336,32 +342,72 @@ impl Transformer {
...
     }
 
+    /// Write a body AnyValueView as a JSON string directly.
+    /// For the common string case, writes raw bytes directly to avoid
+    /// the intermediate String allocation from `extract_string_value`.
+    #[inline]
+    fn write_body_value_json<'a, V: AnyValueView<'a>>(value: &V, out: &mut Vec<u8>) {
+        match value.value_type() {
+            ValueType::String => match value.as_string() {
+                Some(s) => Self::write_json_string(s, out),
```
lalitb: This looks like a behavior regression, unless it’s intentional. Body strings used to go through the `extract_string_value` path, which applies `String::from_utf8_lossy`; this new fast path writes the raw `&[u8]` directly. Since our string views are still byte slices, malformed input can now produce invalid JSON instead of a lossy-but-valid string.

kamilon: Good catch! I've added a fallback. In the normal case this remains 0 allocation, invalid UTF8 drops to lossy mode. I've also added a unit test to validate this going forward. I verified it failed before the fix as well.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
```diff
@@ -336,32 +342,73 @@ impl Transformer {
...
+    /// Falls back to `from_utf8_lossy` for malformed input to guarantee valid JSON.
+    #[inline]
+    fn write_body_value_json<'a, V: AnyValueView<'a>>(value: &V, out: &mut Vec<u8>) {
+        match value.value_type() {
+            ValueType::String => match value.as_string() {
+                Some(s) if std::str::from_utf8(s).is_ok() => Self::write_json_string(s, out),
+                Some(s) => {
+                    let lossy = String::from_utf8_lossy(s);
+                    Self::write_json_string(lossy.as_bytes(), out);
+                }
```
utpilla, line 356: This can be simplified. The two `Some(s)` arms can be collapsed into one. `String::from_utf8_lossy` already returns `Cow::Borrowed` (zero-alloc) for valid UTF-8, so the separate `from_utf8().is_ok()` guard is redundant.

```rust
fn write_body_value_json<'a, V: AnyValueView<'a>>(value: &V, out: &mut Vec<u8>) {
    match value.value_type() {
        ValueType::String => match value.as_string() {
            Some(s) => {
                let valid = String::from_utf8_lossy(s);
                Self::write_json_string(valid.as_bytes(), out);
            }
            None => out.extend_from_slice(b"null"),
        },
        _ => {
            let s = Self::extract_string_value(value);
            Self::write_json_string(s.as_bytes(), out);
        }
    }
}
```

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
```diff
@@ -223,12 +223,12 @@ impl AzureMonitorExporter {
         Ok(())
     }
 
-    async fn handle_logs_view<T: LogsDataView>(
+    async fn handle_logs_view(
```
utpilla: Remove the term `view` from the method name? Maybe rename it to something like `handle_logs` or `handle_log_entries`?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
utpilla, line 314: Pre-existing, but this has the same [invalid-UTF-8 risk](https://github.com/open-telemetry/otel-arrow/pull/2460#discussion_r3011188114) that @lalitb flagged for the body path. Attribute string values from zero-copy protobuf views are raw &[u8] and could contain invalid UTF-8. `write_json_string` passes them through without validation, which could produce invalid JSON.

Not a blocker for this PR since the behavior is unchanged for attributes, but worth a follow-up to add the same `from_utf8_lossy` treatment here.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
```diff
@@ -384,6 +431,85 @@ impl Transformer {
...
+        while len > 0 && digits[len - 1] == b'0' {
+            len -= 1;
+        }
+        out.extend_from_slice(&digits[..len]);
+    }
+
+    /// Write a JSON-quoted RFC3339 timestamp directly to the output buffer.
+    /// Avoids String allocation since RFC3339 contains no JSON-special characters.
+    #[inline]
+    fn write_timestamp_json(time_unix_nano: u64, out: &mut Vec<u8>) {
```
utpilla, line 475: nit: `write_json_hex` reserves capacity upfront, but this method makes ~15 individual `push`/`extend_from_slice` calls without reserving. A timestamp is at most 37 bytes: "YYYY-MM-DDThh:mm:ss.nnnnnnnnn+00:00" (1+4+1+2+1+2+1+2+1+2+1+2+1+9+6+1).

Adding `out.reserve(37)` at the top of the fast path would be consistent with `write_json_hex` and avoid potential reallocations on first use, though in practice the buffer is likely already sized from prior fields.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
```diff
@@ -384,6 +431,85 @@ impl Transformer {
...
+    /// Write a 4-digit zero-padded year directly to output bytes.
+    #[inline]
+    fn write_4_digits(n: u32, out: &mut Vec<u8>) {
+        out.push(b'0' + (n / 1000) as u8);
+        out.push(b'0' + ((n / 100) % 10) as u8);
+        out.push(b'0' + ((n / 10) % 10) as u8);
+        out.push(b'0' + (n % 10) as u8);
+    }
+
+    /// Write nanoseconds with dot prefix, trimming trailing zeros.
```
utpilla: This changes the fractional second format. The old `chrono::to_rfc3339()` always outputs 0, 3, 6, or 9 fractional digits (e.g. 100ms becomes `.100`). The new `write_nanos_trimmed` trims all trailing zeros (e.g. 100ms becomes `.1`).

Both are valid RFC3339, but the output is different:

| Nanos | Old (chrono) | New (trimmed) |
|---|---|---|
| 100_000_000 (100ms) | `.100` | `.1` |
| 100_000 (100us) | `.000100` | `.0001` |
| 1_000_000 (1ms) | `.001` | `.001` |

It would be good to check that this is still accepted by the Azure Log Analytics backend. If it turns out to be an issue, `write_nanos_trimmed` could round `len` up to 3, 6, or 9 to match chrono's behavior.

gouslu:
Thanks for these improvements!

Can we run the bench to have a reference point for future changes to transform? Based on my local bench results this is the perf improvement for transform.rs alone:

image

edit: Particularly bench-contrib
image

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
kamilon: Thanks! Updated.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
kamilon: love it! great catch!

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
kamilon: I don't think it's worth the risk. I've modified the algo to trim in groups of 3. Great catch on this one!

jmacd (approved):

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/fakegen-ame-local.yaml
```diff
@@ -72,7 +72,7 @@ groups:
                       "http.status_code": "HttpStatusCode"
 
               auth:
-                method: "dev"
+                method: "none"
```
drewrelmas: This causes an error on a `validate-config` workflow we have running: https://github.com/open-telemetry/otel-arrow/actions/runs/23820197682/job/69431605750?pr=2460

```
  ❌ rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/fakegen-ame-local.yaml
     
     ⚠️  WARNING: This binary was compiled in debug mode.
        Debug builds are NOT recommended for production, benchmarks, or performance testing.
        Use 'cargo build --release' for optimal performance.
     Error: Custom { kind: Other, error: "Invalid config for component `urn:microsoft:exporter:azure_monitor` in pipeline_group=default pipeline=main node=azure-monitor-exporter: An invalid user configuration occurred: unknown variant `none`, expected one of `managed_identity`, `managedidentity`, `msi`, `cli`, `dev`, `developer`, `development`" }
```

Please make sure that feature is included in the build used for that workflow run.

drewrelmas: New feature should be included [here](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/scripts/validate-configs.sh).

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/auth.rs
```diff
@@ -141,6 +163,8 @@ impl Auth {
                 DeveloperToolsCredentialOptions::default(),
             ))
             .map_err(|e| Error::create_credential(AuthMethod::Development, e))?),
+            #[cfg(feature = "azure-monitor-auth-none")]
```
gouslu: can we achieve this in another way? we can add this option to auth config:

/// Bypass real authentication (sends a dummy token; for local/mock environments)
#[serde(alias = "bypass")]
Bypass,

and what bypass does is to provide a mock bearer token. same idea, but it doesn't create a separate compilation path, which is rather annoying to deal with honestly. bypass implies intent clearly and we can even create a warning log if it starts with bypass to log for any misuses.

drewrelmas: +1 to this suggestion, I'm not sure a workspace level feature is required for something this small. I think we'd like to keep features at the scope of including 'contrib' node types in the build or compiling with certain dependency chains.

Ex: #2343

Another option could be:
```rust
#[serde(alias = "mock")]
Mock,
```

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/fakegen-ame-local.yaml
drewrelmas: Alternatively, see https://github.com/open-telemetry/otel-arrow/pull/2460#discussion_r3018638057

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/auth.rs
```diff
@@ -10,6 +10,28 @@ use futures::future::LocalBoxFuture;
...
+#[cfg(feature = "azure-monitor-auth-none")]
+#[async_trait::async_trait]
+impl TokenCredential for StaticTokenCredential {
+    async fn get_token(
+        &self,
+        _scopes: &[&str],
+        _options: Option<azure_core::credentials::TokenRequestOptions<'_>>,
+    ) -> azure_core::Result<AccessToken> {
+        Ok(AccessToken {
+            token: "static-mock-token".to_string().into(),
```
utpilla: nit: `to_string()` is not needed.

```suggestion
            token: "static-mock-token".into(),
```

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/auth.rs
kamilon: I'll just remove the unauth path entirely.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
```diff
@@ -30,6 +30,12 @@ pub enum AuthMethod {
     /// Use developer tools (Azure CLI, Azure Developer CLI)
     #[serde(alias = "dev", alias = "developer", alias = "cli")]
     Development,
+
+    /// No authentication — sends a static dummy token.
```
cijothomas: not sure of this one. if such a thing is needed, please do it as part of a *separate PR*, and keep this one focused only on the clone-removal/perf-improvement.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
```diff
@@ -100,18 +100,6 @@ fn parse_os_version() -> (String, String) {
     }
 }
 
-#[inline]
```
cijothomas, line 103: please avoid bundling unrelated changes to the same PR.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/fakegen-ame-local.yaml
kamilon: Resolved by removing unauth path

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
kamilon: Removed this path

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
kamilon: Why is this unrelated? Avoids an allocation in this exporter.

cijothomas: my bad. I thought this was doing unrelated refactoring! Thanks for fixing the unwanted allocation even in a non-hot path!

Issue #2465, opened 2026-03-30
Improve control delivery, shutdown robustness, and local deferred work


lquerel:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Objective

This issue describes a multi-step plan to improve node control delivery, inbox behavior, node-local wakeups, and local delayed resume.

A series of PRs will incrementally move the engine in the direction described here.

- node control delivery
- inbox arbitration of control and pdata
- node-local wakeups and delayed resume
- engine-owned shutdown orchestration

This series solves #2431 and prepares for #2432, #2433, #2424, and #2427.

### Rationale

The current runtime mixes several concerns that have different performance and correctness requirements.

In particular:
- forced shutdown should not depend on ordinary control backlog
- node-local deferred work should not create unnecessary global contention
- wakeup-style scheduling and delayed pdata resume are different behaviors and should not share the same overloaded mechanism
- the runtime should remain robust under heavy Ack/Nack traffic in a thread-per-core, single-threaded async model

Durable and batch processors use the DelayedData mechanism in a roundabout way to implement a wake-up mechanism. Points 2 and 3 are intended to address this gap.

### Scope

Planned PR stack:

  1. #2466 
      - introduce the new control-channel design as an isolated crate with benchmarks and documentation
  2. #2469 
      - rename ProcessorMessageChannel / ExporterMessageChannel to ProcessorInbox / ExporterInbox
      - clarify the role of these runtime abstractions before adding new responsibilities
  3. #2470 
      - add keyed wakeup scheduling/cancellation for processors
      - migrate wakeup-style delayed-data uses in batch and durable-buffer
  4. #2471 
      - add requeue_later(...) for processor-local delayed pdata resume
      - migrate retry off the global delayed-data path
  5. #2472
      - delete runtime-managed delayed-data scheduling and related plumbing once migration is complete
  6. Integrate the control-channel design into the engine
      - wire the new control-channel into receivers and node inboxes
      - preserve engine-owned shutdown sequencing and liveness guarantees
  7. Observability and admin UI follow-up
      - add control-channel/inbox-focused metrics and admin UI support once the runtime design is stable

### Acceptance Criteria

- [ ] The work is split into focused PRs with clear scope boundaries
- [ ] Node-local wakeups and delayed resume no longer rely on the old global delayed-data path
[...]

Issue #2431, opened 2026-03-25
Guarantee deadline-forced shutdown delivery under saturated node-control channels


lquerel:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Objective

Ensure that the runtime-control manager's deadline-forced shutdown path still guarantees delivery of `NodeControlMsg::Shutdown` even when node-control channels are already saturated.

### Rationale

The current receiver-first drain design is the right direction, but the deadline-forced branch currently reuses the normal non-blocking `send()` path. If a node-control inbox is full, the forced shutdown message can be deferred into `pending_sends`. The runtime-control manager then exits immediately, which weakens the intended guarantee: the pipeline can decide to force shutdown without ensuring those final shutdown messages were actually delivered.

### Scope

This follow-up covers:
- `RuntimeCtrlMsgManager`
- the deadline-forced branch in `pipeline_ctrl.rs`
- delivery guarantees for `NodeControlMsg::Shutdown` under saturated node-control channels
- tests for forced shutdown under control-channel backpressure

### Acceptance Criteria

- deadline-forced shutdown delivery remains guaranteed even when node-control channels are full
- forced shutdown messages are not left stranded in `pending_sends` when the runtime-control manager exits
- the behavior under saturated node-control channels is explicit and test-covered
- DST and/or unit tests reproduce the saturated-channel case and verify the forced shutdown still reaches target nodes

### Additional Context

See this comment/thread for context -> https://github.com/open-telemetry/otel-arrow/pull/2370#discussion_r2977333043

Issue #2374, opened 2026-03-18
Abstract async runtime usage to allow switching to a different runtime in the future


utpilla:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

## Summary

The engine uses a thread-per-core, shared-nothing design with tokio `current_thread` runtime instances per core. However, node implementations (receivers, processors, and exporters) directly use tokio APIs and macros instead of going through the `EffectHandler` abstraction provided by the engine. This couples the entire node ecosystem to tokio and prevents swapping the async runtime in the future (e.g. to a runtime more suited to the thread-per-core model).

## Problem

`EffectHandler` is the engine's capability-injection boundary for nodes. It already abstracts several async capabilities (periodic timers, TCP/UDP socket creation, metrics reporting, inter-node messaging). However, a significant number of async operations bypass it entirely via direct `tokio::` calls.

An audit of production code in `core-nodes` and `contrib-nodes` reveals [many direct tokio API usages](https://github.com/search?q=repo%3Aopen-telemetry%2Fotel-arrow+%28path%3Arust%2Fotap-dataflow%2Fcrates%2Fcore-nodes+OR+path%3Arust%2Fotap-dataflow%2Fcrates%2Fcontrib-nodes%29+tokio%3A%3A&type=code) across the following categories:

> **Note:** Test code (`#[tokio::test]`, `#[cfg(test)]` modules) is out of scope — tests can continue to use tokio directly.

### 1. Timers and Delays

Direct use of `tokio::time::sleep()`, `tokio::time::interval()`, `tokio::time::timeout()`, and `tokio::time::Instant`:

- `delay_processor` — `tokio::time::sleep()`
- `syslog_cef_receiver` — `tokio::time::interval_at()`, `tokio::time::timeout()`, `tokio::time::Instant`
- `fake_data_generator` — `tokio::time::{Instant, sleep}`
- `otap_exporter` — `tokio::time::sleep()` for backoff
- `azure_monitor_exporter` — `tokio::time::sleep()` / `sleep_until()` / `timeout()` / `Instant` for retry logic, token refresh, periodic export

### 2. `tokio::select!` macro

Used as the primary concurrency combinator in nearly every receiver and exporter event loop:

[...]

PR #2466, opened 2026-03-30
Add standalone control-aware bounded channel for future engine integration

lquerel:
> **Note for reviewers**: I recommend starting the review with the README.md, then the tests, and finally the code to better understand the design and the guarantees provided by this control channel.

## Change Summary

Add a new standalone `control-channel` crate that defines a bounded control-aware channel for OTAP node-control traffic.

This change is intended as a first step toward the broader engine redesign tracked in #2465. It introduces the control-channel design, tests, benchmarks, and documentation in a standalone form so the control-plane  semantics can be reviewed and validated before integration into the engine  runtime.

The new channel is designed for the control-path requirements of the OTAP engine:

- reserved lifecycle delivery for `DrainIngress` and `Shutdown`
- bounded and batched `Ack` / `Nack` handling
- coalesced best-effort control for `TimerTick` / `CollectTelemetry`
- latest-wins `Config`
- bounded fairness and deadline-bounded terminal shutdown progress
- role-specific APIs for receiver vs non-receiver nodes
- a single-owner implementation aligned with the engine's thread-per-core model
- FIFO blocked-sender wakeups for completion-capacity backpressure
- explicit completion metadata support via `AckMsg<PData, Meta>` /
  `NackMsg<PData, Meta>` for future engine unwind-state integration

This PR does not yet wire the new control channel into receivers, processors, exporters, or `pipeline_ctrl`. That integration is planned in the later steps of #2465.

* Part of #2465
* Prepares #2431

- Added unit tests for the standalone `control-channel` crate covering:
  - lifecycle acceptance and ordering
  - completion batching
  - bounded fairness
  - shutdown deadline behavior
  - blocked-sender wake behavior and cancellation safety
  - completion metadata preservation
  - invalid config rejection
- Added a benchmark comparing the standalone control-aware channel against the
  current engine control-channel paths under heavy `Ack` / `Nack` traffic and
  mixed control noise
- Verified with: `cargo xtask check`

### Benchmark results

**Ack/Nack Only**

Mode | Implementation | Time (ms) | Throughput (Melem/s) | Delta vs current
-- | -- | -- | -- | --
Local | current_local | 7.22 | 13.85 | baseline
Local | control_aware | 6.39 | 15.64 | ~+12.9% thrpt / -11.4% time
Shared baseline | current_shared | 16.68 | 5.99 | reference only

**Ack/Nack + Control Noise**

Mode | Implementation | Time (ms) | Throughput (Melem/s) | Delta vs current
-- | -- | -- | -- | --
Local | current_local | 7.51 | 13.31 | baseline
Local | control_aware | 6.49 | 15.42 | ~+15.8% thrpt / -13.6% time
Shared baseline | current_shared | 17.01 | 5.88 | reference only

Notes:
- `control_aware` is the standalone single-owner channel introduced in this PR.
- `current_shared` is included as a reference against the current engine shared path.
- The benchmark still runs on a current-thread runtime and is intended to compare the current engine paths with the standalone single-owner control-channel design before engine integration.
  
No user-facing behavior changes yet.

This PR adds the standalone control channel and its validation/benchmarking surface, but does not yet integrate it into the engine runtime.


rust/otap-dataflow/crates/control-channel/README.md
```diff
@@ -0,0 +1,349 @@
...
+The control channel is meant for node-control traffic, not for pdata transport.
+
+It targets the specific needs of the OTAP engine:
+
+- thread-per-core execution
+- single-threaded async runtimes on the hot path
+- frequent `Ack` and `Nack` traffic when `wait_for_result` is enabled,
+  with batching to amortize receive-side overhead
+- node-local lifecycle transitions such as `DrainIngress` and `Shutdown`
+- a need for bounded memory, bounded-fairness, and explicit terminal progress
```
jmacd:
```suggestion
- a need for bounded memory, bounded-fairness, and explicit termination progress
```

rust/otap-dataflow/crates/control-channel/README.md
```diff
@@ -0,0 +1,349 @@
...
+There are two channel families:
+
+- `receiver_channel(...)`
+  - sender type: `ReceiverControlSender`
+  - receiver event type: `ReceiverControlEvent`
+  - supports both `accept_drain_ingress(...)` and `accept_shutdown(...)`
+- `node_channel(...)`
+  - sender type: `NodeControlSender`
+  - receiver event type: `NodeControlEvent`
+  - supports `accept_shutdown(...)` only
```
jmacd: I think I understand, at least this is how I understand it:

```suggestion
There are two channel families used between the engine controller and the individual nodes. Receivers have a different interface than the other node types, because they begin the shutdown sequence.

- For sending messages to receivers, a controller uses `receiver_channel(...)`
  - channel type: `ReceiverControlSender`, `ReceiverControlReceiver`
  - receiver event type: `ReceiverControlEvent`
  - channel receiver supports both `accept_drain_ingress(...)` and `accept_shutdown(...)`
- For sending messages to other nodes, a controller uses `node_channel(...)`
  - channel type: `NodeControlSender`, `NodeControlReceiver`
  - receiver event type: `NodeControlEvent`
  - channel receiver supports `accept_shutdown(...)` only
```

rust/otap-dataflow/crates/control-channel/README.md
```diff
@@ -0,0 +1,349 @@
...
+
+This split is intentional. `DrainIngress` is receiver-specific lifecycle
+control, so non-receiver nodes cannot express it through the public API.
```
jmacd:
```suggestion
This split is intentional. `DrainIngress` is a receiver-specific lifecycle
control, so non-receiver nodes cannot express it through the public channel receiver API.
```

rust/otap-dataflow/crates/control-channel/README.md
```diff
@@ -0,0 +1,349 @@
...
+
+The crate intentionally exposes one single-owner implementation. That matches the
+thread-per-core execution model the engine is targeting and keeps the control
+state machine behind one receiver-owned queue core.
```
jmacd:
```suggestion
The crate intentionally exposes one single-owner channel implementation. Because of
the thread-per-core execution model, each node has unsynchronized access to the engine's
channel receiver for signalling lifecycle events.
```

rust/otap-dataflow/crates/control-channel/src/tests.rs
```diff
@@ -0,0 +1,832 @@
...
+    assert_eq!(
+        tx.try_send(ControlCmd::CollectTelemetry).unwrap(),
+        SendOutcome::DroppedDuringDrain
+    );
+    assert_eq!(tx.try_send(ack("ack-1")).unwrap(), SendOutcome::Accepted);
+
+    let stats = tx.stats();
+    assert!(stats.drain_ingress_recorded);
+    assert!(!stats.shutdown_recorded);
+    assert_eq!(stats.normal_event_dropped_during_drain_total, 3);
```
jmacd, line 241: What does "normal" relate to, non-lifecycle events?

rust/otap-dataflow/crates/control-channel/src/core.rs
```diff
@@ -0,0 +1,584 @@
...
+    AdmissionClass, CompletionMsg, ControlChannelConfig, ControlChannelStats, ControlCmd,
+    DrainIngressMsg, LifecycleSendResult, Phase, SendOutcome, ShutdownMsg, TrySendError,
+};
+use std::collections::VecDeque;
+use std::time::Instant;
+
+/// Delivery class used by the queue core when it round-robins pending normal
+/// non-completion control work.
+#[derive(Clone, Copy)]
+enum NormalEventClass {
```
jmacd, line 34: I see. Normal is non-completion, non-lifecycle.

rust/otap-dataflow/crates/control-channel/README.md
lquerel: I don't fully agree with this suggestion. Nodes do not necessarily have unsynchronized access to the controller. We still need a local and shared effect handler depending on the nature of the node itself. Among other things, the effect handler is used for communication from the node back to the controller. What is now unsynchronized is the consumption of the node control channel.

rust/otap-dataflow/crates/control-channel/README.md
```diff
@@ -0,0 +1,349 @@
...
+  - channel type: `NodeControlSender`, `NodeControlReceiver`
+  - receiver event type: `NodeControlEvent`
+  - channel receiver supports `accept_shutdown(...)` only
+
+This split is intentional. `DrainIngress` is a receiver-specific lifecycle
+control, so non-receiver nodes cannot express it through the public channel receiver API.
+
+The crate intentionally exposes one single-owner implementation. That matches the
+thread-per-core execution model the engine is targeting and keeps the control
+state machine behind one receiver-owned queue core.
```
jmacd: I'm not sure what it is about these two sentences, here's another attempt:

```suggestion
The crate exposes a single-owner implementation for use by a single engine core. 
This design fits our thread-per-core execution model and keeps the control
state machine behind one receiver-owned or node-owned **queue core**.
```

rust/otap-dataflow/crates/control-channel/src/channel.rs
```diff
@@ -0,0 +1,1036 @@
...
+use std::cell::{Cell, RefCell};
+use std::collections::VecDeque;
+use std::future::Future;
+use std::pin::Pin;
+use std::rc::Rc;
+use std::task::{Context, Poll, Waker};
+use tokio::sync::Notify;
+use tokio::time::{Instant as TokioInstant, Sleep, sleep_until};
+
+// TODO: Consider deduplicating the keyed blocked-sender waiter logic with
```
jmacd, line 59: I take it the code below is copied from there. How come these cases do not use the `slotmap` crate, by the way?

rust/otap-dataflow/crates/control-channel/src/channel.rs
```diff
@@ -0,0 +1,1036 @@
...
+    LifecycleSendResult, NodeControlEvent, ReceiverControlEvent, SendError, SendOutcome,
+    ShutdownMsg,
+};
+use std::cell::{Cell, RefCell};
+use std::collections::VecDeque;
+use std::future::Future;
+use std::pin::Pin;
+use std::rc::Rc;
+use std::task::{Context, Poll, Waker};
+use tokio::sync::Notify;
```
jmacd, line 56: I worry this makes https://github.com/open-telemetry/otel-arrow/issues/2374 even more a moving target.

rust/otap-dataflow/crates/control-channel/src/core.rs
```diff
@@ -0,0 +1,584 @@
...
+/// owns all pending control state and decides admission and delivery order.
+/// It does not perform async waiting, wakeups, or handle management; the outer
+/// channel layer in `channel.rs` adds those behaviors around this state.
+///
+/// `Inner<PData, Meta>` stores pending lifecycle, completion, and normal
+/// control work and emits [`CoreControlEvent`] values in policy order. The
+/// generic parameter `Meta` is carried only inside completion messages
+/// (`Ack`/`Nack`), so the queue core stays agnostic about any unwind or
+/// routing meaning attached to that metadata.
+pub(crate) struct Inner<PData, Meta = ()> {
```
jmacd, line 67: Interesting trick

rust/otap-dataflow/crates/control-channel/src/core.rs
```diff
@@ -0,0 +1,584 @@
...
+    /// Current lifecycle phase. This drives which classes can still be
+    /// admitted and which events may be delivered next.
+    pub(crate) phase: Phase,
+    /// Terminal close flag. Once set, no new sends are accepted and the
+    /// receiver returns `None` after buffered work is drained.
+    pub(crate) closed: bool,
+    // Generation counter used with `Notify` to avoid check-then-sleep races in
+    // sender/receiver wait loops. Any state transition that can unblock a
+    // waiter must bump this value before notifications are observed.
+    pub(crate) version: u64,
```
jmacd, line 79: The comment is a little unclear. I know the keys have generations, and this is also a generation but it is named version, and Notify is a Tokio runtime wakeup mechanism.

rust/otap-dataflow/crates/control-channel/src/channel.rs
```diff
@@ -0,0 +1,1036 @@
...
+    }
+}
+
+struct State<PData, Meta = ()> {
+    /// Single-threaded ownership of the queue core plus the wake state layered
+    /// around it. `Inner` owns all admission, batching, fairness, and
+    /// lifecycle semantics; this wrapper only handles channel-receiver waiting,
+    /// sender liveness, and wake routing.
+    inner: RefCell<Inner<PData, Meta>>,
+    notify: Notify,
```
jmacd, line 276: Comment on how this is used.

rust/otap-dataflow/crates/control-channel/src/channel.rs
```diff
@@ -0,0 +1,1036 @@
...
+        self.state
+            .sender_count
+            .set(self.state.sender_count.get().saturating_add(1));
+        Self {
+            state: Rc::clone(&self.state),
+        }
+    }
+}
+
+impl<PData, Meta> Drop for ControlSenderCore<PData, Meta> {
```
lalitb, line 379: Should `ControlReceiverCore` also close the channel and wake blocked senders on drop? As written, if the receiver task goes away while senders are blocked in `send().await`, I think those waiters can hang indefinitely.

rust/otap-dataflow/crates/control-channel/src/channel.rs
```diff
@@ -0,0 +1,1036 @@
...
+        }
+    }
+
+    fn wake_n(&mut self, mut count: usize) {
+        while count > 0 {
+            let Some(key) = self.queue.pop_front() else {
+                break;
+            };
+            let Some(slot) = self.slots.get_mut(key.index) else {
+                self.queued_stale -= 1;
```
utpilla: `queued_stale` is usize, so this underflows (panics in debug, wraps to `usize::MAX` in release) if it's already 0 when this branch is hit. Same issue with `wake_all` as well.


```suggestion
                self.queued_stale = self.queued_stale.saturating_sub(1);
```

utpilla:
Thanks @lquerel for the thorough design and documentation here. This is well-tested and the behavior guarantees are clearly stated.

I wanted to raise a design alternative for discussion: composing this on top of existing primitives rather than building a custom state machine.

### The idea

Same external API, but internally backed by two channels from `otap-df-channel` plus shared state:

```
lifecycle:    otap_df_channel::mpsc(2)         // DrainIngress + Shutdown
completion:   otap_df_channel::mpsc(capacity)   // Ack/Nack — bounded, backpressured
config:       Rc<RefCell<Option<Value>>>        // latest-wins shared slot
timer_tick:   Rc<Cell<bool>>                    // coalesced shared flag
telemetry:    Rc<Cell<bool>>                    // coalesced shared flag
```

Only lifecycle and completions need real channel semantics. Config and best-effort signals are just shared mutable state, which is what the PR already uses internally, just wrapped inside a custom queue core. The receiver-side policy (lifecycle precedence, batching, fairness, deadline-forced shutdown) would live in the `Inbox` layer that already arbitrates between control and pdata today. Node component authors are unaffected either way. They call `inbox.recv()` and get back typed events regardless of what's behind it.

### Pros

- **Reuses already-reviewed (trusted) code:** The `SenderWaiters` implementation here is pretty similar to `otap-df-channel::mpsc`. Reusing proven implementations reduces risk such as https://github.com/open-telemetry/otel-arrow/pull/2466#discussion_r3019035043

- **Lesser new code to maintain:** ~100-150 lines of receiver-side policy vs ~1600 lines of custom channel + state machine, because backpressure, FIFO wake ordering, and cancellation safety come from existing `otap-df-channel` code rather than being reimplemented.

- **Shared-mode for free:** The current crate is `Rc`/`RefCell` only, but control channels today always use `SharedSender` via `tokio::sync::mpsc`. The existing `ChannelMode` trait already abstracts local vs shared, composing on existing channels gets both modes without a parallel implementation.

- **Policy lives where it's consumed:** The `Inbox` (#2469) already handles control-vs-pdata scheduling today. Adding lifecycle precedence and deadline-forced shutdown there is a natural extension rather than splitting that responsibility across two layers.

### Cons

- **Self-contained testability:** The current design is a clean, isolated policy engine testable independently of the inbox. The alternative distributes policy into the inbox receiver loop.

- **Atomic stats snapshot:** With split state, `stats()` reads from multiple sources. In practice this doesn't matter on a single-threaded runtime (sequential reads without `.await` are effectively atomic), but it's a theoretical difference.

I feel like using the existing `otap-df-channel` implementations with the policy in the Inbox layer would be simpler to implement, maintain, and integrate.

rust/otap-dataflow/crates/control-channel/src/core.rs
```diff
@@ -0,0 +1,584 @@
...
+    /// matter once ingress shutdown has begun. The `DrainIngressMsg::deadline`
+    /// is carried through for the receiver event loop to interpret; unlike
+    /// `Shutdown`, it does not arm queue-core forced-progress behavior here.
+    pub(crate) fn record_drain_ingress(&mut self, msg: DrainIngressMsg) -> LifecycleSendResult {
+        if self.closed {
+            return LifecycleSendResult::Closed;
+        }
+        if self.drain_ingress_recorded {
+            return LifecycleSendResult::AlreadyAccepted;
+        }
```
lalitb, line 278: Curious why there is no `forced-shutdown` check here. Once the deadline expires, graceful time is effectively over,  but a late `DrainIngress` still gets accepted and surfaces before forced shutdown path in `pop_event()`. Since `try_send` already gates on `shutdown_forced` - feels like `record_drain_ingress` should follow the same rule ?

rust/otap-dataflow/crates/control-channel/src/types.rs
```diff
@@ -0,0 +1,377 @@
...
+    }
+}
+
+impl ControlChannelConfig {
+    /// Validates channel configuration.
+    pub fn validate(&self) -> Result<(), ConfigError> {
+        if self.completion_msg_capacity == 0 {
+            return Err(ConfigError::ZeroCompletionMsgCapacity);
+        }
+        if self.completion_batch_max == 0 {
```
ThomsonTan, line 59: Also check and error on `self.completion_batch_max > self.completion_msg_capacity` which is also invalid?

rust/otap-dataflow/crates/control-channel/src/core.rs
```diff
@@ -0,0 +1,584 @@
...
+        if forced {
+            self.completion_abandoned_on_forced_shutdown_total = self
+                .completion_abandoned_on_forced_shutdown_total
+                .saturating_add(self.completion.len() as u64);
+            self.completion.clear();
+        }
+
+        self.clear_normal_pending();
+        self.shutdown_deadline = None;
+        self.shutdown_forced = false;
```
lalitb: nit - this seems inconsistent with the doc comment for this field in `types.rs`:
```rust   
 /// Whether the shutdown deadline has already forced terminal progress.
    pub shutdown_forced: bool,
```

After this line, stats() will report `false` even if forced shutdown did fire. Should this stay set, or should the doc instead say something like "currently forced" ?

rust/otap-dataflow/crates/control-channel/src/channel.rs
lquerel: Indeed, good catch!

lquerel: Fixed in https://github.com/open-telemetry/otel-arrow/pull/2466/commits/b50c9c2dece7fac2053c607437a0663af2af1a20

lquerel:
Thanks @utpilla for the suggestion. I agree the composition approach is interesting, but I'm hesitant to pivot this PR in that direction for a few reasons.

First, the "shared mode for free" benefit is no longer very compelling in the current plan. The control-channel is intended to be integrated as a local-only control path. In practice, the part that still needs local/shared flexibility is the effect handler and some outbound/shared senders, not node-control delivery itself.

Second, moving the policy "into the Inbox" is not as local a change as it sounds. Control is not consumed only by ProcessorInbox / ExporterInbox (future PRs), receiver nodes also consume control directly in their own event loops. So if lifecycle precedence, completion batching, and deadline-forced shutdown move out of the control-channel, that policy either gets duplicated or reintroduced in another shared receiver-side layer. At that point, the simplification is less clear. Let me know if I misinterpreted your proposal.

Third, reusing `otap-df-channel` is not automatically stronger than the current implementation. The control-channel waiter path now has bounded stale-entry compaction, whereas `crates/channel/src/mpsc.rs` still relies on lazy tombstone cleanup only. So on this point, the current control-channel is actually stricter about bounded auxiliary state under cancellation churn.

More broadly, even with composition, a substantial amount of custom policy still remains:
- reserved-capacity lifecycle admission
- Accepted vs AlreadyAccepted lifecycle semantics
- latest-wins config
- coalesced timer / telemetry signals
- shutdown deadline / forced-progress behavior
- stats snapshots

So the alternative reduces some channel mechanics, but it does not remove most of the specialized control policy. It mainly redistributes it.

So my current preference is:
- keep this standalone control-channel design for PR1
- fix correctness issues directly in this design (the ones you and Lalit mentioned)
- keep the crate isolated until the later engine-integration step
- revisit possible reuse/extraction later, once the integration shape is settled

I do think there is useful follow-up work here, especially around deduplicating bounded waiter machinery with `otap-df-channel`, but I would rather do that in a second phase than expand the scope of this PR.

I created the follow GH issue to track this follow-up https://github.com/open-telemetry/otel-arrow/issues/2492

rust/otap-dataflow/crates/control-channel/src/channel.rs
lquerel: Good catch.

lquerel: I applied this suggestion in more than one place, see https://github.com/open-telemetry/otel-arrow/pull/2466/commits/956d53a151f31f77b635518d483ed7ba8bbc21fd

rust/otap-dataflow/crates/control-channel/src/types.rs
lquerel: It is not needed for correctness, but it is good config hygiene so I agree. See this commit https://github.com/open-telemetry/otel-arrow/pull/2466/commits/aa4c06d5ee6a8ba014a43fedf534a705ebcb1bc4

rust/otap-dataflow/crates/control-channel/src/core.rs
lquerel: Good catch working on an improvement

rust/otap-dataflow/crates/control-channel/src/core.rs
lquerel: Good catch. I agree it was inconsistent.

I fixed it by keeping shutdown_forced latched once forced shutdown fires, so stats() stays aligned with the current docs. I also added a regression check for that behavior.

PR #2470, opened 2026-03-31
Add node-local wakeups to ProcessorInbox

lquerel:
## Change Summary

Add a node-local wakeup API for processors and integrate it into `ProcessorInbox`.

This PR introduces keyed, cancelable wakeups that let a processor schedule local wakeup work without sending dummy `pdata` back through the old delayed data path. Due wakeups are surfaced through `ProcessorInbox` as control
  messages.

This is the next step in the redesign tracked in #2465. The goal is to split two different behaviors that were previously mixed together:

- wakeup-style local scheduling
- delayed `pdata` resume

This PR handles the first one only. It migrates the wakeup-style usages in `batch_processor` and `durable_buffer_processor`, while intentionally leaving retry and the old global delayed-data path in place for the next PRs.

* Part of #2465

- Added/updated engine tests covering:
  - local wakeup scheduling
  - wakeup replacement and cancellation
  - `ProcessorInbox` delivery of due wakeups
  - shutdown behavior for pending wakeups
- Added/updated processor tests covering:
  - batch wakeup-driven flush behavior
  - durable-buffer wakeup-driven retry/resume behavior
- Verified with:
  - `cargo xtask check`

No user-facing changes.

This PR is an internal runtime change only.

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
@@ -34,6 +34,7 @@ use otap_df_otap::otap_grpc::otlp::client::{
     LogsServiceClient, MetricsServiceClient, TraceServiceClient,
 };
 use otap_df_otap::pdata::{Context, OtapPdata};
+use otap_df_otap::transport_headers::ValueKind;
 use otap_df_pdata::otlp::logs::LogsProtoBytesEncoder;
 use otap_df_pdata::otlp::metrics::MetricsProtoBytesEncoder;
 use otap_df_pdata::otlp::traces::TracesProtoBytesEncoder;
@@ -47,6 +48,7 @@ use std::future::Future;
 use std::sync::Arc;
 use std::time::Duration;
 use tonic::codec::CompressionEncoding;
+use tonic::metadata::{MetadataKey, MetadataMap, MetadataValue};
 use tonic::transport::Channel;
 
 /// The URN for the OTLP gRPC exporter
@@ -262,12 +264,17 @@ impl Exporter<OtapPdata> for OTLPExporter {
                     let (context, payload) = pdata.into_parts();
                     self.pdata_metrics.inc_consumed(signal_type);
 
+                    // Build gRPC metadata from transport headers if a propagation
+                    // policy is configured. Computed once before signal dispatch.
+                    let metadata = build_grpc_metadata(&effect_handler, &context);
+
                     // Dispatch based on signal type and the concrete payload representation.
                     match (signal_type, payload) {
                         (SignalType::Logs, OtapPayload::OtapArrowRecords(otap_batch)) => {
                             dispatch_otap_export(
                                 otap_batch,
                                 context,
+                                metadata,
                                 SignalType::Logs,
                                 &exporter_id,
                                 &mut logs_proto_buffer,
@@ -286,6 +293,7 @@ impl Exporter<OtapPdata> for OTLPExporter {
                             dispatch_otap_export(
                                 otap_batch,
                                 context,
+                                metadata,
                                 SignalType::Metrics,
                                 &exporter_id,
                                 &mut metrics_proto_buffer,
@@ -304,6 +312,7 @@ impl Exporter<OtapPdata> for OTLPExporter {
                             dispatch_otap_export(
                                 otap_batch,
                                 context,
+                                metadata,
                                 SignalType::Traces,
                                 &exporter_id,
                                 &mut traces_proto_buffer,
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/crates/engine/src/node_local_scheduler.rs
```diff
@@ -0,0 +1,543 @@
...
+    fn repair_heap_at(&mut self, index: usize) {
+        if index > 0 {
+            let parent = (index - 1) / 2;
+            if Self::wakeup_precedes(&self.wakeups[index], &self.wakeups[parent]) {
+                self.sift_up(index);
+                return;
+            }
+        }
+        self.sift_down(index);
+    }
```
jmacd, line 146: Does Rust really not have a crate for this? Hmm (thinking like Golang https://pkg.go.dev/container/heap). Don't get me wrong-- I love heaps! I would put this heap-helper stuff in a new data structure-only file.

jmacd (approved):

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/retry_wakeup_state.rs
```diff
@@ -0,0 +1,460 @@
...
+    pub(super) const fn bundle_ref(self) -> BundleRef {
+        self.bundle_ref
+    }
+
+    pub(super) const fn retry_count(self) -> u32 {
+        self.retry_count
+    }
+}
+
+/// Retry state for a bundle that could not acquire an engine wakeup slot yet.
```
AaronRM: I know this is still a WIP, so not leaving detailed comments yet. Just wanted to chime in early on the approach to hopefully reduce overwork.

My general concern in the `durable_buffer` changes specifically (not the general mechanism) is that we'd be doing something relatively complex with the 'many wakeup slots + OverflowRetry' approach at a time when we can least afford to do so (i.e. under heavy NACK pressure from a network outage).

Instead of one wakeup slot per bundle, could we simplify by only maintaining a single `WakeupSlot` that tracks the earliest pending retry deadline. When a NACK arrives (or the wakeup fires and we need to reschedule), we compute the backoff time and call `set_wakeup(slot, min(current_scheduled, new_retry_at))`. When the wakeup fires, scan Quiver for all bundles whose backoff has elapsed and resend them. We still get exact timing for the next retry without per-bundle slot management. The state the `durable_buffer_processor` maintains reduces to one `Instant` and one wakeup slot. The `OverflowRetry` mechanism goes away in the durable buffer (no longer needed) and we don't need to coordinate between multiple scheduling methods.

lquerel:
@AaronRM  and @JakeDern , could you each review the integration I made of the new wakeup API, respectively in the durable buffer processor and the batch processor, replacing the delayed data control message that was used before?

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/retry_wakeup_state.rs
lquerel: That sounds reasonable, I'll take a look at what I can do.

lquerel: Please take a look at the updated code.

lquerel:
@AaronRM At the durable buffer retry state level, there is still a potential issue with unbounded heap memory usage. With the new wakeup API, I think the situation is better than before, but the heap footprint can still grow with the number of deferred bundles.

During a prolonged transient outage with small bundles, the durable backlog can become large in terms of in-memory retry metadata. I did not try to fix this because it goes well beyond the integration of the wakeup API, but I think a long-term solution would be to move the `retry_at` metadata directly into durable storage so that retry bookkeeping remains bounded in memory.

That said, I don't have your full design in mind, and this may not be the best approach. In any case, I'll leave it to you to take a look.

rust/otap-dataflow/crates/engine/src/effect_handler.rs
```diff
@@ -397,6 +407,48 @@ impl<PData> EffectHandlerCore<PData> {
         })
     }
 
+    /// Set or replace a processor-local wakeup.
+    ///
+    /// Wakeups are keyed by [`WakeupSlot`]. Scheduling the same slot again
+    /// replaces the previous due time for that slot and assigns a new
+    /// scheduler revision for that slot.
```
JakeDern, line 414: Very nice!

rust/otap-dataflow/crates/engine/src/control.rs
```diff
@@ -75,6 +75,26 @@ impl From<Context8u8> for f64 {
...
+#[repr(transparent)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct WakeupSlot(pub u128);
+
+/// Monotonic wakeup revision assigned by the scheduler each time a slot is set.
+///
+/// Re-scheduling an existing slot gives it a new revision. Processors can use
+/// the revision carried back in [`NodeControlMsg::Wakeup`] to distinguish a
+/// current wakeup from a stale delivery for the same slot.
+pub type WakeupRevision = u64;
```
JakeDern, line 97: I'm glad this is part of the engine! I was wondering if I'd have to do a similar thing at the processor level

AaronRM:
> @AaronRM At the durable buffer retry state level, there is still a potential issue with unbounded heap memory usage. With the new wakeup API, I think the situation is better than before, but the heap footprint can still grow with the number of deferred bundles.

Thanks @lquerel. The changes with the single slot redesign now look good to me from the durable_buffer perspective. 👍

To your concern re: unbounded heap growth, I filed #2552 to track that separately. I agree that it's not something your PR should address, although as you note, your changes already improve the situation slightly.

rust/otap-dataflow/crates/engine/src/control.rs
lquerel: I thought this would actually be useful for quite a few use cases.

rust/otap-dataflow/crates/engine/src/node_local_scheduler.rs
jmacd: https://github.com/open-telemetry/otel-arrow/issues/2587

PR #2471, opened 2026-03-31
Add node-local delayed resume to ProcessorInbox

lquerel:
# Change Summary

Add processor-local delayed `pdata` resume via `EffectHandler::requeue_later(...)`, integrated into `ProcessorInbox`, and migrate `retry_processor` from the old global `delay_data(...)` path.

This PR completes the delayed-payload side of the node-local scheduling work. The existing wakeup path remains for lightweight timer-style callbacks, while `requeue_later(...)` is for retry-style retained payloads that must be delivered back to the same processor later.

The goal is to move retry-style deferred work out of the global runtime path and into node-local inbox orchestration, while keeping the old global delayed-data plumbing in place temporarily for compatibility. That keeps this PR focused on local delayed resume and retry migration only; the cleanup and removal of the old global path happens in the next PR.

## What Changed

- Added a bounded processor-local delayed-resume queue in the node-local scheduler.
- Exposed `EffectHandler::requeue_later(when, data)` with payload-preserving rejection.
- Delivered due local delayed resumes through `ProcessorInbox` as `NodeControlMsg::DelayedData`.
- Preserved inbox control/pdata fairness for delayed resumes.
- On processor shutdown, pending delayed resumes are surfaced immediately for drain; new delayed resumes are rejected.
- Migrated `retry_processor` to schedule retries with `requeue_later(...)`.

## Benchmark

Results based on a local Criterion benchmark.

| Scenario | `delay_data` | `requeue_later` | Result |
| --- | ---: | ---: | ---: |
| Schedule only | ~1.247 ms / 10k | ~286.9 us / 10k | ~4.35x faster |
| Schedule + due delivery | ~2.720 ms / 10k | ~1.911 ms / 10k | ~1.42x faster |

The new path removes the runtime-control scheduling hop for retry-style retained pdata and keeps delayed resume state local to the processor inbox/scheduler.

* Part of #2465

- Added scheduler tests for delayed resume ordering, capacity, payload-preserving rejection, and shutdown drain conversion.
- Added inbox tests for delayed-resume delivery, fairness, shutdown rejection, and shutdown drain behavior.
- Updated retry tests for local delayed retry and terminal NACK behavior when `requeue_later(...)` rejects.
- Ran targeted engine checks/tests for the scheduler and inbox behavior. 
- `cargo xtask check`

No user-facing changes.

This PR is an internal runtime change only.


rust/otap-dataflow/crates/engine/src/node_local_scheduler.rs
```diff
@@ -90,7 +111,48 @@ pub(crate) struct NodeLocalWakeupSchedulerMetrics {
...
+        Some(self.cmp(other))
+    }
+}
+
+impl<PData> PartialEq for ScheduledResume<PData> {
+    // Payload identity is not part of scheduling identity.
+    fn eq(&self, other: &Self) -> bool {
+        self.when == other.when && self.sequence == other.sequence
+    }
+}
```
jmacd, line 151: It looks like we could, otherwise comment why we're not using `Reverse<>` instead?
https://doc.rust-lang.org/std/cmp/struct.Reverse.html

jmacd (approved):

PR #2473, opened 2026-03-31
Add heartbeat configuration to Azure Monitor Exporter

gouslu:
## Add heartbeat configuration to Azure Monitor Exporter

Adds a `heartbeat` configuration section to the Azure Monitor Exporter, allowing operators to control heartbeat behavior and override auto-detected system fields.

Part of #2475

### Configuration

```yaml
heartbeat:
  enabled: true       # default: false
  frequency: 30       # seconds, default: 60
  overrides:
    version: "2.0.0-custom"
    computer: "my-host"
    os_type: "Linux"
    os_name: "Ubuntu"
    os_major_version: "22"
    os_minor_version: "04"

### DIFF
--- a/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
@@ -57,8 +57,7 @@ struct HeartbeatRow {
 
 #[inline]
 fn default_heartbeat_version() -> String {
-    std::env::var("AZURE_MONITOR_EXPORTER_HEARTBEAT_VERSION")
-        .unwrap_or_else(|_| env!("CARGO_PKG_VERSION").to_string())
+    env!("CARGO_PKG_VERSION").to_string()
 }
 
 #[inline]
@@ -393,17 +392,6 @@ mod tests {
     fn test_default_heartbeat_version_fallback() {
         let version = default_heartbeat_version();
         assert!(!version.is_empty());
-    }
-
-    #[test]
-    #[allow(unsafe_code)]
-    fn test_default_heartbeat_version_uses_env_override() {
-        const ENV_KEY: &str = "AZURE_MONITOR_EXPORTER_HEARTBEAT_VERSION";
-        // SAFETY: This test runs serially (single-threaded test) and restores the env var.
-        unsafe { std::env::set_var(ENV_KEY, "99.0.0-custom") };
-        let version = default_heartbeat_version();
-        unsafe { std::env::remove_var(ENV_KEY) };
-        assert_eq!(version, "99.0.0-custom");
     }
 
     #[test]

### CODE REVIEW

File: rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs [hunk 1]
```diff
@@ -57,7 +57,8 @@ struct HeartbeatRow {
 
 #[inline]
 fn default_heartbeat_version() -> String {
-    env!("CARGO_PKG_VERSION").to_string()
+    std::env::var("AZURE_MONITOR_EXPORTER_HEARTBEAT_VERSION")
```
cijothomas: this is introducing a magic env variable. I'd suggest to make the heartbeat version just be configurable, and whoever is consuming this crate, can pass the version themselves, including from any env variable (eg: https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/configs/fake-debug-noop-telemetry.yaml#L21)

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
```diff
@@ -80,6 +84,64 @@ fn default_scope() -> String {
...
+/// Heartbeat configuration
+#[derive(Debug, Deserialize, Clone)]
+pub struct HeartbeatConfig {
+    /// Whether heartbeat is enabled (defaults to true)
+    #[serde(default = "default_true")]
+    pub enabled: bool,
+
+    /// Heartbeat frequency in seconds (defaults to 60)
+    #[serde(default = "default_heartbeat_frequency")]
+    pub frequency: u64,
```
drewrelmas: Please consider using `humantime_serde` and `Duration` for this value - it should enable things like configuring `30s` or `2m` instead of restricting unit to seconds.

Example from `durable_buffer_processor`:

https://github.com/open-telemetry/otel-arrow/blob/f8611f8124df885610809c3b99cc2e7434b1a0c2/rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/config.rs#L136-L138

gouslu: very good point!

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
```diff
@@ -57,7 +57,8 @@ struct HeartbeatRow {
...
+        .unwrap_or_else(|_| env!("CARGO_PKG_VERSION").to_string())
```
jmacd: In the end, we kept this. OK! See #2475 anyway.

drewrelmas: Oh - I think this should have been removed - missed it on my re-review, too many PRs is a good problem to have! 😄

gouslu: oh, good catch, will revert to what it used to be

drewrelmas: Now it is actually removed, thanks @gouslu !

PR #2474, opened 2026-03-31
Add basic hashing/substring function call support to query-engine crate

albertlockett:
# Change Summary

Add support for executing functions in the columnar query engine expression evaluation. 

There are three functions now supported:
- `sha256` - computes the sha 256 hash of a string/binary input. result type `DataType::Binary`
- `encode` - encodes a binary input as either hex or base64. result type = `DataType::Utf8` 
- `substring` - extracts a substring of a specified number of characters from a specific starting position in a string. result type `DataType::Utf8` / `DataType::Dict<_, Utf8>` (depends on input)

```kql
logs | set attributes["x"] = encode(sha256(attributes["x"]) "hex")
logs | set attributes["y"] = substring(attributes["y"], 0, 3)
logs | set attributes["y"] = substring(attributes["y"], 3)
```

(btw - open to suggestions if we want to rename these functions, make it possible to hash to hex w/ only one function call, etc.)

For `encode` and `sha256`, we use the datafusion equivalents: ([sha256](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#sha256), [encode](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#encode))

For `substring`, we use a custom implementation loosely based on the datafusion equivalent, with a few minor tweaks:
- supports taking substring from dictionary values directly
- produces a `StringArray` instead of a `StringViewArray`
- start index is 0-based (datafusion uses 1-based start index, similar to Postgres SQL).

Note that, unlike `encode`/`sha256`, `substring` gets parsed [`ScalarExpression::Slice`](https://github.com/open-telemetry/otel-arrow/blob/f8611f8124df885610809c3b99cc2e7434b1a0c2/rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs#L66-L68) instead of a [`ScalarExpression::InvokeFunction`](https://github.com/open-telemetry/otel-arrow/blob/f8611f8124df885610809c3b99cc2e7434b1a0c2/rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs#L46-L47) as this is also how kql parses this expression, and I'm trying to keep the expression tree produced by both parsers as similar as possibe.

Additional parsers changes:
- OPL's function expression parsing now needs access to the functions defined on `PipelineBuilder`, so a `&dyn PipelineBuilder` is now passed throughout many of the expression parsing functions.
- This code cleans up how function definitions were managed by the pipeline builder. Before, we would keep track of max function ID produced by parent/child builders, but now we just have the child builders take a reference to the parent builder and delegate function definition resolution to the root builder.
- The query-engine crate exposes a method called `parsers::default_parser_options` which produces a `ParserOptions` instance that has been populated with the known external function definitions.

* Closes https://github.com/open-telemetry/otel-arrow/issues/2036

Many unit tests

Users of transform processor can now invoke functions `sha256`, `encode` and `substring` in their OPL/KQL programs.


rust/otap-dataflow/crates/query-engine/src/pipeline/functions/substring.rs
```diff
@@ -0,0 +1,633 @@
...
+    ColumnarValue, Documentation, ScalarFunctionArgs, ScalarUDFImpl, Signature,
+};
+use datafusion::scalar::ScalarValue;
+
+/// Replacement for datafusion's [`SubstrFunc`](datafusion::functions::unicode::substr::SubstrFunc)
+/// with a few behaviour changes:
+/// - can operate directly on Dictionary values if the passed source is dictionary encoded and the
+///   start/len arguments are scalars
+/// - returns a `StringArray` instead of a `StringViewArray`. We do this because, OTAP data model
+///   does not support string views.
```
lquerel, line 27: Just out of curiosity, what does it imply that we do not support string views?

albertlockett: My intention here was to point out that in OTAP, we don't have any column types that are have a type of `DataType::Utf8View` (the type of `StringViewArray`). Our string columns are always either `DataType::Utf8` or `Dictionary<_, DataType::Utf8>`.

Datafusion's built-in `substring` function will always produce a `StringViewArray`, but in our case because we don't have any columns that use this type, we'll basically always need to cast it back to `StringArray` before doing the assignment. I figured it was probably faster in most cases to just directly create a `StringArray` as the result.

albertlockett:
> LGTM
> 
> Just out of curiosity, what other functions supported by DataFusion could we easily add to OPL?

There are lots of them which could be easily supported! 
https://datafusion.apache.org/user-guide/sql/scalar_functions.html

Now that we have this PR in place, it should be straight forward to add support for additional functions.

I think we could choose based on what would be most useful. I can foresee some of the string functions like `concat` being good candidates to add next, as well as possibly functions for creating datetimes such as `now`/`current_time`.

lquerel:
> https://datafusion.apache.org/user-guide/sql/scalar_functions.html

I have the impression that we should take advantage of our integration with DataFusion to accelerate support for as many functions as possible in OPL. It strengthens our case for relying on this kind of library.

PR #2478, opened 2026-03-31
AzureMonitor - minor improvement to the example config

cijothomas:
The current config was mapping non-existent attributes/fields, so the transform path was not really executed much. This modifies the mapping to match what is generated by the fake-signal-generator.

Also example command line modified to use 1 core, and release build.
The load-generator is set to "uncapped" mode, so it'll send as much as possible instead of rate limiting.


drewrelmas:
Test failures for a `yaml` change 😢

The windows-latest one at least is another hit that may be resolved with #2476

Requeued

PR #2481, opened 2026-03-31
Gate route_ack/route_nack behind AckNackRouting trait

### LINKED ISSUE #2387: Protect direct callers of EffectHandler route_ack() and route_nack()
### Pre-filing checklist

### Component(s)

### Bug Description

It is possible for callers to directly call the EffectHandler methods (e.g., route_ack and route_nack) when there are ConsumerEffectHandlerExtensions meant to intercept those calls and provide additional functionality. Since these now have the same signature, it's easy to make a mistake.

### Steps to Reproduce

Replace notify_ack and notify_nack with route_ack and route_nack. This compiles. 

### Expected Behavior

The caller has to use notify_ack and notify_nack.

### Actual Behavior

The instrumentation will be broken. The return timestamp is not stamped, making an incorrect _duration metric.

### OTel-Arrow Version

main

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

sjmsft:
# Change Summary

**Moved `route_ack`/`route_nack` from public inherent methods to a `#[doc(hidden)]` `AckNackRouting` trait.**

Previously, any component could call `effect_handler.route_ack(ack)` directly, bypassing `ConsumerEffectHandlerExtension::notify_ack`/`notify_nack`, the wrappers that stamp `return_time_ns` for pipeline duration metrics. This made it easy to accidentally skip timing instrumentation.

Now, `route_ack`/`route_nack` require importing the `AckNackRouting` trait, which is hidden from rustdoc, making accidental bypass unlikely. The intended API remains `notify_ack`/`notify_nack`.

* Closes #2387 

Executed all otap-df-otap tests
cargo test -p otap-df-otap --lib pdata::test

Added 4 new tests.

**New tests** verify the contract between the two layers:
- `notify_ack_stamps_return_time` / `notify_nack_stamps_return_time` confirm that the public `notify_*` methods stamp `return_time_ns` when pipeline timing interests (`ENTRY_TIMESTAMP`) are present.
- `route_ack_does_not_stamp_return_time` / `route_nack_does_not_stamp_return_time` confirm that the lower-level `route_*` methods leave `return_time_ns` at zero, proving the timing logic lives exclusively in the `ConsumerEffectHandlerExtension` layer.

Yes, this is a minor breaking change for downstream code that called route_ack/route_nack directly on an EffectHandler:

Before: route_ack/route_nack were public inherent methods callable without any trait import.
After: They require use otap_df_engine::AckNackRouting; to resolve, and that trait is #[doc(hidden)], so it won't appear in generated docs.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/pdata.rs
+++ b/rust/otap-dataflow/crates/otap/src/pdata.rs
@@ -16,7 +16,6 @@
 use async_trait::async_trait;
 use otap_df_config::PortName;
 use otap_df_config::{SignalFormat, SignalType};
-use otap_df_engine::_private::AckNackRouting;
 use otap_df_engine::control::{AckMsg, CallData, Frame, NackMsg, RouteData, nanos_since_birth};
 use otap_df_engine::error::{Error, TypedError};
 use otap_df_engine::{
@@ -567,6 +566,7 @@ macro_rules! impl_consumer_ext {
         #[async_trait(?Send)]
         impl ConsumerEffectHandlerExtension<OtapPdata> for $handler {
             async fn notify_ack(&self, mut ack: AckMsg<OtapPdata>) -> Result<(), Error> {
+                use otap_df_engine::_private::AckNackRouting;
                 if ack.accepted.has_timing(Interests::ACKS) {
                     ack.unwind.return_time_ns = nanos_since_birth();
                 }
@@ -574,6 +574,7 @@ macro_rules! impl_consumer_ext {
             }
 
             async fn notify_nack(&self, mut nack: NackMsg<OtapPdata>) -> Result<(), Error> {
+                use otap_df_engine::_private::AckNackRouting;
                 if nack.refused.has_timing(Interests::NACKS) {
                     nack.unwind.return_time_ns = nanos_since_birth();
                 }
@@ -683,7 +684,6 @@ mod test {
 
     use crate::testing::{TestCallData, create_test_pdata, next_ack, next_nack};
     use otap_df_channel::mpsc::Channel as LocalChannel;
-    use otap_df_engine::_private::AckNackRouting;
     use otap_df_engine::ConsumerEffectHandlerExtension;
     use otap_df_engine::control::{
         PipelineCompletionMsg, pipeline_completion_msg_channel, runtime_ctrl_msg_channel,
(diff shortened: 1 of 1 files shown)


jmacd:
﻿Thanks for this! The `#[doc(hidden)]` trait approach is a good start. I think we can make the intent stronger by following the `_private` module pattern already found in crates/telemetry/src/internal_events.rs.

Instead of exporting `AckNackRouting` directly from the crate root, could we put it behind a `#[doc(hidden)] pub mod _private` module? Callers would have to write `otap_df_engine::_private::AckNackRouting`.

lalitb:
nit - The tests only cover `LocalProcessorEffectHandler`. This PR moves `route_ack`/`route_nack` behind `AckNackRouting` for all four handler types - local/shared processor and exporter. Could we add equivalent tests for the other three variants? The impls are identical boilerplate so divergence is unlikely, but having coverage across all four would give more confidence.

lalitb (approved):

sjmsft:
> Thanks for this! The `#[doc(hidden)]` trait approach is a good start. I think we can make the intent stronger by following the `_private` module pattern already found in crates/telemetry/src/internal_events.rs.
> 
> Instead of exporting `AckNackRouting` directly from the crate root, could we put it behind a `#[doc(hidden)] pub mod _private` module? Callers would have to write `otap_df_engine::_private::AckNackRouting`.

thanks, sent out next iteration.

sjmsft:
> nit - The tests only cover `LocalProcessorEffectHandler`. This PR moves `route_ack`/`route_nack` behind `AckNackRouting` for all four handler types - local/shared processor and exporter. Could we add equivalent tests for the other three variants? The impls are identical boilerplate so divergence is unlikely, but having coverage across all four would give more confidence.

thanks, sent out next iteration.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -1687,4 +1694,372 @@ mod test {
             "CONSUMER_METRICS entry frame is skipped; ack routes to subscriber at node 0"
         );
     }
+
```
jmacd, line 1697: The new tests are a little heavy. I would like to remove the "route_nack_does_not_stamp_return_time" cases and likewise "route_ack_does_not..."

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -682,13 +683,19 @@ mod test {
 
     use crate::testing::{TestCallData, create_test_pdata, next_ack, next_nack};
     use otap_df_channel::mpsc::Channel as LocalChannel;
-    use otap_df_engine::control::runtime_ctrl_msg_channel;
+    use otap_df_engine::_private::AckNackRouting;
```
jmacd: Let's remove this line. Instead, add this line inside each of the `notify_ack()` and `notify_nack()` definitions, which are inside a `macro_rules!` around line 560.

jmacd (changes requested):

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -566,13 +566,15 @@ macro_rules! impl_consumer_ext {
         #[async_trait(?Send)]
         impl ConsumerEffectHandlerExtension<OtapPdata> for $handler {
             async fn notify_ack(&self, mut ack: AckMsg<OtapPdata>) -> Result<(), Error> {
+                use otap_df_engine::_private::AckNackRouting;
```
jmacd, line 569: 👍

lalitb (approved):

PR #2490, opened 2026-04-01
fix: race condition in quiver's WAL writer cursor validation

### LINKED ISSUE #2487: [otap-df-quiver] WAL: Race condition between tokio::fs::File::write_all and metadata() in cursor validation causes spurious InvalidConsumerCursor errors
### Pre-filing checklist

### Component(s)

## Summary

The `validate_cursor()` function in quiver's WAL writer can fail with `InvalidConsumerCursor("safe offset beyond wal tail")` due to a race condition between tokio's async file write and a subsequent `metadata()` call. This affects production workloads, not just tests.

## Root Cause

In [`crates/quiver/src/wal/writer.rs`](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/quiver/src/wal/writer.rs), `validate_cursor()` queries the OS file length via:

```rust
let file_len = active_file.file_mut()?.metadata().await?.len();
```

Tokio's `File::poll_write()` returns immediately after spawning a blocking write task, before the actual OS `write()` syscall completes. Meanwhile, `metadata()` bypasses the `File`'s internal state machine: it clones the inner `Arc<StdFile>` and calls `fstat()` on a separate blocking-pool task. If the write task hasn't finished, `fstat()` returns the **pre-write** file length.

This is documented behavior of `tokio::fs::File`, which buffers writes and dispatches them to the blocking pool.

## Failure Path

1. `ingest()` -> `append_bundle()` -> `write_all()` returns, but the last OS write is still in-flight on the blocking pool
2. `maybe_flush()` is skipped (flush interval hasn't elapsed; default 25ms)
3. Segment size threshold exceeded -> `finalize_segment_impl()` calls `persist_cursor()`
4. `validate_cursor()` calls `metadata().await?.len()` which returns the **stale** file length
5. The cursor's `safe_offset` (computed from the post-write WAL position) exceeds the stale length
6. Error: `InvalidConsumerCursor("safe offset beyond wal tail")`

## Impact

Ingestion fails with an error that propagates to the caller. The data is actually written correctly — the validation is a false negative. On restart, WAL replay recovers the data, but the transient error causes unnecessary ingest failures.

### Steps to Reproduce

This is a race condition that has been observed in CI builds ([example](https://github.com/open-telemetry/otel-arrow/actions/runs/23797919480/job/69425650640?pr=2467)) in the `startup_deletes_expired_segments_without_loading` test. It may be more likely to reproduce under the following conditions:

- High ingest throughput (flush interval not yet elapsed between write and cursor persist)
- Slow or contended blocking pool (write task delayed)
- Small segment target sizes (frequent finalization)

### Expected Behavior

`validate_cursor()` always correctly validates the cursor relative to WAL file lengths under normal operating conditions and is not subject to timing behavior in `tokio::fs::File`.

### Actual Behavior

```
thread 'engine::tests::startup_deletes_expired_segments_without_loading' panicked at crates/quiver/src/engine.rs:4499:46:
ingest: Wal { source: InvalidConsumerCursor("safe offset beyond wal tail") }
```

### OTel-Arrow Version

4bad53722a4bb7d9111bf2302ec8651c43ad57ca (main)

### Environment

n/a

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

AaronRM:
# Change Summary

Fix a race condition in the WAL cursor validation where `tokio::fs::File::metadata()` can return a stale file length after `write_all()`, causing spurious `InvalidConsumerCursor("safe offset beyond wal tail")` errors during ingestion.

Tokio's `File::poll_write()` returns `Ready` before the blocking-pool write task completes the OS `write()` syscall. `metadata()` bypasses the File's state machine and can observe the pre-write length. The fix replaces the `metadata().await?.len()` call with the in-memory tracked `active_file.len()`, which is updated synchronously after each write.

Since `validate_cursor()` no longer performs I/O, it is changed from `async fn` to `fn`, and the `&mut ActiveWalFile` parameters on `validate_cursor` and `ensure_entry_boundary` are relaxed to `&ActiveWalFile`.

* Closes #2487

- The existing `startup_deletes_expired_segments_without_loading` test exercises the affected code path. It was what originally surfaced the bug in CI.

No. This fixes an internal race condition that could cause transient ingest failures under high throughput or blocking-pool contention. Behavior is otherwise unchanged.

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/wal/writer.rs
+++ b/rust/otap-dataflow/crates/quiver/src/wal/writer.rs
@@ -661,7 +661,7 @@ impl WalWriter {
     /// Call this after downstream has confirmed durability (e.g., segment flush).
     pub(crate) async fn persist_cursor(&mut self, cursor: &WalConsumerCursor) -> WalResult<()> {
         self.coordinator
-            .persist_cursor(&mut self.active_file, cursor)
+            .persist_cursor(&self.active_file, cursor)
             .await
     }
 
@@ -1418,7 +1418,7 @@ impl WalCoordinator {
 
     async fn persist_cursor(
         &mut self,
-        active_file: &mut ActiveWalFile,
+        active_file: &ActiveWalFile,
         cursor: &WalConsumerCursor,
     ) -> WalResult<()> {
         self.validate_cursor(active_file, cursor)?;


rust/otap-dataflow/crates/quiver/src/wal/writer.rs
```diff
@@ -1421,13 +1421,13 @@ impl WalCoordinator {
         active_file: &mut ActiveWalFile,
```
ThomsonTan: nit: `mut` on `active_file` could also be removed.

AaronRM: Fixed. Thanks!

PR #2493, opened 2026-04-01
Revert "fix: Temporarily disable the nightly otap-filter-otap Go collector scenario"

cijothomas:
Reverts open-telemetry/otel-arrow#2396 as we bumped to new Collector image with the fix https://github.com/open-telemetry/otel-arrow/pull/2480


cijothomas:
@JakeDern PTAL

PR #2495, opened 2026-04-01
Add syslog benches for other core functions

kamilon:
# Change Summary

Add benches for a few more syslog core logic functions. This PR is being added to establish a baseline for some incoming changes.

No issue filed.

cargo build and test commands.

No.

### DIFF
--- a/rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
+++ b/rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
@@ -12,22 +12,22 @@
     RAM: 32.0 GB
     | Test                                        | Average time |
     |---------------------------------------------|--------------|
-    | parser_comparison/rfc3164                   | 17.947 ns    |
-    | parser_comparison/rfc5424                   | 35.897 ns    |
-    | parser_comparison/cef                       | 31.826 ns    |
-    | parse_auto_detect/rfc3164                   | 39.528 ns    |
-    | parse_auto_detect/rfc5424                   | 41.345 ns    |
-    | parse_auto_detect/cef                       | 36.844 ns    |
-    | parse_auto_detect/cef_rfc3164               | 115.30 ns    |
-    | parse_auto_detect/cef_rfc5424               | 66.394 ns    |
-    | timestamp_extraction/rfc3164                | 539.92 ns    |
-    | timestamp_extraction/rfc5424                | 22.963 ns    |
-    | cef_extensions/one_extension                | 20.155 ns    |
-    | cef_extensions/ten_extensions               | 212.31 ns    |
-    | cef_extensions/ten_extensions_with_escape   | 190.64 ns    |
-    | rfc3164_arrow_batch_100_msgs                | 95.930 µs    |
-    | rfc5424_arrow_batch_100_msgs                | 47.484 µs    |
-    | cef_arrow_batch_100_msgs                    | 43.238 µs    |
+    | parser_comparison/rfc3164                   | 17.409 ns    |
+    | parser_comparison/rfc5424                   | 35.704 ns    |
+    | parser_comparison/cef                       | 31.316 ns    |
+    | parse_auto_detect/rfc3164                   | 40.016 ns    |
+    | parse_auto_detect/rfc5424                   | 42.375 ns    |
+    | parse_auto_detect/cef                       | 37.960 ns    |
+    | parse_auto_detect/cef_with_rfc3164          | 117.88 ns    |
+    | parse_auto_detect/cef_with_rfc5424          | 67.939 ns    |
+    | timestamp_extraction/rfc3164                | 549.32 ns    |
+    | timestamp_extraction/rfc5424                | 23.182 ns    |
+    | cef_extensions/one_extension                | 20.341 ns    |
+    | cef_extensions/ten_extensions               | 191.42 ns    |
+    | cef_extensions/ten_extensions_with_escape   | 215.20 ns    |
+    | arrow_batch_creation/rfc3164_arrow_batch_100_msgs | 96.281 µs    |
+    | arrow_batch_creation/rfc5424_arrow_batch_100_msgs | 47.681 µs    |
+    | arrow_batch_creation/cef_arrow_batch_100_msgs     | 43.565 µs    |
 */
 
 #![allow(missing_docs)]
@@ -60,9 +60,9 @@ static CEF_WITH_RFC3164_MSG: &[u8] =
 static CEF_WITH_RFC5424_MSG: &[u8] =
     b"<34>1 2003-10-11T22:14:15.003Z mymachine.example.com app - - - CEF:0|Security|threatmanager|1.0|100|worm stopped|10|src=10.0.0.1";
 static CEF_MSG_TEN_EXT: &[u8] =
-    b"CEF:0|Security|threatmanager|1.0|100|worm stopped|10|src=10.0.0.1 dst=2.1.2.2 spt=1232 dpt=1233 proto=TCP act=blocked app=HTTP rt=1234567890 msg=Worm\\=stopped cn1Label=score";
-static CEF_MSG_TEN_EXT_WITH_ESCAPE: &[u8] =
     b"CEF:0|Security|threatmanager|1.0|100|worm stopped|10|src=10.0.0.1 dst=2.1.2.2 spt=1232 dpt=1233 proto=TCP act=blocked app=HTTP rt=1234567890 msg=Worm stopped cn1Label=score";
+static CEF_MSG_TEN_EXT_WITH_ESCAPE: &[u8] =
+    b"CEF:0|Security|threatmanager|1.0|100|worm stopped|10|src=10.0.0.1 dst=2.1.2.2 spt=1232 dpt=1233 proto=TCP act=blocked app=HTTP rt=1234567890 msg=Worm\\=stopped cn1Label=score";
 
 /// Benchmark comparing all three parsers
 fn bench_parser_comparison(c: &mut Criterion) {


rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
```diff
@@ -27,6 +28,12 @@ static RFC5424_MSG: &[u8] =
     b"<34>1 2003-10-11T22:14:15.003Z mymachine.example.com su - ID47 - Message";
 static CEF_MSG: &[u8] =
     b"CEF:0|Security|threatmanager|1.0|100|worm successfully stopped|10|src=10.0.0.1";
+static CEF_WITH_RFC3164_MSG: &[u8] =
```
cijothomas, line 58: not a established practice in this repo, but if you don't mind, can you leave a code comment with the actual numbers similar to how we do in some other repos. Eg: https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-appender-tracing/benches/logs.rs#L1-L24

(this is just for a quick reference)

kamilon: How do you feel about that being in PRs vs checked in to code?

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
```diff
@@ -62,6 +69,82 @@ fn bench_parser_comparison(c: &mut Criterion) {
...
+}
+
+/// Benchmark timestamp extraction
+fn bench_timestamp_extraction(c: &mut Criterion) {
+    let mut group = c.benchmark_group("timestamp_extraction");
+
+    let rfc3164_parsed = parse(RFC3164_MSG).expect("parse RFC3164");
+    let _ = group.bench_function("rfc3164", |b| {
+        b.iter(|| black_box(rfc3164_parsed.timestamp()))
+    });
```
lalitb, line 140: The closure here calls `timestamp` on an immutable precomputed value and only black-box the result. If compiler can prove the call is pure for that input, it can compute it once and reuse the constant. A safer pattern would be:

```rust
b.iter(|| black_box(black_box(&rfc3164_parsed).timestamp()))

```

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
```diff
@@ -62,6 +69,82 @@ fn bench_parser_comparison(c: &mut Criterion) {
...
+
+    let rfc5424_parsed = parse(RFC5424_MSG).expect("parse RFC5424");
+    let _ = group.bench_function("rfc5424", |b| {
+        b.iter(|| black_box(rfc5424_parsed.timestamp()))
+    });
```
lalitb, line 145: same [comment](https://github.com/open-telemetry/otel-arrow/pull/2495/changes#r3019970745)

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
kamilon: I ask because I plan on doing that in each PR. I don't personally see great value in these being checked in as code artifacts since:

1. We all have different machines so we can't directly compare 1:1
2. These should be point in time problems - we find and fix the something and never need the comparison again
3. It's metadata for the PR and not the code

I can be convinced though.

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
kamilon: Due to the way this function works, this happens to not affect the bench. Great catch! Made the change to both calls.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/cef.rs
```diff
@@ -17,7 +17,7 @@ pub struct CefMessage<'a> {
 
 impl CefMessage<'_> {
     /// Parse and iterate over the extensions as key-value pairs
-    pub(super) const fn parse_extensions(&self) -> CefExtensionsIter<'_> {
+    pub const fn parse_extensions(&self) -> CefExtensionsIter<'_> {
```
lalitb: Changing these parser helpers from `pub(super)` / `pub(crate)` to `pub` exposes internal parser APIs from the published `otap-df-core-nodes` crate and makes later parser refactors semver-sensitive. Could we avoid exposing the raw internals here, either by adding a small benchmark-only wrapper, moving this benchmark into the `otap-df-core-nodes` package, or benchmarking through an existing higher-level public entrypoint instead?

lalitb: with benchmark-only wrapper, I mean something like:
```rust
// in `otap-df-core-nodes`

#[cfg(feature = "bench-internals")]
#[doc(hidden)]
pub fn bench_parse_xxx(input: &[u8]) -> Result<BenchResult, ParseError> {
    // call internal parser/helper here
    // return only the minimal data the benchmark needs
}
```

kamilon: Very much did not think about the ramifications of changing the visiblity of these. The best I could come up with (with the help of AI) is to feature flag a struct that exposes these. Only plumbed into bench and hidden from docs.

Does this look like a reasonable solution?

kamilon: Your additional comment didn't show up until I posted mine. Looks like it's the same solution!

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
cijothomas: Check the example reference I shared above. Its not for comparison (and the comment already show hardware on which it was run).

This is just to give a quick sense of how much time things take without having to run them yourself.

(overall perf results from the repo is automatically published to GitHub pages, so this is not strictly required, but something I personally I find valuable)

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
```diff
@@ -62,6 +69,82 @@ fn bench_parser_comparison(c: &mut Criterion) {
...
+
+    group.finish();
+}
+
+/// Benchmark CEF extension iteration
+fn bench_cef_extensions(c: &mut Criterion) {
+    let mut group = c.benchmark_group("cef_extensions");
+
+    let cef_parsed = parse_cef(CEF_MSG).expect("parse CEF");
+    let _ = group.bench_function("three_extensions", |b| {
```
lalitb: nit - the benchmark name `three_extensions` is confusing - it's using `CEF_MSG` which contains only single extension.

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
```diff
@@ -27,6 +28,12 @@ static RFC5424_MSG: &[u8] =
     b"<34>1 2003-10-11T22:14:15.003Z mymachine.example.com su - ID47 - Message";
 static CEF_MSG: &[u8] =
     b"CEF:0|Security|threatmanager|1.0|100|worm successfully stopped|10|src=10.0.0.1";
+static CEF_WITH_RFC3164_MSG: &[u8] =
+    b"<34>Oct 11 22:14:15 mymachine CEF:0|Security|threatmanager|1.0|100|worm stopped|10|src=10.0.0.1";
+static CEF_WITH_RFC5424_MSG: &[u8] =
+    b"<34>1 2003-10-11T22:14:15.003Z mymachine.example.com app - - - CEF:0|Security|threatmanager|1.0|100|worm stopped|10|src=10.0.0.1";
+static CEF_MSG_TEN_EXT: &[u8] =
+    b"CEF:0|Security|threatmanager|1.0|100|worm stopped|10|src=10.0.0.1 dst=2.1.2.2 spt=1232 dpt=1233 proto=TCP act=blocked app=HTTP rt=1234567890 msg=Worm\\=stopped cn1=42 cn1Label=score";
```
lalitb: nit - sorry for nit picking, but this seems to have 11 exrtensions : )

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/mod.rs
```diff
@@ -155,6 +155,45 @@ pub(super) fn parse_priority(input: &[u8]) -> Result<(Priority, &[u8]), ParseErr
     Ok((Priority { facility, severity }, &input[end + 1..]))
 }
 
+/// Benchmark support — exposes internal parser helpers for benchmarking only.
+///
+/// This module is **not** part of the public API and is gated behind the `bench`
+/// Cargo feature. Items here may change or be removed without notice.
+#[cfg(feature = "bench")]
```
cijothomas, line 162: this is a neat trick to avoid public api exposure!

cijothomas:
https://open-telemetry.github.io/otel-arrow/benchmarks/nightly/syslog/ shows our nightly runs.

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
kamilon: Added.

utpilla: @kamilon It looks like you have added the benchmark results to the `src/` files. Could you please move them to file(s) under `benches/`?

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
utpilla: `CEF_MSG_TEN_EXT` includes `\=` in the `msg` value, which mixes escape-handling cost into a general multi-extension throughput benchmark. In practice, most devices emit plain `key=value` pairs without escape sequences, and our optimization focus is on that common case. Consider either:

1. Removing the escape from this input so it measures the common case, or
2. Having two variants: `cef_ten_ext` (no escapes) and `cef_ten_ext_with_escapes`

That way improvements to the common path vs. the escape path can be measured independently.

kamilon: Updated to drop down to 10.

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
kamilon: Fixed.

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
kamilon: Added the second variant. I think seeing both side by side is useful.

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
kamilon: This is done now.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/mod.rs
```diff
@@ -155,6 +155,45 @@ pub(super) fn parse_priority(input: &[u8]) -> Result<(Priority, &[u8]), ParseErr
...
+    impl<'a> CefExtensionsIter<'a> {
+        /// Returns the next key-value extension pair, or `None` when exhausted.
+        pub fn next_extension(&mut self) -> Option<(&[u8], &[u8])> {
+            self.0.next_extension()
+        }
+    }
+
+    /// Start iterating over CEF extensions.
+    #[must_use]
+    pub fn parse_extensions<'a>(msg: &'a CefMessage<'a>) -> CefExtensionsIter<'a> {
```
utpilla: You could mark this method `const`.

```suggestion
    pub const fn parse_extensions<'a>(msg: &'a CefMessage<'a>) -> CefExtensionsIter<'a> {
```

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
```diff
@@ -3,10 +3,38 @@
...
+    | parse_auto_detect/rfc3164                   | 39.528 ns    |
+    | parse_auto_detect/rfc5424                   | 41.345 ns    |
+    | parse_auto_detect/cef                       | 36.844 ns    |
+    | parse_auto_detect/cef_rfc3164               | 115.30 ns    |
+    | parse_auto_detect/cef_rfc5424               | 66.394 ns    |
+    | timestamp_extraction/rfc3164                | 539.92 ns    |
+    | timestamp_extraction/rfc5424                | 22.963 ns    |
+    | cef_extensions/one_extension                | 20.155 ns    |
+    | cef_extensions/ten_extensions               | 212.31 ns    |
+    | cef_extensions/ten_extensions_with_escape   | 190.64 ns    |
```
utpilla: The variable names are swapped. `CEF_MSG_TEN_EXT` contains `Worm\\=stopped` (has escape), while `CEF_MSG_TEN_EXT_WITH_ESCAPE` contains `Worm stopped` (no escape).

With extensions benchmarks should take longer.

jmacd:
> https://open-telemetry.github.io/otel-arrow/benchmarks/nightly/syslog/

I'm confused by the recent changes here, but that can be a different conversation.

PR #2501, opened 2026-04-01
Support filtering attributes with case insensitive key matches

### LINKED ISSUE #2496: Columnar Query Engine ability to do case insensitive equals on attribute keys
In Otel, it's advised that users keep their attribute keys lowercase:
https://opentelemetry.io/docs/specs/semconv/general/naming/#general-naming-considerations

However there may be cases where users don't follow these conventions. The modern world is a cruel and terrible place, replete with untold horrors, viz., users who won't or can't control the letter case of their attribute names.

We should support a mode in the columnar query engine where users who've put themselves in such a situation don't need to perform gymnastics to get their queries to do what they want.

E.g.  In this mode,  a query like this `logs | where attributes["x"] =="y"` would select both logs with attributes `{"x": "y"}` and `{"X": "y"}`. The same principle should apply in the case where the logical expression is used in the condition of the if/else operator.

We should provide a similar solution for nested attribute pipelines. For example, we may consider introducing `=~` as the case insensitive equals operator, similar to what is in KQL.
```
logs | apply attributes {
  if (key =~ "x") {
   ...
  }
}
```

albertlockett:
# Change Summary

There might be some users who do not know a-priori the letter case of their attribute names, but they'd like to do transformations to them using the transform processor anyway. We can imagine a use case where a user would like to drop a log record or attribute by matching the attribute key, but we'd like the match to be case insensitive.

This PR adds support for this by making two changes:

**1. Case insensitive value match added to the columnar query engine & OPL parser**

The KQL parser/record set engine already supports using the `=~` binary operator for case insensitive equality. This PR extends the support for this syntax to OPL and adds the capability to perform case insensitive string equality to the columnar query engine. 

This support enables users to, for example, drop attributes by doing a case insensitive match on the attribute key:
```kql
logs | apply attributes { where not (key =~ "sensitive") }
```

OPL parser also actually had the case-insensitive on `EqualToLogicalExpression` mixed up, so this PR corrects that.

**2. Case insensitive matching of attribute key when filtering**

Here there's actually no new OPL/KQL syntax. Instead there's an option that we can set on the transform processor configuration called `filter_attribute_keys_case_sensitive`, which populates a field on the new `PipelineOptions` struct, used by the columnar query engine's `Pipeline` to control how the queries are planned and executed.

For exmaple, if this flag were set to `false` and we had a query like the following:
```kql
logs | where attributes["x"] = "y"
 ```
If this were set to `false`, logs with attributes `{"x": "y"}` as well as those with attributes `{"X": "y"}` would pass the filter.

**Case Insensitive Match Implementation: Temporary Solution**

In both these cases, we actually plan a datafusion [`BinaryExpr`](https://docs.rs/datafusion/latest/datafusion/logical_expr/struct.BinaryExpr.html) using an operator the case insensitive Like match [`ILikeMatch`](https://docs.rs/datafusion/latest/datafusion/logical_expr/enum.Operator.html#variant.ILikeMatch). In most cases, this actually performs pretty well. I didn't see any regression in our filter benchmarks if I changed them to use this case insensitive match. 

The reason this performs well is because internally in [`arrow_string::predicate`](https://github.com/apache/arrow-rs/blob/1f07e54c1209a512f3ce15f4e576f0ca5d1c8d97/arrow-string/src/predicate.rs#L67-L114) this will get compared using `eq_ignore_ascii_case` _if there are no special like characters_ like `%`, `_`, or escape sequences. If these characters are present, the comparison gets done using case insensitive regex match which is a lot slower! It does this even if the characters are escaped (which incidently this PR does to avoid unexpected match behaviour.)

Eventually we'll probably want to create a special kernel to do this case insensitive equality directly, however I feel that it would be a lot easier to implement this in arrow_string versus crafting something custom in otel-arrow. I'll try to tackle this as a followup.

* Closes #2496

Unit tests

Yes 
-  users now have the `filter_attribute_keys_case_sensitive` flag available as transform processor config
- the `=~` binary operator is also now supported in OPL
- case insensitive equality is actually supported now (before, we ignored this and treated all matches as case sensitive).

### DIFF
--- a/rust/otap-dataflow/crates/opl/src/opl.pest
+++ b/rust/otap-dataflow/crates/opl/src/opl.pest
@@ -118,12 +118,12 @@ multiplicative_expression = {
     unary_expression ~ (multiplicative_op ~ unary_expression)*
 }
 
-rel_op_eq  = @{ "==" }
-rel_op_neq = @{ "!=" }
-rel_op_gte = @{ ">=" }
-rel_op_gt  = @{ ">" }
-rel_op_lte = @{ "<=" }
-rel_op_lt  = @{ "<" }
+rel_op_eq                  = @{ "==" }
+rel_op_neq                 = @{ "!=" }
+rel_op_gte                 = @{ ">=" }
+rel_op_gt                  = @{ ">" }
+rel_op_lte                 = @{ "<=" }
+rel_op_lt                  = @{ "<" }
 rel_op_eq_case_insensitive = @{ "=~" }
 
 rel_op = _{

### CODE REVIEW

File: rust/otap-dataflow/crates/query-engine/src/pipeline.rs [hunk 1]
```diff
@@ -294,7 +322,9 @@ impl Pipeline {
         // lazily plan the pipeline if have not already done so
         if self.planned_pipeline.is_none() {
             let session_ctx = Self::create_session_context();
-            let planner = PipelinePlanner::new();
+            let planner = PipelinePlanner::new().with_filter_attribute_keys_case_sensitive(
+                self.options.filter_attribute_keys_case_sensitive,
```
lquerel, line 326: Is this option also propagated to the `apply_attributes { ... }` stuff?

albertlockett: No, it does not. My intention with this flag was more for filters that have expressions like `attributes["x"]`. Currently if you set this flag to false, and then do something like `logs | apply attributes { where key == "x" }`, that would be a case sensitive match. There is still the option to do `where key =~ "x"` instead inside the `apply attributes` block for a case insensitive match. Happy to revisit the behaviour in a followup if you feel it should propagate?

PR #2502, opened 2026-04-01
feat(temporal_reaggregation_processor): Support for OTLP views, id overflow detection, and general error recovery paths 

JakeDern:
# Change Summary

This is the next major milestone for the temporal reaggregation processor which adds OTLP support and graceful id overflow handling.

The major achievement here, and the reason there is so much churn, is that I needed to pick some strategy that lets us unwind in case we're in the middle of processing some view and hit a limit.

The thing that I settled on was checkpointing the current progress and if we hit some kind of limit halfway through, to simply seal the current batch, discard the partially written data based on the checkpoint, and reprocess with a fresh output batch.

Rather than trying to unwind, I also considered trying to run a pre-flight check to see if everything in the current batch will fit into the output batch. However, it is basically as much work to do that as it is to do the entire aggregation. The reason is that we can't know if everything is going to fit in the output batch until we identify every resource, scope, metric, and stream and see how many unique things we have that need to be aggregated. That is the vast majority of the work required in aggregating and we would need intermediate storage to re-use any of that work. Instead, I chose to optimistically try to pack into the current batch and backtrack if we can't.

* Part of #2422

Unit.

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+## [0.47.0](https://github.com/open-telemetry/otel-arrow/releases/tag/v0.47.0) - 2026-04-02
+
 - Upgrade various Go dependencies.
   [#1767](https://github.com/open-telemetry/otel-arrow/pull/1767),
   [#1802](https://github.com/open-telemetry/otel-arrow/pull/1802),
--- a/collector/cmd/otelarrowcol/main.go
+++ b/collector/cmd/otelarrowcol/main.go
@@ -20,7 +20,7 @@ func main() {
 	info := component.BuildInfo{
 		Command:     "otelarrowcol",
 		Description: "OpenTelemetry Protocol with Apache Arrow development collector, for testing and evaluation",
-		Version:     "0.46.0",
+		Version:     "0.47.0",
 	}
 
 	set := otelcol.CollectorSettings{
--- a/collector/otelarrowcol-build.yaml
+++ b/collector/otelarrowcol-build.yaml
@@ -17,7 +17,7 @@ dist:
 
   # Note: this version number is replaced to match the current release using `sed`
   # during the release process, see ../../../RELEASING.md.
-  version: 0.46.0
+  version: 0.47.0
 
   # Project-internal use: Directory path required for the `make
   # genotelarrowcol`, which the Dockerfile also recognizes.
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/Telemetry.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/Telemetry.md
@@ -1,0 +1,53 @@
+# Syslog CEF Receiver Telemetry
+
+This document lists telemetry emitted directly by the
+`syslog_cef_receiver` component. It includes metric instruments registered
+by the component and log events emitted via `otel_*` log macros.
+
+## Metrics
+
+Metrics are registered under the metric set name `syslog_cef.receiver.metrics`.
+
+| Metric name | Type | Unit | Description | Produced in file |
+| --- | --- | --- | --- | --- |
+| `syslog_cef.receiver.metrics.received_logs_total` | Counter | `{item}` | Total number of log records observed at the socket before parsing. | `crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs` |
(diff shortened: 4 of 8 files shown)


rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/identity.rs
```diff
@@ -304,24 +282,21 @@ impl AttributeHash {
...
-        buf.entries.clear();
+    /// The provided [`HashBuffer`] is reused across calls to avoid repeated
+    /// heap allocations for the hash input bytes. A temporary `Vec` is
+    /// allocated internally for sorting the attributes by key.
+    pub fn compute<A: AttributeView>(buf: &mut HashBuffer, attrs: impl Iterator<Item = A>) -> Self {
         buf.buf.clear();
-        buf.entries.extend(attrs);
 
-        if buf.entries.is_empty() {
+        let mut entries: Vec<A> = attrs.collect();
```
JakeDern, line 287: Sadly, I had to sacrifice being able to reuse the sorting buffer for now. There were major issues with generics here and I would have to duplicate the entire main view processing function so that I could do the lifetime shuffling gymnastics required to turn this buffer into 'static for reuse.

We still reuse the encoding buffer though, which I think is way more important.

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
```diff
@@ -95,11 +182,40 @@ impl local::Processor<OtapPdata> for TemporalReaggregationProcessor {
...
+                        let Err(e) = self.process_metrics_pdata(&pdata) else {
+                            return Ok(());
+                        };
+
+                        match e {
+                            // Id overflows are fine. In this case we seal the current outbound
+                            // batch and then feed the data back into a next one. We do this to
+                            // prevent complex ack/nack scenarios where an input batch has
+                            // representation in multiple output batches.
+                            ProcessPdataError::IdOverflow => {
```
lquerel, line 198: If I understand well this, overflow can happen after part of the current input has already mutated in-memory aggregation state. Handling this safely requires an atomic boundary for the input batch: either rollback all mutations performed after the checkpoint, or stage updates separately and commit only once the whole
input fits (you might think about a better approach). Simply retrying the whole input is only correct if no partial mutations from the failed attempt can leak into the flushed output.

JakeDern: Correct - The way this currently works (and maybe I could write the code in a different way so this is clearer) is that the calls to `flush_at` will clear everything in the current `BatchState` which includes anything we accumulated from processing the failed data. So, we're starting with fresh aggregation state on the second `process_metrics_pdata` call.

To summarize the current procedure is:

1. Checkpoint the current builder positions
2. Process the current batch and append to the builders/update the hashmaps
3. If we hit a failure then finish the current batch, truncate the output to the known good position based on the checkpoint, and clear all the pending aggregation state
4. Retry processing the batch again

JakeDern: I'll also emphasize again the two main drawbacks you probably noticed:

1. We may have some wasted memory in the output otap batch (whatever records were appended before we hit failure)
2. We have to do reprocessing of the same data in some cases

The way to avoid this would be via a pre-flight check to see if what just came in will fit into the current aggregation.  To do the preflight check we would need to compute the Identity for every resource/scope/metric/stream in the and see if we can accommodate them all.

That's very expensive so to avoid wasting that work then we also would want to try to buffer those Identity computations from the pre-flight check so that if it passes, we can re-use all the attribute hashes. It might be reasonable to do this, but it comes at a cost of some complexity and a higher working memory budget for the processor.

albertlockett: I think it's probably it's OK to be a little bit inefficient in the ID overflow case.

We'd only run into an `IdOverflow` in the case when there's such a variety of different time series that we can't aggregate them into a single batch without overflowing the IDs, right?

Having tons of distinct time series could mean the user is doing something inefficient with their telemetry data anyway. For users who really want/need to be in this situation, they could restructure the pipeline, e.,g. maybe put some kind of router processor multiple temporal aggregation processors to group metrics together

JakeDern: >We'd only run into an IdOverflow in the case when there's such a variety of different time series that we can't aggregate them into a single batch without overflowing the IDs, right?

Yes, pretty much.

Also, something I'll bring attention to is the case where someone sends a massive OTLP batch e.g. one that has 128,000 distinct metrics. We wouldn't be able to represent it in a single OTAP batch due to id constraints, so I just nack it.

PR #2504, opened 2026-04-01
fix(geneva_exporter): per-signal metrics, accurate batch tracking, and per-upload latency

lalitb:
Addresses review comments from #2262  on the Geneva exporter's metrics instrumentation.
 
 ### Changes
 
 **1. Fix `try_for_each_concurrent` short-circuit (inflated failure counts)**
 
 Replaced `try_for_each_concurrent` with `buffer_unordered` + collect. All batches are now attempted regardless of individual failures. Successes and failures are counted accurately per-batch using `EncodedBatch::row_count`, rather than attributing all records to either success or failure.
 
 **2. Fix `upload_duration` recording aggregate wall-clock time**
 
 Upload latency is now recorded per individual batch upload, not once for the entire concurrent block. Each upload produces its own timing sample, giving accurate min/max/sum/count statistics.
 
 **3. Rename `unsupported_signals` to `metrics_payloads_dropped`**
 
 Metrics is the only unsupported signal type, so the generic name was misleading.
 
 **4. Split shared counters into per-signal metrics**
 
 All upload/failure/timing counters are now split into `log_*` and `trace_*` variants (e.g., `log_batches_uploaded`, `trace_upload_duration`). This lets operators identify which signal type is failing or experiencing latency. Signal-agnostic counters (`empty_payloads_skipped`, `conversion_errors`, `metrics_payloads_dropped`) remain shared.
 
 ### Known limitation (documented as TODO)
 
 When a payload is split into multiple Geneva batches and some fail, the entire payload is still NACKed and retried - potentially duplicating already-uploaded batches. Geneva assigns a fresh UUID per upload, so there is no  server-side dedup. The Azure Monitor exporter solves this with per-message batch tracking and deferred ACK/NACK; a similar approach should be adopted here in a follow-up.

### DIFF
--- a/rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
@@ -402,7 +402,9 @@ impl GenevaExporter {
                             self.metrics.log_upload_success_duration.record(duration_ms);
                         }
                         SignalType::Traces => {
-                            self.metrics.trace_upload_success_duration.record(duration_ms);
+                            self.metrics
+                                .trace_upload_success_duration
+                                .record(duration_ms);
                         }
                         _ => {}
                     }
@@ -416,7 +418,9 @@ impl GenevaExporter {
                             self.metrics.log_upload_failed_duration.record(duration_ms);
                         }
                         SignalType::Traces => {
-                            self.metrics.trace_upload_failed_duration.record(duration_ms);
+                            self.metrics
+                                .trace_upload_failed_duration
+                                .record(duration_ms);
                         }
                         _ => {}
                     }


rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -288,47 +325,112 @@ impl GenevaExporter {
...
+                            .upload_batch(batch)
+                            .await
+                            .map_err(|e| format!("Failed to upload {:?} batch: {e}", signal_type));
+                        let duration_ms = start.elapsed().as_secs_f64() * 1000.0;
+                        (result, duration_ms, batch_size, row_count)
+                    }
+                })
+                .buffer_unordered(max_concurrent)
+                .collect()
+                .await;
```
utpilla: We should be able to avoid the `vec` allocation here:

```rust
let mut stream = futures::stream::iter(batches.iter())
    .map(|batch| {
        let batch_size = batch.data.len() as u64;
        let row_count = batch.row_count as u64;
        async move {
            let start = Instant::now();
            let result = client
                .upload_batch(batch)
                .await
                .map_err(|e| format!("Failed to upload {:?} batch: {e}", signal_type));
            let duration_ms = start.elapsed().as_secs_f64() * 1000.0;
            (result, duration_ms, batch_size, row_count)
        }
    })
    .buffer_unordered(max_concurrent);

while let Some((result, duration_ms, batch_size, row_count)) = stream.next().await {
    ...
}
```

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -288,47 +325,112 @@ impl GenevaExporter {
...
+                .map(|batch| {
+                    let batch_size = batch.data.len() as u64;
+                    let row_count = batch.row_count as u64;
+                    async move {
+                        let start = Instant::now();
+                        let result = client
+                            .upload_batch(batch)
+                            .await
+                            .map_err(|e| format!("Failed to upload {:?} batch: {e}", signal_type));
+                        let duration_ms = start.elapsed().as_secs_f64() * 1000.0;
```
utpilla: `upload_duration` currently records latency for both successful and failed uploads. A network timeout (e.g., 30-60s) on a failed batch would skew the Mmsc min/max/sum and make it hard to gauge normal upload performance.

We could either record `upload_duration` only on success or split it into two metrics (e.g., `log_upload_duration` for successes and `log_upload_error_duration` for failures) so they don't pollute each other. 

I would favor splitting the metric based on the outcome. Here is a similar discussion: https://github.com/open-telemetry/otel-arrow/pull/2211#discussion_r2942931992

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -288,47 +325,112 @@ impl GenevaExporter {
...
+    /// # Partial-success limitation
+    ///
+    /// TODO: When some batches succeed and others fail, this method returns
+    /// `Err` and the caller NACKs the entire payload. The retry processor then
+    /// resends the whole payload, re-uploading already-successful batches and
+    /// causing duplicates (Geneva assigns a fresh UUID per upload, so there is
+    /// no server-side dedup). The Azure Monitor exporter solves this with
+    /// per-message batch tracking and deferred ACK/NACK (see
+    /// `azure_monitor_exporter/state.rs`). A similar approach should be
+    /// adopted here.
```
utpilla: This TODO references the Azure Monitor exporter's approach, but that pattern doesn't help here. Azure Monitor benefits from it because it mixes records from multiple messages into shared upload batches (via `GzipBatcher`). Geneva doesn't do cross-message batching: one `OtapPdata` maps to `N` batches with no sharing, so when any batch fails the whole message must be NACKed regardless.

As discussed inhttps://github.com/open-telemetry/otel-arrow/pull/1653#discussion_r2778701475, the real fix requires engine-level support for per-batch retry tracking (partial ACK/NACK or exporter-attached retry context on `OtapPdata`).

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
lalitb: Good point. done.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
lalitb: Thanks, done.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
lalitb: Thanks, fixed the TODO.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -152,41 +151,87 @@ pub enum AuthConfig {
...
-    pub bytes_uploaded: Counter<u64>,
+    pub log_bytes_uploaded: Counter<u64>,
+
+    /// Per-upload latency for successful log batches in milliseconds (min/max/sum/count).
+    #[metric(name = "log_upload.success.duration", unit = "ms")]
+    pub log_upload_duration: Mmsc,
+
+    /// Per-upload latency for failed log batches in milliseconds (min/max/sum/count).
+    #[metric(name = "log_upload.failed.duration", unit = "ms")]
+    pub log_upload_error_duration: Mmsc,
```
utpilla: The duration metrics use `#[metric(name = "log_upload.success.duration")]` with dots, while all other metrics in this struct derive their name from the field name (underscores). For consistency, could we drop the `name` override and renaming the fields to match, e.g.:

```rust
#[metric(unit = "ms")]
pub log_upload_success_duration: Mmsc,

#[metric(unit = "ms")]
pub log_upload_failed_duration: Mmsc,
```

This keeps all metric names using the same convention (`otap.exporter.geneva.log_upload_success_duration` vs `otap.exporter.geneva.log_upload.success.duration`).

jmacd: I like this suggestion!

jmacd:
In the long term, I think we should expose an "exporterhelper" interface to make it easy for any exporter to expose common metrics like you've done. I would not block this PR for such an idea, though.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
lalitb: Good point, this is done now.

lalitb:
> In the long term, I think we should expose an "exporterhelper" interface to make it easy for any exporter to expose common metrics like you've done. I would not block this PR for such an idea, though.

Good point, created #2522 to track.

Issue #2462, opened 2026-03-30
Move startup helpers into library crates for external embedding


lalitb:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

## Summary
 
The engine crates are already usable as libraries - `Controller`, `OtelDataflowSpec`, `PipelineFactory`, and node registration via `linkme` all work from external binaries. That's great.
 
 The gap is that several practical startup functions live only in `src/main.rs` and can't be reused without copying them:
 
 - **`validate_engine_components()`** - checks that every node URN in the config maps to a registered component and runs per-component config validation. This is the kind of thing every 
embedding binary needs, and it only uses public APIs (`PipelineFactory` getters, `NodeKind`, factory entry `validate_config`).
 
 - **`apply_cli_overrides()`** - applies core allocation and admin bind overrides to an `OtelDataflowSpec`. Uses only public types (`CoreAllocation`, `HttpAdminSettings`, `ResourcesPolicy`).
 
 - **`system_info()`** - prints registered components and system info. Useful for diagnostics in any distribution.
 
 These are small, stable functions with no private dependencies. Moving them into a library module (e.g. `otap_df_controller::startup` or a thin `otap-df-cli` crate) would let custom binaries share them without copying code that may drift across releases.
 
 This would make the engine meaningfully easier to embed in custom distributions - the same pattern used by projects like [bindplane-otel-collector](https://github.com/observIQ/bindplane-otel-collector) in the Go ecosystem.

### Proposed Solution

- Added above.

### Alternatives Considered

 1. **Copy the functions into each downstream binary.**:  It works, but the functions are non-trivial (~80 lines combined), touch config internals, and would silently diverge from upstream if validation logic changes. Maintainable for one binary, less so for several.
 
 2. **Make `src/main.rs` itself a thin wrapper over library calls.** This is effectively the proposed change - `main.rs` would still exist and work the same way, it would just call into library functions instead of defining them inline. No behavioral change for existing users.
 
 3. **Publish a separate `otap-df-cli` utility crate.** Cleaner separation, but probably overkill for three functions. A module inside `otap-df-controller` or `otap-df-config` is simpler and avoids a new crate.

Issue #2463, opened 2026-03-30
Expose `ObservedStateHandle` from Controller for in-process integrations


lalitb:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

`ObservedStateHandle` is already public, its read methods are public, and it appears to be the intended abstraction for programmatic access to pipeline health. However, when the controller is used as the supported entry point, the handle is created internally and never exposed to the caller.
 
As a result, in-process integrations that need to read pipeline liveness, readiness, or full observed state currently have to go through the admin HTTP server, even though the same data already exists in memory and is already being passed to internal components.
 
 This matters for use cases like:
 
 - Management/reporting agents that need aggregate health
 - Custom supervisors embedded alongside the controller
 - External health checks that want sub-second resolution without HTTP overhead
 
 The admin server itself gets a clone of the handle at [lib.rs](https://github.com/open-telemetry/otel-arrow/blob/1357cedca8c42ba21574dc5dec473288f79ca3ef/rust/otap-dataflow/crates/controller/src/lib.rs#L1283). Exposing the same handle to external code would be a small additive change - the controller's internal behavior doesn't need to change at all.
 
 A few possible shapes:
 
 ```rust
 // Callback on startup
 controller.run_forever_with_handles(config, |handles| { ... });
 
 // Or a builder-style observer
 Controller::new(&FACTORY)
     .with_observer(|handle: ObservedStateHandle| { ... })
     .run_forever(config)?;

```

No strong preference on the API shape — just a supported way to access the handle that already exists.

### Proposed Solution

As above.

### Alternatives Considered

1. Poll the admin HTTP endpoints (`/status`, `/livez`, `/readyz`).
It works, but it adds avoidable HTTP and serialization overhead for data that already exists in-memory and is already exposed through ObservedStateHandle. For integrations running in the same process, this is a less direct abstraction than necessary.

2. Construct a separate ObservedStateStore outside the controller.
The store and handle are public, so external code can create its own instance. In practice, that store would not be wired to the controller’s pipeline lifecycle events, so it would not reflect real runtime state. The missing piece is access to the controller-owned store/handle, not the ability to instantiate another one.

3. Maintain a controller fork. This works mechanically, but it defeats the goal of using the controller as a stable library API. A small additive upstream change is preferable.

lalitb:
#2505 fixes this.

PR #2505, opened 2026-04-01
 feat(otap-dataflow): expose startup helpers as a public library module

lalitb:
Fixes: #2462, #2463

### Summary

- Moves reusable startup functions (`validate_engine_components`, `apply_cli_overrides`, `system_info`) from `src/main.rs` into a new `otap_df_controller::startup` module. This lets custom distribution binaries share the same validation, override, and diagnostics logic without copying code that may drift across releases.
 
-  Adds `run_forever_with_observer` and `run_till_shutdown_with_observer` to `Controller`, exposing the `ObservedStateHandle` via a callback so that in-process integrations can read pipeline liveness and health with zero HTTP overhead.
 
- `src/main.rs` becomes a thin wrapper over these library calls - no behavioral change for existing users. A new "Embedding in Custom Distributions" section is added to the README,  and a runnable `examples/custom_collector.rs` demonstrates the full library embedding pattern end-to-end, including the observer API with an opt-in `--poll-status` flag for periodic health snapshots.

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/builder.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/builder.rs
@@ -121,10 +121,7 @@ impl MetricSignalBuilder {
     /// lengths captured in the checkpoint, discarding rows appended afterward.
     /// This is used by [super::TemporalReaggregationProcessor::flush_at] to
     /// discard some changes if we couldn't fully append an incoming pdata.
-    pub fn finish(
-        &mut self,
-        checkpoint: Option<Checkpoint>,
-    ) -> otap_df_pdata::error::Result<OtapArrowRecords> {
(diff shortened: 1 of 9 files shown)


rust/otap-dataflow/examples/custom_collector.rs
```diff
@@ -0,0 +1,170 @@
...
+        "jemalloc"
+    } else {
+        "system"
+    }
+}
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    // Install the TLS crypto provider before any network operations.
+    otap_df_otap::crypto::install_crypto_provider()
+        .map_err(|e| format!("Failed to install rustls crypto provider: {e}"))?;
```
lquerel, line 111: Just a thought for the future: we could imagine a `prelude` function to handle this kind of initial setup. Or perhaps even better, an attribute macro like `#[engine::main]` on top of `main` to create the engine bootstrap environment.

lalitb: Good point. Created a tracking issue #2519.

PR #2509, opened 2026-04-01
perf: syslog_cef_receiver optimize priority and version parsing

kamilon:
# Change Summary

Parse priority a single time per incoming message.
Parse priority directly from bytes without intermediate string.
Parse version directly from bytes without intermediate string.
Add unit tests for priority parsing and no input case.

Notable bench improvements:
parse_auto_detect/rfc3164: 40.0 ns → 24.5 ns (~39% faster)
parse_auto_detect/rfc5424: 42.4 ns → 37.6 ns (~11% faster)
parse_auto_detect/cef_with_rfc3164: 117.9 ns → 103.2 ns (~12% faster)
parse_auto_detect/cef_with_rfc5424: 67.9 ns → 64.1 ns (~6% faster)

No issue filed.

cargo test and cargo bench.

No.

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/effect_handler.rs
+++ b/rust/otap-dataflow/crates/engine/src/effect_handler.rs
@@ -326,9 +326,12 @@ impl<PData> EffectHandlerCore<PData> {
     /// Send an AckMsg to the runtime control manager for context unwinding.
     /// This will skip if there are no frames.
     ///
-    /// TODO: Note that callers are able to directly invoke route_ack() and route_nack()
-    /// but it means skipping certain frame-related business and they have the same
-    /// signature, so it is easy for callers to do. Find something safer.
+    /// External callers should use [`ConsumerEffectHandlerExtension::notify_ack`]
+    /// and [`ConsumerEffectHandlerExtension::notify_nack`] instead.
+    /// Those wrappers stamp timing information required for correct duration
+    /// metrics. Direct access is gated behind the
+    /// `#[doc(hidden)] pub mod _private` module and the
+    /// [`_private::AckNackRouting`] trait so that accidental bypass is unlikely.
     pub async fn route_ack(&self, ack: AckMsg<PData>) -> Result<(), Error>
     where
         PData: crate::Unwindable,
--- a/rust/otap-dataflow/crates/engine/src/lib.rs
+++ b/rust/otap-dataflow/crates/engine/src/lib.rs
@@ -362,6 +362,33 @@ pub trait ConsumerEffectHandlerExtension<PData> {
     async fn notify_nack(&self, nack: NackMsg<PData>) -> Result<(), Error>;
 }
 
+/// Implementation-detail module re-exporting the internal
+/// [`AckNackRouting`] trait.
+///
+/// **Do not use directly.** Prefer
+/// [`ConsumerEffectHandlerExtension::notify_ack`] /
+/// [`ConsumerEffectHandlerExtension::notify_nack`] which stamp timing
+/// information required for correct duration metrics.
+#[doc(hidden)]
+pub mod _private {
+    use super::*;
+
+    /// Internal routing trait for ack/nack messages.
+    ///
+    /// Callers should use [`ConsumerEffectHandlerExtension::notify_ack`] and
+    /// [`ConsumerEffectHandlerExtension::notify_nack`] instead of calling these
+    /// methods directly. Those wrappers stamp timing information required for
+    /// correct duration metrics before forwarding to `route_ack`/`route_nack`.
+    #[async_trait(?Send)]
+    pub trait AckNackRouting<PData> {
+        /// Routes an ack message to the runtime control manager.
+        async fn route_ack(&self, ack: AckMsg<PData>) -> Result<(), Error>;
+
+        /// Routes a nack message to the runtime control manager.
+        async fn route_nack(&self, nack: NackMsg<PData>) -> Result<(), Error>;
+    }
+}
+
 /// Effect handler extension for adding message source
 #[async_trait(?Send)]
 pub trait MessageSourceLocalEffectHandlerExtension<PData> {
--- a/rust/otap-dataflow/crates/engine/src/local/exporter.rs
+++ b/rust/otap-dataflow/crates/engine/src/local/exporter.rs
@@ -148,22 +148,6 @@ impl<PData> EffectHandler<PData> {
         self.core.start_periodic_telemetry(duration).await
     }
 
-    /// Send an Ack to the runtime control manager for context unwinding.
-    pub async fn route_ack(&self, ack: AckMsg<PData>) -> Result<(), Error>
-    where
-        PData: crate::Unwindable,
-    {
-        self.core.route_ack(ack).await
-    }
-
-    /// Send a Nack to the runtime control manager for context unwinding.
-    pub async fn route_nack(&self, nack: NackMsg<PData>) -> Result<(), Error>
-    where
-        PData: crate::Unwindable,
-    {
-        self.core.route_nack(nack).await
-    }
-
     /// Reports metrics collected by the exporter.
     #[allow(dead_code)] // Will be used in the future. ToDo report metrics from channel and messages.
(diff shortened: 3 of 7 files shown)


rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/mod.rs
```diff
@@ -46,73 +44,101 @@ pub enum ParseError {
 
 /// Parse a syslog message from bytes, automatically detecting the format
 pub(super) fn parse(input: &[u8]) -> Result<ParsedSyslogMessage<'_>, ParseError> {
+    if input.is_empty() {
+        return Err(ParseError::EmptyInput);
+    }
```
utpilla, line 49: This check is not needed. The receiver won't forward zero-length payloads.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/mod.rs
```diff
@@ -139,10 +165,17 @@ pub(super) fn parse_priority(input: &[u8]) -> Result<(Priority, &[u8]), ParseErr
...
+    let mut priority_num: u8 = 0;
+    for &b in priority_bytes {
+        if !b.is_ascii_digit() {
+            return Err(ParseError::InvalidPriority);
+        }
+        priority_num = priority_num
+            .checked_mul(10)
+            .and_then(|v| v.checked_add(b - b'0'))
+            .ok_or(ParseError::InvalidPriority)?;
+    }
```
utpilla, line 174: We could simplify this further by avoiding the branching overhead with `checked_mul` and `checked_add`. We already know we have 1-3 digits with no leading zeros (enforced by the `end < 2 || end > 4` and leading zero checks above). A `u16` accumulator with plain arithmetic would be simpler and avoid the branching overhead on every digit:

```rust
let mut priority_num: u16 = 0;
for &b in priority_bytes {
    if !b.is_ascii_digit() {
        return Err(ParseError::InvalidPriority);
    }
    priority_num = priority_num * 10 + (b - b'0') as u16;
}

if priority_num > 191 {
    return Err(ParseError::InvalidPriority);
}

let priority_num = priority_num as u8;
```

Not a blocker since this change is already an improvement.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/rfc3164.rs
```diff
@@ -53,6 +53,15 @@ pub fn parse_rfc3164(input: &[u8]) -> Result<Rfc3164Message<'_>, ParseError> {
         (None, input)
     };
 
+    parse_rfc3164_with_priority(priority, remaining, input)
+}
+
+/// Parse an RFC 3164 message given a pre-parsed priority and the remaining bytes after `>`.
+pub(super) fn parse_rfc3164_with_priority<'a>(
```
utpilla: The refactoring introduces `parse_rfc3164_with_priority` and `parse_rfc5424_with_priority` as the production code path, but the existing tests still go through the old `parse_rfc3164()` / `parse_rfc5424()` wrappers. This means the new production path (`parse()` → `parse_*_with_priority()`) isn't directly tested by the comprehensive test suites in `rfc3164.rs` and `rfc5424.rs`.

Could we do the following?
1. Remove `parse_rfc3164()` and `parse_rfc5424()`, they're no longer on the production path
2. Update tests in `rfc3164.rs` and `rfc5424.rs` to go through `parse()` instead
3. Update benchmarks: `parser_comparison` becomes redundant with `parse_auto_detect`, so we could remove it. `arrow_batch_creation` should use `bench_support::parse()`.

This way the tests actually exercise the production code path, instead of going through wrapper functions that nothing in production calls.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/mod.rs
kamilon: I had to move this check back after refactoring tests.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/rfc3164.rs
kamilon: I had AI take care of this comment. I'm happy with what it did. Please note it moved a super to crate for the parse method.

I also had it move 2 tests between files since it was really testing a fallback that is more applicable to the other format.

rust/otap-dataflow/benchmarks/benches/syslog_cef_receiver/main.rs
```diff
@@ -10,24 +10,21 @@
...
+    | parse_auto_detect/cef_with_rfc3164                | 107.40 ns    |
+    | parse_auto_detect/cef_with_rfc5424                | 64.228 ns    |
+    | timestamp_extraction/rfc3164                      | 580.73 ns    |
+    | timestamp_extraction/rfc5424                      | 23.816 ns    |
+    | cef_extensions/one_extension                      | 21.113 ns    |
+    | cef_extensions/ten_extensions                     | 200.95 ns    |
+    | cef_extensions/ten_extensions_with_escape         | 219.42 ns    |
+    | arrow_batch_creation/rfc3164_arrow_batch_100_msgs | 102.49 µs    |
+    | arrow_batch_creation/rfc5424_arrow_batch_100_msgs | 51.058 µs    |
+    | arrow_batch_creation/cef_arrow_batch_100_msgs     | 45.241 µs    |
```
lquerel, line 27: How should I interpret this table? As it stands, it gives the impression of a performance regression. Is it because the results are:

* produced from a different machine configuration?
* produced on a machine with other workloads interfering with the measurements?
* reflecting a real regression?

kamilon: This confused me at first too.

It's because the first 3 rows (parser_comparison) are gone now because of the final refactor. It's a net gain for all benches except the last 3 rows (arrow_batch_creation*). The last 3 are now testing more than they did before (auto detect) so it's not apples to apples and even then, it's almost at the level of noise.

VS Code rendered the table in a more obvious way that offset the rows properly. It caught me off guard when I was looking on GitHub.

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/rfc3164.rs
```diff
@@ -16,7 +16,7 @@ pub struct Rfc3164Message<'a> {
     pub(super) input: &'a [u8],
 }
 
-/// Parse an RFC 3164 syslog message from bytes, automatically detecting the format
+/// Parse an RFC 3164 syslog message given a pre-parsed priority and the remaining bytes after `>`.
```
utpilla:
```suggestion
/// Parse an RFC 3164 syslog message given an optional pre-parsed priority and the remaining bytes after `>`.
```

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/mod.rs
```diff
@@ -14,8 +12,8 @@ pub mod rfc5424;
 
 use self::cef::parse_cef;
 use self::parsed_message::ParsedSyslogMessage;
-use self::rfc3164::parse_rfc3164;
-use self::rfc5424::parse_rfc5424;
+use self::rfc3164::parse_rfc3164_with_priority;
+use self::rfc5424::parse_rfc5424_with_priority;
```
utpilla: I realized that now that the old `parse_rfc3164` and `parse_rfc5424` wrappers are removed, we could rename `parse_rfc3164_with_priority` → `parse_rfc3164` and `parse_rfc5424_with_priority` → `parse_rfc5424`.

The `_with_priority` suffix was needed earlier to disambiguate, but with only one function remaining, it's not needed.

PR #2512, opened 2026-04-02
Update the top-level README

jmacd:
# Change Summary

Update README to highlight our progress in Phase 2. We did a lot!

Yes. I tested the example commands in this update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,8 @@ cargo build --bin df_engine --no-default-features
 A [directory of example configurations][EXAMPLE-CONFIGS] provides a
 number of examples (e.g.,
 [syslog-console.yaml][SYSLOG-CONSOLE-YAML]). For example, to receive
-syslog messages on port 5140 and print them to the console:
+syslog messages with our [Syslog/CEF receiver][SYSLOG-CEF] on port
+5140 and print them to the console:
 
 ```bash
 ./target/debug/df_engine -c ./configs/syslog-console.yaml


README.md
```diff
@@ -7,286 +7,246 @@
...
+
+The OTAP Dataflow Engine is **embeddable software**. Our repo builds a
+demonstration `df_engine` artifact with core nodes included and
+features YAML configuration, but really it's designed for safely
+adding OpenTelemetry features and capabilities in other programs,
+anywhere that Rust can be compiled, with fine-grained control over
+memory and CPU resources.
+
+The OTAP Dataflow Engine has built-in support for OTAP and OTLP,
+receivers and exporters, and built-in procsesors featuring batching,
```
cijothomas:
```suggestion
receivers and exporters, and built-in processors featuring batching,
```

README.md
```diff
@@ -7,286 +7,246 @@
...
+./target/debug/df_engine -c ./configs/syslog-console.yaml --num-cores=1
+```
+
+Linux/MacOS users can test this with:
+
+```
+﻿logger -n 127.0.0.1 -P 5140 -d --rfc3164 "hello world"
+```
+
+PowerShell users can test this with:
```
cijothomas, line 156: never realized we can use powershelgl to send a syslog via UdP!

README.md
```diff
@@ -7,286 +7,246 @@
...
-to compression. For very large batch sizes, ZSTD does an excellent job as long
-as the compression window is sufficiently large, but even in this case, the new
-protocol remains superior. As previously mentioned, these compression gains can
-be higher for traffic predominantly containing multivariate metrics.
-
-![Avg % of compressed size improvement of OpenTelemetry Protocol with Apache Arrow over OTLP (zstd compression)](./docs/img/average_improvement_heatmap.png)
-
-For historical Phase 1 details, see the [Phase 1 benchmark
-results](docs/benchmarks-phase1.md) page.
+As a community, we are planning phase 3, see the links below to join us.
```
cijothomas, line 184: any preview of phase3 plans worth mentioning here?

README.md
```diff
@@ -7,286 +7,246 @@
...
+abbreviated **OTAP** for OTel-Arrow Protocol, a column-oriented
+representation for OpenTelemetry data supporting efficient in-memory
+and on-wire telemetry exchange. Where OpenTelemetry's OTLP is a
+row-oriented protocol, OTel-Arrow's OTAP protocol uses Apache Arrow to
+encode telemetry in a columnar format that is more efficient for CPUs
+to work with, because of vectorization, and compresses dramatically
+better, especially over long-lived streams with the use of [Arrow
+IPC][ARROW-IPC] stream encoding.
+
+OTAP maintains 100% compatibility with the OpenTelemetry data model,
```
albertlockett: > 100% compatibility

Is this too strong of a claim? We don't yet support profiles

README.md
```diff
@@ -7,286 +7,246 @@
...
+Apache Arrow libraries follow standard patterns and naming
+conventions, organized around specification documents, and as with
+OpenTelemetry, the libraries have each their own ways to address
+language-specific challenges. Using an Apache Arrow library to
+construct a record batch makes it easy and efficient to exchange data
+across language, process, and network boundaries. You can build a
+record batch in one language, and pass it to another language without
+copying the data. Record batches support zero-copy operations, and for
+network and file-based communications, each library includes a reader
+and writer for [Arrow IPC][ARROW-IPC].
```
albertlockett: nit: I'm on the fence wrt to whether it makes sense to talk about Arrow libraries here. Is the correct understanding that this that this section is trying to give some background on apache arrow, and libraries are a big part of the actual deliverables in this ecosystem so therefore merit mentioning?

If I came to this README as someone who wasn't a developer, I probably wouldn't care about libraries. I'd also like to avoid giving the impression that developers need to be using Arrow libraries as a starting point for interacting with OTel-Arrow

jmacd: Agree -- I got myself onto a tangent.

jmacd: I shortened this to say "record batch" "zero copy" and "Arrow IPC" and nothing more.

README.md
jmacd: I updated to name the three signals we support, adding that Profiles is future work.

README.md
jmacd: I thought about what we might put here, but it's something we should discuss and agree to first. We have a lot of ideas!

README.md
jmacd: I removed newlines. I couldn't find any built-in tools, though `nc` is easy to install. But yes, this is tested!

README.md
```diff
@@ -7,286 +7,244 @@
...
+ourselves](./rust/otap-dataflow/crates/telemetry/README.md) with an
+experimental OpenTelemetry SDK that emits OTAP directly, meaning we
+have an **end-to-end column-oriented telemetry pipeline** in Rust.
+
+Our Golang Collector components [`otelarrowreceiver`][RECEIVER] and
+[`otelarrowexporter`][EXPORTER] have been included in the
+OpenTelemetry Collector-Contrib distribution since [the July 2024
+release of v0.104.0][ARROW-RELEASED].
+
+Our Rust and Go codebases are both used in production. Our project has
```
cijothomas: > used in production

We probably need to add some links/references to make that statement?

README.md
```diff
@@ -7,286 +7,244 @@
...
+better, especially over long-lived streams with the use of [Arrow
+IPC][ARROW-IPC] stream encoding.
+
+OTAP maintains 100% compatibility with the OpenTelemetry data model
+for logs, traces, and metrics, with a straight-forward and non-lossy
+round trip from OTLP to OTAP and back, and support for the
+OpenTelemetry Profiles signal in OTAP is important future work.
+
+In both our Rust and Golang implementations, we support combined OTAP
+and OTLP services on a single port, making OTel-Arrow OTAP/OTLP
```
cijothomas: did you mean we support OTLP and OTAP receiver on a single port? .. This sentence reads a bit odd..

README.md
```diff
@@ -7,286 +7,244 @@
...
+
+```bash
+git clone https://github.com/open-telemetry/otel-arrow.git
+cd otel-arrow/rust/otap-dataflow
+cargo build --bin df_engine --no-default-features
+```
+
+A [directory of example configurations][EXAMPLE-CONFIGS] provides a
+number of examples (e.g.,
+[syslog-console.yaml][SYSLOG-CONSOLE-YAML]). For example, to print
```
cijothomas: to receive syslog messages on port 5140 and then print them to console

README.md
```diff
@@ -7,286 +7,244 @@
...
+syslog records to the console on port 5140:
+
+> [!NOTE]
+> This is insecure! See the [Syslog/CEF Receiver
+> documentation][SYSLOG-CEF] for secure configuration examples.
+
+```bash
+./target/debug/df_engine -c ./configs/syslog-console.yaml --num-cores=1
```
cijothomas: nit: maybe we should mention the relevance of --num-cores=1 ? or omit it?

jmacd: Fixed. I put the num-cores setting into the configuration instead. Thank you! 

```
policies:
  resources:
    core_allocation:
      type: core_count
      count: 1
```

README.md
jmacd: I thought we did! I've removed the entire paragraph. We'll get there one day (FYI @lalitb).

README.md
jmacd: To keep this simple, I removed that sentence.

README.md
```diff
@@ -7,247 +7,199 @@
...
+
+The OTAP Dataflow Engine is **embeddable software**. Our repo builds a
+demonstration `df_engine` artifact with core nodes included and
+features YAML configuration, but really it's designed for safely
+adding OpenTelemetry features and capabilities in other programs,
+anywhere that Rust can be compiled, with fine-grained control over
+memory and CPU resources.
+
+The OTAP Dataflow Engine has built-in support for OTAP and OTLP,
+receivers and exporters, and built-in processors featuring batching,
```
cijothomas, line 36: nit: Mention syslog/cef too

cijothomas: nvm, its already covered below.

README.md
```diff
@@ -7,247 +7,199 @@
...
+ourselves](./rust/otap-dataflow/crates/telemetry/README.md) with an
+experimental OpenTelemetry SDK that emits OTAP directly, meaning we
+have an **end-to-end column-oriented telemetry pipeline** in Rust.
+
+Our Golang Collector components [`otelarrowreceiver`][RECEIVER] and
+[`otelarrowexporter`][EXPORTER] have been included in the
+OpenTelemetry Collector-Contrib distribution since [the July 2024
+release of v0.104.0][ARROW-RELEASED].
+
+Our project has many contributors and over 1500 pull requests merged.
```
cijothomas: I'd remove the count, as it gets stale information.

README.md
```diff
@@ -7,247 +7,199 @@
...
+cd otel-arrow/rust/otap-dataflow
+cargo build --bin df_engine --no-default-features
+```
+
+A [directory of example configurations][EXAMPLE-CONFIGS] provides a
+number of examples (e.g.,
+[syslog-console.yaml][SYSLOG-CONSOLE-YAML]). For example, to receive
+syslog messages on port 5140 and print them to the console:
+
+> [!NOTE]
```
cijothomas: nit: Don't think we need to callout insecure part. OTLP default is also non TLS, and usually its upto end users to decide whether its insecure or not.

PR #2515, opened 2026-04-02
fix: handle DrainIngress in fake_data_generator to unblock graceful shutdown

### LINKED ISSUE #2511: pipeline-perf-test-basic unit test is failing
### Pre-filing checklist

### Component(s)

### Bug Description

The pipeline-perf-test-basic unit test is consistently failing during PR validation runs. Here is an example of a recent failure:

https://github.com/open-telemetry/otel-arrow/actions/runs/23866008369/job/69585200727?pr=2481

Specific error:

`2026-04-01T19:28:11.756Z  ERROR otap-df-state::state.observed_error: Shutdown requested via the `/pipeline-groups/shutdown` endpoint. [pipeline_group_id=default, pipeline_id=main, core_id=2, event_type=**DrainDeadlineReached**]
`
It could potentially be related to https://github.com/open-telemetry/otel-arrow/commit/3dca283731763dc0e6a645afdf444fc1b985bc8a

### Steps to Reproduce

Create a PR and the unit test will fail.

### Expected Behavior

pipeline-perf-test-basic unit test should not fail.

### Actual Behavior

pipeline-perf-test-basic unit test fails.

### OTel-Arrow Version

3dca283731763dc0e6a645afdf444fc1b985bc8a

### Environment

Automated PR integration tests.

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

sjmsft:
# Change Summary

The "Ack nack redesign" PR (3dca2837) introduced a two-phase DrainIngress/ReceiverDrained shutdown protocol but missed updating the fake_data_generator receiver. Without the DrainIngress handler, the message falls into the _ => {} catch-all, notify_receiver_drained() is never called, the pipeline controller never removes the receiver from its pending set, and after the deadline expires it emits DrainDeadlineReached. This was causing pipeline-perf-test-basic to fail consistently.

pipeline-perf-test-basic unit test is failing.

* Closes #2511

fake_data_generator and runtime_control_metrics tests were executed.

No, fake_data_generator is an internal test/load-generation receiver, not a user-facing component.

### DIFF
--- a/rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
+++ b/rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
@@ -81,8 +81,8 @@ pub fn concatenate<const N: usize>(
     }
 
     if items.len() == 1 {
-        for i in 0..N {
-            result[i] = items[0][i].take();
+        for (i, item) in result.iter_mut().enumerate().take(N) {
+            *item = items[0][i].take();
         }
         return Ok(result);
     }
--- a/rust/otap-dataflow/crates/validation/src/lib.rs
+++ b/rust/otap-dataflow/crates/validation/src/lib.rs
@@ -69,6 +69,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
     fn debug_processor() {
         Scenario::new()
             .pipeline(
@@ -130,6 +131,7 @@ mod tests {
     }
 
     #[test]
+    #[ignore]
     fn filter_processor_pipeline() {
         let attr_check = ValidationInstructions::AttributeRequireKeyValue {
             domains: vec![AttributeDomain::Signal],
--- a/rust/otap-dataflow/src/main.rs
+++ b/rust/otap-dataflow/src/main.rs
@@ -372,6 +372,7 @@ connections:
     }
 
     #[test]
+    #[allow(clippy::const_is_empty)]
     fn license_text_is_embedded() {
         assert!(
             !LICENSE_TEXT.is_empty(),
@@ -384,6 +385,7 @@ connections:
     }
 
     #[test]
+    #[allow(clippy::const_is_empty)]
     fn third_party_notices_are_embedded() {
         assert!(
             !THIRD_PARTY_NOTICES.is_empty(),


lalitb:
The sequencing here looks off. In graceful shutdown the runtime does:

`DrainIngress` -> `ReceiverDrained` -> `downstream Shutdown`. 

This change makes fake_data_generator do 

`DrainIngress` -> `ReceiverDrained` -> `wait for Shutdown`, 

but that Shutdown is not part of the normal post-drain receiver path. For this receiver, once ingress is stopped there is no receiver-local work left to preserve, so it should exit directly on DrainIngress rather than report drained and then block waiting for another shutdown message.

lalitb (changes requested):
Please go through the comment [here](https://github.com/open-telemetry/otel-arrow/pull/2515#issuecomment-4179257890).

lalitb:
The correct fix should be:

`DrainIngress` -> `notify_receiver_drained()` -> return `TerminalState` immediately

Something like (not tested):

```rust
  Ok(NodeControlMsg::DrainIngress { deadline, .. }) => {
      otel_info!("fake_data_generator.drain_ingress");                                                                                                                                                                               
      return Ok(TerminalState::new(deadline, [self.metrics.snapshot()]));                                                                                                                                                            
  }    
```

lalitb:
The fix now looks correct. However from CI failures, there looks like a shutdown race in `fake_data_generator` that is easy to hit on slower runners The test config uses `signals_per_second = 1`, so the receiver can sleep for close to 1 second between sends, while the shutdown deadline in `test_telemetry_registries_cleanup` is only `200ms`. That means `DrainIngress` can arrive while the receiver is asleep, the runtime can move into forced shutdown before the receiver handles it, and then `notify_receiver_drained().await?` can fail with Channel is closed.

One option could be to address this in two places:

- make the rate-limit sleep interruptible, since that looks like the root cause here.
```rust
if signals_per_second.is_some() {
    let remaining_time = wait_till - Instant::now();
    if remaining_time.as_secs_f64() > 0.0 {
        tokio::select! {
            biased;

            ctrl_msg = ctrl_msg_recv.recv() => {
                // handle DrainIngress / Shutdown during the rate-limit wait
                // using the same control-message handling as the main loop
            }

            _ = sleep(remaining_time) => {}
        }
    }
}

```
- make `notify_receiver_drained()` best-effort on the terminal `DrainIngress` path, so a late control-plane teardown does not turn shutdown into an error.
```rust
Ok(NodeControlMsg::DrainIngress { deadline, .. }) => {
    otel_info!("fake_data_generator.drain_ingress");
    let _ = effect_handler.notify_receiver_drained().await;
    return Ok(TerminalState::new(deadline, [self.metrics.snapshot()]));
}
```

lquerel:
@sjmsft 

[in addition to the sequence described by @lalitb ]

The exception is deadline-forced shutdown. If the drain deadline expires before the receiver reports drained, the runtime sends `NodeControlMsg::Shutdown { deadline, reason }` to any still-pending receivers.

lalitb (approved):
The updated implementation addresses my previous feedback. The fix correctly follows the drain protocol: DrainIngress -> notify_receiver_drained() -> return TerminalState immediately, without waiting for a subsequent Shutdown. The tokio::select! in the rate-limit sleep is the right approach to ensure prompt shutdown response. Regression test looks good.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
```diff
@@ -268,6 +269,37 @@ impl BatchCache {
...
+    match ctrl_msg {
+        Ok(NodeControlMsg::CollectTelemetry {
+            mut metrics_reporter,
+        }) => {
+            _ = metrics_reporter.report(metrics);
+            Ok(None)
+        }
+        Ok(NodeControlMsg::DrainIngress { deadline, .. }) => {
+            otel_info!("fake_data_generator.drain_ingress");
+            let _ = effect_handler.notify_receiver_drained().await;
```
lalitb:
```suggestion
            effect_handler.notify_receiver_drained().await?;
```

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
```diff
@@ -426,7 +442,18 @@ impl local::Receiver<OtapPdata> for FakeGeneratorReceiver {
...
+                                    // promptly instead of waiting for the full interval.
+                                    tokio::select! {
+                                        biased;
+                                        ctrl_msg = ctrl_msg_recv.recv() => {
+                                            if let Some(terminal) = handle_control_msg(ctrl_msg, &effect_handler, &mut self.metrics).await? {
+                                                return Ok(terminal);
+                                            }
+                                        }
+                                        _ = sleep(remaining_time) => {}
+                                    }
```
lalitb, line 461: The current sleep is making the DrainIngress/Shutdown responsible , but it is also changing the rate-limiting behavior - so any non-terminal control message handled as `Ok(None)` exist the sleep immediately and next batch can be sent before the original `wait_till`. We should replace the line 445-456 above with:
```rust
// Keep the original sleep deadline even if non-terminal control
// messages arrive. Only DrainIngress/Shutdown should interrupt
// the rate-limit wait early.
let sleep_until = sleep(remaining_time);
tokio::pin!(sleep_until);

loop {
    tokio::select! {
        biased;
        ctrl_msg = ctrl_msg_recv.recv() => {
            if let Some(terminal) =
                handle_control_msg(ctrl_msg, &effect_handler, &mut self.metrics).await?
            {
                return Ok(terminal);
            }
        }
        _ = &mut sleep_until => break,
    }
}

```

sjmsft: Please check the new changes + new test.

lalitb (approved):
Thanks.

PR #2527, opened 2026-04-02
feat(otap-dataflow): Add `scope_name` selector to Metric views

### LINKED ISSUE #2500: Allow selection of metric views with scope_name in addition to instrumentation_name
### Pre-filing checklist

### Component(s)

Users of the internal telemetry pipeline may get confusing results if referencing an `ArrowMetricView` with a commonly-used `instrument_name`. As an example, multiple components could use the same `instrument_name`:

* Component 1: `component1.requests.total`
* Component 2: `component2.requests.total`

```rust
selector: MetricSelector {
    instrument_name: Some("requests.total".to_string()),
},
```

It appears the result of this would combine metric values from both types of components.

### Proposed Solution

If we support using metric `scope_name` as an additional selector, we can focus on that metric only from a single component:

```rust
selector: MetricSelector {
    instrument_name: Some("requests.total".to_string()),
    scope_name: Some("component1".to_string())
},
```

### Alternatives Considered

_No response_

### Additional Context

_No response_

drewrelmas:
# Change Summary

Add `scope_name` selector to Metric views
- Optional config alongside existing `instrument_name`
- Fixed yaml indentation of `fake-debug` configs

* Closes #2500

Ran the fake debug configs and queried admin `/metrics` endpoint to see resulting views

No


rust/otap-dataflow/configs/fake-debug-noop-promethueus-telemetry.yaml
```diff
@@ -9,12 +9,23 @@ engine:
...
+            name: "otlp.logs.produced.count"
+            description: "Count of logs produced"
+        - selector:
+            scope_name: "debug.processor.pdata.metrics"
+            instrument_name: "logs.consumed"
+          stream:
+            name: "debug.logs.consumed.count"
+            description: "Count of log messages consumed by debug processor"
+        - selector:
+            scope_name: "fake_data_generator.receiver.metrics"
```
cijothomas: without instrument_name, this matches multiple instruments right? we should show example where a single instrument is only selected.

drewrelmas: That's a good point, let me take another look and pick one that only has one. I'm wondering if there is even a use case in only doing `scope_name` and not specifying `instrument`...

cijothomas: There is, but it's better we only show the case where a *single* instrument gets matched. (Views are complicated beasts, and this multiple selection can result in unexpected results!)

rust/otap-dataflow/configs/fake-debug-noop-promethueus-telemetry.yaml
```diff
@@ -9,12 +9,23 @@ engine:
...
                 path: "/metrics"
-    views:
-      - selector:
-          instrument_name: "logs.produced"
-        stream:
-          name: "otlp.logs.produced.count"
-          description: "Count of logs produced"
+      views:
+        - selector:
+            instrument_name: "logs.produced"
```
cijothomas, line 15: lets modify everything in example to use scope_name+instrument_name to avoid any ambiguous matching.

drewrelmas: done

rust/otap-dataflow/configs/fake-debug-noop-promethueus-telemetry.yaml
drewrelmas: done

rust/otap-dataflow/configs/fake-debug-noop-promethueus-telemetry.yaml
```diff
@@ -9,12 +9,19 @@ engine:
...
                 path: "/metrics"
-    views:
-      - selector:
-          instrument_name: "logs.produced"
-        stream:
-          name: "otlp.logs.produced.count"
-          description: "Count of logs produced"
+      views:
+        - selector:
+            scope_name: "fake_data_generator.receiver.metrics"
```
cijothomas, line 14: for a future PR, I think we should remove the word "metrics" from the scope! its as Meter name, so kinda redundant to use  ".metrics". 

I'll tackle it myself in future.

cijothomas (approved):
LGTM, and good catch on the issue as well.

PR #2529, opened 2026-04-02
feat(config): add URI providers for file: and env: config sources

### LINKED ISSUE #2514: Feature: Add OTel Collector-compatible config URI providers (file:, env:)
### Pre-filing checklist

### Component(s)

Add support for configuration URI schemes to the `df_engine` / data-plane binary, matching the [OTel Collector configuration location](https://opentelemetry.io/docs/collector/configuration/#location) pattern. This allows users to specify config sources using URIs like `--config=file:/path/to/config.yaml` or `--config=env:MY_CONFIG_VAR`.

Currently, the data-plane binary requires a `--config <path>` argument pointing to a file on disk. This works well for Kubernetes deployments with ConfigMaps mounted as volumes, but is less convenient for:

1. **Docker Compose** -- users must create volume mounts even for simple configs
2. **Environment-based config** -- in some deployment patterns, configuration is passed via environment variables (e.g., injected by orchestrators, stored in secrets managers)
3. **Consistency with OTel ecosystem** -- the OTel Collector supports `file:`, `env:`, `yaml:`, `http:`, `https:` schemes, and users familiar with the Collector expect similar flexibility

By adopting the same URI scheme pattern, we:
- Reduce friction for users coming from the OTel Collector
- Enable config-in-env-var workflows (no ConfigMap/volume needed for simple cases)
- Lay the groundwork for future providers (remote config, inline YAML, etc.)
## Relationship to Existing `${env:VAR}` Substitution

The otap-dataflow config system already supports environment variable substitution *inside config content* via `${env:VAR}` patterns (see `env_substitution.rs`). This is useful for interpolating individual secrets or values:

```yaml
# Existing: env var substitution INSIDE config content
exporters:
  otlp:
    endpoint: ${env:OTLP_ENDPOINT}
    headers:
      api-key: ${env:API_KEY:-default-key}
```

The proposed `env:` URI scheme is **complementary, not competing**:

| Feature | Existing `${env:VAR}` | Proposed `--config=env:VAR` |
|---------|----------------------|----------------------------|
| **Scope** | Individual values inside config | Entire config content |
| **Use case** | Inject secrets into a file-based config | Store complete config in env var (no file needed) |
| **Example** | `endpoint: ${env:URL}` in YAML | `--config=env:MY_PIPELINE_CONFIG` |
| **When evaluated** | During config parsing (after file read) | Before config parsing (determines source) |

Both features can be used together: you could load a config from an env var (`--config=env:MY_CONFIG`) that itself contains `${env:SECRET}` placeholders that get substituted during parsing.

### Phase 1: Core Provider Abstraction (this issue)

Implement a config provider abstraction with two initial providers:

| Scheme | Description | Example |
|--------|-------------|---------|
| `file:` | Read config from a file path | `--config=file:/etc/config.yaml` |
| `env:` | Read config **content** from an environment variable | `--config=env:MY_PIPELINE_CONFIG` |
| *(bare path)* | Backwards-compatible, treated as `file:` | `--config=/etc/config.yaml` |

**Behavior:**
- `file:/path/to/config.yaml` -- reads the file at the given path
- `file:relative/path.yaml` -- reads relative to current working directory
- `env:VAR_NAME` -- reads the **content** of the environment variable `VAR_NAME` as the config (not a path to a file)
- Bare paths (no scheme) are treated as `file:` for backwards compatibility

**Fallback when no `--config` is provided:**
Scan well-known paths in order:
1. `/etc/o11y-gateway/config.yaml`
2. `/etc/data-plane/configs/pipeline_config.yaml` (backwards-compat)
3. `/app/config.yaml`

If none found, return an error with a helpful message listing the options.

### Example Usage

```bash
# Explicit file scheme
df_engine --config=file:/etc/o11y-gateway/config.yaml

# Bare path (backwards-compatible)
df_engine --config=/etc/o11y-gateway/config.yaml

# Config content in environment variable
export MY_CONFIG='
version: otel_dataflow/v1
engine: {}
groups:
  default:
    pipelines:
      main:
        nodes:
          receiver:
            type: "urn:otel:receiver:traffic_generator"
            outputs: [out_port]
            config:
              traffic_config:
                max_signal_count: 1000
          noop:
            type: "urn:otel:exporter:noop"
        connections:
          - from: receiver["out_port"]
            to: noop
'
df_engine --config=env:MY_CONFIG
```

### Phase 2: Additional Providers (follow-up issues)

These can be added incrementally after the provider abstraction is in place:

| Scheme | Description | Complexity |
|--------|-------------|------------|
| `yaml:` | Inline YAML from CLI with `::` key separators | Low |
| `http:` | Fetch config from HTTP URL | Medium |
| `https:` | Fetch config from HTTPS URL | Medium |

Support for multiple `--config` flags with merge semantics (like OTel Collector) could also be a future enhancement.

### Suggested Architecture

```rust
/// A provider that can resolve a config URI to config content.
pub trait ConfigProvider: Send + Sync {
    /// Returns true if this provider handles the given scheme (e.g., "file", "env").
    fn scheme(&self) -> &str;

    /// Resolves the URI to config content (YAML/JSON string).
    fn resolve(&self, uri: &str) -> Result<String, ConfigError>;
}

/// Built-in providers
pub struct FileConfigProvider;
pub struct EnvConfigProvider;

/// Resolver that dispatches to the appropriate provider based on URI scheme.
pub struct ConfigResolver {
    providers: Vec<Box<dyn ConfigProvider>>,
}

impl ConfigResolver {
    pub fn resolve(&self, uri: &str) -> Result<String, ConfigError> {
        let (scheme, rest) = parse_uri(uri)?;
        for provider in &self.providers {
            if provider.scheme() == scheme {
                return provider.resolve(rest);
            }
        }
        Err(ConfigError::UnknownScheme(scheme))
    }
}
```

### URI Parsing

Simple scheme detection:
- If URI contains `://` or starts with `scheme:`, extract the scheme
- If URI starts with `/` or `.`, treat as bare path → `file:` scheme
- Otherwise, check for known schemes (`file:`, `env:`)

The OTel Collector uses `:` as the delimiter (not `://`), so `file:/path` and `env:VAR` are valid.

### Error Messages

When config resolution fails, provide helpful error messages:

```
Error: Could not resolve config from 'env:MY_CONFIG': environment variable MY_CONFIG is not set

Hint: Set the MY_CONFIG environment variable with your pipeline configuration, or use:
  --config=file:/path/to/config.yaml
  --config=env:OTHER_VAR_NAME
```

When no `--config` is provided and no well-known path exists:

```
Error: No configuration provided and no config file found at well-known paths.

Searched:
  - /etc/o11y-gateway/config.yaml (not found)
  - /etc/data-plane/configs/pipeline_config.yaml (not found)
  - /app/config.yaml (not found)

Provide a configuration using one of:
  --config=file:/path/to/config.yaml
  --config=env:MY_CONFIG_VAR
```

## Acceptance Criteria

- [ ] `--config=file:/path/to/config.yaml` reads config from file
- [ ] `--config=env:VAR_NAME` reads config content from environment variable
- [ ] Bare paths (`--config=/path/to/file`) work for backwards compatibility
- [ ] Well-known path fallback when no `--config` is provided
- [ ] Helpful error messages when resolution fails
- [ ] Unit tests for both providers
- [ ] Integration test demonstrating `env:` usage
- [ ] Documentation updated with new URI schemes

### Alternatives Considered

_No response_

## References

- [OTel Collector Configuration Location](https://opentelemetry.io/docs/collector/configuration/#location)
- Internal discussion: GitLab issue #97 comments

mateenali66:
Closes #2514

Adds Phase 1 of the config URI provider feature. The goal is to match the OTel Collector's `--config` URI pattern so users coming from the collector get a familiar experience.

**What changed:**
- New `config_provider` module in `otap-df-config` with a `ConfigProvider` trait, `FileConfigProvider`, `EnvConfigProvider`, and a `ConfigResolver` that dispatches by URI scheme
- `--config` now accepts URIs: `file:/path/to/config.yaml`, `env:MY_VAR`, or bare paths (still works for backwards compat)
- When `--config` is omitted, falls back to scanning well-known paths (`/etc/o11y-gateway/config.yaml`, `/etc/data-plane/configs/pipeline_config.yaml`, `/app/config.yaml`)
- Helpful error messages when resolution fails (lists what was searched, shows what to use instead)
- Unit tests for both providers, resolver dispatch, and an integration test that round trips through env var to parsed YAML
- The existing `${env:VAR}` substitution in `env_substitution.rs` is untouched. The two features work at different layers

**Example usage:**

```
df_engine --config=file:/etc/o11y-gateway/config.yaml
df_engine --config=env:MY_PIPELINE_CONFIG
df_engine --config=/path/to/config.yaml  # still works
```

Phase 2 (`yaml:`, `http:`, `https:`) can be added incrementally now that the abstraction is in place.

## Test plan
- [x] `cargo xtask check` passes (fmt, clippy, all workspace tests)
- [ ] Verify `--config file:<path>` reads a local config file
- [ ] Verify `--config env:<VAR>` reads config from an environment variable
- [ ] Verify bare path `--config /path/to/config.yaml` still works
- [ ] Verify omitting `--config` produces the well-known path search error message
- [ ] Verify unknown scheme produces a clear error

### DIFF
--- a/rust/otap-dataflow/crates/config/src/config_provider.rs
+++ b/rust/otap-dataflow/crates/config/src/config_provider.rs
@@ -21,13 +21,26 @@ use std::path::Path;
 /// Fallback config path tried when `--config` is omitted.
 const DEFAULT_CONFIG_PATH: &str = "config.yaml";
 
-/// The result of resolving a config URI: the original source URI and the loaded content.
+/// The serialization format of resolved configuration content.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ConfigFormat {
+    /// YAML format (the default for most config sources).
+    Yaml,
+    /// JSON format (detected by a `.json` file extension).
+    Json,
+}
+
+/// The result of resolving a config URI: the original source URI, the loaded
+/// content, and the detected serialization format.
 #[derive(Debug)]
 pub struct ResolvedConfig {
-    /// The original URI or path string used to load the config (e.g. "file:/etc/config.yaml", "env:MY_VAR").
+    /// The original URI or path string used to load the config
+    /// (e.g. "file:/etc/config.yaml", "env:MY_VAR").
     pub source: String,
     /// The raw configuration content (YAML or JSON string).
     pub content: String,
+    /// The serialization format of `content`, derived from the source URI.
+    pub format: ConfigFormat,
 }
 
 /// A provider that can resolve configuration content from a URI with a specific scheme.
@@ -56,9 +69,15 @@ impl ConfigProvider for FileConfigProvider {
             context: crate::error::Context::default(),
             details: format!("{path}: {e}"),
         })?;
+        let format = if path.ends_with(".json") {
+            ConfigFormat::Json
+        } else {
+            ConfigFormat::Yaml
+        };
         Ok(ResolvedConfig {
             source: uri.to_string(),
             content,
+            format,
         })
     }
 }
@@ -79,6 +98,8 @@ impl ConfigProvider for EnvConfigProvider {
         Ok(ResolvedConfig {
             source: uri.to_string(),
             content,
+            // Env vars are assumed to contain YAML; JSON is not common here.
+            format: ConfigFormat::Yaml,
         })
     }
 }
@@ -211,6 +232,21 @@ mod tests {
         let resolved = provider.resolve(&path).expect("should read file");
         assert_eq!(resolved.content, "hello: world");
         assert_eq!(resolved.source, path);
+        assert_eq!(resolved.format, ConfigFormat::Yaml);
+    }
+
+    #[test]
+    fn file_provider_detects_json_format() {
+        let mut tmp = tempfile::Builder::new()
+            .suffix(".json")
+            .tempfile()
+            .expect("create temp json file");
+        write!(tmp, r#"{{"key": "value"}}"#).expect("write temp file");
+
+        let provider = FileConfigProvider;
+        let path = format!("file:{}", tmp.path().display());
+        let resolved = provider.resolve(&path).expect("should read json file");
+        assert_eq!(resolved.format, ConfigFormat::Json);
     }
 
     #[test]
@@ -237,6 +273,7 @@ mod tests {
         let resolved = provider.resolve(&uri).expect("should read env var");
         assert_eq!(resolved.content, "version: v1");
         assert_eq!(resolved.source, uri);
+        assert_eq!(resolved.format, ConfigFormat::Yaml);
         unsafe {
             env::remove_var(var_name);
         }
--- a/rust/otap-dataflow/src/main.rs
+++ b/rust/otap-dataflow/src/main.rs
@@ -4,7 +4,7 @@
 //! Create and run a multi-core pipeline
 
 use clap::Parser;
-use otap_df_config::config_provider::resolve_config;
+use otap_df_config::config_provider::{ConfigFormat, resolve_config};
 use otap_df_config::engine::{
     HttpAdminSettings, OtelDataflowSpec, SYSTEM_OBSERVABILITY_PIPELINE_ID, SYSTEM_PIPELINE_GROUP_ID,
 };
@@ -309,14 +309,9 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
     println!("{}", system_info());
 
     let resolved = resolve_config(config.as_deref())?;
-    // Detect format from the source URI (strip file: prefix, check extension).
-    let source = &resolved.source;
-    let path_part = source.strip_prefix("file:").unwrap_or(source);
-    let must_use_json = path_part.ends_with(".json");
-    let mut engine_cfg = if must_use_json {
-        OtelDataflowSpec::from_json(&resolved.content)?
-    } else {
-        OtelDataflowSpec::from_yaml(&resolved.content)?
+    let mut engine_cfg = match resolved.format {
+        ConfigFormat::Json => OtelDataflowSpec::from_json(&resolved.content)?,
+        ConfigFormat::Yaml => OtelDataflowSpec::from_yaml(&resolved.content)?,
     };
     apply_cli_overrides(&mut engine_cfg, num_cores, core_id_range, http_admin_bind);
 


rust/otap-dataflow/crates/config/src/config_provider.rs
```diff
@@ -0,0 +1,320 @@
...
+    let colon_pos = uri.find(':')?;
+
+    // If a path separator appears before the colon, treat as bare path.
+    let before_colon = &uri[..colon_pos];
+    if before_colon.contains('/') || before_colon.contains('\\') {
+        return None;
+    }
+
+    Some(before_colon)
+}
```
lquerel, line 216: This code will not recognize Windows path. We need a more OS-independent approach.

rust/otap-dataflow/crates/config/src/config_provider.rs
```diff
@@ -0,0 +1,320 @@
...
+/// A provider that can resolve configuration content from a URI with a specific scheme.
+pub trait ConfigProvider {
+    /// The URI scheme this provider handles (e.g. "file", "env").
+    fn scheme(&self) -> &str;
+
+    /// Resolve the given URI to configuration content as a string.
+    ///
+    /// The `uri` is the full original URI string (including the scheme prefix).
+    /// Implementations should strip their own scheme prefix.
+    fn resolve(&self, uri: &str) -> Result<String, Error>;
```
lquerel: Instead of returning a String, that be nice to return a struct containing both the scheme and the content.

rust/otap-dataflow/crates/config/src/config_provider.rs
```diff
@@ -0,0 +1,320 @@
...
+
+use crate::error::Error;
+use std::path::Path;
+
+/// Well-known config file paths tried when `--config` is omitted.
+const WELL_KNOWN_PATHS: &[&str] = &[
+    "/etc/o11y-gateway/config.yaml",
+    "/etc/data-plane/configs/pipeline_config.yaml",
+    "/app/config.yaml",
+];
```
lquerel: I know these paths are in the original GitHub issue, but I don't think they correspond to anything at the moment. So I suggest removing them, with the possible exception of `config.yaml` in the current directory.

rust/otap-dataflow/crates/config/src/config_provider.rs
mateenali66: fixed in 98775d3 -- replaced the manual prefix checks with `Path::is_absolute()` for cross-platform absolute path detection, and added a guard for single-char alphabetic schemes (Windows drive letters like `C:`). added test cases for both `C:\\config.yaml` and `D:/config.yaml`.

rust/otap-dataflow/crates/config/src/config_provider.rs
mateenali66: done in 98775d3 -- `ConfigProvider::resolve` now returns `ResolvedConfig { source: String, content: String }` where `source` is the original URI string. main.rs uses `resolved.source` to detect JSON vs YAML (checks if the path part ends with `.json`) without re-parsing the URI a second time.

rust/otap-dataflow/crates/config/src/config_provider.rs
mateenali66: good point, the /etc/ paths were from the issue spec but there is nothing actually deployed there yet. simplified in 98775d3 to just `config.yaml` in the current working directory as the single fallback, which is more practical and consistent with how most CLI tools behave.

lquerel:
@mateenali66 You should also document that in `src/README.md` and `configs/README.md`.

The `clap` help text could also be updated.

And finally, there are probably some updates to make in `docs/configuration-model.md` and `crates/config/README.md`.

mateenali66:
updated docs in f638d10, added URI scheme reference tables and examples to src/README.md, configs/README.md, docs/configuration-model.md, and crates/config/README.md. also documented resolve_config and ResolvedConfig in the config crate README. clap help was already updated in the earlier commit (the --config arg description and after_long_help examples show all three URI forms).

rust/otap-dataflow/src/main.rs
```diff
@@ -306,13 +308,22 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     println!("{}", system_info());
 
-    let mut engine_cfg = OtelDataflowSpec::from_file(&config)?;
+    let resolved = resolve_config(config.as_deref())?;
+    // Detect format from the source URI (strip file: prefix, check extension).
+    let source = &resolved.source;
+    let path_part = source.strip_prefix("file:").unwrap_or(source);
+    let must_use_json = path_part.ends_with(".json");
```
lquerel: `ResolveConfig` should expose an enum defining how the content should be interpreted.

lquerel (changes requested):

mateenali66:
added ConfigFormat enum (Yaml | Json) to ResolvedConfig in 1b56b0d.

FileConfigProvider now sets format from file extension (.json -> Json, everything else -> Yaml). EnvConfigProvider defaults to Yaml since env vars almost always carry YAML content.

main.rs dispatches via `match resolved.format` now, no more re-parsing the source URI to check the extension inline.

jmacd:
Sorry, I did look at this PR. LGTM Thank you @mateenali66 and @lquerel.

PR #2530, opened 2026-04-02
feat(temporal_reaggregation_processor): Pass through metrics which cannot be aggregated

JakeDern:
# Change Summary

This PR contains the necessary plumbing to construct a "passthrough" batch and communicate the result of processing a view back up to the main processing loop.

Basically, for each batch we either:

- Aggregate nothing and need to pass the data through as-is
- Aggregate _some_ of the metrics and need to pass through what was not aggregated instead of throwing it away
- Aggregate all of the metrics in the batch

To support this we needed:

- Exemplar support for the passthrough batch (still doesn't exist for the aggregated batch)
- A second set of state for building the passthrough batch

* Part of #2422 

Unit.

No

### DIFF
--- a/.github/workflows/dataflow-engine-binary-size.yml
+++ b/.github/workflows/dataflow-engine-binary-size.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ${{ matrix.platform.runner }}
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
+        uses: step-security/harden-runner@fe104658747b27e96e4f7e80cd0a94068e53901d # v2.16.1
         with:
           egress-policy: audit
 
@@ -88,7 +88,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Download all size reports
-        uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3 # v8.0.0
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           pattern: size-report-*
           merge-multiple: true
@@ -110,7 +110,7 @@ jobs:
           done
 
       - name: Update binary size benchmarks
-        uses: benchmark-action/github-action-benchmark@4bdcce38c94cec68da58d012ac24b7b1155efe8b # v1.20.7
+        uses: benchmark-action/github-action-benchmark@a60cea5bc7b49e15c1f58f411161f99e0df48372 # v1.22.0
         with:
           tool: "customSmallerIsBetter"
           output-file-path: consolidated-sizes.json
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
-      - uses: fossas/fossa-action@c414b9ad82eaad041e47a7cf62a4f02411f427a0 # v1.8.0
+      - uses: fossas/fossa-action@ff70fe9fe17cbd2040648f1c45e8ec4e4884dcf3 # v1.9.0
         with:
           api-key: ${{secrets.FOSSA_API_KEY}}
           team: OpenTelemetry
--- a/.github/workflows/go-ci.yml
+++ b/.github/workflows/go-ci.yml
@@ -29,9 +29,9 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           submodules: true
-      - uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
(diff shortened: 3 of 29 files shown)


rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
```diff
@@ -689,28 +837,139 @@ impl TemporalReaggregationProcessor {
...
+        return Err(ProcessPdataError::IdOverflow);
+    }
+    *counter += 1;
+    Ok(id)
+}
 
-    match data_type {
+/// Lightweight check on a [`DataView`] to determine if it represents an
+/// aggregatable metric type. This avoids computing a full [`MetricId`] which
+/// requires scope/resource context.
```
jmacd, line 941: 👍

jmacd (approved):

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/builder.rs
```diff
@@ -207,63 +269,82 @@ impl MetricSignalBuilder {
...
+        self.ehdp_exemplar_attrs = AttributesRecordBatchBuilder::new();
     }
 
     /// Append resource attributes for a newly seen resource.
     pub fn append_resource<R: ResourceView>(&mut self, id: u16, view: &R) {
         for attr in view.attributes() {
             self.resource_attrs.append_parent_id(&id);
-            let _ = append_attribute_value(&mut self.resource_attrs, &attr);
+            if let Err(e) = append_attribute_value(&mut self.resource_attrs, &attr) {
+                otel_warn!(ATTRIBUTE_ENCODE_FAILED_EVENT, error = %e);
```
lquerel, line 285: I think having a warning message for each attribute is risky, and it could be exploited to DoS the system if the log level is set to warn. I propose that we:

1. Move the warning statement outside the loop.  
2. Inside the loop:  
   - Count the number of errors  
   - Record the first 5 errors if the log level is debug  

3. After the loop, if the number of errors is greater than 0:  
   - Emit a warning message with the total number of errors  
   - If the log level is debug, emit the recorded errors

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
```diff
@@ -284,34 +318,46 @@ impl TemporalReaggregationProcessor {
...
+                    .map_err(|e| ProcessPdataError::ViewCreationFailed { inner: Some(e) })?;
+
+                if !has_aggregatable_metrics(&view) {
+                    return Ok(ProcessResult::NoAggregations);
+                }
+
+                match self.process_view(&view)? {
+                    Some(records) => Ok(ProcessResult::SomeAggregations(records)),
+                    None => Ok(ProcessResult::AllAggregated),
+                }
```
lquerel, line 346: Can't we make this cheaper by folding `has_aggregatable_metrics()` into `process_view()` instead of doing a full preflight scan first? Right now any batch that contains at least one aggregatable metric gets walked twice: once to answer "is there anything aggregatable here?" and then again to actually process it.

lquerel (commented):
I think my comment about logging is important to take into account in this PR.

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
JakeDern: This is to avoid building a passthrough batch, which is expensive.

As we walk the `view` during processing, for every metric we append to either the passthrough batch or the aggregating batch.

If we move this check to that loop and there is nothing that is aggregable, we will end up appending every metric to the passthrough batch and therefore produce a full copy of the input when we could have just sent the original batch through.

I think you could argue that if almost every batch has something aggregable then maybe it's better to move the check to the loop and throw the work away if we get to the end, but I'd argue that I want to see this show up on the profile first 🙂

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/builder.rs
JakeDern: Yeah good point, I wasn't thinking about this as a DoS vector. Will make that update, but have some further thoughts on this.

If you look at why this operation is even fallible at all, it's only because CBOR encoding can fail if the data was not validly encoded in the first place. 

This is one of those things that I sort of wish was checked once on ingest and would let us eliminate the fallibility from the API entirely by the time we get to this point. Things like attribute keys not being valid UTF-8 is another one that can cause problems.

My worry with letting junk data like this get through admission and into our pipelines is that if no component checks it along the way, we might just pass it through and that becomes another DoS vector. A client can send us bad CBOR or UTF-8 data and I don't know there's a guarantee we'll check, so whatever destination we're forwarding to may reject the data instead and therefore cause us to retry, propagate nacks, all the bad stuff.

JakeDern: Any objection to making this update in the next PR? I have a couple of error handling and telemetry related changes to work on that I could lump this in with

lquerel: No objection at all

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
lquerel: Fair enough :-)

PR #2532, opened 2026-04-03
perf: syslog_cef_receiver avoid windowing in cef_with_rfc3164 path

kamilon:
# Change Summary

Remove window scanning in cef_with_rfc3164 path. Instead use pure forward scanning. 50% improvement in bench.

107.40 ns -> 50.547 ns

No issue filed.

cargo test and cargo bench

No.

### DIFF
--- a/.github/workflows/dataflow-engine-binary-size.yml
+++ b/.github/workflows/dataflow-engine-binary-size.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ${{ matrix.platform.runner }}
     steps:
       - name: Harden the runner (Audit all outbound calls)
-        uses: step-security/harden-runner@a90bcbc6539c36a85cdfeb73f7e2f433735f215b # v2.15.0
+        uses: step-security/harden-runner@fe104658747b27e96e4f7e80cd0a94068e53901d # v2.16.1
         with:
           egress-policy: audit
 
@@ -88,7 +88,7 @@ jobs:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
       - name: Download all size reports
-        uses: actions/download-artifact@70fc10c6e5e1ce46ad2ea6f2b72d43f7d47b13c3 # v8.0.0
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           pattern: size-report-*
           merge-multiple: true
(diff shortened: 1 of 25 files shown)


rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/mod.rs
```diff
@@ -109,30 +109,19 @@ fn try_rfc3164_cef<'a>(
...
-                        let mut modified_rfc3164 = rfc3164_msg;
-                        modified_rfc3164.content = Some(cef_message);
-                        return Ok(ParsedSyslogMessage::CefWithRfc3164(
-                            modified_rfc3164,
-                            cef_msg,
-                        ));
-                    }
-                }
+    if let Some(tag) = rfc3164_msg.tag {
+        if tag == b"CEF" && rfc3164_msg.content.is_some() {
```
lalitb: The old code only entered this block when a hostname was present, so `tag=CEF` messages without a hostname used to fall back to plain `Rfc3164`. This change now parses those as `CefWithRfc3164`. That seems reasonable, but it is a behavior change, not just a perf change, so please add a small test for that case.

jmacd: @kamilon let's please add a new test based on @lalitb's observation.

utpilla: nit: You could flatten the nesting by using a single `if` with a let chain:

```rust
if let Some(tag) = rfc3164_msg.tag
    && tag == b"CEF"
    && rfc3164_msg.content.is_some()
{
    // ...
}
```

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/mod.rs
```diff
@@ -109,30 +109,19 @@ fn try_rfc3164_cef<'a>(
...
-                let search_start = hostname_pos + hostname.len();
-                let after_hostname = &input[search_start..];
-
-                // Find "CEF:" in the remaining input
-                if let Some(cef_pos) = after_hostname.windows(4).position(|w| w == b"CEF:") {
-                    // The CEF message starts at this position and goes to the end
-                    let cef_message = &after_hostname[cef_pos..];
-
-                    if let Ok(cef_msg) = parse_cef(cef_message) {
-                        // Update the rfc3164 content to point to the full CEF message
```
utpilla, line 127: No need to remove this comment. Could you add this back?

rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/parser/mod.rs
kamilon: Done.

PR #2535, opened 2026-04-03
azure-monitor-exporter: enforce 50 MiB decompressed payload limit in GzipBatcher

gouslu:
# Change Summary

The Azure Monitor Logs Ingestion API enforces a server-side limit on the decompressed payload size (50 MiB). This limit is checked downstream after the gateway accepts the gzip-compressed request, causing ingestion failures for payloads that pass the 1 MiB compressed-size check but exceed 50 MiB when decompressed.

With highly compressible data (e.g., CEF syslog logs with repetitive templates), the GzipBatcher could easily pack 50+ MiB of decompressed JSON into a ~1 MiB compressed payload, triggering these downstream rejections.

Changes:
- Add `UNCOMPRESSED_LIMIT` (50 MiB) to cap decompressed payload size
- Track `total_uncompressed_size` precisely across flushes (every byte: `[`, commas, data, `]`)
- Finalize batch when cumulative uncompressed input approaches the limit
- Rename constants for readability: `COMPRESSED_LIMIT`, `GZIP_OVERHEAD`, `TARGET_COMPRESSED_LIMIT`
- Raise `MAX_GZIP_FLUSH_COUNT` from 100 to 128 (safety net; unreachable with 50 MiB limit — observed max is ~58)
- Add CEF syslog-style test data generator matching real PAN-OS/CEF workloads
- Add tests verifying decompressed output stays under `UNCOMPRESSED_LIMIT`
- Add test for identical (maximally compressible) entries proving byte-precise tracking

Resolves `PayloadSizeLimitExceeded` errors observed during CEF syslog load testing with highly compressible data patterns.

- `test_uncompressed_limit_enforced_with_identical_entries`: uses 3 rotating minimal JSON objects to verify decompressed output stays within 7 bytes of the 50 MiB limit with byte-precise tracking
- `test_uncompressed_limit_enforced_with_compressible_entries`: uses highly compressible JSON with small varying tails to verify the limit under realistic compression ratios
- `test_cef_syslog_batch_dimensions`: uses CEF syslog-style data matching real PAN-OS payloads to verify both compressed and decompressed dimensions
- All existing compressed-size, JSON-validity, and utilization tests continue to pass

Yes. Batches with highly compressible data will now be smaller (capped at 50 MiB decompressed instead of filling to the ~1 MiB compressed limit). This means more batches will be sent for the same volume of compressible data, but each batch will be accepted by the server without `PayloadSizeLimitExceeded` errors.

### DIFF
--- a/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/gzip_batcher.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/gzip_batcher.rs
@@ -7,6 +7,7 @@ use flate2::write::GzEncoder;
 use std::io::Write;
 
 use super::error::Error;
+
 /// Hard limit for gzip-compressed payload accepted by the Azure Monitor Gateway.
 const COMPRESSED_LIMIT: usize = 1024 * 1024;
 
@@ -19,15 +20,15 @@ const GZIP_OVERHEAD: usize = 4 * 1024;
 const TARGET_COMPRESSED_LIMIT: usize = COMPRESSED_LIMIT - GZIP_OVERHEAD;
 
 /// Hard limit for uncompressed (decompressed) payload size. The Azure Monitor
-/// Logs Ingestion API rejects payloads whose decompressed JSON exceeds 50 MiB
-/// (`MaxDcrChunkReadSizeInBytes`). The batcher finalizes the batch when the
-/// cumulative uncompressed input approaches this limit.
+/// Logs Ingestion API rejects payloads whose decompressed JSON exceeds 50 MiB.
+/// The batcher finalizes the batch when the cumulative uncompressed input
+/// approaches this limit.
 const UNCOMPRESSED_LIMIT: usize = 50 * 1024 * 1024;
 
 /// Maximum number of gzip sync flushes allowed per batch. Each flush adds a
 /// small amount of framing overhead. This cap prevents excessive overhead
-/// accumulation with highly compressible data. Extremely unlikely to reach this limit
-/// due to the uncompressed size limit.
+/// accumulation with highly compressible data. Extremely unlikely to reach
+/// this limit due to the uncompressed size limit.
 const MAX_GZIP_FLUSH_COUNT: usize = 128;
 
 /// Accumulates JSON entries into gzip-compressed batches that stay under a size limit.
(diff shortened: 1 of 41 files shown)


rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/gzip_batcher.rs
```diff
@@ -619,6 +640,216 @@ mod tests {
...
+                });
+
+                let ratio = stats.decompressed_size as f64 / stats.size as f64;
+                let label = format!("cef_syslog/{msg_size}B/lvl{level}");
+                eprintln!(
+                    "{label:<45} {:<10} {:<10} {:<8} {:<8.1} {:<8}",
+                    stats.size, stats.decompressed_size, stats.row_count, ratio, stats.flush_count
+                );
+
+                assert!(
```
lalitb, line 845: This checks the 1 MiB compressed limit, but not the new 50 MiB decompressed limit. Can we also assert `stats.decompressed_size <= UNCOMPRESSED_LIMIT` here?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/gzip_batcher.rs
```diff
@@ -556,9 +572,14 @@ mod tests {
             }
         }
         let batch = batcher.take_pending_batch().unwrap();
+        let mut decoder = GzDecoder::new(&batch.compressed_data[..]);
+        let mut decompressed = Vec::new();
+        _ = decoder.read_to_end(&mut decompressed).unwrap_or(0);
```
lalitb: `unwrap_or(0)` hides gzip decode errors here. Would it be better to fail the test on decode failure instead of silently reporting a decompressed size of 0?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/gzip_batcher.rs
```diff
@@ -129,16 +139,18 @@ impl GzipBatcher {
...
         // Recompute after flush: uncompressed_size was reset so
         // next_size must be recalculated with current state.
         let next_size =
             self.uncompressed_size + structural_overhead + data.len() + finalize_overhead;
-        let must_finalize =
-            next_size > self.remaining_size || self.flush_count >= MAX_GZIP_FLUSH_COUNT;
+        let must_finalize = next_size > self.remaining_size
+            || self.total_uncompressed_size + structural_overhead + data.len() + finalize_overhead
+                >= UNCOMPRESSED_LIMIT
+            || self.flush_count >= MAX_GZIP_FLUSH_COUNT;
```
utpilla: nit: You could simplify this by computing `structural_overhead + data.len() + finalize_overhead` only once and storing it in a variable. It would also make the intent clearer.

```rust
let entry_cost = structural_overhead + data.len() + finalize_overhead;
let next_size = self.uncompressed_size + entry_cost;
let must_finalize = next_size > self.remaining_size
    || self.total_uncompressed_size + entry_cost >= UNCOMPRESSED_LIMIT
    || self.flush_count >= MAX_GZIP_FLUSH_COUNT;
```

PR #2537, opened 2026-04-03
feat(crypto): add support for crypto-native-tls feature and update reqwest dependencies

### LINKED ISSUE #2536: Support platform-native TLS backend (SChannel on Windows, OpenSSL on Linux) via `crypto-native-tls` feature
### Pre-filing checklist

### Component(s)

### Problem

PR #2269 introduced pluggable rustls crypto providers via mutually exclusive feature flags: `crypto-ring`, `crypto-aws-lc`, `crypto-openssl`. All three use rustls as the TLS implementation and differ only in the underlying crypto provider.

This works well for the standalone `df_engine` binary, but creates significant build complexity when attempting to link `otap-df-otap` as a library into a C++ host process esp. across multiple platforms. e.g.:

- **`crypto-aws-lc` fails on Windows aarch64**: `aws-lc-sys`'s GAS-syntax `.S` assembly files are silently ignored by `cl.exe`, producing missing object files (LNK1181).
- **`crypto-aws-lc` on Windows x86\_64** requires `/force:guardehcont` linker workarounds because NASM-assembled objects lack `/guard:ehcont` metadata.
- **`crypto-ring`** works on all platforms but still bundles its own crypto library (ring compiles C/asm from source), adding build complexity and binary size.
- **`crypto-openssl`** requires OpenSSL on Windows (not a system library shipped in Windows).

### Proposed solution

Add a new `crypto-native-tls` feature flag that uses the platform's native TLS stack instead of rustls + a compiled crypto provider:

| Platform | Backend |
|---|---|
| Windows | SChannel (built-in, zero dependencies) |
| Linux | OpenSSL (system library) |
| macOS | Security.framework (built-in) |

This is the same pattern used by `geneva-uploader` and the Azure SDK's `reqwest 0.12` dependency already in otel-arrow's dependency tree.

### Scope

- Add `crypto-native-tls` feature to `otap-df-otap` and the workspace root
- When enabled, `reqwest` uses its `native-tls` feature instead of `rustls-no-provider`
- `install_crypto_provider()` becomes a no-op (no rustls `CryptoProvider` needed)
- The HTTP client builder selects `.use_native_tls()` instead of `.use_rustls_tls()`
- Add CI check for the new feature

### Limitations

- `crypto-native-tls` covers **client-side HTTP/HTTPS** connections only (reqwest-based: OTLP HTTP exporter, Azure Monitor exporter, Geneva uploader)
- **Server-side gRPC TLS** (`experimental-tls` feature) still requires a rustls crypto provider (`crypto-ring`, `crypto-aws-lc`, or `crypto-openssl`). tonic 0.14 does not support native-tls backends. For the embedding use case this is acceptable since the host process handles networking.
- The `default` feature set remains `crypto-ring`. There should be no breaking changes for existing users.

### Additional Context

_No response_

AaronRM:
# Change Summary

Add a new `crypto-native-tls` feature flag that uses the platform's native TLS stack (SChannel on Windows, OpenSSL on Linux, Security.framework on macOS) instead of rustls. This follows the pluggable crypto architecture from PR #2269
and avoids cross-compilation issues with C/asm crypto providers on Windows.

* Closes #2536 

- `cargo test --no-default-features --features crypto-native-tls --workspace` passes (591 tests, same as default features)
- CI step added to `no_default_features_check` job

Yes: new `crypto-native-tls` feature flag available for embedders who want platform-native TLS without compiling a crypto library from source. Default behavior is unchanged (`crypto-ring`). Note that `crypto-native-tls` covers client-side HTTP/HTTPS connections only (reqwest); server-side gRPC TLS (`experimental-tls`) still requires a rustls-based crypto feature.

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/crypto.rs
+++ b/rust/otap-dataflow/crates/otap/src/crypto.rs
@@ -7,11 +7,12 @@
 //! [`CryptoProvider`](rustls::crypto::CryptoProvider) based on compile-time
 //! feature flags.  Exactly one of the `crypto-*` features must be enabled:
 //!
-//! | Feature          | Backend          | Use-case                           |
-//! |------------------|------------------|------------------------------------|
-//! | `crypto-ring`    | `ring`           | Default, backward-compatible       |
-//! | `crypto-aws-lc`  | `aws-lc-rs`      | AWS environments, broader algos    |
-//! | `crypto-openssl` | `rustls-openssl`  | Regulated / FIPS environments      |
+//! | Feature             | Backend          | Use-case                               |
+//! |---------------------|------------------|----------------------------------------|
+//! | `crypto-ring`       | `ring`           | Default, backward-compatible           |
+//! | `crypto-aws-lc`     | `aws-lc-rs`      | AWS environments, broader algos        |
+//! | `crypto-openssl`    | `rustls-openssl` | Regulated / FIPS environments          |
+//! | `crypto-native-tls` | `native-tls`     | Platform-native TLS (SChannel/OpenSSL) |
 
 /// Installs the selected rustls `CryptoProvider` as the process-wide default.
 ///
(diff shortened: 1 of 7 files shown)


AaronRM:
@lalitb and @pritishnahar95 have some context in this area and may have feedback

jmacd (approved):

lquerel (approved):
LGTM

rust/otap-dataflow/crates/otap/src/crypto.rs
```diff
@@ -51,16 +52,29 @@ pub fn install_crypto_provider() -> Result<(), String> {
             .map_err(|_| "crypto provider already installed (openssl)".to_string())?;
     }
 
+    #[cfg(all(
+        feature = "crypto-native-tls",
+        not(feature = "crypto-ring"),
+        not(feature = "crypto-aws-lc"),
+        not(feature = "crypto-openssl")
+    ))]
+    {
```
lalitb, line 61: This looks unsafe with `experimental-tls`: `crypto-native-tls` skips rustls provider installation here, but `experimental-tls` still uses rustls-based code paths that require a default provider. That means the build can succeed and then fail at runtime. Can we disallow that combination or still install a rustls provider in that case?

AaronRM: Setting this back to draft for the time being. Will revisit tomorrow.

lalitb: `experimental-tls` flag has been removed, and TLS is always available now. So , this comment mayn't be valid anymore.

jmacd:
@AaronRM Now I understand what this is doing. Thank you.
#2674

jmacd:
I will re-open a PR, thank you @AaronRM

PR #2510, opened 2026-04-01
docs: Add extension system architecture document

gouslu:
# Change Summary

Adds the Phase 1 architecture document for the extension system ([`rust/otap-dataflow/docs/extension-system-architecture.md`](https://github.com/gouslu/otel-arrow/blob/gouslu/extension-system-doc/rust/otap-dataflow/docs/extension-system-architecture.md)), describing how the [extension system proposal](https://github.com/gouslu/otel-arrow/blob/gouslu/extension-system-doc/rust/otap-dataflow/docs/extension-requirements.md) is implemented.

A working PoC is available on the [PoC branch](https://github.com/gouslu/otel-arrow/tree/gouslu/extension-system-p1-local-shared).

### What this document covers

The document maps each proposal requirement to a concrete Phase 1 implementation approach and introduces two additions beyond the proposal:

- **Active/Passive lifecycle distinction** — type-level enforcement via `Active<E>` / `Passive<E>` wrappers. Passive extensions provide capabilities without spawning a task or allocating control channels.
- **Shared-to-local transparent fallback** — a shared-only extension can serve local consumers automatically via a `SharedAsLocal` adapter. Extension authors who need lock-free local performance can still opt in with a dedicated local variant.

It also covers:
- Ownership model — `Rc` for local (no locks), `Box<Clone+Send>` with `Arc`-wrapped fields for shared (similar to tokio/axum patterns)
- `#[capability]` proc macro — generates local/shared trait variants, sealed impls, registration structs, and coercion functions from a single trait definition
- Type-safe capability resolution — `require_local()` / `require_shared()` with compile-time sealing and 4-step binding validation
- Core types — `ExtensionWrapper`, `ExtensionFactory`, `CapabilityRegistry`, builder pattern, and lifecycle management
- Configuration — YAML schema for `extensions` section and `capabilities` bindings on nodes
- Pipeline lifecycle — extension startup ordering, shutdown sequencing, unused variant cleanup
- Future vision — hierarchical scopes (Phase 2), WASM extensions (Phase 3)
- Phase 1 migration plan — 8 incremental PRs

### Phase 1 — Migration Plan (PRs 1-8)

[...]

PR #2538, opened 2026-04-03
feat(config): extensions section, capabilities bindings, and config-level validation

gouslu:
# Change Summary

This is PR 1 based on https://github.com/open-telemetry/otel-arrow/pull/2510.

Add config parsing for pipeline extensions — the first building block of the extension system ([architecture doc](https://github.com/gouslu/otel-arrow/blob/gouslu/extension-system-doc/rust/otap-dataflow/docs/extension-system-architecture.md)).

Extensions are standalone pipeline components that provide shared capabilities (auth, storage, etc.) to data-path nodes. This PR adds config-level support only — no runtime behavior. Extensions are parsed and validated but not created or started.

**Config changes:**
- `extensions:` section in pipeline config, separate from `nodes:`
- `capabilities:` field on nodes for binding capabilities to extension instances
- `NodeKind::Extension` variant for URN parsing (e.g., `urn:otel:extension:auth`)
- `ExtensionConfig` struct in engine config
- `PipelineConfigBuilder::add_extension()` for programmatic config construction

**Validations added:**
- Capability bindings must reference extensions that exist in the `extensions:` section
- Extensions cannot have `capabilities:` bindings (they provide capabilities, not consume them)
- Extension URN in `nodes:` section → `ExtensionInNodesSection` error
- `extensions:` at group level rejected by serde (`deny_unknown_fields`)
- Duplicate names across nodes and extensions rejected

**Example config:**

    extensions:
      azure_auth:
        type: "urn:microsoft:extension:azure_identity_auth"
        config:
          method: "managed_identity"

    nodes:
      exporter:
        type: "urn:microsoft:exporter:azure_monitor"
        capabilities:
          bearer_token_provider: "azure_auth"

* Part of the extension system initiative (Phase 1, PR 1 of 8)

13 new config tests covering:
- Extensions parsed separately from nodes
- Extension with config and capability bindings
- Extension URN kind detection
- Duplicate extension/node name rejection
- Extension URN in nodes section detection
- Non-extension URN in extensions section detection
- Empty/missing extensions section
- Empty capabilities
- Multiple extensions with capability bindings
- Capability binding to nonexistent extension (rejected)
- Capability binding to existing extension (passes)
- Capabilities on extension itself (rejected)
- Extensions at group level (rejected by serde)

All 168 config tests pass. Example config YAML added to `configs/`.

Yes — pipeline YAML configs now support `extensions:` and `capabilities:` sections. These are optional and have no runtime effect in this PR.

### DIFF
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -2,8 +2,10 @@ coverage:
   status:
     project: # Overall repo coverage
       default:
-        target: auto # Increase
-        threshold: 0.1% # Allow small fluctuation
+        target: 85%
+    patch: # Changed lines in PR
+      default:
+        target: 70%
 
 ignore:
   - "rust/otap-dataflow/crates/quiver-e2e/**" # Test tool, not production code
@@ -25,10 +27,6 @@ component_management:
       name: "query_engine"
       paths:
         - "rust/experimental/query_engine/"
-    - component_id: "syslog_cef_receivers"
-      name: "syslog_cef_receivers"
-      paths:
-        - "rust/experimental/syslog-cef-receivers/"
     - component_id: "otel_arrow_go"
       name: "otel-arrow-go"
       paths:
--- a/.github/workflows/docker-build-check.yml
+++ b/.github/workflows/docker-build-check.yml
@@ -1,0 +1,47 @@
+name: Docker Build Check
+permissions:
+  contents: read
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - 'rust/otap-dataflow/Dockerfile'
+      - 'rust/otap-dataflow/cross-arch-build.sh'
+      - 'rust/otap-dataflow/Cargo.toml'
+      - 'rust/otap-dataflow/Cargo.lock'
+      - 'rust/otap-dataflow/src/**'
+      - 'rust/otap-dataflow/crates/**'
+      - 'THIRD_PARTY_NOTICES.txt'
+  merge_group:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  docker-build:
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@fe104658747b27e96e4f7e80cd0a94068e53901d # v2.16.1
+        with:
+          egress-policy: audit
+
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          submodules: true
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
+
+      - name: Build Docker image
+        run: |
+          cd rust/otap-dataflow
+          docker buildx build \
+            --build-context otel-arrow=../../ \
+            -f Dockerfile \
+            -t df_engine:check \
+            --load \
+            .
--- a/README.md
+++ b/README.md
@@ -269,17 +269,17 @@ repository](https://github.com/open-telemetry/community/blob/main/guides/contrib
 For more information about the approver role, see the [community
 repository](https://github.com/open-telemetry/community/blob/main/guides/contributor/membership.md#approver).
 
-### Emeritus Approvers
-
-- [Alex Boten](https://github.com/codeboten)
-- [Moh Osman](https://github.com/moh-osman3)
-
 ### Triagers
 
 - [Tom Tan](https://github.com/ThomsonTan), Microsoft
 
 For more information about the approver role, see the [community
 repository](https://github.com/open-telemetry/community/blob/main/guides/contributor/membership.md#triager).
+
+### Emeritus
+
+- [Alex Boten](https://github.com/codeboten), Approver
+- [Moh Osman](https://github.com/moh-osman3), Approver
 
 ### Thanks to all of our contributors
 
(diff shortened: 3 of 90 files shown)


kamilon:
The architecture doc link doesn't work. Can you please fix that? I'd like to understand what this is being built against.

jmacd:
I believe this is ready to review because this PR1 is consistent with the original extensions design document. If there are objections to #2510, let them surface now! I will merge that PR before I merge this one, but both are ready to review.

rust/otap-dataflow/crates/config/src/node.rs
```diff
@@ -58,6 +96,23 @@ pub struct NodeUserConfig {
...
+    /// ```yaml
+    /// capabilities:
+    ///   bearer_token_provider: azure_auth
+    /// ```
+    #[serde(
+        default,
+        skip_serializing_if = "HashMap::is_empty",
+        deserialize_with = "deserialize_no_dup_keys"
+    )]
+    pub capabilities: HashMap<String, String>,
```
jmacd: I have a preference for adding type  names instead of Strings, here:

- the Key is a Capability identifier, I would call it CapabilityId, have it a `Cow<'static, str`  like NodeId, see crates/config/src/lib.rs; somewhere, there is a constant caled "key_value_store", I would make it a CapabilityId
- the Value is a NodeId, the node-name identifier type

The `HashMap<CapabilityId, NodeId>` is OK without a named type.

gouslu: I'm looking into it!

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -2502,4 +2612,368 @@ sink:
...
+            .unwrap();
+        assert_eq!(auth_cfg.kind(), NodeKind::Extension);
+        // Any stricter enforcement belongs to downstream engine/runtime validation.
+    }
+
+    #[test]
+    fn test_receiver_urn_in_extensions_section_parsed() {
+        // A receiver URN placed in the `extensions:` section parses
+        // but has receiver kind — the engine should reject this at
+        // build time since the extension factory won't find it.
```
utpilla: This test asserts that a receiver URN in extensions: passes, with a comment saying the engine should reject it at build time. However, nothing currently rejects this end-to-end.

The URN kind check is purely structural and doesn't need the factory registry. It could be added to `validate_capability_bindings`. Just wanted to bring this to attention, you might have planned it for a follow-up PR.

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -631,13 +664,52 @@ impl PipelineConfig {
...
+            }
+        }
+
+        // Check that extensions don't have capability bindings
+        for (ext_id, ext_config) in self.extensions.iter() {
+            if !ext_config.capabilities.is_empty() {
+                errors.push(Error::InvalidUserConfig {
+                    error: format!(
+                        "Extension '{}' has a `capabilities` section, but extensions \
+                         provide capabilities — they don't consume them. \
```
utpilla: nit: The em dash (`—`) in this error message is non-ASCII. It's safest to keep error/log strings ASCII-only so they work reliably with any log storage, parser, or terminal locale.

rust/otap-dataflow/crates/config/src/node.rs
```diff
@@ -60,6 +100,23 @@ pub struct NodeUserConfig {
...
+    /// ```yaml
+    /// capabilities:
+    ///   bearer_token_provider: azure_auth
+    /// ```
+    #[serde(
+        default,
+        skip_serializing_if = "HashMap::is_empty",
+        deserialize_with = "deserialize_no_dup_keys"
+    )]
+    pub capabilities: HashMap<CapabilityId, NodeId>,
```
utpilla, line 118: This enforces one extension per capability per node. Is this intentional? The design doc doesn't state this explicitly.

If an exporter needs `header_setter` functionality from two independent extensions (e.g., one for correlation headers, one for custom metadata headers), the only workaround would be for the user to write a custom extension that combines both. That seems like a heavy ask for what might be a common use case.

jmacd: I noticed this, at some point. I feel that if there are more than one capability, we should follow the Collectr's naming convention and use names like bearer_token/left bearer_token/right if you need more than one. I figure we should tackle this when it happens.

utpilla: I'm okay with tackling this later. However, one thing to keep in mind: the design doc states that capabilities are "known by the engine version and validated during configuration loading," and that external extensions cannot introduce new capability interfaces. This is different from the Go Collector, where any extension can expose any interface and components look them up by ID at runtime.

So `bearer_token/left` and `bearer_token/right` would each need to be pre-registered as known capabilities in the engine core. It's not purely a user naming convention like it would be in the Go Collector. Each new variant requires an engine core change, so this may not be as straightforward to address when the time comes. Just wanted to flag this so it's not overlooked.

rust/otap-dataflow/crates/config/src/pipeline.rs
gouslu: this is such an AI problem, AI loves the em dash, I can't stop it from using it. lol

rust/otap-dataflow/crates/config/src/node.rs
gouslu: Depends on how you resolve them. `/left` and `/right` might just mean that there are 2 different instances and that you get provided with a map hat has both in it or something like that. What would need to be known is the capability id basically, same as today.

rust/otap-dataflow/crates/config/src/pipeline.rs
gouslu: thanks, yeah this would come with a following PR, but adding it now makes sense too.

PR #2539, opened 2026-04-04
[otap-dataflow] add transport header infrastructure

### LINKED ISSUE #2508: Support policy-based transport header extraction and propagation in `OtapPdata`
### Pre-filing checklist

### Component(s)

### Objective

Introduce a policy-based transport header capability that allows header-capable receivers to extract selected inbound transport headers into `OtapPdata` context, and allows header-capable exporters to propagate selected stored headers on egress.

### Rationale

Protocols such as OTLP and OTAP can carry important request-scoped metadata outside the payload itself, for example in HTTP headers or gRPC metadata. That metadata can represent trace context, tenant, auth, routing, correlation, or policy hints that need to survive the pipeline.

`OtapPdata` already has context, but it does not currently preserve transport headers. This makes end-to-end propagation and policy-driven use of request metadata difficult.

A shared policy-based model is the preferred direction because multiple receivers and exporters are expected to need the same capability. Receivers and exporters can still expose node-specific config (by overriding the policy locally), but that config should primarily select or override shared transport-header policy rules rather than
redefine the feature independently at every node.

### Scope

Add a new `transport_headers` policy family and a protocol-neutral transport header abstraction in `OtapPdata` context.

The transport header abstraction should preserve the semantics needed across protocols:

- duplicate header names must be preserved
- matching should use a normalized logical name
- the original wire name should remain available
- values should support both `text` and `binary`
- values should be stored as raw bytes, not as protocol-specific header objects

A conceptual `TransportHeader` entry should contain:

- `name`: normalized logical name used for matching and policy lookup
- `wire_name`: original header or metadata name observed on ingress
- `value_kind`: `text` or `binary`
- `value`: raw bytes

The new `transport_headers` policy family should define:

- extraction rules
- propagation rules
- shared limits and failure handling

Receivers and exporters should activate policy rules from node config. A receiver should be able to opt into extraction rules, and an exporter should be able to opt into propagation rules. Processors should be transport-header transparent unless they explicitly add support for reading or mutating this context.

The design should be protocol-neutral. OTLP/gRPC and OTAP/gRPC are suitable initial targets, but the abstraction should also cover other header-capable protocols.

### Acceptance Criteria

- `OtapPdata` can carry transport headers in context without losing duplicate-name or binary-value semantics.
- A shared `transport_headers` policy family exists in the config model.
- Receivers can activate extraction rules from policy through node config.
- Exporters can activate propagation rules from policy through node config.
- A normal in-pipeline path preserves transport headers through at least one processor.
- An end-to-end pipeline demonstrates extraction on `receiver:otlp` an `receiver:otap`, preservation through a basic processor, and propagation on `exporter:otap`.
- Extraction and propagation are explicit and opt-in. The default behavior is not to forward all inbound headers.
- Tests cover extraction, propagation, filtering, limits, and invalid-header handling.

### Dependencies or Blockers

_No response_

### Additional Context

Proposed policy shape:

```yaml
version: otel_dataflow/v1

groups:
  default:
    pipelines:
      ingest:
        policies:
          header_capture:
            defaults:
              max_entries: 32         # default: 32 captured headers per message
              max_name_bytes: 128     # default: 128 bytes per header name
              max_value_bytes: 4096   # default: 4096 bytes per header value
              on_error: drop          # default: drop the offending captured header

            # Per-header defaults:
            # - store_as: first matched name, normalized
            # - value_kind: text unless protocol-specific inference marks it binary
            # - sensitive: false
            headers:
              - match_names: ["x-tenant-id"]
                store_as: tenant_id

              - match_names: ["x-request-id"]   // store_as default on the extracted header name

              - match_names: ["authorization"]
                sensitive: true

          header_propagation:
            default:
              selector: all_captured  # default: all_captured
              action: propagate       # default: propagate
              name: preserve          # default: preserve stored header name
              on_error: drop          # default: drop the offending outbound header

            # Per-override defaults:
            # - action: propagate
            # - name: preserve
            # - on_error: drop
            overrides:
              - match:
                  stored_names: ["authorization"]
                action: drop

        nodes:
          otlp_ingest:
            type: receiver:otlp
            config:
              protocols:
                grpc:

          otap_ingest:
            type: receiver:otap
            header_capture:    // node level override
              headers:
                - match_names: ["x-request-id"]
                  store_as: request_id
            config:
              listening_addr: "0.0.0.0:50051"

          batch:
            type: processor:batch
            config: {}

          otap_export:
            type: exporter:otap
            config:
              grpc_endpoint: "http://127.0.0.1:60051"

        connections:
          - from: otlp_ingest
            to: batch
          - from: otap_ingest
            to: batch
          - from: batch
            to: otap_export
```

Behavior in this example:
- otlp_ingest uses the pipeline header_capture policy and captures tenant_id, request_id, and authorization.
- otap_ingest overrides header_capture, so it only captures x-request-id and uses tighter limits.
- batch preserves the captured headers unchanged.
- otap_export propagates all captured headers by default.
- authorization is dropped on egress by the propagation override.

c1ly:
# Change Summary

Update the OtapPdata Context with the transport_headers field, so that OtapPdata can carry the headers through the pipeline.

Defined the TransportHeadersPolicy with HeaderCapturePolicy and HeaderPropagationPolicy, update the Policies to have a transport_headers field. Receiver and Exporter nodes can also define HeaderCapturePolicy and HeaderPropagationPolicy respectively, these definitions will override any top level HeaderCapturePolicy and HeaderPropagationPolicy rules.

Exposed the policy to the Receiver and Exporter nodes via the EffectHandler with helper functions that apply the policies on a iterator of key value pairs (for receiver nodes) and transport_headers (for exporter nodes)

* Closes #2508

unit tests and integration tests

no

### DIFF
--- a/rust/otap-dataflow/crates/config/src/transport_headers.rs
+++ b/rust/otap-dataflow/crates/config/src/transport_headers.rs
@@ -312,7 +312,13 @@ mod tests {
 
     #[test]
     fn propagate_all_captured_default() {
-        let policy = HeaderPropagationPolicy::default();
+        let policy = HeaderPropagationPolicy::new(
+            PropagationDefault {
+                selector: PropagationSelector::AllCaptured,
+                ..PropagationDefault::default()
+            },
+            vec![],
+        );
         let mut headers = TransportHeaders::new();
         headers.push(TransportHeader::text(
             "tenant_id",
@@ -327,15 +333,18 @@ mod tests {
 
         let propagated: Vec<_> = policy.propagate(&headers).collect();
         assert_eq!(propagated.len(), 2);
-        assert_eq!(propagated[0].egress_name, "X-Tenant-Id");
-        assert_eq!(propagated[1].egress_name, "X-Request-Id");
+        assert_eq!(propagated[0].header_name, "X-Tenant-Id");
+        assert_eq!(propagated[1].header_name, "X-Request-Id");
     }
 
     #[test]
     fn propagate_override_drops_auth() {
-        let policy = HeaderPropagationPolicy {
-            default: PropagationDefault::default(),
-            overrides: vec![PropagationOverride {
+        let policy = HeaderPropagationPolicy::new(
+            PropagationDefault {
+                selector: PropagationSelector::AllCaptured,
+                ..PropagationDefault::default()
+            },
+            vec![PropagationOverride {
                 match_rule: PropagationMatch {
                     stored_names: vec!["authorization".to_string()],
                 },
@@ -343,7 +352,7 @@ mod tests {
                 name: None,
                 on_error: None,
             }],
-        };
+        );
 
         let mut headers = TransportHeaders::new();
         headers.push(TransportHeader::text(
@@ -359,7 +368,7 @@ mod tests {
 
         let propagated: Vec<_> = policy.propagate(&headers).collect();
         assert_eq!(propagated.len(), 1);
-        assert_eq!(propagated[0].egress_name, "X-Tenant-Id");
+        assert_eq!(propagated[0].header_name, "X-Tenant-Id");
     }
 
     #[test]
@@ -393,18 +402,19 @@ mod tests {
 
         let propagated: Vec<_> = policy.propagate(&headers).collect();
         assert_eq!(propagated.len(), 1);
-        assert_eq!(propagated[0].egress_name, "X-Tenant-Id");
+        assert_eq!(propagated[0].header_name, "X-Tenant-Id");
     }
 
     #[test]
     fn propagate_stored_name_strategy() {
-        let policy = HeaderPropagationPolicy {
-            default: PropagationDefault {
+        let policy = HeaderPropagationPolicy::new(
+            PropagationDefault {
+                selector: PropagationSelector::AllCaptured,
                 name: NameStrategy::StoredName,
                 ..PropagationDefault::default()
             },
-            overrides: vec![],
-        };
+            vec![],
+        );
 
         let mut headers = TransportHeaders::new();
         headers.push(TransportHeader::text(
@@ -415,7 +425,7 @@ mod tests {
 
         let propagated: Vec<_> = policy.propagate(&headers).collect();
         assert_eq!(propagated.len(), 1);
-        assert_eq!(propagated[0].egress_name, "tenant_id");
+        assert_eq!(propagated[0].header_name, "tenant_id");
     }
 
     #[test]
@@ -442,6 +452,6 @@ mod tests {
 
         let propagated: Vec<_> = policy.propagate(&headers).collect();
         assert_eq!(propagated.len(), 1);
-        assert_eq!(propagated[0].egress_name, "X-Tenant-Id");
+        assert_eq!(propagated[0].header_name, "X-Tenant-Id");
     }
 }
--- a/rust/otap-dataflow/crates/config/src/transport_headers_policy.rs
+++ b/rust/otap-dataflow/crates/config/src/transport_headers_policy.rs
@@ -462,7 +462,7 @@ mod tests {
     #[test]
     fn default_propagation_policy() {
         let policy = HeaderPropagationPolicy::default();
-        assert_eq!(policy.default.selector, PropagationSelector::AllCaptured);
+        assert_eq!(policy.default.selector, PropagationSelector::None);
         assert_eq!(policy.default.action, PropagationAction::Propagate);
         assert_eq!(policy.default.name, NameStrategy::Preserve);
         assert_eq!(policy.default.on_error, ErrorAction::Drop);
--- a/rust/otap-dataflow/crates/otap/src/transport_headers.rs
+++ b/rust/otap-dataflow/crates/otap/src/transport_headers.rs
@@ -19,7 +19,7 @@ mod tests {
     use super::*;
     use otap_df_config::transport_headers_policy::{
         CaptureDefaults, CaptureRule, PropagationAction, PropagationDefault, PropagationMatch,
-        PropagationOverride,
+        PropagationOverride, PropagationSelector,
     };
 
     // -- Helper functions for tests ------------------------------------------
@@ -108,7 +108,10 @@ mod tests {
         // ========== Step 4: Simulate exporter propagation ==========
 
         let propagation_policy = HeaderPropagationPolicy::new(
-            PropagationDefault::default(),
+            PropagationDefault {
+                selector: PropagationSelector::AllCaptured,
+                ..PropagationDefault::default()
+            },
             vec![PropagationOverride {
                 match_rule: PropagationMatch {
                     stored_names: vec!["authorization".to_string()],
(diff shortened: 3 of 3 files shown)


rust/otap-dataflow/crates/config/src/transport_headers_policy.rs
```diff
@@ -0,0 +1,516 @@
...
+            return TransportHeaders::new();
+        }
+
+        let defaults = &self.defaults;
+        let mut result = TransportHeaders::with_capacity(defaults.max_entries.min(16));
+
+        for (wire_name, value) in pairs {
+            if result.len() >= defaults.max_entries {
+                break;
+            }
```
albertlockett: I wonder if the signature of this method should return some feedback about whether we failed to consume some of the pairs? Otherwise, it becomes a bit cumbersome, because we'd need to check a) that the pairs iter was consumed and b) that the result length is equal to the iter that was passed.

Maybe a different signature could be something like:
```
    pub fn capture_from_pairs<'a>(
        &self,
        headers: &mut TransportHeaders
        pairs: impl Iterator<Item = (&'a str, &'a [u8])>,
    ) -> Result<()> {
```

c1ly: Added a CaptureStats and PropagationStats that will store number of skipped pairs (due to policy defaults) and number of dropped headers respectively. Updated capture_from_pairs to return CaptureStats and propagate to return PropagationStats

rust/otap-dataflow/crates/config/src/transport_headers_policy.rs
```diff
@@ -115,7 +172,15 @@ impl HeaderCapturePolicy {
...
-        result
+        if skipped_max_entries > 0 || skipped_name_too_long > 0 || skipped_value_too_long > 0 {
+            Some(CaptureStats {
+                skipped_max_entries,
+                skipped_name_too_long,
+                skipped_value_too_long,
+            })
+        } else {
+            None
+        }
```
albertlockett, line 184: nit: it seems like all the stats are conditions that might be caused by an argument that is invalid for this capture policy. I wonder if it would make more sense for this to return something like a `Result<(), CaptureStats>`?

rust/otap-dataflow/crates/config/src/transport_headers_policy.rs
```diff
@@ -0,0 +1,558 @@
...
+/// Selects which captured headers are candidates for propagation.
+#[derive(Debug, Clone, Default, Serialize, Deserialize, JsonSchema, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum PropagationSelector {
+    /// Propagate all captured headers (subject to overrides).
+    #[default]
+    AllCaptured,
+    /// Do not propagate any captured headers by default (overrides may
+    /// still select specific headers).
+    None,
```
lquerel: I think the default variant should be `none` for security/privacy reasons.

rust/otap-dataflow/crates/config/src/transport_headers_policy.rs
```diff
@@ -0,0 +1,558 @@
...
+            })
+        } else {
+            None
+        }
+    }
+
+    /// Find the first capture rule whose `match_names` contains the given
+    /// wire name (case-insensitive comparison).
+    fn find_matching_rule(&self, wire_name: &str) -> Option<&CaptureRule> {
+        let wire_lower = wire_name.to_ascii_lowercase();
```
lquerel: This create an extra allocation we could avoid by using the method `eq_ignore_ascii_case`.
So concretely you can remove this local variable and in the test uses the method `eq_ignore_ascii_case`.

rust/otap-dataflow/crates/config/src/transport_headers_policy.rs
```diff
@@ -0,0 +1,558 @@
...
+            }
+        }
+
+        // Check whether the header passes the default selector.
+        let selected = match &self.default.selector {
+            PropagationSelector::AllCaptured => true,
+            PropagationSelector::None => false,
+            PropagationSelector::Named(names) => {
+                let name_lower = header.name.to_ascii_lowercase();
+                names.iter().any(|n| n.to_ascii_lowercase() == name_lower)
```
lquerel: Same comment than before, you could get rid of the local variable `name_lower` and uses the method `eq_ignore_ascii_case`.

rust/otap-dataflow/crates/config/src/transport_headers_policy.rs
```diff
@@ -0,0 +1,558 @@
...
+    pub match_names: Vec<String>,
+    /// Normalized logical name to store the header under. If omitted,
+    /// defaults to the first matched name lowercased.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub store_as: Option<String>,
+    /// Whether this header contains sensitive data (e.g. auth tokens).
+    /// Sensitive headers may receive special treatment in logging and
+    /// debug output.
+    #[serde(default)]
+    pub sensitive: bool,
```
lquerel, line 257: Add a todo here, as this capability is not yet fully implemented end-to-end.

rust/otap-dataflow/crates/config/src/transport_headers_policy.rs
```diff
@@ -0,0 +1,558 @@
...
+        Self { default, overrides }
+    }
+
+    /// Apply the propagation policy to a set of captured headers in-place,
+    /// removing headers that should not be sent on egress and renaming
+    /// wire names according to the configured name strategy.
+    ///
+    /// Returns `None` when all headers were propagated (none dropped),
+    /// or `Some(PropagationStats)` when one or more headers were removed.
+    pub fn propagate(&self, headers: &mut TransportHeaders) -> Option<PropagationStats> {
```
lquerel: `propagate()` currently allocates a second `TransportHeaders` and clones every selected header’s String/Vec<u8> before export. That adds avoidable heap churn on the egress path even though exporters only need a transient read-only view. Please change the propagation API to expose a borrowed iterator/view over the captured headers and have exporters consume that directly.

Send me a message if you need more details. I'd be happy to help.

rust/otap-dataflow/crates/config/src/transport_headers.rs
```diff
@@ -0,0 +1,457 @@
...
+}
+
+/// An ordered collection of captured transport headers.
+///
+/// Headers are stored as a `Vec` to preserve insertion order and allow
+/// duplicate names (same logical name appearing multiple times), which
+/// is valid in both HTTP and gRPC metadata.
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct TransportHeaders {
+    headers: Vec<TransportHeader>,
```
jmacd, line 99: I was somehow expecting to see an Rc<> or Arc<> here so that cloning a context does not clone the vector of headers. (Discussion topic.)

lquerel: I missed this one. Thank you
@chanly could you create a follow up PR to improve the cloning. Thanks

c1ly: this pr should fix it https://github.com/open-telemetry/otel-arrow/pull/2584

PR #2543, opened 2026-04-05
feat(fake_data_generator): Add per-record variance to static log generation

cijothomas:
## Summary

The static log generator previously produced batches where all records were
nearly identical — same body, same attribute values, same severity, no trace
context. This made payloads unrealistically compressible by Arrow/columnar
encoders (~57:1 zstd ratio), giving misleading throughput numbers in load tests.

## Changes

- **Body**: 50 realistic log message templates (~150 chars each) cycled across
  records. When `log_body_size_bytes` is configured, templates are repeated to
  fill the target size. When 0, body is omitted entirely.
- **Attribute names**: unified pool of 80 OTel semconv attribute names (e.g.
  `thread.id`, `http.route`, `db.query.text`). Overflow names use `attr_N`.
- **Attribute values**: 80-entry pool of realistic strings — URLs, UUIDs,
  hostnames, SQL queries, error messages, user agents — cycled per record.
  Special-cased types for `thread.id` (int), `http.response.status_code`
  (weighted distribution), etc.
- **Severity**: cycles 80% INFO / 15% WARN / 5% ERROR instead of all-INFO.
- **TraceID / SpanID**: random unique IDs per log record, matching real
  log-to-trace correlation behavior.
- **Compression test**: new `test_compression_ratio_is_realistic` asserts
  zstd ratio stays in 3:1–50:1 range (currently ~19:1), guarding against
  regression to the old all-identical regime.
- **Config docs**: updated `log_body_size_bytes` doc to reflect pool behavior.

## Impact

Both `fresh` and `pre_generated` strategies benefit since all variance is
computed at generation time. `PreGenerated` retains zero runtime allocation
cost — only the pre-built batch content is more realistic.

Compression ratio improved from ~57:1 (before) to ~19:1 (after) with
`log_body_size_bytes: 1024, num_log_attributes: 6`.

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
@@ -56,13 +56,235 @@ fn static_metric_attributes() -> Vec<KeyValue> {
     ]
 }
 
-/// Static log attributes (default when `num_log_attributes` is not configured).
-fn static_log_attributes() -> Vec<KeyValue> {
-    vec![
-        KeyValue::new("thread.id", AnyValue::new_int(1)),
-        KeyValue::new("thread.name", AnyValue::new_string("main")),
-    ]
+/// Pool of realistic log attribute names drawn from common OTel semantic conventions.
+/// When generating attributes, names are picked from this pool in order.
+/// If more attributes are requested than the pool contains, extra names are
+/// generated as `attr_N` (where N is the overflow index).
+const ATTR_NAMES: &[&str] = &[
+    "thread.id",
+    "thread.name",
+    "code.function",
+    "code.namespace",
+    "code.filepath",
+    "code.lineno",
+    "log.record.uid",
+    "event.name",
+    "exception.type",
+    "exception.message",
+    "exception.stacktrace",
+    "user.id",
+    "user.name",
+    "user.email",
+    "session.id",
+    "http.request.method",
+    "http.response.status_code",
+    "http.route",
+    "url.full",
+    "url.path",
+    "url.scheme",
+    "server.address",
+    "server.port",
+    "client.address",
+    "client.port",
+    "network.protocol.name",
+    "network.protocol.version",
+    "network.transport",
+    "db.system",
+    "db.namespace",
+    "db.operation.name",
+    "db.query.text",
+    "db.collection.name",
+    "messaging.system",
+    "messaging.operation.type",
+    "messaging.destination.name",
+    "messaging.message.id",
+    "rpc.system",
+    "rpc.service",
+    "rpc.method",
+    "enduser.id",
+    "enduser.role",
+    "enduser.scope",
+    "cloud.provider",
+    "cloud.region",
+    "cloud.availability_zone",
+    "cloud.account.id",
+    "container.id",
+    "container.name",
+    "container.image.name",
+    "container.image.tag",
+    "k8s.pod.name",
+    "k8s.namespace.name",
+    "k8s.deployment.name",
+    "k8s.node.name",
+    "os.type",
+    "os.version",
+    "process.pid",
+    "process.executable.name",
+    "process.command_line",
+    "process.runtime.name",
+    "process.runtime.version",
+    "deployment.environment",
+    "service.namespace",
+    "telemetry.sdk.name",
+    "telemetry.sdk.version",
+    "telemetry.sdk.language",
+    "feature_flag.key",
+    "feature_flag.variant",
+    "gen_ai.system",
+    "gen_ai.request.model",
+    "gen_ai.response.model",
+    "peer.service",
+    "otel.status_code",
+    "otel.status_description",
+    "error.type",
+    "host.name",
+    "host.id",
+    "host.arch",
+    "net.peer.name",
+    "net.peer.port",
+];
+
+/// Thread names to rotate through for realistic cardinality.
+const THREAD_NAMES: &[&str] = &[
+    "main",
+    "worker-1",
+    "worker-2",
+    "worker-3",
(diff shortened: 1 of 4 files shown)


cijothomas:
I was thinking of using the semantic_conventions itself, and generate a bunch of them at startup and then cycle through it - But the existing ones can only produce logs of size ~300 bytes, and we are targeting ~1KB, so we anyway have to put fillers in. This is first step in improve fake-gen, next is to have the perf tests switch to this.

rust/otap-dataflow/crates/otap/tests/durable_buffer_processor_tests.rs
```diff
@@ -1248,6 +1256,55 @@ fn test_durable_buffer_convert_to_arrow_mode() {
     );
 }
 
+/// Same as `test_durable_buffer_convert_to_arrow_mode` but with
+/// `use_trace_context: true` so each log record carries a random trace_id
+/// and span_id.
+///
+/// Currently ignored: the pipeline delivers 0 items when logs contain
```
cijothomas, line 1263: @AaronRM can you take a look why this test fails (after this PR is merged) ?

albertlockett (approved):

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
```diff
@@ -56,14 +56,236 @@ fn static_metric_attributes() -> Vec<KeyValue> {
...
-fn static_log_attributes() -> Vec<KeyValue> {
-    vec![
-        KeyValue::new("thread.id", AnyValue::new_int(1)),
-        KeyValue::new("thread.name", AnyValue::new_string("main")),
-    ]
+/// Pool of realistic log attribute names drawn from common OTel semantic conventions.
+/// When generating attributes, names are picked from this pool in order.
+/// If more attributes are requested than the pool contains, extra names are
+/// generated as `attr_N` (where N is the overflow index).
+const ATTR_NAMES: &[&str] = &[
```
lalitb, line 63: FYI we already have `semconv_signal.rs` that pulls attr names/values from the actual registry - we can see feasibility for reusing that here instead of maintaining a static list, maybe in separate PR.

cijothomas: https://github.com/open-telemetry/otel-arrow/pull/2543#issuecomment-4192695960 Yes good to follow up. For the perf test purposes itself - it does not matter where the attribute are coming, but I think it does make a lot of sense to claim that the perf tests use a load, that matches OTel's own conventions.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
```diff
@@ -56,14 +56,236 @@ fn static_metric_attributes() -> Vec<KeyValue> {
...
+    "tokio-runtime-0",
+    "tokio-runtime-1",
+    "blocking-0",
+    "blocking-1",
+];
+
+/// Realistic attribute value pools for generic string attributes.
+/// Each pool entry is 20–120 chars, resembling real observability values like
+/// paths, IDs, hostnames, URLs, error descriptions, and user agents.
+const ATTR_VALUE_POOL: &[&str] = &[
```
lalitb, line 162: nit: this is a lot of hardcoded strings to maintain - any chance we could generate this from the semconv registry examples at build time instead - Maybe worth evaluating in subsequent PR.

cijothomas: it won't gives us what we need - bigger payloads, more variety to avoid unrealistic compression. We'd still need our own variations added.
If there is some publicly available observability data (eg: From OTel Demo project etc.), I'll try to get that in here in future.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
```diff
@@ -343,42 +651,84 @@ mod tests {
...
+        println!(
+            "Compression: raw={raw_size} bytes, compressed={compressed_size} bytes, ratio={ratio:.1}:1"
+        );
+
+        // Guard against unrealistically compressible data.
+        // Before these changes, identical bodies produced ~57:1 ratios.
+        // With 50 distinct templates and varied attributes, expect ~20:1 to ~45:1
+        // depending on body size. Real systems see ~5:1 to ~10:1, but protobuf
+        // structure overhead keeps synthetic data more compressible.
+        // The key goal: never exceed 50:1 (the old all-identical regime).
```
lalitb: This is confusing - We earlier say "expect ~20:1 to ~45:1" but then the assert allows 50:1 - why not just use 45:1 as the upper bound to match?

lalitb (approved):
Nice work! Few minor nits but nothing blocking — LGTM.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
cijothomas: good catch! Let me fix it to match reality. Tests got stale a bit with some newer things I tried.

cijothomas: fixed in latest commit.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
cijothomas: Discussed offline and resolving. This is the best option for our current testing.

PR #2553, opened 2026-04-06
[WIP] feat(temporal_reaggregation_processor): Support for hashing map and array anyvalue types

JakeDern:
# Change Summary

This PR adds support to the hashing logic for maps and arrays in the identity module.

Unfortunately, I don't yet have support for this for arrow in the View itself. To do this zero-copy for arrow we need to go implement a view upstream in ciborium or implement our own/find another CBOR parser. 

We could also hack something temporary with owned data just to get it to work if we wanted to, but we would have to refactor the types a little bit.

* Part of #2422

Unit

No


JakeDern:
Going to postpone this since there's not a good way to support array/map types for arrow right now

PR #2555, opened 2026-04-06
Record batch_size metric once per batch instead of per HTTP attempt

cijothomas:
The `batch_size` Mmsc metric was recorded inside `try_export()`, which runs on every HTTP attempt including retries. Since the compressed payload is identical across retries (`body.clone()`), this inflated `sum` and `count` without adding useful information.

This PR moves the recording to `export()`, before the retry loop, so it is observed exactly once per batch. Per-attempt status code recording remains in `try_export()` where it belongs.

**Changes:**
- **client.rs**: Move `add_batch_size()` from `try_export()` (per-attempt) to `export()` (once per batch). Remove now-unused `body_len` variable.
- **metrics.rs**: Clarify doc comment to note the metric is recorded once per batch, before the exporter's own HTTP retry loop.

### DIFF
--- a/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
@@ -70,7 +70,7 @@ pub struct AzureMonitorExporterMetrics {
     #[metric(unit = "ms")]
     pub auth_success_latency: Mmsc,
     /// Compressed batch size in bytes (min/max/sum/count).
-    /// Recorded once per batch, before the exporter's own HTTP retry loop.
+    /// Recorded once per batch; HTTP retries do not produce additional observations.
     #[metric(unit = "By")]
     pub batch_size: Mmsc,
     /// Current number of in-flight export requests.


rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/metrics.rs
```diff
@@ -69,7 +69,8 @@ pub struct AzureMonitorExporterMetrics {
     /// Authentication success latency in milliseconds (min/max/sum/count).
     #[metric(unit = "ms")]
     pub auth_success_latency: Mmsc,
-    /// Batch size in bytes (min/max/sum/count).
+    /// Compressed batch size in bytes (min/max/sum/count).
+    /// Recorded once per batch, before the exporter's own HTTP retry loop.
```
utpilla: > exporter's own HTTP retry loop.

Should we make the description more generic without talking about the implementation detail?

```suggestion
    /// Recorded once per batch; HTTP retries do not produce additional observations.
```

utpilla (approved):

gouslu:
Ideally we should be able to attach status code as a dimension for this metric. I am not fully onboard with splitting them for this general idea, but in practice it is not possible to correlate them 1:1 currently, so I think this change makes sense. I would however like to have a todo so that once we support dimensions we do the work to enhance our metrics, so that they are more useful.

cijothomas:
> Ideally we should be able to attach status code as a dimension for this metric. I am not fully onboard with splitting them for this general idea, but in practice it is not possible to correlate them 1:1 currently, so I think this change makes sense. I would however like to have a todo so that once we support dimensions we do the work to enhance our metrics, so that they are more useful.

Yes agreed. The "laclient_*" ones in particular are quite unfortunate today, to be fixed when attribute support is added.

Issue #2540, opened 2026-04-05
Fake generator improvements


cijothomas:
Majority of the perf tests in the repo relies on fake-generator, so it is important to ensure it supports the scenarios properly.
1. In today's saturation+scale tests, the bottleneck is load generator, not engine (as seen by engine not hitting 100%).
2. It takes 3 cores of loadgen to saturate 1 core engine. (The perf machine has 128 cores only, so this is limiting )
3. No entropy - repeating same payload means we get unfair/unrealistic compressions (1:100 ratio). https://github.com/open-telemetry/otel-arrow/pull/2535 recently fixed an issue related to this . But it is better for the loadgenerator to attempt closer-to-real world scenarios by introducing entropy/randomness.
4. We also need a way to measure the efficiency of load generator itself, so improvements can be tracked over time.
5. Need to expand support for metrics, traces.

cijothomas:
https://github.com/open-telemetry/otel-arrow/pull/1877 follow up to this PR is the first step. I'll send a bunch of small PRs to address this issue.

PR #2559, opened 2026-04-06
feat(fake_data_generator): Add SUT saturation benchmark

cijothomas:
## Summary

Add a benchmark script and configs to measure whether the fake data generator can saturate a `df_engine` (SUT) over OTLP gRPC using a single core.

Part of #2540

## Setup

3 processes, 1 pipeline core each:

```
sender(fake-gen → OTLP) → SUT(OTLP recv → OTLP export) → backend(OTLP recv → noop)
     :8080                      :8081                           :8082
```

## Results (Apple M3 Pro)

| Config              | Throughput       | Sender core | SUT core     |
|---------------------|------------------|-------------|--------------|
| static/fresh 1KB    | ~480 K logs/sec  | ~82 %       | **~97 %**    |
| static/pregen 1KB   | ~500 K logs/sec  | ~66 %       | **~95 %**    |

**Conclusion:** `static` + `pre_generated` saturates the SUT to ~95% while using only ~66% of 1 sender core. No need for multi-core load generators.

## Files added

| File | Purpose |
|------|---------|
| `fake_data_generator/README.md` | Benchmark purpose, configs, results, conclusion |
| `fake_data_generator/bench/bench.sh` | Automated benchmark (~3 min) |
| `bench/sender-static-fresh.yaml` | Sender: fake-gen(fresh) → OTLP |
| `bench/sender-static-pregen.yaml` | Sender: fake-gen(pregen) → OTLP |
| `bench/sut-otlp-forward.yaml` | SUT: OTLP recv → OTLP export |
| `bench/backend-noop.yaml` | Backend: OTLP recv → noop |

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/README.md
@@ -1,0 +1,53 @@
+# Fake Data Generator -- SUT Saturation Benchmarks
+
+## Purpose
+
+Verify that a **single fake-gen sender core can saturate a single
+`df_engine` (SUT) core** over OTLP gRPC, so that performance tests
+don't need multiple load-generator cores.
+
+```text
+  sender (1 core)           SUT (1 core)            backend (1 core)
+  fake-gen -> OTLP-export --> OTLP-recv -> OTLP-export --> OTLP-recv -> noop
+       :8080                     :8081                      :8082
+```
+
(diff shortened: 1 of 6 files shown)


cijothomas:
Don't think we'll need to retain this long term, but this is quite useful now, as we are currently struggling to saturate 1 SUT even with multiple fake-senders! And keeping this while we try for better source of synthetic data is useful to make sure if we can easily saturate without load-gen itself taking too much cpu cycles.

PR #2565, opened 2026-04-07
fix(quiver): unify dictionary handling across batches in Arrow IPC format

### LINKED ISSUE #2557: Quiver: Segment finalization fails silently when Arrow batches contain dictionary-encoded columns with varying values
### Pre-filing checklist

### Component(s)

### Bug Description

Quiver's `StreamAccumulator` uses Arrow IPC `FileWriter` to serialize multiple `RecordBatch`es into a single stream. The `FileWriter` rejects dictionary replacement across batches, causing segment finalization to fail. The error is silently swallowed, resulting in data that is ingested but never delivered downstream.

### Impacted Scenarios

- **Durable buffer with `otlp_handling: convert_to_arrow`**: Any OTLP data converted to Arrow format before storage. Logs with trace context, traces, metrics with exemplars, and any signal with high-cardinality attributes will trigger dictionary conflicts across batches.
- **Native Arrow ingestion**: Any `OtapArrowRecords` ingested directly (not via OTLP pass-through) where dictionary-encoded columns carry different values across batches within the same segment.
- **OTLP pass-through mode is NOT affected** - OTLP bytes in pass-thru mode are stored opaque binary, not Arrow-encoded data.

In practice, this means segment finalization breaks as soon as a stream accumulates two or more batches with dictionary-encoded columns containing different values.

### Root Cause

`StreamAccumulator::write_ipc_to()` in `stream_accumulator.rs` uses `FileWriter` with default options. Arrow IPC File format requires a single dictionary per field across all batches. The `FileWriter` hardcodes `error_on_replacement = true` in its `DictionaryTracker`. When batch N has different dictionary values than batch N-1, `ipc_writer.write(batch)` returns `InvalidArgumentError("Dictionary replacement detected...")`.

### Why It Was Hard to Detect

1. `handle_timer_tick` catches `engine.flush()` errors and logs via `otel_warn!` - but returns `Ok(())`
2. `handle_shutdown` does the same with `otel_error!` - also returns `Ok(())`
3. Integration tests have no tracing subscriber, so those log events are invisible
4. No metric tracks segment finalization failures
5. The corrupt segment file (created before the write, left incomplete after failure) occupies the sequence number but is never registered with the segment store. Issue is only encountered on recovery.

### Steps to Reproduce

The `test_durable_buffer_convert_to_arrow_mode_with_trace_context` integration test (currently `#[ignore]`) reproduces this: it generates logs with unique `trace_id`/`span_id` per record, converts to Arrow, and delivers 0 items in 10 seconds.

### Expected Behavior

When multiple `RecordBatch`es with dictionary-encoded columns are ingested into the durable buffer, Quiver should finalize segments successfully and deliver all data downstream. Dictionary-encoded columns with varying values across batches (e.g., unique `trace_id` per log record) should be handled transparently.

### Actual Behavior

Segment finalization fails with `InvalidArgumentError("Dictionary replacement detected when writing IPC file format")`. The error is silently swallowed. Data is consumed (ingested into the open segment) but never produced (delivered downstream), resulting in 0 items delivered and eventual timeout.

### OTel-Arrow Version

092500ce98b89039cbc199b3368be66f4f08ba93

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

AaronRM:
# Change Summary

Fix silent data loss caused by Arrow IPC dictionary replacement rejection during Quiver segment finalization. When multiple RecordBatches with dictionary-encoded columns (e.g., `trace_id`, `span_id`, `scope.name`, `body.str`) accumulate within a segment, Arrow's `FileWriter` rejects batches whose dictionary values differ from the first batch. This caused segment finalization to fail silently, losing all data in the segment.

The fix unifies per-batch dictionaries before writing IPC files by concatenating dictionary values arrays and remapping keys. This handles both top-level dictionary columns and dictionary fields nested inside `Struct` columns (covering all OTAP schema patterns). Key type widening (e.g., `UInt8` -> `UInt16`) is performed when the combined dictionary exceeds the original key type's capacity. If widening would exceed the maximum supported key width (`UInt16`), the dictionary encoding is stripped and the column is stored as the native value type -- ensuring segment finalization succeeds regardless of cardinality.

Additional changes:
- `flush_failures` metric added to both `PersistenceMetrics` and `DurableBufferMetrics` for observability
- `otel_warn` logged when dictionary encoding falls back to native type; `otel_debug` logged on key widening
- Best-effort cleanup of partially-written segment files on flush failure
- Schema fingerprint documented as ingestion-time only (may differ from IPC schema after unification)
- Replace magic metric index numbers with named constants in tests
- Un-ignore `test_durable_buffer_convert_to_arrow_mode_with_trace_context` (this bug was the root cause)
- Add integration tests for traces, metrics, and mixed signals through `convert_to_arrow` mode
- ARCHITECTURE.md updated with dictionary unification strategy, native fallback behavior, supported nesting patterns, and note on future configurability of max key width
- New dependency: `arrow-cast` (same version as other arrow crates, already compiled for workspace)

* Closes #2557

**Quiver unit tests** (10 new tests in `stream_accumulator`):
- Multi-batch dictionary unification with varying values, null keys, 3-batch offset accumulation, and key type widening
- Signed dictionary key types (Int8)
- Struct columns with dictionary children: varying values, null keys, single-batch bypass, key widening
- Non-Utf8 dictionary value types: Dict(UInt8, Int32), Dict(UInt8, FixedSizeBinary(16)), Dict(UInt16, Binary)
- Struct with non-Utf8 dictionary children: Dict(UInt8, Int64) + Dict(UInt8, Binary)
- Native type fallback when cardinality exceeds UInt16 (top-level and struct child)

**Integration tests** (3 new + 1 un-ignored in `durable_buffer_processor_tests`):
- `test_durable_buffer_convert_to_arrow_mode_traces` -- traces-only through Arrow pipeline
- `test_durable_buffer_convert_to_arrow_mode_metrics` -- metrics-only through Arrow pipeline
- `test_durable_buffer_convert_to_arrow_mode_mixed_signals` -- all 3 signal types with trace context
- `test_durable_buffer_convert_to_arrow_mode_with_trace_context` -- un-ignored, now passing

All tests assert `flush_failures == 0` to verify segment finalization succeeds.

No user-facing API changes. The `flush_failures` metric is newly exposed on `DurableBufferMetrics` for operators monitoring segment health. Non-zero values indicate data at risk -- check logs for root cause. New `otel_warn` events (`quiver.dict.native_fallback`) and `otel_debug` events (`quiver.dict.key_widened`) provide visibility into dictionary schema changes during segment finalization.

### DIFF
--- a/rust/otap-dataflow/crates/quiver/src/segment/stream_accumulator.rs
+++ b/rust/otap-dataflow/crates/quiver/src/segment/stream_accumulator.rs
@@ -47,12 +47,14 @@ use arrow_array::types::{
 };
 use arrow_array::{Array, ArrayRef, DictionaryArray, RecordBatch, StructArray};
 use arrow_buffer::ArrowNativeType;
+use arrow_cast::cast;
 use arrow_ipc::writer::{FileWriter, IpcWriteOptions};
 use arrow_schema::{DataType, FieldRef, Fields, Schema, SchemaRef};
 use arrow_select::concat::concat;
 
 use super::error::SegmentError;
 use super::types::{ChunkIndex, MAX_CHUNKS_PER_STREAM, StreamId, StreamKey, StreamMetadata};
+use crate::logging::{otel_debug, otel_warn};
 use crate::record_bundle::{SchemaFingerprint, SlotId};
 
 /// Accumulates `RecordBatch`es for a single `(slot, schema)` stream.
@@ -253,6 +255,10 @@ impl StreamAccumulator {
         self.write_ipc_to(&mut counting)?;
         let byte_length = counting.bytes_written() as u64;
 
+        // Note: schema_fingerprint reflects the *ingestion* schema, which may
+        // differ from the IPC file schema after dictionary unification
+        // (key widening or native-type fallback). This is intentional -- the
+        // fingerprint is used for write-path routing only.
         Ok(StreamMetadata::new(
             self.stream_id,
             self.slot_id,
@@ -330,6 +336,24 @@ impl<W: std::io::Write> std::io::Write for CountingWriter<W> {
 // ─────────────────────────────────────────────────────────────────────────────
 // Dictionary unification helpers
 // ─────────────────────────────────────────────────────────────────────────────
+
+/// Maximum dictionary key width that downstream readers support.
+///
+/// When dictionary unification produces a key type wider than this, the
+/// dictionary encoding is stripped and columns are stored as the native
+/// value type instead. This prevents writing segments that cannot be read
+/// back by the OTAP reader stack (which only supports `UInt8` / `UInt16`
+/// dictionary keys).
+///
+/// TODO: Make this configurable via `QuiverConfig` so non-OTAP embeddings
+/// can set a different limit (e.g., `UInt32` or `UInt64`).
+const MAX_DICT_KEY_TYPE: DataType = DataType::UInt16;
+
+/// Returns `true` if `key_type` exceeds the maximum supported dictionary
+/// key width.
+fn exceeds_max_dict_key(key_type: &DataType) -> bool {
+    key_type_capacity(key_type) > key_type_capacity(&MAX_DICT_KEY_TYPE)
+}
 
 /// Returns `true` if any field in the schema uses dictionary encoding,
 /// including dictionary fields nested inside `Struct` columns.
@@ -427,6 +451,10 @@ fn unify_batch_dicts(
 /// Unifies a single dictionary column across all batches.
 ///
 /// Returns the new per-batch columns and the (possibly updated) field.
+///
+/// If the unified dictionary cardinality exceeds the maximum supported key
+/// width ([`MAX_DICT_KEY_TYPE`]), the dictionary encoding is stripped and
+/// columns are cast to the native value type.
 fn unify_dict_column(
     col_idx: usize,
     field: &FieldRef,
@@ -457,6 +485,31 @@ fn unify_dict_column(
     // Determine the key type that can address all unified values.
     let effective_key_type = widen_key_type(key_type, total_values)?;
 
+    // If the effective key type exceeds the maximum supported width, fall
+    // back to the native value type by casting each batch's dictionary
+    // column. This avoids writing segments that cannot be read back.
+    if exceeds_max_dict_key(&effective_key_type) {
+        otel_warn!(
+            "quiver.dict.native_fallback",
+            field = field.name().as_str(),
+            original_key_type = ?key_type,
+            total_values = total_values,
+            message = "dictionary cardinality exceeds max key width; \
+                       falling back to native type. Consider reducing \
+                       segment.target_size or segment.max_open_duration",
+        );
+        let native_type = value_type.clone();
+        let columns: Vec<ArrayRef> = batches
+            .iter()
+            .map(|batch| {
+                cast(batch.column(col_idx), &native_type)
+                    .map_err(|e| SegmentError::Arrow { source: e })
+            })
+            .collect::<Result<_, _>>()?;
+        let new_field = Arc::new(field.as_ref().clone().with_data_type(native_type));
+        return Ok((columns, new_field));
+    }
+
     // Build remapped columns for each batch.
     let columns: Vec<ArrayRef> = batches
         .iter()
@@ -474,6 +527,13 @@ fn unify_dict_column(
 
     // Update the field if the key type was widened.
     let new_field = if effective_key_type != *key_type {
+        otel_debug!(
+            "quiver.dict.key_widened",
+            field = field.name().as_str(),
+            original_key_type = ?key_type,
+            widened_key_type = ?effective_key_type,
+            total_values = total_values,
+        );
         Arc::new(field.as_ref().clone().with_data_type(DataType::Dictionary(
             Box::new(effective_key_type),
             Box::new(value_type.clone()),
@@ -549,22 +609,54 @@ fn unify_struct_column(
 
             let effective_key_type = widen_key_type(key_type, total_values)?;
 
-            let unified_children: Vec<ArrayRef> = child_arrays
-                .iter()
-                .zip(batch_offsets.iter())
-                .map(|(arr, &offset)| {
-                    remap_dict_keys(arr, key_type, &effective_key_type, &unified_values, offset)
-                })
-                .collect::<Result<_, _>>()?;
-
-            child_data.push(unified_children);
-
-            if effective_key_type != **key_type {
-                new_child_fields.push(Arc::new(child_field.as_ref().clone().with_data_type(
-                    DataType::Dictionary(Box::new(effective_key_type), value_type.clone()),
-                )));
+            // If the effective key type exceeds the max supported width,
+            // fall back to native value type for this struct child.
+            if exceeds_max_dict_key(&effective_key_type) {
+                otel_warn!(
+                    "quiver.dict.native_fallback",
+                    field = %format!("{}.{}", field.name(), child_field.name()),
+                    original_key_type = ?key_type,
+                    total_values = total_values,
+                    message = "dictionary cardinality exceeds max key width; \
+                               falling back to native type",
+                );
+                let native_type: DataType = *value_type.clone();
+                let native_children: Vec<ArrayRef> = child_arrays
+                    .iter()
+                    .map(|arr| {
+                        cast(arr.as_ref(), &native_type)
+                            .map_err(|e| SegmentError::Arrow { source: e })
+                    })
+                    .collect::<Result<_, _>>()?;
+                child_data.push(native_children);
+                new_child_fields.push(Arc::new(
+                    child_field.as_ref().clone().with_data_type(native_type),
+                ));
             } else {
-                new_child_fields.push(Arc::clone(child_field));
+                let unified_children: Vec<ArrayRef> = child_arrays
+                    .iter()
+                    .zip(batch_offsets.iter())
+                    .map(|(arr, &offset)| {
+                        remap_dict_keys(arr, key_type, &effective_key_type, &unified_values, offset)
+                    })
+                    .collect::<Result<_, _>>()?;
+
+                child_data.push(unified_children);
+
+                if effective_key_type != **key_type {
+                    otel_debug!(
+                        "quiver.dict.key_widened",
+                        field = %format!("{}.{}", field.name(), child_field.name()),
+                        original_key_type = ?key_type,
+                        widened_key_type = ?effective_key_type,
+                        total_values = total_values,
+                    );
+                    new_child_fields.push(Arc::new(child_field.as_ref().clone().with_data_type(
+                        DataType::Dictionary(Box::new(effective_key_type), value_type.clone()),
+                    )));
+                } else {
+                    new_child_fields.push(Arc::clone(child_field));
+                }
             }
         } else {
             // Non-dict child: pass through from each struct.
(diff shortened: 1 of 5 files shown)


rust/otap-dataflow/crates/quiver/src/segment/stream_accumulator.rs
```diff
@@ -538,4 +1001,1137 @@ mod tests {
...
+        assert_eq!(batches[1].num_rows(), 200);
+
+        // Spot-check first and last values in each batch.
+        let vals0 = read_dict_strings(&batches[0], 1);
+        assert_eq!(vals0[0].as_deref(), Some("val_a_0"));
+        assert_eq!(vals0[199].as_deref(), Some("val_a_199"));
+
+        let vals1 = read_dict_strings(&batches[1], 1);
+        assert_eq!(vals1[0].as_deref(), Some("val_b_0"));
+        assert_eq!(vals1[199].as_deref(), Some("val_b_199"));
```
drewrelmas, line 1396: This is a really neat way of testing this functionality! Is there any other place in the codebase that deals with multiple record batches like this where schema differences could cause a problem? 

My intuition is that widening is never needed for regular transform scenarios that work on one `RecordBatch` at a time (and also aren't saving data to segment files!)

drewrelmas (approved):
Looks good to my eyes based on the excellent test coverage. Will wait for any additional review before merging this tomorrow.

lquerel:
@albertlockett could you review this PR as it's related to arrow dictionaries. Thanks

rust/otap-dataflow/crates/quiver/src/segment/stream_accumulator.rs
```diff
@@ -304,6 +327,446 @@ impl<W: std::io::Write> std::io::Write for CountingWriter<W> {
...
+            Box::new(value_type.clone()),
+        )))
+    } else {
+        Arc::clone(field)
+    };
+
+    Ok((columns, new_field))
+}
+
+/// Unifies dictionary children inside a `Struct` column across all batches.
```
lalitb, line 548: As per the code`has_dict_columns` will recurse through the nested `struct`s, so it will detect dict in say `Struct<Struct<Dict>>`. But `unify_struct_column` will only process one level of children - so a dict nested inside a multiple nested struct will get unnoticed.

More of a question - Is handling single-level nesting sufficient for all current OTAP schemas?

AaronRM: Yes. The OTAP spec defines struct fields with dot notation (e.g., `resource.schema_url`, `body.str`, `status.code`) and every struct child is a primitive or dictionary type, so this code only needs to support one level of struct nesting to support OTAP. Pushed a commit to make it consistent for now.

However, Quiver can and should support Arrow formats beyond OTAP, so we should come back and address this later. Capturing current support level in Quiver's `ARCHITECTURE.md`.

rust/otap-dataflow/crates/quiver/src/segment/stream_accumulator.rs
```diff
@@ -304,6 +327,449 @@ impl<W: std::io::Write> std::io::Write for CountingWriter<W> {
...
+fn widen_key_type(original: &DataType, total_values: usize) -> Result<DataType, SegmentError> {
+    if total_values <= key_type_capacity(original) {
+        return Ok(original.clone());
+    }
+    // Widening chain preserves signedness.
+    let chain: &[DataType] = match original {
+        DataType::Int8 | DataType::Int16 | DataType::Int32 | DataType::Int64 => {
+            &[DataType::Int16, DataType::Int32, DataType::Int64]
+        }
+        _ => &[DataType::UInt16, DataType::UInt32, DataType::UInt64],
```
lalitb, line 792: There seems to be an issue here - the widening chain allows promotion to `UInt32`/`UInt64`, but the downstream pdata reader stack seems to only support `UInt8` and `UInt16 types`.  As an example here for pdata array:
https://github.com/open-telemetry/otel-arrow/blob/37df685b98ec6e7edc1f0ed84d4d00c72c371bf7/rust/otap-dataflow/crates/pdata/src/arrays.rs#L479-L496

This means the segment can be successfully written with `UInt32`, but failed to be read back to OTAP type. Should we cap the chain at `&[DataType::UInt16]`  so segment write fails gracefully instead of silently producing an unreadable file ?

rust/otap-dataflow/crates/quiver/src/segment/stream_accumulator.rs
AaronRM: Quiver's segment writer is the only code that takes N independent batches (of a common schema) and writes them into a single Arrow IPC file section (embedded within the `.qseg`) but still as separate batches within the IPC file section. The batches have to share a single dictionary for the IPC file section. In the interest of saving CPU cycles during segment writing, Quiver will simply *concatenate each input Dictionary* into a single larger dictionary (i.e. there may be duplicates in the dictionary).

The other similar place I'm aware of that deals with multiple record batches and Dictionary key type unification is `concatenate.rs`, but the problem there is slightly different. In that case, multiple batches are being merged into a single larger `RecordBatch` and the function is attempting to *merge* Dictionaries into a single Dictionary (with no key duplicates).

And your intuition is correct. For everywhere else, none of this detail matters since the Dictionary key type isn't a primary concern and/or the code is working with a single RecordBatch at a time.

rust/otap-dataflow/crates/quiver/src/segment/stream_accumulator.rs
lalitb: Thinking more, I won't block on this issue, as probability for this scenario is low - the dict cardinality of all the batches in one segment should be more than ~65K for this to happen.

AaronRM: Agree that it's somewhat low probability, but *I* would like to block on it 😄 [Pushed one more commit](https://github.com/open-telemetry/otel-arrow/pull/2565/commits/0738b4d8720d505e05ba54b4e688e87f02508cb3) to hard code the UInt16 limit (imposed by OTAP limit) into quiver for now. If we exceed the limit during key-widening, then fall back to the native type. I filed #2570 to track making the limit configurable.

rust/otap-dataflow/crates/quiver/src/segment/stream_accumulator.rs
```diff
@@ -304,6 +333,535 @@ impl<W: std::io::Write> std::io::Write for CountingWriter<W> {
...
+    let mut batch_offsets: Vec<usize> = Vec::with_capacity(batches.len());
+    let mut total_values: usize = 0;
+
+    for batch in batches {
+        let col = batch.column(col_idx);
+        let values = dict_values(col, key_type)?;
+        batch_offsets.push(total_values);
+        total_values += values.len();
+        value_refs.push(values);
+    }
```
albertlockett, line 476: This won't take into account if there are some dictionary values that are the same across dictionaries, will it?

e.g. If I had two columns with dictionary arrays:

```
DictionaryArray {
  keys: [ ... ]
  values: ["a", "b"]
}
DictionaryArray {
  keys: [..]
   values: ["a", "c"]
}
```

won't the `total_values` count be 4 when in reality it should be 3?

AaronRM: Yes, but this is known / by design. We are intentionally *concatenating* the key sets without deduplication to avoid the CPU cost of building a hash set during segment writing.

AaronRM:
> I had one comment/question about how we accumulate the total count of all dict values. I'm thinking maybe it should be count of unique values? Otherwise we end up to avoid expanding the dict unnecessarily, which could lead inefficiencies.
> 
> However, tracking unique dictionary values w/ key remapping is also pretty complicated, and it's difficult to make it work efficiently.

That's exactly why in this change we're just doing strict concatenation, not remapping/dedupe.

> So in lieu of fixing the total_count issue directly, I'm wondering if there's a different approach that could be taken .. The Arrow IPC _file format_ does have this restriction that all batches must have the same dictionary values for some column, however the IPC _streaming format_ doesn't have this same restriction. It allows for dictionary delta/replacement messages. We could maybe try switching the `arrow_ipc::writer::FileWriter` for `arrow_ipc::writer::StreamWriter`, and that might solve the issue (we'd have to switch the reader as well).
> 
> Obviously this is a bigger lift, so I'm fine merging this as is and investigating switching to IPC stream format in future.

Yes, that's a more impactful change. The IPC file format was specifically chosen for Quiver because it supports seeking and memory-mapped reads. The segment reader uses `FileDecoder` with offset-based access, and the entire segment file is designed for zero-copy mmap. The streaming format is sequential read only, which would break the mmap/seek access pattern that Quiver's segment reader relies on for efficient bundle retrieval.

albertlockett:
> > I had one comment/question about how we accumulate the total count of all dict values. I'm thinking maybe it should be count of unique values? Otherwise we end up to avoid expanding the dict unnecessarily, which could lead inefficiencies.
> > However, tracking unique dictionary values w/ key remapping is also pretty complicated, and it's difficult to make it work efficiently.
> 
> That's exactly why in this change we're just doing strict concatenation, not remapping/dedupe.
> 
> > So in lieu of fixing the total_count issue directly, I'm wondering if there's a different approach that could be taken .. The Arrow IPC _file format_ does have this restriction that all batches must have the same dictionary values for some column, however the IPC _streaming format_ doesn't have this same restriction. It allows for dictionary delta/replacement messages. We could maybe try switching the `arrow_ipc::writer::FileWriter` for `arrow_ipc::writer::StreamWriter`, and that might solve the issue (we'd have to switch the reader as well).
> > Obviously this is a bigger lift, so I'm fine merging this as is and investigating switching to IPC stream format in future.
> 
> Yes, that's a more impactful change. The IPC file format was specifically chosen for Quiver because it supports seeking and memory-mapped reads. The segment reader uses `FileDecoder` with offset-based access, and the entire segment file is designed for zero-copy mmap. The streaming format is sequential read only, which would break the mmap/seek access pattern that Quiver's segment reader relies on for efficient bundle retrieval.

If we're not going to switch to the stream format, expanding the dictionaries like this will introduce some inefficiencies. I'm thinking especially for dictionaries we know to be mostly low cardinality, like attribute keys and severity text/number. 

Do you have a long term plan to resolve this?

On the one hand, we could dedupe/remap the keys (like we do in batch processor / attribute processor), but like I mentioned it's a complicated/expensive operation.

I guess the other option would be to somehow change the data layout? e.g. maybe we'd need to map OTAP batches that have different dictionary values in any column to a different slot?

AaronRM:
> Do you have a long term plan to resolve this?
> 
> On the one hand, we could dedupe/remap the keys (like we do in batch processor / attribute processor), but like I mentioned it's a complicated/expensive operation.

I'm running some experiments to measure the CPU cost vs space savings of dedup across different cardinality/overlap profiles to help us make an informed decision. I'll share the results later this week.

AaronRM:
@albertlockett I worked with one of my favorite LLM's to [design and run some experiments to help us decide based on measurements from simulated OTAP data](https://github.com/AaronRM/otel-arrow/blob/expore-dict-dedupe/rust/otap-dataflow/benchmarks/dict-dedup-analysis.md).

TL;DR The recommendation is that we should de-dupe below a certain average cardinality threshold (default: 256) and above that simply concat without dedupe.   Interested to get your thoughts at least [on the Recommendation section](https://github.com/AaronRM/otel-arrow/blob/expore-dict-dedupe/rust/otap-dataflow/benchmarks/dict-dedup-analysis.md#recommendation).

albertlockett:
> @albertlockett I worked with one of my favorite LLM's to [design and run some experiments to help us decide based on measurements from simulated OTAP data](https://github.com/AaronRM/otel-arrow/blob/expore-dict-dedupe/rust/otap-dataflow/benchmarks/dict-dedup-analysis.md).
> 
> TL;DR The recommendation is that we should de-dupe below a certain average cardinality threshold (default: 256) and above that simply concat without dedupe. Interested to get your thoughts at least [on the Recommendation section](https://github.com/AaronRM/otel-arrow/blob/expore-dict-dedupe/rust/otap-dataflow/benchmarks/dict-dedup-analysis.md#recommendation).

Yeah I agree with the recommendation that configurable dict unification is the best solution among those explored. The extra CPU overhead is unfortunate.

On additional question I'd have is - say I had two subsequent batches some dict encoded column, and values:
```
// col from batch 1:
{
  key: [0, 0, 1]
  values: ["a", "b"]
}
// col from batch 2:
{
  key: [0, 1, 0]
  values: ["c", "d"]
}
```

When I read these two batches out of quiver, we wouldn't "de-unify" the dict on the way read path, right? So we'd end up reading something like the following, right?
```
// col from batch 1:
{
  key: [0, 0, 1]
  values: ["a", "b", "c", "d"]
}
// col from batch 2:
{
  key: [2, 3, 2]
  values: ["a", "b", "c", "d"]
}
```
This would be kind of unfortunate because, when we export this, by the OTAP exporter say, we end up transmitting data that's not needed.

Would it be worth exploring an option where we try to do something clever with the data layout? For example, currently a single stream is a standard Arrow IPC File formatted segment, which would have messages:
```
0x00 RecordBatch0
0x10 Dictionary0
0x20 Dictionary1
0x30 RecordBatch1
0x40 RecordBatch2
```

And then a footer with blocks:
```
recordBatches: [ { offset: 0x00 }, {offset: 0x30 }, { offset: 0x40 } ]
dictionaries: [ { offset: 0x10 }, { offset: 0x20 } ]
```

But there's not really anything stopping us from writing whatever we want in the segments eh? For example, maybe we wanted to write the dictionaries for every batch, and when we read the batch, we could also read back the dicts. For example, imagine a data layout like this:
```
0x00 RecordBatch0
0x10 Dictionary0
0x20 Dictionary1
0x30 RecordBatch1
0x40 Dictionary0 // dicts for RecordBatch1
0x50 Dictionary1
0x60 RecordBatch2
0x70 Dictionary0 // dicts for RecordBatch2
0x80 Dictionary1
```
footer
```
recordBatches: [ { offset: 0x00 }, {offset: 0x30 }, { offset: 0x60 } ]
dictionaries: [ { offset: 0x10 }, { offset: 0x20 }, { offset: 0x40 }, { offset: 0x50 }, { offset: 0x70 }, { offset: 0x80 } ]
```

So in our reader, we'd know that there are 2 dictionaries (we know this from the schema). When we read record batch `N`, we use the dictionaries in blocks `num_dicts*N..(num_dicts+1)*N`

Here's a working example of this
https://github.com/albertlockett/otel-arrow/blob/aafd439328431d7605a68be92a6c775cb637b184/rust/otap-dataflow/crates/quiver/src/lib.rs#L171-L390
(there's a few minor hacks that we could maybe make cleaner/more performant with some minor contributions to arrow_ipc).

I think in this case, the CPU should be lower than in the dict deduping case. The segment file would be bigger because we have extra blocks in the footer (although these are quite small), and dictionary values aren't deduplicated (no worse than what we have with the current solution on main). There is extra work happening in the reader to decode new dictionaries for each batch, but I also believe this should be relatively little work due to the "zero copy" nature of arrow.

The CPU might actually be lower on the writer too, since we're resetting the `DictionaryTracker` between batches, we don't need to compare the dicts for each batch to ensure they're the same.

One drawback is that wouldn't be able to just read the read data straight up w/ an arrow IPC file reader because we've done something a bit non-standard. However, I'm not sure this is actually a problem if we always plan to read this using quiver reader. 

The other drawback is that if the reader was disk/object store backed (which it's not currently, looks like we use a memory backed Buffer), we'd have a few extra IOs per batch to read the dictionaries and the record batches, although these could be coalesced.

Note, we don't have to layout the data _exactly_ this recommendation. My intention here was just to point out that we have feasible options for how we read/write the data that could help us avoid the complex work of deduping the dicts in the writer.

TL;DR - fine w/  dedupe in writer as recommended, but might be worth exploring changing the data layout to avoid the problem of orphaned dict values & CPU overhead?

PR #2566, opened 2026-04-07
feat: Add more metrics signal in fake data generator

ThomsonTan:
# Change Summary

Updates the static metrics signal generator from a minimal 2-type (Sum/Gauge) alternation with fixed attributes to a realistic multi-type signal covering all five OTLP metric kinds, with configurable attributes and data points per metric.

* Closes #NNN

The changes are tested via 5 new unit tests

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/indexed_min_heap.rs
+++ b/rust/otap-dataflow/crates/engine/src/indexed_min_heap.rs
@@ -1,0 +1,491 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! A generic indexed binary min-heap with O(1) key lookup and O(log n)
+//! insert, remove, and in-place priority update.
+//!
+//! The heap stores `(K, P)` pairs where `K` is a unique key and `P` is a
+//! priority value.  A companion `HashMap<K, usize>` tracks each key's
+//! current position in the underlying `Vec`, enabling efficient keyed
+//! operations that a standard `BinaryHeap` cannot provide.
+
+use std::collections::HashMap;
+use std::hash::Hash;
+
+/// Outcome of an [`IndexedMinHeap::insert`] call.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum InsertOutcome {
+    /// A new entry was added to the heap.
+    Inserted,
+    /// An existing entry's priority was replaced.
+    Replaced,
+}
+
+/// A binary min-heap that supports O(1) key lookup and O(log n) keyed
+/// insertion, removal, and in-place priority updates.
+///
+/// Entries are `(K, P)` pairs.  The heap is ordered by `P` where the
+/// *smallest* priority sits at the root.  Keys must be unique — inserting
+/// a key that already exists replaces its priority in place.
+///
+/// # Key cloning
+///
+/// Keys are cloned during insert, swap, pop, and removal operations.
+/// Callers should prefer cheap-to-clone key types (e.g. integers).
+//  Using an expensive-to-clone key will add overhead proportional to
+//  tree depth on every heap operation.
+pub(crate) struct IndexedMinHeap<K, P> {
+    entries: Vec<Entry<K, P>>,
+    indices: HashMap<K, usize>,
+}
+
+#[derive(Clone, Debug)]
+struct Entry<K, P> {
+    key: K,
+    priority: P,
+}
+
+impl<K, P> IndexedMinHeap<K, P>
+where
+    K: Eq + Hash + Clone + std::fmt::Debug,
+    P: Ord,
+{
+    /// Creates an empty heap.
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            entries: Vec::new(),
+            indices: HashMap::new(),
+        }
+    }
+
+    /// Returns the number of entries in the heap.
+    #[must_use]
+    pub fn len(&self) -> usize {
+        self.entries.len()
+    }
+
+    /// Returns `true` if the heap contains no entries.
+    #[must_use]
+    pub fn is_empty(&self) -> bool {
+        self.entries.is_empty()
+    }
+
+    /// Returns a reference to the `(key, priority)` pair with the smallest
+    /// priority, or `None` if the heap is empty.
+    #[must_use]
+    pub fn peek(&self) -> Option<(&K, &P)> {
+        self.entries.first().map(|e| (&e.key, &e.priority))
+    }
+
+    /// Returns `true` if the heap contains an entry for `key`.
+    pub fn contains_key(&self, key: &K) -> bool {
+        self.indices.contains_key(key)
+    }
+
+    /// Inserts `key` with the given `priority`.
+    ///
+    /// If `key` already exists its priority is replaced in place and the
+    /// heap property is restored.  Returns whether the key was newly inserted
+    /// or replaced.
+    pub fn insert(&mut self, key: K, priority: P) -> InsertOutcome {
+        if let Some(&index) = self.indices.get(&key) {
+            self.entries[index].priority = priority;
+            self.repair_at(index);
+            InsertOutcome::Replaced
+        } else {
+            let index = self.entries.len();
+            self.entries.push(Entry {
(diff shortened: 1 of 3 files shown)


rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
```diff
@@ -47,13 +49,273 @@ fn static_span_attributes() -> Vec<KeyValue> {
...
+    },
+    MetricDef {
+        name: "http.server.request.duration.summary",
+        description: "HTTP request duration quantiles",
+        unit: "ms",
+        kind: MetricKind::Summary,
+    },
+];
+
+/// Standard histogram bucket boundaries for duration metrics (in ms).
```
cijothomas: lets pick OTEL default bounds?

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
```diff
@@ -609,6 +1111,71 @@ mod tests {
...
+        let mut extra = HashMap::new();
+        _ = extra.insert("tenant.id".to_string(), "prod".to_string());
+        let metrics = static_otlp_metrics(5, Some(&extra));
+        let attrs = &metrics.resource_metrics[0]
+            .resource
+            .as_ref()
+            .unwrap()
+            .attributes;
+        assert!(attrs.iter().any(|kv| kv.key == "tenant.id"));
+    }
```
cijothomas, line 1189: can you add a test to validate the compression ratio, similar to what we added for logs?

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
```diff
@@ -410,42 +690,257 @@ fn static_spans(signal_count: usize) -> Vec<Span> {
...
+            let total_count: u64 = positive_buckets.bucket_counts.iter().sum();
+            let sum = total_count as f64 * 12.5;
+
+            ExponentialHistogramDataPoint::build()
+                .time_unix_nano(timestamp)
+                .start_time_unix_nano(start_time)
+                .count(total_count)
+                .sum(sum)
+                .scale(4)
+                .zero_count(2u64)
```
lalitb: This `zero_count` should also be added to `total_count `?

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
```diff
@@ -410,42 +690,257 @@ fn static_spans(signal_count: usize) -> Vec<Span> {
...
-            }
+            HistogramDataPoint::build()
+                .time_unix_nano(timestamp)
+                .start_time_unix_nano(start_time)
+                .count(total_count)
+                .sum(sum)
+                .bucket_counts(dist.to_vec())
+                .explicit_bounds(HISTOGRAM_BOUNDS.to_vec())
+                .min(midpoints[0])
+                .max(midpoints[dist.len() - 1])
```
lalitb: Is this correct - `min` and `max` are supposed to describe the actual observed values, not just fixed bucket positions.

lalitb:
This looks good for making the static metrics generator much more useful than the previous minimal implementation. One high-level suggestion as subsequent PR - consider making the metric generation more invariant-driven, where one underlying synthetic distribution is chosen first and the related fields are derived from it. That would reduce the chance of internal inconsistencies (as we see now) and make the generated metrics feel more realistic.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
```diff
@@ -410,42 +690,257 @@ fn static_spans(signal_count: usize) -> Vec<Span> {
...
+                2 => Buckets::new(-1, vec![5, 15, 40, 90, 60, 25, 8, 3]),
+                _ => Buckets::new(0, vec![30, 45, 60, 45, 30, 15, 5]),
+            };
+            let total_count: u64 = positive_buckets.bucket_counts.iter().sum();
+            let sum = total_count as f64 * 12.5;
+
+            ExponentialHistogramDataPoint::build()
+                .time_unix_nano(timestamp)
+                .start_time_unix_nano(start_time)
+                .count(total_count)
```
jmacd:
```suggestion
                .count(total_count+zeros)
```

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
jmacd:
```suggestion
                .zero_count(zeros)
```

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
```diff
@@ -410,42 +690,257 @@ fn static_spans(signal_count: usize) -> Vec<Span> {
...
+
+            // Vary the bucket distribution per point
+            let positive_buckets = match point_idx % 4 {
+                0 => Buckets::new(0, vec![100, 80, 40, 15, 5, 2, 1]),
+                1 => Buckets::new(1, vec![20, 50, 120, 80, 30, 10]),
+                2 => Buckets::new(-1, vec![5, 15, 40, 90, 60, 25, 8, 3]),
+                _ => Buckets::new(0, vec![30, 45, 60, 45, 30, 15, 5]),
+            };
+            let total_count: u64 = positive_buckets.bucket_counts.iter().sum();
+            let sum = total_count as f64 * 12.5;
```
jmacd, line 899:
```suggestion
            let sum = total_count as f64 * 12.5;
            let zeros = 2u64;
```
Not that anyone checks--

As this demonstrates, the zero count is redundant information, adding it to total is correct. We should add a data validation check at some point for histograms to verify that counts are consistent. I've taken this point to simply ignore zeros in the histogram aggregator, no need to count them because you have to iterate the buckets and zero count equals the difference between the buckets and the total count.

(As for the sum, we might imagine some kind of sanity check--is the sum in line with the distribution of buckets? However we do not require an exact table lookup in OTel, so this would have to tolerate some error.)

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
ThomsonTan: Done -- switched to the OTel SDK spec defaults `[0, 5, 10, 25, 50, 75, 100, 250, 500, 750, 1000, 2500, 5000, 7500, 10000]` and updated the bucket distributions and midpoints to match the new 16-bucket layout.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
ThomsonTan: Added `test_metrics_compression_ratio_is_realistic

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
ThomsonTan: Fixed -- `min`/`max` are now derived by scanning for the first and last non-zero bucket and using their midpoints, rather than hardcoding `midpoints[0]`/`midpoints[last]`.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
ThomsonTan: Committed locally.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/static_signal.rs
ThomsonTan: Applied all three suggestions. Appreciate the context on zero_count redundancy and the aggregator design.

Issue #2556, opened 2026-04-06
Implement `NodeKind::ProcessorChain`


drewrelmas:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

Consumers of otap-dataflow may have logical operations that consist of multiple invocations of different processors (i.e. optimize performance by using `attributes` processor to efficiently rename attributes and then `transform` processor for a more complicated manipulation).

While this can already be implemented with multiple processors and connections between them, one gap is the metric reporting from this configuration. Many processors opt-in to the [process_duration metrics](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/engine/src/process_duration.rs) like `compute.success.duration`. As it stands today, customers would get individual metric reporting from each individual processor.

### Proposed Solution

There has been an unimplemented [`NodeKind::ProcessorChain`](https://github.com/open-telemetry/otel-arrow/blob/092500ce98b89039cbc199b3368be66f4f08ba93/rust/otap-dataflow/crates/config/src/node.rs#L96) in the repository for a while now which would help satisfy this use case.

https://github.com/open-telemetry/otel-arrow/blob/092500ce98b89039cbc199b3368be66f4f08ba93/rust/otap-dataflow/crates/config/src/node.rs#L80-L97

`ProcessorChain` would enable reporting single, aggregated metrics for all internal nodes that satisfy the necessary conditions.

As an example, report a single `compute.success.duration` as a sum of CPU time used for all processors inside the `ProcessorChain`. This allows consumers of otap-dataflow to control the granularity of reporting on the logical operations they are achieving with multiple processor invocations.

The most in-depth implementation for this feature would require:
* Implement the existing `NodeKind::ProcessorChain`
  * A single node holds `N` sub-processors, calls them sequentially in one `process()`
* Composite `ComputeDuration` wraps the entire chain
  * Sub-processors can still report their own?

#### Duration accumulator on PData

- Accumulate duration in the `Context` stack carried by each `PData` batch
- Each grouped processor adds its elapsed time to the accumulator as data passes through
- Last processor in the group reads the accumulated value and reports the composite metric
- Couples processors to PData internals; "last processor" detection is fragile; adds hot-path overhead

#### Statistical aggregation at metrics layer

- Group node metrics by config label, sum Mmsc values across grouped processors at collection/export time
[...]

PR #2569, opened 2026-04-07
feat: Implement ProcessorChain for composite metric reporting

### LINKED ISSUE #2576: Use `NodeUserConfig` for `ProcessorChain` `processor` configs
Context: https://github.com/open-telemetry/otel-arrow/pull/2569#discussion_r3048298978

In an ideal world, the `ProcessorChain` `processors` would support the exact same config options as regular top-level nodes (named outputs, entity configuration, header_capture, etc).

For the initial implementation, it makes sense to have a separate minimal configuration. 

At the very least, `NodeUserConfig` would need to support an optional `name` property. It isn't required for regular nodes because name (and `node.id`) is already implied by config structure. For `SubProcessors` (which appear in list format) the user can optionally specify an override of the default index-based naming (`chain_name/0`, `chain_name/1`).

### LINKED ISSUE #2579: Initial implementation of `processor_chain`

drewrelmas:
# Change Summary

Implements `ProcessorChain` — a composite node that runs multiple sub-processors sequentially within a single task, eliminating inter-processor channels. The chain reports a single `compute.success.duration` (or`compute.failed.duration`) metric covering the total compute cost of a record batch passing through all sub-processors, while each sub-processor still reports its own individual duration.

## Motivation

When a single logical operation involves multiple internal processors for performance optimization (e.g., attributes -> condense -> recordset KQL), operators need a single duration metric representing the total cost of that logical operation per batch. Without this, the only option is statistical aggregation at the metrics layer, which produces incorrect min/max values (`min(A) + min(B) != min(A+B)`).

The `ProcessorChain` approach gives per-batch composite timing with correct Mmsc distribution (min/max/sum/count).

## Config syntax

```yaml
version: otel_dataflow/v1
engine: {}
groups:
  default:
    pipelines:
      main:
        policies:
          telemetry:
            runtime_metrics: normal
          channel_capacity:
            control:
              node: 100
              pipeline: 100
            pdata: 128

        nodes:
          receiver:
            config:
              traffic_config:
                max_batch_size: 1
                signals_per_second: 1
                log_weight: 100
          insert_A:
            type: processor:attribute
            config:
              actions:
                - action: insert
                  key: pre_chain
                  value: A
          chain:
            type: processor_chain:composite
            config:
              processors:
                insert_B:
                  type: processor:attribute
                  config:
                    actions:
                      - action: insert
                        key: chain_step_1
                        value: B
                insert_C:
                  type: processor:attribute
                  config:
                    actions:
                      - action: insert
                        key: chain_step_2
                        value: C
          insert_D:
            type: processor:attribute
            config:
              actions:
                - action: insert
                  key: post_chain
                  value: D
          debug:
            type: processor:debug
            config:
              verbosity: detailed
              mode: signal
          noop:
            type: exporter:noop
            config:

        connections:
          - from: receiver
            to: insert_A
          - from: insert_A
            to: chain
          - from: chain
            to: insert_D
          - from: insert_D
            to: debug
          - from: debug
            to: noop
```

## Telemetry output

With `runtime_metrics: normal`, this pipeline produces `compute.success.duration` for all 5 of the following:

| `node.id` | `node.type` | `node.urn` |
| --------- | ----------- | ---------- |
| `insert_A` | `processor` | `urn:otel:processor:attribute` |
| `chain` | `processor_chain` | `urn:otel:processor_chain:composite` |
| `chain/insert_B` | `processor` | `urn:otel:processor:attribute` |
| `chain/insert_C` | `processor` | `urn:otel:processor:attribute` |
| `insert_D` | `processor` | `urn:otel:processor:attribute` |

The composite duration is always >= the sum of sub-processor durations (it includes inter-stage overhead).

Querying metrics locally shows the following result:
> curl -s "http://127.0.0.1:8080/api/v1/telemetry/metrics?format=json" | jq '[.metric_sets[] | select(.name == "processor.compute") | {node_id: .attributes["node.id"].String, node_type: .attributes["node.type"].String, node_urn: .attributes["node.urn"].String, success: (.metrics[] | select(.name == "compute.success.duration") | .value | {avg_ms: (if .count > 0 then .sum / .count / 1e6 else 0 end), count})}]'

```json
[
  {
    "node_id": "insert_A",
    "node_type": "processor",
    "node_urn": "urn:otel:processor:attribute",
    "success": {
      "avg_ms": 1.2843131327433628,
      "count": 226
    }
  },
  {
    "node_id": "insert_D",
    "node_type": "processor",
    "node_urn": "urn:otel:processor:attribute",
    "success": {
      "avg_ms": 0.36642488938053097,
      "count": 226
    }
  },
  {
    "node_id": "chain",
    "node_type": "processor_chain",
    "node_urn": "urn:otel:processor_chain:composite",
    "success": {
      "avg_ms": 0.9586159513274336,
      "count": 226
    }
  },
  {
    "node_id": "chain/insert_B",
    "node_type": "processor",
    "node_urn": "urn:otel:processor:attribute",
    "success": {
      "avg_ms": 0.4922827168141593,
      "count": 226
    }
  },
  {
    "node_id": "chain/insert_C",
    "node_type": "processor",
    "node_urn": "urn:otel:processor:attribute",
    "success": {
      "avg_ms": 0.37540364159292033,
      "count": 226
    }
  }
]
```

## Design decisions

- **Vec-backed buffers**: Each intermediate sub-processor (all except the last) gets a buffer `EffectHandler` wired to a shared `Rc<RefCell<Vec<PData>>>` via a `VecSender`. When the sub-processor calls send_message(), the output is pushed directly into the `Vec` — no channel send/recv, waker management, or async polling overhead. These buffer handlers are created once at construction and reused for every batch.
- **Single-item fast path**: For chains of length 1, the chain delegates directly to the sole sub-processor with zero staging overhead. For length ≥ 2, an optimistic fast path threads a single `PData` value through each intermediate stage without any `Vec` operations. Only when a stage produces 0 or 2+ outputs does it fall back to staging vecs (`stage_a`/`stage_b`) which swap roles via `std::mem::swap` and retain heap capacity across calls.
- **Last sub-processor uses real EffectHandler**: The last sub-processor in the chain doesn't need a buffer — it sends directly through the real downstream channel, like any normal processor would.
- **Sub-processor entity registration**: Each sub-processor gets a distinct telemetry entity with its own `node.id` (e.g., `chain/insert_B`) and `node.urn` via `with_node_telemetry_handle` scoping during factory creation. This ensures sub-processor metrics are clearly separated from the chain's composite metrics in telemetry output.
- **CollectTelemetry forwarding**: The real `MetricsReporter` from the `CollectTelemetry` control message is forwarded to sub-processors so their metric snapshots reach the telemetry registry (not an orphaned channel).
- **`NodeKind::ProcessorChain`**: Reuses the existing (previously unused) config variant. Maps to `NodeType::Processor` in the engine so it participates in normal processor wiring.

### ~300µs simulated work per processor (1,000 batches, single-threaded LocalSet):

| Chain len | Chained (ms) | Separate (ms) | Overhead |
|-----------|--------------|---------------|----------|
| 1         | 301.6        | 301.4         | +0.1%    |
| 2         | 603.4        | 604.0         | -0.1%    |
| 3         | 904.9        | 904.9         | 0.0%     |

With realistic per-processor compute (~300µs, approximating processor work), the chain overhead is within noise — effectively zero.

### ~100ns simulated work per processor (1,000 batches, single-threaded LocalSet):

| Chain len | Chained (µs) | Separate (µs) | Overhead |
|-----------|--------------|---------------|----------|
| 1         | 305          | 320           | **-5%**  |
| 2         | 713          | 661           | +8%      |
| 3         | 1,033        | 990           | +4%      |

At trivially low work (100ns per processor), the chain matches or beats separate tasks for `len=1` thanks to a single-processor fast path. For `len>=2`, the remaining ~5-8% gap is the cost of per-stage `Vec`/`RefCell` bookkeeping, which is significant only at these artificially low work levels.

The chain's value is not raw throughput — it's the ability to produce a correct composite duration metric (min/max/sum/count) across all sub-processors, which is impossible with separate processors.

## Future work

- Processor chain visualization in admin UI (expand sub-processors in detail panel): #2575 
- Single `processor_chain` config switch to opt internal nodes out of reporting their own `duration` metrics: #2577

* Progress towards #2556
* Closes #2579
* Closes #2576

Unit tests, benchmarks, and running fake config locally

Yes, users can now configure `processor_chain` elements in their configuration.


drewrelmas:
Marking as draft again, upon further review of the benchmarks I don't think this is a strictly valid comparison at this iteration. Theoretically the chain benchmark should be equivalent or faster than the regular path.

rust/otap-dataflow/crates/engine/src/config.rs
```diff
@@ -171,3 +171,87 @@ impl ExporterConfig {
...
+///     processors:
+///       - type: processor:attribute
+///         config: { ... }
+///       - type: processor:condense_attributes
+///         config: { ... }
+/// ```
+#[derive(Debug, Clone, serde::Deserialize)]
+pub struct ProcessorChainConfig {
+    /// Configurations for the processors in the chain, in order.
+    pub processors: Vec<SubProcessorConfig>,
```
gouslu: seems like this could be NodeUserConfig instead?

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -94,6 +94,22 @@ impl ComputeDuration {
...
+    #[must_use]
+    pub const fn acc_success(&self) -> &Cell<Mmsc> {
+        &self.acc_success
+    }
+
+    /// Returns a reference to the failed accumulator cell.
+    #[must_use]
+    pub const fn acc_failed(&self) -> &Cell<Mmsc> {
+        &self.acc_failed
+    }
```
gouslu: If you want to expose these, maybe limit it to the crate? but I'd say maybe an API like `self.composite_duration.record(elapsed, result.is_ok());` would suffice? I think keeping the logic encapsulated is better in this case.

rust/otap-dataflow/crates/engine/src/config.rs
drewrelmas: This was an intentional decision:
* Chained processor config structure optionally allows setting a node 'name' (used for `node.id`) for each sub-processor in the list which is implied for regular nodes by config structure - `NodeUserConfig` doesn't have a 'name' option.
* SubProcessors (currently) don't support some of the more advanced parts of `NodeUserConfig` like named outputs, header_capture, etc.

I will add a comment here explaining why this choice was made, and tracking in #2576.

rust/otap-dataflow/crates/engine/src/process_duration.rs
drewrelmas: Great point - incorporating in next commit.

gouslu:
Putting here what we discussed offline:

I was wondering if ProcessorChain in this PR can be a ChainProcessor instead :D

### 1. ChainProcessor (user-facing composition)

This is what the PR implements today, but reframed: it's just another processor that allows you to compose multiple processors inside it. It implements the `Processor` trait, it's registered via a normal `ProcessorFactory`, and ideally requires minimal or no engine changes. The engine doesn't need to know it's special — it's a processor of processors and can have more logic in it.

### 2. ProcessorChain (engine-level optimization)

NOTE: This section is exploratory to share an idea, rather than a specific implementation.

This is a separate concept. A processor chain at the engine level would be about **optimization**, not about adding functionality. The idea would be to separate the pure data processing from message handling in processors, something like:

```rust
// Full processor — owns its message loop
trait Processor {
    async fn process(&mut self, channel, effect_handler);
}

// Pure data transform — no message loop, no channels
trait PDataProcessor {
    fn process_pdata(&mut self, pdata: PData) -> Vec<PData>;
}
```

A processor that implements `PDataProcessor` advertises that its data processing is pure and stateless with respect to the message loop. The engine can then automatically fuse linear sequences of `PDataProcessor`s into a single task — no user configuration needed, no new YAML syntax. The engine handles the message loop once at the outer level and just calls `a.process_pdata()` → `b.process_pdata()` → `c.process_pdata()` internally.

This keeps the two concerns separate:
- **ChainProcessor**: user wants to group processors explicitly for composite metrics or organizational reasons — it's just a processor, no engine changes needed.
- **ProcessorChain**: engine detects fusable processors and optimizes them transparently — this is an engine concern, not a user concern.

rust/otap-dataflow/crates/engine/src/config.rs
drewrelmas: Talked more offline - there's an alternative worth pursuing that gets around both of the problems mentioned.
1. Make the `processors` config a map instead of an array (mirroring how top-level nodes work)
2. Disallow the more default features from processor config in the chained case via validation in the chain's create function (in fact, maybe don't need to do this at all)

I will try this out

drewrelmas: Done

lquerel:
Two very brief comments on this PR:
- The benefit of chaining processors cannot be measured only in terms of CPU usage. The main benefit is in memory consumption. By eliminating channels, we reduce the amount of `pdata` accumulated in transit through those channels, and therefore reduce memory usage across the entire pipeline.
- Originally, the idea behind the `NodeKind::ProcessorChain` variant was to allow the engine to automatically infer processor chains in order to transparently eliminate channel overhead, especially memory overhead. What I like about this PR is that, as a first step, we can make this concept explicit and then add support for this optimization in a future PR.
  - The difference I see between the explicit version and the implicit version is the transparency of this optimization at the level of the generated entities and signals. In the current PR, the sub-processors, and therefore the corresponding entities, are scoped by the chain, which seems like the logical behavior a user would expect.
  - For the future implicit version, ideally we should preserve transparency, meaning that the processor entities would remain the original ones, with however one additional entity corresponding to the chain, as in the explicit version.

jmacd:
I like the memory-reduction aspect of the proposal, support on those grounds.

This is a comment on the instrumentation question, not on the processor chain technique. This is a "what if" question--what if we placed a logical stopwatch into the Context; using the helper API we have for measuring compute time, we'll just increment the stopwatch and return it to the context. The pipeline already instruments the point where contexts enter and exit a node. When the context exits we record a process-compute-time measurement. This is a meaningful measurement for individual nodes as well as across any pair of connected nodes, accumulating compute time in the direction of travel.

lalitb:
Nice work, the PR description made the design easy to follow.

One thing I’m still curious about is whether chained processors are expected to get the full effect-handler/runtime behavior. The intermediate stages look lighter-weight than a top-level processor (Vec-backed, default-port-only, no obvious timer/wakeup support), which seems totally fine for simple transforms/filters, but I’m less sure about processors like `retry_processor` or `fanout_processor` that depend on richer runtime behavior. Since the config still accepts a full sub-processor node config, that mismatch would likely only show up at runtime.

If the intent is to support a simpler subset for now, it might be worth either documenting that clearly or rejecting incompatible processors up front in `create_processor_chain`.

gouslu:
> Nice work, the PR description made the design easy to follow.
> 
> One thing I’m still curious about is whether chained processors are expected to get the full effect-handler/runtime behavior. The intermediate stages look lighter-weight than a top-level processor (Vec-backed, default-port-only, no obvious timer/wakeup support), which seems totally fine for simple transforms/filters, but I’m less sure about processors like `retry_processor` or `fanout_processor` that depend on richer runtime behavior. Since the config still accepts a full sub-processor node config, that mismatch would likely only show up at runtime.
> 
> If the intent is to support a simpler subset for now, it might be worth either documenting that clearly or rejecting incompatible processors up front in `create_processor_chain`.

I agree with the distinction of whether a processor supports it or not. In [this comment](https://github.com/open-telemetry/otel-arrow/pull/2569#issuecomment-4209829182) the `PDataProcessor` idea (or sth like that) was meant to address that need. It would be good to have processors distinguish themselves.

utpilla:
Is the primary motivation composite metrics? If so, @jmacd's [context-propagated stopwatch suggestion](https://github.com/open-telemetry/otel-arrow/pull/2569#issuecomment-4216012183) seems better suited and worth exploring. It could achieve correct composite duration metrics (`min`/`max`/`sum`/`count` across a group of processors) without architectural changes. Processors would stay as separate tasks with bounded channels, no engine changes needed, and it could work across any pair of connected nodes rather than only explicitly configured chains. There are implementation details to work out (attaching timing to PData, defining composite groups, handling fan-out), but the core idea avoids the trade-offs below.

I bring this up because the chaining approach introduces a couple of trade-offs that don't exist in the current model:

**Scheduling fairness regression.** The chain converts N separate tasks (each yielding between batches) into a single task that runs all intermediate stages without yielding. A single processor _could_ already starve others with a long-running implementation, but that's a localized bug. Chaining systematically increases uninterrupted CPU time per task. The total compute time is the sum of all sub-processors in the chain, during which the executor cannot schedule receiver accepts, exporter flushes, or timer ticks. The intermediate Vec-backed sends are plain Rust operations, not async channel sends, so they never give the executor a scheduling opportunity between stages. This could be mitigated with `yield_now().await` between stages.

**Unbounded intermediate buffer.** The intermediate Vec buffer is unbounded. `push()` always succeeds with no capacity check. In practice this is fine for well-behaved 1:1 processors (one output per input), which is the common case. However, the existing model's bounded mpsc channels put a ceiling on how much memory one stage can consume at a time. If a processor calls `send_message()` multiple times per input, the channel's capacity limit applies backpressure by yielding when full. The chain removes that ceiling.

**Burden on the user.** Both issues above are regressions from the existing pipeline model. As @lalitb and @gouslu noted, we'd want to restrict which processors can be chained. But validating compatibility isn't just about checking for compute-only processors. Even a chain of purely compute/transform processors has the scheduling fairness issue described above. The safety constraints are architectural (chain length, per-processor compute cost, output cardinality), not just a property of the processor type.

### Chaining has its own merits

The chain's benefit of memory reduction from eliminating intermediate channels is a valid motivator (as @lquerel noted), separate from composite metrics. But that benefit needs to be weighed against the regressions above, which would need guardrails (yield points, buffer bounds, processor compatibility validation) to reach the safety level of the existing model.

drewrelmas:
Thanks @gouslu @lquerel @jmacd @lalitb @utpilla for the feedback. I've written my latest thoughts at https://github.com/open-telemetry/otel-arrow/issues/2556#issuecomment-4239313054 and will be closing this PR while we continue on this design.

PR #2571, opened 2026-04-07
feat(temporal_reaggregation_processor): Add processor microbenchmarks

JakeDern:
# Change Summary

This is a PR to add some microbenchmarks for the temporal_reaggregation_processor. This is nice vs using the e2e framework to isolate the processor and be able to measure improvement beyond %cpu granularity.

* Part of #2422

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/Cargo.toml
+++ b/rust/otap-dataflow/crates/core-nodes/Cargo.toml
@@ -90,11 +90,7 @@ pretty_assertions.workspace = true
 tempfile.workspace = true
 portpicker.workspace = true
 otap-test-tls-certs.workspace = true
-criterion = { workspace = true, features = ["html_reports"] }
-core_affinity.workspace = true
-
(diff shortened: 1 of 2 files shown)


albertlockett (approved):
LGTM!

Out of curiosity, I'd be interested to know what are the results

Also, do you think it makes sense to scan over different amounts of data? E.g. we could run the benchmark for varying metrics per batch / num batches? I guess this is something that could be added in a followup as well

JakeDern:
> LGTM!
> 
> Out of curiosity, I'd be interested to know what are the results
> 
> Also, do you think it makes sense to scan over different amounts of data? E.g. we could run the benchmark for varying metrics per batch / num batches? I guess this is something that could be added in a followup as well

It ended up being not that interesting 😅, but I guess that's a good thing!

Vast majority of time is spent building output batches which I think is a good sign because that's the real work product that this processor needs to do. The rest of the stuff like tracking identities and creating/walking views seems to be relatively low overhead.

image

PR #2572, opened 2026-04-07
feat(temporal_reaggregation_processor): Add Stream cardinality limit setting

JakeDern:
# Change Summary

This PR adds a Stream cardinality limit setting that can be used to help bound the internal data structures used. 

Only the Stream has a limit to keep it simple and because since every other Identity also rolls up into a StreamId this effectively imposes a limit across them all.

* Part of #2422 

Unit.

Yes, new config option.

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/config.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/config.rs
@@ -31,9 +31,7 @@ pub struct Config {
     pub outbound_request_limit: NonZeroUsize,
 
     /// Maximum number of unique metric streams to track in a single aggregating
-    /// batch. When exceeded, the current batch is flushed early and the
-    /// overflowing data is retried into a fresh batch. This provides an
-    /// imperfect bound on memory usage. Default: 32767 (half of u16::MAX).
+    /// batch. When exceeded, the current batch is flushed early.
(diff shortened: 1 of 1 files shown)


JakeDern:
> Is there any feedback to the user that they've hit this limit? If not, do you think that would be useful/needed? I see we increment the flushes_overflow metric, but it's unclear whether it was due to id overflow or stream cardinality exceeded.
> 
> Maybe in either case, the issue is just too many time series so maybe knowing which limit we've hit in this case isn't really all that useful considering the root cause is the same.

Good question! I was thinking the id overflow limit should basically never be hit given I set the max limit for stream cardinality is u16::MAX. I was leaning towards treating them the same since it's just too many metrics either way, but definitely open to adding a separate counter

albertlockett:
> > Is there any feedback to the user that they've hit this limit? If not, do you think that would be useful/needed? I see we increment the flushes_overflow metric, but it's unclear whether it was due to id overflow or stream cardinality exceeded.
> > Maybe in either case, the issue is just too many time series so maybe knowing which limit we've hit in this case isn't really all that useful considering the root cause is the same.
> 
> Good question! I was thinking the id overflow limit should basically never be hit given I set the max limit for stream cardinality is u16::MAX. I was leaning towards treating them the same since it's just too many metrics either way, but definitely open to adding a separate counter

oh yeah gotcha. Yeah that makes sense. I think we can keep this as is with the single counter

PR #2574, opened 2026-04-07
Add the public admin SDK and extract shared admin types

lquerel:
## Change Summary

This PR adds a public Rust admin SDK for the OTAP Dataflow Engine. This first version of the SDK relies on our existing HTTP admin endpoints. However, the SDK's design has been made with future support for other admin interfaces, such as OpAMP, in mind.

Concretely, this PR:
- adds `otap-df-admin-api` as the public integration crate for admin access
- makes the SDK configurable for HTTP and HTTPS, including custom CA and mTLS usage
- exposes the supported public admin endpoints through resource-scoped client methods
- introduces `otap-df-admin-types` as the shared schema crate and rewires `otap-df-admin` to depend on it directly
- updates validation to consume the new SDK
- expands the admin SDK documentation for integrators, including TLS, FIPS-oriented feature guidance, endpoint descriptions, examples, and future live reconfiguration direction.

No specific issue.

These changes were validated with:
- `cargo xtask check`

Yes.

External Rust integrators now have a dedicated public crate, `otap-df-admin-api`, for admin access. The SDK uses a builder plus resource-client API, documents HTTP/TLS/custom-CA/mTLS usage, and re-exports the config model so users only need one public crate.

There are no intended wire-format changes to the existing admin HTTP endpoints. The new `otap-df-admin-types` crate is internal-only and is not intended as a public dependency.

### DIFF
--- a/rust/otap-dataflow/benchmarks/benches/control_channel/main.rs
+++ b/rust/otap-dataflow/benchmarks/benches/control_channel/main.rs
@@ -141,7 +141,8 @@ async fn consume_current_local(
             NodeControlMsg::DrainIngress { .. }
             | NodeControlMsg::MemoryPressureChanged { .. }
             | NodeControlMsg::Shutdown { .. }
-            | NodeControlMsg::DelayedData { .. } => {
+            | NodeControlMsg::DelayedData { .. }
+            | NodeControlMsg::Wakeup { .. } => {
                 panic!("unexpected message in benchmark current local receiver");
             }
         }
@@ -205,7 +206,8 @@ async fn consume_current_shared(
             NodeControlMsg::DrainIngress { .. }
             | NodeControlMsg::MemoryPressureChanged { .. }
             | NodeControlMsg::Shutdown { .. }
-            | NodeControlMsg::DelayedData { .. } => {
+            | NodeControlMsg::DelayedData { .. }
+            | NodeControlMsg::Wakeup { .. } => {
                 panic!("unexpected message in benchmark current shared receiver");
             }
         }
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
@@ -1280,11 +1280,13 @@ mod tests {
 
         let (batch_received_tx, mut batch_received_rx) = tokio::sync::mpsc::channel(1);
         let (server_shutdown_tx, server_shutdown_rx) = tokio::sync::oneshot::channel();
+        let (server_ready_tx, server_ready_rx) = tokio::sync::oneshot::channel();
 
         // Start gRPC server that returns errors
         let listening_addr: SocketAddr = format!("{grpc_addr}:{grpc_port}").parse().unwrap();
         let server_handle = tokio_rt.spawn(async move {
             let tcp_listener = TcpListener::bind(listening_addr).await.unwrap();
+            let _ = server_ready_tx.send(());
             let tcp_stream = TcpListenerStream::new(tcp_listener);
             let error_service = ArrowLogsServiceServer::new(ArrowLogsServiceGrpcErrorMock {
                 sender: batch_received_tx,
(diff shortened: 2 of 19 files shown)


lalitb:
This is very helpful, and close to something we’re doing on our side for embedded/managed collector integrations. One thing I’m trying to clarify: the PR description says the SDK is designed with future support for other admin interfaces in mind, but `AdminBackend` is currently internal-only. Is the intent that alternate backends remain internal for now, or should there eventually be a public extension point for non-HTTP backends?

rust/otap-dataflow/crates/admin-api/src/http_backend.rs
```diff
@@ -0,0 +1,1002 @@
...
+    if let Some(true) = tls.insecure {
+        return Err(Error::ClientConfig {
+            details: "tls.insecure is not supported by the admin SDK; use AdminEndpoint::http(...) for plaintext connections".to_string(),
+        });
+    }
+
+    Ok(())
+}
+
+fn ensure_crypto_provider() -> Result<(), Error> {
```
lalitb, line 473: It may be worth making the crypto provider features mutually exclusive at compile time. With Cargo feature unification, a consumer could intend to use `crypto-openssl` but still end up with `crypto-ring` if both features are enabled. 

```rust
#[cfg(any(
    all(feature = "crypto-ring", feature = "crypto-aws-lc"),
    all(feature = "crypto-ring", feature = "crypto-openssl"),
    all(feature = "crypto-aws-lc", feature = "crypto-openssl"),
))]
compile_error!(
    "crypto provider features are mutually exclusive; enable exactly one"
);
```

rust/otap-dataflow/crates/admin-api/src/client.rs
```diff
@@ -0,0 +1,357 @@
...
+            .pipeline_status(pipeline_group_id, pipeline_id)
+            .await
+    }
+
+    /// Returns the plain-text liveness probe for one pipeline.
+    pub async fn livez(
+        &self,
+        pipeline_group_id: &str,
+        pipeline_id: &str,
+    ) -> Result<pipelines::ProbeResponse, Error> {
```
lalitb: One design question on the future multi-backend story: `pipelines().livez()` / `readyz()` currently return an HTTP-shaped `ProbeResponse` (status_code, body). That works well for the current HTTP client, but seems awkward for something like a future in-process backend, which would have to synthesize HTTP-like values rather than return the probe result directly. Is the public API meant to be transport-agnostic already, or is that flexibility internal-only for now?

lalitb:
> This has me sort of imagining a CLI like `df_enginectl --shutdown`

Love this use-case. A small df_enginectl-style CLI feels like a very natural fit for this SDK. I also feel this is relevant for MCP / agentic tooling. The OpenTelemetry community is already discussing official MCP-based workflows for the Collector ([agentic-workflow.md](https://github.com/open-telemetry/community/blob/main/projects/agentic-workflow.md)), so I can see this SDK being useful as a foundation for external MCP servers on top of the engine admin surface.

lquerel:
> This is very helpful, and close to something we’re doing on our side for embedded/managed collector integrations. One thing I’m trying to clarify: the PR description says the SDK is designed with future support for other admin interfaces in mind, but `AdminBackend` is currently internal-only. Is the intent that alternate backends remain internal for now, or should there eventually be a public extension point for non-HTTP backends?

The use case I have in mind is to support OpAMP between the SDK and our engine once our engine exposes an OpAMP interface.

rust/otap-dataflow/crates/admin-api/src/client.rs
lquerel: Ideally, I'd like to be as agnostic as possible about the underlying protocol. I'll think about a solution and see whether I can include it in this PR or in a follow-up PR.

rust/otap-dataflow/crates/admin-api/src/http_backend.rs
lquerel: See https://github.com/open-telemetry/otel-arrow/pull/2574/commits/0fc12df5db5c3db8db2e69518f554217b52a3373

rust/otap-dataflow/crates/admin-api/src/client.rs
lquerel: Done -> https://github.com/open-telemetry/otel-arrow/pull/2574/commits/66c1d2ee049cf06e9f4b6c4c33027ce6520594c6

PR #2580, opened 2026-04-07
feat(temporal_reaggregation_processor): Update README and ignore Map/KVList types for stream identification purposes

JakeDern:
# Change Summary

This PR updates the README and also changes the identity module to ignore Map/KVList types for stream identification purposes.

The justification for the identity change is that our OTAP view does not support these and it would be difficult to do so as we would need to either convert all the CBOR data to owned types on the side and then borrow them or find/invent a new CBOR implementation that can support a view model.

For now, the limitations are documented and we can maybe alleviate them later.

* Part of #2422 

Unit.

No.

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/identity.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/identity.rs
@@ -341,8 +341,8 @@ impl AttributeHash {
         for attr in &entries {
             match attr.value().map(|v| v.value_type()) {
                 // We're currently treating Maps and list attributes as
-                // non-identifying because we don't have a great way to support
-                // them for OTAP views.
+                // non-identifying because we don't yet support these
+                // attribute types for OTAP views.
                 Some(ValueType::Array) | Some(ValueType::KeyValueList) => {
                     continue;
                 }


rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/identity.rs
```diff
@@ -339,6 +339,15 @@ impl AttributeHash {
         entries.sort_by(|a, b| a.key().cmp(b.key()));
 
         for attr in &entries {
+            match attr.value().map(|v| v.value_type()) {
+                // We're currently treating Maps and list attributes as
+                // non-identifying because we don't have a great way to support
+                // them for OTAP views.
```
albertlockett: nit: we probably can support them by implementing the view over the CBOR serialized data. Maybe we could touch up the comment just so it doesn't seem like something that's actually infeasable?
```suggestion
                // We're currently treating Maps and list attributes as
                // non-identifying because we don't yet support these
                // attribute types for OTAP views.
```

JakeDern:
> LGTM
> 
> I guess maybe we'll document the limitations somewhere for posterity eventually, like on #2422.

Definitely! I have the limitation documented in the README for now, but should probably file an issue as well.

> For the purposes of hashing these attribute types, I wonder if we'd actually be able to get away w/ hashing the raw CBOR binary value instead of building a view that decodes it? I guess we'd need some kind of `fn get_raw_value() -> &[u8]` method on the attrs view for this maybe?

The problem is just that the ciborium library we use only deserializes to owned types, so we could do that in the view and then return references to it if we wanted to.

I did think about just hashing it as binary, but we're supposed to sort by key for map types and it would be inconsistent with any OTLP formatted data which we do have support for these attribute types for.

I figured the safest thing would be to handle both formats the same to prevent any surprises.

albertlockett:
> > LGTM
> > I guess maybe we'll document the limitations somewhere for posterity eventually, like on #2422.
> 
> Definitely! I have the limitation documented in the README for now, but should probably file an issue as well.
> 
> > For the purposes of hashing these attribute types, I wonder if we'd actually be able to get away w/ hashing the raw CBOR binary value instead of building a view that decodes it? I guess we'd need some kind of `fn get_raw_value() -> &[u8]` method on the attrs view for this maybe?
> 
> The problem is just that the ciborium library we use only deserializes to owned types, so we could do that in the view and then return references to it if we wanted to.
> 
> I did think about just hashing it as binary, but we're supposed to sort by key for map types and it would be inconsistent with any OTLP formatted data which we do have support for these attribute types for.
> 
> I figured the safest thing would be to handle both formats the same to prevent any surprises.

Sounds good. We also have the serde_cbor library, which might have an API more suitable to the view implementation.

JakeDern:
> > > LGTM
> > > I guess maybe we'll document the limitations somewhere for posterity eventually, like on #2422.
> > 
> > 
> > Definitely! I have the limitation documented in the README for now, but should probably file an issue as well.
> > > For the purposes of hashing these attribute types, I wonder if we'd actually be able to get away w/ hashing the raw CBOR binary value instead of building a view that decodes it? I guess we'd need some kind of `fn get_raw_value() -> &[u8]` method on the attrs view for this maybe?
> > 
> > 
> > The problem is just that the ciborium library we use only deserializes to owned types, so we could do that in the view and then return references to it if we wanted to.
> > I did think about just hashing it as binary, but we're supposed to sort by key for map types and it would be inconsistent with any OTLP formatted data which we do have support for these attribute types for.
> > I figured the safest thing would be to handle both formats the same to prevent any surprises.
> 
> Sounds good. We also have the serde_cbor library, which might have an API more suitable to the view implementation.

Oh good to know, I only looked at ciborium. If serde_cbor supports a view model, then we can follow up with support for that

PR #2590, opened 2026-04-07
task(temporal_reaggregation_processor): Pull otap id state out to its own construct and remove passthrough batch state from the processor

JakeDern:
# Change Summary

This is just a bit of cleanup to the temporal_reaggregation_processor to pull the minimum common state for building passthrough and aggregated batches out to its own type. I also:
- Removed the passthrough batch state since from the processor itself since the lifetime of that state is only per call to `process_view`.
- Delegate some boilerplate to a derived `Default::default` implementation

* Part of #2422 

Unit.

No.


albertlockett (approved):
Looks reasonable, nice to see that extra un-needed state not being present where it's not needed

jmacd (approved):

PR #2594, opened 2026-04-07
Refactor config structs from internally-tagged enums to flat `type` + `config` for Kubernetes CRD compatibility

### LINKED ISSUE #2592: Config structs should be trivially compatible with K8s CRD schema
### Pre-filing checklist

### Component(s)

There are several enums used in the dataflow engine config which are problematic with Schemars visitor implementation for CRD generation described [here](https://github.com/kube-rs/kube/pull/779) and kubernetes rules around sub-schemas. Moreover, some of these are not totally aligned with the rest of the engine configs generally.

1. Internal Telemetry - MetricsPeriodicExporterConfig (Console or Otlp(OtlpExporterConfig))
2. Transport Header Policy - PropagationSelector (All or None or Named(Vec<String>))
3. Policy - CoreAllocation (AllCores, CoreCount(usize), CoreSet(Vec<CoreRange>))

### Proposed Solution

1. Internal Telemetry - MetricsPeriodicExporterConfig (Console or Otlp(OtlpExporterConfig))

The configs here mirror the go collector, but aren't aligned with how we specify configs for other nodes. We can align these with a struct containing a {type: otlp|console, config: Option<Value>} and then validate the supplied config after deserializing as we do with the other nodes.

2. Transport Header Policy - PropagationSelector (All or None or Named(Vec<String>))

We can switch this to {type: all|none|named, named: Option<Vec<String>>} . It will also eliminate the need for this serde tag stuff that isn't super user friendly:
```
selector: !named 
  - tenant_id
  - x-request-id 
```

3. Policy - CoreAllocation (AllCores, CoreCount(usize), CoreSet(Vec<CoreRange>))

Again we can separate this into a struct with type/strategy and optional count/set fields. This one probably doesn't break anything public facing since we're already taging as `type`

### Alternatives Considered

_No response_

### Additional Context

I have a PR ready for this if someone can assign to me.

clhain:
# Change Summary

&#128680;**This PR Includes Breaking Config API Changes**&#128680;

This branch replaces three internally-tagged/adjacently-tagged enum config types with flat structs using explicit `type` discriminator fields, making them representable as Kubernetes CRD structural schemas.

**What changed:**

- **Metrics exporter configs** (`MetricsPeriodicExporterConfig`, `MetricsPullExporterConfig`): Converted from custom `Deserialize` visitor-based enums (where the exporter name was a map key like `console:` or `otlp: { ... }`) to flat structs with `type` (enum discriminator) + `config` (`serde_json::Value` free-form blob). Exporter-specific configuration is now extracted and validated at runtime via `from_config()` helpers. This mirrors the node specification / configuration logic in the pipeline configs

- **`CoreAllocation`**: Converted from a `#[serde(tag = "type")]` enum with variants like `AllCores`, `CoreCount { count }`, `CoreSet { set }` to a flat struct with a `strategy` enum field plus optional `count` and `set` fields. Convenience constructors (`all_cores()`, `core_count(n)`, `core_set(ranges)`) replace direct enum variant construction.

- **`PropagationSelector`**: Converted from a `#[serde(rename_all)]` enum with variants `AllCaptured`, `None`, `Named(Vec<String>)` to a flat struct with `selector_type` enum + optional `named` list. Includes a new `validate()` method that rejects invalid combinations (e.g., `type: named` without a `named` list).

* Closes #2592

Validation tests added for post-deserialization handling, updated existing tests to new structs.

**Config file format changes** (YAML, breaking):

```yaml
# Before                          # After
exporter:                         exporter:
  otlp:                             type: otlp
    endpoint: "..."                 config:
                                      endpoint: "..."

selector: all_captured            selector:
                                    type: all_captured

selector: !named                  selector:
  - tenant_id                       type: named
  - x-request-id                    named:
                                      - tenant_id
                                      - x-request-id  


core_allocation:                  core_allocation:
  type: core_count                  type: core_count # Non-breaking alias to new `strategy` field.
  count: 4                          count: 4
```

### DIFF
--- a/rust/otap-dataflow/crates/config/src/engine.rs
+++ b/rust/otap-dataflow/crates/config/src/engine.rs
@@ -1792,7 +1792,7 @@ groups:
         let has_properties = spec_schema
             .properties
             .as_ref()
-            .map_or(false, |p| !p.is_empty());
+            .is_some_and(|p| !p.is_empty());
 
         assert!(
             has_properties,


lquerel:
I’d like to get @jmacd ’s and @drewrelmas ’s approval on these changes, but personally I think:
- they have very limited impact in the current state of the project,
- they are important to ensure overall consistency in our configuration model, while also having the benefit of being compatible with the constraints of Kubernetes CRDs,
- we should put in place a **CI check** to verify that this general structure, that is, the explicit use of the type discriminator, is enforced. I don't yet know how to set that up, but we could create a GitHub issue to think it through.

jmacd:
The motivation is great, I also like the outcome.

rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/readers/pull.rs
```diff
@@ -5,52 +5,38 @@
...
-    }
+/// OpenTelemetry Metrics Pull Exporter configuration.
+///
+/// The `type` field selects the exporter, and `config` holds the
+/// exporter-specific configuration as a free-form JSON value.
+/// For `prometheus`, `config` should contain a valid `PrometheusExporterConfig`.
+/// If `config` is omitted, Prometheus defaults (host=0.0.0.0, port=9090,
+/// path=/metrics) are used.
+#[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq)]
+pub struct MetricsPullExporterConfig {
```
lalitb, line 27: Something to be revisited post merge - I think this introduces a validation gap for the new `Prometheus` wrapper shape: if someone puts exporter-specific fields next to `type`: instead of under `config`:, we can now accept the config and silently fall back to Prometheus defaults instead of failing fast. That seems like a regression from the old enum-based shape, where the same mistake would have been rejected.

rust/otap-dataflow/crates/config/src/policy.rs
```diff
@@ -173,6 +173,24 @@ impl Policies {
                 )),
             }
         }
+
+        if let Some(resources) = &self.resources {
+            if let Err(e) = resources.core_allocation.validate() {
+                errors.push(format!("{path_prefix}.resources.core_allocation: {e}"));
+            }
+        }
+        if let Some(transport_headers) = &self.transport_headers {
```
lalitb, line 182: follow up post merge - I like the direction here, but I think the new selector validation wants to be enforced uniformly across all config entry points, not just one path. Otherwise the same invalid selector shape can be rejected in one place and silently accepted in another.

rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/readers/pull.rs
lalitb: Also good to add regression test for this.

rust/otap-dataflow/crates/config/src/policy.rs
clhain: good catches all around - will open a followup shortly. Thanks!

clhain:
Since the merge bounced I just updated here. Added the missing deny unknown fields + regression tests. Added selector validation where missing. 

@lquerel I added a dev dependency to kube-rs/k8s-openapi and a couple tests to just convert it to a CRD as an attempt to address the structural checks you mentioned (should catch other problems we haven't run into yet also). Let me know if this is too heavy and I can remove.

PR #2596, opened 2026-04-08
fix(otlp-receiver): make request_bytes metric consistent across HTTP and gRPC

cijothomas:
The `request_bytes` counter was recording compressed wire size on the HTTP path but decompressed size on the gRPC path (where tonic decompresses before handing bytes to the decoder). Move the HTTP recording to after decompression so both protocols report decompressed payload bytes for successfully received requests.

**Changes:**

- **`otlp_http.rs`**: Move `request_bytes.add()` from before `decode_content_encoding()` to after it, so it records decompressed payload size — matching the gRPC path.
- **`otlp_metrics.rs`**: Update doc comment to clarify the metric measures decompressed bytes for successfully received requests.

### DIFF
--- a/rust/otap-dataflow/crates/config/src/transport_headers.rs
+++ b/rust/otap-dataflow/crates/config/src/transport_headers.rs
@@ -15,6 +15,7 @@
 //! - Normalized logical names for policy matching
 
 use std::fmt;
+use std::sync::Arc;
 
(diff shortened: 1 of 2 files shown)


jmacd:
https://github.com/open-telemetry/otel-arrow/issues/2603

PR #2618, opened 2026-04-09
Add live pipeline reconfiguration and shutdown control to the admin API

### LINKED ISSUE #2617: Add live pipeline reconfiguration and shutdown control to the admin API
### Pre-filing checklist

### Component(s)

### Objective

Enable operators to manage logical pipelines in a running OTAP Dataflow Engine instance through the admin API, without restarting the engine. This includes creating a new pipeline in an existing group, updating an existing pipeline, resizing pipeline core allocation, detecting no-op updates, and shutting down an individual pipeline.

### Rationale

Today, pipeline topology and resource changes require engine restarts or full config redeploys, which increases operational cost and risk. The engine needs a resident control plane that can safely apply pipeline changes in place, track progress, expose status to operators, and preserve service continuity during updates.

### Scope

- Add admin API endpoints under `/api/v1/groups/...` for pipeline create/update and shutdown.
- Refactor the controller into a long-lived runtime manager with live pipeline registry/state.
- Support pipeline update modes:
  - create for new logical pipelines
  - rolling replace for topology/config changes
  - resize for resource-only core allocation changes
  - noop for identical effective configs
- Add rollout and shutdown operation tracking with status lookup endpoints.
- Extend runtime/state identity with deployment generations so overlapping instances remain distinguishable.
- Preserve existing `/status` compatibility while surfacing rollout metadata.
- Add operator documentation for live reconfiguration behavior and API usage.

### Acceptance Criteria

- `PUT /api/v1/groups/{group}/pipelines/{id}` can create a new pipeline in an existing group.
- The same endpoint can update an existing pipeline with a health-gated rolling cutover.
- Resource-only core allocation changes perform resize behavior without restarting unchanged cores.
- Identical effective pipeline updates return a successful noop result without restarting instances.
- `POST /api/v1/groups/{group}/pipelines/{id}/shutdown` shuts down a logical pipeline and exposes shutdown progress.
- Rollout and shutdown status can be queried through dedicated status endpoints.
- Overlapping old/new instances are distinguishable in observed state and status through deployment generation.
- Existing status payloads remain backward compatible, with rollout metadata added rather than replacing current fields.
- Controller and admin tests cover create, replace, resize, noop, shutdown, rollback/failure handling, and endpoint behavior.
- Operator documentation explains the workflow and API examples.

### Dependencies or Blockers

#831 will be supported in a future PR.

### Additional Context

- This work is pipeline-scoped only. Group-level multi-pipeline rollouts are out of scope (for now).
- Runtime config persistence is out of scope; changes are applied in memory only.

lquerel:
## Change Summary

This PR adds live pipeline lifecycle control to the OTAP Dataflow Engine admin API.

The controller now runs as a resident runtime manager and supports in-memory pipeline operations on a running engine:

- create a new pipeline in an existing group
- replace an existing pipeline with a health-gated serial rolling cutover
- resize a pipeline when only the effective core allocation changes
- detect identical updates and return a successful `noop` result
- shut down an individual logical pipeline and track shutdown progress

The PR also adds rollout and shutdown status resources, extends runtime and observed-state tracking with deployment generations so overlapping instances remain distinguishable, updates the Rust admin SDK with typed live-control methods/outcomes, and documents the feature.

## Design Decisions

This PR keeps live reconfiguration scoped to one logical pipeline at a time, keyed by `(group, pipeline)`. Reconfiguration mutations go through one declarative endpoint, `PUT /api/v1/groups/{group}/pipelines/{id}`, while pipeline shutdown remains a separate operation through `POST /api/v1/groups/{group}/pipelines/{id}/shutdown`.

The controller classifies each reconfiguration request as `create`, `noop`, `resize`, or `replace` instead of exposing separate start, scale, and update APIs.

Topology and node-configuration changes use a serial rolling cutover with overlap: start the new instance on one core, wait for `Admitted` and `Ready`, then drain the old instance on that core. This preserves continuity while fitting the existing runtime model, without requiring a full second serving fleet or a separate traffic-switching layer.

Pure core-allocation changes use a dedicated internal resize path so scale up/down only starts or stops the delta cores and leaves unchanged cores running.

Runtime mutations remain intentionally narrow. This PR keeps changes in memory only and rejects updates that would require topic-broker reconfiguration or broader engine/group-level policy mutation. Runtime config persistence remains out of scope.

Operations are explicit and observable through rollout and shutdown ids. Terminal operation history is intentionally bounded in memory, so old rollout/shutdown ids may eventually return `404` or `Ok(None)` from the SDK after retention pruning.

Observed status is deployment-generation-aware. During overlapping rollouts, `/status.instances` preserves old and new runtime instances, while aggregate readiness/liveness still uses the selected serving generation per core. After controller work completes, superseded observed instances are compacted so status memory does not grow unbounded across rollouts.

The controller also hardens lifecycle cleanup around live operations: worker panics are converted into terminal rollout/shutdown failure states, runtime exits are reported through supervised bookkeeping, and shutdown drain completion waits for upstream closure rather than queue emptiness alone.

* Closes #2617

Commands run:

- `cargo xtask check`

Automated coverage includes:

- rollout planning for create, replace, resize, scale up/down, noop, topic-mutation rejection, and conflict handling
- rollout execution success/failure paths, rollback behavior, worker panic cleanup, and bounded operation retention
- per-pipeline shutdown planning, progress tracking, conflict handling, timeout behavior, and worker panic cleanup
- observed-state generation selection, overlap-aware `/status.instances`, compaction of superseded generations, and shutdown-terminal status behavior
- admin HTTP handlers for reconfigure, rollout status, shutdown, shutdown status, operation errors, wait timeouts, and missing retained operation ids
- Rust admin SDK decoding for typed reconfigure/shutdown outcomes, operation rejection errors, pipeline details/status, and rollout/shutdown polling
- engine shutdown-drain behavior where downstream nodes wait for upstream channel closure before completing shutdown drain
- route/doc compatibility checks through the full workspace validation suite

Manual validation covered:

- creating a new pipeline through the admin API
- replacing a pipeline with a topology/config change
- resizing a pipeline by changing only core allocation
- submitting an identical config and observing a noop rollout
- shutting down a pipeline and observing shutdown tracking/status
- checking generation-aware status during overlapping rollout behavior

Yes.

- The admin API now supports live pipeline create/update/resize/noop via `PUT /api/v1/groups/{group}/pipelines/{id}`.
- Pipeline shutdown is available via `POST /api/v1/groups/{group}/pipelines/{id}/shutdown`.
- Rollout and shutdown progress can be queried through dedicated status endpoints.
- New pipeline-scoped admin routes are under `/api/v1/groups/...`.
- Pipeline status payloads now include deployment-generation-aware instance data and rollout metadata while preserving the existing aggregate status shape.
- Terminal rollout and shutdown ids are retained only within a bounded in-memory window.
- The Rust admin SDK now exposes typed clients, request options, operation outcomes, and error models for live reconfiguration and shutdown.
- Operator documentation was added in `rust/otap-dataflow/docs/admin/live-reconfiguration.md`.
- Support for the existing WebSocket log stream endpoint `/api/v1/telemetry/logs/stream` is preserved/restored in this branch.


rust/otap-dataflow/crates/admin-api/src/client.rs
```diff
@@ -158,37 +172,54 @@ pub struct AdminClient {
...
             backend: self.backend.as_ref(),
         }
     }
 
-    /// Returns the pipeline-group-scoped resource client.
+    /// Returns the group-scoped resource client for fleet-style status and shutdown operations.
     #[must_use]
-    pub fn pipeline_groups(&self) -> PipelineGroupsClient<'_> {
-        PipelineGroupsClient {
+    pub fn groups(&self) -> GroupsClient<'_> {
```
lquerel, line 208: This is a breaking change compared to the previous version, but in my opinion it's still early enough to make this kind of change without worrying about users of this SDK.

rust/otap-dataflow/crates/controller/src/live_control/runtime.rs
```diff
@@ -0,0 +1,405 @@
...
+
+        for (deployed_key, sender) in senders {
+            sender
+                .try_send_shutdown(
+                    Instant::now() + Duration::from_secs(timeout_secs.max(1)),
+                    "global shutdown".to_owned(),
+                )
+                .map_err(|err| ControlPlaneError::Internal {
+                    message: err.to_string(),
+                })?;
```
lalitb, line 361: this seems to fail fast after first `try_send_shutdown` error, and the senders that appears later in the iteration never receives the shutdown signal. Is this intended?

rust/otap-dataflow/src/main.rs
```diff
@@ -234,10 +234,6 @@ fn parse_core_id_range(s: &str) -> Result<CoreRange, String> {
 }
 
 fn main() -> Result<(), Box<dyn std::error::Error>> {
-    #[cfg(all(not(tarpaulin_include), feature = "dhat-heap"))]
```
lalitb, line 237: This seems to be removed?

rust/otap-dataflow/docs/admin/live-reconfiguration.md
```diff
@@ -0,0 +1,366 @@
...
+```bash
+BASE=http://127.0.0.1:8085
+GROUP=topic_multitenant_isolation
+PIPE=tenant_c_pipeline
+```
+
+Inspect the current committed config and observed runtime state:
+
+```bash
+curl -s "$BASE/groups/$GROUP/pipelines/$PIPE" | jq .
```
lalitb, line 238: nit - should `BASE` have `api/v1`, or else add it in curl path here?

rust/otap-dataflow/crates/controller/src/live_control/planning.rs
```diff
@@ -0,0 +1,930 @@
...
+        &self,
+        pipeline_group_id: &str,
+        pipeline_id: &str,
+        request: &ReconfigureRequest,
+    ) -> Result<CandidateRolloutPlan, ControlPlaneError> {
+        let pipeline_group_id: PipelineGroupId = pipeline_group_id.to_owned().into();
+        let pipeline_id: PipelineId = pipeline_id.to_owned().into();
+        let pipeline_key = PipelineKey::new(pipeline_group_id.clone(), pipeline_id.clone());
+
+        let (live_config, current_record) = {
```
lalitb, line 164: More of a design question: what consistency model do we want for concurrent live reconfiguration across different logical pipelines? This clones `live_config` under lock for validation, but commits from other pipelines can land before this rollout commits. Should live config preserve whole-config invariants across concurrent updates, or should operators treat live reconfiguration as serialized only per `(group, pipeline)`?

rust/otap-dataflow/crates/controller/src/live_control/execution.rs
```diff
@@ -0,0 +1,773 @@
...
+                "starting",
+                None,
+            );
+            let deployed_key = self
+                .launch_regular_pipeline_instance(
+                    &plan.resolved_pipeline,
+                    *core_id,
+                    plan.target_generation,
+                )
+                .map_err(|err| RolloutExecutionError::Failed(err.to_string()))?;
```
lalitb, line 260: In case of error, should we be cleaning up instances launched on earlier cores in this loop - just the way we do in the readiness loop below?

rust/otap-dataflow/crates/controller/src/live_control/execution.rs
```diff
@@ -0,0 +1,773 @@
...
+        panic: Box<dyn Any + Send>,
+    ) {
+        let report = PanicReport::capture("rollout worker", panic, Some(thread_name), None, None);
+        let failure_reason = report.summary_message();
+        self.update_rollout(pipeline_key, rollout_id, |rollout| {
+            rollout.state = RolloutLifecycleState::Failed;
+            rollout.failure_reason = Some(failure_reason.clone());
+        });
+        self.report_controller_worker_panic(pipeline_key, "rollout", rollout_id, &report);
+        self.finish_rollout(pipeline_key, rollout_id);
```
lalitb, line 57: `finish_rollout` seems to only prunes already-exited instances. how about the instances the worker already launched - they stay active without any logical pipeline pointing at them. Would a best-effort teardown of instances belonging to the panicking operation's generations, before marking the rollout terminal, be reasonable here?

rust/otap-dataflow/src/main.rs
lquerel: My bad. Fixed in https://github.com/open-telemetry/otel-arrow/pull/2618/commits/9d4018fa0cd1d5f3fe23c4804ebd00b70b770d8e

rust/otap-dataflow/crates/controller/src/live_control/runtime.rs
lquerel: Thanks, this was not intended. I updated `request_shutdown_all` so global shutdown is now best-effort across the full active-instance snapshot: one `try_send_shutdown` failure no longer prevents later instances from receiving shutdown.

I also added a regression test covering the original concern and the retry/idempotency behavior.

https://github.com/open-telemetry/otel-arrow/pull/2618/commits/49ea270408b94460fadeee70993400d6919552fa

rust/otap-dataflow/docs/admin/live-reconfiguration.md
lquerel: https://github.com/open-telemetry/otel-arrow/pull/2618/commits/68f510c69d40cba9cee2b3b75cbf0d5ace10fcf5

rust/otap-dataflow/crates/controller/src/live_control/planning.rs
lquerel: Great feedback! I clarified the current consistency model in the docs.

This PR intentionally keeps live reconfiguration serialized per logical pipeline, keyed by (group, pipeline), and does not claim whole-config serializability. The candidate is validated against a cloned full config snapshot, but commit only patches the accepted pipeline back into the latest live config.

I also documented the future evolution path: if/when we support mutable group-level invariants, we can widen the conflict scope to the pipeline group and continue using 409 Conflict without changing the existing per-pipeline endpoint or response schema. Engine-level reconfiguration would likely be a separate operation surface.

I created the following issue to track these future improvements: https://github.com/open-telemetry/otel-arrow/issues/2738

Doc changes: https://github.com/open-telemetry/otel-arrow/pull/2618/commits/6d9ec311561b17203d99f86cd2f5c8e0378c082c

rust/otap-dataflow/crates/controller/src/live_control/execution.rs
lquerel: Good catch! The create path already cleaned up candidates after readiness failure, but not if a later core failed during the initial launch loop. I updated the launch-error path to mark the failed core and tear down any candidate instances launched on earlier cores before returning the rollout failure. Since this is a create rollout, there is no previous generation to roll back to, so the correct cleanup is best effort shutdown of the partially launched candidate set.

See https://github.com/open-telemetry/otel-arrow/pull/2618/commits/b2f59071e90d7af07901b7dda8516da169759976

rust/otap-dataflow/crates/controller/src/live_control/execution.rs
lquerel: Another good point! I updated the rollout worker panic path to best effort shut down active candidate instances from the rollout's uncommitted target generation before marking the rollout terminal and clearing the conflict. The cleanup intentionally skips the target generation if it has already been committed as the active serving generation, so a late bookkeeping panic does not take down healthy serving instances. I also added regression tests for both the uncommitted candidate cleanup case and the committed generation protection case, and documented these controller safety behaviors in `live-reconfiguration.md`.

See https://github.com/open-telemetry/otel-arrow/pull/2618/commits/a76e1508bc06d9c78913e5b645b5a98724908283

lquerel:
@lalitb Thanks, for all this very valuable feedback.

rust/otap-dataflow/src/main.rs
```diff
@@ -280,11 +281,6 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
 
     let controller = Controller::new(&OTAP_PIPELINE_FACTORY);
     let result = controller.run_forever(engine_cfg);
-    #[cfg(all(not(tarpaulin_include), feature = "dhat-heap"))]
-    {
-        dhat_finish();
-    }
-
```
jmacd, line 287: btw, what's this about?

rust/otap-dataflow/crates/controller/src/live_control/execution.rs
```diff
@@ -0,0 +1,853 @@
...
+                None,
+            );
+
+            let new_key = self
+                .launch_regular_pipeline_instance(
+                    &plan.resolved_pipeline,
+                    *core_id,
+                    active_generation,
+                )
+                .map_err(|err| RolloutExecutionError::Failed(err.to_string()))?;
```
lalitb: Shouldn't this go through `rollback_resize_rollout` instead of returning Failed directly? If an earlier core already started and became ready, it'll be left running when this fails.

Same pattern is there in `run_replace_rollout` too.

lalitb (approved):
LGTM. I left one lifecycle cleanup question around partial resize/replace launch failures, but otherwise the engine/control-plane changes look thoughtfully designed and well covered. Looking forward to using live pipeline reconfiguration :)

rust/otap-dataflow/crates/controller/src/live_control/execution.rs
lquerel: Good catch. I updated both `run_resize_rollout` and `run_replace_rollout` so a launch failure after earlier rollout progress now goes through the existing rollback helpers instead of returning Failed directly. That means already started cores are cleaned up rather than being left running. I also added regression tests covering resize rollback cleanup and replace rollback cleanup for previously activated added cores.

https://github.com/open-telemetry/otel-arrow/pull/2618/commits/791ba2b2f973767df458d728e825215686df3f6c

rust/otap-dataflow/crates/controller/src/live_control/planning.rs
```diff
@@ -0,0 +1,933 @@
...
+            }
+        })?;
+        if let serde_json::Value::Object(map) = &mut current_pipeline {
+            let _ = map.remove("policies");
+        }
+        if let serde_json::Value::Object(map) = &mut candidate_pipeline {
+            let _ = map.remove("policies");
+        }
+
+        Ok(current_pipeline == candidate_pipeline
```
gouslu: nit: The correctness of this comparison depends on `PipelineConfig`'s internal shape (which fields matter for topology vs. which are resource policy), but that knowledge lives here in the planner rather than on the type itself. Could this be encapsulated as a method on `PipelineConfig` such as `eq_topology(&other)` / `eq_ignoring_resources(&other)`, built on `PartialEq` (which `PipelineConfig` already derives)? That keeps the "what counts as the same topology" definition next to the field definitions, so anyone adding a field sees the equivalence method right there and is much more likely to make a deliberate choice, and this can also help drop the `serde_json::Value` round-trip and `Result` arm because a pure comparison with `PartialEq` can't fail like `serde_json::to_value`.

rust/otap-dataflow/crates/controller/src/live_control/planning.rs
```diff
@@ -0,0 +1,933 @@
...
+            && current.policies.channel_capacity == candidate.policies.channel_capacity
+            && current.policies.health == candidate.policies.health
+            && current.policies.telemetry == candidate.policies.telemetry
+            && current.policies.transport_headers == candidate.policies.transport_headers)
+    }
+
+    /// Compares two resolved pipelines for exact runtime equivalence.
+    pub(super) fn resolved_pipeline_matches(
```
gouslu: nit: see my nit above :D

rust/otap-dataflow/crates/controller/src/live_control/runtime.rs
```diff
@@ -0,0 +1,476 @@
...
+                Some(RuntimeInstanceExit::Error(error)) => Err(error.message),
+                None => Err(err.to_string()),
+            };
+        }
+        self.release_instance_control_sender(deployed_key);
+        Ok(())
+    }
+
+    /// Waits until a specific deployed instance exits or the deadline expires.
+    pub(super) fn wait_for_instance_exit(
```
gouslu, line 297: nit: Seems like `Condvar` could be used here to eliminate polling and potential (but very minor) delays on serial instance-exit waits.

jmacd (approved):

rust/otap-dataflow/src/main.rs
lquerel: I think you looked at an older version of this PR, because that code had been accidentally removed during a merge with main. I restored it yesterday.

rust/otap-dataflow/crates/controller/src/live_control/planning.rs
lquerel: Good point. Moved this comparison out of the planner and onto the config types. `ResolvedPipelineConfig` now owns both `runtime_matches(...)` and `runtime_shape_matches_ignoring_resources(...)`, built on `PipelineConfig::eq_ignoring_policies(...)` and `ResolvedPolicies::eq_ignoring_resources(...)`. That removes the `serde_json` round-trip and keeps the "what counts as topology vs resource-only" decisions next to the owning structs so future field additions are harder to miss.

https://github.com/open-telemetry/otel-arrow/pull/2618/commits/0e9cd1261f72cc98f446379ef51ab579e9fd463e

rust/otap-dataflow/crates/controller/src/live_control/planning.rs
lquerel: See https://github.com/open-telemetry/otel-arrow/pull/2618/commits/0e9cd1261f72cc98f446379ef51ab579e9fd463e

rust/otap-dataflow/crates/controller/src/live_control/runtime.rs
lquerel: https://github.com/open-telemetry/otel-arrow/pull/2618/commits/1fe6227210702c69b7492847c488df2a82fa88f8

lquerel: Hum in fact -> https://github.com/open-telemetry/otel-arrow/pull/2618/commits/bfcfc0f314ea25d24e2f4819593894b64606fcb8

PR #2620, opened 2026-04-09
feat(otap-dataflow): add live internal log streaming to the admin UI

lalitb:
**Description**

This PR adds live internal log viewing to the embedded admin UI.

It introduces:

- a WebSocket log stream endpoint for retained internal logs
- server-side pause/resume, backfill, and filtering for the stream
- an “Internal Logs” panel in the admin UI

**Implementation notes:**

- `log_tap` now supports live subscribers in addition to retained-log queries
- the stream reuses the admin log schema used by /api/v1/telemetry/logs
- the log panel remains disconnected by default and only opens the stream when the user explicitly clicks Connect

**User-facing behavior:**

- view internal logs live in the admin UI
- pause/resume server-side delivery
- resync from retained logs after drops
- filter by minimum level and text

When admin UI opened, the log panel is disabled. 

Screenshot 2026-04-12 at 10 59 41 AM

When enabled:

Screenshot 2026-04-12 at 11 01 03 AM


lquerel:
Could you add a screenshot showing the controls to display / hide the log tap panel.

lalitb:
> Could you add a screenshot showing the controls to display / hide the log tap panel.

I added that now. This PR originally did not include a separate show/hide control for the Internal Logs panel, but it does now, and I’ve also added screenshots showing the display/hide control and the panel state.

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1269,6 +1238,468 @@ fn escape_prom_help(s: &str) -> String {
...
+
+    /// Build a `LogFilter` from optional client-supplied strings.
+    fn from_params(
+        search_query: Option<String>,
+        minimum_timestamp: Option<String>,
+        minimum_level: Option<String>,
+    ) -> Self {
+        let minimum_timestamp = minimum_timestamp.and_then(|s| {
+            chrono::DateTime::parse_from_rfc3339(&s)
+                .ok()
```
jmacd, line 1360: This looks like it might suppress an error? Wondering if we should return `Result<Self>`

jmacd (approved):
I must express that I'm not very qualified as a reviewer for the Javascript aspects here. I think it's OK given the relatively internal and developer-facing nature of this code. 

One thing I'd like to make sure of (not blocking) is that users who do not want the admin component have a way to build without it.

PR #2622, opened 2026-04-09
feat(engine): add ExtensionFactory, ExtensionWrapper, and extension lifecycle types

gouslu:
# Change Summary

Add the core engine types for the extension lifecycle system ([Phase 1, PR 2](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/extension-system-architecture.md#pr-2----engine-extensionfactory--extensionwrapper--builder)).

Extensions are PData-free components that provide cross-cutting services
(e.g., authentication, storage) to data-path nodes. This PR introduces
the foundational types for defining, building, and starting extensions —
fully separated from the node/DAG infrastructure.

**New types (config crate):**
- `ExtensionId` — separate ID namespace from `NodeId`
- `ExtensionUrn` — 3-segment URN format (`urn:<namespace>:<id>`), simpler than node's 4-segment `NodeUrn` (`urn:<namespace>:<kind>:<id>`) since kind is implicit from the `extensions:` config section
- `ExtensionUserConfig` — extension-specific config (`type`, `description`, `config` only — no output ports, wiring, or header policies). Uses `#[serde(deny_unknown_fields)]` to reject node-specific fields at parse time
- `PipelineExtensions` — collection type for pipeline extensions, parallel to `PipelineNodes`
- `DuplicateExtension` — separate error variant for duplicate extension IDs (distinct from `DuplicateNode`)

**New types (engine crate):**
- `ExtensionFactory` — PData-free factory struct with `create` (returns `ExtensionBundle`) and `validate_config`
- `ExtensionWrapper` — enum with `Local` / `Shared` variants (first layer). Each variant holds an `ExtensionLifecycle<E, R>` with `Active` / `Passive` variants (second layer). This two-level structure makes impossible states unrepresentable
- `ExtensionLifecycle<E, R>` — generic lifecycle enum; `Active` bundles the extension trait object with its control channel sender and receiver, `Passive` has no channels
- `ExtensionBundle` — opaque return type from the builder, holding at most one local and one shared `ExtensionWrapper`. Private fields enforce construction only via the builder
- `ExtensionBundleBuilder` — builder with `with_local()` / `with_shared()` + `build()`. Enforces at least one variant present and rejects same-type dual registration via `TypeId` guard
- `ControlChannel<R>` — generic shutdown-aware control channel. Concrete types `local::extension::ControlChannel` (with `LocalReceiver`) and `shared::extension::ControlChannel` (with `SharedReceiver`) are defined in their respective modules, matching how pipeline nodes handle control channels
- `ControlReceiver` — internal trait enabling the generic `ControlChannel<R>` to work with both local and shared receivers without code duplication
- `Active<E>` / `Passive<E>` — newtype wrappers that signal lifecycle intent at the type level
- `EffectHandler` — extension identity and metrics reporter (forward-compatible extension point for the `start()` signature)
- `ExtensionControlMsg` — PData-free control message enum (Config, CollectTelemetry, Shutdown)
- `ExtensionControlSender` — control sender for extensions, using `ExtensionId`
- `local::Extension` / `shared::Extension` — lifecycle traits (`Rc<Self>` vs `Box<Self>`), each taking their own `ControlChannel` type

**Modified existing types:**
- `ExtensionAlreadyExists` error variant added (uses `ExtensionId`, not `NodeId`)
- `PipelineFactory::new()` now accepts extension factories as a 4th parameter
- `pipeline_factory` proc macro generates `EXTENSION_FACTORIES` distributed slice
- `wrap_control_channel_metrics` generalized to accept `&str` name instead of `&NodeId`, enabling reuse by both nodes and extensions
- `control_channel_id()` similarly generalized from `&NodeId` to `&str`
- `PipelineConfig` extensions field changed from `PipelineNodes` to `PipelineExtensions`
- Extension URN validation moved from runtime `NodeKind` check to parse-time `ExtensionUrn` format enforcement
- Removed redundant `canonicalize_plugin_urns` for extensions — `ExtensionUrn` normalizes at parse time via `#[serde(try_from)]`
- `PipelineConfigBuilder` tracks `duplicate_extensions` separately from `duplicate_nodes`

**Design decisions:**
- Extensions are fully separated from nodes — zero `NodeType`, `NodeId`, `NodeUrn`, `NodeKind`, or `NodeUserConfig` references in extension code
- `ExtensionUrn` uses a simpler 3-segment format (`urn:<ns>:<id>`) since the `<kind>` segment is redundant for extensions. This also prevents cross-kind misconfiguration: a 4-segment node URN placed in `extensions:` is rejected at parse time
- Two-level enum design: `ExtensionWrapper` (Local/Shared) × `ExtensionLifecycle` (Active/Passive) — eliminates impossible states without `Option`s
- `ExtensionBundle` replaces `Vec<ExtensionWrapper>` — exactly 0-1 local + 0-1 shared, private fields, builder-enforced invariants
- Same-type guard: runtime `TypeId` check prevents registering the same concrete type as both local and shared (avoids duplicate lifecycles on shared `Arc` state)
- Local extensions use local channels (`LocalSender`/`LocalReceiver` via `LocalMode`), shared extensions use shared channels (`SharedSender`/`SharedReceiver` via `SharedMode`) — matching how pipeline nodes handle their control channels
- Shutdown grace-period: `ControlChannel::recv()` continues delivering Config/CollectTelemetry messages during the grace period, using a 2-branch `select!` racing the deadline against incoming messages
- `build()` and `start()` return `Result` with `Error::InternalError` instead of panicking
- `Clone` bound on `Active<E>` and `Passive<E>` shared providers is retained for the capability system (next PR) which needs `Box<dyn CloneAnySend>` for type-erased registration

* Part of the extension system Phase 1 rollout (PR 2 of 8)

- 27 unit tests in `extension.rs` (engine) covering: all lifecycle variants (local/shared × active/passive), builder validation (empty, same-type guard), dual-type creation, start/shutdown for both local and shared, control channel behavior (immediate/delayed shutdown, config ordering, grace-period message delivery, channel close), effect handler, control sender, telemetry metrics for both active and passive paths, `ExtensionBundle` iteration
- 10 unit tests in `extension_urn.rs` (config) covering: full URN parsing, short form expansion, case insensitivity, rejection of 4-segment node URN format, empty/invalid segments, serde roundtrip, `From<&str>`
- 2 unit tests in `extension.rs` (config) covering: YAML deserialization, rejection of unknown fields
- 4 unit tests in `lib.rs` (engine) covering: `ExtensionFactory` name/clone/validate_config, `ExtensionAlreadyExists` error variant
- All existing engine tests and config tests pass
- `cargo clippy` with `-D warnings` passes for both `otap-df-config` and `otap-df-engine`
- `cargo fmt --all -- --check` passes

Extension URN format changed from `urn:<namespace>:extension:<id>` to `urn:<namespace>:<id>`. This is a simplification — the `:extension:` segment was redundant since extensions are always declared in the `extensions:` config section.


rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1198 @@
...
+
+// Active<E> shared: requires Extension + Clone + Send
+impl<E: shared_ext::Extension + Clone + Send + 'static> sealed_provider::SealedShared
+    for Active<E>
+{
+}
+
+impl<E: shared_ext::Extension + Clone + Send + 'static> SharedProvider for Active<E> {
+    fn decompose(self) -> SharedDecomposed {
+        let ext: Box<dyn shared_ext::Extension> = Box::new(self.0);
```
lalitb: Do we need `Clone` on `E` ?  As `decompose` appears to consume the value into a `Box`. If cloning isn't actually needed elsewhere, removing that bound can make API more rust-idiomatic.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1198 @@
...
+}
+
+// Passive<E> shared: no Extension bound needed
+impl<E: Clone + Send + 'static> sealed_provider::SealedShared for Passive<E> {}
+
+impl<E: Clone + Send + 'static> SharedProvider for Passive<E> {
+    fn decompose(self) -> SharedDecomposed {
+        SharedDecomposed {
+            extension: None,
+            type_id: TypeId::of::<E>(),
```
lalitb, line 239: `Passive<E>` looks like it's registering the instance, but `decompose()` drops it and only keeps the `TypeId` (which is itself discarded after the same-type guard in build()).
Could you add a doc note on Passive clarifying that the value isn't retained? That way anyone building on this before the capability PR lands won't be surprised.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1198 @@
...
+            .control_receiver
+            .expect("start() called on passive extension — this is a bug");
+        let ctrl_chan = ControlChannel::new(control_receiver);
+
+        match (self.local_extension, self.shared_extension) {
+            (Some(_), Some(_)) => {
+                // The engine starts each active lifecycle independently — it
+                // should never call start() on a wrapper that still holds both.
+                // The runtime pipeline will take each variant separately.
+                unreachable!(
```
lalitb: `build()` accepts the dual-active shape and allocates two control channels for it, but `start()` treats it as `unreachable!()`. Even for an incremental PR, it'd be good to keep  these consistent - either reject dual-active in `build()` until subsequent PR adds support, or return a proper error from `start()` instead of panicking.

lalitb (approved):
LGTM. Would good to have one more eye, in case I missed something obvious :)

gouslu:
> LGTM. Would good to have one more eye, in case I missed something obvious :)

thanks! added comments to the code to address your feedback. all of these changes are intended, including dual-active. when there is an explicit local implementation, they can't share the same lifecycle, so the idea is that engine will wire things up, but if there are no components need either local or shared variant, they will be dropped. only the ones that are used will be started.

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: clone is needed for shared variants as it is how we will be storing type erased capability registrations. local one uses Rc<>.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1209 @@
...
+
+    /// Print an info message to stdout.
+    pub async fn info(&self, message: &str) {
+        use tokio::io::{AsyncWriteExt, stdout};
+        let mut out = stdout();
+        let _ = out.write_all(message.as_bytes()).await;
+        let _ = out.write_all(b"\n").await;
+        let _ = out.flush().await;
+    }
+}
```
jmacd: I would prefer to see an otel_info! log in places where this might be used.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1209 @@
...
+
+    #[tokio::test]
+    async fn test_effect_handler_info() {
+        let node_id = test_node("eh_info");
+        let metrics_reporter = test_metrics_reporter();
+        let handler = EffectHandler::new(node_id, metrics_reporter);
+        // Just verify it doesn't panic; output goes to stdout.
+        handler.info("test message").await;
+    }
+
```
jmacd, line 1038:
```suggestion
```

jmacd (approved):

rust/otap-dataflow/crates/engine/src/node.rs
```diff
@@ -59,6 +59,8 @@ pub enum NodeType {
     Processor,
     /// Represents a node that exports data to an external destination.
     Exporter,
+    /// Represents a node that provides shared capabilities to other components.
+    Extension,
```
lquerel: Why mix the concepts of extensions and PData nodes? In our configuration model, extensions and nodes are clearly two sibling concepts, so grouping both under `NodeType` is not ideal.

rust/otap-dataflow/crates/engine/src/local/extension.rs
```diff
@@ -0,0 +1,45 @@
...
+/// # Thread Safety
+///
+/// The local `Extension` trait does NOT require the `Send` bound, allowing
+/// use of `Rc`, `RefCell`, and other !Send types within a single-threaded
+/// `LocalSet`.
+///
+/// # Ownership
+///
+/// `start` takes `Rc<Self>` so the same instance can serve both the background
+/// task and capability consumers without cloning internal state.
```
lquerel, line 30: Nice.

rust/otap-dataflow/crates/engine/src/local/extension.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// Extensions are started BEFORE receivers, processors, and exporters.
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`Error`] if an unrecoverable error occurs.
+    async fn start(
+        self: Rc<Self>,
+        ctrl_chan: ControlChannel,
+        effect_handler: EffectHandler,
+    ) -> Result<TerminalState, Error>;
```
lquerel, line 44: It’s cool to find the same approach again.

rust/otap-dataflow/crates/engine/src/control.rs
```diff
@@ -705,6 +705,69 @@ where
...
+}
+
+/// A control sender for a single extension.
+///
+/// Stored separately from [`ControlSenders`] because extensions use
+/// [`ExtensionControlMsg`] (PData-free) rather than [`NodeControlMsg<PData>`].
+pub struct ExtensionControlSender {
+    /// Unique identifier of the extension.
+    #[allow(dead_code)] // Used by runtime pipeline in a follow-up PR.
+    pub(crate) node_id: NodeId,
```
lquerel: Identifying an extension with a `NodeId` seems confusing to me. An extension is not a DAG/dataflow node, nor even an entity declared in the `nodes` section of our configuration, so it's definitely not identified by a `NodeId`. I think we should introduce an `ExtensionId` here.

We shouldn't be able to build a map keyed by NodeId that mixes nodes and extensions because (1) the values are not of the same kind, and (2) in theory, the ID scopes for extensions and nodes are disjoint, so nothing should prevent an extension from having the same name as a DAG node.

rust/otap-dataflow/crates/engine/src/error.rs
```diff
@@ -337,6 +337,13 @@ pub enum Error {
         node: NodeName,
     },
 
+    /// The specified extension already exists in the pipeline.
+    #[error("The extension `{extension}` already exists")]
+    ExtensionAlreadyExists {
+        /// The name of the extension that already exists.
+        extension: NodeId,
```
lquerel: `NodeId` -> `ExtensionId`

rust/otap-dataflow/crates/engine/src/exporter.rs
```diff
@@ -223,7 +223,7 @@ impl<PData> ExporterWrapper<PData> {
                 ..
             } => {
                 let (control_sender, control_receiver) =
-                    wrap_control_channel_metrics::<LocalMode, PData>(
+                    wrap_control_channel_metrics::<LocalMode, NodeControlMsg<PData>>(
```
lquerel, line 226: I'll wait to see the rest, but I'm not sure I understand why introducing extensions requires changing this declaration. I wonder whether all of this isn't simply coming from mixing up nodes and extensions.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1209 @@
...
+// ── EffectHandler ───────────────────────────────────────────────────────────
+
+/// The effect handler for extensions.
+///
+/// Provides extensions with node identity and basic I/O. Extensions manage
+/// their own timers directly via `tokio::time` rather than through the engine's
+/// timer infrastructure, keeping the extension system fully PData-free.
+#[derive(Clone)]
+pub struct EffectHandler {
+    node_id: NodeId,
```
lquerel: See my previous comment.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1209 @@
...
+pub struct Active<E>(pub E);
+
+/// Wraps an extension type to signal it is passive (no lifecycle).
+///
+/// No task is spawned, no control channel is created.
+///
+/// TODO: In this PR the wrapped value is consumed by `decompose()` and
+/// only its `TypeId` is retained for the same-type guard. The value itself
+/// will be stored in a later PR when the capability system is added.
+pub struct Passive<E>(pub E);
```
lquerel, line 185: I wonder whether, fundamentally, we don't have three kinds of extensions that could be summarized by the following enum:

```rust
enum ExtensionLifecycle {
    Passive,
    Local(Rc<dyn local::Extension>),              // assigned to a task
    Shared(Box<dyn shared::Extension>),     // assigned to a task
}
```

That would also avoid having a `start` method for extensions that don't have an associated task.

rust/otap-dataflow/crates/engine/src/extension.rs
lquerel: I think that if we follow a slightly different approach, like the one mentioned [here](https://github.com/open-telemetry/otel-arrow/pull/2622/changes#r3077385456), we can avoid this kind of issue.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1209 @@
...
+// ── ExtensionWrapper ────────────────────────────────────────────────────────
+
+/// Wrapper for extension instances in the pipeline engine.
+///
+/// Extensions are NOT generic over PData — they operate exclusively on
+/// [`ExtensionControlMsg`], keeping the extension system entirely decoupled
+/// from the data-plane type.
+pub struct ExtensionWrapper {
+    node_id: NodeId,
+    user_config: Arc<NodeUserConfig>,
```
lquerel: Another example of reuse that really doesn't seem like a good idea to me at all. An extension does not share the same configuration as a node in a DAG. An extension has no outputs, does not share the same policies, etc. This seems to me like the definitive proof that extensions and nodes need to be separated.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1209 @@
...
+}
+
+// ── ExtensionWrapper ────────────────────────────────────────────────────────
+
+/// Wrapper for extension instances in the pipeline engine.
+///
+/// Extensions are NOT generic over PData — they operate exclusively on
+/// [`ExtensionControlMsg`], keeping the extension system entirely decoupled
+/// from the data-plane type.
+pub struct ExtensionWrapper {
```
lquerel: Here we have a struct instead of an enum. What is the reason for that? As a result, we end up with a whole bunch of `Option`s to represent both the local and shared forms of extensions. The final structure takes up more memory than its enum equivalent, so I’m not sure I understand the rationale.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1209 @@
...
+    ///
+    /// - Panics if neither `with_local` nor `with_shared` was called.
+    /// - Panics if both were called with the same concrete type.
+    pub fn build(self) -> ExtensionWrapper {
+        let has_local = self.local_extension.is_some() || self.local_type_id.is_some();
+        let has_shared = self.shared_extension.is_some() || self.shared_type_id.is_some();
+        assert!(
+            has_local || has_shared,
+            "ExtensionWrapper must have at least one variant (local or shared)"
+        );
```
lquerel: With the enum-based approach, we wouldn’t end up with this kind of representation that’s impossible to handle.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1209 @@
...
+    shared_control_receiver: Option<SharedReceiver<ExtensionControlMsg>>,
+    telemetry: Option<NodeTelemetryGuard>,
+}
+
+// ── Builder ──────────────────────────────────────────────────────────────────
+
+/// Builder for `ExtensionWrapper`.
+///
+/// At least one variant (local or shared) must be added before calling `build()`.
+pub struct ExtensionWrapperBuilder {
```
lquerel: Same remark, why not using a enum here ?

lquerel (changes requested):
I've reviewed half of the PR. I think there are some interesting things, but before going any further, I'd like us to address what seems to me to be an important issue: an extension is not a DAG node. An extension should not have the same configuration as a node (as is the case here), should not use the same kind of ID, and should not be part of the `NodeType` variants. In my view, this creates all sorts of problems, which end up showing up as a series of tests and panic handling, even though we have a type system precisely to help us avoid this kind of impossible representation.

gouslu:
> I've reviewed half of the PR. I think there are some interesting things, but before going any further, I'd like us to address what seems to me to be an important issue: an extension is not a DAG node. An extension should not have the same configuration as a node (as is the case here), should not use the same kind of ID, and should not be part of the `NodeType` variants. In my view, this creates all sorts of problems, which end up showing up as a series of tests and panic handling, even though we have a type system precisely to help us avoid this kind of impossible representation.

Thanks for the comments. A lot of them seem to be related to me re-using node related code, and I thought of completely separating extensions from nodes, but I didn't want to get ahead of myself with doing so. I was partly expecting a feedback as such. Will make the necessary changes and get back.

lalitb:
I approved too early and missed an important point already reflected in the design doc. On a second pass, I agree with @lquerel that extensions and nodes are separate sibling concepts, not variants of the same engine-level node abstraction.

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: The struct approach was chosen to support dual-registration (both local and shared variants in one wrapper). However it creates the issues you pointed out. I'll work on addressing this using an enum based approach.

rust/otap-dataflow/crates/engine/src/exporter.rs
gouslu: Extensions control channels are PData-free, so this change generalizes `wrap_control_channel_metrics` to accept any message type and any name (id), since now ExtensionId and NodeId are separated as well.

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: Builder allows for multiple variants. So you can have local or shared or both. Enum here doesn't match that design. Updated the comment on the struct to clarify.

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: deleted info entirely from extensions implementation, we can use otel_info instead.

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: deleted the info method.

gouslu:
> I've reviewed half of the PR. I think there are some interesting things, but before going any further, I'd like us to address what seems to me to be an important issue: an extension is not a DAG node. An extension should not have the same configuration as a node (as is the case here), should not use the same kind of ID, and should not be part of the `NodeType` variants. In my view, this creates all sorts of problems, which end up showing up as a series of tests and panic handling, even though we have a type system precisely to help us avoid this kind of impossible representation.

Addressed the comments, separated nodes from extensions entirely, added ExtensionLifecycle and got rid of the impossible states being representable problem.

rust/otap-dataflow/crates/engine/src/extension.rs
jmacd: Cool.

jmacd (approved):

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,969 @@
...
+use otap_df_telemetry::otel_debug;
+use otap_df_telemetry::reporter::MetricsReporter;
+use std::any::TypeId;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::time::Instant;
+use tokio::time::{Sleep, sleep_until};
+
+/// Name type for extensions — separate from node names.
+pub type ExtensionName = otap_df_config::ExtensionId;
```
lquerel: nit: Why having both ExtensionId and ExtensionName?

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,969 @@
...
+
+// ── ControlChannel ──────────────────────────────────────────────────────────
+
+/// A channel for receiving control messages for extensions.
+///
+/// When a `Shutdown` message arrives with a future deadline, the channel waits
+/// until the deadline expires, then returns the `Shutdown`.
+pub struct ControlChannel {
+    control_rx: Option<SharedReceiver<ExtensionControlMsg>>,
```
lquerel, line 50: nit: Just for maintainability, I think we should explain why we have a shared receiver here. I believe the reason is that an extension can be used by both local and shared nodes.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,969 @@
...
+                tokio::select! {
+                    biased;
+                    _ = sleep_until_deadline.as_mut().expect("set above") => {
+                        let shutdown = self.pending_shutdown
+                            .take()
+                            .expect("pending_shutdown must exist");
+                        self.shutdown();
+                        return Ok(shutdown);
+                    }
+                }
```
lquerel, line 112: There is only one branch in this `select!`, so I don’t really see the point of using it.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,969 @@
...
+                        }
+                        self.shutting_down_deadline = Some(deadline);
+                        self.pending_shutdown = Some(ExtensionControlMsg::Shutdown { deadline, reason });
+                        continue;
+                    }
+                    Ok(msg) => return Ok(msg),
+                    Err(e)  => return Err(e),
+                },
+            }
+        }
```
lquerel: Same here. There is only one branch in this `select!`, so I don’t really see the point of using it.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,969 @@
...
+    }
+
+    fn shutdown(&mut self) {
+        self.shutting_down_deadline = None;
+        drop(self.control_rx.take().expect("control_rx must exist"));
```
lquerel: nit: drop is useless here because in this context `drop(self.control_rx.take().expect("control_rx must exist"));` is effectively the same as `self.control_rx.take().expect("control_rx must exist");` :-)

rust/otap-dataflow/crates/engine/src/extension.rs
lquerel: @gouslu Why Active<E> and Passive<E> still there?

lquerel: Ok, after reading the rest of the code, I get the impression that these markers are here so that in the builder we can have something like `with_local(Active(...))` instead of `with_local_active(...)`. 
Not against the approach but not sure that makes a big difference in terms of usability.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,969 @@
...
+                    control_receiver: r,
+                }
+            }
+        };
+        self
+    }
+
+    /// Returns `ExtensionControlSender`(s) for sending control messages.
+    #[allow(dead_code)]
+    pub(crate) fn extension_control_senders(&self) -> Vec<crate::control::ExtensionControlSender> {
```
lquerel: Why a `Vec<crate::control::ExtensionControlSender>` instead of an `Option<crate::control::ExtensionControlSender>` ?

rust/otap-dataflow/crates/engine/src/extension.rs
lquerel: We still have a struct here, not an enum. Why was this comment marked as resolved?

lquerel (approved):
This new version is definitely much cleaner. I noted a few minor issues here and there, but I'm approving on the assumption that these issues will be resolved, or discussed if needed, before merging into main.

rust/otap-dataflow/crates/engine/src/lib.rs
```diff
@@ -213,6 +215,46 @@ impl<PData> NamedFactory for ExporterFactory<PData> {
...
+    pub description: &'static str,
+    /// URL to the extension's documentation.
+    pub documentation_url: &'static str,
+    /// A function that creates a new extension instance.
+    pub create: fn(
+        pipeline: PipelineContext,
+        name: otap_df_config::ExtensionId,
+        ext_config: Arc<otap_df_config::extension::ExtensionUserConfig>,
+        extension_config: &ExtensionConfig,
+    ) -> Result<ExtensionWrapper, otap_df_config::error::Error>,
```
lalitb: I’m confused by this API shape. `ExtensionWrapperBuilder::build()` can return two wrappers for a dual local+shared extension, but `ExtensionFactory::create` can only return a single ExtensionWrapper.  If dual-variant extensions are meant to be supported in this PR, shouldn’t the factory boundary return a collection as well? Otherwise the builder can represent something the factory API can’t hand back to the engine.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,969 @@
...
+                        .expect("pending_shutdown must exist");
+                    self.shutdown();
+                    return Ok(shutdown);
+                }
+
+                if sleep_until_deadline.is_none() {
+                    sleep_until_deadline = Some(Box::pin(sleep_until(dl.into())));
+                }
+
+                tokio::select! {
```
lalitb, line 96: I think this branch is the problem: once shutdown is latched, `recv()` stops polling `control_rx` and only waits for the deadline. So any Config / CollectTelemetry messages sent during the grace period just sit in the channel and get dropped when `shutdown()` drops the receiver. Was the intent to stop serving control messages after shutdown is queued? I was expecting the channel to keep handling them until the deadline fires.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,969 @@
...
+
+    /// Build the `ExtensionWrapper`(s).
+    ///
+    /// Returns one wrapper for single-variant extensions, or two wrappers
+    /// for dual-type extensions (one local, one shared).
+    ///
+    /// # Panics
+    ///
+    /// Panics if neither `with_local` nor `with_shared` was called.
+    pub fn build(self) -> Vec<ExtensionWrapper> {
```
lalitb: I may be missing it, but I don’t see the same-type guard anymore. SharedDecomposed / LocalDecomposed still carry a TypeId, but `build()` never compares them before creating both wrappers.  As written, this seems to allow registering the same implementation as both local and shared, which the PR description says should be rejected.

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -437,6 +438,100 @@ impl IntoIterator for PipelineNodes {
...
+    /// Canonicalize extension type URNs.
+    fn canonicalize_plugin_urns(
+        &mut self,
+        pipeline_group_id: &PipelineGroupId,
+        pipeline_id: &PipelineId,
+    ) -> Result<(), Error> {
+        use crate::extension::ExtensionUrn;
+        for (ext_id, ext) in self.0.iter_mut() {
+            let mut updated = (**ext).clone();
+            let normalized = ExtensionUrn::parse(updated.r#type.as_ref()).map_err(|e| {
```
lalitb: This looks like a re-parse of an already-parsed `ExtensionUrn`, so I’m not sure it can actually canonicalize anything new here. Also, if this ever does fail, mapping it to `InvalidNodeType` seems misleading now that extensions have their own type/id space.

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -1066,29 +1137,23 @@ impl PipelineConfigBuilder {
...
         id: S,
-        node_type: U,
+        ext_type: U,
         config: Option<Value>,
     ) -> Self {
         let id = id.into();
-        let node_type = node_type.into();
+        let ext_type = ext_type.into();
         if self.extensions.contains_key(&id) {
             self.duplicate_nodes.push(id.clone());
```
lalitb: This still records duplicate extension IDs in `duplicate_nodes`, so they’ll be reported later as DuplicateNode. Since extensions are now a separate concept, should this have its own duplicate-extension path instead

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,969 @@
...
+/// Wraps an extension type to signal it is passive (no lifecycle).
+///
+/// In this PR the wrapped value is consumed by `decompose()` and only its
+/// `TypeId` is retained. The value itself will be stored when the capability
+/// system is added.
+pub struct Passive<E>(pub E);
+
+/// Decomposed result of a shared extension provider.
+#[doc(hidden)]
+pub struct SharedDecomposed {
```
lalitb, line 189: If these types aren’t meant to be public API yet, I’d make them `pub(crate)` for now instead of `pub + #[doc(hidden)]`. Right now they’re technically exposed to downstream crates, just undocumented.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,969 @@
...
+    }
+
+    /// Starts the extension lifecycle.
+    pub async fn start(self, metrics_reporter: MetricsReporter) -> Result<TerminalState, Error> {
+        let ext_name = self.name.clone();
+        let effect_handler = EffectHandler::new(self.name, metrics_reporter);
+
+        match self.lifecycle {
+            ExtensionLifecycle::Passive => {
+                panic!("start() called on passive extension — this is a bug")
```
lalitb: Small API concern: start() returns Result<TerminalState, Error>, but this branch still panics on Passive. f this is an invariant we expect callers to violate sometimes during integration, it may be better to return an Error here instead of crashing.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,969 @@
...
+        let name = self.name.as_ref();
+
+        self.lifecycle = match self.lifecycle {
+            ExtensionLifecycle::Passive => ExtensionLifecycle::Passive,
+            ExtensionLifecycle::LocalActive {
+                extension,
+                control_sender,
+                control_receiver,
+            } => {
+                let (s, r) = wrap_control_channel_metrics::<SharedMode, ExtensionControlMsg>(
```
lalitb: Small maintainability nit: could we add a comment on why `LocalActive` still uses `SharedMode` here? I think it’s because extension control stays on the shared engine control path, but that wasn’t immediately obvious.

lalitb:
Good to fix atleast these one's I commented - the factory/builder mismatch for dual-variant extensions, the shutdown grace-period behavior in ControlChannel::recv(), and the missing same-type guard in build(). The rest of my comments can be follow-up cleanup.

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: Making extension lifecycle an enum and assigning as a field had solved the problem I was trying to solve with the impossible states, and I marked it as resolved. Making the wrapper enum seemed to cause a bit more code duplication, so I didn't go for that. I feel that it is either or, not both. Do you think I am misunderstanding your suggestion? Otherwise, what direction would you prefer here?

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: seems like an AI doing unnecessary things, will get rid of it.

rust/otap-dataflow/crates/engine/src/lib.rs
gouslu: good catch, thanks, better to catch it now than later in a future PR.

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: very good catch, thanks!

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: On another thought, I have done the changes you requested, in a slightly different way, and made it so that wrapper is Local/Shared and lifecycle is Active/Passive, that way they are are different sets of configurations and added an `ExtensionBundle` type to return local and/or shared result from the factory. With these I am trying to achieve these:

- Extension factory can return either a local or a shared or both, not a vector.
- If extension factory returns shared, it also doubles as a local
- All of these variation combinations are possible:
  - `with_local(Active<Rc<LocalVariant>>).with_shared(Active<SharedVariant>)`
  - `with_local(Active<Rc<LocalVariant>>).with_shared(Passive<SharedVariant>)`
  - `with_local(Passive<Rc<LocalVariant>>).with_shared(Active<SharedVariant>)`
  - `with_local(Passive<Rc<LocalVariant>>).with_shared(Passive<SharedVariant>)`
  - `with_shared(Active<SharedVariant>)` // doubles as local
  - `with_shared(Passive<SharedVariant>)` // doubles as local
  - `with_local(Active<Rc<LocalVariant>>)`
  - `with_local(Passive<Rc<LocalVariant>>)`
- Dual registrations don't allow same concrete type, because if it did, then they'd allow the situation where authors might clone one instance and registering as local separately, which would be running 2 lifecycles on same Arc fields without a good reason
- Shared only registration generates a SharedAsLocal fallback (future PR) that uses the same instance, therefore the same lifecycle.

These design decisions lead me to return something like:

```rust
struct ExtensionWrapper {
  local: Option<ExtensionWrapper>,
  shared: Option<ExtensionWrapper>,
  ...
}
```

Combined with the builder it allowed me to write the logic based on whether local or shared available, fallback logic etc. As you requested, I converted ExtensionWrapper into enum, but I still need to return it as such:

```rust
struct ExtensionBundle {
  local: Option<ExtensionWrapper>,
  shared: Option<ExtensionWrapper>,
}
```

`ExtensionBundle` fields are private. the builder is the only construction path, ensuring `local` always holds a `Local` variant and `shared` always holds a `Shared` variant. Builder still has type id check to avoid local and shared to be the same type.

Does this work for you?

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: Extensions are owned by the engine, capabilities are given to the nodes. I can split control channel to local/shared, which I didn't do before, because it didn't seem necessary given that controlchannel is very low volume. But, I'll address this by splitting it so that local extensions get local receiver.

gouslu: Now local extensions use local receiver.

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: not using drop makes me do `let _`, either works I guess. did change it nonetheless.

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: control channels are split now!

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1253 @@
...
+                    biased;
+                    _ = sleep_until_deadline.as_mut().expect("set above") => {
+                        let shutdown = self
+                            .pending_shutdown
+                            .take()
+                            .expect("pending_shutdown must exist");
+                        self.shutdown();
+                        return Ok(shutdown);
+                    }
+                    msg = self.control_rx.as_mut().expect("checked above").recv() => {
```
lalitb, line 131: I think there’s one shutdown edge case here introduced with latest changes: once a delayed shutdown is latched in `pending_shutdown`, if the sender side closes before the deadline, this branch returns `RecvError::Closed` immediately instead of eventually returning the pending Shutdown.

I was expecting a latched shutdown to still win even if the channel closes afterward.

rust/otap-dataflow/crates/engine/src/extension.rs
```diff
@@ -0,0 +1,1253 @@
...
+        let dl = Instant::now() + std::time::Duration::from_millis(50);
+        let sender = SharedSender::mpsc(tx);
+        sender
+            .send(ExtensionControlMsg::Shutdown {
+                deadline: dl,
+                reason: "dl".into(),
+            })
+            .await
+            .unwrap();
+        // Keep sender alive so control_rx doesn't close before the deadline.
```
lalitb, line 1103: This seems to be working around the same delayed-shutdown edge case in `recv()`: the sender has to stay alive until the deadline for the pending shutdown to be returned.

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: addressed with https://github.com/open-telemetry/otel-arrow/pull/2622/changes/ad648fcac7b865625c3bd0834743c490f1b929e1

rust/otap-dataflow/crates/engine/src/extension.rs
gouslu: added `test_ctrl_delayed_shutdown_sender_dropped`

lquerel (approved):
I'm fine with the last updates.

Issue #2507, opened 2026-04-01
Auto generated telemetry documentation


andborja:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

The process to create and document telemetry is expensive and error prone. A mechanism to add and document telemetry in an automated way would be ideal.

### Proposed Solution

Implement a mechanism similar to mdatagen used in the go Opentelemetry Collector that generates metrics and documentation based on a metadata.yaml input file.
Some examples of manual telemetry documentation of a component can be found at:
- https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/engine/Telemetry.md
- https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/Telemetry.md

### Alternatives Considered

Manual documentation as it is done today

lquerel:
I'm totally in favor of auto-generating the documentation for our telemetry. It was mentioned in the project here: https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/telemetry/README.md#7-tooling-driven-validation-and-documentation-with-weaver

The idea was to rely on a format that is officially supported by OpenTelemetry, not `mdatagen`, but rather the semantic conventions format supported by two SIGs: the SemConv SIG and the Weaver SIG, of which I’m a member.

Andres, I can provide more details on these two OTel projects if you want.

jmacd:
Look at the documentation in https://github.com/open-telemetry/otel-arrow/tree/main/rust/otap-dataflow/docs/telemetry, has more detail on how we can integrate Weaver for metric code generation.

Issue #2411, opened 2026-03-23
Add first-class OTAP support for multivariate metric sets


lquerel:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Objective

Add native OTAP model, schema, payload, and transport support for multivariate metric sets so internal telemetry can represent them directly instead of flattening them into univariate form.

### Rationale

Our internal telemetry SDK treats multivariate metric sets as a first-class concept, but OTLP and OTAP do not yet have first-class support for them in the current implementation.

This leaves us relying on translations or univariate fallbacks, which limits efficiency and makes some semantics lossy.

### Scope

- define the OTAP payload and schema representation for multivariate metric sets
- support encode and decode paths
- support transport handling where needed
- preserve compatibility with internal telemetry metric-set semantics

### Acceptance Criteria

- [ ] OTAP has a non-placeholder representation for multivariate metric sets
- [ ] encode and decode paths support native metric-set transport
- [ ] internal telemetry metric sets can be exported without first flattening everything into ad hoc univariate equivalents
- [ ] follow-on work can build pipeline-native aggregation and detectors on top of this support

### Dependencies or Blockers

- #395

jmacd:
https://github.com/open-telemetry/otel-arrow/blob/main/docs/multivariate-design.md

Issue #2405, opened 2026-03-23
Unify internal telemetry around the OTAP pipeline


lquerel:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Objective

Converge internal telemetry toward a simpler architecture centered on our own pipeline:

- keep observed state authoritative for status and health until exporter-based replacements are ready
- keep the current registry and admin metrics endpoint in the short term
- allow short-term detector outputs to land on the current registry/admin path
- bridge metric export through our own telemetry pipeline instead of the direct OpenTelemetry client SDK path
- later add native OTAP support for metric sets and move aggregation, detectors, and Prometheus into pipeline nodes
- align signal names and attributes with the telemetry guidelines already defined in the project

### Rationale

Our internal telemetry is still split across several parallel paths instead of being unified around the OTAP pipeline.

Today we have all of the following in play:
- an internal telemetry pipeline for logs
- a dedicated observed-state/status path for engine lifecycle and health semantics
- a registry-backed metrics path exposed by an HTTP admin endpoint
- optional OpenTelemetry SDK metric readers/exporters configured from `engine.telemetry.metrics`
- OTAP metric handling still has explicit gaps around multivariate payload support.

We also already have project [telemetry guidelines](https://github.com/open-telemetry/otel-arrow/tree/main/rust/otap-dataflow/docs/telemetry) for naming, attributes, entity modeling, and stability, but implementation is not yet fully aligned with them.

### Scope

- #2406  
   This should happen early so new and migrated signals move closer to the documented model instead of adding more drift.
    
- #2407   
   Short-term detector outputs should land on the current registry/admin path.

- #2408   
   This is the stage-1 export bridge for OTLP or OTAP.

- #2409   
   Do this after the univariate bridge exists.

- #2410   
   This makes the current health/status semantics exportable without changing the source of truth.

- #2411   
   Replace the bridge with a native model.

- #2412   
   This is the final consolidation step.

### Acceptance Criteria

- [ ] All sub-tasks have been completed

### Dependencies or Blockers

- #1294 Export OTLP metrics from internal instrumentation
- #1378 Internal telemetry metrics dispatcher
- #1872 Proposal 9 - Rationalization and simplification of the proliferation of settings related to internal concerns

jmacd:
All of these statements and sub-issues look good to me. I will add a few in separate sub-issues, briefly summarized:

[...]

PR #2623, opened 2026-04-10
Internal metrics telemetry pipeline and SDK design

### LINKED ISSUE #1378: Internal telemetry metrics dispatcher

jmacd:
# Change Summary

Part of #1950; a design only.

Part of #2405

Fixes #1378 

Part of #2411

Part of #2507 

Proposes an OTAP-direct metrics SDK in four phases.


rust/otap-dataflow/docs/internal-metrics-sdk.md
```diff
@@ -0,0 +1,497 @@
...
+        requirement_level: recommended
+      - ref: signal
+        requirement_level: optional
+    x-otap-levels:
+      basic:
+        disabled: true
+      normal:
+        dimensions: [outcome]
+      detailed:
+        dimensions: [outcome, signal]
```
lquerel, line 128: This won’t be recognized by Weaver as is. For this kind of thing, I think the best approach is to use the annotation mechanism.

```yaml
groups:
  - id: metric.consumer.items
    type: metric
    metric_name: consumer.items
    instrument: counter
    unit: "{item}"
    brief: "Items consumed by a node."
    attributes:
      - ref: outcome
        requirement_level: recommended
      - ref: signal
        requirement_level: optional
    annotations:
      x-otap-levels:
        basic:
          disabled: true
        normal:
          dimensions: [outcome]
        detailed:
          dimensions: [outcome, signal]
```

rust/otap-dataflow/docs/internal-metrics-sdk.md
```diff
@@ -0,0 +1,497 @@
...
+
+    #[metric(unit = "{item}")]
+    consumed_failed: Counter<u64>,
+
+    #[metric(unit = "{item}")]
+    consumed_refused: Counter<u64>,
+
+    // ... and more (e.g., duration)
+}
+```
```
lquerel, line 91: To cover all aspects of metric sets, I suggest adding another typical kind of metric set like the one below (in complement). In this example, the metrics in the set cannot only be distinguished by one or two dimensions such as outcome and signal type.

```rust
/// Engine-wide metrics emitted once per engine instance.
#[metric_set(name = "engine.metrics")]
pub struct EngineMetrics {
    /// Process-wide Resident Set Size — physical RAM currently used by the process.
    /// Matches what external tools report (e.g. `kubectl top pod`, `htop`, `ps rss`).
    #[metric(unit = "{By}")]
    pub memory_rss: ObserveUpDownCounter<u64>,

    /// Process-wide CPU utilization as a ratio in [0, 1], normalized across all
    /// logical CPU cores on the system (not just engine-assigned cores).
    /// Aligned with the OTel semantic convention `process.cpu.utilization`.
    ///
    /// The `cpu.mode` attribute is not set; this reports combined user + system time.
    #[metric(unit = "{1}")]
    pub cpu_utilization: Gauge<f64>,

    /// Process-wide memory limiter state encoded as `0=normal`, `1=soft`, `2=hard`.
    #[metric(unit = "{state}")]
    pub memory_pressure_state: Gauge<u64>,

    /// Most recent process-wide memory limiter sample, in bytes.
    #[metric(unit = "{By}")]
    pub process_memory_usage_bytes: Gauge<u64>,

    /// Effective process-wide memory limiter soft limit, in bytes.
    #[metric(unit = "{By}")]
    pub process_memory_soft_limit_bytes: Gauge<u64>,

    /// Effective process-wide memory limiter hard limit, in bytes.
    #[metric(unit = "{By}")]
    pub process_memory_hard_limit_bytes: Gauge<u64>,
}
```

rust/otap-dataflow/docs/internal-metrics-sdk.md
```diff
@@ -0,0 +1,497 @@
...
+    // ... and more (e.g., duration)
+}
+
+#[derive(Debug, Default, Clone)]
+pub enum ConsumedItemsByOutcomeAndSignal {
+    #[default]
+    Basic,                       // disabled
+    Normal(Box<[Counter; 3]>),   // 1 dimension: outcome
+    Detailed(Box<[Counter; 9]>), // 2 dimensions: outcome x signal
+}
```
lquerel, line 153: I think I need to see the code as a whole to understand.

What I'm wondering is: why don't we have more or less the same code as before, except that instead of being generated by a macro, it would be generated by Weaver from the semantic conventions?

rust/otap-dataflow/docs/internal-metrics-sdk.md
```diff
@@ -0,0 +1,497 @@
...
+```
+
+Note the use of `Box<_>` in the enum variant ensures that metric level
+actually controls how much memory is used by instrumentation.
+
+In another file, we will define the translation from the `v1` to `v0`
+schema, which we will eventually deprecate. Separately, we will
+define how to generate either the `v1` or the `v0` schema from the
+current instrumentation.
```
lquerel, line 173: I’m confused, but probably I just don't have the full picture yet.

rust/otap-dataflow/docs/internal-metrics-sdk.md
```diff
@@ -0,0 +1,497 @@
...
+The two schemas may be configured as (using a scalar value for a
+single schema selection):
+
+```yaml
+engine:
+  telemetry:
+    scopes:
+      metrics:
+        metrics.otap.consumer:
+          schema_url: otap-dataflow/consumer@v1
```
lquerel, line 298: I don't understand. Are we supposed to list all the metric sets in the configuration? That seems strange to me.

rust/otap-dataflow/docs/internal-metrics-sdk.md
```diff
@@ -0,0 +1,497 @@
...
+single set of attributes and timestamp, significantly improving
+compression and row efficiency. This remains future work in the OTAP
+specification.
+
+## Phase 3: Remove OpenTelemetry SDK
+
+With all `#[metric_set]` definitions replaced by generated code, we
+remove the OpenTelemetry SDK from the OTAP Dataflow dependencies.
+Users configure either the Builtin Prometheus support or an ITS
+internal telemetry pipeline (e.g., batch processor, OTLP exporter).
```
lquerel, line 415: I'm not saying this is necessarily a better plan, but note that removing the OpenTelemetry SDK could be done as early as phase 1, even before introducing semantic conventions and code generation with Weaver.

lquerel (commented):
I haven't understood everything yet. I'll probably need to have a discussion with you to better understand the details of your plan :-)

What I had in mind was something like:
- Support for bounded dynamic attributes (e.g. outcome, signal type) in the current metric set system
- Converting the `TelemetryRegistry` into an internal metric receiver
- Removing the OpenTelemetry SDK
- Creating the semantic conventions
- Replacing the code generated by the `metric_set` macro with code generated by Weaver

jmacd:
I will build more prototype as I incorporate the feedback.

PR #2624, opened 2026-04-10
 feat: Remove experimental-tls feature flag and make TLS always available 

lalitb:
### Summary                                 
                                          
  Remove the `experimental-tls` feature gate and make TLS support available by
  default across OTAP Dataflow.                                                                                                                                                           
   
  ### What changed                                                                                                                                                                        
                  
  - Removed the `experimental-tls` feature wiring from the workspace and all                                                                                                              
    affected crates
  - Made the core TLS dependencies in `otap-df-otap` unconditional
  - Removed feature-gated TLS fallback paths and the obsolete
    `TlsFeatureDisabled` error variant    
  - Made existing TLS tests compile and run by default
  - Updated configs, scripts, and docs to stop referring to `experimental-tls`                                                                                                            
  - Added a binary-level compile-time guard in `df_engine` so normal builds must
    enable exactly one crypto provider:                                                                                                                                                   
    - `crypto-ring`
    - `crypto-aws-lc`                                                                                                                                                                     
    - `crypto-openssl`                        
                                                                                                                                                                                          
  ### Notes       
                                                                                                                                                                                          
  - This change does not alter the existing `crypto-*` feature flags; it only
    removes the compile-time gate around TLS availability.                                                                                                                                
  - `tonic/tls-native-roots` was intentionally not made unconditional. Native
    trust anchors are loaded directly via `rustls_native_certs` in the TLS helper
    paths, so this is not an omission.

### DIFF
--- a/.github/workflows/rust-ci.yml
+++ b/.github/workflows/rust-ci.yml
@@ -376,6 +376,9 @@ jobs:
       - name: cargo check benchmarks --no-default-features
         run: cargo check -p benchmarks --benches --no-default-features
         working-directory: ./rust/${{ matrix.folder }}
+      - name: cargo test otap integration targets --no-run
+        run: cargo test -p otap-df-otap --tests --no-run
+        working-directory: ./rust/${{ matrix.folder }}
 
   # Required matrix combinations for test_and_coverage: otap-dataflow on ubuntu and windows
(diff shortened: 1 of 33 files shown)


lquerel (approved):
LGTM

PR #2626, opened 2026-04-10
Columnar Query Engine/OPL Parser support for `TextScalarExpression::Capture`

### LINKED ISSUE #2578: OPL/Columnar query engine support `TextScalarExpression`
In our AST expression tree, we have `TextScalarExpression` representing operations: concat, join, replace and regex capture group:
https://github.com/open-telemetry/otel-arrow/blob/72fba8d2a94cd5e20403875e9214756a86cd405f/rust/experimental/query_engine/expressions/src/scalars/text_scalar_expression.rs#L7-L23

We should add support to these for the columnar query engine & OPL parser.

The first three (concat, join and replace) are very straight forward because there are datafusion scalar expressions for these (join is called `concat_ws`).

regex capture group is a bit more complicated, because there's no datafusion equivalent so we'll need to write something custom. See the implementation of https://github.com/apache/datafusion/blob/main/datafusion/functions/src/regex/regexpinstr.rs for inspiration

albertlockett:
# Change Summary

Partially support `TextScalarExpression::Capture` in columnar query engine.
https://github.com/open-telemetry/otel-arrow/blob/b5f0814099566c119a29aa8465a137e04adbeeb4/rust/experimental/query_engine/expressions/src/scalars/text_scalar_expression.rs#L7-L11

This currently only supports identifying the capture group by number or name.

In KQL, the `extract` function gets parsed into this type:
```kql
logs | set severity_text = extract(".*severity=(.*).*", 1, event_name)
```

In OPL, I added a function called `regexp_capture` that has similar arguments:
```kql
logs | set severity_text = regexp_capture(event_name, ".*severity=(.*).*", 1)
```

Named capture groups are also supported:
```kql
logs | set attributes["x"] = regexp_capture(attributes["y"], ".*(?P<mygroup>.....+) ", "mygroup")
```

The order of the arguments are slightly different from KQL's `extract` function. This follows the convention used in datafusion's [regexp scalar functions](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#regular-expression-functions) (which are based on various functions available in SQL) where the first argument is the string and second is the pattern.

Underneath the hood, this is implemented using a ScalarUDF called `regexp_substr`, that accepts the same signature found in many SQL dialetcs (such as https://docs.snowflake.com/en/sql-reference/functions/regexp_substr), with the additional capability that this can be called using a named group instead of a group number. This is probably overkill for just the capture group extraction case, but I was thinking it was best to do something flexible/standard and that we could probably try donating this to upstream datafusion eventually.

This function is also available to be called in OPL/KQL directly:
```kql
logs | extend attributes["s1"] = regexp_substr(attributes["attr"], "hell.")
```

The `regex_substr` function is optimized for the case all the arguments are scalars, except for the first argument which is the source string column. If any of the other arguments are not scalar, it casts the array into either a  string array or Int64 array. This keeps the imlementation straight forward, and having a ton of edge cases for dict handling.

There's currently some funny logic in the handling for the optional parameters that I'm not proud of. Datafusion functions support optional arguments (via multiple signatures), but our expression tree expects every function to have a well-defined number of parameters, and the defaults are actually supplied to the parser state. Currently KQL is using these  - if params are supplied in the query, it fills them in. OPL parser isn't doing this and just delegates to the function to figure it out internally. In the future, I'll probably try to fix up the expressions we use for function signature in our expression tree to add the concept of optional parameters. 

* Closes #2578

unit

Yes - these functions are available for use in transform processor.

### DIFF
--- a/rust/otap-dataflow/crates/opl/src/parser/expression.rs
+++ b/rust/otap-dataflow/crates/opl/src/parser/expression.rs
@@ -5,14 +5,15 @@ use std::sync::LazyLock;
 
 use data_engine_expressions::{
     AndLogicalExpression, BinaryMathematicalScalarExpression, BooleanScalarExpression,
-    CollectionScalarExpression, CombineScalarExpression, ContainsLogicalExpression,
-    DoubleScalarExpression, DoubleValue, EqualToLogicalExpression, Expression,
-    GreaterThanLogicalExpression, GreaterThanOrEqualToLogicalExpression, IntegerScalarExpression,
-    IntegerValue, InvokeFunctionArgument, InvokeFunctionScalarExpression, JoinTextScalarExpression,
-    ListScalarExpression, LogicalExpression, MatchesLogicalExpression, MathScalarExpression,
-    NotLogicalExpression, NullScalarExpression, OrLogicalExpression, QueryLocation,
-    ReplaceTextScalarExpression, ScalarExpression, SliceScalarExpression, SourceScalarExpression,
-    StaticScalarExpression, StringScalarExpression, TextScalarExpression, ValueAccessor,
+    CaptureTextScalarExpression, CollectionScalarExpression, CombineScalarExpression,
+    ContainsLogicalExpression, DoubleScalarExpression, DoubleValue, EqualToLogicalExpression,
+    Expression, GreaterThanLogicalExpression, GreaterThanOrEqualToLogicalExpression,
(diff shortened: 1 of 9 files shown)


lquerel:
An idea for later: it would be good to come up with a syntax that lets us extract multiple groups so we can assign them to multiple fields or attributes.

lquerel (approved):
LGTM

PR #2627, opened 2026-04-10
refactor(engine): extract IndexedMinHeap from node-local wakeup scheduler

### LINKED ISSUE #2587: Extract reusable IndexedMinHeap from node_local_scheduler
### Pre-filing checklist

### Component(s)

﻿The NodeLocalScheduler in crates/engine/src/node_local_scheduler.rs contains a hand-rolled indexed min-heap (~80 lines of sift_up/sift_down/swap_entries/repair_heap_at/remove_heap_entry) that supports O(log n) replacement and cancellation. This is tightly coupled to the scheduler.

### Proposed Solution

﻿We should extract it into a generic IndexedMinHeap<K, V> in a standalone data-structures module (or crate). Benefits:

 - Reuse: pipeline_ctrl.rs and fanout_processor both use BinaryHeap<Reverse<…>>, might benefit
 - Testability: heap invariants get their own focused unit tests, independent of wakeup scheduling logic.
 - Readability: NodeLocalScheduler shrinks to pure scheduling semantics.

### Alternatives Considered

﻿Alternatively, evaluate whether an existing crate provides the same indexed-heap semantics we need.

### Additional Context

https://github.com/open-telemetry/otel-arrow/pull/2470#discussion_r3034500008

sjmsft:
# Change Summary

Extracts the hand-rolled binary heap (Vec<ScheduledWakeup> + HashMap<WakeupSlot, usize>) from NodeLocalScheduler into a generic, reusable IndexedMinHeap<K, P> data structure with its own module and test suite.

* Closes #2587 

cargo test -p otap-df-engine -- "indexed_min_heap|node_local_scheduler"

Existing unit tests in node_local_scheduler and new unit tests in indexed_min_heap.

No. All changes are internal to otap-df-engine. No downstream crate or user-facing behavior changes.

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/indexed_min_heap.rs
+++ b/rust/otap-dataflow/crates/engine/src/indexed_min_heap.rs
@@ -14,7 +14,7 @@ use std::hash::Hash;
 
 /// Outcome of an [`IndexedMinHeap::insert`] call.
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub enum InsertOutcome {
+pub(crate) enum InsertOutcome {
     /// A new entry was added to the heap.
     Inserted,
     /// An existing entry's priority was replaced.
@@ -27,7 +27,7 @@ pub enum InsertOutcome {
 /// Entries are `(K, P)` pairs.  The heap is ordered by `P` where the
 /// *smallest* priority sits at the root.  Keys must be unique — inserting
 /// a key that already exists replaces its priority in place.
-pub struct IndexedMinHeap<K, P> {
+pub(crate) struct IndexedMinHeap<K, P> {
     entries: Vec<Entry<K, P>>,
     indices: HashMap<K, usize>,
 }
--- a/rust/otap-dataflow/crates/engine/src/indexed_min_heap.rs
+++ b/rust/otap-dataflow/crates/engine/src/indexed_min_heap.rs
@@ -95,7 +95,10 @@ where
             InsertOutcome::Replaced
         } else {
             let index = self.entries.len();
-            self.entries.push(Entry { key: key.clone(), priority });
+            self.entries.push(Entry {
+                key: key.clone(),
+                priority,
+            });
             assert!(
                 self.indices.insert(key, index).is_none(),
                 "new key should not already exist in index map"
--- a/rust/otap-dataflow/crates/engine/src/indexed_min_heap.rs
+++ b/rust/otap-dataflow/crates/engine/src/indexed_min_heap.rs
@@ -1,0 +1,481 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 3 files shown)


rust/otap-dataflow/crates/engine/src/indexed_min_heap.rs
```diff
@@ -0,0 +1,481 @@
...
+    Replaced,
+}
+
+/// A binary min-heap that supports O(1) key lookup and O(log n) keyed
+/// insertion, removal, and in-place priority updates.
+///
+/// Entries are `(K, P)` pairs.  The heap is ordered by `P` where the
+/// *smallest* priority sits at the root.  Keys must be unique — inserting
+/// a key that already exists replaces its priority in place.
+pub struct IndexedMinHeap<K, P> {
```
lalitb: Making this `pub` does make `IndexedMinHeap` and `InsertOutcome` part of the crate’s public API. If the current goal is only internal reuse within `otap-df-engine`, I’d lean toward keeping it `pub(crate)` for now and only making it public once there’s a concrete external use case.

rust/otap-dataflow/crates/engine/src/node_local_scheduler.rs
```diff
@@ -459,15 +331,7 @@ mod tests {
...
-        let moved_index = scheduler
-            .wakeup_indices
-            .get(&WakeupSlot(3))
-            .copied()
-            .expect("rescheduled slot should still be tracked");
-        assert!(
-            moved_index > 0,
-            "rescheduled slot should be a non-root entry"
-        );
+        assert!(scheduler.heap.contains_key(&WakeupSlot(3)));
```
lalitb, line 334: This test got a bit weaker in the refactor. Before, it checked that the rescheduled wakeup had moved off the root before cancellation. Now it only checks `contains_key`, which shows the entry is still there but doesn’t verify that specific case anymore. Can we add that check back, or replace it with something equivalent like asserting that `WakeupSlot(3)` is not the current root?

rust/otap-dataflow/crates/engine/src/indexed_min_heap.rs
```diff
@@ -0,0 +1,481 @@
...
+
+#[derive(Clone, Copy, Debug)]
+struct Entry<K, P> {
+    key: K,
+    priority: P,
+}
+
+impl<K, P> IndexedMinHeap<K, P>
+where
+    K: Eq + Hash + Copy + std::fmt::Debug,
```
lalitb: `K: Copy` rules out common owned key types like `String` or `Arc<str>`. Since this is being exposed as a generic reusable structure, that feels more restrictive than necessary. The implementation only seems to rely on copying keys in a few spots (`swap`, `pop`, `remove_at`), so switching this to `Clone` looks like it would make the API more broadly usable without adding much complexity.

lalitb (approved):
LGTM. The `fmt` CI is failing, so need to fix that to enable merge.

jmacd (approved):

PR #2629, opened 2026-04-10
test: improve handling of channel-send shutdown in durable_buffer_processor_tests

AaronRM:
# Change Summary

Refactor how test code determines acceptable shutdown results in `durable_buffer_processor_tests.rs`. Don't rely on error message matching.

n/a

Verified that the tests pass locally.

No.

### DIFF
--- a/rust/otap-dataflow/crates/otap/tests/durable_buffer_processor_tests.rs
+++ b/rust/otap-dataflow/crates/otap/tests/durable_buffer_processor_tests.rs
@@ -638,11 +638,10 @@ where
 fn is_acceptable_shutdown_result(
     run_result: &Result<Vec<()>, otap_df_engine::error::Error>,
 ) -> bool {
-    match run_result {
-        Ok(_) => true,
-        Err(otap_df_engine::error::Error::ChannelSendError { error }) => error.contains("closed"),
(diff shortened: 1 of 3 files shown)


drewrelmas (approved):

rust/otap-dataflow/crates/otap/tests/durable_buffer_processor_tests.rs
```diff
@@ -634,6 +630,17 @@ where
...
 // ─────────────────────────────────────────────────────────────────────────────
 
+/// Return true when pipeline shutdown completed cleanly or hit an expected
+/// shutdown-time channel close race.
+fn is_acceptable_shutdown_result(
+    run_result: &Result<Vec<()>, otap_df_engine::error::Error>,
+) -> bool {
+    matches!(
+        run_result,
+        Ok(_) | Err(otap_df_engine::error::Error::ChannelSendError { .. })
```
lalitb: This broadens the accepted shutdown behavior beyond the previous test semantics. Not a blocker since this is only test code, but could we keep the helper and only accept the closed-channel case?

lalitb (approved):

lquerel (approved):

PR #2631, opened 2026-04-11
Fix topic shutdown and mixed publish semantics

### LINKED ISSUE #2630: Topic shutdown under backpressure and mixed-topic publish semantics need cleanup
# Change Summary

Topic-based pipelines currently have a cluster of shutdown, backpressure, and mixed-delivery semantic problems that make rollout and shutdown behavior unreliable under load.

The main symptoms observed during live validation were:
- topic receiver drain could report drained while the receiver still kept running
- topic exporter could block shutdown when queue_on_full: block was configured
- mixed-topic try_publish could deliver to broadcast subscribers even when balanced delivery rejected the same message
- mixed async publish could reserve permits across balanced groups in a convoy-prone way while still waiting for another group

These issues are closely related because they all stem from node-side emulation of topic waiting and admission behavior instead of letting the topic runtime own those semantics.

### 1. Topic receiver drain behavior was not shutdown-safe
During receiver-first shutdown, the topic receiver could acknowledge ingress drain too early while still polling or otherwise remaining active. Under sustained traffic this could leave replace rollouts, pipeline shutdown, or group shutdown waiting until drain deadlines expired.

### 2. Topic exporter could block coordinated shutdown under backpressure
When queue_on_full: block was used, the topic exporter could remain stuck behind blocked publish behavior and fail to respond promptly to shutdown. That prevented clean downstream drain and contributed to pipeline shutdown timeouts.

### 3. Mixed-topic try_publish semantics were inconsistent
For mixed topics, the non-blocking path could publish to the broadcast ring before balanced-group admission had succeeded. That meant broadcast subscribers could observe a message that the balanced side had already rejected with DroppedOnFull, which broke the all-or-nothing backpressure model and diverged from the async publish behavior.

### 4. Mixed async publish could convoy unrelated balanced groups
The async mixed publish path acquired one balanced-group permit at a time and held earlier permits while waiting on later groups. Under partial saturation, one slow group could unnecessarily reserve capacity in unrelated groups even though no message had been admitted yet.

## Expected Direction

The cleanup should keep waiting and admission semantics inside the topic runtime rather than duplicating them in topic nodes.

That implies:
- topic-owned blocked publish waiting
- delivery-lease handling on the subscribe side
- shutdown-responsive receiver and exporter behavior without node-side retry loops on the hot path
- all-or-nothing semantics for mixed publish across balanced and broadcast delivery
- no partial permit reservation across balanced groups while waiting for another group

## Why This Matters

These problems directly affect correctness of mixed-topic delivery semantics, predictability of backpressure behavior, debuggability of rollout and shutdown behavior under load, and graceful shutdown reliability in topic-heavy topologies.

## Validation Signals

The issues were surfaced by live validation of topic-heavy scenarios, especially multi-tenant topic topologies with sustained traffic, where:
- replace rollout could get stuck in draining-old and fail on drain deadline
- pipeline shutdown could time out while waiting for drain
- group shutdown could time out under coordinated shutdown pressure

## Follow-up

A focused PR has been extracted to address this topic-runtime and topic-node behavior independently from the larger live-reconfiguration work.

lquerel:
## Change Summary

This PR extracts the topic-runtime fixes and refactorings from the larger live-reconfiguration branch into a standalone change.

The core design decision in this PR is to move topic-specific waiting and delivery semantics back into the topic runtime instead of emulating them in the topic receiver and topic exporter. Before this change, the node layer carried its own retry/wait loops to handle blocked publish, downstream backpressure, and shutdown/drain behavior. **That made shutdown under load fragile and allowed topic semantics to diverge across code paths**.

Concretely, this PR does four things:

1. It makes the topic runtime own blocked publish waiting.
   - `publish(...)` / `publish_tracked(...)` now wait inside the topic implementation instead of relying on node-side polling loops.
   - This removes the old timer-driven retry behavior from the topic exporter hot path and keeps `queue_on_full: block` wake-driven and shutdown-responsive.

2. It introduces delivery leases on the subscribe side and uses them in the topic receiver.
   - The receiver now distinguishes “the topic produced a message” from “the receiver successfully handed it downstream”.
   - That allows `DrainIngress` to stop polling the topic immediately while still waiting for already-forwarded tracked messages to reach their terminal Ack/Nack outcome.
   - This fixes the previous shutdown bug where the topic receiver could report drained state while continuing to run under load.

3. It tightens mixed-topic semantics so publish behavior is consistent across code paths.
   - Mixed `try_publish` is now all-or-nothing across balanced and broadcast delivery.
   - If balanced delivery cannot admit a message, broadcast subscribers do not receive it either.
   - Mixed async publish was also changed to avoid holding balanced-group permits while waiting on another full group, which removes the previous partial-permit convoy behavior.

4. It keeps the refactor performance-oriented rather than purely structural.
   - The topic receiver/exporter fast paths still try immediate forward/publish first.
   - The in-memory delivery-lease path now uses specialized inline storage instead of per-message boxing on the common path.
   - The opaque fallback is still preserved for future non-in-memory backends, but the in-memory backend no longer pays that allocation cost per delivered message.

**This PR does not add new public routes, config knobs, or admin APIs.** The existing topic publish/subscribe APIs remain source-compatible; the main changes are semantic correctness under backpressure/shutdown and simplification of where the waiting logic lives.

* Closes [#2630](https://github.com/open-telemetry/otel-arrow/issues/2630)

- `cargo xtask check`

Key topic/runtime guarantees covered by tests:
- Balanced topics deliver each message exactly once within a subscriber group.
- Balanced topics preserve per-subscriber ordering.
- Broadcast topics deliver all messages to all subscribers in order.
- Broadcast lag handling is isolated to the lagging subscriber and does not block healthy subscribers.
- Mixed async publish does not expose a message to broadcast subscribers before balanced admission succeeds.
- Mixed `try_publish` is all-or-nothing: if balanced admission fails, broadcast delivery does not happen.
- Mixed async publish does not reserve permits in unrelated balanced groups while waiting on a full group.
- Dropping a blocked mixed publish does not leak a partial broadcast delivery.
- Blocking tracked publish does not leak in-flight capacity when canceled.
- Delivery leases preserve tracked message outcome semantics, including abort-to-nack behavior.
- The in-memory delivery lease fast path uses specialized inline storage, while the opaque fallback path still works.
- Topic receiver `DrainIngress` exits promptly when Ack/Nack propagation is disabled.
- Topic receiver `DrainIngress` waits for already-forwarded tracked work to complete when Ack/Nack propagation is enabled.
- Topic receiver drain can interrupt a blocked downstream forward without hanging shutdown.
- Topic exporter shutdown interrupts blocked untracked publish under `queue_on_full: block`.
- Topic exporter shutdown interrupts blocked tracked publish under `queue_on_full: block`.
- Topic exporter shutdown force-resolves buffered in-flight pdata it still owns instead of hanging.
- End-to-end topic exporter -> topic receiver flow still transfers pdata correctly.
- End-to-end topic receiver source tagging still works when enabled.

In addition, this topic work was exercised in the larger live-reconfiguration branch against the scenario that originally exposed the shutdown issue (`topic_multitenant_isolation.yaml`), including replace rollout, per-pipeline shutdown, and group shutdown flows.

Yes.

- Topic-based pipelines shut down more reliably under backpressure.
- Topic receiver drain behavior is now stricter and better aligned with Ack/Nack propagation semantics.
- Mixed-topic publish is now all-or-nothing across balanced and broadcast delivery.
- There are no new public APIs, config knobs, or route changes in this PR.

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/topic_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/topic_exporter/mod.rs
@@ -31,7 +31,8 @@ use otap_df_telemetry::{otel_info, otel_warn};
 use otap_df_telemetry_macros::metric_set;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
-use std::collections::HashMap;
+use std::collections::{HashMap, VecDeque};
+use std::future;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Duration;
@@ -95,6 +96,34 @@ pub struct TopicExporter {
     metrics: MetricSet<TopicExporterMetrics>,
 }
 
+/// One upstream pdata message waiting for topic admission under `queue_on_full: block`.
+struct PendingPublish {
+    data: OtapPdata,
+    published: Arc<OtapPdata>,
+    should_track_end_to_end: bool,
+}
+
+impl PendingPublish {
+    /// Build the retained publish attempt state for one upstream pdata message.
+    fn new(data: OtapPdata, should_track_end_to_end: bool) -> Self {
+        Self {
+            published: Arc::new(data.clone_without_context()),
+            data,
+            should_track_end_to_end,
+        }
+    }
+}
+
+/// Result of one non-blocking attempt to admit a pending publish.
+enum PendingPublishProgress {
+    Published,
+    Pending,
+}
+
+/// Retry cadence while a topic publish remains blocked after an immediate
+/// fast-path admission attempt already failed.
+const PENDING_PUBLISH_RETRY_DELAY: Duration = Duration::from_millis(1);
+
 /// Declares the topic exporter as a local exporter factory.
 #[allow(unsafe_code)]
 #[distributed_slice(OTAP_EXPORTER_FACTORIES)]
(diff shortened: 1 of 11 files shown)


rust/otap-dataflow/crates/otap/tests/durable_buffer_processor_tests.rs
```diff
@@ -601,13 +601,17 @@ where
...
+    // return a NACK after shutdown has already started tearing down downstream
+    // channels. Depending on where the close is observed, the error may come
+    // from the control channel or from the data path.
     let is_acceptable_shutdown = match &run_result {
         Ok(_) => true,
-        Err(e) => e.to_string().contains("Channel is closed"),
+        Err(e) => {
+            let message = e.to_string();
+            message.contains("Channel is closed") || message.contains("downstream channel closed")
+        }
```
lquerel: This fix is temporary, and I haven't fundamentally changed the existing approach. In the long term, we need a better fix that doesn’t rely on the content of the error message.

rust/otap-dataflow/crates/core-nodes/src/exporters/topic_exporter/mod.rs
```diff
@@ -152,6 +167,204 @@ impl TopicExporter {
...
+            Pin<Box<dyn Future<Output = (u64, TrackedPublishOutcome)> + Send>>,
+        >,
+    ) -> Result<Option<BlockedPublish>, Error> {
+        let should_track_end_to_end = ack_propagation_mode == TopicAckPropagationMode::Auto
+            && data.has_ack_or_nack_interests();
+
+        match queue_on_full {
+            TopicQueueOnFullPolicy::Block => {
+                let published = Arc::new(data.clone_without_context());
+                if should_track_end_to_end {
```
lquerel, line 266: Preserve a cheap uncontended fast path: only retain a blocked publish when the topic runtime reports real backpressure.

rust/otap-dataflow/crates/core-nodes/src/receivers/topic_receiver/mod.rs
```diff
@@ -233,14 +248,276 @@ impl local::Receiver<OtapPdata> for TopicReceiver {
...
+        let mut telemetry_cancel_handle = Some(
+            effect_handler
+                .start_periodic_telemetry(Duration::from_secs(1))
+                .await?,
+        );
+        let mut draining_deadline: Option<Instant> = None;
         let mut draining_reason: Option<String> = None;
-        let mut drained_notified = false;
+        let mut pending_tracked_message_ids = HashSet::new();
+        let mut pending_forward: Option<PendingForward> = None;
```
lquerel, line 259: These represent two different handoff stages:
- `pending_tracked_message_ids` are deliveries that were already forwarded and committed locally, but whose downstream Ack/Nack has not been bridged back to the topic runtime yet.
- `pending_forward` is one leased topic delivery that is still trying to enter the downstream pipeline and therefore must not be committed yet.

rust/otap-dataflow/crates/core-nodes/src/receivers/topic_receiver/mod.rs
```diff
@@ -345,87 +623,76 @@ impl local::Receiver<OtapPdata> for TopicReceiver {
...
-                                    otel_warn!(
-                                        "topic_receiver.downstream_backpressure",
-                                        node = receiver_id.name.as_ref(),
-                                        topic = config.topic.as_ref(),
-                                        blocked_ms = blocked_for.as_millis() as u64,
-                                        message = "Topic receiver blocked while forwarding to downstream pipeline channel"
-                                    );
+                                match effect_handler.try_send_message_with_source_node(pdata) {
+                                    Ok(()) => {
+                                        delivery.commit();
```
lquerel, line 662: Commit the topic lease only after the downstream pipeline accepts the message. That keeps drain precise: an unadmitted message can still be aborted locally instead of turning into tracked in-flight work.

rust/otap-dataflow/crates/engine/src/topic/topic.rs
```diff
@@ -635,33 +691,71 @@ impl<T: Send + Sync + 'static> MixedTopic<T> {
...
-                sender
-                    .send(envelope.clone())
-                    .await
-                    .map_err(|_| TopicClosed)?;
+            for (group, permit) in groups.as_ref().iter().zip(permits.into_iter()) {
+                send_queued_envelope(&group.tx, envelope.clone(), permit)?;
             }
         }
 
+        self.broadcast_ring.publish(Envelope {
```
lquerel, line 753: Broadcast is intentionally last so mixed async publish has the same visible contract as mixed try_publish: no broadcast subscriber can observe a message before the balanced side has admitted it.

jmacd (approved):
There is a small amount of confusion over the "lease" which if I read correctly is also called "permit". Since the "permit" term is already there, I suggest replacing the few "lease" uses with "permit".

lalitb (approved):
LGTM, the topic semantics seems much cleaner and more correct now.

lquerel:
@jmacd @lalitb thanks for the review. I make this renaming before merging. The next PR will be BIG in multiple dimensions...

PR #2632, opened 2026-04-11
Bridge columnar query engine's expr eval into filter code (Part 1)

albertlockett:
# Change Summary

Bridges the columnar query engine's expression evaluation into the filter evaluation.

Before this change, we only had the capability in the columnar query engine to evaluate filters where at least one side of some relative expression was a literal (e.g. `severity_text == "ERROR`). This is clearly very limiting, and there's a whole class of logical expressions that we wouldn't evaluate, such as comparing two fields (`attributes["x"] > attributes["y"]`) or using functions / arithmetic on either side of expression (such as `substring(event_name, 0, 3) == substring(severity_text, 0, 3)`).

This is part 1 of what will be multiple PRs to complete this work. 

Future PRs will include:
- Type mismatches - currently if the expressions on either side of the relative binary don't evaluate to the same type (or comparable types), the execution fails with an error. This will include correct handling of nulls (e.g. something evaluating to `null == null` should probably result in `true`)
- Bridging the code in the opposite direction - currently we don't support evaluating logical expressions outside the context of a filter. E.g., we don't yet support `logs | set attributes["x"] = attributes["y"] > attributes["z"]`

* Relates to #1508

Unit tests

Yes - these new filtering capabilities are now available in transform processor.

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
@@ -122,7 +122,7 @@ impl DataScope {
     /// Rules:
     /// - Any scope can combine with StaticScalar (constants)
     /// - Same scopes can combine (e.g., Root + Root), because the row order is the same.
-    fn can_combine(&self, other: &Self) -> bool {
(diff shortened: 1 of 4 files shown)


lquerel (approved):
LGTM

PR #2635, opened 2026-04-12
Support function calls where args have differing row orders

### LINKED ISSUE #2634: Columnar Query Engine function calls with columns of different row order
Consider calling a function in OPL or KQL like `my_func(severity_text, attributes{"x"])`

In these cases, the severity_text would have the "row order" of the root record batch, where attributes["x"] would have the "row order" of however the parent_ids were sorted in rows where key == "x".

by "row order" here, I mean the logical signal (e.g. log, span, etc.) represented by each row.

Currently the columnar query engine doesn't support this and will just return an error. We should fix this.

albertlockett:
# Change Summary

Adds capability to the columnar query engine to invoke functions where the columns passed as arguments have a different "row order"..

Consider calling a function in OPL or KQL like `my_func(severity_text, attributes["x"])`. In these cases, the `severity_text` would have the "row order" of the root record batch, where `attributes["x"]` would have the "row order" of however the `parent_id`s  were sorted in the Log Attrs batch rows where `key == "x"`. 

(We can think of "row order" here as order we'd encounter a row belonging to some signal (e.g. log, span, etc.) represented as we scan the column).

Currently the columnar query engine doesn't handle the case where function args have different "row order" and will just return an error. This PR resolves the issue. 

It does this by determining if the arguments to some function have different "row order" (in the parlance of the engine's expressions planning, we say they have incompatible `DataScope`s), and if the scopes are incompatible we must align the rows in each column by performing one or more joins before calling the function.

Noe: the engine's planned expressions are a tree of "scoped exprs" where each node in the tree represents a datafusion expression operating on a single data scope. While evaluating the tree at each node, we take data from the source and possibly perform "joins" on the input (depending on the data source for this node), to create an input record batch for the datafusion expression. A 2-way join was already implemented for binary expressions (using in arithmetic, for example).

This PR adds a new kind of data source called `MultiJoin` representing an arbitrary number of input expressions that must be joined, and having the convention that the resulting record batch columns will be named like "arg0", "arg1", ... "argn". The planner is changed to create this kind of join for as the data source to a scoped expr node that evaluates a function if any of the args have incompatible `DataScope`s. The scoped expr node now has logic to perform this `MultiJoin` when it encounters data source of this type. The mutli join logic is implemented in the `pipeline::expr::join` module, using the same join utilities we use for binary expressions.

* Closes #2634

Unit tests

Yes, users of transform processor can now pass to functions columns as arguments that in OTAP would have different row orders.

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
@@ -1198,6 +1198,11 @@ fn validate_attribute_assign_cardinality(
             validate_attribute_assign_cardinality(dest_attrs_id, dest_query_location, left)?;
             validate_attribute_assign_cardinality(dest_attrs_id, dest_query_location, right)?;
         }
+        LogicalExprDataSource::MultiJoin(children) => {
+            for child in children {
+                validate_attribute_assign_cardinality(dest_attrs_id, dest_query_location, child)?;
+            }
+        }
     }
 
     Ok(())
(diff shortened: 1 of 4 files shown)


lquerel (approved):
LGTM

PR #2636, opened 2026-04-12
[query-engine] Slice expression bug fix

CodeBlanch:
## Changes

* Switch `Slice` expression (arrays and strings) to use `Slice(source, start, [length])` instead of `Slice(source, start_inclusive, [end_exclusive])` to match KQL [substring](https://learn.microsoft.com/kusto/query/substring-function?view=azure-data-explorer) behavior.

/cc @albertlockett @drewrelmas


CodeBlanch:
@albertlockett It seems [your substring implementation](https://github.com/open-telemetry/otel-arrow/blob/5868ff15d7307129796bf35ba0f322a08a8d3586/rust/otap-dataflow/crates/query-engine/src/pipeline/functions/substring.rs#L95) was already treating `range_end_exclusive` as `length` so this change should be good for you too can you confirm?

albertlockett (approved):

PR #2650, opened 2026-04-14
feat: Add feature flag (crypto-symcrypt) to use symcrypt as the rustls crypto backend

AaronRM:
# Change Summary

* Add support for the `rustls-symcrypt` cryptography provider, enabling the [SymCrypt backend](https://github.com/microsoft/rustls-symcrypt?tab=readme-ov-file) as an option (off by default) for the rustls crypto provider if building for Windows or Linux. It updates the dependency lists, feature flags, and provider installation logic to allow selection of `rustls-symcrypt` as a mutually exclusive option alongside the existing providers (`ring`, `aws-lc-rs`, and `openssl`).
* macOS is excluded as `rustls-symcrypt` currently only supports Windows and Linux
* Update the `reqwest` dependency to explicitly disable default features.

Part of #2674 

* Confirmed that `test_tls_reload_integration` (which performs real TLS handshakes) passes when the `crypto-symcrypt` feature flag is enabled:

``` bash
cargo test -p otap-df-otap --test tls_reload --no-default-features --features "experimental-tls,test-utils,crypto-symcrypt" test_tls_reload_integration -- --nocapture
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.28s
warning: the following packages contain code that will be rejected by a future version of Rust: buf_redux v0.8.4, multipart v0.18.0
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running tests/tls_reload.rs (target/debug/deps/tls_reload-127615fe1f941564)

running 1 test
test tests::test_tls_reload_integration ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.03s
```

There are no user-facing changes by default. If a user would build with the `rustls-symcrypt` feature flag enabled, then SymCrypt would be used as the crypto backend.


jmacd (approved):

rust/otap-dataflow/crates/admin-api/src/http_backend.rs
```diff
@@ -506,7 +507,7 @@ fn ensure_crypto_provider() -> Result<(), Error> {
         )))]
         {
             Err(
-                "HTTPS admin connections require one of the admin SDK crypto features: \
+                "Admin client construction requires one of the admin SDK crypto features: \
```
lalitb, line 522: should admin-api crate also expose "crypto-symcrypt" just like otap crate?

rust/otap-dataflow/crates/otap/Cargo.toml
```diff
@@ -90,11 +90,15 @@ quiver = { workspace = true }
...
 
+[target.'cfg(any(target_os = "linux", target_os = "windows"))'.dependencies]
+rustls-symcrypt = { workspace = true, optional = true }
+
 [features]
 test-utils = ["rustls/ring", "otap-df-pdata/testing"]
 crypto-ring = ["rustls/ring", "tokio-rustls/ring"]
 crypto-aws-lc = ["rustls/aws_lc_rs", "tokio-rustls/aws-lc-rs"]
 crypto-openssl = ["dep:rustls-openssl", "tonic/tls-native-roots"]
+crypto-symcrypt = ["dep:rustls-symcrypt", "tonic/tls-native-roots"]
```
lalitb, line 101: `rusttls-symcrypt` is target-gated above, but this feature is still exposed globally. On macOS, users can still enable crypto-symcrypt and get an indirect build failure. We can  add `compile_error!` for non -Linux/non-Windows targets, likely alongside the other crypto feature guards in `rust/otap-dataflow/src/main.rs`.

lalitb (approved):
LGTM. Couple of comments, nothing blocking.

rust/otap-dataflow/crates/otap/Cargo.toml
AaronRM: Added a compile_error check for non-(Windows|Linux)

rust/otap-dataflow/crates/admin-api/src/http_backend.rs
AaronRM: Yes, added.

PR #2651, opened 2026-04-14
feat(temporal_reaggregation_processor): Use wakeup mechanism instead of timer ticks to trigger flushes

### LINKED ISSUE #2422: Metrics temporal reaggregation processor

JakeDern:
# Change Summary

This PR changes the timed flush mechanism from a timer tick to a scheduled wakeup. This has the advantage of allowing us to reset the timer whenever we flush data for some other reason, such as overflowing the max stream count. We can also now wait to schedule a wakeup until the time when we start aggregating a batch.

* Closes #2422

Unit.

No.


jmacd (approved):

JakeDern:
Test failures look flakey to me, ARM only and not in anything touched in this PR

lquerel (approved):
LGTM

What did you think of the Wakeup API? Do you see anything that could be improved?

JakeDern:
> LGTM
> 
> What did you think of the Wakeup API? Do you see anything that could be improved?

Wakeup API was quite nice! Mostly We were just missing some testing facilities that I wanted, but were no problem to add.

Two things that just came to mind in general while I was working with it:

- I think wakeup interactions have somewhat subtle correctness properties similar to acknack and I continue to think we need a better method for exercising these behaviors exhaustively to make sure there is no possibility we get confused or stuck inside a node with a pathological message sequence. 
- I don't love the idea of nodes computing the current time themselves to set their wakeup, I think we may want to consider providing an api for that to be able to control it for testing

Issue #2657, opened 2026-04-14
OPL use tagged string syntax for date time and regex literals.


albertlockett:
Currently we are using the syntax `datetime(1991-01-01)` for date times, but I got some feedback that having these kind of special functions that execute and validate at compile time is a programming language anti-pattern.

The suggestion, which I agree with, is to use a tagged string literal for these: `date_time"1991-01-01T00:00:00Z`. This should be parsed and validted at compile time.

This also gives us an opportunity to: have a similar syntax for regexs. e.g. `re"hell. world"`

At the same time we can make the decision for OPL to be more restrictive with the date formats it accepts. KQL supports a [ton of date formats](https://github.com/open-telemetry/otel-arrow/blob/1620e5346367e62e95a1bfd292e65abc09bb93c4/rust/experimental/query_engine/expressions/src/primitives/date_utils.rs#L10-L24) because it needs to maintain compatibility with the existing implementation, but OPL has no such restriction, so we can be more restrictive and just support something smaller such as ISO 8601

PR #2658, opened 2026-04-14
OPL parse datetimes as tagged literal containing ISO 8601 datetime

albertlockett:
# Change Summary

Change the parsing of datetime literals in OPL to use a tagged string literal that contains an ISO 8601 date time. For example: `date_time"1991-01-01T00:00:00-0400"`.

The motivation for the syntax change: We want something that is clearly validated at compile time. The old syntax where we were using an a function, seems like a function which are usually executed at runtime. It seemed strange to have this one datetime function be special evalutated at compile time.

We'll also use the same tagged string pattern for regex in a future PR.

The motivation behind the format change: I felt it was best to be restrictive in the date formats we support. KQL has to support many additional formats for compatibility with the existing implementation, but OPL does not have this restriction so we can be more specific about the format supported.

* Part of #2657

Unit tests

Yes - OPL programs in transform processor config with date literals will need to migrate the syntax from `datetime(1991-01-01)` to `date_time"1991-01-01T00:00:00"`.

### DIFF
--- a/rust/otap-dataflow/Cargo.toml
+++ b/rust/otap-dataflow/Cargo.toml
@@ -103,6 +103,7 @@ hashbrown = { version = "0.17", features = ["equivalent"] }
 http = "1.4"
 humantime = "2.3.0"
 humantime-serde = "1.1.1"
+iso8601 = "0.6.3"
 itoa = "1"
 itertools = "0.14.0"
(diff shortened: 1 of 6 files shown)


lquerel (approved):
LGTM

PR #2662, opened 2026-04-15
fix(temporal_reaggregation_processor): Only accept pdata when there are at least 3 outbound slots available

### LINKED ISSUE #2653: Temporal reaggregation processor should only accept pdata if there are 3 outbound slots available and not 2
### Pre-filing checklist

### Component(s)

### Bug Description

The temporal reaggregation processor `accept_pdata` implementation gates on having one inbound and two outbound slots available, the maximum required to accommodate an incoming pdata.

This does not account for, however, the outbound slot required by the processor when a flush signal comes in. So, we can have the following situation:

1. Accept pdata with exactly two slots left
2. Allocate an outbound slot for non aggregable data to be passed through
3. Overflow the existing aggregating batch (via stream count or ID overflow), thereby triggering a flush and using a second outbound slot
4. Place the incoming data in the pending buffer after flushing
5. Have a timer tick/wakeup message come in while the outbound slots are full which causes the processor to crash. 

When step (5) happens we could choose to nack the data, but applying backpressure through the engines queues is the generally better solution than getting ourselves into a situation where we have to nack.

The solution is to change this to requiring three outbound slots available before accepting pdata.

### Steps to Reproduce

-

### Expected Behavior

-

### Actual Behavior

-

### OTel-Arrow Version

-

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

JakeDern:
# Change Summary

Fixes the attached issue by bumping the required slots to 3 instead of 2.

* Closes #2653

Unit and [something else](https://github.com/JakeDern/otel-arrow/blob/metrics-trag-processor-spec-3/rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/spec/acknack3.qnt#L551-L559)?

Minimum outbound slots is now 3.


lquerel (approved):
LGTM

PR #2666, opened 2026-04-15
OPL Parser add support for regex literal as tagged string 

### LINKED ISSUE #2657: OPL use tagged string syntax for date time and regex literals.

albertlockett:
# Change Summary

Adds support to parse a regex literal as a tagged string. The tag is `r`. For example:

```kql
logs | where matches(event_name, r"^err.*")
```

The difference between this and simply defining the regex as a string (e.g. like `matches(event_name, "^err.*")`) is that the parser will validate the the regex while parsing, versus lazily converting it to a pipeline later as the pipeline is constructed.

This code slightly refactors the grammar of OPL. Before we had a rule for parsing datetime_expression specifically, but now we just have a for tagged literals and both regexs and datetimes are parsed using this rule. The organization of the parser code is changed accordingly.

* Closes #2657

Unit tests

Yes - this new regex literal syntax would be available for users who write OPL prorgrams when configuring the transform processor.


rust/otap-dataflow/crates/opl/src/parser/expression.rs
```diff
@@ -663,6 +662,54 @@ fn parse_primitive_expression(
...
+            let regex = Regex::new(pattern).map_err(|e| {
+                ParserError::SyntaxError(
+                    query_location.clone(),
+                    format!("Invalid regex literal '{pattern}': {e}"),
+                )
+            })?;
+            StaticScalarExpression::Regex(RegexScalarExpression::new(query_location, regex))
+        }
+        _unknown_tag => {
+            todo!()
```
lquerel: Not a big deal for this PR, but it's very easy to make our pipeline panic with an unrecognized tab.

lquerel (approved):
LGTM.
I'm not completely sure, but it seems that with this PR, the `matches` function accepts both an unchecked string and a string tagged with `r"..."`. Shouldn't we, in order to fully benefit from this typing mechanism, accept only strings with the proper tags for these functions? It's definitely something we could revisit in a future PR.

rust/otap-dataflow/crates/opl/src/parser/expression.rs
albertlockett: oops yeah I definitely meant to handle this. fixed in [90e6a72](https://github.com/open-telemetry/otel-arrow/pull/2666/commits/90e6a72c59ce68b2ddbfaca3bd32bf45e046aa25)

PR #2669, opened 2026-04-15
feat(otap-dataflow): Implement `processor_chain:inlined` for channel elimination and composite metric reporting

### LINKED ISSUE #2576: Use `NodeUserConfig` for `ProcessorChain` `processor` configs

### LINKED ISSUE #2577: Add processor_chain config switches to enable/disable sub-processor metric reporting
A user of the `processor_chain` feature may want to disable separate metric reporting for sub-processors (i.e. only track the chain's `compute.success.duration`, not individual node datapoints).

### LINKED ISSUE #2579: Initial implementation of `processor_chain`

drewrelmas:
# Change Summary

## Motivation

`ProcessorChain` runs multiple sub-processors sequentially within a single task, eliminating inter-processor channels. This gives per-batch composite timing with correct min/max/sum/count distribution — something impossible with separate processors where `min(A) + min(B) != min(A+B)`.

The initial implementation (#2569) used Vec-backed buffer `EffectHandler`s. Design review identified two issues: `is_chainable()` was too weak (a boolean that could silently break), and expanding stages needed structural prevention. This PR replaces the Vec-based approach with a narrow `InlineProcessor` trait whose shape makes both problems impossible to express.

### `InlineProcessor` trait

```rust
pub enum InlineOutput<PData> { Forward(PData), Drop }

pub trait InlineProcessor<PData> {
    fn process_inline(&mut self, data: PData) -> Result<InlineOutput<PData>, Error>;
    fn compute_duration(&self) -> Option<&ComputeDuration> { None }
    fn on_config(&mut self, _config: serde_json::Value) {}
    fn collect_telemetry(&mut self, _reporter: &mut MetricsReporter) {}
}
```

Structurally prevents unsafe-for-chain operations: no multi-output (caps at 0 or 1), no `EffectHandler` (no fan-out/timers), synchronous only.

### Factory-level chainability

`ProcessorFactory.create_inline` replaces `is_chainable()`. Non-inline processors are rejected at config validation time.

### `ProcessorChainNode`

Records a composite `ComputeDuration` covering the full sequence. Sub-processor timing is handled internally via `ComputeDuration::timed()`. Uses `consume_budget().await` between stages for cooperative scheduling — only yields when Tokio's task budget (128) is exhausted, paying ~1-2ns per stage instead of the old ~1µs `yield_now()`.

### Sub-stage telemetry

`enable_sub_processor_telemetry: true` (default `false`) controls whether sub-processor metrics are reported. Sub-processors get scoped `node.id` (e.g., `chain/insert_B`).

### NACK support

The chain clones original PData for NACK when `has_frames()` is true. On error, sends NACK via `route_nack` directly (not `notify_nack`, which requires concrete `OtapPdata`).

### Producer metrics via `PrepareSourceSend`

Standalone processors call `send_message_with_source_node()` to stamp source metadata for producer metrics — but that's only implemented for concrete `OtapPdata`, not generic `PData`. A new `PrepareSourceSend` trait in the engine crate (mirroring `ReceivedAtNode`) bridges the gap. The chain calls `prepare_source_send()` + `router.send_default_stamped()` so `produced.success` is correctly recorded. Adds `StampOutputPort + PrepareSourceSend` bounds that ripple to `PipelineFactory`, `HyperEdgeWiring`, and controller validation.

## Config syntax

```yaml
chain:
  type: processor_chain:inlined
  config:
    enable_sub_processor_telemetry: true
    processors:
      insert_B:
        type: processor:attribute
        config:
          actions: [{ action: insert, key: step_1, value: B }]
      insert_C:
        type: processor:attribute
        config:
          actions: [{ action: insert, key: step_2, value: C }]
```

## Performance

Side-by-side using `configs/fake-classic-and-chain-noop.yaml` — 3 attribute processors, 100-log batches at 1000/s.

### Telemetry — release build (700 batches of 100 logs at 1000/s)

| Node | consumed | produced | compute avg |
|---|---|---|---|
| **chained/receiver** | — | 700 | — |
| **chained/chain** | 700 | 700 | 330µs |
| chain/insert_A | — | — | 60µs |
| chain/insert_B | — | — | 40µs |
| chain/insert_C | — | — | 40µs |
| **chained/noop** | 700 | — | — |
| **classic/receiver** | — | 700 | — |
| **classic/insert_A** | 700 | 700 | 90µs |
| **classic/insert_B** | 700 | 700 | 40µs |
| **classic/insert_C** | 700 | 700 | 40µs |
| **classic/noop** | 700 | — | — |

The composite chain duration (330µs) vs sub-processor sum (140µs) has a ~190µs gap. This is explained by deserialization: the first `process_inline` converts `OtlpBytes → OtapArrowRecords` (~120µs per the `pdata_deserialization` benchmark for 100 logs). Stages 2 and 3 receive `OtapArrowRecords` already, so their `try_into()` is a no-op enum unwrap. The remaining ~70µs is `send_message`, `prepare_source_send`, and chain bookkeeping. The classic pipeline has the same deserialization cost in `insert_A` but it's invisible — `effect_handler.timed()` wraps only the core transform, and there's no wrapping timer at the task-loop level. See the added deserialization benchmark that supports this theory.

### Memory — release build

| Metric | Classic | Chained | Diff |
|---|---|---|---|
| memory.usage (bytes) | 4,039,600 | 3,007,240 | **-26%** |
| memory.allocated (bytes) | 1,257,057,520 | 1,224,277,936 | -3% |
| Channel receivers | 9 | 5 | -4 |

### Microbenchmarks

Microbenchmarks use `String` PData with busy-spin work — no Arrow deserialization — isolating pure chain vs channel mechanics.

**~100µs work/processor** (1,000 batches):

| N | Classic (ms) | Chained (ms) | Diff |
|---|---|---|---|
| 1 | 101.6 | 101.5 | -0.1% |
| 2 | 201.9 | 202.2 | +0.1% |
| 3 | 303.3 | 305.2 | +0.6% |
| 10 | 1,012.2 | 1,011.5 | -0.1% |

**~100ns work/processor** (1,000 batches):

| N | Classic (µs) | Chained (µs) | Diff |
|---|---|---|---|
| 1 | 208 | 224 | +7% |
| 2 | 449 | 414 | **-8%** |
| 3 | 674 | 619 | **-8%** |
| 10 | 2,180 | 2,134 | **-2%** |

Chain beats separate at N>=2 by eliminating ~200ns/stage channel overhead. `consume_budget()` costs ~1-2ns (thread-local check) vs the old ~1µs `yield_now()`. The fact that these benchmarks show zero chain overhead confirms that the ~190µs gap in the telemetry section is not from chain mechanics — it's from `OtlpBytes → OtapArrowRecords` deserialization, which only exists with real `OtapPdata`.

### Deserialization benchmark (`pdata_deserialization`)

Isolated `OtlpBytes → OtapArrowRecords` conversion (the `try_into()` call inside each processor, outside the `timed()` block):

| Batch size | Time |
|---|---|
| 10 logs | ~56µs |
| 100 logs | ~115µs |
| 1000 logs | ~1.0ms |

The 100-log result (~115µs) directly accounts for most of the ~190µs telemetry gap. Only the first processor in the chain pays this cost — subsequent stages receive `OtapArrowRecords` already (no-op enum unwrap). The classic pipeline has the same cost in `insert_A` but no metric captures it because it happens outside of the `timed()` scope.

Relevant code in `attributes_processor`: https://github.com/open-telemetry/otel-arrow/blob/ac4a60e9875c995eafa1f4a9e8f54c7d9ef10b3e/rust/otap-dataflow/crates/core-nodes/src/processors/attributes_processor/mod.rs#L360-L377

## Follow-ups

- **Transform processor inlining**: Blocked by async pipeline execution and multi-output routing. Pure filter/project/set transforms without `route_to` could be supported.
- **`processor_chain:channeled`**: A variant using channels for processors with expansion behavior. The `:inlined` scope leaves the door open.

- Contributes to #2556
- Closes #2576, #2577, #2579

Unit tests, benchmarks, local runs of sample config

Yes, users can now configure `processor_chain:inlined` nodes


jmacd:
>  the chain's yield_now() between stages dominates (~700-800ns per yield vs 100ns of actual work)

Makes me want to avoid `yield_now()`, or maybe make the call conditional on cpu time.

drewrelmas:
> > the chain's yield_now() between stages dominates (~700-800ns per yield vs 100ns of actual work)
> 
> Makes me want to avoid `yield_now()`, or maybe make the call conditional on cpu time.

I happened upon https://docs.rs/tokio/latest/tokio/task/coop/fn.consume_budget.html which seems to help avoid wasted yields. Have updated the PR and description with new benchmark numbers that show reduction of the overhead seen when benchmarking classic vs chained on low-work scenarios.

drewrelmas:
One significant drawback of this approach with `process_inline()` is requiring each participating processor to have a largely duplicate internal implementation. I am still iterating locally to see if I land at something better.

I am thinking of actually bringing back the 'LocalSender' effect handler approach from #2569 to help here.

gouslu:
Two design-level thoughts:

1. **Processor chaining config.** I wonder if `processor_chain:inlined` would be better as an engine-internal optimization rather than a user-facing topology primitive. If the motivation is composite metrics, maybe we could express that intent directly with a separate grouping mechanism at the config level, which is different than processor chaining in my opinion.

2. **Dual implementation surface.** With every inline-capable processor now maintaining both `Processor::process` and `InlineProcessor::process_inline`, I'm a bit concerned about the long-term maintenance cost as the two paths could drift apart. It also seems like if a processor can implement process_inline and process both, why not just do process_inline, it seems like it probably doesn't need the full process trait. Engine can still provide all the benefits of a regular processor for inline only processors by wrapping the effect_handler, control_channel, metrics etc. concerns.

To be clear, the PR solves a real problem. Correct composite timing across a processor sequence isn't possible with separate per-node metrics, and the benchmarks are solid. I think that this PR is not far off from what I'm thinking, which is whether we can get to the same outcome through a simple/complex processor split rather than explicit chaining config.

In other words, if what we need is composite metrics, let's find a way to group them and emit them together from any processor layout. And if what we need is performance gains, then address it separately by providing a separate interface for inline processors only, if you also agree. It could still work with the existing internals and all that, we just need a "common processor" defined by the engine that wraps inline processors, which later can get selectively (based on pipeline config) chained by the engine during wiring.

drewrelmas:
I closed this PR in favor of #2747 which implements a much smaller-scoped solution for composite metrics. The `processor_chain` design is a big item which requires some more thought.

PR #2675, opened 2026-04-15
run validation ci manually

c1ly:
I am seeing the validation tests acting up again in the validation ci job, setting the ci job to be a manual trigger instead of automatically running on every PR


.github/workflows/rust-validation-tests.yml
```diff
@@ -14,14 +14,7 @@ permissions:
...
 on:
-  push:
-    branches:
-      - main
-  pull_request:
-    branches:
-      - main
-  merge_group:
-  # workflow_dispatch:
+  workflow_dispatch:
```
jmacd, line 17: workflow_dispatch was commented before. I take it this becomes required as we remove the three fields above?

jmacd (approved):

.github/workflows/rust-validation-tests.yml
c1ly: that is correct

PR #2683, opened 2026-04-15
fix: Pass through generation_strategy to load generator for benchmarks 

### LINKED ISSUE #2681: Batch processor nightly benchmarks should be using fresh data generation and not pre_generated
### Pre-filing checklist

### Component(s)

### Bug Description

Batch processor nightly benchmarks use pre_generated data from the fake_data_generator and not fresh data which makes the compression rate look way better than it would in real life.

I attempted to set this, but there is a parameter drilling disconnect between `loadgen/config.yaml.jq` and `df-loadgen-steps-docker.yaml` that needs to be corrected

### Steps to Reproduce

Run the batch processor nightly benchmarks for some signal along with a baseline

### Expected Behavior

Observe a slightly better compression rate with a batch processor vs without

### Actual Behavior

Observe an insane compression rate of 90% which is basically equal to `1 - (records_per_batch / total_records)`.

### OTel-Arrow Version

NA

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

JakeDern:
# Change Summary

This fixes the disconnect in the attached issue by adding `generation_strategy` config to `df-loadgen-steps-docker.yaml`.

* Closes #2681

Results before:
image

Results after:

image

No

### DIFF
--- a/.github/workflows/rust-validation-tests.yml
+++ b/.github/workflows/rust-validation-tests.yml
@@ -14,14 +14,7 @@ permissions:
   contents: read
 
 on:
-  push:
-    branches:
-      - main
-  pull_request:
-    branches:
-      - main
-  merge_group:
-  # workflow_dispatch:
+  workflow_dispatch:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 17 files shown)


lquerel (approved):
Thanks

cijothomas (approved):

JakeDern:
Looks like arm is still flakey

albertlockett (approved):

PR #2685, opened 2026-04-16
fix(fake_data_generator): Make signal production metrics more reliable by not cancelling send futures when there is backpressure

### LINKED ISSUE #2678: fake_data_generator metric counts can be incorrect for a variety of reasons
### Pre-filing checklist

### Component(s)

### Bug Description

The fake data generator cancels futures when a control message comes in and can frequently throw away metrics if downstream latency is high. There are better ways to handle the logic.

Also, there are minor problems where the metric calculations for the fake_data_receiver can be incorrect due to:

- Partial failures where a send is split into multiple batches (unlikely in practice but technically feasible)
- Rounding errors which can cause over reporting in generated metrics

These undermine exact accounting.

### Steps to Reproduce

1. Run any OTAP->BATCH->OTAP benchmark scenario
2. Observe the metrics

### Expected Behavior

Signal counts are consistent across the loadgen, engine, and backend instances

### Actual Behavior

The fake data generator in the loadgen instance reports 0.

### OTel-Arrow Version

All

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

JakeDern:
# Change Summary

This PR tries to make signal production metrics more reliable under backpressure. Previously any time there was a control message we would cancel the send future. This could happen a lot of there was backpressure in a benchmark scenario and in some cases, we just report 0 produced signals all the time even if we are sending some but not all of them on each cycle.

This PR creates the future up front then polls a reference in the select statement so that it's not dropped if a control message comes in before it completes. This means that every time we generate a batch of data to send, we will poll until we send all of it and then update the metrics.

This is not perfect because we won't update the metrics until after all the sends complete, at least we don't report 0 and the total reported count is accurate. I would rather increment the metrics at the point of sending, but that would be a larger refactoring due to borrow issues.

As listed in the attached issue, there is also a little bit of margin of error in terms of reported metrics counts, this PR also addresses that by counting the items we export and propagating that back up the stack, so we report exactly what was successfully exported. Note that we don't really have to worry about partial failures due to sending being infallible unless the channel is closed.

* Closes #2678

Running locally, here is what the metrics look like after the fix:

image

This is what it looked like before:

image

No.


lquerel (approved):
Thank you for this fix!

PR #2694, opened 2026-04-17
fix(dataflow): isolate blocked routes in content and signal type routers

### LINKED ISSUE #2693: Exclusive routers can stall healthy routes when one selected output is blocked
### Pre-filing checklist

### Component(s)

### Bug Description

`content_router` can lose liveness when one selected output route becomes blocked.

After the router resolves a message to a matched route or `default_output`, it currently awaits forwarding to that selected output. If that route is full or no longer making progress, the router task blocks on that one send.

Because the router processes messages serially, traffic for unrelated healthy routes behind the same router can also stop making progress. This turns a route-local failure into a router-wide liveness problem.

### Steps to Reproduce

  1. Configure a pipeline with `content_router` and multiple outputs, for example
     one route per tenant plus an optional `default_output`.
  2. Connect one route to a downstream pipeline that can be stopped or allowed to
     fill up.
  3. Keep another route connected to a healthy downstream pipeline.
  4. Send traffic that selects both routes.
  5. Stop the downstream consumer for one selected route, or let that route fill.
  6. Continue sending traffic for both the blocked route and the healthy route.

### Expected Behavior

  When the selected route is blocked, only that message should be refused.

  - The blocked-route message should be rejected explicitly.
  - The router should remain live.
  - Traffic for other healthy routes should continue to flow.

### Actual Behavior

  Once messages start selecting the blocked route, the router can block on that
  selected output.

  - Traffic for the blocked route stops.
  - Traffic for unrelated healthy routes behind the same router can also stop.
  - The router no longer provides route-level isolation under backpressure or
    downstream shutdown.

### OTel-Arrow Version

495588ee584201ea956c15c6fc102dc3465de675

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

lquerel:
# Change Summary

This PR fixes a router liveness bug identified during the live reconfiguration and shutdown work.

The issue was that `content_router` could block the whole router task when one selected output route became blocked, which prevented unrelated healthy routes from continuing to make progress. While addressing that bug, this PR also applies the same selected-route admission model to `signal_type_router`, since it has the same exclusive-routing shape: each input message selects at most one output route.

For both `content_router` and `signal_type_router`:

- selected matched/named/default routes now use non-blocking admission
- `Closed` selected routes produce an immediate retryable route-local NACK with `NackCause::RouteClosed`
- `Full` selected routes follow `admission_policy.on_full`
- the default `reject_immediately` policy emits an immediate retryable route-local NACK with `NackCause::RouteFull`
- the optional `backpressure` policy parks at most one message per blocked selectable route, keeps healthy routes flowing, and pauses pdata admission only when all selectable routes are currently blocked
- the processor returns `Ok(())` for route-local rejection paths, so one blocked route does not fail or stall the whole router task
- router-owned parked work is retryable-NACKed with `NackCause::NodeShutdown` when shutdown starts

This PR also:

- adds engine-level selected-route admission outcomes used by non-blocking sends
- adds a shared core-nodes exclusive-router scheduler for the two exclusive routers
- adds route-full and route-closed rejection telemetry
- adds regression coverage for route `Full` / `Closed` behavior, route isolation, backpressure, shutdown, and Ack/Nack propagation behavior
- documents processor classes in `docs/processors.md` and the router-specific guarantees in the `content_router` and `signal_type_router` READMEs

* Closes #2693

- `cargo xtask check`

Yes.

Operators now get explicit retryable route-local NACKs when the selected route in `content_router` or `signal_type_router` is unavailable (`Closed`) or rejected by the configured full-route admission policy.

This PR also adds optional router-local configuration:

```yaml
  admission_policy:
    on_full: reject_immediately # or "backpressure"
```

reject_immediately is the default. backpressure allows a router to tolerate temporarily slow routes by parking one message per blocked route while unrelated healthy routes continue to flow.


rust/otap-dataflow/crates/engine/src/control.rs
```diff
@@ -164,6 +164,20 @@ impl<PData> AckMsg<PData> {
     }
 }
 
+/// The NACK message.
```
jmacd:
```suggestion
/// The NACK message cause.
```

rust/otap-dataflow/docs/processors.md
```diff
@@ -0,0 +1,124 @@
...
+- `stateful`: processing depends on local state retained across calls.
+- `ack-aware`: the processor subscribes to downstream Ack/Nack outcomes.
+- `wakeup-driven`: the processor uses local wakeups to resume deferred work.
+- `admission-gated`: the processor can return `accept_pdata() == false`.
+- `buffering`: the processor retains data locally before forwarding.
+- `batching`: the processor combines or splits data by size or time.
+- `retrying`: the processor reschedules failed delivery attempts.
+- `may-drop`: the processor can intentionally drop input data.
+- `must-explicitly-ack-or-nack-deferred-work`: the processor owns deferred work
+  and must resolve it explicitly during success, failure, or shutdown paths.
```
jmacd, line 98: Looks good.

jmacd (approved):

rust/otap-dataflow/docs/processors.md
lquerel: Yes, I'm trying to establish a classification of processors that we could later use as a basis for automated checks, in order to validate that the properties and guarantees associated with each processor class are properly satisfied and enforced.

PR #2696, opened 2026-04-17
feat(config): add yaml: and http: URI providers

### LINKED ISSUE #2598: feat(config): Phase 2 config URI providers — yaml: and http:
### Pre-filing checklist

### Component(s)

Rust OTAP dataflow (`rust/otap-dataflow/`)

This is the planned Phase 2 follow-up to #2514 (implemented in #2529). Phase 1 delivered the `ConfigProvider` trait abstraction along with `file:` and `env:` providers. The original issue explicitly deferred these additional providers to a follow-up once the abstraction was in place.

### Proposed Solution

Extend the provider abstraction introduced in #2529 with two additional providers and improved fallback behaviour:

#### New providers

| Scheme | Description | Complexity |
|--------|-------------|------------|
| `yaml:` | Inline YAML supplied directly on the CLI | Low |
| `http:` | Fetch config from an HTTP URL (unauthenticated GET) | Medium |

---

#### `yaml:` provider

Follows the [OTel Collector `yaml:` convention](https://opentelemetry.io/docs/collector/configuration/#yaml-provider) — the content after `yaml:` is treated as a literal YAML string, with `::` used as a key path separator for inline overrides:

```bash
# Full inline YAML config
df_engine --config=yaml:version::otel_dataflow/v1

# Used to override a specific field from a base config (once merge support lands)
df_engine --config=file:/etc/base.yaml --config=yaml:engine::{}
```

---

#### `http:` provider

Performs an unauthenticated HTTP GET to fetch the config content. The response body is treated as YAML (or JSON if the `Content-Type` is `application/json`). A configurable timeout should be applied (default: 30s). Standard HTTP redirects are followed.

```bash
df_engine --config=http://config.internal:8080/pipeline.yaml
```

`https:`, authentication support (Bearer token, mTLS), and multi-config merge are intentionally deferred to Phase 3.

---

### Acceptance Criteria

- [ ] `--config=yaml:<inline-yaml>` parses inline YAML using `::` key path separator (OTel Collector convention)
- [ ] `--config=http://...` fetches config via unauthenticated HTTP GET with 30s default timeout
- [ ] HTTP: follows redirects, detects JSON vs YAML from `Content-Type`
- [ ] Helpful error message listing searched paths when all fallbacks fail
- [ ] Unit tests for both new providers
- [ ] `parse_scheme` updated to handle `http://` (note `://` vs `:` separator)

### Alternatives Considered

- `https:` support was considered but deferred to Phase 3 alongside authentication (Bearer, mTLS) since the two concerns go together and add TLS backend complexity (cert validation, CA bundles, feature flags).
- Multi-config merge (`--config` specified multiple times with deep merge semantics) is also deferred to Phase 3.

### Additional Context

- Phase 1 tracking issue: #2514
- Phase 1 implementation: #2529
- OTel Collector `yaml:` provider docs: https://opentelemetry.io/docs/collector/configuration/#yaml-provider
- OTel Collector multiple config docs: https://opentelemetry.io/docs/collector/configuration/#multiple-config-files

mateenali66:
Fixes #2598. Phase 2 follow-up to #2529 (which added the ConfigProvider trait and the `file:`/`env:` providers).

## What this adds

**`yaml:` provider** — inline YAML on the CLI following the OTel Collector convention. `::` expands into nested YAML, and the segment after the final `::` is treated as a trailing YAML fragment (typically `key: value`). No `::` means the content is passed through as literal YAML.

```
yaml:version: otel_dataflow/v1            ->  version: otel_dataflow/v1
yaml:exporters::debug::verbosity: detailed ->  exporters:
                                                 debug:
                                                   verbosity: detailed
yaml:engine::{}                            ->  engine:
                                                 {}
```

**`http:` provider** — unauthenticated HTTP GET with a 30s timeout. Response body is treated as YAML unless `Content-Type` is `application/json`. Standard HTTP redirects are followed. Uses `reqwest::blocking` so config resolution can happen at startup without a tokio runtime.

## Out of scope (deferred to Phase 3 per the issue)

- `https:` and TLS configuration
- Bearer / mTLS authentication
- Configurable timeouts
- Multi-config merge semantics (which is where the `::` inline-override example from the issue body really lands)

## Tests

- `yaml:` provider: literal content, single-level key path, nested key path, flow-style value, round-trip through `default_resolver` and `serde_yaml`
- `http:` provider: yaml body, JSON content-type detection, non-2xx status error. Uses `wiremock` as an async mock server with `tokio::task::spawn_blocking` to drive the blocking reqwest client.
- `parse_scheme` already accepts `http://` from Phase 1 (test is in #2529). Added a `yaml:` case alongside.

## Validation

`cargo test -p otap-df-config` -> 273 tests pass.
`cargo clippy -p otap-df-config --all-targets -- -D warnings` -> clean.
`cargo check -p otap-df-engine -p otap-df-controller` -> clean (config crate still builds for its downstream consumers).

`cargo xtask quick-check` fails on an unrelated pre-existing `df_engine` main.rs issue: `static GLOBAL: Jemalloc = Jemalloc;` is used without the same `not(clippy)` cfg guard that gates the `use tikv_jemallocator::Jemalloc;` import. Reproduces on upstream main with this branch stashed, so not caused by this PR. Happy to file a separate issue if useful.

## Note on rustls

The workspace's `reqwest` uses the `rustls-no-provider` feature, so callers must install a crypto provider before any `reqwest::blocking::Client` is built. Production code already handles this via `otap-df-otap/crypto-ring`. The config crate's http tests install `rustls::crypto::ring::default_provider()` lazily via a `std::sync::Once`, and pull in `rustls` with the `ring` feature only as a dev-dependency.


lquerel:
That was fast!

Can you take into consideration the following comment https://github.com/open-telemetry/otel-arrow/issues/2598#issuecomment-4271067645 Thanks

mateenali66:
> That was fast!
> 
> Can you take into consideration the following comment [#2598 (comment)](https://github.com/open-telemetry/otel-arrow/issues/2598#issuecomment-4271067645) Thanks

Thanks @lquerel , faster than the backoff will be. I was already prepared and was looking forward to triage.

exp backoff 500ms doubling to a 60s cap, retries on network errors and any non-2xx, configurable via `HttpConfigProvider::with_max_attempts(n)`, default u64::MAX.

rust/otap-dataflow/crates/config/src/config_provider.rs
```diff
@@ -104,6 +110,193 @@ impl ConfigProvider for EnvConfigProvider {
...
+    }
+
+    /// Build a provider with an explicit cap on retry attempts. `max_attempts`
+    /// counts the first try as an attempt, so `1` disables retries entirely.
+    #[must_use]
+    pub fn with_max_attempts(max_attempts: u64) -> Self {
+        let client = reqwest::blocking::Client::builder()
+            .timeout(Duration::from_secs(30))
+            .build()
+            .expect("reqwest blocking client should always build");
```
lquerel: It is preferable to use `expect` only for errors that are not expected. In this specific case, this is user input, so we should use the error-handling mechanism (`Result`) in order to provide a unified error presentation.

rust/otap-dataflow/crates/config/src/config_provider.rs
```diff
@@ -104,6 +110,193 @@ impl ConfigProvider for EnvConfigProvider {
...
+    /// Sleep duration before the next attempt, doubling until capped.
+    fn backoff_for(attempt: u64) -> Duration {
+        // Cap the shift at 20 so the multiplier stays well below u32::MAX
+        // before we clamp to MAX_BACKOFF below.
+        let shift = attempt.min(20) as u32;
+        let multiplier = 1u64 << shift;
+        let scaled = Self::BASE_BACKOFF
+            .checked_mul(multiplier.try_into().unwrap_or(u32::MAX))
+            .unwrap_or(Self::MAX_BACKOFF);
+        scaled.min(Self::MAX_BACKOFF)
```
lquerel: Could be simplified as follow:

```rust
    let shift = attempt.min(20) as u32;
    let multiplier = 1u32 << shift;

    Self::BASE_BACKOFF
        .checked_mul(multiplier)
        .unwrap_or(Self::MAX_BACKOFF)
        .min(Self::MAX_BACKOFF)
```

Explanation:
- `multiplier.try_into().unwrap_or(u32::MAX)` is unnecessary in practice because 1u64 << 20 is only 1_048_576, which always fits in u32
- the comment about staying below u32::MAX is true, but misleading, because with a cap of 20 you are not even close to the limit
- the saturation behavior depends on `checked_mul`, which is fine, but the code is more complicated than needed

rust/otap-dataflow/crates/config/src/config_provider.rs
```diff
@@ -221,6 +417,16 @@ mod tests {
...
+    use std::sync::Once;
+
+    /// Install a rustls crypto provider so reqwest can build a blocking client.
+    /// Production code installs this at process startup; tests do it lazily.
+    static CRYPTO_INIT: Once = Once::new();
+    fn ensure_crypto_provider() {
+        CRYPTO_INIT.call_once(|| {
+            let _ = rustls::crypto::ring::default_provider().install_default();
+        });
+    }
```
lquerel, line 429: We already have a similar initialization in `main`. Is it necessary to declare it again here?

rust/otap-dataflow/crates/config/src/error.rs
```diff
@@ -233,13 +233,27 @@ pub enum Error {
...
+    )]
     #[diagnostic(code(data_plane::config_uri_unknown_scheme), url(docsrs))]
     ConfigUriUnknownScheme {
         /// The unrecognized scheme prefix.
         scheme: String,
     },
 
+    /// An http: config URI could not be fetched.
+    #[error(
+        "Could not fetch config from '{uri}': {details}\n\nHint: Check the URL is reachable and responds within 30s. `https://` and authenticated endpoints are not supported yet."
```
lquerel: The error message is no longer accurate with the added retry mechanism

lquerel (changes requested):
Thank you for these quick contributions. I just noted a few minor issues that need to be addressed before merging.

rust/otap-dataflow/crates/config/src/config_provider.rs
mateenali66: dropped the `expect`. `HttpConfigProvider::new` and `with_max_attempts` now return `Result<Self, Error>` with a new `ConfigHttpClientBuildFailed` variant, and `default_resolver` propagates that through.

rust/otap-dataflow/crates/config/src/config_provider.rs
mateenali66: applied your suggestion. also dropped the stale comment about `u32::MAX` since the cap at 20 makes it irrelevant.

rust/otap-dataflow/crates/config/src/config_provider.rs
mateenali66: kept the local helper because `otap_df_otap::crypto` isn't reachable here: `otap-df-otap` already depends on `otap-df-config`, so pulling it in as a dev-dep would create a workspace cycle. Happy to extract the init into a shared test-utility crate if you'd prefer, but that's a separate refactor.

rust/otap-dataflow/crates/config/src/error.rs
mateenali66: rewrote the message to describe the retry behavior instead of the "30s" claim, and split off a `ConfigHttpClientBuildFailed` variant for the construction failure path.

mateenali66:
@lquerel ping when you have a moment. pushed the feedback fixes in `6de5aa4`:

- dropped the `.expect(...)` on the reqwest client; `HttpConfigProvider::new` and `with_max_attempts` now return `Result<Self, Error>` with a new `ConfigHttpClientBuildFailed` variant, threaded through `default_resolver`
- replaced the manual `checked_shl` pattern with your suggested `shift.min(20)` + `BASE.checked_mul` chain
- rewrote the stale timeout error message and split the client-build failure into its own error variant
- kept the local rustls crypto init helper because moving it out would create a workspace cycle (`otap-df-otap` already depends on `otap-df-config`); happy to extract into a shared test utility crate if you'd prefer

the `build_nonrequired (otap-dataflow, ubuntu-24.04-arm)` failure is unrelated to the PR (non-required CI, all required builds and clippy pass on arm64). would appreciate a re-review when you get a slot.

mateenali66:
@lquerel could you take another pass when you have a slot? pushed the requested fixes in `6de5aa4` on apr 23 (return `Result` from `HttpConfigProvider::new`/`with_max_attempts` with a new `ConfigHttpClientBuildFailed` variant, exp backoff with `shift.min(20)` + `checked_mul`, reworked timeout error). all required CI is green, the only red check is `build_nonrequired (ubuntu-24.04-arm)` which is unrelated runner flake. stale bot pinged the PR yesterday so figured i'd nudge before it auto-closes.

jmacd:
@mateenali66 I've looked over your PR. Thank you! I tasked an agent with a first pass review, and it leads me to think we should separate the YAML from the HTTP factories, that would make it easier to move forward. Here's the raw output from my agent:

## Summary

Cleanly structured PR. The `yaml:` provider is straightforward and largely
correct. The `http:` provider, however, contains **scope creep with a serious
UX regression** that should be addressed before merging.

## Blocking concerns

1. **Infinite-retry-by-default is a behavioral regression not requested by the
   issue.** `HttpConfigProvider::new()` defaults to `max_attempts = u64::MAX`
   with no logging during retry. Issue #2598 does **not** ask for retries — the
   OTel Collector's `http:` confmap provider also does not retry. Concrete
   failure mode: user types `http://confg.internal/...` (typo); the CLI hangs
   forever, capped at 60s sleeps, completely silent. There is no Ctrl-C friendly
   behavior beyond signal-interruptible sleep, no diagnostic, no progress.
   - Recommended: default to **1 attempt** (fail-fast) and let supervisors /
     restart policies handle the "wait for server" use case. If retries are
     kept, default to a small bounded count (e.g. 3) and emit a warning log per
     failure.
   - Alternatively: drop the retry loop entirely from this PR; it is unrequested
     scope and can be added later with proper logging/observability and tests
     for cancellation.

2. **Issue acceptance criteria say "configurable timeout (default: 30s)".** PR
   hard-codes 30s with no setter. The PR description lists "Configurable
   timeouts" as deferred, but the issue lists it as in-scope. Worth either
   honoring it or getting explicit sign-off to defer.

3. **`default_resolver()` now returns `Result` purely because of HTTP client
   construction**, even when the user is only resolving `file:` / `env:` /
   `yaml:` URIs. Tests had to add `ensure_crypto_provider()` to unrelated tests
   (`bare_path_treated_as_file`, `unknown_scheme_returns_error`,
   `yaml_provider_parses_as_yaml`) — strong smell. Suggested fix: lazily
   construct the reqwest client inside `HttpConfigProvider::resolve` (or via
   `OnceLock`) so `default_resolver()` stays infallible and only HTTP URIs need
   a crypto provider. This is a noticeable API improvement vs. the current `?`
   propagation everywhere.

## Correctness / non-blocking

4. **`expand_key_path` doesn't validate the resulting YAML.** `yaml:::value`
   produces `:\n  value` (invalid). `yaml:foo:bar::baz` produces
   `foo:bar:\n  baz` which YAML will read as nested keys, not what the user
   likely meant. Consider running `serde_yaml::from_str` on the expansion and
   surfacing a friendly error if it fails — gives a much better diagnostic than
   a downstream parse error far from the URI.

5. **Backoff math is correct but obtuse.** `1u32 << attempt.min(20)` * 500ms =
   up to ~524s, then `.min(60s)`. Works, but `attempt.min(7)` is the actual
   saturation point — clearer to cap there. Minor.

6. **`HttpConfigProvider::resolve` for `max_attempts == 0`** synthesizes a fake
   error rather than treating 0 as a programmer error. Either accept it as "0
   disables retries entirely" (consistent with the doc on `with_max_attempts`
   which says "1 disables retries"), or `debug_assert!(max_attempts > 0)`.
   Current behavior is fine but the doc/code disagree about whether 0 is
   meaningful.

7. **Plaintext HTTP fetch of pipeline config** is a real MITM risk (attacker
   can inject arbitrary pipeline definitions). Should be called out prominently
   in the rustdoc and probably the user-facing docs/release notes — currently
   only mentioned obliquely. The deferral of `https:` makes this more important
   to flag, not less.

8. **Tests**: `http_provider_retries_on_transient_failure` adds 1.5s of real
   sleep to the test suite (500ms + 1s). Acceptable but worth knowing. Consider
   exposing `BASE_BACKOFF` as a field for test injection if more retry tests
   are added.

## Nits

- `parse_scheme` test added one `yaml:` case; consider also
  `https://example.com` returning `Some("https")` to exercise the "unknown
  scheme" error path documented in the new error message.
- `expand_key_path` builds indentation with byte-by-byte `push(' ')` loops;
  `out.push_str(&" ".repeat(depth * 2))` reads better. Cosmetic.
- Error hint mentions `otap_df_otap::crypto::install_crypto_provider()` —
  verify path string stays accurate (it does, per
  `rust/otap-dataflow/src/main.rs:244`).

## What I verified

- `rust/otap-dataflow/src/main.rs:244` already installs the crypto provider
  before `resolve_config`, so production callers won't hit
  `ConfigHttpClientBuildFailed`.
- `parse_scheme` already returns `Some("http")` for `http://...` (test on line
  383 of original file).
- Workspace `reqwest` uses `rustls-no-provider` (Cargo.toml:226), so the rustls
  install requirement is real.
- YAML expansion math checked manually — output for the documented examples is
  correct.

## Recommendation

Request changes on (1), (2), and (3) at minimum. Items (4)–(8) are good-to-have.
The `yaml:` half could land independently as-is.

lquerel (approved):
LGTM

PR #2699, opened 2026-04-17
test: Add JUnit XML upload and flaky test tracking workflow

AaronRM:
# Change Summary

- Implemented JUnit XML result uploads for both required and non-required tests in the Rust-CI workflow.
- Created a new workflow for detecting flaky tests from JUnit XML artifacts, which runs daily and on-demand.
- The flaky test tracker parses JUnit results, identifies flaky tests, and creates or updates a tracking issue with a summary.

n/a

n/a. This is adding a new flaky test detection workflow. No changes to product code.

No. Test/infra only.


lquerel (approved):
LGTM

.github/workflows/scripts/parse_flaky.py
```diff
@@ -0,0 +1,383 @@
...
+    if not artifacts_path.exists():
+        print("No artifacts directory found", file=sys.stderr)
+        return test_results
+
+    for xml_file in artifacts_path.rglob("*.xml"):
+        # Extract run ID from path (junit-artifacts/run-<id>/...)
+        run_id = "unknown"
+        for part in xml_file.parts:
+            if part.startswith("run-"):
+                run_id = part.removeprefix("run-")
```
lalitb, line 60: I’m trying to understand the `run_id` logic here. A single rust CI workflow run can contain multiple distinct jobs across OSes and test partitions, but they all share the same run_id. If a test fails consistently on one platform and passes consistently on another within the same CI run, this logic would observe both a pass and a fail under one run_id and could incorrectly classify the test as flaky. Maybe more specific context is needed with run_id + `\<some meta-data\>` in that case.

.github/workflows/rust-ci.yml
```diff
@@ -703,6 +719,32 @@ jobs:
...
+      - test_nonrequired
+    permissions:
+      checks: write
+    steps:
+      - name: Download all JUnit XML artifacts
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          pattern: junit-xml-*
+          path: junit-results
+          merge-multiple: true
```
lalitb: Will this end up overwriting files in the report job? Each matrix artifact seems to contain a `junit.xml`, and this step downloads all of them with merge-multiple: true into a single directory?

lalitb (approved):
LGTM. Couple of comments - mostly doubts :)

.github/workflows/rust-ci.yml
AaronRM: Yes, you're correct. Removing merge-multiple so that we use per-artifact directories and avoid overwriting `junit.xml` files.

.github/workflows/scripts/parse_flaky.py
AaronRM: Pushing an update which changes the logic to be "For each OS that has failures, check if that same OS also has passes. Only if at least one OS has mixed results is the test flagged as flaky."

PR #2702, opened 2026-04-17
Transport header support in OTLP/Fake Data receivers

### LINKED ISSUE #2692: support transport headers in receivers
### Pre-filing checklist

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/), CI / Build

### Objective

Update the OTLP Receiver and Fake Data Receiver to support transport headers policy.

For the OTLP receiver it should be able to extract data from grpc headers and append transport headers on to OtapPdata

For the fake data receiver it should create transport headers and append them onto the generated OtapPdata

### Rationale

Currently there is no receivers that supports the transport header policy while we have a OTLP exporter that does. We need have a receiver node support the transport header feature to take full advantage of the transport header policy, so we will update the OTLP receiver to work with the OTLP exporter.
We also need to support the transport headers for the fake data receiver to allow for additional testing

### Scope

- Update OTLP Receiver to read grpc headers when there is a capture policy defined, and append transport headers to OtapPdata context.
- Add tests for OTLP receiver confirming that otap pdata has transport headers (when capture policy + grpc headers)
- Update fake data receiver config to allow user to provide a list of keys for the transport headers, user can provide a string value, if not a random string value will be generated for that key value pair. 
- Add tests to confirm that fake data receiver creates otap pdata with transport headers

### Acceptance Criteria

- otlp receiver extracts data from grpc headers when capture policy is defined and appends transport headers onto the otapdata context
- tests that confirm otlp receiver reads data from grpc headers and the otap pdata context has transport headers upon inspection.
- fake data receiver generates transport headers based on config
-  tests confirm that generated otappdata from the fake data receiver contains transport headers

### Dependencies or Blockers

_No response_

### Additional Context

_No response_

c1ly:
# Change Summary

Update the OTLP receiver to use the capture header policy when defined, extracts data from grpc/http headers

Update the Fake Data Generator to allow user to specify transport headers to tack on to generated OtapPdata, users can provide key/value pairs or just the key (random value will be generated).

* Closes #2692

unit tests

no


jmacd (approved):

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server_new.rs
```diff
@@ -365,12 +366,31 @@ impl UnaryService<OtapPdata> for OtapBatchService {
...
+        // Capture transport headers synchronously before moving the effect handler
+        // into the async block, avoiding a clone of the capture policy.
+        if let Some(policy) = effect_handler.capture_policy() {
+            let mut transport_headers = TransportHeaders::new();
+            let pairs = metadata.iter().map(|kv| match kv {
+                tonic::metadata::KeyAndValueRef::Ascii(key, value) => {
+                    (key.as_str(), value.as_bytes())
+                }
+                tonic::metadata::KeyAndValueRef::Binary(key, value) => {
+                    (key.as_str(), value.as_encoded_bytes())
```
lalitb, line 385: This seems to be an issue -  `value.as_encoded_bytes()` is the wire/base64 form for binary gRPC metadata, not the decoded raw bytes. That means we store encoded bytes in transport_headers, and any downstream gRPC propagation will encode them again and corrupt `*-bin` headers. We should capture the decoded bytes here instead.

lalitb: Please also add a test for a *-bin metadata key to verify we store raw bytes rather than the base64 wire form.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
```diff
@@ -316,6 +317,37 @@ impl BatchCache {
...
+    for (key, value) in config_headers {
+        let resolved_value: Vec<u8> = match value {
+            Some(v) => v.as_bytes().to_vec(),
+            None => {
+                let mut buf = [0u8; 16];
+                rand::RngExt::fill(&mut rand::rng(), &mut buf);
+                buf.to_vec()
+            }
+        };
+        headers.push(TransportHeader::text(
```
lalitb, line 342: This seems a bit off to me, since it will also mark gRPC binary headers as "Text", which feels inconsistent with their intended semantics. Should we infer the header kind here?

c1ly: updated to determine the header type based on whether the header has -bin suffix

rust/otap-dataflow/crates/otap/src/otap_grpc/otlp/server_new.rs
c1ly: updated to use to_bytes which provides the decoded raw bytes https://docs.rs/tonic/latest/src/tonic/metadata/value.rs.html#129-131

lalitb (approved):
thanks.

PR #2705, opened 2026-04-18
Crypto library selection condition simplification

sapatrjv:
# Change Summary

Part of #2537


lalitb (approved):
LGTM; this is a good simplification. 
As a follow-up, it would be good to add a single CI feature-matrix validation, since changing these crypto features currently requires updates in three places:
- Cargo.toml, 
- the compile-time guard in src/main.rs, and 
- the runtime provider selection in crates/otap/src/crypto.rs.

PR #2716, opened 2026-04-20
Include response body in RateLimited/ServerError Display for azure_monitor_exporter

cijothomas:
The `Display` impl for `Error::RateLimited` and `Error::ServerError` previously omitted the HTTP response body, losing actionable details when these errors were logged at WARN level via Display formatting (`%`).

For example, when the main export path exhausts retries on a 429, the WARN log showed:
```
Export failed after 5 attempts: Rate limited
```

With this change it now shows:
```
Export failed after 5 attempts: Rate limited: {"error":{"code":"QuotaLimitExceeded","message":"Allowed data-rate per data collection rule: 2GB/minute"}}
```

This lets operators identify *which* Azure Monitor quota was hit (data-rate vs request-rate per DCR) without enabling debug logging.

Additionally, the heartbeat failure WARN log in `exporter.rs` is changed from Debug format (`?e`) to Display format (`%e`), aligning with the project's [events guide](rust/otap-dataflow/docs/telemetry/events-guide.md) recommendation to prefer Display for `thiserror` types at warn/error severity.

**Changes:**
- `error.rs`: Include `{body}` in `#[error(...)]` for `RateLimited` and `ServerError` variants; update corresponding test assertions.
- `exporter.rs`: Change heartbeat send failure WARN log from `error = ?e` (Debug) to `error = %e` (Display).

### DIFF
--- a/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
@@ -584,7 +584,7 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
                     if let Some(ref mut hb) = self.heartbeat {
                         match hb.send().await {
                             Ok(_) => otel_debug!("azure_monitor_exporter.heartbeat.sent"),
-                            Err(e) => otel_warn!("azure_monitor_exporter.heartbeat.send_failed", error = ?e),
+                            Err(e) => otel_warn!("azure_monitor_exporter.heartbeat.send_failed", error = %e),
                         }
                     }
                 }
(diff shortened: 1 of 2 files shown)


rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
```diff
@@ -584,7 +584,7 @@ impl Exporter<OtapPdata> for AzureMonitorExporter {
                     if let Some(ref mut hb) = self.heartbeat {
                         match hb.send().await {
                             Ok(_) => otel_debug!("azure_monitor_exporter.heartbeat.sent"),
-                            Err(e) => otel_warn!("azure_monitor_exporter.heartbeat.send_failed", error = ?e),
+                            Err(e) => otel_warn!("azure_monitor_exporter.heartbeat.send_failed", error = %e),
```
drewrelmas, line 587: Could you please provide concrete example of before/after moving from `?` to `%` in PR description?

drewrelmas (approved):

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/exporter.rs
jmacd: Good question, the before-case ? looks like Debug, which is not derived for the `Error` type. The `thiserror` machinery derives Display, so this looks right.

jmacd (approved):

PR #2717, opened 2026-04-20
  Add Linux user_events Receiver

### LINKED ISSUE #2698: Proposal: Add Linux user_events Receiver for Telemetry from Applications Co-Located with the Collector
### Pre-filing checklist

### Component(s)

## Summary

  Add a receiver for application telemetry emitted through Linux [`user_events`](https://docs.kernel.org/trace/user_events.html) tracepoints.

  This receiver targets telemetry emitted by applications through Linux `user_events` tracepoints. It is **_not intended_** for generic kernel/host event collection or Linux perf/profile ingestion.

  ## Motivation

For applications running on the same Linux host as the collector, `user_events` provides a native `tracepoint`-based event path through the Linux tracing/perf stack. Events can be consumed from kernel-managed per-CPU ring buffers via perf_event_open, avoiding the local TCP/UDP/HTTP/gRPC stack. This offers lower-overhead, lower-latency same-host ingestion with bounded in-kernel buffering, but not durable persistence.

  Producer compatibility anchor:
  - Rust exporter: [`opentelemetry-user-events-logs`](https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-user-events-logs)
  - C++ exporter: [`opentelemetry-user_events-exporter`](https://github.com/open-telemetry/opentelemetry-cpp-contrib/tree/main/exporters/user_events)
  - .Net exporter [`OpenTelemetry.Exporter.Geneva`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/main/src/OpenTelemetry.Exporter.Geneva/README.md)

  Related decoding surface:
  - EventHeader decoding crate: [`tracepoint_decode`](https://docs.rs/tracepoint_decode/latest/tracepoint_decode/)

This proposal is to add this as part of `contrib-nodes` crate under feature flag. 

## Data flow:

```mermaid
flowchart TB
    A[Application A threads]
    B[Application B threads]
    C[Application C threads]

    U[Linux user_events tracepoint writes]

    A --> U
    B --> U
    C --> U

    R0[CPU 0 perf ring buffer]
    R1[CPU 1 perf ring buffer]
    R2[CPU 2 perf ring buffer]
    RN[CPU N perf ring buffer]

    U --> R0
    U --> R1
    U --> R2
    U --> RN

    P0[df-engine pipeline 0<br/>pinned to CPU 0<br/>Receiver → Processors → Exporters]
    P1[df-engine pipeline 1<br/>pinned to CPU 1<br/>Receiver → Processors → Exporters]
    P2[df-engine pipeline 2<br/>pinned to CPU 2<br/>Receiver → Processors → Exporters]
    PN[df-engine pipeline N<br/>pinned to CPU N<br/>Receiver → Processors → Exporters]

    R0 --> P0
    R1 --> P1
    R2 --> P2
    RN --> PN
```
 
### Proposed Solution

- As above

### Alternatives Considered

- N/A

### Additional Context

- N/A

lalitb:
# Change Summary

Adds a `contrib` receiver for ingesting events through Linux `user_events` tracepoint.
 The receiver reads configured Linux user_events tracepoints through per-CPU perf sessions, decodes supported log payloads, and emits OTAP log records into the pipeline. It includes batching, late tracepoint registration, downstream-full overflow handling, memory-pressure ingress shedding, drop/loss metrics, and Linux-only feature gating.

Detailed configuration and runtime behavior are documented in- 
`crates/contrib-nodes/src/receivers/userevents_receiver/README.md`.

* Closes #2698


rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,523 @@
+# Linux Userevents Receiver
+
+**URN:** `urn:microsoft:receiver:userevents`
```
cijothomas: do we need to include microsoft in the name? There is no vendor specific logic right?

cijothomas: ability to do MS CommonSchema - I see we do that, but its optional, and users are free to do their own schema mapping (even if its not implemented today). I think we should make that more clear, and then remove Microsoft from the urn.

lalitb: Yes was going to raise this question while opening the PR - good you already raised :) - I am in agreement to make it vendor-agnostic naming with assumption that in future users can being their own mapping.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,530 @@
...
+  ring:      R0     R1     R2     R3     R4     R5     R6     R7
+              ^      ^      ^      ^      ^      ^      ^      ^
+              |      |      |      |      |      |      |      |
+  pipeline:  P0     P1     P2     P3     P4     P5     P6     P7
+
+  P0 reads only R0. P1 reads only R1. ... P7 reads only R7.
+```
+
+If a CPU has no corresponding pipeline, writes on that CPU are not collected by
+this receiver deployment. They are not seen and then discarded; no receiver is
```
cijothomas, line 106: lets make it clear that there is no metric emitted by engine about the missing events in other CPU - they are totally lost and silently.

Perhaps also suggest that the running engine on --cores should be done carefully when using this receiver,

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,530 @@
...
+  placement, such as `numactl --interleave=all`.
+- The receiver is changed to read perf rings for CPUs other than its assigned
+  pipeline core. Today `session.rs` intentionally constructs
+  `cpu_ids = vec![pipeline.core_id()]`.
+
+## Platform Support
+
+This receiver is **Linux-only**.
+
+It does **not** work on macOS or Windows because `user_events` is a Linux
```
cijothomas, line 191: can you add the kernel version requirement? 6.3 I think? 
and mention the Azure Linux and Ubuntu releases that include this, so anyone trying this would know which exact distro to try this!

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,530 @@
...
+kernel tracing feature. Windows support would require a separate ETW receiver.
+
+## Current Scope
+
+Current implementation supports:
+
+- single-tracepoint and multi-tracepoint configuration
+- `common_schema_otel_logs` as the only supported decode format
```
cijothomas: nit: Given we mean Microsoft Common Schema, lets make it explicit that it is Microsoft schema. Without qualifier, it could be misinterpreted as some"common" format.!

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,530 @@
...
+    config:
+      subscriptions:
+        - tracepoint: "user_events:myprovider_L2K1"
+          format:
+            type: common_schema_otel_logs
+        - tracepoint: "user_events:app_L2K1"
+          format:
+            type: common_schema_otel_logs
+      session:
+        per_cpu_buffer_size: 1048576
```
cijothomas: unit?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,530 @@
...
+Tracepoints must be configured with the `user_events:` group prefix. The receiver
+rejects other groups because collection always uses the Linux `user_events`
+tracefs group:
+
+```text
+user_events:<provider>_L<level>K<keyword>
+```
+
+The `_L<level>K<keyword>` suffix follows EventHeader tracepoint naming. The
+receiver uses `level` only as a fallback when PartB does not provide
```
cijothomas: nit: Avoid Common Schema terms like PartB in main doc.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,530 @@
...
+| `5` | `DEBUG` / `5` |
+
+If the suffix is absent or malformed, Common Schema payloads with no
+`severityNumber` are emitted without a fallback severity.
+
+## Decode
+
+### `common_schema_otel_logs`
+
+Intended for payloads produced by
```
cijothomas: mention that this is just an example or reference implementation producing events in MS CommonSchema format.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,530 @@
...
+PartC fields are emitted as flat attributes using their original names
+(e.g. `user_name`, `user_email`) with their source types preserved
+(`Int`/`Bool`/`Double`/`Str`).
+
+The receiver intentionally does **not** emit receiver-internal
+transport/diagnostic fields such as tracepoint name, provider name,
+EventHeader level/keyword, CPU, PID/TID, sample id, payload size, body
+encoding, or decode mode. These describe the receiver itself rather than the
+application payload; surfacing them as OTLP log attributes would pollute
+downstream backends (e.g. Geneva turns each attribute into a dynamic backend
```
cijothomas: remove Geneva mentions.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,530 @@
...
+  dropped rather than flushed.
+- `overflow.on_downstream_full = drop` is currently the only mode; the perf
+  drain loop is never blocked on downstream.
+
+## Runtime Behavior
+
+### Backpressure
+
+`user_events` perf rings cannot be backpressured like a socket. For that
+reason, the receiver defaults to dropping when downstream is full instead of
```
cijothomas: is it receiver who is dropping, or the kernel itself doing it?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/decoder.rs
```diff
@@ -0,0 +1,1610 @@
...
+                } else if name == b"ext_dt_spanId" {
+                    let value = item.value().bytes();
+                    if let Some(bytes) = parse_hex_bytes_ascii::<8>(value) {
+                        cs.span_id = Some(bytes);
+                    } else {
+                        cs.span_id_text = item_as_string(&item);
+                    }
+                } else if name == b"ext_dt_traceFlags" {
+                    cs.trace_flags = item_as_u32(&item);
+                } else if matches!(name, b"ext_cloud_role" | b"ext_cloud_roleInstance") {
```
ThomsonTan: Should this be decoded instead of being ignored?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
lalitb: - Receiver drops already-drained batches when downstream is full, counted as `dropped_downstream_full`.
  - Kernel/perf ring overruns before the receiver drains are counted separately as `lost_perf_samples`.

Clarified in the document.

lalitb:
This is ready for review. cc - @utpilla @cijothomas @ThomsonTan.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/arrow_records_encoder.rs
```diff
@@ -0,0 +1,252 @@
...
+    /// Returns true when the builder is empty.
+    #[must_use]
+    pub(super) const fn is_empty(&self) -> bool {
+        self.curr_log_id == 0
+    }
+
+    /// Appends a decoded userevents record.
+    pub(super) fn append(&mut self, record: DecodedUsereventsRecord) {
+        self.logs.append_time_unix_nano(record.time_unix_nano);
+        self.logs.body.append_str(record.body.as_bytes());
```
cijothomas, line 57: interesting..body is `str`? do we support complex body like arrays , maps etc?

cijothomas: also, body can be int or float simple types too right?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/metrics.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// Number of samples dropped because the downstream channel was full.
+    #[metric(unit = "{item}")]
+    pub dropped_downstream_full: Counter<u64>,
+    /// Number of samples dropped due to process-wide memory pressure.
+    #[metric(unit = "{item}")]
+    pub dropped_memory_pressure: Counter<u64>,
+    /// Number of samples dropped because no matching subscription was found.
+    #[metric(unit = "{item}")]
+    pub dropped_no_subscription: Counter<u64>,
+    /// Number of samples that failed CS decoding and fell back to base64.
```
cijothomas: such metrics tie us too much to CS. We expect it to be general purpose, so maybe the metrics can be "decode_failed", and decoding could be simple k-v pair or following cs or some other format.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/metrics.rs
```diff
@@ -0,0 +1,43 @@
...
+    #[metric(unit = "{item}")]
+    pub cs_decode_fallbacks: Counter<u64>,
```
cijothomas: fallbacks -> failures?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/metrics.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// Number of lost samples reported by the perf ring.
```
cijothomas, line 32: curious if "perf" is strictly involved or it is just one of the ways to listen.. Is there more technically accurate term.. like kernel_dropped_samples?

cijothomas: also, is 1 sample - 1 log record? if yes, lets clarify it in the metric desc.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/mod.rs
```diff
@@ -0,0 +1,933 @@
...
+/// Keep the receiver identity vendor-neutral. The current implementation only
+/// supports the Common Schema OTLP logs decoder, but the receiver is intended
+/// to support pluggable schema mappings in the future.
+pub const USEREVENTS_RECEIVER_URN: &str = "urn:otel:receiver:userevents";
+
+#[derive(Debug, Clone, Default, Deserialize)]
+#[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
+enum FormatConfig {
+    #[default]
+    CommonSchemaOtelLogs,
```
cijothomas:
```suggestion
    MicrosoftCommonSchemaLogs,
```

rust/otap-dataflow/crates/contrib-nodes/Cargo.toml
```diff
@@ -46,14 +49,28 @@ azure_core = { workspace = true, optional = true, features = ["reqwest"] }
...
 rand = { workspace = true, optional = true }
 ryu = { workspace = true, optional = true }
 reqwest = { workspace = true, optional = true, features = ["rustls-no-provider"] }
 sysinfo = { workspace = true, optional = true }
 urlencoding = { workspace = true, optional = true }
 
+[target.'cfg(target_os = "linux")'.dependencies]
+# TODO: Remove this pinned git dependency once one-collect is upstream in the
+# normal dependency graph and no longer needs to be pulled directly here.
+one_collect = { git = "https://github.com/microsoft/one-collect.git", rev = "2d33e674348168f4a19bd3ab40d92f1c556407ec", features = ["scripting"], optional = true }
```
cijothomas, line 61: nit: put the date of the commit here in comment, so it is to tell which date's version we are consuming here.

rust/otap-dataflow/crates/contrib-nodes/README.md
```diff
@@ -14,6 +18,18 @@ This crate contains optional (feature-gated) contrib processors and exporters.
...
+
+| Node | URN | Module |
+| ---- | --- | ------ |
+| userevents_receiver | `urn:otel:receiver:userevents` | `src/receivers/userevents_receiver/` |
+
+#### userevents_receiver
+
+- Reads Linux `user_events` tracepoints through per-CPU perf sessions
+- Supports single-tracepoint and multi-tracepoint configuration
+- Supports the `common_schema_otel_logs` decode format for supported log payloads
```
cijothomas:
```suggestion
- Supports the `microsoft_common_schema_logs` decode format for supported log payloads
```

1. I think it's best to quality CS with Microsoft, as it is hard for users to known what is "common" about it.
2.CommonLogs is just CS Logs - its inspired by OTel Logs data model, but not tied to it, so I suggest avoiding the term OTel Logs, Just "Logs" is good.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+# Linux Userevents Receiver
+
+**URN:** `urn:otel:receiver:userevents`
+
+This receiver ingests Linux
+[`user_events`](https://docs.kernel.org/trace/user_events.html) tracepoints
+through `perf_event_open` and converts them into OTAP logs for downstream
+processing.
+
+> **Note:** The current decoder supports the Microsoft Common Schema OTLP logs
```
cijothomas:
```suggestion
> **Note:** The current decoder supports the Microsoft Common Schema logs
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+- one receiver instance per pipeline thread
+- one perf session per assigned CPU
+- one bounded drain loop per receiver
+- no shared hot-path state across pipeline threads
+
+## Architecture
+
+`user_events` writes enter Linux tracing as user-space tracepoint events. The
+kernel stores trace records in per-CPU ring buffers; each receiver instance
+drains the ring for the CPU/pipeline core it owns, decodes the EventHeader
```
cijothomas: out of curiosity-what if 2 instances of df_engine is run on a node - will they both read the same events from the ring buffer? Nothing wrong with that, but making sure we anticipate such scenarios and make no assumptions otherwise.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+the per-CPU buffers independently. The receiver consumes those buffers via
+`perf_event_open` / `one_collect`; it does not implement the kernel ring-buffer
+algorithm itself.
+
+## CPU Pinning and Coverage
+
+**Placement rule:** a pipeline assigned to core `K` opens the perf ring for
+CPU `K` only. If an application thread writes a `user_events` record while the
+kernel has scheduled it on CPU `X`, that record lands in the CPU `X` ring and
+is only visible to the receiver instance that owns CPU `X`. No other receiver
```
cijothomas: receiver instance that owns CPU 'x" feels incorrect... Receiver runs on CPU X, and it only reads from that CPU X. It cannot assume ownership of ring buffer X as another df_engine instance could also run.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+```text
+  user_events write on CPU K
+             |
+             v
+       CPU K perf ring
+             |
+             v
+  receiver pipeline pinned to CPU K
+
+  No other receiver pipeline reads CPU K's ring.
```
cijothomas:
```suggestion
  No other receiver pipeline in this df_ngine instance reads CPU K's ring.
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+```
+
+For an 8-core host, coverage looks like this:
```
cijothomas: nit: user 4 core example to make it slightly easier to read.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+  pipeline:  P0     P1     P2     P3     P4     P5     P6     P7
+
+  P0 reads only R0. P1 reads only R1. ... P7 reads only R7.
+```
+
+If a CPU has no corresponding pipeline, writes on that CPU are not collected by
+this receiver deployment. They are not seen and then discarded; no receiver is
+attached to that CPU's ring. The engine does not emit a metric for these
+missing records because they are never observed by the receiver; they are lost
+silently.
```
cijothomas: nit: do mention that silently refers to silent from df_engine standpoint. The sender could very well check is_listener_available() and use that to emit its own metric, so it won't be totally silent loss... Maybe too much details? Won't hurt!

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+
+| Configuration                    | Writer pinned? | Result                                                     |
+|----------------------------------|----------------|------------------------------------------------------------|
+| `--num-cores 1`                  | No             | Racy; writes on CPUs 1..N not collected by this receiver   |
```
cijothomas: what does "Racy" means?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
cijothomas: I don't see the version listed still?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+
+It does **not** work on macOS or Windows because `user_events` is a Linux
+kernel tracing feature. Windows support would require a separate ETW receiver.
+
+## Current Scope
+
+Current implementation supports:
+
+- single-tracepoint and multi-tracepoint configuration
+- `common_schema_otel_logs`, which decodes Microsoft Common Schema OTLP logs,
```
cijothomas:
```suggestion
- `common_schema_otel_logs`, which decodes Microsoft Common Schema logs,
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+          enabled: true
+          poll_interval_ms: 100
+      drain:
+        max_records_per_turn: 1024
+        max_bytes_per_turn: 1048576
+        max_drain_ns: 2ms  # duration string: 2ms, 500us, 1000000ns
+      batching:
+        max_size: 512
+        max_duration: 50ms
+      overflow:
```
cijothomas, line 233: does other receivers has such a setting to define overflow behavior?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+        per_cpu_buffer_size: 1048576  # bytes
+        late_registration:
+          enabled: true
+          poll_interval_ms: 100
+      drain:
+        max_records_per_turn: 1024
+        max_bytes_per_turn: 1048576
+        max_drain_ns: 2ms  # duration string: 2ms, 500us, 1000000ns
+      batching:
+        max_size: 512
```
cijothomas, line 231: 512 is count of items? or size? the config is not clear...

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+          format:
+            type: common_schema_otel_logs
+        - tracepoint: "user_events:app_L2K1"
+          format:
+            type: common_schema_otel_logs
+      session:
+        per_cpu_buffer_size: 1048576  # bytes
+```
+
+Exactly one of `tracepoint` or `subscriptions` must be configured.
```
cijothomas: the example has tracepoint as a sub-setting within subscriptions.. so not sure what this statement is referring to . Is that a stale comment?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+TODO: Wire `session.wakeup_watermark` into the perf ring setup once
+`one_collect` exposes wakeup/readiness and watermark configuration for
+tracepoint sessions.
+
+### Configuration Reference
+
+| Field | Default | Description |
+| --- | --- | --- |
+| `tracepoint` | none | Single tracepoint shorthand. Must use `user_events:<event>`. Mutually exclusive with `subscriptions`. |
+| `subscriptions` | none | List of tracepoints. Each entry must use `user_events:<event>` and `format.type: common_schema_otel_logs`. |
```
cijothomas: could we avoid subscriptions concept, and just make tracepoint config a [], so users can put 1 or more tracepoints there? Sorry I am missing why we have subscription concept.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+| `format.type` | `common_schema_otel_logs` | Only supported decode format. |
+| `session.per_cpu_buffer_size` | `1048576` | Requested per-CPU perf ring size in bytes. Rounded by the underlying perf/ring setup. |
```
cijothomas: nit: is it common to specify size in bytes? If yes, do we support human friendly settings like 10KB?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+| `session.wakeup_watermark` | `262144` | Reserved for future one_collect wakeup support; currently ignored. |
+| `session.late_registration.enabled` | `false` | When true, keep retrying if the tracepoint is not registered yet. |
```
cijothomas, line 307: add some explanation when this is relevant. like the engine starts up first, and apps start much later, so engine won't find the tracepoints when it starts and hence it has to keep poling.

(or is there an alternate where the engine just creates the tracepoint itself..?)

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+| `session.late_registration.poll_interval_ms` | `1000` | Retry interval for late tracepoint registration. |
+| `drain.max_records_per_turn` | `1024` | Maximum records popped from the receiver's pending queue per drain turn. |
+| `drain.max_bytes_per_turn` | `1048576` | Maximum payload bytes popped per drain turn. |
+| `drain.max_drain_ns` | `2ms` | Total drain-turn budget. Accepts duration strings such as `2ms`, `500us`, or `1000000ns`; must be greater than zero. |
+| `batching.max_size` | `512` | Flush once this many logs are buffered in the current Arrow batch. |
```
cijothomas, line 312: add a note about why *this* batching is required. (something like sending one arrow batch with 1 record is inefficient etc..)

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+
+```text
+user_events:<provider>_L<level>K<keyword>
+```
+
+The `_L<level>K<keyword>` suffix follows EventHeader tracepoint naming. The
+receiver uses `level` only as a fallback when the payload does not provide
+`severityNumber`:
+
+| EventHeader level | Fallback severity |
```
cijothomas: this feels an arbitrary mapping..lets link to OTel Log data model, so this makes clear sense.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+| EventHeader name / PartA.name | fallback for typed `event_name` |
+| PartA.time | `time_unix_nano` |
+| PartA.ext_dt_traceId | typed `trace_id` when valid hex; otherwise `trace.id` attribute |
+| PartA.ext_dt_spanId | typed `span_id` when valid hex; otherwise `span.id` attribute |
+| PartA.ext_dt_traceFlags | log `flags` |
+| PartB.name | preferred typed `event_name` |
+| PartB.severityNumber | `severity_number` |
+| PartB.severityText | `severity_text` |
+| PartB.body | log `body` |
+| PartB.eventId | typed `eventId` attribute |
```
cijothomas: I think we can drop this thing... OTel Rust producer also need to drop this. EventName is serving the needs currently, adding a numerical id, specially treated inside attribute is additional complexity nobody outside us may understand!

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+| PartC scalar fields | flat log attributes with original field names and basic value types |
```
cijothomas: TODO for supporting non-basic? or we can make a statement that user_events only support basic types, and producers are expected to produce that only?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+
+Decode failure means the payload could not be interpreted as the supported
+EventHeader/Common Schema log shape. Examples include malformed EventHeader
+bytes, `__csver__` missing or not first, `__csver__ != 0x400`, missing PartB,
+PartB `_typeName != "Log"`, duplicate PartA/PartB/PartC structs, unknown PartB
+fields, or invalid nesting. On failure, the receiver emits the raw payload as a
+base64 body and increments `cs_decode_fallbacks`.
```
cijothomas: nice!

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+
+## Output Shape
+
+The receiver emits OTAP logs. Payload data is represented as typed OTLP log
+fields and flat application attributes; receiver transport/debug metadata is
+not emitted as log attributes.
+
+The only PartB field emitted as a log attribute is:
```
cijothomas: as suggestd elsewhere, its probably best to remove this special casing.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+
+- `eventId` (typed Int, when the Common Schema payload carries PartB.eventId)
+
+Typed OTLP log fields (not attributes) also carry:
+
+- `body`
+- `severity_number` / `severity_text`
+- `event_name` (prefers PartB.name, falls back to EH.Name / PartA.name)
+- `time_unix_nano` (from PartA.time when present, else the perf sample timestamp)
```
cijothomas: shouldn't;t we also check for observed_ts, before falling back to perf sample timestamp?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+
+PartC fields are emitted as flat attributes using their original names
+(e.g. `user_name`, `user_email`) with their source types preserved
+(`Int`/`Bool`/`Double`/`Str`).
+
+The receiver intentionally does **not** emit receiver-internal
+transport/diagnostic fields such as tracepoint name, provider name,
+EventHeader level/keyword, CPU, PID/TID, sample id, payload size, body
+encoding, or decode mode. These describe the receiver itself rather than the
+application payload; surfacing them as OTLP log attributes would pollute
```
cijothomas: I think pid/tid are quite useful, and describes the app itself right? worth considering, but not crucial for v1.

cijothomas: pid is important I think, as multiple processes can exist and this is only way to distinguish?
anyway, not blocker for the 1st PR.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,558 @@
...
+## Linux Requirements
+
+This receiver requires all of the following on the host:
+
+- Linux kernel **6.4 or later**, built with `CONFIG_USER_EVENTS=y`
+- tracefs available, typically under `/sys/kernel/tracing`
+- permission to read tracefs metadata
+- permission to use `perf_event_open` for the configured tracepoints
+
+`user_events` was merged in Linux 6.4; earlier kernels do not expose
```
cijothomas, line 504: nice. I think I left a comment about documenting this elsewhere, but just saw you already had that here!

rust/otap-dataflow/crates/contrib-nodes/tests/userevents_exporter_receiver_e2e.rs
```diff
@@ -0,0 +1,209 @@
...
+            let expected_event_id = userevents_common_schema_emitter::EVENT_ID.to_string();
+            assert_eq!(
+                attrs.get("eventId").map(String::as_str),
+                Some(expected_event_id.as_str())
+            );
+        })
+    }
+}
+
+#[ignore = "requires Linux kernel user_events support and write permissions to tracing/user_events_data"]
```
cijothomas: don't we have that in the Github Linux runners? We run user_events tests in CI in OTel Rust Contrib.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/mod.rs
```diff
@@ -0,0 +1,933 @@
...
+const DEFAULT_BATCH_MAX_SIZE: u16 = 512;
+const DEFAULT_BATCH_MAX_DURATION: Duration = Duration::from_millis(50);
+const DEFAULT_LATE_REGISTRATION_POLL: Duration = Duration::from_secs(1);
+
+/// URN for the Linux userevents receiver.
+///
+/// Keep the receiver identity vendor-neutral. The current implementation only
+/// supports the Common Schema OTLP logs decoder, but the receiver is intended
+/// to support pluggable schema mappings in the future.
+pub const USEREVENTS_RECEIVER_URN: &str = "urn:otel:receiver:userevents";
```
utpilla: I think we should aim for a generic `user_events` receiver rather than one that bakes in Common Schema mapping. A few thoughts on how this could be structured:

**1. Start with what the Linux ecosystem already supports.**

A generic `user_events` receiver should, at minimum, decode events the same way standard Linux tools (perf, ftrace, etc.) do (using the tracefs format string declared at registration time). If a producer registers `"my_event u32 status; char[] endpoint"`, the receiver should emit those as typed attributes on a log record. The `tracepoint_decode` crate already supports this via its `TraceEventFormat` decoding style. I don't think in its current state, the PR handles this path. It only decodes EventHeader payloads.

**2. Extend with EventHeader decoding as an opt-in feature.**

`tracepoint_decode` auto-detects whether a tracepoint uses standard format-string fields or the EventHeader self-describing convention (by checking if the first field is named `eventheader_flags`). We could support both decode paths in the receiver, with EventHeader gated behind a feature flag. For standard events, attributes come from the format-string fields. For EventHeader events, attributes come from the self-describing metadata. Both land as typed attributes on OTAP/OTLP log records. There is no semantic interpretation at the receiver level.

This gives us a genuinely generic receiver that works with any `user_events` event. The EventHeader feature flag lets us keep that path opt-in since EventHeader is a custom userspace convention.

**3. Model the Common Schema mapping as a processor.**

The CS-specific logic (checking `__csver__ == 0x400`, promoting `PartB.body` → `LogRecord.body`, parsing `PartA.ext_dt_traceId` → typed trace ID, mapping `ext_cloud_role` → `service.name`, etc.) is essentially attribute transformation: promote, rename, delete. This fits naturally as a processor in contrib-nodes. The receiver emits decoded EventHeader fields as attributes; the CS processor promotes the known ones to OTAP/OTLP typed fields and strips the originals.

This separation would allow for a generic receiver with a vendor-neutral name and scope, while the custom Common Schema interpretation can reside in a contrib processor node. We could even consider making `user_events` reciever a core node at that point.

reyang: > **1. Start with what the Linux ecosystem already supports.**

+100

> **2. Extend with EventHeader decoding as an opt-in feature.**

+1

> **3. Model the Common Schema mapping as a processor.**

Not sure, I guess there are many ways to do this. Having said that, I think it'll be great to focus on building a great user_events receiver first, and experiment with it to see what can be done to fit the Microsoft need (extension, etc.). I don't think anyone here would want to build a user_events receiver based on Microsoft's specific need.

lalitb: That makes sense; I’ll align the PR around a generic Linux `user_events` receiver first, with tracefs-format decoding as the baseline and EventHeader handled as an explicit extension point, while keeping Microsoft/CommonSchema-specific mapping outside the receiver boundary.

lalitb:
Moving to draft for refactor as discussed [here](https://github.com/open-telemetry/otel-arrow/pull/2717#discussion_r3140772146)

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/mod.rs
lalitb: This is done. Thanks @utpilla  - really nice feedback.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
lalitb: No, all other receivers can manage back pressure (except probably syslog-udp).

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
lalitb: this code is stale, we only have one way of setting tracepoint now.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
lalitb: Thanks. I have added back both.

rust/otap-dataflow/crates/contrib-nodes/tests/userevents_exporter_receiver_e2e.rs
lalitb: Yes, the CI was not stable, and I didn't want to add another test for now. Will do this separately.

lalitb:
> Moving to draft for refactor as discussed [here](https://github.com/open-telemetry/otel-arrow/pull/2717#discussion_r3140772146)

Ready for review now. thanks for all the comments.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/decoder.rs
```diff
@@ -0,0 +1,704 @@
...
+//! ```
+//!
+//! In `tracefs` mode, the producer-defined data is decoded with the field
+//! offsets, sizes, locations, and type names from the tracefs `format` file.
+//! This is the standard Linux tracepoint model used by perf/ftrace. After a
+//! successful tracefs decode, downstream processors receive the decoded OTAP log
+//! attributes, not the original raw tracepoint sample bytes. Unknown static
+//! fields may be preserved as per-field base64 string attributes.
+//!
+//! In `event_header` mode, the producer-defined data is decoded as an
```
ThomsonTan: nit: should be `EventHeader` mode?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
lalitb: Currently bytes only, matching the perf ring API. We can add  KB/MB parsing later as separate PR. Added TODO for this.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
lalitb: This is covered now in the Late Registration section below. The engine does not create user_events tracepoints; producers register them, so this mode covers engine-starts-first scenarios.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
lalitb: This is stale after the refactor. The receiver no longer maps `EventHeader` level to OTLP severity; Common Schema promotion uses PartB.severityNumber directly.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
lalitb: Good point. The internals section shows where batching happens, but not why. I have added a short note that batching avoids one Arrow payload per perf sample and balances throughput with latency via max_size/max_duration.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
lalitb: This is covered now in the event_header section. The receiver currently emits only scalar values: strings, signed integers, booleans, and floats. Arrays/blobs/non-scalar EventHeader encodings are explicitly not emitted yet. Have added TODO for now.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
lalitb: Agree. EventName is enough for now, and treating PartB.eventId specially adds unnecessary schema coupling. I have dropped the promotion to eventId and preserved PartB.eventId as a normal attribute if present.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
lalitb: Yes, updated that.  The PartA.time already carries the producer-side timestamp fallback, including observed timestamp. The receiver now keeps the perf sample timestamp as observed_time_unix_nano.

lalitb:
Discussed offline with @utpilla  - and we agreed to split this PR into two parts
- One PR for Linux user-events receiver ( more generic)
- Another PR for MS common schema processor (Microsoft specific)

Closing this PR, will be raising two separate PRs now.

PR #2721, opened 2026-04-21
[query-engine] Tweak slice validation errors

CodeBlanch:
Relates to #2636

# Changes

* Tweak the slice validation error messages

### DIFF
--- a/rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
+++ b/rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
@@ -1168,7 +1168,7 @@ impl SliceScalarExpression {
             Err(ExpressionError::TypeMismatch(
                 query_location.clone(),
                 format!(
-                    "Range {name} for a slice expression should be an integer type, encountered '{}' value",
+                    "Range {name} for a slice expression should be an integer type, encountered '{}' type",
                     value.get_value_type()
                 ),
             ))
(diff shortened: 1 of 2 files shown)


rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
```diff
@@ -1181,8 +1186,7 @@ impl SliceScalarExpression {
             return Err(ExpressionError::ValidationFailure(
                 query_location.clone(),
                 format!(
-                    "{name} slice index starts at '{range_start}' but target ends at index '{}'",
-                    target_length - 1
```
CodeBlanch, line 1185: When target has a length of `0` this `usize - 1` op leads to overflow. Switched the error message to just show the length to avoid the issue.

rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs
```diff
@@ -1158,14 +1158,19 @@ impl SliceScalarExpression {
...
+                    ),
                 ));
             }
             Ok(v as usize)
         } else {
             Err(ExpressionError::TypeMismatch(
                 query_location.clone(),
-                format!("Range {name} for a slice expression should be an integer type"),
+                format!(
+                    "Range {name} for a slice expression should be an integer type, encountered '{}' value",
```
drewrelmas:
```suggestion
                    "Range {name} for a slice expression should be an integer type, encountered '{}' type",
```

drewrelmas (approved):
Nice catch! Left one nit on the `TypeMismatch` case

PR #2723, opened 2026-04-21
feat: Make fake data generator pacing more reliable and update sql reports to query correct time frame

### LINKED ISSUE #2713: Produced signals rate in benchmarks is off by as much as 20%
### Pre-filing checklist

### Component(s)

### Bug Description

Metrics for all of our benchmarks show that the produced signals rate is off by as much as 20%. For example, in the 100klrps test, the metrics are showing ~120klrps. It also varies from run to run and scenario to scenario which compromises the results.

Image

### Steps to Reproduce

View the 100klrps benchmarks at https://open-telemetry.github.io/otel-arrow/benchmarks/continuous/.

### Expected Behavior

Almost exactly 100klrps.

### Actual Behavior

Closer to 120klrps.

### OTel-Arrow Version

NA

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

JakeDern:
# Change Summary

This PR is an attempt to solve #2713 with three connected changes.

The first is to make the reported data production rate coming out of the fake data generator smoother and more reliable especially under backpressure by using tokio intervals which are less prone to drift and have a catch-up mechanism.

The second makes sure we report metrics in smaller increments after every batch is exported. #2685 made some progress in this area, but as noted there it was an incomplete solution which held back metrics for prolonged periods of time in addition to having to heap allocate the send futures. I went after the full solution to the problem in this PR. Now we exclusively do non-blocking sends one batch at a time and always bump counters in-between on successful sends.

The third is against the rate calculations which were using the total produced signals from the run, but computing the rate from the observation time interval only. This means that we artificially inflated the overall rate by making it look like it took a shorter time to achieve than it did. We correct this by computing the rate for the observation window only. 

Other small fixes:

- Added two production modes for the fake data generator which can do either a smooth production loop or an open loop in case the interval to do a smooth loop is too small.
- Updated some try_from implementations.

* Closes #2713

Ran the benchmarks locally and we can see that the produced and received rates are _very_ close to 100klrps now as opposed to 120k.

image

New fake data generator `production_mode` setting.

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/config.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/config.rs
@@ -404,6 +404,20 @@ impl TrafficConfig {
     pub const fn num_data_points_per_metric(&self) -> Option<usize> {
         self.num_data_points_per_metric
     }
+
+    /// Validate semantic invariants that serde cannot express.
+    ///
+    /// Currently checks that at least one signal weight is non-zero so the
+    /// producer has something to generate.
+    pub fn validate(&self) -> Result<(), otap_df_config::error::Error> {
+        if self.metric_weight + self.trace_weight + self.log_weight == 0 {
+            return Err(otap_df_config::error::Error::InvalidUserConfig {
+                error: "at least one of metric_weight, trace_weight, or log_weight must be > 0"
+                    .to_string(),
+            });
+        }
+        Ok(())
+    }
 }
 
 const fn default_signals_per_second() -> Option<usize> {
@@ -758,5 +772,30 @@ mod tests {
             Some(&None),
             "null value should parse as None"
         );
+    }
+
+    #[test]
+    fn validate_rejects_all_zero_weights() {
+        let cfg = super::TrafficConfig::new(Some(10), None, 100, 0, 0, 0);
+        let result = cfg.validate();
+        assert!(
+            result.is_err(),
+            "all-zero signal weights should be rejected"
+        );
+    }
+
+    #[test]
+    fn validate_accepts_at_least_one_nonzero_weight() {
+        let cfg = super::TrafficConfig::new(Some(10), None, 100, 0, 0, 1);
+        cfg.validate().expect("one non-zero weight should pass");
+
+        let cfg = super::TrafficConfig::new(Some(10), None, 100, 1, 0, 0);
+        cfg.validate().expect("one non-zero weight should pass");
+
+        let cfg = super::TrafficConfig::new(Some(10), None, 100, 0, 1, 0);
+        cfg.validate().expect("one non-zero weight should pass");
+
+        let cfg = super::TrafficConfig::new(Some(10), None, 100, 1, 1, 1);
+        cfg.validate().expect("all non-zero weights should pass");
     }
 }
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/config.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/config.rs
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! Implementation of the configuration of the fake signal receiver
-//!
 
 use serde::de::Deserializer;
 use serde::{Deserialize, Serialize};
@@ -49,14 +48,21 @@ pub enum GenerationStrategy {
     /// - Timestamps and IDs will repeat (stale)
     /// - Lowest CPU cost, maximum throughput
     PreGenerated,
+    // Future: Templates variant — pre-generate signal templates, clone and update
+    // timestamps/IDs per batch for moderate CPU cost with fresh data per batch.
+    // Not yet implemented.
+    // Templates,
+}
 
-    /// Pre-generate signal templates, clone and update timestamps/IDs per batch
-    /// - Templates cloned per batch
-    /// - Fresh timestamps and unique IDs
-    /// - Moderate CPU cost, good balance
-    ///
-    /// TODO: Not yet implemented - currently behaves like Fresh
-    Templates,
+/// Controls how signal batches are paced within each one-second production run.
+#[derive(Clone, Debug, Default, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum ProductionMode {
+    /// Spread batches evenly across the second using a fixed-interval ticker.
+    #[default]
+    Smooth,
+    /// Produce all batches as fast as possible without pacing.
+    Open,
 }
 
 /// A single resource-attribute set with an optional batch weight.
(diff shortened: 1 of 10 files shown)


rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/config.rs
```diff
@@ -49,14 +48,21 @@ pub enum GenerationStrategy {
     /// - Timestamps and IDs will repeat (stale)
     /// - Lowest CPU cost, maximum throughput
     PreGenerated,
+    // Future: Templates variant — pre-generate signal templates, clone and update
```
JakeDern, line 51: I commented out this variant since it's not implemented and doesn't work. I think we can uncomment it when we actually implement it.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/config.rs
```diff
@@ -49,14 +48,21 @@ pub enum GenerationStrategy {
...
-    Templates,
+/// Controls how signal batches are paced within each one-second production run.
+#[derive(Clone, Debug, Default, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(rename_all = "snake_case")]
+pub enum ProductionMode {
+    /// Spread batches evenly across the second using a fixed-interval ticker.
+    #[default]
+    Smooth,
+    /// Produce all batches as fast as possible without pacing.
+    Open,
```
lquerel, line 65: Nice

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
```diff
@@ -108,267 +106,154 @@ impl FakeGeneratorReceiver {
...
-        }
-    }
+    async fn run_open(
+        mut self: Box<Self>,
+        mut ctrl_msg_recv: local::ControlChannel<OtapPdata>,
+        handler: &local::EffectHandler<OtapPdata>,
+        mut producer: TrafficProducer,
+        mut run_ticker: Interval,
+    ) -> Result<TerminalState, Error> {
+        'start: loop {
```
lquerel, line 173: Long time, I didn't see this syntax used!

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
```diff
@@ -108,267 +106,154 @@ impl FakeGeneratorReceiver {
...
+            // This is the start of each traffic run which has two phases
+            let mut current_run = producer.next_run();
+
+            // First phase is the open export phase where we pump data as fast as
+            // possible one chunk at a time while checking for control messages
+            // inbetween. If we hit backpressure then we just skip that chunk of data
```
lquerel:
```suggestion
            // in between. If we hit backpressure then we just skip that chunk of data
```

tools/pipeline_perf_test/test_suites/integration/configs/integration_report_traces.yaml
```diff
@@ -139,30 +139,31 @@ queries:
             ON wr.timestamp BETWEEN ow.start_ts AND ow.stop_ts
         ),
 
-        full_range_aggregates AS (
+        observation_window_aggregates AS (
```
JakeDern, line 142: We were taking counts from the entire run and treating them as though it ran for just the observation window which artificially shrank the time and inflated the throughput metrics

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
JakeDern: Yeah it's rare but sometimes the situation calls for it. I think it's a nice way to express certain control flow patterns, this one included

lquerel (approved):
LGTM

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
```diff
@@ -108,212 +107,153 @@ impl FakeGeneratorReceiver {
...
+
+                        continue 'start;
+                    }
 
-    /// Generate OTLP logs
-    fn generate_logs(&self, count: usize, batch_index: u64) -> OtlpProtoMessage {
-        match self {
-            SignalGenerator::SemanticConventions(registry) => {
-                OtlpProtoMessage::Logs(semconv_signal::semconv_otlp_logs(count, registry))
+                    else => {
```
ThomsonTan: The `else =>` arm here seems unreachable because both other branches use irrefutable patterns.

JakeDern: Ah thanks - I hadn't quite gotten around to testing that part yet and it doesn't behave how I thought. I was looking for `std::future::ready(())`

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/config.rs
cijothomas: thanks!

tools/pipeline_perf_test/test_suites/integration/configs/integration_report_metrics.yaml
```diff
@@ -139,30 +139,31 @@ queries:
...
-            metric_name,
-            core_id,
-            MAX(value) AS max_value,
-            MIN(value) AS min_value
-          FROM with_rate
-          GROUP BY component_name, metric_name, core_id
+            wr.component_name,
+            wr.metric_name,
+            wr.core_id,
+            MAX(wr.value) - MIN(wr.value) AS total_increase
```
lalitb, line 147: Can you double-check this? I may be missing something, but since `value` is cumulative, I think `MAX(wr.value) - MIN(wr.value)` can undercount by dropping the first in-window delta.

lalitb: Same pattern for logs and traces too.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
lalitb: My C/C++ background gave me mild `goto` flashbacks ;)

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
```diff
@@ -98,222 +98,228 @@ impl FakeGeneratorReceiver {
...
+                    biased;
+
+                    msg = ctrl_msg_recv.recv() => {
+                        if let Some(terminal) = handle_control_msg(msg, handler, &mut self.metrics).await? {
+                            return Ok(terminal);
+                        }
+                    }
+
+                    _ = run_ticker.tick() => {
+                        otel_debug!(
```
jmacd, line 204: This is very nice. I would like to see (somehow) a mechanism to log this at info level every-so-often.

jmacd (approved):

tools/pipeline_perf_test/test_suites/integration/configs/integration_report_metrics.yaml
JakeDern: Could you help me understand a bit what you mean by "dropping the first in-window delta"?

I think we are prone to undercounting because `MIN(wr.value)` was going to be scraped sometime after the observation window start and `MAX(wr.value)` would similarly be sometime before the end. 

If that's what you mean, then I agree and I think we could instead try to compute the rate using the time associated with the min/max values so it's on a direct boundary. 

Though I think we might still get a bit of error with this because the engine also aggregates every 1s internally and we might have scraped `MAX(wr.value)` just before it updated.

lalitb: Yes, that boundary undercount is what I mean, but specifically on the left edge of the window.


  Because `value` is cumulative, the first row inside the observation window is not a zero/baseline for the window. `MAX(value) - MIN(value)` effectively discards the delta represented by the first in-window sample. For example, if the samples are:

t=9s value=1000
  t=10s value=2000   -- first sample inside the observation window
  t=11s value=3000

  then `MAX - MIN` over the window gives 1000, but the rows ending at 10s and 11s represent two increments of 1000 each. Depending on where the true observation boundary falls, the first delta may need interpolation/proration, but dropping it entirely undercounts.

JakeDern: Ok yeah, we're totally aligned on that. I have further fine tuning to the metrics and reports that need to be done right after this: https://github.com/open-telemetry/otel-arrow/issues/2731.

Would you mind if I defer this update to the next PR? This calculation at least brings us much closer since we're overcounting by ~20% right now 😄

lalitb: Make sense. Thanks.

lalitb (approved):

albertlockett (approved):

PR #2727, opened 2026-04-21
feat(config): add optional field to EngineConfig

### LINKED ISSUE #2561: Allow applications using df-engine as a library to carry their own engine-level config
### Problem
 
 The engine config is intentionally strict - any unknown field in the YAML is rejected at parse time [deny_unknown_fields](
 https://github.com/open-telemetry/otel-arrow/blob/c66317f995e9b14e02fa6e4eab428594a70f685d/rust/otap-dataflow/crates/config/src/engine.rs#L60). The [configuration model docs](
 https://github.com/open-telemetry/otel-arrow/blob/c66317f995e9b14e02fa6e4eab428594a70f685d/rust/otap-dataflow/docs/configuration-model.md#L155-L157) acknowledge this and mention it could be relaxed if there are good use cases for extensibility. This is one such use case - 

 If you build a collector binary that embeds the dataflow engine and need your own engine-level configuration - for remote management, auth, fleet coordination, etc. - there's no clean way to carry it today. You either:
 
 1. Fork the config crate and add fields directly - creates a maintenance burden and prevents using otel-arrow as a clean dependency.
 2. Pre-parse the YAML, strip your custom keys, then feed cleaned YAML to `OtelDataflowSpec::from_yaml()` - fragile and breaks if the config format evolves.
 
 Node-level configs don't have this problem. They use [`serde_json::Value`](https://github.com/open-telemetry/otel-arrow/blob/c66317f995e9b14e02fa6e4eab428594a70f685d/rust/otap-dataflow/crates/config/src/node.rs#L59), so each node factory handles its own typed config independently. But there's no equivalent escape hatch at the engine level.
 
 ### Proposal
 
 Add a single optional field to `EngineConfig`:
 
 ```rust
 /// Opaque metadata for applications that embed the dataflow engine.
 /// The engine ignores this field entirely — embedding binaries can
 /// read namespaced keys for their own engine-level concerns.
 #[serde(default, skip_serializing_if = "HashMap::is_empty")]
 pub consumer: HashMap<String, serde_json::Value>,
```

This lets embedding applications place their own config under engine.consumer:

```yaml
 engine:
   http_admin:
     bind_address: "127.0.0.1:8080"
   consumer:
     remote_management:
       server_url: "ws://mgmt.example.com/v1"
       heartbeat_interval_secs: 10
     custom_auth:
       provider: "oidc"
       token_endpoint: "https://auth.example.com/token"
```

The engine doesn't interpret any of it. The embedding binary reads what it needs:

```rust
 if let Some(mgmt_cfg) = engine_cfg.engine.consumer.get("remote_management") {
     let settings: MyMgmtSettings = serde_json::from_value(mgmt_cfg.clone())?;
     // ...
 }
```

## Use case

Custom collector binaries that embed the dataflow engine and need engine-level config for concerns the engine itself doesn't handle - remote management, authentication, fleet coordination, custom telemetry routing, etc.

pritishnahar95:
Allow applications embedding the dataflow engine to carry their own engine-level configuration under `engine.custom`. The engine ignores this field entirely; embedding binaries can read namespaced keys for concerns like remote management, auth, or fleet coordination.

# Change Summary

Add an optional `custom: HashMap<String, serde_json::Value>` field to `EngineConfig`. This gives embedding applications an escape hatch for engine-level config without forking the config crate or pre-parsing YAML. The field defaults to an empty map and is omitted from serialized output when empty.

* Closes #2561

Three new unit tests in `engine.rs`:
- `from_yaml_accepts_custom_config` — parses a config with multiple namespaced custom keys and verifies values
- `custom_defaults_to_empty` — confirms the field defaults to an empty map when omitted
- `custom_roundtrips_through_json` — serializes to JSON and deserializes back, verifying data is preserved

Yes. A new optional `custom` key is available under the `engine` section of the YAML/JSON config. Example:

```yaml
engine:
  custom:
    remote_management:
      server_url: "ws://mgmt.example.com/v1"
      heartbeat_interval_secs: 10
    custom_auth:
      provider: "oidc"
      token_endpoint: "https://auth.example.com/token"
```

Existing configs are unaffected since the field defaults to empty.

### DIFF
--- a/rust/otap-dataflow/crates/config/src/engine.rs
+++ b/rust/otap-dataflow/crates/config/src/engine.rs
@@ -18,6 +18,7 @@ use crate::policy::{ChannelCapacityPolicy, Policies, TelemetryPolicy};
 use crate::topic::{TopicImplSelectionPolicy, TopicSpec};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
+use serde_json::Value;
 use std::collections::HashMap;
 
 pub use self::resolve::{
@@ -77,6 +78,15 @@ pub struct EngineConfig {
     /// Engine observability declarations.
     #[serde(default)]
     pub observability: EngineObservabilityConfig,
+
+    /// Opaque metadata for applications that embed the dataflow engine.
+    ///
+    /// The engine ignores this field entirely — embedding binaries can
+    /// read namespaced keys for their own engine-level concerns
+    /// (remote management, auth, fleet coordination, etc.).
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    #[schemars(extend("x-kubernetes-preserve-unknown-fields" = true))]
+    pub custom: HashMap<String, Value>,
 }
 
 /// Engine-wide topic runtime settings.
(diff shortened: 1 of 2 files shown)


lquerel (approved):
LGTM

rust/otap-dataflow/crates/config/src/engine.rs
```diff
@@ -77,6 +78,15 @@ pub struct EngineConfig {
...
     #[serde(default)]
     pub observability: EngineObservabilityConfig,
+
+    /// Opaque metadata for applications that embed the dataflow engine.
+    ///
+    /// The engine ignores this field entirely — embedding binaries can
+    /// read namespaced keys for their own engine-level concerns
+    /// (remote management, auth, fleet coordination, etc.).
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    #[schemars(extend("x-kubernetes-preserve-unknown-fields" = true))]
```
lalitb, line 88: Do we actually need Kubernetes-specific schema metadata here, or should `engine.custom` stay backend/platform agnostic?

lalitb (approved):
LGTM, with non-blocking comment.

rust/otap-dataflow/crates/config/src/engine.rs
pritishnahar95: The annotation is a no-op outside Kubernetes -- it only affects the JSON Schema output generated by `schemars`. Without it, if the schema is ever used as a K8s CRD, the API server would reject arbitrary nested keys under `custom` as unknown fields. The same annotation is already used on [node.config: Value](https://github.com/open-telemetry/otel-arrow/blob/c66317f995e9b14e02fa6e4eab428594a70f685d/rust/otap-dataflow/crates/config/src/node.rs#L58) for the same reason, with a comment explaining why. Happy to drop it if we want to defer K8s compatibility, but it's zero-cost to include now and avoids a follow-up fix later.

lalitb: fair enough. thanks.

pritishnahar95:
Check: "[Rust-CI / build_nonrequired (otap-dataflow, ubuntu-24.04-arm) (pull_request)](https://github.com/open-telemetry/otel-arrow/actions/runs/24800579031/job/72581792731?pr=2727)" has been failing repeatedly for this PR as well as other PRs. Failure does not look to be related to this PR, does not show any logs

image

image

See a few similar failures in other PRs:
- https://github.com/open-telemetry/otel-arrow/actions/runs/24669111757/job/72134618483?pr=2710
- https://github.com/open-telemetry/otel-arrow/actions/runs/24761298125/job/72445472340?pr=2717
- https://github.com/open-telemetry/otel-arrow/actions/runs/24697600492/job/72233676636?pr=2719

For the last job linked here, I was able to see logs. Exit code 143 = SIGTERM (the process was killed). The job ran for 50 minutes and was terminated — it seems to have hit the CI timeout.
The build was still compiling (you can see it got through many crates but hadn't finished). 

image

Can we have the PR checked in skipping past the unrelated issue seen in other PRs too?

jmacd (approved):

PR #2732, opened 2026-04-22
feat(engine): Extension System - Capability Registry & Resolver

gouslu:
# Extension system — Phase 1 (capabilities, registry, builder, proc macro)

Implements the Phase-1 extension/capability system for the OTAP dataflow
engine. Extensions are first-class config siblings of nodes; nodes
explicitly bind to extension instances via named capabilities, and
receive typed handles resolved once at factory time — no hot-path
registry lookups.

Tracking docs:
[`docs/extension-system-architecture.md`](rust/otap-dataflow/docs/extension-system-architecture.md)
(rewritten in this PR).

### `#[capability]` proc macro (`engine-macros`)

- New `capability.rs` expansion: from a single `#[capability] trait
  Foo { ... }` source it generates `local::Foo` (`!Send`-friendly) and
  `shared::Foo` (`Send + Clone`) trait variants plus a `SharedAsLocal`
  adapter and an `ExtensionCapability` impl. The dual variants are
  derived from one source, so authors can't accidentally let local
  and shared semantics diverge.
- New `pipeline_factory.rs` expansion to build the static
  `PipelineFactory` registry used by `main.rs`.
- All emitted paths use fully-qualified `::std::...` /
  `::async_trait::...` / `::otap_df_engine::...` so generated code is
  hygienic in any caller crate.

### Capability registry (`engine::capability`)

- `CapabilityRegistry`: typed-keyed (`(extension_name, TypeId)`)
  storage with **typestate-enforced** single `.shared()` / `.local()`
  registration per builder — duplicates are unrepresentable rather
  than runtime errors.
- Two execution models: native local (`Rc<dyn Local>`, lock-free) and
  native shared (`Box<dyn Shared>`, `Send + Clone`). A shared-only
  extension serves local consumers transparently via the
  `SharedAsLocal` adapter generated by the proc macro.
- Two **instance policies** chosen at build time, invisible to
  consumers: `.cloned()` (clone a stored prototype) and
  `.constructed()` (per-consumer construction via a closure;
  Passive-only — `Active + Constructed` is statically rejected).
- `resolve_bindings`: walks a node's `capabilities:` declaration and
  produces a per-node `Capabilities` bundle with all bindings
  resolved, surfacing config errors (unknown extension, unknown
  capability, capability not provided by bound extension, multiple
  bindings for the same capability).
- `Capabilities`: per-node consumer API with `require_local`,
  `require_shared`, `optional_local`, `optional_shared`. Instances
  are minted lazily at the call site, not at resolution time.
- `ConsumedTracker`: cross-node, per-(capability, extension)
  consumption flags driving `drop_local()` / `drop_shared()` cleanup
  for extensions no node ever claimed.

### One-shot consumption contract

A binding is claimable **at most once per node**, regardless of
execution model. The guard is the `Cell<Option<_>>::take()` on each
resolved entry's `produce` closure — no auxiliary flag.

- Same accessor twice → `CapabilityAlreadyConsumed`.
- Different accessors on a SharedAsLocal-fallback binding share one
  underlying entry, so claiming either side consumes the other
  naturally.
- Different accessors on a native-dual binding (extension registered
  both native local **and** native shared) take and drop the
  alternative entry's `produce` closure on success, so the
  per-binding contract holds uniformly. The cross-node tracker is
  only flipped by actual consumption, not by invalidation, so
  `drop_*` cleanup remains correct.

### Documentation

- `docs/extension-system-architecture.md`: rewritten to describe the
  capability-based design, the local/shared duality, instance
  policies, Active vs Passive lifecycle, and the typestate builder.

## Tests

New, focused unit tests cover:

- Registry: typestate single-registration, duplicate rejection,
  `SharedAsLocal` adapter freshness per node, double-`Box` envelope
  for shared `produce`.
- `resolve_bindings`: every error path (unknown extension / unknown
  capability / capability not provided / wrong extension), local-only
  and shared-only binding shapes, fallback path, native-dual path.
- One-shot contract: second-call rejection on each of `require_local`,
  `require_shared`, `optional_local`, `optional_shared`; fallback
  cross-side rejection; native-dual cross-side rejection (both
  directions).
- `ConsumedTracker`: per-extension consumption flags, with the
  invariant that mere invalidation does not flip a bucket.
- Proc-macro end-to-end: `local-only`, `shared-only`, and `dual`
  forms of `extension_capabilities!` against the registry.

## Validation

```text
cargo xtask check
✅ Cargo workspace structure complies with project policies.
✅ Formatting completed successfully.
✅ Clippy linting passed without warnings.
✅ All tests passed successfully.
```


rust/otap-dataflow/crates/engine-macros/src/capability.rs
```diff
@@ -0,0 +1,439 @@
...
+//! - Visibility modifiers on the trait
+//!
+//! # Unsupported (rejected with compile-time errors)
+//!
+//! These limitations are fundamental to the type-erased `HashMap<TypeId, Entry>`
+//! registry design and the `SharedAsLocal` adapter delegation pattern:
+//!
+//! - **Trait-level generics / lifetimes** — `TypeId::of::<T>()` requires a
+//!   concrete monomorphized type; a generic registration struct cannot produce
+//!   a single `TypeId` for the registry.
```
jmacd: I can't follow this or what it means as a limitation.

rust/otap-dataflow/crates/engine/src/capability/registry/entry.rs
```diff
@@ -0,0 +1,171 @@
...
+// ── Resolved entries (per-node) ─────────────────────────────────────────────
+
+/// A resolved local capability entry for a specific node.
+pub(crate) struct ResolvedLocalEntry {
+    /// Per-node produce closure, cloned from the registry entry.
+    pub(crate) produce: Box<dyn LocalProduce>,
+    /// Per-node one-shot consumption flag for the consumer-side
+    /// "claim each binding at most once per node" contract enforced
+    /// by [`Capabilities`](super::Capabilities).
+    pub(crate) claimed: Cell<bool>,
```
jmacd: Curiosity question: can we use `Cell<Option<Box<dyn LocalProduce>>>` to encode the claimed state as Some vs None?

rust/otap-dataflow/docs/extension-system-architecture.md
```diff
@@ -148,35 +156,54 @@ themselves, and they never touch pipeline data directly.
...
+   `.active()` stage only accepts types implementing
+   `Extension`; the `.passive()` stage does not require
+   that trait.
+
+4. **Instance policy (Passive only).** After `.passive()`,
+   the extension author picks an instance policy. The
+   choice is **provider-side only** -- capability
+   consumers call `require_shared()` / `require_local()`
+   and cannot observe which policy was used.
+   - `.cloned()` -- each consumer receives an independent
```
lalitb: This wording seems too broad for the local case. Shared `.cloned()` clones the stored value per consumer, but local `.cloned()` clones the Rc handle, so local consumers in the same pipeline instance share one underlying extension object.

rust/otap-dataflow/crates/engine/src/extension/builder.rs
```diff
@@ -0,0 +1,469 @@
...
+
+impl ActiveStage {
+    /// Register the shared (Send) variant.
+    #[must_use]
+    pub fn shared<E>(mut self, extension: E) -> Self
+    where
+        E: shared_ext::Extension + Clone + Send + 'static,
+    {
+        debug_assert!(
+            self.parent.shared.is_none(),
```
ThomsonTan: This invariant check disappears in release builds. A second `.shared(...)` or `.local(...)` overwrites the first side silently. Consider make this a real invariant.

rust/otap-dataflow/crates/engine/src/capability/registry/capabilities.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// Resolve a **required** local capability.
+    ///
+    /// Returns `Rc<dyn C::Local>`. If the capability was registered by a
+    /// shared-only extension, the `SharedAsLocal` adapter is returned
+    /// transparently — the caller always gets a local trait object.
+    ///
+    /// # One-shot contract
+    ///
+    /// Each capability binding may be claimed **at most once per
+    /// node** across all four accessors (`require_local`,
```
lalitb: This one-shot contract wording seems slightly stricter than the  implementation. The docs say a capability can be claimed once per node across all accessors, but the shared-to-local fallback appears to allow one local claim and one shared claim for the same binding. That may be the intended behavior, but the docs/error text should probably describe it explicitly so node authors know what to expect.

rust/otap-dataflow/crates/engine/src/extension/builder.rs
```diff
@@ -0,0 +1,469 @@
...
+            "ExtensionBundleBuilder: .shared(...) called more than once",
+        );
```
lalitb: Minor builder API question: duplicate `.shared(...)` / `.local(...)` calls currently panic only in debug builds and silently use the last registration in release builds. Since this is construction-time behavior, should duplicates be recorded and returned as a deterministic `.build()` error instead? That would make misuse behave the same in debug and release without affecting the data hot path.

lalitb (approved):
LGTM. I left a few comments around docs/API-contract clarity and one construction-time builder edge case. Please address or acknowledge those before merge.

rust/otap-dataflow/crates/engine/src/extension/wrapper.rs
```diff
@@ -0,0 +1,618 @@
...
+    ///
+    /// # Errors
+    ///
+    /// Propagates any error returned by the capability registry
+    /// (e.g. duplicate `(capability, extension)` registration).
+    pub fn register_into(
+        &self,
+        capabilities: &crate::capability::ExtensionCapabilities,
+        registry: &mut crate::capability::registry::CapabilityRegistry,
+    ) -> Result<(), crate::capability::registry::Error> {
```
ThomsonTan, line 601: `register_into()` should fail fast when `ExtensionCapabilities` advertises capabilities for an execution-model side that the returned `ExtensionBundle` does not contain.

Right now, if `capabilities.shared` is non-empty but `self.shared` is `None` (or the same for local), registration silently skips that side. That turns an extension-author bug into a later config-resolution error like “extension does not provide capability”, which points reviewers/users at the wrong layer.

Can we add explicit checks before the `if let Some(shared/local)` blocks and return `InternalError` for metadata/runtime mismatch? Shared-only with local fallback remains valid; this only rejects advertising a side that was not actually built.

rust/otap-dataflow/docs/extension-system-architecture.md
gouslu: Authors know what they pass as a value when the choose cloned, that makes the cloning behavior explicit to them based on whether they choose local or shared. What they pass into local is Rc<> whereas the other is a type that derives clone. What you pass in as a value is what gets cloned.

rust/otap-dataflow/crates/engine/src/capability/registry/capabilities.rs
gouslu: Intended behavior is if you call require_local, you can't call require_shared on the same trait again. The idea is that you can get each capability exactly once, regardless of whether it is local or shared. So, I think what I need to look into is whether this is enforced consistently or not. Will investigate.

rust/otap-dataflow/docs/extension-system-architecture.md
lalitb: That makes sense. My only point is that the doc wording was a bit confusing to me because “independent clone of a stored prototype” can sound like the underlying local object is cloned too. Maybe just clarify that .cloned() clones the value passed to the builder, so for local that means cloning the Rc handle.

rust/otap-dataflow/crates/engine/src/capability/registry/capabilities.rs
lalitb: That sounds good. If the intended contract is exactly once per capability regardless of local/shared, then my read is that the fallback path may not enforce it consistently today because local and shared resolved entries appear to have separate claimed cells. Thanks for checking.

rust/otap-dataflow/docs/extension-system-architecture.md
```diff
@@ -958,12 +1154,18 @@ Add type-safe capability resolution:
...
-  `drop_local()` / `drop_shared()`
+- `SharedAsLocal` adapter (deferred, per-node) with
+  fallback-bucket tracker accounting
+- Two-level consumption tracking: per-node
+  `claimed: Cell<bool>` on each resolved entry, plus
+  `tracker_consumed: Rc<Cell<bool>>` feeding the
+  registry-wide `ConsumedTracker`
 - `KNOWN_CAPABILITIES` link-time registration
 - Sealed traits + `ExtensionCapability` with
   `Local`/`Shared` associated types
```
jmacd: Nothing is wrong, but IMO this document is overly detailed. I would cut to around line 300, everything before "### Import Paths" is good and worth reading! 📖

jmacd (approved):

rust/otap-dataflow/crates/engine/src/extension/wrapper.rs
gouslu: builder now collects errors and reports at build()

rust/otap-dataflow/crates/engine/src/extension/builder.rs
gouslu: Typestate builder now prevents it

rust/otap-dataflow/crates/engine/src/capability/registry/capabilities.rs
gouslu: fixed

rust/otap-dataflow/crates/engine/src/extension/builder.rs
gouslu: typestate builder now prevents a second shared/local

rust/otap-dataflow/crates/engine/src/capability/registry/entry.rs
gouslu: fixed this with a bigger change that makes require local go through require shared in the fallback mode; because both fixing the bug lalit raised and doing this required that bigger change.

rust/otap-dataflow/crates/engine/src/capability/registry/tracker.rs
```diff
@@ -0,0 +1,131 @@
...
+use std::cell::Cell;
+use std::collections::HashMap;
+use std::rc::Rc;
+
+/// Tracks which capability variants were consumed by node factories.
+///
+/// Created alongside per-node [`Capabilities`](super::Capabilities)
+/// during [`resolve_bindings`](super::resolve_bindings). After all node
+/// factories have run, the engine inspects this tracker to determine
+/// which extension variants are unused and can be dropped.
```
lquerel: That may be true, but are we certain that an extension has no usefulness when it is not associated with any node? Couldn't we imagine an extension that provides a global-level service? I don't have a concrete example in mind right now, but I'm raising the question here to get reactions.

rust/otap-dataflow/crates/engine/src/capability/registry/tracker.rs
```diff
@@ -0,0 +1,131 @@
...
+///
+/// Keyed by `(capability TypeId, extension ID)` so multiple providers of
+/// the same capability are tracked independently. The `Rc<Cell<bool>>`
+/// for a given key is shared across all nodes that bind to that
+/// provider — once any of them consumes the capability the cell is set.
```
lquerel: And also the same extension can appears multiple times in this list as a single extension can expose multiple capabilities. Right?

rust/otap-dataflow/crates/engine/src/capability/registry/resolve.rs
```diff
@@ -0,0 +1,167 @@
...
+    registry: &CapabilityRegistry,
+    known_extensions: &HashSet<ExtensionId>,
+    tracker: &mut ConsumedTracker,
+) -> Result<Capabilities, Error> {
+    // Build lookup from capability name → KnownCapability
+    let known_caps: HashMap<&str, &crate::capability::KnownCapability> =
+        crate::capability::KNOWN_CAPABILITIES
+            .iter()
+            .map(|kc| (kc.name, kc))
+            .collect();
```
lquerel, line 50: Shouldn't we detect cases where two extensions are declared with the same names? That shouldn't happen, but I'm not actually sure anything prevents it.

rust/otap-dataflow/crates/engine/src/capability/registry/tracker.rs
gouslu: that's right, I think we should add it in this PR. will make that update.

rust/otap-dataflow/crates/engine/src/capability/registry/tracker.rs
gouslu: supposed to be that way, let me verify.

gouslu: verified, will add a comment to clarify.

rust/otap-dataflow/crates/engine/src/capability/registry/resolve.rs
gouslu: it is rejected at config layer, but can add a comment here.

lquerel (approved):
LGTM. Thanks

PR #2733, opened 2026-04-22
Validation framework - transport header support

### LINKED ISSUE #2704: support transport headers with the validation framework
### Pre-filing checklist

### Component(s)

Go OTel Arrow Receiver/Exporter (go/)

### Objective

Update the validation framework to allow for validating transport headers and to allow users to configure the use of transport headers for testing.

Users should be able to specify transport headers to use for a Generator pipeline (map to fake data generator).
 
By default the Capture pipeline should have a transport header policy defined that accepts all headers.

Users should also be able to configure transport header policies (pipeline level) for their SUV pipeline (of course they can also directly specify these policies at the node level in their pipeline config file)

### Rationale

Currently the validation framework has no way of integrating transport headers in the tests and has no way of inspecting the contents of transport headers to validate the the presence of specified key values

### Scope

- Update Generator struct to allow user to configure transport headers to insert into the test
- Update Generator template to insert transport header configuration into fake data generator
- Update Capture template with default capture header policy (no overrides, accepts all headers)
- Add a new ValidationInstruction that checks transport headers for specific keys and key/value pairs
- Update Pipeline struct to allow user to configure transport header policy
- Update validation template to insert transport header policy for SUV pipeline

### Acceptance Criteria

- Can run a validation test with defined transport headers being added and can confirm they exist via transport header ValidationInstruction

### Dependencies or Blockers

_No response_

### Additional Context

_No response_

c1ly:
# Change Summary

- Updated Pipeline, Capture, Generator structs with transport header fields and new functions to allow user to configure transport headers and transport header policies to use

- Updated templates to insert transport headers and transport header policy

* Closes #2704

unit tests and validation test

no


rust/otap-dataflow/crates/validation/src/validation_exporter.rs
```diff
@@ -213,15 +232,17 @@ impl Exporter<OtapPdata> for ValidationExporter {
...
                             } else if self.control_indices.contains(&node_index) {
                                 self.control_msgs.push(msg);
-                                self.validate_and_record();
+                                // self.validate_and_record();
                             }
                         } else if self.control_indices.is_empty() {
                             self.suv_msgs.push((msg, time_elapsed));
-                            self.validate_and_record();
+                            self.suv_transport_headers.push(transport_headers);
+                            // self.validate_and_record();
```
jmacd: Is this intentional?

c1ly: oop meant to delete that commented out line. Yes those commented out self.validate_and_record() calls aren't necessary since we call already call self.validate_and_record() during the CollectTelemetry stage

rust/otap-dataflow/crates/validation/src/validation_exporter.rs
```diff
@@ -113,10 +118,22 @@ pub static VALIDATION_EXPORTER_FACTORY: ExporterFactory<OtapPdata> = ExporterFac
 
 impl ValidationExporter {
     /// Run the configured validations and update metrics.
+    ///
+    /// The `OtlpProtoMessage` projection is built once here so that
+    /// multiple [`ValidationInstructions`] can share it without
+    /// redundant cloning.
     fn validate_and_record(&mut self) {
+        let suv_msgs: Vec<OtlpProtoMessage> =
+            self.suv_msgs.iter().map(|(msg, _)| msg.clone()).collect();
```
albertlockett: nit: it seems like the comment being added here is documentation for the implementation, not the method. Maybe it should be moved inside the function definition?
```suggestion
    fn validate_and_record(&mut self) {
        // The `OtlpProtoMessage` projection is built once here so that
        // multiple [`ValidationInstructions`] can share it without
        // redundant cloning.
        let suv_msgs: Vec<OtlpProtoMessage> =
            self.suv_msgs.iter().map(|(msg, _)| msg.clone()).collect();
```

rust/otap-dataflow/crates/validation/README.md
```diff
@@ -446,6 +479,97 @@ it should receive control signals from
...
   - each generator label must match a label passed to `add_generator`
 
+### Transport Headers
+
+Validate that transport headers survive the full pipeline chain from generator
+through the system-under-validation to the capture pipeline.
+
+> **NOTE:** Only OTLP (`otlp_grpc`) receivers and exporters currently support
+> transport header capture and propagation. OTAP receivers/exporters do **not**
+> support transport headers.
```
albertlockett: We could support transport header configuration eventually right? e.g. using the `headers` field on `BatchArrowRecords`? https://github.com/open-telemetry/otel-arrow/blob/bc4cec8b9dd31adbf3a207903eb652c16499467c/proto/opentelemetry/proto/experimental/arrow/v1/arrow_service.proto#L52-L61

nit: I wonder if the language should be softened to suggest that we may support transport headers in the future, so as not to give the impression that it's technically infeasible?

```suggestion
> transport header capture and propagation. OTAP receivers/exporters do **not**
> yet support transport headers.
```

rust/otap-dataflow/crates/validation/src/validation_types/mod.rs
```diff
@@ -365,8 +418,38 @@ mod tests {
...
+
+    #[test]
+    fn transport_header_require_key_serialization_check() {
+        let instruction = ValidationInstructions::TransportHeaderRequireKey {
+            keys: vec!["x-tenant-id".into()],
+        };
+        let yaml = serde_yaml::to_string(&instruction).expect("serialize");
+        let back: ValidationInstructions = serde_yaml::from_str(&yaml).expect("deserialize");
+        assert_eq!(back, instruction);
+    }
```
albertlockett, line 449: Should we also call `validate` on these instructions? That way we get coverage coverage over the fact that `validate` dispatches the correct validation function for each variant of `ValidationInstructions`?

albertlockett (commented):
Looks mostly good to me @c1ly. nice addition to the validation test framework.

I had a few minor feedbacks, let me know what you think

lquerel (approved):
LGTM

PR #2735, opened 2026-04-22
bugfix: Rename pending_sends_buffered metrics to disambiguate them.

### LINKED ISSUE #2734: Metric description conflict: duplicate pending_sends_buffered metric across runtime-control and completion scopes
### Pre-filing checklist

### Component(s)

### Bug Description

Description:

When the Prometheus pull exporter is configured, every scrape logs:

"Instrument description conflict, using existing"

Two metric sets in control_plane_metrics.rs declare a metric with the same name pending_sends.buffered but different descriptions:

- pipeline.runtime_control (RuntimeControlMetrics): https://github.com/open-telemetry/otel-arrow/blob/ebf2403b5810741def8e0725e17e3fafbefc2458/rust/otap-dataflow/crates/engine/src/control_plane_metrics.rs#L76
- pipeline.completion (PipelineCompletionMetrics): https://github.com/open-telemetry/otel-arrow/blob/ebf2403b5810741def8e0725e17e3fafbefc2458/rust/otap-dataflow/crates/engine/src/control_plane_metrics.rs#L253

The opentelemetry-prometheus exporter flattens all OTel scopes into a single Prometheus namespace keyed by metric name. When both scopes produce a metric named pending_sends_buffered, the first description seen wins and the second is silently dropped (see validate_metrics in the exporter):

https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-prometheus/src/lib.rs#L452

Although this is only a description conflict (both are Gauge, so no data is lost), it produces a noisy warning on every Prometheus scrape and means one of the two metrics loses its description.

Suggested fix:

Rename the metrics to disambiguate them.

Both introduced in commit 3dca283.

### Steps to Reproduce

Configure a prometheus exporter for internal telemetry and run an instance of dataflow engine. Check the logs.

### Expected Behavior

The conflict to be resolved, so the log message is no longer present.

### Actual Behavior

The log message is present every time a scrapper is called over the prometheus endpoint.

### OTel-Arrow Version

latest

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

andborja:
# Change Summary
Rename pending_sends_buffered metrics to disambiguate them and remove its conflict.

* Closes [#2734](https://github.com/open-telemetry/otel-arrow/issues/2734)

Locally running the collector.

No

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/control_plane_metrics.rs
+++ b/rust/otap-dataflow/crates/engine/src/control_plane_metrics.rs
@@ -73,7 +73,7 @@ pub(crate) struct RuntimeControlMetrics {
     /// full node-control inboxes instead of blocking the runtime.
     ///
     /// Level: `basic`.
-    #[metric(name = "pending_sends.buffered", unit = "{message}")]
+    #[metric(name = "control_pending_sends.buffered", unit = "{message}")]
     pub pending_sends_buffered: Gauge<u64>,
     /// Number of ordinary per-node timers currently registered.
     ///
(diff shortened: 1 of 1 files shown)


cijothomas (changes requested):
I left a comment in the issue. It is *not* a bug to use same metric name in different meter/scopes.

https://github.com/open-telemetry/otel-arrow/issues/2734#issuecomment-4301599621

andborja:
Not able to reproduce the issue. Abandoning this PR.

Issue #2697, opened 2026-04-17
Remove and ban indirect use of openssl-sys crate


jmacd:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Description

The use of `openssl` in as a `crypto-openssl` provider happens indirectly through OTel Weaver and then through Git Oxide (gix). 

On Linux platforms this is no big deal. On Windows, it causes a build from source at around 7 minutes, more than double everything else we have to build in otap-dataflow.

The solution requires a fix in Weaver. Ideally, when Weaver is exposed as a library it will use the no-tls variation of request-blocking; then in the main.rs file, before networking starts, inject the chosen crypto provider.

After we finish this work, we should I think add a ban in deny.toml to prevent the 7 minute OpenSSL build from happening by future mistake.

### Additional Context

https://github.com/open-telemetry/weaver/pull/1302
https://github.com/open-telemetry/weaver/pull/1355

jmacd:
Thanks @sapatrjv for the Weaver fix.

michaelvanstraten:
@jmacd I would prefer if we removed it entirely instead of just windows.

PR #2739, opened 2026-04-23
Use latest weaver build that has the exclusion of openssl build in case of windows platforms. In case of windows platforms it uses SChannel TLS instead of natively building openssl.

sapatrjv:
# Change Summary


sapatrjv:
Part of https://github.com/open-telemetry/otel-arrow/issues/2697.

Issue #2736, opened 2026-04-22
Support OTTL transform statements in the transform processor


marcsnid:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

The transform processor currently supports KQL and OPL query languages (see
`Query` enum in `crates/core-nodes/src/processors/transform_processor/config.rs`).
We need to add support for a subset of
[OTTL (OpenTelemetry Transformation Language)](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/pkg/ottl/LANGUAGE.md)
to allow users to express metric and log transformations using the same syntax
as the OpenTelemetry Collector's `transform` processor.

Users migrating from the OTel Collector `transform` processor need to express
attribute manipulation, pattern extraction, conditional assignment, and value
formatting using OTTL syntax. A minimal viable subset covers the most common
real-world use cases.

### Proposed Solution

Add an OTTL front-end to the existing transform processor by extending the OTTL pest grammar (`rust/experimental/query_engine/ottl-parser/`) to parse a targeted subset of OTTL statements, then lower them to the existing expression AST (`SetTransformExpression`, `ConditionalDataExpression`, `Variable`, `Attached`, etc.) in `rust/otap-dataflow/crates/query-engine/` — adding new AST nodes only for `merge_maps`, `ExtractPatterns`, and `FormatTime`.
Wire this in via a new `OttlQuery` variant on the `Query` enum in `crates/core-nodes/src/processors/transform_processor/config.rs`, so users can configure OTTL statements alongside the existing KQL/OPL options.

### Already have AST equivalents (need OTTL parser to AST mapping)

| Primitive | Existing AST Node | Notes |
|---|---|---|
| `set(target, value) where <cond>` | `ConditionalDataExpression` wrapping `SetTransformExpression` | The OTTL `set(...) where ...` pattern maps directly to a `ConditionalDataExpression` branch containing a `SetTransformExpression`. This is a single conditional-set operation, **not** a separate set + filter. This is the key building block for most statements in the target subset. |
| `set(target, value)` (unconditional) | `SetTransformExpression` | Used when no `where` clause is present. |
| `==`, `!=` | `EqualToLogicalExpression` (+ `Not`) | |
| `>` | `GreaterThanLogicalExpression` | |
| `and` | `AndLogicalExpression` | |
[...]

PR #2746, opened 2026-04-23
[query-engine] Rename log-specific methods in bridge

CodeBlanch:
Relates to #2736

# Changes

* Adds "logs" into names of bridge methods which are specific to logs processing

# Details

Prepping for #2736 which seems like it will require "metrics" APIs in bridge.


albertlockett (approved):
lgtm @CodeBlanch

PR #2747, opened 2026-04-23
Implement pdata-propagated stopwatch timer across processors

drewrelmas:
# Change Summary

Alternative to the full `ProcessorChain` implementation under discussion at #2556

After some conversations, trying to simplify back to focusing on ONLY the composite duration problem in a way that isn't directly tied to the channel elimination strategy / full implementation of `ProcessorChain`

## Motivation

Per-node `processor.compute.*.duration` reports each processor in isolation. Operators also need a single metric for the combined compute of a *range* of processors (e.g. five sequential attribute processors as one number).

Distributed **stopwatches** are engine-managed metrics that sum per-message compute time across configurable processor node pairs on the forward path.

## Design

**What's measured.** The wall-clock time spent inside each processor's `process()` between successive sends. Queue wait and channel latency between nodes are excluded.

**How it's measured.** Immediately before each `process()` call, the engine arms an `Instant` send-marker on the EffectHandler. On every send (`send_message`, `try_send_message`, `_to`, and `_with_source_node*` variants), the unified `ProcessorSendHook` reads the elapsed time since the marker, advances the marker to "now", and adds the delta to the message's on-PData `stopwatch_compute_ns` accumulator. No processor cooperation needed — `timed()` is not required for stopwatch participation, and is used solely to split per-node compute into success/failed.

**Non-overlapping ranges.** At most one stopwatch is active per message. The on-PData accumulator is a single `Option<NonZeroU64>`.

**Declarative YAML config.**

```yaml
policies:
  telemetry:
    runtime_metrics: normal
    stopwatches:
      - name: ingest_pipeline
        start_node: attr1
        stop_node: attr5
```

Each stopwatch gets its own metric entity with `stopwatch.name`, `stopwatch.start_node`, and `stopwatch.stop_node` attributes.

## How It Works

1. **Build time** — `build_stopwatch_state()` resolves config node names to indices, validates that endpoints are processor nodes (local or shared; receivers and exporters are rejected) and that no node is reused across roles. If `runtime_metrics` does not include `PROCESS_DURATION`, all stopwatches are skipped with a single warning. Each processor's `EffectHandler` is assigned its start/stop role and (for stop nodes) the metric set to report into.
2. **Marker arming** — the engine loop calls `effect_handler.begin_process_timing()` before each `process()` call when stopwatches are active.
   - Local: `Rc<Cell<Option<Instant>>>`.
   - Shared: `Arc<Mutex<Option<Instant>>>` (mutex contention is bounded to the per-processor sequential `process()` loop).
3. **Forward-path accumulation** — `OtapPdata::before_processor_send` (the unified `ProcessorSendHook` impl, called by both the plain `send_message*` family and the `send_message_with_source_node*` family on local and shared processor handlers) invokes `stopwatch_accumulate`, which calls `take_elapsed_since_send_marker_ns()` (single read-and-advance step) and:
   - **Start node:** initialises the on-PData accumulator with the 1 ns sentinel.
   - **Any node within an active range:** adds the delta.
   - **Stop node:** takes the total via `take_stopwatch_compute()` and records into the local stop accumulator (`Cell<Mmsc>` for local, `Arc<Mutex<Mmsc>>` for shared).
4. **Periodic reporting** — on `CollectTelemetry` and at shutdown, the engine loop calls `effect_handler.report_stopwatch()`, which drains the stop accumulator into its `MetricSet` and reports it.

## Metrics Produced

| Metric Set  | Metric Name                  | Unit | Attributes                                                                      |
|-------------|------------------------------|------|---------------------------------------------------------------------------------|
| `stopwatch` | `stopwatch.compute.duration` | ns   | `stopwatch.name`, `stopwatch.start_node`, `stopwatch.stop_node`, pipeline attrs |

`Mmsc` (min/max/sum/count), delta temporality.

## Live Validation

Demo: `configs/fake-stopwatch-demo.yaml` (5 attribute processors, one stopwatch spanning the chain).

```bash
curl -s 'http://127.0.0.1:8080/api/v1/telemetry/metrics?format=json&reset=true' | jq '
  [.metric_sets[] | select(.name == "stopwatch" or .name == "processor.compute")]
  | .[] | {
      name: .name,
      identity: (.attributes["node.id"].String // .attributes["stopwatch.name"].String),
      duration: (
        .metrics[]
        | select(
            (.name == "compute.success.duration") or
            (.name == "stopwatch.compute.duration")
          )
        | .value
        | if type == "object" and .count > 0 then {
            avg_ms: (.sum / .count / 1e6),
            count
          } else empty end
      )
    }'
```

The stopwatch average is at least the sum of per-processor `processor.compute` averages — it also covers `process()` time outside `timed()` closures (message handling, send-side bookkeeping):

| Metric              | Node / Stopwatch                  | Avg (ms)  | Count  |
|---------------------|-----------------------------------|-----------|--------|
| `processor.compute` | attr1                             | 1.305     | 40     |
| `processor.compute` | attr2                             | 0.908     | 40     |
| `processor.compute` | attr3                             | 1.207     | 40     |
| `processor.compute` | attr4                             | 0.521     | 40     |
| `processor.compute` | attr5                             | 0.519     | 40     |
| **`stopwatch`**     | **ingest_pipeline** (attr1→attr5) | **6.121** | **40** |

## Limitations / Follow-ups

- **Non-overlapping only.** Build-time validation rejects nodes reused across roles (e.g. `a→b` + `b→c`); runtime detection warns and resets if topologically overlapping ranges (e.g. 1→3 + 2→4) appear.
- **Requires `runtime_metrics: normal` or higher.** Gated behind the same `PROCESS_DURATION` interest as `processor.compute`. Lower levels skip all stopwatches at build time with one warning.
- **Processors only.** Receivers and exporters are rejected at build time (no incoming-message frame / no further send path).
- **No topology ordering validation.** Start-before-stop is not enforced; a misconfigured stopwatch simply never fires.
- **Shared processors with worker tasks.** The marker captures only time spent inside `process()` itself. Processors that fan work out to spawned tasks and return early can still report accurate per-message compute via `timed()` from inside their workers.
- **Send-bounded measurement.** Each processor's contribution is the wall-clock span from `begin_process_timing` (or its previous send) through its next `send_message`. Two consequences:
  - **Dropped messages contribute nothing.** A processor that consumes a message without sending it never reaches a stop node, so no stopwatch observation is recorded — by design, the metric counts only complete transits.
  - **Post-final-send work is not attributed.** Cleanup performed after the last `send_message` in `process()` is not captured. For typical stateless processors this is a trivial return path; processors that do significant post-send work should perform it before the final send if they want it measured.

* Related to #2556
* Contributes to #2782

Unit tests, local run of sample config

Yes, users could configure stopwatch metrics across multiple processors

### DIFF
--- a/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
+++ b/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
@@ -1,15 +1,17 @@
 # Demonstrates the distributed stopwatch feature.
 #
-# Five attribute processors form a sequential processing chain. Two stopwatches
-# measure aggregate synchronous compute duration across different sub-ranges:
+# Five attribute processors form a sequential processing chain. A stopwatch
+# measures aggregate synchronous compute duration across the full chain:
 #
 #   stopwatch "ingest_pipeline" - covers attr1 through attr5 (full chain)
-#   stopwatch "enrich_stage"    - covers attr2 through attr4 (middle three)
 #
 # Stopwatches sum per-message compute.*.duration (the time spent inside timed()
 # closures) across each processor in the range, excluding queue wait and async I/O.
-# Metrics appear as stopwatch.compute.success.duration and
-# stopwatch.compute.failed.duration (MMSC histograms in nanoseconds).
+# Metrics appear as stopwatch.compute.duration (MMSC histograms in
+# nanoseconds).
+#
+# Note: only non-overlapping stopwatch ranges are supported. A message can be
+# inside at most one stopwatch range at a time.
 #
 # Query metrics while running:
 #   curl -s 'http://127.0.0.1:8080/api/v1/telemetry/metrics?format=json' \
@@ -33,9 +35,6 @@ groups:
               - name: ingest_pipeline
                 start_node: attr1
                 stop_node: attr5
-              - name: enrich_stage
-                start_node: attr2
-                stop_node: attr4
 
         nodes:
           receiver:
--- a/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
+++ b/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
@@ -3,8 +3,8 @@
 # Five attribute processors form a sequential processing chain. Two stopwatches
 # measure aggregate synchronous compute duration across different sub-ranges:
 #
-#   stopwatch "ingest_pipeline" — covers attr1 through attr5 (full chain)
-#   stopwatch "enrich_stage"    — covers attr2 through attr4 (middle three)
+#   stopwatch "ingest_pipeline" - covers attr1 through attr5 (full chain)
+#   stopwatch "enrich_stage"    - covers attr2 through attr4 (middle three)
 #
 # Stopwatches sum per-message compute.*.duration (the time spent inside timed()
 # closures) across each processor in the range, excluding queue wait and async I/O.
--- a/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
+++ b/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
@@ -1,0 +1,106 @@
+# Demonstrates the distributed stopwatch feature.
+#
+# Five attribute processors form a sequential processing chain. Two stopwatches
+# measure aggregate synchronous compute duration across different sub-ranges:
+#
+#   stopwatch "ingest_pipeline" — covers attr1 through attr5 (full chain)
+#   stopwatch "enrich_stage"    — covers attr2 through attr4 (middle three)
+#
+# Stopwatches sum per-message compute.*.duration (the time spent inside timed()
+# closures) across each processor in the range, excluding queue wait and async I/O.
+# Metrics appear as stopwatch.compute.success.duration and
+# stopwatch.compute.failed.duration (MMSC histograms in nanoseconds).
+#
+# Query metrics while running:
+#   curl -s 'http://127.0.0.1:8080/api/v1/telemetry/metrics?format=json' \
+#     | jq '[.metric_sets[] | select(.name == "stopwatch" or .name == "processor.compute")]'
+
+version: otel_dataflow/v1
+engine: {}
+groups:
+  default:
+    pipelines:
+      main:
+        policies:
+          channel_capacity:
+            control:
+              node: 100
+              pipeline: 100
+            pdata: 128
+          telemetry:
+            runtime_metrics: normal
+            stopwatches:
+              - name: ingest_pipeline
+                start_node: attr1
+                stop_node: attr5
+              - name: enrich_stage
+                start_node: attr2
+                stop_node: attr4
+
+        nodes:
+          receiver:
+            type: receiver:traffic_generator
+            config:
+              traffic_config:
+                max_batch_size: 10
+                signals_per_second: 10
+                log_weight: 100
+              registry_path: https://github.com/open-telemetry/semantic-conventions.git[model]
+
+          attr1:
+            type: processor:attribute
+            config:
+              actions:
+                - action: insert
+                  key: pipeline.stage
+                  value: "1"
+
+          attr2:
+            type: processor:attribute
+            config:
+              actions:
+                - action: insert
+                  key: pipeline.stage2
+                  value: "2"
+
+          attr3:
+            type: processor:attribute
+            config:
+              actions:
+                - action: insert
+                  key: pipeline.stage3
+                  value: "3"
+
+          attr4:
+            type: processor:attribute
+            config:
+              actions:
+                - action: insert
+                  key: pipeline.stage4
+                  value: "4"
+
+          attr5:
+            type: processor:attribute
+            config:
+              actions:
+                - action: insert
+                  key: pipeline.stage5
+                  value: "5"
+
+          noop:
+            type: exporter:noop
+            config:
+
+        connections:
+          - from: receiver
+            to: attr1
+          - from: attr1
+            to: attr2
+          - from: attr2
+            to: attr3
+          - from: attr3
+            to: attr4
+          - from: attr4
+            to: attr5
+          - from: attr5
+            to: noop
--- a/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
+++ b/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
@@ -5,8 +5,13 @@
 #
 #   stopwatch "ingest_pipeline" - covers attr1 through attr5 (full chain)
 #
-# Stopwatches sum per-message compute.*.duration (the time spent inside timed()
-# closures) across each processor in the range, excluding queue wait and async I/O.
+# Stopwatches sum per-processor wall-clock compute time across each node in the
+# range. The engine arms an Instant send-marker before each `process()` call
+# and advances it on every `send_message`; the elapsed delta is accumulated
+# onto the message's stopwatch context and recorded at the stop node. This
+# captures synchronous compute plus any await/I/O that occurs inside
+# `process()` between sends; it excludes time the message spends queued in
+# inbound channels.
 # Metrics appear as stopwatch.compute.duration (MMSC histograms in
 # nanoseconds).
 #
--- a/rust/otap-dataflow/crates/engine/src/effect_handler.rs
+++ b/rust/otap-dataflow/crates/engine/src/effect_handler.rs
@@ -27,7 +27,7 @@ use tokio::net::{TcpListener, UdpSocket};
 #[derive(Clone, Copy)]
 pub enum SourceTagging {
     /// Disabled means no source node-id will be automatically
-    /// inserted for nodes that do not not otherwise subscribe to
+    /// inserted for nodes that do not otherwise subscribe to
     /// Ack/Nack.
     Disabled,
 
@@ -264,6 +264,18 @@ impl<PData> EffectHandlerCore<PData> {
         metrics: &mut MetricSet<M>,
     ) -> Result<(), TelemetryError> {
         self.metrics_reporter.report(metrics)
+    }
+
+    /// Reports processor-local wakeup scheduler metrics, if this processor uses
+    /// the local wakeup service.
+    pub fn report_local_scheduler_metrics(
+        &self,
+        metrics_reporter: &mut MetricsReporter,
+    ) -> Result<(), TelemetryError> {
+        if let Some(scheduler) = &self.local_scheduler {
+            scheduler.report_metrics(metrics_reporter)?;
+        }
+        Ok(())
     }
 
     /// Re-usable function to send a runtime control message. This returns a reference
--- a/rust/otap-dataflow/crates/engine/src/effect_handler.rs
+++ b/rust/otap-dataflow/crates/engine/src/effect_handler.rs
@@ -3,6 +3,9 @@
 
 //! Common foundation of all effect handlers.
 
+use std::sync::Mutex;
+use std::sync::atomic::{AtomicU64, Ordering};
+
 use crate::Interests;
 use crate::completion_emission_metrics::CompletionEmissionMetricsHandle;
 #[cfg(any(test, feature = "test-utils"))]
@@ -14,6 +17,7 @@ use crate::control::{
 use crate::error::Error;
 use crate::node::NodeId;
 use crate::node_local_scheduler::NodeLocalSchedulerHandle;
+use crate::stopwatch::{StopwatchId, StopwatchMetrics};
 use crate::{WakeupError, WakeupSetOutcome};
 use otap_df_channel::error::SendError;
 use otap_df_telemetry::error::Error as TelemetryError;
@@ -47,7 +51,6 @@ impl SourceTagging {
 /// Common implementation of all effect handlers.
 ///
 /// Note: This implementation is `Send`.
-#[derive(Clone)]
 pub(crate) struct EffectHandlerCore<PData> {
     pub(crate) node_id: NodeId,
     // ToDo refactor the code to avoid using Option here.
@@ -64,11 +67,49 @@ pub(crate) struct EffectHandlerCore<PData> {
     node_interests: Interests,
     /// Optional processor-local wakeup scheduler.
     local_scheduler: Option<NodeLocalSchedulerHandle>,
+    /// Accumulated synchronous compute duration (nanoseconds) for the
+    /// current PData message.  Reset to 0 before each `process()` call
+    /// and written into the frame when the processor sends the message
+    /// downstream.  Uses `AtomicU64` so the core remains `Send + Sync`.
+    pub(crate) per_message_compute_ns: AtomicU64,
+    /// Stopwatch IDs for which this node is a start node.
+    /// The processor's send path initialises accumulators for these IDs.
+    pub(crate) stopwatch_start_ids: Vec<StopwatchId>,
+    /// (StopwatchId, MetricSet) pairs for stopwatches where this node is the
+    /// stop node.  At send time the accumulated total for each ID is taken
+    /// from PData and recorded into the corresponding MetricSet.
+    /// Wrapped in `Mutex` for interior mutability from `&self` in send paths.
+    pub(crate) stopwatch_stop_metrics: Mutex<Vec<(StopwatchId, MetricSet<StopwatchMetrics>)>>,
+}
+
+impl<PData: Clone> Clone for EffectHandlerCore<PData> {
+    fn clone(&self) -> Self {
+        Self {
+            node_id: self.node_id.clone(),
+            runtime_ctrl_msg_sender: self.runtime_ctrl_msg_sender.clone(),
+            pipeline_completion_msg_sender: self.pipeline_completion_msg_sender.clone(),
+            metrics_reporter: self.metrics_reporter.clone(),
+            completion_emission_metrics: self.completion_emission_metrics.clone(),
+            source_tag: self.source_tag,
+            node_interests: self.node_interests,
+            local_scheduler: self.local_scheduler.clone(),
+            per_message_compute_ns: AtomicU64::new(
+                self.per_message_compute_ns.load(Ordering::Relaxed),
+            ),
+            stopwatch_start_ids: self.stopwatch_start_ids.clone(),
+            stopwatch_stop_metrics: Mutex::new(
+                self.stopwatch_stop_metrics
+                    .lock()
+                    .unwrap_or_else(|p| p.into_inner())
+                    .clone(),
+            ),
+        }
+    }
 }
 
 impl<PData> EffectHandlerCore<PData> {
     /// Creates a new EffectHandlerCore with node_id and a metrics reporter.
-    pub(crate) const fn new(node_id: NodeId, metrics_reporter: MetricsReporter) -> Self {
+    pub(crate) fn new(node_id: NodeId, metrics_reporter: MetricsReporter) -> Self {
         Self {
             node_id,
             runtime_ctrl_msg_sender: None,
@@ -78,6 +119,9 @@ impl<PData> EffectHandlerCore<PData> {
             source_tag: SourceTagging::Disabled,
             node_interests: Interests::empty(),
             local_scheduler: None,
+            per_message_compute_ns: AtomicU64::new(0),
+            stopwatch_start_ids: Vec::new(),
+            stopwatch_stop_metrics: Mutex::new(Vec::new()),
         }
     }
 
@@ -126,6 +170,16 @@ impl<PData> EffectHandlerCore<PData> {
         self.node_interests = interests;
     }
 
+    /// Sets stopwatch start/stop roles for this node.
+    pub(crate) fn set_stopwatch_roles(
+        &mut self,
+        start_ids: Vec<StopwatchId>,
+        stop_metrics: Vec<(StopwatchId, MetricSet<StopwatchMetrics>)>,
+    ) {
+        self.stopwatch_start_ids = start_ids;
+        self.stopwatch_stop_metrics = Mutex::new(stop_metrics);
+    }
+
     /// Returns the precomputed node interests.
     ///
     /// Includes SOURCE_TAGGING when source tagging is enabled.
@@ -142,6 +196,21 @@ impl<PData> EffectHandlerCore<PData> {
     #[must_use]
     pub(crate) fn node_id(&self) -> NodeId {
         self.node_id.clone()
+    }
+
+    /// Reset the per-message compute accumulator to zero.
+    ///
+    /// Called at the start of each PData message processing cycle so that
+    /// `timed()` calls within a single `process()` invocation accumulate
+    /// fresh values.
+    pub(crate) fn reset_per_message_compute_ns(&self) {
+        self.per_message_compute_ns.store(0, Ordering::Relaxed);
+    }
+
+    /// Return the accumulated per-message compute duration in nanoseconds.
+    #[must_use]
+    pub(crate) fn per_message_compute_ns(&self) -> u64 {
+        self.per_message_compute_ns.load(Ordering::Relaxed)
     }
 
     /// Print an info message to stdout.
--- a/rust/otap-dataflow/crates/engine/src/effect_handler.rs
+++ b/rust/otap-dataflow/crates/engine/src/effect_handler.rs
@@ -3,9 +3,6 @@
 
 //! Common foundation of all effect handlers.
 
-use std::sync::Mutex;
-use std::sync::atomic::{AtomicU64, Ordering};
-
 use crate::Interests;
 use crate::completion_emission_metrics::CompletionEmissionMetricsHandle;
 #[cfg(any(test, feature = "test-utils"))]
@@ -17,7 +14,6 @@ use crate::control::{
 use crate::error::Error;
 use crate::node::NodeId;
 use crate::node_local_scheduler::NodeLocalSchedulerHandle;
-use crate::stopwatch::{StopwatchId, StopwatchMetrics};
 use crate::{WakeupError, WakeupSetOutcome};
 use otap_df_channel::error::SendError;
 use otap_df_telemetry::error::Error as TelemetryError;
@@ -51,6 +47,7 @@ impl SourceTagging {
 /// Common implementation of all effect handlers.
 ///
 /// Note: This implementation is `Send`.
+#[derive(Clone)]
 pub(crate) struct EffectHandlerCore<PData> {
     pub(crate) node_id: NodeId,
     // ToDo refactor the code to avoid using Option here.
@@ -67,44 +64,6 @@ pub(crate) struct EffectHandlerCore<PData> {
     node_interests: Interests,
     /// Optional processor-local wakeup scheduler.
     local_scheduler: Option<NodeLocalSchedulerHandle>,
-    /// Accumulated synchronous compute duration (nanoseconds) for the
-    /// current PData message.  Reset to 0 before each `process()` call
-    /// and written into the frame when the processor sends the message
-    /// downstream.  Uses `AtomicU64` so the core remains `Send + Sync`.
-    pub(crate) per_message_compute_ns: AtomicU64,
-    /// Stopwatch IDs for which this node is a start node.
-    /// The processor's send path initialises accumulators for these IDs.
-    pub(crate) stopwatch_start_ids: Vec<StopwatchId>,
-    /// (StopwatchId, MetricSet) pairs for stopwatches where this node is the
-    /// stop node.  At send time the accumulated total for each ID is taken
-    /// from PData and recorded into the corresponding MetricSet.
-    /// Wrapped in `Mutex` for interior mutability from `&self` in send paths.
-    pub(crate) stopwatch_stop_metrics: Mutex<Vec<(StopwatchId, MetricSet<StopwatchMetrics>)>>,
-}
-
-impl<PData: Clone> Clone for EffectHandlerCore<PData> {
-    fn clone(&self) -> Self {
-        Self {
-            node_id: self.node_id.clone(),
-            runtime_ctrl_msg_sender: self.runtime_ctrl_msg_sender.clone(),
-            pipeline_completion_msg_sender: self.pipeline_completion_msg_sender.clone(),
-            metrics_reporter: self.metrics_reporter.clone(),
-            completion_emission_metrics: self.completion_emission_metrics.clone(),
-            source_tag: self.source_tag,
-            node_interests: self.node_interests,
-            local_scheduler: self.local_scheduler.clone(),
-            per_message_compute_ns: AtomicU64::new(
-                self.per_message_compute_ns.load(Ordering::Relaxed),
-            ),
-            stopwatch_start_ids: self.stopwatch_start_ids.clone(),
-            stopwatch_stop_metrics: Mutex::new(
-                self.stopwatch_stop_metrics
-                    .lock()
-                    .unwrap_or_else(|p| p.into_inner())
-                    .clone(),
-            ),
-        }
-    }
 }
 
 impl<PData> EffectHandlerCore<PData> {
@@ -119,9 +78,6 @@ impl<PData> EffectHandlerCore<PData> {
             source_tag: SourceTagging::Disabled,
             node_interests: Interests::empty(),
             local_scheduler: None,
-            per_message_compute_ns: AtomicU64::new(0),
-            stopwatch_start_ids: Vec::new(),
-            stopwatch_stop_metrics: Mutex::new(Vec::new()),
         }
     }
 
@@ -170,16 +126,6 @@ impl<PData> EffectHandlerCore<PData> {
         self.node_interests = interests;
     }
 
-    /// Sets stopwatch start/stop roles for this node.
-    pub(crate) fn set_stopwatch_roles(
-        &mut self,
-        start_ids: Vec<StopwatchId>,
-        stop_metrics: Vec<(StopwatchId, MetricSet<StopwatchMetrics>)>,
-    ) {
-        self.stopwatch_start_ids = start_ids;
-        self.stopwatch_stop_metrics = Mutex::new(stop_metrics);
-    }
-
     /// Returns the precomputed node interests.
     ///
     /// Includes SOURCE_TAGGING when source tagging is enabled.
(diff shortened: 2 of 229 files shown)


rust/otap-dataflow/crates/engine/src/local/processor.rs
```diff
@@ -177,12 +177,62 @@ impl<PData> EffectHandler<PData> {
...
+    /// Returns the stopwatch IDs for which this node is a start node.
+    #[must_use]
+    pub fn stopwatch_start_ids(&self) -> &[crate::stopwatch::StopwatchId] {
+        &self.core.stopwatch_start_ids
+    }
+
+    /// Record `total` nanoseconds for each stopwatch where this node is
+    /// the stop node and the given ID matches, then report the snapshot
+    /// so the telemetry dispatcher can pick it up.
+    pub fn record_stopwatch_stop(&self, id: crate::stopwatch::StopwatchId, total: u64) {
```
utpilla: This reports a metric snapshot immediately on every PData at the stop node, then clears the `MetricSet`. The existing `ComputeDuration` metric follows a different pattern. It accumulates observations into an `Mmsc` and only reports when `CollectTelemetry` arrives periodically.

Why not follow the same model here? Accumulate into the `MetricSet` and let the periodic `StopwatchState::report()` (which is already wired up in the dispatcher) drain it. That keeps the send path lighter and avoids introducing a second reporting pattern.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -41,6 +43,12 @@ pub struct Context {
     /// `None` when no headers have been captured (the common case, zero
     /// additional allocation).
     transport_headers: Option<TransportHeaders>,
+    /// Active stopwatch accumulators keyed by stopwatch ID.
+    ///
+    /// Populated by start nodes, accumulated by every node's send path,
+    /// consumed by stop nodes.  Supports multiple overlapping ranges.
```
utpilla: Is there a use case for overlapping stopwatch ranges?

Without that, this simplifies significantly. You would only need an `Option<u64>` instead of `Vec<(StopwatchId, u64)>`, no `StopwatchId` on PData, and the accumulator methods become simpler as well.

rust/otap-dataflow/crates/engine/src/effect_handler.rs
```diff
@@ -64,11 +67,49 @@ pub(crate) struct EffectHandlerCore<PData> {
     node_interests: Interests,
     /// Optional processor-local wakeup scheduler.
     local_scheduler: Option<NodeLocalSchedulerHandle>,
+    /// Accumulated synchronous compute duration (nanoseconds) for the
+    /// current PData message.  Reset to 0 before each `process()` call
+    /// and written into the frame when the processor sends the message
+    /// downstream.  Uses `AtomicU64` so the core remains `Send + Sync`.
+    pub(crate) per_message_compute_ns: AtomicU64,
```
utpilla: `EffectHandlerCore` previously had no `sync` primitives and derived Clone. This PR adds `AtomicU64` and `Mutex`, which requires a manual Clone impl and adds overhead on every processor. including the local path where synchronization is unnecessary.

Since no processor in the codebase implements the shared (`Send`) path, I'd suggest scoping stopwatch support to local processors only. Put these fields on `local::EffectHandler` as `Cell<u64>` / `RefCell` instead of on the core. This keeps the core sync-primitive-free and zero-cost. If shared processor support is needed later, it can be added to the shared wrapper with the appropriate primitives at that point.

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -74,6 +78,7 @@ impl ComputeDuration {
     pub fn timed<T, E>(
         &self,
         interests: Interests,
+        capture: Option<&AtomicU64>,
```
utpilla: If the stopwatch fields move to the local `EffectHandler` wrapper as suggested in https://github.com/open-telemetry/otel-arrow/pull/2747#discussion_r3145242323, this `capture: Option<&AtomicU64>` parameter becomes unnecessary. The local wrapper can read the elapsed time from the return value of `timed()` and write to its own `Cell` directly.

rust/otap-dataflow/crates/engine/src/effect_handler.rs
drewrelmas: This is a great suggestion and simplifies a lot, I plan to follow it.

rust/otap-dataflow/crates/otap/src/pdata.rs
drewrelmas: I was originally thinking a customer may want to define one stopwatch over all processors and additional stopwatches between certain internal ranges.

My own personal use case does not require overlapping ranges - I will incorporate this feedback.

rust/otap-dataflow/crates/engine/src/process_duration.rs
drewrelmas: Addressed

rust/otap-dataflow/crates/engine/src/local/processor.rs
drewrelmas: I have gone back over the model and I think I've updated to follow the regular processor `ComputeDuration` model.

Is [this what you were imagining](https://github.com/open-telemetry/otel-arrow/pull/2747/changes#diff-9fa1d4c99b9a3ef8ef83cbaf36e0088406ea2c8965d5dcc2065a0ea40f2d57cb)? Identifying `CollectTelemetry` messages and choosing that moment to report stopwatch accumulator `Mmsc`?

utpilla: Yes.

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -75,7 +80,7 @@ impl ComputeDuration {
         &self,
         interests: Interests,
         f: impl FnOnce() -> Result<T, E>,
-    ) -> Result<T, E> {
+    ) -> (Result<T, E>, u64) {
```
utpilla: `ComputeDuration::timed()` now always returns `(Result<T, E>, u64)` instead of just `Result<T, E>`. This means the local `EffectHandler::timed()` always accumulates elapsed nanos into `per_message_compute_ns` on every processor, on every message even in pipelines with no stopwatches configured at all. A user who never configures a stopwatch still gets the changed `timed()` return type, the per-message accumulation, and the `reset_per_message_compute_ns()` call in the engine loop.

I'd suggest we keep `ComputeDuration::timed()` at its original `-> Result<T, E>` signature and add a separate `timed_with_elapsed() -> (Result<T, E>, u64)` for the stopwatch path. The effect handler can gate on a `stopwatch_accumulation_enabled: bool` flag (set at build time for processors within a stopwatch range). The existing `local::EffectHandler::timed()` wrapper would branch like this:

```rust
pub fn timed<T, E>(
    &self,
    cd: &ComputeDuration,
    f: impl FnOnce() -> Result<T, E>,
) -> Result<T, E> {
    if self.stopwatch_accumulation_enabled {
        let (result, elapsed_ns) = cd.timed_with_elapsed(
            self.core.node_interests(), f
        );
        if elapsed_ns > 0 {
            self.per_message_compute_ns
                .set(self.per_message_compute_ns.get().saturating_add(elapsed_ns));
        }
        result
    } else {
        cd.timed(self.core.node_interests(), f)
    }
}
```

This keeps the no-stopwatch path clean.

drewrelmas: I do see the value in this, but I'm hesitant about adding a per-node `stopwatch_accumulation_enabled` at build-time on each processor. This requires added code to do graph traversal and determine which nodes are in/out of scope. It is doable but feels a little heavy. The current stopwatch impl is more free-form (start recording at one node, stop recording at another, and assume accumulation happened in the middle).

Do you think it is acceptable to have a pipeline-level flag instead? It would have extra cost from any nodes not participating in any stopwatch for that pipeline, but in my opinion the most common case for stopwatch would be spanning the majority of a pipeline.

utpilla: > Do you think it is acceptable to have a pipeline-level flag instead?

Yeah, I think that's an acceptable approach.

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -94,6 +94,37 @@ impl ComputeDuration {
...
 
+    /// Time a synchronous, fallible closure and return the elapsed
+    /// nanoseconds alongside the result.
+    ///
+    /// Like [`timed`](Self::timed), but also returns the wall-clock
+    /// nanoseconds consumed by `f` (0 when timing is disabled).
+    /// Used by the stopwatch path to accumulate per-message compute
+    /// duration.
+    #[inline]
+    pub fn timed_with_elapsed<T, E>(
```
jmacd: Is `timed_with_elapsed()` needed separately from `timed()`? I am wondering why we might need both.

drewrelmas: @jmacd please see https://github.com/open-telemetry/otel-arrow/pull/2747#discussion_r3150275877

Utkarsh had a good point that we don't always want to accumulate, especially when there are pipelines that don't participate in stopwatches. I tried to incorporate the feedback, but without a full graph traversal it is currently limited to a per-pipeline switch.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -41,6 +42,12 @@ pub struct Context {
     /// `None` when no headers have been captured (the common case, zero
     /// additional allocation).
     transport_headers: Option<TransportHeaders>,
+    /// Active stopwatch accumulator (nanoseconds).
+    ///
+    /// `Some(ns)` when a message is inside a stopwatch range (between
+    /// start and stop nodes).  At most one stopwatch can be active at
+    /// a time (non-overlapping ranges).
+    stopwatch_compute_ns: Option<u64>,
```
jmacd: I recommend `Option<NonZeroU64>`, it will save a little bit of space.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -641,6 +681,57 @@ impl_consumer_ext!(otap_df_engine::shared::exporter::EffectHandler<OtapPdata>);
...
+                self.record_stopwatch_stop(total);
+            }
+        }
+    };
+}
+
+impl_stopwatch_handler!(otap_df_engine::local::processor::EffectHandler<OtapPdata>);
+impl_stopwatch_handler!(otap_df_engine::shared::processor::EffectHandler<OtapPdata>);
+impl_stopwatch_handler!(otap_df_engine::local::receiver::EffectHandler<OtapPdata>);
+impl_stopwatch_handler!(otap_df_engine::shared::receiver::EffectHandler<OtapPdata>);
```
jmacd: Do we need the shared variations of these? Do the receivers measure stopwatches? (I thought it was only processors.)

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -94,6 +94,37 @@ impl ComputeDuration {
...
+                &self.acc_success
+            } else {
+                &self.acc_failed
+            };
+            let mut val = acc.get();
+            val.record(elapsed);
+            acc.set(val);
+            (result, elapsed as u64)
+        } else {
+            (f(), 0)
```
ThomsonTan: Seems the configured stopwatches could silently report zero durations whenever `runtime_metrics` is `basic` or `none`, because `Interests::from_metric_level()` does not include `PROCESS_DURATION` at those levels. Since `stopwatches` are independent telemetry config, I think `timed_with_elapsed()` should always measure and return elapsed time when called, while only gating the `processor.compute.duration` MMSC recording on `PROCESS_DURATION`.

drewrelmas: I was originally thinking that stopwatch enablement (which mainly measures composite processor duration) should be enabled/disabled based on `runtime_metrics`. You are suggesting it is an independent feature, and the metric level should only gate individual processor duration reporting?

One other way to avoid silent ignore of stopwatch is adding a WARN log emission stating "Stopwatch was configured but ignored due to runtime_metrics level {level}"

rust/otap-dataflow/crates/otap/src/pdata.rs
drewrelmas: Ack, can easily do

rust/otap-dataflow/crates/otap/src/pdata.rs
drewrelmas: I think I've found a better way to do this (with Copilot assistance!). I've added a `maybe_stopwatch_accumulate()` macro so we don't need the no-op impls on `shared::processor`, `local::receiver`, and `shared::receiver`.

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
```diff
@@ -214,6 +216,25 @@ impl<PData: 'static + Debug + Clone + ReceivedAtNode + Unwindable> RuntimePipeli
...
+
+        // Collect the set of processor node indices for stopwatch validation.
+        let processor_indices: HashSet<usize> =
+            processors.iter().map(|p| p.node_id().index).collect();
+
+        // Build stopwatch state and per-node role assignments up front.
+        let stopwatch_state = build_stopwatch_state(
+            &telemetry_policy,
+            &node_name_to_index,
+            &processor_indices,
```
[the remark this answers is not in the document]
drewrelmas: Following this advice

rust/otap-dataflow/crates/engine/src/stopwatch.rs
```diff
@@ -0,0 +1,356 @@
...
+                "Stopwatch start/stop nodes must be processors, skipping"
+            );
+            continue;
+        }
+
+        // Non-overlapping: a node cannot be the start (or stop) of more than
+        // one stopwatch.  The second definition would silently overwrite the
+        // first accumulator on PData.
+        if start_nodes.contains(&start_idx) || stop_nodes.contains_key(&stop_idx) {
+            otap_df_telemetry::otel_warn!(
```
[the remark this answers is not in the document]
drewrelmas: Also following this, updating tests to cover this case

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -641,22 +688,78 @@ impl_consumer_ext!(otap_df_engine::shared::exporter::EffectHandler<OtapPdata>);
...
-    ($async_attr:meta, $trait_name:ident, $handler:ty, $id_method:ident) => {
+    ($async_attr:meta, $trait_name:ident, $handler:ty, $id_method:ident, $stopwatch:ident) => {
         #[$async_attr]
         impl $trait_name<OtapPdata> for $handler {
             async fn send_message_with_source_node(
                 &self,
                 mut data: OtapPdata,
             ) -> Result<(), TypedError<OtapPdata>> {
                 data.prepare_source_send(self.node_interests(), self.$id_method().index);
+                maybe_stopwatch_accumulate!($stopwatch, self, &mut data);
```
lalitb: Can we move the stopwatch hook lower down so all local processor sends of `OtapPdata` go through it, not just the source-node helpers? Some processors call `timed()` and then send via plain `send_message` / `send_message_to`, so their compute time is missing from the sum. If one of those processors is a stop node, the stopwatch may never record.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -641,22 +688,78 @@ impl_consumer_ext!(otap_df_engine::shared::exporter::EffectHandler<OtapPdata>);
 
 /* --------  effect handler extensions (shared, local) -------- */
 
+/// Trait abstracting the stopwatch-related methods available on all EffectHandler types.
+trait StopwatchHandler {
+    fn per_message_compute_ns(&self) -> u64;
+    fn is_stopwatch_start(&self) -> bool;
+    fn is_stopwatch_stop(&self) -> bool;
+    fn record_stopwatch_stop(&self, total: u64);
+}
```
lalitb: Nit: could we simplify this now that stopwatch handling is only wired for local processors? The trait/macro makes this look like a generic extension point, but it only has one real impl today. A helper that takes the local processor effect handler directly may be easier to follow. If the plan is to generalize this soon, feel free to ignore.

rust/otap-dataflow/crates/engine/src/stopwatch.rs
```diff
@@ -0,0 +1,420 @@
...
+        // Non-overlapping: a node cannot appear in either role across
+        // stopwatches. The second definition would silently overwrite the
+        // first accumulator on PData.
+        if start_nodes.contains(&start_idx)
+            || stop_nodes.contains_key(&start_idx)
+            || start_nodes.contains(&stop_idx)
+            || stop_nodes.contains_key(&stop_idx)
+        {
+            otap_df_telemetry::otel_warn!(
+                "stopwatch.config.overlap",
```
lalitb: I think this only catches reused endpoint nodes, not ranges that overlap in the graph. For example, `a -> c` and `b -> d` have distinct endpoints, so both can be accepted, but at runtime they share the same single accumulator and can overwrite each other.

Should we reject graph-overlapping ranges at build time too, or make the unsupported case more explicit?

lalitb:
Thanks for the detailed PR description. The goal makes sense to me: a single compute-duration metric across a processor range is useful.

One concern I see if  making the YAML contract line up with runtime behavior. Today the YAML can be accepted, but the reported stopwatch may not mean what the user expects. For example, a stop node before the start node may never fire, a processor that does not call `timed()` contributes `0`, and cloning/fan-out can make the accumulator semantics unclear.

  Could we make these cases louder where possible? Some look rejectable at build time, like unreachable or graph-overlapping ranges. Others, like processors that do not call `timed()` or clone/fan-out behavior, may need explicit warnings, docs, or processor capability metadata. A stopwatch that reports a misleading value is harder to debug than one that refuses to start, since users will tend to trust the metric once it appears.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -41,6 +44,13 @@ pub struct Context {
     /// `None` when no headers have been captured (the common case, zero
     /// additional allocation).
     transport_headers: Option<TransportHeaders>,
+    /// Active stopwatch accumulator (nanoseconds).
+    ///
+    /// `Some(ns + 1)` when a message is inside a stopwatch range (between
+    /// start and stop nodes). The +1 bias represents active-with-zero while
+    /// preserving `Option<NonZeroU64>` niche optimization.
+    /// At most one stopwatch can be active at a time (non-overlapping ranges).
+    stopwatch_compute_ns: Option<NonZeroU64>,
```
jmacd: Recommending to leave out the +1 adjustment. Duration measurements are required to be >0 ns. It's OK to use `elapsed.max(1)` at the callsite to ensure this, OTOH

rust/otap-dataflow/crates/otap/src/pdata.rs
drewrelmas: Great catch! Before pushing, would like your thoughts on potential solution. I can introduce a `LocalProcessorSendHook` trait (with a default no-op of `before_local_processor_send()` ) and then `OtapPdata` provides real implementation that accumulates the stopwatch? Then, `send_message`, `try_send_message`, `send_message_to`, and `try_send_message_to` all invoke that function before passing to `router`?

rust/otap-dataflow/crates/otap/src/pdata.rs
drewrelmas: Yes - this is an even better solution than the one I found in response to https://github.com/open-telemetry/otel-arrow/pull/2747#discussion_r3162362692. Will do this.

rust/otap-dataflow/crates/engine/src/stopwatch.rs
drewrelmas: I see your point and think that we should add the graph traversal to improve validation. I would prefer to do this in a follow-up PR. Opened #2784 to track.

rust/otap-dataflow/crates/otap/src/pdata.rs
drewrelmas: Easily done as well

drewrelmas:
> Could we make these cases louder where possible? Some look rejectable at build time, like unreachable or graph-overlapping ranges. Others, like processors that do not call `timed()` or clone/fan-out behavior, may need explicit warnings, docs, or processor capability metadata. A stopwatch that reports a misleading value is harder to debug than one that refuses to start, since users will tend to trust the metric once it appears.

Tracking follow-up in #2785 to make sure relevant processors participate in `timed()`. We can consider emitting a warning when a processor that does not participate in `timed()` is included in a stopwatch range.

jmacd (approved):

rust/otap-dataflow/crates/engine/src/process_duration.rs
ThomsonTan: Yes. I think `runtime_metrics` should gate the broad per-processor runtime metrics, like `processor.compute.duration`.

But `stopwatches` are explicit config: the user named a start node, a stop node, and asked for `stopwatch.compute.duration`. If we accept that config, we should either measure it correctly or clearly skip/reject it.

The current behavior is the risky middle ground: the stopwatch can be configured and registered, but at `runtime_metrics: basic` `timed_with_elapsed()` returns `0`, so the metric looks enabled while reporting incorrect zero durations.

So either fix is OK:

1. Always return real elapsed time from `timed_with_elapsed()` when the stopwatch path calls it, while only gating `processor.compute.duration` on `PROCESS_DURATION`.
2. Or skip/reject stopwatch config at `runtime_metrics: none/basic` with a WARN and avoid registering a misleading zero-valued metric.

I slightly prefer option 1 because it lets users enable one targeted composite duration without enabling all per-processor duration metrics.

rust/otap-dataflow/crates/engine/src/stopwatch.rs
```diff
@@ -0,0 +1,420 @@
...
+                "stopwatch.config.invalid",
+                name = sw_config.name,
+                start = sw_config.start_node,
+                stop = sw_config.stop_node,
+                "Stopwatch references unknown node(s), skipping"
+            );
+            continue;
+        };
+
+        // Stopwatches only work on local processors (they own the Cell<Mmsc>
```
lquerel: What happens if the chain of processors is local -> shared -> local ?

rust/otap-dataflow/crates/otap/src/pdata.rs
drewrelmas: I've done another refactor based on some of Laurent's feedback to support both local and shared processors.

rust/otap-dataflow/crates/engine/src/stopwatch.rs
drewrelmas: The latest iteration should support this. Since the accumulation happens on `Pdata` objects, it doesn't really matter if the data flows through local and shared processors. Each one will add to the accumulation in their own way.

rust/otap-dataflow/crates/engine/src/process_duration.rs
drewrelmas: I've added a WARN message for this case and confirmed we don't register a misleading zero-valued metric. It feels the most correct to me from user expectation even though we don't technically depend on `PROCESS_DURATION` to compute the stopwatch value.

rust/otap-dataflow/crates/engine/src/local/processor.rs
```diff
@@ -124,6 +128,27 @@ pub struct EffectHandler<PData> {
...
+    /// When false, `begin_process_timing` and
+    /// `take_elapsed_since_send_marker_ns` are no-ops.
+    stopwatches_active: bool,
+    /// Metric set for the stopwatch where this node is the stop node.
+    /// `None` when this node is not a stopwatch stop node.
+    /// Reported periodically via [`report_stopwatch`], not on every message.
+    stopwatch_stop_metric: Option<MetricSet<StopwatchMetrics>>,
+    /// Local Mmsc accumulator for stopwatch stop observations.
+    /// Drained into `stopwatch_stop_metric` on periodic telemetry collection.
+    stopwatch_stop_acc: Cell<Mmsc>,
```
jmacd: These 5 fields have corresponding entries in both the local and shared stopwatch state. I've begun thinking of this change as an important first step, but it seems also likely that we will evolve how stopwatches are instrumented in the future. 

Not required! I think this would be more readable if we created an `enum StopwatchState { Local { these 5 fields }, Shared { those 5 fields } }` and moved the stopwatch fields into a common location.

lquerel: Shouldn’t we create a structure that groups the metric set and the MMSC, so that both can sit behind the `Option` already used around the metric set? That would avoid creating an MMSC for nodes that are not terminal.

rust/otap-dataflow/crates/engine/src/shared/processor.rs
```diff
@@ -123,6 +126,33 @@ pub struct EffectHandler<PData> {
...
+    /// Metric set for the stopwatch where this node is the stop node.
+    /// Only meaningful on the engine-owned handler; clones see `None` in
+    /// practice but the field is replicated harmlessly.
+    stopwatch_stop_metric: Option<MetricSet<StopwatchMetrics>>,
+    /// Shared Mmsc accumulator for stopwatch stop observations.
+    /// Drained and reported on `CollectTelemetry` and shutdown.
+    ///
+    /// `Mutex` should be fine here — contention is bounded to the stop
+    /// node and only on messages with an active stopwatch.
+    stopwatch_stop_acc: Arc<Mutex<Mmsc>>,
```
lquerel: Same here

rust/otap-dataflow/crates/engine/src/effect_handler.rs
```diff
@@ -68,7 +68,7 @@ pub(crate) struct EffectHandlerCore<PData> {
 
 impl<PData> EffectHandlerCore<PData> {
     /// Creates a new EffectHandlerCore with node_id and a metrics reporter.
-    pub(crate) const fn new(node_id: NodeId, metrics_reporter: MetricsReporter) -> Self {
+    pub(crate) fn new(node_id: NodeId, metrics_reporter: MetricsReporter) -> Self {
```
lquerel: Why removing the const?

rust/otap-dataflow/crates/engine/src/lib.rs
```diff
@@ -397,6 +398,39 @@ impl StampOutputPort for String {
...
+    }
+}
+
+impl StopwatchAccumulation for String {
+    fn start_stopwatch(&mut self) {}
+    fn add_stopwatch_compute(&mut self, _ns: u64) {}
+    fn take_stopwatch_compute(&mut self) -> Option<u64> {
+        None
+    }
+}
```
lquerel: Why do we implement `StopwatchAccumulation` for `()` and `String` ?

rust/otap-dataflow/crates/engine/src/process_duration.rs
```diff
@@ -12,6 +12,13 @@
 //! recorded into separate instruments so operators can distinguish
 //! compute time from error-path time.
 //!
+//! This API complements, but is not required for, the engine's automatic
+//! per-message stopwatch. [`ComputeDuration::timed`] provides the
```
lquerel, line 16: In practice, we'll need to see whether we still need this `timed` method or not.

rust/otap-dataflow/crates/engine/src/processor.rs
```diff
@@ -27,15 +27,57 @@ use crate::node::{Node, NodeId, NodeWithPDataReceiver, NodeWithPDataSender};
...
 use std::sync::Arc;
 use std::time::Duration;
 
+/// Abstract processor `EffectHandler` surface used by per-message
+/// bookkeeping hooks (currently: distributed stopwatch accumulation).
+///
+/// Implemented by both `local::processor::EffectHandler<PData>` and
+/// `shared::processor::EffectHandler<PData>` so that PData-side hooks can
+/// be written once, generic over handler kind.
+pub trait ProcessorHandler {
```
lquerel: This trait seems very specific to the stopwatch mechanism. I wonder whether we shouldn't give it a name that’s a bit less generic.

rust/otap-dataflow/crates/engine/src/local/processor.rs
drewrelmas: This is a good idea - I'm thinking of something similar with a `StopwatchState<Marked, Accumulator>` struct that shares the primitives and uses necessary types for local/shared implementations:

```rust
pub(crate) struct StopwatchState<Marker, Accumulator>

pub(crate) type SharedStopwatchState =
    StopwatchState<Arc<Mutex<Option<Instant>>>, Arc<Mutex<Mmsc>>>;

pub(crate) type LocalStopwatchState =
    StopwatchState<Rc<Cell<Option<Instant>>>, Cell<Mmsc>>;
```

rust/otap-dataflow/crates/engine/src/processor.rs
```diff
@@ -546,18 +595,35 @@ impl<PData> ProcessorWrapper<PData> {
...
+                        && matches!(
+                            &msg,
+                            Message::Control(NodeControlMsg::CollectTelemetry { .. })
+                        )
+                    {
+                        effect_handler.report_stopwatch();
+                    }
+                    if matches!(&msg, Message::PData(_)) {
+                        effect_handler.begin_process_timing();
+                    }
```
lquerel: I suggest revisiting this code. At the top level, you should have an `if` so that the `match` is only evaluated when we are in a stopwatch case. The first branch of that `match` would detect the `CollectMetric` control message, and the second branch would start the stopwatch when handling `pdata`.

rust/otap-dataflow/crates/engine/src/processor.rs
```diff
@@ -546,18 +595,35 @@ impl<PData> ProcessorWrapper<PData> {
...
                     processor.process(msg, &mut effect_handler).await?;
                 }
                 // Cancel periodic collection
                 _ = telemetry_cancel_handle.cancel().await;
                 // Collect final metrics before exiting
+                effect_handler.report_stopwatch();
```
lquerel: Shoud we call `report_stopwatch` even if `is_stopwatch_stop` is false?

rust/otap-dataflow/crates/engine/src/processor.rs
```diff
@@ -580,18 +646,35 @@ impl<PData> ProcessorWrapper<PData> {
...
+                        && matches!(
+                            &msg,
+                            Message::Control(NodeControlMsg::CollectTelemetry { .. })
+                        )
+                    {
+                        effect_handler.report_stopwatch();
+                    }
+                    if matches!(&msg, Message::PData(_)) {
+                        effect_handler.begin_process_timing();
+                    }
```
lquerel: See previous comment

rust/otap-dataflow/crates/engine/src/stopwatch.rs
```diff
@@ -0,0 +1,440 @@
...
+/// (e.g. `runtime_metrics: basic` or `none`), all configured stopwatches
+/// are skipped with a single warning so users see a clear signal rather
+/// than silent zero-duration metrics.
+pub(crate) fn build_stopwatch_state(
+    telemetry_policy: &TelemetryPolicy,
+    node_name_to_index: &HashMap<String, usize>,
+    processor_indices: &HashSet<usize>,
+    node_interests: Interests,
+    pipeline_context: &PipelineContext,
+) -> StopwatchState {
```
lquerel: I’m surprised not to see a `Result` here returning `StopwatchState` along with the engine’s error enum. If this build function detects that a start/stop node does not exist or is not a processor, then it should return an error so that we can report the configuration as invalid, rather than just logging it.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -340,6 +354,45 @@ impl otap_df_engine::StampOutputPort for OtapPdata {
...
 }
 
+impl StopwatchAccumulation for OtapPdata {
+    fn start_stopwatch(&mut self) {
+        if self.context.stopwatch_compute_ns.is_some() {
+            otap_df_telemetry::otel_warn!(
+                "stopwatch.overlap",
+                "start_stopwatch called while another stopwatch is active; \
+                 overlapping ranges are not supported — previous accumulator discarded"
+            );
```
lquerel, line 371: That should be detected during the construction phase and not in the hot path.

rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
```diff
@@ -0,0 +1,105 @@
...
+#
+# Five attribute processors form a sequential processing chain. A stopwatch
+# measures aggregate synchronous compute duration across the full chain:
+#
+#   stopwatch "ingest_pipeline" - covers attr1 through attr5 (full chain)
+#
+# Stopwatches sum per-message compute.*.duration (the time spent inside timed()
+# closures) across each processor in the range, excluding queue wait and async I/O.
+# Metrics appear as stopwatch.compute.duration (MMSC histograms in
+# nanoseconds).
```
lquerel: It's no longer the case, right?

lquerel (changes requested):
I think we are very close. I made a bunch of suggestion to either cleanup the code or to improve memory usage.

rust/otap-dataflow/crates/engine/src/lib.rs
drewrelmas: These are never concretely used, only the `OtapPdata` implementation, will remove them. I was following along with the default impls of `ReceivedAtNode()` and `StampOutputPort()` without too much reflection which DO have `()` and `String` implementations.

rust/otap-dataflow/crates/engine/src/processor.rs
drewrelmas: Good suggestion, will push again with `StopwatchEffectHandler`

rust/otap-dataflow/crates/engine/src/processor.rs
drewrelmas: No, there is no need to. Wrapping in a conditional.

rust/otap-dataflow/crates/engine/src/stopwatch.rs
drewrelmas: This is a great point, incorporating.

rust/otap-dataflow/crates/otap/src/pdata.rs
drewrelmas: See https://github.com/open-telemetry/otel-arrow/issues/2784, follow-up from other comments. This is definitely doable at build-time, just requires some boilerplate code for testing reachability which I thought could be separated from core implementation.

rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
drewrelmas: Correct, updated

rust/otap-dataflow/crates/engine/src/effect_handler.rs
drewrelmas: Leftover change from earlier implementation attempt that touched `EffectHandlerCore`, reverting.

lquerel (approved):
LGTM

Issue #2752, opened 2026-04-24
Support logical expressions that check the type of fields or the stream elements


albertlockett:
In OPL / OTAP Query Engine, it would be great if we could use an expression to check the type of some signal when doing a filter. For example, if we could do:

```kotlin
signals | 
if (is Log) {
 // ...
} else if (is Metric) {
 // ... etc.
```

or:
```kotlin
if (attributes["x"] is String) {
  // ...
}
```

The rational for this is:
- a) be able to write a single program that processes logs/metrics/spans separately (e.g. in a single transform processor)
- b) eventually when the parser/planner has more type awareness, it may be smart enough to regognize that when the stream type is a `signal` (could be logs, traces or metrics), we'll want to disallow accessing fields that are not in the intersection of these types. This type of syntax could be used to cast the stream type of the if/else branch into the type (much like https://kotlinlang.org/docs/typecasts.html#smart-casts).
- c) Similar to point b, when the parser/planner may eventually become smart enough to give type warnings in the case where some field is not guaranteed to be the type expected by the expression. For exaple, when passing an AnyValue to a function like `substring`. In these cases, we may be able to use this syntax to ensure the type will be correct and get rid of the warning.

PR #2754, opened 2026-04-24
Add capability in query-engine to filter by type of signal

albertlockett:
# Change Summary

Adds the capability to use a syntax like `is <signal type>` in logical expressions in OPL. This can be used for example when we want to have a single program that does different operations to each signal:
```kotlin
signals |
if (is Log) {
 // ...
} else  if (is Metric) {
  // ...
} else if (is Span) {
  // ...
}
```

Alternatively, we can also do things like this to just keep/drop certain signal types:
```kql
signals | where is Log
signals | where not(is Log)
```

* Part of #2752

Unit tests

Yes this syntax is now available for use in transform processor

 

## Future work:

In a followup, I'll add support for checking where a particular field is of some type. 
```
logs | if (attributes["x"] is String) {
  // ...
```
There are some TODOs in this PR to add support for this in the near future.

Also, when we eventually add more capability to the parser/planner to be type aware, it may have the capability to reject the use of invalid field accesses for certain signal types, and the syntax in this PR will allow users to get around these compile time checks. However, that's not implemented as part of this PR.

### DIFF
--- a/rust/otap-dataflow/crates/opl/src/parser/expression.rs
+++ b/rust/otap-dataflow/crates/opl/src/parser/expression.rs
@@ -7,9 +7,9 @@ use data_engine_expressions::{
     AndLogicalExpression, BinaryMathematicalScalarExpression, BooleanScalarExpression,
     CaptureTextScalarExpression, CollectionScalarExpression, CombineScalarExpression,
     ContainsLogicalExpression, DateTimeScalarExpression, DoubleScalarExpression, DoubleValue,
-    EqualToLogicalExpression, Expression, GetTypeScalarExpression, GreaterThanLogicalExpression,
-    GreaterThanOrEqualToLogicalExpression, IntegerScalarExpression, IntegerValue,
-    InvokeFunctionArgument, InvokeFunctionScalarExpression, JoinTextScalarExpression,
+    EqualToLogicalExpression, Expression, GetRecordTypeScalarExpression,
+    GreaterThanLogicalExpression, GreaterThanOrEqualToLogicalExpression, IntegerScalarExpression,
+    IntegerValue, InvokeFunctionArgument, InvokeFunctionScalarExpression, JoinTextScalarExpression,
     ListScalarExpression, LogicalExpression, MatchesLogicalExpression, MathScalarExpression,
     NotLogicalExpression, NullScalarExpression, OrLogicalExpression, QueryLocation,
     RegexScalarExpression, ReplaceTextScalarExpression, ScalarExpression, SliceScalarExpression,
@@ -215,13 +215,9 @@ pub(crate) fn parse_type_check_expression(
         // IF there are two rules, we have an expression like (is <Type>) meaning we're checking
         // that an element of the stream is some type
         2 => {
-            let type_check_expr = ScalarExpression::GetType(GetTypeScalarExpression::new(
-                type_check_rule_query_location.clone(),
-                ScalarExpression::Source(SourceScalarExpression::new(
-                    type_check_rule_query_location.clone(),
-                    ValueAccessor::new_with_selectors(Vec::new()),
-                )),
-            ));
+            let type_check_expr = ScalarExpression::GetRecordType(
+                GetRecordTypeScalarExpression::new(type_check_rule_query_location.clone()),
+            );
 
             let type_name_rule = inner_rules.nth(1).expect("two rules");
             let type_check_expected_type = ScalarExpression::Static(
--- a/rust/otap-dataflow/crates/opl/src/parser/expression.rs
+++ b/rust/otap-dataflow/crates/opl/src/parser/expression.rs
@@ -7,7 +7,7 @@ use data_engine_expressions::{
     AndLogicalExpression, BinaryMathematicalScalarExpression, BooleanScalarExpression,
     CaptureTextScalarExpression, CollectionScalarExpression, CombineScalarExpression,
     ContainsLogicalExpression, DateTimeScalarExpression, DoubleScalarExpression, DoubleValue,
-    EqualToLogicalExpression, Expression, GreaterThanLogicalExpression,
+    EqualToLogicalExpression, Expression, GetTypeScalarExpression, GreaterThanLogicalExpression,
     GreaterThanOrEqualToLogicalExpression, IntegerScalarExpression, IntegerValue,
     InvokeFunctionArgument, InvokeFunctionScalarExpression, JoinTextScalarExpression,
     ListScalarExpression, LogicalExpression, MatchesLogicalExpression, MathScalarExpression,
@@ -137,8 +137,8 @@ pub(crate) fn parse_and_expression(
     let left_expr = parse_next_child_rule(
         pipeline_builder,
         &mut inner_rules,
-        Rule::rel_expression,
-        parse_rel_expression,
+        Rule::type_check_expression,
+        parse_type_check_expression,
     )
     .transpose()?
     .ok_or_else(|| no_inner_rule_error(query_location.clone()))?;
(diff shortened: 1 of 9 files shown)


rust/otap-dataflow/crates/opl/src/parser/expression.rs
```diff
@@ -198,6 +198,66 @@ impl From<LogicalOrScalarExpr> for ScalarExpression {
...
+        // If there is one rule, it's simply a relative expression.
+        1 => {
+            // we have a simple relative expression, with
+            let inner_rule = inner_rules.next().expect("one rule");
+            parse_rel_expression(inner_rule, pipeline_builder)
+        }
+        // IF there are two rules, we have an expression like (is <Type>) meaning we're checking
+        // that an element of the stream is some type
+        2 => {
+            let type_check_expr = ScalarExpression::GetType(GetTypeScalarExpression::new(
```
CodeBlanch: This is really not what `GetType` was intended for. `GetType` should return something defined on `enum Value`. As in the system-type of the thing being accessed. When applied to the "source" it _should_ return `Map`. We treat the root thing as a "map" even though it is special.

For determining "signal type" why not just introduce a sort of virtual thing? Assuming somewhere you have mapping for "known" things like `serverity_text` or `severity_number` couldn't you just return a static if you see `signal_type` off the root?

albertlockett: > This is really not what GetType was intended for. GetType should return something defined on enum Value. As in the system-type of the thing being accessed. When applied to the "source" it should return Map. We treat the root thing as a "map" even though it is special.

In Open-Telemetry, a map is a different data type than a Log, a Trace, or a Metric. Map would be for something like attributes. I believe in the OTel context we shouldn't just be treating the root as a 'map',  so I don't actually think `GetType` should return 'map' when the source is one of the root telemetry signals.

> For determining "signal type" why not just introduce a sort of virtual thing? Assuming somewhere you have mapping for "known" things like serverity_text or severity_number couldn't you just return a static if you see signal_type off the root?

How would we use what you're suggesting wrt to `signal_type`? e.g. are you suggesting we have a virtual field on the root, and use it like this?
```kql
logs | where signal_type == "Log"
```
I see some issues with that::
- a) this seems strange because `signal_type` isn't part of the Open Telemetry data model
- b) we'd need to have special handling for this field when it's used outside a filter context (e.g. it can't be assigned).

All that said, I'm open to not using `GetType` here if it's not being used as you intended. Would you find it acceptable if I just parsed this as into a special external function that the OTAP query engine knew how to interpret?

albertlockett: Or were you thinking more like `signal_type` should be like an internal-only virtual column that is only used in the AST produced by the parser? E.g. internally we'd parse something like `is Log` into something like this:
```rs
LogicalExpression::EqualTo(EqualToLogicalExpression::new(
    SourceScalarExpression::Source(SourceScalarExpression::new(
        query_location,
        ScalarExpression::Source(SourceScalarExpression::new(
            query_location
            ValueAccessor::new_with_selectors(vec![
                StaticScalarExpression::String(StringScalarExpression::new(
                    query_location,
                    "signal_type"
                )),
            ]),
        )),
    )),
    ScalarExpression::Static(
            StaticScalarExpression::String(StringScalarExpression::new(
                to_query_location(&type_name_rule),
                type_name_rule.as_str(),
            )),
     )
)
```
And then the query-engine just knows to look for this special expression and treat it like checking the signal type?

CodeBlanch:
> In Open-Telemetry, a map is a different data type than a Log, a Trace, or a Metric. Map would be for something like attributes. I believe in the OTel context we shouldn't just be treating the root as a 'map', so I don't actually think GetType should return 'map' when the source is one of the root telemetry signals.

From OTel perspective I agree with you. Except expression tree doesn't know anything about OTel 😄 `GetType` in AST means "give me the query type of x." `GetType(severity_text)` should return `String`. `GetType(severity_number)` should be `Integer`. `GetType(attributes['attr1'])` would be whatever type is resolved at runtime. The idea returning `Map` for `GetType(source)` is it best represents the root thing. Not a perfect fit for sure.

> Would you find it acceptable if I just parsed this as into a special external function that the OTAP query engine knew how to interpret?

`signal_type` is an OPL thing so if you want to mount it to a special function all good with me.

> And then the query-engine just knows to look for this special expression and treat it like checking the signal type?

I'm OK with this too. Seems like heavy lifting having to parse the tree and look for it, but not impossible.

You could also introduce some new expression in AST if you wanted to. `GetRecordType()` or something along those lines. Doesn't accept any parameters just implicitly uses the "current context" aka "source" aka "the thing being processed currently" and returns a string identifying some friendly name for the root ("log", "metric", "trace", or plural versions is probably OK too).

albertlockett:
>  From OTel perspective I agree with you. Except expression tree doesn't know anything about OTel 😄 GetType in AST means "give me the query type of x." GetType(severity_text) should return String. GetType(severity_number) should be Integer. GetType(attributes['attr1']) would be whatever type is resolved at runtime. The idea returning Map for GetType(source) is it best represents the root thing. Not a perfect fit for sure.

Could we not consider that `GetType` is meant to "give me the query type of x." _in the context in which this expression is evaluated_? If the record set engine considers that returning `Map` for `GetType(source)` is the best thing, then that's what it can do. If the OTAP query engine wants knows that its source has definite type (`Log`/`Span`/`Metric`), why could it not return that? 

> > And then the query-engine just knows to look for this special expression and treat it like checking the signal type?
>
> I'm OK with this too. Seems like heavy lifting having to parse the tree and look for it, but not impossible.
>
> You could also introduce some new expression in AST if you wanted to. GetRecordType() or something along those lines.

I'm reticent to add yet another expression type like `GetRecordType` just for this purpose. Also would this not this have the same issue where, if the record set engine considers the source a Map, this `GetRecordType` expression would also return Map when it is evaluated over there?

It's not really too complex to parse this `signal_type` into a special LogicalExpression -- it's basically the same as what this PR is currently doing, I would just be looking for this virtual field name as opposed to looking for the `GetType` variant of `ScalarExpression`. I'll actually probably just move forward with this solution so we don't have to change what `GetType(source)` philosophically represents.

CodeBlanch:
> Could we not consider that GetType is meant to "give me the query type of x." in the context in which this expression is evaluated? If the record set engine considers that returning Map for GetType(source) is the best thing, then that's what it can do. If the OTAP query engine wants knows that its source has definite type (Log/Span/Metric), why could it not return that?

The idea is I can take some language. KQL, OPL, OTTL, MyCustomLang and spit out an expression tree. Then I can take that tree and run it using some engine. I could use KQL with RecordSet or OPL with QueryEngine. But what if I want to take KQL and use QueryEngine or take OPL and use RecordSet? The engines _shouldn't_ have different behavior. If it will only ever be possible to use OPL + QueryEngine and never any other combination fine to define a different behavior for GetType in QueryEngine.

> I'm reticent to add yet another expression type like GetRecordType just for this purpose. Also would this not this have the same issue where, if the record set engine considers the source a Map, this GetRecordType expression would also return Map when it is evaluated over there?

If you added `GetRecordType` I assume it would throw not_supported\todo in RecordSet engine. If we wanted to add support, it wouldn't be too difficult. There is a trait `Record` (IIRC) in RecordSet it would need some API (`fn get_record_type(&self) -> &str`) which the engine would use to retrieve the value when it sees a `GetRecordType()` in the tree.

albertlockett:
> > Could we not consider that GetType is meant to "give me the query type of x." in the context in which this expression is evaluated? If the record set engine considers that returning Map for GetType(source) is the best thing, then that's what it can do. If the OTAP query engine wants knows that its source has definite type (Log/Span/Metric), why could it not return that?
> 
> The idea is I can take some language. KQL, OPL, OTTL, MyCustomLang and spit out an expression tree. Then I can take that tree and run it using some engine. I could use KQL with RecordSet or OPL with QueryEngine. But what if I want to take KQL and use QueryEngine or take OPL and use RecordSet? The engines _shouldn't_ have different behavior. If it will only ever be possible to use OPL + QueryEngine and never any other combination fine to define a different behavior for GetType in QueryEngine.
> 
> > I'm reticent to add yet another expression type like GetRecordType just for this purpose. Also would this not this have the same issue where, if the record set engine considers the source a Map, this GetRecordType expression would also return Map when it is evaluated over there?
> 
> If you added `GetRecordType` I assume it would throw not_supported\todo in RecordSet engine. If we wanted to add support, it wouldn't be too difficult. There is a trait `Record` (IIRC) in RecordSet it would need some API (`fn get_record_type(&self) -> &str`) which the engine would use to retrieve the value when it sees a `GetRecordType()` in the tree.

Since there's a desire to maintain the consistent behaviour between engines, the safest behaviour will be to add this new `GetRecordType` as you've described. This is what I will do. Otherwise, record set engine would need to also have logic for the virtual `signal_type` field.

lquerel (approved):
LGTM

jmacd (approved):

PR #2755, opened 2026-04-24
feat(metrics): add scope_attributes to MetricSelector for view filtering

### LINKED ISSUE #2742: Expand MetricSelector to support scope_attribute view selectors
### Pre-filing checklist

### Component(s)

Follow-up to #2500 

The OpenTelemetry Rust SDK exposes 7 selector dimensions on `&Instrument` for view matching:

| Dimension | SDK accessor | Supported in otel-arrow |
|---|---|---|
| Instrument name | `instrument.name()` | **Yes** |
| Scope name | `instrument.scope().name()` | **Yes** |
| Instrument kind | `instrument.kind()` | No |
| Instrument unit | `instrument.unit()` | No |
| Scope version | `instrument.scope().version()` | No |
| Scope schema URL | `instrument.scope().schema_url()` | No |
| Scope attributes | `instrument.scope().attributes()` | No |

Our `MetricSelector` struct currently only supports `instrument_name` and `scope_name`. This limits the expressiveness of declarative view configuration.

https://github.com/open-telemetry/otel-arrow/blob/446e2510d411d9944a3a37faba577e2029677e8b/rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/views.rs#L19-L27

Similarly, `MetricStream` only supports `name` and `description`, while the SDK's `Stream::builder()` also supports `unit`, `aggregation`, `allowed_attribute_keys`, and `cardinality_limit`.

### Use Case

- **`scope_attributes`**: Filter by deployment environment or other scope-level metadata, e.g., only apply a view to instruments from scopes tagged with `env=production`.

### Immediate work (high priority)

Add `scope_attributes` to `MetricSelector`:

```rust
pub struct MetricSelector {
    pub instrument_name: Option<String>,
    pub scope_name: Option<String>,
    pub scope_attributes: Option<HashMap<String, String>>,  // new
}
```

And add corresponding filter clauses in `DeclarativeView::to_view_funtion()` in `views_provider.rs`.

Example YAML config:

```yaml
views:
  - selector:
      scope_name: "my.library"
      scope_attributes:
        env: "production"
    stream:
      description: "Production histograms"
```

### Low priority work

> [!WARNING]
> The support for OpenTelemetry Rust SDK views is not planned long-term in otap-dataflow, see #2623, which proposes eventually replacing views with equivalent ability in the internal telemetry system. Therefore, it is not super critical to implement the lower priority pieces of the issue unless there is an immediate use case.

#### Stream (lower priority, follow-up)

Expand `MetricStream` to support `unit`, `aggregation`, `allowed_attribute_keys`, and `cardinality_limit`.

#### Other Selectors (lower priority, follow-up)

`instrument_unit`, `scope_version`, and `scope_schema_url` could be added later if use cases arise. These are rarely needed in practice.

`instrument_kind` was originally considered a high priority item / easy win, but feedback from @cijothomas on this issue raised a good point that we may not want to support it, at least for now.

### Alternatives Considered

N/A

### Additional Context

* [Reference to opentelemetry-rust `instrument.rs`](https://github.com/open-telemetry/opentelemetry-rust/blob/8e95e16be60acc64b27f846598267f508b74ce5d/opentelemetry-sdk/src/metrics/instrument.rs#L110-L130)

pritishnahar95:
Expand MetricSelector to support scope_attributes as a new selector dimension. When configured, the view only applies to instruments whose scope contains all specified attribute key-value pairs.

# Change Summary

Add `scope_attributes: Option<HashMap<String, String>>` to `MetricSelector` in the config crate, and implement corresponding AND-based filter logic in `DeclarativeView::to_view_funtion()` in the telemetry crate. Each configured key-value pair must be present on the instrument's scope for the view to match.

* Closes #2742

- Added `test_view_config_with_scope_attributes` in `views.rs` to verify YAML deserialization of the new field.
- Added `test_views_provider_configure_with_scope_attributes` in `views_provider.rs` to verify that a view with scope attributes can be registered on the meter provider.
- Updated all existing test struct literals to include `scope_attributes: None`.

Yes. Users can now configure `scope_attributes` in their YAML view selectors to filter by scope-level metadata:

```yaml
views:
  - selector:
      scope_name: "my.library"
      scope_attributes:
        env: "production"
    stream:
      description: "Production histograms"
```

### DIFF
--- a/rust/otap-dataflow/configs/fake-debug-noop-telemetry.yaml
+++ b/rust/otap-dataflow/configs/fake-debug-noop-telemetry.yaml
@@ -15,8 +15,6 @@ engine:
         - selector:
             scope_name: "fake_data_generator.receiver.metrics"
             instrument_name: "logs.produced"
-            scope_attributes:
-              feature_flag: "experimental"
           stream:
             name: "otlp.logs.produced.count"
             description: "Count of logs produced"
--- a/rust/otap-dataflow/configs/fake-debug-noop-telemetry.yaml
+++ b/rust/otap-dataflow/configs/fake-debug-noop-telemetry.yaml
@@ -16,7 +16,7 @@ engine:
             scope_name: "fake_data_generator.receiver.metrics"
             instrument_name: "logs.produced"
             scope_attributes:
-              version: "0.0.1"
+              feature_flag: "experimental"
           stream:
             name: "otlp.logs.produced.count"
             description: "Count of logs produced"
--- a/rust/otap-dataflow/configs/fake-debug-noop-telemetry.yaml
+++ b/rust/otap-dataflow/configs/fake-debug-noop-telemetry.yaml
@@ -15,6 +15,8 @@ engine:
         - selector:
             scope_name: "fake_data_generator.receiver.metrics"
             instrument_name: "logs.produced"
+            scope_attributes:
+              version: "0.0.1"
           stream:
             name: "otlp.logs.produced.count"
             description: "Count of logs produced"
--- a/rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/views.rs
+++ b/rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/views.rs
@@ -8,6 +8,8 @@ use std::collections::HashMap;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
+use crate::pipeline::telemetry::AttributeValue;
+
 /// OpenTelemetry Metrics View configuration.
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema, PartialEq)]
 pub struct ViewConfig {
@@ -29,7 +31,8 @@ pub struct MetricSelector {
     /// The instrumentation scope attributes to match.
     /// When set, the view only applies to instruments whose scope contains all
     /// of the specified attribute key-value pairs.
-    pub scope_attributes: Option<HashMap<String, String>>,
+    #[serde(default)]
+    pub scope_attributes: HashMap<String, AttributeValue>,
 }
 
 /// OpenTelemetry Metric Stream configuration.
@@ -134,9 +137,12 @@ mod tests {
             "#;
         let config: ViewConfig = serde_yaml::from_str(yaml_str).unwrap();
         assert_eq!(config.selector.scope_name.as_deref(), Some("my.library"));
-        let attrs = config.selector.scope_attributes.unwrap();
+        let attrs = &config.selector.scope_attributes;
         assert_eq!(attrs.len(), 1);
-        assert_eq!(attrs.get("feature_flag").unwrap(), "experimental");
+        assert_eq!(
+            attrs.get("feature_flag").unwrap(),
+            &AttributeValue::String("experimental".to_string())
+        );
         assert_eq!(
             config.stream.description.as_deref(),
             Some("Experimental library metrics")
--- a/rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/views.rs
+++ b/rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/views.rs
@@ -128,7 +128,6 @@ mod tests {
             selector:
               scope_name: "my.library"
               scope_attributes:
-                version: "0.0.1"
                 feature_flag: "experimental"
             stream:
               description: "Experimental library metrics"
@@ -136,8 +135,7 @@ mod tests {
         let config: ViewConfig = serde_yaml::from_str(yaml_str).unwrap();
         assert_eq!(config.selector.scope_name.as_deref(), Some("my.library"));
         let attrs = config.selector.scope_attributes.unwrap();
-        assert_eq!(attrs.len(), 2);
-        assert_eq!(attrs.get("version").unwrap(), "0.0.1");
+        assert_eq!(attrs.len(), 1);
         assert_eq!(attrs.get("feature_flag").unwrap(), "experimental");
         assert_eq!(
             config.stream.description.as_deref(),
--- a/rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/views.rs
+++ b/rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/views.rs
@@ -3,6 +3,8 @@
 
 //! Metrics views level configurations.
 
+use std::collections::HashMap;
+
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
@@ -24,6 +26,10 @@ pub struct MetricSelector {
     /// The instrumentation scope (meter) name to match.
     /// When set, the view only applies to instruments created under this scope.
     pub scope_name: Option<String>,
+    /// The instrumentation scope attributes to match.
+    /// When set, the view only applies to instruments whose scope contains all
+    /// of the specified attribute key-value pairs.
+    pub scope_attributes: Option<HashMap<String, String>>,
 }
 
 /// OpenTelemetry Metric Stream configuration.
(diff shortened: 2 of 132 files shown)


rust/otap-dataflow/crates/telemetry/src/otel_sdk/meter_provider/views_provider.rs
```diff
@@ -54,6 +54,18 @@ impl DeclarativeView {
                 return None;
             }
 
+            if let Some(scope_attributes) = &config.selector.scope_attributes {
```
cijothomas: consider modify one existing example config to show how this can be used.

rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/views.rs
```diff
@@ -115,4 +121,27 @@ mod tests {
...
         );
     }
+
+    #[test]
+    fn test_view_config_with_scope_attributes() {
+        let yaml_str = r#"
+            selector:
+              scope_name: "my.library"
+              scope_attributes:
+                env: "production"
```
cijothomas: nit: these are typically Resource level attributes, not scope level. Immaterial for tests, but just sharing!

rust/otap-dataflow/configs/fake-debug-noop-telemetry.yaml
```diff
@@ -15,6 +15,8 @@ engine:
         - selector:
             scope_name: "fake_data_generator.receiver.metrics"
             instrument_name: "logs.produced"
+            scope_attributes:
+              version: "0.0.1"
```
cijothomas: version is not stored as attribute, but a top-level field, like scope_name. Are you sure this works as-is?

tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker-filtered.yaml
```diff
@@ -116,7 +116,7 @@ steps:
       run:
         pre:
           - send_http_request:
-              url: http://localhost:8085/api/v1/pipeline-groups/shutdown?wait=true&timeout_secs={{drain_timeout_secs | default(60)}}
+              url: http://localhost:8085/api/v1/groups/shutdown?wait=true&timeout_secs={{drain_timeout_secs | default(60)}}
```
cijothomas, line 119: thanks for fixing this. I generally recommend to make it to own PR, to make it easy to review PRs (and also avoid unrealated changes getting stuck on each other!

rust/otap-dataflow/configs/fake-debug-noop-telemetry.yaml
```diff
@@ -15,6 +15,8 @@ engine:
         - selector:
             scope_name: "fake_data_generator.receiver.metrics"
             instrument_name: "logs.produced"
+            scope_attributes:
+              feature_flag: "experimental"
```
ThomsonTan: This example looks like it will not match any current internal metric. The dispatcher creates the meter with `global::meter(descriptor.name)`, so the scope has the metric-set name but no `feature_flag` attribute; node/entity attributes are recorded as data-point attributes. Can the `scope_attributes` be moved to a separate example backed by a real `InstrumentationScope::builder(...).with_attributes(...)` meter?

cijothomas: it is okay to keep the same example itself, just use a valid attribute.

rust/otap-dataflow/crates/telemetry/src/otel_sdk/meter_provider/views_provider.rs
```diff
@@ -54,6 +54,18 @@ impl DeclarativeView {
...
+                let scope_attrs: Vec<_> = instrument.scope().attributes().collect();
```
cijothomas: this allocates a new vector during instrument creation. I'd normally consider that okay, given instrument creation is meant to be done once and re-used. However, this repo does not follow that practice, and keep creating instruments during each reporting. 
its best to avoid the allocation here.

rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/views.rs
```diff
@@ -115,4 +121,25 @@ mod tests {
...
+                feature_flag: "experimental"
+            stream:
+              description: "Experimental library metrics"
+            "#;
+        let config: ViewConfig = serde_yaml::from_str(yaml_str).unwrap();
+        assert_eq!(config.selector.scope_name.as_deref(), Some("my.library"));
+        let attrs = config.selector.scope_attributes.unwrap();
+        assert_eq!(attrs.len(), 1);
+        assert_eq!(attrs.get("feature_flag").unwrap(), "experimental");
+        assert_eq!(
```
cijothomas, line 146: the test is not actually executing the code path of applying view selector via scope... its an existing issue in the repo - not a blocker for this PR, but ideally we should make sure test coverage is good.

pritishnahar95: In lieu of the other important bug fix going in this PR, will skip this particular point since this is a wider problem. Sounds like it needs a separate issue to be created.

rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/views.rs
```diff
@@ -24,6 +26,10 @@ pub struct MetricSelector {
     /// The instrumentation scope (meter) name to match.
     /// When set, the view only applies to instruments created under this scope.
     pub scope_name: Option<String>,
+    /// The instrumentation scope attributes to match.
+    /// When set, the view only applies to instruments whose scope contains all
+    /// of the specified attribute key-value pairs.
+    pub scope_attributes: Option<HashMap<String, String>>,
```
jmacd: The resource argument configured in config/src/pipeline/telemetry.rs uses 

```
    /// Resource attributes to associate with telemetry data.
    #[serde(default)]
    pub resource: HashMap<String, AttributeValue>,
```

I wonder if we should prefer the same here? (`HashMap<String, AttributeValue>` w/o Option)?

rust/otap-dataflow/configs/fake-debug-noop-telemetry.yaml
pritishnahar95: Skipping this for now, since we dont have any attributes added to scope presently.

rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/views.rs
pritishnahar95: The other fields in Metric Selector all use Option, so I was using that for consistency, but I can make the switch.
I am also willing to switch to AttributeValue. I do have a question though.
How do we want to handle array matching when an attribute value is an array? 
1. Should array specified in selector - scope attribute be a subset of the array values in the actual metric's scope attribute? 
2. Should it be an exact match?
3. Do we not support arrays in scope attribute values, since that is unlikely to be used anyway?

rust/otap-dataflow/crates/telemetry/src/otel_sdk/meter_provider/views_provider.rs
```diff
@@ -54,6 +54,18 @@ impl DeclarativeView {
...
+                for (key, value) in scope_attributes {
+                    let matched = scope_attrs
+                        .iter()
+                        .any(|kv| kv.key.as_str() == key && kv.value.to_string() == *value);
```
lalitb: `Value::to_string()` allocates for every comparison. Since werebuilds SDK instruments during telemetry reporting, this predicate can run repeatedly. I believe there would be easy way to avoid the allocation.

pritishnahar95: Thanks for catching this! This is outdated code now, addressed by refactor in new commit.

rust/otap-dataflow/crates/telemetry/src/otel_sdk/meter_provider/views_provider.rs
pritishnahar95: addressed

cijothomas (approved):
LGTM, need to rebase from main and fix CI/link issues(minor) before merge.

rust/otap-dataflow/crates/config/src/pipeline/telemetry/metrics/views.rs
pritishnahar95: going to skip array support for now, since IMO this seems unlikely to be used. We can add support later if needed.

jmacd: I see! Yes, separate confusion over when a container needs an option. Thank you.
Array support definitely is not important.

jmacd (approved):

PR #2756, opened 2026-04-24
Add dfctl CLI and TUI for OTAP Dataflow Engine administration

lquerel:
# Change Summary

Adds `dfctl`, an admin SDK-based command-line tool for operating local and remote OTAP Dataflow Engines.

The CLI supports engine, group, pipeline, telemetry, rollout, shutdown, reconfiguration, diagnosis, bundle, watch, shell completion, automation friendly output, and an interactive TUI for operational workflows.

More details can be find here -> https://github.com/lquerel/otel-arrow/blob/962a01e30116433e448ed58b6d8b820e1bcdcd3a/rust/otap-dataflow/crates/enginectl/README.md

N/A

- `cargo xtask check`

Yes, the new CLI by itself.

image

image


rust/otap-dataflow/crates/enginectl/src/commands/catalog.rs
```diff
@@ -0,0 +1,1172 @@
...
+    })
+}
+
+/// Builds a complete catalog snapshot from the current Clap command tree.
+///
+/// This is intentionally recomputed instead of cached or generated at build
+/// time so the catalog always reflects the binary being executed.
+fn build_catalog() -> CommandCatalog {
+    let mut root = Cli::command();
+    root.build();
```
lalitb, line 77: Nice, I like the idea of generating the command catalog from Clap!

rust/otap-dataflow/crates/enginectl/src/ui/app/recipes.rs
```diff
@@ -0,0 +1,971 @@
...
+//! current TUI pane, making the UI a learning surface for automation. This
+//! module also owns modal transitions that are closely tied to those actions,
+//! such as opening action menus, staging config drafts, and preparing
+//! confirmation or scale editors.
+
+use super::*;
+
+impl AppState {
+    /// Builds the equivalent CLI command recipe for the active view and tab.
+    pub(crate) fn current_command_recipe(&self) -> CommandRecipe {
```
lalitb, line 16: Really like this. The TUI showing the matching `dfctl` command makes it a bridge to scripts and agents, not just a UI. And good call to redact the client key path.

lalitb (approved):
High-level pass looks good to me. The `dfctl` code is isolated as a separate crate/binary and appears to use the admin SDK rather than reaching into the engine/controller internals, so I don’t see an obvious regression risk for the existing runtime path.

I like the overall shape: it works for humans, scripts, and agent workflows from the start. The TUI showing the matching dfctl command, while redacting sensitive args, is a nice touch.

rust/otap-dataflow/crates/enginectl/src/ui/app/recipes.rs
lquerel: We have to be ready for the modern days!

rust/otap-dataflow/crates/enginectl/src/commands/catalog.rs
```diff
@@ -0,0 +1,1172 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Generated command catalog for humans, scripts, and AI agents.
```
jmacd, line 4: 🚀

jmacd (approved):
This looks useful, well thought out, and manually tested, and it is separate from the engine code in a way that was designed. Since I recall suggesting `dfctl` making sense after the admin client API was introduced, makes sense to me.

PR #2758, opened 2026-04-25
[fix] correct filter using expressions where each side returns nulls and/or mixed types

### LINKED ISSUE #1394: [experimental] Columnar query engine initial implementation
In https://github.com/open-telemetry/otel-arrow/pull/1342 we implemented a version of our columnar query engine (which I'm still somewhat hacking on). I thought it would be good to have an issue to track future work on this including feature gaps, performance optimizations, future ideation/experimentation, etc.

(note: list below is still a work in progress. what's documented below is not a comprehensive list of TODOs. More may be added to it as future and for some of these we can/will make dedicated issues).

**Feature gaps:**
- Additional signal support
   - [ ] Metrics
   - [ ] Traces
- Attribute transformations:
   - [ ] Set attribute
     - [ ] from literals(`extend attributes["X"] ="Y"`)
     - [ ] from other expressions (things like `extend attributes["event"] = <some_field> / some_func(<some_field>) / <etc.>`)
   - [ ] Rename attributes (`project-rename`)
   - [ ] Drop attributes (`project-away`)
- Filtering:
  - [x] Literal on RHS of binary expression not supported (`where "WARN" == severity_text` does not work)
  - [ ] Filter by body

**Plan Construction & Execution:** (I will add issues to expand on the bullet points in near-future)
- [x] #1417 
- [x] #1409 
- [x] #1410
- [x] ~Perf: remove the `window(row_number())` on the root batch scan~ (resolved in #1477, this is no longer needed)
- [ ] #1390 (Optimize the post-filtering applied to child batches)
- [x] #1494 

**Further Exploration**
- [ ] supporting nested streams pipelines (iterating metrics datapoints)
- [ ] filtering by nested streams (filter metrics by datapoints)
- [ ] support fork (copy data and send down two pipelines)

### LINKED ISSUE #2659: Columnar Query engine comparing fields with different types/nulls does not have correct behaviour
Since https://github.com/open-telemetry/otel-arrow/pull/2632 we have the capability to use any expressions in logical expressions in filters in the columnar query engine. For example:
```
logs | where severity_text == <some expr>
```

If the types on the left or the right don't evaluate to the same type, we return throw an execution error at runtime
```
logs | where severity_text == severity_number
```

I think the correct behaviour would probably be for this to resolve as false. Then we could also maybe catch these using static analysis and fold this a static `false` (we can only do this if we know the type a-priori, won't work for all expressions).

Similarly, if we have expressions the evaluate to null on both sides. E.g. `attributes["x"] == attributes["y"]` where both these attributes are null, currently evaluates to `false` but I feel it should evaluate to `true`.

albertlockett:
# Change Summary

Corrects the behaviour of filtering expressions where each side of a relative expression returns either nulls, types that are not equivalent and/or struct columns representing `AnyValues`.

For example, in a program like this:
```kql
logs | where attributes["x"] == attributes["y"]
```

In this case, we don't know at compile time what is the type of the left or right side of the `==` expression, so we evaluate the expressions on either side and then attempt to compare the results. Before this change, we were effectively just naively using the `eq` compute kernel (technically this would get called inside the [`BinaryExpr`](https://docs.rs/datafusion/latest/datafusion/physical_expr/expressions/struct.BinaryExpr.html)). This means that we'd produce incorrect results or errors in cases where:
- the type on the left / right were not the same (we returned an error)
- both sides were null (we return `false`, but `null == null`should return `true`)
- if one side was a struct representing an `AnyValue` (such as log body, although attributes may also be coerced into a struct like this), we would also just return an error.

This PR changes the implementation to use a special `compare` kernel which handles all these cases. Effectively the logic it applies for each row would be:
- if the types do not match, evaluate as false
- null handling:
  - if both sides are `null` and the operator is `==`, return `true`
  - if exactly one side is `null` and the operator is `!=` return `true`
  - otherwise return `false`

The `compare` handles comparing the appropriate values columns for `AnyValue`s represented as struct columns. It also makes the determination if one side is effectively a `null` for purpose of comparison -- it considers the row null when any of the following conditions are met:
- the struct column is null
- the value in the type column is `Empty`
- the value columns elected by a non-`Empty` type column value is null

* Closes #2659
* Closes #1394

Unit tests

No - this is basically just a bug fix for how OTAP query engine handles some filter expressions.

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
@@ -61,8 +61,8 @@ use crate::pipeline::expr::{
 };
 use crate::pipeline::planner::{AttributesIdentifier, ColumnAccessor};
 use crate::pipeline::project::anyval::{
-    fill_null_type_as_empty, is_any_value_data_type, project_any_value_columns,
-    wrap_as_any_value_struct,
+    attempt_coerce_value_column_from_any_value_struct_column, fill_null_type_as_empty,
+    is_any_value_data_type, wrap_as_any_value_struct,
 };
 use crate::pipeline::project::{ProjectedSchemaColumn, Projection};
 use crate::pipeline::state::ExecutionState;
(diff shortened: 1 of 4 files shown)


lquerel (approved):
LGTM

PR #2763, opened 2026-04-27
Fix batch byte sizing and wakeup state

lquerel:
# Change Summary

Fixes batch processor behavior for OTLP byte-sized batches by comparing `min_size` against pending bytes instead of item count. Also avoids unnecessary wakeup set/cancel work for immediate size flushes, clears stale timer state after full drains, and adds wakeup scheduler metrics for attribution.

* Closes #NNN

- cargo xtask check
- controlled benchmark

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
@@ -53,7 +53,7 @@ use otap_df_pdata::{
     OtapArrowRecords, OtapPayload, OtapPayloadHelpers, OtlpProtoBytes, error::Error as PDataError,
     otap::batching::make_item_batches, otlp::batching::make_bytes_batches,
 };
-use otap_df_telemetry::instrument::Counter;
+use otap_df_telemetry::instrument::{Counter, Mmsc};
 use otap_df_telemetry::metrics::MetricSet;
 use otap_df_telemetry_macros::metric_set;
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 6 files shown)


rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
```diff
@@ -539,6 +542,31 @@ pub struct BatchProcessorMetrics {
...
+    flush_timer_lateness_duration: Mmsc,
+    /// Number of output batches emitted by each flush
+    #[metric(unit = "{batch}")]
+    flush_output_batches: Mmsc,
+    /// Number of primary signal items emitted by each flush
+    #[metric(unit = "{item}")]
+    flush_output_items: Mmsc,
+    /// Number of bytes emitted by each flush when byte size is known
+    #[metric(unit = "By")]
+    flush_output_bytes: Mmsc,
```
lquerel, line 568: Maybe in the future we could move these metrics to `MetricLevel::Detailed`, but for now I've kept them to make analyzing the wakeup API easier.

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
```diff
@@ -879,23 +907,29 @@ where
...
 
         self.buffer
             .inputs
             .accept(payload, BatchPortion::new(inkey, items));
 
+        let pending_size = self.buffer.inputs.size_by(self.fmtcfg.sizer)?;
+
         // Flush based on size when the batch reaches the lower limit.
-        if timeout != Duration::ZERO && self.buffer.inputs.items < self.fmtcfg.lower_limit() {
+        if timeout != Duration::ZERO && pending_size < self.fmtcfg.lower_limit() {
```
lquerel, line 927: The comparison wasn't valid because it was comparing a number of items to a quantity of bytes. As a result, the batch trigger based on byte size was never firing.

JakeDern: For what it's worth, there are probably more issues with bytes and/or request-based sizers. The lower-level batching code (concatenation/splitting) doesn't deal with bytes yet.

JakeDern: (I considered these features unimplemented in #1926)

JakeDern (approved):

PR #2764, opened 2026-04-27
Otap exporter stream concurrency

lquerel:
# Change Summary

Improves OTAP exporter progress under backpressure by draining completion updates while waiting for stream queue capacity. Adds async attribution metrics for stream enqueue, encode, correlation, response wait, inflight depth, and latest-window export latency percentiles.

Part of https://github.com/open-telemetry/otel-arrow/issues/1324

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
@@ -371,9 +371,14 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
                 .accept_compressed(encoding);
         }
 
-        // TODO comment on the purpose of these
-        // TODO import so can use as just "channel" here
-        // TODO check if we can use our local channel since we are already using `tokio::task::spawn_local`.
+        // Each signal type is exported through one or more long-lived stream
+        // workers. The exporter task only converts incoming pdata and enqueues
+        // it into a bounded per-stream queue; the worker owns the gRPC request
+        // stream and response correlation for that queue.
+        //
+        // Keeping these queues bounded preserves backpressure. Increasing
+        // `streams_per_signal` adds independently progressing gRPC streams
+        // instead of hiding pressure behind a deeper single queue.
         let stream_queue_capacity = self.config.stream_queue_capacity;
         let streams_per_signal = self.config.streams_per_signal;
         let (pdata_metrics_tx, mut pdata_metrics_rx) = tokio::sync::mpsc::channel(64);
@@ -384,7 +389,9 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
         )
         .then(|| arrow_ipc::CompressionType::ZSTD);
 
-        // TODO check if we can expose/use spawn_local method in the effect handler
+        // Tonic clients are cheap to clone because they share the underlying
+        // Channel. Each clone below is used by exactly one worker, which lets
+        // the workers drive separate streaming RPCs concurrently.
         let (logs_senders, logs_handles) = spawn_stream_workers(
             arrow_logs_client,
             SignalType::Logs,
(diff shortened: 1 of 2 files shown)


rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
```diff
@@ -53,6 +56,96 @@ use config::Config;
...
+    /// Median outbound gRPC export response duration for the latest telemetry interval.
+    #[metric(name = "export.rpc.duration.p50", unit = "ns")]
+    pub export_rpc_duration_p50_ns: Gauge<f64>,
+    /// 90th percentile outbound gRPC export response duration for the latest telemetry interval.
+    #[metric(name = "export.rpc.duration.p90", unit = "ns")]
+    pub export_rpc_duration_p90_ns: Gauge<f64>,
+    /// 99th percentile outbound gRPC export response duration for the latest telemetry interval.
+    #[metric(name = "export.rpc.duration.p99", unit = "ns")]
+    pub export_rpc_duration_p99_ns: Gauge<f64>,
+}
```
lquerel, line 102: I will move these metrics in MetricLevel::Detailed once the benchmark season will be behind us...

albertlockett (approved):
LGTM - nice addition

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
albertlockett: Sounds good. With `MetricLevel::Detailed` we'd be able to avoid the computation overhead of computing the histogram quantiles if the user isn't interested in these detailed metrics?

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
```diff
@@ -245,11 +475,23 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
...
+                            SignalType::Traces => least_loaded_stream_sender(&traces_senders),
                         };
+                        self.enqueue_stream_batch(
+                            sender,
+                            pdata,
+                            message,
+                            &mut pdata_metrics_rx,
+                            &effect_handler,
+                        )
+                        .await?;
```
lalitb, line 494: Small concurrency question: we pick the least-loaded stream once, then pass only that sender into enqueue_stream_batch. If that queue is full, we wait on that queue’s `reserve()`. Could this still block behind one stream even if another stream for the same signal type frees up first? Wondering if we should re-check the least-loaded queue after a full queue wait, or if this is an intentional simplification.

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
```diff
@@ -471,27 +845,48 @@ async fn handle_res_stream(
...
-                        if let Some(pdata) = correlation_rx.recv().await {
-                            _ = pdata_metrics_tx.send(PDataMetricsUpdate::IncFailed(signal_type, pdata)).await;
+                        if let Some(correlated) = correlation_rx.recv().await {
+                            _ = pdata_metrics_tx
+                                .send(PDataMetricsUpdate::IncFailed(
+                                    signal_type,
+                                    correlated.pdata,
+                                    Some(elapsed_nanos(correlated.sent_at)),
+                                ))
+                                .await;
```
lalitb, line 889: This may be pre-existing, but since this PR is touching the response correlation path: on a stream error we still fail only one correlated pdata and then break. Could there be additional already-yielded batches in `correlation_rx` that would be dropped without ACK/NACK?

lalitb (approved):
LGTM. The comments are more of edge case, and pre-existing - nothing blocking. Thanks.

jmacd:
FYI https://github.com/open-telemetry/otel-arrow/issues/1324 maybe fixed

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
jmacd: This should be a Histogram, IMO we're just not there yet. https://github.com/jmacd/rust-expohisto

jmacd (approved):

PR #2765, opened 2026-04-27
Improve OTAP exporter stream progress metrics

lquerel:
# Change Summary

Adds configurable `streams_per_signal` support to run multiple OTAP gRPC streams per signal type. Batches are routed to the least-loaded stream queue, allowing large-payload workloads to reduce single-stream response-path bottlenecks.

* Closes #NNN

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
@@ -34,11 +34,13 @@ use otap_df_pdata::proto::opentelemetry::arrow::v1::{
     arrow_metrics_service_client::ArrowMetricsServiceClient,
     arrow_traces_service_client::ArrowTracesServiceClient,
 };
+use otap_df_telemetry::instrument::{Gauge, Mmsc};
 use otap_df_telemetry::metrics::MetricSet;
 use otap_df_telemetry::{otel_error, otel_info};
+use otap_df_telemetry_macros::metric_set;
 use serde_json::Value;
 use std::sync::Arc;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use tokio::sync::mpsc::{Receiver, Sender};
 use tonic::transport::Channel;
 use tonic::{IntoStreamingRequest, Response, Status, Streaming};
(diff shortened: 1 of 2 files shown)


albertlockett (commented):
@lquerel  Is this a duplicate of https://github.com/open-telemetry/otel-arrow/pull/2764, just with fewer commits?

lquerel:
Yes, one is built on the other.

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
```diff
@@ -227,7 +446,12 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
                         _ = metrics_handle.await;
                         _ = traces_handle.await;
                         _ = timer_cancel_handle.cancel().await;
-                        return Ok(TerminalState::new(deadline, [self.pdata_metrics]))
+                        self.export_latency_window
+                            .report_into(&mut self.async_metrics);
+                        return Ok(TerminalState::new(
+                            deadline,
+                            [self.pdata_metrics.snapshot(), self.async_metrics.snapshot()],
+                        ))
```
lalitb, line 454: Small shutdown concern: this may already be an existing issue, but the new async metric updates make it more visible. After the stream tasks are awaited, there may still be `PDataMetricsUpdate`s sitting in `pdata_metrics_rx`. Should we drain those before reporting/snapshotting the final metrics?

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
```diff
@@ -101,6 +196,126 @@ impl OTAPExporter {
...
+                    return Ok(());
+                }
+                Err(tokio::sync::mpsc::error::TrySendError::Full(item)) => {
+                    pending = Some(item);
+                }
+                Err(tokio::sync::mpsc::error::TrySendError::Closed(_item)) => {
+                    self.async_metrics
+                        .stream_enqueue_duration_ns
+                        .record(elapsed_nanos(enqueue_start));
+                    return Ok(());
```
lalitb, line 287: This silent drop may have existed before too, but this new enqueue path makes it easier to handle. If the stream queue is closed, `_item` still has the `pdata`, so I think we should mark it failed and send a NACK instead of returning Ok(())

jmacd (approved):

jmacd:
#2764 is in the merge queue, can we close this?

lquerel:
#2764 merged

PR #2766, opened 2026-04-27
Improve smooth load generator observability

lquerel:
# Change Summary

Improves smooth load generator pacing by using nanosecond intervals and skipped missed ticks. Adds smooth-mode observability for behind runs, remaining work, tick lateness, payload generation, and send backpressure. Also adds the performance experiment log and diagrams documenting the methodology and conclusions.

* Closes #NNN


rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
```diff
@@ -137,29 +162,83 @@ impl FakeGeneratorReceiver {
...
-                                remaining=current_run.len(),
+                                remaining=remaining_batches,
+                                remaining_items,
                             );
+                        } else if !run_completed {
+                            self.metrics.smooth_runs_completed.inc();
+                            run_completed = true;
+                        }
+
+                        if next_pdata.is_some() {
```
JakeDern, line 187: Is the intent here to not move into the next run until the current run is complete? I think we have two different behaviors with that depending on whether `next_pdata` is `None` or `Some`. 

If it's `None` then we'll break and go onto the next run, but `current_run.next()` might still return `Some()` so we won't finish the current run even if there are still items left.

if `next_pdata` is `Some` then we skip the reset for this tick and use the entire next tick to try and finish the current run.

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
```diff
@@ -137,29 +162,83 @@ impl FakeGeneratorReceiver {
...
 
-                        let Some(payload) = current_run.next() else {
-                            continue;
+                        let channel_result = match next_pdata.take() {
+                            Some(pdata) => {
+                                self.metrics.smooth_payload_send_retry.inc();
+                                let send_start = StdInstant::now();
+                                let result = self.export_pdata(handler, pdata)?;
+                                self.metrics
+                                    .smooth_payload_send_duration_ns
```
JakeDern, line 207: Do we expect the send duration to vary given that we're using the non-blocking api?

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
```diff
@@ -448,18 +527,22 @@ impl local::Receiver<OtapPdata> for FakeGeneratorReceiver {
             .await?;
 
         let run_len = producer.run_len();
-        let batch_duration_millis = 1000u64.div_euclid(run_len as u64);
 
         // We consume one tick here because it's always immediately ready and would
         // make us think we're lagging;
         let mut run_ticker = interval(Duration::from_secs(1));
+        run_ticker.set_missed_tick_behavior(MissedTickBehavior::Skip);
```
JakeDern, line 534: I think this is a good change - We're not supposed to be playing catchup in-between runs

rust/otap-dataflow/crates/core-nodes/src/receivers/fake_data_generator/mod.rs
```diff
@@ -137,29 +162,83 @@ impl FakeGeneratorReceiver {
...
+                    scheduled = batch_ticker.tick() => {
+                        let tick_lateness = tokio::time::Instant::now()
+                            .saturating_duration_since(scheduled);
+                        self.metrics
+                            .smooth_batch_tick_lateness_duration_ns
+                            .record(duration_nanos(tick_lateness));
 
-                        let Some(payload) = current_run.next() else {
-                            continue;
+                        let channel_result = match next_pdata.take() {
```
lalitb, line 201: Nice, this makes the full-channel case much easier to see. Instead of just skipping a tick, we keep the payload and retry it, so backpressure shows up in the new metrics. One small thing to double-check: whether a run that was already marked behind can later also get counted as completed.

lalitb (approved):

jmacd (approved):

PR #2767, opened 2026-04-27
Remove unused dependencies

utpilla:
# Change Summary

Remove unused dependencies

### DIFF
--- a/rust/otap-dataflow/Cargo.toml
+++ b/rust/otap-dataflow/Cargo.toml
@@ -36,11 +36,8 @@ otap-df-controller.workspace = true
 otap-df-core-nodes.workspace = true
 otap-df-engine.workspace = true
 otap-df-otap.workspace = true
-thiserror.workspace = true
-serde_json.workspace = true
 clap.workspace = true
 mimalloc = { workspace = true, optional = true }
-sysinfo.workspace = true
 dhat = { workspace = true, optional = true }
 cfg-if.workspace = true
 
--- a/rust/otap-dataflow/benchmarks/Cargo.toml
+++ b/rust/otap-dataflow/benchmarks/Cargo.toml
@@ -18,7 +18,6 @@ otap-df-otap = { workspace = true }
 [dev-dependencies]
 criterion = { workspace = true, features = ["html_reports", "async_tokio"] }
 tonic = { workspace = true }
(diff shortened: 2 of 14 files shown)


lalitb (approved):
Nice. Would be good to add [udeps](https://github.com/est31/cargo-udeps) in CI as followup to restrict adding unused dependencies.

cijothomas (approved):
Nice! Agree with Lalit's suggestion of enforcing this at CI itself to prevent accidental additions in future.

utpilla:
> Nice. Would be good to add [udeps](https://github.com/est31/cargo-udeps) in CI as followup to restrict adding unused dependencies.

I looked into both `cargo-udeps` and `cargo-machete`:

`cargo-udeps` is more accurate (compiler-based) but requires nightly. We'd need to pin a nightly date and run it as `continue-on-error` since nightly can break independently. Also, our mutually exclusive crypto features (`crypto-ring` / `crypto-aws-lc` / `crypto-symcrypt`) mean we can't just use `--all-features`. We'd need multiple runs or pick a default feature set.

`cargo-machete` Runs on stable, fast (text scan, no compile), can be a hard gate. Has some false positives with `serde` attributes (e.g., `humantime-serde`) but those are handled with `[package.metadata.cargo-machete] ignored = [...]`.

Maybe we should use `cargo-machete` as the CI gate and occasional `cargo-udeps runs` locally for deeper checks?

lalitb:
> Maybe we should use `cargo-machete` as the CI gate and occasional `cargo-udeps runs` locally for deeper checks?

Also have a look into `cargo-shear` - we use this in otel-rust in place of cargo-machete - https://github.com/open-telemetry/opentelemetry-rust/pull/3247.

utpilla:
> > Maybe we should use `cargo-machete` as the CI gate and occasional `cargo-udeps runs` locally for deeper checks?
> 
> Also have a look into `cargo-shear` - we use this in otel-rust in place of cargo-machete - [open-telemetry/opentelemetry-rust#3247](https://github.com/open-telemetry/opentelemetry-rust/pull/3247).

@lalitb Yeah that's a better idea.

albertlockett (approved):

lquerel (approved):

PR #2772, opened 2026-04-28
Isolation of setup-protoc jobs from RUST-CI

### LINKED ISSUE #2768: Isolate `protoc` install into a targeted Rust-CI job
### Pre-filing checklist

### Component(s)

CI / Build

### Objective

As noted in [the otel-arrow-dev slack](https://cloud-native.slack.com/archives/C08RRSJR7FD/p1777317282006939), we really only need `protoc` installed if we were going to run a targeted `cargo xtask compile-proto` to detect:
* `.proto` changed without regenerating `.rs`
* manual edits to generated `.rs` without matching `.proto` changes

We don't need to run the [setup-protoc](https://github.com/arduino/setup-protoc) action in most of our Rust-CI jobs.

### Rationale

Simplify / speed up majority of CI jobs that don't need `protoc`

### Scope

- Remove `setup-protoc` from `Rust-CI` jobs that don't need it
- Add a targeted `compile_proto` job (similar to [`structure_check`](https://github.com/open-telemetry/otel-arrow/blob/bc4cec8b9dd31adbf3a207903eb652c16499467c/.github/workflows/rust-ci.yml#L379-L401)) that keeps the `setup-protoc` and also runs:
  > cargo xtask compile-proto
  git diff --exit-code rust/otap-dataflow/crates/pdata/src/proto

### Acceptance Criteria

PR making above scoped changes is merged

### Dependencies or Blockers

_No response_

### Additional Context

_No response_

luckymachi:
# Change Summary

As mentioned in issue [2768](https://github.com/open-telemetry/otel-arrow/issues/2768), `setup-protoc` jobs are dropped in favour of a targeted `compile_proto` job.

This issue closes issue [2768](https://github.com/open-telemetry/otel-arrow/issues/2768)


.github/workflows/rust-ci.yml
```diff
@@ -400,12 +379,11 @@ jobs:
...
         working-directory: ./rust/${{ matrix.folder }}
 
-  # Verify the workspace builds with --no-default-features to catch broken
-  # optional feature gates (e.g., dev-tools, crypto-*).
-  no_default_features_check:
+  compile_proto:
     strategy:
       fail-fast: false
       matrix:
+        # Only otap-dataflow has a structure check
```
drewrelmas: Suggest removing comment from copy/paste
```suggestion
```

rust/otap-dataflow/crates/pdata/src/proto/opentelemetry.proto.collector.logs.v1.rs
```diff
@@ -3,23 +3,23 @@
 #[derive(crate::otlp::Message)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct ExportLogsServiceRequest {
-    #[prost(message, repeated, tag="1")]
+    #[prost(message, repeated, tag = "1")]
```
drewrelmas, line 6: Can I assume that these changes to the `.rs` files are actually formatting changes missing in the main branch?

I assume it has been a while since someone ran and committed `compile-proto` and some tooling has updated resulting in these changes?

luckymachi: Yes, all changes related to `src/proto/` are formatting ones; they all tackle specifically changing `x=y` to `x = y`, none of the content has been altered

drewrelmas: Interestingly I tried the `compile-proto` on my machine from main branch and it didn't produce any diff. Perhaps the CI machine uses a different version/tool than I have... Did you get this diff when running locally?

luckymachi: Locally `compile-proto` did produce a diff for me (the exact ones in this commit) because my `/proto` dir got updated accordingly. In theory it should give the diff right now because `main` has the old `x=y` format; perhaps it could be good to add an extra `cargo fmt` within `compile-proto`?

lalitb: I'd be careful about adding `cargo fmt` here. `compile-proto` already turns off formatting on purpose - `xtask/src/genproto.rs` sets `cfg.format(false)` . The intent seems to be that the committed `.rs` files should match the raw generator output exactly.

lalitb: One small suggestion: it might be worth pinning the protoc version in `arduino/setup-protoc` so CI uses a stable compiler and a future upstream default change can't cause unrelated failures here.

Side note - since `Cargo.lock` isn't checked in (which is fine), `prost-build` and `tonic-prost-build` can also pick different versions on different machines. That's probably why `compile-proto` gives a diff for some people but not others. Pinning `protoc` won't fully fix that, but it's one less thing that can drift.

luckymachi: Sounds reasonable! Is there an specific protoc version that would be preferred for CI? https://github.com/protocolbuffers/protobuf/releases

lalitb: I'd suggest using 34.1, that's the current stable protobuf release - https://github.com/protocolbuffers/protobuf/releases/tag/v34.1

luckymachi: Awesome, should be good to go now. Also pinned the version in `rust-validation-tests.yml` and `rust-bench.yml` to keep it consistent 👍

.github/workflows/rust-bench.yml
```diff
@@ -49,6 +50,7 @@ jobs:
       - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b # v3.0.0
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+          version: "34.1"
```
drewrelmas: I don't think we need `protoc` in the `rust-bench` workflow for same reason, can you please remove?

.github/workflows/rust-validation-tests.yml
```diff
@@ -42,6 +42,7 @@ jobs:
       - uses: arduino/setup-protoc@c65c819552d16ad3c9b72d9dfd5ba5237b9c906b # v3.0.0
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
+          version: "34.1"
```
drewrelmas: Same here, potentially don't need it

drewrelmas (approved):
LGTM, provided `protoc` action is removed in additional places it is not needed

.github/workflows/rust-validation-tests.yml
luckymachi: Great points; just committed a new version removing `protoc` from both files 👍

lalitb (approved):

PR #2776, opened 2026-04-28
Validation framework fix equivalence check for metrics

### LINKED ISSUE #2786: assert_equivalent fails for metrics
### Pre-filing checklist

### Component(s)

### Bug Description

Validation tests that check that metrics are unchanged via the equivalence are failing.

This is due to the fact that when Buckets = None they are set to Some(Buckets { offset = 0, bucket_count: []})
which is not accounted for in the canonicalize_exp_histogram_data_point function.

The canonicalize_exp_histogram_data_point_function needs to be updated to handle this edge case

### Steps to Reproduce

Run validation test with metrics and otap receiver or exporter

```
    #[test]
    fn validation_metrics_otap_to_otlp() {
        Scenario::new()
            .pipeline(
                Pipeline::from_file("./validation_pipelines/otap-otlp.yaml")
                    .expect("failed to read in pipeline yaml"),
            )
            .add_generator(
                "traffic_gen",
                Generator::metrics()
                    .fixed_count(500)
                    .otap_grpc("receiver")
                    .core_range(1, 1)
                    .static_signals(),
            )
            .add_capture(
                "validate",
                Capture::default()
                    .otlp_grpc("exporter")
                    .validate(vec![ValidationInstructions::Equivalence])
                    .control_streams(["traffic_gen"])
                    .core_range(2, 2),
            )
            .run()
            .expect("metrics otap-to-otlp validation failed");
    }
```

### Expected Behavior

Should pass with no problems

### Actual Behavior

Validation test fail, equivalence check shows that metric output doesn't match metric input values

### OTel-Arrow Version

8a0e66ad7a59d2f5cb6477f64db12b3668938b9b

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

c1ly:
# Change Summary

This adds additional validation tests to check that all signals are preserved through otlp/otap receiver/exporters.

Adds canonicalize_buckets fn to set values from None to Some(Buckets {offset: 0, bucket_counts: vec![]}). Before the validation tests involving metrics and otap receiver/exporters would fail, but now they are succeeding

* Closes #2786

validation test

no


rust/otap-dataflow/crates/pdata/src/testing/equiv/metrics.rs
```diff
@@ -249,6 +249,22 @@ fn canonicalize_histogram_data_point(dp: &mut HistogramDataPoint) {
...
 }
 
+/// Normalize absent buckets to the default Buckets struct.
+/// The OTAP Arrow encoding cannot distinguish `None` from
+/// `Some(Buckets { offset: 0, bucket_counts: [] })` because
+/// `BucketsRecordBatchBuilder` stores both as a non-null struct
+/// with default values. Canonicalize `None` -> `Some(default)` so
+/// the equivalence check is not affected by this known
+/// encode/decode limitation.
+fn canonicalize_buckets(buckets: &mut Option<Buckets>) {
```
JakeDern, line 259: Are any histogram fields like `bucket_counts` or `explicit_bounds` also affected by this? Wondering if we have more mapping from `None -> vec![]` to do for those.

JakeDern (approved):
Thanks Chanly! Just left one small question about whether this affects histogram as well

rust/otap-dataflow/crates/pdata/src/testing/equiv/metrics.rs
c1ly: no those fields aren't affected

albertlockett (approved):

PR #2777, opened 2026-04-29
feat: Add OTAP metric-name filtering support

ThomsonTan:
# Change Summary

Adds metric-name filtering to the OTAP filter processor and documents how to configure it.

The filter exists for traces/logs signal, but remains TODO for metrics.

* Closes #NNN

Added test and passed locally.


rust/otap-dataflow/crates/pdata/src/otap/filter/metrics.rs
```diff
@@ -0,0 +1,267 @@
...
+    ///
+    /// Returns tuple of (filtered batch, metrics_consumed, metrics_filtered).
+    pub fn filter(
+        &self,
+        metrics_payload: OtapArrowRecords,
+        pool: &mut IdBitmapPool,
+    ) -> Result<(OtapArrowRecords, u64, u64)> {
+        let payload_type = metrics_payload.root_payload_type();
+        let metrics = metrics_payload
+            .root_record_batch()
```
lalitb: I believe this should handle an empty metrics payload before requiring the root record batch. Empty OTLP metrics requests encode to `Metrics::default()` with no `UnivariateMetrics` batch, so a pipeline with the default no-op metric filter would now fail with `RecordBatchNotFound` instead of passing the empty batch through.

rust/otap-dataflow/crates/pdata/src/otap/filter/metrics.rs
```diff
@@ -0,0 +1,267 @@
...
+        } else if let Some(include_config) = &self.include
+            && self.exclude.is_none()
+        {
+            include_config.create_filter(&metrics_payload, false)?
+        } else {
+            return Ok((metrics_payload, num_rows, 0));
+        };
+
+        let filtered_count = num_rows - metric_filter.true_count() as u64;
+        let filtered = filter_otap_batch(&metric_filter, &metrics_payload, pool)?;
```
lalitb, line 92: Can we add one direct OTAP test for this cascade? The current tests prove the round-tripped OTLP metrics look right, but they do not directly show that child batches like data points, exemplars, and attrs are pruned when a metric is removed. A small assertion on the child batch row counts would make this path much safer.

ThomsonTan: Good point, added a direct OTAP cascade test in `test_filter_prunes_child_batches` for this.

rust/otap-dataflow/crates/core-nodes/src/processors/filter_processor/README.md
```diff
@@ -77,7 +87,15 @@ config:
           value: false
 ```
 
-Currently we don't support metric filtering
+For a runnable metric-name filter pipeline, see
+[`configs/fake-metric-filter-debug-noop.yaml`](../../../../configs/fake-metric-filter-debug-noop.yaml).
```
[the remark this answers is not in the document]
ThomsonTan: Fixed.

rust/otap-dataflow/configs/fake-metric-filter-debug-noop.yaml
```diff
@@ -0,0 +1,50 @@
...
+              control:
+                node: 100
+                pipeline: 100
+              pdata: 128
+
+        nodes:
+          receiver:
+            type: receiver:traffic_generator
+            config:
+              data_source: static
```
lalitb: please check this once.  `traffic_generator` only accepts `data_source: synthetic` or `semantic_conventions`; `static` is not a valid enum value. Probably switch to `synthetic` so the runnable metric-filter example actually starts?

lalitb (approved):
Thanks.

rust/otap-dataflow/configs/fake-metric-filter-debug-noop.yaml
ThomsonTan: Thanks for the catch, fix it.

rust/otap-dataflow/crates/pdata/src/otap/filter/metrics.rs
```diff
@@ -0,0 +1,368 @@
...
+                    let value_scalar = StringArray::new_scalar(name);
+                    arrow::compute::kernels::cmp::eq(&names_column, &value_scalar)
+                        .expect("can compare string name column to string scalar")
+                }
+            };
+            filter = arrow::compute::or_kleene(&name_filter, &filter)
+                .map_err(|e| Error::ColumnLengthMismatch { source: e })?;
+        }
+
+        let filter = nulls_to_false(&filter);
```
jmacd, line 140: This caught my attention--I wonder if there is a potential optimization to avoid the call to `nulls_to_false` here (and other places in filter processor). Non-blocking, if you think the answer is yes, please file an issue. My guess is that for the `invert` branch below, the arrow::compute::not() call could be fused with nulls_to_false (which is a call to and_kleen(is_not_null, value)).

jmacd (approved):

Issue #2774, opened 2026-04-28
Pipeline perf test fails with 404 on shutdown endpoint


pritishnahar95:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Bug Description

# Pipeline perf test fails with 404 on shutdown endpoint

## Description

The pipeline perf test (`pipeline_perf_test_basic`) is consistently failing because the shutdown HTTP requests return `404 Not Found`.

```
HTTP request failed: 404 Client Error: Not Found for url:
http://localhost:8085/api/v1/pipeline-groups/shutdown?wait=true&timeout_secs=60
```

see https://github.com/open-telemetry/otel-arrow/actions/runs/25075767629/job/73467812180?pr=2755
https://github.com/open-telemetry/otel-arrow/actions/runs/25073419796/job/73468516227?pr=2772

## Root Cause

A recent renaming to the admin API routes from `/pipeline-groups/...` to `/groups/...` in `crates/admin/src/pipeline_group.rs`, seems to have broken the the pipeline perf test, since test YAML templates were not updated to match.

**Server-side route (current):** `/api/v1/groups/shutdown`
**Test template URL (stale):** `/api/v1/pipeline-groups/shutdown`

## Affected Files

- `tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker.yaml` (3 occurrences)
- `tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker-filtered.yaml` (3 occurrences)
- `tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker-otel.yaml` (1 occurrence)

## Fix

Replace all instances of `/api/v1/pipeline-groups/shutdown` with `/api/v1/groups/shutdown` in the above files.

### Steps to Reproduce

See test run in https://github.com/open-telemetry/otel-arrow/actions/runs/25075767629/job/73467812180?pr=2755

### Expected Behavior

Pipeline perf test to succeed without any errors

### Actual Behavior

The pipeline perf test (`pipeline_perf_test_basic`) is consistently failing because the shutdown HTTP requests return `404 Not Found`.

### OTel-Arrow Version

685040edd2eff9828044cccbbf74a4ebd93f4b2b

### Configuration

```yaml

```

### Log Output

```shell

```

PR #2779, opened 2026-04-29
Make pipeline perf test a required CI check

cijothomas:
## Summary

Make the pipeline performance test a required CI check so that PRs which break the perf test are caught before merge.

> **Dependency**: #2780 must be merged first (it fixes the currently broken perf test).

#2774 is an example of the kind of breakage this prevents — a route rename broke the perf test but the PR still merged because the perf test was not a required check.

### Changes

- **rust-ci.yml**: Add `pipeline_perf_test` job (runs on `ubuntu-latest`) and include it in `rust-required-status-check` aggregator
- **pipeline-perf-on-label.yaml**: Simplify to only run on dedicated Oracle bare-metal hardware when `pipelineperf` label is present — the basic validation path is removed since `rust-ci.yml` now covers it

### Motivation

The pipeline perf test has been broken by merged PRs several times because it was not a required check. This change ensures that if a PR breaks the perf test (e.g. build failures, config issues, test infrastructure breakage), it is caught before merge.


JakeDern (approved):
Yes please 🙂

Image

albertlockett (approved):

PR #2787, opened 2026-04-30
Add Linux user_events receiver

### LINKED ISSUE #2698: Proposal: Add Linux user_events Receiver for Telemetry from Applications Co-Located with the Collector

lalitb:
# Change Summary

Split out from the original combined PR #2717.

  Adds a contrib receiver for Linux user_events tracepoints.

  The receiver supports two decode formats:
  - `tracefs`: generic Linux tracepoint decoding based on the static schema exposed in tracefs format files.
  - `event_header`: decodes self-describing EventHeader payloads. EventHeader comes from the Microsoft LinuxTracepoints-Rust project, but is open and usable by any producer.

  The receiver uses `one_collect` for `perf`/`user_events` collection, and supports single or multiple subscriptions, batching, late registration, and basic receiver metrics. It is behind the `userevents-receiver` feature flag.

  Also adds a Linux CI smoke test that registers and reads both tracefs and EventHeader user_events tracepoints when kernel support is available.

* Closes #2698

Added Linux CI smoke test

Yes, a receiver.


rust/otap-dataflow/crates/contrib-nodes/README.md
```diff
@@ -14,6 +18,18 @@ This crate contains optional (feature-gated) contrib processors and exporters.
 Feature flags are grouped into aggregate categories and individual node flags.
 Aggregate flags enable all nodes in their category.
 
+### Receivers
+
+| Node | URN | Module |
+| ---- | --- | ------ |
+| userevents_receiver | `urn:otel:receiver:userevents` | `src/receivers/userevents_receiver/` |
```
utpilla: Let's use the official name with the underscore.

```suggestion
| Node | URN | Module |
| ---- | --- | ------ |
| userevents_receiver | `urn:otel:receiver:user_events` | `src/receivers/user_events_receiver/` |
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/mod.rs
```diff
@@ -0,0 +1,6 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+/// Linux userevents receiver.
+#[cfg(feature = "userevents-receiver")]
+pub mod userevents_receiver;
```
utpilla: This receiver is Linux-only. This will let you remove unnecessary stub implementations in `session.rs` and `one_collect_adpater.rs`.

```suggestion
#[cfg(all(feature = "userevents-receiver", target_os = "linux"))]
pub mod userevents_receiver;
```

rust/otap-dataflow/crates/contrib-nodes/Cargo.toml
```diff
@@ -14,29 +14,32 @@ workspace = true
...
 serde_json.workspace = true
 thiserror.workspace = true
 linkme.workspace = true
+tokio.workspace = true
+tracing.workspace = true
+base64 = { workspace = true, optional = true }
 chrono = { workspace = true, optional = true }
-geneva-uploader = { workspace = true, optional = true }
 humantime-serde = { workspace = true, optional = true }
+tracepoint_decode = { version = "0.5.0", optional = true }
```
utpilla: Could we feature gate the EventHeader related code and have only `tracefs` support by default?

rust/otap-dataflow/crates/contrib-nodes/Cargo.toml
utpilla, line 1: The three new crate dependencies (`tracepoint_decode`, `eventheader_dynamic`, `tracepoint`) specify versions inline instead of using `workspace = true` breaking the convention used in the repo.

Also, since these are Linux-only crates:

- `tracepoint_decode` should go under `[target.'cfg(target_os = "linux")'.dependencies]` (like `one_collect` already does)
- `eventheader_dynamic` and `tracepoint` should go under `[target.'cfg(target_os = "linux")'.dev-dependencies]`

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/arrow_records_encoder.rs
```diff
@@ -0,0 +1,256 @@
...
+        self.logs.scope.append_id_n(0, log_record_count);
+        self.logs.scope.append_name_n(None, log_record_count);
+        self.logs.scope.append_version_n(None, log_record_count);
+        self.logs
+            .scope
+            .append_dropped_attributes_count_n(0, log_record_count);
+
+        self.logs.append_schema_url_n(None, log_record_count);
+        self.logs
+            .append_dropped_attributes_count_n(0, log_record_count);
```
jmacd, line 117: These empty-value columns are being filled in because the builder requires it. Nothing actionable, but I think we should find a lighter-weight way to build that would leave these columns optional.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md
```diff
@@ -0,0 +1,550 @@
...
+`linux.userevents.payload_base64` attribute.
+
+Only scalar EventHeader values that map to the receiver's current attribute
+types are surfaced: strings, signed integers, booleans, and floating point
+values. EventHeader arrays, binary blobs, and other non-scalar encodings are
+not emitted as attributes yet.
+
+TODO: Add support for non-scalar EventHeader values once there is a stable
+mapping to OTLP log attributes.
+
```
jmacd, line 362: Suggest adding links to several ways to produce these events with OpenTelemetry or non-OTel SDKs.

I'm aware of is https://github.com/microsoft/LinuxTracepoints-Rust and(your own) https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-user-events-logs for OTel-Rust.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/arrow_records_encoder.rs
```diff
@@ -0,0 +1,256 @@
...
+            .append_trace_id(record.trace_id.as_ref() as Option<&TraceId>);
+        _ = self
+            .logs
+            .append_span_id(record.span_id.as_ref() as Option<&SpanId>);
+
+        // Receiver-internal transport/diagnostic fields are intentionally not
+        // represented on decoded records or emitted as downstream attributes:
+        // the Ingestion backend treats OTLP log attributes as backend columns,
+        // so surfacing per-record diagnostics there would pollute the
+        // application schema.
```
jmacd: This comment probably belongs inside from_tracefs() in decoder.rs.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
```diff
@@ -0,0 +1,749 @@
...
+    /// Optional severity text.
+    pub severity_text: Option<Cow<'static, str>>,
+    /// Optional OTLP log flags.
+    pub flags: Option<u32>,
+    /// Optional W3C trace id (16 bytes).
+    pub trace_id: Option<[u8; 16]>,
+    /// Optional W3C span id (8 bytes).
+    pub span_id: Option<[u8; 8]>,
+    /// Additional structured attributes, preserving the source typed value.
+    pub attributes: Vec<(Cow<'static, str>, DecodedAttrValue)>,
```
jmacd, line 147: For this field and for body and event_name above, these look like intermediate copies of the data, they stand out because of the memory allocations required. Not blocking this, but I can imagine filing an issue that will reorganize this receiver to call directly in to the Arrow builders. For this field, we would write directly into the LogAttrs builder, for example.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
```diff
@@ -0,0 +1,749 @@
...
+//! +----------------------+----------+----------+----------+
+//! ```
+//!
+//! EventHeader is decoded structurally and vendor-neutrally here. Structs are
+//! flattened into dot-separated attribute names, e.g. `PartB.body` or
+//! `Request.path`, but this module does not attach semantic meaning to any
+//! particular struct or field name. Schema-specific interpretation, such as
+//! Microsoft Common Schema `PartA`/`PartB`/`PartC` promotion, belongs in a
+//! processor. If EventHeader decoding fails, only the user payload region is
+//! preserved as `linux.userevents.payload_base64`.
```
jmacd: Please add a reference URL (best you can) to the user_events and tracefs payload structures decoded in this file.

jmacd (commented):
Partial review. This is super!

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
```diff
@@ -0,0 +1,550 @@
...
+
+### Subscriptions
+
+```yaml
+nodes:
+  ingest:
+    type: receiver:userevents
+    config:
+      subscriptions:
+        - tracepoint: "user_events:myprovider_L2K1"
```
hestolz: considering this is the `user_events` receiver, I would not expect support for tracepoints from any subsystem other than `user_events:`; should that prefix be dropped?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/one_collect_adapter.rs
```diff
@@ -0,0 +1,502 @@
...
+
+#[cfg(target_os = "linux")]
+#[derive(Debug, Clone, Copy)]
+struct PerfTimeAnchor {
+    unix_nano: u64,
+    perf_nano: u64,
+}
+
+#[cfg(target_os = "linux")]
+impl PerfTimeAnchor {
```
gouslu, line 55: AI review flagged that qpc_time() appears to be a monotonic/hardware clock while SystemTime::now() is CLOCK_REALTIME (wall clock), and that the two can drift apart over time, gradually from NTP slewing, or instantly from step adjustments like date -s or VM resume. Since the anchor is captured once at `OneCollectUserEventsSession::open()` and reused for the session lifetime, long-running pipelines could see emitted timestamps drift from real wall clock. Is this a concern worth re-anchoring periodically?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md
```diff
@@ -0,0 +1,550 @@
...
+      subscriptions:
+        - tracepoint: "user_events:myprovider_L2K1"
+          format:
+            type: tracefs
+      session:
+        per_cpu_buffer_size: 1048576  # bytes
+        late_registration:
+          enabled: true
+          poll_interval_ms: 100
+      drain:
```
hestolz, line 224: I don't see the equivalent implemented in LinuxTracepoints-Rust, and it's only viable for EventHeader tracepoints, but the tracepoint-control c++ has a [`PreregisterTracepointDefinition`](https://github.com/microsoft/LinuxTracepoints/blob/9ff8fe86f9a33ac606605ae2510283ef062ddf6a/libtracepoint-control-cpp/include/tracepoint/TracepointCache.h#L163) which removes the need to poll for late registrations.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/mod.rs
lalitb: Good point. Done.

rust/otap-dataflow/crates/contrib-nodes/Cargo.toml
lalitb: Done. thanks.

rust/otap-dataflow/crates/contrib-nodes/Cargo.toml
lalitb: Done. Moved all three to workspace dependencies and kept them Linux-only:

  - tracepoint_decode is now under Linux target dependencies.
  - eventheader_dynamic and tracepoint are now under Linux target dev-dependencies.
  - The EventHeader decode path is behind the new userevents-eventheader feature, while userevents-receiver keeps tracefs support by default.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/arrow_records_encoder.rs
lalitb: Agreed. I left the current builder calls in place for now since the record builders need row-aligned arrays before finish(), but added a TODO here to replace these placeholder appends with lighter-weight builder support for absent optional resource/scope/log columns.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md
lalitb: Good point. Added a Producers section with links to `LinuxTracepoints-Rust` and the OTel Rust user-events logs exporter. I also added receiver-local producer examples for manual validation: one `tracefs` producer and one `EventHeader` producer, each with a matching debug-pipeline config.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/arrow_records_encoder.rs
lalitb: Moved it into `from_tracefs()` where the `common_*` fields are actually skipped. That keeps the rationale next to the behavior it explains.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
lalitb: Agreed. I added a TODO on the decoded record shape noting that these owned intermediate values should eventually be replaced by writing directly into the Arrow builders, especially LogAttrs, once we reorganize this path.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
lalitb: Added references near the decoder module docs for the Linux `user_events` registration/write ABI and the `tracefs` event format structure:
  - https://docs.kernel.org/trace/user_events.html
  - https://docs.kernel.org/trace/events.html#event-formats

rust/otap-dataflow/crates/contrib-nodes/src/receivers/userevents_receiver/README.md
lalitb: Good point. I changed the examples to use bare event names, e.g. myprovider_L2K1, and the receiver now normalizes them internally to `user_events:<event>`. The explicit `user_events:<event>` form is still accepted, but other tracepoint groups are rejected. 
Thanks for the reviewing this.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/one_collect_adapter.rs
lalitb: Yes, the concern makes sense. I don’t think we should add periodic re-anchoring in this PR, though, because refreshing the anchor after a wall-clock step can also make emitted timestamps jump forward or backward. I added a TODO to prefer a `one_collect`-owned `sample-time` to realtime conversion API, and to revisit re-anchoring only once we have an explicit policy for wall-clock steps.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md
lalitb: Thanks, added a TODO in the late-registration section. I kept the current behavior as polling because that is what the Rust path supports today, but noted that `EventHeader` producers may be able to avoid polling once the Rust collection path exposes support equivalent to the C++ `PreregisterTracepointDefinition` API.

rust/otap-dataflow/crates/contrib-nodes/Cargo.toml
```diff
@@ -45,14 +47,33 @@ azure_core = { workspace = true, optional = true, features = ["reqwest"] }
...
+# normal dependency graph and no longer needs to be pulled directly here. The
+# pinned commit is from 2026-04-10.
+one_collect = { git = "https://github.com/microsoft/one-collect.git", rev = "2d33e674348168f4a19bd3ab40d92f1c556407ec", features = ["scripting"], optional = true }
+tracepoint_decode = { workspace = true, optional = true }
+
 [features]
+contrib-receivers = [
+    "userevents-receiver",
+]
+userevents-receiver = [
```
utpilla:
```suggestion
user_events-receiver = [
```

rust/otap-dataflow/crates/contrib-nodes/Cargo.toml
```diff
@@ -45,14 +47,33 @@ azure_core = { workspace = true, optional = true, features = ["reqwest"] }
...
+    "dep:base64",
+    "dep:chrono",
+    "dep:humantime-serde",
+    "dep:one_collect",
+]
+userevents-eventheader = [
```
utpilla:
```suggestion
user_events-eventheader = [
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md
```diff
@@ -0,0 +1,631 @@
...
+**URN:** `urn:otel:receiver:user_events`
+
+This receiver ingests Linux
+[`user_events`](https://docs.kernel.org/trace/user_events.html) tracepoints
+through `perf_event_open` and converts them into OTAP logs for downstream
+processing.
+
+> **Note:** This receiver is vendor-neutral. It performs Linux `user_events`
+> collection and structural decoding only; schema-specific mappings such as
+> Microsoft Common Schema should be modeled outside the receiver.
```
utpilla: No need to mention Microsoft Common Schema explicitly here.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md
```diff
@@ -0,0 +1,631 @@
...
+- The receiver is changed to read perf rings for CPUs other than its assigned
+  pipeline core. Today `session.rs` intentionally constructs
+  `cpu_ids = vec![pipeline.core_id()]`.
+
+## Platform Support
+
+This receiver is **Linux-only**.
+
+It does **not** work on macOS or Windows because `user_events` is a Linux
+kernel tracing feature. Windows support would require a separate ETW receiver.
```
utpilla: No need to talk about ETW Receiver here since this is README is not about general kernel tracing support. It's specific to user_events.

```suggestion
kernel tracing feature.
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/mod.rs
```diff
@@ -0,0 +1,1483 @@
...
+#[derive(Debug, Deserialize, Clone, Default)]
+#[serde(deny_unknown_fields)]
+struct LateRegistrationConfig {
+    /// Enables periodic retries for subscriptions whose tracepoints do not
+    /// exist yet.
+    #[serde(default)]
+    enabled: bool,
+    /// Delay, in milliseconds, between late-registration retry attempts.
+    #[serde(default = "default_late_registration_poll_ms")]
+    poll_interval_ms: u64,
```
utpilla: Could we use an `Option<Duration>` here instead of two fields: a bool and a time interval? That would make invalid states unrepresentable: `enabled: false` plus a custom `poll_interval_ms` would be allowed but it's invalid.

lalitb:
Thanks @utpilla - these are valid comments, and addressed.

Commit: 66fab1bdf Address user_events receiver review feedback

  Changes made:

  - Renamed features to user_events-receiver / user_events-eventheader.
  - Removed README mentions of Microsoft Common Schema and ETW.
  - Replaced late_registration.enabled + poll_interval_ms with session.late_registration_poll_interval: Option<Duration>.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md
```diff
@@ -0,0 +1,628 @@
...
+## Runtime Behavior
+
+### Backpressure
+
+`user_events` perf rings cannot be backpressured like a socket. For that
+reason, the receiver does not block the perf drain loop when downstream is
+full; it drops already-drained batches and reports them as
+`dropped_downstream_full`. Separately, if the kernel/perf ring overruns before
+the receiver drains it, that loss is reported by the perf path and counted as
+`lost_perf_samples`.
```
utpilla: I think the receiver should still block here. The fact that `user_events` perf rings can't be back pressured isn't a reason to skip backpressure entirely. It's a reason to apply it at the channel boundary instead.

The receiver shares a core with the downstream nodes under the engine's thread-per-core, current-thread runtime. When downstream is saturated, every cycle the receiver spends on parse, pop, and decode of records that immediately get dropped is CPU taken away from the very downstream task that is the bottleneck. With `send().await`, the receiver would park and let the downstream task run, which feels like the natural backpressure path on a single-threaded runtime. Try-send plus a drop counter does the opposite: it keeps the receiver doing more work in the situation where it should be doing less.

There is also an observability point. The current design keeps the kernel ring drained even under sustained saturation, so kernel-side loss counters like `lost_perf_samples` and `PERF_RECORD_LOST_SAMPLES` could stay at zero. Producers and standard Linux perf tooling might never get to see that the write rate is too high or that `per_cpu_buffer_size` is undersized. The drops only show up in our internal counter, which is harder to discover. With `send().await`, saturation surfaces at the kernel boundary that already has standard observability, which feels like the more honest signal.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
```diff
@@ -0,0 +1,788 @@
...
+        }
+    }
+}
+
+fn until_nul(bytes: &[u8]) -> &[u8] {
+    bytes
+        .iter()
+        .position(|byte| *byte == 0)
+        .map_or(bytes, |end| &bytes[..end])
+}
```
utpilla, line 319: Could we use tracepoint_decode for the `tracefs` path too? You're already using it for `eventheader` format. It already provides format parsing (`PerfEventFormat::parse`) and field extraction (`PerfFieldFormat::get_field_value_with_data_and_reader`), including correct handling of `data_loc` / `rel_loc` and endianness.

Reusing it lets us lean on the upstream review and test coverage instead of maintaining a parallel implementation here.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/metrics.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Metrics for the Linux userevents receiver.
+
+use otap_df_telemetry::instrument::Counter;
+use otap_df_telemetry_macros::metric_set;
+
+/// Internal telemetry for the Linux userevents receiver.
+#[metric_set(name = "userevents.receiver.metrics")]
```
utpilla:
```suggestion
#[metric_set(name = "user_events.receiver.metrics")]
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
```diff
@@ -0,0 +1,788 @@
...
+        if let Some(thread_id) = value.thread_id {
+            attributes.push((
+                Cow::Borrowed("linux.userevents.thread.id"),
+                DecodedAttrValue::Int(i64::from(thread_id)),
+            ));
+        }
+
+        Self {
+            time_unix_nano: i64::try_from(value.timestamp_unix_nano).unwrap_or(i64::MAX),
+            body: String::new(),
```
utpilla: `body` is always set to `String::new()`, so every record gets an empty body string that then gets encoded into the Arrow body column. Neither `tracefs` nor `EventHeader` has a natural source for a body value, since per the README, schema-specific promotion belongs in a processor downstream of the receiver.

Could we change `body: String` to `body: Option<String>` (matching the existing flags: `Option<u32>` pattern) and have the encoder call `body.append_null()` for `None`? `LogsBodyBuilder` already supports this, and its `finish()` omits the body column from the resulting `RecordBatch` when all rows are null, same as it does for flags today.

That eliminates the dead allocation per record while still leaving room for a downstream processor to populate body later.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/session.rs
```diff
@@ -0,0 +1,238 @@
...
+
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct SessionDrainStats {
+    pub received_samples: u64,
+    pub dropped_no_subscription: u64,
+    pub lost_samples: u64,
+    pub dropped_pending_overflow: u64,
+}
+
+mod imp {
```
utpilla: The `mod imp { ... } pub(super) use imp::{...}` wrapping looks like leftover scaffolding from when the Linux gating lived at this layer. Now that the entire `user_events_receiver` module is gated at the parent (`#[cfg(all(feature = "userevents-receiver", target_os = "linux"))]` in `receivers/mod.rs`), the inner module isn't doing anything anymore.

Could we inline the contents at file scope and drop the `mod imp` wrapper?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/session.rs
```diff
@@ -0,0 +1,238 @@
...
+                dropped_no_subscription: 0,
+                lost_samples: drained.lost_samples,
+                dropped_pending_overflow: drained.dropped_pending_overflow,
+            };
+            for event in drained.events {
+                let EventSource::UserEvents(source) = event.source;
+                if source.subscription_index >= self.inner.subscription_count() {
+                    stats.dropped_no_subscription += 1;
+                    continue;
+                }
```
utpilla: This branch looks structurally unreachable. Indices are assigned in `OneCollectUserEventsSession::open` via `subscriptions.iter().enumerate()` and captured in each event callback at session-open time, and subscriptions can't be added or removed mid-session. The README also says `dropped_no_subscription` "should normally stay zero".

Could we either reduce this to a `debug_assert!` and drop the metric, or add a comment noting it's defense-in-depth against a future adapter regression? Right now, the metric and counter exist but can't really be exercised.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md
lalitb: Thanks, this makes sense. I had originally treated the perf ring as the place we should keep draining because it cannot be backpressured like a socket, but your point about the channel boundary is right for this runtime.

I changed the receiver to await the downstream send instead of try_send + drop. That lets the receiver yield on saturation, and it deliberately makes the kernel perf ring the overflow point so standard perf/user_events loss observability still works. I also removed the downstream-full drop metric/config and added `downstream_send_blocked_ns` so we still have a signal when the receiver is blocked on downstream capacity.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
lalitb: Agreed. Reusing `tracepoint_decode` for` tracefs` is the better long-term direction, especially for `data_loc`/`rel_loc` and endianness.

I did not fold that rewrite into this PR because it is a larger decoder change with its own test surface. I added an inline note documenting the current same-host/live-decoding assumption and left the
`tracepoint_decode` migration as a follow-up with explicit `rel_loc`/`data_loc` and cross-endian coverage.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
lalitb: Agreed. I changed body to `Option<String>` and now append null when there is no natural body value.

Small clarification: `String::new()` itself was not a heap allocation, but it did make every record encode a non-null empty body value. With None, all-null batches can omit the body column, matching the flags pattern.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/session.rs
lalitb: Agreed, that wrapper was leftover scaffolding. I inlined the contents at file scope.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/session.rs
lalitb: Agreed that this should be structurally unreachable with the current adapter. I kept the guard as defense-in depth for future adapter changes and added a comment explaining that; the metric should remain zero unless that invariant is broken.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
utpilla: The migration is mostly a **delete**, not an expansion: drop `tracefs_field_bytes`, `decode_tracefs_field`, `TracefsField`, `TracefsFieldLocation`, and the `LocationType` mirror in the adapter, then replace the type-name `match` with a `PerfItemValue` `match` similar to the existing `item_as_any_scalar_value` we already use for EventHeader. Net is a code shrink, and the new code reuses an upstream-tested API rather than adding to the maintenance surface.

There are already two latent correctness issues in the current code that the migration would fix as a side effect:

1. **`StaticString` is not bounded by `field.size`.** `tracefs_field_bytes` does `event_data.get(field.offset..).map(until_nul)` for `StaticString` — it ignores `field.size` and reads forward until the first NUL or end-of-buffer. A malformed payload with no NUL in a `char[N]` field overruns into adjacent fields. Compare to `StaticUtf16String` two lines down, which correctly bounds by `field.size`.

2. **`loc_bytes.len() != 4` is structurally unreachable** in the `DynRelative` / `DynAbsolute` branch. The slice's length is always `field.size`, so the check fires only when `field.size != 4`, meaning it silently relies on `field.size == 4` for dynamic locations rather than validating it.

This PR already carries 18 TODOs across the user_events files and I'd be a bit hesitant to add another one specifically for a code path with two known-incorrect behaviors that an upstream library already handles. Since the handwritten decoder was introduced by this PR and hasn't shipped yet, removing it now is also cheaper than after release.

Could we do the migration in this PR instead of as a follow-up?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/one_collect_adapter.rs
```diff
@@ -0,0 +1,458 @@
...
+    pub thread_id: Option<u32>,
+    pub event_data: Vec<u8>,
+    pub user_data_offset: usize,
+    pub fields: Arc<[TracefsField]>,
+    pub source: EventSource,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[non_exhaustive]
+pub(crate) enum EventSource {
```
utpilla: The `EventSource` / `UserEventsSource` wrapping carries one `usize` (`subscription_index`) through two layers of types and a `non_exhaustive` enum, with no second variant in sight. Could we inline `subscription_index: usize` directly on `CollectedEvent` and drop both the enum and the wrapper struct?

Same shape concern for `UserEventsSubscription` / `UserEventsSessionConfig` in this file. They mirror `SubscriptionConfig` / `SessionConfig` from `mod.rs` field-for-field with no transformation, so the adapter's `open` could take the user-facing types directly. Net is roughly 30 lines deleted across this file and session.rs, no behavior change.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/mod.rs
```diff
@@ -0,0 +1,6 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+/// Linux user_events receiver.
```
utpilla, line 4: The feature flag and metric meter were renamed to `user_events-*` and `user_events.receiver.metrics`, but a lot of internal references still use the unspaced `userevents` form. Quick search finds 118 occurrences across the receiver. They fall into a few groups:

1. **Type names**: `UsereventsReceiver`, `DecodedUsereventsRecord`, `RawUsereventsRecord`, `UsereventsSession`, `UsereventsReceiverConfig`, `UsereventsReceiverMetrics`. Should be `UserEventsReceiver`, `DecodedUserEventsRecord`, etc., to match the kernel feature name.
2. **Module-level doc comments** in `mod.rs`, `decoder.rs`, `session.rs` (`//! Linux userevents ...`).
3. **User-facing config error messages** like `"userevents receiver requires at least one subscription"` (~11 in `mod.rs`).
4. **Log messages** in `otel_info!` / `otel_warn!` calls (`"Linux userevents receiver started"`, `"failed to drain Linux userevents perf ring"`, etc.).
5. **Emitted log attribute names**: `linux.userevents.process.pid`, `linux.userevents.thread.id`, `linux.userevents.payload_base64` in `decoder.rs`. These are part of the OTAP output contract, so renaming them is also a wire-format change worth aligning with the kernel name and the `user_events.receiver.metrics` meter.
6. **README references** to the same attribute names.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
lalitb: I agree `tracepoint_decode` is the right long-term direction, but I think doing that migration in this PR would add more risk than it removes.

`tracepoint_decode 0.5.0` does not appear to be a clean drop-in replacement for this `tracefs` path yet: tracefs `bool` / `float` / `double` still need extra handling, `one_collect` does not expose the raw format text so we would need a second format-file read, and `__rel_loc` needs careful validation/workaround.

Given that, I think the safer fix for this PR is to address the two concrete issues directly: bound `StaticString` by `field.size`, and explicitly require dynamic location fields to be size 4. Then we can do the full `tracepoint_decode` migration separately once the upstream tracefs gaps are resolved.

EDIT - as of now, fixed these two concrete issues in the latest commit.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/one_collect_adapter.rs
lalitb: I inlined `subscription_index` on `CollectedEvent` and removed the `EventSource` / `UserEventsSource` wrapper since that was only carrying one value with no other source variants.

For `UserEventsSubscription` / `UserEventsSessionConfig`, I would prefer to leave those adapter-facing types for now. They are not exact mirrors of the user config: `SubscriptionConfig` also carries the decode `format`, `SessionConfig` has receiver-level settings like `late_registration_poll_interval` / `wakeup_watermark`, and the adapter config derives `cpu_ids` from the pinned pipeline CPU. Keeping those small types keeps the one_collect boundary narrower and avoids passing user-facing config fields into the adapter that it should not consume.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/mod.rs
lalitb: Good point. Fixed these now.

.github/workflows/rust-ci.yml
```diff
@@ -196,6 +196,24 @@ jobs:
           retention-days: 30
           if-no-files-found: ignore
 
+  userevents_linux_smoke:
```
utpilla:
```suggestion
  user_events_linux_smoke:
```

.github/workflows/rust-ci.yml
utpilla, line 851: Add the check for user_events smoke test as well? Similar to what you did for host-metrics-semconv?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/mod.rs
```diff
@@ -0,0 +1,1528 @@
...
+
+        true
+    }
+
+    #[tokio::test(flavor = "current_thread")]
+    async fn user_events_linux_e2e_smoke_when_available() {
+        if std::env::var_os(RUN_USEREVENTS_E2E_ENV).is_none() {
+            return;
+        }
+
```
utpilla, line 868: The smoke test opens the perf session on CPU 0 but the `tokio::test(flavor = "current_thread")` runtime isn't pinned. If the OS schedules the test thread off CPU 0 during `tracepoint_state.write(...)`, the sample lands in a different ring than the receiver is watching, and the 2-second `drain_ready` timeout panics. Works today on idle GitHub runners, but might lead to flakiness? Could we pin the smoke test to CPU 0?

```rust
if !core_affinity::set_for_current(core_affinity::CoreId { id: 0 }) {
    eprintln!("user_events smoke skipped: could not pin to CPU 0");
    return;
}
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
```diff
@@ -0,0 +1,837 @@
...
+        )];
+
+        let decoded = decode_tracefs(fields, vec![0; 8]);
+
+        assert!(decoded.attributes.is_empty());
+    }
+
+    #[cfg(feature = "user_events-eventheader")]
+    #[test]
+    fn eventheader_decodes_microsoft_common_schema_shape_as_flattened_attributes() {
```
utpilla, line 735: `item_as_any_scalar_value` has 12 decoding paths but the test only covers integers and strings. Specifically untested:

`ENC_VALUE32`+ `FMT_FLOAT` → `f32::from_bits`
`ENC_VALUE64` + `FMT_FLOAT` → `f64::from_bits`
`ENC_VALUE8` + `FMT_BOOLEAN`
`ENC_VALUE32` + `FMT_BOOLEAN`

It would be nice to add test coverage for them as well. The float paths have a risk of a wrong byte-reinterpretation (e.g., `to_i32` instead of `to_u32` before `from_bits`) would silently corrupt values.

utpilla (approved):
LGTM overall. I left a comment about including the smoke test in `rust-required-status-check`.

Thanks for working on this @lalitb! Great work!

lalitb:
> LGTM overall. I left a comment about including the smoke test in `rust-required-status-check`.
> 
> Thanks for working on this @lalitb! Great work!

Thanks @utpilla for the thorough review! Addressed all remaining points in https://github.com/open-telemetry/otel-arrow/pull/2787/commits/e01f7676362e0a8cf28510d3f321a99e922cbf87

rust/otap-dataflow/configs/user-events-tracefs-debug.yaml
```diff
@@ -0,0 +1,57 @@
...
+          debug:
+            type: processor:debug
+            config:
+              verbosity: detailed
+              mode: signal
+              signals:
+                - logs
+              output: /tmp/user_events_tracefs_debug.log
+
+          noop:
```
cijothomas, line 53: do we need this?

lalitb: Yes, it's required as `processor:debug` is a processor, not a terminal node.

rust/otap-dataflow/configs/user-events-tracefs-debug.yaml
```diff
@@ -0,0 +1,57 @@
...
+engine: {}
+groups:
+  default:
+    pipelines:
+      main:
+        # Linux user_events -> debug processor pipeline for manual tracefs E2E
+        # validation.
+        #
+        # Build:    cargo build --features user_events-receiver
+        # Run:      ./target/debug/df_engine --config configs/user-events-tracefs-debug.yaml --num-cores 1
```
cijothomas, line 11: trying to recollect the conclusion of earlier discussion - this runs on 1 core only - does it accept events from all cpu cores? or example has to run on this core itself for this to work?

lalitb: Yes - with `--num-cores 1` the receiver only drains CPU 0's ring, so the producer has to run on CPU 0 too (hence the `taskset -c 0` in the example).

cijothomas: I am not seeing anything specific about producer that makes it pick the same processor core as the receiver.. ? 
Since df-engine supports accepting the exact core it runs (and not just --num-cores 1), should we use that for producer+consumer and do a comment here so users would know why it is important?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/metrics.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Metrics for the Linux user_events receiver.
+
+use otap_df_telemetry::instrument::Counter;
+use otap_df_telemetry_macros::metric_set;
+
+/// Internal telemetry for the Linux user_events receiver.
+#[metric_set(name = "user_events.receiver.metrics")]
```
cijothomas: we recently did a bunch of renaming. The ".metrics" suffix should be removed. And the name should be receiver.user_events. (The second part is still waiting for directional approval).

If you are waiting for other reviews, it may be better to make them right now. Alternate of follow up PRs also totally fine.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md
```diff
@@ -0,0 +1,629 @@
...
+```text
+<provider>_L<level>K<keyword>
+```
+
+The explicit `user_events:<event>` form is also accepted and normalized
+internally. Other tracefs groups are rejected.
+
+EventHeader-style names such as `<provider>_L<level>K<keyword>` are accepted,
+but the generic receiver does not interpret the level or keyword as OTLP
+severity.
```
cijothomas, line 333: if a user want to accept events on particular severity only, then they need to manually name the tracepoints using this convention and then configure listener to pick those names only - just checking if my understanding is correct. 
(I guess this will be diff in ETW where severity can be specified directly in the config)

cijothomas: or do we plan to recognize the severity from the EH style names as a feature?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
```diff
@@ -0,0 +1,917 @@
...
+            let loc_end = field.offset.checked_add(field.size)?;
+            let loc_bytes = event_data.get(field.offset..loc_end)?;
+            let loc = u32::from_ne_bytes(loc_bytes.try_into().ok()?);
+            let len = (loc >> 16) as usize;
+            let mut start = (loc & 0xFFFF) as usize;
+            if field.location == TracefsFieldLocation::DynRelative {
+                start = start.checked_add(loc_end)?;
+            }
+            let end = start.checked_add(len)?;
+            event_data.get(start..end).map(until_nul)
```
ThomsonTan: Should `until_nul` be applied here as this current arm could not be NUL terminated string?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/mod.rs
```diff
@@ -0,0 +1,1540 @@
...
+    builder: &mut ArrowRecordsBuilder,
+    drained_records: &mut Vec<RawUserEventsRecord>,
+    drain_stats: SessionDrainStats,
+    batch_cfg: &BatchConfig,
+) -> Result<(), Error> {
+    let received_samples = drain_stats.received_samples;
+    let mut dropped_memory_pressure: u64 = 0;
+    let mut dropped_no_subscription = drain_stats.dropped_no_subscription;
+
+    for raw in drained_records.drain(..) {
```
ThomsonTan: When mid-batch `flush_batch(...).await?` errors (e.g. downstream send fails during shutdown), the `Drain` iterator and in-flight batch are dropped silently with no `dropped_*` attribution - should we capture `in_flight = builder.len()` and add `in_flight + iter.count()` to a new `dropped_send_error` counter before propagating, and apply the same at the trailing-flush sites in `start()` / `DrainIngress`?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md
lalitb: Yes, that is correct. Today the receiver only subscribes by tracepoint name and does not infer severity from EventHeader-style names. We can add severity recognition as a follow-up if needed.

rust/otap-dataflow/configs/user-events-tracefs-debug.yaml
lalitb: Good point. This was in README.md, but now I updated the debug configs to make this explicit.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/metrics.rs
lalitb: Done. I changed the metric set name to `receiver.user_events` and updated the README metrics section to match.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/decoder.rs
lalitb: Fixed. Dynamic loc fields should not be NUL-trimmed unconditionally. I changed the path so we preserve the declared byte range first, and only apply string trimming when the field is actually decoded as a string.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/mod.rs
lalitb: Fixed. This was a real accounting gap. I added a `dropped_send_error` counter and now count both the in-flight batch on send failure and the remaining drained records in the mid-batch flush path before propagating the error. I also added focused tests for the send-failure cases.

lalitb:
@copilot resolve the merge conflicts in this pull request

jmacd (approved):
This is great to see. I would request that for the many TODOs in this code that we open issues to explain -- I'm curious for more information about them all.

PR #2788, opened 2026-04-30
Add Microsoft Common Schema log processor

lalitb:
# Change Summary

 Split out from the original combined PR #2717.

  Adds a contrib processor that promotes decoded Microsoft Common Schema log attributes into typed OTLP log fields.

  This processor is independent of the user_events receiver. It expects Common Schema fields to already be decoded into log attributes, for example from EventHeader payloads.

 Also adds unit tests for supported promotion behavior and pass-through cases.

* Closes towards #2698 

Unit tests

Yes, adds a processor.


rust/otap-dataflow/crates/contrib-nodes/src/processors/microsoft_common_schema_processor/README.md
```diff
@@ -0,0 +1,79 @@
+# Microsoft Common Schema Processor
```
gouslu, line 1: are there any public documentation links to what is "Microsoft Common Schema"?

cijothomas: nope. But reference implementations are all public. (OTel Rust, C++, .NET etc)

gouslu: it would help to document what "common schema" is.

rust/otap-dataflow/crates/contrib-nodes/src/processors/microsoft_common_schema_processor/mod.rs
```diff
@@ -0,0 +1,1027 @@
...
+        let Some(values) = keys.values().as_any().downcast_ref::<StringArray>() else {
+            return false;
+        };
+        return (0..keys.len()).any(|index| {
+            if !keys.is_valid(index) {
+                return false;
+            }
+            let value_index = keys.keys().value(index) as usize;
+            values.is_valid(value_index) && values.value(value_index) == target
+        });
```
albertlockett: This will compare the target with the dictionary value multiple times if the value is present in the dictionary multiple times, wont it?

Would it be faster in cases where the array is long, but low cardinality, to scan the dictionary values, and then scan the keys array for indices pointed to the dict value?

```
for (i, value) in values.iter().flatten().enumerate() {
  if value == target {
    let key = i as u8;
    let has_key = key in keys.keys().iter().flatten().find(|k| k == k);
    if has_key {
      return true
    }
  }
}
```

Two other things to consider - it would slightly unusual (all though not impossible) to get a dictionary column that has a dict value with no key pointing at it. We might save time here by only checking if the dict values array contains the target, and then assume there's a dict key pointing to the dict value. On the off-chance we assume wrongly, we do some extra work.

rust/otap-dataflow/crates/contrib-nodes/src/processors/microsoft_common_schema_processor/mod.rs
```diff
@@ -0,0 +1,1027 @@
...
+//! `Logs` and `LogAttrs` record batches. A native implementation needs a
+//! reusable cross-batch transform that can read attributes by `parent_id`, set
+//! root log columns such as body, event name, severity, trace/span IDs, flags,
+//! and timestamp, then rebuild `LogAttrs` with promoted fields removed and
+//! remaining fields preserved. Existing attribute/transform processors cannot
+//! express that today: they operate on attribute batches with rename/delete and
+//! literal insert/upsert operations, but they do not move per-log attribute
+//! values into root log columns or parse Common Schema values into typed log
+//! fields. Until that Arrow-native promotion path exists, Arrow-native
+//! pipelines pay an Arrow-to-OTLP conversion here.
```
albertlockett: We could eventually do this using the query-engine crate. Some of the transformations we're doing are even supported today!

For example we could use an OPL program like following to do some of this:
```kql
logs | 
if (attributes["PartA.name"] != null and attributes["PartA.name"] ! ="") {
  set name = attributes["PartA.name"]
} |
if (attributes["PartB.name"] != null and attributes["PartB.name"] ! ="") {
  set name = attributes["PartB.name"]
} |
if (attributes["PartB.body"] != null) {
  set body = attributes["PartB.body"]
}
exclude attributes["PartA.name"], attributes["PartB.name"] , attributes["PartB.body"] // drop the attributes
```

There are probably some minor gaps we have currently, including:
- decoding hex/checking byte array length
- parsing datetimes


All that said, I think at least trying query-engine might be a worthwhile approach down the line.

albertlockett (approved):
LGTM @lalitb . One minor comment for improving perf in `arrow_log_attrs_have_key` simply for your consideration

rust/otap-dataflow/crates/contrib-nodes/src/processors/microsoft_common_schema_processor/mod.rs
lalitb: Good point. I changed this to scan the dictionary values once, record the matching value indexes, and then scan the row keys for a reference to one of those indexes. That avoids repeatedly comparing the
  same dictionary value for long, low-cardinality arrays.

  I kept the extra key scan instead of assuming the dictionary value is referenced, so an orphan dictionary value still won’t produce a false positive. Added a small regression test for that case too.

rust/otap-dataflow/crates/contrib-nodes/src/processors/microsoft_common_schema_processor/mod.rs
lalitb: Makes sense. I updated the TODO to call out query-engine as the likely future path here, since it already supports parts of this flow like setting root fields from attributes and dropping promoted attrs.

rust/otap-dataflow/crates/contrib-nodes/src/processors/microsoft_common_schema_processor/README.md
```diff
@@ -0,0 +1,79 @@
...
+emit Microsoft Common Schema records, such as the Linux `user_events` receiver
+using `event_header` format.
+
+## Configuration
+
+The processor currently has no options:
+
+```yaml
+processors:
+  common_schema:
```
cijothomas, line 16:
```suggestion
  microsoft_common_schema:
```

rust/otap-dataflow/crates/contrib-nodes/src/processors/microsoft_common_schema_processor/README.md
```diff
@@ -0,0 +1,79 @@
...
+Unknown configuration fields are rejected.
+
+## Promotion Rules
+
+Only records with `__csver__ == 0x400` and `PartB._typeName == "Log"` are
+promoted. Other records are forwarded unchanged.
+
+| Common Schema attribute | OTLP destination |
+| --- | --- |
+| `PartA.time` | `time_unix_nano` |
```
cijothomas, line 30: do we handle the situation where `time_unix_nano` already was populated? Similar for other fields?

rust/otap-dataflow/crates/contrib-nodes/src/processors/microsoft_common_schema_processor/mod.rs
```diff
@@ -0,0 +1,1058 @@
...
+//! reusable cross-batch transform that can read attributes by `parent_id`, set
+//! root log columns such as body, event name, severity, trace/span IDs, flags,
+//! and timestamp, then rebuild `LogAttrs` with promoted fields removed and
+//! remaining fields preserved. The query-engine crate is a promising path for
+//! an Arrow-native implementation because it already supports some of these
+//! transformations, such as setting root log fields from attributes and
+//! dropping promoted attributes. Remaining gaps include Common Schema-specific
+//! parsing and validation such as decoding hex trace/span IDs, checking byte
+//! lengths, and parsing datetimes. Until that Arrow-native promotion path
+//! exists, Arrow-native pipelines pay an Arrow-to-OTLP conversion here.
```
jmacd, line 25: Please file an issue, thanks for explaining. I might title this "OTAP-direct Microsoft Common Schema processor".

There are probably two ways to go: (1) Arrow-based implementation of MCS; the would be better to convert OTLP to OTAP vs decode/encode protobuf?; (2) A query engine form as @albertlockett suggested (https://github.com/open-telemetry/otel-arrow/pull/2788#discussion_r3170473575)

rust/otap-dataflow/crates/contrib-nodes/src/processors/microsoft_common_schema_processor/mod.rs
jmacd: Thanks @albertlockett!

rust/otap-dataflow/crates/contrib-nodes/src/processors/microsoft_common_schema_processor/mod.rs
```diff
@@ -0,0 +1,1058 @@
...
+                if promote_microsoft_common_schema_log(log) {
+                    stats.records_promoted = stats.records_promoted.saturating_add(1);
+                }
+            }
+        }
+    }
+    stats
+}
+
+fn arrow_logs_have_no_csver(records: &OtapArrowRecords) -> bool {
```
jmacd, line 286:
```suggestion
/// Quick test that helps avoid decoding and encoding OTLP protobufs where no record has 
/// a common schema version.
fn arrow_logs_have_no_csver(records: &OtapArrowRecords) -> bool {
```

jmacd (approved):

lalitb:
converting to draft, as we are still discussing - this functionally Is very MS specific, and if this can be added in genera exporter/uploader.

lalitb:
> converting to draft, as we are still discussing - this functionally Is very MS specific, and if this can be added in genera exporter/uploader.

closing this. This would be eventually handled directly in geneva exporter through geneva-uploader library ( refer - https://github.com/open-telemetry/opentelemetry-rust-contrib/pull/620)

Issue #2725, opened 2026-04-21
OTLP protobuf message size limit


jmacd:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

At several points in our code, we construct a `ProtoBuffer` object and encode data in OTLP into a buffer.

The buffer size is not currently limited. We use a 4-byte placeholder at locations, which technically allows up to 256MiB per message; however nothing enforces this limit.

### Proposed Solution

We have an `impl TryFrom<OtapArrowRecords> for OtlpProtoBytes` block, no limit.

The easiest solution is to impose a single limit across the whole binary. The natural limit would b 256MiB, as it stands.

### Alternatives Considered

The limit with a 3-byte placeholder is 2MiB, the limit with a 2-byte placeholder is 16KiB, and the 1-byte limit is 128.

If we could agree to make a 2MiB limit in all situations, this would be easy.

jmacd:
Consider using a `try_from_limited()`. Could be thread-level, could be configurable.
or `try_convert_with_options()`.

PR #2792, opened 2026-04-30
Internal logging safety: protocol buffer size limits

### LINKED ISSUE #1746: Improvements in OTLP encoding logic required for internal logging safety

jmacd:
# Change Summary

Makes protocol buffer generation limited. Use a stack-allocated array (256 bytes) for logged attributes, a fixed constant.

~Replaces `try_into` for OtapPayloads with `try_into_with_options` and `try_into_with_default` to apply default settings.~
Moved to #2811.

The default size is large to avoid surprises (256MiB, the hard max implied by 4-byte placeholders). In a future PR we can tighten this and add configurable limits that apply by default anywhere OTLP protos are generated in a pipeline.  

Implements `dropped_attributes` accounting for records the internal logging pipeline.

Implements string truncation for string-valued fields, with `[...]` inserted if possible. Truncated fields are counted in dropped_attributes. This applies a measure of fairness to the fields of a log record as follows: when the string value is inserted it is allowed to use up to 50% of the remaining bytes or be truncated. 

Makes the `ProtoBuffer` padding width be determined by the available space, removes a macro used to set padding width.

Fixes #1746 
Part of #2725

A new benchmark was added, a heavy self-tracing workload with 10 string attributes and body, deliberately overflowing the 256B inline budget. The old code would re-allocate a Vec to log this event, the new code does not.

| Benchmark                 | upstream/main | jmacd/logs_sdk |       Δ |
| ------------------------- | ------------: | -------------: | ------: |
| `new_record`              |      170.4 µs |       183.2 µs |  +7.5%  |
| `encode_proto`            |      195.3 µs |       195.2 µs |  ~0%    |
| `encode_proto_with_scope` |      206.0 µs |       225.6 µs |  +9.5%  |
| `encode_and_format`       |      862.1 µs |       559.6 µs | −35.1%  |
| `format_with_entity`      |      934.3 µs |       627.5 µs | −32.8%  |

There is some cost for protocol writers, this is the cost of safety. For console writers, this is a win because we use a stack-allocated buffer for the attributes encoding. Note that `raw_error!` is now allocation-free.

Yes. We begin to observe dropped logs attributes instead of permitting very large log records in the internal logs pipeline.

For the normal data path, e.g., OTAP to OTLP conversions, the encoding is all-or-none. We will see encode failures instead of exceeding the 256MiB limit.

## Open questions?

For the internal logging path, a call to `with_max_remaining()` ensures that each logged field can only consume half the remaining buffer space. This only applies to the internal logging path. The normal OTLP logging path would greedily consume the whole buffer, thought at this time we do not set a limit below the maximum so this will be unlikely except for internal logging events. This could be refined to be more restrictive, e.g., to say log attribute values cannot exceed 1024 bytes as opposed to half the remaining space. Another policy would be greedier. Either way, this can be refined by adding to ConversionOptions.


rust/otap-dataflow/crates/pdata/src/otlp/common.rs
```diff
@@ -326,114 +323,508 @@ impl<'a> TryFrom<&'a RecordBatch> for AnyValueArrays<'a> {
...
-/// Buffer for encoding protobuf bytes.
-#[derive(Debug, Default)]
-pub struct ProtoBuffer {
-    buffer: Vec<u8>,
+/// Suffix appended to truncated strings.
+pub const TRUNCATION_SUFFIX: &[u8] = "[...]".as_bytes();
+
+/// Returned when a buffer write cannot fit within the size limit.
+///
+/// Used as the error type in [`EncodeResult`] returns from
```
albertlockett: nit: reads a bit funny

```suggestion
/// Used as the error type in [`EncodeResult`] which is returned from
```

rust/otap-dataflow/crates/pdata/src/otlp/common.rs
```diff
@@ -326,114 +323,508 @@ impl<'a> TryFrom<&'a RecordBatch> for AnyValueArrays<'a> {
...
+    #[inline]
+    fn encode_bytes(&mut self, field_tag: u64, val: &[u8]) -> EncodeResult {
+        self.encode_field_tag(field_tag, wire_types::LEN)?;
+        self.encode_varint(val.len() as u64)?;
+        self.try_extend(val)
+    }
+
+    /// Encode a string field, truncating with a `[...]` suffix if it won't fit.
+    ///
+    /// Returns `Err(Dropped)` if not even a minimal truncated form fits.
```
albertlockett: This also returns `Err(Dropped)` if a truncated form would fit, right? Should this comment be updated? To me it kind of makes it seem like it won't return `Err(Dropped)` if it writes truncated data

rust/otap-dataflow/crates/pdata/src/otlp/common.rs
```diff
@@ -326,114 +323,508 @@ impl<'a> TryFrom<&'a RecordBatch> for AnyValueArrays<'a> {
...
+            2 => self.encode_len_delimited_with::<2, E, _>(field_tag, f),
+            3 => self.encode_len_delimited_with::<3, E, _>(field_tag, f),
+            _ => self.encode_len_delimited_with::<4, E, _>(field_tag, f),
+        }
+    }
+
+    /// Implementation of [`encode_len_delimited`](Self::encode_len_delimited)
+    /// with an explicit `B`-byte placeholder width.
+    #[inline]
+    fn encode_len_delimited_with<const B: usize, E, F>(
```
albertlockett: nit: just double checking - this be `encode_len_delimited_width` or is `with` intentional?

rust/otap-dataflow/crates/pdata/src/otlp/common.rs
```diff
@@ -460,96 +891,131 @@ impl AsMut<[u8]> for ProtoBuffer {
...
+pub struct StackProtoBuffer<const N: usize> {
+    buf: [u8; N],
+    pos: usize,
+    limit: usize,
+}
+
+impl<const N: usize> StackProtoBuffer<N> {
+    /// Construct a stack buffer bounded to `N` bytes.
+    #[must_use]
+    pub const fn with_inline() -> Self {
```
albertlockett: nit: should this just be `new` instead of `with_ineline`? Seems like it would be a more idiomatic name for a constructor and it has alignment with the name of the constructor for the other variant of this type `Protobuf`

albertlockett (commented):
LGTM @jmacd . A few minor questions/comments but nothing blocking

albertlockett (approved):

albertlockett:
The protobuf improvements to `otlp::common` are very cool!

rust/otap-dataflow/crates/pdata/src/otlp/common.rs
jmacd: Thanks 🦅

rust/otap-dataflow/crates/pdata/src/otlp/common.rs
jmacd: Fixed

rust/otap-dataflow/crates/pdata/src/otlp/common.rs
jmacd: I made it `default()`. Not sure when to use new() vs default() but this is one less method.

PR #2794, opened 2026-04-30
Add capability to evaluate field-level type checks

### LINKED ISSUE #2752: Support logical expressions that check the type of fields or the stream elements

albertlockett:
# Change Summary

Part 2 of what was started in https://github.com/open-telemetry/otel-arrow/pull/2754 ...

Adds the capability to OPL and the OTAP query-engine to evaluate logical expressions that check the type of some field. For example:
```kql
logs | where attributes["x"] is String
logs | where attributes["y"] is Integer
```

This i useful for cases where users may wish to do something with a field of unknown type (an `AnyValue`) and the operation would fail if the value was not the correct type. For example, maybe I wish to redact log body only if it is text, when I know that sometimes the body will actually be an int:
```kotlin
logs | if (body is String) {
  set body = concat(substring(body, 0, 3), "****") // <-- substring only accepts string type
}
```

In this case we use the `GetType` expression which is only supposed to return a value from the `ValueType` enum (see discussion here: https://github.com/open-telemetry/otel-arrow/pull/2754#discussion_r3139794049), so currently only these values are supported. Although there are in fact many other types supported by OTel (e.g. specific int types like uint32, some etc.) the set of types in `ValueType` seem to be enough for most useful expressions, at least in the context of what the engine currently supports.

* Closes #2752

Unit tests

Yes - this type of expressions is now available in transform processor.


rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
```diff
@@ -535,6 +538,66 @@ impl FilterPlan {
...
+                }
+
+                if expr_source.expr_type.is_concrete() {
+                    // since the type we're not equal, and we know what type exactly the
+                    // expression will return, we know the types do not match here:
+                    return Ok(Some(FilterPlan::from(lit(false))));
+                }
+
+                // call the is_type function on the result of the expression
+                expr_source.logical_expr = Expr::ScalarFunction(ScalarFunction::new_udf(
```
lalitb, line 590: For the AnyValue-backed case, could we check the AnyValue `type` tag directly instead of going through the generic expression/UDF path?

For something like `attributes["x"] is String`, we do not need the actual value column. The per-row AnyValue type tag already tells us whether it is a string. That would avoid splitting by concrete type, running the UDF, and stitching the result back together for what may be a common filter path.

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
```diff
@@ -535,6 +538,66 @@ impl FilterPlan {
...
+                })?;
+
+                let expected_type = match value_type {
+                    ValueType::Boolean => ExprLogicalType::Boolean,
+                    ValueType::DateTime => ExprLogicalType::TimestampNanosecond,
+                    ValueType::Double => ExprLogicalType::Float64,
+                    ValueType::Integer => ExprLogicalType::AnyInt,
+                    ValueType::String => ExprLogicalType::String,
+                    ValueType::TimeSpan => ExprLogicalType::DurationNanoSecond,
+                    ValueType::Array | ValueType::Map | ValueType::Null | ValueType::Regex => {
```
lalitb, line 568: Should we support `Array`, `Map`, and `Null` here too?

Those should only need the `AnyValue` type tag, even if evaluating map/slice values is not supported yet.

lalitb (approved):
LGTM. I left a couple of non-blocking thoughts around the AnyValue-backed path.

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
```diff
@@ -535,6 +538,66 @@ impl FilterPlan {
...
+                };
+
+                if expr_source.expr_type == expected_type {
+                    // here we know the expression returns the type we are seeking. This may be
+                    // a trivial expression such as `severity_text is String`
+                    return Ok(Some(FilterPlan::from(lit(true))));
+                }
+
+                if expr_source.expr_type.is_concrete() {
+                    // since the type we're not equal, and we know what type exactly the
```
jmacd:
```suggestion
                    // since the types were not equal, and we know what type exactly the
```

jmacd (approved):

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
albertlockett: Oh yeah you're 100% right about that @lalitb. I'll document that in followup issue related to performance improvement.

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
albertlockett: Yes we should support them eventually. I was thinking it would be best to come back and add support for them when the engine actually supports these types of attributes. 

Although, as you pointed out [here](https://github.com/open-telemetry/otel-arrow/pull/2794#discussion_r3190466699), for `AnyValue` it probably makes sense to check the type column instead of inspecting the type of the values column, so in that case we probably could support these types easily. I will document that as a followup

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
albertlockett: opened https://github.com/open-telemetry/otel-arrow/issues/2841 for this

rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs
albertlockett: opened https://github.com/open-telemetry/otel-arrow/issues/2841 for this

PR #2796, opened 2026-04-30
[PerfTest] Fix shutdown hooks: use wait=false with readyz polling (#2795)

### LINKED ISSUE #2795: [PerfTest] Shutdown hooks with wait=true cause fatal suite abort when pipelines don't drain in time
## Pre-filing checklist

## Component(s)

## Bug Description

The `df-loadgen-steps-docker*.yaml` test step templates use `POST /api/v1/pipeline-groups/shutdown?wait=true&timeout_secs=60` to gracefully shut down the load generator, engine, and backend after metrics are collected.

However, `wait=true` blocks until all pipelines terminate. The admin HTTP server **stays alive after pipeline shutdown by design** (`run_forever` parks the main thread indefinitely), so if pipelines don't fully drain within `timeout_secs`, the endpoint returns **504 Gateway Timeout**.

Since the `send_http_request` hooks have no `on_error` configuration and `raise_for_status` defaults to `true`, this 504 causes the orchestrator to **abort the entire test suite** -- losing all metrics collected during the observation window.

## Steps to Reproduce

1. Run a test suite using `df-loadgen-steps-docker.yaml` with a workload that takes longer than `drain_timeout_secs` (default: 60s) to fully drain
2. Observe the suite fails at the "Stop Load Generator", "Stop Engine", or "Stop Backend" step with:
   ```
   requests.exceptions.HTTPError: 504 Server Error: Gateway Timeout for url: 
   http://localhost:808X/api/v1/pipeline-groups/shutdown?wait=true&timeout_secs=60
   ```
3. The report step never runs -- collected metrics are lost

This is reproducible in downstream deployments using the orchestrator framework, particularly with:
- Kafka exporter scenarios where the producer flush takes longer than 60s
- CI environments with Docker-in-Docker overhead increasing drain times
- Historical commits where the admin API path differs (returning 404 instead of 504)

## Expected Behavior

The suite should continue to the report step after shutdown, even if pipelines don't drain within `timeout_secs`. Metrics collected during the observation window should always be preserved.

## Actual Behavior

The suite aborts at the shutdown step with a 504 or 404 error, losing all collected metrics. The `when: always` report step never runs because the orchestrator treats the 504 as a fatal unrecoverable error.

## OTel-Arrow Version

Introduced in #1803 (commit `47264a5b`). Present in current HEAD.

## Affected Files

- `tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker.yaml` -- 3 shutdown hooks (ports 8085, 8086, 8087)
- `tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker-filtered.yaml` -- 3 shutdown hooks  
- `tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker-otel.yaml` -- 1 shutdown hook (port 8085)

## Proposed Fix

Switch from blocking `wait=true` to non-blocking `wait=false` with active readyz polling:

1. `POST /api/v1/pipeline-groups/shutdown?wait=false` -- returns 202 immediately, initiating pipeline drain in the background
2. Poll `GET /api/v1/readyz` every 5s until it returns `503` (pipelines terminated) or the endpoint stops responding
3. Add `on_error: { continue: true }` so 404/connection errors (e.g. service already exited) are non-fatal
4. The polling loop exits cleanly with a warning after 300s (the container will be killed by `destroy` regardless)

Example for one component:

```yaml
  - name: Stop Engine
    hooks:
      run:
        pre:
          - send_http_request:
              url: http://localhost:8086/api/v1/pipeline-groups/shutdown?wait=false
              method: POST
              headers:
                "Content-Type": "application/json"
              timeout: 10
              on_error:
                continue: true
          - run_command:
              command: >-
                for i in $(seq 1 60); do
                  status=$(curl -sf -o /dev/null -w "%{http_code}"
                    http://localhost:8086/api/v1/readyz 2>/dev/null || echo "000");
                  if [ "$status" = "503" ] || [ "$status" = "000" ]; then
                    echo "Engine pipelines stopped after ${i}s (status=${status})";
                    exit 0;
                  fi;
                  sleep 5;
                done;
                echo "WARNING: Engine still running after 300s, proceeding anyway";
                exit 0
    action:
      no_op: {}
```

This approach:
- Never aborts the suite due to shutdown timing -- collected metrics are always preserved
- Provides clear feedback ("Engine pipelines stopped after Ns")
- Handles 404 (older API paths) and connection refused (process already exited) gracefully
- Is consistent with the design of `run_forever` -- the process doesn't exit on pipeline shutdown

I'm happy to submit a PR with this change if it looks reasonable. /cc @cijothomas

daviddahl:
## Summary

Fixes #2795.

The `df-loadgen-steps-docker*.yaml` test step templates used `wait=true` on shutdown hooks, which blocks until all pipelines terminate. The admin server **stays alive after pipeline shutdown by design** (`run_forever` parks the main thread), so `wait=true` returns **504 Gateway Timeout** when drain takes longer than `timeout_secs`.

Since `send_http_request` defaults to `raise_for_status: true` with no `on_error` configured, this 504 aborts the entire suite before the report step -- losing all metrics collected during the observation window.

## Changes

Switch from blocking `wait=true` to non-blocking `wait=false` with active readyz polling across all three affected template files:

| File | Hooks fixed |
|---|---|
| `df-loadgen-steps-docker.yaml` | 3 (ports 8085, 8086, 8087) |
| `df-loadgen-steps-docker-filtered.yaml` | 3 (ports 8085, 8086, 8087) |
| `df-loadgen-steps-docker-otel.yaml` | 1 (port 8085) |

**Pattern for each shutdown hook:**
1. `POST .../shutdown?wait=false` -- returns 202 immediately, drain runs in background
2. Poll `/readyz` every 5s until 503 (pipelines terminated) or endpoint stops responding
3. `on_error: continue: true` so 404/connection errors (service already exited) are non-fatal
4. Polling loop exits with a warning after 300s (container will be killed by `destroy` regardless)

## Testing

Verified in downstream deployments using the orchestrator framework across a range of workloads. The polling approach provides clear log output ("Engine pipelines stopped after Ns") and never aborts the suite due to shutdown timing.

/cc @cijothomas


tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker.yaml
```diff
@@ -112,17 +112,39 @@ steps:
...
       run:
         pre:
           - send_http_request:
-              url: http://localhost:8085/api/v1/pipeline-groups/shutdown?wait=true&timeout_secs={{drain_timeout_secs | default(60)}}
+              url: http://localhost:8085/api/v1/pipeline-groups/shutdown?wait=false
               method: POST
               headers:
                 "Content-Type": "application/json"
-              timeout: {{drain_timeout_secs | default(60) | int + 5}}  # Client timeout slightly longer than server
+              timeout: 10
```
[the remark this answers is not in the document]
cijothomas: +1 - lets keep the timeout

tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker.yaml
```diff
@@ -144,11 +181,24 @@ steps:
...
+              on_error:
+                continue: true
+          - run_command:
+              command: >-
+                for i in $(seq 1 60); do
+                  if ! curl -sf http://localhost:8087/api/v1/telemetry/metrics?reset=false >/dev/null 2>&1; then
```
cijothomas: this looks inconsistent- it is querying for metrics endpoint after shutdown as a way to poll if shutdown is done? /readyz is what we use elsewhere in the PR.

daviddahl: oops, switching to /readyz.

tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker.yaml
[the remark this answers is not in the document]
daviddahl: Fixed -- `timeout_secs` restored on all shutdown URLs.

tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker.yaml
daviddahl: Fixed -- all three now poll `/readyz` consistently.

daviddahl:
All feedback addressed in latest push (rebased on current main):
1. `timeout_secs` restored on all URLs
2. Backend polling uses `/readyz` (was `/telemetry/metrics`)
3. Elapsed time: `${i}s` → `$((i * 5))s`
4. Log messages: "stopped" → "no longer ready"
5. URLs updated to `/api/v1/groups/` (post-rebase)

@cijothomas ready for re-review when you have a chance.

daviddahl:
Fixed CI failure -- conflict resolution during rebase dropped `method: POST` from the `send_http_request` blocks. Restored in latest push.

tools/pipeline_perf_test/test_suites/integration/templates/test_steps/df-loadgen-steps-docker-filtered.yaml
```diff
@@ -110,17 +110,33 @@ steps:
...
+                for i in $(seq 1 60); do
+                  status=$(curl -sf -o /dev/null -w "%{http_code}"
+                    http://localhost:8085/api/v1/readyz 2>/dev/null || echo "000");
+                  if [ "$status" = "503" ] || [ "$status" = "000" ]; then
+                    echo "Load generator no longer ready after $((i * 5))s (status=${status})";
+                    exit 0;
+                  fi;
+                  sleep 5;
+                done;
+                echo "WARNING: Load generator still running after 300s, proceeding anyway";
```
JakeDern: I'm probably missing something, but could you explain how this is this different from setting `raise_for_status: false` and bumping the drain timeout from 60s to 300s?

I'm also wondering under which circumstances is draining taking longer than a minute? That's a pretty long time to flush the queues when we know the df engine can process millions of records/sec 😜

daviddahl: I think the problem is on our side where we are testing with Kafka, etc

JakeDern: Is setting `raise_for_status: false` and bumping the timeout a valid solution to this problem instead of writing this shell script in all the polling steps?

daviddahl: You're right that `raise_for_status: false` + a longer `timeout_secs` would also prevent the suite abort. The polling approach adds live log feedback and early exit, but it's more code.

The drain timing issue came up in downstream deployments with heavier exporter configurations where flush times regularly exceeded the 60s default. The other common failure was 404 from the API path rename when testing against historical commits.

Happy to simplify to `raise_for_status: false` if the team prefers -- the polling is nice-to-have for observability but not strictly necessary.

daviddahl:
Simplified -- dropped the polling, just `raise_for_status: false` + `on_error: { continue: true }` on the existing hooks.

JakeDern (approved):
LGTM, thanks @daviddahl!

albertlockett (approved):

PR #2804, opened 2026-05-01
fix(engine): centralize telemetry timer management in runtime manager

### LINKED ISSUE #1305: Periodic telemetry API improvement

cijothomas:
Fixes https://github.com/open-telemetry/otel-arrow/issues/1305

## Motivation

Our idle perf tests were showing surprisingly high CPU usage. Investigation revealed that each node was independently calling `start_periodic_telemetry(Duration::from_secs(1))`, triggering internal metric collection (syscalls like `getrusage`, jemalloc stats, tokio worker metrics, channel snapshots) every second across all nodes. At 1s intervals, the telemetry overhead itself dominated idle CPU measurements, producing misleading results.

## Changes

Centralizes telemetry timer registration in the runtime control manager so all nodes use the configured `engine.telemetry.reporting_interval` instead of each node managing its own timer independently. This removes ~250 lines of per-node boilerplate across 15 nodes, with shutdown cancellation handled centrally.

Also bumps the idle perf test to `reporting_interval: 5s` (with matching 5s Prometheus scrape) for a more realistic deployment baseline. With centralized timing and the 5s interval, idle CPU numbers look significantly better — engine CPU drops ~2.3x and pipeline CPU drops ~6.4x compared to the previous 1s configuration.

Adds validation rejecting zero-duration `reporting_interval` to prevent accidental spin loops.

**Notes**
- `perf_exporter.config.frequency` is now silently ignored; deprecation/cleanup is a follow-up.
- `start_periodic_telemetry` API is still public — TBD whether to remove.
- TODO comment added in `pipeline_ctrl.rs` for an eager-timer-registration race flagged in review (telemetry can queue ahead of `Shutdown` in a slow-starting node's bounded control channel); to be addressed via a node-ready signal in a follow-up.


JakeDern (approved):
Looks good to me

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
```diff
@@ -1247,7 +1259,7 @@ mod tests {
     use tokio::task::LocalSet;
     use tokio::time::timeout;
 
-    const TEST_CONTROL_PLANE_METRICS_FLUSH_INTERVAL: Duration = Duration::from_millis(10);
+    const TEST_CONTROL_PLANE_METRICS_FLUSH_INTERVAL: Duration = Duration::from_secs(3600);
```
lquerel, line 1262: Why this change?

lquerel (approved):
Thank you for cleaning up this part of our internal telemetry reporting!

rust/otap-dataflow/crates/engine/src/pipeline_ctrl.rs
cijothomas: This is the per-node `CollectTelemetry` timer interval (new constant, not `TEST_CONTROL_PLANE_METRICS_FLUSH_INTERVAL`). Since timers are now pre-registered in the constructor, a low value floods control channels with `CollectTelemetry` messages during tests, causing 8 test failures. Setting it to 3600s keeps them dormant; tests that need telemetry drive it manually.

PR #2810, opened 2026-05-01
  [Geneva Exporter] Encode logs using LogsDataView

lalitb:
## Description

  Updates the otel-arrow Geneva exporter to encode log payloads through the shared `LogsDataView` abstraction instead of converting logs back through OTLP proto objects.

  ## Changes

  - Encode OTAP Arrow logs through `OtapLogsView`, the OTAP implementation of `LogsDataView`.
  - Encode OTLP byte logs through `RawLogsData`, the OTLP bytes implementation of `LogsDataView`.
  - Call the existing `geneva-uploader` `encode_and_compress_logs(&view)` API for both log paths.
  - Patch `otap-df-pdata` / `otap-df-pdata-views` so `geneva-uploader` uses the workspace view traits.
  - Add a TODO for compressed byte accounting until `EncodedBatch::data_len()` is exposed upstream.

  - `cargo fmt --all`
  - `CARGO_NET_GIT_FETCH_WITH_CLI=true cargo check -p otap-df-contrib-nodes --features "geneva-exporter otap-df-otap/crypto-ring"`

None.


rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -614,18 +585,13 @@ impl GenevaExporter {
                             message = "Uploading logs to Geneva using OTLP path"
                         );
 
-                        // Decode OTLP bytes to ResourceLogs
-                        let logs_request =
-                            ExportLogsServiceRequest::decode(&bytes[..]).map_err(|e| {
-                                self.metrics.conversion_errors.inc();
-                                format!("Failed to decode logs request: {}", e)
-                            })?;
+                        let logs_view = RawLogsData::new(bytes.as_ref());
```
AaronRM: The change from `ExportLogsServiceRequest::decode` => `RawLogsData::new` removes OTLP input validation and is the cause of [the geneva_exporter_emits_nack_for_decode_failure test failure in the CI run](https://github.com/open-telemetry/otel-arrow/actions/runs/25237722220/job/74008790593?pr=2810). This is a must fix regression.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
AaronRM, line 1: This PR should also enable the `test_geneva_exporter_creates_view_from_otap_records` and `test_geneva_exporter_handles_missing_logs_batch` tests which are commented out.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -372,7 +371,9 @@ impl GenevaExporter {
         // to avoid an intermediate Vec allocation.
         let mut stream = futures::stream::iter(batches.iter())
             .map(|batch| {
-                let batch_size = batch.data.len() as u64;
+                // TODO: restore compressed byte accounting after geneva-uploader exposes
```
AaronRM: Is there an issue tracking this?

AaronRM (changes requested):

rust/otap-dataflow/crates/pdata/src/views/otap/logs.rs
```diff
@@ -149,6 +149,24 @@ impl<'a> TryFrom<&'a OtapArrowRecords> for OtapLogsView<'a> {
...
+            })?;
+
+        let resource_attrs = records.get(ArrowPayloadType::ResourceAttrs);
+        let scope_attrs = records.get(ArrowPayloadType::ScopeAttrs);
+        let log_attrs = records.get(ArrowPayloadType::LogAttrs);
+
+        Self::new(logs_batch, resource_attrs, scope_attrs, log_attrs)
+    }
+}
+
```
utpilla: No need to add this method. We should be able to use the existing `OtapArrowRecords` to `OtalpLogsView` conversion method defined above. You could update the exporter to use [@ bindings](https://doc.rust-lang.org/book/ch19-03-pattern-syntax.html#using--bindings) and use the existing conversion method. Something like this:

```rust
match payload {
    // OTAP Arrow path: encode logs through LogsDataView without converting back to OTLP.
    OtapPayload::OtapArrowRecords(otap_records) => {
        match otap_records {
            // `@` binding: matches the `Logs` variant but keeps the whole
            // `OtapArrowRecords` value bound to `otap_records`, so we can pass it
            // to `OtapLogsView::try_from(&OtapArrowRecords)` without moving the
            // inner `Logs` out of the enum.
            otap_records @ OtapArrowRecords::Logs(_) => {
                otel_info!(
                    "geneva_exporter.upload",
                    message = "Uploading logs to Geneva using OTAP record views"
                );
                ...
            }
            ...
        }
    }
}
```

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
utpilla: I think we could add a `RawLogsData::try_new(buf) -> Result<Self, Error>` method that walks the top-level proto framing once (single linear pass, no allocations, no deep field decoding). Exporter then uses `try_new` instead of `new`.

I suggest adding it to `RawLogsData` because this check would be needed by any component that works on OTLP bytes directly.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -483,54 +484,24 @@ impl GenevaExporter {
...
-                        let otlp_bytes: OtlpProtoBytes =
-                            OtapPayload::OtapArrowRecords(OtapArrowRecords::Logs(otap_records))
-                                .try_into()
-                                .map_err(|e| {
-                                    self.metrics.conversion_errors.inc();
-                                    format!("Failed to convert OTAP to OTLP: {:?}", e)
-                                })?;
-
-                        let OtlpProtoBytes::ExportLogsRequest(bytes) = otlp_bytes else {
+                        let logs_view = OtapLogsView::try_from(&otap_records).map_err(|e| {
```
utpilla, line 512: The new path constructs `OtapLogsView::try_from(&otap_records)` directly. `OtapLogsView::new` then calls `group_by_resource_id`, `group_by_scope_id`, and `build_attribute_index`, all of which read raw `RESOURCE.ID` / `SCOPE.ID` / `parent_id` columns. If the records arrived from an OTAP gRPC receiver, those IDs are still delta-encoded and `parent_id` is quasi-delta encoded, so resource/scope grouping and attribute lookup will be wrong with no error and no diagnostic.
- For OTLP→OTAP via the local batch processor, IDs are freshly assigned and not transport-encoded, so that path is fine.
- For the syslog receiver path mentioned in the doc comment, same reasoning — likely fine.
- For urn:otap:receiver:arrow → batch → geneva, this regresses to silent wrong output.


We should do two things:
1. Add a centralized check in `OtapLogsView::new` using the existing column-encoding metadata (`is_column_encoded(...)` in `transport_optimize.rs`). Refuse to build the view over still-encoded IDs and return a typed error. Something like:

```rust
// In OtapLogsView::new
if is_column_encoded(RESOURCE_ID_COL_PATH, schema) != Some(false)
   || is_column_encoded(SCOPE_ID_COL_PATH, schema) != Some(false)
   || /* parent_id encoding on attr batches */ ...
{
    return Err(Error::TransportOptimizedIdsNotDecoded);
}
```
2. The exporter then calls `otap_records.decode_transport_optimized_ids()?` before `OtapLogsView::try_from(...)`

albertlockett: Great catch @utpilla - that's exactly right, and I think the solution you have suggested is the right way to go. Minor question wrt to point 1: I wonder if the check should be in `new` or `try_from` constructor? `new` is currently infallible. 

If we simply called `otap_records.decode_transport_optimized_ids()?`, that would be enough for to correct the behaviour for PR. Maybe we could document/implement the changes needed to the `OtapLogsView` constructor in another issue/PR?

lalitb:
Thanks @AaronRM @utpilla, these are fair comments.
Lesson learned: I trusted Friday-evening-me to review this properly before raising it. That was clearly too much trust. I’ll clean this up properly and push an update.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
utpilla: > If we simply called otap_records.decode_transport_optimized_ids()?, that would be enough for to correct the behaviour for PR.

Yes, I think just for this PR, calling `otap_records.decode_transport_optimized_ids()?` should be good enough.

> Maybe we could document/implement the changes needed to the OtapLogsView constructor in another issue/PR?

Year we could create another issue/PR for the centralized check. 

> Minor question wrt to point 1: I wonder if the check should be in new or try_from constructor? new is currently infallible.

I was wondering if instead of adding a new method should we make the `new` method fallible. That would change a lot of call sites but that would enforce that no component erroneously runs into this issue. What do you think?

lalitb: For this PR, I kept the fix narrow: Geneva now calls `decode_transport_optimized_ids()` before constructing `OtapLogsView`, per your suggestion.

 I can file a follow-up issue for the broader centralized guard. One clarification for that follow-up: `OtapLogsView::new` is already fallible today (`Result<Self, Error>`), so adding the guard there would not require changing its signature. The design question is whether the guard belongs in `new` for strongest enforcement, or only in `TryFrom<&OtapArrowRecords>` for a narrower change.

lalitb: > I can file a follow-up issue for the broader centralized guard. 

Added #2824 to track. We can discuss there if needed.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -484,54 +495,29 @@ impl GenevaExporter {
...
-                        //     .encode_and_compress_logs_view(&logs_view)
-                        //     .map_err(|e| format!("Failed to encode logs from view: {}", e))?;
-
-                        // Fallback path: Convert OTAP Arrow → OTLP bytes
+                    mut otap_records @ OtapArrowRecords::Logs(_) => {
                         otel_info!(
-                            "geneva_exporter.convert",
-                            message = "Converting OTAP logs to OTLP bytes (fallback path)"
+                            "geneva_exporter.upload",
+                            message = "Uploading logs to Geneva using OTAP record views"
```
utpilla: nit: We could be consistent with the error message phrasing "using OTAP recover views" vs "using OTLP path"

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -543,7 +529,7 @@ impl GenevaExporter {
                         otel_info!(
                             "geneva_exporter.upload",
                             count = batches_uploaded,
-                            message = "Successfully uploaded log batches to Geneva (OTAP fallback)"
+                            message = "Successfully uploaded log batches to Geneva (OTAP views)"
```
utpilla: nit: We can be consistent with error message phrasing for "Uploading ... **using OTAP record views**" and "Successfully uploaded...**Geneva (OTAP views)**"

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
utpilla: Same as https://github.com/open-telemetry/otel-arrow/pull/2810#discussion_r3184766474

We can be consistent. Either say "using ..." or use brackets/

rust/otap-dataflow/crates/pdata/src/views/otlp/bytes/logs.rs
```diff
@@ -46,14 +46,72 @@ impl<'a> RawLogsData<'a> {
     pub const fn new(buf: &'a [u8]) -> Self {
         Self { buf }
     }
+
+    /// Create a new instance after validating top-level protobuf wire framing.
+    ///
+    /// This validates the outer `ExportLogsServiceRequest` envelope without allocating or
+    /// recursively decoding nested `ResourceLogs` bodies.
```
utpilla: This reads like the function only peeks at the first few bytes. In practice it does a full linear pass over `buf`: every field tag is parsed, every LEN field has its length verified against the buffer end, every fixed field's bounds are checked. What it doesn't do is interpret the LEN payloads as nested messages. We could be more specific here.

```rust
/// Construct a [`RawLogsData`] after validating top-level protobuf wire framing.
///
/// Cost: a single linear walk of `buf` (`O(buf.len())`) with no allocations.
/// Each field tag is decoded and each length-delimited / fixed field is bounds-checked
/// against the end of `buf`. LEN payloads are not recursively decoded as messages —
/// nested `ResourceLogs` / `ScopeLogs` / `LogRecord` content is interpreted lazily on
/// access via the view APIs.
```

utpilla (approved):
Left a few minor non-blocking suggestions! Thanks @lalitb!

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
lalitb: I created an issue in otel-rust-contrib repo, and added the link as comment here.

lalitb:
@AaronRM - Can you have a look when possible - as you have requested the changes, and it is now resolved.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -411,7 +414,9 @@ impl GenevaExporter {
                     }
                     succeeded += 1;
                     records_ok += row_count;
-                    bytes_ok += batch_size;
+                    if let Some(batch_size) = batch_size {
```
ThomsonTan, line 417: Would this branch ever be reached? It looks like `batch_size` is always returned as `None` from the stream above, so this `Some` match will never succeed.

AaronRM (approved):

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
lalitb: It would be eventually reached (and the metrics emitted) once the support to return  batch_size` is added in geneva-uploader as mentioned in comment in line 375.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs
```diff
@@ -484,54 +495,29 @@ impl GenevaExporter {
...
-                        //     .await;
-                        // let logs_view = OtapLogsView::try_from(&otap_records)
-                        //     .map_err(|e| format!("Failed to build logs view: {}", e))?;
-                        // let batches = self
-                        //     .geneva_client
-                        //     .encode_and_compress_logs_view(&logs_view)
-                        //     .map_err(|e| format!("Failed to encode logs from view: {}", e))?;
-
-                        // Fallback path: Convert OTAP Arrow → OTLP bytes
+                    mut otap_records @ OtapArrowRecords::Logs(_) => {
```
jmacd, line 501: Today is a good day to learn new Rust syntax.

jmacd (approved):

PR #2822, opened 2026-05-04
fix(perf-test): flush loadgen metrics during run, not only on thread exit

cijothomas:
The Python load generator's three worker threads only flushed `logs_produced` / `failed` / `bytes_sent` to the shared metrics dict on thread exit, so every Prometheus scrape during the observation window saw `0` — causing the [nightly syslog dashboard](https://open-telemetry.github.io/otel-arrow/benchmarks/nightly/syslog/) to publish `logs_produced_rate=0` and `dropped_logs_percentage=0` for every syslog test since #2723 (2026-04-23). This PR flushes accumulators once per send-loop iteration; verified locally that `logs_produced_rate` is now `~12k` (UDP) / `~95k–200k` (TCP) instead of `0`.

Current state where logs count shows 0.
image


cijothomas:
@JakeDern PTAL.

JakeDern:
LGTM. Just curious - What format are these exported in? In the future we might want to compute these rates based off of timestamps attached to these metric data points rather than when we get/scrape them in the orchestrator. 

I think that would let us also go back to sending this data all at once if we wanted to by sending two datapoints after the run is over - A `0` with a timestamp at the start of the window and then the entire cumulative total with the timestamp at the end.

JakeDern (approved):

jmacd:
> In the future we might want to compute these rates based off of timestamps attached to these metric data points

Sign me up!

tools/pipeline_perf_test/load_generator/loadgen.py
```diff
@@ -406,21 +420,40 @@ def syslog_tcp_worker_thread(self, thread_id: int, args: dict) -> None:
...
-        total_bytes_sent = 0
-        total_late_batches = 0
+        # Accumulate metrics locally between flushes to keep lock contention low
+        # while still publishing during the observation window (not only on exit).
+        batch_sent = 0
+        batch_failed = 0
+        batch_bytes_sent = 0
+        batch_late_batches = 0
+
+        def flush_batch_metrics():
```
jmacd, line 430: Is this method duplicated? I wonder if we can factor it out.

cijothomas:
> LGTM. Just curious - What format are these exported in? In the future we might want to compute these rates based off of timestamps attached to these metric data points rather than when we get/scrape them in the orchestrator.
> 
> I think that would let us also go back to sending this data all at once if we wanted to by sending two datapoints after the run is over - A `0` with a timestamp at the start of the window and then the entire cumulative total with the timestamp at the end.

The loadgen exposes this in a very minimal Prometheus format. We could make a future improvement to have it send just two datapoints (at start and end) and adjust the rest of the orchestrator to react to it.

Issue #2099, opened 2026-02-23
OTLP & OTLP Exporter ignore shutdown deadline


albertlockett:
In both the gRPC & HTTP implementations of the OTLP exporter, when we receive the shutdown message we try to finish any inflight requests before returning from the `start` method.

https://github.com/open-telemetry/otel-arrow/blob/51a8fbab84a3819181adfd3010bcd261198aa94e/rust/otap-dataflow/crates/otap/src/otlp_http_exporter/mod.rs#L255-L269

https://github.com/open-telemetry/otel-arrow/blob/51a8fbab84a3819181adfd3010bcd261198aa94e/rust/otap-dataflow/crates/otap/src/otlp_exporter.rs#L227-L245

This could avoid blocking the shutdown for longer than the expected deadline. These should probably race some timeout future as suggested here: https://github.com/open-telemetry/otel-arrow/pull/2070#discussion_r2833286199

albertlockett:
Note - we should be careful about if/how we write unit tests for this to ensure they're reliable.

Parquet exporter has a similar shutdown behaviour to what we want here, where it races a timeout future:
https://github.com/open-telemetry/otel-arrow/blob/7aff0cf9268d4f9dbc8afaca22d13879aa07a0cf/rust/otap-dataflow/crates/otap/src/parquet_exporter.rs#L246-L269

However the shutdown timeout test is notoriously flakey, so we probably want to come up with a better pattern
https://github.com/open-telemetry/otel-arrow/blob/7aff0cf9268d4f9dbc8afaca22d13879aa07a0cf/rust/otap-dataflow/crates/otap/src/parquet_exporter.rs#L874-L880

gyanranjanpanda:
i would love to work on this

albertlockett:
hey @gyanranjanpanda I think @luckymachi is already working on it.

AvinashDevX:
I'd like to pick this up. Will make the shutdown path in both `otlp_exporter.rs` and `otlp_http_exporter/mod.rs` race the drain against the shutdown deadline (per the discussion linked above), so an in-flight request that overruns the deadline gets cut loose instead of blocking shutdown indefinitely.

luckymachi:
[...]

PR #2826, opened 2026-05-04
Improved exporter shutdown handling

luckymachi:
# Change Summary

Modified the shutdown handling for both the HTTP and gRPC exporters to make them resemble more than parquet implementation. This includes a modification of `InFlightExports` to be able to get the signal types in case a timeout happens.

This closes issue #2099 

It is planned to improve existing tests later either in this PR or on a new one exclusively for the exporters

None


rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
```diff
@@ -709,18 +750,28 @@ fn make_export_future(
     }
 }
 
+/// Wrapper designed to be able to properly get the signal type of an export in case of shutdown.
+/// [`FuturesUnordered`] provides no way to inspect pending futures, so without this wrapper
+/// there is no way to know which signal types are still in flight at any given point.
+struct InFlightExport<Output> {
+    signal_type: SignalType,
+    fut: Pin<Box<dyn Future<Output = Output>>>,
+}
```
albertlockett: nit: If we ever add additional fields to this, or us the type outside the context of shutdown handling, then the commentary about this being specifically signal_type/shutdown become less relevant. Also, the commentary about what we cannot do simply w/ `FuturesUnordered`, makes sense in the context of explaining this diff but makes less sense when considering this code in isolation (imo).

Maybe we could update the comments to something more general:
```suggestion
/// Wrapper designed to be able to retrieve information about the opaque export future.
struct InFlightExport<Output> {
    /// The signal type of the batch being exported
    signal_type: SignalType,
    fut: Pin<Box<dyn Future<Output = Output>>>,
}
```

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
```diff
@@ -235,19 +250,45 @@ impl Exporter<OtapPdata> for OTLPExporter {
...
-                            grpc_clients.release(client);
+
+                    let mut timeout = Delay::new(deadline.duration_since(Instant::now())).fuse();
+
+                    // Stop telemetry loop concurrently with flushing; do not block shutdown on cancel
+                    let mut timer_cancelled = false;
+
+                    loop {
+                        // If the exports are completed in time, we exit the loop to evaluate the
+                        // rest of the possible races
```
albertlockett: > to evaluate the rest of the possible races

What are the rest of the possible races we evaluate? I actually think we should be exiting the loop, then returning `TerminalState` right?

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
```diff
@@ -235,19 +250,45 @@ impl Exporter<OtapPdata> for OTLPExporter {
...
+                                }
+                                return Err(Error::IoError {
+                                    node: exporter_id.clone(),
+                                    error: std::io::Error::from(ErrorKind::TimedOut),
+                                })
+                            },
+                        };
                     }
-                    _ = timer_cancel_handle.cancel().await;
-                    return Ok(TerminalState::new(deadline, [self.pdata_metrics]));
```
albertlockett, line 304: I think we should still be returning `Ok(TerminalState::new(deadline, [self.pdata_metrics]))` after handling the shutdown signal, shouldn't we?

albertlockett (commented):
Looks like a good start @luckymachi, thanks for taking on this work!



Also, would you mind adding some unit tests to simulare the timeout behaviour and verify that we return metrics indicating that some requests failed? Starting with the otlp_http_exporter might be easier. You'll see in this test suite [we setup mock http servers](https://github.com/open-telemetry/otel-arrow/blob/8b4e674f6909f04cdd6726c42e3ad0f3281b3d6e/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/mod.rs#L770-L978), so you could create one that responds slowly, send some pdata, then send the shutdown signal, then verify the metrics. [Here's a test](https://github.com/open-telemetry/otel-arrow/blob/8b4e674f6909f04cdd6726c42e3ad0f3281b3d6e/rust/otap-dataflow/crates/core-nodes/src/processors/transform_processor/mod.rs#L675-L696) with an example of how to validate the metrics for you to go by.

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
luckymachi: Yeah that's a miss on my part; if all the exports are completed then we just exit the loop 👍

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
luckymachi: Yep, a miss from my part too. I'm keeping only the `return` since `timer_cancel_handle` is handled in an earlier part of the code now

utpilla:
> # Change Summary
> Modified the shutdown handling for both the HTTP and gRPC exporters to make them resemble more than parquet implementation. This includes a modification of `InFlightExports` to be able to get the signal types in case a timeout happens.
> 
> ## What issue does this PR close?
> This closes issue #2009
>... 

Updated the PR description to use the correct issue number (2009 -> 2099).

albertlockett:
reopening by request of contributor

luckymachi:
Greetings everybody! Hopefully this final implementation is good to go. Other than the tests themselves, some notable changes from the previous commits include adding additional checks when checking if `inflight_exports` is empty and removing `futures::future::pending::<()>().await` in `timer_cancel_fut`/`tel_timer_cancel_fut` construction.
- In the first case, it would happen that in case `inflight_exports` is empty, it'd get passed to `futures::select_biased!` as a valid case. However, `select_biased!` polls branches in the order they're listed and picks the first one that's ready, so an always-ready `next_completion()` branch would win on every single poll once the queue drains, starving the `tel_timer_cancel_fut` and `timeout` branches below forever. The loop would spin synchronously (it'd poll `next_completion()`, get `Ready(None)`, do nothing, loop back to the top, poll again) without ever returning `Poll::Pending` up to the executor. On this local/single-threaded runtime  it wedges the task entirely, since the timeout `Delay` and the telemetry-timer cancel future never get polled again.
- In the second case, the construction is done exactly once, when `cancel()` completes, then (thanks to `.fuse()`) returns `Pending` forever on every subsequent poll without re-running the body. Following the `cancel().await` with `futures::future::pending::<()>().await` here prevents this future from ever resolving to `Ready` so that the `timer_cancelled` flag below can never become `true`, which means shutdown always runs out the full `deadline` and returns `Err(TimedOut)` even when every export already finished successfully.
- The tests themselves used the Parquet suite as the main base for their design; as of right now all necessary tests pass! :)

I'd really appreciate your feedback and reviews, many thanks in advance guys! :D

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
```diff
@@ -235,18 +249,63 @@ impl Exporter<OtapPdata> for OTLPExporter {
...
-                        if let Some(completed) = inflight_exports.next_completion().await {
-                            let client = finalize_completed_export(
-                                completed,
-                                &effect_handler,
-                                &mut self.pdata_metrics,
-                            )
-                            .await;
-                            grpc_clients.release(client);
+
+                    let mut timeout = Delay::new(deadline.duration_since(Instant::now())).fuse();
```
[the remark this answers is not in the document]
jmacd: The comment looks valid.

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
jmacd: This I think will require a test?

jmacd (changes requested):
This branch will require a pretty major merge operation to resolve conflicts.

PR #2827, opened 2026-05-04
[Web Admin UI] Feature gate admin live log WebSocket

lalitb:
# Change Summary

This change adds a feature flag around the admin UI live log WebSocket path.

The WebSocket protocol requires `SHA-1` as part of the opening handshake to compute `Sec-WebSocket-Accept`. This is protocol validation, not use of `SHA-1` as a cryptographic security primitive. However, automated scanning systems can sometimes flag the transitive `sha1` crate because they cannot reliably distinguish this protocol-mandated use from general-purpose cryptographic use.

Making this path optional lets builds that do not need live log streaming avoid pulling in the WebSocket dependency chain.

  The feature remains enabled by default for normal development builds. When disabled, retained logs remain controlled by `engine.telemetry.logs.tap.enabled` and can still be queried through the HTTP logs endpoint when configured; only the live WebSocket streaming path is gated. In the admin UI, the Internal Logs panel is still rendered, but the live-stream controls are disabled when the WebSocket feature is not compiled in.

This is intended as a short-term containment change. We should separately evaluate the longer-term design: keep WebSockets with this feature boundary, move the log stream to SSE plus HTTP controls, use HTTP polling, or choose another approach that better fits the deployment requirements.

Manual test for admin UI - with and without flag.

Yes - if the flag is disabled, the Internal Logs panel is still rendered, but the live-stream controls are disabled.

### DIFF
--- a/rust/otap-dataflow/Cargo.toml
+++ b/rust/otap-dataflow/Cargo.toml
@@ -251,7 +251,7 @@ geneva-uploader = "0.4.0"
 
 [features]
 # ToDo When jemalloc is enabled, we could use jemalloc_pprof to profile where memory is allocated. See https://crates.io/crates/jemalloc_pprof
-default = ["jemalloc", "crypto-ring", "dev-tools"]
+default = ["jemalloc", "crypto-ring", "dev-tools", "admin-live-logs-ws"]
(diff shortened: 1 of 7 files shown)


lquerel (approved):
LGTM

jmacd (approved):

PR #2829, opened 2026-05-04
fix(otlp_grpc_exporter): Set accept_compressed in addition to send_compressed

### LINKED ISSUE #2828: otlp_grpc exporter doesn't set accept_compressed
### Pre-filing checklist

### Component(s)

### Description

We set `send_compressed` in the exporter, but not `accept_compressed` so if responses are streamed back with the same compression as they were sent then we reject it.

https://github.com/open-telemetry/otel-arrow/blob/b91157891d6b1a4b3d3198d24bb42c6d0209be02/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs#L822

### Additional Context

I found this when working on the benchmarks, see these warns from the exporter in the loadgen:

Image

But at the same time my system under test was successfully processing the messages.

JakeDern:
# Change Summary

Set `accept_compressed` so that we can process responses if the replies are compressed with the same codec.

* Closes #2828

I tried them locally and verified the warnings in the issue went away.

No.


lquerel (approved):
LGTM

jmacd (approved):

Issue #2782, opened 2026-04-29
Implement pdata-propagated stopwatch across processors


drewrelmas:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

See majority of context on #2556. The main goal is ability to produce a composite duration for a combination of processors. This is really a related sub-problem of the full `ProcessorChain` feature and can be accomplished without pursuing the full channel elimination / memory reduction that chains would bring.

### Proposed Solution

Follow up on the context-propagated stopwatch design as a shorter-term solution instead of doing the entire `ProcessorChain` implementation.

drewrelmas:
Closing this as completed for now

PR #2839, opened 2026-05-05
feat(otap-dataflow): Add stopwatch `signals.incoming` and `signals.outgoing` metrics

### LINKED ISSUE #2837: Enable additional stopwatch metrics to track signal consumed and produced
Follow-up to #2747 

In addition to stopwatch duration, an operator may want to measure the overall signal input and output from a range of processors.

This is somewhat similar to the `channel_metrics` `consumed` and `produced`:

https://github.com/open-telemetry/otel-arrow/blob/8b4e674f6909f04cdd6726c42e3ad0f3281b3d6e/rust/otap-dataflow/crates/engine/src/channel_metrics.rs#L77-L122

However, those metrics only track channel messages, not the actual count of signals passing through the channel.

Need to add the following stopwatch metrics:
- `stopwatch.items.consumed` (emitted from stopwatch start node)
- `stopwatch.items.produced` (emitted from stopwatch stop node)

https://github.com/open-telemetry/otel-arrow/blob/8b4e674f6909f04cdd6726c42e3ad0f3281b3d6e/rust/otap-dataflow/crates/engine/src/stopwatch.rs#L31-L44

This will let us build a picture of:
```
100 log signals in
->
transform 1 (filtering)
-> 
transform 2 (doing some renaming)
->
78 log signals out
```
in addition to tracking combined `duration` of both transforms

drewrelmas:
# Change Summary

Follow-up to #2747

Adds two MMSC metrics on the existing `stopwatch` metric set so operators can compare signal volume in vs. out across a stopwatch range, alongside the existing combined compute duration. "Signals" here means individual log records, spans, or metric data points (`OtapPdata::num_items()`).

| Metric | Recorded | Why |
|---|---|---|
| `stopwatch.signals.incoming` | At the start node, **before** `process()` runs | Any filter/drop in the start processor itself does not undercount entry volume. |
| `stopwatch.signals.outgoing` | At the stop node, **after** `process()` completes | Reflects what actually leaves the range. |

Implemented as two metric set types (`StopwatchStartMetrics`, `StopwatchStopMetrics`) sharing one entity per stopwatch — mirrors the `ChannelSenderMetrics` / `ChannelReceiverMetrics` precedent. Each role registers its own `MetricSet` against the same entity and drains its own accumulator on the periodic `CollectTelemetry` tick and at shutdown.

To capture the incoming count pre-process, the existing `ProcessorSendHook` trait is renamed to `FlowMeasurementHook` and gains a default-no-op `after_processor_receive` method. The engine run loops (Local + Shared) call it immediately after `inbox.recv_when(...)` returns a `Message::PData`, before `begin_process_timing` and `process()`. `OtapPdata` overrides it to drive the start-side counter; test PData stand-ins (`()`, `String`, `TestMsg`) get blanket no-op impls.

The two hooks fire from different surfaces by design, matching the asymmetric flow control of a processor:

| Hook | Fires from | Cardinality per `process()` | Captures |
|---|---|---|---|
| `after_processor_receive` | Engine run loop | Exactly 1 (1 dequeue per iteration) | True pre-process input volume |
| `before_processor_send` | Effect handler `send_message[_to]` | 0..N (drop, pass-through, or fan out) | What actually leaves |

**Behavior change:** removed the `PROCESS_DURATION` gate in `build_stopwatch_state`. Stopwatches are already explicit opt-in via the telemetry policy YAML; the gate was redundant and signal counts don't need the timing path. Pipelines with stopwatches under `runtime_metrics: basic`/`none` will now run them instead of silently skipping.

## Demo

`configs/fake-stopwatch-demo.yaml` now includes a 1-in-3 `processor:log_sampling` node inside the stopwatch range so `signals.outgoing` is visibly smaller than `signals.incoming`.

```bash
cargo run --bin df_engine -- --config configs/fake-stopwatch-demo.yaml
curl -s 'http://127.0.0.1:8080/api/v1/telemetry/metrics?format=json' \
  | jq '.metric_sets[] | select(.name == "stopwatch")'
```

Sample output (truncated, after ~38 collection cycles at 10 signals/sec):

```json
{
  "name": "stopwatch.signals.incoming",
  "value": { "min": 10.0, "max": 10.0, "sum": 380.0, "count": 38 }
}
{
  "name": "stopwatch.compute.duration",
  "value": { "min": 2859829.0, "max": 6619768.0, "sum": 170014602.0, "count": 38 }
}
{
  "name": "stopwatch.signals.outgoing",
  "value": { "min": 3.0, "max": 4.0, "sum": 127.0, "count": 38 }
}
```

Reading: 380 signals entered the range (38 batches × 10 signals), 127 left it (≈1/3, matching the sampler ratio), and `compute.duration` averages ~4.47 ms per batch across the chain (170014602 ns / 38). Both signal-count metrics share the same `stopwatch.name` / `start_node` / `stop_node` attributes as the duration metric, so they correlate without joins.

* Related to #2782 
* Closes #2837 

Unit Tests / Local runs

1. Stopwatch duration metric will now be tracked and emitted even on `runtime_metrics: basic/none`.
2. New Stopwatch metrics for `consumed` and `produced`

### DIFF
--- a/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
+++ b/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
@@ -3,7 +3,7 @@
 # Five attribute processors form a sequential processing chain, with a 1-in-3
 # log sampler in the middle so that produced < consumed:
 #
-#   stopwatch "ingest_pipeline" - covers attr1 through attr5 (full chain)
+#   stopwatch "ingest_pipeline" - covers attr1 through attr4 (full chain)
 #
 # Stopwatches sum per-processor wall-clock compute time across each node in the
 # range. The engine arms an Instant send-marker before each `process()` call
@@ -75,7 +75,7 @@ groups:
                   value: "2"
 
           # Drops ~2 of every 3 log records so that items.produced
-          # at attr5 is visibly smaller than items.consumed at attr1.
+          # at attr4 is visibly smaller than items.consumed at attr1.
           sampler:
             type: processor:log_sampling
             config:
@@ -97,8 +97,8 @@ groups:
             config:
               actions:
                 - action: insert
-                  key: pipeline.stage5
-                  value: "5"
+                  key: pipeline.stage4
+                  value: "4"
 
           noop:
             type: exporter:noop
--- a/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
+++ b/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
@@ -19,7 +19,7 @@
 #   - stopwatch.items.produced   (MMSC, items observed at the stop node)
 #
 # With the sampler in place, `items.consumed` (at attr1) reflects the full
-# input volume while `items.produced` (at attr5) reflects roughly 1/3 of it,
+# input volume while `items.produced` (at attr4) reflects roughly 1/3 of it,
 # so the difference is directly visible.
 #
 # Note: only non-overlapping stopwatch ranges are supported. A message can be
--- a/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
+++ b/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
@@ -1,7 +1,7 @@
 # Demonstrates the distributed stopwatch feature.
 #
-# Five attribute processors form a sequential processing chain. A stopwatch
-# measures aggregate synchronous compute duration across the full chain:
+# Five attribute processors form a sequential processing chain, with a 1-in-3
+# log sampler in the middle so that produced < consumed:
 #
 #   stopwatch "ingest_pipeline" - covers attr1 through attr5 (full chain)
 #
@@ -12,8 +12,15 @@
 # captures synchronous compute plus any await/I/O that occurs inside
 # `process()` between sends; it excludes time the message spends queued in
 # inbound channels.
-# Metrics appear as stopwatch.compute.duration (MMSC histograms in
-# nanoseconds).
+#
+# Metrics emitted on the `stopwatch` metric set:
+#   - stopwatch.compute.duration (MMSC histogram, nanoseconds)
+#   - stopwatch.items.consumed   (MMSC, items observed at the start node)
+#   - stopwatch.items.produced   (MMSC, items observed at the stop node)
+#
+# With the sampler in place, `items.consumed` (at attr1) reflects the full
+# input volume while `items.produced` (at attr5) reflects roughly 1/3 of it,
+# so the difference is directly visible.
 #
 # Note: only non-overlapping stopwatch ranges are supported. A message can be
 # inside at most one stopwatch range at a time.
@@ -39,7 +46,7 @@ groups:
             stopwatches:
               - name: ingest_pipeline
                 start_node: attr1
-                stop_node: attr5
+                stop_node: attr4
 
         nodes:
           receiver:
@@ -67,15 +74,17 @@ groups:
                   key: pipeline.stage2
                   value: "2"
 
-          attr3:
-            type: processor:attribute
+          # Drops ~2 of every 3 log records so that items.produced
+          # at attr5 is visibly smaller than items.consumed at attr1.
+          sampler:
+            type: processor:log_sampling
             config:
-              actions:
-                - action: insert
-                  key: pipeline.stage3
-                  value: "3"
+              policy:
+                ratio:
+                  emit: 1
+                  out_of: 3
 
-          attr4:
+          attr3:
             type: processor:attribute
             config:
               actions:
@@ -83,7 +92,7 @@ groups:
                   key: pipeline.stage4
                   value: "4"
 
-          attr5:
+          attr4:
             type: processor:attribute
             config:
               actions:
@@ -101,10 +110,10 @@ groups:
           - from: attr1
             to: attr2
           - from: attr2
+            to: sampler
+          - from: sampler
             to: attr3
           - from: attr3
             to: attr4
           - from: attr4
-            to: attr5
-          - from: attr5
             to: noop
--- a/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
+++ b/rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
@@ -15,11 +15,11 @@
 #
 # Metrics emitted on the `stopwatch` metric set:
 #   - stopwatch.compute.duration (MMSC histogram, nanoseconds)
-#   - stopwatch.items.consumed   (MMSC, items observed at the start node)
-#   - stopwatch.items.produced   (MMSC, items observed at the stop node)
+#   - stopwatch.signals.incoming   (MMSC, items observed at the start node)
+#   - stopwatch.signals.outgoing   (MMSC, items observed at the stop node)
 #
-# With the sampler in place, `items.consumed` (at attr1) reflects the full
-# input volume while `items.produced` (at attr4) reflects roughly 1/3 of it,
+# With the sampler in place, `signals.incoming` (at attr1) reflects the full
+# input volume while `signals.outgoing` (at attr4) reflects roughly 1/3 of it,
 # so the difference is directly visible.
 #
 # Note: only non-overlapping stopwatch ranges are supported. A message can be
@@ -45,7 +45,7 @@ groups:
             runtime_metrics: normal
             stopwatches:
               - name: ingest_pipeline
-                start_node: attr1
+                start_node: sampler
                 stop_node: attr4
 
         nodes:
@@ -57,6 +57,18 @@ groups:
                 signals_per_second: 10
                 log_weight: 100
               registry_path: https://github.com/open-telemetry/semantic-conventions.git[model]
+
+          # Drops ~2 of every 3 log records so that signals.outgoing
+          # at attr4 is visibly smaller than signals.incoming at sampler.
+          # Used as beginning of stopwatch range to prove that signals.incoming
+          # is reported before start node's process() begins.
+          sampler:
+            type: processor:log_sampling
+            config:
+              policy:
+                ratio:
+                  emit: 1
+                  out_of: 3
 
           attr1:
             type: processor:attribute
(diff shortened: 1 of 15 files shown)


rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -711,15 +711,22 @@ fn stopwatch_accumulate<H: StopwatchEffectHandler>(handler: &H, data: &mut OtapP
     if !is_start && !is_stop && !data.has_active_stopwatch() {
         return;
     }
+    let items = data.num_items() as u64;
```
jmacd: This is appropriately guarded by having a stopwatch, but we only need to know it for start and stop nodes, and it's not a free call. I've considered idealized ways to compute and cache the number of items in the PData context so that we can calculate this lazily and avoid repeatedly counting items.

Without such ideals, maybe we can push `data.num_items() as u64` to the two call sites to avoid computing it at every node.

drewrelmas: Great suggestion, done this.

rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
```diff
@@ -1,7 +1,7 @@
 # Demonstrates the distributed stopwatch feature.
 #
-# Five attribute processors form a sequential processing chain. A stopwatch
-# measures aggregate synchronous compute duration across the full chain:
+# Five attribute processors form a sequential processing chain, with a 1-in-3
+# log sampler in the middle so that produced < consumed:
 #
 #   stopwatch "ingest_pipeline" - covers attr1 through attr5 (full chain)
```
cijothomas: nit: only upto attr4 now

rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
```diff
@@ -67,23 +74,25 @@ groups:
                   key: pipeline.stage2
                   value: "2"
 
-          attr3:
-            type: processor:attribute
+          # Drops ~2 of every 3 log records so that items.produced
+          # at attr5 is visibly smaller than items.consumed at attr1.
```
cijothomas:
```suggestion
          # at attr4 is visibly smaller than items.consumed at attr1.
```

rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
```diff
@@ -39,7 +46,7 @@ groups:
             stopwatches:
```
cijothomas, line 46: a nit observation. I missed reviewing the original PR, but stopwatch made sense for that PR. Now that we are not just timing, but also counting, the name feels a bit off..

drewrelmas: That's a great point - I can take a poll at the SIG meeting later today. Options:
- `segment`
- `probe`
- `transit`

Would prefer to do rename in separate PR if we decide on a term.

cijothomas: don't have great suggestions, but consider including something to tell its only applicable for processors.

Just sharing something:
```yaml
telemetry:
  runtime_metrics: normal
  processor_metrics: 
    ranges:                  
      - name: ingest_pipeline
        start_processor(or node): attr1
        end_processor(or node): attr4
```

Not for this PR.

rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
```diff
@@ -12,8 +12,15 @@
 # captures synchronous compute plus any await/I/O that occurs inside
 # `process()` between sends; it excludes time the message spends queued in
 # inbound channels.
-# Metrics appear as stopwatch.compute.duration (MMSC histograms in
-# nanoseconds).
+#
+# Metrics emitted on the `stopwatch` metric set:
+#   - stopwatch.compute.duration (MMSC histogram, nanoseconds)
+#   - stopwatch.items.consumed   (MMSC, items observed at the start node)
```
cijothomas: what do you think of naming them items_incoming and items_outgoing instead?

cijothomas: Operators/users typically "how many items went in vs came out of this segment?", so incoming/outgoing feels more aligned with mental model. (could be my personal preference!)

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -712,14 +712,25 @@ fn stopwatch_accumulate<H: StopwatchEffectHandler>(handler: &H, data: &mut OtapP
         return;
     }
     let delta_ns = handler.take_elapsed_since_send_marker_ns();
+
+    // Note that num_items() is only called when required at stopwatch start and stop.
+    // It is intentionally not called on every node to minimize performance overhead.
     if is_start {
+        let items: u64 = data.num_items() as u64;
```
lalitb: One thing I want to confirm: the PR description says `stopwatch.items.consumed` is `OtapPdata::num_items()` at entry, but this records it from the processor send hook. If the start node filters, splits, batches, or otherwise changes the number of items, this would not be the entry count anymore.  Is that intended, or should the consumed count be captured before the start processor runs?

lalitb (approved):
Thanks.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -712,14 +712,25 @@ fn stopwatch_accumulate<H: StopwatchEffectHandler>(handler: &H, data: &mut OtapP
...
         data.start_stopwatch();
+        if items > 0 {
+            handler.record_stopwatch_start_items(items);
```
utpilla: `items.consumed` is counted here during `send_message`, after the start processor has already processed the data. If the start processor filters or drops items, this reports the post-processing count, not the pre-processing count. For a metric named "consumed" (items entering the range), shouldn't this count be captured before `process()` runs instead. In the engine loop alongside `begin_process_timing()`?

rust/otap-dataflow/crates/engine/src/stopwatch.rs
```diff
@@ -23,24 +24,30 @@ use otap_df_telemetry::instrument::Mmsc;
...
-///
-/// Future extension: this set could also record signal-item counts
-/// at the start and stop nodes — `stopwatch.items.consumed` (Mmsc,
-/// observed at the start node from `OtapPdata::num_items()`) and
-/// `stopwatch.items.produced` (observed at the stop node).
+/// Metric set emitted by the start node of a stopwatch range.
+#[metric_set(name = "stopwatch")]
+#[derive(Debug, Default, Clone)]
+pub struct StopwatchStartMetrics {
+    /// Number of signal items entering the stopwatch range.
```
cijothomas: "items" - could we clarify we mean logrecord/span/metric-points by items by updating the Description?

rust/otap-dataflow/crates/otap/src/pdata.rs
utpilla: nit: Not from this PR, but the `1ns` sentinel means every stopwatch observation is off by exactly `1ns`. This could be fixed by subtracting 1. Not a practical issue at microsecond-scale measurements just a nit for correctness.

```suggestion
            .map(|acc| acc.get() - 1)
```

rust/otap-dataflow/configs/fake-stopwatch-demo.yaml
drewrelmas: I do like `items.incoming` and `items.outgoing`. I had initially chosen produced/consumed to mirror the channel metrics, but I think that isn't semantically correct with what these metrics are measuring. We're not adding/removing data, simply snapshotting it. I will update.

rust/otap-dataflow/crates/otap/src/pdata.rs
drewrelmas: You are correct, the `incoming` count should be taken before processor starts, I forgot to add a parallel `ProcessorReceiveHook`. Same thing that Utkarsh observed in https://github.com/open-telemetry/otel-arrow/pull/2839#discussion_r3190936088. Will update.

rust/otap-dataflow/crates/otap/src/pdata.rs
drewrelmas: Great catch, see https://github.com/open-telemetry/otel-arrow/pull/2839#discussion_r3191139047. Will fix this.

drewrelmas: Pushed, please see https://github.com/open-telemetry/otel-arrow/pull/2839/commits/052a8bc98deb3e5e6fbf1bc9299a164f8ff8b98a. Went with a `FlowMeasurementHook` with both `after_processor_receive()` and `before_processor_send()`.

rust/otap-dataflow/crates/otap/src/pdata.rs
drewrelmas: Pushed, please see https://github.com/open-telemetry/otel-arrow/pull/2839/commits/052a8bc98deb3e5e6fbf1bc9299a164f8ff8b98a. Went with a `FlowMeasurementHook` with both `after_processor_receive()` and `before_processor_send()`.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -720,13 +721,34 @@ fn stopwatch_accumulate<H: StopwatchEffectHandler>(handler: &H, data: &mut OtapP
...
     }
+
+    /// At the stopwatch start node, count items *entering* the range —
+    /// i.e. before `process()` runs and may filter or drop them. This
+    /// gives the true input volume to compare against the stop-node
+    /// output volume recorded in [`stopwatch_accumulate`].
+    fn after_processor_receive<H: StopwatchEffectHandler>(&mut self, handler: &H) {
+        if handler.is_stopwatch_start() {
+            let signals: u64 = self.num_items() as u64;
+            if signals > 0 {
```
ThomsonTan: The `if signals > 0` here (and the same at L747) silently drops empty batches, but `record_stopwatch_stop(total)` above is unconditional. So `signals.outgoing.count` no longer matches `compute.duration.count`, and `sum/count` stops meaning "items per batch" once any filter can produce a 0-item batch — which is exactly the demo's 1-in-3 sampler scenario. It also hides 0-in/0-out traversals, which are a useful starvation/over-filter signal.


```suggestion
        handler.record_stopwatch_stop_signals(data.num_items() as u64);
```

Also probably add a unit test where `num_items() == 0` asserting all three `count`s match.

drewrelmas: Really interesting edge case. I was assuming that the engine would stop forwarding an empty batch, but looking through the code again I think this isn't accurate. I've updated to remove the `if signals > 0` condition.

jmacd (approved):

drewrelmas:
Going to go ahead and merge this PR to get started on the proposed renaming, follow-ups can be done at that time!

PR #2840, opened 2026-05-05
  Add Linux host metrics receiver for OTAP Dataflow

### LINKED ISSUE #2741: Implement Linux v1 OTAP-native host_metrics receiver
### Pre-filing checklist

### Component(s)

### Objective

Implement an OTAP-native `host_metrics` receiver for the OTAP Dataflow Engine.

The v1 goal is a Linux-first, low-overhead host metrics receiver that emits host-level metrics directly as OTAP metric records, supports first-class container host filesystem collection via `root_path`, and avoids known complexity and overhead patterns learned from the Go Collector `hostmetricsreceiver`.

Semantic conventions are a day-one requirement. The receiver must emit metrics aligned with current OpenTelemetry system semantic conventions by default: [Semantic conventions for system metrics](https://opentelemetry.io/docs/specs/semconv/system/system-metrics/)

The receiver should register as `urn:otel:receiver:host_metrics` and live under `crates/core-nodes/src/receivers/host_metrics`.

### Rationale

The existing Go Collector hostmetricsreceiver is feature rich, but user and maintainer feedback points to several recurring design problems:
- generic multi-OS libraries can hide OS-specific semantics and repeated work
- independent scrapers can duplicate /proc, /sys, mount, and process reads
- per-process metrics can create high CPU overhead and high-cardinality output
- root_path behavior has historically been difficult to keep consistent
- scraper errors can be noisy and difficult to attribute
- semantic convention migration is harder when collection and reporting are tightly coupled

The OTAP Dataflow Engine has different constraints and opportunities:
- thread-per-core execution
- local !Send receivers
- direct OTAP Arrow record support
- existing strict typed configuration
- existing receiver telemetry through MetricSet
- topic-based pipeline splitting for singleton collection plus multicore downstream processing

The receiver should therefore be designed as a host-observation subsystem, not as a direct line-by-line port of the Go implementation.

When **engine-level extensions become available**, the host observation service should move behind an engine-scoped shared extension. At that point, one engine-wide host observer can own the host snapshot/cache/eBPF state once per process, while one or more receiver nodes consume from that shared observer and project metrics into their pipelines. Until then, v1 keeps singleton collection safe through topology and startup validation.

### Step 1: v1 Linux Core Host Metrics

Implement the first production-oriented Linux receiver with these properties:
- Local !Send receiver implementing local::Receiver<OtapPdata>.
- Direct OtapArrowRecords::Metrics emission using existing pdata metric record builders.
- Native config shape, not strict Go config parity.
- One shared Linux scrape snapshot per due tick.
- Clear separation between:
   - collection from /proc, /sys, filesystem APIs;
   - normalization/delta calculation;
   - semantic convention metric projection.
- First-class HostRoot / HostView abstraction for root_path.
- Singleton deployment contract for host-wide metrics.
- Receiver-level self-observability.
- Day-one semantic convention alignment for emitted system metrics.

v1 metric families:
- CPU
- memory
- load
- paging/swap
- system uptime
- disk counters
- filesystem usage
- network interface counters
- aggregate process counts only

Example configuration:

```yaml
groups:
  host:
    pipelines:
      collect:
        policies:
          resources:
            core_allocation:
              type: core_count
              count: 1
        nodes:
          host_metrics:
            type: receiver:host_metrics
            config:
              collection_interval: 60s
              initial_delay: 1s
              host_view:
                root_path: /
                validation: fail_selected
              families:
                cpu:
                  interval: 30s
                memory:
                  interval: 30s
                load:
                  interval: 30s
                paging:
                  interval: 60s
                system:
                  interval: 60s
                disk:
                  interval: 60s
                filesystem:
                  interval: 5m
                  include_virtual_filesystems: false
                network:
                  interval: 60s
                  include_connection_count: false
                processes:
                  interval: 60s
                  mode: summary
          publish:
            type: exporter:topic
            config:
              topic: host_metrics
        connections:
          - from: host_metrics
            to: publish
```

Unlike the Go Collector pattern, users should not need to run multiple hostmetrics receiver instances just to collect different metric families at different frequencies. This receiver supports per-family intervals inside one singleton receiver. The receiver scheduler groups all families due at the same tick, builds one shared host snapshot for that due set, and emits a metrics batch without duplicating host-wide collection.

Explicitly out of scope for v1:
- per-PID process metrics;
- top-k or grouped process metrics;
- eBPF;
- io_uring;
- NFS;
- network connection counts;
- macOS support;
- Windows support;
- entity-event logs.

### Step 2: v2 Linux Advanced Capabilities

Track separately in a follow-up issue.

Potential scope:
- per-process metrics as an explicit opt-in capability
- top-k process reporting by CPU, RSS, or I/O
- grouped process aggregates by executable, cgroup, service, or container
- optional io_uring for high-volume procfs process enrichment
- optional eBPF provider for process lifecycle, network connection state, block I/O latency, or attribution-heavy signals
- procfs fallback when eBPF or io_uring is unavailable

### Step 3: v2 OS Portability

Track separately in a follow-up issue.

Potential scope:
- macOS native collectors using platform APIs such as sysctl, libproc, getfsstat, and statfs
- Windows native collectors using PDH, IP Helper APIs, and targeted Windows system APIs
- per-OS capability detection
- semantic parity tests across supported platforms
- no generic multi-OS crate on the default hot path unless benchmarks and semantic tests justify it

### v1 Functional Criteria

- host_metrics receiver is registered as `urn:otel:receiver:host_metrics`.
- Receiver can be configured in an OTAP Dataflow pipeline and emits metrics downstream.
- Receiver emits direct OTAP metric records, not prost OTLP metric object graphs on the steady path.
- Linux root_path is applied consistently through a single HostRoot / HostView abstraction.
- Enabled metric families read only the required Linux sources.
- Multiple metric families sharing a source do not independently reread the same source in one scrape tick.
- Per-family intervals are supported so users can collect different metric families at different frequencies without deploying duplicate receivers.
- Aggregate process counts do not emit per-PID series.
- Errors include the metric family and low-cardinality error class.
- Expected partial errors are counted and rate-limited rather than logged noisily every interval.

### v1 Deployment Criteria

- The receiver fails fast by default when configured in a pipeline allocated to more than one core.
- The error message explains that host-wide collection must run in a one-core source pipeline.
- A recommended topology is documented:
  - one-core host_metrics collection pipeline;
  - topic exporter;
  - multicore downstream topic receiver pipeline for batching/export.
- Duplicate host collection inside one engine process is guarded by a startup-only lease keyed by normalized host view / root_path.
- The design documents the future migration path to an engine-scoped host observation extension once engine-level extensions are available.

### v1 Semantic Criteria

- Emitted metrics use current **OpenTelemetry system semantic conventions** by default: https://opentelemetry.io/docs/specs/semconv/system/system-metrics/
- Resource attributes include stable host identity where available.
- Metric names, units, temporality, and attributes are covered by tests.
- Known Go receiver behavior bugs do not need to be preserved; migration deltas are documented.

### v1 Performance Criteria

- Benchmarks compare v1 Linux families against the Go Collector hostmetricsreceiver on the same host.
- Acceptance is based on lower CPU/allocation trend than the Go receiver for equivalent families.
- No unbounded RSS growth during repeated scrapes.
- The implementation avoids generic multi-OS system metrics crates on the Linux hot path.

### v1 Test Criteria

- Unit tests cover Linux parsers using fixture data.
- Unit tests cover scheduler behavior and per-family intervals.
- Unit tests cover HostRoot path resolution and root-path validation.
- Receiver harness tests cover initial delay, scrape tick, control priority, drain, shutdown, and telemetry reporting.
- Semantic tests decode emitted OTAP metrics and verify expected metric shape.
- Config tests verify strict unknown-field rejection and invalid configuration errors.

### Dependencies or Blockers

- Existing OTAP metric record builders must be sufficient for direct metric emission. If gaps are found, add minimal builder support before implementing receiver projection.
- The receiver should use existing MetricSet infrastructure for self-observability.
- The singleton deployment guard may require a small process-local registry or lease mechanism.
- The recommended one-core collection plus topic fan-out topology depends on existing topic receiver/exporter behavior.
- Engine-level extensions are not required for v1, but should become the preferred home for the shared host observation service once available.

### Additional Context

Design choices for v1:

- Prefer custom Linux /proc and /sys readers over generic multi-OS crates on the hot path.
- Keep collection, normalization, and emission separate from day one.
- Use direct OTAP Arrow metrics emission for lower allocation overhead.
- Keep v1 cardinality bounded by excluding per-PID metrics.
- Support per-family intervals inside one singleton receiver instead of requiring multiple receiver instances for different scrape frequencies.
- Treat eBPF and io_uring as future optional providers, not v1 requirements.
- Treat Windows and macOS as separate future work, because those implementations require different APIs, tests, and capability models.
- Plan for a future engine-scoped host observation extension to centralize host collection and allow multiple pipelines to consume shared host snapshots safely.

Recommended follow-up issues:

1. host_metrics v2: Linux advanced process, eBPF, and io_uring capabilities
2. host_metrics v2: macOS and Windows native collector support

lalitb:
# Change Summary

  Adds a Linux host metrics receiver for OTAP Dataflow that collects host-level `system.*` metrics from procfs/sysfs and emits OTAP Arrow metrics directly.

  Highlights:

  - Implements the Phase 1 host metric families with current semconv coverage: CPU, memory, paging/swap, system uptime, disk, filesystem, network, and aggregate process summary.
  - Emits metrics using OpenTelemetry semantic conventions pinned to schema `1.41.0`.
  - Centralizes semconv metric/attribute constants and adds a semconv drift check.
  - Builds OTAP Arrow records directly without constructing intermediate OTLP/protobuf metric objects.
  - Uses OTAP metric rows efficiently by grouping device/interface datapoints under shared metric handles.
  - Supports per-family intervals inside one singleton receiver.
  - Enforces the one-core host collection model and duplicate receiver lease guard.
  - Supports host root views for container/DaemonSet deployments, including host network namespace handling via `/proc/1/net/dev`.
  - Keeps `system.process.count` limited to registered `process.state` values; Linux `procs_blocked` is intentionally not emitted until semconv defines a matching state.

  Notes:

  - `load` is intentionally not emitted in this PR because current OpenTelemetry system semantic conventions do not define a stable system load metric.
  - Receiver self-observability is implemented with the current `MetricSet` support. Per-family / error-class labelled internal telemetry is a follow-up because the internal telemetry API does not currently support attributes on individual metric observations.

* Closes #2741

  - Rust unit/config/projection tests
  - Semconv drift check against OpenTelemetry semantic-conventions `v1.41.0`
  - df_engine runtime validation on Ubuntu
  - End-to-end validation: host metrics receiver → OTLP exporter → OpenTelemetry Collector → Prometheus → Grafana

Yes, a receiver.

Screenshot 2026-05-04 at 9 48 05 PM
Screenshot 2026-05-04 at 9 48 20 PM


rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/otap_builder.rs
```diff
@@ -0,0 +1,348 @@
...
+        let id = self.curr_metric_id;
+        self.metrics.append_id(id);
+        self.metrics.append_metric_type(metric_type as u8);
+        self.metrics.append_name(name.as_bytes());
+        self.metrics.append_description(&[]);
+        self.metrics.append_unit(unit.as_bytes());
+        match metric_type {
+            MetricType::Sum => {
+                self.metrics
+                    .append_aggregation_temporality(Some(AGGREGATION_TEMPORALITY_CUMULATIVE));
```
cijothomas, line 159: wondering why we hard code cumulative temporality here?

lalitb: Yes, this is intentional. For host metrics, the values we read from procfs/sysfs are absolute cumulative values or current totals, not per-scrape deltas.  If we add a genuinely delta source later, we can add a separate delta path then.

.github/workflows/rust-ci.yml
```diff
@@ -656,6 +656,32 @@ jobs:
...
+        with:
+          workspaces: ./rust/otap-dataflow
+      - name: Run host metrics semconv drift check
+        env:
+          OTAP_HOST_METRICS_SEMCONV_REGISTRY: ${{ github.workspace }}/semantic-conventions/model
+        run: |
+          cargo test -p otap-df-core-nodes \
+            --features dev-tools,otap-df-otap/crypto-ring \
+            emitted_phase1_metric_shapes_match_weaver_semconv --lib -- --ignored
+        working-directory: ./rust/otap-dataflow
```
lquerel, line 683: This can be done in a future PR, but I think this kind of check should be possible with Weaver.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/mod.rs
```diff
@@ -0,0 +1,1758 @@
...
+}
+
+fn default_root_path() -> PathBuf {
+    PathBuf::from("/")
+}
+
+/// Configuration for the host metrics receiver.
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct Config {
```
lquerel: Since the configuration for this receiver is fairly extensive, I suggest moving this configuration related code into a dedicated `config.rs` file.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/mod.rs
```diff
@@ -0,0 +1,1758 @@
...
+        }
+        let regexes = if match_type == MatchType::Regexp {
+            let mut regexes = Vec::with_capacity(values.len());
+            for value in &values {
+                regexes.push(Regex::new(value).map_err(|err| {
+                    otap_df_config::error::Error::InvalidUserConfig {
+                        error: format!("invalid host metrics regexp filter {value:?}: {err}"),
+                    }
+                })?);
+            }
```
lquerel: This can absolutely be done in another PR, but I think we should use a `RegexSet` here to optimize pattern matching.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/mod.rs
```diff
@@ -0,0 +1,1758 @@
...
+        })?;
+        RuntimeConfig::try_from(config).map(|_| ())
+    },
+};
+
+#[cfg(target_os = "linux")]
+impl HostMetricsReceiver {
+    /// Creates a new host metrics receiver.
+    pub fn new(config: Config) -> Result<Self, otap_df_config::error::Error> {
+        validate_supported_platform()?;
```
lquerel: How can we execute this code if the entire implementation is guarded by `#[cfg(target_os = "linux")]`?

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/otap_builder.rs
```diff
@@ -0,0 +1,348 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Direct OTAP Arrow record construction for host metrics.
```
lquerel, line 4: @albertlockett or @JakeDern could you take a look at this part of the PR. Thanks

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/procfs.rs
```diff
@@ -0,0 +1,3796 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Linux procfs-backed host metric source.
```
lquerel: For maintainability reasons, it would be good to have a `procfs` module with submodules, to avoid ending up with one giant file.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/procfs.rs
```diff
@@ -0,0 +1,3796 @@
...
+            counter_tracker: CounterTracker::default(),
+            boot_time_unix_nano: None,
+            resource: None,
+        };
+        source.apply_startup_validation()?;
+        Ok(source)
+    }
+
+    /// Collects one host snapshot for the due family set.
+    pub fn scrape_due(&mut self, due: ProcfsFamilies) -> io::Result<HostScrape> {
```
lquerel: What worries me a little is that this method is not async. It may not be a problem in this specific case, but we should make sure that:
1. it cannot block, and
2. the time spent in this method is not too long or dependent on the system configuration.

lquerel (commented):
It is really great to see this critical receiver getting close to being integrated natively into our engine.

My 3 main pieces of feedback are:
- scraping `procfs`, which has the potential to block the async runtime for too long if there are many network interfaces, disks, CPUs, and so on (one slow or hung mount could potentially leave that thread blocked indefinitely).
- observe_key inserts and updates per series counter state, but nothing ever removes entries for devices or interfaces that disappear. On long running nodes with churny veth, loop, or ephemeral block devices, states grows monotonically, which violates the bounded resource requirement and gradually increases hash/allocation cost on every scrape. 
- the size of the files, which makes maintainability problematic.

.github/workflows/rust-ci.yml
lalitb: Agree. This test already uses the Weaver crates to resolve the semconv registry, but it still runs through a Rust fixture because we are checking the actual OTLP emitted by the receiver, not only static metadata. Moving this to a more direct Weaver/CLI-based validation sounds good for a follow-up PR.

jmacd (approved):
The host metrics receiver is one of the high-value components in the Collector, and it has a high degree of configurability. It has many "Scrapers" defined, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/scraping-receivers.md as an overview.

See an example [metadata.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/memoryscraper/metadata.yaml) where *generated* instruments are defined, which determines compile-time views much like in my PR #2623 ans gives operators control over attribute dimensions. Like @lquerel's comment, I don't want to block this PR. We should consider it unstable and aim to converge with the Collector's hostmetrics receiver, which is close to stable.

jmacd:
I think it will help (maybe now) to replace "Family" with "Scraper" as a nod to Collector's terminology.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/mod.rs
lalitb: Done. I moved the receiver configuration and runtime config conversion into a dedicated config.rs file, and kept the public config types re-exported from mod.rs so the external surface stays the same.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/mod.rs
lalitb: Done in this PR. I switched the compiled regexp filter to use RegexSet, so we compile one set per filter and use a single set match during scraping.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/mod.rs
lalitb: Good catch. I cleaned this up so the factory entry points are available on all platforms, while the Linux receiver implementation stays Linux-only. On non-Linux, create/validate now return the unsupported-platform config error directly instead of relying on a Linux-only constructor path.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/procfs.rs
lalitb: Done. I split `procfs` into a module with smaller submodules:
- `paths` for host path resolution
- `readings` for `procfs`/`sysfs` parsing and `statvfs`
- `projection` for building the emitted OTAP metric data
-  tests

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/procfs.rs
lalitb: I looked into this with the df-engine threading model in mind. I kept the `procfs`/`sysfs` reads synchronous because they are short kernel virtual-file reads served from in-memory kernel state. Routing them through
`tokio::fs` would mostly hand the same reads to `Tokio`’s global blocking pool, adding overhead without giving us a stronger bound, and weakening core locality.

The part that can really block is `statvfs` on remote/userspace mounts. I tightened that path in this PR: `statvfs` is isolated behind a dedicated bounded worker thread using `sync_channel(1)`, each mount has a timeout, the filesystem scrape has a total budget, and remote/FUSE filesystems are skipped by default unless explicitly opted in via `include_remote_filesystems`.

 Does this direction look reasonable to you?

lalitb:
> It is really great to see this critical receiver getting close to being integrated natively into our engine.
> 
> My 3 main pieces of feedback are:
> 
> * scraping `procfs`, which has the potential to block the async runtime for too long if there are many network interfaces, disks, CPUs, and so on (one slow or hung mount could potentially leave that thread blocked indefinitely).
> * observe_key inserts and updates per series counter state, but nothing ever removes entries for devices or interfaces that disappear. On long running nodes with churny veth, loop, or ephemeral block devices, states grows monotonically, which violates the bounded resource requirement and gradually increases hash/allocation cost on every scrape.
> * the size of the files, which makes maintainability problematic.

Thanks Laurent, this feedback was very helpful.

  I addressed these in the latest commits:

  - Split the large config/procfs code into smaller modules for maintainability.
  - Added pruning for disappeared disk/network counter series, while preserving state across transient read failures.
  - Kept `procfs`/`sysfs` reads synchronous because they are short kernel virtual-file reads served from in-memory kernel state.
  - Tightened the risky `statvfs` path: it now runs behind a dedicated bounded worker, has per-mount timeouts, a total filesystem scrape budget, and remote/FUSE filesystems are skipped by default unless explicitly
  enabled.

  Please let me know if this direction looks reasonable to you.

lalitb:
> The host metrics receiver is one of the high-value components in the Collector, and it has a high degree of configurability. It has many "Scrapers" defined, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/scraping-receivers.md as an overview.
> 
> See an example [metadata.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/memoryscraper/metadata.yaml) where _generated_ instruments are defined, which determines compile-time views much like in my PR #2623 ans gives operators control over attribute dimensions. Like @lquerel's comment, I don't want to block this PR. We should consider it unstable and aim to converge with the Collector's hostmetrics receiver, which is close to stable.

Thanks @jmacd , that makes sense. This PR is intended to cover the Step 1 / v1 Linux core scope from #2741.

I agree we should keep evolving this incrementally and continue converging with the Collector hostmetrics receiver where it makes sense. The Collector-style `metadata.yaml` / generated-instrument model and operator-controlled attribute dimensions seem like good follow-up work, since that is broader than the v1 scope and probably needs its own design pass.

We can align the long-term config and emitted metric surface there without blocking this first native integration.

lalitb:
> I think it will help (maybe now) to replace "Family" with "Scraper" as a nod to Collector's terminology.

Great point @jmacd  - That makes sense as a convergence point. We kept `family` in this PR because it follows the v1 scheduler/config terminology from #2741, and renaming it now would add churn without changing behavior.

`scraper` is closer to Collector terminology, so we can consider that as part of the follow-up alignment work around the Collector hostmetrics model ?

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/procfs/mod.rs
```diff
@@ -0,0 +1,544 @@
...
+        {
+            Ok(Some(proc_stat)) => parse_stat(proc_stat, clk_tck),
+            Ok(None) => StatSnapshot::default(),
+            Err(err) => {
+                record_partial_error(&mut partial_errors, &mut first_error, err);
+                StatSnapshot::default()
+            }
+        };
+        if stat.boot_time_unix_nano != 0 {
+            self.boot_time_unix_nano = Some(stat.boot_time_unix_nano);
```
ThomsonTan, line 182: It falls back to `now_unix_nano()` for `start_time_unix_nano` whenever `/proc/stat` cannot be read, but `validate_selected_paths` only opens `/proc/stat` when cpu/system/processes are enabled. So a host root with valid `meminfo`/`diskstats`/`net/dev` but no readable `stat` passes validation and then emits cumulative metrics whose `start_time` shifts every scrape, looking like a reset on every interval. Suggest validating `/proc/stat` whenever any cumulative family is enabled, and capturing a stable receiver-start fallback once for `validation: none`.

lalitb: Good catch. I fixed this in the latest commit:
- `/proc/stat` is now validated whenever a cumulative family is enabled.
- The fallback start time is captured once when the receiver is created, instead of using per-scrape `now`.
-  Added tests for both validation and stable fallback behavior.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/procfs.rs
lalitb: Further discussed with @utpilla as well - And also added cooperative yield points around the heavier scrape phases, so a large scrape does not run as one uninterrupted task on the current-thread runtime. So the direction is -  keep short procfs/sysfs reads synchronous, bound the risky `statvfs` path, and yield between larger scrape phases.

lquerel: Right now the synchronous scrape approach is probably completely fine for v1, especially since:
* process metrics are currently limited to aggregate host summaries instead of full per-process scraping
* remote/userspace filesystem collection is opt-in
* `statvfs` calls already have timeout protection
* collection is centralized into a shared scrape path instead of multiple independent scrapers repeatedly traversing procfs/sysfs

So I don’t think this PR should necessarily change direction immediately.

That said, I suspect we will eventually want to move toward a model where:
* the runtime thread remains responsible for scheduling, control, projection, and downstream flow control
* while the procfs/sysfs/statvfs traversal itself executes inside a dedicated bounded blocking subsystem

Something roughly like:
* runtime schedules scrape tick
* dedicated scraper worker builds a `HostSnapshot`
* runtime projects/emits OTAP metrics

The idea would be to keep the existing one-core collection model while isolating potentially long-running synchronous host inspection work behind explicit resource and latency boundaries:
* max 1 in-flight scrape
* bounded channels
* scrape deadlines/budgets
* overrun policies
* stronger self-observability around scrape latency/cost

lquerel:
> The host metrics receiver is one of the high-value components in the Collector, and it has a high degree of configurability. It has many "Scrapers" defined, see https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/scraping-receivers.md as an overview.
> 
> See an example [metadata.yaml](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/memoryscraper/metadata.yaml) where _generated_ instruments are defined, which determines compile-time views much like in my PR #2623 ans gives operators control over attribute dimensions. Like @lquerel's comment, I don't want to block this PR. We should consider it unstable and aim to converge with the Collector's hostmetrics receiver, which is close to stable.

@jmacd, we tried to take into account the extensive feedback that was given on the Go version when designing this new host metrics receiver.

For example, instead of having many largely independent scrapers repeatedly traversing `procfs` / `sysfs`, we have a single scraper that inspects `procfs` / `sysfs` in one pass. That is significantly more efficient. This is why the concept of a "family" was introduced: **1 scraper -> n families**.

We also chose to support the host-related semantic conventions directly, and more generally I think we are trying to adopt the semantic conventions + Weaver ecosystem as natively as possible. So I am not sure that the `metadata.yaml` direction is the one we want to pursue.

lquerel:
> I think it will help (maybe now) to replace "Family" with "Scraper" as a nod to Collector's terminology.

I disagree with this one. See my previous comment.

lquerel (approved):
I am aligned with this first version. The proposal I made in that [comment](https://github.com/open-telemetry/otel-arrow/pull/2840#discussion_r3203092771) can absolutely be explored in a future PR.

PR #2843, opened 2026-05-05
ci(codecov): wait for all uploads; use auto project target; raise patch to 80%

### LINKED ISSUE #2591: The build is always red
### Pre-filing checklist

### Component(s)

CI / Build

### Description

codecov/patch is often showing as broken
pipeline-perf-check sometimes times out

Image

### Additional Context

(It's bad!)

jmacd:
Codecov was reporting a failing project status (e.g. 56% vs target 85%) before all test jobs had even finished, because it computed the status from the first upload that arrived. We upload coverage from three jobs: rust-ci 'test' (Linux x86_64), rust-ci 'coverage', and go-ci 'test_and_coverage'. Set codecov.notify.after_n_builds: 3 (with wait_for_ci: true) so the status is computed only after all three reports have been received.

Also relax the thresholds:
- project: switch to 'target: auto' with a 1% threshold so PRs only fail if they meaningfully decrease overall coverage, rather than needing the whole repo to hit a fixed 85%.
- patch: raise to 80% (from 70%) for changed lines in a PR.

Fixes #2591

### DIFF
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -1,11 +1,22 @@
+codecov:
+  notify:
+    # Wait for all coverage uploads before posting status. We currently
+    # upload from three jobs: rust-ci `test` (Linux x86_64), rust-ci
+    # `coverage`, and go-ci `test_and_coverage`. Without this, codecov
+    # posts a (failing) partial status as soon as the first upload
+    # arrives, before the other test jobs have even finished.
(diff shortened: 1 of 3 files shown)


drewrelmas (approved):
I didn't know that `after_n_builds` was an option! Looks great!

albertlockett (approved):

PR #2846, opened 2026-05-05
chore: Rename `stopwatches` to `flow_metrics` to better match supported metrics

### LINKED ISSUE #2845: Rename `stopwatches` to `flow_metrics`
After #2839, the term `stopwatch` doesn't fully capture the provided behavior (which includes counting signal items).

As suggested in Slack [on this thread](https://cloud-native.slack.com/archives/C08RRSJR7FD/p1778007437437759), move to a configuration model like the following:
```yaml
flow_metrics:
  - name: transform_path
    bounds:
      start_node: processor1
      end_node: processorN
    metrics:
      - duration
      - signals_incoming
      - signals_outgoing
```

This config structure is flexible for additional improvements:
- Allowing declaration of bounds using node labels (instead of node name)
- Easily extendable for `messages_incoming` and `messages_outgoing` metrics

drewrelmas:
# Change Summary

## Motivation

The original `stopwatch` feature measured aggregate per-message compute duration across a contiguous range of processor nodes. As item-count metrics were added in #2839 at the start and end of each range, the "stopwatch" name became misleading — the feature now records three distinct measurements over a flow. The rename reframes the feature around the *flow* it observes and lets operators opt into individual measurements.

## What changed

- Terminology: "stopwatch" → "flow_metrics"; `stop_node` → `end_node`.
- Config: `telemetry.stopwatches` → `telemetry.flow_metrics`, with `start_node`/`end_node` nested under `bounds`, and a new optional `metrics` selector (`compute_duration`, `signals_incoming`, `signals_outgoing`). Omitting `metrics` enables all three.
- Metrics emitted on the `flow` metric set:
  - `flow.compute.duration` (ns)
  - `flow.signals.incoming` (items, at the start node)
  - `flow.signals.outgoing` (items, at the end node)

Runtime measurement semantics are unchanged: per-message wall-clock time inside `process()` is accumulated via the `Instant` send-marker advanced on every `send_message`.

## Example

```yaml
telemetry:
  flow_metrics:
    - name: ingest_pipeline
      bounds:
        start_node: sampler
        end_node: attr4
      # optional; omit to enable all
      metrics: [compute_duration, signals_incoming, signals_outgoing]
```

This config structure is flexible for additional improvements:

- Allowing declaration of bounds using node labels (instead of node name)
- Easily extendable for `messages_incoming` and `messages_outgoing` metrics

* Closes #2845

Unit test / local runs

Yes, config contract for `stopwatches` becomes `flow_metrics`


rust/otap-dataflow/crates/config/src/policy.rs
```diff
@@ -266,28 +270,82 @@ pub struct TelemetryPolicy {
     /// shared control-plane telemetry.
     #[serde(default = "default_metric_level_basic")]
     pub runtime_metrics: MetricLevel,
-    /// Distributed stopwatches that sum per-message compute duration across
+    /// Distributed flow_measurements that sum per-message compute duration across
     /// a range of processor nodes.
     #[serde(default)]
-    pub stopwatches: Vec<StopwatchConfig>,
+    pub flow_measurements: Vec<FlowMeasurementConfig>,
```
cijothomas: rest of the settings are all "_metrics", so should we use same ending for flow also?

drewrelmas: Agree, will rename in this manner

jmacd (approved):

Issue #2818, opened 2026-05-04
[Epic] OPL/OTAP query engine scalar functions


albertlockett:
**Help wanted:** help us fill in the standard-library of functions for OPL / OTAP query-engine. Please comment on a sub-issue if you'd like to work on it. For questions and help, comment on the issue or reach out in the #otel-arrow-dev channel in [CNCF Slack](https://www.cncf.io/case-studies/slack/)

**WIP:** I will continue to document the list of functions we want to support in sub-issues

---

There are many ways that users may wish to implement transformations on values in their OPL programs. Many of these transformations can be expressed as function calls. 

The intention of this epic is to list the set of functions for which we'd like to add support in OPL and the OTAP query-engine. 

The selection is mostly driven by [what is available in OTTL](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/ottl/ottlfuncs#converters) and can be augmented by what we feel may be generally useful. _Please feel free to comment if there's a function you would like to see supported!_

### How to add function support

This section will give instructions for how to add support for a new function in OPL and OTAP query-engine. The procedure is mostly prescriptive -- if you implementing one of the subtasks, please follow these instructions, but also look at the subtask for any additional commentary. Also, check back on this issue before taking on new tasks, as the instructions may change as the engine evolves.

#### Background:

The general flow here is:
- 1. OPL Parser converts the stringified query into an AST of expressions. The expressions are defined in the [expressions crate](https://github.com/open-telemetry/otel-arrow/tree/main/rust/experimental/query_engine/expressions) and for function calls in particular we are basically always going to be parsing into a [`ScalarExpression`](https://github.com/open-telemetry/otel-arrow/blob/main/rust/experimental/query_engine/expressions/src/scalars/scalar_expressions.rs)
- 2. The query-engine's planner modules are responsible for converting the AST into a query plan. The main entrypoint for planning scalar expressions specifically is [`ExprLogicalPlanner::plan_scalar_expr`](https://github.com/open-telemetry/otel-arrow/blob/34c171b819eb69c1fce176e54072729c747b0cd0/rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs#L258-L400).
- 3. The planned expressions goes on and gets executed by some `PipelineStage` 

Generally for supporting a function call, we may need to make changes in either the parser, or the planner, or both. 

At a high level there are two basic patterns:
[...]

PR #2853, opened 2026-05-06
feat(query-engine): add uuid() and uuidv7() scalar functions

### LINKED ISSUE #2833: [OPL/OTAP query-engine functions] uuid generation
Add support for generating a unique UUID per row in OPL / OTAP query-engine.
```
logs | set attributes["my.log.id"] = uuid() // v4 uuid
logs | set attributes["my.log.id"] =  uuidv7()
```
For general implementation instructions see https://github.com/open-telemetry/otel-arrow/issues/2818

The UDFs we should use are:
https://datafusion.apache.org/user-guide/sql/scalar_functions.html#uuid

We'll need to create a custom scalar UDF to support UUID v7. For inspiration, see the custom scalar UDFs that have already been created:
https://github.com/open-telemetry/otel-arrow/tree/main/rust/otap-dataflow/crates/query-engine/src/pipeline/functions
See also the implementation of the v4 uuid function:
https://github.com/apache/datafusion/blob/main/datafusion/functions/src/string/uuid.rs

This should be parsed into an InvokeFunctionExpr as no special AST expression specifically exists for this capability.

nicknikolakakis:
# Change Summary

Add two zero-arg scalar functions in the OTAP query-engine, per #2833:

- `uuid()`: UUID v4 string per row, backed by DataFusion's built-in [`string::uuid`](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#uuid) UDF.
- `uuidv7()`: UUID v7 string per row, implemented as a custom volatile `ScalarUDF` (DataFusion has no v7 equivalent) using `uuid::Uuid::now_v7`.

```kql
logs | extend attributes["my.log.id"] = uuid()
logs | extend attributes["my.log.id"] = uuidv7()
```

## Zero-arg function support

`uuid()` / `uuidv7()` are the first 0-arg functions in the engine, so this also lifts the prior `"Only functions with one or more arguments currently supported"` guard in `ExprLogicalPlanner::plan_function_invocation`.

Zero-arg invocations now plan with `DataScope::Root` (instead of `DataScope::StaticScalar`). This is important for volatile UDFs: `Volatility::Volatile` UDFs like `uuid()` produce one new value per row in the input batch, but evaluating them in `StaticScalar` scope would feed them the empty placeholder batch (or collapse the result into a `ColumnarValue::Scalar` that gets broadcast to every row, so all rows would get the same UUID). Routing through `Root` makes the function evaluate against the root batch and produce per-row distinct UUIDs.

This pattern should generalize to other zero-arg volatile functions (e.g. `now()` mentioned in the original TODO) and to non-volatile zero-arg functions, since broadcasting an identical scalar across `num_rows` rows is equivalent to producing `num_rows` identical values.

## Files

- `consts.rs`, `parser.rs`: register `uuid` and `uuidv7` as external functions with zero param placeholders.
- `pipeline/expr.rs`: import DataFusion's `uuid()` and the new `uuidv7()` UDF, add them to `DataFusionFunctionDef::from_func_name`, and handle the zero-arg path in `plan_function_invocation`.
- `pipeline/functions/uuidv7.rs` (new): custom `UuidV7Func` mirroring DataFusion's `UuidFunc` shape but using `Uuid::now_v7`.
- `pipeline/functions.rs`: wire the new module via `make_udf_function!`.
- `crates/query-engine/Cargo.toml`: depend on the workspace `uuid` crate (already pulled in transitively, just needs to be declared).

* Closes https://github.com/open-telemetry/otel-arrow/issues/2833
* Part of #2818

Three layers of tests, all in this PR:

- `pipeline::functions::uuidv7` UDF tests: verify per-row uniqueness, zero-row case, argument rejection, and that produced UUIDs parse as v7.
- `pipeline::expr` planner tests (`test_function_invocation_uuid_v4`, `test_function_invocation_uuid_v7`): drive `ExprLogicalPlanner::plan_scalar_expr` with a 0-arg `InvokeFunctionScalarExpression`, execute against a 3-log OTAP batch, and check the resulting `StringArray` contains 3 distinct UUIDs of the expected version.
- `pipeline::assign` end-to-end tests (`test_set_attr_to_uuid_v{4,7}_function_call_result_{opl,kql}_parser`): run `logs | extend attributes["my.log.id"] = uuid()` (and `uuidv7()`) through both OPL and KQL parsers and the full pipeline, assert each log row receives its own distinct UUID of the right version.

`cargo xtask check` passes locally for the affected crates. There is one unrelated network-dependent failure in `otap-df-core-nodes` (`fake_data_generator` tests cloning `open-telemetry/semantic-conventions` over Git) that occurs on my machine regardless of these changes.

Yes: users of the transform processor / query-engine can now invoke `uuid()` and `uuidv7()` in OPL/KQL programs, primarily for assigning a unique identifier per record:

```kql
logs | extend attributes["my.log.id"] = uuid()
logs | extend attributes["trace.log.id"] = uuidv7()
```


albertlockett (approved):
Thanks, this is great @nicknikolakakis !

I had forgotten we lacked the support for zero-arg funcs, so thanks for adding this. nice set of tests as well

Issue #2748, opened 2026-04-23
Admin Prometheus text formatter does not follow OTel OTLP-to-Prometheus translation rules


cijothomas:
## Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

## Description

The custom Prometheus text formatter in the admin crate (`format_prometheus_text` in `crates/admin/src/telemetry.rs`) produces valid Prometheus text exposition format, but does not follow the [OTel OTLP-to-Prometheus translation rules](https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#otlp-metric-points-to-prometheus). Currently missing:

- **`_total` suffix for counters** (e.g., `successful_rows` should be `successful_rows_total`)
- **Unit suffixes** (e.g., `_bytes`, `_seconds`)
- **`otel_scope_name` / `otel_scope_version` labels** (uses a custom `set="..."` label instead)
- **`target_info` metric** for resource attributes

The [OTLP-to-Prometheus mapping spec](https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#otlp-metric-points-to-prometheus) is currently being stabilized, so this is a good time to align the formatter before the rules are finalized.

## Steps to Reproduce

1. Run the engine with any config
2. `curl http://127.0.0.1:8080/api/v1/telemetry/metrics`
3. Compare output against the [OTLP Metric points to Prometheus](https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#otlp-metric-points-to-prometheus) spec

## Expected Behavior

Prometheus text output follows OTel OTLP-to-Prometheus translation rules so standard Prometheus tooling and dashboards work as expected.

## Actual Behavior

Output is valid Prometheus text format but does not apply the OTel-specified name translations or label conventions.

cijothomas:
@pritishnahar95 has volunteered to do this!

pritishnahar95:
@drewrelmas can you please assign this issue to me?

PR #2857, opened 2026-05-06
fix(admin): Prometheus text formatter OTel spec compliance (#2748)

pritishnahar95:
# Change Summary

Brings the admin Prometheus text formatter into compliance with the OTel
OTLP-to-Prometheus translation spec. Output now matches what a real Prometheus
scraper expects: counters carry the `_total` suffix, units are translated from
UCUM to Prometheus base names and surfaced via `# UNIT` metadata, instrumentation
scope is exposed through the standard `otel_scope_name` / `otel_scope_version`
labels (replacing the ad-hoc `set=` label), and resource attributes are exposed
as a `target_info` gauge. Attribute ordering is now deterministic, helpers are
shared between the live (`format_prometheus_text`) and aggregated
(`agg_prometheus_text`) renderers, and resource attributes are threaded through
`AppState` (inert at the controller for now — passed as `HashMap::new()` until
the controller has resource attributes to hand in).

Highlights:

- Add `_total` suffix for counters (with `has_total_suffix` guard so we never
  emit `_total_total`).
- Add UCUM → Prometheus unit suffixes covering simple units, compound rates
  (e.g. `By/s` → `bytes_per_second`), and `{...}` annotation stripping.
- Replace the custom `set=` label with `otel_scope_name` /
  `otel_scope_version`.
- Emit a `target_info` gauge built from resource attributes.
- Emit `# UNIT` metadata lines next to `# HELP` / `# TYPE`.
- Sort attribute keys for deterministic output.
- Share helpers between `format_prometheus_text` and `agg_prometheus_text` so
  the two renderers cannot drift.
- Thread `resource_attributes` through `AppState` (inert: `HashMap::new()` at
  the controller call site, with a `TODO(#2748)` to wire real values in).

Addresses, but does not close PR #2748

* N/A

- 34 unit tests in `crates/admin/src/telemetry.rs`, including a new end-to-end
  integration test (`test_format_prometheus_text_e2e_otel_compliance`) that
  registers a real metric set, accumulates values, runs them through
  `format_prometheus_text`, and asserts every spec rule (target_info presence
  and sorted labels, `_total` for counters, no `_total_total`, no legacy
  `set=`, `# UNIT` for non-dimensionless units, `otel_scope_name` /
  `otel_scope_version` on every series, attribute key sanitization).
- Targeted unit tests for the helpers: `sanitize_prom_metric_name` /
  `sanitize_prom_label_key` (including underscore-collapse), `build_prom_metric_name`
  (`_total` rules, no double suffix), `ucum_to_prometheus_unit` (simple,
  compound rates, bracketed annotations), `strip_curly_braces`,
  `escape_prom_label_value`.
- Live endpoint smoke test: built `otap-df-controller`, hit `/metrics`, and
  verified the rendered text includes `_total`, `# UNIT`, `otel_scope_name` /
  `otel_scope_version`, sorted attributes, and contains no `_total_total` and
  no `set=`.

Run locally:

```bash
cargo test -p otap-df-admin --lib telemetry
```

Yes — the `/metrics` Prometheus text output changes shape:

- Counter metric names gain a `_total` suffix (e.g. `http_requests` →
  `http_requests_total`). Existing counters that already ended in `_total` are
  unchanged (no double-suffix).
- Non-dimensionless units are appended to the metric name and surfaced via
  `# UNIT`, e.g. `memory_usage` with UCUM `By` becomes `memory_usage_bytes`
  with `# UNIT memory_usage_bytes bytes`.
- The custom `set="..."` label is removed. Instrumentation scope is now
  exposed via `otel_scope_name` and `otel_scope_version` labels on every
  series.
- A `target_info` gauge is emitted when resource attributes are present
  (currently inert: the controller passes an empty map, so `target_info` is
  not emitted in the default deployment until the controller is wired up).
- Attribute keys/values appear in deterministic (sorted) order.

Dashboards or alerts that referenced the old metric names without `_total`,
or that filtered on `set=`, will need to be updated.

### DIFF
--- a/rust/otap-dataflow/crates/admin/src/telemetry.rs
+++ b/rust/otap-dataflow/crates/admin/src/telemetry.rs
@@ -338,9 +338,19 @@ async fn get_metrics(
         }
         OutputFormat::Prometheus => {
             let body = if q.reset {
-                format_prometheus_text(&state.metrics_registry, true, Some(now.timestamp_millis()))
+                format_prometheus_text(
+                    &state.metrics_registry,
+                    true,
+                    Some(now.timestamp_millis()),
+                    &state.resource_attributes,
+                )
             } else {
-                format_prometheus_text(&state.metrics_registry, false, Some(now.timestamp_millis()))
+                format_prometheus_text(
+                    &state.metrics_registry,
+                    false,
+                    Some(now.timestamp_millis()),
+                    &state.resource_attributes,
+                )
             };
             let mut resp = body.into_response();
             // Prometheus text exposition format 0.0.4
@@ -407,7 +417,11 @@ pub async fn get_metrics_aggregate(
             Ok(resp)
         }
         OutputFormat::Prometheus => {
-            let body = agg_prometheus_text(&groups, Some(now.timestamp_millis()));
+            let body = agg_prometheus_text(
+                &groups,
+                Some(now.timestamp_millis()),
+                &state.resource_attributes,
+            );
             let mut resp = body.into_response();
             let _ = resp.headers_mut().insert(
                 header::CONTENT_TYPE,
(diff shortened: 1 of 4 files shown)


rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1038,26 +1161,32 @@ fn format_prometheus_text(
...
                 &mut base_labels,
-                "set=\"{}\"",
+                "otel_scope_name=\"{}\",otel_scope_version=\"\"",
                 escape_prom_label_value(descriptor.name)
             );
         }
-        for (key, value) in attributes.iter_attributes() {
+        // Sort attributes for deterministic output
+        let mut sorted_attrs: Vec<_> = attributes.iter_attributes().collect();
+        sorted_attrs.sort_by(|a, b| a.0.cmp(b.0));
```
lalitb: This one may be pre-existing, but since this PR is making the Prometheus formatter more OTel compliant, it may be worth handling here or tracking separately. If two original labels like http.method and http_method both sanitize to http_method, the OTel mapping says their values should be merged with ; instead of emitting duplicate labels - 

https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/#metric-attributes

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1226,32 +1287,241 @@ fn sanitize_prom_metric_name(s: &str) -> String {
...
+    if matches!(instrument, Instrument::Counter) && !has_total_suffix(&name) {
+        name.push_str("_total");
+    }
+
+    name
+}
+
+/// Returns true if `name` already ends with `_total` as a proper suffix
+/// (preceded by `_` or the entire name is `"total"`).
+fn has_total_suffix(name: &str) -> bool {
```
gouslu, line 1543: should this be case insensitive?

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1226,32 +1287,243 @@ fn sanitize_prom_metric_name(s: &str) -> String {
...
+];
+
+/// Builds a Prometheus metric name with proper unit and type suffixes per OTel spec.
+fn build_prom_metric_name(base_name: &str, unit: &str, instrument: Instrument) -> String {
+    let mut name = sanitize_prom_metric_name(base_name);
+
+    // Append unit suffix if applicable and not already present.
+    if let Some(unit_word) = ucum_to_prometheus_unit(unit) {
+        let unit_suffix = format!("_{unit_word}");
+        if !name.ends_with(&unit_suffix) {
```
lalitb: Small edge case: since this function already tries to handle names that end in `_total`, it should also handle that case when a unit suffix is added. Otherwise a counter named `errors_total` with unit `By` becomes `errors_total_bytes_total`, even though the unit suffix should come before `_total`.

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1226,32 +1287,243 @@ fn sanitize_prom_metric_name(s: &str) -> String {
...
+    ("grams", "second", "grams_per_second"),
+];
+
+/// Builds a Prometheus metric name with proper unit and type suffixes per OTel spec.
+fn build_prom_metric_name(base_name: &str, unit: &str, instrument: Instrument) -> String {
+    let mut name = sanitize_prom_metric_name(base_name);
+
+    // Append unit suffix if applicable and not already present.
+    if let Some(unit_word) = ucum_to_prometheus_unit(unit) {
+        let unit_suffix = format!("_{unit_word}");
```
gouslu: nit: this allocates a small `String` per metric per scrape just to probe the suffix. Since `unit_word` is `&'static str` ASCII and `name` is ASCII after 
sanitization, we can check the suffix without allocating:
 
 ```suggestion
         let already_present = name
             .strip_suffix(unit_word)
             .is_some_and(|prefix| prefix.ends_with('_'));
         if !already_present {
             name.push('_');
             name.push_str(unit_word);
         }
 ```

lalitb (approved):
Nice work. LGTM - only small correctness/spec edge cases.

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -688,20 +702,190 @@ fn agg_line_protocol_text(groups: &[AggregateGroup], timestamp_millis: Option<i6
...
 
-fn agg_prometheus_text(groups: &[AggregateGroup], timestamp_millis: Option<i64>) -> String {
+/// Emits `target_info` gauge from resource attributes into the output buffer.
+fn emit_target_info(out: &mut String, resource_attributes: &HashMap<String, String>) {
+    if resource_attributes.is_empty() {
+        return;
+    }
+    let _ = writeln!(out, "# HELP target_info Target metadata");
+    let _ = writeln!(out, "# TYPE target_info gauge");
+    let mut sorted_attrs: Vec<_> = resource_attributes.iter().collect();
```
gouslu: what happens if two resource attribute keys sanitize to the same Prometheus label key (e.g. service.name and service-name both → service_name)? Looks like both would be written into the same target_info{} series, which I believe Prometheus rejects as a duplicate label. Is that a realistic input here, and if so should we dedupe by sanitized key?

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1038,26 +1161,32 @@ fn format_prometheus_text(
...
                      attributes: &dyn AttributeSetHandler,
                      metrics_iter: MetricsIterator<'_>| {
-        // Render labels from attributes + set label
+        // Render labels from attributes + otel_scope_name/version labels
         let mut base_labels = String::new();
         if !descriptor.name.is_empty() {
             let _ = write!(
                 &mut base_labels,
-                "set=\"{}\"",
+                "otel_scope_name=\"{}\",otel_scope_version=\"\"",
```
gouslu: Should otel_scope_version be omitted when there's no version available, rather than emitted as ""? My read of the OTLP -> Prometheus translation spec is that the label is only added when a non-empty scope version exists. Since MetricsDescriptor doesn't carry a version today, every series ends up with otel_scope_version="". Is that intentional? (Same on line 888.)

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -1526,6 +1526,9 @@ impl<
                     telemetry_registry,
                     memory_pressure_state,
                     log_tap_handle,
+                    // TODO: Wire real OTel resource attributes from the engine configuration
+                    // so that `target_info` is emitted with service.name, etc. See issue #2748.
+                    HashMap::new(),
```
ThomsonTan, line 1531: This keeps `target_info` inert in the real controller path. The formatter now supports emitting it when `resource_attributes` is non-empty, but default `/metrics` scrapes reach this `HashMap::new()`, so operators still will not see `target_info` even though the PR closes #2748. Can we either wire the real controller/engine resource attributes here, or leave #2748 open and track that wiring separately?

rust/otap-dataflow/crates/admin/src/telemetry.rs
cijothomas: I don't think such scenario is applicable here. This is not a prometheus exporter per-se - it just exposes internal metrics in that format, and we control/validate the telemetry schema.

cijothomas: why sorting? Also, `collect()` allocates. Given Resource is static/fixed, we should be pre-serializing it and reusing, instead of doing this in every scrape.

cijothomas (changes requested):
See https://github.com/open-telemetry/otel-arrow/pull/2857/changes#r3198098723

We should be able to pre-generate or cache entire target_info/Resource section given they won't change in engine. 

Also, I suggest to split to smaller PRs.
Eg:
PR1 - fix counter's "_total" suffix
PR2 - convert `By` → `_bytes` etc for Unit
PR3 - Support scope and replace "set="
PR4 - handle Resource/Target_info
PR5 - properly export MMSC following Histogram semantics
etc...

rust/otap-dataflow/crates/controller/src/lib.rs
cijothomas: +1

rust/otap-dataflow/crates/admin/src/telemetry.rs
cijothomas: This is a general scenario - the prometheus endpoint is not accepting and exporting arbitrary metrics. It's only for our own internal stuff, which we'd make sure would never cause such issues. (weaver/other tooling/CI checks etc will ensure them. Maynot be there today, but we can design this prometheus exposition with those assumptions to keep things simpler)

pritishnahar95:
this will be split into smaller PRs to address Cijo's feedback. Marking it as draft for now

pritishnahar95:
This PR was split into 2 separate PRs, and they have been merged, so closing this one now

PR #2860, opened 2026-05-06
feat(engine): wire extensions and capabilities into runtime pipeline

gouslu:
# Change Summary
 
 Part 4 of the Extension System (P1) series. Wires the previously
 landed Capability Registry & Resolver (#2732) into the runtime
 pipeline so extensions are actually instantiated, started, and
 shut down by the engine, and so consumer nodes can resolve their
 capability bindings at build time.
 
 Highlights:
 
 - **Runtime wiring** in `runtime_pipeline.rs`: extension lifecycle
   is invoked before any data-path node is spawned, and `Shutdown`
   is delivered to extensions only after the data path drains
   ("started first, shut down last"). Active and passive extensions
   are handled separately; failures abort startup cleanly.
 - **Local capability ownership aligned with shared** via a
   Box-clone factory pattern, removing the prior asymmetry between
   the two trait variants.
 - **Two reference test capabilities** under
   `crates/engine/src/testing/capability/`: `NoOpStateless` and
   `NoOpStateful`. They exercise every codegen path of the
   `#[capability]` proc macro (`&self` × {sync, async}, `&mut self`
   × {sync, async}, borrowed/owned returns, etc.). Test-only; they
   intentionally live under `testing/` rather than the public
   `capability/` surface.
 - **Comprehensive end-to-end test suite** at
   `crates/engine/tests/extension_e2e.rs` (26 tests) covering:
   passive/active/background extensions, lifecycle ordering and
   shutdown ordering, fail-fast on extension errors, dual-variant
   pruning, one-shot capability enforcement (all accessor
   combinations), shared mutable state across consumers via
   `Arc`/`Rc` for both local and shared trait variants, async
   `&mut self` invocation through boxed handles, and active
   extensions mutating shared state observed by capability
   consumers.
 - **Architecture doc** updated with a precise statement of the
   start-first/shut-down-last invariant (it orders lifecycle
   *calls*, not init completion) and a noted future consideration
   to add an opt-in readiness probe if/when an extension needs an
   init-complete guarantee.
 - **URN unification**: extension URNs now use the canonical
   4-segment form `urn:<namespace>:extension:<id>` (mirroring the
   receiver/processor/exporter convention), with a short form
   `extension:<id>`. The shared parser core lives in a new
   private `crates/config/src/urn.rs`; `node_urn.rs` and
   `extension_urn.rs` delegate to it with disjoint accepted-kind
   sets so the two URN types cannot be confused. As a consequence,
   `NodeKind::Extension` and the now-unreachable
   `Error::ExtensionInNodesSection` are removed. Misplacement
   errors include actionable hints (e.g. *"declare under
   `extensions:` instead of `nodes:`"*).
 - All in-tree node factories (receivers, processors, exporters
   in `core-nodes` and `contrib-nodes`) updated to accept the new
   `&Capabilities` parameter; existing factories that don't depend
   on any capability simply ignore it.
 
 - New `extension_e2e.rs` integration test (26 tests) exercises the
   wiring end-to-end against synthetic receivers/processors/
   exporters/extensions.
 - New unit tests in `urn.rs` cover the shared parser core and the
   misplacement-error hints; existing `extension_urn` and
   `node_urn` tests updated to assert the canonical 4-segment form.
 - Pipeline-level regression tests cover rejecting extension URNs
   in the `nodes:` section and node URNs in the `extensions:`
   section.
 - `cargo xtask check` (structure check + `fmt` + `clippy --workspace
   --all-targets -- -D warnings` + `cargo test --workspace`) passes
   cleanly. No new clippy warnings.
 
 Yes:
 
 - **Extension URN format**: extension URNs now use
   `urn:<namespace>:extension:<id>` (4-segment) instead of the
   pre-existing 3-segment `urn:<namespace>:<id>`. Short form
   `extension:<id>` (expands to `urn:otel:extension:<id>`) is
   available as a developer convenience. Existing 3-segment
   extension URNs in pipeline configs must be updated. The
   previously-bundled `configs/fake-with-extension.yaml` was an
   orphan (its URN had no registered `ExtensionFactory` anywhere
   in the binary, and it had no test/script/doc consumers) and
   was removed in `482feb22c`; the canonical 4-segment shape is
   covered by the `test_extension_with_config_and_capabilities`
   unit test in `crates/config/src/pipeline.rs`. A runnable demo
   config can land in a follow-up alongside a real factory.
 - **New extension authoring surface**: `Extension` trait,
   `ExtensionWrapper::builder` typestate, the
   `extension_capabilities!` macro, and the test capabilities
   `NoOpStateless` / `NoOpStateful` (under `testing/capability/`)
   are now reachable for external extension authors. The
   architecture doc captures the lifecycle contract.
 - **Node factory signature** now includes `&Capabilities` as a
   parameter; existing custom factories will need to accept (and
   may ignore) this new argument


rust/otap-dataflow/crates/engine/src/capability/registry/tests.rs
```diff
@@ -913,14 +913,15 @@ fn test_end_to_end_shared_constructed_policy_mints_independent_instances() {
...
-        "urn:test:counter".into(),
+        "urn:test:extension:counter".into(),
         serde_json::Value::Null,
     ));
     let runtime_config = ExtensionConfig::new("counter");
 
     // ConstructedImpl: each `start()` increments a shared counter so
     // we can confirm the factory ran per-consumer.
     #[derive(Clone)]
+    #[allow(dead_code)] // counter Arc is held to keep refcount-based debugging easy
```
jmacd: I think I understand the comment, but not `#[dead_code]`. Is it dead?

rust/otap-dataflow/crates/engine-macros/src/capability.rs
```diff
@@ -426,16 +443,21 @@ pub(crate) fn expand_capability(args: CapabilityArgs, trait_item: ItemTrait) ->
...
+                ::std::boxed::Box::new(adapter)
             }
         }
 
         // Registers the capability in the `KNOWN_CAPABILITIES` distributed
         // slice at link time, so the engine can enumerate all capabilities
         // compiled into the binary (by name, description, and TypeId) without
         // needing an explicit registration call.
+        //
+        // `#[allow(unsafe_code)]` is required because `linkme::distributed_slice`
```
jmacd, line 560: (Curiousity: how did this not trigger a warning before?)

rust/otap-dataflow/crates/engine/src/capability/registry/tests.rs
```diff
@@ -1410,3 +1411,227 @@ fn test_resolve_bindings_background_not_a_provider() {
...
+    let mut handle_a = caps_a.require_local::<MutSelfCap>().unwrap();
+    let mut handle_b = caps_b.require_local::<MutSelfCap>().unwrap();
+
+    // Mutating A should not affect B.
+    assert_eq!(handle_a.bump(), 1);
+    assert_eq!(handle_a.bump(), 2);
+    assert_eq!(handle_b.bump(), 1, "consumer B must start from prototype");
+    assert_eq!(handle_a.bump(), 3, "consumer A must keep its own counter");
+    assert_eq!(handle_b.bump(), 2, "consumer B must keep its own counter");
+}
```
jmacd, line 1636: Thanks for these tests, they helped wrap my head around the Box-to-Rc change happening in this PR.

rust/otap-dataflow/crates/engine/src/testing/capability/no_op_stateful.rs
```diff
@@ -0,0 +1,76 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! `NoOpStateful` test capability.
```
jmacd, line 4: Should this be added in engine/src/testing?

rust/otap-dataflow/crates/engine/src/testing/capability/no_op_stateless.rs
```diff
@@ -0,0 +1,58 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! `NoOpStateless` test capability.
```
jmacd, line 4: Should this be added in engine/src/testing?

rust/otap-dataflow/crates/engine/tests/extension_e2e.rs
```diff
@@ -0,0 +1,3770 @@
...
+//! 3. **Active spawn ordering** — an active extension's `start()` runs
+//!    BEFORE data-path nodes, but capability *construction* (the
+//!    receiver factory's `create()` body) runs *before* `start()` is
+//!    invoked at all (because `create()` is build-time).
+//! 4. **Fail-fast on extension error** — an active extension whose
+//!    `start()` returns immediately aborts the pipeline; data-path
+//!    drain is not awaited.
+//! 5. **Shutdown ordering** — extensions receive
+//!    `ExtensionControlMsg::Shutdown` only after data-path nodes have
+//!    drained.
```
jmacd, line 28: This looks thorough. Haven't read carefully. :)

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
```diff
@@ -492,30 +535,58 @@ impl<PData: 'static + Debug + Clone + ReceivedAtNode + Unwindable + FlowMetricHo
...
-                                task_results.push(res);
-                            }
-                            Ok(Err(e)) => {
-                                // A task returned an error
-                                return Err(e);
+                    loop {
+                        // `biased;`: prefer data-path completions when both
+                        // arms are simultaneously ready. Functionally
+                        // optional — kept to make intent explicit.
+                        tokio::select! {
```
jmacd: `select!` statements always make me nervous, but this one's easy to read. 😀

jmacd (approved):
The change from Box<> to Rc<> occupies most of this PR, makes sense.

rust/otap-dataflow/crates/engine/src/capability/registry/tests.rs
gouslu: seems like a leftover from some previous testing while I was iterating on it, removed it.

rust/otap-dataflow/crates/engine-macros/src/capability.rs
gouslu: there were no users of the macro prior to this.

rust/otap-dataflow/docs/extension-system-architecture.md
```diff
@@ -23,7 +23,7 @@ is addressed in the Phase 1 implementation:
 | Multiple implementations of same capability | `CapabilityRegistry` keyed by `(extension_name, TypeId)` -- different extensions can provide the same capability |
 | Multiple configured instances | `extensions:` section in YAML, each with a unique name; nodes bind by name in `capabilities:` |
 | Existing config model integration | Extensions are siblings to `nodes` in the pipeline config hierarchy |
-| Preserve performance model (thread-per-core) | Local extensions use `Rc` (no locks); shared extensions use `Clone + Send` with `Arc`-wrapped state |
+| Preserve performance model (thread-per-core) | Local extensions use `Rc<RefCell<T>>` for shared state (no locks); shared extensions use `Clone + Send` with `Arc`-wrapped state |
```
lalitb: Not introduced by this PR, but since this PR makes the runtime behavior real, can we clarify the Phase 1 shared-extension contract here?

If I understand correctly, the extension-requirement document describes `shared` as "one runtime instance shared across cores", while this architecture doc says pipeline-scoped shared extensions are still instantiated per pipeline instance/per-core. Which behavior should extension authors rely on for P1?

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
```diff
@@ -492,30 +535,58 @@ impl<PData: 'static + Debug + Clone + ReceivedAtNode + Unwindable + FlowMetricHo
...
+                            biased;
+                            Some(result) = futures.next(), if !futures.is_empty() => {
+                                match result {
+                                    Ok(Ok(res)) => task_results.push(res),
+                                    Ok(Err(e)) => return Err(e),
```
lalitb: Do we want to still send `Shutdown` to extensions before returning here? Right now, if a node or another extension errors, any already started extensions seem to be dropped without seeing the normal shutdown message. For extensions that own sockets, files, or background work, it would be safer if the error path still gave them the same cleanup signal before the pipeline exits.

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
```diff
@@ -492,30 +535,58 @@ impl<PData: 'static + Debug + Clone + ReceivedAtNode + Unwindable + FlowMetricHo
...
                             }
-                            Err(e) => {
-                                // JoinError (panic or cancellation)
-                                return Err(Error::JoinTaskError {
-                                    is_canceled: e.is_cancelled(),
-                                    is_panic: e.is_panic(),
-                                    error: e.to_string(),
-                                });
+                            Some(result) = extension_lifecycle.next_completion(), if !extension_lifecycle.is_empty() => {
+                                match result {
```
lalitb: Should this wait be bounded by the shutdown grace period? We pass a shutdown deadline to extensions, but the runtime still waits here until the extension task finishes. If an extension misses or ignores `Shutdown`, the pipeline can hang forever. Either enforcing the grace timeout here, or documenting that the deadline is only cooperative, would make the contract clearer.

rust/otap-dataflow/crates/controller/src/startup.rs
```diff
@@ -115,13 +115,6 @@ pub fn validate_pipeline_components<PData: 'static + Clone + Debug>(
                 .get_exporter_factory_map()
                 .get(urn_str)
                 .map(|f| f.validate_config),
-            NodeKind::Extension => {
```
lalitb, line 118: Now that extensions have their own factory registry, should static validation cover them too? We removed the old `NodeKind::Extension` path because extensions are no longer nodes, but I don’t see a replacement pass over `pipeline_cfg.extension_iter()`. So an unknown extension type or bad extension config may still get past validation and only fail later during startup.

rust/otap-dataflow/docs/extension-system-architecture.md
gouslu: Shared across cores in the thread per core model approach will be made possible once an extension can be declared at a higher level, such as pipeline group or engine level. We are not far off from that. This doesn't align with the original requirements document word-by-word, but I think it is a better approach. Because thread-per-core model is there to allow execution/usage in different contexts, meanwhile sharing across cores is more of a concern for at which level the extension is declared in my opinion. Shared is the only mode an extension can be used if it is declared at a higher level, so in that sense it aligns.

rust/otap-dataflow/configs/fake-with-extension.yaml
```diff
@@ -15,7 +15,7 @@ groups:
       main:
         extensions:
           sample_kv_store:
-            type: "urn:otap:sample_shared_key_value_store"
+            type: "urn:otap:extension:sample_shared_key_value_store"
```
lalitb: Is this config meant to run? I don’t see a registered extension for this URN, so it looks like this would fail with `UnknownExtension`.

rust/otap-dataflow/docs/extension-system-architecture.md
gouslu: And, yes, I can document this. I think implementation document should be the reference for behavior.

lalitb: Thanks, that makes sense. Maybe we should update the requirements doc too, since it currently reads like Phase 1`shared` itself means cross-core sharing. Your explanation is clearer: Phase 1 `shared` is per pipeline instance, and real cross-core sharing comes later with group/engine-level extensions.

rust/otap-dataflow/crates/engine-macros/src/capability.rs
```diff
@@ -446,3 +468,50 @@ pub(crate) fn expand_capability(args: CapabilityArgs, trait_item: ItemTrait) ->
...
+
+    #[test]
+    fn accepts_ref_self() {
+        assert!(validate("trait Cap { fn get(&self) -> u32; }").is_none());
+    }
+
+    #[test]
+    fn accepts_ref_self_with_lifetime() {
+        assert!(validate("trait Cap { fn get<'a>(&'a self) -> &'a str; }").is_none());
+    }
```
lalitb, line 594: Can we add a test for method type generics too? Since the generated handles are `Box<dyn ...>`, I’m not sure `fn get<T>(&self, key: T)` is actually dyn-compatible.

rust/otap-dataflow/crates/engine-macros/src/capability.rs
```diff
@@ -247,7 +264,7 @@ pub(crate) fn expand_capability(args: CapabilityArgs, trait_item: ItemTrait) ->
                 .inputs
                 .iter()
                 .filter_map(|arg| {
-                    if let syn::FnArg::Typed(pat_type) = arg {
+                    if let FnArg::Typed(pat_type) = arg {
```
lalitb, line 366: Small follow-up since this PR is touching the macro support surface: this existing panic could maybe become a `syn::Error`. If a user writes a destructured argument like 

`fn set(&self, (a, b): (u64, u64))`, 

they would get a proc-macro panic today instead of a clear compile error. The fix would be to reject that parameter shape with a normal `syn::Error` message.

rust/otap-dataflow/docs/extension-system-architecture.md
```diff
@@ -152,6 +166,29 @@ themselves, and they never touch pipeline data directly.
    nodes have drained. Passive extensions (no lifecycle)
    skip spawning entirely.
 
+   *Scope of the guarantee.* This orders **lifecycle
+   calls**, not init completion. `start()` is async, so
+   invoking it merely enqueues a future; the extension's
+   init body runs concurrently with the data path once
```
lalitb, line 172: Not sure if I missed this in the document, but could we also mention the runtime expectation for active/background extensions here? Since `start()` runs on the same per-core async runtime as the data path, extension authors should avoid blocking or CPU-heavy work there, or move it to a bounded/off-thread path. Otherwise an extension can starve the pipeline on that core.

rust/otap-dataflow/crates/controller/src/startup.rs
gouslu: good catch

rust/otap-dataflow/configs/fake-with-extension.yaml
gouslu: this was for cfg validation, but with added validations but no real capabilities/extensions it fails. we can add it back later, deleting for now.

rust/otap-dataflow/crates/engine-macros/src/capability.rs
gouslu: Method types generics are not supported and validation added for that, as well as tests for the validation.

rust/otap-dataflow/crates/engine/src/lib.rs
```diff
@@ -870,12 +963,134 @@ impl<PData: 'static + Clone + Debug> PipelineFactory<PData> {
...
+                    return kept;
+                }
+
+                // Category 3 / 3a: per-variant consumption.
+                // A variant is "consumed" iff it exists in the bundle
+                // AND its tracker slot was flipped to `true` (i.e., the
+                // ext_id is *absent* from the unconsumed set).
+                let local_present = bundle.local().is_some();
+                let shared_present = bundle.shared().is_some();
+                let local_consumed = local_present && !unconsumed_local.contains(&ext_id);
```
lalitb: I think this pruning logic may be a little too broad for extensions that provide more than one capability.

For example, if one extension provides both `cap_a` and `cap_b`, and a node consumes only `cap_a`, we should still keep that extension variant. But because this collapses the unconsumed entries down to just `ExtensionId`, an unused `cap_b` can make the whole extension look unconsumed and get dropped.

Would it be safer to keep the variant if any capability from that extension/variant was consumed, and only drop it when none of its capabilities were consumed?

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -0,0 +1,232 @@
...
+                passive.push(ext_wrapper);
+                continue;
+            }
+            if let Some(sender) = ext_wrapper.extension_control_sender() {
+                shutdown_senders.push(sender);
+            }
+            let ext_metrics_reporter = metrics_reporter.clone();
+            let ext_id = ext_wrapper.name();
+            let fut = async move {
+                match ext_wrapper.start(ext_metrics_reporter).await {
```
lalitb: Should we preserve/report the extension `TerminalState` here like we do for receiver/exporter tasks?

Right now `ExtensionWrapper::start()` returns a `TerminalState`, but this path turns a successful extension completion into `Ok(())`. If an extension returns final metric snapshots in its terminal state, those snapshots seem to be dropped instead of being reported through the `MetricsReporter`.

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -0,0 +1,232 @@
...
+        let deadline = Instant::now() + EXTENSION_SHUTDOWN_GRACE;
+        self.shutdown_deadline = Some(deadline);
+        for sender in &self.shutdown_senders {
+            // `try_send` is intentional: the extension's control
+            // channel is a small mpsc and we don't want shutdown
+            // broadcast to block the runtime loop. Drop on full is
+            // acceptable — the channel's only other writer is the
+            // dispatcher, which has already terminated by the time
+            // this is called.
+            let _ = sender.sender.try_send(ExtensionControlMsg::Shutdown {
```
lalitb: Can we double-check the shutdown path here?

If `try_send` fails, we silently drop the shutdown message, so later it will just look like the extension ignored shutdown.

Also, even when this send succeeds, I think `ControlChannel::recv()` holds a `Shutdown` with a future deadline until that deadline expires. Since this sends `now + EXTENSION_SHUTDOWN_GRACE`, the extension may only receive shutdown at the end of the grace window, not at the start.

Should shutdown be delivered right away with the deadline as the cleanup budget? And should we at least log if sending shutdown fails?

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -0,0 +1,232 @@
...
+                deadline,
+                reason: SHUTDOWN_REASON.to_string(),
+            });
+        }
```
lquerel: This sends `Shutdown` with a future deadline, but `extension::ControlChannel::recv()` currently holds future-deadline shutdowns until the deadline expires before returning them to the extension. That means a well behaved extension waiting on `recv()` does not see shutdown until its cleanup budget is already exhausted, and normal pipeline shutdown waits roughly `EXTENSION_SHUTDOWN_GRACE` before extensions can exit.

Can we either deliver extension `Shutdown` immediately while carrying the future deadline, or send an already due shutdown here and keep the grace period only as the runtime drain timeout? As written, the shutdown deadline is acting like a delivery delay rather than a cleanup budget.

rust/otap-dataflow/crates/engine/src/lib.rs
```diff
@@ -870,12 +963,134 @@ impl<PData: 'static + Clone + Debug> PipelineFactory<PData> {
...
+        //
+        // A bundle's two variants (local + shared) are evaluated
+        // independently in 3/3a — a SharedAsLocal-fallback bundle
+        // (shared-only) only ever populates `unconsumed_shared`, so the
+        // local check is automatically a no-op for it.
+        let bound_extensions: HashSet<otap_df_config::ExtensionId> = config
+            .node_iter()
+            .flat_map(|(_, node_config)| node_config.capabilities.values().cloned())
+            .collect();
+        let unconsumed_local: HashSet<otap_df_config::ExtensionId> = consumed_tracker
```
lquerel: Does collapsing `ConsumedTracker::unconsumed_local()` into a `HashSet<ExtensionId>` lose capability-level information for extensions that provide more than one capability?

The tracker appears to be keyed by `(capability TypeId, extension ID)`, so an extension providing multiple capabilities can show up once per capability. If a node binds two capabilities from the same extension but only consumes one, would the still-unconsumed capability keep the extension ID in this set and make `local_consumed` false below, even though another capability from that variant was consumed?

If so, should this use "any consumed for this extension variant" semantics, or otherwise only drop the variant when all relevant capability entries for that variant are unconsumed?

rust/otap-dataflow/crates/engine/src/lib.rs
gouslu: good catch, thanks

lalitb (approved):
LGTM. thanks

rust/otap-dataflow/crates/engine-macros/src/capability.rs
```diff
@@ -49,6 +56,13 @@
 //!   implementations could have different associated types, making a single
 //!   registry entry impossible.
 //! - **Associated constants** — same fundamental issue as associated types.
+//! - **Receiver shapes** — only `&self` and `&mut self` are supported.
```
lquerel: The updated docs say unsupported receiver shapes are rejected with compile-time errors and that only `&self` / `&mut self` are supported, but `validate_trait` does not appear to explicitly validate the receiver shape yet. Unsupported receivers will probably still fail through generated `dyn Trait` / adapter code, but that is less aligned with the fail-fast validation added for method generics and destructured parameters.

Could we reject no-self functions, consuming `self`, and arbitrary self types like `self: Box<Self>` directly in `validate_trait` so these cases get the same clear macro-level diagnostic?

rust/otap-dataflow/crates/engine-macros/src/capability.rs
lquerel, line 160: This error message still says "use method-level generics instead" but method-level type and const generics are now rejected below. Could this be narrowed to say that method-level lifetimes are supported, while type/const parameters should be expressed with concrete method signatures or split capability traits?

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -0,0 +1,235 @@
...
+        if tokio::time::timeout_at(drain_deadline, drain)
+            .await
+            .is_err()
+        {
+            otel_warn!(
+                "extension.shutdown.timeout",
+                grace_secs = EXTENSION_SHUTDOWN_GRACE.as_secs(),
+                remaining = self.futures.len()
+            );
+        }
```
lquerel, line 248: We probably need metrics in addition to logs. A misbehaving active/background extension can monopolize the per-core runtime or ignore shutdown, but this lifecycle path appears to only emit `otel_warn!` events for task errors/timeouts. Runtime metrics may show pipeline-level starvation and the wrapped control channel may show pressure, but neither clearly attributes the issue to a specific extension.

Could we add per-extension lifecycle metrics such as shutdown duration, shutdown timeout count, active task count, and possibly control-channel lag/depth labeled by extension id so operators can detect and isolate misbehaving extensions?

Could be done in a future PR. Please create a GH issue to track this if you select this option.

rust/otap-dataflow/crates/controller/src/lib.rs
lquerel, line 287: This controller-side validation path still only iterates `pipeline_cfg.node_iter()`, so extension URNs and extension-specific config are not validated here. `startup::validate_pipeline_components()` now has an extension validation pass, but **live reconfiguration** calls this controller method from `live_control/planning.rs`, meaning a candidate pipeline with an unknown or invalid extension can pass planning and only fail later during runtime build.

Could we mirror or share the extension validation pass here so startup and live-control enforce the same validation boundary? Otherwise this is a validation bypass / availability risk for control-plane updates involving extensions.

Related reference: crates/controller/src/live_control/planning.rs:210

rust/otap-dataflow/crates/config/src/pipeline.rs
lquerel, line 453: [could be in another PR]

Security-hardening question: should `PipelineExtensions` reject duplicate extension IDs during YAML/JSON deserialization? The capabilities map has a custom duplicate-key deserializer because plain `HashMap` deserialization can silently overwrite earlier entries. Since nodes bind capabilities by extension ID, duplicate `extensions.auth` entries could make the effective provider config differ from what a reviewer/operator expects.

Should extensions use a duplicate-key rejecting deserializer as well, especially for auth/credential-style extensions?

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
```diff
@@ -492,32 +538,94 @@ impl<PData: 'static + Debug + Clone + ReceivedAtNode + Unwindable + FlowMetricHo
...
+                    // outer cleanup, giving us an "async finally"
+                    // shape: cleanup always runs, then the loop's
+                    // result is propagated.
+                    let loop_result: Result<Vec<_>, Error> = async {
+                        let mut task_results = Vec::new();
+                        loop {
+                            // `biased;`: prefer data-path completions when both
+                            // arms are simultaneously ready. Functionally
+                            // optional — kept to make intent explicit.
+                            tokio::select! {
```
lquerel, line 602: This appears to defeat the bounded extension shutdown path on normal pipeline shutdown. Once `futures` becomes empty, we broadcast `Shutdown` below, but this loop keeps selecting on `extension_lifecycle.next_completion()`. If an extension ignores shutdown or never returns, the loop awaits that extension forever and never reaches `drain_until_deadline()` at line 626.

Should the loop break immediately after broadcasting shutdown when the data-path futures are empty, and let `drain_until_deadline()` own the bounded wait for remaining extension tasks?

rust/otap-dataflow/crates/engine/src/extension/wrapper.rs
```diff
@@ -81,97 +88,33 @@ impl<R: ControlReceiver> ControlChannel<R> {
...
-                Err(e) => return Err(e),
+        let rx = self.control_rx.as_mut().ok_or(RecvError::Closed)?;
+        match rx.recv().await {
+            Ok(ExtensionControlMsg::Shutdown { deadline, reason }) => {
+                // Close the underlying receiver so subsequent recv()
+                // calls return Closed and the extension's event loop
+                // exits instead of attempting to dispatch further
+                // control messages past Shutdown.
+                let _ = self.control_rx.take();
+                Ok(ExtensionControlMsg::Shutdown { deadline, reason })
```
lquerel, line 113: Is `Shutdown` supposed to be delivered immediately with a future deadline, or only when that deadline expires? This stores a future-deadline shutdown and only returns it to the extension once the deadline has elapsed, so a well behaved extension waiting on `ctrl.recv()` cannot begin cleanup until its cleanup budget is already gone.

If the deadline means "finish cleanup by this time", should `ControlChannel` return `Shutdown` immediately and let the extension use the deadline to bound its own cleanup?

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -0,0 +1,235 @@
...
+        self.shutdown_broadcast_fired = true;
+
+        let deadline = Instant::now() + EXTENSION_SHUTDOWN_GRACE;
+        self.shutdown_deadline = Some(deadline);
+        for sender in &self.shutdown_senders {
+            // Non-blocking: data path has already drained so the
+            // small mpsc should have room. Warn on the rare failure
+            // (extension backed up or already torn down) rather than
+            // silently dropping the cleanup signal.
+            if let Err(e) = sender.sender.try_send(ExtensionControlMsg::Shutdown {
```
lquerel: Dropping a failed `try_send` seems risky for shutdown. Even if no other writer is active after the dispatcher terminates, the channel may still be full with earlier control messages if the extension is slow to receive. In that case a cooperative extension may never see `Shutdown` and will be force-dropped after the drain timeout.

Could we at least log `Full`/`Closed`, or use a bounded async send / priority shutdown path so graceful shutdown is not silently skipped?

I’d also add a regression test for an extension that never completes after data-path drain. The current tests cover cooperative extensions, but they don’t prove that EXTENSION_SHUTDOWN_GRACE actually bounds a misbehaving extension on the normal shutdown path.

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
```diff
@@ -219,6 +232,39 @@ impl<PData: 'static + Debug + Clone + ReceivedAtNode + Unwindable + FlowMetricHo
...
+        // state at capability construction time, or have the capability
+        // surface a not-ready error/default until init progresses).
+        //
+        // TODO: Revisit when an extension actually needs an init-complete
+        // guarantee. Likely shape: opt-in readiness probe registered at
+        // build time (e.g. `builder.with_readiness_probe()` returns a
+        // handle the extension fires from `start()`); the engine awaits
+        // only the registered probes via `try_join_all` before spawning
+        // data-path tasks. Extensions that don't opt in keep today's
+        // behavior with zero overhead.
```
lquerel, line 288: Important aspect for a future PR

lquerel (approved):
I think the PR is broadly in good shape overall, so I’m approving it. It would still be good to either address the comments I added or, for some of them, create GitHub issues for tracking.

The following is more for a future PR, but I don’t think the extension/live-reconfiguration interaction is fully finalized yet:

1. Validation mismatch: startup validation includes extensions, but the live-control validation path appears to call `validate_engine_components_with_factory()` in `crates/controller/src/live_control/planning.rs:210`, and that method currently validates nodes only in `crates/controller/src/lib.rs:287`. So live reconfiguration may accept invalid extension config until runtime build fails.

2. Readiness is probably too early for active extensions: the pipeline thread reports ready after `PipelineFactory::build()` in `crates/controller/src/lib.rs:2077`, before `RuntimePipeline::run_forever()` starts polling active/background extension `start()` futures. That means live rollout may switch generations before extension startup has actually succeeded.

3. Old/new extension overlap during replace: for common cores, live replace launches the target generation before shutting down the previous one in `crates/controller/src/live_control/execution.rs:554`. So old and new extension instances can overlap. That is fine for stateless/passive extensions, but risky for extensions owning exclusive resources like sockets, files, external leases, or singleton clients.

4. Shutdown correctness matters for live reconfig: the extension shutdown behavior is now part of the live-reconfiguration correctness contract. If an extension ignores shutdown, or if shutdown delivery is delayed/dropped, live replace/shutdown can time out because the runtime thread may not exit cleanly.

rust/otap-dataflow/crates/engine/src/lib.rs
gouslu: already fixed and added `test_tracker_consumed_local_any_capability_consumed` test

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
gouslu: this is also fixed, we send shutdown immediately now.

rust/otap-dataflow/crates/controller/src/lib.rs
gouslu: added it!

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
gouslu: Addressed in 6ce6bf43d, bounded send().await with timeout (warns on full/closed/timeout) + regression test for a stuck extension.

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
gouslu: tracking -> https://github.com/open-telemetry/otel-arrow/issues/3039

rust/otap-dataflow/crates/config/src/pipeline.rs
gouslu: addressed

rust/otap-dataflow/crates/config/src/pipeline.rs
```diff
@@ -2890,13 +2888,15 @@ connections:
 
     #[test]
     fn test_extension_urn_format_rejected_in_nodes_section() {
-        // Extension URNs use a 3-segment format (`urn:<ns>:<id>`) which is
-        // incompatible with the 4-segment node URN format (`urn:<ns>:<kind>:<id>`).
-        // Placing an extension-style URN under `nodes:` is a parse error.
+        // Both node and extension URNs share the canonical 4-segment shape
+        // `urn:<ns>:<kind>:<id>`, but the kind segment is disjoint:
+        // NodeUrn rejects `extension`, ExtensionUrn rejects
```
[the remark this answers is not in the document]
gouslu: we don't have capabilities or extensions to test, will be added back later.

PR #2861, opened 2026-05-06
feat(query-engine): fast-path is_type checks on AnyValue columns

### LINKED ISSUE #2841: [OTAP Query Engine]: Improve type checking for `AnyValue`s
In https://github.com/open-telemetry/otel-arrow/pull/2794 we added the capability to check the type of a field/attribute using a syntax like:
```kql
logs | where attributes["x"] is String
```

However, for heterogenous typed `AnyValue` structs columns, we will split the column by concrete type, evaluate the `is_type` UDF on each segment, and stitch the result back together. This is inefficient!

The better solution when we are checking the type of one of these columns would simply be to grap the the type column from the AnyValue struct and check that the value is some type.

### What to do:

In the `is_type` UDF, check if the arg data type is the any value struct (by calling `is_any_value_data_type`), and if so, go down a different execution path:
https://github.com/open-telemetry/otel-arrow/blob/0f2aec6367fdd7c9eea126923f6a2eab352a8599/rust/otap-dataflow/crates/query-engine/src/pipeline/functions/is_type.rs#L74

To ensure this gets called with the unpartitioned struct column, ensure that when invoking the is_type UDF we return `true` here:
https://github.com/open-telemetry/otel-arrow/blob/0f2aec6367fdd7c9eea126923f6a2eab352a8599/rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs#L849-L854

Now that we have this capability, we should also be able to support value type `Array`, `Map`, and `Null` through this type check mechanism. Relax the type restriction here and add tests covering type checks for attributes of these types.
https://github.com/open-telemetry/otel-arrow/blob/0f2aec6367fdd7c9eea126923f6a2eab352a8599/rust/otap-dataflow/crates/query-engine/src/pipeline/filter.rs#L568-L574

ThomsonTan:
# Change Summary

Previously `is_type` on a heterogeneous `AnyValue` struct column went through the generic path: split the column by concrete subtype, evaluate the UDF on each homogeneous segment, then stitch the boolean results back together. For a check that only depends on the discriminator column, this is wasteful.

* Closes #2841 

Unit-test

Yes, user-facing behavior changes in the OTAP query engine:

These now work for AnyValue attributes via the discriminator-column fast path

```
logs | where attributes["complex"] is Array
logs | where attributes["complex"] is Map
logs | where attributes["complex"] is Null
```


rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
```diff
@@ -87,6 +87,8 @@ use crate::pipeline::expr::join::{join, multi_join};
 use crate::pipeline::expr::types::{
     ExprLogicalType, coerce_arithmetic, nested_struct_field_type, root_field_type,
 };
+use crate::pipeline::functions::is_type::IsTypeFunc;
+use crate::pipeline::functions::{arity_range, regexp_substr, substring};
```
lalitb: nit - this import is duplicated by next line. Is it needed?
EDIT - also CI is failing for this, so indeed to be removed :)

lalitb (approved):
LGTM. Would be good to have another look from @albertlockett just in case.

albertlockett (approved):
LGTM @ThomsonTan - thanks for this fix. Good to merge once the clippy issue is resolved

PR #2862, opened 2026-05-06
feat(query-engine): implement log10 scalar function

### LINKED ISSUE #2832: [OPL/OTAP query-engine functions] logarithm
Add support for computing the logarithm of a numeric value in the OPL / OTAP query-engine.
```kql
logs | attributes["log_duration_ms"] = log(attributes["duration_ms"])
```

For general implementation instructions see https://github.com/open-telemetry/otel-arrow/issues/2818

The UDF we should use are:
https://datafusion.apache.org/user-guide/sql/scalar_functions.html#log

This should be parsed into an InvokeFunctionExpr as no special AST expression specifically exists for this capability.

Note - in the future we _may_ wish to give an optional base to this, but for this issue it's probably not strictly necessary.  For now, we probably only need to support passing a single argument and which will compute the default base10 logarithm.

gyanranjanpanda:
Closes #2832

# Implementing `log` Scalar Function

## Objective
Add support for the `log()` mathematical scalar function in the OPL/OTAP query engine.

## Changes Made
- Added `math_expressions` feature to `datafusion` dependency in `rust/otap-dataflow/crates/query-engine/Cargo.toml`.
- Registered the `log` function name in `rust/otap-dataflow/crates/query-engine/src/consts.rs`.
- Added the `log` external function to the parser configuration in `rust/otap-dataflow/crates/query-engine/src/parser.rs`.
- Mapped the `log` function to the corresponding DataFusion UDF in `rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs`.
- Ensured arguments to `log` are coerced to `Float64` in `expr.rs` to correctly interface with DataFusion's scalar UDF logic.
- Added comprehensive unit tests in `rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs` to verify that the `log` function computes correct values and works seamlessly with both KQL and OPL parsers.

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
@@ -1749,7 +1749,7 @@ mod test {
             OtlpProtoMessage,
             opentelemetry::{
                 arrow::v1::ArrowPayloadType,
-                common::v1::{AnyValue, InstrumentationScope, KeyValue},
+                common::v1::{AnyValue, InstrumentationScope, KeyValue, any_value},
                 logs::v1::{LogRecord, LogsData, ResourceLogs, ScopeLogs},
                 metrics::v1::Metric,
                 resource::v1::Resource,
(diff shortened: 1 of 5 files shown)


albertlockett (commented):
@gyanranjanpanda looks like there are unrelated changes. This is basically the conly commit we want, right?
https://github.com/open-telemetry/otel-arrow/pull/2862/commits/52328e2249db98623d50309ca2c5a8bbf232b048

I think you might need to do something like..
```
git reset --hard origin/main
git cherry-pick 52328e2249db98623d50309ca2c5a8bbf232b048
git push -f 
```

gyanranjanpanda:
i did it @albertlockett  thank u for wonderful guidance

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
```diff
@@ -820,6 +828,7 @@ impl DataFusionFunctionDef {
         // upstream in datafusion_functions)
         Some(match func_name {
             ENCODE_FUNC_NAME => Self::new(encode(), ExprLogicalType::String, false, None),
+            LOG_FUNC_NAME => Self::new(log(), ExprLogicalType::Float64, true, None),
```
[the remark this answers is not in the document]
albertlockett: I agree w/ Codex's feedback here that this could cause some confusion about the base of the logarithm
https://github.com/open-telemetry/otel-arrow/pull/2862#discussion_r3197213265

I'm not sure I'm ready to rule on what the best base for `log` should be, and so for now I think we just avoid making the call by being explicit. Here's what I propose we do:

1. change the name of the func being added from `log` to `log10`
2. Use the log10 scalarUDF instead of log https://datafusion.apache.org/user-guide/sql/scalar_functions.html#log10

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
```diff
@@ -567,9 +568,18 @@ impl ExprLogicalPlanner {
...
                 self.plan_function_args(scalar_arg_exprs.into_iter(), functions)?;
 
+            if func_name.as_ref() == LOG_FUNC_NAME {
+                // DataFusion's SQL planner would insert this coercion for log(), but here we
+                // construct the UDF expression directly.
+                arg_exprs = arg_exprs
+                    .into_iter()
+                    .map(|arg_expr| cast(arg_expr, DataType::Float64))
+                    .collect();
+            }
```
albertlockett: I think we should take this out. We don't do this for other functions, even if the UDF has a coercible signature. In the future we _may_ add some code to coerce automatically, but for now I think we should let this just fail on an invalid type rather than having a one-of hard-coded coercion. 

Users want to ensure this is infallible have the ability to check the type before they call this, but ensuring they do so is outside the scope of this PR.

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -4805,6 +4805,68 @@ mod test {
...
+        let log_ratio = log_0
+            .attributes
+            .iter()
+            .find(|kv| kv.key == "log_ratio")
+            .and_then(|kv| kv.value.as_ref())
+            .and_then(|value| value.value.as_ref());
+        let Some(any_value::Value::DoubleValue(log_ratio)) = log_ratio else {
+            panic!("expected log_ratio to be a double attribute");
+        };
+        assert!((*log_ratio - 2.0).abs() < f64::EPSILON);
```
albertlockett: couldn't the comparison be simply `assert_eq!(*log_ratio, 100.0f64.log10());` ?

walkthrough.md
```diff
@@ -0,0 +1,12 @@
+# Implementing `log` Scalar Function
```
albertlockett: I don't think we need this waklthrough file. Can you please remove it?

rust/otap-dataflow/crates/query-engine/examples/scalar_functions.rs
```diff
@@ -0,0 +1,77 @@
+// Copyright The OpenTelemetry Authors
```
albertlockett: Hey @gyanranjanpanda this example seems unrelated to the change in this PR. I think we should remove it as well.

albertlockett (commented):
Looks mostly good @gyanranjanpanda , but I think there are some additional files included in the current changeset that should be removed

albertlockett (approved):

PR #2864, opened 2026-05-06
docs: encourage small, single-purpose PRs in CONTRIBUTING.md

cijothomas:
Adds a "Keep Pull Requests Small and Incremental" section to CONTRIBUTING.md to set explicit expectations for small, single-purpose PRs.


JakeDern:
@lquerel 😜

JakeDern (approved):

cijothomas:
> @lquerel 😜

I have left some flexibility in the wording to account for major architectural changes etc:

_Larger PRs are
  acceptable when genuinely unavoidable, but the description should explain
  why splitting was not practical._

drewrelmas:
We could consider a PR workflow that counts the number of changed lines, classifies as tiny, small, medium, or large, and leaves a suggestion comment if medium or large.

Would not be surprised if there is already a github action out there somewhere to do this.

drewrelmas (approved):

CONTRIBUTING.md
```diff
@@ -127,13 +127,55 @@ Join the OpenTelemetry
...
+  3. Integrate the no-op into the pipeline / runtime.
+  4. Add behavior in small, independently reviewable slices, each with
+     tests.
+  5. Documentation and examples.
+- **Use a tracking issue** for multi-PR features so the overall design and
+  progress remain visible to reviewers across the sequence of PRs.
+
+Reviewers may ask authors to split a PR before reviewing it. PRs that
+bundle unrelated changes, or that are large enough that no single reviewer
+can give every component close attention, are likely to be sent back for
```
lalitb: Maybe soften this slightly?

Instead of saying larger PRs are “likely to be sent back,” we could say reviewers may ask for a split when the PR is hard to review thoroughly.

That keeps reviewer discretion clear without making larger PRs sound automatically suspect.

lalitb:
Agree with the intent. One small suggestion: could we make `reviewability` the main rule, and treat `line-count` as a supporting guideline?

Some larger PRs are cohesive or mostly mechanical and easier to review as one unit, while some small PRs can still be tricky.

Maybe: small and single-purpose by default, but larger is okay when the change is cohesive, clearly scoped, and the description explains why splitting would make review harder, not easier.

This is close to what is already written. It just shifts the emphasis from “unavoidable” to “cohesive.”

lquerel (approved):

lalitb (approved):

PR #2110, opened 2026-02-25
refactor: flip URN format from urn:ns:id:kind to urn:ns:kind:id

pritishnahar95:
# Change Summary

Refactor URN format: flip component name and type ordering from `urn:<namespace>:<id>:<kind>` to `urn:<namespace>:<kind>:<id>`.

The new ordering follows a general-to-specific hierarchy — you first specify what kind of component it is (receiver, processor, exporter), then which one. This makes URNs more intuitive to read and discover.

**Before:**
```
urn:otel:otlp:receiver
urn:otel:batch:processor
urn:otel:noop:exporter
urn:microsoft:geneva:exporter
```

**After:**
```
urn:otel:receiver:otlp
urn:otel:processor:batch
urn:otel:exporter:noop
urn:microsoft:exporter:geneva
```

The shortcut form is also updated from `<id>:<kind>` to `<kind>:<id>` (e.g., `receiver:otlp`).

Changes span the following:
- Core URN parsing logic (`node_urn.rs`): updated `parse()`, `build_node_urn()`, `split_segments()`, error messages, and all doc-comments
- All URN constant definitions across receiver, processor, and exporter implementations
- All YAML configuration files (quoted and unquoted URN values)
- All test fixtures (JSON and YAML) and inline test strings
- All documentation (`urns.md`, `configuration-model.md`, `otlp-receiver.md`, crate READMEs)
- Added a test case verifying the old format is now rejected

## What issue does this PR close?

* Closes [#2108](https://github.com/open-telemetry/otel-arrow/issues/2108)

## How are these changes tested?

- All existing unit tests in `otap-df-config` pass with zero failures.
- Updated test assertions to validate the new URN format.
- Added explicit test case confirming old format (`urn:otel:otlp:receiver`) is rejected.
- Full workspace build (`cargo check --workspace`) compiles cleanly.

## Are there any user-facing changes?

Yes. All URN references in pipeline configuration files must use the new `urn:<namespace>:<kind>:<id>` format. The shortcut form changes from `<id>:<kind>` to `<kind>:<id>`. For example:
- `urn:otel:otlp:receiver` → `urn:otel:receiver:otlp`
- `otlp:receiver` → `receiver:otlp`

Existing configurations using the old format will be rejected with a clear error message pointing to the URN documentation.

PR #1948, opened 2026-02-04
Standardization of urns, validation and usage

daviddahl:
# Change Summary

  Standardized otap-df-otap node URNs to the canonical urn:<namespace>:<id>:<kind> format, added strict parsing/normalization
  (including OTel shortcut support), updated component constants/configs/templates/docs to match, and documented otelcol config
  compatibility design and URN rules.

  ## What issue does this PR close?

  - Closes #1831

  ## How are these changes tested?

  - cargo test (per local confirmation)
  - Added unit/config tests for URN normalization and legacy URN rejection in otap_df_config

  ## Are there any user-facing changes?

  Yes. Configuration now enforces canonical URN format and accepts the OTel shortcut form; legacy URNs are rejected with a doc-linked
  error message.

PR #2872, opened 2026-05-07
refactor(core-nodes): rename fake_data_generator module to traffic_generator

cijothomas:
# Change Summary

Renames the `fake_data_generator` receiver module to `traffic_generator` so that the module folder and Rust symbols match the URN id (`urn:otel:receiver:traffic_generator`). This brings the receiver in line with the convention used by every other receiver/processor/exporter in `core-nodes`, where the module name matches the URN id segment.

This was [flagged in PR #1857 review feedback](https://github.com/open-telemetry/otel-arrow/pull/1857#discussion_r2718063522). The URN was last touched by [#1948](https://github.com/open-telemetry/otel-arrow/pull/1948) and [#2110](https://github.com/open-telemetry/otel-arrow/pull/2110), but the module folder and Rust symbol names were not realigned at the time.

## Scope

To keep this PR focused, the public observability surface (metric set name, log event names, semconv scope, `FakeSignalReceiverMetrics` type) is intentionally **not** renamed here. A follow-up PR will rename the metrics and related strings.

`cargo check`, the renamed module's unit tests, and the otap integration / validation tests that depend on it all pass locally.

The Rust module path and the URN `pub const` constant are renamed. The URN string value itself is unchanged, so YAML/config consumers are unaffected.


lquerel (approved):
Thanks!

Issue #2531, opened 2026-04-02
Meter names should not need the word "metrics" in it


cijothomas:
https://github.com/open-telemetry/otel-arrow/pull/2527#discussion_r3030821794

PR #2879, opened 2026-05-07
Rename otlp.receiver.metrics meter to otlp.receiver

cijothomas:
Drops the redundant trailing `.metrics` from the OTLP receiver's meter/scope name. A meter name already names a set of metrics, so `otlp.receiver.metrics` was tautological in scrape output and view selectors.

Part of #2531. First (smallest) of a series of per-component renames; this one is fully self-contained.

Instrument names are unchanged — only the scope/meter name.


lquerel (approved):
LGTM

JakeDern (approved):

Issue #2884, opened 2026-05-07
Opt-in items and bytes counts at the node level.


jmacd:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Description

We have node-level metrics for requests consumed and produced following OTel Collector semantic conventions.

We also want item counters and bytes counters of the same form.

### Additional Context

https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md defines otelcol.{component_type}.produced.items for example.

We should support an opt-in metric for bytes and items.

Note that it may be possible to compute num_items and num_bytes once per (Context, Payload), if done carefully, so that we only recompute the number of items once per request (/mutation) in the pipeline. However, this is not a requirement.

drewrelmas:
I don't necessarily see this as conflicting with the work in #2859, they target two separate things:
1. Per-component signal counts (parity with Collector metrics)
2. Per-flow signal counts (Operator-defined logical groups)

drewrelmas:
Partially solved in #3437, follow-up PR could add bytes metrics.

drewrelmas:
I was thinking about receiver and exporter `bytes` counters as well, and I think we need to come to some decisions regarding this issue.

The Collector [component universal telemetry RFC](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md#auto-instrumented-metrics) calls the protocol-neutral measurement `size`, not `bytes`. It defines it as the protobuf size of the PData crossing a component boundary.

I propose implementing:

- `node.producer.produced.size{signal,outcome}`
- `node.consumer.consumed.size{signal,outcome}`

with unit `By`, disabled by default.

Size should use the same node telemetry policy mechanism as item counts, but with an independent option:

```yaml
policies: telemetry:
    item_counts: true
    size: true
```

The settings should remain independent because their costs differ. Item counting is generally a structural operation, while canonical protobuf sizing may require a full traversal or OTAP/Arrow-to-OTLP conversion solely for measurement. Users should be able to enable item counts without paying the sizing cost.

This separates two concepts:

- `size`: canonical uncompressed OTLP protobuf size, comparable across components and exporter protocols.
- `bytes`: a concrete transport or storage representation, such as compressed HTTP, ArrowStream, Kafka payload, or Parquet storage bytes.

[...]

Issue #2859, opened 2026-05-06
Extend `flow_metrics` for opt-in signal count auto-instrumentation


drewrelmas:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

#### Motivation

otel-arrow already instruments data movement at the **channel** level (message/batch counts). Consumers typically also need it at the **signal-item** level — counts of individual log records, metric data points, or spans, especially at decision points where some are kept and others dropped. Channel metrics can't answer "how many records did this receiver accept and reject?" because batch sizes vary and decisions are per-record.

`flow_metrics` is the right shape for signal-item counts (uniform, opt-in, attribute-rich, view-friendly), but four limitations prevent it from covering these cases. This issue proposes four small, independent extensions so `flow_metrics` can serve as the engine's universal opt-in auto-instrumentation surface across every node kind.

#### Why extend `flow_metrics` rather than add bespoke per-component counters

Two properties to preserve:

- **Off by default.** Per-record counting is hot-path cost. `flow_metrics` is configured per-flow at deployment, so users pay only where they've asked. Bespoke always-on counters tax every deployment.
- **Component authors only add metrics specific to their component.** Topology metrics (`incoming`/`outgoing`/`compute.duration`) come free from being in a flow; decision outcomes come from one capability declaration plus `record_kept`/`record_dropped` calls — no new instruments, MeterProvider plumbing, or naming conventions per component.

#### Limitations to fix

1. **Receivers and exporters can't be flow_metrics endpoints.** `build_flow_metrics_state` rejects them — but they're the natural boundary nodes.
2. **`start_node == end_node` is rejected.** Many real cases are inherently single-node (a self-validating receiver; a single-stage filter).
3. **All flow_metrics instruments emit under one `flow` Meter.** OTel View selectors filter by scope name + scope attributes + instrument name — distinct views can't target distinct flow purposes today since every flow shares the single `flow` scope. `FlowAttributeSet` already places `flow.name` / `flow.start_node` / `flow.end_node` on scope attributes (`crates/engine/src/flow_metric.rs:59-74`); the gap is operator control over a per-flow purpose differentiator that View selectors can match on.
4. **No decision-outcome metric sets.** `incoming`/`outgoing` are *topology* metrics (data crossing edges); they don't capture *decisions* (records the node explicitly judged). The two coincide for some nodes but diverge in important cases — see below.

[...]

PR #2885, opened 2026-05-07
Avoid OtapPdata::num_items in batch_processor

### LINKED ISSUE #2882: Avoid using num_items for batch_processor bookkeeping
### Pre-filing checklist

### Component(s)

### Bug Description

Part of #2881 

### Steps to Reproduce

Batch processor can use the sizer's weight function instead of num_items for matching inputs/outputs.

### Expected Behavior

O(1) bookkeeping

### Actual Behavior

Batch processor can use the sizer's weight function instead of num_items for matching inputs/outputs.

### OTel-Arrow Version

HEAD

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

jmacd:
# Change Summary

Fixes #2882. Uses the sizer's natural weight which is O(1) for batch processor bookkeeping. OTLP batching pays a penalty otherwise, due to an (not strictly necessary) allocation in addition to the traversal.


rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
```diff
@@ -1670,33 +1622,19 @@ mod tests {
         (telemetry_registry, metrics_reporter, phase)
     }
 
-    /// Helper to verify consumed and produced item metrics
+    /// Helper to verify that batch counters were incremented.
     fn verify_item_metrics(
         telemetry_registry: &TelemetryRegistryHandle,
         signal: SignalType,
-        expected_items: usize,
+        _expected_items: usize,
```
JakeDern: Can we whack this parameter if we don't need it?

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
```diff
@@ -546,9 +526,6 @@ pub struct BatchProcessorMetrics {
     /// Number of input requests pending at flush time
     #[metric(unit = "{request}")]
     flush_pending_requests: Mmsc,
-    /// Number of primary signal items pending at flush time
-    #[metric(unit = "{item}")]
```
JakeDern, line 550: Nice, thanks for removing these!

JakeDern (approved):
Cool, thanks!

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
```diff
@@ -738,30 +712,18 @@ impl BatchProcessor {
         effect: &mut local::EffectHandler<OtapPdata>,
         request: OtapPdata,
     ) -> Result<(), EngineError> {
-        let items = request.num_items();
-
-        if items == 0 {
+        if request.is_empty() {
```
utpilla: This subtly changes the semantics of the empty check. Not in the happy path, but `is_empty()` would now admit payloads that the old `num_items() == 0` predicate dropped, for e.g., an OTLP request with non-empty bytes but no log records (a `ResourceLogs` with `scope_logs: []`), or an OTAP Metrics batch where the `UnivariateMetrics` batch is empty but the data-point batches aren't (`is_empty()` only checks `UnivariateMetrics`, while `num_items()` sums across all data-point batches).

Once admitted with `weight == 0`, these payloads:

- get counted in `consumed_batches_*` instead of `dropped_empty_records` (so the drop metric undercounts and the consumed metric overcounts),
- consume an inbound request slot,
- don't contribute to the size threshold for the flush

We could do the drop in `accept_payload` after `batch_size()` to preserve the original semantics. It would naturally extend to `Sizer::Bytes`/`Requests` later too:

```rust
let weight = self.fmtcfg.sizer.batch_size(&payload)?;
if weight == 0 {
    self.metrics.dropped_empty_records.inc();
    let pdata = OtapPdata::new(ctx, payload.into());
    effect.notify_ack(AckMsg::new(pdata)).await?;
    return Ok(());
}
```

lalitb: I think this can miss non-empty OTLP envelopes with zero records. For example, an OTLP `LogsData` can contain a `resource_logs` entry but no `log_records`, so the encoded byte buffer is not empty while `num_items()` is still `0`. In forced OTAP mode that can enter the item-sized buffer with weight `0` and later be treated as empty, so I think we still need a zero-item check before accepting it.

lalitb: Oh, I missed that @utpilla had already called this out a couple of minutes earlier. Please ignore my duplicate comment :)

drewrelmas:
In general, I think the intention behind this change is related to issue #2859. If the `num_items()` is an expensive operation, we should:
1. Try to optimize it as best we can
2. (More importantly) only do signal count reporting at specific engine-level config-requested points and discourage individual components from calling it.

jmacd:
@drewrelmas I filed https://github.com/open-telemetry/otel-arrow/issues/2884 to talk about the same topic -- I would not add items/bytes counters to the flow measurement/stopwatch thing, I would add them to individual nodes, allow us to enable on a node-by-node basis at detailed metric level.

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
jmacd: Thanks. I changed the parameter to use expected (input, output) counts.

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
jmacd: Good catch. This leads me to drop the handling and counting of empty records. :-)

OpenTelemetry explicitly permits this behavior (with a MAY) & discourages this from happening (with a SHOULD NOT). See https://github.com/open-telemetry/opentelemetry-proto/pull/622

jmacd: I think we're better off not bothering to test for empty records. To be consistent, I've removed that branch entirely, dropped the corresponding metric. See https://github.com/open-telemetry/opentelemetry-proto/pull/622

lquerel (approved):
LGTM

lquerel:
@drewrelmas Determining the number of signals/items is only costly for pdata with pure OTLP protobytes backend. For regular OTAP based pdata the cost is probably negligeable.

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
jmacd: @utpilla requested to continue dropping the weight=0 requests, which I've done. however, since the metric would then be miseleading, we left it out.

Issue #2834, opened 2026-05-05
[OPL/OTAP query-engine functions] additional hash algorithms


albertlockett:
Add support for generating a hash of some value per row in OPL / OTAP query-engine. We already support sha256, so this work will be extending to support additional hashing algorithms.
```kql
logs | set attributes["x"] = sha512(attributes["x"])
```

For parity with OTTL should add: fnv, murmur3, md5, sha1/512, xxh3/128

For general implementation instructions see https://github.com/open-telemetry/otel-arrow/issues/2818

The UDFs we should use are already existing
https://datafusion.apache.org/user-guide/sql/scalar_functions.html#hashing-functions

We'll need to create a custom scalar UDF to support sha1, xxh and fnv. For inspiration, see the custom scalar UDFs that have already been created:
https://github.com/open-telemetry/otel-arrow/tree/main/rust/otap-dataflow/crates/query-engine/src/pipeline/functions
See also the implementation of the hash function:
https://github.com/apache/datafusion/blob/main/datafusion/functions/src/crypto/md5.rs

This should be parsed into an InvokeFunctionExpr as no special AST expression specifically exists for this capability.

SzymonIwaniuk:
Hi @albertlockett Could you assign this ticket to me? Thanks 🚀

albertlockett:
Thanks @SzymonIwaniuk - assigned to you!

albertlockett:
closed in #2887

PR #2887, opened 2026-05-07
feat(query-engine): add hash functions fnv, murmur3, md5, sha1, sha512, xxh3, xxh128

SzymonIwaniuk:
# Change Summary
  Add seven hash functions to the OTAP query-engine, per #2834:
  - `md5()`: MD5 digest, backed by DataFusion's built-in [`crypto::md5`](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#md5) UDF.
  - `sha1()`: SHA-1 digest, implemented as a custom `ScalarUDF` using the `sha1` crate (DataFusion has no SHA-1 equivalent).
  - `sha512()`: SHA-512 digest, backed by DataFusion's built-in [`crypto::sha512`](https://datafusion.apache.org/user-guide/sql/scalar_functions.html#sha512) UDF.
  - `fnv()`: FNV-1a 64-bit hash, implemented as a custom `ScalarUDF`.
  - `murmur3()`: MurmurHash3 32-bit hash, implemented as a custom `ScalarUDF`.
  - `xxh3()`: XXH3 64-bit hash, implemented as a custom `ScalarUDF` using `xxhash-rust`.
  - `xxh128()`: XXH3 128-bit hash, implemented as a custom `ScalarUDF` using `xxhash-rust`.
  ```kql
  logs | extend attributes["hash"] = encode(md5(attributes["body"]), "hex")
  logs | extend attributes["bucket"] = murmur3(attributes["service.name"])
  logs | extend attributes["sig"] = xxh3(attributes["message"])
  ```
  ## Files
  - `consts.rs`, `parser.rs`: register all 7 functions as external functions with `param_placeholders(1)`.
  - `pipeline/expr.rs`: import DataFusion's `md5()`, `sha512()` and the new custom UDFs, add them to `DataFusionFunctionDef::from_func_name` with `requires_dict_downcast: true`.
  - `pipeline/functions/fnv.rs` (new): custom `FnvHashFunc` - FNV-1a 64-bit, returns `Int64`.
  - `pipeline/functions/murmur3.rs` (new): custom `Murmur3HashFunc` - MurmurHash3 32-bit, returns `Int64`.
  - `pipeline/functions/sha1.rs` (new): custom `Sha1Func` using the `sha1` crate, returns `Binary`.
  - `pipeline/functions/xxh3.rs` (new): custom `Xxh3Func` using `xxhash-rust`, returns `Int64`.
  - `pipeline/functions/xxh128.rs` (new): custom `Xxh128Func` using `xxhash-rust`, returns `Binary` (16 bytes big-endian).
  - `pipeline/functions.rs`: module declarations and `make_udf_function!` registrations.
  - `crates/query-engine/Cargo.toml`: add `sha1` and `xxhash-rust` workspace dependencies.
  - Closes sub-issue of #2834
  - Part of #2818
  Two layers of tests, all in this PR:
  - Unit tests in each custom UDF module (`fnv`, `murmur3`, `sha1`, `xxh3`, `xxh128`): verify scalar string input, binary input, and null handling.
  - `pipeline::assign` end-to-end tests for every function, exercising both OPL and KQL parsers through the full pipeline. Binary-returning functions (`md5`, `sha1`, `sha512`, `xxh128`) are wrapped with `encode(..., "hex")` and
  the output hex string is asserted. Integer-returning functions (`fnv`, `murmur3`, `xxh3`) assert the `Int64` value directly.
  `cargo xtask quick-check` passes clean.
  Yes users of the transform processor / query-engine can now call these hash functions in OPL/KQL programs:
  ```kql
  logs | extend attributes["id"] = encode(sha1(attributes["body"]), "hex")
  logs | extend attributes["bucket"] = fnv(attributes["service.name"])
  logs | extend attributes["sig"] = xxh3(attributes["message"])
  logs | extend attributes["hash"] = encode(xxh128(attributes["message"]), "hex")


rust/otap-dataflow/Cargo.toml
```diff
@@ -211,6 +211,7 @@ weaver_forge = { git = "https://github.com/open-telemetry/weaver.git", tag = "v0
 weaver_resolved_schema = { git = "https://github.com/open-telemetry/weaver.git", tag = "v0.21.2"}
 weaver_resolver = { git = "https://github.com/open-telemetry/weaver.git", tag = "v0.21.2"}
 weaver_semconv = { git = "https://github.com/open-telemetry/weaver.git", tag = "v0.21.2"}
+sha1 = { version = "0.10", features = ["oid"] }
```
lalitb: Can we avoid making `sha1` a default dependency here, or gate just the `sha1()` function behind a feature? `otap-df-query-engine` is pulled into the normal `df_engine` build through core-nodes/transform-processor, so this makes SHA-1 part of the default engine dependency graph. Also, the workspace `sha1` dependency enables `oid`, which does not look used by this implementation.

lalitb: for reference, the concerns on sha-1 - #2827

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
```diff
@@ -839,6 +843,13 @@ impl DataFusionFunctionDef {
             FORMAT_DATETIME_FUNC_NAME => Self::new(to_char(), ExprLogicalType::String, false, None),
             RTRIM_FUNC_NAME => Self::new(rtrim(), ExprLogicalType::String, true, None),
             SHA256_FUNC_NAME => Self::new(sha256(), ExprLogicalType::Binary, true, None),
+            MD5_FUNC_NAME => Self::new(md5(), ExprLogicalType::Binary, true, None),
```
lalitb: Can we check this return type? DataFusion `md5()` already returns a hex string, not raw bytes like `sha256()`. So marking it as `Binary` may be wrong.

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -4805,6 +4805,306 @@ mod test {
...
+                .attributes(vec![
+                    KeyValue::new("str_attr", AnyValue::new_string("y")),
+                    KeyValue::new("binary_attr", AnyValue::new_bytes(b"418")),
+                ])
+                .finish(),
+        ]);
+        let query = r#"logs | extend
+          attributes["str_attr"] = encode(md5(attributes["str_attr"]), "hex"),
+          attributes["binary_attr"] = encode(md5(attributes["binary_attr"]), "hex")
+      "#;
```
lalitb, line 4820: Related to the `md5()` return type: DataFusion `md5()` already returns a hex string, so should these be `md5(attributes["..."])` directly instead of wrapping with `encode(..., "hex")`?

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
albertlockett: This is correct, we should have `String` here, but it turns out this actually returns a `DataType::Utf8View`
https://github.com/apache/datafusion/blob/aca4d1377bf9785e6fa9a154c579cc761ca7b54b/datafusion/functions/src/crypto/md5.rs#L88-L90

We will actually fail if we try to assign this as is done in the test:
https://github.com/open-telemetry/otel-arrow/blob/4857f2baa0ccebb76a6c379df75e8a80da7170cb/rust/otap-dataflow/crates/pdata/src/otap/transform/upsert_attributes.rs#L299-L301

I think this should actually be:
```suggestion
            MD5_FUNC_NAME => Self::new(md5(), ExprLogicalType::String, true, Some(DataType::Utf8)),
```

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
albertlockett: This is correct - but will need to add change from [this comment](https://github.com/open-telemetry/otel-arrow/pull/2887/files#r3204210116) for this test to pass

rust/otap-dataflow/crates/query-engine/src/pipeline/functions.rs
```diff
@@ -13,12 +13,21 @@ use datafusion::logical_expr::{self as datafusion_expr, TypeSignature};
...
 mod contains;
+mod fnv;
 pub(crate) mod is_type;
+mod murmur3;
 mod regexp_substr;
+mod sha1;
 mod substring;
-mod uuidv7;
+mod xxh128;
+mod xxh3;
```
albertlockett, line 25: This line got removed in the merge. Can we add it back?
```suggestion
mod xxh3;
mod uuidv7;
```

rust/otap-dataflow/crates/query-engine/src/pipeline/functions/murmur3.rs
```diff
@@ -0,0 +1,260 @@
...
+
+fn hash_scalar(scalar: &ScalarValue) -> Result<Option<i64>> {
+    match scalar {
+        ScalarValue::Utf8(v) | ScalarValue::LargeUtf8(v) => {
+            Ok(v.as_deref().map(|s| murmur3_32(s.as_bytes())))
+        }
+        ScalarValue::Utf8View(v) => Ok(v.as_deref().map(|s| murmur3_32(s.as_bytes()))),
+        ScalarValue::Binary(v) | ScalarValue::LargeBinary(v) => {
+            Ok(v.as_ref().map(|b| murmur3_32(b)))
+        }
```
albertlockett, line 139: Why do we support different types when handling a scalar versus handling an Array?

For the Arrray case, it looks like we only support `Utf8`, `LargeUtf8` and `Binary`. I think these supported types should maybe be consistent, unless there's a good reason for them not to be? Same comment for fnv, sha1 and xxh

rust/otap-dataflow/crates/query-engine/src/pipeline/functions.rs
SzymonIwaniuk: Sorry, I added uuidv7 back.

rust/otap-dataflow/crates/query-engine/src/pipeline/functions/murmur3.rs
SzymonIwaniuk: Good catch, made the array path consistent with the scalar path by adding `Utf8View` and `LargeBinary` support to hash_array in all five udf files.

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
SzymonIwaniuk: Fixed.

rust/otap-dataflow/Cargo.toml
SzymonIwaniuk: I gated `sha1` behind an `sha1-hash` cargo feature. The module, UDF registration, parser, and tests are all behind `#[cfg(feature = "sha1-hash")]`

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
SzymonIwaniuk: Done, I changed the return type to`ExprLogicalType::String` with `Some(DataType::Utf8)` and removed the `encode()` wrapper from the test, also updated the expected hash values accordingly.

SzymonIwaniuk:
Hey, all changes addressed and `cargo test -p otap-df-query-engine` unit tests pass locally.
```
test result: ok. 623 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.14s

   Doc-tests otap_df_query_engine

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

rust/otap-dataflow/Cargo.toml
lquerel: @SzymonIwaniuk @lalitb That could be done in a future PR but I’m slightly concerned by the growing number of SHA-1 related features. While the current usages are valid (e.g. WebSocket protocol compatibility or non-cryptographic hash functions in OPL), multiplying fine-grained features could make the build matrix and dependency story harder to reason about over time.

I’d recommend converging toward a single high-level compatibility feature (e.g. sha1-compat) controlling all SHA-1 usage globally, instead of component-specific flags. Internally, all SHA-1 usage should go through a shared utility module with explicit documentation clarifying that it is used for protocol compatibility/non-security purposes only.

Please open a GH issue to track this if not integrated in this PR. Thanks.

SzymonIwaniuk: @lquerel @lalitb I agree with the consolidation of SHA-1 related feature flags. I think this would be better addressed in a follow-up issue rather than in this PR, as it involves changes across multiple components. Feel free to assign it to me once the issue is created after this PR. @lalitb I'd like to know what you think about this approach, would that work for you?

SzymonIwaniuk:
Hey @albertlockett, I fixed the bad merge in assign.rs, could you please re-run CI? I ran the full suite before pushing, so it should finally work.

rust/otap-dataflow/Cargo.toml
lalitb: @SzymonIwaniuk @lquerel I agree it's better to have a single higher-level compatibility feature, plus shared documentation around permitted non-security SHA-1 usage.
For now, can we keep the local flag for this PR (as it is already there). After this merges, I can raise a follow-up PR to consolidate the SHA-1-related flags under a unified compatibility feature. Internally for us, SHA-1 usage gets flagged regardless of whether it is crypto-related or only used for compatibility/non-security hashing. Also cc @AaronRM as he has some similar experience in handling md-5 usage.

rust/otap-dataflow/crates/query-engine/src/pipeline/functions.rs
albertlockett: No need to appolgize, thanks for fixing :)

albertlockett (approved):

lalitb (approved):

PR #2888, opened 2026-05-07
Rename engine.metrics and pipeline.metrics meters

cijothomas:
Drops the redundant trailing `.metrics` from the engine-wide and per-pipeline meter/scope names:

- `engine.metrics` → `engine`
- `pipeline.metrics` → `pipeline`

A meter name already names a set of metrics, so the trailing `.metrics` was tautological in scrape output and view selectors. Instrument names are unchanged — only the scope/meter names.

Part of #2531. Follow-up to #2879 (which renamed `otlp.receiver.metrics` → `otlp.receiver`). The remaining per-component renames (core-nodes receivers/processors/exporters, contrib-nodes, validation, docs sweep) will land as separate smaller PRs.

## Breaking change for downstream consumers

Anything that selects metrics by **scope/meter name** must be updated. The instrument names are unchanged — only the scope name changes.

Examples that need updating:

- View configurations that filter by `ScopeName` (e.g. `ScopeName: "engine.metrics"` → `ScopeName: "engine"`, `ScopeName: "pipeline.metrics"` → `ScopeName: "pipeline"`).
- Prometheus relabeling/alerting that keys off the `set="…"` label.
- Dashboards or queries that group by OTLP scope name.

Effective emitted name mapping (`<scope>.<instrument>`):

| Before | After |
|---|---|
| `engine.metrics.memory_rss` | `engine.memory_rss` |
| `engine.metrics.cpu_utilization` | `engine.cpu_utilization` |
| `pipeline.metrics.uptime` | `pipeline.uptime` |
| `pipeline.metrics.memory_usage` | `pipeline.memory_usage` |
| `pipeline.metrics.cpu_utilization` | `pipeline.cpu_utilization` |
| (…and the rest of the `pipeline.*` set) | (analogous) |

## In-repo consumers updated in this PR

- Rust: `enginectl` TUI (`refresh.rs`, `tests.rs`), `admin-types` wire-shape roundtrip tests
- Admin UI JS: `engine-metrics.js`, `main.js`, `charts-controller.js`, `selectors-ui.js`, `polling-controller.test.mjs`
- Python: `scripts/engine-metrics.py`
- Docs: `engine/telemetry.md`, `docs/admin/architecture.md`, `docs/memory-limiter-phase1.md`, `docs/telemetry/metrics-guide.md`


drewrelmas:
@cijothomas do you think this is a widespread problem across the repo that we should tackle?

For example, I think I am guilty of something similar in the newly-added `flow_metrics`:
https://github.com/open-telemetry/otel-arrow/blob/fbf013edfb32b5c1f4922f1cfdf1ecad10dfa2a8/rust/otap-dataflow/crates/engine/src/flow_metric.rs#L30-L56

The instrument name doesn't really need to also include `flow`.

EDIT: I should have finished reading the PR description:
> The remaining per-component renames (core-nodes receivers/processors/exporters, contrib-nodes, validation, docs sweep) will land as separate smaller PRs.

drewrelmas (approved):
LGTM - noting that it is a breaking change for existing consumers, but I think it has positive outcome

gouslu:
I was wondering if "metric_set" should do some validation and maybe print a warning for the redundancy or something like that, at least for dev builds? Does such a guardrail make sense and can it be added in a way that it makes sense, or is it too restrictive? Think about it as linting :), but for metric names

cijothomas:
> I was wondering if "metric_set" should do some validation and maybe print a warning for the redundancy or something like that, at least for dev builds? Does such a guardrail make sense and can it be added in a way that it makes sense, or is it too restrictive? Think about it as linting :), but for metric names

Good point! I'm tracking it via https://github.com/open-telemetry/otel-arrow/issues/1613 where entire internal metrics would be coming a internal-conventions.yaml and code auto-generated by weaver, and CI gated! We are not there yet, but that is the long term plan.

PR #2900, opened 2026-05-07
fix(admin): apply OTel→Prometheus name & unit suffix rules (#2748)

pritishnahar95:
# Change Summary

Apply the OpenTelemetry → Prometheus name & unit suffix rules in the admin
HTTP server's Prometheus text exposition.

Per the OpenTelemetry spec for Prometheus exposition:

- Counter metric names must end in `_total`.
- Unit suffix derived from UCUM (e.g. `By` → `_bytes`, `By/s` →
  `_bytes_per_second`) is inserted between the base name and `_total`.
- A base name that already ends in `_total` must not push the unit suffix
  after it (`errors_total` + `By` → `errors_bytes_total`, not
  `errors_total_bytes_total`).

This PR introduces `build_prom_metric_name` and the supporting UCUM lookup
tables/helpers, and routes both `format_prometheus_text` and
`agg_prometheus_text` through it. `sanitize_prom_metric_name` also now
collapses consecutive `_` per spec §Metric Names.

This PR is the first of two carved out of the original PR #2748. It is
intentionally focused on metric-name and unit-suffix rules so reviewers can
evaluate that surface in isolation. Scope-label rename (`set=` →
`otel_scope_name=`), `target_info` rendering and caching, and label-key
collision merging are split into a follow-up PR.

Partially addresses #2748. The remainder (scope label, `target_info`, label
collision handling) will be addressed in a follow-up PR.

* Refs #2748

- 11 new unit tests covering:
  - `build_prom_metric_name` for counters with units, gauges with units,
    `_total` suffix preservation, the `<base>_<unit>_total` ordering fix,
    and the `subtotal` (not a real `_total` suffix) edge case.
  - `has_total_suffix` case-insensitivity.
  - `ucum_to_prometheus_unit` for simple units, bracketed annotations
    (`{packet}/s`), compound rate units (`By/s`, `KiBy/s`, `m/s`), and the
    intentional `By/m` rejection (UCUM `m` is meters, not minutes).
  - `strip_curly_braces` including unbalanced-brace handling.
  - `sanitize_prom_metric_name` underscore collapsing.
- The existing `test_agg_prometheus_mmsc_metrics` was updated to assert the
  new `_milliseconds` unit suffix on its sub-metrics, exercising the new
  path through `agg_prometheus_text`.
- Full crate test suite: 33/33 telemetry lib tests pass.
- `cargo fmt --all -- --check` clean.
- `cargo clippy -p otap-df-admin --all-targets --all-features` clean.

Yes — the Prometheus text exposed at `/metrics` and
`/metrics/aggregate` changes shape:

- Counters now end in `_total` (and only one `_total`, even when the
  source metric name already ended in `_total`).
- Metric names gain a unit suffix derived from the metric's declared UCUM
  unit, e.g. a metric named `request_duration` with unit `ms` is now
  exposed as `request_duration_milliseconds` (or
  `request_duration_milliseconds_total` for counters). MMSC sub-metrics
  pick up the unit suffix as well, so `request_duration_min` becomes
  `request_duration_milliseconds_min`.
- Metrics whose unit is empty or `1` (dimensionless) are unchanged.

Downstream Prometheus scrape consumers that hard-coded the previous
unit-less metric names will need to update their queries. This is the
spec-compliant naming and aligns the admin endpoint with what other
OpenTelemetry → Prometheus exporters produce.


rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1226,10 +1226,249 @@ fn sanitize_prom_metric_name(s: &str) -> String {
...
+        };
+    }
+
+    // Compose: ucum_simple_unit(numerator) + "_per_" + denom_word.
+    // Uses a static lookup to avoid heap allocation in the return value.
+    let numer_word = ucum_simple_unit(numerator)?;
+    COMPOUND_RATE_CACHE
+        .iter()
+        .find(|(n, d, _)| *n == numer_word && *d == denom_word)
+        .map(|(_, _, result)| *result)
```
lalitb, line 1363: Please check this once - I believe This cache makes rate units incomplete. `ucum_simple_unit()` accepts more units than `COMPOUND_RATE_CACHE` covers.

lalitb (approved):
LGTM. One small comment to check on the compound unit cache.

rust/otap-dataflow/crates/admin/src/telemetry.rs
pritishnahar95: Fixed through filling up cache using a macro

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1226,10 +1226,263 @@ fn sanitize_prom_metric_name(s: &str) -> String {
...
+        if ch == '_' {
+            if !prev_underscore {
+                collapsed.push('_');
+            }
+            prev_underscore = true;
+        } else {
+            collapsed.push(ch);
+            prev_underscore = false;
+        }
+    }
```
jmacd, line 1244:
```
 use itertools::Itertools;

 let collapsed: String = out
     .chars()
     .dedup_by(|a, b| *a == '_' && *b == '_')
     .collect();
```

Can we do this instead? (Untested!)

jmacd:
```
 let mut chars = out.chars().peekable();
 let collapsed: String = std::iter::from_fn(|| {
     let ch = chars.next()?;
     if ch == '_' {
         while chars.peek() == Some(&'_') {
             chars.next();
         }
     }
     Some(ch)
 }).collect();
```

pritishnahar95: The current approach was taken to avoid allocations with collect() and still collapse `_` in a single pass

jmacd (approved):

rust/otap-dataflow/crates/admin/src/telemetry.rs
jmacd: Thanks @pritishnahar95

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1226,10 +1226,263 @@ fn sanitize_prom_metric_name(s: &str) -> String {
...
+
+    // Append unit suffix if applicable and not already present.
+    // The check is done byte-wise to avoid allocating a temporary String:
+    // both `name` (post-sanitization) and `unit_word` are ASCII.
+    if let Some(unit_word) = ucum_to_prometheus_unit(unit)
+        && !ends_with_underscore_word(&name, unit_word)
+    {
+        if !name.is_empty() {
+            name.push('_');
```
cijothomas, line 1450: sanitize_prom_metric_name could make `name`  already ending with an underscore `_` ? should we avoid double underscore?

pritishnahar95: Good catch, will address in second PR. See: https://github.com/open-telemetry/otel-arrow/pull/2904

Issue #2458, opened 2026-03-30
Exponential histogram aggregator data structure


jmacd:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/), Go OTel Arrow Receiver/Exporter (go/)

### Is your feature request related to a problem?

This issue covers how we instrument with an exponential histogram in the otap-dataflow codebase. Currently we have a Counter type which is backed by an unsynchronized uint64. The idea is to replace this with Histogram, a small data structure that aggregates into an OTel exponential histogram and supports swapping, copying and aggregation through our metrics path.

### Proposed Solution

I've developed a data structure roughly as described above, https://github.com/jmacd/rust-expohisto (still a work in progress), that is meant to serve as a reference implementation for OpenTelemetry. We will not be able to use this with the OpenTelemetry SDK, therefore progress is blocked by #1950. When we have a pure internal metrics pipeline, we can copy OTel exponential histograms directly as OTLP or OTAP.

### Alternatives Considered

We use an asynchronous API to report measurements to the OTel SDK. The OTel SDK has no API for reporting pre-aggregated histogram values. This is an open question for OpenTelemetry. Since we have our own instrmentation API and internal pipeline, this is not the same question for us. See:

https://github.com/open-telemetry/opentelemetry-specification/issues/2713

### Additional Context

The Rust code mentioned above is inspired by https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/expohisto (also authored by me).

Issue #2428, opened 2026-03-24
Exponential histogram support in OTAP-Dataflow


jmacd:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

Introduce an exponential histogram instrument for observing the dataflow engine.

Related to #2051 Part of #2405.

### Proposed Solution

As an OpenTelemetry contributor, I have designed the Go exponential histogram reference implementation and helped write the data model specification. It is time for an OpenTelemetry Rust exponential histogram reference implementation with the following properties: zero allocation and exact table lookup.

### Alternatives Considered

The OTel-Rust SDK has an exponential histogram data type. As a user, I want to instrument my own code and make telemetry directly from histogram data.

### Additional Context

We would replace the current Mmsc aggregator with 

```
enum Histogram {
    Mmsc(Mmsc),
    Histogram(Box<Histogram>),
}
```

with support for fixed or variable sizes of data.

PR #2904, opened 2026-05-08
fix(admin): add target_info, scope label, and label-collision merging (#2748)

pritishnahar95:
fix(admin): add target_info, scope label, and label-collision merging (#2748)

Per the OpenTelemetry spec for Prometheus exposition:
- Replace the ad-hoc `set="<scope>"` label with `otel_scope_name`. `otel_scope_version` is omitted when empty (MetricsDescriptor does not yet carry a version).
- Emit a `target_info` gauge derived from resource attributes. Pre-render once at server startup and cache as `Arc<str>` on AppState so the hot path is a single `push_str`. Empty attribute map yields an empty block (which the spec mandates).
- Merge label values whose original keys collide after sanitization (joined with `;`) and collapse consecutive `_` in sanitized label keys per spec §Metric Attributes.
- Extract per-metric emission into `emit_scalar_metric`, `emit_mmsc_metric`, and `emit_sample_line` helpers; add `# UNIT` metadata lines for typed metrics. Histogram→gauge fallback is documented at the call site (proper histogram family requires buckets/sum/count which the registry doesn't store yet).

The controller passes `HashMap::new()` for now; wiring real resource attributes is tracked under `TODO(#2748)` and follows once the controller exposes them.

# Change Summary

This is **PR-2 of 2** splitting the original Prometheus text-formatter OTel-spec compliance work (#2748). PR-1 (#2900, merged) handled metric name and unit suffix rules. This PR completes the spec compliance by addressing scope identification, resource identity (`target_info`), and label-key sanitization edge cases.

Highlights:
- **Scope label**: emits `otel_scope_name="<scope>"` instead of the ad-hoc `set="<scope>"` so downstream Prometheus consumers can identify the originating instrumentation scope per the OTel/Prometheus interop spec.
- **`target_info` gauge**: rendered once at admin server startup from the supplied resource attribute map and cached as an `Arc<str>` on `AppState`. Each scrape pays only a single `push_str` — no per-scrape allocation, no formatting, no locking.
- **Label sanitization & collision merging**: keys like `http.method` and `http_method` both sanitize to `http_method`; their values are joined with `;` rather than silently overwriting one another. Consecutive underscores in sanitized keys are collapsed to a single `_`, matching the existing rule for metric names.
- **Hot-path refactor**: per-metric emission split into `emit_scalar_metric`, `emit_mmsc_metric`, and `emit_sample_line` so each formatter (Prometheus / JSON / line-protocol) reuses the same shape. `# UNIT` metadata lines are now emitted for typed metrics.
- **Documented fallback**: histograms still render as a gauge of the count, with an inline comment at the call site explaining that a full `_bucket` / `_sum` / `_count` family requires bucket boundaries the registry does not yet store.

* Part of #2748 (PR-1 merged in #2900; this PR completes the remaining spec-compliance items).

- 45/45 `otap-df-admin` lib tests pass, including:
  - `test_format_prometheus_text_e2e_otel_compliance` — end-to-end fixture asserting `# HELP` / `# TYPE` / `# UNIT` ordering, `otel_scope_name` placement, `target_info` block, and `_total` suffix on counters.
  - `test_sanitize_and_merge_label_pairs_collisions_use_semicolon` — verifies `;`-joined merging of values whose original keys collide after sanitization.
  - `test_sanitize_and_merge_label_pairs_distinct_keys_unchanged` — guards against false-positive merging.
  - `test_sanitize_prom_label_key_collapses_underscores` — verifies the `_+` → `_` rule on label keys.
- `cargo fmt -p otap-df-admin --check` clean.
- `cargo clippy -p otap-df-admin --all-targets -- -D warnings` clean.
- Manually scraped a running admin server against a Prometheus 2.x instance to confirm the output parses without warnings and `target_info` joins correctly via `* on (job, instance) group_left(...) target_info`.

Yes — the `/metrics` endpoint output changes in three ways visible to scrapers:

1. The label `set="<scope>"` is replaced by `otel_scope_name="<scope>"`. Dashboards and alerts that grouped by `set` must be updated.
2. A new `target_info` gauge is emitted (empty block when no resource attributes are configured, which is the current default — the controller still passes `HashMap::new()` pending follow-up).
3. Label keys that previously collided silently are now merged with `;` separators rather than one value overwriting the other. This is a correctness fix; the previous behavior was non-deterministic.

No configuration changes are required. The change is opt-in only in the sense that it affects output of the existing admin Prometheus endpoint that already had to be enabled to be scraped.

### DIFF
--- a/rust/otap-dataflow/crates/admin/src/telemetry.rs
+++ b/rust/otap-dataflow/crates/admin/src/telemetry.rs
@@ -1321,6 +1321,16 @@ fn sanitize_prom_metric_name(s: &str) -> String {
             prev_underscore = false;
         }
     }
+    // Strip a trailing underscore so callers that append unit / `_total`
+    // suffixes don't produce double underscores (e.g. `foo.` → `foo_` →
+    // `foo__bytes`). If stripping leaves an empty string, fall back to
+    // the placeholder name used for fully-invalid inputs.
+    if collapsed.ends_with('_') {
+        let _ = collapsed.pop();
+    }
+    if collapsed.is_empty() {
+        return "metric".to_string();
+    }
     collapsed
 }
 
@@ -3104,6 +3114,29 @@ mod tests {
     #[test]
     fn test_sanitize_prom_metric_name_collapses_underscores() {
         assert_eq!(sanitize_prom_metric_name("foo__bar___baz"), "foo_bar_baz");
+    }
+
+    #[test]
+    fn test_sanitize_prom_metric_name_strips_trailing_underscore() {
+        // A trailing non-alphanumeric character (e.g. `.`, `-`) sanitizes to
+        // `_`. If left in place, downstream callers that append `_<unit>` or
+        // `_total` would produce double underscores (e.g.
+        // `foo.` → `foo_` → `foo__bytes`). `sanitize_prom_metric_name` strips
+        // the trailing `_` so suffix-appending callers don't have to.
+        assert_eq!(sanitize_prom_metric_name("foo."), "foo");
+        assert_eq!(sanitize_prom_metric_name("foo___"), "foo");
+        assert_eq!(sanitize_prom_metric_name("req.count."), "req_count");
+        // After stripping, the unit suffix joins cleanly with a single `_`.
+        assert_eq!(
+            build_prom_metric_name("foo.", "By", Instrument::Gauge),
+            "foo_bytes"
+        );
+        assert_eq!(
+            build_prom_metric_name("req.count.", "By", Instrument::Counter),
+            "req_count_bytes_total"
+        );
+        // All-invalid input still falls back to the placeholder.
+        assert_eq!(sanitize_prom_metric_name("..."), "metric");
     }
 
     // ---------------------------------------------------------------------
--- a/rust/otap-dataflow/crates/admin/src/telemetry.rs
+++ b/rust/otap-dataflow/crates/admin/src/telemetry.rs
@@ -338,9 +338,19 @@ async fn get_metrics(
         }
         OutputFormat::Prometheus => {
             let body = if q.reset {
-                format_prometheus_text(&state.metrics_registry, true, Some(now.timestamp_millis()))
+                format_prometheus_text(
+                    &state.metrics_registry,
+                    true,
+                    Some(now.timestamp_millis()),
+                    &state.target_info,
+                )
             } else {
-                format_prometheus_text(&state.metrics_registry, false, Some(now.timestamp_millis()))
+                format_prometheus_text(
+                    &state.metrics_registry,
+                    false,
+                    Some(now.timestamp_millis()),
+                    &state.target_info,
+                )
             };
             let mut resp = body.into_response();
             // Prometheus text exposition format 0.0.4
@@ -407,7 +417,8 @@ pub async fn get_metrics_aggregate(
             Ok(resp)
         }
         OutputFormat::Prometheus => {
-            let body = agg_prometheus_text(&groups, Some(now.timestamp_millis()));
+            let body =
+                agg_prometheus_text(&groups, Some(now.timestamp_millis()), &state.target_info);
             let mut resp = body.into_response();
             let _ = resp.headers_mut().insert(
                 header::CONTENT_TYPE,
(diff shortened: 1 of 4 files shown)


rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1038,112 +1171,58 @@ fn format_prometheus_text(
...
+        // `otel_scope_version` is emitted only when a non-empty version is
+        // available; the current `MetricsDescriptor` does not carry one, so
+        // the label is omitted entirely (per OTel→Prometheus spec: only
+        // labels with values are emitted).
         let mut base_labels = String::new();
         if !descriptor.name.is_empty() {
             let _ = write!(
                 &mut base_labels,
-                "set=\"{}\"",
+                "otel_scope_name=\"{}\"",
```
lalitb, line 1202: I know this follows the old `set` label pattern, but `otel_scope_name` is different because it is a spec label. If a metric also has an attribute like `otel.scope.name`, that can clean up to the same label name. Since this label is written separately from the merge logic below, we can still emit duplicate `otel_scope_name` labels instead of merging them.

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1038,112 +1171,58 @@ fn format_prometheus_text(
...
         }
-        for (key, value) in attributes.iter_attributes() {
+        // Merge values for keys that collide after sanitization (per
+        // OTel→Prometheus spec). Emission order is unspecified.
+        let merged = sanitize_and_merge_label_pairs(
+            attributes
+                .iter_attributes()
+                .map(|(k, v)| (k, v.to_string_value())),
+        );
+        for (key, value) in &merged {
```
lalitb, line 1216: I believe this makes the label order unstable. The old code sorted attributes before writing them, but now the same metric can print labels in a different order across runs.

Can we sort the merged labels before emitting them? Same pattern looks present in the aggregate formatter too.

lalitb (approved):
Approved from my side, assuming the label ordering / collision handling comments are addressed before merge.

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -688,115 +699,237 @@ fn agg_line_protocol_text(groups: &[AggregateGroup], timestamp_millis: Option<i6
...
+            // Rendering as the Prometheus histogram family
+            // (`_bucket{le=...}`/`_sum`/`_count`) would require fabricating
+            // bucket data we don't have, so we emit a `gauge` reflecting the
+            // raw stored scalar. This is a known limitation: not spec-compliant
+            // for OTel Histograms (the spec mandates the histogram family) and
+            // potentially misleading because the gauge value's meaning depends
+            // on what the producer chose to put in `snapshot_values()`.
+            // Proper handling requires extending `MetricValue` with a variant
+            // carrying buckets/sum/count.
+            Instrument::Histogram => "gauge",
```
jmacd, line 746: FWIW I have a start on this https://github.com/jmacd/rust-expohisto

rust/otap-dataflow/crates/admin/src/telemetry.rs
pritishnahar95: I actually undid the sort for perf reasons in the old code, since it was happening every time metrics were scraped.
@cijothomas @jmacd what do you think?

rust/otap-dataflow/crates/admin/src/telemetry.rs
pritishnahar95: addressed

rust/otap-dataflow/crates/admin/src/telemetry.rs
pritishnahar95: I had pushed a new commit in the larger single PR to undo sorting for perf reasons (preventing extra allocations and iterations), since this is being done for every metric scrape.

@cijothomas @jmacd what do you think?

rust/otap-dataflow/crates/admin/src/telemetry.rs
pritishnahar95: nice, let's create an issue to incorporate this when it's ready :)

jmacd: Will do!  The code there is practically ready. #2428 #2458 
We can't connect pre-aggregated expo-histos with the OTel SDK, so we'll do this after we have an internal telemetry pipeline.

rust/otap-dataflow/crates/admin/src/telemetry.rs
jmacd: I'm not aware of any requirement to sort Prometheus output within a family. Families have to be kept together-- this looks OK.

jmacd (approved):

rust/otap-dataflow/crates/admin/src/telemetry.rs
pritishnahar95: Thanks!

rust/otap-dataflow/crates/admin/src/telemetry.rs
pritishnahar95: Cool, will keep this as is for now

PR #2906, opened 2026-05-08
Fix timeouts in flaky test_no_partitioning UT

wbutler:
Working on a previous merge, I got caught up in a merge validation failure on `test_no_partitioning` in `parquet_exporter`.

Looking at the source, the 1s timeout is defined in `mod.rs` at:

```rust
        test_runtime
            .set_exporter(exporter)
            .run_test(logs_scenario(
                num_rows,
                Instant::now().add(Duration::from_secs(1)),
                Duration::from_secs(1),
            ))
```

The result gets fed into the async block within the `logs_scenario` helper func as `shutdown_timeout`:

```rust
    fn logs_scenario(
        num_rows: usize,
        shutdown_timeout: Instant,
    ) -> impl FnOnce(TestContext<OtapPdata>) -> Pin<Box<dyn Future<Output = ()>>> {
        move |ctx| {
            Box::pin(async move {
                let mut consumer = Consumer::default();
                let otap_batch = consumer
                    .consume_bar(&mut fixtures::create_simple_logs_arrow_record_batches(
                        SimpleDataGenOptions {
                            num_rows,
                            ..Default::default()
                        },
                    ))
                    .unwrap();

                ctx.send_pdata(OtapPdata::new_default(
                    OtapArrowRecords::Logs(from_record_messages(otap_batch).unwrap()).into(),
                ))
                .await
                .expect("Failed to send  logs message");

                ctx.send_shutdown(shutdown_timeout, "test completed")
                    .await
                    .unwrap();
            })
        }
    }
```

But because the `Instant` is determined at the call time of `logs_scenario`, the shutdown timeout clock is already running during setting up the exporter and running the actual test.

## Changes

The fix is to pass in a `Duration` and then properly start the clock when we call `send_shutdown` by converting to an `Instant` at that time. This is consistent with the pattern in other UT's, in the module, like  `test_adaptive_schema_dict_upgrade_write`, `test_metrics`, `test_adaptive_schema_optional_columns`, and others, which all compute the `Instant` at the call time of `send_shutdown`.

This fix also addresses `test_with_partitioning`, which uses the same helper and the same anti-pattern.

## Validation

The following commands run clean:

`cargo check --workspace`
`cargo test -p otap-df-core-nodes -- parquet_exporter`
`cargo clippy -p otap-df-core-nodes --all-targets -- -D warnings`
`cargo fmt --all -- --check`
`cargo xtask quick-check`

## Notes

`cargo check -p otap-df-core-nodes` fails on trying to resolve `ring` in `crypto`. I'm not going after that as non-germane to this fix.


lalitb (approved):
LGTM. Nice catch !

wbutler:
As far as I can tell, this is fully baked and ready for merge. The config validation failure appears unrelated. Feedback, asks, etc. are very welcome.

jmacd:
Thanks @wbutler!

PR #2907, opened 2026-05-08
feat(admin): wire engine resource attributes into target_info (#2748)

### LINKED ISSUE #2748: Admin Prometheus text formatter does not follow OTel OTLP-to-Prometheus translation rules

pritishnahar95:
feat(admin): wire engine resource attributes into target_info (#2748)

Replace the `HashMap::new()` placeholder in the controller with the engine's `telemetry.resource` map so `target_info` carries real resource metadata.

Push the typed `AttributeValue` enum down to `render_target_info` instead of pre-flattening to strings at the API boundary. Stringification now happens once at startup, in the only function that needs the flat form, via a new `AttributeValue::to_string_value()` on the config crate.

Conversion is round-trip lossless: scalars use Display (Ryu for f64), arrays render as bare JSON arrays of the inner Vec (preserving element type and order, avoiding the externally-tagged enum form).

Hot scrape path is unchanged: `target_info` is still pre-rendered once and cached as `Arc<str>`.

Adds tests for scalar formatting, f64 bit-equal round-trip across edge cases (PI, 1e±300, EPSILON, -0.0), bare-JSON array encoding, and an end-to-end `render_target_info` test exercising every variant including escaped JSON arrays in Prometheus label values.

Closes the `TODO(#2748)` for resource attribute wiring.

# Change Summary

This is the **resource-attribute follow-up to #2748**, building on the merged PR-1 (#2900 — name/unit suffix rules) and PR-2 (#2904 — scope label, `target_info` block, label-collision merging). PR-2 introduced the `target_info` gauge but the controller was passing `HashMap::new()` with a `TODO(#2748)` because the resource attributes had not yet been threaded through. This PR closes that TODO.

The change has two parts:

1. **Wiring (1 line of behavior change in the controller):** the admin server is now spawned with `engine_config.engine.telemetry.resource.clone()` instead of the empty placeholder, so any `engine.telemetry.resource` keys declared in YAML now appear as `target_info` labels on every Prometheus scrape.

2. **Type-correctness refactor at the admin API boundary:** previously, `pub async fn run(...)` accepted `HashMap<String, String>`, forcing every caller to pre-flatten typed values. The signature now accepts `HashMap<String, AttributeValue>` (the same enum already defined in `otap_df_config::pipeline::telemetry`). Stringification is moved down into `render_target_info` — the only function that actually needs the flat form. This eliminates a parallel string-only type at the boundary, matches the pattern already used elsewhere in the codebase, and means future callers can pass typed values without first lossy-flattening them.

The new helper `AttributeValue::to_string_value()` is the single source of truth for the typed → string conversion. It is round-trip lossless for every variant:

* `String` — cloned verbatim (no behavioral change vs. the prior code).
* `Bool` — `"true"` / `"false"`.
* `I64` — decimal `Display`.
* `F64` — Rust's `Display`, which uses **Ryu** under the hood: the shortest decimal representation that round-trips back to a bit-equal `f64`. Verified by a dedicated test against `PI`, `1e±300`, `EPSILON`, and `-0.0`.
* `Array` — encoded as a **bare JSON array** of the inner `Vec` (e.g. `["edge","us-west"]`, `[1,2,3]`, `[true,false]`). The inner `Vec` is serialized directly so the output is not the externally-tagged `{"String":[...]}` form that serializing the enum variant would yield. Generic JSON tooling, OTel SDKs, and Prometheus consumers all parse the result without special-casing.

The hot scrape path is unchanged: `render_target_info` is still called exactly once, at admin server startup, and the result is cached as `Arc<str>` on `AppState` and `push_str`'d on every scrape. No per-scrape allocation, no per-scrape conversion, no per-scrape locking.

* Closes #2748 

* New unit tests on `AttributeValue::to_string_value()` in `otap-df-config`.
* New end-to-end test on the renderer in `otap-df-admin`
  (`test_format_prometheus_text_e2e_otel_compliance` extension).
* New regression test for collision-merge ordering in `otap-df-admin`
  (`test_sanitize_and_merge_label_pairs_collision_order_is_lexicographic`),
  covering multiple `Vec` permutations and a `HashMap` input to verify
  the lex-by-original-key ordering required by the OTel→Prometheus spec.
* Comment update on existing `test_sanitize_and_merge_label_pairs_collisions_use_semicolon`.

* Full results:
  * `cargo test -p otap-df-admin --lib` → 49 passed (47 prior + 2 new).
  * `cargo test -p otap-df-config --lib` → all green, 3 new tests added.
  * `cargo clippy -p otap-df-admin --all-targets --no-deps -- -D warnings` → clean.
  * `cargo build -p otap-df-{admin,controller,config}` → clean.

Yes — but additive only. Resource attributes declared under `engine.telemetry.resource` in the YAML config now flow into the `target_info` gauge labels at `/metrics`. Example:

```yaml
engine:
  telemetry:
    resource:
      service.name: "edge-collector"
      service.instance.port: 4317
      deployment.canary: true
      service.tags: ["edge", "us-west"]
```

…now produces (label order is unspecified):

```
# HELP target_info Target metadata
# TYPE target_info gauge
target_info{service_name="edge-collector",service_instance_port="4317",deployment_canary="true",service_tags="[\"edge\",\"us-west\"]"} 1
```

Previously, the `target_info` block was always empty, so any consumer that was relying on its emptiness will see real labels for the first time. No config schema change — the `engine.telemetry.resource` field already existed and was already being consumed by the OTel SDK metrics path; this PR just stops dropping it on the Prometheus path.

### DIFF
--- a/rust/otap-dataflow/crates/admin/src/telemetry.rs
+++ b/rust/otap-dataflow/crates/admin/src/telemetry.rs
@@ -19,6 +19,7 @@ use axum::response::{IntoResponse, Response};
 use axum::routing::get;
 use axum::{Json, Router};
 use otap_df_admin_types::telemetry as api;
+use otap_df_config::pipeline::telemetry::AttributeValue as ResourceAttributeValue;
 use otap_df_telemetry::attributes::{AttributeSetHandler, AttributeValue};
 use otap_df_telemetry::descriptor::{Instrument, MetricValueType, MetricsDescriptor, MetricsField};
 use otap_df_telemetry::event::LogEvent;
@@ -844,7 +845,12 @@ fn emit_sample_line(
 /// empty (per OTel→Prometheus spec, `target_info` is only emitted when there
 /// is metadata to expose). Intended to be called once at server startup; the
 /// resulting string is then prepended verbatim to every Prometheus scrape.
-pub(crate) fn render_target_info(resource_attributes: &HashMap<String, String>) -> String {
+///
+/// Accepts `AttributeValue` (the same type used in the engine telemetry
+/// config) so callers do not need to pre-flatten typed values to strings.
+pub(crate) fn render_target_info(
+    resource_attributes: &HashMap<String, ResourceAttributeValue>,
+) -> String {
     if resource_attributes.is_empty() {
         return String::new();
     }
@@ -854,7 +860,7 @@ pub(crate) fn render_target_info(resource_attributes: &HashMap<String, String>) 
     let merged = sanitize_and_merge_label_pairs(
         resource_attributes
             .iter()
-            .map(|(k, v)| (k.as_str(), v.clone())),
+            .map(|(k, v)| (k.as_str(), v.to_string_value())),
         // No reserved keys: `target_info` is a separate metric line and
         // does not carry `otel_scope_*` labels, so no collision is possible.
         &[],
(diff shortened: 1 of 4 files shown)


rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -854,7 +860,7 @@ pub(crate) fn render_target_info(resource_attributes: &HashMap<String, String>)
     let merged = sanitize_and_merge_label_pairs(
         resource_attributes
             .iter()
-            .map(|(k, v)| (k.as_str(), v.clone())),
+            .map(|(k, v)| (k.as_str(), v.to_string_value())),
```
lalitb, line 863: Small concern: now that real resource attributes are passed here, collisions can actually happen. For example, `service.name` and `service_name` both become the same Prometheus label.

Since this starts from a `HashMap`, the merge order may not be stable. Can we sort by the original key before merging, so the joined value is deterministic and matches the OTel Prometheus rule?

Spec ref: https://opentelemetry.io/docs/specs/otel/compatibility/prometheus_and_openmetrics/

  > OpenTelemetry keys to map to the same Prometheus key. In such cases, the values MUST be concatenated together, separated by `;`, and ordered by the lexicographical order of the original keys.

lalitb (approved):
LGTM. Small comment worth fixing before merge. Nice work !

rust/otap-dataflow/crates/admin/src/telemetry.rs
pritishnahar95: Thanks for pointing this out, good catch!
Added a new commit to fix this.

jmacd (approved):
💯

PR #2914, opened 2026-05-09
Honor the console_fallback setting for engine.observed_state.logging_events send policy

### LINKED ISSUE #2916: Console logging fallback in ITS mode not honored
### Pre-filing checklist

### Component(s)

## Symptom

When the engine is configured with logging providers in ITS mode
(`engine.telemetry.logs.providers.{global,engine}: its`), and the
internal-telemetry pipeline is unable to drain its log channel as
fast as the application emits log events, every dropped event
triggers a `raw_error!("Channel full, dropping observed event",
event = ?event)` call that writes a multi-line debug rendering of the
dropped log record directly to **stderr**, regardless of the engine's
`engine.observed_state.logging_events.console_fallback` setting.

The default for `logging_events.console_fallback` is `false`
([`crates/config/src/observed_state.rs:50`]) — i.e. the engine's
documented default is "do not print to stderr when an internal log
event is dropped". The defaulted value is honored on the
`obs_state_store` reporter path, but the ITS reporter is constructed
separately in `crates/telemetry/src/lib.rs:264,267` with
`SendPolicy::default()`, which has `console_fallback: true`. The
engine config field is silently ignored.

### Steps to Reproduce

See above.

### Expected Behavior

The setting should disable logging the channel-full message.

### Actual Behavior

It logs a lot of noise, observed through experiments.

### OTel-Arrow Version

HEAD

### Environment

any

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

jmacd:
# Change Summary

`console_fallback: false`

would not disable telemetry from logging calls that overflow, due to a default setting.

The correct field `engine.observed_state.logging_events`.

Fixes https://github.com/open-telemetry/otel-arrow/issues/2916

### DIFF
--- a/rust/otap-dataflow/crates/telemetry/src/lib.rs
+++ b/rust/otap-dataflow/crates/telemetry/src/lib.rs
@@ -234,6 +234,7 @@ impl InternalTelemetrySystem {
         config: &TelemetryConfig,
         telemetry_registry: TelemetryRegistryHandle,
         console_async_reporter: Option<ObservedEventReporter>,
+        its_reporter_policy: SendPolicy,
         context_fn: LogContextFn,
         log_tap_handle: Option<log_tap::InternalLogTapHandle>,
     ) -> Result<Self, Error> {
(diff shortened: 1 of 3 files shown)


rust/otap-dataflow/crates/telemetry/src/lib.rs
```diff
@@ -261,10 +262,10 @@ impl InternalTelemetrySystem {
         let (its_reporter, its_settings) = if config.logs.providers.uses_its_provider() {
             let (sender, logs_receiver) = flume::bounded(config.reporting_channel_size);
             let reporter = if let Some(log_tap) = &log_tap_handle {
-                ObservedEventReporter::new(SendPolicy::default(), sender)
+                ObservedEventReporter::new(logging_send_policy.clone(), sender)
```
lalitb, line 265: nit - Could we add one regression test for this path with `ProviderMode::ITS` and `console_fallback: false`?

lalitb (approved):

PR #2919, opened 2026-05-10
feat: Replace static musl build on alpine with glibc based build on gcr distroless

### LINKED ISSUE #2918: Replace static musl build + alpine image with glibc build + distroless gcp image
### Pre-filing checklist

### Component(s)

### Description

Benchmarking has demonstrated a huge performance gap in terms of CPU utilization for many scenarios when using our static musl build with Alpine as compared to a build that links with glibc on something like gcp distroless. For example, here is a comparison of an OTAP/OTLP passthrough benchmark with zstd compression:

 | Protocol | Rate | Recv (logs/s) alpine | Recv (logs/s) distroless | CPU% alpine | CPU% distroless | CPU Δ | RAM MiB alpine | RAM MiB distroless | RAM Δ |
  |-------|-----:|---------------------:|-------------------------:|------------:|----------------:|------:|---------------:|-------------------:|------:|
  | OTLP  | 100k | 100,314 | 101,857 | 14.47 | 8.51  | -41.2% | 9.89  | 11.02 | +11.5% |
  | OTLP  | 200k | 198,891 | 202,524 | 25.58 | 13.52 | -47.1% | 10.18 | 11.75 | +15.4% |
  | OTLP  | 300k | 302,476 | 300,145 | 34.70 | 18.35 | -47.1% | 10.92 | 11.36 | +4.0%  |
  | OTLP  | 400k | 408,467 | 402,204 | 44.06 | 22.55 | -48.8% | 11.07 | 11.83 | +6.9%  |
  | OTAP  | 100k | 100,133 | 99,295  | 14.82 | 6.74  | -54.5% | 13.05 | 13.94 | +6.8%  |
  | OTAP  | 200k | 201,921 | 200,429 | 26.03 | 11.05 | -57.5% | 11.57 | 12.27 | +6.0%  |
  | OTAP  | 300k | 297,408 | 301,856 | 36.32 | 15.49 | -57.4% | 12.65 | 12.77 | +0.9%  |
  | OTAP  | 400k | 407,929 | 396,594 | 49.16 | 20.22 | -58.9% | 12.63 | 13.34 | +5.6%  |

In this benchmark we trade 1-15% higher memory for a 50% reduction in CPU. A clear win for most scenarios. 

### Additional Context

_No response_

JakeDern:
# Change Summary

This PR changes our standard df_engine image from being a static build based on musl + alpine to being based on glibc on gcr distroless.

Part of this change requires updates to the mount paths for orchestrator config files as the home directory is now `nonroot` instead of `dataflow`.

* Closes #2918

To test this, I did local runs of all the nightly/continuous/comparison dashboard suites using the new image + mount path changes.

I did my best to test cross compiling with the new targets for ARM as well:

image

Yes, the runtime image has been changed for the repo dockerfile and that comes along with some new expectations for mount paths.


lquerel (approved):
LGTM

.github/workflows/dataflow-engine-binary-size.yml
```diff
@@ -53,7 +53,7 @@ jobs:
 
       - name: Get binary size and create report
         run: |
-          SIZE=$(docker run --rm --entrypoint stat df_engine:${{ matrix.platform.name }} -c %s /home/dataflow/df_engine)
+          SIZE=$(docker run --rm --entrypoint stat df_engine:${{ matrix.platform.name }} -c %s /home/nonroot/df_engine)
```
lalitb: I believe `stat` will not be available in the distroless runtime image. Should we instead create a container from the image, copy `/home/nonroot/df_engine` out, and run `stat` on the host?

lalitb: Another option is to measure the binary size in the build stage before copying it into the distroless runtime image, but copying it out would keep this workflow checking the final image contents.

lalitb (approved):
Nice work on this. I authored #2214, and I think this is the better fix overall. #2214 fixed the immediate musl build issue. This PR avoids that problem completely by moving the image to glibc + distroless, which feels cleaner long term.

I left one small comment about the binary size workflow, but overall this direction looks good to me.

jmacd (approved):

rust/otap-dataflow/Dockerfile
```diff
@@ -37,21 +37,11 @@ ENV FEATURES=${FEATURES}
 # Build dataflow engine
 RUN ./cross-arch-build.sh
 
-# The runtime image
-FROM docker.io/alpine:3.23@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
+# Runtime image
+FROM gcr.io/distroless/cc-debian13:nonroot
```
ThomsonTan: Pin the runtime image by digest?

.github/workflows/dataflow-engine-binary-size.yml
JakeDern: Good call - I changed it to create a container and then use `docker cp` with `wc -c`

rust/otap-dataflow/Dockerfile
JakeDern: Done

albertlockett (approved):

JakeDern:
I ran a little experiment and indeed after this PR we use a huge amount of memory to build compared to before - Now the question is why:

```
  ┌─────────────────────────┬───────────┬──────────────────┐
  │         Branch          │ Wall time │ Peak memory used │
  ├─────────────────────────┼───────────┼──────────────────┤
  │ main (9a664bf71)        │ 6m 30s    │ 4.85 GiB         │
  ├─────────────────────────┼───────────┼──────────────────┤
  │ glibc-image (1165f5dbf) │ 8m 22s    │ 18.51 GiB        │
  └─────────────────────────┴───────────┴──────────────────┘
```

lalitb:
> I ran a little experiment and indeed after this PR we use a huge amount of memory to build compared to before - Now the question is why:
> 
> ```
>   ┌─────────────────────────┬───────────┬──────────────────┐
>   │         Branch          │ Wall time │ Peak memory used │
>   ├─────────────────────────┼───────────┼──────────────────┤
>   │ main (9a664bf71)        │ 6m 30s    │ 4.85 GiB         │
>   ├─────────────────────────┼───────────┼──────────────────┤
>   │ glibc-image (1165f5dbf) │ 8m 22s    │ 18.51 GiB        │
>   └─────────────────────────┴───────────┴──────────────────┘
> ```

My wild guess is that the final link step is more memory hungry for glibc as compared to musl. Probably can validate by adding temporary memory sampling around the `cargo build --release` step and checking whether the peak happens near the final `rustc`/linker invocation.

JakeDern:
> > I ran a little experiment and indeed after this PR we use a huge amount of memory to build compared to before - Now the question is why:
> > ```
> >   ┌─────────────────────────┬───────────┬──────────────────┐
> >   │         Branch          │ Wall time │ Peak memory used │
> >   ├─────────────────────────┼───────────┼──────────────────┤
> >   │ main (9a664bf71)        │ 6m 30s    │ 4.85 GiB         │
> >   ├─────────────────────────┼───────────┼──────────────────┤
> >   │ glibc-image (1165f5dbf) │ 8m 22s    │ 18.51 GiB        │
> >   └─────────────────────────┴───────────┴──────────────────┘
> > ```
> 
> My wild guess is that the final link step is more memory hungry for glibc as compared to musl. Probably can validate by adding temporary memory sampling around the `cargo build --release` step and checking whether the peak happens near the final `rustc`/linker invocation.

Removing the config.toml overrides for the linker brought it back down. I kept them originally and changed them to `-gnu` because I wasn't sure why they were in there, but then took them out as an experiment which was successful. My guess is that specifying those overrides clobbers more settings/flags than it appears and that somehow influences the memory usage during link time

PR #2930, opened 2026-05-11
ETW receiver with control path and session

swashtek:
# Change Summary

ETW receiver initial PR

- Control path hanges
- ETW session to consume using one-collect git commit

* Closes #NNN

Validation using cargo check, run etc. 
Local validation with ETW console yaml

`cargo run --features etw-receiver -- -c configs/etw-console.yaml`

N/A


rust/otap-dataflow/Cargo.toml
```diff
@@ -216,6 +215,7 @@ xxhash-rust = { version = "0.8", features = ["xxh3"] }
 zip = "=8.6.0"
 byte-unit = { version = "5.2.0", features = ["serde"] }
 cpu-time = "1.0.0"
+one_collect = { git = "https://github.com/swashtek/one-collect.git", rev = "c839578" }
```
swashtek: @utpilla Setting up PR WIP as I should complete the change in main in one-collect before merging this PR.

rust/otap-dataflow/configs/etw-console.yaml
```diff
@@ -0,0 +1,43 @@
...
+            config:
+              session_name: "OtelArrowETW"
+              providers:
+                # Microsoft-Windows-Kernel-Process
+                # - guid: "22fb2cd6-0e7b-422b-a0c7-2fad1fd0e716"
+                #  level: information
+                # Windows DNS Client (generates frequent events)
+                # - guid: "1c95126e-7eea-49a9-a3fe-a378b03ddb4d"
+                #  level: verbose
+                # RusticDemoProvider (custom TraceLogging provider)
```
cijothomas: lets remove this and keep one that works by default. The above ones are good to keep!

rust/otap-dataflow/configs/etw-console.yaml
```diff
@@ -0,0 +1,43 @@
+# ETW → Console pipeline
+#
+# Subscribes to Windows ETW providers and prints events to the console.
+#
+# Usage (run as Administrator for ETW access):
+#   cargo run -- -c configs/etw-console.yaml
+#
+# The receiver captures raw ETW events from the configured providers
+# and logs a summary to stdout. Full Arrow encoding is a TODO.
```
cijothomas:
```suggestion
# and emits summary as an internal log/event. Full Arrow encoding is a TODO.
```

rust/otap-dataflow/crates/core-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,306 @@
...
+        Error::InternalError {
+            message: format!("invalid GUID string '{s}': {e}"),
+        }
+    })?;
+
+    Ok(Guid::from_u128(val))
+}
+
+/// Map the receiver's [`TraceLevel`] to the `one_collect` ETW level constant.
+fn trace_level_to_etw(level: &TraceLevel) -> u8 {
```
cijothomas: const?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,306 @@
...
+            let level = trace_level_to_etw(&p.level);
+            Ok((guid, level, p.keywords))
+        })
+        .collect::<Result<Vec<_>, Error>>()?;
+
+    // Shared stop flag between the receiver and the session thread.
+    let stop_flag = Arc::new(AtomicBool::new(false));
+    let stop_flag_clone = stop_flag.clone();
+
+    let session_name = config.session_name.clone();
```
cijothomas: if a session name is left unclosed (eg: crash a previous run without clean exit), will it linger around? should the new start try to clean up the previous?

rust/otap-dataflow/crates/core-nodes/src/receivers/mod.rs
```diff
@@ -20,5 +20,5 @@ pub mod otap_receiver;
 /// OTLP receiver.
 pub mod otlp_receiver;
 
-/// Host metrics receiver.
-pub mod host_metrics_receiver;
+/// ETW (Event Tracing for Windows) receiver.
+pub mod etw_receiver;
```
utpilla: 1. Let's add the ETW receiver to contrib node instead of core node.
2. Since the receiver is Windows only, let's compile the entire receiver code only if the `target_os` is Windows. You could check the user_events receiver for reference. It's a Linux-only receiver: https://github.com/open-telemetry/otel-arrow/pull/2787/changes#diff-32b7335bd1be4a58766e6ac8549ec92c01bfe9faa9d1795113fd9a230af37a0bR5-R6

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
swashtek: Great catch, one-collect's `TraceSession::start()` already handles it. It stops any existing session with the same name before creating a new one, so lingering sessions from crash/without clean exit are automatically cleaned up on next start.

rust/otap-dataflow/crates/core-nodes/src/receivers/mod.rs
swashtek: Great, thanks for pointing that out. 
I have moved the ETW receiver to contrib-nodes.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+        ))
+    },
+    wiring_contract: otap_df_engine::wiring_contract::WiringContract::UNRESTRICTED,
+    validate_config: otap_df_config::validation::validate_typed_config::<Config>,
+};
+
+// ── Receiver trait implementation ────────────────────────────────────────────
+
+#[async_trait(?Send)]
+impl local::Receiver<OtapPdata> for EtwReceiver {
```
utpilla: Let's implement the fanout in this PR. The receiver is registered with `ReceiverWrapper::local`, so the engine [calls](https://github.com/open-telemetry/otel-arrow/blob/a9e6735caf596fcd1c903efa5004b933d0b18593/rust/otap-dataflow/crates/engine/src/lib.rs#L1542) `factory.create` once per pinned core. `start_etw_session` runs from `start()`, so each core spins up its own session under the same `session_name`. `StartTrace` rejects everyone after the first with `ERROR_ALREADY_EXISTS`, and the failure is currently silent: the error from `parse_until` is captured but not surfaced, so the affected cores' channels just close with no log line. The example config uses `core_count: 1`, which hides this; bumping to 2+ breaks the receiver on every core except one.

```text
Current (broken on multi-core):

  core A                core B                core C
  start_etw_session     start_etw_session     start_etw_session
       v                     v                     v
  StartTrace OK         ALREADY_EXISTS        ALREADY_EXISTS
                        (silently swallowed)  (silently swallowed)
```

## Target shape: one session, N channels

ETW gives exactly one `ProcessTrace` callback thread per session, and no per-CPU buffers like Linux `user_events`. To match the engine's per-core receiver model the session should be a process-global singleton with N independent consumer ends (one per receiver instance) and round-robin fan-out from the callback thread.

```text
ProcessTrace OS thread  (lazily spawned on first factory call)
callback: txs[next].try_send(data); next = (next + 1) % N
      |          |          |
    tx[0]      tx[1]      tx[2]
      v          v          v
     mpsc       mpsc       mpsc
      v          v          v
 +--------+ +--------+ +--------+
 | core A | | core B | | core C |
 | rx[0]  | | rx[1]  | | rx[2]  |
 +--------+ +--------+ +--------+
```

## Here are some suggested changes:

- **Session creation moves to factory time.** Built once on first arrival into `factory.create`, not in `start()`.
- **N consumers handed out, one per per-core receiver.** `num_cores` from `PipelineContext` is identical across calls, so the pop count matches the channel count.
- **Round-robin in the callback.** Per-core ring full only drops events bound for that core.
- **Session lives until process exit.** Drop `SessionHandle` + `AtomicBool` plumbing;
- **`Mutex<Option<_>>` over `LazyLock`/`OnceLock`.** Init is fallible (GUID parse, thread spawn) and state needs post-init mutation, so init-once primitives don't fit without an extra `Mutex` layer anyway.

Sketch of the moving pieces:

```rust
// session.rs
static SESSION: Mutex<Option<Vec<Receiver<EtwEventData>>>> = Mutex::new(None);

pub fn take_consumer(cfg: &Config, num_cores: usize)
    -> Result<Receiver<EtwEventData>, Error>
{
    let mut g = SESSION.lock().expect("poisoned");
    if g.is_none() {
        let (txs, rxs): (Vec<_>, Vec<_>) =
            (0..num_cores).map(|_| mpsc::channel(EVENT_CHANNEL_CAPACITY)).unzip();
        spawn_etw_thread(cfg, txs)?;     // moves all N senders into the callback
        *g = Some(rxs);
    }
    g.as_mut().unwrap().pop().ok_or(...)
}

// callback (inside spawn_etw_thread)
let mut next = 0usize;
session.set_raw_event_callback(move |event| {
    let _ = txs[next].try_send(build_event_data(event));
    next = (next + 1) % txs.len();
});

// mod.rs - runs at factory time, not start time
fn from_config(pipeline: PipelineContext, value: &Value) -> Result<Self, _> {
    let cfg = parse_and_validate(value)?;
    let event_rx = session::take_consumer(&cfg, pipeline.num_cores())?;
    Ok(Self { config: cfg, metrics, event_rx })
}
```

TDH decoding and Arrow batch construction would anyway be a follow up but the fan-out work would serve as a good base for that work to build on.

swashtek: Yes, thank you so much for detailed information. 
Added multi core fan out and set core count to 4 for now. Lmk your thoughts.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+    /// process-global ETW session.
+    ///
+    /// Called at **factory time** (once per allocated core).  The first call
+    /// lazily initialises the singleton session and spawns the `ProcessTrace`
+    /// thread.
+    fn from_config(
+        pipeline: PipelineContext,
+        config: &Value,
+    ) -> Result<Self, otap_df_config::error::Error> {
+        let cfg: Config = serde_json::from_value(config.clone()).map_err(|e| {
```
utpilla: Two small things in `mod.rs` / `session.rs` that I think are worth tidying up.

## 1. Pull the validation rules out of `from_config`

Right now `from_config` is doing three different jobs in one block: deserializing, checking the two domain rules, and acquiring the channel from the session. The domain checks would read better as a method on `Config`:

```rust
impl Config {
    fn validate(&self) -> Result<(), otap_df_config::error::Error> {
        if self.providers.is_empty() { return Err(/* ... */); }
        for (i, p) in self.providers.iter().enumerate() {
            match (&p.name, &p.guid) {
                (Some(_), Some(_)) => return Err(/* xor */),
                (None, None)       => return Err(/* required */),
                _ => {}
            }
        }
        Ok(())
    }
}
```

and `from_config` becomes:

```rust
let cfg: Config = serde_json::from_value(config.clone())?;
cfg.validate()?;
let event_rx = session::take_consumer(&cfg, num_cores)?;
```

The reason I'd push for this:

- The rules are now unit-testable without standing up a pipeline. `#[test] fn name_xor_guid()` is a one-liner. Today you'd need a YAML harness to hit those branches.
- When someone adds the next rule (`session_name` non-empty, max providers, whatever), there's an obvious place for it. Right now it would get bolted onto the loop in `from_config` next to the unrelated channel-acquisition code.
- `from_config` reads as "parse, check, acquire" instead of mixing all three.

## 2. Drop the duplicate `(None, None)` check in `session::resolve_provider_guid`

That check is also done in `from_config`, so by the time `resolve_provider_guid` runs the case is already impossible:

```rust
fn resolve_provider_guid(cfg: &ProviderConfig) -> Result<Guid, Error> {
    if let Some(guid_str) = &cfg.guid { return parse_guid(guid_str); }
    if let Some(name) = &cfg.name     { return Err(/* not implemented */); }
    Err(Error::InternalError {                                     // duplicate
        message: "provider must specify either 'name' or 'guid'".to_string(),
    })
}
```

`unreachable!("validated upstream")` is a more honest spelling.

Same goes for the `(Some, Some)` case: today `resolve_provider_guid` silently picks `guid` and ignores `name`, which would quietly mask a bug if the upstream check ever drifted. Worth turning into a `debug_assert!` or just removing once cleanup 1 is in.

The two functions currently agree on the rules by accident; this pins down the contract so the session layer is allowed to assume validation has happened.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+/// events into OTAP Arrow log records.
+///
+/// Each per-core instance holds its own `event_rx` obtained from the
+/// process-global singleton session at factory time.
+struct EtwReceiver {
+    config: Config,
+    metrics: Rc<RefCell<MetricSet<EtwReceiverMetrics>>>,
+    /// Per-core consumer channel from the singleton ETW session.
+    /// Wrapped in `Option` so it can be moved out in `start()`.
+    event_rx: Option<tokio::sync::mpsc::Receiver<EtwEventData>>,
```
utpilla: Could we drop the `Option` around `event_rx`? An `EtwReceiver` without a channel isn't a meaningful state. There's nothing it can do. We shouldn't let the type represent it. Today it can: `event_rx: Option<...>` plus `take()` in `start` permits a `None` variant whose only purpose is producing the unreachable `"duplicate start?"` `InternalError`. `start` consumes `Box<Self>`, so the "second start" it guards against can't happen; the `Option` exists purely to dodge a borrow conflict between `&self` on `run_event_loop` and `&mut event_rx` as its argument.

Make `run_event_loop` take `&mut self` and read `self.event_rx` directly:

```rust
struct EtwReceiver {
    config: Config,
    metrics: Rc<RefCell<MetricSet<EtwReceiverMetrics>>>,
    event_rx: tokio::sync::mpsc::Receiver<EtwEventData>,  // no Option
}

async fn run_event_loop(
    &mut self,
    ctrl_chan: &mut local::ControlChannel<OtapPdata>,
    effect_handler: &local::EffectHandler<OtapPdata>,
    telemetry_timer_handle: TelemetryTimerCancelHandle<OtapPdata>,
    // ... self.event_rx.recv() inside the select! ...
}
```

This lets the struct's type matches reality and the unreachable `InternalError` branch goes away, `run_event_loop` loses an argument. `metrics` is `Rc<RefCell<…>>` so `self.metrics.borrow_mut()` still works inside `&mut self`; field-level disjoint borrows make `&mut self.event_rx` in the `select!` arm fine.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+
+        loop {
+            tokio::select! {
+                biased; // Prioritise control messages over data.
+
+                ctrl_msg = ctrl_chan.recv() => {
+                    match ctrl_msg {
+                        Ok(NodeControlMsg::DrainIngress { deadline, .. }) => {
+                            let _ = telemetry_timer_handle.cancel().await;
+                            effect_handler.notify_receiver_drained().await?;
```
utpilla: Add a TODO here so that we don't forget to flush the data before notifying.

```suggestion
                            // TODO: drain buffered events before notifying
                            effect_handler.notify_receiver_drained().await?;
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,353 @@
...
+/// so exactly `num_cores` calls are expected.
+///
+/// # Errors
+///
+/// Returns an error if:
+/// - Provider GUID parsing fails (first call only).
+/// - The ETW session thread cannot be spawned (first call only).
+/// - All consumers have already been handed out (more calls than `num_cores`).
+/// - The session lock is poisoned (indicates a prior panic).
+pub fn take_consumer(
```
utpilla: Need not be public.

```suggestion
pub(super) fn take_consumer(
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+
+    /// Trace level filter. Defaults to `information`.
+    #[serde(default)]
+    pub level: TraceLevel,
+
+    /// Optional keywords bitmask to further filter events.
+    /// When omitted, all keywords are matched.
+    #[serde(default)]
+    pub keywords: Option<u64>,
+}
```
utpilla: These could be marked private along with the `Config` struct.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+
+impl ProviderConfig {
+    /// Returns the display label for this provider (for logging and error messages).
+    pub fn display_id(&self) -> &str {
```
utpilla: Is this method needed?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
utpilla: I think `subscribe` would be a better name for this method.


```suggestion
pub fn subscribe(
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,353 @@
...
+                    enabler.ensure_keyword(*kw);
+                }
+            }
+
+            // Round-robin index, atomically updated from the callback.
+            // Using `AtomicUsize` even though the callback runs on a single
+            // thread because the closure is `Fn` (not `FnMut`); interior
+            // mutability via atomics avoids requiring a `Mutex` in the
+            // hot path.
+            let next = AtomicUsize::new(0);
```
utpilla: The closure is `FnMut` so we don't need an `AtomicUsize`.

https://github.com/swashtek/one-collect/blob/4657c633789d171d6652e45b63ec0e3399a304b2/one_collect/src/etw/mod.rs#L432-L440

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,353 @@
...
+            // process shutdown.  When it exits, all senders are dropped,
+            // closing the channels and signalling the async receiver loops.
+        })
+        .map_err(|e| Error::InternalError {
+            message: format!("failed to spawn ETW session thread: {e}"),
+        })?;
+
+    // The JoinHandle is intentionally leaked — the session thread runs for
+    // the lifetime of the process.
+    drop(_join_handle);
```
utpilla: `drop(_join_handle)` doesn't leak anything. Dropping a `JoinHandle` detaches the thread (only `mem::forget` would actually leak it). And the _-prefixed binding would fall out of scope on the next line anyway, so the explicit `drop(...)` is a no-op. The whole block collapses to `let _ = std::thread::Builder::new()...spawn(...).map_err(...)?;` adn the comment can be updated to "Detach the session thread; it runs for the lifetime of the process."

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,353 @@
...
+                "provider name resolution is not yet implemented; \
+                 please specify a GUID instead of name '{name}'. \
+                 You can find a provider's GUID via `logman query providers \"{name}\"`"
+            ),
+        });
+    }
+
+    Err(Error::InternalError {
+        message: "provider must specify either 'name' or 'guid'".to_string(),
+    })
```
utpilla: Both `resolve_provider_guid` and `parse_guid` methods are called to validate the configuration. These should be returning `InvalidUserConfig` error instead `InternalError`. `InternalError` are meant to be used for engine issues.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// ── GUID parsing ─────────────────────────────────────────────────────────────
+
+/// Parse a GUID string in the standard `xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`
+/// format (with or without surrounding braces/hyphens) into a
+/// [`one_collect::Guid`].
+fn parse_guid(s: &str) -> Result<Guid, Error> {
+    // Strip optional braces.
+    let s = s.trim().trim_start_matches('{').trim_end_matches('}');
```
utpilla: Do we want to allow braces?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,353 @@
...
+
+    // Collect only hex digits.
+    let hex: String = s.chars().filter(|c| c.is_ascii_hexdigit()).collect();
```
utpilla: This would allow input to have any non-hex character anywhere.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
swashtek: This is great, I didn't realize that making entire `self` mutable solves the problem. 
Out of curiosity, will this be an issue if the method is called concurrently?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,353 @@
...
+    }
+
+    guard
+        .as_mut()
+        .and_then(|rxs| rxs.pop())
+        .ok_or_else(|| Error::InternalError {
+            message: "all ETW consumer channels have been handed out; \
+                      num_cores mismatch or duplicate factory calls"
+                .to_string(),
+        })
```
utpilla: This `ok_or_else` fires for two distinct conditions, conflated into one message:

- `guard.is_none()`: never initialized (logically can't happen here because we just initialized above, but the chain doesn't know that).
- `pool.pop()` returned `None`: pool empty.

Pulling them apart gives clearer diagnostics:

```rust
let pool = guard.as_mut().expect("initialised above");
pool.pop().ok_or_else(|| Error::InternalError {
    message: format!(
        "ETW consumer pool exhausted (capacity {}); engine requested more receivers than num_cores",
        // optional: track and report initial size
    ),
})
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,353 @@
...
+pub fn take_consumer(
+    config: &Config,
+    num_cores: usize,
+) -> Result<mpsc::Receiver<EtwEventData>, Error> {
+    let mut guard = SESSION.lock().map_err(|e| Error::InternalError {
+        message: format!("ETW session lock poisoned: {e}"),
+    })?;
+
+    if guard.is_none() {
+        // First call — initialise the session.
```
utpilla: Use US English to be consistent.

```suggestion
        // First call — initialize the session.
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,353 @@
...
+    #[test]
+    fn parse_guid_with_braces() {
+        let guid = parse_guid("{22fb2cd6-0e7b-422b-a0c7-2fad1fd0e716}").expect("valid GUID");
+        assert_eq!(guid.data1, 0x22fb2cd6);
+    }
+
+    #[test]
+    fn parse_guid_uppercase() {
+        let guid = parse_guid("22FB2CD6-0E7B-422B-A0C7-2FAD1FD0E716").expect("valid GUID");
+        assert_eq!(guid.data1, 0x22fb2cd6);
```
utpilla: Add assertions for `data2`, `data3`, and `data4` as well.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -0,0 +1,416 @@
...
+
+        let num_cores = pipeline.num_cores();
+        let metrics = pipeline.register_metrics::<EtwReceiverMetrics>();
+
+        // Acquire this core's consumer channel from the singleton session.
+        // The first call initialises the session; subsequent calls pop from
+        // the pre-allocated pool.
+        let event_rx =
+            session::take_consumer(&cfg, num_cores).map_err(|e| {
+                otap_df_config::error::Error::InvalidUserConfig {
```
utpilla: `InternalError` seems better suited here.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
utpilla: > Out of curiosity, will this be an issue if the method is called concurrently?

No. `&mut self` is an exclusive borrow, so the compiler won't let two callers hold it at once; concurrent calls on the same instance aren't representable.

And there's no shared instance to race on: the engine is thread-per-core shared-nothing, so each core constructs its own `EtwReceiver` and runs `start` on that instance on its own `current_thread` Tokio runtime.

`start` already takes `mut self: Box<Self>`, which consumes the receiver, so "two starts on one instance" is also impossible. The `"duplicate start?" InternalError` was guarding an invariant the type system already enforces; with `&mut self` it just moves fully to compile time, and the branch becomes dead code.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
swashtek: Removing it.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
swashtek: Looks like this is stale, removing it.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
swashtek: Generally when we copy & paste GUID from windows tools, the braces are copied. This will be to make that case handle. To make it strict on format of input, I will remove it.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
swashtek: There is no `InternalError` in `otap_df_config` , are you referring `InternalError` from `otap_df_engine` ?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,423 @@
...
+                let txs = txs.clone();
+                let mut wide_event = one_collect::event::Event::new(0, "otap_wide".to_string());
+                {
+                    let ext = wide_event.extension_mut();
+                    *ext.provider_mut() = *guid;
+                    *ext.level_mut() = *level;
+                    *ext.keyword_mut() = keywords.unwrap_or(0);
+                }
+
+                wide_event.add_callback({
```
utpilla: It looks like you have changed the callback method registration code in the last commit. I'm okay with the refactor from `set_raw_event_callback` to per-provider `wide_event` registrations. 

However, these changes replaced the single round-robin counter with one counter per provider, which could underutilize cores in startup. Each callback closure now captures its own `local_next`. In long-run steady state both schemes give uniform per-core load. The regression is at startup: every provider's counter is 0, so the first event of every provider piles on core 0, the second on core 1, and so on. With M providers all stepping in lock-step from 0, the early cores are oversubscribed and the late cores stay idle until each provider's counter has walked past them.

Example: 4 cores, 2 providers each emitting 3 events (`A1, B1, A2, B2, A3, B3`):

```text
per-provider counters:           shared counter:
  A1 -> A=0 -> core 0              A1 -> 0 -> core 0
  B1 -> B=0 -> core 0              B1 -> 1 -> core 1
  A2 -> A=1 -> core 1              A2 -> 2 -> core 2
  B2 -> B=1 -> core 1              B2 -> 3 -> core 3
  A3 -> A=2 -> core 2              A3 -> 4 -> core 0
  B3 -> B=2 -> core 2              B3 -> 5 -> core 1
  result: 2, 2, 2, 0               result: 2, 2, 1, 1
```

Same pattern recurs any time multiple providers' counters happen to be equal `mod num_txs` and they all burst together — most reliably at session startup, when every counter is exactly 0. The original `set_raw_event_callback` design had one shared counter so the distribution was uniform.

We can fix this by sharing the counter across callbacks. `one_collect` requires one `Event` registration per provider, so we can't collapse to a single closure. We can share the counter (and the `Vec<Sender>`) across closures via `Rc<Cell<usize>>` / `Rc<Vec<...>>`. All callbacks dispatch on the single `ProcessTrace` thread, so `Cell` is safe — no atomics, no locking.

```rust
use std::cell::Cell;
use std::rc::Rc;

let next: Rc<Cell<usize>> = Rc::new(Cell::new(0));
let txs:  Rc<Vec<mpsc::Sender<EtwEventData>>> = Rc::new(txs);

for (guid, level, keywords) in &resolved_providers {
    let mut wide_event = one_collect::event::Event::new(0, "otap_wide".into());
    {
        let ext = wide_event.extension_mut();
        *ext.provider_mut() = *guid;
        *ext.level_mut()    = *level;
        *ext.keyword_mut()  = keywords.unwrap_or(0);
    }

    let ancillary = ancillary.clone();
    let next      = Rc::clone(&next);
    let txs       = Rc::clone(&txs);

    wide_event.add_callback(move |_event_data| {
        let anc = ancillary.borrow();
        let data = EtwEventData {
            provider_id: anc.provider().to_bytes(),
            timestamp:   anc.time(),
            process_id:  anc.pid(),
            thread_id:   anc.tid(),
            // TODO: populate event_id/opcode/level/keywords/version once
            // WindowsEventExtension exposes EVENT_DESCRIPTOR.
            event_id: 0, opcode: 0, version: 0, level: 0, keywords: 0,
        };
        drop(anc);

        let i = next.get();
        next.set(i.wrapping_add(1));
        let _ = txs[i % txs.len()].try_send(data);
        Ok(())
    });

    session.add_event(wide_event, None);
}
```

Also collapses N `Vec<Sender>` clones to N refcount handles, as a side benefit.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
utpilla: I just realized `from_confg` would only return `otap_df_config` `Error`.

Let's leave this for now. We could revisit this later.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,423 @@
...
+//!  | core 0 | | core 1 | | core 2 |
+//!  | rx[0]  | | rx[1]  | | rx[2]  |
+//!  +--------+ +--------+ +--------+
+//! ```
+//!
+//! ## Integration with `one_collect`
+//!
+//! Instead of using the low-level `set_raw_event_callback` (which bypasses
+//! `one_collect`'s event routing), we use the **provider-wide event** mechanism
+//! via [`EtwSession::add_wide_event`].  This registers a catch-all handler for
```
utpilla: This API is not being called in the code.

swashtek: Yes this is not existing functionality in one-collect, changed the comment to `add_event` which is currently used.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
swashtek: Added the change. Thanks for catching the regression.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,422 @@
...
+        )));
+    }
+
+    unreachable!("validated upstream: provider must specify either 'name' or 'guid'")
+}
+
+// ── Singleton session state ──────────────────────────────────────────────────
+
+/// Process-global session state.  Initialized on the first call to
+/// [`take_consumer`]; subsequent calls pop one receiver from the pool.
```
utpilla:
```suggestion
/// [`subscribe`]; subsequent calls pop one receiver from the pool.
```

utpilla (approved):
Thanks for working on this @swashtek!

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -0,0 +1,517 @@
...
+                            return Ok(TerminalState::new(deadline, [snapshot]));
+                        }
+                        Ok(NodeControlMsg::Shutdown { deadline, .. }) => {
+                            let _ = telemetry_timer_handle.cancel().await;
+                            otel_info!(
+                                "etw_receiver.shutdown",
+                                message = "ETW receiver shutting down",
+                            );
+                            let snapshot = self.metrics.borrow().snapshot();
+                            return Ok(TerminalState::new(deadline, [snapshot]));
```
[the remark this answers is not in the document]
utpilla: @swashtek Created issue #3033 to track this. This can be done in a follow up PR.

utpilla:
@open-telemetry/arrow-approvers The CI Check for `validate-configs` is failing for `etw-console.yaml` because ETW receiver is a Windows-only component and would not work on Linux runners. I have created an issue #3041 to track the changes needed to CI workflow to accommodate this. We could address them in a follow-up PR.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,410 @@
...
+
+// ── Singleton session state ──────────────────────────────────────────────────
+
+/// Process-global session state.  Initialized on the first call to
+/// [`subscribe`]; subsequent calls pop one receiver from the pool.
+///
+/// We use `Mutex<Option<Vec<…>>>` rather than `OnceLock` / `LazyLock` because:
+/// - Initialization is fallible (GUID parsing, thread spawn).
+/// - We need post-init mutation (`Vec::pop`).
+static SESSION: Mutex<Option<Vec<mpsc::Receiver<EtwEventData>>>> = Mutex::new(None);
```
lalitb, line 191: Question: should this singleton track the config it was initialized with? Since `SESSION` is process-global and the pool is consumed with `pop()`, a later ETW receiver or restart in the same process may not behave as expected. If full scoped/resettable state is follow-up work, a clear error for mismatched/additional subscriptions would still make this safer.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,410 @@
...
+
+                    Ok(())
+                });
+
+                session.add_event(wide_event, None);
+            }
+
+            // `parse_until` blocks on `ProcessTrace`.  We never signal stop,
+            // so the session runs until the process exits.
+            let _result = session.parse_until(&session_name, || false);
```
lalitb, line 302: Can we avoid dropping this result? If ETW fails to start or exits because of permissions/session/provider errors, the receiver currently looks initialized and later just sees a closed channel. At minimum this should log the error, and ideally startup failures should be surfaced before the receiver is considered ready.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -0,0 +1,410 @@
...
+
+                    // Drop the borrow before sending.
+                    drop(anc);
+
+                    let i = next.get();
+                    next.set(i.wrapping_add(1));
+
+                    // Best-effort send; if this core's channel is full,
+                    // drop the event for that core only.
+                    let _ = txs[i % txs.len()].try_send(data);
```
lalitb, line 292: This seems to drops the ETW event silently when the per-core channel is full. Since this is the main backpressure/loss path in the scaffold, can we record a drop counter or at least log it in a rate-limited way? Also the comment says “for that core only”, but the event is assigned to one core, so this drops it from the pipeline.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
utpilla: @lalitb A lot of this (the APIs being called) might be changed when TDH decoding support is added to `one-collect`.

@swashtek For now, you could add a TODO here to log the result so that we remember to do it if we continue to keep this code in future.

rust/otap-dataflow/crates/contrib-nodes/Cargo.toml
```diff
@@ -55,15 +55,17 @@ reqwest = { workspace = true, optional = true, features = ["rustls-no-provider"]
...
 
+[target.'cfg(windows)'.dependencies]
+one_collect = { workspace = true, optional = true }
+
 [target.'cfg(target_os = "linux")'.dependencies]
-# TODO: Remove this pinned git dependency once one-collect is upstream in the
-# normal dependency graph and no longer needs to be pulled directly here. The
-# pinned commit is from 2026-04-10.
-one_collect = { git = "https://github.com/microsoft/one-collect.git", rev = "9292caacaddf9ff9e4fbdf77bc62b5ec25494c84", features = ["scripting"], optional = true }
+one_collect = { workspace = true, optional = true }
```
lalitb, line 62: nit - small cleanup
```yaml
  [target.'cfg(any(windows, target_os = "linux"))'.dependencies]
  one_collect = { workspace = true, optional = true }
```

rust/otap-dataflow/Cargo.toml
```diff
@@ -217,6 +217,7 @@ xxhash-rust = { version = "0.8", features = ["xxh3"] }
 zip = "=8.6.0"
 byte-unit = { version = "5.2.0", features = ["serde"] }
 cpu-time = "1.0.0"
+one_collect = { git = "https://github.com/microsoft/one-collect.git", rev = "cfe3f78" }
```
lalitb, line 220: nit -  Can we use the full commit SHA here - would keep this consistent and easier to audit later.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
utpilla: The issue with restarts and hot-reload is being tracked in #3033. I think some of what you're saying would also be covered in it.

> a clear error for mismatched/additional subscriptions would still make this safer.

Could you explain in more detail what you're looking for here?

lalitb (approved):
LGTM. Since this is an iterative PR, most of my comments can be handled in follow-ups.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
lalitb: > The issue with restarts and hot-reload is being tracked in https://github.com/open-telemetry/otel arrow/issues/3033. I think some of what you're saying would also be covered in it.

Thanks, #3033 covers the restart/shutdown lifecycle part.

> Could you explain in more detail what you're looking for here?

What I meant here is only the case where the singleton is already running and another ETW receiver comes in with a different session name or provider list. Today it seems like that second receiver would just reuse the first session. I was wondering if we should detect that and return a clear error, instead of making it look like the second config was applied.

drewrelmas:
Setting to merge on the hope that fixing #3041 is a fast follow-up to prevent confusing check failures for unrelated PRs.

Issue #2901, opened 2026-05-07
[Comparison Dashboard] Publish comparison site next to existing benchmarks


JakeDern:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Other

### Objective

Automatically build and publish the comparison dashboard site alongside the existing benchmarks. Since there can only be one github pages site per repo we need to publish from the same root and via the same method (pushing data into a branch).

The main idea is roughly the following to minimize work/churn in the short term:

- All data and generated site files are checked into the same `benchmarks` branch that we use today
- Data will be (for now) manually PR'd into a `/comparison-data` directory next to the `benchmarks` folder where our continuous/nightly benchmarks go
- A github action will automatically trigger on changes in `/comparison-data` on the benchmarks branch and build the site + PR it into the benchmarks branch
- The comparison dashboard have a landing page at the `/compare` path with the detailed comparison pages living at `/compare/<comparison_slug>`

### Rationale

To make the comparison dashboard public and start exercising the code there.

### Scope

- Add arguments to `dashboard.py` script to specify the data publishing directory
- Add arguments to `dashboard.py` to specify the target directory for the site build
- Action to automatically build the site and PR the results which will be triggered on changes to the `/comparison-data` directory

### Acceptance Criteria

.

PR #2937, opened 2026-05-12
feat(comparison_dashboard): Add a workflow for building/publishing the comparison dashboard when code is updated on main

JakeDern:
# Change Summary

This is the first part of #2901 which is to automatically build and publish the site.

We also need to add a workflow to the benchmarks branch that triggers when there are changes to the data there.

* Part of #2901

Been testing these on my fork - Seems to work there, fingers crossed.

Example run: https://github.com/JakeDern/otel-arrow/actions/runs/25747403028

Yes, this will start publishing the stub site at `https://open-telemetry.github.io/otel-arrow`


lquerel (approved):
LGTM

PR #2938, opened 2026-05-12
feat(comparison_dashboard): Run main build + publish workflow when data is committed to the benchmarks branch

JakeDern:
# Change Summary

Trigger the workflow added in #2937 when data is pushed to this branch.

Note that apparently some people can push directly to the benchmarks branch which I did accidentally this morning because my origin was the wrong remote. I'm not sure if that's intended or not and hence is why this PR looks like it's just updating things.

* Part of #2901

Been testing on my fork, seems to work - Fingers crossed.

No.


tools/.gitignore
```diff
@@ -0,0 +1 @@
+comparison_dashboard
```
JakeDern, line 1: Common to have comparison_dashboard stuff floating around like the .site and .data dirs for testing.

lquerel (approved):
LGTM

PR #2940, opened 2026-05-12
fix(load_generator): honor --serve-port in pre-flight port check

ThomsonTan:
# Change Summary

`loadgen.py --serve` always probed the hard-coded `FLASK_PORT` (5001) before calling `app.run(port=args.serve_port)`, so `--serve-port` was effectively ignored by the pre-flight check. Three observable failure modes:

- `--serve-port 8080` with 8080 busy: check passed, then Flask died with a late `OSError: address already in use`.
- `--serve-port 8080` with an unrelated process on 5001: check raised, refusing to start a server that would have bound fine.
- Error message hard-coded `FLASK_PORT` in the string, so it named the wrong port either way.

Also tightened `is_port_in_use` itself: it used `connect_ex` against `0.0.0.0`, which only sees loopback listeners and tests reachability instead of bindability -- the property actually relevant to `app.run`.

* Closes #NNN

Tested locally.


cijothomas (approved):
LGTM

lquerel (approved):

PR #2956, opened 2026-05-13
fix(admin): group Prometheus metrics contiguously per spec

### LINKED ISSUE #2945: DFE admin server prometheus metrics format is not spec compliant
### Pre-filing checklist

### Component(s)

### Bug Description

 The otap-dataflow engine's admin Prometheus endpoint emits non-spec-compliant
 text. Strict downstream parsers silently drop most of the samples for any
 metric that is reported from more than one pipeline-thread entity (i.e. any
 metric on a multi-core component). In the dashboard this manifests as
 logs_produced_rate plateauing at one core's throughput while
 logs_received_rate (single-core backend, unaffected) keeps scaling.

 What the Prometheus exposition spec requires (text format
 (https://prometheus.io/docs/instrumenting/exposition_formats/)):
 all sample lines for a given metric must appear as one contiguous group,
 preceded by at most one # HELP and one # TYPE directive. A parser that
 sees the same metric name re-opened later in the stream without a
 preceding # TYPE may treat the new occurrence as a separate unknown-type
 family.

 What the engine actually does
 (rust/otap-dataflow/crates/admin/src/telemetry.rs:1182): format_prometheus_text
 iterates the telemetry registry per (metric_set, entity) pair and, for each
 entity, emits its full block of fields in sequence. With N entities sharing
 the same metric set (e.g. 10 pipeline threads each owning a copy of
 TrafficGeneratorReceiverMetrics), every metric in that set is re-opened N
 times, interleaved with the other fields. The global seen: HashSet<String>
 in emit_scalar_metric (line 714) gates # HELP / # TYPE to first sight,
 so re-opened groups have no directive.

 Evidence -- live scrape of df_engine:latest during the 1000k dashboard
 test (10 loadgen cores), curl .../telemetry/metrics?format=prometheus:

```
 # line 195
 # HELP logs_produced_total Number of logs generated.
 # TYPE logs_produced_total counter
 # line 197
 logs_produced_total{...,core_id="2"} 580000
 # lines 198-392: spans_produced_total, metrics_produced_total,
 #                channel/exporter blocks (each with its own HELP/TYPE)
 # line 393 -- block re-opens with NO preceding TYPE
 logs_produced_total{...,core_id="9"} 598000
 # 8 more re-opens at lines 430, 511, 692, 729, 918, 959, 1024, 1109
```

 Feeding that text into prometheus_client.parser.text_string_to_metric_families
 (the parser the orchestrator scraper uses) yields:

```
 ('logs_produced',       'counter', 'logs_produced_total')   # 1 sample, core 2
 ('logs_produced_total', 'unknown', 'logs_produced_total')   # 1 sample per re-open x9
```

 Because counter detection requires a # TYPE directive, the parser strips
 _total only from the first family. The scraper then routes the 9
 re-opened cores to a different OTel instrument than the first, and the SQL
 filter metric_name = 'logs_produced' matches only one core. The diagnostic
 log_metric_core_counts table in the dashboard SQL report confirms:

 | component_name   | metric_name   | core_count | core_ids |
 | backend-service  | logs          |          1 | 1        |   (1-core component, correct)
 | load-generator   | logs_produced |          1 | 3        |   (should be 10: cores 2..11)

 Why now: #2900 "fix(admin): apply OTel→Prometheus name & unit suffix
 rules" (99b02d611, 2026-05-08) started suffixing counters with _total.
 Pre-#2900 the sample name was just logs_produced, so the parser had no
 _total to strip and all re-opened groups merged onto the same family.name
 despite the underlying spec violation. The multi-core dashboard suites
 (#2889, 2026-05-07) had a ~1-hour window of correct behavior before
 #2900 flipped them red.

### Steps to Reproduce

Run any comparison_dashboard suite at higher than 100k throughput so that multiple loadgen cores are activated.

### Expected Behavior

100k * N logs produced rate

### Actual Behavior

100k logs produced rate

### OTel-Arrow Version

NA

### Environment

_No response_

### Configuration

```yaml

```

### Log Output

```shell

```

### Additional Context

_No response_

cijothomas:
## Problem

The Prometheus exposition format requires all sample lines for a given metric name to appear as one contiguous group, preceded by at most one `# HELP` and one `# TYPE` directive. The previous implementation iterated per-entity, emitting each entity's full metric set inline. With N entities sharing the same metric set (e.g. 10 pipeline-thread cores), every metric was re-opened N times non-contiguously. Strict parsers (like `prometheus_client.parser.text_string_to_metric_families`) split re-opened groups into separate unknown-type families, causing multi-core metrics like `logs_produced` to report only one core's throughput.

## Solution

Replace the direct-emission approach with a **two-phase collect-then-emit** pattern:

1. **Collect phase**: During the registry visit, samples are grouped by metric name into a `PromGroupedMetrics` structure that preserves insertion order.
2. **Emit phase**: After all entities are visited, each metric family is emitted as a contiguous block: `# HELP` → `# UNIT` → `# TYPE` → all sample lines.

Both `format_prometheus_text` and `agg_prometheus_text` are updated.

## Testing

- All 50 existing + new tests pass (`cargo xtask check` clean)
- Added `test_format_prometheus_text_multi_entity_contiguous_grouping` that registers two entities sharing a metric set and verifies:
  - Exactly one `# HELP` and `# TYPE` per metric name
  - All lines (directives + samples) for each metric are contiguous (no gaps)
  - Both entities' samples are present

Fixes #2945

### DIFF
--- a/rust/otap-dataflow/crates/admin/src/telemetry.rs
+++ b/rust/otap-dataflow/crates/admin/src/telemetry.rs
@@ -599,6 +599,74 @@ fn format_lp_value(value: MetricValue, value_type: Option<MetricValueType>) -> S
     }
 }
 
+/// Metadata for a Prometheus metric family (HELP, UNIT, TYPE directives).
+struct PromMetricMetadata {
+    help: Option<String>,
+    unit: Option<String>,
+    prom_type: String,
+}
+
+/// A single metric family group: metadata + collected sample lines.
+struct PromMetricGroup {
+    metadata: PromMetricMetadata,
+    samples: Vec<String>,
+}
+
+/// Collects metric samples grouped by metric name, preserving insertion order.
+/// After all entities are visited, `emit` writes contiguous metric families.
+struct PromGroupedMetrics {
+    /// Metric names in insertion order.
+    order: Vec<String>,
+    /// Metric name → group.
+    groups: HashMap<String, PromMetricGroup>,
+}
+
+impl PromGroupedMetrics {
+    fn new() -> Self {
+        Self {
+            order: Vec::new(),
+            groups: HashMap::new(),
+        }
+    }
+
+    /// Returns a mutable reference to the group for `metric_name`, creating it
+    /// with the given metadata factory if it doesn't exist yet.
+    fn get_or_insert(
+        &mut self,
+        metric_name: &str,
+        metadata_fn: impl FnOnce() -> PromMetricMetadata,
+    ) -> &mut PromMetricGroup {
+        match self.groups.entry(metric_name.to_string()) {
+            Entry::Occupied(e) => e.into_mut(),
+            Entry::Vacant(e) => {
+                self.order.push(metric_name.to_string());
+                e.insert(PromMetricGroup {
+                    metadata: metadata_fn(),
+                    samples: Vec::new(),
+                })
+            }
+        }
+    }
+
+    /// Emits all collected metrics as contiguous Prometheus text families.
+    fn emit(self, out: &mut String) {
+        for name in &self.order {
+            if let Some(group) = self.groups.get(name) {
+                if let Some(ref help) = group.metadata.help {
+                    let _ = writeln!(out, "# HELP {name} {help}");
+                }
+                if let Some(ref unit) = group.metadata.unit {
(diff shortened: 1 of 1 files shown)


lquerel (approved):
LGTM

JakeDern (approved):

PR #2957, opened 2026-05-13
Validate otel_* event names at compile time

cijothomas:
Adds a `const fn validate_event_name` invoked from each `otel_info!` / `otel_warn!` / `otel_debug!` / `otel_error!` / `otel_event!` macro arm, so calls with empty or whitespace-containing event names fail `cargo check` instead of silently producing logs whose entire description is shoved into `event.name`.

Zero runtime cost (compile-time only). Existing call sites that violated the rule are updated to use a short identifier with the descriptive text moved to a `message = "..."` field.

The local macro copies in the `quiver` crate are intentionally left for a follow-up.

### DIFF
--- a/rust/otap-dataflow/crates/telemetry/src/internal_events.rs
+++ b/rust/otap-dataflow/crates/telemetry/src/internal_events.rs
@@ -232,6 +232,7 @@ macro_rules! otel_event {
 #[macro_export]
 macro_rules! raw_error {
     ($name:expr $(, $($fields:tt)*)?) => {{
+        const _: () = $crate::_private::validate_event_name($name);
         use $crate::self_tracing::ConsoleWriter;
         let now = std::time::SystemTime::now();
         let record = $crate::__log_record_impl!($crate::_private::Level::ERROR, $name $(, $($fields)*)?);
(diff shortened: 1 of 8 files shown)


JakeDern (approved):
Guilty as charged, thank you 🙂

lalitb:
Also update the example here - https://github.com/open-telemetry/otel-arrow/blob/b645a269cc18cba6310ca15648a990f4bdc94e68/rust/otap-dataflow/crates/telemetry/README.md?plain=1#L70

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
lalitb, line 233: Should raw_error! also use stable event names, or is it intentionally allowed to keep human-readable fallback messages?

lalitb (approved):
LGTM, with nit comments

rust/otap-dataflow/crates/telemetry/src/internal_events.rs
cijothomas: Good catch — fixed them too. (Quiver also needs it; will send a separate PR.)

PR #2964, opened 2026-05-13
implement struct column assignment and validation logic

### LINKED ISSUE #2838: [OTAP query-engine]: Support assignment of "struct" columns
In OTAP data model we have a handful of columns that are struct columns. These include notably columns for resource and instrumentation scope.

In the columnar query-engine, we should support assignment to these columns:
```
logs | set resource.schema_url = "http://my-schema.com"
logs | set instrumentation_scope.name = "some_scope_name"
```

### Validity rules

We should not allow assignments to fields where the relationship between the destination and the source is 1:many. That in OTEL data-model, there is a conceptual hierarchy of `RESOURCE ==1:many==> SCOPE ==1:many==> LOG/SPAN/METRIC`.

This means that no field on `scope`/`log` can be used to assign a field on `resource` for example. No field on `log` could be used to assign the value of `scope`. E.g. `set instrumentation_scope.name = attributes["x"]` is not allowed.

We already have [some logic to validate this when assigning attributes](https://github.com/open-telemetry/otel-arrow/blob/8b4e674f6909f04cdd6726c42e3ad0f3281b3d6e/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs#L1444-L1506), so we must do something similar for these struct fields.

### Implementation Instructions

Implement the rules described above in `validate_assign`:
https://github.com/open-telemetry/otel-arrow/blob/8b4e674f6909f04cdd6726c42e3ad0f3281b3d6e/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs#L1430-L1439

The main entrypoint to the `AssignPipelineStage` is this method. Here there is a variant of `ColumnAccessor` called `StructCol` which is not handled. It can be handled in a similar way to what we do for regular non-struct columns:
https://github.com/open-telemetry/otel-arrow/blob/8b4e674f6909f04cdd6726c42e3ad0f3281b3d6e/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs#L791-L815

E.g. can implement similar methods to what we do for non-struct columns. See [`assign_to_root`](https://github.com/open-telemetry/otel-arrow/blob/8b4e674f6909f04cdd6726c42e3ad0f3281b3d6e/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs#L187-L325) and [`assign_null_root_column`](https://github.com/open-telemetry/otel-arrow/blob/8b4e674f6909f04cdd6726c42e3ad0f3281b3d6e/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs#L385-L420). 

The type validation will need to be implemented using the same patterns as these two methods but for fields within the struct columns. See section five of the [otap-spec.md](https://github.com/open-telemetry/otel-arrow/blob/main/docs/otap-spec.md#5-payload-specifications) for the column types.

We can probably also refactor the main logic inside the function [`try_upsert_column`](https://github.com/open-telemetry/otel-arrow/blob/8b4e674f6909f04cdd6726c42e3ad0f3281b3d6e/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs#L1663-L1730) to a separate method that can accept `arrow::datatype::Fields` and `Vec<ArrayRef>` (columns) and return the updated fields & columns.

### Tests
Here is roughly the list of test cases that should be added:

- assign struct field from static (e.g. `resource.schema_url = "something"`)
- assign struct field  where dest:source relation is 1:many relation fails e.g. not allowed to do `resource.schema_url = logs.attributes["x"]` or `resource. dropped_attributes_count = instrumentation_scope. dropped_attributes_count `)
- assign struct from a dest:source relation is many:1 is OK (e.g. `instrumentation_scope. dropped_attributes_count = resource. dropped_attributes_count`, `instrumentation_scope.name = resource.attributes["name"]`)
- assign struct field from it's own attributes `instrumentation_scope.name = instrumentation_scope.attributes["name"]`)
- assign the wrong type is not allowed (e.g. not allowed to do `resouce.dropped_attribute_count = "hello"`)
- assign null (e.g. `resource.schema_url = null`), ensure this drops the column from the OTAP batch
- assign the result of a function call (e.g. `instrumentation_scope.name = concat(instrumentation_scope.name, "-v", instrumentation_scope.version)
- assign when the struct column did not already exist 
- assign when the struct field did not already exist

guancioul:
# Change Summary

This PR implements execution and validation logic for assigning values to nested struct fields in the OTAP columnar query engine, enabling queries such as:
```
logs | set resource.schema_url = "http://my-schema.com"
logs | set instrumentation_scope.name = "some_scope_name"
```

### Validation (`validate_assign` in `assign.rs`)

- Add `StructCol` arm to the `validate_assign` dispatch
- Add `is_valid_struct_field` to check that the target field name actually exists in the struct schema
- Add type validation to reject assignments where the source value type does not match the destination field type
- Add `validate_struct_col_assign_cardinality` to enforce the OTAP 1:many hierarchy constraint (`RESOURCE ==1:many==> SCOPE ==1:many==> LOG/SPAN/METRIC`); e.g. `set resource.schema_url = attributes["x"]` is rejected because log-level data cannot be used to assign a resource-level field

### Execution (`execute` in `assign.rs`)

- Add `StructCol` arm to the execution dispatch to handle struct field assignment
- Add `assign_to_struct_column` — handles non-null assignment to a struct field with row-level alignment between the source array and the target struct column
- Add `assign_null_struct_field` — handles `null` assignment by dropping the target field from the struct in the OTAP batch

### Scalar Integer Coercion

- Coerce static scalar integer literals to the destination field's Arrow type (e.g. an `AnyInt` literal `42` is cast to `UInt32` when assigning to `instrumentation_scope.dropped_attributes_count`), mirroring the same cast already done for root-level column assignment

### DataScope Enhancement

- Extend `DataScope` with a new `RootParent` variant to represent the cardinality level of a value read from a struct column field (e.g. reading `resource.schema_url` produces `RootParent(ResourceAttrs)`)
- Previously, reading a struct field like `resource.schema_url` as a source expression was treated as `Root` (log-level) scope — the same as any plain column — so the cardinality validator could not distinguish resource-level struct field reads from log-level reads
- With `RootParent`, the validator can now correctly compare the cardinality level of the **source expression** (e.g. `resource.schema_url` → `RootParent(ResourceAttrs)`) against the destination struct column (e.g. assigning to `instrumentation_scope`) and reject invalid assignments accordingly

### Spec Update (`docs/otap-spec.md`)

- Update the Spans and Metrics column tables to reflect that `scope.name`, `scope.version`, and `resource.schema_url` support `Dict(u8), Dict(u16)` optional encoding (was previously missing from those two tables, though already present in the Logs table)

- Closes #2838

- Assign a struct field from a static literal (e.g. `resource.schema_url = "something"`)
- Assign a non-string (UInt32) struct field from a scalar literal (e.g. `instrumentation_scope.dropped_attributes_count = 42`)
- Reject assignment where source is at a lower cardinality level than the destination (e.g. `resource.schema_url = attributes["x"]`)
- Allow assignment where source is at a higher or equal cardinality level (e.g. `instrumentation_scope.name = resource.schema_url`)
- Assign a struct field from its own sibling attributes (e.g. `instrumentation_scope.name = instrumentation_scope.attributes["name"]`)
- Reject assignment of the wrong type (e.g. `resource.dropped_attributes_count = "hello"`)
- Assign `null` to a struct field and verify the field is dropped from the batch
- Assign the result of a function expression (e.g. `instrumentation_scope.name = substring(instrumentation_scope.name, 0, 6)`)
- Assign when the struct column does not yet exist in the batch
- Assign when the target field does not yet exist within an existing struct column

Yes. Users can now write OPL/KQL queries that assign values to nested struct fields such as `resource.schema_url`, `instrumentation_scope.name`, `instrumentation_scope.version`, and `instrumentation_scope.dropped_attributes_count`.

### DIFF
--- a/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
+++ b/rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
@@ -18,14 +18,14 @@ use std::rc::Rc;
 use std::sync::Arc;
 
 use arrow::array::{
-    Array, ArrayRef, BooleanArray, DictionaryArray, Float64Array, Int64Array, NullArray,
+    Array, ArrayRef, AsArray, BooleanArray, DictionaryArray, Float64Array, Int64Array, NullArray,
     RecordBatch, StringArray, StructArray, UInt8Array, UInt16Array,
 };
 use arrow::buffer::BooleanBuffer;
 use arrow::compute::kernels::cmp::{eq, neq};
 use arrow::compute::kernels::merge::merge;
 use arrow::compute::{and_not, cast, filter, max, take};
-use arrow::datatypes::{DataType, Field, Schema, UInt16Type};
+use arrow::datatypes::{DataType, Field, Fields, Schema, UInt16Type};
 use async_trait::async_trait;
 use data_engine_expressions::QueryLocation;
 use datafusion::config::ConfigOptions;
@@ -53,7 +53,7 @@ use crate::pipeline::expr::join::{
     RootToAttributesJoin,
 };
 use crate::pipeline::expr::types::{
-    ExprLogicalType, root_field_supports_dict_encoding, root_field_type,
+    ExprLogicalType, nested_struct_field_type, root_field_supports_dict_encoding, root_field_type,
 };
 use crate::pipeline::expr::{
     DataScope, ExprPhysicalPlanner, LogicalExprDataSource, PhysicalExprEvalResult,
@@ -419,6 +419,147 @@ impl AssignPipelineStage {
         Ok(otap_batch)
     }
 
+    /// try to assign an all-null value to a field within a struct column in the root record batch.
+    /// In practice, this just means removing the field from the struct. This will return an error
+    /// if it turns out the field is not nullable. If the struct column or the field does not exist,
+    /// this is a no-op.
+    fn assign_null_struct_field(
+        &self,
+        mut otap_batch: OtapArrowRecords,
+        struct_col_name: &str,
+        field_name: &str,
+    ) -> Result<OtapArrowRecords> {
+        let root_batch = match otap_batch.root_record_batch() {
+            Some(rb) => rb,
+            None => {
+                // nothing to do
+                return Ok(otap_batch);
+            }
+        };
+
+        let schema = root_batch.schema_ref();
+        let maybe_struct_col = schema.fields().find(struct_col_name);
+        if let Some((struct_col_index, struct_field)) = maybe_struct_col {
+            // the column should always be a struct, but guard against schema corruption
+            let DataType::Struct(struct_fields) = struct_field.data_type() else {
+                return Err(Error::ExecutionError {
+                    cause: format!(
+                        "expected struct column '{struct_col_name}' to have DataType::Struct, \
+                        found {:?}",
+                        struct_field.data_type()
+                    ),
+                });
+            };
+
+            let maybe_found_field = struct_fields.find(field_name);
+            if let Some((field_index, field)) = maybe_found_field {
+                if field.is_nullable() {
+                    let struct_array = root_batch.column(struct_col_index).as_struct();
+                    let mut new_fields = struct_fields.to_vec();
+                    _ = new_fields.remove(field_index);
+                    let mut new_columns = struct_array.columns().to_vec();
+                    _ = new_columns.remove(field_index);
+
+                    let new_struct = Arc::new(StructArray::new(
+                        Fields::from(new_fields),
+                        new_columns,
+                        struct_array.nulls().cloned(),
+                    ));
+
+                    let root_payload_type = otap_batch.root_payload_type();
+                    otap_batch.set(
+                        root_payload_type,
+                        try_upsert_column(struct_col_name, new_struct, root_batch)?,
+                    )?;
+                } else {
+                    return Err(Error::ExecutionError {
+                        cause: format!(
+                            "cannot assign null to non-nullable field '{field_name}' \
+                            on struct column '{struct_col_name}'"
+                        ),
+                    });
+                }
+            }
+        }
+
+        Ok(otap_batch)
+    }
+
+    fn assign_to_struct_column(
+        &self,
+        mut otap_batch: OtapArrowRecords,
+        mut eval_result: PhysicalExprEvalResult,
+        dest_scope: &Rc<DataScope>,
+        dest_column_name: &str,
+        dest_field_name: &str,
+    ) -> Result<OtapArrowRecords> {
+        let root_batch = match otap_batch.root_record_batch() {
+            Some(rb) => rb,
+            None => {
+                // nothing to do
+                return Ok(otap_batch);
+            }
+        };
+
+        let root_payload_type = otap_batch.root_payload_type();
+        let column_supports_dict_encoding = nested_struct_field_supports_dict_encoding(
+            dest_column_name,
+            dest_field_name,
+            root_payload_type,
+        );
+
+        if let ColumnarValue::Array(values) = &eval_result.values {
+            if is_any_value_data_type(values.data_type()) {
+                let coerced = attempt_coerce_value_column_from_any_value_struct_column(values)?;
+                eval_result.values = ColumnarValue::Array(coerced);
+            }
+        }
+
+        let mut values = eval_result_to_array(
+            &eval_result.values,
+            column_supports_dict_encoding,
+            root_batch.num_rows(),
+        )?;
+
+        // Check if the source rows are already aligned with the root batch rows.
+        // Scalars broadcast to any row count; Root and StructField both live in the root batch
+        // so their row order matches. If the source is an Attributes batch, it has fewer rows
+        // (one per scope/resource) than the root batch (one per log/span/metric), so we need
+        // a join to expand and reorder the values to match the root batch row count.
+        let already_aligned = eval_result.data_scope.is_scalar()
+            || eval_result.data_scope.as_ref() == dest_scope.as_ref()
+            || matches!(
+                eval_result.data_scope.as_ref(),
+                DataScope::Root | DataScope::StructField(_)
+            );
+
+        if !already_aligned {
+            let DataScope::Attributes(attrs_id, _) = eval_result.data_scope.as_ref() else {
+                unreachable!("unexpected data_scope")
+            };
+
+            let join_exec = RootToAttributesJoin::new(*attrs_id);
+            let vals_take_indices = join_exec.rows_to_take(
+                &PhysicalExprEvalResult::new(
+                    ColumnarValue::Scalar(ScalarValue::Null),
+                    Rc::clone(dest_scope),
+                    root_batch,
+                ),
+                &eval_result,
+                &OtapArrowRecords::Logs(Logs::default()),
+            )?;
+
+            values = take(&values, &vals_take_indices, None)?;
+        }
+
+        otap_batch.set(
+            root_payload_type,
+            try_upsert_struct_col(dest_column_name, dest_field_name, values, root_batch)?,
+        )?;
+
+        Ok(otap_batch)
+    }
+
     fn assign_to_attributes(
         &mut self,
         mut otap_batch: OtapArrowRecords,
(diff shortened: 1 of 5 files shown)


rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -419,6 +419,160 @@ impl AssignPipelineStage {
...
+                    let mut new_fields = struct_fields.to_vec();
+                    _ = new_fields.remove(field_index);
+                    let mut new_columns = struct_array.columns().to_vec();
+                    _ = new_columns.remove(field_index);
+
+                    let new_struct = Arc::new(StructArray::new(
+                        Fields::from(new_fields),
+                        new_columns,
+                        struct_array.nulls().cloned(),
+                    ));
```
albertlockett: If there are no remaining columns, then can we remove the struct column completely?

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -419,6 +419,160 @@ impl AssignPipelineStage {
...
+                return Ok(otap_batch);
+            }
+        };
+
+        let root_payload_type = otap_batch.root_payload_type();
+        let column_supports_dict_encoding = nested_struct_field_supports_dict_encoding(
+            dest_column_name,
+            dest_field_name,
+            root_payload_type,
+        );
```
albertlockett: I don't think the root payload type is needed as an argument here. Presumably this would be passed because we may want to support dictionary encoding on some struct field _only_ if the batch is some specific signal type - but I actually think for all signal types the struct fields that support dictionary encoding is the same. e.g. if `<scope_col>.<scope_field>` supports dicts encoding for logs, it will also support it for traces/metrics.

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -1505,6 +1691,108 @@ fn validate_attribute_assign_cardinality(
...
+        (struct_name, field_name, payload_type),
+        (consts::RESOURCE, consts::SCHEMA_URL, ArrowPayloadType::Logs)
+            | (
+                consts::RESOURCE,
+                consts::SCHEMA_URL,
+                ArrowPayloadType::Spans
+            )
+            | (consts::SCOPE, consts::NAME, ArrowPayloadType::Logs)
+            | (consts::SCOPE, consts::VERSION, ArrowPayloadType::Logs)
+    )
```
albertlockett, line 1718: See my other comment at the call-site - we don't need to consider the payload type arg here. dictionary encoding decision for the field should be the same for all signal types.

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
```diff
@@ -142,6 +142,11 @@ pub(crate) enum DataScope {
     /// A special data scope indicating the data is produced from a static scalar value defined
     /// in the input expression tree, rather than data from the OTAP batch.
     StaticScalar,
+
+    /// A field read from a struct column in the root record batch (e.g., resource.schema_url or
+    /// instrumentation_scope.name). Physically the data lives in the root batch (same as Root),
+    /// but the payload type records the hierarchy level for cardinality validation.
+    StructField(ArrowPayloadType),
```
albertlockett: I wonder if we could call this something like `RootParent` or something instead of `StructField`? I feel `StructField` might be confusing because logs.body is also a struct array on the root logs record batch, but for purposes of validating assignment cardinality using scope we actually want to treat it as `Root`.

Also, I feel we could introduce an enum specifically for determining which struct field we're accessing? This would preclude the possibility of creating an invalid scope by providing an `ArrowPayloadtype` variant that is neither `ResourceAttrs` nor `ScopeAttrs`.

e.g. maybe something like: 
```
enum RootParentStruct {
  Resource
  Scope
}
```

albertlockett (commented):
Thanks for taking on this work @guancioul! I had a few feedbacks, please take a look

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
guancioul: Yes, I missed this case. Fixed — when the last field is removed, we now drop the struct column entirely instead of leaving an empty StructArray.

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
guancioul: The implementation is following [otap-spec.md](https://github.com/open-telemetry/otel-arrow/blob/main/docs/otap-spec.md#51-logs) "Optimized Encodings" column. which differs per signal type. For example, `scope.name` and `scope.version` support dict encoding for Logs but not for Spans or Metrics. If we want uniform behavior across all signal types, the spec would need to be updated first. Or did I miss something?

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
guancioul: Same as above.

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
guancioul: You're right that `StructField` was misleading, `logs.body` is also a struct on the root batch but should be treated as `Root`, so the name was ambiguous. Introducing a dedicated `RootParentStruct` enum is also a great suggestion to make invalid states unrepresentable. Both changes have been applied. Thank you!

rust/otap-dataflow/crates/query-engine/src/pipeline/expr.rs
```diff
@@ -153,8 +165,14 @@ impl DataScope {
...
     /// - Any scope can combine with StaticScalar (constants)
     /// - Same scopes can combine (e.g., Root + Root), because the row order is the same.
+    /// - Root and RootParent can combine because both live in the root record batch.
     pub(crate) fn can_combine(&self, other: &Self) -> bool {
-        self.is_scalar() || other.is_scalar() || (self == other)
+        if self.is_scalar() || other.is_scalar() {
+            return true;
+        }
+        let self_in_root = matches!(self, Self::Root | Self::RootParent(_));
+        let other_in_root = matches!(other, Self::Root | Self::RootParent(_));
```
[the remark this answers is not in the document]
guancioul: The specific scenario described — mixing `RootParent(Resource)` and `RootParent(Scope)` in a binary expression — is not reachable for string fields (e.g. `resource.schema_url`, `scope.name`), because `coerce_arithmetic` returns `None` for `String + String`, causing a planning error before `can_combine` is even relevant. What do you think? Is this worth addressing in this PR?

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
guancioul: Base on the Copliot https://github.com/open-telemetry/otel-arrow/pull/2964#discussion_r3273673991, I looked at the actual `ScopeBuilder` and `ResourceBuilder` implementations — they both use `dict8` for `scope.name` and `scope.version` regardless of signal type (the same `ScopeBuilder::new()` is reused by both `LogsRecordBatchBuilder` and `TracesRecordBatchBuilder`). However, [otap-spec.md](https://github.com/open-telemetry/otel-arrow/blob/main/docs/otap-spec.md#51-logs) shows no dict encoding for `scope.name`/`scope.version` on Traces, which contradicts the builder behavior. So there's a discrepancy between the spec and the implementation. My current `nested_struct_field_supports_dict_encoding` follows the spec. Which side should we align to? And if we align to the builder behavior, should the spec be updated as well?

albertlockett: I think the spec document is incorrect in this case. These struct columns should be dictionary encoded for traces & metrics

guancioul: Thanks for the clarification! I've updated [otap-spec.md](https://github.com/open-telemetry/otel-arrow/blob/main/docs/otap-spec.md) to reflect that `scope.name`, `scope.version`, and `resource.schema_url` support `Dict(u8)/Dict(u16)` for Spans and Metrics as well, and simplified `nested_struct_field_supports_dict_encoding` to drop the `payload_type` parameter.

albertlockett (approved):
LGTM, Thanks again @guancioul

guancioul:
Thanks for the review @albertlockett , I have updated the description to match all the changes.

albertlockett:
Thank you @guancioul !

PR #2966, opened 2026-05-13
feat: add num_connections config to OTLP gRPC exporter

cijothomas:
## Summary

Add a `num_connections` configuration option to the OTLP gRPC exporter that controls how many independent TCP connections (tonic Channels) are created per pipeline.

## Problem

When the receiver uses `SO_REUSEPORT` across multiple cores, the kernel distributes **new TCP connections** (not individual RPCs) across listener sockets. With the previous behavior of 1 gRPC channel per pipeline, this caused severe core imbalance — e.g., with 2 engine cores: one core at 60% and another at 94%.

## Solution

- Added `num_connections` config field (default: 1) to the OTLP gRPC exporter
- When `num_connections > 1`, creates N independent tonic Channels, each establishing its own TCP connection
- Rewrote `GrpcClientPool` to use a FIFO `VecDeque` for round-robin distribution of gRPC clients across channels
- Updated saturation test templates to set `num_connections = num_cores * 4`

## Results

With `num_connections` set appropriately:
- Core imbalance fixed: 60%/94% → 99%/99%
- 2-core throughput improved from 0.90× to 1.36× of 1-core baseline

| Config | logs/sec | Scaling | Core balance |
|--------|----------|---------|--------------|
| 1-core, 1 conn (old) | 164,727 | baseline | N/A |
| 2-core, 1 conn (old) | 148,685 | 0.90× | 60%/94% |
| 1-core, 4 conns (new) | 177,461 | baseline | 99.6% |
| 2-core, 8 conns (new) | 241,964 | 1.36× | 95.4% avg, balanced |

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
@@ -44,6 +44,7 @@ use otap_df_telemetry::instrument::Counter;
 use otap_df_telemetry::metrics::MetricSet;
 use otap_df_telemetry::{otel_debug, otel_info, otel_warn};
 use serde::Deserialize;
+use std::collections::VecDeque;
 use std::future::Future;
 use std::sync::Arc;
 use tonic::codec::CompressionEncoding;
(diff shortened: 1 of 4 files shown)


cijothomas:
Re-opening from fork

PR #2977, opened 2026-05-14
Validate quiver otel_* event names at compile time

cijothomas:
Follow-up to #2957, which intentionally skipped the local macro copies in the `quiver` crate.

Zero runtime cost (compile-time only). All existing quiver call sites already use short stable identifiers (e.g. `quiver.wal.rotate`), so no call-site updates were required.


AaronRM:
LGTM. Thanks @cijothomas!

AaronRM (approved):

lalitb (approved):

albertlockett (approved):

PR #2982, opened 2026-05-15
perf: add egress_bytes_per_log metric to benchmark reports

cijothomas:
## Summary

Two changes to benchmark report metrics:

### 1. Add `egress_bytes_per_log` metric
Adds a derived metric (bytes/log) computed as `network_tx_bytes_rate / logs_received_rate`. This makes it easy to assess whether compression ratios in tests are realistic.

For a ~150 byte log record, realistic values should be ~35-50 bytes/log. Values like ~27 bytes/log indicate unrealistic compression from low-entropy test data (e.g., replayed identical payloads).

### 2. Switch network metrics from bytes/sec to MB/s
Replaces `network_tx_bytes_rate_avg` and `network_rx_bytes_rate_avg` (bytes/sec) with `network_tx_mb_per_sec` and `network_rx_mb_per_sec` (MB/s) for readability. Raw bytes/sec values like 2,689,390 are hard to interpret at a glance; 2.56 MB/s is immediately meaningful.

### Files changed
- `integration/configs/integration_report_logs.yaml`
- `integration/configs/integration_report_metrics.yaml`
- `integration/configs/integration_report_traces.yaml`
- `comparison_dashboard/reports/report_logs.yaml`
- `comparison_dashboard/reports/report_metrics.yaml`
- `comparison_dashboard/reports/report_traces.yaml`

Relates to #2540

### DIFF
--- a/tools/pipeline_perf_test/test_suites/comparison_dashboard/reports/report_logs.yaml
+++ b/tools/pipeline_perf_test/test_suites/comparison_dashboard/reports/report_logs.yaml
@@ -323,7 +323,7 @@ queries:
 
           UNION ALL
 
-          SELECT 'network_tx_bytes_rate_avg' AS name, 'bytes/sec' AS unit, average_rate AS value,
+          SELECT 'network_tx_mb_per_sec' AS name, 'MB/s' AS unit, average_rate / 1048576.0 AS value,
                 metadata_row."test.suite" || '/' || metadata_row."test.name" || ' - Network Utilization' AS extra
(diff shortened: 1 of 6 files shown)


jmacd (approved):

JakeDern:
For the bytes/sec change my $0.02 would be to leave the report data in bytes/sec which and let the dashboard round/reformat into MiB/s or GiB/s.  This is what the comparison dashboard does today in some sections:

image

PR #2983, opened 2026-05-15
Better config-time validation for overlapping flow metrics configuration

### LINKED ISSUE #2784: Better config-time validation for overlapping stopwatch configuration
Follow-up from https://github.com/open-telemetry/otel-arrow/pull/2747#discussion_r3163147891 and https://github.com/open-telemetry/otel-arrow/pull/2747#discussion_r3183718018.

We should implement a simple graph traversal to detect overlapping stopwatches at configuration build time instead of simply emitting a WARN message.

Relevant location in `stopwatch.rs`: 

https://github.com/open-telemetry/otel-arrow/blob/8b4e674f6909f04cdd6726c42e3ad0f3281b3d6e/rust/otap-dataflow/crates/engine/src/stopwatch.rs#L238-L255

Current WARN log in `pdata.rs`: 

https://github.com/open-telemetry/otel-arrow/blob/8b4e674f6909f04cdd6726c42e3ad0f3281b3d6e/rust/otap-dataflow/crates/otap/src/pdata.rs#L358-L372

wbutler:
Closes #2784.

This change implements a BFS walk of the processor pipeline at construction time to validate that flow metrics do not collide.

- In `runtime_pipeline.rs`, introduce a helper func to capture a flat edge list vector of pipeline connections and pass the result as a new param on `build_flow_metric_state`. Saves us from having to pass in some part of `_config` into that function and keeps data types better separated.
- In `flow_metrics.rs`:
  - In `build_flow_metric_state`, add a new data structure that captures start\end pairs for registered flow metrics.
  - at the end of the main loop, iff the count of flow metrics > 1
    - convert the list of edges into a one-to-many adjacency data structure
    - for each flow metric
      - do a BFS to find all the nodes between the start and the end
      - check for nodes in this set that also appear as a start node in the data structure above.
      - If found, throw an error, halting pipeline construction.

## Tests

Within `flow_metrics.rs`, this change adds 10 unit tests, as follows:
- 7 tests exercise the interleaving detection logic on synthetically created pipeline topologies.
  - Three expected-pass tests include:
    - Linear topology with disjoint ranges
    - Branching topology with disjoint ranges
    - Linear topology with a single flow metric
  - Four expected-reject tests include:
    - Linear topology with interleaving
    - Linear topology with interleaving, reverse declaration order
    - Diamond topology with interleaving
    - Linear topology with fully nested ranges
- 3 tests exercise a helper function directly.

## Validation

The following commands pass cleanly:

`cargo check -p otap-df-engine`
`cargo test -p otap-df-engine` (contains new tests)
`cargo clippy -p otap-df-engine --all-targets -- -D warnings`
`cargo fmt --all -- --check`
`cargo xtask quick-check`


wbutler:
@drewrelmas should get a look at this since it bubbles up to his flow metrics work. Feedback & comments are very welcome.

rust/otap-dataflow/crates/engine/src/flow_metrics.rs
```diff
@@ -296,6 +301,13 @@ pub(crate) fn build_flow_metric_state(
         signals_outgoing_metrics.push(signals_outgoing_metric);
         let _ = end_nodes.insert(end_idx, id);
         let _ = start_nodes.insert(start_idx, id);
+        resolved_ranges.push((start_idx, end_idx, flow_config.name.clone()));
+    }
+
+    // Validate that flow_metric ranges don't interleave with each other.
+    if resolved_ranges.len() > 1 {
```
lalitb, line 308: This looks like the right direction. One concern: should we also fail config if the end node is not reachable from the start node?

wbutler: We certainly can! It's not my high-level feature, so I didn't want to overstep, but if @drewrelmas or the community thinks it's a good idea, I'm glad to extend the PR.

wbutler: @lalitb WDYT re adding in this PR or filing an issue and I'll do it in a second PR? I don't really have an opinion; I'm happy to do either.

lalitb: @wbutler - No strong opinion either, but in general ,we can keep this scope to overlap validation, and separate follow-up for rest of the static validations :)

lalitb (approved):
LGTM.

wbutler:
Created https://github.com/open-telemetry/otel-arrow/issues/3009 to capture the follow-up hardening for unreachable end nodes.

wbutler:
Ready for merge, IMO.

jmacd (approved):

PR #2984, opened 2026-05-15
fix: stop metrics otap encode panic

larry-c-ef:
Fix a panic in the metrics otap encoder on otlp bytes due to cursor misuse. Fix similar error due to column count mismatch when the resource field was blank.

Before this PR, we'd call `scope_metric.metrics().count()`, and then also `for metric in scope_metric.metrics()` on the same scope_metrics object. The problem is the first `count()` had side effects of updating a shared index into the protobuf bytes buffer that changed the value of the second call to `metrics()`, frequently leaving you with inconsistent row counts for some of the columns. (The `MetricsRecordBatchBuilder.finish()` code panics if some columns have a different row-count than the first column.)

After this PR, we only iterate over `scope_metric.metrics()` once.

In a related change, it's possible that the `resource` field is `None`; this PR updates the ‘if let Some(resource) = resource_metric.resource()` with an `else` branch that updates the `dropped_attributes_count` column, and also pulls out the updates to the `schema_url` and `id` columns to above the `if`, since they needed to happen either way.

# Change Summary

- Avoid using a shared OTLP proto-bytes metrics() cursor twice: drop the pre-loop metrics().count() (which was followed by reusing the same metrics() iterator). Instead, only iterate over metrics() once and append scope/schema columns per metric.
- Handle ResourceMetrics with no resource by still appending resource rows so Arrow struct lengths stay consistent.

These changes fix a panic due to column mismatch in MetricsRecordBatchBuilder::finish.

n/a, did not create an issue first.

Two new unit tests, which include raw bytes from a real request, generated from the opentelemetry-demo.

No.


larry-c-ef:
I created this draft so my company could get the ball rolling on the CLA. I will revise it later.

lquerel:
Hi @larry-c-ef, Sorry for the delayed response, and thank you for this contribution. I have no idea why this PR has remained open for so long. It may have slipped under our radar because it is still marked as a draft.

Could you please address the few cosmetic issues flagged by the CI pipeline and mark the PR as ready for review so that we can merge it? You can use the cargo xtask check command locally to reproduce most of the issues reported by CI.

lquerel:
Hi @larry-c-ef, Thank you again for this contribution, and apologies for how slowly we responded. The core diagnosis and most of the implementation in this PR were sound, particularly the missing-resource handling and the need to avoid inconsistent metrics traversal. This work directly informed #3582.

I’m closing this draft in favor of #3582 so we can unblock #3565. Since this PR remained in draft, and given the long delay on our side, we weren’t sure whether you were still available to continue working on it. We also couldn’t take over the branch directly because the repository settings don’t allow maintainers to push changes to contributor PR branches.

#3582 carries forward the main approach from this PR, with a few additional changes and regression tests identified during review. Closing this PR is therefore procedural and does not mean the original work was incorrect.

Thank you for identifying the issue and preparing the initial fix, and sorry again that we weren’t more responsive.

larry-c-ef:
> apologies for how slowly we responded

@lquerel Please do not worry about that. I deliberately left it in draft because I didn't think I'd have the bandwidth to respond to comments; I had just enough to keep it up to date with `main` so it wouldn't be auto-closed due to inactivity. The fault was entirely mine. I'm glad y'all were able to make use of the core code.

PR #2989, opened 2026-05-17
chore(ci): add PR size labeler workflow

cijothomas:
Adds a GitHub Actions workflow that automatically labels pull requests by size (`size/XS` through `size/XL`) based on changed line count, and posts a friendly reminder comment when a PR hits `size/XL` (500+ lines).

Uses [`CodelyTV/pr-size-labeler`](https://github.com/CodelyTV/pr-size-labeler) (390+ stars, MIT, actively maintained), pinned to a commit SHA per OSSF Scorecard best practices.

### Thresholds (Kubernetes / Prow-aligned)

| Label | Lines changed |
|---|---|
| `size/XS` | 0–9 |
| `size/S` | 10–29 |
| `size/M` | 30–99 |
| `size/L` | 100–499 |
| `size/XL` | 500+ (triggers reminder comment) |

### Notes

- Triggered via `pull_request_target` (same pattern as the existing `labeler.yml`). No PR code checkout occurs, so this is safe with fork PRs.
- Generated/vendored files (`go.sum`, `Cargo.lock`, `*.pb.go`, `testdata/`, `vendor/`) are ignored to avoid skewing counts.
- Draft PRs are skipped.
- Does not fail CI — only labels and (for XL) comments.

### Follow-up (if approved)

The labels (`size/XS` ... `size/XL`) will need to be created in the repo (one-time admin task). Suggested color scheme matches Kubernetes' Prow defaults.

Follow-up to https://github.com/open-telemetry/otel-arrow/pull/2864#issuecomment-4392806533 from @drewrelmas.


.github/workflows/pr-size-label.yml
```diff
@@ -0,0 +1,56 @@
...
+          s_label: 'size/S'
+          s_max_size: '29'
+          m_label: 'size/M'
+          m_max_size: '99'
+          l_label: 'size/L'
+          l_max_size: '499'
+          xl_label: 'size/XL'
+          fail_if_xl: 'false'
+          message_if_xl: >
+            **This PR is `size/XL` (500+ changed lines).** Large PRs are
```
drewrelmas, line 32: I would love for the Github Insights tab to give me concrete numbers to back up my gut feeling, but I think the `XL` size should probably be around 1500 lines in our repo based on historical trend.

gouslu: The action seems to cap L size to 1000 lines, so what you are asking seems impossible, despite being 100% in agreement with you. Especially considering that with AI, chloggen, comments, reaching to 500 lines is much easier than before.

**Btw, here is our stats as calculated by copilot based on last 300 PRs:**

### Sample summary

Source: 300 most-recent merged PRs in [`open-telemetry/otel-arrow`](https://github.com/open-telemetry/otel-arrow) (raw `additions + deletions`, no ignore list applied).

| Statistic | Value |
| --- | ---: |
| Sample size (`n`) | 300 |
| Mean | 2,991 |
| Median | 92 |
| p75 | 438 |
| p90 | 1,201 |
| p95 | 4,420 |
| p99 | 91,009 |
| Max | 219,221 |

> Mean is heavily skewed by a handful of dashboard/data-dump PRs (max ≈ 219k lines). The median (92) and p75 (438) are the right shape signals.

### Bucket distribution: PR-proposed vs. XL ≥ 1000

Only the L → XL boundary changes; XS / S / M brackets are identical.

| Bucket | PR-proposed range | Count | %     | Bumped range | Count | %     |
| ------ | ----------------- | ----: | ----: | ------------ | ----: | ----: |
| XS     | ≤ 9               |    55 | 18.3% | ≤ 9          |    55 | 18.3% |
| S      | 10 – 29           |    36 | 12.0% | 10 – 29      |    36 | 12.0% |
| M      | 30 – 99           |    66 | 22.0% | 30 – 99      |    66 | 22.0% |
| L      | 100 – 499         |    77 | 25.7% | 100 – 999    |   107 | 35.7% |
| **XL** | **≥ 500**         |  **66** | **22.0%** | **≥ 1000** |  **36** | **12.0%** |

**Net effect of moving the XL cutoff from 500 → 1000:** XL drops from **1-in-5** PRs to **1-in-8**; the 30 displaced PRs all fall into L. XS/S/M are untouched.

gouslu: If we calibrate for our repo history, 1000 for XL seems good, and we should probably adjust XS, S, M, L, XL accordingly. If we see that 1000 is still too much, we can reduce this gradually, but I think it'd be disruptive to start with 500, which is like almost every 1 in 4 PRs.

.github/workflows/pr-size-label.yml
```diff
@@ -0,0 +1,56 @@
...
+            If the size is unavoidable (e.g., a large refactor or generated
+            code update), please explain the reasoning in the PR description.
+            See
+            [CONTRIBUTING.md](https://github.com/open-telemetry/otel-arrow/blob/main/CONTRIBUTING.md#keep-pull-requests-small-and-incremental)
+            for guidelines on keeping PRs small.
+          files_to_ignore: |
+            "go.sum"
+            "Cargo.lock"
+            "*.lock"
+            "*.pb.go"
```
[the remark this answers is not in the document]
ThomsonTan: Besides the quotes mentioned by copilot, the pattern `go.sum`, `Cargo.lock` also don't work because they are not in the repo root. Need wildchard to match the prefix.

jmacd (approved):

.github/workflows/pr-size-label.yml
cijothomas: Thanks for the feedback. One thing to note is - just because the repo has been doing big PRs in the past does not mean we should continue that practice. I'll check few new PRs and adjust this surely.

.github/workflows/pr-size-label.yml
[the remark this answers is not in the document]
cijothomas: Thanks for catching this — fixed in #3313.

Issue #2858, opened 2026-05-06
Logs receiver for systemd / journald logs


jmacd:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

On UNIX hosts, we are interested in a journald logs receiver for otap-dataflow. 

### Proposed Solution

A low-code way to do this would be to exec the "journald" command, which supports paging through the logs with a cursor, the --after-cursor flag and the __CURSOR attribute on each entry. These logs can be emitted as JSON. This would be inefficient for us. 

https://systemd.io/ has the official documentation on this format, which can be accessed directly through existing C libraries with `#include <systemd/sd-journal.h>`.

### Alternatives Considered

We should study how other logs collectors do this more specifically. We will likely encounter a desire to co-locate the receiver with the CPU and NUMA region where the journal disk is attached.

jmacd:
Probably connected with / overlaps with https://github.com/open-telemetry/otel-arrow/issues/2844

lalitb:
I think this should be treated as separate journald receiver, aligned with #2844 but not implemented as a filelog variant.

The source model is different. Filelog discovers normal files, reads bytes, frames records, handles rotation, and checkpoints by file identity + byte offset. Journald should not watch or tail `.journal` files directly. It should read structured entries through [sd-journal](https://www.freedesktop.org/software/systemd/man/latest/sd-journal.html) and checkpoint by the opaque journald cursor (`__CURSOR`).
 As #2844 evolves, I think we should keep the reusable contracts source-neutral where practical, so journald can share the same receiver-level model without being forced into file-specific mechanics. The likely shared pieces are bounded backpressure, Ack/Nack-aware progress, checkpoint envelope/storage conventions, lifecycle/drain behavior, and generic partition assignment / placement hints.

The source-specific parts remain different: 
- filelog: file discovery, inode/device identity, byte offsets, framing, rotation
- journald: journal source selection, `sd-journal` iteration, cursor handling, retention/cursor-loss behavior

For `journald`, discovery is simpler: the initial partition can be a logical source such as `journal/system`. Later this could extend to user journals or journal namespaces. If #2844 produces a generic assignment extension, journald can use it to assign logical journal partitions to receiver instances, but it should not depend on file-discovery mechanics.

Proposed path:
 
1. Start with a short Linux-only design doc.
[...]

PR #3001, opened 2026-05-18
[Journald receiver]  Add journald receiver design for OTAP dataflow

lalitb:
### Fixes
Towards #2858 

### Description

Adds the design document for an OTAP-native journald receiver. 
This covers the intended v1 shape at a high level: 
    - local Linux journald ingestion, 
    - the worker/async execution model, 
    - Ack-driven cursor checkpointing, 
    - bounded handoff behavior,

The doc is intended to review the architecture and operational contract before the implementation PR.


rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+| --- | --- |
+| Source API | `sd-journal` via a small internal `libsystemd` FFI wrapper loaded at runtime |
+| Progress unit | Opaque journald cursor (`__CURSOR`) |
+| First implementation | Linux-only, default local system journal, single-instance source pipeline (one core) |
+| Delivery model | At-least-once from the last committed cursor |
+| Checkpoint advance | Only after downstream Ack and durable checkpoint write |
+| Backpressure | Stop calling `sd_journal_next()` when the bounded handoff is full |
+| Semantic processing | Kept out of the receiver; processors do normalization, parsing, routing |
+| Blocking calls | Isolated on a dedicated worker thread, never on the df-engine async task |
+| NUMA | Not addressed in v1; pinning/co-location is future work |
```
lquerel, line 34: Really nice summary

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+```mermaid
+flowchart LR
+    J["systemd journal<br/>accessed via sd-journal API"] --> W["JournalWorker<br/>blocking thread"]
+    W -->|"bounded batches<br/>batch_id + cursor range"| A["Async receiver task<br/>df-engine runtime"]
+    A --> P["Processors<br/>parse / enrich / route"]
+    P --> E["Exporters"]
+    E -->|"Ack / Nack"| A
+    A -->|"Commit { cursor }"| W
+    W --> C[("Checkpoint file<br/>namespace -> cursor")]
+```
```
lquerel, line 110: Thanks for this diagram

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+receiver into explicit `PRIORITY=N` matches. The default should include all
+levels `0..=7`; it should not silently drop debug entries.
+
+Filter changes are not retroactive. If filters are widened after a checkpoint
+exists, the receiver resumes from the existing cursor and does not backfill
+older entries that now match.
+
+## Field Projection
+
+The receiver performs only mechanical OTLP projection. It preserves native
```
lquerel:
```suggestion
The receiver performs only mechanical OTAP projection. It preserves native
```

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+emits OTAP log records. It reads through the `sd-journal` API, not by tailing
+`.journal` files and not by execing `journalctl`.
+
+The receiver is a source-specific sibling to the proposed filelog receiver in
+[#2844](https://github.com/open-telemetry/otel-arrow/issues/2844). It is **not**
+a filelog variant: its progress unit is an opaque journald cursor and its
+source API is `sd-journal`, not file discovery and byte offsets. It should not
+depend on the #2844 filelog assignment extension landing first.
+
+## Core Decisions
```
lquerel, line 22: I think this design would benefit from a small "Go receiver compatibility / classification" matrix, aligned with the reference-informed design guide. For each major Go receiver behavior, we should say whether we preserve, improve, decompose, defer, or reject it. Important rows: `journalctl` backend, `start_at`, cursor storage, priority default, `units`, `matches`, `grep`, `dmesg`, `directory/files/root_path`, `operators`, and field projection. This would make it clear which differences are intentional OTAP-native choices rather than accidental incompatibilities.

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+
+| Decision | Choice |
+| --- | --- |
+| Source API | `sd-journal` via a small internal `libsystemd` FFI wrapper loaded at runtime |
```
lquerel: Even if we only implements `sd-journal`, I'd suggest introducing a small `JournalSource` abstraction now. Something like `open`, `next_entry`, `wait`, `seek_after_cursor`, and `close` would give us a `FakeJournalSource` for deterministic tests and keep the receiver logic independent from the concrete sd-journal backend. It also leaves room for future `journalctl`, offline-file, or remote journal sources without rewriting the receiver.

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+    W --> C[("Checkpoint file<br/>namespace -> cursor")]
+```
+
+Flow:
+
+1. At startup, the async receiver task validates configuration, takes the
+   process-local namespace lease, and starts one blocking worker thread for the
+   assigned journal source.
+2. The worker opens `sd-journal`, applies filters, seeks from the checkpoint or
+   configured start position, and builds bounded log batches.
```
lquerel: We should define checkpoint resume as "start after the committed cursor", not merely "seek from checkpoint". With `sd_journal_seek_cursor`, the implementation needs to seek, call `sd_journal_next`, use `sd_journal_test_cursor` to detect whether the selected entry is the committed cursor, and skip it if so. Otherwise we can either replay the last committed entry or mishandle a vacuumed/stale cursor. Suggested addition: a short pseudo-code block for `seek_after_committed_cursor(cursor)` plus an explicit invalid/stale cursor policy.

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+| Backpressure | Stop calling `sd_journal_next()` when the bounded handoff is full |
+| Semantic processing | Kept out of the receiver; processors do normalization, parsing, routing |
+| Blocking calls | Isolated on a dedicated worker thread, never on the df-engine async task |
+| NUMA | Not addressed in v1; pinning/co-location is future work |
+
+The first implementation uses runtime-loaded `libsystemd` FFI because the
+receiver needs only a narrow `sd-journal` API surface and should keep
+`libsystemd` as a runtime dependency for this receiver path, instead of adding
+native `pkg-config`/linker setup to normal builds. If `libsystemd.so.0` is not
+available when the receiver starts, startup fails with an explicit error.
```
lquerel, line 40: Since this design intentionally uses sd-journal instead of execing journalctl, the container/host-journal story should be documented more explicitly. The Go receiver can rely on journalctl_path / root_path patterns, but with libsystemd the receiver needs the container image to provide a compatible libsystemd.so.0, the host journal directories to be mounted, and the process to have sufficient permissions. Could we add a dedicated "Container / host journal access" section and define the supported config shape? For example: `journal.root_path: /host` using `sd_journal_open_directory(..., SD_JOURNAL_OS_ROOT...`

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+   the worker without additional journal reads.
+
+In v1 the receiver reads one logical journal source per receiver instance
+(default: the local system journal). When #2844 introduces assignment, startup
+namespace selection can be replaced with assignment events without changing the
+read, Ack, or checkpoint model.
+
+## Startup and Instance Model
+
+`namespace` is the stable OTAP source identifier. In v1, only `system` is
```
lquerel: I'd split the stable OTAP source identifier from the actual systemd journal namespace. Right now namespace: system appears to mean both "OTAP source id" and "systemd namespace selection". In systemd, the default namespace is selected by passing NULL, while named namespaces are actual systemd namespace names => suggested shape: source_id: system plus journal.namespace: null, or rename the systemd-specific field to journal_namespace. This will avoid future confusion with the Go receiver's namespace option.

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+blocking filesystem I/O (`write`, `fsync`, `rename`). None of these run on the
+df-engine per-core async pipeline task.
+
+The receiver uses one long-lived blocking worker thread per assigned namespace:
+
+- worker owns the `sd_journal*` handle
+- async task owns the engine `EffectHandler`, lifecycle state, and Ack tracker
+- bounded worker-to-async channel carries completed batches
+- bounded async-to-worker `sync_channel` carries commit, rewind, and shutdown
+  commands
```
lquerel, line 181: The bounded sync_channel is fine on the worker side, but we should explicitly state that the async task must never perform a blocking send on the df-engine runtime thread. Commits, rewinds, and shutdown should use try_send, an async channel, or another non-blocking handoff. If the command channel is full, that should be treated as a lifecycle/protocol error rather than blocking the single-threaded runtime.

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+    Wait --> Next
+    Build --> Ready{Batch ready?}
+    Ready -->|no| Next
+    Ready -->|yes| Send["try_send batch"]
+    Send -->|sent| Next
+    Send -->|full| Hold["Hold batch<br/>do not read more"]
+    Hold --> PollShutdown["Poll shutdown command"]
+    PollShutdown -->|shutdown| Stop["Stop worker"]
+    PollShutdown -->|none| Send
+```
```
lquerel, line 221: The read loop diagram currently sends a batch and then immediately goes back to Next. That conflicts with the later statement that v1 allows only one in-flight batch per namespace. With max_in_flight_batches = 1, after a successful handoff the worker should stop reading until the batch receives a terminal Ack/Nack and the commit/rewind decision is processed. Otherwise we either have more than one in-flight batch or we need to define ordered multi-batch cursor commits.

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+## Field Projection
+
+The receiver performs only mechanical OTLP projection. It preserves native
+journal fields as attributes and leaves semantic-convention mapping to
+processors.
+
+| OTAP field | Source |
+| --- | --- |
+| `body` | `MESSAGE`, unset when missing |
+| `time_unix_nano` | default `__REALTIME_TIMESTAMP`; `_SOURCE_REALTIME_TIMESTAMP` can be a future option |
```
lquerel: Since this receiver uses sd-journal directly, I'd avoid describing timestamp/cursor extraction as reading `__REALTIME_TIMESTAMP` / `__CURSOR` fields. Those are address fields exposed in external formats. The design should specify the APIs instead: `time_unix_nano <- sd_journal_get_realtime_usec() * 1000` and internal `cursor <- sd_journal_get_cursor()`. Also worth stating that data returned by `sd_journal_get_data` / enumeration must be copied before advancing the journal pointer.

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+| `severity_number` | derived from `PRIORITY` |
+| `attributes` | all other native journal fields, key names preserved |
```
lquerel: "All other native journal fields" needs explicit duplicate-field and binary-field rules. Journald permits binary values and non-unique field names. Suggested rule: single occurrence becomes a scalar attribute; repeated fields become an array; binary values are preserved as OTAP bytes when supported, otherwise base64 encoded with an explicit encoding marker. For repeated `MESSAGE`, we could use the first value as body and preserve all values in an attribute.

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+| `2` | critical | `FATAL2` |
+| `3` | error | `ERROR` |
+| `4` | warning | `WARN` |
+| `5` | notice | `INFO2` |
+| `6` | info | `INFO` |
+| `7` | debug | `DEBUG` |
+
+Binary `MESSAGE` handling depends on the current OTAP body/attribute support.
+If bytes can be represented directly, preserve bytes. Otherwise encode as
+base64 and mark the encoding explicitly; do not lossy-decode.
```
lquerel: Please add hard resource limits for field and entry extraction. sd_journal_set_data_threshold is only a hint, not a safety boundary, and large compressed fields can still be expensive.
Suggested config => max_entry_bytes, max_field_bytes, max_fields_per_entry, and large_field_policy: drop_and_count

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+If the journal is on tmpfs, overlayfs, a bind mount, or a device that cannot be
+resolved, the NUMA node should be reported as unknown.
+
+Future goal:
+
+```text
+journal storage NUMA node -> journald worker thread -> same-node pipeline
+```
+
+## Implementation Scope
```
lquerel, line 581: Could we include receiver self-telemetry in the first implementation scope? This receiver will be operationally sensitive. Useful counters/gauges: entries read/emitted/dropped, batches emitted/acked/nacked, checkpoint commits/failures, invalid/stale cursor events, permission/partial-journal warnings, field drops, large-field drops, last entry timestamp, worker restarts, and backpressure duration

lquerel (approved):
I really like the approach. I added a few feedback comments to help consolidate the overall design. Thanks.

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -0,0 +1,454 @@
...
+
+The first implementation uses runtime-loaded `libsystemd` FFI because the
+receiver needs only a narrow `sd-journal` API surface and should keep
+`libsystemd` as a runtime dependency for this receiver path, instead of adding
+native `pkg-config`/linker setup to normal builds. If `libsystemd.so.0` is not
+available when the receiver starts, startup fails with an explicit error.
+
+## Journald vs Filelog
+
+Journald is intended to be a separate receiver, not a filelog variant. Even
```
ThomsonTan, line 52: Does this imply that the jornald and filelog receiver are totally separated?

lalitb: yes - they can share the reusable components.

lalitb:
lquerel Thanks for the detailed review. I agree with the feedback and pushed an update addressing the comments here -  https://github.com/open-telemetry/otel-arrow/pull/3001/commits/7a68c1318ea9a1925922bdb0aec69c6ba3db1dd2.

jmacd (approved):

PR #3006, opened 2026-05-18
feat(comparison_dashboard): Create logs, metrics, and traces DFE comparisons

JakeDern:
# Change Summary

Update the manifest to point to three new comparisons for the dataflow engine - One each for logs, metrics, and traces. This is in preparation to publish logs/metrics/traces results for the dataflow engine.

Note that this PR does not publish data, we will do that separately.

## What issue does this PR close?

* Closes #2992

## How are these changes tested?

image

## Are there any user-facing changes?

Yes, new charts on the dashboard.

PR #3008, opened 2026-05-18
feat(comparison_dashboard): Specify tests for comparison in the manifest and comparison metadata

### LINKED ISSUE #3005: [Comparison Dashboard] Define visible tests for a dashboard in comparison definitions
### Pre-filing checklist

### Component(s)

Other

### Objective

To be able to define the names of tests to be plotted in the comparison definition rather than be inferred from the existing data/results.

### Rationale

We test many loadgen rates for different engines. Sometimes we may want to plot more or less of those rates depending on:

 - How many rates fit on the graph 
 - What rates show interesting data for a particular test
 - Whether or not some test suites have data for a particular rate
 
For example, we may only be able to fit 8 rates comfortably on a chart given the number of dimensions that we have. One engine may top out at around 400k signals/sec and it makes sense to plot 100k, 200k, 300k, 400k to demonstrate the scaling characteristics of that engine.

A second engine may top out at around 2000k signals/sec and it makes sense to then plot a higher range like 500k, 1000k, 1500k, 2000k on the same chart.

This will also give additional flexibility in what shows up on the x-axis e.g. we could instead compare on commit hashes to do a historical comparison instead of one based on rate.

### Scope

Add a section `tests` to the chart to explicitly configure the names of tests that should be plotted in addition to other metadata:
 
- `name`: The name of the test
- `loadgen_rate`: The expected loadgen rate for the test
- `label`: The label to use for the test on the chart

### Acceptance Criteria

.

### Dependencies or Blockers

_No response_

### Additional Context

_No response_

JakeDern:
# Change Summary

Updates the site so that the tests to be plotted for comparison are specified via some metadata rather than be inferred/hardcoded into the site.

Waiting on #3006 for a rebase.

* Closes #3005 

image

No.

### DIFF
--- a/tools/comparison_dashboard/comparisons/dfe_logs_baselines.yaml
+++ b/tools/comparison_dashboard/comparisons/dfe_logs_baselines.yaml
@@ -2,6 +2,23 @@ slug: dfe_logs_baselines
 name: Dataflow Engine Logs
 description: Baseline DFE logs runs across OTAP, OTLP, and OTLP/HTTP with none, gzip, and zstd compression.
 
+tests:
+  - name: 100k
+    loadgen_rate: 100000
+  - name: 200k
+    loadgen_rate: 200000
+  - name: 300k
+    loadgen_rate: 300000
+  - name: 400k
+    loadgen_rate: 400000
+  - name: 600k
+    loadgen_rate: 600000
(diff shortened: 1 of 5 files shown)


lquerel (approved):
LGTM

PR #3011, opened 2026-05-18
feat(comparison_dashboard): Specify metrics metadata via manifest

### LINKED ISSUE #3007: [Comparison Dashboard] Define metric labels in the manifest and visible metrics in the comparison definition
### Pre-filing checklist

### Component(s)

Other

### Objective

This is in the same vein of making the comparison dashboard a little more generic as #3005. The specific goals for this task are to allow:

1. Specifying metric labels and display text without modifying app.js code
2. Allow modifying the visible metrics on a per comparison basis

### Rationale

Raw metric names such as `produced_logs_signals` are not friendly for reading, so the dashboard has to have some mapping from metric name -> display name in order to get nice things on the chart. 

Additionally, we have to make some choices about which metrics are suitable to be displayed in the chart. Some metrics such as `test_duration` are not always useful to have on the chart and just clutter the dropdown menu.

Finally, it's conceivable that we would want to make different selections about what metrics go in the chart for different comparisons that use different methodologies.

All of these parameters today are hardcoded into the app.js code. Instead, we should make them specifiable via manifest metadata like most other things.

### Scope

- Add a new `metrics` field to the manifest that allows specifying the required metadata for each metric
- Add a new `metrics.chart` field to the comparison definitions that allow specifying the metrics that should be displayed on the chart
- Update all manifests and comparisons appropriately
- Update the build script to place this information in relevant data.js files
- Update the site code to read this information from data.js files

### Acceptance Criteria

.

### Dependencies or Blockers

_No response_

### Additional Context

_No response_

JakeDern:
# Change Summary

This PR makes the dashboard a little more generic by also pulling out test metric related parameters to the manifest.

Waiting on #3008 for a rebase.

* Closes #3007

image

No.

### DIFF
--- a/tools/comparison_dashboard/dashboard.py
+++ b/tools/comparison_dashboard/dashboard.py
@@ -70,6 +70,7 @@ class Manifest:
     comparison_files: list[Path]  # absolute paths to comparison YAMLs
     variables: dict               # template variables passed to Jinja at top level
     meta: dict                    # allowed values per meta key (key -> [values])
+    metrics: list                 # ordered list of {name, label} for dashboard metrics
 
 def load_manifest(manifest_path: Path) -> Manifest:
     """Load and validate the manifest file. Resolves all listed paths."""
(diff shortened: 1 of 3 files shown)


lquerel (approved):
LGTM

JakeDern:
This PR is ready now.

PR #3021, opened 2026-05-18
docs: document saturation test workload characteristics

cijothomas:
Document that saturation tests use static 1KB log bodies with realistic entropy (512 unique bodies), distinguishing them from other tests that use semantic_conventions (~300 byte logs). Also removes the stale TODO and adds scaling efficiency formula explanation with link to the scaling-efficiency benchmark page.


lquerel (approved):
LGTM

PR #3026, opened 2026-05-19
Add uncompressed bytes-per-log metric to traffic generator benchmarks

### LINKED ISSUE #2987: Add uncompressed bytes-per-log metric to traffic generator
## Summary

Add a metric to the traffic generator receiver that exposes the average uncompressed protobuf size per log record produced. This would make it easy to compute compression ratios from benchmark reports without manual estimation.

## Context

The traffic generator already has access to `OtapPayload::num_bytes()` (serialized protobuf size) and `num_items()` at the point where it sends data. The metrics struct (`TrafficGeneratorReceiverMetrics`) already exposes counters via Prometheus, which the orchestrator scrapes.

Currently, a human reviewing benchmark results has to estimate the uncompressed log size from knowledge of the data source config (body size, attribute count, trace context). Exposing this as a metric would make the compression ratio self-documenting in reports.

## Proposed approach

1. Add a gauge or running average metric (e.g., `bytes_per_item_produced`) to `TrafficGeneratorReceiverMetrics`
2. In `export_pdata()`, compute `payload.num_bytes() / payload.num_items()` and record it
3. Wire the metric into the orchestrator report SQL to produce an `uncompressed_bytes_per_log` entry in `gh_actions_benchmark`

This enables a derived compression ratio: `egress_bytes_per_log / uncompressed_bytes_per_log`.

Relates to #2540

cijothomas:
Closes #2987

Adds a `logs_bytes_produced` counter metric to the traffic generator receiver that tracks the total protobuf-encoded (uncompressed) bytes of log payloads produced. The benchmark report SQL then computes `uncompressed_bytes_per_log` from this counter, enabling direct comparison of uncompressed payload size against the egress (compressed) bytes per log.

### Changes
- **metrics.rs**: Added `logs_bytes_produced: Counter<u64>` with unit `By`
- **mod.rs**: Record payload bytes in `export_pdata()` for log signals (captured before ownership move)
- **integration_report_logs.yaml** & **report_logs.yaml**: Added `logs_bytes_produced` to metric filter and `uncompressed_bytes_per_log` computed metric to report SQL


cijothomas (commented):
Addressing review feedback:

1. **manifest.yaml**: Added `uncompressed_bytes_per_log` (and the also-missing `egress_bytes_per_log` from PR #2982) to `tools/comparison_dashboard/manifest.yaml`.

2. **Test for logs_bytes_produced**: The existing tests don't verify any metrics counters (`logs_produced`, `spans_produced`, `metrics_produced`) — they validate message content/structure only. Adding a counter test just for `logs_bytes_produced` would be inconsistent with the existing pattern.

rust/otap-dataflow/crates/core-nodes/src/receivers/traffic_generator/metrics.rs
```diff
@@ -13,6 +13,11 @@ pub struct TrafficGeneratorReceiverMetrics {
     /// Number of logs generated.
     #[metric(unit = "{log}")]
     pub logs_produced: Counter<u64>,
+    /// Total uncompressed bytes of log payloads produced (protobuf-encoded size before
+    /// compression). Together with `logs_produced`, this yields the average uncompressed
+    /// bytes per log record for compression-ratio analysis.
+    #[metric(unit = "By")]
+    pub logs_bytes_produced: Counter<u64>,
```
albertlockett, line 20: nit: I'm wondering if the comments here about these logs specifically being uncompressed or the fact that we use these to calculate compression ratio belong here in this component? It feels like a comment that makes sense in the context of this PR, but generally the fact that some other component might compress these bytes or that we might compute a ratio from this seems outside the scope of concern of this component. WDYT? I'll admit I'm not deeply steeped in best practices for describing metrics, and this is simply my opinion

rust/otap-dataflow/crates/core-nodes/src/receivers/traffic_generator/mod.rs
```diff
@@ -380,12 +380,18 @@ impl TrafficGeneratorReceiver {
     ) -> Result<Result<u64, OtapPdata>, Error> {
         let signal = pdata.signal_type();
         let count = pdata.num_items() as u64;
+        let payload_bytes = pdata.payload_ref().num_bytes();
         match handler.try_send_message_with_source_node(pdata) {
             Ok(()) => {
                 match signal {
                     otap_df_config::SignalType::Traces => self.metrics.spans_produced.add(count),
                     otap_df_config::SignalType::Metrics => self.metrics.metrics_produced.add(count),
```
albertlockett, line 389: Will we eventually want/need the same for metrics and spans?

albertlockett (approved):
LGTM @cijothomas . A few minor comments/questions, but nothing I'd consider blocking

PR #3028, opened 2026-05-19
Skip continuous benchmark when only docs/non-rust files change

cijothomas:
Adds `paths` filter to the continuous pipeline performance test workflow so it only runs when relevant code is modified:

- `rust/**`
- `tools/pipeline_perf_test/**`
- `tools/comparison_dashboard/**`
- The workflow file itself

This avoids wasting expensive bare-metal runner time on commits that only touch markdown, Go code, or other unrelated files. `workflow_dispatch` still allows manual runs anytime.


lquerel (approved):
Good idea

PR #3034, opened 2026-05-19
Fix deterministic failure of integration_env_var_to_yaml on arm64 partition 2/3

cijothomas:
## Problem

`otap-df-config::config_provider::tests::integration_env_var_to_yaml` fails deterministically in CI on `test_nonrequired (otap-dataflow, ubuntu-24.04-arm, 2)` (nextest partition 2/3), including on `main`. Examples:

- https://github.com/open-telemetry/otel-arrow/actions/runs/26081151000 (main)
- https://github.com/open-telemetry/otel-arrow/actions/runs/26075261669 (main)
- https://github.com/open-telemetry/otel-arrow/actions/runs/26077655441 (PR #3031)

Panic:
```
thread 'config_provider::tests::integration_env_var_to_yaml' panicked at reqwest-0.13.3/src/blocking/client.rs:1522:5:
```
```
thread 'reqwest-internal-sync-runtime' panicked at reqwest-0.13.3/src/async_impl/client.rs:2461:5:
event loop thread panicked
```

This is the standard rustls *"no default crypto provider installed"* panic.

## Root cause

The test calls `resolve_config(Some("env:..."))` → `default_resolver()` → constructs an `HttpConfigProvider` which builds a `reqwest::blocking::Client`. rustls requires a process-wide default crypto provider, which the test module installs lazily via `ensure_crypto_provider()`. Other tests that build the default resolver call this helper first (`bare_path_treated_as_file`, `unknown_scheme_returns_error`), but `integration_env_var_to_yaml` did not. Whether it passes depends on whether some other test in the same process happened to install the provider first — which on the arm64 partition 2/3 ordering, never happens.

## Fix

Call `ensure_crypto_provider()` at the start of `integration_env_var_to_yaml`, matching the other tests that use `default_resolver()`.


AaronRM:
FWIW these failures were introduced when #2696 merged yesterday.

jmacd (approved):

PR #3035, opened 2026-05-19
perf: keep network rate metrics in bytes/sec for dashboard formatting

cijothomas:
Follow-up to #2982 per [review suggestion](https://github.com/open-telemetry/otel-arrow/pull/2982#issuecomment-4464515872): keep report data in bytes/sec and let the dashboard handle rounding/reformatting into MiB/s or GiB/s as needed.


lquerel (approved):
LGTM

PR #3042, opened 2026-05-19
Standardize loadgen max_batch_size to 512 (OTel SDK default)

cijothomas:
Standardizes the loadgen `max_batch_size` to 512 across all benchmark tests (100kLRPS, passthrough, saturation max-throughput, and templates). This matches the OTel SDK batch processor default and aligns with the scaling efficiency tests which already used 512.

Previously these tests used 1000, which does not reflect real-world SDK behavior. The batch-size sweep nightly test (which intentionally varies batch sizes) and syslog tests (different loadgen) are unchanged.

### 100kLRPS (rate-limited to ~95k logs/sec, single core)

| Test | CPU Before | CPU After | Throughput |
|------|-----------|-----------|------------|
| OTLP-ATTR-OTLP | 64.7% | 64.7% | ~95k (unchanged) |
| OTLP-ATTR-OTAP | 65.0% | 65.0% | ~95k (unchanged) |
| OTAP-ATTR-OTAP | 23% | 38% | ~95k (unchanged) |
| OTAP-BATCH-OTAP | 18% | 33% | ~90k (unchanged) |

CPU increase for OTAP tests is expected — smaller batches mean less compression efficiency per batch in Arrow columnar format.

### Saturation Max Throughput (single core, CPU-bound)

| Test | Before (batch=1000) | After (batch=512) | Change |
|------|--------------------|--------------------|--------|
| OTLP | ~360k logs/sec | ~330k logs/sec | −8% |
| OTAP | ~2.6M logs/sec | ~1.8M logs/sec | −30% |

### Passthrough (single core, CPU-bound)

| Test | Before (batch=1000) | After (batch=512) | Change |
|------|--------------------|--------------------|--------|
| OTLP | ~607k logs/sec | ~568k logs/sec | −6% |
| OTAP | ~2.6M logs/sec | ~1.9M logs/sec | −28% |

Peak throughput reduction in saturation tests is expected — smaller batches mean more per-batch framing overhead and less compression efficiency. OTAP (Arrow columnar) is more affected than OTLP because it benefits more from larger batch sizes.

Validated across 4 CI runs on bare-metal — results are consistent (±2% variance).


JakeDern:
The efficiency match what I'd expect based on some tests I did a while back with batch sizes of 1500, 2000, and 5000 which saw very slight efficiency gains.

512 seems reasonable to test if that's generally what the SDKs default to - Probably worth throwing a batch processor in with similar defaults to what the OTC sets on batch processor/exporter helper and see how much of that perf we get back in each case.

For future tests, I wonder if we should consider adding batching as a standard part of every pipeline. It seems most common to me that that batching would be enabled, and I feel like batching being included in every exporter via the exporterhelper for OTC (though I think it's off by default) supports that direction overall.

JakeDern (approved):

cijothomas:
@jmacd @lquerel Please review this. Since this affects perf numbers, I want to make sure we are ok with this standadization. Of course, we have other tests specifically targetting batch size variations, as noted in PR desc.

JakeDern:
Now that we have some updated batch size data for both DFE and OTC, it would seem to me that the point of diminishing returns for OTC is around 1024 and for DFE it's around 2048. Both see performance improvements beyond this point, but definitely much smaller ones:

https://open-telemetry.github.io/otel-arrow/compare/log_attr_processor_insert_multi_batch/

I don't know that this information should change anything about this PR, but it's interesting that we have more per batch overhead, about double that of OTC, based on these results. Maybe that's a challenge to us to profile those tests and see if we can bring that down 🙂.

Issue #3002, opened 2026-05-18
If crypto features aren’t specified, use SymCrypt on Windows and aws-lc-rs on other platforms.


sapatrjv:
> The default crypto provider for Rustls (starting from version 0.23) is aws-lc-rs. Previously, ring was the default provider; however, it has been superseded because aws-lc-rs provides broader platform support, improved performance characteristics, and official FIPS certification, which is critical for compliance-focused and enterprise environments. The shift also aligns Rustls with a more actively maintained and forward-looking cryptographic backend that better supports diverse deployment targets and long-term security requirements.

However, the function ensure_crypto_provider is still using ring as the default provider, which creates an inconsistency with the current Rustls defaults.

> Specifically for Windows platforms, symcrypt should be used as the default crypto provider. This choice aligns with Windows-native cryptographic primitives and leverages deep Azure and Microsoft ecosystem integration. Using symcrypt also enables reuse of the existing internal crypto stack, reduces duplication, and ensures better compatibility with platform security features, updates, and validation processes.

lalitb:
Thanks for raising this. I think there may be two related but separate points here:
 
1. The `ring` -> `aws-lc-rs` part makes sense to me, since that matches rustls 0.23’s default provider change.
2. Could you help clarify the source or requirement for making `SymCrypt` the default specifically on Windows? My understanding is that `SymCrypt` is available through a third-party rustls provider, but it is not rustls’ upstream Windows default. If this is based on an OTAP/Azure/platform requirement, it would be helpful to call that out separately from the rustls default alignment.

sapatrjv:
You are right—SymCrypt is not the upstream default for Windows. However, SymCrypt is closely aligned with Windows-native cryptographic primitives and is designed to integrate deeply with the Azure and broader Microsoft ecosystem. This tight integration brings benefits such as better alignment with the Windows security model, consistent API behavior, and easier long-term maintenance within Microsoft-managed environments. As a result, SymCrypt is particularly well suited for Windows-specific scenarios where platform-native support, compliance, and ecosystem compatibility are important considerations. Given this, my thinking is to choose the most appropriate cryptographic implementation with respect to the platform’s in-default security posture.

PR #3043, opened 2026-05-19
feat: use aws-lc-rs as default crypto library instead of ring

sapatrjv:
Switch the default cryptographic provider from ring to aws-lc-rs across all crates that previously defaulted to crypto-ring. The crypto-ring feature remains available for opt-in use.

# Change Summary

The default crypto provider for Rustls (starting from version 0.23) is aws-lc-rs. Previously, ring was the default provider; however, it has been superseded because aws-lc-rs provides broader platform support, improved performance characteristics, and official FIPS certification, which is critical for compliance-focused and enterprise environments.

Related to https://github.com/open-telemetry/otel-arrow/issues/3002

cargo xtask check

 No


rust/otap-dataflow/crates/admin-api/Cargo.toml
```diff
@@ -16,7 +16,7 @@ workspace = true
 rustls-symcrypt = { workspace = true, optional = true }
 
 [features]
-default = ["http-client", "crypto-ring"]
+default = ["http-client", "crypto-aws-lc"]
```
lalitb, line 19: One more place may need updating: https://github.com/open-telemetry/otel-arrow/blob/a46fb9831c1ae90ec0429b7ad5b90054e10b9e90/rust/otap-dataflow/crates/otap/src/crypto.rs#L35

If `aws-lc` is now the default, should `crypto-aws-lc` be checked before `crypto-ring` in `install_crypto_provider()`? Otherwise, when Cargo feature unification enables both features, this function will still install the ring provider because `crypto-ring` is checked first.

sapatrjv: Agree. There are more changes and planning to checkin by EOW. Thats why I have marked PR as Draft.

PR #3046, opened 2026-05-19
Platform gating for specific configs

### LINKED ISSUE #3041: validate-configs CI check has no platform-gating, fails on platform-specific configs
## Summary

`scripts/validate-configs.sh` discovers every `version: otel_dataflow/v1` YAML in the repo and validates each one against a single `df_engine` binary. There is no notion of "this config is only valid on platform X", so any platform-specific YAML breaks the script on every other platform's CI runner.

This is currently failing the CI check for validate-configs on PR #2930 (ETW receiver) on the Linux CI run with:

```text
❌ rust/otap-dataflow/configs/etw-console.yaml
   Error: Custom { kind: Other, error: "Unknown receiver component `urn:otel:receiver:etw` in pipeline_group=default pipeline=main node=etw" }
```

## Root cause

Two things have to line up for a component to be registered, and neither is satisfied on the Linux runner for the ETW receiver:

1. The `etw-receiver` Cargo feature has to be enabled in the build. The script's default build line is `cargo build --features azure,aws,contrib-exporters,contrib-processors,contrib-receivers,recordset-kql-processor,azure-monitor-exporter,geneva-exporter,condense-attributes-processor,resource-validator-processor,user_events-eventheader` — `etw-receiver` is absent.
2. The receiver module is also gated on `target_os = "windows"` in [`crates/contrib-nodes/src/receivers/mod.rs`](crates/contrib-nodes/src/receivers/mod.rs):

   ```rust
   #[cfg(all(feature = "etw-receiver", target_os = "windows"))]
   pub mod etw_receiver;
   ```

   So even if the feature were enabled, the `#[distributed_slice(OTAP_RECEIVER_FACTORIES)]` registration would not be present on Linux. The receiver depends on ETW which is only on Windows, so this gate cannot be dropped.

The script discovers `configs/etw-console.yaml` like any other YAML and feeds it to a binary that has no concept of the ETW receiver, producing "Unknown receiver component".

The same trap exists in the opposite direction for `configs/user-events-tracefs-debug.yaml`, which is Linux-only by design (depends on `tracepoint`/`eventheader`). It happens to validate today because the runner is Linux, but the script offers no protection if and when validation moves to a Windows or macOS runner.

## Current behaviour

`scripts/validate-configs.sh`:

- Globs every `*.yaml`/`*.yml` containing `version: otel_dataflow/v1`.
- Runs `$BINARY --config "$config" --validate-and-exit` against each one.
- Treats any non-zero exit as failure and aborts the script with a summary.

There is no per-config metadata or filename convention to tell the script which platform a YAML targets.

## Proposed fix

Adopt an explicit, opt-in marker the script can read without parsing YAML. The most general option is a `# platform:` comment at the top of any platform-specific config:

```yaml
# platform: windows-only
version: otel_dataflow/v1
...
```

Permitted values: `windows-only`, `linux-only`, `macos-only` (extend as needed). Configs without the marker are platform-agnostic and validated everywhere, preserving current behaviour for the ~40 existing YAMLs.

The script grows a small dispatch around the validate call:

```bash
required_platform=$(sed -n 's/^# platform: //p' "$config" | head -1)
host_os="$(uname -s)"

case "$required_platform" in
    "")
        ;;  # platform-agnostic — always validate
    windows-only)
        case "$host_os" in
            MINGW*|MSYS*|CYGWIN*) ;;
            *) echo "  ⏭  $REL_PATH (skipped: requires Windows)"; SKIPPED=$((SKIPPED+1)); continue ;;
        esac ;;
    linux-only)
        if [ "$host_os" != Linux ]; then
            echo "  ⏭  $REL_PATH (skipped: requires Linux)"; SKIPPED=$((SKIPPED+1)); continue
        fi ;;
    macos-only)
        if [ "$host_os" != Darwin ]; then
            echo "  ⏭  $REL_PATH (skipped: requires macOS)"; SKIPPED=$((SKIPPED+1)); continue
        fi ;;
    *)
        echo "  ❌ $REL_PATH (invalid platform marker: $required_platform)"; FAILED=$((FAILED+1)); continue ;;
esac
```

Add a `SKIPPED` tally to the final summary so the skips are visible.

Apply the marker to the two known platform-specific configs:

- `configs/etw-console.yaml` — `# platform: windows-only`
- `configs/user-events-tracefs-debug.yaml` — `# platform: linux-only`

## Out of scope but worth following up

- `etw-receiver` is not in `validate-configs.sh`'s `--features` list. Even on a Windows runner the script would never actually exercise the ETW receiver. Add it once the platform-skip mechanism is in place.
- No CI workflow currently runs `validate-configs.sh` on Windows. The Linux skip means `configs/etw-console.yaml` is effectively unvalidated on every platform. A Windows job (or a `runs-on: windows-latest` matrix entry) that runs the same script would close that gap.
- Consider extending the marker to take multiple values (`# platform: linux-only, macos-only`) if a config is non-Windows but otherwise portable.

## Acceptance criteria

- `scripts/validate-configs.sh` exits 0 on a Linux runner with `configs/etw-console.yaml` present.
- The summary line shows the skip count, e.g. `Config validation: 39/40 passed, 0 failed, 1 skipped.`
- A YAML carrying an unknown `# platform:` value fails the script with a clear message.
- `configs/user-events-tracefs-debug.yaml` is marked `linux-only` and continues to validate on Linux.

swashtek:
# Change Summary

The change adds platform gating for validate config script. 
Validate-configs CI check has no platform-gating, fails on platform-specific configs. 

* Closes #3041

Validate config check

No


swashtek:
Closing in favor of #3060

PR #3047, opened 2026-05-20
df_engine: MVP weaver live-check for in-tree semconv events (#1613)

cijothomas:
Initial wiring for #1613: validates one internal event (`tls.handshake.failed`) end-to-end via `weaver registry live-check` in CI. Establishes the registry → emit → live-check mechanism so subsequent PRs can backfill the remaining events and tighten gates.

Builds on #3049 (event_name / InstrumentationScope.name encoder change), which is already merged.

## Changes

- New in-tree semconv registry at `rust/otap-dataflow/semconv/` with one `type: event` group. `manifest.yaml` pulls upstream `semantic-conventions@v1.41.0` via `dependencies:` (no second checkout in CI).
- New `configs/internal-events-otlp.yaml` wires `internal_telemetry` → `otlp_grpc` and a TLS-enabled `otlp` receiver whose handshakes are failed by plaintext HTTP from CI.
- New workflow `.github/workflows/df-engine-internal-observability.yml` holding the live-check job. Kept out of `rust-ci.yml` to convey long-term intent and avoid polluting the rust workflow as the registry grows. **Not in required status checks yet**, mirroring the staged rollout used for the host-metrics live-check. The assert step is registry-driven: it discovers every declared event and fails if any received zero samples or has event-level violations.

## Verified locally

`weaver registry check` clean; `cargo xtask check` green; end-to-end smoke produced `tls.handshake.failed` samples with 0 event-level violations.

## Deferred (follow-ups under #1613)

Backfill remaining event names; attribute-level alignment (e.g. `error` → `error.type` per OTel semconv); `InstrumentationScope.version`; xtask static drift check; promote workflow to required. Severity declaration in semconv is blocked on open-telemetry/weaver#1004 (the wire already carries `severity_number = 13`).

Follow-up: adopt `weaver-live-check-{start,stop}` composite actions once open-telemetry/weaver#1448 merges to drop ~half the workflow boilerplate.

PR #3049, opened 2026-05-20
telemetry: encode event_name as bare name; emit crate via InstrumentationScope.name

cijothomas:
Align df_engine's OTLP log output with the OpenTelemetry logs data model: emit the crate name via `InstrumentationScope.name` instead of prepending it to `LogRecord.event_name`. Stands on its own merit irrespective of the parallel weaver live-check work (#3047, will rebase once this lands).

## Today

`encode_event_name` writes `LogRecord.event_name` as `"<target>::<name>"` (e.g. `otap-df-otap::tls.handshake.failed`). `InstrumentationScope.name` is left empty. This conflates two distinct OTel concepts: the emitting library identity (scope) and the stable event identifier (event.name).

## With this change

```
InstrumentationScope.name  = "otap-df-otap"           (new: was empty)
LogRecord.event_name       = "tls.handshake.failed"   (was "otap-df-otap::tls.handshake.failed")
```

Matches the OTel logs spec: `instrumentation_scope.name` identifies the emitting library; `event.name` is a stable, library-independent identifier.

## Out of scope, left unchanged

- The console formatter still renders `target::name` for stdout/stderr -- humans tailing logs see the same string as before. Only OTLP-on-the-wire is affected.
- `InstrumentationScope.version` is still empty (could later be populated from `env!("CARGO_PKG_VERSION")`).
- The admin/log_tap path already separated `target` and `event_name` (`crates/admin/src/telemetry.rs:214-215`); no change needed there. This PR brings the OTLP exporter in line with what admin already does.

### DIFF
--- a/rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
+++ b/rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
@@ -81,13 +81,17 @@ impl<'buf, B: BoundedBuf> DirectLogRecordEncoder<'buf, B> {
 }
 
 /// Encode the event name from callsite metadata.
-/// Format: "target::name (file:line)" or "target::name" if no file/line.
+///
+/// Emits the bare `callsite.name()` (the first argument to `otel_info!` /
+/// `otel_warn!` / …). The crate name (`callsite.target()`) is **not**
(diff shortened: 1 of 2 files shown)


lquerel (approved):
LGTM

PR #3051, opened 2026-05-20
Add initial OPL specification

gbracha:
# First draft of OPL specification

LaTeX sources for an initial draft specification for OPL (the OTEL Programming Language). Includes the spec and a style file, and, for the convenience of those who do not wish to bother running LaTeX, a PDF.

None.

It's a spec, not executable code.

Not until this spec is approved and the implementation integrated. Quite a way off.


lquerel:
**Message for reviewers**

I would like to personally thank @gbracha, who agreed to help us formalize the initial OPL draft that @albertlockett and I started a few months ago. Gilad is a recognized expert in the field of programming languages. He has authored or contributed to the specification of several languages, including Java, Dart, and Newspeak. So we are in very good hands!

This specification will be complemented by a companion document intended to provide a less formal version of the spec, with many examples.

albertlockett:
Removing stale label - waiting on CLA process

albertlockett:
Note for reviewers/those interested in OPL, we have a user-guide too which is a gentler introduction than the spec: https://github.com/open-telemetry/otel-arrow/pull/3182

rust/otap-dataflow/docs/opl/OPLSpec.tex
```diff
@@ -0,0 +1,1229 @@
...
+
+\REMOVE{} can also be referenced using the predefined alias \PROJECTREMOVE.
+
+\subsection{Rename}
+\label{rename}
+
+
+The \RENAME{} operator allows for renaming a set of dynamic attributes of a stream eleement.
+
+\RENAME{} can also be referenced using the predefined alias \PROJECTRENAME.
```
albertlockett: @gbracha , after discussing a bit with @lquerel we have decided that a good syntax for rename will be:

```
logs | rename attributes "old.key1" as "new.key1", "old.key2" as "new.key2" //, etc.
```

Given that this diverges from the `project-rename` operator from KQL which inspired this operator, we've decided that `project-rename` shouldn't be an  alias for rename.

Perhaps for now, we can just drop this phrase about the alias.

New syntax also documented in https://github.com/open-telemetry/otel-arrow/issues/3224

rust/otap-dataflow/docs/opl/OPLSpec.tex
```diff
@@ -0,0 +1,1229 @@
...
+\subsection{Extend}
+\label{extend}
+
+The \EXTEND{} operator allows the assignment of an element of a map.
+
+{\em The purpose of \EXTEND{} is to add or mutate dynamic attributes of a stream element.}
+
+{\em Or, equivalently, to an element of a list of key-value pairs.}
+
+Let $x$ be the current element of the input stream $s$. The operator call $\EXTEND{}$  $id[k] = e$ produces an element $x'$ such that $x'$ is equivalent to $x$ except for the value of the field $id$, whose value in $x'$ is the same as $x$, except that $x'.id[k] = v$ where $v$ is the result of evaluating $e$.
```
albertlockett: I wonder if we actually need a separate `extend` operator?

Currently the way we've implemented things is that `extend` is simply an alias for `set`. This means it's valid to write things like the following, which to me seems natural:
```
logs | set attribute["foo"] = "hello"
```

However it also means it is possible to write something like this:
```
logs | extend severity_text = "hello"
```
which is maybe a bit awkward, as it's not as if we're adding a new column. logs will always have a severity_text field as defined in the OTel data-model. However, I think we can maybe live with this awkwardness as it would feel noram for folks who are familiar with KQL

gbracha: I see several options here. We can leave things as is, if it helps KQL users and we think that justifies what is, ultimately, a wart. We can eliminate extend, which is the cleanest and simplest option (we could also eliminate set, but that seems to encourage the wrong terminlogy). Or, we could interpret extend to mean that we expect this to be a true extension, and it fail sif it attempts to overwrite and existing entry. This is a bit brittle, and I don't know if it agrees with the KQL extend.

rust/otap-dataflow/docs/opl/OPLSpec.tex
gbracha: ok.

jmacd (approved):
This document is a pleasure to read.

rust/otap-dataflow/docs/opl/OPLSpec.tex
albertlockett: > We can eliminate extend, which is the cleanest and simplest option

I think eliminating `extend` the best decision

lquerel:
@gbracha, I think the only remaining blocker is the CLA. I believe @albertlockett identified what was causing the issue. Could you please complete the CLA so we can merge the PR?

gbracha:
> @gbracha, I think the only remaining blocker is the CLA. I believe @albertlockett identified what was causing the issue. Could you please complete the CLA so we can merge the PR?

Completed (My 2nd attempt). Awaiting approval on F5's side.

PR #3059, opened 2026-05-20
Unified expression evaluation

### LINKED ISSUE #2664: OTAP query engine support `LogicalExpression` within `ScalarExpression`
In #2632 we added the ability to use any kind of scalar expression when evaluating logical expression for a filter. In this PR, we want to add support for bridging these expression evaluation paths in the other direction. 

Currently if I write an OPL/KQL program like this, it will not return an error:
```kql
logs | 
extend attribute["bad.thing.happen"] = severity_text == "ERROR" or severity_text == "WARN"
```
We should have support for this type of expression.

### LINKED ISSUE #3003: [query-engine]: unify expression evaluation implementations
The query engine, through no particular malice or incompetence, has arrived at a state where it has two systems that do more or less the same thing.

The filter path (filter.rs) evaluates predicates on OTAP data, which may contain expressions. The expression path (expr.rs) also evaluates expressions on OTAP data which may contain predicates. They do this using entirely different mechanisms, different data structures, and different philosophies about how one ought to relate rows in one RecordBatch to rows in another. The filter path works in bitmap space (for efficiency) whereas the expression path materializes joined RecordBatches (for flexibility). They were both, at the time of their creation, reasonable solutions.

The trouble is that each system may need to fall back on the other, creating a dependencies which neither system was originally designed to accommodate. 

## Why

Artistically it's disharmony that I cannot abide. There are also practical reasons to unify these code paths:

- The scope problem keeps getting re-solved.  Every module that touches this data develops its own private theory of how to navigate these relationships
-  Adding support for logical expressions in the expression path means either building a bridge between two systems that were never meant to be bridged, or duplicating logic that already exists elsewhere in a slightly different form
- The current system is built around the root-to-attributes relationship, but we have more parent-child relationships to be added: datapoint attributes, span event attributes, span links. Extending the current approach to handle these would mean extending four different scope-handling strategies in four different modules
- Cross-cutting concerns have nowhere to live

## The Proposal

- Replace both systems with one. A unified execution tree called ScopedOp that subsumes the filter path's bitmap-algebraic efficiency and the expression path's value-producing generality into a single structure. The key ideas, briefly:
- ScopedValue: a value paired with metadata about where it came from and how it relates to other batches. One type to replace PhysicalExprEvalResult.
- ScopedOp: a tree of operations — Eval, JoinAndEval, BitmapAnd, BitmapOr, BitmapNot. Each node can be asked to produce either values (execute_as_value) or bitmaps (execute_as_id_mask). The caller picks the mode that suits its needs. Bitmap chains stay in bitmap space without unnecessary materialization. Value chains produce actual arrays.
- align: a standalone function for re-mapping values from one scope to another via ID column relationships. One function, one place, no more scattered alignment logic.
- Generalized DataScope: replace the hardcoded root/attributes distinction with something that can handle arbitrary parent-child relationships in the OTAP data model.

## How We Get There (w/out breaking everything)

- Phase 1 — Introduce ScopedValue, ScopedOp, LeafEval, BatchPredicate, and align. Implement both execution modes. Write standalone tests. Touch nothing that already exists.
- Phase 2A — Build a unified planner that produces ScopedOp trees from the input AST, coexisting peacefully with the old planners.
- Phases 2B–2D — Migrate consumers (filter, assign, conditional) one at a time to use the new system.
- Phase 2E — Thread IdBitmapPool properly through bitmap nodes.
- Phase 2F — Remove the old types. A moment of silence for ScopedLogicalExpr, ScopedPhysicalExpr, FilterPlan, FilterExec, Composite, and their many friends.
- Phase 3 — Generalize DataScope for arbitrary parent-child relationships.
- Phase 4 — Final cleanup

albertlockett:
_Note to reviewers: it's a lot easier to read in split mode as everything has really been moved around to different files. I could try to break this up as well if it would make it easier to reivew._

# Change Summary

Replaces two systems for evaluating expressions (the filter path and the expression path) with a single system that is used in both places. Before this change, the filter.rs module had a system for evaluating expressions that produced bools and sometimes combined intermediate results using bitmaps of IDs, wheras the expr module did everything else (albeit, with fewer optimizations).

This made it hard to mix logical exprs (which the filter code handles) with the general expression code as the paradigms just didn't mesh very well.

I feel the most sensible thing is just to have one single module that evaluates all expressions, while trying to preservie the special optimizations for logical exprs. This change regorganizes the code to add this single expression evaluation path. As a happy side effect, we also now get seemless integration between logical exprs and all the other expression types "for free", which means we can do things like:

```kql
logs | extend attributes["is_hello"] = attributes["attr"] == "hello"

logs | where contains(concat(attributes["username"], attributes["email"]), "albert")
```

This PR also reduces the amount of code in expr eval by ~30% and also introduces a simpler type called `ScopedValue` for the result of an expression evaluation, which I think will make it a lot easier to support variables in the future.

**New Architecture:**
All expression and predicate evaluation now flows through `ScopedExpr`, a tree structure in pipeline/expr/. Each node supports two execution modes:
- `execute_as_value`:  ScopedValue (values + scope metadata + ID columns). Used for assignment, arithmetic, function arguments
- `execute_as_id_mask` → IdMask bitmap. Used for filtering, boolean combination. Allows chains of AND/OR/NOT to stay in ID bitmap space without materializing intermediate selection vectors (e.g. `BooleanArray`s)

Each `ScopedExpr` is one of the variants:
- `Eval(DfExpr)` Evaluate a DataFusion expression on a single scope's RecordBatch
- `Eval(BatchPredicate)` Batch-level predicate (e.g., signal type check for `is Log`)
- `JoinAndEval` Cross-scope: materialize children, join, evaluate
- `BitmapAnd/Or/Not` Boolean combination via IdMask algebra

The `ExprPlanner` builds these trees from the expression AST, applying optimizations at planning time.

**New Stuff**
| Location | Change |
|----------|--------|
| `pipeline/expr.rs` | Core types: `ScopedExpr`, `ScopedValue`, `LeafEval`, `BatchPredicate`, `DataScope` |
| `pipeline/expr/eval.rs` | `execute_as_value` for each `ScopedExpr` variant |
| `pipeline/expr/bitmap.rs` | `execute_as_id_mask` for each `ScopedExpr` variant |
| `pipeline/expr/planner.rs` | `ScopedOpPlanner` — builds `ScopedExpr` trees from `ScalarExpression` / `LogicalExpression` AST |
| `pipeline/filter.rs` | `FilterPipelineStage` uses `ScopedExpr`  + `ScopedValue` |
| `pipeline/assign.rs` | `AssignPipelineStage` uses `ScopedExpr` + `ScopedValue` |
| `pipeline/conditional.rs` | Branch predicates use `ScopedExpr` |
| `pipeline/planner.rs` | Routes filter, assign, and conditional to `ExprPlanner` |
| `pipeline/id_mask.rs` | `IdMask` extracted to a shared module (was inline in `filter.rs`) |

**What didn't change**
- Join mechanics (expr/join.rs) — parent_id-based cross-scope alignment
- AnyValue split-evaluate-stitch pattern
- Column-level projection
- All observable behaviour — all existing tests pass without modification

**Known Follow-ups**
- The old filter/compare.rs module (custom comparison kernel handling AnyValue struct comparisons, type mismatches, and null == null semantics) was removed as part of cleaning up the old filter path. The new path relies on DataFusion's standard comparison kernels. AnyValue-vs-AnyValue comparisons (e.g., attributes["x"] == attributes["y"]) may have subtle behavioral differences and deserve dedicated test coverage.
- Need to add an outer join for relalignment the coalesce operation (and add tests

* Closes #3003
* Closes #2664

Unit

No


albertlockett:
Attribute insertion benchmarks didnt change (which is to be expected, as this code basically does the same thing), but filtering benchmarks do improve slightly. The new code paths are a bit more aggressive about not materializing intermediate selection vectors.

bench | main | new code | percent change
-- | -- | -- | --
simple_field_filter/batch_size/32 | 6.6677 | 6.1898 | -7.17%
simple_field_filter/batch_size/1024 | 11.986 | 10.673 | -10.95%
simple_field_filter/batch_size/8192 | 47.937 | 43.522 | -9.21%
simple_attr_filter/batch_size/32 | 9.0604 | 8.382 | -7.49%
simple_attr_filter/batch_size/1024 | 20.232 | 18.255 | -9.77%
simple_attr_filter/batch_size/8192 | 100.57 | 89.652 | -10.86%
attr_or_attr_filter/batch_size/32 | 12.029 | 11.364 | -5.53%
attr_or_attr_filter/batch_size/1024 | 28.007 | 26.676 | -4.75%
attr_or_attr_filter/batch_size/8192 | 140.22 | 139.07 | -0.82%
attr_and_prop_filter/batch_size/32 | 10.045 | 9.4661 | -5.76%
attr_and_prop_filter/batch_size/1024 | 18.178 | 16.559 | -8.91%
attr_and_prop_filter/batch_size/8192 | 76.982 | 70.054 | -9.00%
attr_and_attr_filter/batch_size/32 | 4.181 | 4.1191 | -1.48%
attr_and_attr_filter/batch_size/1024 | 8.6431 | 8.488 | -1.79%
attr_and_attr_filter/batch_size/8192 | 41.45 | 40.603 | -2.04%
attr_and_or_together_filter/batch_size/32 | 15.77 | 14.288 | -9.40%
attr_and_or_together_filter/batch_size/1024 | 28.603 | 26.108 | -8.72%
attr_and_or_together_filter/batch_size/8192 | 124.33 | 111.63 | -10.21%

rust/otap-dataflow/crates/query-engine/src/pipeline/expr/bitmap.rs
```diff
@@ -0,0 +1,287 @@
...
+
+                    // create a bitmap of the IDs that pass the filter
+                    let bool_values = boolean_arr.values();
+                    let mut bitmap = pool.acquire();
+                    for (start, end) in
+                        BitSliceIterator::new(bool_values.inner().as_slice(), 0, boolean_arr.len())
+                    {
+                        for idx in start..end {
+                            bitmap.insert(id_col.value(idx) as u32);
+                        }
```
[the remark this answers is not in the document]
albertlockett: I made the change here so the expr evaluation will always combine the null bitmask into the selection vec before returning it, which means we don't need to worry about the nulls in this location.

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -1356,6 +1353,36 @@ fn decompose_any_value_upsert<'a>(
...
+/// Returns the `downcast_dicts` option from the inner `LeafEval::DatafusionExpr` projection
+/// options, if this is an `Eval(DatafusionExpr)` node. Returns `false` otherwise.
+fn requires_dict_downcast(expr: &ScopedExpr) -> bool {
+    match expr {
+        ScopedExpr::Eval {
+            eval: LeafEval::DatafusionExpr {
+                projection_opts, ..
+            },
+            ..
+        } => projection_opts.downcast_dicts,
```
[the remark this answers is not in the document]
albertlockett: In this case, the only place where this is used is in a location where the ScopeExpr variant could only be an eval (because it's working on a single attrs record batch). I changed the name of the method to make it apparant it only considers the leaf variants.

rust/otap-dataflow/crates/query-engine/src/pipeline/expr/planner.rs
```diff
@@ -0,0 +1,2630 @@
...
+    fn try_plan_fused_attr_contains(
+        &self,
+        haystack: &PlannedOp,
+        needle: &PlannedOp,
+    ) -> Result<Option<ScopedExpr>> {
+        // Haystack must be an attribute access, needle must be a static string literal
+        let (attrs_id, key) = match haystack.expr.eval_scope() {
+            Some(DataScope::Attribute(id, key)) => (*id, key.clone()),
+            _ => return Ok(None),
+        };
```
lquerel, line 1222: **[Finding from my automated review] As it looked correct to me, I’m copying the finding below.**

P1 correctness: `crates/query-engine/src/pipeline/expr/planner.rs:1218` and `crates/query-engine/src/pipeline/expr/planner.rs:1330` fuse `contains`/`matches` for any `DataScope::Attribute` haystack, but then evaluate against raw `col("str")` at lines 1243 and 1342. This drops transformations such as `lower(attributes["x"]) contains "foo"` or `lower(attributes["x"]) matches "foo"`, silently changing results. These paths should use the same `is_simple_attr_value_column(...)` guard already used by fused comparisons; otherwise they should fall back to normal expression planning.

rust/otap-dataflow/crates/query-engine/src/pipeline/expr/eval.rs
```diff
@@ -0,0 +1,768 @@
...
+
+/// Convert a `ScopedValue` into a `JoinInput` for use with the join module.
+///
+/// For root-scoped values, this looks up `scope_ids` and `resource_ids` from the root batch.
+/// For other scopes, these fields are left as `None`.
+pub(crate) fn scoped_value_to_join_input(
+    sv: ScopedValue,
+    otap_batch: &OtapArrowRecords,
+) -> Result<JoinInput> {
+    let is_root = sv.scope == DataScope::Root;
```
lquerel: [Issue from my automated review] Since it looked correct to me, I’m copying the finding below.

P1 runtime correctness: `crates/query-engine/src/pipeline/expr/eval.rs:467` only treats `DataScope::Root` as root when building `JoinInput`, while `crates/query-engine/src/pipeline/expr/join.rs:171` and `crates/query-engine/src/pipeline/expr/join.rs:277` only support `Root + Attribute`, not `RootParent(_) + Attribute`. Expressions combining resource/scope struct fields with attributes, e.g. `resource.schema_url == resource.attributes["schema_url"]`, can plan as `JoinAndEval` but then fail with invalid join scopes. `RootParent(_)` should be treated like root for join metadata and join strategy selection while preserving the appropriate result scope.

lquerel (approved):
Man, this is great stuff in terms of simplification and generalization!

I left you two pieces of feedback coming from my automated review. My manual review didn’t find anything else!

rust/otap-dataflow/crates/query-engine/src/pipeline/expr/eval.rs
albertlockett: good catch, fixed in 92c13cd63 !

rust/otap-dataflow/crates/query-engine/src/pipeline/expr/planner.rs
albertlockett: Good catch, fixed in [2b2859e](https://github.com/open-telemetry/otel-arrow/pull/3059/commits/2b2859e5dead23b240899cdf4ba3b6397b6e68d8) !

PR #3061, opened 2026-05-20
Add per-extension lifecycle metrics

### LINKED ISSUE #3039: Extension lifecycle follow-ups
Tracking the deferred follow-ups from @lquerel's review on #2860. The correctness fixes, macro-validation hardening, and regression tests have all been addressed in the PR itself; the items below are scoped to separate work because each is self-contained.
 
 ## 📊 Observability
 
 - [ ] **Per-extension lifecycle metrics.** A misbehaving active/background extension can monopolize a per-core runtime or ignore `Shutdown`, but the lifecycle path only emits `otel_warn!` events (`extension.shutdown.timeout`, `extension.shutdown.send_failed`, `extension.shutdown.send_timeout`). Runtime metrics show pipeline-level starvation and the wrapped control channel shows pressure, but neither clearly attributes the issue to a specific extension. Add per-extension metrics labeled by extension id — shutdown duration, shutdown timeout count, active task count, control-channel lag/depth — so operators can detect and isolate misbehaving extensions. ([review comment](https://github.com/open-telemetry/otel-arrow/pull/2860#discussion_r3262829745))

 ## 🏗️ Larger design follow-ups (separate PRs / design discussion)
 
 - [ ] **Extension readiness probes for init-complete guarantee.** Today, capability *construction* happens at build time but `start()` is async, so capability consumers can observe a pre-init state until deferred init (open connection, load config, warm cache) completes. Likely shape: opt-in readiness probe registered at build time (e.g. `builder.with_readiness_probe()` returns a handle the extension fires from `start()`); the engine awaits only the registered probes via `try_join_all` before spawning data-path tasks. Extensions that don't opt in keep today's behavior with zero overhead. ([review comment](https://github.com/open-telemetry/otel-arrow/pull/2860#discussion_r3262926578) — TODO already noted at `runtime_pipeline.rs:255`)
 
 - [ ] **(Discussion) Priority shutdown channel — engine-wide, not extension-only.** Suggested for extensions during review; concluded out of scope for #2860. No existing node type (receiver / processor / exporter) uses a dedicated shutdown channel — they all multiplex `NodeControlMsg::Shutdown` through the same mpsc as every other control message, and the data path even stashes future-deadline `Shutdown` in `pending_shutdown` to keep delivering other messages until the deadline. Introducing a priority channel just for extensions would be inconsistent with the rest of the engine. If priority shutdown is desirable, it warrants a cross-cutting design that applies to all node types.

gouslu:
# Change Summary

Adds per-extension lifecycle metrics for the OTAP engine, emitted from a new `ExtensionContext` and recorded by `ExtensionLifecycle`. The metric set lets operators tell apart extensions that shut down cleanly, timed out, failed to receive the `Shutdown` signal, exited early, errored, or panicked — and measure how long shutdown took for the ones that actually received it.

* Closes #3039

- Unit tests in `crates/engine/src/extension_lifecycle.rs` covering:
  - Bounded `drain_until_deadline` for a stuck extension
  - Per-extension completion recording (happy path)
  - `task.error` on `Err` return
  - `task.early_exit` when an extension exits before `Shutdown` is delivered (clean exit and error exit, with negative assertions that `shutdown.sent` / `shutdown.send_failed` stay at 0)
  - `shutdown.duration` skipped when `Shutdown` was never sent
  - `shutdown.timeout` skipped when `Shutdown` was never sent
  - No misleading `shutdown.duration` recorded for a timed-out variant
  - Panic attribution via `catch_unwind` → `task.error`
  - `active` gauge transitions on terminal paths (including timeout-with-abort and `Drop`)
  - Per-`(id, variant)` isolation in the registry (no cross-attribution)
  - `broadcast_shutdown` skips already-finalized entries (no spurious `shutdown.send_failed`)
- `cargo xtask check` passes (workspace build, clippy, tests, doctests).
- `cargo test -p otap-df-engine --lib extension_lifecycle` → 27/27 pass.

Yes — new internal telemetry metrics are now emitted per extension, keyed by `(extension.id, extension.variant)` (pipeline-scoped today):

| Metric | Unit | Kind | Meaning |
| --- | --- | --- | --- |
| `extension.lifecycle.active` | `{1}` | Gauge | `1` while the task is running, `0` on any terminal path (including `Drop` and timeout-with-abort). |
| `extension.lifecycle.shutdown.sent` | `{1}` | Gauge | `1` if the `Shutdown` control message was successfully sent, `0` otherwise. Canonical gate for interpreting `shutdown.duration`. |
| `extension.lifecycle.shutdown.duration` | `ns` | Gauge | Time from successful `Shutdown` send to task completion. Only set when `shutdown.sent == 1`. |
| `extension.lifecycle.shutdown.timeout` | `{1}` | Counter | Extension did not complete within the drain deadline. Gated on send so it doesn't double-count with send failures. Surviving tasks are aborted before this is recorded so `active=0` matches reality. |
| `extension.lifecycle.shutdown.send_failed` | `{1}` | Counter | `Shutdown` send failed — control channel closed, or the send exceeded `EXTENSION_SHUTDOWN_SEND_TIMEOUT` (500ms). Both failure modes are unified under this counter. |
| `extension.lifecycle.task.early_exit` | `{1}` | Counter | Extension terminated before `Shutdown` could be delivered (lifecycle-contract violation). The broadcast skips already-finalized extensions to avoid spurious `send_failed` attribution. |
| `extension.lifecycle.task.error` | `{1}` | Counter | Extension future returned `Err`. Panics are caught via `catch_unwind` and surfaced through this path. |

Implementation notes:
- New `ExtensionContext` (`crates/engine/src/extension/context.rs`) with `from_pipeline` / `from_pipeline_parts` constructors. A private `AttributeSet` enum keeps the scope (pipeline today, engine/group later) opaque to consumers. `ExtensionAttributeSet` lives in `crates/engine/src/attributes.rs` and is keyed by `(extension.id, extension.variant)` so different variants of the same id stay isolated in the registry.
- Lifecycle metrics are accumulated directly into the registry via `accumulate_metric_set_snapshot` (no intermediate channel), and the `active` gauge is canonicalized each cycle to survive the framework's `clear_values`.
- `shutdown.duration` (in nanoseconds) is anchored to a per-send `Instant::now()` taken **after** each successful `Shutdown` send (not broadcast start), so it isn't inflated by send-queueing time.
- `broadcast_shutdown` filters out already-finalized entries via `is_finalized()` so an extension that crashed/exited early is recorded as `task.early_exit` rather than `shutdown.send_failed`.
- Extension task body is wrapped in `AssertUnwindSafe(...).catch_unwind()` so panics are attributed to the correct extension id rather than fanning out to all pending extensions.
- On `JoinError` (panic or abort observed by the runtime), the per-extension entry is finalized to avoid later timeout misattribution.
- When the drain deadline fires, surviving extension `JoinHandle`s are aborted before `shutdown.timeout` is recorded so the published `active=0` reflects the runtime's view (task will not be polled again).
- `Drop` aborts surviving futures, flips `active=0` / `finalized=true` for non-finalized entries, and unregisters metric sets and entities in order.


rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -159,37 +391,67 @@ impl ExtensionLifecycle {
...
+                    // `shutdown.delivered` to 1 so observers can gate
+                    // duration interpretation on a single boolean.
+                    // If the entry was already finalized (the
+                    // extension completed before `broadcast_shutdown`
+                    // was even called and `record_completion` ran via
+                    // the main select loop), leave it alone — that
+                    // earlier finalize is authoritative and the
+                    // extension never participated in cooperative
+                    // shutdown.
+                    let delivered_at = Instant::now();
```
lalitb: Could we make the `shutdown.duration` timestamp fully per-extension? Right now it looks like the timestamp is recorded only after all shutdown sends finish, so a slow or timed-out send for one extension can affect the measured duration for another extension. Capturing the timestamp inside each successful send would keep each extension’s duration independent.

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -105,34 +252,97 @@ impl ExtensionLifecycle {
...
+            // this code. Then attach the scope-agnostic lifecycle
+            // metric set.
+            let entity_key = extension_context.register_extension_entity(&ext_id);
+            registered_entities.push(entity_key);
+            let mut lifecycle_metrics =
+                registry.register_metric_set_for_entity::<ExtensionLifecycleMetrics>(entity_key);
+            // Mark the extension active immediately so a snapshot taken
+            // before the task is scheduled still observes it.
+            lifecycle_metrics.active.set(1);
+            let _ = metrics_reporter.clone().report(&mut lifecycle_metrics);
```
lalitb: I think `active` can be reported incorrectly after the first flush. `MetricsReporter::report()` clears the local `MetricSet` after sending the snapshot, so after the initial `active=1` report this stored metric set is back to zero. Later, send-failure/send timeout reports can include `active=0` even though the extension task is still running. A clean pre-shutdown completion can also fail to publish `active=0` because the local gauge is already zero and `needs_flush()` may skip the report.

Could we keep lifecycle state separately, or re-set the current gauge values before every lifecycle report?

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -105,34 +252,97 @@ impl ExtensionLifecycle {
             }
             let ext_metrics_reporter = metrics_reporter.clone();
             let ext_id = ext_wrapper.name();
+
+            // Register a per-extension entity through the
+            // `ExtensionContext` so the parent attribute hierarchy
+            // (pipeline today, possibly engine later) stays opaque to
+            // this code. Then attach the scope-agnostic lifecycle
+            // metric set.
+            let entity_key = extension_context.register_extension_entity(&ext_id);
```
lalitb: This map is keyed only by `ExtensionId`, but a bundle can have both local and shared active variants with the same extension id. In that case the second `ext_metrics.insert(ext_id.clone(), ...)` overwrites the first variant's entry, while both tasks and metric sets were registered. Completion/timeout attribution then collapses both variants into one entry, and the overwritten metric set may not be unregistered.

Should the lifecycle key include the execution variant as well, for example `(extension_id, local/shared)`?

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -245,13 +522,256 @@ impl ExtensionLifecycle {
...
+                let _ = reporter.report(&mut entry.metrics);
+            }
+        }
+        // Unregister per-extension metric sets and entities so a
+        // dropped lifecycle doesn't leak registry slots, mirroring
+        // `PipelineMetricsMonitor::Drop`.
+        for entry in self.ext_metrics.values() {
+            let _ = self
+                .registry
+                .unregister_metric_set(entry.metrics.metric_set_key());
```
lalitb: One concern with unregistering here: `report()` only enqueues a snapshot to the metrics channel. If `Drop` unregisters the metric set before the collector consumes that snapshot, `accumulate_metric_set_snapshot()` will not find the key and the final lifecycle update can be dropped. That matters for terminal values like `active=0`, `shutdown.duration`, `shutdown.timeout`, and `task.error`.

Do we need to defer unregistering these extension metric sets until pending snapshots are collected, or otherwise flush/drain before removing the registry entries?

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -4,135 +4,235 @@
...
             }
             let ext_metrics_reporter = metrics_reporter.clone();
             let ext_id = ext_wrapper.name();
+            let key: ExtensionKey = (ext_id.clone(), variant);
+
+            // Register a per-lifecycle entity keyed by `(id, variant)`
+            // so local and shared variants are tracked as distinct
+            // entities everywhere — registry, scrapers, and the
+            // per-lifecycle metric set below.
+            let entity_key = extension_context.register_extension_entity(&ext_id, variant);
```
lalitb, line 160: I think this makes the active gauge unreliable for long running extensions. We publish `active=1` once during spawn, but registry metric values are reset after export. If nothing republishes `active=1` while the extension keeps running, the extension can look inactive or disappear after the first scrape. Can we tie this into the normal reporting path or reassert the gauge on each collection?

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -143,94 +243,145 @@ impl ExtensionLifecycle {
...
+                        Ok(Err(e)) => BroadcastSendResult::SendFailed(format!("{e}")),
+                        Err(_elapsed) => BroadcastSendResult::SendFailed(format!(
+                            "send did not complete within {}ms",
+                            EXTENSION_SHUTDOWN_SEND_TIMEOUT.as_millis(),
+                        )),
+                    };
+                    (key, result)
+                })
+            });
+        let results: Vec<_> = futures::future::join_all(sends).await;
```
lalitb, line 321: I believe this duration can be inflated by other extensions. We record each extension's send time, but then wait for all shutdown sends to finish before polling completions again. If one extension exits quickly while another send blocks for 500ms, the fast extension's shutdown.duration includes that unrelated wait. Can we capture the actual task completion time or avoid blocking completion observation behind the full broadcast join?

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -4,135 +4,235 @@
...
+            let _ = pending.insert(key.clone());
+
+            let fut_ext_id = ext_id.clone();
             let fut = async move {
-                match ext_wrapper.start(ext_metrics_reporter.clone()).await {
-                    Ok(terminal_state) => {
+                // `catch_unwind` so a panic in user code is attributed
+                // to *this* extension instead of surfacing as an
+                // opaque `JoinError(panic)`.
+                let outcome = AssertUnwindSafe(ext_wrapper.start(ext_metrics_reporter.clone()))
```
lalitb, line 176: Can you check this once, as it seems catching the panic here changes the runtime failure classification. Previously an extension panic would surface as a JoinError with `is_panic=true`; now it becomes a normal `InternalError` and is counted like `task.error`. That hides the difference between an extension returning `Err` and the task actually panicking. Can we keep the per-extension attribution without losing the panic classification?

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
lalitb, line 160: Separate from the active gauge behavior, this path also looks unconditional. If `runtime_metrics` is set to none, should we still register extension lifecycle entities and publish `extension.lifecycle` metrics? Other runtime metrics seem to respect `telemetry_policy.runtime_metrics`, so I think this should be gated the same way.

gouslu:
thanks for the feedback. besides the bugs in it, I have other things I don't like about this design and will close this PR and work on a different approach.

PR #3064, opened 2026-05-21
[ETW Receiver] Per-session-name isolation and event drop observability

### LINKED ISSUE #3065: [ETW Receiver] Singleton should track its config and reject mismatched subscribers
### Pre-filing checklist

### Component(s)

## Singleton should track its config and reject mismatched subscribers

**Review comment:** [discussion_r3268320828](https://github.com/open-telemetry/otel-arrow/pull/2930#discussion_r3268320828), [discussion_r3268428162](https://github.com/open-telemetry/otel-arrow/pull/2930#discussion_r3268428162)

### Summary

The process-global `SESSION` singleton is initialized with the first
`receiver:etw` node's config, but a second node with a different
`session_name` or provider list silently reuses (or fails on) the first
session. There is no guard that detects or rejects mismatched configurations.

Lalit's specific concern: "What I meant here is only the case where the
singleton is already running and another ETW receiver comes in with a different
session name or provider list. Today it seems like that second receiver would
just reuse the first session. I was wondering if we should detect that and
return a clear error, instead of making it look like the second config was
applied."

### Current implementation

`SESSION` is a `Mutex<Option<Vec<Receiver>>>` with no record of which config
created it. When `subscribe` is called:

- If `guard.is_none()`: the session is initialized with **this** config.
- If `guard.is_some()`: the function skips initialization entirely and
  just pops a receiver from the pool — regardless of whether the caller's
  config matches the one that created the pool.

A second node therefore either (a) silently shares the first node's
session or (b) fails with a misleading "pool exhausted" `InternalError`.

```rust
// session.rs L336–L360
static SESSION: Mutex<Option<Vec<mpsc::Receiver<EtwEventData>>>> = Mutex::new(None);

pub(super) fn subscribe(
    config: &Config,
    num_cores: usize,
) -> Result<mpsc::Receiver<EtwEventData>, Error> {
    let mut guard = SESSION.lock().map_err(|e| Error::InternalError {
        message: format!("ETW session lock poisoned: {e}"),
    })?;

    if guard.is_none() {
        // First call — initialize the session.
        let (txs, rxs): (Vec<_>, Vec<_>) = (0..num_cores)
            .map(|_| mpsc::channel(EVENT_CHANNEL_CAPACITY))
            .unzip();
        spawn_etw_session(config, txs)?;
        *guard = Some(rxs);
    }

    let pool = guard.as_mut().expect("session initialized above");
    pool.pop().ok_or_else(|| Error::InternalError {
        message: "ETW consumer pool exhausted; engine requested more receivers than num_cores"
            .to_string(),
    })
}
```

### Proposed approach

Replace the singleton with a `HashMap<String, SessionEntry>` keyed by
`session_name`. Each entry owns its own channel pool and is created once.
A second node with the **same** `session_name` gets a clear
`InvalidUserConfig` error ("session_name '…' is already in use"). A second
node with a **different** `session_name` initializes an independent kernel
session.

```rust
struct SessionEntry {
    pool: Vec<mpsc::Receiver<EtwEventData>>,
}

static SESSIONS: Mutex<HashMap<String, SessionEntry>> = Mutex::new(HashMap::new());

pub(super) fn subscribe(config: &Config, num_cores: usize)
    -> Result<mpsc::Receiver<EtwEventData>, Error>
{
    let mut guard = SESSIONS.lock().map_err(|e| Error::InternalError {
        message: format!("ETW sessions lock poisoned: {e}"),
    })?;

    let entry = match guard.entry(config.session_name.clone()) {
        Entry::Vacant(v) => {
            let (txs, rxs): (Vec<_>, Vec<_>) = (0..num_cores)
                .map(|_| mpsc::channel(EVENT_CHANNEL_CAPACITY))
                .unzip();
            spawn_etw_session(config, txs)?;
            v.insert(SessionEntry { pool: rxs })
        }
        Entry::Occupied(o) => o.into_mut(),
    };

    entry.pool.pop().ok_or_else(|| {
        Error::ConfigError(Box::new(
            otap_df_config::error::Error::InvalidUserConfig {
                error: format!(
                    "ETW session_name '{}' is already in use; \
                     each receiver:etw node must specify a distinct session_name",
                    config.session_name,
                ),
            },
        ))
    })
}
```

### Tradeoffs

- Does **not** distinguish "same `session_name` + identical providers"
  from "same `session_name` + different providers" — both are rejected.
  Allowing the former would require sizing the pool for the union of all
  node replicas, which the factory contract does not support cleanly.
- A controller bug that calls `subscribe` more than `num_cores` times for
  a single node would be misreported as a session-name collision.
  Acceptable for the scaffold.
- Utpilla noted that the restart/hot-reload lifecycle is being tracked
  separately in [#3033](https://github.com/open-telemetry/otel-arrow/issues/3033),
  which covers some of the broader session lifecycle concerns.

### Additional Context

_No response_

swashtek:
# Change Summary

ETW multi session for different names, logs error on same name session. 
Event drop observability

* Closes #3065

cargo run --features etw-receiver -- -c .\configs\etw-multi-session.yaml 

No


rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -42,20 +42,24 @@
...
 use tokio::sync::mpsc;
 
 use super::{Config, ProviderConfig, TraceLevel};
 
 /// Channel capacity for ETW events sent from the blocking session thread to
 /// each per-core async receiver loop.  A bounded channel provides implicit
 /// backpressure: when a core's channel is full the round-robin callback skips
-/// that core for the current event (the event is dropped for that core only).
+/// that core for the current event (the event is dropped entirely from the
+/// pipeline).
```
utpilla: Let's update the doc comment here to make the expectation clear:

```rust
/// backpressure: when the target core's channel is full the event is dropped
/// from the pipeline entirely.  The round-robin index still advances, so the
/// next event continues to the following core (no retry on another core).
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -287,9 +305,21 @@ fn spawn_etw_session(config: &Config, txs: Vec<mpsc::Sender<EtwEventData>>) -> R
...
-                    // drop the event for that core only.
-                    let _ = txs[i % txs.len()].try_send(data);
+                    // Best-effort send; if this core's channel is full the
+                    // event is dropped from the pipeline entirely (each event
+                    // is assigned to exactly one core via round-robin).
+                    if txs[i % txs.len()].try_send(data).is_err() {
+                        let count = dropped.get() + 1;
+                        dropped.set(count);
+                        // Rate-limited log: first drop, then every 10,000th.
+                        if count == 1 || count.is_multiple_of(10_000) {
```
utpilla: This dropped count should be reported as a metric instead of log. It's not as straightforward. We can just use `MetricSet` as it is as this code is being run in a background thread. Could you add a TODO here to consider using metrics here later?

swashtek: Added.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -314,55 +375,188 @@ fn spawn_etw_session(config: &Config, txs: Vec<mpsc::Sender<EtwEventData>>) -> R
...
-    }
+            spawn_etw_session(config, txs)?;
 
-    let pool = guard.as_mut().expect("session initialized above");
-    pool.pop().ok_or_else(|| Error::InternalError {
-        message: "ETW consumer pool exhausted; engine requested more receivers than num_cores"
-            .to_string(),
+            v.insert(SessionEntry { pool: rxs })
+        }
+        Entry::Occupied(o) => o.into_mut(),
```
lalitb: This reuse may be too broad for a misconfigured pipeline. For example:

```text
receiver A -> session_name: "OtelArrowETW" -> provider: KernelProcess
receiver B -> session_name: "OtelArrowETW" -> provider: PowerShell
```
If B is accidentally configured with the same `session_name`, it can get a channel from A’s existing session until the pool is exhausted, so it may receive `KernelProcess` events instead of failing clearly. Can we check that the existing session belongs to the same receiver/config before reusing it?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -299,7 +351,16 @@ fn spawn_etw_session(config: &Config, txs: Vec<mpsc::Sender<EtwEventData>>) -> R
 
             // `parse_until` blocks on `ProcessTrace`.  We never signal stop,
             // so the session runs until the process exits.
-            let _result = session.parse_until(&session_name, || false);
+            // TODO: Surface startup failures via a oneshot readiness channel
+            // once the one-collect API stabilizes (TDH decoding work).
+            let result = session.parse_until(&session_name, || false);
+            if let Err(ref e) = result {
+                otel_error!(
+                    "etw.parse_until.failed",
```
lalitb, line 366: This logs the failure, but `subscribe()` has already returned success once the thread was spawned. If `parse_until` fails immediately, the pipeline can still look like it started successfully and then just receive no ETW events. Can we send a startup success/failure signal back to `subscribe()` before accepting the receiver as ready?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
utpilla: This is a valid issue. A proper fix for this has to reject duplicate `session_name`s before any factory call runs, i.e. at pipeline-validation time. Today's `validate_config` hook is per-node only and has no view of other nodes' configs. Closing the gap requires a framework-level change, and should be addressed separately.

This PR is still helpful in that it now supports multiple ETW nodes as long as they have different session names. It doesn't invalidate nodes with same `session_name`s and the doc comments etc. should be updated to reflect the same.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
swashtek: Currently there's a gap between `subscribe()` returning success and `parse_until` actually starting the ETW kernel session. If `parse_until` fails immediately (e.g., `ERROR_ACCESS_DENIED` because the process isn't running as Administrator, or `ERROR_ALREADY_EXISTS` because another process owns that session name), the pipeline looks healthy but receives zero events, it is a silent failure.

The existing code already has the TODO for this, it needs coordination with `one-collect` API stabilization:
```rust
// TODO: Surface startup failures via a oneshot readiness channel
// once the one-collect API stabilizes (TDH decoding work).
```

### The proposed implementation approach

Since `parse_until` **blocks on success** (it calls `ProcessTrace` which doesn't return until the session stops) but **returns immediately on failure**, a timeout-based readiness pattern works:

```
spawn_etw_session():
  1. Create std::sync::mpsc::channel for startup signal
  2. Spawn OS thread:
     a. Set up EtwSession, enable providers, register callbacks
     b. Call parse_until() — blocks on success, returns Err on failure
     c. If parse_until returns: send result via startup_tx
  3. Back in spawn_etw_session, wait on startup_rx with timeout (e.g. 500ms):
     - Timeout expired → parse_until is still blocking → session started OK ✓
     - Received Err(e) → parse_until failed immediately → propagate error ✗
     - Disconnected → thread panicked → propagate error ✗
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
swashtek: Added config-mismatch detection on session reuse, this prevents accidental reuse of same `session_name` and silently receives wrong events.

utpilla: This kind of cross-node validation is a more generic problem that should be solved at an engine level. I have created a tracking issue for it: #3072

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
lalitb: Makes sense. I’m okay with this being follow-up work if this PR is intentionally iterative. The current change logs `parse_until()` failure, but does not yet propagate it back to `subscribe()`, so the startup readiness gap still remains and should stay tracked clearly.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
lalitb: Thanks, that makes sense. The config-mismatch detection on session reuse fix should be fine, and cross-node validation can be covered separately.

lalitb (approved):
LGTM for the PRs scope.  Should be merged once @utpilla approves this.

One small cleanup: some comments in `mod.rs` still describe the ETW session as a process-global singleton; they should probably be updated to say per-session-name registry.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
utpilla: There is still some references to process-wide singleton design which is being removed in the PR.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -234,12 +255,14 @@ fn spawn_etw_session(config: &Config, txs: Vec<mpsc::Sender<EtwEventData>>) -> R
...
-            // is safe — no atomics or locking needed.  Sharing the counter
-            // ensures uniform core distribution even at startup when
+            // Shared round-robin counter, drop counter, and sender list.
+            // All provider callbacks run on the single `ProcessTrace` thread,
+            // so `Cell` is safe — no atomics or locking needed.  Sharing the
+            // counter ensures uniform core distribution even at startup when
             // multiple providers would otherwise all start at index 0.
             let next: Rc<Cell<usize>> = Rc::new(Cell::new(0));
+            let dropped: Rc<Cell<u64>> = Rc::new(Cell::new(0));
+            let closed_logged: Rc<Cell<bool>> = Rc::new(Cell::new(false));
```
utpilla: `closed_logged` is a single global flag for the whole session, so the info log fires once and is permanently suppressed for any subsequent closure. That suppresses spam but you can't distinguish "core 0 panicked early" from "all cores closing during normal shutdown". Any later closure on a different core is silent.

Could we make it per-core? Something like:

```rust
let closed_logged: Rc<Vec<Cell<bool>>> =
    Rc::new((0..txs.len()).map(|_| Cell::new(false)).collect());
```

and in the callback:

```rust
Err(mpsc::error::TrySendError::Closed(_)) => {
    let core = i % txs.len();
    if !closed_logged[core].get() {
        closed_logged[core].set(true);
        otel_info!("etw.event.channel_closed", core = core);
    }
}
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
utpilla: Ah just noticed @lalitb commented about it already in https://github.com/open-telemetry/otel-arrow/pull/3064#pullrequestreview-4367399154.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
swashtek: Fixed, thanks.

utpilla (approved):

jmacd (approved):

PR #3075, opened 2026-05-23
fix(otap): guard OTAP views against encoded transport IDs

### LINKED ISSUE #2824: [OtapLogsView] centralize transport-optimized ID guard
### Pre-filing checklist

### Component(s)

### Objective

### Description:

Follow-up from #2810.   

`OtapLogsView` assumes OTAP Arrow ID columns are already decoded/plain before it builds grouping and attribute indexes.                                      
                                                                                                                                                                 
If transport-optimized OTAP records from an OTAP gRPC receiver reach `OtapLogsView`, then `group_by_resource_id`, `group_by_scope_id`, and `build_attribute_index` may read delta/quasi-delta encoded columns: 
    - `resource.id`
    - `scope.id`
    - attribute `parent_id`                                                                                                                                       That can produce silently wrong resource/scope grouping and attribute lookup.                                                                                

 #2810 fixed the Geneva exporter narrowly by calling `decode_transport_optimized_ids()` before constructing `OtapLogsView`. This issue tracks the broader centralized guard so future callers cannot accidentally build a view over still-encoded IDs.                                                

#### Open design question:

  - Add the guard in `OtapLogsView::new` for strongest enforcement. `new` is already fallible today (`Result<Self, Error>`), so this would add a validation failure mode without changing its signature.                                                                                                                 
  - Or add the guard only in `TryFrom<&OtapArrowRecords>`, which is narrower but still leaves direct `OtapLogsView::new(...)` callers vulnerable.
    
#### Likely callers to audit/patch if the guard is centralized:
                                                                                                                                            
- `azure_monitor_exporter`                                                                                                                                   
- `content_router`                                                                                                                                           
- `console_exporter`                                                                                                                                         
- `resource_validator`                                                                                                                                       
 - any future `OtapLogsView` caller

#### Suggested error:                                                                                                                                                                                                                                                                                                             

```rust                                                                                                                                                      
Error::TransportOptimizedIdsNotDecoded {                                                                                                                     
     payload_type: ArrowPayloadType,                                                                                                                          
     column: String,                                                                                                                                          
 }
```                                                                                                                                                            
#### Suggested behavior: 
                                                                                                                                                                                                                                                                                                          
- Detect encoded resource.id, scope.id, and attribute parent_id using existing column encoding metadata.                                                     
- Return a typed error instead of constructing a view over still-encoded IDs.                                                                                
- Ensure callers that can receive OTAP gRPC records call decode_transport_optimized_ids() before constructing `OtapLogsView`.                                  
                                                                                                                                                   

### Rationale

-

### Scope

-

### Acceptance Criteria

-

### Dependencies or Blockers

-

### Additional Context

-

yeomin4242:
# Change Summary

### Problem

`OtapLogsView`, `OtapMetricsView`, and `OtapTracesView` build grouping and child
indexes from OTAP Arrow ID columns. If transport-optimized records reach these
views before IDs are decoded, the views can read encoded `id`, `resource.id`,
`scope.id`, or `parent_id` columns as plain IDs.

### Solution

Add fail-fast guards before each OTAP view builds indexes, and decode
transport-optimized OTAP Arrow records in callers that can receive them before
constructing the view.

Changes:
- Add `Error::TransportOptimizedIdsNotDecoded`.
- Add shared helpers for checking plain-encoded ID columns.
- Guard logs, metrics, and traces view construction before grouping/indexing.
- Decode OTAP Arrow logs/metrics in affected node callers before constructing views.
- Add transport guard tests for logs, metrics, and traces.

This covers Metrics and Traces in the same PR because they share the same failure
mode as Logs, and the issue discussion called out fixing the remaining views as
acceptable.

* Closes #2824

- `cargo fmt --all`
- `cargo test -p otap-df-pdata views::otap --lib`
- `cargo test -p otap-df-pdata test_metrics_all_fields_otap_view --lib`
- `cargo test -p otap-df-pdata test_traces_all_fields_otap_view --lib`
- `cargo check -p otap-df-pdata`
- `cargo check -p otap-df-core-nodes --features otap-df-otap/crypto-ring`
- `cargo xtask check`

No external API or configuration changes. OTAP views now return a typed error
instead of constructing over still-encoded transport ID columns.


rust/otap-dataflow/crates/core-nodes/src/exporters/console_exporter/mod.rs
```diff
@@ -139,14 +139,20 @@ impl ConsoleExporter {
...
                 }
             },
-            OtapPayload::OtapArrowRecords(records) => match OtapLogsView::try_from(records) {
-                Ok(logs_view) => {
-                    self.formatter.print_logs_data(&logs_view).await;
-                }
-                Err(e) => {
-                    otel_error!("console.logs_view.otap_create_failed", error = ?e, message = "Failed to create OTAP logs view");
+            OtapPayload::OtapArrowRecords(records) => {
+                let mut records = records.clone();
```
jmacd: Is the clone necessary?

lalitb: This pattern appears across several call sites; we could centralize or decode an owned payload instead of making each view consumer clone and materialize IDs?

yeomin4242: @jmacd I think Yes, the clone is needed here because decoding mutates the records while the original `pdata` still needs to remain available.

@lalitb Agreed, the repeated clone+decode+view pattern is noisy; I’ll centralize it in a small helper.

lalitb:
@yeomin4242  - I’m still concerned about the overall decode-before-view strategy here. The guards are the right correctness boundary, but the caller-side fix now makes borrowed view consumers clone and decode the whole `OtapArrowRecords` before building the view. Some of these paths, like routing/validation, only need resource-level data, so decoding the full logs/metrics payload can add avoidable latency and memory pressure on hot paths. Can we move this into an owned/in-place decode path or provide narrower decoded/resource-only view helpers so callers do not have to materialize more of the payload than they actually inspect?

yeomin4242:
@lalitb 

Thanks. I updated the hot resource-only logs paths to avoid `DecodedOtapArrowRecords` and use a narrower `DecodedOtapLogsResources` helper instead. It only clones/decodes `Logs.resource.id` and `ResourceAttrs.parent_id`, then builds an `OtapLogsResourcesView`, so routing/validation no longer decode `ScopeAttrs`, `LogAttrs`, or the rest of the logs payload just to inspect resources.

I kept the full-view guards as the correctness boundary for `OtapLogsView`, and added a test with encoded `ScopeAttrs` + `LogAttrs` to verify the resource-only view only requires the resource-level columns.

rust/otap-dataflow/crates/pdata/src/views/otap/logs/resources/view.rs
```diff
@@ -0,0 +1,137 @@
...
+        ensure_plain_encoded_columns(ArrowPayloadType::Logs, logs_batch, &[RESOURCE_ID_COL_PATH])?;
+        if let Some(batch) = resource_attrs {
+            ensure_plain_encoded_columns(
+                ArrowPayloadType::ResourceAttrs,
+                batch,
+                &[consts::PARENT_ID],
+            )?;
+        }
+
+        let resource_groups = group_by_resource_id(logs_batch);
```
lalitb: Since this resource-only view only uses the resource id later and ignores the `RowGroup`, do we need to build full row groups here? This can still allocate based on the number of log records even though routing/validation only need the distinct resources.

rust/otap-dataflow/crates/pdata/src/views/otap/logs/resources/decoded.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// the transport-optimized IDs needed to build a resource-only view.
+    pub fn clone_and_decode(records: &OtapArrowRecords) -> Result<Self, Error> {
+        let logs_batch = records
+            .get(ArrowPayloadType::Logs)
+            .ok_or(Error::RecordBatchNotFound {
+                payload_type: ArrowPayloadType::Logs,
+            })?;
+        let logs_batch = remove_transport_optimized_resource_id(logs_batch)?;
+
+        let resource_attrs = records
```
lalitb, line 62: This is better than decoding the full logs payload, but it still decodes all resource attributes before routing/validation looks for one configured key. Could this be made more lazy or key-focused so large resource attribute batches do not all get materialized up front?

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
```diff
@@ -610,19 +610,35 @@ impl TemporalReaggregationProcessor {
...
-                Ok(view) => self.process_view(effect_handler, &view).await,
-                Err(e) => {
-                    otel_warn!(telemetry::VIEW_CREATION_FAILED_EVENT, error = %e);
-                    self.metrics.batches_rejected.inc();
-                    let msg = format!("Failed to create view: {:#}", e);
-                    effect_handler
-                        .notify_nack(NackMsg::new_permanent(msg, pdata))
-                        .await?;
+            OtapPayload::OtapArrowRecords(records) => {
+                let decoded_records = match DecodedOtapArrowRecords::clone_and_decode(records) {
```
lalitb, line 637: This decodes and builds the full metrics view before `process_view()` checks whether the batch has any aggregatable metrics. If the batch has none, we later pass the original payload through unchanged, so this decode work is wasted. Can we do the aggregatable check before the full decode/view setup?

rust/otap-dataflow/crates/pdata/src/views/otap.rs
```diff
@@ -12,7 +12,43 @@ pub mod common;
...
+/// Prefer [`DecodedOtapLogsResources`] when a logs caller only needs
+/// resource-level data.
+pub struct DecodedOtapArrowRecords {
+    records: OtapArrowRecords,
+}
+
+impl DecodedOtapArrowRecords {
+    /// Clone borrowed OTAP Arrow records and decode transport-optimized IDs.
+    pub fn clone_and_decode(records: &OtapArrowRecords) -> Result<Self> {
+        let mut records = records.clone();
```
lalitb, line 56: Do we have a cheap way to skip this clone when the records are already plain? Some in process paths may already have decoded IDs, so this helper may still clone the whole record set before finding that out.

rust/otap-dataflow/crates/pdata/src/views/otap/logs/resources/view.rs
yeomin4242: The resource-only logs view now collects only distinct resource IDs instead of building `RowGroup`s, since routing/validation only needs resource-level iteration.

rust/otap-dataflow/crates/pdata/src/views/otap/logs/resources/decoded.rs
yeomin4242: The router/validator paths now use a keyed resource decode helper, so `ResourceAttrs` is filtered to the configured key before parent IDs are decoded. I also added coverage for both matching-key and missing-key cases, including the empty filtered batch path.

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
yeomin4242: I added a root metrics preflight before full OTAP decode/view construction. It checks only `metric_type`, `aggregation_temporality`, and `is_monotonic`; if no aggregatable metrics are present, the processor forwards the original pdata unchanged and skips child ID decode/full view construction on that passthrough path.

rust/otap-dataflow/crates/pdata/src/views/otap.rs
yeomin4242: `DecodedOtapArrowRecords` now uses borrowed-or-owned storage: already-plain records are borrowed directly, while transport-optimized records still take the existing clone+decode path.

yeomin4242:
@lalitb 

Thanks for the review. I addressed the four efficiency points in separate commits:

- `862eebdb`: avoids building `RowGroup`s in the resource-only logs view and collects only distinct resource IDs.
- `17057059`: makes resource routing/validation use keyed resource-attr decoding, filtering `ResourceAttrs` to the configured key before parent ID decode. Added tests for both matching-key and missing-key/empty-filtered-batch cases.
- `87766714`: adds a root metrics preflight before full OTAP decode/view construction. If no aggregatable metrics are present, the original pdata is forwarded unchanged and child ID decode/full view construction is skipped on that passthrough path.
- `16fbb86a`: skips cloning in `DecodedOtapArrowRecords` when transport IDs are already plain by borrowing the records directly, while keeping clone+decode for transport-optimized records.

ThomsonTan:
This is a user-facing fix that also adds public API (`DecodedOtapArrowRecords`, `DecodedOtapLogsResources`, `otap_metrics_have_aggregatable_metrics`, a new `Error` variant) and changes view behavior (typed error instead of building over still-encoded IDs), but I don't see a `.chloggen/` entry. Probably add one?

yeomin4242:
> This is a user-facing fix that also adds public API (`DecodedOtapArrowRecords`, `DecodedOtapLogsResources`, `otap_metrics_have_aggregatable_metrics`, a new `Error` variant) and changes view behavior (typed error instead of building over still-encoded IDs), but I don't see a `.chloggen/` entry. Probably add one?

This repo seems to use the top-level CHANGELOG.md Unreleased section rather than .chloggen, so I’ll add an entry there.

PR #3083, opened 2026-05-25
[Attributes Processor] Add attribute `hash` action.

### LINKED ISSUE #3055: [Attributes processor] Add hash action for stable attribute pseudonymization
### Pre-filing checklist

### Component(s)

The attributes processor currently supports actions such as `rename`, `delete`, `insert`, and `upsert`. It would also be useful to support hashing existing attribute values for cases where operators need stable pseudonymization instead of deletion or constant masking.

### Proposed Solution

 This issue proposes adding a `hash` action:

```yaml
actions:
 - action: hash
   key: user.email
   algorithm: sha256
   salt: "${HASH_SALT}"
```

Expected behavior:

  - If the target attribute exists, replace its value with a deterministic hash.
  - If the target attribute does not exist, do not add a new attribute.
  - Support at least `sha256` initially.
  - Validate unsupported algorithms at startup.
  - Avoid logging raw attribute values or salts.
  - Document how salts should be supplied securely.

### Alternatives Considered

  - Use `delete`
    This removes the raw value, but also removes the attribute entirely, so operators lose the ability to correlate records by that field.

  - Use `mask`
    A constant mask value is useful for redaction, but it does not preserve stable correlation. All values become the same placeholder.

  - Use `upsert`
    `upsert` can replace a value with a configured constant, but it cannot derive a per-record hash from the original value. It can also create the attribute when it was absent, which is not the desired update-only behavior.

 - Use the transform processor
    The transform processor may be able to express some hashing use cases, but a dedicated hash action keeps common pseudonymization workflows simple and explicit in the attributes processor.

  - Do hashing before telemetry reaches OTAP dataflow
    This works for some deployments, but it pushes common pseudonymization logic into every producer or upstream collector instead of providing a reusable processor capability.

### Additional Context

This is related to the proposed `mask` action #3054, but serves a different use case. `mask` removes the original value and replaces it with a constant placeholder. `hash` removes the original value while preserving stable correlation for the same input value.

The salt should be supplied through whatever secret/config mechanism the engine supports, for example environment-variable expansion if supported by the config loader. The implementation should avoid logging raw values or salts.

lalitb:
# Change Summary

Adds support for the attributes processor `hash` action.

`hash` replaces existing scalar attribute values with lowercase hex SHA-256 hashes computed as `sha256(salt || value)`. It supports all scalar attribute types and leaves non-scalar or null value unchanged.

The implementation keeps hashing in the existing attribute mutation path, documents the byte encoding used for scalar values, redacts salts from debug output, and materializes transport-optimized attribute batches before rewriting values.

* Closes #3055 

  - `cargo +1.95 fmt --all`
  - `cargo +1.95 check -p otap-df-pdata`

Yes. Users can configure a new attributes processor `hash` action to replace existing scalar attribute values with lowercase hex SHA-256 hashes. Non-scalar and null values are left unchanged.

```yaml
processors:
  attributes/hash:
    actions:
      - action: hash
        key: user.email
        algorithm: sha256
        salt: "${env:ATTRIBUTE_HASH_SALT}"
```

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/attributes_processor/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/attributes_processor/mod.rs
@@ -49,7 +49,7 @@ use otap_df_engine::message::Message;
 use otap_df_engine::node::NodeId;
 use otap_df_engine::process_duration::ComputeDuration;
 use otap_df_engine::processor::ProcessorWrapper;
-use otap_df_otap::{OTAP_PROCESSOR_FACTORIES, pdata::OtapPdata};
+use otap_df_otap::{OTAP_PROCESSOR_FACTORIES, opaque_string::OpaqueString, pdata::OtapPdata};
 use otap_df_pdata::TryIntoWithOptions;
 use otap_df_pdata::otap::transform::apply_attribute_transform;
 use otap_df_pdata::otap::{
@@ -127,58 +127,17 @@ pub enum Action {
         algorithm: HashAlgorithm,
         /// Salt prepended to the attribute bytes before hashing.
         ///
-        /// Stored as a [`RedactedString`] so that derived `Debug` output does not leak the salt
+        /// Stored as an [`OpaqueString`] so that derived `Debug` output does not leak the salt
         /// into logs, panics, or diagnostic dumps. The salt is still round-trippable via
         /// `Serialize`/`Deserialize`, so configuration reload and persistence are unaffected.
         #[serde(default)]
-        salt: RedactedString,
+        salt: OpaqueString,
     },
 
     /// Other actions are accepted for forward-compatibility but ignored.
     /// These variants allow deserialization of Go-style configs without effect.
     #[serde(other)]
     Unsupported,
-}
-
-/// A wrapper around `String` whose `Debug` representation redacts the inner value.
-///
-/// Use this for configuration fields that hold secrets (salts, tokens, ...). `Serialize` and
-/// `Deserialize` are transparent: the value round-trips through JSON/YAML as a plain string,
-/// but any code that formats the wrapper with `{:?}` will see `RedactedString(<redacted>)` rather
-/// than the secret. `Display` is intentionally not implemented so accidental `{}` formatting is
-/// rejected at compile time.
-#[derive(Clone, Default, Serialize, Deserialize)]
-#[serde(transparent)]
-pub struct RedactedString(String);
-
-impl RedactedString {
-    /// Wraps a `String` so it is redacted in `Debug` output.
-    #[must_use]
-    pub fn new(value: String) -> Self {
-        Self(value)
-    }
-
-    /// Returns the underlying string, consuming the wrapper.
-    #[must_use]
-    pub fn into_inner(self) -> String {
-        self.0
-    }
-
-    /// Returns a borrowed reference to the underlying string.
-    #[must_use]
-    pub fn as_str(&self) -> &str {
-        &self.0
-    }
-}
-
-impl std::fmt::Debug for RedactedString {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        if self.0.is_empty() {
-            f.write_str("RedactedString(\"\")")
-        } else {
-            f.write_str("RedactedString(<redacted>)")
-        }
-    }
 }
 
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
@@ -415,7 +374,7 @@ impl AttributesProcessor {
         &self,
         records: &mut OtapArrowRecords,
         signal: SignalType,
-    ) -> Result<(u64, u64, u64, u64, u64), EngineError> {
+    ) -> Result<(u64, u64, u64, u64, u64, u64), EngineError> {
         let mut deleted_total: u64 = 0;
         let mut renamed_total: u64 = 0;
         let mut inserted_total: u64 = 0;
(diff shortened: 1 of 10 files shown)


rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3595,9 +3667,215 @@ fn apply_inserts_and_upserts<T: ArrowPrimitiveType>(
...
+        return Ok(());
+    };
+    let Ok(value_type) = AttributeValueType::try_from(value_type) else {
+        builder.append_null();
+        return Ok(());
+    };
+
+    let mut hasher = Sha256::new();
+    hasher.update(&hash_spec.salt);
+    match value_type {
```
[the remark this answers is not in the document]
lalitb: done - Reused a pre-salted Sha256 state per hash key and clone it per row.

rust/otap-dataflow/crates/core-nodes/src/processors/attributes_processor/mod.rs
```diff
@@ -104,12 +105,85 @@ pub enum Action {
...
+/// A wrapper around `String` whose `Debug` representation redacts the inner value.
+///
+/// Use this for configuration fields that hold secrets (salts, tokens, ...). `Serialize` and
+/// `Deserialize` are transparent: the value round-trips through JSON/YAML as a plain string,
+/// but any code that formats the wrapper with `{:?}` will see `RedactedString(<redacted>)` rather
+/// than the secret. `Display` is intentionally not implemented so accidental `{}` formatting is
+/// rejected at compile time.
+#[derive(Clone, Default, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct RedactedString(String);
```
albertlockett: We have a type `OpaqueString` which is currently in the cloud_auth module. Maybe we could reuse that?
https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/otap/src/cloud_auth/opaque_string.rs

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3595,9 +3667,221 @@ fn apply_inserts_and_upserts<T: ArrowPrimitiveType>(
...
+        AttributeValueType::Double => attr_arrays
+            .attr_double
+            .as_ref()
+            .and_then(|col| col.value_at(row))
+            .is_some(),
+        AttributeValueType::Bytes => attr_arrays
+            .attr_bytes
+            .as_ref()
+            .and_then(|col| col.slice_at(row))
+            .is_some(),
```
albertlockett: Rather than looking up the value, it might be faster to call `is_valid` which will just check the validity bitmap for the array.

```suggestion
        AttributeValueType::Str => attr_arrays
            .attr_str
            .as_ref()
            .map(|col| col.is_valid(row))
            .unwrap_or_default(),
        AttributeValueType::Bool => attr_arrays
            .attr_bool
            .as_ref()
            .map(|col| col.is_valid(row))
            .unwrap_or_default(),
        AttributeValueType::Int => attr_arrays
            .attr_int
            .as_ref()
            .map(|col| col.is_valid(row))
            .unwrap_or_default(),
        AttributeValueType::Double => attr_arrays
            .attr_double
            .as_ref()
            .map(|col| col.is_valid(row))
            .unwrap_or_default(),
        AttributeValueType::Bytes => attr_arrays
            .attr_bytes
            .as_ref()
            .map(|col| match &col {
                crate::arrays::ByteArrayAccessor::FixedSizeBinary(c) => c.is_valid(row),
                crate::arrays::ByteArrayAccessor::Binary(c) => c.is_valid(row),
            })
            .unwrap_or_default(),
```

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
```diff
@@ -3595,9 +3667,221 @@ fn apply_inserts_and_upserts<T: ArrowPrimitiveType>(
...
+                eq(&key_column, &StringArray::new_scalar(attrs_key)).map_err(|_| {
+                    Error::UnsupportedStringColumnType {
+                        data_type: key_column.data_type().clone(),
+                    }
+                })?;
+            let hashable_key_mask = hashable_scalar_key_mask(&existing_key_mask, &any_value_arrays);
+            let existing_parent_ids = filter(&parent_ids_col, &hashable_key_mask)
+                .map_err(|e| Error::ColumnLengthMismatch { source: e })?;
+            // TODO: avoid materializing parent IDs for update-only transforms once
+            // upsert_attributes can derive update counts without this scratch Vec.
```
albertlockett, line 3766: Agreed, this would be a nice optimization

albertlockett (approved):
nice addition, thanks again @lalitb !

rust/otap-dataflow/crates/core-nodes/src/processors/attributes_processor/mod.rs
lalitb: Good point, thanks. I moved `OpaqueString` out of `cloud_auth` into a shared `otap` module and reused it for the hash salt here.

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
lalitb: Good point. updated accordingly, thanks :)

rust/otap-dataflow/crates/pdata/src/otap/transform.rs
lalitb: Thanks, created #3085 to track this.

PR #3084, opened 2026-05-25
[Attributes Processor] Add attribute update action

### LINKED ISSUE #3054: [Attributes processor] Add mask action for attribute redaction
### Pre-filing checklist

### Component(s)

The attributes processor currently supports actions such as `rename`, `delete`, `insert`, and `upsert`. A common operational use case is replacing sensitive attribute values with a constant placeholder rather than deleting the key entirely.  

This is useful for redaction workflows where operators want to preserve the presence of a field while removing its sensitive value.

### Proposed Solution

This issue proposes adding a `mask` action:

```yaml
  actions:
    - action: mask
      key: user.email
      replacement: "[REDACTED]"
    - action: mask
      key: user.id
      replacement: "[REDACTED]"
    - action: mask
      key: client.ip
      replacement: "0.0.0.0"
```
  Expected behavior:

  - If the target attribute exists, replace its value with the configured replacement.
  - If the target attribute does not exist, do not add a new attribute.
  - If `replacement` is omitted, use a documented default such as `[REDACTED]`.
  - Existing actions remain unchanged.

### Alternatives Considered

  - Use `delete`
    This removes the sensitive value, but also removes the attribute entirely. Some redaction workflows want to preserve that the field was present while removing the original value.

  - Use `upsert`
    `upsert` can replace a value with `[REDACTED]`, but it can also create the attribute when it was absent. The desired mask behavior is update-only: replace the value only when the target attribute already exists, and do not
  create the attribute when it was absent.

  - Use the transform processor
    The transform processor can express constant replacement, but a dedicated `mask` action keeps common redaction workflows simple and explicit in the attributes processor.

### Additional Context

_No response_

lalitb:
# Change Summary

Adds support for the attributes processor `update` action.

`update` replaces existing attribute values for matching keys without inserting missing attributes. This provides the generic primitive needed for redaction-style replacements while preserving existing-only semantics.

The implementation reuses the existing attribute mutation path and handles transport-optimized attribute batches by materializing `parent_id` before value changes that can alter equality runs.

* Closes #3054 

 - `cargo +1.95 fmt --all`
  - `cargo +1.95 check -p otap-df-pdata`

Yes. Users can configure a new attributes processor `update` action to replace existing attribute values for matching keys without inserting missing attributes.

```yaml
processors:
  attributes/update:
    actions:
      - action: update
         key: secret
         value: "[MASKED]"
```

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/attributes_processor/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/attributes_processor/mod.rs
@@ -762,7 +762,7 @@ mod tests {
                         .and_then(|kv| kv.value.as_ref()),
                     Some(&AnyValue::new_string("[MASKED]"))
                 );
-                assert!(!log_attrs.iter().any(|kv| kv.key == "missing"));
+                assert!(!log_attrs.iter().any(|kv| kv.key == "must_not_insert"));
                 assert!(
                     log_attrs.iter().any(|kv| kv.key == "keep"
                         && kv.value == Some(AnyValue::new_string("unchanged")))
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/attributes_processor/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/attributes_processor/mod.rs
@@ -12,9 +12,10 @@
 //! - `delete`: Removes an attribute by key.
 //! - `insert`: Inserts a new attribute (only if the key doesn't already exist).
 //! - `upsert`: Inserts a new attribute, or updates the value if the key already exists.
+//! - `update`: Updates an existing attribute only if the key already exists.
 //!
 //! Unsupported actions are ignored if present in the config:
-//! `update` (value update), `hash`, `extract`, `convert`.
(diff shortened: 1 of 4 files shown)


rust/otap-dataflow/crates/core-nodes/src/processors/attributes_processor/mod.rs
```diff
@@ -630,13 +666,111 @@ mod tests {
...
+
+                let log_attrs = &decoded.resource_logs[0].scope_logs[0].log_records[0].attributes;
+                assert_eq!(
+                    log_attrs
+                        .iter()
+                        .find(|kv| kv.key == "secret")
+                        .and_then(|kv| kv.value.as_ref()),
+                    Some(&AnyValue::new_string("[MASKED]"))
+                );
+                assert!(!log_attrs.iter().any(|kv| kv.key == "missing"));
```
albertlockett: Should this be `must_not_insert`, since that is what we pass in the processor config?
```suggestion
                assert!(!log_attrs.iter().any(|kv| kv.key == "must_not_insert"));
```

albertlockett (approved):
LGTM - thanks @lalitb

PR #3093, opened 2026-05-26
[Host Metrics Receiver] Isolate host metrics scraping in bounded worker

### LINKED ISSUE #3000: [Host metrics receiver] Isolate host metrics procfs/sysfs scraping behind a bounded scraper worker
### Description

As discussed here: https://github.com/open-telemetry/otel-arrow/pull/2840#discussion_r3203092771

The current host metrics receiver keeps scheduling, control handling, snapshot collection, projection, and downstream send in one local receiver task. This is acceptable for v1 because process metrics are aggregate-only, remote/userspace filesystems are opt-in, and `statvfs` already uses a bounded timeout-protected worker.

As a follow-up, we should move the synchronous host inspection work behind an explicit bounded scraper subsystem. The runtime task should remain responsible for scheduling scrape ticks, control messages, projection, and downstream flow control, but `procfs`/`sysfs`/`statvfs` traversal should execute in a dedicated bounded worker that returns a `HostSnapshot`.

The follow-up should preserve the current one-core/singleton collection model while adding clearer resource and latency boundaries:

  - max one in-flight scrape
  - bounded request/result channels
  - scrape deadline or budget
  - defined overrun behavior when a scrape is still running at the next tick
  - runtime shutdown/drain not blocked indefinitely behind host filesystem reads
  - self-observability for scrape latency, overrun/drop/busy counts, and worker failures

This should not move the whole receiver off-runtime. Only the blocking host inspection path should be isolated; scheduling, counter/projection policy, `OTAP` encoding, and downstream send should remain on the receiver side unless there is a strong reason to move them.

lalitb:
# Change Summary

Moves host metrics collection to a bounded dedicated scraper worker so the async runtime thread is no longer responsible for blocking procfs/sysfs/statvfs traversal.

The runtime task keeps ownership of scheduling, control handling, projection/encoding, and downstream flow control. The scraper worker builds the host scrape result and returns it through a bounded handoff.

## Design

This follows the model discussed in #3000:

  - runtime schedules scrape ticks
  - dedicated worker performs blocking host inspection
  - runtime projects/emits OTAP metrics
  - at most one scrape is accepted/in flight
  - bounded request/result handoff
  - explicit timeout/overrun behavior
  - worker failures are observable and terminal where appropriate

* Closes #3000 

- Focused host metrics tests
- Core nodes compile check
- Full workspace check
- E2E VM validation:
    `host_metrics -> OTLP exporter -> OpenTelemetry Collector -> Prometheus -> Grafana`


rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/mod.rs
```diff
@@ -512,18 +550,65 @@ impl local::Receiver<OtapPdata> for HostMetricsReceiver {
...
-                    match source.scrape_due(due).await {
+                    match scrape_worker.try_start_scrape(due) {
                         Ok(scrape) => {
+                            scheduler.advance_due(now);
+                            if let Some(metrics) = metrics.as_mut() {
+                                metrics.scrapes_started.add(1);
+                                metrics.families_scraped.add(due_family_count(due));
+                            }
+                            in_flight = Some(InFlightScrape {
+                                result: Box::pin(timeout(SCRAPE_WORKER_TIMEOUT, scrape)),
```
lquerel: Shouldn't we derive this delay from the host metrics receiver configuration?

lquerel (approved):
LGTM. I just made a small suggestion.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/mod.rs
lalitb: Oh good catch, thanks - I now changed it to derive the scrape timeout from the configured interval for the due families.

PR #3110, opened 2026-05-27
Tighten flow_metrics attribute naming to follow repo conventions

drewrelmas:
# Change Summary

Observed a few deviations from common practices in the repo in the `flow` implementation. This PR updates the standard metric attributes attached to all flow metrics:
- `flow.name` -> `flow.id` (to be more in line with `node.id`)
- `flow.start_node` -> `flow.node.start` (attribute labels normally use dot notation)
- `flow.end_node` -> `flow.node.end` (same)

In addition, updated the user-facing yaml contract from `name` -> `id` to match the attribute set.

Another motivation for this is an offline conversation with @jmacd in which he described flows as `virtual nodes`. That is the main motivation for changing `name` to `id`, to bring it in closer parity with node telemetry.

N/A, minor fix not tracked in an issue

Tests

This is a one-time breaking change in the config declaration of flows, which should be low-impact since it is such a new capability.

### DIFF
--- a/rust/otap-dataflow/configs/fake-flow-metrics-demo.yaml
+++ b/rust/otap-dataflow/configs/fake-flow-metrics-demo.yaml
@@ -44,13 +44,13 @@ groups:
           telemetry:
             runtime_metrics: normal
             flow_metrics:
-              - name: ingest_pipeline
+              - id: ingest_pipeline
                 bounds:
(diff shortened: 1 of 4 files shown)


cijothomas (approved):
LGTM

jmacd (approved):

Issue #1749, opened 2026-01-09
Issue Triage Process and Labels


drewrelmas:
As our issue list has grown, there are a number of stale, duplicate, and no longer needed items.

In addition, our label set has a number of items without any associated issues. Conversely, many issues have no labels.

Opening this as a topic of discussion of how to achieve the following general goals:
- Every issue has at least one label (ideally)
- `good first issue` and `help wanted` are accurately linked for newcomers to the repo to find proper work opportunities
- Limit the number of stale issues and close those no longer needed
- Every PR is linked to an issue (ideally)

It is open to discussion how this group wants to use existing Project: https://github.com/orgs/open-telemetry/projects/139 along with Status assignment

drewrelmas:
As a first step, I've created the `stale` label and will try to start applying it on issues created before the Rust effort began in this repository (everything before ~Feb 2025).

ThomsonTan:
For the issues on and after Feb 2025, could we add action to label them as stale automatically after sometime without any update? Like in below example.

https://github.com/open-telemetry/community/blob/e74afc151a9db50e8fb3a05532d9831186e97b0a/.github/workflows/issue-management-stale-action.yml#L18

drewrelmas:
Planned agenda for 2025-01-13 SIG:

- Component-specific labels (`exporter/otlp`, `processor/attributes`) or only generic (`exporter`, `processor`)?
  - Rearranging `crates/otap/src` into `processor/`, `exporter/`, `receiver`, `contrib/processor/`, `contrib/exporter/`, and `contrib/receiver` 
- Issue 'Type' usage (`Feature`, `Task`, and `Bug`)
- `pipeline`, `engine-capability`, `flow-control`, `data-processing` duplication?

albertlockett:
>Every PR is linked to an issue (ideally)

For having all PRs linked to issue, we could try having a PR template that encourages including an issue number. 

For example, here's what arrow-rs uses:
https://github.com/apache/arrow-rs/blob/f122d7786837d1fc6d93ad45c13067fbfe93eb3e/.github/pull_request_template.md?plain=1#L1-L7

ThomsonTan:
There are [50 labels](https://github.com/open-telemetry/otel-arrow/labels) in total in this repo. I’d like to propose a label renaming pass to make the labels more structured and easier to scan/filter.

## Goal
[...]

PR #3112, opened 2026-05-27
chore(repo): Overhaul otel-arrow Changelog generation process to use `chloggen`

drewrelmas:
# Change Summary

## Changelog process
 
Adopts [chloggen](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/.chloggen) with **per-language changelogs**:
 
 - `go/CHANGELOG.md` (moved from root; the old file was Go-only despite the name).
 - `rust/otap-dataflow/CHANGELOG.md` (was empty; now seeded).
 - Root `CHANGELOG.md` is a pointer file.
 
Each tree has its own `.chloggen/` directory and a short `components:` allowlist. New `Makefile` targets (`chlog-install`, `chlog-new-go`, `chlog-new-rust`, `chlog-validate`, `chlog-preview`, `chlog-update`) pinned to chloggen `v0.30.0`.
 
New `.github/workflows/changelog.yml` requires a `.chloggen/*.yaml` entry on every PR. Bypass: bot actors, `dependencies`/`chore` label, or `chore` in title. Edits to existing entries also count.
 
 Contributor docs updated: `CONTRIBUTING.md`, `rust/otap-dataflow/CONTRIBUTING.md`, and PR template.
 
 ## PR auto-labeling refresh
 
 Renames and expands `.github/labeler.yml` to use #1749's taxonomy (`lang:*`, `area:*`, `ci-repo`, `release`, `documentation`). Path globs updated to match current code layout (e.g. opl/ottl moved under `query-engine-languages/`).
 
 > Labels must exist in GitHub before they apply; missing ones produce warnings, not failures.
 
 ## Out of scope (follow-ups)
 
 - Release-workflow integration (Stage 2)
 - PR-title standardization (Stage 3)

Progress towards:
- #1423
- #1749

Local runs of `chloggen` commands

Yes, added a new CI workflow to track Changelog entry addition/modification. Not currently enforced, in trial mode to start.


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,84 @@
...
+  cancel-in-progress: true
+
+jobs:
+  changelog:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor != 'dependabot[bot]' && github.actor != 'renovate[bot]' && github.actor != 'otelbot' }}
+    env:
+      PR_HEAD: ${{ github.event.pull_request.head.sha }}
+      # Reusable expression: true when any bypass condition is set
+      SKIP: ${{ contains(github.event.pull_request.labels.*.name, 'dependencies') || contains(github.event.pull_request.labels.*.name, 'chore') || contains(github.event.pull_request.title, '[chore]') }}
```
cijothomas: [chore], chore: - lets do both. 
see https://github.com/open-telemetry/opentelemetry-specification/pull/4865

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,84 @@
...
+# required (e.g. internal refactors, dep bumps), include "[chore]" in
+# the PR title or apply the "chore" label.
+#
+# Modeled on opentelemetry-collector-contrib/.github/workflows/changelog.yml.
+
+name: changelog
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
```
cijothomas:
```suggestion
    types: [opened, synchronize, reopened, labeled, unlabeled, edited]
```

.github/pull_request_template.md
```diff
@@ -17,3 +17,17 @@ We highly recommend correlation of every PR to an issue
...
+### Changelog
+
+<!--
+If this PR has user-facing changes, add a `.chloggen/*.yaml` entry via
+`make chlog-new-go FILENAME=<short-name>` (Go changes) or
+`make chlog-new-rust FILENAME=<short-name>` (Rust changes). See
+go/.chloggen/README.md or rust/otap-dataflow/.chloggen/README.md.
+
+If a changelog entry is not required, include `[chore]` in the PR title
+or apply the `chore` label.
```
cijothomas: nits:
apply label - only triggers/above can do it, so we can probably not mention this at all in the template. (for those with permissions, they would likely know about this.). Or maybe just use PR title alone and avoid label totally.

.github/workflows/changelog.yml
drewrelmas: Sure, flipping to `contains(..., 'chore')` which should handle both

CONTRIBUTING.md
```diff
@@ -83,6 +83,46 @@ Examples of `<type>`: `feat` (new feature), `fix` (bug fix), `docs`
...
+[`rust/otap-dataflow/.chloggen/README.md`](./rust/otap-dataflow/.chloggen/README.md)
+for the full guide and the allowed `component:` values (defined in the
+`config.yaml` of each directory).
+
+**Skipping a changelog entry.** Use any of the following when the change is
+not user-facing (build chores, internal refactors, doc-only edits, dep bumps):
+
+- Include `chore` in the PR title.
+- Apply the `chore` label.
+- For dependency-update PRs: Renovate auto-applies the `dependencies` label
```
cijothomas, line 123: if the dependency is part of the rust binary, then bumping their version is worth change logging?

.github/pull_request_template.md
drewrelmas: Fair, I am removing it from the PR template. I think we should keep label functionality just in case, doesn't hurt.

CONTRIBUTING.md
drewrelmas: I worry about clogging up the Changelog with (mostly) silent dependency upgrades, we get a TON of Renovate PRs. Perhaps we can automate this away in the Release process and add one additional entry listing all dependencies PRs together. I would file this away under future improvement.

cijothomas: agree! A single entry during release process make sense.

.github/labeler.yml
```diff
@@ -1,28 +1,163 @@
...
+# --- Query engine -------------------------------------------------------
+
+area:query-engine:
+  - changed-files:
+      - any-glob-to-any-file:
+          - "rust/otap-dataflow/crates/query-engine/**"
+          - "rust/otap-dataflow/crates/query-engine-languages/**"
+          - "rust/experimental/query_engine/**"
+
+area:query-engine/columnar:
```
albertlockett: if we're changing the label, I might suggest `area:query-engine/otap` to distinguish it from any other query engines that may process data in a columnar representation. The query-engine in otap-dataflow is specifically focused on processing OTAP (it's not a generic engine like recordset for eample).

```suggestion
area:query-engine/otap:
```

albertlockett (approved):

jmacd (approved):

rust/otap-dataflow/CONTRIBUTING.md
```diff
@@ -28,6 +28,22 @@ the full check path, likely because many of the longest tests are concentrated
...
+
+User-facing Rust changes are recorded in
+[`CHANGELOG.md`](./CHANGELOG.md). Changelog entries are added per PR as YAML
+files under [`.chloggen/`](./.chloggen/) in this directory and collapsed into
+the CHANGELOG at release time.
+
+From the repo root:
+
+```bash
+make chlog-new-rust FILENAME=my-short-description
```
cijothomas: nit: add an example

CONTRIBUTING.md
```diff
@@ -83,6 +83,50 @@ Examples of `<type>`: `feat` (new feature), `fix` (bug fix), `docs`
...
+make chlog-new-go   FILENAME=my-change   # for Go changes
+make chlog-new-rust FILENAME=my-change   # for Rust changes
+# edit the generated file - set change_type, component, note, issues
+make chlog-validate
+make chlog-preview                       # optional: render without writing
+```
+
+See [`go/.chloggen/README.md`](./go/.chloggen/README.md) and
+[`rust/otap-dataflow/.chloggen/README.md`](./rust/otap-dataflow/.chloggen/README.md)
+for the full guide and the allowed `component:` values (defined in the
```
cijothomas, line 114: this is pointing to the rust/contributing for "full" guide. but rust one is pointing back here for "full" guide 🤣 that is a circular reference bug!

rust/otap-dataflow/CONTRIBUTING.md
drewrelmas: I did add an example in this PR itself under both Go and Rust. To be honest, the `make` commands for this aren't very useful, you can just copy/paste/rename TEMPLATE.yaml which produces same result.

I can't leave a persistent example because the process consumes all yaml files.

lalitb:
LGTM. Just to be noted that the current release process would be broken until Stage 2 wires it to the new per-language changelog flow.

lalitb (approved):

CONTRIBUTING.md
drewrelmas: I don't think the Rust one points back here? Maybe you were looking at an older iteration?

drewrelmas:
> LGTM. Just to be noted that the current release process would be broken until Stage 2 wires it to the new per-language changelog flow.

That is true, but I plan to jump on that as soon as this merges and I believe I'm the only one that has queued releases in the past year 😄

CONTRIBUTING.md
cijothomas: > See [`.chloggen/README.md`](./.chloggen/README.md) and the project-level
> [CONTRIBUTING][] "Changelog entries" section for the full workflow.

^ is from the rust/otap-dataflow/contributing.md file, which is pointing to the main contributing.md file?

rust/otap-dataflow/CONTRIBUTING.md
cijothomas: I meant an example command itself like 
`make chlog-new-rust FILENAME=exporter-fix-data-loss`

it is unclear what is short-name otherwise. could it have spaces? Is that going to become filename.yaml?
New contributors would find this a lot of steps just to get a changelog. (I faced this myself in other repos myself)

CONTRIBUTING.md
drewrelmas: Ahh I see what you mean, fixing!

rust/otap-dataflow/CONTRIBUTING.md
drewrelmas: Got it! I think much improved in https://github.com/open-telemetry/otel-arrow/pull/3112/commits/14c9197d4ae6bf5eafe1c9ae48c9d169e34b7f53

CONTRIBUTING.md
drewrelmas: Fixed in https://github.com/open-telemetry/otel-arrow/pull/3112/commits/14c9197d4ae6bf5eafe1c9ae48c9d169e34b7f53

lquerel (approved):
LGTM

ThomsonTan:
Just curious whether there is any alternative of `chloggen` tool in in go? I removed the go toolset from my devbox for a while and seems I need to get it back.

drewrelmas:
@ThomsonTan the process should be designed in such a way that individual contributors don't need the Go toolset. Simply copying the TEMPLATE.yaml file is fine, only the release automation will need to run the other `make` commands.

PR #3115, opened 2026-05-27
fix(flaky-test-tracker): use fixed issue number for flaky-test-tracker

AaronRM:
# Change Summary

Problem: Last night, the flaky-test-tracker workflow created duplicate issue #3101 instead of updating existing issue #2720, due to GitHub's search/label-filter APIs being stale and failing to return the open issue.

Solution: Hard-code the tracking issue number (#2720) instead of searching for it. This eliminates dependency on flaky GitHub search indexes.

n/a

Will validate after the next scheduled run tonight.

No. This is an internal infra change.


cijothomas (approved):

.github/workflows/flaky-test-tracker.yml
```diff
@@ -89,33 +94,35 @@ jobs:
...
-            | head -1)
-
-          if [[ -n "$EXISTING_ISSUE" ]]; then
-            echo "Updating existing issue #$EXISTING_ISSUE"
-            gh issue edit "$EXISTING_ISSUE" --body "$BODY"
+          FLAKY_COUNT="${{ steps.parse.outputs.flaky_count }}"
+          FLAKY_COUNT="${FLAKY_COUNT:-0}"
+
+          echo "Updating tracking issue #$FLAKY_ISSUE_NUMBER (flaky_count: $FLAKY_COUNT)"
+          gh issue edit "$FLAKY_ISSUE_NUMBER" \
```
ThomsonTan, line 109: Should the direct issue-number lookup apply to the report comparison path too?

The PR now writes to the canonical tracking issue by number, which fixes the duplicate-creation path. But `parse_flaky.py` still reads the previous report body through `gh issue list --label ... --state open --search ...` here:

https://github.com/open-telemetry/otel-arrow/blob/7e9b8f342c6717d08f85d0fc9ab3275f595bdac1/.github/workflows/scripts/parse_flaky.py#L329

Would it make sense to pass `FLAKY_ISSUE_NUMBER` into the parse step and have `get_previous_flaky_names()` use:

```bash
gh issue view "$FLAKY_ISSUE_NUMBER" --json body --jq .body
```

AaronRM: Good catch, thanks!

lquerel (approved):

PR #3134, opened 2026-05-29
[Journald receiver] Implement journald ingestion

### LINKED ISSUE #2858: Logs receiver for systemd / journald logs

lalitb:
# Change Summary

Adds the Rust `receiver:journald` implementation for Linux `systemd-journald` ingestion.

This includes `sd-journal` based log reading, batching, OTAP Arrow log projection, Ack/Nack-driven checkpoint advancement, durable cursor checkpointing, extraction limits, receiver metrics, documentation, and changelog entry.

The README and design doc document current operational limits, including one active owner per host journal source/checkpoint identity, process-local duplicate protection only, deferred cross-process locking, and deferred production hardening for partial journal access, first-checkpoint anchoring, and aggregate batch byte limits.

  * Closes #2858 

  * Added unit tests for config validation, checkpoint read/write, corrupt checkpoint handling, log projection, extraction limits, lease behavior, and Ack/Nack/drain state handling.
  * Ran targeted Rust checks locally:
    * `cargo check -p otap-df-core-nodes --features otap-df-otap/crypto-ring`
    * `cargo test -p otap-df-core-nodes --features otap-df-otap/crypto-ring journald_receiver`
    * `cargo clippy -p otap-df-core-nodes --features otap-df-otap/crypto-ring --all-targets -- -D warnings`
  * Ran markdown and sanity checks for docs/changelog updates.
  *   Manually validated on an Ubuntu VM using real `systemd-journald` input through the `df_engine` pipeline to local debug/noop output. Covered basic ingestion, unit filtering, stdout/stderr, metadata mapping, start positions, checkpoint restart behavior, checkpoint deletion behavior, invalid config, missing permissions smoke, lifecycle shutdown, and high volume smoke.

  Yes. This adds a Linux-only `receiver:journald` component for reading logs from `systemd-journald`.

  ### Changelog

### DIFF
--- a/rust/otap-dataflow/Cargo.toml
+++ b/rust/otap-dataflow/Cargo.toml
@@ -218,7 +218,7 @@ xxhash-rust = { version = "0.8", features = ["xxh3"] }
 zip = "=8.6.0"
 byte-unit = { version = "5.2.0", features = ["serde"] }
 cpu-time = "1.0.0"
-one_collect = { git = "https://github.com/microsoft/one-collect.git", rev = "f655a30aec1d15467501ae9f735defb2d6b86581" }
+one_collect = { git = "https://github.com/microsoft/one-collect.git", rev = "293b7d3b135f862bf3a128a98b2a9715c97f3994" }
 
 azure_core = {version = "1.0.0", default-features = false }
 azure_identity = {version = "1.0.0", default-features = false }
--- a/rust/otap-dataflow/Cargo.toml
+++ b/rust/otap-dataflow/Cargo.toml
@@ -218,7 +218,7 @@ xxhash-rust = { version = "0.8", features = ["xxh3"] }
 zip = "=8.6.0"
 byte-unit = { version = "5.2.0", features = ["serde"] }
 cpu-time = "1.0.0"
-one_collect = { git = "https://github.com/microsoft/one-collect.git", rev = "293b7d3b135f862bf3a128a98b2a9715c97f3994" }
+one_collect = { git = "https://github.com/microsoft/one-collect.git", rev = "f655a30aec1d15467501ae9f735defb2d6b86581" }
 
 azure_core = {version = "1.0.0", default-features = false }
 azure_identity = {version = "1.0.0", default-features = false }
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
@@ -126,6 +126,31 @@ config:
     on_nack: rewind
 ```
 
+## Configuration Options
+
+| Field | Type | Default | Description |
+| ----- | ---- | ------- | ----------- |
+| `source_id` | string | `system` | Stable source identifier used for checkpoint paths and telemetry labels. |
+| `journal.root_path` | path | `/` | Local or mounted host root used for `sd-journal` access. |
+| `journal.namespace` | string | unset | Named namespaces are rejected in v1. |
+| `units` | list | `[]` | Exact `_SYSTEMD_UNIT` matches. |
+| `identifiers` | list | `[]` | Exact `SYSLOG_IDENTIFIER` matches. |
+| `priorities` | list | unset | Exact journald `PRIORITY` values to include. When unset, no priority filter is installed. |
+| `max_priority` | enum | unset | Shorthand for all priorities up to the selected level. Mutually exclusive with `priorities`. |
+| `start_at` | enum | `end` | `end` reads new entries only when no checkpoint exists; `beginning` reads existing entries. |
+| `batch.max_records` | integer | `1024` | Maximum log records per emitted batch. |
+| `batch.max_flush_period` | duration | `200ms` | Maximum time to hold a partial batch. |
+| `extraction.max_entry_bytes` | byte size | `1MiB` | Maximum copied bytes per journal entry. |
+| `extraction.max_field_bytes` | byte size | `256KiB` | Maximum copied bytes per field value. |
+| `extraction.max_fields_per_entry` | integer | `256` | Maximum copied fields per journal entry. |
+| `extraction.large_field_policy` | enum | `drop_and_count` | Oversized fields are dropped and counted. |
+| `checkpoint.directory` | path | `${engine.state_dir}/journald` | Root directory for durable cursor checkpoints. |
+| `checkpoint.max_in_flight_batches` | integer | `1` | Must be `1` in v1. |
+| `checkpoint.on_nack` | enum | `rewind` | Either `rewind` or `fail`. |
+| `checkpoint.max_consecutive_failures` | integer | `5` | Consecutive checkpoint write failures before failing the source. |
+| `wait_timeout` | duration | `1s` | Bounds idle wait and shutdown responsiveness. |
+| `drain_timeout` | duration | `5s` | Drain deadline budget; must exceed `wait_timeout`. |
+
 ## Telemetry
 
 These tables list telemetry emitted directly by this node. Common engine
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
@@ -181,6 +181,10 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
   Startup fails clearly when journal files are present but unreadable.
 - Must run in a one-core source pipeline. Use `receiver:journald` followed by a
   topic exporter to fan out to multicore downstream processing.
+- Run one active journald receiver per host journal source
+  (`journal.root_path` + namespace) unless duplicate collection is intentional.
+  The receiver rejects duplicate journal sources inside one process, but v1 does
+  not coordinate ownership across separate collector processes.
 - Named journal namespaces are not supported in v1.
 - Kernel ring-buffer (`dmesg`) ingestion is not supported by this receiver.
 - `checkpoint.max_in_flight_batches` must be `1` in v1.
@@ -189,7 +193,9 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
 - Only one receiver in a process can target the same concrete journal source.
 - Duplicate journald fields are emitted as repeated same-key attributes in the
   first implementation. Array coalescing is planned as a follow-up.
-- Cross-process duplicate readers are not prevented in v1.
+- Cross-process source locking is not implemented in v1; run one owner for a
+  checkpoint identity (`source_id` and checkpoint directory) to avoid cursor
+  races.
 - NUMA pinning and placement are future work.
 
 ## Related Docs
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
@@ -185,6 +185,9 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
   (`journal.root_path` + namespace) unless duplicate collection is intentional.
   The receiver rejects duplicate journal sources inside one process, but v1 does
   not coordinate ownership across separate collector processes.
+  Filters such as units, identifiers, and priorities do not define separate
+  source ownership; two receivers with different filters but the same journal
+  root/namespace still conflict in the same process.
 - Named journal namespaces are not supported in v1.
 - Kernel ring-buffer (`dmesg`) ingestion is not supported by this receiver.
 - `checkpoint.max_in_flight_batches` must be `1` in v1.
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
@@ -10,25 +10,42 @@
 
 ## Overview
 
-Linux-only receiver for local `systemd-journald` entries. It targets structured
-journal records, uses journald source selection such as units, identifiers, and
-priorities, and is designed around journald cursors for checkpointed progress.
+Linux-only receiver for local `systemd-journald` entries backed by the
+`sd-journal` API. It reads structured journal records through runtime-loaded
+`libsystemd.so.0`, uses journald source selection such as units, identifiers,
+and priorities, and emits OTAP log records for downstream processing.
 
-The current implementation is the first receiver slice: it registers the
-factory, validates configuration, enforces a process-local source lease, and
-handles lifecycle, drain, shutdown, and telemetry control messages. The blocking
-`sd-journal` worker, batch handoff, ACK/NACK tracking, and checkpoint
-persistence are follow-up work.
+The receiver does not exec `journalctl` and does not read `.journal` files
+directly. It uses journald cursors for progress tracking and advances the
+durable checkpoint only after downstream Ack.
 
 ## Getting Started
 
 Start at the end of the default system journal:
 
 ```yaml
-type: receiver:journald
-config:
-  source_id: system
-  start_at: end
+groups:
+  host:
+    pipelines:
+      collect:
+        policies:
+          resources:
+            core_allocation:
+              type: core_count
+              count: 1
+        nodes:
+          journald:
+            type: receiver:journald
+            config:
+              source_id: system
+              start_at: end
+          publish:
+            type: exporter:topic
+            config:
+              topic: journald_logs
+        connections:
+          - from: journald
+            to: publish
 ```
 
 ## Configuration
@@ -62,6 +79,16 @@ config:
     max_records: 1024
     # Maximum time to hold a partial batch before flushing (default: 200ms).
     max_flush_period: 200ms
+
+  extraction:
+    # Maximum copied bytes per journal entry (default: 1MiB).
+    max_entry_bytes: 1MiB
+    # Maximum copied bytes per field value (default: 256KiB).
+    max_field_bytes: 256KiB
+    # Maximum copied fields per journal entry (default: 256).
+    max_fields_per_entry: 256
+    # Behavior for oversized fields: "drop_and_count" (default).
+    large_field_policy: drop_and_count
 
   checkpoint:
     # Root directory for cursor checkpoint files (default:
@@ -125,6 +152,7 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
 ## Limits
 
 - Linux only.
+- Requires `libsystemd.so.0` and permission to read the selected journal.
 - Must run in a one-core source pipeline. Use `receiver:journald` followed by a
   topic exporter to fan out to multicore downstream processing.
 - Named journal namespaces are not supported in v1.
@@ -133,9 +161,10 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
 - `wait_timeout` must be no more than `5s`, and `drain_timeout` must be greater
   than `wait_timeout`.
 - Only one receiver in a process can target the same concrete journal source.
-- The current implementation does not yet emit journal records; the blocking
-  `sd-journal` worker, batch handoff, ACK/NACK tracking, and checkpoint
-  persistence are follow-up work.
+- Duplicate journald fields are emitted as repeated same-key attributes in the
+  first implementation. Array coalescing is planned as a follow-up.
+- Cross-process duplicate readers are not prevented in v1.
+- NUMA pinning and placement are future work.
 
 ## Related Docs
 
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
@@ -178,6 +178,7 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
 
 - Linux only.
 - Requires `libsystemd.so.0` and permission to read the selected journal.
+- Many `unsafe` regions, everywhere Rust calls the C library.
   Startup fails clearly when journal files are present but fully unreadable;
   failing closed on partially readable journal trees is planned for production
   hardening.
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
@@ -178,7 +178,9 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
 
 - Linux only.
 - Requires `libsystemd.so.0` and permission to read the selected journal.
-  Startup fails clearly when journal files are present but unreadable.
+  Startup fails clearly when journal files are present but fully unreadable;
+  failing closed on partially readable journal trees is planned for production
+  hardening.
 - Must run in a one-core source pipeline. Use `receiver:journald` followed by a
   topic exporter to fan out to multicore downstream processing.
 - Run one active journald receiver per host journal source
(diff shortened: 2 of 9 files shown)


rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/checkpoint.rs
```diff
@@ -0,0 +1,175 @@
...
+
+pub(crate) fn read_cursor(path: &Path) -> Result<Option<String>, String> {
+    let bytes = match std::fs::read(path) {
+        Ok(bytes) => bytes,
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => return Ok(None),
+        Err(err) => {
+            return Err(format!(
+                "failed to read checkpoint {}: {err}",
+                path.display()
+            ));
```
lquerel: Why not using a structured error here like we do in the rest of the project?

lquerel: Same remark for the other errors below.

rust/otap-dataflow/Cargo.toml
```diff
@@ -218,7 +218,7 @@ xxhash-rust = { version = "0.8", features = ["xxh3"] }
 zip = "=8.6.0"
 byte-unit = { version = "5.2.0", features = ["serde"] }
 cpu-time = "1.0.0"
-one_collect = { git = "https://github.com/microsoft/one-collect.git", rev = "f655a30aec1d15467501ae9f735defb2d6b86581" }
+one_collect = { git = "https://github.com/microsoft/one-collect.git", rev = "293b7d3b135f862bf3a128a98b2a9715c97f3994" }
```
lquerel: Normal?

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/checkpoint.rs
lalitb: Good point. I used `String` because the worker boundary was carrying simple failure messages, but the checkpoint helper should preserve structured errors and only format at the receiver boundary. I’ll fix that before marking the PR ready for review.

rust/otap-dataflow/Cargo.toml
lalitb: Will revert this, thanks for noticing.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/arrow_records_encoder.rs
```diff
@@ -0,0 +1,254 @@
...
+        }
+    }
+
+    pub(crate) const fn len(&self) -> u16 {
+        self.curr_log_id
+    }
+
+    pub(crate) fn append(&mut self, entry: &JournalEntry) {
+        self.logs
+            .append_time_unix_nano(entry.realtime_unix_nano as i64);
```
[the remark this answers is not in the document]
lalitb: Valid. I changed this to use `i64::try_from(...).unwrap_or(i64::MAX)` so a pathological future timestamp cannot wrap negative.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/mod.rs
```diff
@@ -213,42 +727,354 @@ impl local::Receiver<OtapPdata> for JournaldReceiver {
...
-                        source_id = config.source_id.as_str()
-                    );
+                }, if drain_deadline.is_some() => {
+                    let deadline = drain_deadline.expect("drain deadline must be set");
+                    let _ =
+                        send_worker_command(&worker.cmd_tx, WorkerCommand::Shutdown, &effect_handler).await;
+                    drop(event_rx);
+                    join_worker(worker, &effect_handler).await?;
                     effect_handler.notify_receiver_drained().await?;
                     return Ok(terminal_state(deadline, &metrics));
```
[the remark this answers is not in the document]
lalitb: Valid. I added an explicit `journald_receiver.drain_timeout` warning when the drain deadline fires with pending batches still awaiting Ack/Nack, including the pending batch count.

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -257,6 +266,11 @@ downstream backpressure, the async task races the blocked send against lifecycle
 control messages and the worker polls its command channel while holding a full
 handoff batch.
 
+During drain, a batch that has already been read from journald but is queued or
+currently being sent downstream is not abandoned. Drain first stops additional
+worker reads, then lets the current batch reach a terminal Ack/Commit,
+Nack/Rewind, or configured fail outcome before the receiver reports drained.
```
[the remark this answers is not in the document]
lalitb: Valid. I updated the design doc to clarify that drain attempts to wait for terminal completion, but if the drain deadline expires first the receiver shuts down without advancing the checkpoint and uncommitted entries may replay on restart.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/mod.rs
```diff
@@ -213,42 +727,354 @@ impl local::Receiver<OtapPdata> for JournaldReceiver {
...
+                                                    std::future::pending::<()>().await;
+                                                }
+                                            }, if drain_deadline.is_some() => {
+                                                let deadline = drain_deadline.expect("drain deadline must be set");
+                                                let _ =
+                                                    send_worker_command(&worker.cmd_tx, WorkerCommand::Shutdown, &effect_handler).await;
+                                                drop(event_rx);
+                                                join_worker(worker, &effect_handler).await?;
+                                                effect_handler.notify_receiver_drained().await?;
+                                                return Ok(terminal_state(deadline, &metrics));
```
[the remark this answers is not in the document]
lalitb: Valid. I added a `journald_receiver.drain_timeout` warning for the blocked-send deadline path as well, including the affected `batch_id`.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/arrow_records_encoder.rs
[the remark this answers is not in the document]
lalitb: Valid. Fixed by using `i64::try_from(entry.realtime_unix_nano).unwrap_or(i64::MAX)` instead of a direct cast, so pathological future timestamps cannot wrap negative.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/mod.rs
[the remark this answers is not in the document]
lalitb: Valid. Fixed by emitting `journald_receiver.drain_timeout` when the drain deadline fires with pending batches still awaiting Ack/Nack. The warning includes the pending batch count and makes clear that checkpoints are not advanced.

rust/otap-dataflow/docs/journald-receiver.md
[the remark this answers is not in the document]
lalitb: Valid. Fixed  by updating the design doc: drain attempts to let already-read batches reach terminal completion, but if the deadline expires first the receiver shuts down without advancing the checkpoint and uncommitted entries may replay on restart.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/mod.rs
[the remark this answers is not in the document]
lalitb: Valid. Fixed by adding a `journald_receiver.drain_timeout` warning in the blocked-send deadline path, including the affected `batch_id`.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
```diff
@@ -0,0 +1,657 @@
...
+            });
+        }
+        Ok(())
+    }
+
+    fn inspect_journal_path(path: &Path, depth: usize, summary: &mut JournalAccessSummary) {
+        const MAX_DEPTH: usize = 4;
+        if depth > MAX_DEPTH {
+            return;
+        }
```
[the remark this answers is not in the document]
lalitb: Thanks, but I’m going to leave this as-is for now. The extra walk is startup-only and depth-bounded, while keeping the full scan preserves the access summary we need for diagnostics and for the planned partial-access warning/fail-closed hardening. Short-circuiting would be a small optimization, but it makes that follow up easier to miss or implement incorrectly.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
```diff
@@ -0,0 +1,657 @@
...
+        let metadata = match std::fs::metadata(path) {
+            Ok(metadata) => metadata,
+            Err(err) if err.kind() == std::io::ErrorKind::NotFound => return,
+            Err(err) => {
+                if err.kind() == std::io::ErrorKind::PermissionDenied {
+                    summary.unreadable_directories =
+                        summary.unreadable_directories.saturating_add(1);
+                }
+                record_first_error(summary, path, &err);
+                return;
```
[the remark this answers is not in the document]
lalitb: I don’t think this is a real issue for normal Linux journald paths. If a `.journal` file itself lacks read permission, `metadata()` should still succeed and the later `File::open()` path counts it as an unreadable file. A `PermissionDenied` from `metadata()` usually means a parent directory is not traversable, so treating it as an unreadable directory is the right diagnostic. I’d leave this as-is.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/checkpoint.rs
```diff
@@ -0,0 +1,271 @@
...
+    dir.sync_all()
+        .map_err(|source| CheckpointError::FsyncDirectory {
+            path: parent.to_path_buf(),
+            source,
+        })
+}
+
+#[cfg(not(target_os = "linux"))]
+fn sync_parent_directory(_parent: &Path) -> Result<(), CheckpointError> {
+    Ok(())
```
jmacd, line 164: Should we file any issues?

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
```diff
@@ -0,0 +1,156 @@
...
+| `checkpoint.max_in_flight_batches` | integer | `1` | Must be `1` in v1. |
+| `checkpoint.on_nack` | enum | `rewind` | Either `rewind` or `fail`. |
+| `checkpoint.max_consecutive_failures` | integer | `5` | Consecutive checkpoint write failures before failing the source. |
+| `wait_timeout` | duration | `1s` | Bounds idle wait and shutdown responsiveness. |
+| `drain_timeout` | duration | `5s` | Drain deadline budget; must exceed `wait_timeout`. |
+
+## Current Limits
+
+- Linux only.
+- Requires `libsystemd.so.0` and permission to read the selected journal.
```
jmacd, line 181:
```suggestion
- Requires `libsystemd.so.0` and permission to read the selected journal.
- Many `unsafe` regions, everywhere Rust calls the C library.
```

jmacd (approved):

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/checkpoint.rs
lalitb: I don’t think we need a follow-up issue for this receiver. `receiver:journald` is Linux-only.  If we later reuse this checkpoint helper for a cross-platform receiver or file-log receiver, we can revisit platform-specific directory durability then.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
lalitb: Good point. I reworded this a bit: instead of calling out `unsafe` directly, the README now says we use direct `sd-journal` FFI calls to `libsystemd.so.0`, so this is limited to Linux environments with that library available. That felt like the more useful operator-facing limit

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
```diff
@@ -0,0 +1,657 @@
...
+                root_path: root_path.to_path_buf(),
+                journal_files: summary.journal_files,
+                unreadable_files: summary.unreadable_files,
+                unreadable_directories: summary.unreadable_directories,
+                first_error: summary
+                    .first_error
+                    .unwrap_or_else(|| "permission denied".to_owned()),
+            });
+        }
+        if root_path != Path::new("/") && summary.visible_directories == 0 {
```
lquerel: This code exempts the default root / from the "no visible journal directory" preflight failure, but the design doc says no visible journal directory is a startup failure. On a Linux host/container with no /run/log/journal or /var/log/journal, sd_journal_open can succeed with an empty handle, making the receiver look idle instead of failing clearly. I'd either remove the / exemption or explicitly document that only non-root mounted host roots fail this preflight.

lquerel (approved):
I just left one minor comment. Once it's addressed, I think we can merge it.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
lalitb: Good catch.  Fixed in https://github.com/open-telemetry/otel-arrow/pull/3134/commits/f2eaf225282fcb686a38aa0e7483c5fa2e87ba42 . The preflight now treats the default local root the same as mounted host roots, so startup fails clearly when no journal directories are visible. CI is passing now.

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -293,6 +309,14 @@ If the committed cursor is invalid, stale, or vacuumed, v1 fails closed and
 does not silently fall back to tail or head. The operator must remove the
 checkpoint or choose an explicit recovery action.
 
+Before the first successful checkpoint exists, `start_at: end` has no durable
+resume anchor. If the process crashes after entries are read from journald but
+before the first cursor commit succeeds, restart applies `start_at` again and
+may skip those uncommitted entries. While the process stays alive, a Nack before
+the first checkpoint replays the retained in-flight batch instead of reopening
+at the live tail. A production follow-up should add an initial durable anchor or
+document an operator policy for first-start loss tolerance.
```
hestolz, line 318: One option could be to run `sd_journal_seek_tail` on startup and immediately committing the resulting cursor

lalitb:
@copilot resolve the merge conflicts in this pull request

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
```diff
@@ -0,0 +1,157 @@
...
+```
+
+## Configuration Options
+
+| Field | Type | Default | Description |
+| ----- | ---- | ------- | ----------- |
+| `source_id` | string | `system` | Stable source identifier used for checkpoint paths and telemetry labels. |
+| `journal.root_path` | path | `/` | Local or mounted host root used for `sd-journal` access. |
+| `journal.namespace` | string | unset | Named namespaces are rejected in v1. |
+| `units` | list | `[]` | Exact `_SYSTEMD_UNIT` matches. |
```
hestolz: does `[]` imply all units (and identifiers as per below)? worth clarifying in description. somewhat ambiguous because you can have an empty `_SYSTEMD_UNIT` (e.g. if `_TRANSPORT` is `audit`)

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
```diff
@@ -0,0 +1,157 @@
...
+| `batch.max_records` | integer | `1024` | Maximum log records per emitted batch. |
+| `batch.max_flush_period` | duration | `200ms` | Maximum time to hold a partial batch. |
+| `extraction.max_entry_bytes` | byte size | `1MiB` | Maximum copied bytes per journal entry. |
+| `extraction.max_field_bytes` | byte size | `256KiB` | Maximum copied bytes per field value. |
+| `extraction.max_fields_per_entry` | integer | `256` | Maximum copied fields per journal entry. |
+| `extraction.large_field_policy` | enum | `drop_and_count` | Oversized fields are dropped and counted. |
+| `checkpoint.directory` | path | `${engine.state_dir}/journald` | Root directory for durable cursor checkpoints. `${engine.state_dir}` expands to `$OTAP_DF_STATE_DIR` or `.otap-state` when unset. |
+| `checkpoint.max_in_flight_batches` | integer | `1` | Must be `1` in v1. |
+| `checkpoint.on_nack` | enum | `rewind` | Either `rewind` or `fail`. |
+| `checkpoint.max_consecutive_failures` | integer | `5` | Consecutive checkpoint write failures before failing the source. |
```
hestolz: is there a provision for unbounded retries? e.g. `0` or `-1`

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -523,6 +569,7 @@ Initial severity mapping:
 | `libsystemd.so.0` load failure | startup failure; this receiver has no `journalctl` fallback |
 | `sd_journal_open` / permission failure | startup failure; not treated as an empty stream |
 | invalid `journal.root_path` / no visible journal directory | startup failure; operator must fix the mount or config |
+| partially readable journal tree | v1 may start if at least one journal file is readable; production hardening should fail closed or emit a mandatory warning/metric |
 | checkpoint missing | apply `start_at` |
 | checkpoint corrupt / unknown version | fail closed; operator must remove or migrate it |
 | cursor vacuumed / stale | fail closed; operator must remove the checkpoint or choose an explicit recovery action |
```
hestolz, line 574: feel free to move to follow-ups section:

this should not be a failure, it is recoverable; seeking a cursor that points to a rotated/vacuumed entry will [return the nearest available newer entry](https://www.freedesktop.org/software/systemd/man/latest/sd_journal_get_cursor.html#)
```
Seeking to a cursor position without the specific entry being available locally will seek to the next closest (in terms of time) available entry.
```

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -523,6 +569,7 @@ Initial severity mapping:
 | `libsystemd.so.0` load failure | startup failure; this receiver has no `journalctl` fallback |
 | `sd_journal_open` / permission failure | startup failure; not treated as an empty stream |
 | invalid `journal.root_path` / no visible journal directory | startup failure; operator must fix the mount or config |
+| partially readable journal tree | v1 may start if at least one journal file is readable; production hardening should fail closed or emit a mandatory warning/metric |
 | checkpoint missing | apply `start_at` |
 | checkpoint corrupt / unknown version | fail closed; operator must remove or migrate it |
```
hestolz, line 573: also a followup-type item; i could see an operator wanting configurability over this. some would want automatic fallback (e.g. to the behavior as defined by `start_at`, so that a fresh cursor could be written and self-heal the corruption) rather than letting this fail the whole engine

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -557,6 +604,33 @@ the v1 signal set includes:
...
+- introducing a `JournalSource` trait and `FakeJournalSource` test boundary
+- expanding receiver self-telemetry for stale cursors, permission warnings,
+  worker restarts, last-entry timestamp, and backpressure duration
+- failing closed or warning clearly on partially readable journal trees
+- an initial durable resume anchor for `start_at: end` before the first
+  checkpoint commit
+- aggregate batch byte limits in addition to per-entry and per-field extraction
+  limits
+- named systemd journal namespace support
+- arbitrary journald match expressions
```
hestolz, line 634: +1, a passthrough into `sd_journal_add_match` (and adjacent match-controlling functions) would be super powerful!

rust/otap-dataflow/docs/journald-receiver.md
lalitb: Yes, that makes sense as one possible implementation for the initial-anchor follow-up. I’m leaving it out of this PR because immediately committing a tail cursor changes first-start checkpoint semantics, but I documented the follow-up so we can handle it explicitly.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
lalitb: No, v1 does not support unbounded retries. The config currently requires `max_consecutive_failures` to be greater than zero, and I clarified that in the README now.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/README.md
lalitb: Good point. `[]` means no filter is installed for that field, not “match empty _SYSTEMD_UNIT”. I clarified this for both units and identifiers in the README.

rust/otap-dataflow/docs/journald-receiver.md
lalitb: Agreed this is recoverable. The current v1 code verifies the exact checkpoint cursor and fails closed if it is no longer present, so I kept the current behavior documented and added configurable stale/vacuumed cursor recovery to the follow-up list.

rust/otap-dataflow/docs/journald-receiver.md
lalitb: Agreed. That is the direction I’d expect for the arbitrary-match follow-up: expose a controlled passthrough to `sd_journal_add_match` and related match controls.

rust/otap-dataflow/docs/journald-receiver.md
lalitb: Agreed. Current v1 fails closed for corrupt or unsupported checkpoint envelopes, but I added configurable recovery/fallback to the follow-up list so operators can explicitly opt into behavior like falling back to `start_at` and writing a fresh cursor.

hestolz (approved):

rust/otap-dataflow/docs/journald-receiver.md
```diff
@@ -485,23 +515,39 @@ the receiver must copy the required field names and values before calling
 Attribute projection rules:
 
 - a single occurrence of a field becomes a scalar attribute
-- repeated occurrences of the same field become an array attribute in journal
-  enumeration order
 - the first `MESSAGE` value becomes the OTAP body; all `MESSAGE` values are
-  also preserved in a native journal attribute using the duplicate-field rules
+  also preserved in a native journal attribute
 - binary values are preserved as OTAP bytes when supported; otherwise they are
   base64 encoded with an explicit encoding marker
```
ThomsonTan: Is this base64 encoding fallback implemented? Didn't see it in the encoder.

lalitb: Good catch. don’t think we need fallback in the receiver path because OTAP/OTLP can preserve non-UTF-8 values as bytes. I updated the design doc to remove the base64 fallback claim; any base64/string fallback should be handled later at an exporter or text-only boundary if needed.

Issue #3138, opened 2026-05-29
Add `user_agent` config option to shared `HttpClientSettings` and `GrpcClientSettings`


drewrelmas:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

Add an optional `user_agent` field to the shared `HttpClientSettings` and `GrpcClientSettings` structs so any HTTP or gRPC exporter (and receiver) can inject a custom User-Agent string via config.

The `azure_monitor_exporter` recently added a per-exporter `user_agent` field (#3137). This is useful for server-side request attribution, but the feature is only available to that single exporter. Promoting `user_agent` into the shared client settings would make it available to `otlp_http_exporter`, `otlp_grpc_exporter`, `otap_exporter`, and any future HTTP/gRPC-based node without duplicating the config field in each one.

### Proposed Solution

**`HttpClientSettings`** (`crates/otap/src/otlp_http/client_settings.rs`):
- Add `user_agent: Option<String>`
- Apply in `client_builder()` via `reqwest::ClientBuilder::user_agent()` (replaces the default reqwest UA)

**`GrpcClientSettings`** (`crates/otap/src/otap_grpc/client_settings.rs`):
- Add `user_agent: Option<String>`
- Apply in `build_endpoint_from_uri()` via `tonic::transport::Endpoint::user_agent()` (prepends to the default `tonic/x.x.x` UA)

### Behavior

| Transport | `user_agent` set | Resulting header |
|---|---|---|
| HTTP (reqwest) | `"my-app/1.0"` | `my-app/1.0` |
| HTTP (reqwest) | not set | reqwest default |
| gRPC (tonic) | `"my-app/1.0"` | `my-app/1.0 tonic/x.x.x` |
| gRPC (tonic) | not set | `tonic/x.x.x` |

Note the semantic difference: reqwest **replaces** the User-Agent while tonic **prepends** to its default. This should be documented on the config field.

### Example config

```yaml
# otlp_http_exporter
type: urn:otel:exporter:otlp_http
config:
  http:
    user_agent: "my-pipeline/2.1"
  endpoint: https://collector.example.com

# otlp_grpc_exporter
type: urn:otel:exporter:otlp_grpc
config:
  grpc:
    user_agent: "my-pipeline/2.1"
    grpc_endpoint: https://collector.example.com:4317
```

[...]

PR #3139, opened 2026-05-29
feat(azure_monitor_exporter): Add optional config for UserAgent value

### LINKED ISSUE #3137: Allow optional configuration of UserAgent in `azure_monitor_exporter`
### Pre-filing checklist

### Component(s)

### Objective

For motivation similar to #2355, would like to allow Operators to customize behavior of the `reqwest` HTTP client used internally to export logs to Azure Monitor via populating their own UserAgent values.

### Rationale

Enrich certain telemetry streams within Azure Monitor with UserAgent information, driving parity with other solutions.

### Scope

- Add `user_agent` as an optional config to `azure_monitor_exporter`
- Use that value in construction of HTTP clients

### Acceptance Criteria

- Operators can set their own UserAgent values for `azure_monitor_exporter`.

### Dependencies or Blockers

N/A

### Additional Context

I briefly considered some shared solution that would enable all http-based exporters to customize UserAgent, perhaps by adding to `HttpClientSettings`, which is used by the `otlp_http_exporter`.

https://github.com/open-telemetry/otel-arrow/blob/d624d457aedfdfa11eb8b3d7db98fbc57576a6a8/rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs#L21-L24

However, the `azure_monitor_exporter` currently does not expose this level of control to Operators, instead choosing specific client values on behalf of:

https://github.com/open-telemetry/otel-arrow/blob/d624d457aedfdfa11eb8b3d7db98fbc57576a6a8/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/client.rs#L66-L77

There could be a follow-up to add `HttpClientSettings` to the `azure_monitor_exporter` config, but that is a larger decision compared to the very self-contained change of adding an optional config directly to existing `Config` struct.

drewrelmas:
# Change Summary

Add `user_agent` config to `azure_monitor_exporter`.

As noted in the issue #3137, there is a potential follow-up here to add the same in a convenient place for all other exporters. Opened #3138 to propose this wider scope. For the moment, we are not using the shared client settings in `azure_monitor_exporter`.

* Closes #3137

Unit test and local run

Yes, Operators using `azure_monitor_exporter` may set their own UserAgent values for all HTTP requests sent to Azure Monitor

### Changelog


rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
```diff
@@ -184,6 +184,11 @@ pub struct ApiConfig {
     /// - 9: maximum compression (highest CPU, marginal size reduction over 6)
     #[serde(default = "default_gzip_compression_level")]
     pub gzip_compression_level: u32,
+
+    /// Custom User-Agent string to send with HTTP requests (optional).
+    /// When set, this value is used as the User-Agent header on all outgoing requests
+    /// (including heartbeat). When not set, the default reqwest User-Agent is used.
+    pub user_agent: Option<String>,
```
lalitb, line 192: nit - Should we validate this in `Config::validate()`? I don’t think we need full `User-Agent` grammar validation, but it would be useful to reject values that cannot be represented as a normal HTTP header string, for example by checking `HeaderValue::from_str(user_agent)` plus `to_str().is_ok()`. That would give operators a clearer `api.user_agent` config error. We may also want to reject an empty string if that is not intended.

lalitb (approved):

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
drewrelmas: Good idea, added.

rust/otap-dataflow/.chloggen/ame-user-agent.yaml
```diff
@@ -0,0 +1,7 @@
+change_type: enhancement
+
+component: pipeline
```
cijothomas, line 3: one thing about changelog- could we make it per crate instead of bundling most of them into pipeline? (not for this PR, just noticed while in this!)

drewrelmas: @cijothomas to clarify, do you mean add a larger list of `components` here:

https://github.com/open-telemetry/otel-arrow/blob/d624d457aedfdfa11eb8b3d7db98fbc57576a6a8/rust/otap-dataflow/.chloggen/config.yaml#L18-L29

My thinking was to avoid a contributor needing to think hard about what component to use by making the list small 😄

gouslu (approved):

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
```diff
@@ -235,6 +241,23 @@ impl Config {
...
+            }
+            if HeaderValue::from_str(ua)
+                .ok()
+                .and_then(|v| v.to_str().ok().map(|_| ()))
+                .is_none()
+            {
+                return Err(Error::Config(
+                    "Invalid configuration: user_agent contains characters that cannot be represented as an HTTP header value".to_string(),
+                ));
+            }
```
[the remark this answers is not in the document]
lalitb: Make sense -  Validating with `HeaderValue::from_str()` should be sufficient.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/heartbeat.rs
```diff
@@ -112,14 +112,18 @@ fn default_heartbeat_os_type() -> String {
...
             .pool_max_idle_per_host(MAX_IDLE_CONNECTIONS_PER_HOST)
             .pool_idle_timeout(Duration::from_secs(90))
-            .tcp_nodelay(true)
-            .build()
-            .map_err(Error::CreateClient)?;
+            .tcp_nodelay(true);
+
+        if let Some(ref ua) = config.user_agent {
+            builder = builder.user_agent(ua);
+        }
```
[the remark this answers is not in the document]
drewrelmas: Addressed in slightly different fashion.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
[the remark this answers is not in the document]
drewrelmas: Done

rust/otap-dataflow/.chloggen/ame-user-agent.yaml
drewrelmas: We could consider splitting `pipeline` out into `core-nodes` and `contrib-nodes`

jmacd: +1 this is the hardest part in the collector, which has a component per "node" in our terms, which is obvious except when it's not, so for shared packages and extension interfaces you often have to guess and learn by trial and error.

jmacd (approved):

PR #3143, opened 2026-05-29
feat(engine): Add extension lifecycle monitor + metrics, decouple from PipelineContext

gouslu:
## Change Summary

Adds per-extension lifecycle telemetry and decouples the extension subsystem
from `PipelineContext` so extensions can be hosted at any scope.

- **Scope-agnostic extensions.** New `ExtensionContext` carrying only
  `ControllerContext`; `ExtensionFactory::create` no longer takes
  `PipelineContext`. `broadcast_shutdown` now takes a caller-supplied reason.
- **Per-variant telemetry.** `ExtensionAttributeSet` gains
  `extension_variant` (`local` | `shared`); dual-variant extensions register
  two distinct entities with independent `EntityTelemetryGuard` ownership so
  dropping one variant only tears down its own entity tree.
- **Per-scope attribution.** `ExtensionAttributeSet` carries a nested
  `ExtensionScopeAttributeSet` (today: `pipeline(group, pipeline_id, core,
  num_cores)`) so the same extension `id+variant` hosted at different scopes
  mints distinct `EntityKey`s. Same shape will extend cleanly to engine/group
  scopes.
- **Channel attribute split.** `ChannelAttributeSet` split into
  `NodeChannelAttributeSet` and `ExtensionChannelAttributeSet`; fixes a leak
  where extension channel `MetricSet` keys were never tracked via the entity
  handle.
- **`ExtensionMetricsMonitor`.** New monitor wired into the lifecycle pump
  emits counters (`spawned` / `completed` / `shutdown_sent` / errors /
  timeouts) and a state gauge per extension entity. Hardening:
  - `register()` is the single source of `spawned` credit — it synchronously
    sets `state = Spawned` and increments the `spawned` counter, eliminating
    a class of undercount bugs by construction (no transient pre-Spawned
    state to drop events against).
  - `wait_all_spawned()` is preserved as a pure scheduling barrier so node
    tasks cannot observe partially-initialized extensions.
  - Biased `select!` ordering (started → completions → ticks); sticky-
    terminal monotonic state machine; `mark_stragglers_as_timeout`
    reconciles non-terminal entries at shutdown; `Drop`-time registry
    unregister. Each scope owns its own monitor + lifecycle so ticks and
    shutdowns never cross scopes.
  - State gauge discriminants pinned (`Spawned=1, ShutdownSent=2,
    CompletedOk=3, Failed=4, TimedOut=5`) with `0` reserved as vacant for
    dashboard compatibility; enforced by an exhaustive-match test that
    fails to compile if a new variant is added without updating the
    contract.

Closes "Per-extension lifecycle metrics." on Extension lifecycle follow-ups
#3039.

- ~30+ new unit tests across `extension/tests.rs`, `extension_lifecycle.rs`,
  and `extension_monitor.rs` covering per-variant isolation, `CollectTelemetry`
  round-trip and interval gating, `mark_stragglers_as_timeout` reconciliation,
  sticky-terminal `Completed`, state-gauge refresh, `Drop`-time unregister,
  cross-scope isolation (distinct scopes — pipelines / cores / groups today —
  for both shutdown broadcast and collect-telemetry fanout), live
  two-pipelines-through-shared-reporter, panicking `CollectTelemetry`
  handler, snapshot-after-`Drop` identity, `wire_telemetry` dual
  register/release cycle, `broadcast_shutdown` idempotency, passive
  extensions excluded from monitor, state-gauge integer encoding stability,
  `try_send` failure non-skew, panic-before-`notify_started` surfaced
  through `wait_all_spawned`, and structural invariants ensuring
  `ShutdownSent` and fast completions never undercount `spawned`.
- Existing extension lifecycle and `extension_e2e` integration tests
  updated and passing.
- `cargo test -p otap-df-engine` green (446 lib + 28 e2e).
- `cargo fmt --all` and `cargo clippy -p otap-df-engine --all-targets
  -- -D warnings` clean.

**Breaking** (out-of-tree consumers will get compile errors):

- `ExtensionFactory::create` signature changed: no longer receives
  `PipelineContext`; it now receives `ExtensionContext`. Out-of-tree
  extensions need to be updated.
- `PipelineContext::register_channel_entity` renamed to
  `register_node_channel_entity`; `channel_attribute_set` renamed to
  `node_channel_attribute_set`. The `ChannelAttributeSet` struct is split
  into `NodeChannelAttributeSet` and `ExtensionChannelAttributeSet`.

**New telemetry surface:** `extension.lifecycle.*` metrics with
`extension_id` + `extension_variant` attributes.

**No config or wire-format changes.**


lalitb:
Early thought before doing a deeper review: the direction here looks good, and I agree with removing `PipelineContext` from `ExtensionFactory::create`.

The one model issue I’d look at before merge is that `ExtensionContext` seems scope-less, not scope-agnostic.

Scope-agnostic means extension code should not care whether the extension is hosted by a pipeline, engine, group, or node. But the engine still needs to remember which scope owns the extension, especially for telemetry attribution.

For example, today extensions are still pipeline-hosted. So lifecycle metrics keyed only by this are not enough:

```text
extension_id = auth
extension_variant = shared
```

Two different pipelines can both define an extension named auth:

```text
pipeline = traces
extension_id = auth
extension_variant = shared


pipeline = metrics
extension_id = auth
extension_variant = shared
```

 Without the pipeline/host scope in the attributes, those two become indistinguishable. Since `TelemetryRegistryHandle::register_entity` dedupes by attribute set, they can also be registered as the same entity.

I think `ExtensionContext` should carry a host-scope identity, and `ExtensionAttributeSet` should include that scope. The exact implementation could be an enum, trait, or small helper type; only the pipeline case needs to exist
 today. Please don’t speculatively add engine/group/node variants until there is a real caller. The important part is that adding those scopes later is non-breaking and does not lose telemetry attribution.

One related PR-body note: it says the new telemetry surface is keyed by extension_id + extension_variant. I think that should be host_scope + extension_id + extension_variant; otherwise the scope-agnostic model loses the identity of the scope that owns the extension.

gouslu:
> Early thought before doing a deeper review: the direction here looks good, and I agree with removing `PipelineContext` from `ExtensionFactory::create`.
> 
> The one model issue I’d look at before merge is that `ExtensionContext` seems scope-less, not scope-agnostic.
> 
> Scope-agnostic means extension code should not care whether the extension is hosted by a pipeline, engine, group, or node. But the engine still needs to remember which scope owns the extension, especially for telemetry attribution.
> 
> For example, today extensions are still pipeline-hosted. So lifecycle metrics keyed only by this are not enough:
> 
> ```
> extension_id = auth
> extension_variant = shared
> ```
> 
> Two different pipelines can both define an extension named auth:
> 
> ```
> pipeline = traces
> extension_id = auth
> extension_variant = shared
> 
> 
> pipeline = metrics
> extension_id = auth
> extension_variant = shared
> ```
> 
> Without the pipeline/host scope in the attributes, those two become indistinguishable. Since `TelemetryRegistryHandle::register_entity` dedupes by attribute set, they can also be registered as the same entity.
> 
> I think `ExtensionContext` should carry a host-scope identity, and `ExtensionAttributeSet` should include that scope. The exact implementation could be an enum, trait, or small helper type; only the pipeline case needs to exist today. Please don’t speculatively add engine/group/node variants until there is a real caller. The important part is that adding those scopes later is non-breaking and does not lose telemetry attribution.
> 
> One related PR-body note: it says the new telemetry surface is keyed by extension_id + extension_variant. I think that should be host_scope + extension_id + extension_variant; otherwise the scope-agnostic model loses the identity of the scope that owns the extension.

Thanks, I was already in the process of adding generic context info, should be addressed in f2045b32.

rust/otap-dataflow/crates/engine/src/extension_monitor.rs
```diff
@@ -0,0 +1,1025 @@
...
+
+struct ExtensionMonitorEntry {
+    key: ExtensionKey,
+    state: ExtensionRuntimeState,
+    lifecycle_metrics: MetricSet<ExtensionLifecycleMetrics>,
+    control_sender: Option<ExtensionControlSender>,
+    registry: TelemetryRegistryHandle,
+}
+
+impl Drop for ExtensionMonitorEntry {
```
lalitb, line 106: I think there is still a shutdown-time telemetry race here. The snapshots are enqueued, so  if the collector processes the queued snapshot after this unregister, the registry no longer has the metric key and the final snapshot can be ignored. Could we keep extension lifecycle metric sets registered until the collector has drained, or directly accumulate/flush the final lifecycle snapshot before unregistering?

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
lalitb, line 637: If a node task returns `Err` or panics here, this loop returns early. The cleanup path then broadcasts extension `Shutdown` before the remaining data-path/control tasks are explicitly cancelled or joined. Since this all runs inside the `LocalSet`, those remaining tasks may still be alive while extension-owned resources are already being torn down.

Example: an auth/cache/discovery extension is used by receivers/exporters. If one processor fails, the extension may receive shutdown while another task is still unwinding or still holding a capability handle backed by that extension.

Could we either preserve the same “data path settles first, extensions stop last” ordering on error paths, or document that the stop-last guarantee only applies to normal drain?

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
```diff
@@ -260,6 +275,29 @@ impl<PData: 'static + Debug + Clone + ReceivedAtNode + Unwindable + FlowMetricHo
...
+                    ext_ctx.clone(),
+                    EXTENSION_MONITOR_TICK_INTERVAL,
+                    EXTENSION_MONITOR_COLLECT_TELEMETRY_INTERVAL,
+                )
+            } else {
+                crate::extension_monitor::ExtensionMetricsMonitor::disabled(ext_ctx.clone())
+            }
+        };
+
         // Lifecycle invariant: "extensions start first, shut down last".
```
lalitb, line 301: I’d be careful with the “extensions start first” guarantee here. The current code schedules extension futures before node futures, but that does not mean the extension has completed initialization before nodes begin running.

That distinction matters for real extensions. For example, an auth extension may do this inside `start()`:

  ```rust
  async fn start(...) {
      let token = fetch_token().await;
      self.state.set_token(token);
      // then handle control messages
  }
```

A receiver using that auth capability can start before `fetch_token().await` completes, so it may observe empty/default/not-ready state.

The `TODO` about a readiness probe is the right direction. Until that exists, can we make the public docs explicit that extensions are scheduled before nodes, but capability state must be safe before async start() initialization completes? If we want a stronger contract, we probably need an opt-in readiness handle that the engine waits on before starting data-path tasks.

File: rust/otap-dataflow/crates/engine/src/attributes.rs [hunk 3]
```diff
@@ -97,6 +97,68 @@ pub struct PipelineAttributeSet {
...
+        id: impl Into<Cow<'static, str>>,
+    ) -> Self {
+        Self {
+            kind: kind.into(),
+            id: id.into(),
+        }
+    }
+
+    /// Pipeline-host scope. Id is
+    /// `"<group>/<pipeline>/core/<core>/gen/<generation>"`.
```
lalitb: This fixes the earlier scope-less issue, which is good. But I think encoding the whole host scope as one opaque `scope.id` string is still weaker than the rest of the telemetry model.

Pipeline and node telemetry use typed/composed attributes. For example, `NodeAttributeSet` composes `PipelineAttributeSet`, so operators can filter directly by `pipeline_group_id`, `pipeline_id`, deployment generation, engine/resource attrs, etc. Extension metrics now require parsing `scope.id` to answer the same questions.

There is also a small ambiguity risk because pipeline group/id values are plain strings. For example:

  ```text
  group = "a/b", pipeline = "c"
  group = "a",   pipeline = "b/c"

Both can produce the same delimiter-shaped prefix.

Could pipeline-hosted extensions compose PipelineAttributeSet directly for now? Future engine/group/node scopes can add their own typed scope attribute sets when there is a real caller.

rust/otap-dataflow/crates/engine/src/extension/wrapper.rs
```diff
@@ -39,25 +39,12 @@ use std::sync::Arc;
...
-/// further control messages.
-///
-/// Immediate delivery is appropriate here: the runtime broadcasts
-/// `Shutdown` to extensions only after every data-path node has
-/// drained (see [`crate::extension_lifecycle`]), so the full
-/// `deadline - now()` window is available for cleanup. This differs
-/// from the data-path node control channel in [`crate::message`],
-/// which stashes `Shutdown` and keeps delivering other control
-/// messages until the deadline so nodes can drain in-flight PData.
+/// `Shutdown` is delivered immediately and the `deadline` field is the
```
lalitb: The docs say `Shutdown` is delivered immediately, but it looks like shutdown uses the same FIFO control channel as `CollectTelemetry`. If telemetry messages are already queued, shutdown sits behind them and the extension may not observe it until part of the shutdown budget is already gone.

Example: a slow extension has a few `CollectTelemetry` messages queued. During pipeline shutdown we enqueue `Shutdown`, but the extension receives the telemetry messages first. That weakens the shutdown deadline semantics.

Could we either make shutdown priority/preemptive, drain/drop older telemetry messages when shutdown is queued, or soften the doc to say shutdown is delivered through the normal control queue?

rust/otap-dataflow/crates/engine/src/extension_monitor.rs
```diff
@@ -0,0 +1,1025 @@
...
+            entry.state = ExtensionRuntimeState::Spawned;
+            entry.lifecycle_metrics.spawned.add(1);
+            entry
+                .lifecycle_metrics
+                .state
+                .set(ExtensionRuntimeState::Spawned as u64);
+        }
+    }
+
+    fn on_shutdown_sent(&mut self, key: &ExtensionKey) {
```
lalitb, line 209: This transition only records `shutdown_sent` when the current state is `Spawned`. I think that can undercount real shutdowns.

A fast-draining pipeline can broadcast extension shutdown before the extension task has been polled and before the `Spawned` signal is observed. In that case the extension really did receive a shutdown send, but the monitor entry is still `Pending`, so this branch ignores it.

Could `ShutdownSent` be counted for `Pending` as well, or could `on_completed` reconcile `Pending -> Completed` as “the task did run”? Otherwise short-lived/fast-drain pipelines can report `completed_ok` with `spawned = 0` and `shutdown_sent = 0`, which is confusing operationally.

rust/otap-dataflow/crates/engine/src/extension/wrapper.rs
```diff
@@ -148,6 +129,29 @@ impl EffectHandler {
...
 
+// ── Variant kind ─────────────────────────────────────────────────────────────
+
+/// Identifies a physical extension variant within an [`ExtensionBundle`].
+/// A single `ExtensionId` may register both `Local` and `Shared` variants.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum ExtensionVariant {
+    /// `!Send` variant scheduled onto the host `LocalSet`.
+    Local,
+    /// `Send` variant runnable on multi-threaded executors.
```
lalitb, line 172: nit - Could we make this comment more clear - Something like: “Shared means the extension/capability is `Send`; the lifecycle task is still run on the host pipeline `LocalSet` today.”

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -117,43 +121,130 @@ impl ExtensionLifecycle {
...
 
     /// Broadcasts `Shutdown` to all active+background extensions.
     /// Idempotent. Sends fan out concurrently, each bounded by
-    /// [`EXTENSION_SHUTDOWN_SEND_TIMEOUT`]; the deadline carried in
-    /// the message is `now() + EXTENSION_SHUTDOWN_GRACE` (a fresh
-    /// cleanup window, not a continuation of the pipeline deadline).
-    pub async fn broadcast_shutdown(&mut self) {
+    /// [`EXTENSION_SHUTDOWN_SEND_TIMEOUT`]; `reason` is propagated
+    /// verbatim (defaulting to [`DEFAULT_SHUTDOWN_REASON`]).
+    pub async fn broadcast_shutdown(&mut self, reason: Option<&str>) {
```
lalitb: One shutdown edge case worries me here. `CollectTelemetry` and `Shutdown` share the same bounded extension control channel. If a telemetry message is already queued and the extension is busy, the `Shutdown` send can time out.

Because `shutdown_broadcast_fired` is set before the send attempts, the later cleanup call will not retry. So a low-priority telemetry message can block the only graceful shutdown signal, and the extension may end up timing out instead of cleaning up normally.

Could we either prioritize `Shutdown`, drop stale telemetry before shutdown, use a separate telemetry path, or only latch shutdown delivery after the send succeeds?

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
```diff
@@ -612,48 +632,53 @@ impl<PData: 'static + Debug + Clone + ReceivedAtNode + Unwindable + FlowMetricHo
...
-                                        Err(e) => return Err(Error::JoinTaskError {
-                                            is_canceled: e.is_cancelled(),
-                                            is_panic: e.is_panic(),
-                                            error: e.to_string(),
-                                        }),
+                                event = extension_lifecycle.next_event() => {
+                                    match event {
+                                        crate::extension_lifecycle::LifecycleEvent::Completion(result) => {
+                                            match result {
+                                                Ok(Ok(())) => {}
```
lalitb, line 649: I’m not sure a clean `Ok(())` from an active extension should be treated as harmless while the pipeline is still running.

For real extensions, `start()` is usually the long-lived maintenance loop for things like auth tokens, discovery state, leases, or caches. If that loop exits early, the capability handles held by nodes may still exist, but the background task that keeps them fresh is gone. Operators would see `CompletedOk`, which sounds healthy, while the pipeline may now be using stale state.

Could we make the contract explicit here? For example, active extension completion before shutdown could either be treated as a pipeline error, or recorded as a separate `early_exit` state/counter so it does not look like a normal healthy completion.

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -117,43 +121,130 @@ impl ExtensionLifecycle {
...
+    /// Records the completion outcome in the monitor and reshapes the
+    /// join result. Panic/cancel paths recover the key from
+    /// `task_id_to_key`.
+    fn route_joined(
+        monitor: &mut ExtensionMetricsMonitor,
+        task_id_to_key: &mut HashMap<task::Id, ExtensionKey>,
+        joined: Result<(ExtensionKey, Result<(), Error>), JoinError>,
+    ) -> Result<Result<(), Error>, JoinError> {
+        match joined {
+            Ok((key, res)) => {
```
lalitb, line 293: When an extension task completes, we remove it from `task_id_to_key`, but its sender stays in `shutdown_senders`.

Later cleanup broadcasts `Shutdown` to every saved sender, including extensions whose task already completed and whose receiver was dropped. That can produce an `extension.shutdown.send_failed` warning after the real completion/failure already happened, which makes the shutdown path look broken even when the extension was simply already terminal.

Could we remove completed keys from `shutdown_senders`, or have `broadcast_shutdown` skip entries that are already in a terminal monitor state?

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
gouslu: for start guarantees, we'll have a readiness probe. but it is not in the scope of this PR. start first here means the start is triggered first. added a spawn barrier for now so that at least we wait until all extensions are spawned, but not necessarily completed initialization.

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
gouslu: done, documented, and created a TODO to fix the unhappy path. will create an issue for it as well once reviewed.

rust/otap-dataflow/crates/engine/src/attributes.rs
gouslu: added PipelineAttributeSet to ExtensionScopeAttributeSet.

rust/otap-dataflow/crates/engine/src/extension/wrapper.rs
gouslu: I made shutdown a oneshot channel and is delivered with priority.

rust/otap-dataflow/crates/engine/src/extension_monitor.rs
gouslu: added wait_all_spawned to address this.

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
gouslu: early exit is treated as an error now.

lalitb:
~~One integration question - receivers/processors/exporters already have a static self-registration path through distributed slices, so an out-of-repo component crate can be linked into the final binary and contribute its factory automatically.~~

~~For extensions, this PR adds `ExtensionFactory`, but the factory set appears to be passed into `PipelineFactory::new(...)` as an explicit `&'static [ExtensionFactory]`. I don’t see an equivalent public `OTAP_EXTENSION_FACTORIES` distributed slice yet.~~

~~Is that intentional for this phase? If external linked crates are expected to bring their own extensions, we probably need either the same static registration surface as nodes, or docs that say extension factories must be manually included by the final binary integrator. Otherwise “extensions” may sound pluggable while the current support is really “host extensions whose factories are already wired into this binary.”~~

Update: Please ignore this.  I missed that the PR already wires this for extension. Good work!

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -117,120 +107,273 @@ impl ExtensionLifecycle {
...
+                shutdown_channels,
+                monitor,
+                started_rx,
+                outstanding_spawn_signals,
+                ..
+            } = self;
+            if *outstanding_spawn_signals == 0 {
+                return Ok(());
+            }
+            tokio::select! {
```
lalitb, line 240: I think this barrier can miss an extension that exits immediately after it starts. Since the select handles the `started_rx` signal before checking task completion, it can return `Ok(())` once all “started” signals are seen, even if one of those tasks has already failed or returned. Can we check any ready completions before returning so nodes don’t start after an extension has already died?

lalitb (approved):
LGTM overall. I left one comment about a startup race, but not a blocker.

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -117,120 +107,273 @@ impl ExtensionLifecycle {
...
+                }
+            }
+            tokio::select! {
+                biased;
+                Some(_key) = started_rx.recv() => {
+                    *outstanding_spawn_signals = outstanding_spawn_signals.saturating_sub(1);
+                    continue;
+                }
+                Some(joined) = futures.next() => {
+                    *outstanding_spawn_signals = outstanding_spawn_signals.saturating_sub(1);
```
ThomsonTan: A fast-failing task drives the counter down by 2 instead of 1 (1 on start and 1 on failure). The counter reaches 0 too early, and the barrier thinks "everyone started" when one task hasn't. `saturating_sub` just stops it from going negative — it hides the mistake instead of fixing it.

gouslu: right, will fix with a hashmap

jmacd (approved):

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
```diff
@@ -260,6 +275,29 @@ impl<PData: 'static + Debug + Clone + ReceivedAtNode + Unwindable + FlowMetricHo
...
+        // completion events and drives the monitor's `CollectTelemetry`
+        // fan-out from its own tick cadence.
+        //
+        // Construction enters the runtime so `tokio::time::interval_at`
+        // can attach to the timer driver; the guard is dropped before
+        // `rt.block_on` so the runtime can be re-entered there.
+        let ext_ctx = pipeline_context.extension_context();
+        let ext_monitor = {
+            let _enter = rt.enter();
+            if telemetry_policy.pipeline_metrics {
```
lquerel, line 290: When `pipeline_metrics` is false and the extension monitor is disabled, what keeps active extension control senders alive until shutdown? It looks like the monitor no longer retains the cloned sender, and `ExtensionWrapper::start(self, ...)` consumes the wrapper, so an extension waiting on `ctrl.recv()` may see `Closed` before shutdown is initiated. Should this path have a non-metrics owner for the sender, or a regression test with `pipeline_metrics: false`?

rust/otap-dataflow/crates/engine/src/extension/wrapper.rs
```diff
@@ -435,15 +540,19 @@ impl ExtensionWrapper {
                     ExtensionLifecycle::Active {
                         extension,
                         control_receiver,
+                        shutdown_receiver,
```
lquerel, line 554: Should the `EntityTelemetryGuard` stored on the wrapper live for the full duration of the active extension task? Since this arm does not bind `telemetry`, it looks like the guard may be dropped when `start()` begins, which could unregister extension/control-channel telemetry while the extension is still running. Should the guard be held until the awaited extension `start()` returns?

lquerel (commented):
Overall it's really good. However, I have 2 questions that could be potential issues.

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
gouslu: Good point, ExtensionLifecycle now owns the control_sender and added regression tests.

rust/otap-dataflow/crates/engine/src/extension/wrapper.rs
gouslu: ExtensionLifecycle now takes and holds the guard and drops it after start returns.

lquerel (approved):
LGTM. Thanks, Gokhan, for this additional round in the complicated world of extensions.

PR #3144, opened 2026-05-29
feat(flow_metrics): Add purpose scope attribute to flow_metric config to enable differentiation of flows

drewrelmas:
# Change Summary

(Attempt 2, see #3132 for context. Copilot identified real issue with that simpler implementation)

 ## Add `flow.purpose` scope attribute to flow metrics

 ### What

 Adds an optional `purpose` field to flow-metric configuration (`FlowMetricConfig` in `crates/config/src/policy.rs`). When set, it is emitted as the **`flow.purpose` instrumentation-scope attribute** on every metric the flow produces.

 To do this correctly, this PR adds a small, general mechanism for declaring that an attribute belongs on the OpenTelemetry **instrumentation scope** rather than on each data point:

 - A new `#[attribute(key = "...", scope)]` marker in `telemetry-macros`. The derive records these keys in a new `scope_keys` field on`AttributesDescriptor` (and merges them across `#[compose]`d attribute sets).
 - A descriptor-backed `AttributeSetHandler::is_scope_attribute(key)` helper, so the designation survives type erasure through `EntityAttributeSet`.
 - The metrics dispatcher (`crates/telemetry/src/metrics/dispatcher.rs`) now partitions an entity's attributes into `(scope, data_point)`. Scopeattributes are attached to the meter via `global::meter_with_scope(InstrumentationScope::builder(name).with_attributes(...))`; everything else stays on `counter.add` / `histogram.record` as before.
 - `flow.purpose` (in `FlowAttributeSet`, `crates/engine/src/flow_metrics.rs`) is marked `scope`.

 This is change **A** from #2859 ("per-flow scope"): the smallest, backward-compatible step that lets OTel View selectors distinguish flows thatshare the single `flow` instrumentation scope. It builds on #2755, which enabled `scope_attributes` for Metric View filtering.

 ### Why

 All flow metrics emit under one shared `flow` scope. OTel View selectors match on **scope name + scope attributes + instrument name**, so todaythey cannot target one flow's metrics without also matching every other flow's same-named instrument. As we start declaring flows for different*flavors* of work, their metrics collide in a single View.

 `flow.purpose` gives operators a per-flow differentiator that selectors can key on, **without introducing a new scope per flow**. Crucially,because View selectors match on `instrument.scope().attributes()`, the value must be emitted on the instrumentation scope — emitting it as adata-point attribute (an earlier iteration of this PR) would never match a `scope_attributes` selector.

 ### Backward compatibility

 Flows that do not configure a `purpose` are unchanged: an unset/empty `flow.purpose` is **omitted entirely** (the dispatcher drops empty scopeattributes), so the instrumentation-scope identity for existing flows stays exactly as before. No new attribute or label appears on existingmetrics.

 ### Example — two flows that need to be told apart

 A pipeline declares two flows over different processor stages. Both enable `compute_duration`, so without a differentiator a `compute.duration`View selector would match — and conflate — both:

 ```yaml
 policies:
   telemetry:
     flow_metrics:
       # A sampling stage: records not selected by the sampler are dropped.
       # Operator cares about all flow metrics (duration, incoming, and outgoing).
       - id: sampling
         bounds: { start_node: log_sampler, end_node: log_sampler }
         purpose: filter
         metrics: [compute_duration, signals_incoming, signals_outgoing]

       # A downstream enrichment/transform stage. Operator only cares about how
       # long the transforms takes.
       - id: enrich
         bounds: { start_node: transform_a, end_node: transform_b }
         purpose: transform
         metrics: [compute_duration]
 ```

 Both flows emit under scope `flow`. With `flow.purpose` on the scope, Views route each flow's instruments to distinct streams:

 ```yaml
 views:
   - selector: { scope_name: flow, scope_attributes: { flow.purpose: filter }, instrument_name: compute.duration }
     stream:   { name: sampler_duration }
   - selector: { scope_name: flow, scope_attributes: { flow.purpose: filter }, instrument_name: signals.incoming }
     stream:   { name: sampler_incoming_items }
   - selector: { scope_name: flow, scope_attributes: { flow.purpose: filter }, instrument_name: signals.outgoing }
     stream:   { name: sampled_records_kept }
   - selector: { scope_name: flow, scope_attributes: { flow.purpose: transform }, instrument_name: compute.duration }
     stream:   { name: enrich_duration }
 ```

 Without `flow.purpose`, a `scope_name: flow` + `instrument_name: compute.duration` selector would match **both** flows and report mixed durationdata.

 A runnable demonstration is included at `rust/otap-dataflow/configs/flow-purpose-demo.yaml`.

 * Progress towards #2859

Unit tests

Yes — operators can now specify a flow `purpose`, emitted as the `flow.purpose` scope attribute. Flows without a `purpose` are unaffected.

### Changelog

### DIFF
--- a/rust/otap-dataflow/crates/telemetry-macros/src/lib.rs
+++ b/rust/otap-dataflow/crates/telemetry-macros/src/lib.rs
@@ -325,6 +325,8 @@ pub fn derive_metric_set_handler(input: TokenStream) -> TokenStream {
 ///   - `#[attributes(name = "my.attributes.name")]`
 ///     Field attributes:
 ///   - `#[attribute(key = "field.key")]` (optional, defaults to field name with dots)
+///   - `#[attribute(key = "field.key", scope)]` to emit the attribute on the
+///     OpenTelemetry instrumentation scope instead of on each data point
 ///   - `#[compose]` for fields that implement AttributeSetHandler
 #[proc_macro_derive(AttributeSetHandler, attributes(attributes, attribute, compose))]
 pub fn derive_attribute_set_handler(input: TokenStream) -> TokenStream {
@@ -413,6 +415,7 @@ pub fn derive_attribute_set_handler(input: TokenStream) -> TokenStream {
     let mut attr_field_descriptions = Vec::new();
     let mut attr_field_types: Vec<proc_macro2::TokenStream> = Vec::new();
     let mut attr_iter_values: Vec<proc_macro2::TokenStream> = Vec::new();
+    let mut attr_field_scope_keys: Vec<String> = Vec::new();
 
     let mut composed_field_idents = Vec::new();
 
@@ -478,6 +481,16 @@ pub fn derive_attribute_set_handler(input: TokenStream) -> TokenStream {
         }
         let derived_key = ident.to_string().replace('_', ".");
         let final_key = key_attr.unwrap_or(derived_key);
+
+        // Detect `#[attribute(..., scope)]` to mark this attribute as a
+        // scope-level (instrumentation-scope) attribute.
+        let is_scope_field = field
+            .attrs
+            .iter()
+            .any(parse_attribute_field_is_scope);
+        if is_scope_field {
+            attr_field_scope_keys.push(final_key.clone());
+        }
 
         // Determine attribute value type based on field type
         let (attr_type, iter_value) = match &field.ty {
@@ -555,6 +568,9 @@ pub fn derive_attribute_set_handler(input: TokenStream) -> TokenStream {
 
     let desc_ident = format_ident!("ATTRIBUTES_DESCRIPTOR");
 
+    // Local scope-attribute keys (from `#[attribute(key = "...", scope)]`).
+    let local_scope_keys = &attr_field_scope_keys;
+
     // Generate the descriptor and iterator implementation
     if composed_field_idents.is_empty() {
         // Simple case: no composition - keep the original approach
@@ -570,6 +586,7 @@ pub fn derive_attribute_set_handler(input: TokenStream) -> TokenStream {
                                 r#type: #attr_field_types
                             } ),*
                         ],
+                        scope_keys: &[ #( #local_scope_keys ),* ],
                     };
                     &#desc_ident
                 }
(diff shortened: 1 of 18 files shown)


rust/otap-dataflow/crates/telemetry-macros/src/lib.rs
```diff
@@ -325,6 +325,8 @@ pub fn derive_metric_set_handler(input: TokenStream) -> TokenStream {
 ///   - `#[attributes(name = "my.attributes.name")]`
 ///     Field attributes:
 ///   - `#[attribute(key = "field.key")]` (optional, defaults to field name with dots)
+///   - `#[attribute(key = "field.key", scope)]` to emit the attribute on the
```
cijothomas, line 328: from a quick look, the direction is reasonable. If one were emitting metrics using any OTel sdk, they have the free to create Meter with attributes, and then use that meter to create instrument etc. Since our macros abstract those things, this seems a reasonable thing to let the user convey if they want some attributes to be on scope vs metric.

rust/otap-dataflow/crates/telemetry/src/descriptor.rs
```diff
@@ -106,4 +106,10 @@ pub struct AttributesDescriptor {
     pub name: &'static str,
     /// Ordered attribute field metadata.
     pub fields: &'static [AttributeField],
+    /// Keys of attributes that must be emitted on the OpenTelemetry
+    /// instrumentation scope rather than as data-point attributes. Populated by
+    /// the `#[derive(AttributeSetHandler)]` macro from fields marked
+    /// `#[attribute(key = "...", scope)]` (composed sets contribute their own
+    /// scope keys). Empty when the set has no scope-level attributes.
+    pub scope_keys: &'static [&'static str],
```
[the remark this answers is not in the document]
drewrelmas: Addressed

rust/otap-dataflow/crates/telemetry-macros/src/lib.rs
cijothomas: @drewrelmas I think it is easier to review if we do the change to support scope into its own PR, and then do the flow_metric to adopt it in subsequent PR. it is easier to review them independently :)

drewrelmas: Sure thing, I'll close this one for now and send the smaller versions next week.

PR #3146, opened 2026-05-30
feat:(host metrics receiver) Add opt-in Linux per-process host metrics

### LINKED ISSUE #3066: [Host metrics receiver] Add opt-in Linux per-process metrics
### Objective

  Add opt-in Linux per-process metrics to the OTAP-native `host_metrics` receiver.

  The original host metrics issue (#2741) intentionally kept v1 bounded to aggregate process counts only and listed per-process metrics as future v2 work. This issue tracks that follow-up separately.

  The Go Collector hostmetrics [`process`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/hostmetricsreceiver/internal/scraper/processscraper/documentation.md) scraper emits per-process CPU, memory, disk I/O, threads, uptime, paging faults, context switches, and related process resource attributes. We should not copy that surface blindly, because per-process metrics can be expensive and high-cardinality.

  ### Proposed scope

  Add an explicit opt-in process mode, for example:

  ```yaml
  families:
    processes:
      enabled: true
      mode: summary_and_per_process
      process:
        include:
          match_type: regexp
          names: ["df_engine", "otelcol"]
        max_processes: 100
        labels:
          pid: false
          command: true
          executable_name: true
          executable_path: false
          command_line: false
          owner: false
          cgroup: false
          parent_pid: true
        metrics:
          cpu_time: true
          cpu_utilization: true
          memory_usage: true
          memory_virtual: true
          disk_io: true
          threads: true
          uptime: true

```

  ### Design requirements

  - Per-process collection must be disabled by default.
  - Cardinality controls are required before enabling this feature.
  - Track processes by (pid, start_time) so CPU utilization does not mix reused PIDs.
  - Respect host_view.root_path.
  - Handle permission failures without noisy logs.
  - Avoid blocking the receiver indefinitely while walking /proc.
  - Keep aggregate mode: summary behavior unchanged.

  ### Initial metrics

  Start with a small Linux subset:

  - process.cpu.time
  - process.cpu.utilization
  - process.memory.usage
  - process.memory.virtual
  - process.disk.io
  - process.threads
  - process.uptime

  Optional follow-up metrics can include context switches, open file descriptors, paging faults, cgroup, owner, executable path, and command line.

  ### Acceptance criteria

  - Existing aggregate process count behavior is unchanged by default.
  - Per-process metrics are emitted only when explicitly configured.
  - Include/exclude filters work.
  - max_processes or equivalent cardinality guard is enforced.
  - PID reuse is handled safely.
  - Permission errors are counted/rate-limited, not logged noisily every scrape.
  - Parser and projection tests cover the emitted process metric shape.

### PR DESCRIPTION
**Pull request by lalitb**
# Change Summary

Adds opt-in Linux per-process metrics to the OTAP-native `host_metrics` receiver.

Per-process collection is disabled by default and only runs when `processes.mode: summary_and_per_process` is configured. The existing aggregate process summary behavior remains unchanged.

The implementation adds bounded per-process CPU, memory, disk I/O, thread count, and uptime metrics with include/exclude filters, `max_processes`, PID reuse protection via `(pid, start_time)`, `host_view.root_path` support, and quiet handling for expected `/proc/<pid>` race/permission failures.

* Closes #3066

* `cargo fmt --all`
* `cargo check -p otap-df-core-nodes --all-features`
* `cargo test -p otap-df-core-nodes --all-features receivers::host_metrics_receiver`
* `cargo clippy -p otap-df-core-nodes --all-targets --all-features -- -D warnings`
* `npx markdownlint-cli docs/host-metrics-receiver.md`
* `python3 ../../tools/sanitycheck.py`
* Linux VM validation:
    * `cargo test -p otap-df-core-nodes --all-features receivers::host_metrics_receiver`
    * `cargo clippy -p otap-df-core-nodes --all-targets --all-features -- -D warnings`

Yes. Users can now opt in to Linux per-process host metrics through the `processes` family configuration.

 Example:

  ```yaml
  families:
    processes:
      enabled: true
      mode: summary_and_per_process
      process:
        include:
          match_type: strict
          names: ["df_engine", "otelcol"]
        max_processes: 100
        labels:
          pid: true
          command: true
          executable_name: true
          parent_pid: true
        metrics:
          cpu_time: true
          cpu_utilization: true
          memory_usage: true
          memory_virtual: true
          disk_io: true
          threads: true
          uptime: true
```

### Changelog

* [x] Added a `.chloggen/*.yaml` entry,

### Screenshot

host-process-metrics-grafana-latest


lalitb:
@lquerel - Can you have a look into the PR when possible. thanks.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/procfs/mod.rs
```diff
@@ -504,6 +578,203 @@ impl ProcfsSource {
...
     }
 
+    fn read_processes(
+        &mut self,
+        host_start_time_unix_nano: u64,
+        now_unix_nano: u64,
+        partial_errors: &mut u64,
+        first_error: &mut Option<io::Error>,
+    ) -> Vec<ProcessMetrics> {
+        let mut processes = Vec::new();
```
lquerel: [suggestion] To avoid creating this unbounded `Vec`, you could use a bounded top-N selection structure based on the conditions used in your `sort`.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/procfs/readings.rs
```diff
@@ -783,6 +892,22 @@ pub(super) fn clock_ticks_per_second() -> f64 {
     if ticks > 0 { ticks as f64 } else { 100.0 }
 }
 
+pub(super) fn available_process_cpu_count() -> f64 {
+    std::thread::available_parallelism()
+        .map(|count| count.get() as f64)
+        .unwrap_or(1.0)
+}
```
lquerel: This uses `std::thread::available_parallelism()` as the denominator for `process.cpu.utilization`, but the receiver can scrape a different host view through `host_view.root_path`. In containerized host scraping, this may return the collector/container CPU availability instead of the scraped host's logical CPU count, making utilization wrong. Might be better to derive the CPU count from the same procfs/cpuinfo host view used for the scrape.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/config.rs
```diff
@@ -383,6 +414,90 @@ pub enum ProcessMode {
...
+    /// Emit process.cgroup.
+    pub cgroup: bool,
+    /// Emit process.parent_pid.
+    pub parent_pid: bool,
+}
+
+impl Default for ProcessLabelsConfig {
+    fn default() -> Self {
+        Self {
+            pid: false,
```
lquerel: With `process.pid` disabled by default, multiple same-command processes can emit indistinguishable per-process metric series. For example, two `otelcol` processes with the same parent would have the same default emitted labels, while the receiver still tracks counters internally by PID/start time.
You could make a stable process identity non-optional, or otherwise aggregate/dedupe by the emitted label set so exported series cannot collide.

lquerel (approved):
Looks great!

I made a few comments that I think are important to take into account. I'm approving the PR because I know you'll address them.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/procfs/mod.rs
lalitb: Good point. Have now changed this to keep a bounded top-N heap while scanning `/proc`, using the same CPU time, memory, and PID ordering as before. The final output is still sorted the same way, but we no longer keep every matching process in memory before truncating. - https://github.com/open-telemetry/otel-arrow/pull/3146/commits/7100ca0a830bd15796b1ba631eb64375cd9b9a3e

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/procfs/readings.rs
lalitb: Good catch. Have updated now so the per-process CPU utilization denominator comes from the scraped host view’s `/proc/cpuinfo` instead of `std::thread::available_parallelism()` -   https://github.com/open-telemetry/otel-arrow/pull/3146/commits/7100ca0a830bd15796b1ba631eb64375cd9b9a3e

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/config.rs
lalitb: Agreed. I changed `process.pid` to be enabled by default and made it required when `processes.mode=summary_and_per_process`. That keeps exported per-process series distinguishable instead of only tracking uniqueness internally by PID/start time - https://github.com/open-telemetry/otel-arrow/pull/3146/commits/7100ca0a830bd15796b1ba631eb64375cd9b9a3e

lalitb:
> Looks great!
> 
> I made a few comments that I think are important to take into account. I'm approving the PR because I know you'll address them.

Thanks @lquerel. These were very good catches. I’ve addressed them now.

rust/otap-dataflow/crates/core-nodes/src/receivers/host_metrics_receiver/config.rs
```diff
@@ -383,6 +414,90 @@ pub enum ProcessMode {
...
+        }
+    }
+}
+
+/// Per-process metric controls.
+#[derive(Clone, Copy, Debug, Deserialize, Serialize)]
+#[serde(default, deny_unknown_fields)]
+pub struct ProcessMetricsConfig {
+    /// Emit process.cpu.time.
+    pub cpu_time: bool,
```
drewrelmas, line 474: @lalitb not necessarily a blocker on this PR, but wondering if we should align on user-facing config strategy for optional metrics produced by different components. I raise this as the approach `flow_metrics` currently uses is an opt-in list instead of individual per-metric bools: https://github.com/open-telemetry/otel-arrow/blob/e72e1aabf49d70f5aaa11f0c8117b28864685453/rust/otap-dataflow/configs/fake-flow-metrics-demo.yaml#L53

```yaml
flow_metrics:
  - id: all_metrics
    bounds:
      start_node: node1
      end_node: node2
  - id: only_signals_outgoing
    bounds:
      start_node: node3
      end_node: node4
    metrics: [signals_outgoing]
```

lalitb: Good point. There are already other optional metric configs, both inside host metrics and outside this PR, that don’t follow the `flow_metrics` list pattern - I’ll handle this as a separate follow-up for host metrics.

Issue #3071, opened 2026-05-22
[User events receiver]  Add per-tracepoint throttling/governance


lalitb:
# Description

The Linux `user_events` receiver can subscribe to multiple tracepoints in one receiver/session. In the current one_collect/perf setup, each tracepoint has its own perf event fd, but those events are redirected into the same per-CPU session output ring/reader.

That means tracepoints in the same receiver session share the same receive path, drain limits, batching, downstream backpressure boundary, and receiver-level metrics.

This can create shared-fate behavior: a high-volume tracepoint can consume shared capacity and impact quieter tracepoints in the same session.

It would be useful to add per-tracepoint governance while keeping the shared receiver/session model. For example:
  - configurable event or byte limits per tracepoint
  - per-tracepoint throttled/dropped metrics
  - clear behavior when one tracepoint exceeds its limit
  - enforcement as early as practical in the receive path

This would give better fairness without requiring a separate receiver/session for every tracepoint.

## Alternatives Considered

Use separate receiver/session instances for tracepoints that need isolation. This provides stronger receiver-level isolation and avoids shared receiver budgets, but it duplicates perf session resources and output/read-side buffering. That may be acceptable for a small number of tracepoints, but can become expensive when many tracepoints need independent limits.

## Additional Context

  Roughly, the current shape is:

  ```text
  CPU K
    |
    v
  one receiver/session
    |
    +-- tracepoint A perf event fd --\
    +-- tracepoint B perf event fd ----> shared per-CPU output/read path
    +-- tracepoint C perf event fd --/
                                        |
                                        v
                             shared drain/batch/downstream budgets

```

So the goal is not to create a separate receiver/session for every tracepoint, but to add per-tracepoint limits before one noisy tracepoint consumes the shared receiver capacity.

hestolz:
To add some additional motivation / context, the multi-session approach provides isolation between would-be "noisy neighbor" tracepoints, but also risks oversizing the fixed perf ringbuffers. E.g. if you know the incoming data has high burst characteristics, you should ensure the ringbuffer is sized for that peak load; when multiple tracepoints with potentially correlated peaks share a session, that compounds the problem. Another contributing factor is the current otap-dataflow model where the one_collect/perf reader shares the same thread as all the other pipeline nodes on that CPU; I imagine that gives more opportunity for the reads to be delayed by other work versus having a dedicated thread.

[...]

PR #3153, opened 2026-06-01
feat(Linux user_events receiver) - Add per-subscription pending governance 

lalitb:
# Change Summary

Adds Phase 1 per-subscription pending governance for the Linux `user_events` receiver.

This introduces optional per-subscription pending limits:

  - `subscriptions[].limits.max_pending_events`
  - `subscriptions[].limits.max_pending_bytes`

When no effective subscription limits are configured, the receiver keeps the existing shared FIFO pending queue behavior. When any subscription has effective limits, the receiver uses per-subscription pending queues with round-robin drain under the existing global `session.max_pending_events` and `session.max_pending_bytes` ceilings.

Admission is enforced in the `one_collect` callback before `event_data().to_vec()`, so rejected samples avoid payload heap allocation.

This improves receiver-side fairness for pending queue capacity and drain order. It does not provide kernel-ring isolation: perf-ring loss and one_collect parse work remain shared by tracepoints in the same receiver/session.

  * Addresses receiver-side pending governance for #3071

  - `cargo fmt --all`
  - `git diff --check`
  - `cargo check -p otap-df-contrib-nodes --features user_events-receiver,otap-df-otap/crypto-ring --all-targets`
  - `cargo clippy -p otap-df-contrib-nodes --features user_events-receiver,otap-df-otap/crypto-ring --all-targets -- -D warnings`
  - Linux VM:
    - user_events adapter tests: 13 passed
    - user_events config validation subset: 6 passed
    - manual noisy/quiet tracepoint test confirmed quiet records pass with noisy tracepoint pending limits enabled

Yes. The Linux `user_events` receiver now supports optional per-subscription pending limits:

  ```yaml
  subscriptions:
    - tracepoint: "user_events:noisy_tracepoint"
      format:
        type: tracefs
      limits:
        max_pending_events: 1024
        max_pending_bytes: 4194304
```

When subscriptions[].limits is present, at least one pending cap must be set and configured cap values must be greater than zero.

Phase 1 keeps dropped_pending_overflow aggregate-only. Per-tracepoint/per-cap drop attribution is left for a follow-up.

### Changelog

### CODE REVIEW

**Comment by lalitb** [comment]
@hestolz -  tagging you in case you would like to review this against the requirements we discussed.

File: rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md
Comment (hestolz) [inline rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md]: 1. should `session.max_pending_*` move to `session.limits.max_pending_*` to mimic the new options?
2. a future QoL suggestion: something like `session.limits.apply_equally` (bad name but you get the idea) where rather than cx doing the math to break up the max_pending_events per subscription, we just give them even weight within the configured session-wide limit, and apply round-robin drain

File: rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md [hunk 1]
```diff
@@ -298,6 +316,8 @@ tracepoint sessions.
 | --- | --- | --- |
 | `subscriptions` | none | Required non-empty list of `user_events` tracepoints. Each entry may be either `<event>` or `user_events:<event>`. |
 | `subscriptions[].format.type` | `tracefs` | Decode format for one subscription. `tracefs` is always supported; `event_header` requires the `user_events-eventheader` feature. |
+| `subscriptions[].limits.max_pending_events` | none | Optional per-subscription pending event cap. When `limits` is present, at least one pending cap must be set, and configured cap values must be greater than zero. |
+| `subscriptions[].limits.max_pending_bytes` | none | Optional per-subscription pending raw payload byte cap. Counts raw `event_data` bytes, matching `session.max_pending_bytes`. When any subscription has effective limits, per-subscription queues and round-robin drain are enabled under the global session caps. |
```
hestolz:
```suggestion
| `subscriptions[].limits.max_pending_bytes` | none | Optional per-subscription pending raw payload byte cap. Counts raw `event_data` bytes, matching `session.max_pending_bytes`. When any subscription has effective limits (byte- and/or event-based), per-subscription queues and round-robin drain are enabled under the global session caps. |
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/mod.rs
```diff
@@ -320,6 +346,37 @@ impl UserEventsReceiver {
         Ok(())
     }
 
+    fn validate_subscription_limits(
```
hestolz, line 370: is there any valid reason the subscription-specific limits should be allowed to sum to greater than the session limits? if not, let's at least warn if the receiver has been configured that way, or even error out (or normalize per subscription so they total appropriately?)

gouslu:
This is a net win to avoid starvation of low-volume subs if user configures this correctly. Adding a hard-enforced min, with max as a ceiling bounded by what's physically reachable on top of min, would give users who configure this a good safety net that is worth adding in my opinion.

- sum(min) <= global_max
- shared_pool = global_max - sum(min)
- per sub: max <= min + shared_pool (so max stays within what's physically reachable)

Example:

sub a: min=100  max=500
sub b: min=200  max=300
sub c: min=50   max=700
global_max: 1000

shared_pool = 1000 - 350 = 650
effective ceiling per sub = own min + shared_pool
  a: 100 + 650 = 750  (configured 500, ok)
  b: 200 + 650 = 850  (configured 300, ok)
  c:  50 + 650 = 700  (configured 700, at ceiling)

Should we consider adding a min per sub (either with this or a future PR) that defaults to 0 so that users can have a safety net when they configure this receiver?

edit: I think there should be an option to configure per sub min and a global min per sub, where both default to 0 and per sub min takes precedence.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/mod.rs
gouslu, line 370: I think we need to validate per sub max doesn't exceed global max.

gouslu: I think that's a valid pattern to overcommit the limits. Each sub has a theoretical upper bound, but they can't all reach that value at the same time. How allocation is managed under contention is the more important aspect, that's where starvation can happen. This is also one of the reasons why I think we should also have a min limit where sum of all mins is strictly enforced to stay under global max.

ThomsonTan:
Sounds like this addresses part of #3071 instead of closing it, because per-tracepoint "dropped" metrics are not in this PR.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/README.md
lalitb: Good point -  I moved the session-wide pending caps under `session.limits.*` so they line up with the per-subscription `subscriptions[].limits.*` shape.

The automatic fair-share mode also makes sense, but I’d keep that as future work. This PR already enables per-subscription caps plus round-robin drain when limits are configured; a session-level “split this equally across subscriptions” option would be a useful QoL follow-up once we settle the exact config name and semantics.

lalitb:
> This is a net win to avoid starvation of low-volume subs if user configures this correctly. Adding a hard-enforced min, with max as a ceiling bounded by what's physically reachable on top of min, would give users who configure this a good safety net that is worth adding in my opinion.
> 
> * sum(min) <= global_max
> * shared_pool = global_max - sum(min)
> * per sub: max <= min + shared_pool (so max stays within what's physically reachable)
> 
> Example:
> 
> sub a: min=100 max=500 sub b: min=200 max=300 sub c: min=50 max=700 global_max: 1000
> 
> shared_pool = 1000 - 350 = 650 effective ceiling per sub = own min + shared_pool a: 100 + 650 = 750 (configured 500, ok) b: 200 + 650 = 850 (configured 300, ok) c: 50 + 650 = 700 (configured 700, at ceiling)
> 
> Should we consider adding a min per sub (either with this or a future PR) that defaults to 0 so that users can have a safety net when they configure this receiver?
> 
> edit: I think there should be an option to configure per sub min and a global min per sub, where both default to 0 and per sub min takes precedence.

Thanks. That model makes sense, but I’d prefer to keep `min_pending_*` guarantees as a follow-up. This PR only adds max caps plus round-robin drain; adding reserved minimums changes admission semantics and needs separate validation like `sum(min) <= session.limits.*`. Will create a follow-up issue to track this.

lalitb:
> Sounds like this addresses part of #3071 instead of closing it, because per-tracepoint "dropped" metrics are not in this PR.

Yes, that’s fair. This PR addresses the receiver-side pending governance part of #3071, but per-tracepoint dropped metrics are still future work, so I’ll avoid marking the issue as fully closed by this PR. Have also updated the PR desc to reflect that.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/user_events_receiver/mod.rs
lalitb: I think overcommit is valid here because `session.limits.*` remains the hard aggregate ceiling. I added a `README` clarification that per-subscription caps may intentionally overcommit the session-wide caps, and that these caps do not reserve minimum capacity for a subscription.

lalitb: This is intentional. The per-subscription values are ceilings, not reservations, and `session.limits.*` remains the hard aggregate cap. So a subscription limit greater than the session limit cannot exceed the global queue bound; it just means that subscription is constrained by the session cap first. I added README wording to make that overcommit behavior explicit.

utpilla (approved):

PR #3154, opened 2026-06-01
feat(observability): Add a general mechanism for defining instrumentation-scope metric attributes

drewrelmas:
# Change Summary

## Add a general mechanism for instrumentation-scope attributes

 Plumbing-only half of a two-part split (from #3144); a follow-up PR wires it into `flow_metrics`.

 ### What

- New **set-level** `#[attribute_set(name = "...", scope)]` marker in `telemetry-macros`. Marking a set `scope`designates **every attribute that set directly declares** as an instrumentation-scope attribute. The derive records thosekeys in a new `scope_keys` field on `AttributesDescriptor` (merged, sorted, and de-duplicated across `#[compose]`d sets).Composed children are governed by their own marker, not the parent's.
 - `AttributeSetHandler::is_scope_attribute(key)`, descriptor-backed so it survives type erasure through `&dynAttributeSetHandler` and `EntityAttributeSet`.
 - The metrics dispatcher partitions attributes into `(scope, data_point)`. Scope attributes attach to the meter via`meter_with_scope(...)`; the rest stay on `counter.add` / `histogram.record`. Empty-string scope values are omitted sothey don't perturb scope identity.
 - `EntityAttributeSet` equality/hash now include `scope_keys` (compared order-independently), so sets differing only inscope designation aren't merged during registry de-dup.

 No set is marked `scope` yet, so this is behavior-preserving. Other file changes are mechanical (`scope_keys: &[]`).

 ### Why

 OTel View selectors match on scope name + scope attributes + instrument name (see https://github.com/open-telemetry/otel-arrow/pull/2755). This lets chosen attributes be emitted on the scope rather than per data point. Split out from the `flow.purpose` work so the reusable mechanism is reviewable on its own.

### Example

```rust
```rust
 /// Scope-level attributes, lifted onto the OpenTelemetry instrumentation
 /// scope instead of each data point. The set-level `scope` flag marks every
 /// field it declares as a scope attribute.
 #[attribute_set(name = "node.scope", scope)]
 #[derive(Debug, Clone, Default, Hash)]
 pub struct NodeScopeAttributes {
     /// Scope attribute: bounded, known at node creation. OTel View selectors
     /// keyed on `scope_attributes` can match it (e.g. all `exporter` nodes).
     /// Omitted entirely when empty.
     #[attribute(key = "node.kind")]
     pub kind: Cow<'static, str>, // "receiver" | "processor" | "exporter"
 }

 /// A pipeline node's metric attributes. `node.name` stays a data-point
 /// attribute; the composed scope set contributes `node.kind` on the scope.
 #[attribute_set(name = "node.attrs")]
 #[derive(Debug, Clone, Default, Hash)]
 pub struct NodeAttributeSet {
     /// A regular data-point attribute (emitted on each measurement).
     #[attribute(key = "node.name")]
     pub name: Cow<'static, str>,

     #[compose]
     pub scope: NodeScopeAttributes,
 }
```

What the derive produces from this — a descriptor whose `scope_keys` captures the composed scope set's keys:

```rust
// Generated (illustrative):
 AttributesDescriptor {
     name: "node.attrs",
     fields: &[
         AttributeField { key: "node.name", r#type: AttributeValueType::String, /* ... */ },
         AttributeField { key: "node.kind", r#type: AttributeValueType::String, /* ... */ },
     ],
     scope_keys: &["node.kind"], // <-- drives is_scope_attribute(...)
 };
```

* Progress towards #2859 

Unit tests

Developers may now emit scope-level metric attributes

### Changelog

### DIFF
--- a/rust/otap-dataflow/crates/telemetry-macros/src/lib.rs
+++ b/rust/otap-dataflow/crates/telemetry-macros/src/lib.rs
@@ -325,6 +325,8 @@ pub fn derive_metric_set_handler(input: TokenStream) -> TokenStream {
 ///   - `#[attributes(name = "my.attributes.name")]`
 ///     Field attributes:
 ///   - `#[attribute(key = "field.key")]` (optional, defaults to field name with dots)
+///   - `#[attribute(key = "field.key", scope)]` to emit the attribute on the
+///     OpenTelemetry instrumentation scope instead of on each data point
 ///   - `#[compose]` for fields that implement AttributeSetHandler
 #[proc_macro_derive(AttributeSetHandler, attributes(attributes, attribute, compose))]
 pub fn derive_attribute_set_handler(input: TokenStream) -> TokenStream {
@@ -413,6 +415,7 @@ pub fn derive_attribute_set_handler(input: TokenStream) -> TokenStream {
     let mut attr_field_descriptions = Vec::new();
     let mut attr_field_types: Vec<proc_macro2::TokenStream> = Vec::new();
     let mut attr_iter_values: Vec<proc_macro2::TokenStream> = Vec::new();
+    let mut attr_field_scope_keys: Vec<String> = Vec::new();
 
     let mut composed_field_idents = Vec::new();
 
@@ -478,6 +481,13 @@ pub fn derive_attribute_set_handler(input: TokenStream) -> TokenStream {
         }
         let derived_key = ident.to_string().replace('_', ".");
         let final_key = key_attr.unwrap_or(derived_key);
+
+        // Detect `#[attribute(..., scope)]` to mark this attribute as a
+        // scope-level (instrumentation-scope) attribute.
+        let is_scope_field = field.attrs.iter().any(parse_attribute_field_is_scope);
+        if is_scope_field {
+            attr_field_scope_keys.push(final_key.clone());
+        }
 
         // Determine attribute value type based on field type
         let (attr_type, iter_value) = match &field.ty {
@@ -555,6 +565,9 @@ pub fn derive_attribute_set_handler(input: TokenStream) -> TokenStream {
 
     let desc_ident = format_ident!("ATTRIBUTES_DESCRIPTOR");
 
+    // Local scope-attribute keys (from `#[attribute(key = "...", scope)]`).
+    let local_scope_keys = &attr_field_scope_keys;
+
     // Generate the descriptor and iterator implementation
     if composed_field_idents.is_empty() {
         // Simple case: no composition - keep the original approach
@@ -570,6 +583,7 @@ pub fn derive_attribute_set_handler(input: TokenStream) -> TokenStream {
                                 r#type: #attr_field_types
                             } ),*
                         ],
+                        scope_keys: &[ #( #local_scope_keys ),* ],
                     };
                     &#desc_ident
                 }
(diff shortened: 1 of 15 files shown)


jmacd:
Compare and contrast:
- The proposal in https://github.com/open-telemetry/otel-arrow/pull/2623 was roughly that we begin using code generation tools instead of macros to declare instrumentation schema
- I feel mildly confused that we're splitting attributes into two categories: my mental model for a scope in the code is an EntityKey. I sort of expect to see a new entity defined for each scope in the output, instead we're mixing scope and metric attributes.

During research for same PR linked above, I explored how to convert metric/attribute sets directly into OTAP. It is easiest to generate one scope per combined metric/attribute set, and that makes me question was is being gained here, aside from a pathway to the OTel SDK view feature. I would prefer to see us implement the code generation and implement the views feature ourselves.

rust/otap-dataflow/crates/telemetry-macros/src/lib.rs
```diff
@@ -325,6 +325,8 @@ pub fn derive_metric_set_handler(input: TokenStream) -> TokenStream {
 ///   - `#[attributes(name = "my.attributes.name")]`
 ///     Field attributes:
 ///   - `#[attribute(key = "field.key")]` (optional, defaults to field name with dots)
+///   - `#[attribute(key = "field.key", scope)]` to emit the attribute on the
```
cijothomas: I think we should give some guidance to users about when to use which, in the context of df_engine.
1. When to go for scope vs metric point. Given we rely 100% on bound instruments in the engine, there shouldn't be any perf or cardinality issues, so the motivation should be based on something else. (the view based filtering in flow_metric case).
2. Moving an attribute from metric point to scope, or vice-versa must be documented as breaking change.
3. https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/telemetry/attributes-guide.md#2-entity-attributes This is not stable guidance , but it says scope is to be used for entity attributes. Might be a separate follow up.

lquerel:
@drewrelmas @jmacd @cijothomas - Shouldn't we invert the logic?

I think that in our model, most attributes within a metric set identify the entity they are attached to. So in my view, most of these attributes are actually scope-level attributes. An attribute that is not scope-level is therefore somewhat of an exception in our model.

I would therefore suggest introducing a new macro to identify these non-scoped attributes, which would also significantly reduce the size of the upcoming PR.

For ref:
- https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/telemetry/attributes-guide.md#2-entity-attributes
- https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/telemetry/entity-model.md

rust/otap-dataflow/crates/telemetry-macros/src/lib.rs
drewrelmas: I did add some docs info about when to choose scope vs datapoint as well as a note about breaking change.

Need to think a little more about the `entity-attributes` guidance, Laurent mentioned something similar below.

drewrelmas:
@jmacd, @cijothomas, and @lquerel, thanks for looking at this. I think these comments converge on similar point, so let me respond in one place.

@lquerel's suggestion of inverting the logic makes sense to me, we can flip to use an opt-out marker. I agree with the direction that all entity attributes should be `scope`-level by default, but I'm wondering what we can practically do in-code to help enforce that practice.

I don't believe this is diverging from the future code generation goal, it could produce the same `descriptor` metadata. This is purely an interim bridge to the SDK we have currently.

I can either do the inversion in this PR or as a follow-up depending if we agree on the direction!

EDIT: I think this should happen in a separate PR because we should also look at all existing `attribute_sets` in the repo and triage what fields are `entity` level vs not.

EDIT 2: Based on a brief look, it is very possible we should **only** be using scope-level attributes via `attribute_sets`, not even allowing datapoint-level.

cijothomas:
+1 to keeping the attributes at scope-level. They are the natural fit. 
(We'll need to modify the admin/prometheus output so that they'll emit these as scope (or keep a backcompat flag to continue emitting as metric point level) - perf tests heavily rely on this.

drewrelmas:
Closing to pursue a different direction, likely around making the default behavior of `attribute_set` fields landing on `scope` attributes instead of signal attributes.

PR #3168, opened 2026-06-02
fix(observability): Emit resource/entity metric attributes at proper OpenTelemetry layer

### LINKED ISSUE #3161: Self-telemetry attributes emitted as metric data-point attributes instead of Resource/Scope layers
### Pre-filing checklist

### Component(s)

### Bug Description

Self-telemetry attributes are emitted on the wrong OpenTelemetry layer. All attributes — including process/host **resource** identity (`host_id`, `container_id`, `process_instance_id`) and **entity/scope** identity (pipeline, node,channel, flow) — are attached as **data-point attributes** rather than being split across the OTel Resource and Instrumentation Scope layers as the data model and the OTel/Prometheus compatibility spec require.

Specifically:
 - The SDK metrics dispatcher uses `global::meter(name)` and attaches every attribute to each data point; nothing isplaced on the instrumentation scope.
 - The admin Prometheus endpoint lifts only the scope *name* (`otel_scope_name`) but emits every attribute as a bareinline per-metric label. The `target_info` gauge exists but is fed solely from the (empty-by-default) `telemetry.resource` config map, so process/host identity is never emitted at the Resource layer.

 The auto-detected resource values (`host.id`, `container.id`, a unique process instance id) are structurally`#[compose]`d into the entity attribute sets (via `ResourceAttributeSet` → `EngineAttributeSet`), so they are flattenedonto every metric and are indistinguishable from entity attributes at emission time.

### Steps to Reproduce

1. Build and run the engine with any pipeline with custom resource fields provided:

```yaml
version: otel_dataflow/v1
engine:
  http_admin:
    bind_address: "127.0.0.1:8080"
  telemetry:
    resource:
      service.name: "resource-demo"
      deployment.environment.name: "dev-proof"
    metrics:
      readers:
        - periodic:
            interval: "5s"
            exporter:
              type: console
groups:
  demo:
    pipelines:
      main:
        nodes:
          gen:
            config:
              traffic_config:
                max_batch_size: 10
                signals_per_second: 10
                log_weight: 100
          sink:
            type: exporter:noop
            config: {}
        connections:
          - from: [ gen ]
            to: [ sink ]
```

2. Scrape the admin endpoint and compare `target_info` against one pipeline metric:
```bash 
curl -s 'http://localhost:8080/api/v1/telemetry/metrics?format=prometheus&reset=false' | grep -E '^(target_info|uptime)\{'
```
   
Observed output:

> target_info{deployment_environment_name="dev-proof",service_name="resource-demo"} 1
uptime{otel_scope_name="pipeline",deployment_generation="0",host_id="<computer-name>",numa_node_id="0",pipeline_id="main",container_id="",pipeline_group_id="demo",core_id="0",process_instance_id="AGPIKQMXQR2KFMZJBQFJRMWSBE"} 4.001954808 1780352138700

Two problems are visible on the uptime line:

- Resource attrs are on the `metric` `datapoint` instead of the resource `target_info`: `host_id`, `process_instance_id`, and `container_id` are inline labels.
  - See the declaration of `resource.attrs` `attribute_set` here: 
    https://github.com/open-telemetry/otel-arrow/blob/d4afab713cf7d661082a82a43319f134c8947cd9/rust/otap-dataflow/crates/engine/src/attributes.rs#L47-L60 
- Entity attrs not scoped: `pipeline_id`, node/channel etc. are bare
 labels rather than `otel_scope_*` labels (only `otel_scope_name` is lifted).

The same thing may also be confirmed via the console exporter configured in the sample pipeline above:

```log
2026-06-01T22:25:11.883Z  INFO  otap-df-controller::pipeline.core_allocation: [pipeline_group_id=demo, pipeline_id=main, num_cores=1, core_allocation=[1 cores]]
2026-06-01T22:25:11.886Z  INFO  otap-df-admin::endpoint.start: Admin HTTP server listening [bind_address=127.0.0.1:8080]
Metrics
Resource
         ->  service.name=String(Owned("resource-demo"))
         ->  deployment.environment.name=String(Owned("dev-proof"))
        Instrumentation Scope #0
                Name         : engine
Metric #0
                Name         : memory.rss
                Description  : Process-wide Resident Set Size — physical RAM currently used by the process. Matches what external tools report (e.g. `kubectl top pod`, `htop`, `ps rss`).
                Unit         : {By}
                Type         : Gauge
                Gauge DataPoints
                StartTime    : 2026-06-01 22:25:12.887116
                EndTime      : 2026-06-01 22:25:16.882209
                DataPoint #0
                        Value        : 86581248
                        Attributes   :
                                 ->  process.instance.id: AGPIKST4ZFZFDGUAJZ4UMZX5JY
                                 ->  host.id: <computer name>
                                 ->  container.id:
```

### Expected Behavior

Attributes should be placed on their correct OTel layer:

 - **Resource** (`host.id`, `container.id`, `service.instance.id`) → emitted on the SDK `Resource` (OTLP) and populated `target_info` gauge (Prometheus), once per process.
 - **Scope** (pipeline / node / channel / flow identity, plus `core_id`, `numa_node_id`, `deployment_generation`,`pipeline_group_id`) → emitted on the instrumentation scope (OTLP) and as `otel_scope_*` labels (Prometheus).
 - **Data point** → only genuine per-measurement dimensions (none exist today, so data points should carry no attributes).

 So each metric line should carry `otel_scope_*` labels for entity identity only, and a single `target_info{host_id=...,container_id=...,service_instance_id=...} 1` line should carry resource identity.

### Actual Behavior

Every attribute — resource and entity alike — is emitted as a data-point attribute. On the admin Prometheus endpoint these appear as bare inline labels on every metric line; `target_info` is empty/absent because the resource config map is empty by default. Resource identity is therefore smeared across every metric series instead of being emitted once at the resource layer, and no attributes are placed on the instrumentation scope.

### Additional Context

See https://github.com/open-telemetry/otel-arrow/pull/3154 for additional context. Specifically, @lquerel's comment at https://github.com/open-telemetry/otel-arrow/pull/3154#issuecomment-4595192645 referencing the `otap-dataflow` telemetry docs.

drewrelmas:
# Change Summary

### Problem
 All self-telemetry attributes were emitted as **data-point attributes**, regardless of whether theydescribed the process/host (Resource), the in-process entity (Scope), or the individual measurement (Datapoint). This produced high-cardinality per-series labels, prevented `target_info`-based joins,and did not follow the OpenTelemetry data model.

### Change
 Each attribute is now emitted on its correct OpenTelemetry layer, for **both metrics and logs**:
 | Category | Source | OTLP export (metrics + logs) | Admin Prometheus |
 | --- | --- | --- | --- |
 | **Resource** | process/host identity: `host.id`, `container.id`, `service.instance.id`(auto-detected) + config `engine.telemetry.resource` | `Resource` on `ResourceMetrics` /`ResourceLogs` | `target_info{...}` gauge |
 | **Scope** | entity identity: pipeline/node/channel/flow + `core.id`, `numa.node.id`,`deployment.generation` | `InstrumentationScope.attributes` | `otel_scope_*` labels |
 | **Datapoint** | none today | data-point / log-record attributes | inline labels |

 Key points:
 - Process/host resource attributes are auto-detected and injected into the resolved `engine.telemetry.resource` map; **config-provided keys take precedence**, empty values are omitted.
 - The **same** resolved resource map feeds the SDK metrics `Resource`, the OTLP log `Resource` bytes, and the admin `target_info` gauge — so resource identity is consistent across metrics, logs, and theadmin endpoint.
 - Entity attributes are rendered as `otel_scope_*` labels per the OTel→Prometheus spec; reserved keys (`otel_scope_name`/`version`/`schema_url`) are never overridden.
 - Engine-global metrics carry **no** scope attributes — identity comes entirely from the Resource layer.

### Validation

Running the same test pipeline described in #3161, we see the following from Prometheus:

```log
target_info{host_id="<computer name>",service_instance_id="AGPIRRISSR6UDHSZJSAHIKJV3U",service_name="resource-demo",deployment_environment_name="dev-proof"} 1
cpu_utilization{otel_scope_name="pipeline",otel_scope_pipeline_group_id="demo",otel_scope_core_id="0",otel_scope_pipeline_id="main",otel_scope_deployment_generation="0",otel_scope_numa_node_id="0"}0.00135 
```

The Console OTLP exporter shows similar results:

```log
Resource
     ->  service.instance.id=String(Owned("AGPIRRISSR6UDHSZJSAHIKJV3U"))
     ->  deployment.environment.name=String(Owned("dev-proof"))
     ->  service.name=String(Owned("resource-demo"))
     ->  host.id=String(Owned("<computer name>"))
...
Instrumentation Scope #0
     Name         : channel.sender
     Scope Attributes:
          ->  channel.id: gen:control
          ->  node.id: gen
          ->  node.urn: urn:otel:receiver:traffic_generator
          ->  node.type: receiver
          ->  pipeline.id: main
          ->  pipeline.group.id: demo
          ->  deployment.generation: 0
          ->  core.id: 0
          ->  numa.node.id: 0
 Metric #0
     Name         : send.count
     ...
     DataPoint #0
         Value        : 1
         Attributes   :          # <- empty: datapoint layer carries nothing
```

* Closes #3161

Local pipeline runs and unit tests

**Yes, this is a breaking change for any downstream consumers of otap-dataflow via Prometheus or OTLP metric exporter**

### Changelog

### DIFF
--- a/rust/otap-dataflow/.chloggen/resource-scope-datapoint-attributes.yaml
+++ b/rust/otap-dataflow/.chloggen/resource-scope-datapoint-attributes.yaml
@@ -1,0 +1,22 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: observability
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Self-telemetry attributes are now split across the correct OpenTelemetry layers instead of all being emitted as data-point attributes
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [3161]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Process/host identity (`host.id`, `container.id`, `service.instance.id`) moves
+  to the Resource layer (SDK `Resource` / Prometheus `target_info`); entity
+  identity moves to the instrumentation scope (`otel_scope_*` labels). Resource
+  attributes are no longer per-metric labels (query `target_info`), and entity
+  attributes are now `otel_scope_`-prefixed on the admin Prometheus output.
--- a/rust/otap-dataflow/.chloggen/resource-scope-datapoint-attributes.yaml
+++ b/rust/otap-dataflow/.chloggen/resource-scope-datapoint-attributes.yaml
@@ -6,7 +6,7 @@ change_type: breaking
 component: observability
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
-note: Self-telemetry attributes are now split across the correct OpenTelemetry layers instead of all being emitted as data-point attributes
+note: Self-telemetry metric attributes are now split across the correct OpenTelemetry layers instead of all being emitted as data-point attributes
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
 issues: [3161]
@@ -15,8 +15,10 @@ issues: [3161]
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: |
-  Process/host identity (`host.id`, `container.id`, `service.instance.id`) moves
-  to the Resource layer (SDK `Resource` / Prometheus `target_info`); entity
-  identity moves to the instrumentation scope (`otel_scope_*` labels). Resource
-  attributes are no longer per-metric labels (query `target_info`), and entity
-  attributes are now `otel_scope_`-prefixed on the admin Prometheus output.
+  Metric entity identity moves to the instrumentation scope (`otel_scope_*`
+  labels); logs already emitted entity identity on scope. Process/host identity
+  (`host.id`, `container.id`, `service.instance.id`) is now applied
+  consistently to the Resource layer for both metrics and logs (SDK `Resource`
+  / Prometheus `target_info`). Resource attributes are no longer per-metric
+  labels (query `target_info`), and metric entity attributes are now
+  `otel_scope_`-prefixed on the admin Prometheus output.
--- a/rust/otap-dataflow/crates/admin/src/telemetry.rs
+++ b/rust/otap-dataflow/crates/admin/src/telemetry.rs
@@ -1784,7 +1784,10 @@ where
         let final_key = if prefix.is_empty() {
             sanitized
         } else {
-            format!("{prefix}{sanitized}")
+            let mut key = String::with_capacity(prefix.len() + sanitized.len());
+            key.push_str(prefix);
+            key.push_str(&sanitized);
+            key
         };
         // Skip keys that collide with separately-emitted scope/reserved
         // labels. Comparison is on the final label key after prefixing.
--- a/rust/otap-dataflow/crates/admin/src/telemetry.rs
+++ b/rust/otap-dataflow/crates/admin/src/telemetry.rs
@@ -938,8 +938,9 @@ pub(crate) fn render_target_info(
         resource_attributes
             .iter()
             .map(|(k, v)| (k.as_str(), v.to_string_value())),
+        "",
         // No reserved keys: `target_info` is a separate metric line and
-        // does not carry `otel_scope_*` labels, so no collision is possible.
+        // resource attributes remain unprefixed.
         &[],
     );
     let mut labels = String::new();
@@ -985,15 +986,17 @@ fn agg_prometheus_text(
                 escape_prom_label_value(&g.name)
             );
         }
-        // Merge values for keys that collide after sanitization (per
-        // OTel→Prometheus spec). Emission order is unspecified — Prometheus
-        // treats labels as an unordered set. Drop attribute keys that
-        // sanitize to the reserved `otel_scope_*` names already emitted
-        // above to avoid duplicate-label rejection by Prometheus.
+        // Scope attributes become `otel_scope_<key>` labels. Merge values
+        // for keys that collide after sanitization (per OTel→Prometheus
+        // spec). Emission order is unspecified — Prometheus treats labels as
+        // an unordered set. Drop attribute keys whose prefixed labels collide
+        // with reserved `otel_scope_*` names already emitted above to avoid
+        // duplicate-label rejection by Prometheus.
         let merged = sanitize_and_merge_label_pairs(
             g.attributes
                 .iter()
                 .map(|(k, v)| (k.as_str(), v.to_string_value())),
+            "otel_scope_",
             RESERVED_SCOPE_LABEL_KEYS,
         );
         for (k, v) in &merged {
@@ -1288,14 +1291,16 @@ fn format_prometheus_text(
                 escape_prom_label_value(descriptor.name)
             );
         }
-        // Merge values for keys that collide after sanitization (per
-        // OTel→Prometheus spec). Emission order is unspecified. Drop
-        // attribute keys that sanitize to the reserved `otel_scope_*`
-        // names already emitted above.
+        // Scope attributes become `otel_scope_<key>` labels. Merge values
+        // for keys that collide after sanitization (per OTel→Prometheus
+        // spec). Emission order is unspecified. Drop attribute keys whose
+        // prefixed labels collide with reserved `otel_scope_*` names already
+        // emitted above.
         let merged = sanitize_and_merge_label_pairs(
             attributes
                 .iter_attributes()
                 .map(|(k, v)| (k, v.to_string_value())),
+            "otel_scope_",
             RESERVED_SCOPE_LABEL_KEYS,
         );
         for (key, value) in &merged {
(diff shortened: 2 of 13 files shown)


rust/otap-dataflow/docs/telemetry/attributes-guide.md
```diff
@@ -89,6 +89,40 @@ Do not duplicate information:
...
+
+| Category | OTLP export (metrics + logs) | Admin Prometheus endpoint (metrics only) |
+| --- | --- | --- |
+| Resource | `Resource` on `ResourceMetrics` / `ResourceLogs` | `target_info{...}` gauge labels |
+| Entity | `InstrumentationScope.attributes` on scope metrics / scope logs | `otel_scope_<key>` labels on each series |
+| Signal-specific | Data-point attributes (metrics) / log-record attributes (logs) | Inline (unprefixed) data-point labels |
+
+Notes:
+
+- The engine auto-detects and injects process/host resource attributes
```
cijothomas, line 106: for a future discussion - I don't think engine should do this like this. we could use the notion of ResourceDetectors and let operators pick the ones they want.

rust/otap-dataflow/.chloggen/resource-scope-datapoint-attributes.yaml
```diff
@@ -0,0 +1,22 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: observability
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Self-telemetry attributes are now split across the correct OpenTelemetry layers instead of all being emitted as data-point attributes
```
cijothomas: only for metrics right? logs were already following the right approach?

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1773,14 +1781,18 @@ where
     let mut merged: HashMap<String, String> = HashMap::with_capacity(entries.len());
     for (key, value) in entries {
         let sanitized = sanitize_prom_label_key(key);
+        let final_key = if prefix.is_empty() {
+            sanitized
+        } else {
+            format!("{prefix}{sanitized}")
+        };
```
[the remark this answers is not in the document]
drewrelmas: Incorrect suggested changes, would not compile! Handling myself

rust/otap-dataflow/.chloggen/resource-scope-datapoint-attributes.yaml
drewrelmas: That is correct, I've improved the verbiage here. Logs were already doing the right thing.

rust/otap-dataflow/crates/admin/src/telemetry.rs
```diff
@@ -1791,15 +1806,14 @@ where
...
-/// not a collision. Add it here when version emission is implemented.
-const RESERVED_SCOPE_LABEL_KEYS: &[&str] = &["otel_scope_name"];
+/// separately from per-metric attributes; per the spec, scope attributes whose
+/// prefixed keys collide with these reserved identity labels are dropped to
+/// avoid duplicate-label scrape errors.
+const RESERVED_SCOPE_LABEL_KEYS: &[&str] = &[
+    "otel_scope_name",
+    "otel_scope_version",
+    "otel_scope_schema_url",
+];
```
lalitb, line 1816: Not sure if I am reading this correct - but seems `version` or `schema.url` gets dropped even though there is no duplicate label to avoid. Could we either only reserve the labels we currently emit, or actually emit scope version/schema URL separately before reserving them?

rust/otap-dataflow/crates/engine/src/context.rs
```diff
@@ -204,20 +207,35 @@ impl ControllerContext {
...
+    }
+
+    /// Returns the auto-detected process/host resource attributes mapped to
+    /// OpenTelemetry semantic-convention keys. Empty values are omitted.
+    /// Keys: `host.id`, `container.id`, `service.instance.id`.
+    #[must_use]
+    pub fn resource_attributes(&self) -> Vec<(String, String)> {
+        let mut out = Vec::new();
+        if !self.host_id.is_empty() {
+            out.push(("host.id".to_string(), self.host_id.to_string()));
```
lalitb, line 249: As per the [specs](https://opentelemetry.io/docs/specs/semconv/registry/attributes/host/#host-id), `host.id` should be the unique host identifier:

``
Unique host ID. For Cloud, this must be the instance_id assigned by the cloud provider. For non-containerized systems, this should be the machine-id. See the table below for the sources to use to determine the machine-id based on operating system
``

However `self.host_id` here actually gets [populated](https://github.com/open-telemetry/otel-arrow/blob/5613cf0a998cb0d484ada4ebd5ac396285ffeab5/rust/otap-dataflow/crates/engine/src/context.rs#L43) from `/etc/hostname` or `$HOSTNAME` both of which are not correct identifiers.  Could we map this value to `host.name`, or only populate `host.id` when we detect a real machine-id?

rust/otap-dataflow/crates/admin/src/telemetry.rs
drewrelmas: Thanks for the pointer, yes we do only have `otel_scope_name` right now, but I imagined it would be a painful breaking change later if a component was emitting `version` and all of a sudden we start overriding it by introducing our own. I thought it could be best to just reserve those keys ahead of time.

rust/otap-dataflow/crates/engine/src/context.rs
lalitb: I like the idea of resource detector [here](https://github.com/open-telemetry/otel-arrow/pull/3168#discussion_r3342892613), and also good handle this issue in future as part of adding that support.

drewrelmas: Agree, opened sub-issue #3176 to keep this PR focused.

drewrelmas: After reading up more on resource detector, I closed #3176 and opened #3177 as the proper direction, but I think it is still separate from this PR.

rust/otap-dataflow/docs/telemetry/attributes-guide.md
drewrelmas: Opened #3177 to track this

tools/pipeline_perf_test/test_suites/integration/configs/integration_report_logs.yaml
```diff
@@ -91,7 +91,7 @@ queries:
         metrics_filtered AS (
             SELECT *,
             "metric_attributes.component_name" AS component_name,
-            COALESCE("metric_attributes.core_id", 'N/A') AS core_id,
+            COALESCE("metric_attributes.otel_scope_core_id", 'N/A') AS core_id,
```
lalitb, line 94: This YAML update makes sense to keep the reports working with the new Prometheus labels. One follow-up question: should we also normalize `otel_scope_*` labels in the perf scraper into `scope_attributes`, so future reports don't have to keep querying scope identity through `metric_attributes`?

rust/otap-dataflow/crates/admin/src/telemetry.rs
lalitb: Makes sense. Reserving the full scope identity label set now avoids creating a compatibility trap for users before version/schema URL emission is added.

lalitb (approved):
LGTM. Thanks.

tools/pipeline_perf_test/test_suites/integration/configs/integration_report_logs.yaml
drewrelmas: I'm actually not sure about this, @cijothomas? Didn't we mention on the SIG call yesterday that our prometheus exporter wasn't exactly spec-compliant (it still puts scope attributes into metric datapoint line of prometheus output?)
```log
target_info{host_id="<computer name>",service_instance_id="AGPIRRISSR6UDHSZJSAHIKJV3U",service_name="resource-demo",deployment_environment_name="dev-proof"} 1
cpu_utilization{otel_scope_name="pipeline",otel_scope_pipeline_group_id="demo",otel_scope_core_id="0",otel_scope_pipeline_id="main",otel_scope_deployment_generation="0",otel_scope_numa_node_id="0"}0.00135 
```

drewrelmas: Will get this fix in now as-is and can always address this in a follow-up.

PR #3180, opened 2026-06-02
fix: replace OPL `exclude`/`date_time` keywords to match spec

albertlockett:
# Change Summary

In https://github.com/open-telemetry/otel-arrow/pull/3051 we are adding a specification for OPL. It has some minor keyword differences from the current implementation.

The keyword for the operation it specifies to remove attributes is `remove`, but currently we use `exclude`. I think `remove` is a more sensible name, so we'll make this change. (`project-away` will remain an alias`).

```
// before this would be `exclude attributes["x"]`
logs | remove attributes["x"]
```

The tag we use for timestamp literals also changes to match the spec, from `date_time` to `timestamp`. This seems more sensible as well, because that is what the arrow type is called. E.g., now timestamp/datetime literals will be defined like `timestamp"..."`

* Relates to #3051 

Unit tests

Yes - this is a breaking change to OPL syntax
 

### Changelog


lquerel (approved):
LGTM

rust/otap-dataflow/.chloggen/opl-spec-keyword-corrections.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: breaking
+component: query-engine
+note: changes keyword "exclude" to "remove", and changes "date_time" literal tag to "timestamp" to align with nascent spec
+issues: [3051]
```
ThomsonTan: nit: end with new line.

PR #3181, opened 2026-06-02
pdata: handle nanosecond duration/timestamp cardinality in concatenate

ragumarimuthu-git:
## Summary
This pull request fixes a bug in the `pdata` component related to cardinality estimation for Arrow `Duration` and `Timestamp` types. Previously, the code would panic when handling these types with certain time units. The update ensures that all Arrow time units are correctly supported for both `Duration` and `Timestamp` types, and adds comprehensive tests to cover these cases.

Bug fix and improved handling for Arrow time types:

* Updated `estimate_cardinality` in `concatenate.rs` to handle all Arrow `Duration` and `Timestamp` time units (`Second`, `Millisecond`, `Microsecond`, `Nanosecond`) without panicking.
* Added imports for all relevant Arrow time unit types (`DurationSecondType`, `TimestampMillisecondType`, etc.) to support the new logic.


rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
```diff
@@ -1099,6 +1130,57 @@ mod schema_tests {
...
 
+    #[test]
+    fn test_cardinality_duration_all_time_units() {
+        for unit in [
+            arrow_schema::TimeUnit::Second,
+            arrow_schema::TimeUnit::Millisecond,
+            arrow_schema::TimeUnit::Microsecond,
+            arrow_schema::TimeUnit::Nanosecond,
+        ] {
+            test_cardinality_for_type(&[250], &DataType::Duration(unit), Some(DataType::UInt8));
```
lalitb, line 1141: I think this test still misses the new code path. With only 250 values, we return early before running the new timestamp/duration cardinality logic. Can we add a case with more than 255 values, like `[250, 10]`, so this actually tests the fix?

lalitb (approved):
LGTM once the test coverage gap above is fixed. Thanks @ragumarimuthu-git for contribution.

rust/otap-dataflow/crates/pdata/src/otap/transform/concatenate.rs
```diff
@@ -1596,6 +1722,26 @@ mod schema_tests {
...
+                    }
+                    arrow_schema::TimeUnit::Microsecond => {
+                        Arc::new(DurationMicrosecondArray::from(values))
+                    }
+                    arrow_schema::TimeUnit::Nanosecond => {
+                        Arc::new(DurationNanosecondArray::from(values))
+                    }
+                }
+            }
+            DataType::Timestamp(_, _) => {
```
ThomsonTan: Nit: consider `DataType::Timestamp(_unit, _tz)` instead of `DataType::Timestamp(_, _)` to make it self-documenting that both fields matter (the full `value_type` is forwarded to `cast`). Not blocking.

ThomsonTan (approved):

rust/otap-dataflow/.chloggen/pdata-temporal-cardinality-fix.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: bug_fix
+component: pdata
+note: Handle Duration and Timestamp cardinality estimation across all Arrow time units and add timezone-preserving timestamp test coverage in concatenate schema tests.
```
cijothomas: Handle Duration and Timestamp cardinality estimation across all Arrow time units in concatenate (previously panicked).

^ just this would be good. Changlogs don't need to talk about test additions, we use this for generated end-user facing changelogs only.

PR #3184, opened 2026-06-02
chore(repo): Reduce stale PR time to 14+7days

drewrelmas:
# Change Summary

Follow-up from general agreement in SIG meeting to help keep PR list shorter

N/A

N/A

N/A

### Changelog

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,9 +1,9 @@
-name: 'Mark stale issues and pull requests'
+name: "Mark stale issues and pull requests"
 
 on:
   schedule:
     # Run daily at 1:30 UTC
-    - cron: '30 1 * * *'
+    - cron: "30 1 * * *"
   workflow_dispatch:
 
 permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -24,19 +24,20 @@ jobs:
...
+          stale-issue-label: "stale"
           days-before-issue-stale: 180
           days-before-issue-close: 30
-          exempt-issue-labels: 'do-not-stale'
+          exempt-issue-labels: "do-not-stale"
           # Pull request settings
           stale-pr-message: >
             This pull request has been marked as stale due to lack of recent activity.
-            It will be closed in 30 days if no further activity occurs.
+            It will be closed in 7 days if no further activity occurs.
```
cijothomas: nit: it'd be nice if we ack that the lack of activity in the PR is due to the repo maintainers being busy, so it won't sound like the contributor did something wrong..

cijothomas: https://github.com/open-telemetry/opentelemetry-rust/blob/main/.github/workflows/stale-pr.yml could use some ideas from here, if that reads better.

utpilla (approved):

.github/workflows/stale.yml
```diff
@@ -24,19 +24,20 @@ jobs:
...
             If this PR is still relevant, please comment or push new commits to keep it active.
-          stale-pr-label: 'stale'
-          days-before-pr-stale: 14
-          days-before-pr-close: 30
-          exempt-pr-labels: 'do-not-stale'
+            If your PR was closed, you can reopen it to continue working on it.
+          stale-pr-label: "stale"
+          days-before-pr-stale: 7
+          days-before-pr-close: 7
+          exempt-pr-labels: "do-not-stale"
```
lalitb, line 41: I missed the community meeting where this was discussed, so please treat this as just my personal take if this was already agreed there.

` 7+7` feels a bit aggressive to me for a blanket PR stale policy. I understand the intent may be to encourage faster maintainer review and keep the PR queue manageable, but in practice the stale action is applied to the PR author’s contribution. If the delay is mostly on the maintainer/reviewer side, closing the PR can feel like the author is being penalized for something outside their control.

Most of the OpenTelemetry repos I checked either use a longer blanket PR window or apply `7+7` only when the next action is clearly on the author. `otel-dotnet` being only exception with `7+7` broad rule.

Would it make sense to either keep a broader `14+14` policy, or use `7+7` only when the next action is clearly on the author, for example when the PR has a `needs author feedback` label?

drewrelmas: That's a reasonable point and follows somewhat to Cijo's comment above - we want to be careful on the messaging. Perhaps `14+7` is a safer middle ground? And we can definitely tweak the message further. Let me take another iteration at it.

.github/workflows/stale.yml
drewrelmas: I took this almost verbatim @cijothomas, I think it is a great improvement!

.github/workflows/stale.yml
drewrelmas: I've updated to `14+7` and used almost the verbatim messaging Cijo linked from OTel Rust.

lalitb (approved):

JakeDern (approved):

cijothomas (approved):
Thanks!

PR #3186, opened 2026-06-03
Add user_agent config to shared HttpClientSettings and GrpcClientSettings

### LINKED ISSUE #3138: Add `user_agent` config option to shared `HttpClientSettings` and `GrpcClientSettings`

c-valdebenito:
# Change Summary

Add an optional user_agent field to the shared HttpClientSettings and GrpcClientSettings structs so any HTTP or gRPCexporter (and receiver) can inject a custom User-Agent string via config without duplicating the field in each node.

Expected behavior

 | Transport | `user_agent` set | Resulting header |
 |-----------|-----------------|------------------|
 | HTTP (reqwest) | `"my-app/1.0"` | `my-app/1.0` |
 | HTTP (reqwest) | not set | reqwest default |
 | gRPC (tonic) | `"my-app/1.0"` | `my-app/1.0 tonic/x.x.x` |
 | gRPC (tonic) | not set | `tonic/x.x.x` |

* Closes #3138 

6 new unit tests (3 per transport): deserialization, builder/endpoint success, default is None. All pass via cargo test -p otap-df-otap --lib -- user_agent. Clippy clean with cargo clippy -p otap-df-otap --features crypto-ring --lib -- -D warnings.

Yes, exporters already using the shared HttpClientSettings or GrpcClientSettings (e.g. otlp_http_exporter,otlp_grpc_exporter, otap_exporter) can now set user_agent in their config to customize the User-Agent header on outbound requests.

### Changelog

### DIFF
--- a/rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
@@ -229,7 +229,7 @@ impl GrpcClientSettings {
     /// Validates the settings at config load time.
     ///
     /// Checks that `user_agent`, when set, is non-empty and contains only
-    /// characters valid in an HTTP header value (visible ASCII, 32–127).
+    /// characters valid in an HTTP header value (visible ASCII, 32-127).
     pub fn validate(&self) -> Result<(), GrpcEndpointError> {
         if let Some(ua) = &self.user_agent {
             if ua.trim().is_empty() {
--- a/rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
@@ -145,6 +145,13 @@ pub struct GrpcClientSettings {
     #[serde(default)]
     #[doc(hidden)]
     pub proxy: Option<ProxyConfig>,
+
+    /// Custom User-Agent header for outbound gRPC requests. When set, tonic
+    /// **prepends** this value to its default `tonic/x.x.x` User-Agent (e.g.
+    /// `my-app/1.0 tonic/0.12.x`). When not set, only the default tonic
+    /// User-Agent is sent.
+    #[serde(default)]
+    pub user_agent: Option<String>,
 }
 
 /// Error returned when building a gRPC [`Endpoint`] (including TLS/mTLS setup).
@@ -342,6 +349,9 @@ impl GrpcClientSettings {
         if let Some(timeout) = self.timeout {
             endpoint = endpoint.timeout(timeout);
         }
+        if let Some(ua) = &self.user_agent {
+            endpoint = endpoint.user_agent(ua.as_str())?;
+        }
 
         Ok(endpoint)
     }
@@ -505,6 +515,7 @@ impl Default for GrpcClientSettings {
             buffer_size: None,
             startup_check: StartupCheck::default(),
             proxy: None,
+            user_agent: None,
         }
     }
 }
@@ -600,6 +611,23 @@ mod tests {
     }
 
     #[test]
+    fn test_user_agent_deserialization() {
+        let settings: GrpcClientSettings = serde_json::from_str(
+            r#"{ "grpc_endpoint": "http://localhost:4317", "user_agent": "my-app/1.0" }"#,
+        )
+        .unwrap();
+
+        assert_eq!(settings.user_agent.as_deref(), Some("my-app/1.0"));
+    }
+
+    #[test]
+    fn test_default_has_no_user_agent() {
+        let settings = GrpcClientSettings::default();
+
+        assert_eq!(settings.user_agent, None);
+    }
+
+    #[test]
     fn effective_concurrency_limit_clamps_to_one() {
         let settings: GrpcClientSettings = serde_json::from_str(
             r#"{ "grpc_endpoint": "http://localhost:4317", "concurrency_limit": 0 }"#,
@@ -607,6 +635,18 @@ mod tests {
         .unwrap();
 
         assert_eq!(settings.effective_concurrency_limit(), 1);
+    }
+
+    #[test]
+    fn test_user_agent_applied_to_endpoint() {
+        let settings = GrpcClientSettings {
+            grpc_endpoint: "http://localhost:4317".to_string(),
+            user_agent: Some("my-app/1.0".to_string()),
+            ..GrpcClientSettings::default()
+        };
+
+        let endpoint = settings.build_endpoint().unwrap();
+        let _ = endpoint;
     }
 
     #[tokio::test]
--- a/rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
@@ -611,6 +611,55 @@ mod tests {
     }
 
     #[test]
+    fn validate_rejects_empty_user_agent() {
+        let settings = GrpcClientSettings {
+            user_agent: Some(String::new()),
+            ..GrpcClientSettings::default()
+        };
+
+        assert!(matches!(
+            settings.validate(),
+            Err(GrpcEndpointError::InvalidConfig(_))
+        ));
+    }
+
+    #[test]
+    fn validate_rejects_whitespace_only_user_agent() {
+        let settings = GrpcClientSettings {
+            user_agent: Some("   ".to_string()),
+            ..GrpcClientSettings::default()
+        };
+
+        assert!(matches!(
+            settings.validate(),
+            Err(GrpcEndpointError::InvalidConfig(_))
+        ));
+    }
+
+    #[test]
+    fn validate_rejects_non_ascii_user_agent() {
+        let settings = GrpcClientSettings {
+            user_agent: Some("bad\nvalue".to_string()),
+            ..GrpcClientSettings::default()
+        };
+
+        assert!(matches!(
+            settings.validate(),
+            Err(GrpcEndpointError::InvalidConfig(_))
+        ));
+    }
+
+    #[test]
+    fn validate_accepts_valid_user_agent() {
+        let settings = GrpcClientSettings {
+            user_agent: Some("my-app/1.0".to_string()),
+            ..GrpcClientSettings::default()
+        };
+
+        assert!(settings.validate().is_ok());
+    }
+
+    #[test]
     fn effective_concurrency_limit_clamps_to_one() {
         let settings: GrpcClientSettings = serde_json::from_str(
             r#"{ "grpc_endpoint": "http://localhost:4317", "concurrency_limit": 0 }"#,
--- a/rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
+++ b/rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
@@ -6,6 +6,7 @@
 use crate::compression::CompressionMethod;
 use crate::otap_grpc::proxy::ProxyConfig;
 use crate::tls_utils;
+use http::header::HeaderValue;
 use hyper_util::rt::TokioIo;
 use otap_df_config::byte_units;
 use otap_df_config::tls::TlsClientConfig;
@@ -173,6 +174,10 @@ pub enum GrpcEndpointError {
     #[error("invalid grpc_endpoint: {0}")]
     InvalidEndpoint(String),
 
+    /// Invalid configuration value detected at validation time.
+    #[error("invalid configuration: {0}")]
+    InvalidConfig(String),
+
     /// DNS resolution failed during a `startup_check: dns` check.
     #[error("startup dns check failed for \"{host}\": {source}")]
     DnsCheckFailed {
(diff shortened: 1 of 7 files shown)


rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
```diff
@@ -64,6 +64,12 @@ pub struct HttpClientSettings {
         alias = "compression_method"
     )]
     pub compression: Option<CompressionMethod>,
+
+    /// Custom User-Agent header for outbound HTTP requests. When set, reqwest
+    /// **replaces** its default User-Agent with this value. When not set, the
+    /// default reqwest User-Agent is used.
```
drewrelmas: IIRC technically the default if not provided is empty (`reqwest` doesn't add any value). Can you please reflect that in the comment here?

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -108,6 +108,13 @@ pub struct GrpcClientSettings {
     #[serde(default)]
     #[doc(hidden)]
     pub proxy: Option<ProxyConfig>,
+
+    /// Custom User-Agent header for outbound gRPC requests. When set, tonic
+    /// **prepends** this value to its default `tonic/x.x.x` User-Agent (e.g.
+    /// `my-app/1.0 tonic/0.12.x`). When not set, only the default tonic
+    /// User-Agent is sent.
```
drewrelmas, line 153: Is it possible to get a unit test that demonstrates this behavior (the `tonic/` portion)? It may be more involved than just checking the resulting endpoint. Maybe standing up a fake server and verifying `user_agent` of sample requests against it?

drewrelmas (commented):
Thanks for your first contribution @c-valdebenito! Some small questions below.

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
```diff
@@ -89,6 +95,10 @@ impl HttpClientSettings {
                 self.effective_concurrency_limit(),
             ));
 
+        if let Some(ua) = &self.user_agent {
+            client_builder = client_builder.user_agent(ua.as_str());
```
utpilla, line 122: Could we validate `user_agent` at config load?

Right now, neither `HttpClientSettings` nor `GrpcClientSettings` validates user_agent before passing it down. Two classes of values would silently slip through and surface late:

- Empty or whitespace-only: sending an explicit empty `User-Agent` is rarely intentional and defeats server-side attribution. [`HeaderValue::from_str("")`](https://docs.rs/reqwest/latest/reqwest/header/struct.HeaderValue.html#method.from_str) and `HeaderValue::from_str(" ")` both succeed, so this needs its own check.
- Invalid header bytes (CR, LF, NUL, TAB, non-ASCII): `HeaderValue::from_str` enforces visible ASCII only (32-127) and rejects these.

Today both paths would already fail at exporter startup. HTTP via `ClientBuilder::build()` inside `HttpClientPool::try_new`, gRPC via `build_endpoint_from_uri()`. The downside isn't when it fails but how: the user sees an opaque `reqwest::Error` or `tonic::transport::Error` that doesn't name `user_agent` or the rule it broke. A typed config error keeps the message consistent with the Azure Monitor exporter and makes the failure self-describing.

Azure Monitor Exporter does something [similar](https://github.com/open-telemetry/otel-arrow/blob/3a482a342201f71bae5e100998c2c16d05ac0f9d/rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs#L244-L255). Could we add a validate method inside the settings? Concretely:

```rust
impl HttpClientSettings {
    pub fn validate(&self) -> Result<(), ConfigError> {
        if let Some(ua) = &self.user_agent {
            if ua.trim().is_empty() {
                return Err(/* user_agent must be non-empty */);
            }
            if HeaderValue::from_str(ua).is_err() {
                return Err(/* user_agent must contain only visible ASCII (32-127) */);
            }
        }
        Ok(())
    }
}
```

(Same shape on `GrpcClientSettings`.) Each consuming exporter's `Config::validate()` then calls `self.client.validate()?`, so misconfigurations fail at pipeline startup with a clear, transport-agnostic message instead of mid-flight.

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
```diff
@@ -231,6 +241,46 @@ impl Default for HttpClientSettings {
...
+
+        assert_eq!(
+            settings.user_agent.as_deref(),
+            Some("otap-custom-agent/1.0")
+        );
+    }
+
+    #[tokio::test]
+    async fn test_user_agent_applied_to_client_builder() {
+        crate::crypto::ensure_crypto_provider();
```
utpilla, line 379: This won't be needed to run this test.

```suggestion
```

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
```diff
@@ -231,6 +241,50 @@ impl Default for HttpClientSettings {
...
+    use wiremock::matchers;
+    use wiremock::{Mock, MockServer, ResponseTemplate};
+
+    #[tokio::test]
+    async fn test_user_agent_sent_on_wire() {
+        crate::crypto::ensure_crypto_provider();
+
+        // Verify default has no user_agent
+        let defaults = HttpClientSettings::default();
+        assert_eq!(defaults.user_agent, None);
```
ThomsonTan, line 383: The unset-case contract "no User-Agent header is sent when not set" seems tested; `assert_eq!(defaults.user_agent, None)` only checks the struct field, not the wire. Could we assert (reusing the existing `MockServer`) that a default `HttpClientSettings` yields a request with **no** `User-Agent`? wiremock has no "header absent" matcher, so a tiny custom `Respond` checking `req.headers.get("user-agent").is_none()` works. Separate from the empty/invalid-value tests — this locks down the *unset* path.

c-valdebenito: great idea! Thanks @ThomsonTan , will update!

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
c-valdebenito: great idea @utpilla ! will implement

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
```diff
@@ -231,6 +268,96 @@ impl Default for HttpClientSettings {
...
+
+        assert!(
+            !saw_ua.load(Ordering::SeqCst),
+            "Expected no User-Agent header when user_agent is unset, but one was sent"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_user_agent_sent_on_wire() {
+
```
utpilla, line 380: There is still a new line left which cargo fmt is complaining about. I'm going to apply this suggestion to delete the new line.

```suggestion
```

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
```diff
@@ -64,9 +65,37 @@ pub struct HttpClientSettings {
...
+    #[serde(default)]
+    pub user_agent: Option<String>,
 }
 
 impl HttpClientSettings {
+    /// Validates the settings at config load time.
+    ///
+    /// Checks that `user_agent`, when set, is non-empty and contains only
+    /// characters valid in an HTTP header value (visible ASCII, 32–127).
+    pub fn validate(&self) -> Result<(), HttpClientError> {
```
utpilla, line 81: Thanks for adding the `validate` methods to both HTTP and gRPC client settings!

However, there are two main gaps that shouls be addressed:

1. No production caller\: Neither `HttpClientPool::try_new` nor any exporter (`otlp_http_exporter`, `otlp_grpc_exporter`, `otap_exporter`) invokes `HttpClientSettings::validate()` or `GrpcClientSettings::validate()`. The factories register `validate_config: validate_typed_config::<Config>`, which is static-only. It just checks deserialization. So, a bad `user_agent` still surfaces as an opaque `reqwest::Error` / `tonic::transport::Error` at startup
2. No test caller:The empty / whitespace / non-ASCII rejection paths in `validate()` are unreachable from any test, so the logic itself is uncovered.

Could we:

- Call `self.config.http.validate()?` (and `self.config.grpc.validate()?`) once in each consuming exporter's constructor / `start()`, before `HttpClientPool::try_new` / `build_endpoint_with_tls`; and
- Add unit tests directly on the settings struct, e.g.:

```rust
#[test]
fn validate_rejects_empty_user_agent() {
    let s = HttpClientSettings { user_agent: Some(String::new()), ..Default::default() };
    assert!(matches!(s.validate(), Err(HttpClientError::InvalidConfig(_))));
}

#[test]
fn validate_rejects_whitespace_only_user_agent() { /* "   " */ }

#[test]
fn validate_rejects_non_ascii_user_agent() { /* "bad\nvalue" */ }

#[test]
fn validate_accepts_valid_user_agent() { /* "my-app/1.0" */ }
```

Same with `GrpcClientSettings` as well. That closes the loop on both the wiring and the coverage.

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -291,6 +303,28 @@ impl GrpcClientSettings {
         Ok(())
     }
 
+    /// Validates the settings at config load time.
+    ///
+    /// Checks that `user_agent`, when set, is non-empty and contains only
+    /// characters valid in an HTTP header value (visible ASCII, 32–127).
```
utpilla: Let's use the ASCII hyphen instead of the non ASCII en-dash.

```suggestion
    /// characters valid in an HTTP header value (visible ASCII, 32-127).
```

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
```diff
@@ -64,9 +65,37 @@ pub struct HttpClientSettings {
...
+    /// header is sent (reqwest does not add one by default).
+    #[serde(default)]
+    pub user_agent: Option<String>,
 }
 
 impl HttpClientSettings {
+    /// Validates the settings at config load time.
+    ///
+    /// Checks that `user_agent`, when set, is non-empty and contains only
+    /// characters valid in an HTTP header value (visible ASCII, 32–127).
```
utpilla: Same as this: https://github.com/open-telemetry/otel-arrow/pull/3186#discussion_r3384714865

```suggestion
    /// characters valid in an HTTP header value (visible ASCII, 32-127).
```

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
utpilla: > Call self.config.http.validate()? (and self.config.grpc.validate()?) once in each consuming exporter's constructor / start(), before HttpClientPool::try_new / build_endpoint_with_tls; and

Following up on this. I think my original suggestion was a bit misleading. Calling `validate()` from start would catch bad config, but only after the controller has already accepted the pipeline and begun bringing nodes up. On a live reconfigure that means we'd tear down a healthy node and replace it with one that fails on startup.

An even better place is the factory's `validate_config` pointer, which the controller invokes at both initial load and live reconfigure, before any node is started. A small free function per exporter that deserializes `Config` and calls `cfg.grpc.validate()` / `cfg.http.validate()`, wired in as `validate_config`, gives us:

- fail-fast rejection before deployment,
- a typed `InvalidUserConfig` error attributed to the offending node (what admin/CLI tooling already renders),
- and a pure-function entry point that's trivially unit-testable (no runtime or client builder needed).

`validate_typed_config` only covers `serde` and is documented as static-only, so semantic checks like `user_agent` are better enforced in the per-exporter `validate_config` wrapper rather than in `start`.

c-valdebenito: sounds great, made the changes!

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
utpilla:
```suggestion
    /// characters valid in an HTTP header value (visible ASCII, 32-127).
```

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -214,6 +226,28 @@ fn validate_grpc_endpoint(endpoint: &str) -> Result<(), String> {
 }
 
 impl GrpcClientSettings {
+    /// Validates the settings at config load time.
+    ///
+    /// Checks that `user_agent`, when set, is non-empty and contains only
+    /// characters valid in an HTTP header value (visible ASCII, 32–127).
```
utpilla:
```suggestion
    /// characters valid in an HTTP header value (visible ASCII, 32-127).
```

utpilla (approved):
@c-valdebenito Thanks for working on this!

drewrelmas (approved):
LGTM!

Issue #2783, opened 2026-04-29
Add Windows ETW Receiver for telemetry from applications co-located with otap-dataflow


utpilla:
## Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

## Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

## Summary

Add a receiver for application telemetry emitted through Windows [Event Tracing for Windows (ETW)](https://learn.microsoft.com/en-us/windows/win32/etw/event-tracing-portal) sessions. The receiver will support TraceLogging, TraceLoggingDynamic, and manifest-based event formats.

This receiver targets telemetry emitted through ETW providers on the same Windows host as the collector. A single ETW session can subscribe to multiple providers, and [`one_collect`](https://github.com/microsoft/one-collect) manages the full session lifecycle (creation, provider subscription, buffer configuration, `ProcessTrace` thread, lost event callbacks, graceful shutdown). Events are consumed via the `ProcessTrace` callback, decoded using TDH APIs, and distributed to per-core pipelines through per-core `tokio::sync::mpsc` channels.

Related decoding surface:

- TDH APIs: Available via the [`windows`](https://crates.io/crates/windows)/ [windows-sys](https://crates.io/crates/windows-sys) crate

ETW session management library:

- [`one_collect`](https://github.com/microsoft/one-collect) -- Manages the full ETW session lifecycle including session creation, provider subscription, `ProcessTrace` thread management, buffer configuration, lost event callbacks, and graceful shutdown. Without it, the receiver would need raw Win32 API calls (`StartTrace`, `EnableTraceEx2`, `OpenTrace`, `ProcessTrace`, `StopTrace`) with careful error handling.

This proposal is to add this as part of the `contrib-nodes` crate under a feature flag.

## Motivation

For applications running on the same Windows host as the collector, ETW provides a native kernel-managed event path. Events can be consumed from an ETW session via a `ProcessTrace` callback, avoiding the local TCP/UDP/HTTP/gRPC stack. This offers lower-overhead, lower-latency same-host ingestion with bounded kernel-side buffering.

Key advantages over network-based ingestion:

- **No network stack overhead** -- no TCP/gRPC/HTTP framing, no protobuf serialization, no listener sockets, no TLS termination (events still require TDH decoding from the ETW buffer format)
- **Kernel-managed buffering** absorbs producer bursts (events are dropped when buffers fill, not back-pressured)

### Use cases

[...]

PR #3187, opened 2026-06-03
ETW receiver: add TDH field decoding, Arrow conversion, and batched flush

swashtek:
Towards #2783 

# Change Summary
This covers the three main areas in the branch:

1. __TDH field decoding__ — TraceLogging schema discovery via `TdhGetEventInformation`, field extraction via skip-chains
2. __Arrow conversion__ — `EtwArrowRecordsBuilder` mapping ETW events to OTAP Arrow log record batches (severity, attributes, timestamps)
3. __Batched flush__ — configurable `max_size` / `max_duration` batch policy with timer-based and size-based flush in the async event loop

* Closes #NNN

No

### Changelog

* [ ] Added a `.chloggen/*.yaml` entry, OR this PR is a `chore` (indicated in title).


rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -64,17 +139,36 @@ const EVENT_CHANNEL_CAPACITY: usize = 4096;
...
+#[derive(Debug, Clone)]
+pub struct DecodedField {
+    /// Field name (e.g. `"ProcessId"`, or `"Parent.ChildField"` for nested structs).
+    pub name: String,
+    /// Type name matching one_collect conventions (e.g. `"u32"`, `"string"`, `"wstring"`).
+    pub type_name: String,
+    /// Raw field data bytes copied from the event payload.
+    /// Empty for unsupported or zero-length fields.
+    pub data: Vec<u8>,
+}
```
utpilla, line 182: The current `DecodedField { name: String, type_name: String, data: Vec<u8> }` defers type interpretation to the encoder, which then string-matches on `type_name` and `data.len()` to reconstruct a typed value. A few issues with this:

1. Silent type drift: If `one_collect` ever renames a type string (e.g. `"unsigned int"` → `"u32_le"`), the encoder's match arm stops firing and the field silently falls through to the `_ => AttrValue::Bytes(...)` hex-string path. The compiler can't catch this; only a test that covers the exact `(type_name, len)` pair will.

2. Missing TDH types degrade silently: `TDH_INTYPE_BOOLEAN`, `TDH_INTYPE_GUID`, `TDH_INTYPE_FILETIME`, `TDH_INTYPE_SID`, `TDH_INTYPE_HEXINT*`, etc. aren't in the match. They all become hex strings downstream with no signal that decoding wasn't fully supported.

3. Wasted allocations on the ETW callback hot path. For a `u32` field we currently allocate `String` (name) + `String` (type_name) + `Vec<u8>` (data) on the single `ProcessTrace` thread, only for the consumer to undo two of those to recover an `i64`. The `user_events_receiver` already solved the same problem with a typed enum:

```rust
pub enum DecodedAttributeValue { Str(String), Int(i64), Bool(bool), Double(f64) }
```

Decoder produces the typed variant once; encoder is a small exhaustive match. Adding a new variant is a compile error at every match site. There are no silent fall-throughs.

I would suggest move the TDH type → value mapping from `decode_field_value` in `arrow_records_encoder.rs` into `extract_decoded_fields` in `session.rs`, and define an `EtwAttributeValue` enum modeled on `DecodedAttributeValue`. `DecodedField` then becomes `{ name: String, value: EtwAttributeValue }` and the encoder collapses to a match statement.

This keeps encoding distributed across consumer cores (no change there), eliminates the duplicate `type_name: String` allocation per field, removes consumer-side byte re-parsing, and centralizes TDH type knowledge next to the decoder where it belongs.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -75,6 +85,11 @@ use std::time::Duration;
 /// URN for the ETW receiver.
 pub const ETW_RECEIVER_URN: &str = "urn:otel:receiver:etw";
 
+// ── Defaults ─────────────────────────────────────────────────────────────────
+
+const DEFAULT_BATCH_MAX_SIZE: u16 = 512;
+const DEFAULT_BATCH_MAX_DURATION: Duration = Duration::from_millis(50);
```
utpilla: Let's use `100ms` as the default duration?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -118,6 +133,32 @@ struct ProviderConfig {
...
+///
+/// Decoded ETW events are accumulated into OTAP log batches before they are
+/// sent downstream.  Batching improves throughput but increases the number of
+/// decoded records that can be lost if the process exits before a flush.
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+#[serde(deny_unknown_fields)]
+struct BatchConfig {
+    /// Maximum number of log records per emitted OTAP batch.
+    #[serde(default = "default_batch_max_size")]
+    max_size: u16,
```
utpilla: Let's use `NonZeroU16` as the type for `max_size`. That would enforce that the user provides valid values.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -220,28 +284,75 @@ impl EtwReceiver {
...
+) -> Result<(), Error> {
+    if builder.is_empty() {
+        return Ok(());
+    }
+
+    let item_count = u64::from(builder.len());
+
+    let payload = std::mem::take(builder)
+        .build()
+        .map_err(|error| Error::ReceiverError {
```
utpilla: If the build fails, we should emit the forward failed metric.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -220,28 +284,75 @@ impl EtwReceiver {
...
             event_rx,
         })
     }
 }
 
+// ── Batch flush helpers ──────────────────────────────────────────────────────
+
+/// Flush the current Arrow batch downstream via the effect handler.
+///
+/// Resets `builder` to a fresh empty builder on success.  On failure the
```
utpilla: The builder seems to be reset regardless of outcome (success or failure).

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -251,7 +362,16 @@ impl EtwReceiver {
...
+                            // Drain remaining events from the channel before
+                            // flushing the final batch.
+                            while let Ok(event) = self.event_rx.try_recv() {
+                                self.metrics.borrow_mut().received_events_total.inc();
+                                builder.append(&event);
+                                if builder.len() >= batch_max_size {
+                                    flush_batch(effect_handler, &self.metrics, &mut builder).await?;
+                                }
+                            }
+                            flush_batch(effect_handler, &self.metrics, &mut builder).await?;
```
utpilla: Two issues here:

**1. Drain doesn't honour the deadline.**

`while let Ok(event) = self.event_rx.try_recv()` only exits when the channel is momentarily empty. The producer thread running `ProcessTrace` keeps round-robining events into our `tx`, so on a busy provider this loop can spin past the `deadline` indefinitely. The runtime expects `notify_receiver_drained` by the deadline; overshooting risks a forced kill mid-flush.

**2. `flush_batch` uses awaiting `send`, which can block on backpressure.**

During drain, downstream may be slow or already shutting down. `effect_handler.send_message_with_source_node(...).await` will park the task indefinitely if the downstream channel is full, again, blowing past the deadline. We should use `try_send_message_with_source_node` for exactly this reason.

#### Suggested fix

Two changes:

- **Bound the drain by 90% of the deadline** (mirroring the syslog CEF TCP pattern at [`syslog_cef_receiver/mod.rs`](rust/otap-dataflow/crates/core-nodes/src/receivers/syslog_cef_receiver/mod.rs)), capped at a small max budget so a generous deadline doesn't stall shutdown.
- **Use `try_send` for both flushes** so we never block on a slow downstream during drain. Skip the metric increment on success and add a `received_events_forward_failed` increment on `TrySendError::Full` / `Closed`.

```rust
const MAX_DRAIN_BUDGET: Duration = Duration::from_secs(1);

Ok(NodeControlMsg::DrainIngress { deadline, .. }) => {
    let _ = telemetry_timer_handle.cancel().await;

    let now = std::time::Instant::now();
    let remaining = deadline.saturating_duration_since(now);
    let drain_budget = std::cmp::min(remaining * 9 / 10, MAX_DRAIN_BUDGET);
    let drain_deadline = now + drain_budget;

    while std::time::Instant::now() < drain_deadline {
        match self.event_rx.try_recv() {
            Ok(event) => {
                self.metrics.borrow_mut().received_events_total.inc();
                builder.append(&event);
                if builder.len() >= batch_max_size {
                    try_flush_batch(effect_handler, &self.metrics, &mut builder)?;
                }
            }
            Err(_) => break, // Empty or Disconnected
        }
    }

    // Flush the trailing partial batch (no-op if already empty).
    try_flush_batch(effect_handler, &self.metrics, &mut builder)?;

    let snapshot = self.metrics.borrow().snapshot();
    return Ok(TerminalState::new(deadline, [snapshot]));
}
```

Where `try_flush_batch` is a sync sibling of `flush_batch` that calls `try_send_message_with_source_node` instead of awaiting `send_message_with_source_node`.

utpilla: The "right" fix would be to stop the producer for this core before draining, so `try_recv` reliably reaches `Disconnected`. That requires plumbing through [session.rs], a per-slot `unsubscribe()` that drops this core's `Sender` from the shared `txs` table (which today is `Rc<Vec<Sender>>` on the `ProcessTrace` thread and would need to become `Arc<[ArcSwapOption<Sender>]>` or similar). That's a session-lifecycle change with hot-path implications and belongs in its own PR. The deadline-bounded drain above is a safe interim fix.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -220,28 +284,75 @@ impl EtwReceiver {
...
         ctrl_chan: &mut local::ControlChannel<OtapPdata>,
         effect_handler: &local::EffectHandler<OtapPdata>,
         telemetry_timer_handle: TelemetryTimerCancelHandle<OtapPdata>,
     ) -> Result<TerminalState, Error> {
-        let mut event_count: u64 = 0;
         let mut channel_alive = true;
+        let mut builder = EtwArrowRecordsBuilder::new();
+        let batch_max_size = self.batching.max_size;
+
+        let mut flush_interval = time::interval(self.batching.max_duration);
```
utpilla: Use `interval_at` instead of `interval` here otherwise `flush_interval.tick()` would fire immediately.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -281,43 +401,34 @@ impl EtwReceiver {
                     }
                 }
 
+                // Timer-based batch flush
+                _ = flush_interval.tick() => {
```
utpilla: Include the channel check for timer branch as well.

```suggestion
                _ = flush_interval.tick(), if channel_alive => {
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -281,43 +401,34 @@ impl EtwReceiver {
...
+                                flush_batch(effect_handler, &self.metrics, &mut builder).await?;
+                            }
                         }
                         None => {
                             // Channel closed — the ETW session thread has
                             // exited (process shutdown or unrecoverable error).
-                            // Stop polling event_rx; only handle control
-                            // messages from now on.
+                            // Flush any remaining events, then stop polling.
+                            flush_batch(effect_handler, &self.metrics, &mut builder).await?;
```
utpilla: You want to use the `try_flush_batch` method here as well as mentioned in the [comment](https://github.com/open-telemetry/otel-arrow/pull/3187#discussion_r3378229750) for `DrainIngress`. We don't want to block when shutting down or stopping the receiver.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -220,28 +284,75 @@ impl EtwReceiver {
...
+            source_detail: format_error_sources(&error),
+        })?;
+
+    let pdata = OtapPdata::new_todo_context(payload.into());
+    if let Err(error) = effect_handler.send_message_with_source_node(pdata).await {
+        metrics
+            .borrow_mut()
+            .received_events_forward_failed
+            .add(item_count);
+        return Err(error.into());
```
utpilla: The only error reachable on the awaiting `send_message_with_source_node(...).await` path is `SendError::Closed` which happens when the downstream node is torn down. `SendError::Full` (backpressure) is impossible here because the awaiting send parks until a slot opens instead of returning.

Rather than propagating this via `?` and terminating the receiver on a single transient downstream closure, prefer to treat it as a per-batch loss event: increment `received_events_forward_failed`, log, and keep the receiver alive. The control plane (`DrainIngress`/`Shutdown`) decides when the receiver actually stops; data-path errors shouldn't take it down unilaterally. Check Syslog CEF Receiver for reference.

That way the receiver survives transient downstream restarts which might be a thing in future.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/arrow_records_encoder.rs
```diff
@@ -0,0 +1,427 @@
...
+        // Duplicate the event name as an attribute so it is preserved in
+        // exporters whose schema does not include the OTel `event_name`
+        // log record field (e.g. the Parquet exporter writes only the
+        // standard OTAP columns and omits `event_name`).
+        if !event.event_name.is_empty() {
+            self.append_attr(
+                "etw.event_name",
+                AttrValue::Str(Cow::Borrowed(&event.event_name)),
+            );
+        }
```
utpilla: Let's fix Parquet Exporter to export the event name instead of duplicating this in ETW Receiver?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/arrow_records_encoder.rs
```diff
@@ -0,0 +1,427 @@
...
+            guid[7],
+            guid[8],
+            guid[9],
+            guid[10],
+            guid[11],
+            guid[12],
+            guid[13],
+            guid[14],
+            guid[15],
+        );
```
utpilla: Could we use a stack buffer to avoid the allocation caused by `format!`?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/arrow_records_encoder.rs
```diff
@@ -0,0 +1,427 @@
...
+        self.append_attr("etw.provider_id", AttrValue::Str(Cow::Owned(provider_guid)));
+
+        // Activity ID — only emit when non-zero (provider set a correlation ID)
+        if event.activity_id != [0u8; 16] {
```
utpilla, line 215: Same as https://github.com/open-telemetry/otel-arrow/pull/3187#discussion_r3378387278

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/arrow_records_encoder.rs
```diff
@@ -0,0 +1,427 @@
...
+    fn append_attr(&mut self, key: &str, value: AttrValue<'_>) {
+        self.log_attrs.append_key(key);
+        match value {
+            AttrValue::Str(s) => self.log_attrs.any_values_builder.append_str(s.as_bytes()),
+            AttrValue::Int(i) => self.log_attrs.any_values_builder.append_int(i),
+            AttrValue::Double(d) => self.log_attrs.any_values_builder.append_double(d),
+            AttrValue::Bool(b) => self.log_attrs.any_values_builder.append_bool(b),
+            AttrValue::Bytes(b) => {
+                // Encode raw bytes as hex string for observability
+                let hex: String = b.iter().map(|byte| format!("{byte:02x}")).collect();
```
utpilla: Could we reduce the unnecessary allocations caused by `format!` and `collect()` by using manual lookup? Something like this:

```rust
const HEX: &[u8; 16] = b"0123456789abcdef";
AttrValue::Bytes(b) => {
    let mut hex = Vec::with_capacity(b.len() * 2);
    for &byte in b {
        hex.push(HEX[(byte >> 4) as usize]);
        hex.push(HEX[(byte & 0x0f) as usize]);
    }
    self.log_attrs.any_values_builder.append_str(&hex);
}
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -85,12 +205,34 @@ pub struct EtwEventData {
...
+    pub decoded_fields: Vec<DecodedField>,
+    /// Raw copy of the event's `UserData` payload.
+    ///
+    /// Provided for consumers that need access to the full uninterpreted
+    /// event bytes (e.g. for binary-level forwarding or custom decoding).
+    #[expect(
+        dead_code,
+        reason = "retained for future binary-level forwarding or manifest decoding"
+    )]
+    pub user_data: Vec<u8>,
```
utpilla: This is requiring a big allocation for every event.

Could we drop the field for now? And bring it back only when there is an ask for it? Or if we want to keep it we want to gate this behind a config flag (`capture_raw_payload: bool`, default `false`) and make it `Option<Vec<u8>>`.

rust/otap-dataflow/.gitignore
```diff
@@ -35,3 +35,7 @@ resolved-schema.yaml
 
 # Ignore the temporary repo dir created by `cargo xtask history` command
 history-temp-repo/
+
+# ETW receiver Parquet/durable-buffer output
+etw_output/
+etw_buffer/
```
utpilla: Do we need this in `.gitignore` or could it have been skipped locally?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -184,6 +326,192 @@ fn resolve_provider_guid(cfg: &ProviderConfig) -> Result<Guid, Error> {
...
+) -> Vec<DecodedField> {
+    let mut fields = Vec::with_capacity(format.fields().len());
+
+    for field in format.fields() {
+        // `try_get_field_data_closure` may panic with `todo!()` for
+        // unsupported LocationType variants (e.g. DynAbsolute).
+        // Since we're called from an `extern "system"` ETW callback
+        // that cannot unwind, we must catch any panic here to prevent
+        // the process from aborting.
+        let data = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
```
utpilla: Do we need `catch_unwind` here? Looks like this is guarding against a panic that can't actually happen via TDH:

- try_get_field_data_closure` itself doesn't panic, all its failure paths just return `None`.
- The `todo!()`s are in `get_data_with_offset_direct`, and only hit for `LocationType::DynRelative` / `DynAbsolute`.
- TDH's `intype_to_field_info` (in `one_collect/src/etw/tdh.rs`) only produces `Static`, `StaticString`, `StaticUTF16String`, and `StaticLenPrefixArray`. The `Dyn*` variants are a Linux tracefs thing (`__rel_loc`), not ETW.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -184,6 +326,192 @@ fn resolve_provider_guid(cfg: &ProviderConfig) -> Result<Guid, Error> {
...
+/// code unit and substituting U+FFFD for invalid surrogate pairs.
+fn decode_utf16le(data: &[u8]) -> String {
+    let u16_iter = data
+        .chunks_exact(2)
+        .map(|c| u16::from_ne_bytes([c[0], c[1]]))
+        .take_while(|&c| c != 0);
+    char::decode_utf16(u16_iter)
+        .map(|r| r.unwrap_or('\u{FFFD}'))
+        .collect()
+}
```
utpilla, line 492: Let's optimize this method considering it's called on the hot-path:

1. Pre-allocate the `String`: `.collect()` reallocs as it grows; a capacity hint avoids that.
2. ASCII fast path: Most ETW string fields (paths, identifiers, English log lines) are pure ASCII, where every high byte is 0. We can detect that with one scan and skip the surrogate-decode state machine entirely.
3. Name vs. behavior. Function is called `decode_utf16le` but uses `from_ne_bytes`. Correct on Windows (always LE), but `from_le_bytes` matches the name and compiles to the same code on LE hosts.
4. Trailing odd byte. chunks_exact(2) already discards it; rounding len down explicitly makes the intent obvious.

What are your thoughts on something like this?

```rust
fn decode_utf16le(data: &[u8]) -> String {
    // Round down to whole code units; ignore a trailing odd byte.
    let len = data.len() & !1;
    let bytes = &data[..len];

    // ASCII fast path: stop at first NUL or first non-ASCII code unit.
    let ascii_end = bytes
        .chunks_exact(2)
        .position(|c| c[0] == 0 || c[1] != 0)
        .map(|i| i * 2)
        .unwrap_or(len);

    if ascii_end == len {
        // All ASCII up to end-or-NUL: byte copy, no surrogate logic.
        let mut out = String::with_capacity(ascii_end / 2);
        for chunk in bytes[..ascii_end].chunks_exact(2) {
            out.push(chunk[0] as char);
        }
        return out;
    }

    // Mixed / non-ASCII: full UTF-16 decode.
    let mut out = String::with_capacity(len / 2);
    let u16_iter = bytes
        .chunks_exact(2)
        .map(|c| u16::from_le_bytes([c[0], c[1]]))
        .take_while(|&c| c != 0);
    out.extend(char::decode_utf16(u16_iter).map(|r| r.unwrap_or('\u{FFFD}')));
    out
}
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -283,28 +627,97 @@ fn spawn_etw_session(config: &Config, txs: Vec<mpsc::Sender<EtwEventData>>) -> R
...
-                    // available directly; event_id/version/level/keywords
-                    // come from the full_data bytes passed via EventData.
+                    // Extract event descriptor fields from the raw EVENT_RECORD.
+                    // AncillaryData exposes id/opcode/version directly; for
+                    // level and keywords we read from the EVENT_RECORD pointer.
+                    let event_id = anc.id();
+                    let opcode = anc.op_code();
+                    let version = anc.version();
+
+                    let (level, keywords) = match anc.record() {
```
utpilla: `anc.record()` is currently called multiple times in this callback. Each call returns the same `Option<&EVENT_RECORD>` We could simplify this and we can fold them into a single `if let Some(record) = anc.record()` and handle every record-derived field in one place.

```rust
let mut level = 0u8;
let mut keywords = 0u64;
let mut activity_id = [0u8; 16];
let mut event_name = String::new();
let mut decoded_fields = Vec::new();
let mut user_data = Vec::new();

if let Some(record) = anc.record() {
    level = record.EventHeader.EventDescriptor.Level;
    keywords = record.EventHeader.EventDescriptor.Keyword;

    let g = &record.EventHeader.ActivityId;
    activity_id[0..4].copy_from_slice(&g.data1.to_ne_bytes());
    activity_id[4..6].copy_from_slice(&g.data2.to_ne_bytes());
    activity_id[6..8].copy_from_slice(&g.data3.to_ne_bytes());
    activity_id[8..16].copy_from_slice(&g.data4);

    if !record.UserData.is_null() && record.UserDataLength != 0 {
        user_data = unsafe {
            std::slice::from_raw_parts(
                record.UserData as *const u8,
                record.UserDataLength as usize,
            )
        }
        .to_vec();
    }

    if let Ok(result) = decoder.borrow_mut().decode(record) {
        event_name = result.event_name.unwrap_or("").to_owned();
        decoded_fields = extract_decoded_fields(
            result.event_data.format(),
            result.event_data.event_data(),
        );
    }
}
```

This way we get some benefits:
- Drops the `(decoded_fields, event_name, user_data)` tuple shuffle.
- Collapses the two identical `TdhDecodeError` arms (`NotFound` and `_e` both produce empty defaults) into a single `if let Ok(...)`.
- Reads top-to-bottom: declare defaults, fill from the record once, build `EtwEventData`.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -184,6 +326,192 @@ fn resolve_provider_guid(cfg: &ProviderConfig) -> Result<Guid, Error> {
...
+        // StaticLenPrefixArray — `data` contains only the UTF-8 content bytes.
+        ("counted_string", _) => {
+            let s = String::from_utf8_lossy(data);
+            EtwAttributeValue::Str(s.trim_end_matches('\0').to_owned())
+        }
+        // Counted UTF-16 strings (TDH_INTYPE_COUNTEDSTRING, in_type 300).
+        ("counted_wstring", _) if data.len() >= 2 => EtwAttributeValue::Str(decode_utf16le(data)),
+        ("counted_wstring", _) => EtwAttributeValue::Str(String::new()),
+        // UTF-16LE strings, trim null terminator.
+        ("wstring", _) if data.len() >= 2 => EtwAttributeValue::Str(decode_utf16le(data)),
```
utpilla, line 422: We might not need the `>=2` check here since `decode_utf16le` would return and empty string for those cases anyway, right?

That would simplify it to:

```rust
("wstring" | "counted_wstring", _) => EtwAttributeValue::Str(decode_utf16le(data)),
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -184,6 +326,192 @@ fn resolve_provider_guid(cfg: &ProviderConfig) -> Result<Guid, Error> {
...
+        ("string", _) => {
+            let s = String::from_utf8_lossy(data);
+            EtwAttributeValue::Str(s.trim_end_matches('\0').to_owned())
+        }
+        // Counted ANSI/UTF-8 strings (TDH_INTYPE_COUNTEDANSISTRING, in_type 301).
+        // The u16 byte-count prefix has already been consumed by the framework's
+        // StaticLenPrefixArray — `data` contains only the UTF-8 content bytes.
+        ("counted_string", _) => {
+            let s = String::from_utf8_lossy(data);
+            EtwAttributeValue::Str(s.trim_end_matches('\0').to_owned())
```
utpilla: Minor: For invalid UTF-8 input this allocates twice: `from_utf8_lossy` allocates a `String`, then `to_owned` allocates a second one from the trimmed slice.

Trim the NULs off the byte slice first:

```rust
fn decode_ansi(data: &[u8]) -> String {
    let trimmed = data.split(|&b| b == 0).next().unwrap_or(data);
    String::from_utf8_lossy(trimmed).into_owned()
}
```

Then the two string arms collapse:

```rust
("string" | "counted_string", _) => EtwAttributeValue::Str(decode_ansi(data)),
```

One alloc instead of two on the invalid path, same single alloc on the ASCII path.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/arrow_records_encoder.rs
swashtek: Event name is not generic field across different receivers e.g. Syslog receiver, so parquet exporter would be taking this fix for ETW receiver. If that's fine, we can make a change (obviously there will be guards) 
I will create change once you confirm this.

swashtek: Please ignore my previous comment. I realized that optional OTAP column missing from incoming batch is materialized to null. Hence moving `event_name` to exporter makes more sense.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
swashtek: Agreed, don't need `catch_unwind` anymore as panic is unreachable.

rust/otap-dataflow/crates/core-nodes/src/exporters/parquet_exporter/schema.rs
utpilla, line 1: Let's create a separate PR for this.

swashtek: Reverted the change, will bring as a separate PR.

utpilla (approved):
Thanks @swashtek for this PR!

rust/otap-dataflow/.chloggen/etw-receiver-tdh-decoder.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: pipeline
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add TDH (Trace Data Helper) decoder for ETW tracelogging events in the ETW receiver
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [3187]
```
utpilla:
```suggestion
issues: [2783]
```

Let's use the issue instead.

Issue #2301, opened 2026-03-12
Continuous benchmark for low-memory use


jmacd:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Other

### Is your feature request related to a problem?

The continuous benchmarks measure loads that are meant for relatively large servers, proxies, telemetry gateways. We are interested in tests for a small OTAP Dataflow engine configuration, particularly one that runs on a single CPU.

### Proposed Solution

See https://github.com/open-telemetry/otel-arrow/pull/2212, a quick overview of how to test a low-memory configuration. At a minimum, we:

- Single-core engine setup
- Small PData channel size
- Backpressure
- Small concurrency factor
- Small admission control settings

### Alternatives Considered

We are struggling to test configurations that run on very small machines, think system-on-chip. We need reliable pipelines that run in 512MiB, we need reliable pipelines that run in 32MiB. As small as we can go, we should.

### Additional Context

When we have a small-memory configuration designed and tested, we should add it to the continuous benchmarks run nightly reporting total low-memory usage and performance.

Most importantly, the continuous benchmark harness should not require modifications to enable these tests. It is important to control all aspects of the test, including load generation, concurrency, blocking and dropping behavior at all levels of the pipeline, and to verify the end-to-end behavior of the system. Remember [Little's Law](https://en.wikipedia.org/wiki/Little%27s_law): we are bound by physics, careful not to measure self-induced load from self-diagnostic logging, and so on. 

Then, put it through an optimization cycle:

- Run the low-memory continuous benchmark under memory profiling
- Inspect the memory profile, looking for waste and or excessive allocations
- For large allocations, ensure a configuration knob exists to lower that configuration
- Repeat the process

PR #3195, opened 2026-06-03
Add consistent memory limits, concurrency, and queue sizes to OTC configs

jmacd:
# Change Summary

This sets explicit memory limits, concurrency, and queue sizes for the OTC being compared with DFE in our compare site.

Part of https://github.com/lquerel/opentelemetry.io/pull/1
Part of #2301

We have always envisioned the OTLP and OTAP protocols being served on one port, which the Phase 1 components do. This changes the study to use the Phase-1 `otelarrowreceiver` as the OTLP receiver because it applies consistent/unified admission limits to both streams and otherwise behaves like the core OTLP receiver (it was forked from). This will reduce the ~10x memory explosion that we've measured for overloaded OTC (with less impact on performance than memorylimiter has).

Other notes:

Sets admission control to 128MiB in flight w/ 32MiB waiting before fail-fast (_this is the current default, which formerly applied to OTC OTAP receivers in the dashboard, makes it apply to OTC OTLP receivers consistently_)
Sets export/stream concurrency to 5 to match the DFE settings.
Sets queue_size to 100 to match DFE settings.
For OTC/OTAP exporters, use num_streams equal to 5; set num_consumers to 100 i.e. queue_size.
For OTC/OTLP exporters, use num_consumers equal to 5.


rust/otap-dataflow/.chloggen/otelcol-compare-settings.yaml
```diff
@@ -0,0 +1,23 @@
...
+issues: [2301]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  This is to place a realistic limit that is above what 1 CPU can encode/decode for
+  OTC OTLP and OTAP tests, which have demonstrated the effectiveness of the Phase 1
+  request-bytes admission controller. That receiver makes the OTLP and OTAP tests
+  comparable, will reduce ~10x memory used in the overloaded OTLP tests.
```
JakeDern, line 23: I would maybe treat this as a chore or revise this subtext

tools/pipeline_perf_test/test_suites/comparison_dashboard/templates/otelcol/otlp-otlp.yaml.j2
```diff
@@ -1,8 +1,11 @@
 receivers:
-  otlp:
+  otelarrow:
     protocols:
       grpc:
         endpoint: 0.0.0.0:4317
+    admission:
+      request_limit_mib: {{ admission_request_mib | default(128) }}
+      waiting_limit_mib: {{ admission_waiting_mib | default(32) }}
```
JakeDern, line 8: Just curious - Do these admission limits take into account the process memory or just put limits on some buffer?

tools/pipeline_perf_test/test_suites/comparison_dashboard/templates/otelcol/otlp-attr-rename.yaml.j2
```diff
@@ -1,8 +1,11 @@
 receivers:
-  otlp:
+  otelarrow:
```
JakeDern, line 2: I did notice different performance characteristics using the otelarrow receiver vs the otlp receiver in the saturation tests we were discussing.

I'm cool with this change especially since it looks like we were already using it for the continuous benches, but we might want to justify somewhere why we're using this over the standard otlp receiver in our benchmarks.

tools/pipeline_perf_test/test_suites/comparison_dashboard/templates/otelcol/otap-otap.yaml.j2
```diff
@@ -13,12 +16,18 @@ exporters:
     arrow:
       payload_compression: {{ payload_compression | default("none") }}
```
utpilla, line 17: Not related to this PR, but OTAP runs don't seem to be using `payload_compression`. Is that intentional? 

It still keeps the OTC-vs-DFE comparison fair but might understate OTAP's compression effeciency OTAP-vs-OTLP comparison.

tools/pipeline_perf_test/test_suites/comparison_dashboard/templates/otelcol/otlphttp-otlphttp.yaml.j2
```diff
@@ -15,6 +15,10 @@ exporters:
       block_on_overflow: {{ block_on_overflow | default(true) }}
       sizer: {{ sending_queue_sizer | default("requests") }}
       queue_size: {{ sending_queue_size | default(100) }}
+      num_consumers: {{ sending_queue_otlp_consumers | default(5) }}
```
ThomsonTan, line 18: This template keeps the core `otlphttp` receiver with no admission control, so it's the one scenario without the 128/32 MiB cap. Is that intentional?

PR #3197, opened 2026-06-04
chore: Intentionally break internal Event to test Weaver guards

cijothomas:
Testing out if weaver is catching these!


cijothomas:
image
Okay! It di fail. Worth noting that copilot also pointed out this - https://github.com/open-telemetry/otel-arrow/pull/3197#discussion_r3352844644

Closing, as this was done just to validate if weaver if catching it properly and to improve weaver tooling upstream!

PR #3229, opened 2026-06-05
feat(otap): preserve OtapPdata::peer_addr across merging processors

### LINKED ISSUE #3228: Preserve `OtapPdata::peer_addr` across merging processors
### Pre-filing checklist

### Component(s)

### Description

# Preserve `OtapPdata::peer_addr` across merging processors

Follow up to #3220

## Component(s)

Rust OTAP dataflow (`rust/otap-dataflow/`)

## Is your feature request related to a problem?

[#3220](https://github.com/open-telemetry/otel-arrow/issues/3220) added a
receiver-observed peer address on `OtapPdata` and wired it through every
receiver that owns a real socket. Receivers do their part: an inbound batch
arrives with `pdata.peer_addr() == Some(addr)`.

The follow-up problem is on the processor side.
Two processors — batch_processor and temporal_reaggregation_processor —
emit derived output by combining one or more inputs into a fresh
`OtapPdata` built with a brand-new `Context`. On those paths the peer
address set by the receiver is dropped, so any consumer placed downstream
of one of these processors will observe `peer_addr() == None` even though
the originating socket was known one hop earlier.

An audit of every processor in `crates/core-nodes/src/processors/` and
`crates/contrib-nodes/src/processors/` produced the following matrix:

| Processor | Emit pattern | `peer_addr` preserved? |
| --- | --- | --- |
| `attributes_processor` | reuses input `Context` | ✅ |
| `batch_processor` (flush path, [mod.rs#L1053](../../rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs#L1053)) | `OtapPdata::new(Context::default(), records.into())` | ❌ **MERGE** |
| `condense_attributes_processor` | reuses input `Context` | ✅ |
| `content_router` | passthrough | ✅ |
| `debug_processor` | sink only | n/a |
| `delay_processor` | passthrough | ✅ |
| `durable_buffer_processor` | reuses input `Context` | ✅ |
| `fanout_processor` | clones input pdata per fanout port | ✅ |
| `filter_processor` | reuses input `Context` | ✅ |
| `log_sampling_processor` | reuses input `Context` | ✅ |
| `recordset_kql_processor` | reuses input `Context` | ✅ |
| `resource_validator_processor` | validator only | n/a |
| `retry_processor` | requeues original pdata | ✅ |
| `signal_type_router` | passthrough | ✅ |
| `temporal_reaggregation_processor` (subscriber path, [mod.rs#L673](../../rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs#L673)) | `OtapPdata::new(Context::with_capacity(1), …)` | ❌ **MERGE** |
| `transform_processor` | reuses input `Context` | ✅ |

Two processors fall into the **MERGE** bucket. Both also rebuild
`transport_headers` from scratch (they have the same drop bug — preserved
here for symmetry, even though `transport_headers` is not the focus of this
issue):

- `batch_processor` — combines N input requests into one output batch on
  flush.
- `temporal_reaggregation_processor` — combines metrics points from many
  inputs over a time window into one aggregate output.

## Proposed Solution

For each merging processor, decide on the correct merge semantics for
request-scoped transport facts and apply them consistently. The decision
factors are:

1. **Did all merged inputs share the same `peer_addr`?**
   If yes, forward that address on the output. Single-peer streams (the
   common case in tail-sampling-style deployments where one collector
   instance owns one client) keep working as if no merge happened.

2. **Did the merged inputs come from different peers?**
   Then there is no single originating peer to attribute the output to.
   Leave `peer_addr` as `None`. Downstream processors that depend on
   `peer_addr` will naturally degrade to "unknown peer" instead of
   accidentally tagging the batch with one arbitrary peer's address.

A minimal helper on `Context` (or `OtapPdata`) can encapsulate the rule —
roughly:

```rust
/// Returns `Some(addr)` only when every contributing input observed the
/// same peer address; returns `None` if any input was peerless or if
/// distinct peers contributed.
fn merge_peer_addr<'a, I: IntoIterator<Item = Option<SocketAddr>>>(
    inputs: I,
) -> Option<SocketAddr>
```

Processors then accumulate the peer addresses of contributing inputs
alongside the data they merge and feed them through this helper when
constructing the output `Context`.

### Required properties

- **Opt-in cost only**: processors that do not merge (the majority — 12 of
  the 14 functional processors today) pay nothing because they already reuse
  the input `Context`.
- **No silent attribution errors**: never tag a merged batch with one
  arbitrary contributor's `peer_addr`.
- **Symmetric with `transport_headers`**: the same merging policy should
  apply to other request-scoped facts living on `Context`. Solving it once
  on the merge helper covers both.

### Non-goals

- Designing per-record (rather than per-batch) attribution. If a use case
  needs that, it belongs on the payload, not on the request envelope.
- Changing receiver behavior. Receivers already do the right thing after
  [#3220](https://github.com/open-telemetry/otel-arrow/issues/3220).

## Alternatives Considered

- **Always forward the first contributor's `peer_addr`.** Rejected:
  silently misattributes multi-peer batches.
- **Always drop `peer_addr` in merging processors.** Rejected: makes the
  common single-peer case (where the answer is unambiguous) useless to
  downstream processors.
- **Tag every output record with its originating peer.** Out of scope here;
  belongs in payload-level enrichment.

## Additional Context

This issue is the follow-up flagged in a [comment](https://github.com/open-telemetry/otel-arrow/pull/3225#issuecomment-4635855843) in PR
[#3225](https://github.com/open-telemetry/otel-arrow/pull/3225). It does not
block #3225, which is the wire-level propagation across receivers. The
present issue covers the processor-side preservation needed before
downstream consumers (notably the k8sattributes processor) can rely on
`peer_addr` end-to-end.

### Additional Context

_No response_

pritishnahar95:
# Change Summary

Preserve `OtapPdata::peer_addr` across the two processors that previously
dropped it when merging multiple inputs into one output batch
(`batch_processor` and `temporal_reaggregation_processor`).

Single-peer streams now retain the receiver-observed peer address end-to-end.
Multi-peer merges intentionally keep `peer_addr` as `None` to avoid attributing
a merged batch to one arbitrary contributor.

- Closes #3228

- Unit test for the merge rule (empty / single / all-equal / any-`None` /
  distinct contributors).
- Three end-to-end tests per processor covering: single peer preserved,
  mixed peers dropped, peerless contributor drops. For
  `temporal_reaggregation_processor` an additional test covers the
  partial-passthrough subscriber path.

Yes. Processors that consult `OtapPdata::peer_addr()` now observe the correct 
peer address downstream of `batch_processor` and `temporal_reaggregation_processor` 
for single-peer streams instead of always seeing `None`.

### Changelog

- [x] Added `rust/otap-dataflow/.chloggen/preserve-peer-address-merging-processors.yaml`.

### DIFF
--- a/rust/otap-dataflow/.chloggen/preserve-peer-address-merging-processors.yaml
+++ b/rust/otap-dataflow/.chloggen/preserve-peer-address-merging-processors.yaml
@@ -1,0 +1,26 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: pipeline
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Preserve `OtapPdata::peer_addr` across merging processors (`batch_processor`, `temporal_reaggregation_processor`) by merging contributing inputs' peer addresses via the new `Context::merge_peer_addr` helper.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [3228]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Adds `Context::merge_peer_addr`, which returns `Some(addr)` only when every
+  contributing input observed the same peer address and `None` otherwise.
+  `batch_processor` and `temporal_reaggregation_processor` — the two processors
+  that rebuild `OtapPdata` with a fresh `Context` when merging multiple inputs
+  — now apply this helper so single-peer streams retain the receiver-observed
+  peer address end-to-end, while multi-peer merges keep `peer_addr` as `None`
+  to avoid misattribution.
--- a/rust/otap-dataflow/.chloggen/preserve-peer-address-merging-processors.yaml
+++ b/rust/otap-dataflow/.chloggen/preserve-peer-address-merging-processors.yaml
@@ -19,7 +19,7 @@ issues: [3228]
 subtext: |
   Adds `Context::merge_peer_addr`, which returns `Some(addr)` only when every
   contributing input observed the same peer address and `None` otherwise.
-  `batch_processor` and `temporal_reaggregation_processor` — the two processors
+  `batch_processor` and `temporal_reaggregation_processor` - the two processors
   that rebuild `OtapPdata` with a fresh `Context` when merging multiple inputs
   — now apply this helper so single-peer streams retain the receiver-observed
   peer address end-to-end, while multi-peer merges keep `peer_addr` as `None`
--- a/rust/otap-dataflow/.chloggen/preserve-peer-address-merging-processors.yaml
+++ b/rust/otap-dataflow/.chloggen/preserve-peer-address-merging-processors.yaml
@@ -21,6 +21,6 @@ subtext: |
   contributing input observed the same peer address and `None` otherwise.
   `batch_processor` and `temporal_reaggregation_processor` - the two processors
   that rebuild `OtapPdata` with a fresh `Context` when merging multiple inputs
-  — now apply this helper so single-peer streams retain the receiver-observed
+  - now apply this helper so single-peer streams retain the receiver-observed
   peer address end-to-end, while multi-peer merges keep `peer_addr` as `None`
   to avoid misattribution.
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
@@ -48,7 +48,7 @@ use otap_df_engine::{
 };
 use otap_df_otap::OTAP_PROCESSOR_FACTORIES;
 use otap_df_otap::accessory::slots::{Key as SlotKey, State as SlotState};
-use otap_df_otap::pdata::{Context, OtapPdata};
+use otap_df_otap::pdata::{Context, OtapPdata, PeerAddrMerger};
 use otap_df_pdata::TryIntoWithOptions;
 use otap_df_pdata::{
     OtapArrowRecords, OtapPayload, OtapPayloadHelpers, OtlpProtoBytes, error::Error as PDataError,
@@ -1439,7 +1439,16 @@ where
         let remaining = output_batches.pop().expect("has last");
         let last_input = from_inputs.context.last().expect("has last");
         let last_weight = sizer.batch_size(&remaining).expect("known size");
-        let new_part = BatchPortion::new(last_input.inkey, last_input.peer_addr, last_weight);
+        // The retained partial output may aggregate records from multiple
+        // input requests, but the existing context bookkeeping only tracks
+        // input portions, not which inputs contributed to which output
+        // batch. Reusing `last_input.peer_addr` here would be wrong when
+        // inputs from multiple peers fed the final output batch: it would
+        // misattribute the retained portion (and any future merge that
+        // consumes it) to one arbitrary contributor. Conservatively drop
+        // the address — merge_peer_addr will then propagate `None`, the
+        // safe default.
+        let new_part = BatchPortion::new(last_input.inkey, None, last_weight);
 
         from_inputs.weight -= last_weight;
 
@@ -1461,8 +1470,9 @@ where
         let mut out = Vec::new();
         // Track every contributing portion's peer_addr, even ones that do not
         // route ack/nack, so the merged output `peer_addr` is correct when
-        // some (or all) inputs had no subscribers.
-        let mut peers: Vec<Option<SocketAddr>> = Vec::new();
+        // some (or all) inputs had no subscribers. Folded incrementally to
+        // avoid allocating a `Vec` on every flush.
+        let mut peer_merger = PeerAddrMerger::new();
 
         while weight > 0 && contexts.pos < contexts.inputs.len() {
             let bp = contexts.inputs.get_mut(contexts.pos).expect("valid");
@@ -1471,7 +1481,7 @@ where
             bp.weight -= take;
             weight -= take;
             let peer_addr = bp.peer_addr;
-            peers.push(peer_addr);
+            peer_merger.push(peer_addr);
 
             if bp.weight == 0 {
                 contexts.pos += 1;
@@ -1486,7 +1496,7 @@ where
             }
         }
 
-        let merged_peer = Context::merge_peer_addr(peers);
+        let merged_peer = peer_merger.finish();
         let routed = (!out.is_empty()).then_some(out);
         (routed, merged_peer)
     }
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
@@ -1439,16 +1439,24 @@ where
         let remaining = output_batches.pop().expect("has last");
         let last_input = from_inputs.context.last().expect("has last");
         let last_weight = sizer.batch_size(&remaining).expect("known size");
-        // The retained partial output may aggregate records from multiple
-        // input requests, but the existing context bookkeeping only tracks
-        // input portions, not which inputs contributed to which output
-        // batch. Reusing `last_input.peer_addr` here would be wrong when
-        // inputs from multiple peers fed the final output batch: it would
-        // misattribute the retained portion (and any future merge that
-        // consumes it) to one arbitrary contributor. Conservatively drop
-        // the address — merge_peer_addr will then propagate `None`, the
-        // safe default.
-        let new_part = BatchPortion::new(last_input.inkey, None, last_weight);
+        // Compute the retained portion's peer_addr from the input portions
+        // that actually contributed to this output batch. Inputs are emitted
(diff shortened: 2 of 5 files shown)


rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
```diff
@@ -1415,7 +1439,7 @@ where
         let remaining = output_batches.pop().expect("has last");
         let last_input = from_inputs.context.last().expect("has last");
         let last_weight = sizer.batch_size(&remaining).expect("known size");
-        let new_part = BatchPortion::new(last_input.inkey, last_weight);
+        let new_part = BatchPortion::new(last_input.inkey, last_input.peer_addr, last_weight);
```
lalitb: can you check the logic once -   I think this can misattribute a retained partial batch. The leftover batch can contain data from multiple input peers after rebatching, but we store only the last input's peer_addr on it. If that leftover is merged later, mixed-peer data can incorrectly become `Some(last_peer)` instead of `None`

rust/otap-dataflow/crates/core-nodes/src/processors/temporal_reaggregation_processor/mod.rs
```diff
@@ -240,6 +241,13 @@ pub struct TemporalReaggregationProcessor {
     /// flushed.
     pending_flush: Vec<CallData>,
 
+    /// Receiver-observed peer addresses of every input whose data contributed
+    /// to [Self::builder]. Tracked independently of subscriber state so that
+    /// flushes from no-subscriber inputs can also forward `peer_addr` (via
+    /// [`Context::merge_peer_addr`]) when all contributing inputs agreed.
+    /// Drained on every flush alongside the builder.
+    aggregated_peer_addrs: Vec<Option<SocketAddr>>,
```
lalitb: This keeps one peer entry per aggregating input for the whole window, but flush only needs the final merged peer result. Could we track this incrementally instead of storing a ` Vec`? That would avoid growth across large aggregation windows and the realloc after `mem::take`.

pritishnahar95: good catch, addressed, thanks!

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
pritishnahar95: good catch, addressed

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
```diff
@@ -1415,7 +1439,16 @@ where
...
+        // The retained partial output may aggregate records from multiple
+        // input requests, but the existing context bookkeeping only tracks
+        // input portions, not which inputs contributed to which output
+        // batch. Reusing `last_input.peer_addr` here would be wrong when
+        // inputs from multiple peers fed the final output batch: it would
+        // misattribute the retained portion (and any future merge that
+        // consumes it) to one arbitrary contributor. Conservatively drop
+        // the address — merge_peer_addr will then propagate `None`, the
+        // safe default.
+        let new_part = BatchPortion::new(last_input.inkey, None, last_weight);
```
lalitb: I think this may be too conservative. A retained partial from a single peer gets requeued with `peer_addr = None`, so a later same-peer merge can lose a valid peer address. Can we preserve `last_input.peer_addr` here or cover this retained-partial case with a test?

utpilla: Agree the single-peer case is worth fixing. Just flagging that any fix here needs to handle three cases:

- **Leftover from one input (one input got split across two output batches):** the peer is clear and we should keep it.
- **Leftover spanning multiple inputs with different peers:** mixed-peer data, so it must be None to avoid misattribution.
- **Leftover spanning multiple inputs that happen to share the same peer:** we can keep that shared peer.

So `last_input.peer_addr` alone isn't enough, since case 2 would mislabel mixed-peer data with the last contributor's peer ([the original bug](https://github.com/open-telemetry/otel-arrow/pull/3229#discussion_r3376587332)). Worth covering all three with tests too.

rust/otap-dataflow/crates/otap/src/pdata.rs
```diff
@@ -355,6 +373,63 @@ impl Context {
...
+/// Use when contributing inputs are observed one-at-a-time as part of an
+/// existing pass over the input list (e.g. a processor's flush loop). Push
+/// each input's `peer_addr` exactly once with [`PeerAddrMerger::push`], then
+/// call [`PeerAddrMerger::finish`] to obtain the merged result.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct PeerAddrMerger {
+    candidate: Option<SocketAddr>,
+    started: bool,
+    poisoned: bool,
+}
```
utpilla: Consider modeling `PeerAddrMerger` as a three-variant enum instead of `{ candidate: Option<SocketAddr>, started: bool, poisoned: bool }`. The merger really only has three states: empty, one agreed-upon address, or poisoned.

```rust
pub enum PeerAddrMerger {
    Empty,
    Single(SocketAddr),
    Poisoned,
}
```

This is one of the places Rust's type system really pays off, so worth using:

1. **Illegal states become unrepresentable:** The current struct allows four `(candidate, poisoned)` combinations, but only three are valid (`Some(_) + poisoned=true` is invalid). The code carefully avoids that state at runtime, but the type doesn't enforce it. With the enum, the invalid state can't be constructed at all.

2. **The merge rule reads directly from the variants:** push becomes a single exhaustive match over `(self, peer_addr)` that mirrors the rule line-for-line. No `started` bookkeeping, no if `self.poisoned { return; }` guard, no risk of forgetting to clear candidate when poisoning. And match exhaustiveness means the compiler catches any missed case if the rule ever evolves.

The enum based approach just leans on the type system to do the work the current code does by hand.

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
pritishnahar95: addressed

rust/otap-dataflow/crates/otap/src/pdata.rs
pritishnahar95: addressed

rust/otap-dataflow/.chloggen/preserve-peer-address-merging-processors.yaml
```diff
@@ -0,0 +1,26 @@
...
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [3228]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Adds `Context::merge_peer_addr`, which returns `Some(addr)` only when every
+  contributing input observed the same peer address and `None` otherwise.
+  `batch_processor` and `temporal_reaggregation_processor` — the two processors
```
utpilla: Use ASCII hyphen character instead of non-ASCII em dashes.

```suggestion
  `batch_processor` and `temporal_reaggregation_processor` - the two processors
```

rust/otap-dataflow/.chloggen/preserve-peer-address-merging-processors.yaml
```diff
@@ -0,0 +1,26 @@
...
+  that rebuild `OtapPdata` with a fresh `Context` when merging multiple inputs
+  — now apply this helper so single-peer streams retain the receiver-observed
```
utpilla:
```suggestion
  - now apply this helper so single-peer streams retain the receiver-observed
```

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
```diff
@@ -1433,15 +1474,22 @@ where
         &mut self,
         mut weight: usize,
         contexts: &mut MultiContext,
-    ) -> Option<Vec<BatchPortion>> {
+    ) -> (Option<Vec<BatchPortion>>, Option<SocketAddr>) {
```
utpilla, line 1477: (non-blocking, design thought)

This PR adds the first instance of carrying a `Context` field across the merge boundary in `batch_processor` and `temporal_reaggregation_processor`. Everything else still resets to `Context::default()` after a merge. The current shape (extra tuple slot in `drain_context`, extra `peer_addr` field on `BatchPortion`, parallel field on the temporal aggregator) works well for a single field.

If/when a second `Context` field needs the same treatment, it might be worth exploring a more extensible design so that each new mergeable field doesn't require touching every call site in the same pattern. Not something to act on now since there's no concrete second case today, and the right shape will be clearer once one appears. Just flagging it for awareness.

utpilla (approved):

rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor/mod.rs
pritishnahar95: Good point, thanks for raising this!

lalitb:
Quick scope question: since this PR closes #3228, should we open a follow-up issue to track the `transport_headers` part mentioned there, so someone can address it later? This PR fixes `peer_addr`, but the same fresh `Context` merge paths still seem to drop headers.

lalitb (approved):
LGTM. Thanks for the PR.

pritishnahar95:
> Quick scope question: since this PR closes #3228, should we open a follow-up issue to track the `transport_headers` part mentioned there, so someone can address it later? This PR fixes `peer_addr`, but the same fresh `Context` merge paths still seem to drop headers.

@lalitb `transport_headers` is not covered by #3228. We should file a separate issue for it
See this note in #3228 
image

PR #3232, opened 2026-06-07
feat(parquet-exporter): add object store retry config and flush metrics

### LINKED ISSUE #3052: [Parquet Exporter] Add retry configuration and flush health metrics
### Pre-filing checklist

### Component(s)

The Parquet exporter currently relies on the underlying object-store client behavior for transient write failures. Operators do not have an explicit exporter-level retry configuration surface or enough metrics to understand archive write health, such as whether files are being flushed, closed, and uploaded successfully.

### Proposed Solution

This issue proposes making object-store-backed Parquet exports more production-operable by:
- Exposing optional retry settings in the Parquet exporter config.
- Supporting knobs such as max retries, initial backoff, max backoff, backoff
  multiplier, and retry timeout where supported by the object-store backend.
- Adding flush/write health counters, for example:
  - `flush_attempts`
  - `flush_success`
  - `flush_failures`
- Preserving backward compatibility by making retry config optional.
- Documenting how these metrics should be interpreted.

### Why this is useful:

Object-store writes can fail transiently due to network timeouts, 5xx responses, throttling, or temporary service issues. Operators need both controlled retry behavior and visibility into whether Parquet files are being successfully finalized.

### Alternatives Considered

- **Rely on the generic retry processor**
    This does not fully cover Parquet/object-store write failures because the exporter may fail while flushing, closing, or uploading files after it has already accepted telemetry into exporter state. 
- **Rely only on object-store client defaults**
    This keeps the exporter simple, but operators do not get an explicit config surface or clear metrics showing whether final file writes are succeeding, retrying, or failing.
- **Handle retries only outside the engine**
    External retry or restart logic can help with process-level failures, but it cannot provide exporter-level visibility into individual flush/write attempts and final archive health.

### Additional Context

- This is generic exporter operability work and is not tied to any specific deployment model.
- Pipeline-level retry does not replace exporter/object-store retry because the failure happens while the exporter finalizes and uploads files after accepting telemetry into exporter state.

lalitb:
## Change Summary

  Adds optional object-store retry settings to the Parquet exporter for cloud-backed storage such as S3 and Azure Blob.

  The retry settings are passed through to the `object_store` crate. This does not add Parquet writer-level retry loops.

  Also adds flush outcome metrics for file flush attempts, successes, and failures.

  * closes #3052 

  * Added unit tests for retry config deserialization, defaults, validation, and translation to `object_store::RetryConfig`.
  * Added tests for local file storage behavior with absent, valid, and invalid retry config.
  * Added AWS/Azure builder smoke tests behind their feature flags.
  * Added writer tests for flush attempt/success/failure attribution, including concurrent close failures.
  * Ran `cargo xtask check`.

  Yes. Parquet exporter configs can now include an optional `retry` block, and the exporter now reports flush attempt/success/failure metrics.

  Example:

  ```yaml
  retry:
    max_retries: 10
    init_backoff: "200ms"
    max_backoff: "30s"
    backoff_base: 2.0
    retry_timeout: "2min"
```

  ### Changelog

### CODE REVIEW

File: rust/otap-dataflow/crates/otap/src/object_store.rs [hunk 1]
```diff
@@ -14,11 +15,148 @@ use crate::cloud_auth;
...
+use object_store::{BackoffConfig, RetryConfig};
 
 /// Azure object storage
 #[cfg(feature = "azure")]
 pub mod azure;
 
+/// Retry settings for object-store-backed storage.
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
+#[serde(try_from = "RetryOptionsUnchecked")]
+pub struct RetryOptions {
```
albertlockett, line 28: this is a nice pattern

albertlockett (approved):
nice work @lalitb , LGTM

PR #3233, opened 2026-06-07
Fix gRPC DNS startup check proxy selection

lalitb:
## Change Summary

Fixes the gRPC `startup_check: dns` path so it skips local DNS resolution only when the target endpoint would actually use a configured proxy.

Previously, the check skipped whenever any proxy was configured. That was too broad: for example, an `http://` endpoint does not use `HTTPS_PROXY`, so local DNS is still required unless `HTTP_PROXY` or `ALL_PROXY` applies.

  * Part of #3222

  * `cargo test -p otap-df-otap startup_check_dns_not_skipped_when_proxy_does_not_apply_to_endpoint_scheme`
  * `cargo check -p otap-df-otap --features crypto-ring`
  * `python3 tools/sanitycheck.py`
  * `cargo xtask check`

  Yes. `startup_check: dns` now matches the real proxy selection path and catches DNS failures when a configured proxy does not apply to the endpoint scheme.

  ### Changelog

### DIFF
--- a/rust/otap-dataflow/.chloggen/grpc-dns-proxy-selection.yaml
+++ b/rust/otap-dataflow/.chloggen/grpc-dns-proxy-selection.yaml
@@ -1,0 +1,19 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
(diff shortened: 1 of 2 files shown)


albertlockett (approved):
Thanks for this fix @lalitb !

Issue #3094, opened 2026-05-26
Kubernetes attributes processor


jmacd:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

This is widely needed for our OTAP pipeline in particular file log and journald receivers:
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/k8sattributesprocessor/README.md

### Proposed Solution

Implement this in Rust -- it is high value.

### Alternatives Considered

None considered.

### Additional Context

https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/ai/ai-assisted-component-development.md

pritishnahar95:
I would like to take up this issue

PR #3256, opened 2026-06-09
doc: Add design doc for Resource Enricher Processor

pritishnahar95:
# Change Summary

* Addresses #3094 

### Changelog


rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,889 @@
...
+The Kubernetes attributes processor enriches OTAP telemetry with resource
+attributes derived from Kubernetes objects (pods, namespaces, nodes, and
+optionally deployments / statefulsets / daemonsets / jobs / replicasets /
+cronjobs). It associates each inbound resource with a pod using a small,
+ordered set of association rules, then projects configured metadata, labels,
+and annotations onto the resource's attribute set.
+
+The processor responsibility is split into two pieces:
+
+1. **`k8s_metadata_extension`** -- an active, shared extension that owns the
```
gouslu: Would that make sense to have a local optimization for it as well? Extensions can provide both local and shared implementations, if it makes sense to have a local version of it as well, for optimization, it is worth noting here. It can also be a TBD.

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,889 @@
...
+   Kubernetes client, runs the watch/informer event loop, maintains the
+   in-memory metadata cache, and exposes a `KubernetesMetadataLookup`
+   capability.
+2. **`k8s_attributes_processor`** -- an inline single-route processor that
+   resolves each inbound `OtapPdata` resource to a pod via the capability and
+   writes the configured attributes onto Arrow resource attribute batches.
+
+The split is forced by two engine constraints that have no Go analog:
+
+- **Per-core duplication.** Processors are instantiated per pipeline
```
gouslu: I am a little bit confused, just because this bullet point seems to imply that a shared extension would be sharing the data across the processors. Today that is not possible. What it means is that, a shared extension has a "Send" boundary, which makes it share-able but not necessarily shared, whereas a local extension does not have "Send" boundary. What makes an extension instance shared across cores or pipeline is the declaration scope of it. Plan is to make it possible to declare it at the pipeline group or engine level, so that it can be shared across pipelines, which is only possible to do with a shared extension. I also think that a shared extension that is regularly updating the cache for attributes would require some sort of lock or synchronization primitive or some atomic value to ensure that there is no race condition. So, in hot path this may have implications for performance, just to keep in mind.

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,889 @@
...
+
+1. **`k8s_metadata_extension`** -- an active, shared extension that owns the
+   Kubernetes client, runs the watch/informer event loop, maintains the
+   in-memory metadata cache, and exposes a `KubernetesMetadataLookup`
+   capability.
+2. **`k8s_attributes_processor`** -- an inline single-route processor that
+   resolves each inbound `OtapPdata` resource to a pod via the capability and
+   writes the configured attributes onto Arrow resource attribute batches.
+
+The split is forced by two engine constraints that have no Go analog:
```
gouslu: Having this as an extension is also good for some other components where we might need it. For example, azure monitor exporter can make use of it for certain internal telemetry it emits.

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,889 @@
...
+  row in [Core Decisions](#core-decisions) for the lookup guarantees.
+
+The extension boundary is where blocking and async work belongs; the
+processor stays a synchronous reader of a published snapshot.
+
+## Goals
+
+The v1 capability must:
+
+- enrich logs, metrics, and traces with the default Kubernetes resource
```
gouslu: Does it mean that the "KubernetesMetadataLookup" will enrich the data we pass to it, or will it just provide the data so that processor can do the enrichment? Latter seems to me better for re-usability purposes.

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,889 @@
...
+  data path -- the surface exposes no such method, and the processor crate
+  has no direct Kubernetes client dependency. See the "Hot-path contract"
+  row in [Core Decisions](#core-decisions) for the lookup guarantees.
+
+The extension boundary is where blocking and async work belongs; the
+processor stays a synchronous reader of a published snapshot.
+
+## Goals
+
+The v1 capability must:
```
gouslu: if there is a need for capability to provide readiness signal, so that by the time the pipeline is running the data is already available and cached, that is a pending work. Just noting it here, because it seems like it might be useful in this case.

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,889 @@
...
+  remains bounded by the pod count and `pod_delete_grace_period`.
+
+## Open Questions
+
+1. **Cross-pipeline extension sharing.** The Phase 1 extension system
+   instantiates extensions per pipeline instance (per core). The
+   process-local shared backend in
+   [Thread-per-Core Integration](#thread-per-core-integration) gives
+   one-watcher-per-process semantics inside that constraint. Is this an
+   acceptable interim shape, or should the extension system grow an
```
gouslu: I don't think we need to wait for it to grow to engine-scope, but I think that a local optimization for the extension is definitely needed. If a node asks for local extension, but local is not available for the context, shared is provided as a fallback. This ensures that if extension is defined in the pipeline scope require_local will give the local version and provide necessary optimizations, and when extension is defined in higher scopes, shared fallback will be given.

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,889 @@
...
+   to the semantic conventions' "how to compute" notes, which themselves
+   have known ambiguity. Confirm this is acceptable as a v2 follow-up.
+4. **Composite-key index size under user-defined sources.** A user who
+   adds a high-cardinality attribute (e.g. `service.name`) as an
+   association source will multiply the index size proportionally and
+   may hit the documented caveat that multiple pods can share that
+   value. v1 keeps one shared metadata entry per shape and lets the last
+   write win. Should we instead reject high-cardinality sources with a
+   `last_writer_wins_warning`?
+5. **Readiness-probe hook in the extension framework.** Implementing
```
gouslu: We should add the readiness probe. It is a low hanging fruit, and can be a pre-req for this task.

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,889 @@
...
+Kubernetes watcher per core would multiply both API server load and cache
+memory, neither of which is acceptable for this component.
+
+The extension declares itself `Active + Shared` and keeps its per-core
+state empty. All persistent state -- the Kubernetes client, the per-kind
+reflector stores, the published lookup snapshot (see
+[`KubernetesMetadataLookup` capability](#kubernetesmetadatalookup-capability)),
+and the pod-delete grace queue -- lives behind a single
+`Arc<KubernetesState>` held in a process-shared
+`OnceLock<Arc<KubernetesState>>` keyed by the extension config's hash.
```
lalitb: I think this needs a clearer ownership model. A single `OnceLock` cannot really be keyed by config hash unless there is another registry around it. Also, if we support more than one `k8s_metadata` extension with different scopes, we need to make sure they do not accidentally share the same backend. Can we spell out whether this is a keyed registry, and how entries are cleaned up when the extension shuts down?

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,889 @@
...
+Per-core extension instances each hold an `Arc::clone` of that backend;
+the per-core wrapper itself is empty.
+
+Initialization is gated by a `tokio::sync::OnceCell` inside the shared
+`KubernetesState`. Inside `start()`, the first per-core extension to
+reach the cell builds the Kubernetes client, constructs the watchers,
+and spawns the background task on the core that called it; subsequent
```
lalitb: This part feels risky to me. If the shared watcher is started by whichever core happens to initialize first, then the backend lifecycle is tied to an arbitrary per-core extension instance. It is not clear what happens if that instance shuts down or restarts while other cores still use the shared cache. I think the watcher needs a clear owner, either the extension lifecycle itself or an explicit engine/group-scoped owner.

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,889 @@
...
+## Kubernetes Watch Model
+
+- One watcher per resource kind, using streaming initial lists and watch
+  bookmarks. Watchers reconnect on `410 Gone`, network failure, and auth
+  failure with exponential backoff.
+- `api.qps` / `api.burst` are applied as a token-bucket rate limiter on
+  the client; both list and watch HTTP requests pass through the limiter.
+- The watcher task runs on the same per-core async runtime as the
+  extension instance that initialized it. All Kubernetes client I/O is
+  non-blocking; no dedicated blocking worker thread is required.
```
lalitb: Non-blocking I/O helps, but the watcher also has CPU work: decoding watch events, updating metadata, and rebuilding indexes. Since this runs on the same per-core runtime as the data path, large Kubernetes updates could add latency to telemetry processing. Can we either move this work off the data-path runtime, or explain why the expected CPU cost is small enough?

lalitb:
@pritishnahar95  -  Thanks for putting this together. This is going in a good direction, and the design doc already covers a lot of important details. I left a few early comments only on the major sharing/lifecycle pieces, since those are easier to adjust while the PR is still draft. Will do a more thorough review once the PR is ready for review.

rust/otap-dataflow/docs/k8s-attributes-processor.md
pritishnahar95: IMO, each local copy of the extension with a pipeline per-core replica would have the exact same cache, so I chose to only keep one shared implementation and avoid per-core duplication of memory. Even from the point of view of overall compute, we are basically adding more compute cost to have identically configured watchers watch for K8s events and build and maintain the cache. The only benefit is to have each per core copy of consumer node read from the cache without using Arc primitives.

rust/otap-dataflow/docs/k8s-attributes-processor.md
pritishnahar95: I would like to ideally have the extension's data shared between processors through a well supported engine primitive. But since that is not available in phase 1, and in order to make progress, I am proposing a workaround using ArcSwap primitives. See `## Thread-per-Core Integration` section in the doc for more details.

rust/otap-dataflow/docs/k8s-attributes-processor.md
pritishnahar95: KubernetesMetadataLookup will simply provide pod/workload metadata based on attributes specified by processor. The responsibility of enriching telemetry data rests with the processor.

rust/otap-dataflow/docs/k8s-attributes-processor.md
pritishnahar95: Thanks for pointing this out, it's captured later in the doc. 
I haven't delved into how extensions are implemented and integrated by engine in pipelines, do you think it would be possible to implement extension readiness in a way that pipelines are not started, until extension is not ready? No worries if you don't have an answer ready, I will probably look at this later myself.

rust/otap-dataflow/docs/k8s-attributes-processor.md
pritishnahar95: Thanks for raising this, it's a very good point! The extension lifecycle does not currently support sharing data across pipeline per-core replicas, and we will have explicit engine/pipeline-group scope ownership available, only in later phases in extension support. What I have proposed is a workaround, but we can always migrate K8sMetadataLookup extensions to pipeline-group or engine scoped extensions, once we have that support.

rust/otap-dataflow/docs/k8s-attributes-processor.md
gouslu: A collector with a single pipeline/core can make use of that additional optimization and not pay for any synchronization or atomics. For single pipeline/core collectors this would be a win/win.

rust/otap-dataflow/docs/k8s-attributes-processor.md
gouslu: Yes, that makes sense and was already aligned with how I thought it would work!

pritishnahar95: sounds good, will update doc, stating that design will assume this feature has been added to extensions

rust/otap-dataflow/docs/k8s-attributes-processor.md
pritishnahar95: sounds good, will update doc, stating that design will assume this feature has been added to extensions and we can add `wait_for_metadata`

rust/otap-dataflow/docs/k8s-attributes-processor.md
pritishnahar95: synced with Gokhan offline, pipeline-group or engine scoped extensions are in the works. It makes sense to wait a little, avoid this workaround, and start from the long term solution, will update this doc accordingly.

rust/otap-dataflow/docs/k8s-attributes-processor.md
pritishnahar95: Switching to pipeline-group or engine scoped extensions will automatically take care of this.
Based on conversation with @gouslu, my understanding is that pipeline-group or engine scoped extensions will be scheduled on a different OS thread/runtime, so this should be addressed.

rust/otap-dataflow/docs/k8s-attributes-processor.md
pritishnahar95: @gouslu will switch to pipeline-group or engine scoped extensions based on our offline discussion.
We will skip this intermediate solution using OnceLock.

lquerel:
@pritishnahar95 @lalitb @gouslu 

I like the direction of splitting Kubernetes metadata acquisition into k8s_metadata_extension and keeping the processor hot-path-only. **I'm wondering whether we should push this one step further and make the processor generic!**

Instead of implementing a Kubernetes-specific resource mutation processor, **we could introduce a generic resource_enricher_processor that consumes metadata lookup capabilities exposed by extensions.** The first provider would be `k8s_metadata_extension`, but the same processor could later consume `aws_metadata_extension`, `gcp_metadata_extension`, `azure_metadata_extension`, or other inventory/metadata providers.

In this model, the extension owns provider-specific discovery, authentication, caching, indexes, refresh/watch behavior, and degraded state. The processor owns the generic mechanics: association rule evaluation, conflict policy, projection into resource attributes, Arrow mutation planning, provider ordering, hit/miss/conflict metrics, and hot-path safety.

This would keep the Kubernetes behavior familiar, possibly through a k8s_attributes preset, while avoiding a one-off processor architecture.

rust/otap-dataflow/docs/k8s-attributes-processor.md
pritishnahar95: Will skip this solution and start from the long term solution based on pipeline-group or engine scoped extensions, will update this doc accordingly.

gouslu:
> This would keep the Kubernetes behavior familiar, possibly through a k8s_attributes preset, while avoiding a one-off processor architecture.

Thanks for breaking it down like that. I think it makes a lot of sense. I was thinking about how this could be made more generic purpose without making it too generic, but didn't have an answer to it for myself. This seems like a sweet spot for making it generic without making it too generic.

Since these extension(s) will be used in the hot path, we need to design the capability surface such that it is suitable for hot-path optimizations both for implementers and the consumers.

lalitb:
> Instead of implementing a Kubernetes-specific resource mutation processor, **we could introduce a generic resource_enricher_processor that consumes metadata lookup capabilities exposed by extensions.

I like this direction. One thing worth making explicit is the provider contract: the generic processor should not call Kubernetes-shaped methods like `pod()`, `namespace()`, or `workload()`. It should pass a prepared lookup key and get back a provider-neutral enrichment record/snapshot, so future providers can plug in without changing the processor hot path.

lalitb:
> Since these extension(s) will be used in the hot path, we need to design the capability surface such that it is suitable for hot-path optimizations both for implementers and the consumers.

Agree. For the hot path, I think the capability should look like a precomputed lookup table: the processor builds a small lookup key, does one synchronous lookup against an immutable snapshot, and gets back a ready-to-project record. Any watcher state, index rebuild, cache refresh, or provider-specific branching should stay off that path.

rust/otap-dataflow/docs/k8s-attributes-processor.md
lalitb: Makes sense to avoid the per-core workaround. One thing I would still like the updated doc to make explicit is the runtime contract for group/engine-scoped extensions: whether watcher decoding, cache updates, and index rebuilds run off the data-path runtime, or just under a different ownership scope. The ownership fix is clear, but the latency isolation depends on that scheduling detail.

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,1158 @@
...
+| Default extracted metadata | `k8s.namespace.name`, `k8s.pod.name`, `k8s.pod.uid`, `k8s.pod.start_time`, `k8s.deployment.name`, `k8s.node.name` |
+| Authentication | `service_account` (in-cluster) and `kube_config` (out-of-cluster); no `auth_type: none` in v1 |
+| Pod-delete grace | Default 120s |
+| Watch resync | Default `0s` (disabled); event-driven only, matches large-cluster guidance |
+| Live reconfiguration | Receives `NodeControlMsg::Config { config }` like `attributes_processor`. Extension and per-pipeline processor reloads are independent units of change since the extension lives outside the pipeline. See [Live Reconfiguration](#live-reconfiguration). |
+| Index-shape registration | The extension cannot see processor configs at its own `start()` (the framework wires extensions and consumers in one direction; see [Index-shape registration](#index-shape-registration)). Each processor calls `register_shape()` on its capability handle at its own `start()` and on every `NodeControlMsg::Config` reload. The first lookup for a newly-registered shape may miss until the extension's writer back-fills it over the existing reflector contents; registration is idempotent. |
```
lalitb: > The first lookup for a newly-registered shape may miss until the extension's writer back-fills it

This makes config application eventually consistent: startup succeeds, telemetry flows, and enrichment may miss until async back-fill completes. That feels like a core user-facing contract, not just an implementation detail. Can we define whether shape registration can fail or block startup/reload when the required lookup state is not ready?

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,1158 @@
...
+
+## Filter and Exclude
+
+`filter.node` / `filter.node_from_env_var` narrow the pod watcher with a
+`spec.nodeName=<node>` field selector. The namespace, node, and workload
+watchers are not node-narrowed.
+
+`filter.namespace` narrows the pod watcher to a single namespace at API
+scope (not via a selector). The namespace and node watchers remain at
+cluster scope and silently degrade to empty when RBAC denies them (see
```
lalitb: I don’t think RBAC denial should look the same as “metadata does not exist.” If configured extraction depends on namespace/node/workload metadata, the extension should expose a degraded state or clear error so operators know enrichment is incomplete because of permissions.

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,1158 @@
...
+The association source plugs into the existing composite-key index: the
+writer indexes pods by `[(connection.ip, <pod_status_pod_ip>)]`
+alongside the other default shapes, with the same `hostNetwork` skip
+rule as `k8s.pod.ip` (see
+[`KubernetesMetadataLookup` capability](#kubernetesmetadatalookup-capability)).
+No new index machinery is needed.
+
+### `passthrough`
+
+`passthrough: true` stamps `k8s.pod.ip` from `OtapPdata::peer_addr()`
```
lalitb: My thought - Passthrough should probably be able to run without starting or binding the Kubernetes metadata extension. Its main value is that an agent can stamp pod IP without Kubernetes API access or cache/watch cost. If the group extension still starts watchers, passthrough loses that operational benefit.

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,1158 @@
...
+  rebuilds all introduce scheduling points. Keeping that work behind a
+  synchronous, non-blocking capability surface (`KubernetesMetadataLookup`)
+  makes it impossible to accidentally call Kubernetes API code from the
+  data path -- the surface exposes no such method, and the processor crate
+  has no direct Kubernetes client dependency.
+
+The extension boundary is where blocking and async work belongs; the
+processor stays a synchronous reader of a published snapshot.
+
+The sharing unit is the pipeline group, not the engine. Different groups
```
lalitb: Group scope makes sense for Kubernetes filters, but I’m not sure it should be baked into the enrichment architecture as the only valid provider scope. Some metadata providers may be naturally engine-wide, while Kubernetes may be group scoped. It may be cleaner to keep provider scope independent from the processor/enrichment model.

ThomsonTan (commented):
Duplicated comment.

pritishnahar95:
> @pritishnahar95 @lalitb @gouslu
> 
> I like the direction of splitting Kubernetes metadata acquisition into k8s_metadata_extension and keeping the processor hot-path-only. **I'm wondering whether we should push this one step further and make the processor generic!**
> 
> Instead of implementing a Kubernetes-specific resource mutation processor, **we could introduce a generic resource_enricher_processor that consumes metadata lookup capabilities exposed by extensions.** The first provider would be `k8s_metadata_extension`, but the same processor could later consume `aws_metadata_extension`, `gcp_metadata_extension`, `azure_metadata_extension`, or other inventory/metadata providers.
> 
> In this model, the extension owns provider-specific discovery, authentication, caching, indexes, refresh/watch behavior, and degraded state. The processor owns the generic mechanics: association rule evaluation, conflict policy, projection into resource attributes, Arrow mutation planning, provider ordering, hit/miss/conflict metrics, and hot-path safety.
> 
> This would keep the Kubernetes behavior familiar, possibly through a k8s_attributes preset, while avoiding a one-off processor architecture.

@lquerel this sounds like a good idea. I have added a design doc for the generic processor, tentatively titled "resource enricher processor". This is a first draft of the design. If this makes sense, happy to pivot to this fully, and delete design doc for K8s attribute processor.
cc @lalitb @gouslu @jmacd

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,1158 @@
...
+      k8s_metadata:
+        type: extension:k8s_metadata
+        config:
+          auth_type: service_account            # service_account | kube_config
+          kube_config:
+            path: ~                              # optional, falls back to KUBECONFIG / ~/.kube/config
+            context: ~                           # optional
+          api:
+            qps: 5
+            burst: 10
```
jmacd: FYI I will be proposing a standard configuration model for token-bucket rate limits (in the near future); I hope we can revise this,

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1001 @@
...
+| Conflict-policy application | processor |
+| Lookup hit/miss telemetry, attribute-write counts | processor |
+| Index-spec derivation and registration | processor |
+| Live reconfiguration of association + projection rules | processor |
+
+The extension owns nothing that depends on a specific pipeline's data
+schema. The processor owns nothing that depends on the provider's runtime
+state or vocabulary.
+
+## `MetadataLookup` capability
```
jmacd, line 263: FYI https://github.com/open-telemetry/otel-arrow/issues/3275 calls for a similar mechanism.

ThomsonTan (commented):
Duplicated comment.

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1001 @@
...
+
+**Target crate:** `crates/contrib-nodes`
+**Target processor module:** `crates/contrib-nodes/src/processors/resource_enricher_processor/`
+**Target extension module (first provider):** `crates/contrib-nodes/src/extensions/k8s_metadata_extension/`
+
+This design follows
+[Reference-Informed OTAP-Native Capability Design](ai/reference-informed-otap-native-capability-design.md).
+It assumes:
+
+- the engine supports **shared extensions** (engine-scoped or
```
lalitb: Can we add a short “Required engine support” section here? The current text says the design assumes shared extensions at group/engine scope and a separate extension runtime, but I am not sure which of those are available today versus future framework work.

My understanding is that the processor can use an extension capability today, but the “one shared Kubernetes metadata cache/watcher per group or engine” part depends on broader extension scopes and runtime isolation. If that is correct, it would help to state it explicitly so reviewers can separate the processor design from the engine work it depends on.

rust/otap-dataflow/docs/k8s-attributes-processor.md
```diff
@@ -0,0 +1,1158 @@
+# Kubernetes Attributes Processor Design
```
lalitb: Is the intent to eventually remove this Kubernetes-specific processor design, or keep it only as a Kubernetes provider appendix for the generic `resource_enricher_processor` design?

The PR discussion seems to have moved toward the generic processor model, so it would help to make the relationship between the two docs explicit.

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1001 @@
...
+    /// Config-time only. Declare what to index by and what to store on
+    /// behalf of `registrant`. Set-replacing per registrant: the
+    /// extension stores `Map<RegistrantId, IndexSpec>` and the
+    /// effective spec is the union across all registrants. Diffs the
+    /// union before vs. after, back-fills additions, drops removals,
+    /// and starts/stops discovery sources accordingly. Idempotent per
+    /// registrant. Returns immediately; back-fill runs on the
+    /// extension's runtime. Never called from `process()`.
+    /// See [Index spec registration](#index-spec-registration).
+    fn set_index_spec(&self, registrant: RegistrantId, spec: &IndexSpec);
```
lalitb, line 350: Should this call have a way to return validation errors, or should there be a separate validation step before it? The processor is generic, so it can parse the config, but it does not know which fields a provider actually supports.

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1001 @@
...
+struct RegistrantId {
+    pipeline_group: PipelineGroupId,
+    pipeline:       PipelineId,
+    node:           NodeId,
+}
+
+// What the processor declares to the extension at config time.
+// Both fields are derived from the processor's own config:
+//   key_shapes    <- associations[].sources   (names only)
+//   record_fields <- projection[].from        (dotted paths; globs ok)
```
lalitb: Can we make the mapping between resource attribute names and provider field names explicit? In the examples, associations use resource attributes like  `k8s.pod.uid`, but projections use provider fields like `pod.uid`. Those are really the same value, but with different names.

The Kubernetes metadata extension needs to know that `k8s.pod.uid` maps to `pod.uid`, and it also needs to handle special cases like normalizing `container.id`. It would help to define where this mapping lives, otherwise a valid-looking lookup key could miss just because the processor and provider used different names for the same field.

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1001 @@
...
+processor one stable view across an entire `process()` call, and a
+config-time `set_index_spec()` / `clear_index_spec()` pair that
+processors use to declare — per registrant — what to index by and what
+fields to materialize (see [Index spec registration](#index-spec-registration)
+for why this lives on the capability surface rather than inside
+extension config, and why it is multi-registrant).
+
+```rust
+// A single lookup key: an AND of (resource_attribute_name, value) pairs.
+type LookupKey = Vec<(String, String)>;
```
lalitb: Can we tighten the hot-path allocation claim here? The doc says lookups are allocation-free, but `LookupKey = Vec<(String, String)>` suggests the processor may allocate a Vec and clone strings while building the lookup key for each resource.

If the intent is truly allocation-free lookup, should the key use borrowed values or a small stack-backed structure instead? Otherwise it may be better to say the snapshot read/cache lookup is lock-free, while key construction may still allocate.

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1001 @@
...
+  to another group's group-scoped extension are rejected.
+- Within a scope unit, an extension `name:` is unique; two declarations
+  with the same name at the same scope unit are rejected.
+
+## Association
+
+Association is the rule engine that maps `OtapPdata` resource attributes
+to an `EnrichmentRecord` in the extension's cache. The semantics are:
+
+- **Rule selection by attribute presence.** The processor evaluates the
```
lalitb, line 776: Can we define how association source predicates are handled in the generic design? Go has a Kubernetes specific rule where `host.name` is only used for pod lookup if the value is actually an IP address. In this generic design, association sources seem to be selected only by attribute presence.

For cases like this, we need somewhere to express “use this source only if the value is an IP.” Should that live in processor config, provider validation, or provider lookup logic?

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1001 @@
...
+  `keep_existing`.
+
+## Connection-Peer Association
+
+Some receivers observe a connection peer address that is only available at
+the receiver — by the time a batch reaches a processor, the originating
+socket is gone — so it is propagated on `OtapPdata` itself via
+`OtapPdata::peer_addr()`.
+
+A `sources_from_connection:` entry binds the peer IP to a named resource
```
lalitb: Can we keep connection peer IP as lookup context unless the user explicitly enables stamping/passthrough? The peer IP is not always the pod IP; it may be a proxy, sidecar,= gateway, NAT, or load balancer hop.

Using it as a lookup hint is fine, but automatically writing it as a resource attribute like `k8s.pod.ip` could make the telemetry resource incorrect. I think the design should separate “use peer address for association” from “persist peer address onto the resource.”

rust/otap-dataflow/docs/k8s-attributes-processor.md
pritishnahar95: Sure, we can. The only caveat is whatever rate limit model we come up with, it needs to be compatible with this(specifically K8s metadata lookup) extension.

rust/otap-dataflow/docs/k8s-attributes-processor.md
pritishnahar95: Removed K8s attribute processor doc, now that we have some positive momentum for the alternative component

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1001 @@
...
+      # `from` is a record-field name (or glob). `to` is the destination
+      # resource-attribute name (template-expandable).
+      projection:
+        - { from: "pod.uid",          to: "k8s.pod.uid" }
+        - { from: "pod.name",         to: "k8s.pod.name" }
+        - { from: "pod.start_time",   to: "k8s.pod.start_time" }
+        - { from: "namespace.name",   to: "k8s.namespace.name" }
+        - { from: "deployment.name",  to: "k8s.deployment.name" }
+        - { from: "node.name",        to: "k8s.node.name" }
+        - { from: "pod.labels.*",     to: "k8s.pod.label.{name}" }
```
ThomsonTan: `pod.labels.*` projects every label/annotation on an object into resource attributes, but those keys are usually set by workload owners, not the collector operator, so attribute count and names are effectively unbounded and user-controlled. That risks runaway memory and high-cardinality cost downstream (annotations can also be very large). The memory section's "only stores listed fields" doesn't help when the field is a wildcard. Suggest adding a cap (max attrs per resource, max key length), an optional label allow/deny list, and a enrich.projection.dropped counter, and calling out the untrusted-key risk explicitly.

pritishnahar95: Good point, removed wildcards and regex for now

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: makes sense, updated

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: I think any such logic will reside in the provider. There are no generic predicates in rule selection for now.

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: good point, updated

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: updated

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: added a method for validation and updated workflow

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: added a new section for this

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1231 @@
...
+The lookup snapshot is one struct published as a single immutable
+snapshot:
+
+```rust
+// Extension-internal type.
+struct MetadataTables {
+    /// The only public-facing index. Stored with owned keys (one
+    /// (String, String) pair per element of the key's KeyShape) so the
+    /// HashMap owns its data, but the hot-path lookup looks the entry
+    /// up by a borrowed `LookupKey<'_>` — i.e. the processor does not
```
lalitb: Can we explain how this lookup stays allocation-free? The stored key is owned (`Vec<(String, String)>`), but the lookup key is   borrowed (`&str`). I don’t think `HashMap::get` supports that automatically, so it would help to mention whether this needs a custom key type, `Borrow`, or another lookup path.

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1231 @@
...
+that group bind by the local name and see a different cache.
+
+The Kubernetes provider's catalog publishes the OpenTelemetry K8s
+semantic-convention attribute names — see
+[semantic-conventions/docs/resource/k8s](https://github.com/open-telemetry/semantic-conventions/tree/main/docs/resource/k8s).
+So `k8s.pod.uid`, `k8s.pod.ip`, `k8s.pod.name`, `k8s.namespace.name`,
+`k8s.deployment.name`, `k8s.node.name`, `container.id`, etc. are the
+valid names that may appear in a processor's `associations:` and
+`attributes:` blocks when binding this provider. Per-key label and
+annotation paths are exposed as `k8s.pod.labels.<key>` and
```
lalitb, line 720: Can we clarify how validation works for labels and annotations? Fixed K8s attributes can be in a catalog, but label and annotation keys are created by workloads and may not be known ahead of time. I think the design needs to say how names like `k8s.pod.labels.app` are accepted without requiring the key to already exist in the cache.

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1231 @@
...
+   client, and returns.
+
+### Live Reconfiguration
+
+The processor handles `NodeControlMsg::Config { config }` (same
+mechanism as `attributes_processor` et al.): re-parse, recompile plans,
+re-derive and re-validate the `IndexSpec`, and on success call
+`set_index_spec(registrant, spec)` under the stable `RegistrantId`. A
+validation failure leaves the previous plan in place and is surfaced as
+a reload error. The extension subscribes to the same message for its
```
lalitb: Can we clarify the reload path for extensions? The processor uses `NodeControlMsg::Config`, but extensions use their own control channel. This sentence makes it sound like the extension also receives node config messages, which seems different from the extension architecture doc.

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1231 @@
...
+  `set_index_spec()` back-fills never compete with the data path. The
+  capability surface enforces this contract: every method is synchronous
+  and returns immediately (`snapshot()` is one atomic load, `lookup()`
+  is one map read, `set_index_spec()` schedules background work and
+  returns). The processor never awaits extension work, even at
+  `start()`, and the processor crate has no provider-client dependency.
+
+## Required engine support
+
+The processor side of this design is implementable on the engine as it
```
lalitb, line 68: Can we clarify whether this can actually run before group/engine-scoped extensions are available? This section says the processor can work today, but later the config rules (line 707) reject per-pipeline scope, which is the only scope available today. I think the doc should make that dependency clear.

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1231 @@
...
+Flow on the data path:
+
+1. Receiver emits `OtapPdata` with `resource` attributes that include at
+   least one of the configured association sources (e.g. `k8s.pod.uid`).
+2. Processor walks `associations` rules in order. The first rule whose
+   sources are all present on the resource is used. Each rule's source
+   values are joined into a `LookupKey`.
+3. On a hit, the processor reads the returned `EnrichmentRecord` and
+   writes the configured attributes onto the resource attribute Arrow
+   batch in place, honoring the conflict policy.
```
lalitb: Can we clarify the write cost here? Arrow batches are immutable, so enrichment will likely rebuild or replace the resource attribute batch rather than update it in place. That matters for the hot-path cost of this processor.

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1231 @@
...
+and the rule is skipped, the same way an absent resource attribute is.
+This keeps the first-match-wins semantics consistent.
+
+### Caveats
+
+- Connection IP is not always the originating client's IP: with NAT,
+  load balancers, service meshes, sidecars, or gateway hops the
+  observed peer is the last network hop. Because the value is only a
+  match key and is never written back, the worst case is a lookup miss,
+  not a corrupted resource attribute.
```
lalitb: I don’t think the worst case is only a miss. If traffic comes through a gateway, sidecar, NAT, or another collector, the peer IP may still match a real pod, but the wrong one. Then enrichment could attach the forwarder’s metadata to many resources. Can we make that risk explicit?

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1231 @@
...
+      #       it (mapping to the pod whose `status.podIP` matches).
+      #       See Connection-Peer Association.
+      associations:
+        - sources:
+            - { from: resource_attribute, name: k8s.pod.uid }
+        - sources:
+            - { from: resource_attribute, name: k8s.pod.name }
+            - { from: resource_attribute, name: k8s.namespace.name }
+        - sources:
+            - { from: resource_attribute, name: k8s.pod.ip }
```
lalitb, line 612: Can we define what happens when a lookup key is not unique? For example, pod IP can be reused, and hostNetwork pods can share the node IP. If two records produce the same key, the design should say whether that is rejected, last-writer-wins, or reported as an ambiguous match.

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1231 @@
...
+
+### Conflict policy
+
+When a destination resource attribute already exists on the resource, the
+configured `conflict_policy` decides:
+
+- `keep_existing` (default) — leave the existing value; skip the write.
+- `overwrite` — replace the existing value with the value from the
+  enrichment record.
+- `merge_string_lists` — when both the existing value and the value
```
lalitb: Can we define how list values are represented and merged in OTAP attributes? If string lists require serialized attribute values, then `merge_string_lists` may need decode, merge, and re-encode behavior rather than a simple scalar write. It would help to make that contract clear.

rust/otap-dataflow/docs/resource-enricher-processor.md
lalitb: That sounds reasonable for v1. I agree we probably do not need generic predicate syntax yet. Can we just make the doc explicit that provider-specific predicates and normalization live inside the provider, while the generic processor only checks source presence and calls lookup?

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: added a new section: `### Write cost (Arrow immutability)`

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: refactored doc to remove confusion, this will only run with engine or group scoped extensions.

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: addressed, rewrote this part

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: good catch, rewrote this point

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: added a new section for non-unique keys: `#### Non-unique lookup keys`

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: Dropped `merge` option for now, we can revisit this later

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: see new section -> `#### How the borrowed lookup stays allocation-free`

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: Cache, will have this key, but provider catalog needs to allow for this, updated catalog section for this. 
See - `### Provider catalog`

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1366 @@
...
+  extension is rejected.
+
+A provider may support both scopes; the operator picks at config time.
+
+## Telemetry
+
+The processor is a single concrete component, so its metrics are fixed.
+It reports through the `processor.resource_enricher` set:
+
+| Metric | Type | Labels | Notes |
```
lalitb: One more small thing: can we align this telemetry table with the current `MetricSet` model?  I think today `MetricSet` supports fixed metric fields plus metri set/entity attributes, not different labels per individual metric observation.

For v1, maybe these should be fixed counters, or the doc should call out labeled metric support as extra framework work.

rust/otap-dataflow/docs/resource-enricher-processor.md
```diff
@@ -0,0 +1,1366 @@
...
+future work):
+
+- **Broader extension scopes — pipeline-group scope and engine scope.**
+  Today the only available scope is *pipeline* scope, which the
+  framework documents as per-core. A processor can bind a shared
+  capability today, but "exactly one discovery loop and one cache per
+  group (or per process)" depends on the framework adding group-scoped
+  and engine-scoped extensions. The processor design here is
+  scope-agnostic by construction — it does not name a scope, it just
+  binds a shared capability — and works correctly under per-pipeline
```
lalitb: > and works correctly under per-pipeline per-core extensions;

Can we make this more explicit? This says the design works under per-pipeline extensions, but later the config rules say per-pipeline scope is rejected. Since pipeline scope is the only thing available today, I think the doc should clearly say whether this design waits for group/engine-scoped extensions, or whether pipeline scope is allowed as a temporary duplicated mode.

lalitb (approved):
Approving, two design clarifications posted above - good to address before merge. The overall design direction looks good to me.

lalitb (approved):

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: good catch, thanks! refactored metrics

rust/otap-dataflow/docs/resource-enricher-processor.md
pritishnahar95: addressed in new commit, clarified this part.

jmacd (approved):

PR #3140, opened 2026-05-29
chore(dependencies): Reduce frequency of Go/Rust digest dependency updates

drewrelmas:
# Change Summary

In the past few week alone, we've had 9 PRs updating various Go and Rust digest references, sometimes on every commit to the upstream repo. Examples:
- https://github.com/open-telemetry/otel-arrow/pull/3092
- https://github.com/open-telemetry/otel-arrow/pull/3091
- https://github.com/open-telemetry/otel-arrow/pull/3099
- https://github.com/open-telemetry/otel-arrow/pull/3136

This creates some churn in the repo, want to limit each dependency to only spawn once a week.

(Note: if a dependency upgrade is needed immediately, can always override the schedule by hitting checkbox for relevant update on #417)

## What issue does this PR close?

N/A

## How are these changes tested?

Renovate config passes validation

## Are there any user-facing changes?

No

### Changelog

* [x] Added a `.chloggen/*.yaml` entry, OR this PR is a `chore` (indicated in title).

PR #3257, opened 2026-06-09
chore(deps): Isolate digest updates while keeping once a week pattern

drewrelmas:
# Change Summary

It is taxing to isolate the breaking change from #3235.

In this specific case, it is the `geneva` dependency: https://github.com/open-telemetry/otel-arrow/actions/runs/27225496774/job/80391437570?pr=3235

```log
error[E0063]: missing field `obo_event_map` in initializer of `geneva_uploader::GenevaClientConfig`
   --> crates/contrib-nodes/src/exporters/geneva_exporter/mod.rs:301:29
    |
301 |         let client_config = GenevaClientConfig {
    |                             ^^^^^^^^^^^^^^^^^^ missing `obo_event_map`
```

#3140 attempted to limit the number of digest PRs we get, but grouping them all together probably isn't the right move.

Instead, still get digest updates from Go and Rust once a week, but treat each one as a separate PR. That way, we can easily merge non-breaking ones and immediately have visibility into which breaking ones.

N/A

N/A

N/A

### Changelog

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -84,10 +84,16 @@
       ]
     },
     {
-      // limit Go and Rust git/module digest-only updates to once a week
-      // these trickle in frequently, so consolidate them
-      "groupName": "go and rust digest updates",
-      "matchManagers": ["gomod", "cargo"],
+      // limit Go git/module digest-only updates to once a week
+      "matchManagers": ["gomod"],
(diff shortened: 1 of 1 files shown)


albertlockett (approved):
:+1:  thanks @drewrelmas

PR #2375, opened 2026-03-19
Refactor crates/otap to crates/common (fixes #2362)

perashanid:
Fixes #2362

## Summary
Renamed the `crates/otap` folder to `crates/common` to better reflect its purpose as shared infrastructure rather than OTAP-specific node implementations.

## Changes
- Renamed folder: `crates/otap` → `crates/common`
- Renamed package: `otap-df-otap` → `otap-df-common`
- Updated all Cargo.toml dependencies (6 files)
- Updated all Rust source file imports (50+ files)
- Updated documentation references

## Rationale
After moving all core/contrib nodes out of `crates/otap`, the remaining code is miscellaneous shared infrastructure (transport, utilities, test support) used by multiple components. The new name `common` accurately represents this purpose.

## Breaking Changes
External users depending on `otap-df-otap` will need to update to `otap-df-common`.

Issue #2362, opened 2026-03-18
Refactor remaining crates/otap content to relevant crates


drewrelmas:
Now that all core/contrib nodes are moved out of `crates/otap`, most of what remains is 'miscellaneous code' that is mostly used by multiple nodes or as test infrastructure.

Consider the proper home for these code pieces. It is possible to rename the `otap` crate to something like `common` to better represent its purpose.

For particulars, see these comments on issue #1847:
* https://github.com/open-telemetry/otel-arrow/issues/1847#issuecomment-3785124745
* https://github.com/open-telemetry/otel-arrow/issues/1847#issuecomment-3785259366

jmacd:
With apologies to @perashanid, I have removed "good first issue". I would like to see us consider the sub-modules of the crates/otap area independently, and second, I would like to see us plan to reduce the impact of these changes. 

For the first, I would like to see the contents of `crates/otap/src/pdata.rs` move into `crates/pdata`, but there are pitfalls and this could require a bit of untangling. When I looked at the file, it occurs to me that maybe the Context type belongs in `crates/engine`.

For the second, if we were to move the Context type (into crates/engine/src/context.rs makes sense) or the OtapPdata type (into crates/pdata/src/pdata.rs or lib.rs?), we can do so with less impact by leaving behind type aliases, and then following with PRs that only adjust `use` statements and remove unused aliases.

drewrelmas:
I still want to do some cleanup inside `otap` but I think we've accepted for the moment it is still our `common` crate.

PR #3261, opened 2026-06-10
[otap-dataflow] add kafka receiver into contrib-nodes

### LINKED ISSUE #3248: [otap-dataflow] Add Kafka Receiver
### Kafka Receiver

The Kafka receiver consumes messages from one or more Kafka topics and routes them into the pipeline as traces, metrics, or logs based on topic matching.

#### Per-Signal Topic Routing

Each signal type (traces, metrics, logs) can subscribe to different topics. Topics can be specified as literal strings or regex patterns (prefixed with `^`). Exclusion patterns provide negative filtering, allowing broad regex subscriptions with specific topic opt-outs.

```yaml
traces:
  topics: ["^otlp-traces-.*"]
  exclude_topics: ["^otlp-traces-internal-.*"]
metrics:
  topics: ["otlp-metrics"]
logs:
  topics: ["otlp-logs-prod", "otlp-logs-staging"]
```

#### Encoding Formats

Supports two encoding formats per signal:

- `otlp_proto` (default) — Standard OTLP protobuf
- `otap_proto` — OTel Arrow (OTAP) protobuf

#### Per-Message Format Detection

When `message_format_header` is configured, the receiver checks each incoming Kafka message for a header matching the configured key. If the header value is `otlp` or `otap`, it overrides the per-signal encoding default. This enables mixed-format topics where different producers use different encodings.

```yaml
message_format_header: "MessageFormat"
```

When not configured, no header-based detection is performed and the per-signal `encoding` is always used.

#### Offset Management

Two commit modes:

- **`manual`** (default, at-least-once) — Tracks every consumed message per partition. Only the lowest un-acknowledged offset is committed, preventing offset skipping when downstream acks arrive out of order.
- **`auto`** (at-most-once) — Delegates to rdkafka's auto-commit.

Optional idempotent processing (`enable_idempotency`) skips messages whose offset is already in-flight or below the high-water mark.

#### Header Extraction into Resource Attributes

The `resource_attrs_from_headers` config maps Kafka header keys to resource attribute keys with typed parsing:

```yaml
resource_attrs_from_headers:
  x-tenant-id:
    key: "tenant.id"
    value_type: "string"
  x-priority:
    key: "message.priority"
    value_type: "int"
```

Supports `string`, `bool`, `int`, and `float` value types. Extracted attributes are upserted onto the resource (existing attributes with the same key are overwritten).

#### Transport Header Capture

When a capture policy is configured at the pipeline level, matching Kafka headers are captured into the pdata context as `TransportHeaders`. These headers are preserved through the pipeline and can be used by downstream nodes (e.g., the Kafka exporter for dynamic topic routing or partition key generation).

#### Consumer Group Configuration

- Rebalancing strategies: `range`, `round_robin`, `cooperative_sticky`
- Static group membership via `group_instance_id`
- Configurable session timeout, heartbeat interval, and fetch tuning
- Transaction isolation level: `read_uncommitted` or `read_committed`

#### Authentication and TLS

- SASL mechanisms: AWS MSK IAM OAUTHBEARER (with automatic token refresh), PLAIN, SCRAM-SHA-512
- Full mTLS support with optional certificate verification skip
- Security protocol auto-inferred from auth/TLS combination (PLAINTEXT, SSL, SASL_PLAINTEXT, SASL_SSL)

#### Additional configuration

`consumer_config` accepts arbitrary librdkafka key-value pairs for advanced tuning. First-class config fields always take precedence on conflict.

c1ly:
# Change Summary

Add Kafka Receiver implementation that takes inspiration from the go and rotel versions.

Add kafka_util that shares common functions and data types with the kafka receiver and exporter

* Closes #3248 

unit tests and integration tests with kafka broker (requires docker container)

no user face changes


rust/otap-dataflow/Cargo.toml
```diff
@@ -65,6 +65,8 @@ data_engine_expressions = { path = "../experimental/query_engine/expressions" }
 data_engine_kql_parser = { path = "../experimental/query_engine/kql-parser" }
 data_engine_parser_abstractions = { path = "../experimental/query_engine/parser-abstractions"}
 
+aws-config = "1.0"
+aws-msk-iam-sasl-signer = "1.0.1"
```
albertlockett, line 72: I think maybe the AWS MSK IAM Auth mechanism should be behind the `aws` feature (similar to what we do for s3 object_store support).

rust/otap-dataflow/crates/contrib-nodes/src/common/mod.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Shared functions and data types for contrib node implementations.
+
+/// Shared Kafka utilities for Kafka receiver and exporter.
+#[cfg(any(feature = "kafka-receiver"))]
+pub mod kafka_util;
```
albertlockett: nit: can we just call this `kafka` instead of `kafka_util`? `util` strikes me as an unstructured grab-bag of miscellaneous utilities, but the `common::kafka` module actually has some structure to it (auth, security, etc)

rust/otap-dataflow/Cargo.toml
albertlockett: Changes will also be needed in the common/kafka_utils auth.rs, aws.rs and mod.rs to make the rust code paths also behind this feature.

albertlockett (commented):
Super contribution @c1ly ! Haven't gone through everything yet. Here are a few initial comments. Will continue reviewing this afternoon

rust/otap-dataflow/crates/contrib-nodes/src/common/mod.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Shared functions and data types for contrib node implementations.
```
drewrelmas, line 4: This is less of a question on the actual `kafka` implementation and more of a question for the other @open-telemetry/arrow-maintainers. I think this relates to:
- #1847 
- #2362

Do we envision `contrib-nodes` and `core-nodes` crates both having `common` submodules? 

As an exercise, I am imagining we move [grpc client_settings](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs) and [http client_settings](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs) into `crates/core-nodes/common` since they are used primarily by exporters there. However, what would happen then if a `contrib` exporter also wanted to use the same structs? We would introduce a dependency on `core` in `contrib`. The same argument could potentially be made in reverse.

One alternative is a `common` crate (which was actually previously attempted by a contributor in #2375). Instead of simply renaming `crates/otap` to `crates/common`, we could start small, with this shared `kafka` code being the first resident.

albertlockett: Thanks for pointing this out @drewrelmas !

If I understand correctly, the issue here that this is introducing a new pattern, where shared code for contrib-nodes is in this `common` module, but shared code for core-nodes is in the `otap` crate, and in the long-term we'd like to have a less dis-jointed pattern?

This is tricky, and I don't have extremely strong feelings about it. I'm leaning toward having `core-nodes/common` and `contrib-nodes/common`. 

Would be interested to hear what others have to say as well

> Do we envision contrib-nodes and core-nodes crates both having common submodules?

I hadn't thought that far ahead, but I suppose it could. 

> As an exercise, I am imagining we move grpc client_settings and http client_settings into crates/core-nodes/common since they are used primarily by exporters there. 

That seems reasonable.

> However, what would happen then if a contrib exporter also wanted to use the same structs? We would introduce a dependency on core in contrib. 

I don't have a strong objection to that. 

> The same argument could potentially be made in reverse.

The issue with this, assuming contrib-nodes is already depending on core-nodes, would be that it creates a circulate dependency right? IMO if something is a common utility that is used by both core and contrib nodes, then it belongs in core-nodes crate (assuming we don't introdce the `common` crate as you suggested).

> One alternative is a `common` crate

I find this somewhat less favourable. My inclination (based more on my feelings than technical merit) would be to have code that is only used in contrib-nodes (as these kafka utilities would be) contained within contrib-nodes, and not have them be a dependency of core-nodes.

drewrelmas: > have code that is only used in contrib-nodes (as these kafka utilities would be) contained within contrib-nodes

I'm also fine with continuing this approach until we hit a case where we need something from `contrib` in `core`, at which point we should just move whatever it is into `core` instead.

TL;DR, it is fine to have a dependency on `core` in `contrib`, but we should avoid the other way around. Thanks for the reply!

drewrelmas: I will mark this as resolved to avoid derailing this review, but to anyone else with ideas here, feel free to continue over slack or `Discussions`

rust/otap-dataflow/crates/contrib-nodes/src/common/mod.rs
lquerel: @drewrelmas @albertlockett @c1ly to add to this discussion, it's likely that we'll need part of the code related to Kafka connectivity and topic management as part of our topic concept, which currently only offers an in-memory implementation.

I don't yet know what form this will take (perhaps a topic extension?) but it's very likely that we'll have the kind of discussion shared in this thread in the future!

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,2673 @@
...
+    traces_topic_regexes: Vec<Option<Regex>>,
+    metrics_topic_regexes: Vec<Option<Regex>>,
+    logs_topic_regexes: Vec<Option<Regex>>,
+    /// Pre-compiled exclude topic regexes for each signal.
+    traces_exclude_regexes: Vec<Regex>,
+    metrics_exclude_regexes: Vec<Regex>,
+    logs_exclude_regexes: Vec<Regex>,
+    // TODO: add this back once we can reset it without re-creation: https://github.com/open-telemetry/otel-arrow/issues/1669
+    // used to decode otap bytes
+    // pdata_consumer: PdataConsumer,
```
albertlockett, line 204: This was completed in https://github.com/open-telemetry/otel-arrow/pull/1718 . We should be able to have one `PdataConsumer` and call `reset_streams` instead of creating a new `PdataConsumer` for each message.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,2673 @@
...
+
+        // Route the topic to the correct signal decoder. Supports both literal
+        // topic names and regex patterns (prefixed with `^`), exclude patterns,
+        // per-signal encoding, and multiple topics per signal type.
+        let mut pdata = if matches_any_topic(
+            self.config.traces_topics(),
+            &self.traces_topic_regexes,
+            topic,
+        ) && !matches_any_exclude(&self.traces_exclude_regexes, topic)
+        {
```
albertlockett, line 334: Maybe an idea for a future optimization: we check every message against the list of configured topics, and also ensure it is not in the list of excluded topics, so that we can figure out the signal type. 

I'm not sure how much time is spent making this determination - possibly not very much, but I suppose it would depend on the number of configured topics/regexes. If profiling reveals that we spent a lot of time here, maybe we could cache the signal type determination (or whether the topic is excluded).

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,2673 @@
...
+                // 2. Periodic offset commit (safety-net timer)
+                // will only trigger if user provides a commit interval
+                _ = async {
+                    match commit_interval.as_mut() {
+                        Some(interval) => interval.tick().await,
+                        None => std::future::pending().await,
+                    }
+                } => {
+                    self.commit_offsets(&consumer, &receiver_id)?;
+                }
```
albertlockett: Is there a reason we don't use EffectHandler::start_periodic_timer and then commit when we receive `NodeControlMessage::TimerTick`?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,2673 @@
...
+        MessageFormat::OtlpProto => Ok(OtapPdata::new(
+            Context::default(),
+            OtlpProtoBytes::ExportLogsRequest(Bytes::copy_from_slice(data)).into(),
+        )),
+        MessageFormat::OtapProto => {
+            let mut bar =
+                BatchArrowRecords::decode(data).map_err(|e| EngineError::PdataConversionError {
+                    error: e.to_string(),
+                })?;
+            let mut pdc = PdataConsumer::default();
```
albertlockett, line 1080: related to my other comment above - we should be able to reuse the consumer now that https://github.com/open-telemetry/otel-arrow/pull/1718 has landed

rust/otap-dataflow/crates/contrib-nodes/src/common/mod.rs
albertlockett: right, great point! I'm in favour of keeping this stuff in the contrib-nodes/common for now, and then when we're ready to tackle kafka backed topics we can move it to some sensible shared location

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
albertlockett: oh nevermind actually I'm mistaken. I was thinking about the Producer to which we added `reset_streams`. We might get a marginal benefit doing the same thing with the consumer, but that change would be unrelated to this PR so we can leave this TODO as is for now.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
albertlockett: nevermind, I am mistaken! I was thinking about the producer.

rust/otap-dataflow/crates/contrib-nodes/src/common/mod.rs
utpilla: The lean in this [thread](https://github.com/open-telemetry/otel-arrow/pull/3261#discussion_r3389694664) so far seems towards using submodules in both `core-nodes` and `contrib-nodes` with a "promote to core when both tiers need it" rule. I think that's right for some cases and wrong for others, and the line worth drawing is:

**Shared code should live in a scope that matches its consumer set's natural ownership boundary.** A submodule works well when every consumer lives in the same crate and shares its ownership, release, and feature-gating story. The moment that stops being true, having a separate crate is a better approach.

Four cases that cover what we'll keep running into:

1. **Single consumer:** inline. No shared module.
2. **Multiple components within one tier, single owner, single release** (e.g. core-internal glue between sibling core components): submodule. Core ships as one unit; the ceremony of a separate crate isn't paid back.
3. **Multiple components with independent ownership or feature-gating** (e.g. a contrib integration family such as kafka with a receiver and exporter, or shared code across multiple contrib integrations): separate crate, but pathed under the tier it belongs to (e.g. `crates/contrib-nodes/common/<integration>/`). Still its own workspace member with its own `Cargo.toml`, but under `contrib-nodes/` so it travels with contrib if/when contrib is extracted to a separate repo. A submodule here couples integrations into one compile unit, one clippy report, one PR diff scope, and forces `#[cfg(any(feature = "...-receiver", feature = "...-exporter"))]` gates to be kept in sync as consumers come and go.
4. **Consumers across the core/contrib boundary** (e.g. Laurent's comment above about the topic abstraction needing Kafka connectivity; also grpc/http client settings if/when both tiers want them): extract just that piece to a top-level crates/common/<x>/. Don't preemptively populate it; the trigger should be two real cross-tier consumers, not a speculative one. The "promote to core when contrib needs it" alternative drags core's dep graph and feature surface into contrib for a small utility, turns core's internal helpers into a de-facto stable API for contrib consumers

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
c1ly: didn't know we can use the effect handler to start a periodic timer, updated to use this and the timer tick to trigger commits

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
c1ly: updated to include a todo comment describe possible enhancement in the future

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,2672 @@
...
+                                let (topic_id, partition, offset) =
+                                    decode_calldata(&ack_msg.unwind.route.calldata);
+                                // Resolve the dynamic topic ID back to
+                                // the actual topic name via the registry.
+                                if let Some(name) = self.topic_registry.name_for(topic_id) {
+                                    let advanced = self
+                                        .offset_tracker
+                                        .acknowledge(name, partition, offset);
+                                    if advanced {
+                                        self.commit_offsets(&consumer, &receiver_id)?;
```
lalitb: I think commit failures should be recoverable here, instead of bubbling out of receiver loop and stop the receiver.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,2672 @@
...
+        consumer: StreamConsumer<C>,
+    ) -> Result<TerminalState, EngineError> {
+        // Start periodic telemetry collection
+        let telemetry_cancel_handle = effect_handler
+            .start_periodic_telemetry(Duration::from_secs(1))
+            .await?;
+        let topics = self.config.all_topics();
+
+        // Subscribe to the configured topics
+        consumer.subscribe(&topics).map_err(|e| {
```
lalitb, line 636: I’m a bit worried about manual offset tracking during consumer-group rebalances. The receiver keeps pending offset state in memory, but I don’t see callbacks that commit/clear that state when partitions are revoked or reassigned. Could we add explicit rebalance handling here so the receiver does not keep committing offsets for partitions it no longer owns?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,2672 @@
...
+                                        .acknowledge(name, partition, offset);
+                                    if advanced {
+                                        self.commit_offsets(&consumer, &receiver_id)?;
+                                    }
+                                }
+                            }
+                        },
+                        Ok(NodeControlMsg::Nack(nack_msg)) => {
+                            self.metrics.nacks_received.add(1);
+                            // Treat nack as ack (advance past failed message).
```
lalitb, line 746: This is a bit confusing and took me some time to understand. Can we document this behavior more clearly? Manual commit mode is described as at-least-once, but a `nack` is treated like an ack here, so a message that failed downstream can still be skipped permanently once the offset advances.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,2672 @@
...
+    }
+
+    /// Get or assign a `u8` ID for the given topic name.
+    fn get_or_assign(&mut self, topic: &str) -> u8 {
+        // if topic hasn't been seen yet then we assign topic a id
+        if let Some(&id) = self.name_to_id.get(topic) {
+            return id;
+        }
+        // get id for a new topic
+        let id = self.id_to_name.len() as u8;
```
lalitb: This uses a `u8` topic ID, so it can only represent 256 distinct topic names. With regex topic subscriptions, the receiver may discover more real topics over time, and then this cast can wrap around. Could we use a wider ID or return a clear error before assigning a duplicate ID?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/metrics.rs
```diff
@@ -0,0 +1,71 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Metrics for the Kafka Receiver node.
+
+use otap_df_telemetry::instrument::Counter;
+use otap_df_telemetry_macros::metric_set;
+
+/// Metrics for the Kafka Receiver.
+#[metric_set(name = "kafka.receiver.metrics")]
```
lalitb: Small naming consistency question: other contrib receiver metric sets seem to use `receiver.<name>` style names, while this uses `kafka.receiver.metrics`. Should this be `receiver.kafka` before the metric names become user-visible?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,2672 @@
...
+                            if idempotent
+                                && self
+                                    .offset_tracker
+                                    .is_known_offset(&topic, partition, offset)
+                            {
+                                self.metrics.idempotent_skips.add(1);
+                                continue;
+                            }
+
+                            match self.process_kafka(data, capture_policy) {
```
lalitb, line 833: Small performance question: `process_kafka` does the decode and optional header injection inline on the receiver loop. For large OTAP/OTLP payloads, could that keep the receiver task busy long enough to slow down polling/control-message handling? Not necessarily something to change in v1, but should we consider offloading heavier decode work or tracking it as a follow-up?

ThomsonTan (commented):
This is duplicated comment.

ThomsonTan:
Add an entry to .chloggen/?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
c1ly: extended to u32 and emits a error log when wrap around is detected

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/metrics.rs
c1ly: updated to receiver.kafka to match convention

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
c1ly: checks for error and emits a error log instead

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
c1ly: added custom client context to handle rebalancing

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
c1ly: updated documentation, added section mentioning the use of the retry processor to add retry support for the kafka receiver

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
c1ly: decoding could take long given larger payloads, added a todo

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
lalitb: Thanks, this is the right direction.

 I still see one gap. The rebalance callback queues revoked partitions, but `commit_offsets()` still commits the whole tracker until the receive loop processes that queue.

If TimerTick, shutdown commit, or poison pill handling runs before that cleanup, this receiver can still commit offsets for a partition it no longer owns.

Can you scope commits to the current assigned partitions, or drain revoked partitions before any commit path runs?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,3286 @@
...
+                            // track the offset even on decode failure.
+                            let topic = data.topic().to_owned();
+                            let partition = data.partition();
+                            let offset = data.offset();
+
+                            // Throughput metrics: count every received
+                            // message and its payload size.
+                            self.metrics.messages_received.add(1);
+                            if let Some(payload) = data.payload() {
+                                self.metrics.bytes_received.add(payload.len() as u64);
```
lalitb, line 799: I see `messages_received` and `bytes_received` are incremented twice in the same Ok(data) branch. That makes the `receiver.kafka` counters report double the real consumed messages and bytes.

Can you remove one of the two metric blocks?

lalitb:
Thanks @c1ly , this is a big improvement. The receiver is much easier to reason about now. I left one blocking comment on the [rebalance commit path](https://github.com/open-telemetry/otel-arrow/pull/3261#discussion_r3476170953). Once that is addressed, I think this should be close.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
c1ly: `commit_offsets()` now drains the revoked partitions before getting any committable offsets and committing

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
c1ly: removed duplicate metric block

gouslu:
Extensions are supported now so auth can be an extension, as it seems like both #3261 and #3262 share the same auth code.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/README.md
```diff
@@ -0,0 +1,925 @@
...
+The tables below summarize
+the **feature gaps** relative to the Go receiver; the
+[offset/error-handling mapping](#offset-and-error-handling) and
+[Known gaps](#known-gaps--behavioral-differences) further down cover the
+offset-commit and retry differences.
+
+##### Signals
+
+The Go receiver also consumes a `profiles` signal. This receiver supports
+`traces`, `metrics`, and `logs` only -- there is no `profiles` signal.
```
lquerel:
```suggestion
The Go receiver also consumes a `profiles` signal. This receiver supports
`traces`, `metrics`, and `logs` only -- support for `profiles` has not yet been added..
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,3424 @@
...
+                                );
+                            }
+                            // Bound staleness of the rebalance commit snapshot
+                            // to the commit interval.
+                            self.refresh_committable_snapshot();
+                        },
+                        Err(e) => {
+                            return Err(EngineError::ChannelRecvError(e));
+                        }
+                        _ => {
```
lquerel, line 778: You should explicitly handle the `NodeControlMsg::DrainIngress` event. The engine's receiver-first shutdown protocol sends DrainIngress and waits for notify_receiver_drained() before shutting down downstream nodes. Because this receiver ignores the message, it continues polling Kafka and admitting new data until the shutdown deadline forces termination.

Please handle DrainIngress explicitly: stop Kafka ingress, perform a bounded drain/final commit, call notify_receiver_drained(), and remain responsive to the eventual Shutdown message.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/config.rs
```diff
@@ -0,0 +1,2616 @@
...
+            _ = config.set("group.instance.id", instance_id);
+        }
+
+        // Commit settings derived from CommitConfig
+        let auto_commit = matches!(self.commit.mode, CommitMode::Auto);
+        if auto_commit {
+            // When interval_ms is None in auto mode, default to 0 (commit as
+            // often as possible).
+            let interval = self.commit.interval_ms.unwrap_or(0);
+            _ = config.set("auto.commit.interval.ms", interval.to_string());
```
lquerel: In librdkafka, `auto.commit.interval.ms=0` disables periodic commits; it does not mean "commit as often as possible". Consequently, `commit.mode: auto` with no explicit interval stores offsets in memory but never commits them to Kafka.
Please either omit this property so librdkafka retains its positive default, or supply an explicit positive default. The current test expecting "0" should be updated as well.

Reference: https://docs.confluent.io/platform/current/clients/librdkafka/html/md_CONFIGURATION.html

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,3424 @@
...
+        // *and* an explicit interval was configured. When no interval is
+        // set in manual mode, offsets are committed purely via ack/nack.
+        // The timer delivers `NodeControlMsg::TimerTick` on the control
+        // channel, which is handled in the main loop below.
+        if manual_commit {
+            if let Some(ms) = self.config.commit_interval_ms() {
+                let _commit_timer_handle = effect_handler
+                    .start_periodic_timer(Duration::from_millis(ms))
+                    .await?;
+            }
```
lquerel, line 665: Configuration validation currently accepts `commit.interval_ms: 0`. In manual mode that reaches this call as Duration::ZERO. The engine's recurring timer loop calculates the next expiration by adding the duration; with zero, the expiration never advances and fire_due loops indefinitely on the single pipeline thread.

Please reject zero during configuration validation, consistent with the README's documented > 0 requirement, and change the test that currently asserts zero is valid.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,3424 @@
...
+        }
+        // Drop any partitions revoked by the rebalance callback since the last
+        // reconcile *before* building the commit list, so we never commit an
+        // offset for a partition this consumer no longer owns.
+        self.purge_revoked_partitions();
+        let tpl = self.offset_tracker.committable_tpl();
+        if tpl.count() == 0 {
+            return Ok(());
+        }
+        match consumer.commit(tpl, CommitMode::Sync) {
```
lquerel: `CommitMode::Sync` blocks the caller until the broker finishes processing the request. This method is invoked from the normal ACK/NACK path, potentially once for every advancing offset, so broker latency can stall data processing, control messages, telemetry, and shutdown on the pipeline's single-thread runtime.

Please use asynchronous commits for steady-state progress and observe failures through the consumer commit callback. Any synchronous final flush should be bounded, deadline-aware, and kept off the pipeline core.

Reference: https://docs.rs/rdkafka/latest/rdkafka/consumer/enum.CommitMode.html

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/rebalance.rs
```diff
@@ -0,0 +1,588 @@
...
+            .partitions_revoked
+            .fetch_add(revoked.len() as u64, Ordering::Relaxed);
+    }
+
+    /// Handle a `post_rebalance` assign: replace the assigned set with the new
+    /// assignment.
+    fn handle_assign(&self, tpl: &TopicPartitionList) {
+        let assigned_partitions = topic_partitions(tpl);
+        let mut assigned = self.lock_assigned();
+        assigned.clear();
```
lquerel: Kafka can rebalance by reporting only what changed instead of returning every partition the consumer still owns. For example, suppose this consumer already owns partitions 0 and 1; after a rebalance it keeps partition 0 and gains partition 2. Kafka may report only [2] because partition 0 never moved.

Clearing the assignment here would leave the receiver believing it owns only partition 2, even though it still owns partition 0. ACK/NACK feedback for partition 0 would then be rejected as belonging to a revoked partition, preventing its offsets from being committed.

Please maintain the assignment by applying additions and removals incrementally, or query `base_consumer.assignment()` after the rebalance and store that complete assignment.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
```diff
@@ -0,0 +1,3424 @@
...
+    /// the start of every commit (via [`commit_offsets`](Self::commit_offsets)),
+    /// so no commit path can ever persist an offset for a partition this
+    /// consumer no longer owns — even if the revocation was queued by the
+    /// callback after the last loop-top reconcile (e.g. just before a
+    /// `TimerTick`, shutdown commit, or poison-pill advance).
+    fn purge_revoked_partitions(&mut self) {
+        if self.config.is_auto_commit() {
+            return;
+        }
+        let revoked = self.rebalance_state.drain_revoked();
```
lquerel, line 473: This queue identifies a revocation only by topic and partition. It does not record which assignment the revocation belonged to. A partition can be revoked during a rebalance and later assigned back to the same consumer before this queue is processed.

For example:
1. Partition 0 is revoked and added to this queue.
2. Kafka assigns partition 0 back to this consumer.
3. The receiver accepts and tracks a new record from partition 0.
4. This method processes the old queued revocation and deletes the newly created tracking state.

The new record’s ACK then has no matching tracked offset. A later record can advance the committed offset past work that is still outstanding, breaking the receiver’s at-least-once guarantee.

Please distinguish different ownership periods (e.g. with an assignment generation) and ensure revocations are applied before accepting records from a newer assignment. ACK/NACK metadata should carry the same generation so feedback from an earlier assignment cannot affect the current one.

lquerel (changes requested):
Thank you for this substantial contribution. This PR establishes a strong foundation for an experimental Kafka receiver, with a broad configuration surface, useful documentation, and good initial unit-test coverage!

I'm requesting some changes for the correctness and runtime issues raised in the inline comments. Addressing them is important to preserve offset guarantees, rebalance correctness, runtime responsiveness, and graceful shutdown behavior.

Looking ahead, moving the receiver from experimental toward production-ready will require more extensive validation. This does not need to happen in this PR, but I suggest tracking the following areas:
- [ ] Offset guarantees: manual and automatic commits, out-of-order ACK/NACKs, poison messages, commit failures, restarts, and process crashes.
- [ ] Consumer-group rebalancing: adding/removing consumers, partitions returning to the same consumer, incremental reassignments, and in-flight records crossing assignment changes.
- [ ] Lifecycle behavior: drain and shutdown under sustained traffic, bounded shutdown deadlines, final commits, and broker unavailability during shutdown.
- [ ] Failure recovery: broker restarts, network interruptions, authentication failures, commit timeouts, and prolonged outages.
- [ ] Backpressure and performance: throughput and latency benchmarks, slow downstream consumers, high partition counts, bounded memory usage, and verification that broker operations never stall the pipeline core.
- [ ] Routing and payload correctness: all supported encodings, malformed payloads, topic regular expressions and exclusions, header extraction, and mixed-signal configurations.
- [ ] Compatibility: representative Kafka versions, Linux/macOS/Windows builds, TLS and supported authentication modes.
- [ ] Operational visibility: validation that metrics and error reporting accurately distinguish expected filtering, transient failures, rebalances, commit failures, and data-processing errors.

A focused integration suite using a real Kafka broker (especially for offset, rebalance, failure, and shutdown scenarios) will be the most important next step in building confidence. Overall, this is promising work, and I'm looking forward to seeing the receiver mature through that validation.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
c1ly: use commit mode async

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/rebalance.rs
c1ly: calls assignment to get the full list of owned partitions

rust/otap-dataflow/crates/contrib-nodes/src/receivers/kafka_receiver/receiver.rs
c1ly: added a generation tracker for commit messages and the offset tracker so each partition stores their generation   (version number). Partitions get a new generation number when they are assigned or reassigned during rebalancing

lquerel (approved):
Thanks for these changes.

PR #3262, opened 2026-06-10
[otap-dataflow] add kafka exporter into contrib-nodes

### LINKED ISSUE #3249: [otap-dataflow] Add Kafka Exporter
### Kafka Exporter

The Kafka exporter takes pipeline data (traces, metrics, logs) and produces it to Kafka topics.

#### Per-Signal Configuration

Each signal type has its own topic, encoding, dynamic topic routing, and partitioning settings. At least one signal must be configured. Unconfigured signal types are permanently nacked (non-retryable).

```yaml
traces:
  topic: "otlp-traces"
  encoding: "otlp_proto"
metrics:
  topic: "otlp-metrics"
  encoding: "otap_proto"
logs:
  topic: "otlp-logs"
  encoding: "otlp_proto"
```

#### Encoding Formats

Same two formats as the receiver: `otlp_proto` (default) and `otap_proto`.

#### Message Format Header

When `message_format_header` is configured, the exporter writes a header on every outgoing Kafka message with the encoding format value (`otlp` or `otap`). This allows downstream consumers (including the Kafka receiver) to detect the encoding per-message.

```yaml
message_format_header: "MessageFormat"
```

When not configured the default value `MessageFormat` is used. Transport headers with the same name as the configured format header are automatically excluded from propagation to avoid collision.

#### Dynamic Topic Routing

Per-signal `topic_from_transport_header` specifies a normalized header name. When the header is present in the pdata context, its value becomes the destination Kafka topic instead of the static `topic` field.

```yaml
traces:
  topic: "otlp-traces"
  topic_from_transport_header: "x_target_topic"
```

Invalid header-derived topics (empty, `.`, `..`, invalid characters, exceeds 249 chars) fall back to the static topic. Each signal can route independently using different header keys.

#### Partition Key from Transport Headers

When `partition_by_transport_headers` is enabled per-signal, all transport headers from the pdata context are serialized into a deterministic binary key used as the Kafka message key.

```yaml
traces:
  topic: "otlp-traces"
  partition_by_transport_headers: true
```

Properties of the partition key:

- **Order-independent** — Headers are sorted by `(name, value)` before serialization
- **Case-insensitive** — Uses normalized (lowercase) header names
- **Unambiguous** — Length-prefixed binary format with no delimiter collision risk
- **Null key when empty** — When no transport headers are present, the Kafka key is null, ensuring true round-robin under all partitioner strategies

#### Transport Header Propagation

When a propagation policy is configured at the pipeline level and the pdata context carries transport headers, matching headers are propagated as Kafka record headers on the outgoing message.

#### Partitioning Strategies

Configurable librdkafka partitioner algorithm (default: `consistent_random`):

`random`, `consistent`, `consistent_random`, `murmur2`, `murmur2_random`, `fnv1a`, `fnv1a_random`

#### Producer Configuration

- `compression` — `gzip`, `snappy`, `lz4`, `zstd`
- `required_acks` — `none` (fire-and-forget), `one` (leader ack), `all` (all ISR ack)
- `timeout_ms`, `max_message_bytes`, `linger_ms` — Standard producer tuning

#### Authentication and TLS

Same capabilities as the receiver: SASL (AWS MSK IAM, PLAIN, SCRAM-SHA-512), mTLS, auto-inferred security protocol.

#### Additional configuration

`producer_config` accepts arbitrary librdkafka key-value pairs. The exporter warns when keys conflict with first-class config fields. The `partitioner` key is included in the managed key set to prevent silent overwrites.

c1ly:
# Change Summary

Add Kafka Exporter implementation that takes inspiration from the go and rotel versions.

Add kafka_util that shares common functions and data types with the kafka receiver and exporter

* Closes #3249 

unit tests and integration tests with kafka broker (requires docker container)

no user face changes


lquerel:
@c1ly I merged yesterday a series of README.md files related to the documentation of nodes. Could you update the README.md in contrib-nodes and add/update a README.md file per node in their respective module directory. Thanks

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/encoder.rs
```diff
@@ -0,0 +1,94 @@
...
+        let otlp_bytes = OtlpProtoBytes::ExportTracesRequest(bytes.clone().into());
+        let payload = OtapPayload::OtlpBytes(otlp_bytes);
+
+        let result = encode_to_otlp_bytes(payload).expect("encoding should succeed");
+        assert_eq!(result, bytes);
+    }
+
+    // NOTE: Full roundtrip conversion tests for OtapArrowRecords <-> OtlpProtoBytes/BatchArrowRecords are covered upstream in the
+    // otel-arrow OSS repo: https://github.com/open-telemetry/otel-arrow
+    // No additional test coverage needed here (see MR discussion).
```
albertlockett: I think this coment can now be removed

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/partitioner.rs
```diff
@@ -0,0 +1,339 @@
...
+    None
+}
+
+// TODO: Explore `partition_by_trace_id` — partition traces by hex-encoded trace ID.
+//   Trace IDs are 16-byte FixedSizeBinary values in the OTAP Arrow Spans schema.
+//   A single OTAP batch can contain spans with different trace IDs, so implementing
+//   this requires splitting the batch into sub-batches grouped by trace ID (returning
+//   something like `Vec<(String, RoaringBitmap)>`) before sending each sub-batch to
+//   Kafka with its own partition key. Empty/zero trace IDs should map to an empty
+//   partition key (round-robin).
```
albertlockett, line 99: :+1: this would be good to document as a github issue so we don't lose track of the requirement.

FYI I was planning on implementing something like this (probably in query-engine) where we could partition the batch based on some arbitrary OPL expression. I bet we could reuse that for this, rather than building something bespoke

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/exporter.rs
```diff
@@ -0,0 +1,917 @@
+//! Core Kafka exporter implementation.
```
albertlockett, line 4: Can you add the OTEL copyright header to these rust files?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/partitioner.rs
albertlockett: FYI documented an issue for how we might pre-partition the batch upstream of this processor using OPL/Query-engine https://github.com/open-telemetry/otel-arrow/issues/3273

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/topic_router.rs
```diff
@@ -0,0 +1,343 @@
...
+            if validate_kafka_topic(&topic).is_ok() {
+                metrics.inc_topic_from_header();
+                return Cow::Owned(topic);
+            }
+            // Invalid topic from header -- fall back to static config.
+            tracing::warn!(
+                header_topic = %topic,
+                static_topic = signal_config.topic(),
+                "invalid Kafka topic from transport header, falling back to static topic"
+            );
```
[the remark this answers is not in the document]
c1ly: added todo

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/exporter.rs
```diff
@@ -0,0 +1,936 @@
...
+        let topic = TopicRouter::resolve(signal_config, &context, &mut self.metrics);
+
+        // Clone payload for ack/nack reporting -- send_to_kafka consumes the original.
+        let payload_for_reporting = payload.clone();
+
+        // Encode payload to bytes using the per-signal encoding.
+        // This block borrows &mut self.pdata_producer so it must complete
+        // before we borrow self.config again for the topic reference below.
+        let payload_bytes = match encoding {
+            MessageFormat::OtlpProto => encoder::encode_to_otlp_bytes(payload.clone())?,
```
lalitb: If encoding fails, seems we are return early without sending `ack` or `nack` here. Is this intended?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/topic_router.rs
```diff
@@ -0,0 +1,346 @@
...
+        context: &Context,
+        metrics: &mut KafkaExporterMetrics,
+    ) -> Cow<'a, str> {
+        // Priority 1: Transport header for this signal type
+        if let Some(topic) = Self::resolve_from_header(signal_config, context) {
+            if validate_kafka_topic(&topic).is_ok() {
+                metrics.inc_topic_from_header();
+                return Cow::Owned(topic);
+            }
+            // Invalid topic from header -- fall back to static config.
```
lalitb: I’m not sure we should silently fall back to the static topic when the routing header is present but invalid. That can send data to a different topic than the request asked for, which seems risky for tenant/request-based routing. I’d prefer treating this as an export failure, likely a permanent `nack`, so the bad header is visible instead of rerouting the data.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/partitioner.rs
```diff
@@ -0,0 +1,342 @@
...
+    // only in original casing produce the same partition key.
+    let mut sorted: Vec<_> = headers.iter().collect();
+    sorted.sort_unstable_by(|a, b| a.name.cmp(&b.name).then_with(|| a.value.cmp(&b.value)));
+
+    let mut buf = Vec::new();
+    for header in &sorted {
+        buf.extend_from_slice(&(header.name.len() as u32).to_be_bytes());
+        buf.extend_from_slice(header.name.as_bytes());
+        buf.extend_from_slice(&(header.value.len() as u32).to_be_bytes());
+        buf.extend_from_slice(&header.value);
```
lalitb: This may have sensitive auth data, seems it will get persisted to Kafka, and/or visible to consumers. Better to hash the serialize header in the case?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/exporter.rs
```diff
@@ -0,0 +1,936 @@
...
+            .start_periodic_telemetry(Duration::from_secs(1))
+            .await?;
+
+        let ack_nack_reporter = EffectHandlerReporter::new(&effect_handler);
+
+        // Main event loop
+        loop {
+            match inbox.recv().await? {
+                Message::PData(pdata) => {
+                    _ = self
```
lalitb, line 527: This appears to wait for each Kafka delivery before reading the next pdata message, so the exporter only has one Kafka record in flight at a time. That is simple and gives\ backpressure, but it may limit throughput quite a bit compared with the Go exporter and rotel patterns that allow multiple sends in flight. Is that intentional for the first version, or should we track delivery futures separately and ack/nack as they complete?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/producer.rs
```diff
@@ -0,0 +1,715 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+/// This is a port of the rdkafka ThreadedProducer and FutureProducer located at https://github.com/fede1024/rust-rdkafka/blob/master/src/producer/future_producer.rs.
```
lalitb, line 4: Since this file is a port/copy of `rust-rdkafka`’s producer implementation, can we preserve the upstream MIT copyright/license notice somewhere appropriate.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/metrics.rs
```diff
@@ -0,0 +1,202 @@
...
+///
+/// Metric set name: `gateway.exporter.kafka.pdata.metrics`
+///
+/// Individual metric names (after field name translation `_` → `.`):
+/// - `logs.exported` / `logs.failed`
+/// - `metrics.exported` / `metrics.failed`
+/// - `traces.exported` / `traces.failed`
+/// - `acks.received`
+/// - `nacks.received`
+#[metric_set(name = "gateway.exporter.kafka.pdata.metrics")]
```
lalitb: Small consistency question: this metric set is named `gateway.exporter.kafka.pdata.metrics`, but other contrib nodes use names like `exporter.azure_monitor`, `receiver.user_events`, and `processor.resource_validator`. Should this follow the same shape, e.g. `exporter.kafka` before the metric names become visible to users?

ThomsonTan (commented):
Add an entry to .chloggen/.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/topic_router.rs
c1ly: updated to permanent nack when invalid dynamic topic is found

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/exporter.rs
c1ly: nope, updated to nack on encoding fail

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/metrics.rs
c1ly: updated to exporter.kafka to match convention

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/partitioner.rs
c1ly: partition key is now created by hashing the header

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/exporter.rs
c1ly: this is intentional for the first version, will do a follow up on the kafka receiver and exporter that improves their throughput. Added a todo note

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/producer.rs
c1ly: updated third_party_notice.txt

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/config.rs
```diff
@@ -0,0 +1,1756 @@
...
+    "debug",
+];
+
+/// Per-signal exporter configuration.
+///
+/// Each signal type (traces, metrics, logs) can optionally have its own topic
+/// and encoding format. When a signal is `None` in [`KafkaExporterConfig`],
+/// that signal type will not be exported.
+#[derive(Debug, Clone, Deserialize, PartialEq)]
+pub struct SignalConfig {
```
lalitb, line 49: nit - Should this also use `#[serde(deny_unknown_fields)]`? The top-level config rejects unknown fields and the README says unknown configuration fields are rejected, but unknown fields inside `traces` / `metrics` / `logs` would still be ignored here. That could hide typos like `topic_from_transpot_header` and silently disable routing/partitioning.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/exporter.rs
```diff
@@ -0,0 +1,1036 @@
...
+
+        headers
+    }
+
+    /// Exports a single PData message to Kafka with Ack/Nack support.
+    ///
+    /// Uses the [`TopicRouter`] to resolve the destination topic(s):
+    /// 1. Transport header (highest priority)
+    /// 2. Resource attribute (may split the batch)
+    /// 3. Static per-signal topic (fallback)
```
lalitb: Looks like this comment is still stale. The current router only supports transport header routing and static fallback; resource-attribute routing / batch splitting is not implemented. Can we update this so the generated docs do not describe a routing mode that is not present yet?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/config.rs
```diff
@@ -0,0 +1,1756 @@
...
+
+    /// The transport header name for dynamic topic routing, if set.
+    #[must_use]
+    pub fn topic_from_transport_header(&self) -> Option<&str> {
+        self.topic_from_transport_header.as_deref()
+    }
+
+    /// Set the transport header name for dynamic topic routing.
+    #[must_use]
+    pub fn with_topic_from_transport_header(mut self, key: impl Into<String>) -> Self {
```
lalitb, line 117: Should we normalize this header name here, or validate that it is already normalized? The router compares this config value against `header.name`, which is normalized, so a natural config like `X-Target-Topic` would silently never match and would fall back to the static topic.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/exporter.rs
```diff
@@ -0,0 +1,1036 @@
...
+    fn build_kafka_headers(
+        encoding: MessageFormat,
+        format_header_key: &str,
+        context: &otap_df_otap::pdata::Context,
+        effect_handler: Option<&EffectHandler<OtapPdata>>,
+    ) -> OwnedHeaders {
+        let mut headers = OwnedHeaders::new();
+
+        // Always write the message format header.
+        let format_value = match encoding {
```
lalitb, line 286: Can we confirm this is the intended behavior? Issue #3249 says no format header is written when `message_format_header` is not configured, but the current config gives it a default and this path always writes it. Either behavior seems fine, but the issue/design and implementation should match.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/config.rs
c1ly: added #[serde(deny_unknown_fields)]

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/config.rs
c1ly: updated the try_from for the exporter config and exporter config builder to normalize the user's input so we can safely compare against header.name

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/exporter.rs
c1ly: updated comment and other stale documentation relating to the topic router

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/exporter.rs
c1ly: updated issue, message_format_header should use default value `MessageFormat` when the user doesn't set it

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/topic_router.rs
```diff
@@ -0,0 +1,410 @@
...
+    fn resolve_from_header(signal_config: &SignalConfig, context: &Context) -> Option<String> {
+        // `topic_from_transport_header` is pre-normalized (lowercased) in
+        // `KafkaExporterConfig::try_from`, matching how transport headers store
+        // their logical names, so a plain equality check is sufficient here.
+        let header_key = signal_config.topic_from_transport_header()?;
+        let transport_headers = context.transport_headers()?;
+
+        for header in transport_headers.iter() {
+            if header.name == *header_key {
+                return header.value_as_str().map(String::from);
```
lalitb: I think this still falls back to the static topic when the routing header exists but its value is not valid UTF-8.

`value_as_str()` returns None here. Then `resolve_from_header` treats the header as missing, so the exporter uses the static topic instead of rejecting the message.

I believe we should treat this as a routing error and permanent nack. That keeps malformed dynamic-topic headers from sending data to the fallback topic.

lalitb:
Thanks @c1ly, this is much clearer now. Most of the earlier comments look addressed. I left one blocking [comment](https://github.com/open-telemetry/otel-arrow/pull/3262#discussion_r3476257737) on the dynamic topic header path. Once that is fixed, I think this is close.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/topic_router.rs
c1ly: checked for invalid utf-8 string now, if invalid emit as topic routing error so the message is permanently nacked

gouslu:
Extensions are supported now so auth can be an extension, as it seems like both #3261 and #3262 share the same auth code.

lalitb:
> Extensions are supported now so auth can be an extension, as it seems like both #3261 and #3262 share the same auth code.

Thanks @gouslu - that's a good point. I think we can keep these two PRs focused and handle extension-based auth in a follow-up. That follow-up can cover the new capability, the extension, config binding, and the rdkafka ClientContext wiring together.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/kafka_exporter/exporter.rs
```diff
@@ -0,0 +1,1042 @@
...
+            .headers(headers)
+            .payload(&payload_bytes);
+        // only set the partition key if it isn't none
+        if let Some(ref key) = partition_key {
+            record = record.key(key);
+        }
+
+        // Send to Kafka with timeout
+        let timeout = Duration::from_millis(self.config.timeout_ms());
+        match self.producer.send(record, timeout).await {
```
lquerel, line 426: Awaiting the delivery future inline prevents this exporter from reading shutdown and telemetry control messages until Kafka completes the delivery. Because timeout_ms is also assigned to message.timeout.ms and currently accepts 0 (which librdkafka defines as an infinite timeout) a broker outage can make this exporter unable to shut down indefinitely.

Please reject zero and unreasonably large timeouts (I suggest 30s max). On shutdown, stop accepting data, drain until the supplied deadline, then purge and nack anything remaining.

lquerel (approved):
I'm approving this PR because the Kafka exporter is explicitly marked as **Experimental**, with the understanding that this is not yet an endorsement for production use. Before promoting its stability, we should establish a consolidation plan with focused integration, failure-injection, and load tests covering:
- Security: constrain dynamic topic routing, prevent uncontrolled topic creation, avoid sensitive-header fingerprinting, and sanitize adversarial values in telemetry.
- Shutdown and live reconfiguration: verify deadline-aware cancellation, draining, flushing, producer-thread termination, and replacement while deliveries are in flight.
- Retry correctness: validate integration with the Retry Processor, including transient/permanent error classification, retry exhaustion, and duplicate-delivery behavior.
- Backpressure and resource bounds: exercise bounded concurrency, queue saturation, prolonged broker outages, and memory behavior under sustained load.
- Delivery semantics: test ACK/NACK propagation, ordering, partitioning, timeouts, and potentially-persisted messages across broker and network failures.
- Kafka integration: add broker-backed success and failure tests covering supported encodings, acknowledgements, compression, TLS, SASL, and AWS MSK IAM.
- Telemetry: verify metric names, units and record counts, final shutdown snapshots, and safe structured events.
- Configuration and packaging: validate timeout limits, producer escape-hatch interactions, feature gates, and optional dependency isolation.
- Performance and scalability: benchmark throughput, latency, polling overhead, and multi-core behavior with slow and unavailable brokers.

These areas should be tracked as follow-up work and treated as exit criteria for moving the exporter beyond Experimental status.

PR #3268, opened 2026-06-11
feat(filter-processor): add path and kept-count observability metrics

### LINKED ISSUE #3053: [Filter processor]  Add path and kept-count observability metrics
### Description

Related to #2859 and #2884. These issues cover the broader direction for flow-level kept/dropped signal counters and opt-in node-level item/byte counters. This request is for the filter processor’s more specific observability gaps.

This is intended as a narrow filter-processor improvement that can be delivered independently of the broader flow/node telemetry work in #2859 and #2884.

It would be useful to have filter-processor metrics that show both item-level filtering results and which configured filter paths are being evaluated. 

Existing filtered counters should remain unchanged. The missing pieces are counters such as:

  - log/span signals kept
  - log/span/metric batches seen by the filter processor
  - log/span include-configured batches
  - log/span exclude-configured batches

The include/exclude configured-batch counters should be documented carefully as “this batch went through a configured include/exclude path,” not “records matched this rule.”

ThomsonTan:
# Change Summary

Adds new metrics to the `processor.filter.pdata` metric set to close the observability gaps described in #3053, while leaving existing `*_signals_filtered` counters unchanged.

* Closes #3053 

Added test and and passed local run.

### Changelog


rust/otap-dataflow/crates/core-nodes/src/processors/filter_processor/mod.rs
```diff
@@ -203,14 +242,23 @@ impl local::Processor<OtapPdata> for FilterProcessor {
...
+                        self.metrics
+                            .log_signals_kept
+                            .add(signals_consumed.saturating_sub(signals_filtered));
                     }
                     SignalType::Traces => {
                         self.metrics.span_signals_consumed.add(signals_consumed);
                         self.metrics.span_signals_filtered.add(signals_filtered);
+                        self.metrics
+                            .span_signals_kept
+                            .add(signals_consumed.saturating_sub(signals_filtered));
```
lalitb, line 270: I think this new kept counter exposes an existing inconsistency in the log/span filter return values. With no include/exclude filter configured, the processor keeps all records, but the new `*.signals.kept` metric can report 0. Can we add coverage for the pass-through case and make the kept metric reflect what was actually forwarded?

rust/otap-dataflow/crates/core-nodes/src/processors/filter_processor/mod.rs
```diff
@@ -142,6 +180,7 @@ impl local::Processor<OtapPdata> for FilterProcessor {
 
                 let mut arrow_records: OtapArrowRecords = payload.try_into_with_default()?;
                 arrow_records.decode_transport_optimized_ids()?;
+                self.record_batch_path_metrics(signal);
```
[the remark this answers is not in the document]
lalitb: I think the main issue here is that the metric meaning is not clear yet. If `batches_seen` means “received by this processor,” then it should be recorded before decode/conversion as you said. But the include/exclude configured-path counters probably should not imply that filtering actually ran successfully. Can we make the naming/docs precise for these two cases?

rust/otap-dataflow/.chloggen/filter-processor-path-and-kept-metrics.yaml
```diff
@@ -0,0 +1,25 @@
...
+note: Add filter-processor observability metrics for kept signals, batches seen, and configured include/exclude paths
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [3053]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Adds `log/metric/span_signals_kept`, `log/metric/span_batches_seen`, and
```
lalitb: Can we use the exact emitted metric names here instead of the shorthand, and also update the filter processor `README.md` telemetry table? These are user-facing metrics, so the release note/docs should show the names users will actually query.

rust/otap-dataflow/crates/core-nodes/src/processors/filter_processor/metrics.rs
```diff
@@ -19,6 +67,16 @@ pub struct FilterPdataMetrics {
     #[metric(unit = "{span}")]
     pub span_signals_consumed: Counter<u64>,
 
+    /// Number of log signals kept (consumed minus filtered)
+    #[metric(unit = "{log}")]
+    pub log_signals_kept: Counter<u64>,
```
drewrelmas, line 87: Not a problem specifically about this PR, but we should figure out how to standardize per-signal metric naming conventions.

For example, `durable_buffer` has the following:

https://github.com/open-telemetry/otel-arrow/blob/f3cfd7f866fe6ec6f8ef23520286f385de43ca4c/rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs#L185-L215

`unit`= `{log_record|data_point|span}`
`name` = `{<>_log_records|<>_metric_points|<>_spans}`

cc @cijothomas who I'm sure has thoughts on this matter

drewrelmas: Food for thought, here are existing OTC naming practices: https://opentelemetry.io/docs/collector/internal-telemetry/#additional-detailed-level-metrics

ThomsonTan: I will create a follow up task for standardizing the naming.

rust/otap-dataflow/.chloggen/filter-processor-path-and-kept-metrics.yaml
ThomsonTan: Changed to the exact emitted metric names.

One thing to flag: the existing README rows already use underscores (`processor.filter.pdata.log_signals_consumed`) rather than the canonical dotted names, and the same underscore convention is used across the other node READMEs. I'll switch the filter table to the dotted names here so it matches what users query; happy to file a follow-up for the sibling READMEs rather than widen this PR.

rust/otap-dataflow/crates/core-nodes/src/processors/filter_processor/mod.rs
ThomsonTan: Fixed, the log/span pass-through branch now returns `filtered = 0` so `*.signals.kept` reflects what's forwarded, with pass-through tests added.

rust/otap-dataflow/crates/core-nodes/src/processors/filter_processor/mod.rs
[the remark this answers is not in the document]
ThomsonTan: 1. `*_batches_seen` now means "received": `record_batch_path_metrics` is called *before* conversion/decode, so a batch that later fails to convert/decode is still counted. Docs updated to say "counted on receipt, before conversion/decode".
2. The include/exclude `*_configured_batches` counters are now documented as "received while an include/exclude filter was configured, and does not imply filtering ran successfully or that any records matched". They're incremented on receipt alongside `*_batches_seen`, not as a signal that filtering completed.

README telemetry table synced to match.

jmacd:
@drewrelmas should we update this PR to reflect the progress in https://github.com/open-telemetry/otel-arrow/pull/3454?

drewrelmas:
Yes - @ThomsonTan could you please take a look at https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/telemetry/item-attributes.md and update this implementation accordingly?

drewrelmas:
Discussed offline with @ThomsonTan, this PR is superseded by #4045, am closing it.

PR #3269, opened 2026-06-11
Feature/durable buffer metrics fresh

### LINKED ISSUE #3117: Enhance `durable_buffer` metrics

Dipanshusinghh:
Closes #3117

How are these changes tested?
Verified locally with the following commands:
cargo fmt --all -- --check        ✓ no formatting issues
cargo clippy -p otap-df-core-nodes --tests --features otap-df-otap/crypto-ring   ✓ no warnings or errors
cargo test -p otap-df-core-nodes --features otap-df-otap/crypto-ring              ✓ 723 passed, 0 failed
Key tests passing:

test_storage_utilization_reporting — ok
test_per_signal_dropped_expired_metrics — ok

Are there any user-facing changes?
Yes. The following new metrics are now emitted by durable_buffer_processor:

storage_utilization — storage used/cap ratio (0.0–1.0)
dropped_log_records — log records lost via DropOldest policy
dropped_spans — spans lost via DropOldest policy
dropped_metric_datapoints — metric datapoints lost via DropOldest policy
expired_log_records — log records lost via MaxAge policy
expired_spans — spans lost via MaxAge policy
expired_metric_datapoints — metric datapoints lost via MaxAge policy

Existing dropped_items and expired_items metrics are preserved for backward compatibility.

Changelog

 Added a .chloggen/*.yaml entry, OR this PR is a chore (indicated in title).

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/bundle_adapter.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/bundle_adapter.rs
@@ -82,7 +82,7 @@ const fn to_slot_id(_signal_type: SignalType, payload_type: ArrowPayloadType) ->
 }
 
 /// Convert signal type to OTLP slot ID (for opaque binary storage)
-const fn to_otlp_slot_id(signal_type: SignalType) -> SlotId {
+pub(crate) const fn to_otlp_slot_id(signal_type: SignalType) -> SlotId {
     SlotId::new(match signal_type {
         SignalType::Logs => otlp_slots::OTLP_LOGS,
         SignalType::Traces => otlp_slots::OTLP_TRACES,
@@ -1007,7 +1007,7 @@ mod tests {
         // HashMap iteration order is not guaranteed, but all slots should be from same signal
         #[rustfmt::skip]
         let records = OtapArrowRecords::Traces(traces!(
-            (Spans, 
+            (Spans,
                 ("id", UInt16, vec![0u16, 1])),
             (SpanEvents,
                 ("id", UInt32, vec![0u32, 1]),
--- a/rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs
@@ -304,6 +304,35 @@ pub struct DurableBufferMetrics {
     /// Data may still be recoverable via WAL replay on restart.
     #[metric(unit = "{error}")]
     pub flush_failures: Counter<u64>,
+
+    // ─── Appended metrics (Do not shift index) ──────────────────────────────
+    /// Current storage utilization ratio (0.0 to 1.0).
+    #[metric(unit = "{1}")]
+    pub storage_utilization: Gauge<f64>,
+
+    /// Total log records lost due to force-dropped segments (DropOldest policy).
+    #[metric(unit = "{log_record}")]
+    pub dropped_log_records: ObserveCounter<u64>,
+
+    /// Total spans lost due to force-dropped segments (DropOldest policy).
+    #[metric(unit = "{span}")]
+    pub dropped_spans: ObserveCounter<u64>,
+
+    /// Total metric data points lost due to force-dropped segments (DropOldest policy).
+    #[metric(unit = "{data_point}")]
+    pub dropped_metric_datapoints: ObserveCounter<u64>,
+
+    /// Total log records lost due to expired segments (max_age retention).
+    #[metric(unit = "{log_record}")]
+    pub expired_log_records: ObserveCounter<u64>,
+
+    /// Total spans lost due to expired segments (max_age retention).
+    #[metric(unit = "{span}")]
+    pub expired_spans: ObserveCounter<u64>,
+
+    /// Total metric data points lost due to expired segments (max_age retention).
+    #[metric(unit = "{data_point}")]
+    pub expired_metric_datapoints: ObserveCounter<u64>,
 }
 
 // ─────────────────────────────────────────────────────────────────────────────
@@ -420,7 +449,7 @@ struct CachedSegmentMetrics {
 ///
 /// # Segment Metrics Cache
 ///
-/// To avoid per-tick allocations in `recompute_queued_counters`, this struct
+/// To avoid per-tick allocations in `recompute_metrics`, this struct
 /// maintains a `segment_cache` that maps finalized segment sequences to their
 /// pre-computed per-bundle signal classification.  Because segments are
 /// **immutable** after finalization, the cache entry for a given segment
@@ -729,7 +758,7 @@ impl DurableBuffer {
                 Ok((engine, subscriber_id)) => {
                     // Seed queued gauges immediately so they reflect
                     // persisted state before the first telemetry tick.
-                    self.recompute_queued_counters(&engine, &subscriber_id);
+                    self.recompute_metrics(&engine, &subscriber_id);
 
                     self.engine_state = EngineState::Ready {
                         engine,
@@ -777,7 +806,7 @@ impl DurableBuffer {
     /// 3. `open_segment_metrics` — engine open-segment lock (brief snapshot).
     ///
     /// After snapshots are taken, all iteration is lock-free.
-    fn recompute_queued_counters(&mut self, engine: &QuiverEngine, subscriber_id: &SubscriberId) {
+    fn recompute_metrics(&mut self, engine: &QuiverEngine, subscriber_id: &SubscriberId) {
         self.segment_cache_generation = self.segment_cache_generation.wrapping_add(1);
         let current_generation = self.segment_cache_generation;
 
@@ -867,11 +896,15 @@ impl DurableBuffer {
                 }
             };
 
+            let mut seg_logs = 0;
+            let mut seg_metrics = 0;
+            let mut seg_spans = 0;
+
             // Fast path: no bundles resolved → use precomputed totals.
             if progress.resolved_count() == 0 {
-                logs += summary.total_logs;
-                metrics += summary.total_metrics;
-                spans += summary.total_spans;
+                seg_logs = summary.total_logs;
+                seg_metrics = summary.total_metrics;
+                seg_spans = summary.total_spans;
             } else {
                 // Slow path: iterate per-bundle, skipping resolved.
                 for (idx, &(item_count, signal)) in summary.bundles.iter().enumerate() {
@@ -887,14 +920,18 @@ impl DurableBuffer {
 
                     if !progress.is_resolved(BundleIndex::new(bundle_idx)) {
                         match signal {
-                            Some(SignalType::Logs) => logs += item_count,
-                            Some(SignalType::Metrics) => metrics += item_count,
-                            Some(SignalType::Traces) => spans += item_count,
+                            Some(SignalType::Logs) => seg_logs += item_count,
+                            Some(SignalType::Metrics) => seg_metrics += item_count,
+                            Some(SignalType::Traces) => seg_spans += item_count,
                             None => {}
                         }
                     }
                 }
             }
+
+            logs += seg_logs;
+            metrics += seg_metrics;
+            spans += seg_spans;
         }
 
         // Step 3: add items from the open (accumulating) segment.
@@ -921,6 +958,58 @@ impl DurableBuffer {
         // Step 4: evict cache entries for segments no longer tracked.
         self.segment_cache
             .retain(|seq, _| progress_snapshot.contains_key(&SegmentSeq::new(*seq)));
+
+        self.metrics
+            .dropped_segments
+            .observe(engine.force_dropped_segments());
+        self.metrics
+            .dropped_bundles
+            .observe(engine.force_dropped_bundles());
+        self.metrics
+            .dropped_items
+            .observe(engine.force_dropped_items());
+        self.metrics
+            .expired_bundles
+            .observe(engine.expired_bundles());
+        self.metrics.expired_items.observe(engine.expired_items());
+
+        // Update dropped and expired metrics by iterating the engine's per-slot snapshots
+        // and aggregating by signal type via signal_type_from_slot_id(). This handles all
+        // slot ranges (Arrow 10-45 and OTLP 60-62) without hardcoding any slot IDs here.
+        let mut dropped_logs = 0u64;
+        let mut dropped_metrics = 0u64;
+        let mut dropped_spans = 0u64;
+        for (slot, count) in engine.force_dropped_items_per_slot_snapshot() {
+            match signal_type_from_slot_id(slot) {
+                Some(SignalType::Logs) => dropped_logs += count,
+                Some(SignalType::Metrics) => dropped_metrics += count,
+                Some(SignalType::Traces) => dropped_spans += count,
+                None => {}
+            }
+        }
+        self.metrics.dropped_log_records.observe(dropped_logs);
+        self.metrics
+            .dropped_metric_datapoints
+            .observe(dropped_metrics);
+        self.metrics.dropped_spans.observe(dropped_spans);
+
+        let mut expired_logs = 0u64;
+        let mut expired_metrics = 0u64;
+        let mut expired_spans = 0u64;
+        for (slot, count) in engine.expired_items_per_slot_snapshot() {
+            match signal_type_from_slot_id(slot) {
+                Some(SignalType::Logs) => expired_logs += count,
+                Some(SignalType::Metrics) => expired_metrics += count,
+                Some(SignalType::Traces) => expired_spans += count,
+                None => {}
+            }
+        }
+        self.metrics.expired_log_records.observe(expired_logs);
+        self.metrics
+            .expired_metric_datapoints
+            .observe(expired_metrics);
+        self.metrics.expired_spans.observe(expired_spans);
+
         self.metadata_load_warned_segments
             .retain(|seq| progress_snapshot.contains_key(&SegmentSeq::new(*seq)));
         self.enforce_segment_cache_bound();
@@ -1871,6 +1960,14 @@ impl otap_df_engine::local::processor::Processor<OtapPdata> for DurableBuffer {
                     {
                         self.metrics.storage_bytes_used.set(used);
                         self.metrics.storage_bytes_cap.set(cap);
+
+                        let utilization = if cap > 0 {
+                            used as f64 / cap as f64
+                        } else {
+                            0.0
+                        };
+                        self.metrics.storage_utilization.set(utilization);
+
                         self.metrics.dropped_segments.observe(dropped_segs);
                         self.metrics.dropped_bundles.observe(dropped_buns);
                         self.metrics.dropped_items.observe(dropped_items);
@@ -1881,7 +1978,7 @@ impl otap_df_engine::local::processor::Processor<OtapPdata> for DurableBuffer {
                         // registry. This is the single source of truth for
                         // these gauges, correctly accounting for ACKs,
                         // force-drops, and expiry.
-                        self.recompute_queued_counters(&engine, &subscriber_id);
+                        self.recompute_metrics(&engine, &subscriber_id);
                     }
 
                     metrics_reporter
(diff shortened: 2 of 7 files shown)


drewrelmas:
@Dipanshusinghh please take a look at the test failures in `ubuntu-latest` and `macos-latest`. I understand it may be difficult to check them locally if you're developing on Windows.

I believe you should focus attention on these:
```log
  TRY 3 FAIL [   2.032s] (2578/4711) otap-df-otap::durable_buffer_processor_tests test_durable_buffer_mixed_transient_and_permanent_nacks
  TRY 3 FAIL [   1.831s] (2579/4711) otap-df-otap::durable_buffer_processor_tests test_durable_buffer_permanent_nack_rejects_without_retry
  TRY 3 FAIL [   2.205s] (2690/4711) otap-df-otap::durable_buffer_processor_tests test_durable_buffer_retries_on_nack
```

Dipanshusinghh:
Hi @drewrelmas, Ihve completed the requested changes and rechecked the tests could you please review once more? If any further changes are needed please let me know thanks!

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/bundle_adapter.rs
```diff
@@ -113,6 +113,11 @@ pub(crate) fn signal_type_from_slot_id(slot: SlotId) -> Option<SignalType> {
     from_slot_id(slot).map(|(st, _)| st)
 }
 
+pub(crate) const ARROW_LOGS_SLOT: SlotId = SlotId::new(30);
+pub(crate) const ARROW_TRACES_SLOT: SlotId = SlotId::new(40);
+pub(crate) const ARROW_METRICS_SLOT_1: SlotId = SlotId::new(10);
+pub(crate) const ARROW_METRICS_SLOT_2: SlotId = SlotId::new(11);
```
drewrelmas: Please see the following from `arrow.v1.rs`:

https://github.com/open-telemetry/otel-arrow/blob/f3cfd7f866fe6ec6f8ef23520286f385de43ca4c/rust/otap-dataflow/crates/pdata/src/proto/opentelemetry.proto.experimental.arrow.v1.rs#L35-L39

and from `otap.rs` where we specify what Slots to look at in the `num_items()` function:

https://github.com/open-telemetry/otel-arrow/blob/f3cfd7f866fe6ec6f8ef23520286f385de43ca4c/rust/otap-dataflow/crates/pdata/src/otap.rs#L522-L527

https://github.com/open-telemetry/otel-arrow/blob/f3cfd7f866fe6ec6f8ef23520286f385de43ca4c/rust/otap-dataflow/crates/pdata/src/otap.rs#L539-L546

For the purposes of counting number of metric items, I believe you want to use Slots `[11, 12, 13, 14]`, not `[10, 11]`.

To prove this hypothesis, I recommend either updating `test_per_signal_dropped_expired_metrics` or adding 2 new tests that exercise all 3 signal types. I believe the Metric test would fail with the current iteration.

drewrelmas (changes requested):
This looks very close to complete, but had a concern about the Metrics counting. Please take a look, thanks!

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/bundle_adapter.rs
AaronRM: Rather than fix the numbers, I would suggest dropping these constants entirely and avoid hardcoding slots in the processor. Attribute each dropped/expired bundle's `item_count()` to its primary payload slot in the engine, then map slot->signal in the processor via the existing `signal_type_from_slot_id()`.

rust/otap-dataflow/crates/quiver/src/segment_store.rs
```diff
@@ -146,12 +148,18 @@ impl SegmentHandle {
             .map(|entry| entry.item_count())
             .sum::<u64>();
 
+        let mut items_per_slot = HashMap::new();
+        for stream in reader.streams() {
+            *items_per_slot.entry(stream.slot_id).or_insert(0) += stream.row_count;
```
AaronRM: `items_per_slot` sums `stream.row_count`, but that's not the logical item count for OTLP pass-through (the default). `OtlpBytesAdapter` stores each request as one binary row, so `row_count` for slots 60/61/62 is the bundle count, not the record count, while the bundle's `item_count()` (`num_items()`) holds the real value. The per-signal counters therefore undercount and won't reconcile with dropped_items/expired_items. Prefer attributing each bundle's `item_count()` to its primary payload slot instead of summing raw row_counts. That fixes OTLP *and* Arrow metrics with one path and keeps quiver OTel-agnostic via the descriptor.

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs
```diff
@@ -2842,4 +2920,252 @@ mod tests {
...
+                    reported_utilization,
+                    used as f64 / cap as f64,
+                    "reported utilization should match used / cap"
+                );
+            })
+            .validate(|_| async {});
+    }
+
+    #[tokio::test]
+    async fn test_per_signal_dropped_expired_metrics() {
```
AaronRM, line 3054: This test only exercised the Logs path (and only the case where `item_count` == `row_count`, so it misses two bugs mentioned in other comments (metrics + OTLP Passthrough). Need to also cover the following scenarios:

- OTAP Metrics with >1 data point types (asserting that count == `num_items()`)
- OTAP Traces
- OTLP Passthrough Bundles
- Reconciliation assertions for dropped_log_records + dropped_spans + dropped_metric_datapoints == dropped_items   and expired_log_records + expired_spans + expired_metric_datapoints == expired_items

rust/otap-dataflow/crates/quiver/src/segment/stream_accumulator.rs
```diff
@@ -765,13 +765,13 @@ fn dict_values<'a>(col: &'a ArrayRef, key_type: &DataType) -> Result<&'a dyn Arr
...
 fn key_type_capacity(dt: &DataType) -> usize {
     match dt {
-        DataType::Int8 => i8::MAX as usize + 1,
-        DataType::Int16 => i16::MAX as usize + 1,
-        DataType::Int32 => i32::MAX as usize + 1,
-        DataType::Int64 => i64::MAX as usize + 1,
-        DataType::UInt8 => u8::MAX as usize + 1,
-        DataType::UInt16 => u16::MAX as usize + 1,
-        DataType::UInt32 => u32::MAX as usize + 1,
+        DataType::Int8 => (i8::MAX as usize).saturating_add(1),
```
AaronRM: Related to this PR? If not, let's revert and change as a separate issue and/or PR.

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs
```diff
@@ -304,6 +304,35 @@ pub struct DurableBufferMetrics {
     /// Data may still be recoverable via WAL replay on restart.
     #[metric(unit = "{error}")]
     pub flush_failures: Counter<u64>,
+
```
AaronRM, line 307: New metrics need to be added to the durable_buffer_processor's `telemetry.md`

AaronRM (dismissed):

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs
Dipanshusinghh: Hi @drewrelmas @AaronRM, Ihve addressed all the feedback. Here's a summary:

bundle_adapter.rs: removed hardcoded ARROW_METRICS_SLOT_1 / ARROW_METRICS_SLOT_2 constants entirely
segment_store.rs: reverted items_per_slot row_count logic completely
engine.rs / mod.rs: dynamic slot mapping via bundle_metadata + item_count() for both OTAP and OTLP passthrough.
stream_accumulator.rs: Unrelated Int8 change reverted
telemetry.md: All 7 new metrics documented
4 new tests added: Metrics, Spans, OTLP Passthrough, Reconciliation

Test results: core-nodes 753 passed ✅ | quiver 437 passed ✅ | cargo fmt clean ✅

image

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs
```diff
@@ -2842,4 +2939,785 @@ mod tests {
...
+            BundleDescriptor, PayloadRef, RecordBundle, SchemaFingerprint, SlotDescriptor, SlotId,
+        };
+
+        struct SimpleBundle {
+            descriptor: BundleDescriptor,
+            batch: RecordBatch,
+            fingerprint: SchemaFingerprint,
+            slot_id: SlotId,
+            item_count: u64,
+        }
```
drewrelmas: Thanks for adding tests for all 3 signal types! I note there is a lot of duplicated content for each test, is there a way to re-use things ( `use` statements, `SimpleBundle`, `impl RecordBundle`, test setup, etc) to improve readability? Theoretically the same test harness can be used 3 times with 3 slightly different inputs.

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/bundle_adapter.rs
```diff
@@ -113,6 +113,11 @@ pub(crate) fn signal_type_from_slot_id(slot: SlotId) -> Option<SignalType> {
     from_slot_id(slot).map(|(st, _)| st)
 }
 
+#[allow(dead_code)]
+pub(crate) const ARROW_LOGS_SLOT: SlotId = SlotId::new(30);
+#[allow(dead_code)]
+pub(crate) const ARROW_TRACES_SLOT: SlotId = SlotId::new(40);
+
```
drewrelmas: Please remove these as no longer needed

Dipanshusinghh: Thanks again 🙂4 this I will update all things!

Dipanshusinghh: Hi @drewrelmas, addressed your feedback-

Removed ARROW_LOGS_SLOT and ARROW_TRACES_SLOT constants from bundle_adapter.rs
Refactored duplicate test code into shared make_simple_bundle helper — use statements, SimpleBundle struct, and impl RecordBundle are now defined once and reused across all tests

All tests passing: 753 passed, 0 failed ✅..

drewrelmas: @Dipanshusinghh perhaps some updates to the tests were not included? I still see duplicate copies hanging around on the latest revision.

Dipanshusinghh: Hi @drewrelmas
, the duplicate SimpleBundle definitions have now been fully removed struct SimpleBundle, impl RecordBundle, and all use statements are defined once in the shared tests module and reused across all tests via make_simple_bundle and setup_test_processor helpers tests passing:- 753

Dipanshusinghh:
Ptal @drewrelmas  thnks

rust/otap-dataflow/crates/otap/tests/durable_buffer_processor_tests.rs
```diff
@@ -365,8 +365,23 @@ const IDX_QUEUED_LOG_RECORDS: usize = 27;
...
+#[allow(dead_code)]
+const IDX_DROPPED_SPANS: usize = 33;
+#[allow(dead_code)]
+const IDX_DROPPED_METRIC_DATAPOINTS: usize = 34;
+#[allow(dead_code)]
+const IDX_EXPIRED_LOG_RECORDS: usize = 35;
+#[allow(dead_code)]
+const IDX_EXPIRED_SPANS: usize = 36;
+#[allow(dead_code)]
+const IDX_EXPIRED_METRIC_DATAPOINTS: usize = 37;
```
drewrelmas: I believe these constants should also be removed, I see they're marked as `dead_code`

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -1342,12 +1390,21 @@ impl QuiverEngine {
...
                 bundles_dropped += bundles as u64;
                 items_dropped += items;
             }
+            if let Ok(metadata) = self.segment_store.bundle_metadata(*seq) {
+                for bundle in metadata {
+                    if let Some(&slot) = bundle.slot_ids.first() {
+                        *items_dropped_per_slot.entry(slot).or_insert(0) += bundle.item_count;
+                    }
+                }
+            }
```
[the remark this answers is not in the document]
drewrelmas: @Dipanshusinghh I think this point still stands, so I'll leave the thread unresolved.

 ```rust
 if let Some(&slot) = bundle.slot_ids.first() {
     *items_dropped_per_slot.entry(slot).or_insert(0) += bundle.item_count;
 }
 ```

 This attributes the bundle's `item_count` to `slot_ids.first()`. However:

 1. `slot_ids` is unordered (it comes from a `HashMap`), so `.first()` returns a non-deterministic slot for the bundle.
 2. Real OTAP bundles always include the shared slots `RESOURCE_ATTRS(1)` and `SCOPE_ATTRS(2)` alongside the per-signal slots.

 If `.first()` happens to return a shared slot, the processor classifies it as `None` and the bundle's items are silently dropped from every per-signal counter. For example, a logs bundle with 100 records occupies slots `[1, 2, 30, 31]`  (ResourceAttrs, ScopeAttrs, Logs, LogAttrs) with `item_count = 100`. The correct result is `dropped_log_records += 100`:

 ```text
 .first() == 30      → {30: 100}          → dropped_logs = 100   ✅ correct (lucky ordering)
 .first() == 1       → { 1: 100}          → None, items vanish   ❌ dropped_logs = 0
 add to each slot    → {30: 100, 31: 100} → 30→Logs + 31→Logs    ❌ dropped_logs = 200 (double)
 find_map per bundle → Some(Logs)         → dropped_logs += 100  ✅ correct, always
 ```

 A persisted bundle carries exactly **one** signal, spread across the shared slots (1/2) and one or more per-signal slots — so it should be classified and counted **once**, not per-slot. Iterating all slots over-counts (slots 30 and 31 both map to  Logs); `.first()` under-counts when it lands on a shared slot.

 Instead of `.first()`, consider having the engine expose the raw per-bundle data and classifying once in the processor:

 ```rust
 // instead of force_dropped_items_per_slot: RwLock<HashMap<SlotId, u64>>
 dropped_bundles_pending: RwLock<Vec<(Vec<SlotId>, u64)>>,   // (slot_ids, item_count)
 expired_bundles_pending: RwLock<Vec<(Vec<SlotId>, u64)>>,

 // in the loop, instead of .first()
 if let Ok(metadata) = self.segment_store.bundle_metadata(*seq) {
     let mut pending = self.dropped_bundles_pending.write();
     for bundle in metadata {
         pending.push((bundle.slot_ids, bundle.item_count));
     }
 }
 ```

 Then in the processor you can reuse the existing `find_map` pattern to classify each bundle by signal **once** and add its `item_count` to the matching per-signal total:

 https://github.com/open-telemetry/otel-arrow/blob/678b5d63f2cf952d2418827f552d6501c2d5d188/rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs#L823-L839

 This keeps the engine OTel-agnostic (it never interprets a `SlotId`). The alternative — stamping a canonical primary slot at write time and plumbing it through `BundleMetadata` — also works, but it touches the manifest format, so draining raw  `(slot_ids, item_count)` and classifying in the processor is the lighter change.

 It would also be good to add a test that ingests a real multi-slot bundle (with ResourceAttrs/ScopeAttrs present) and asserts the reconciliation equality — that's the case the current single-slot tests don't cover.

Dipanshusinghh:
@drewrelmas @AaronRM 
Replaced .first() with drain_dropped_bundles_pending / drain_expired_bundles_pending (Vec<(slot_ids, item_count)>)
Processor classifies each bundle once via find_map(signal_type_from_slot_id) — no over/under counting
Dead IDX_* constants removed from tests
Cumulative totals tracked in DurableBuffer struct fields
PTAL ! thanks.

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/telemetry.md
```diff
@@ -62,6 +62,35 @@ The effective instrument name is `otap.processor.durable_buffer.<field_name>`.
...
+
+### Per-Signal Expired Items (max_age policy)
+
+| Metric name | Unit | Instrument | Description | Produced in file |
+| --- | --- | --- | --- | --- |
+| `otap.processor.durable_buffer.expired_log_records` | `{log_record}` | ObserveCounter | Cumulative log records lost due to `max_age` segment expiry. Aggregated across all Arrow and OTLP log slots. | `crates/otap/src/durable_buffer_processor/mod.rs` |
+| `otap.processor.durable_buffer.expired_spans` | `{span}` | ObserveCounter | Cumulative spans lost due to `max_age` segment expiry. Aggregated across all Arrow and OTLP trace slots. | `crates/otap/src/durable_buffer_processor/mod.rs` |
+| `otap.processor.durable_buffer.expired_metric_datapoints` | `{data_point}` | ObserveCounter | Cumulative metric data points lost due to `max_age` segment expiry. Aggregated across all Arrow and OTLP metric slots. | `crates/otap/src/durable_buffer_processor/mod.rs` |
+
+
```
drewrelmas, line 93: nit: Remove extra newline
```suggestion
```

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/telemetry.md
```diff
@@ -62,6 +62,35 @@ The effective instrument name is `otap.processor.durable_buffer.<field_name>`.
 | `otap.processor.durable_buffer.expired_bundles` | `{bundle}` | ObserveCounter | Cumulative bundles lost due to `max_age` segment expiry. Non-zero values indicate data aged out before delivery. | `crates/otap/src/durable_buffer_processor/mod.rs` |
 | `otap.processor.durable_buffer.expired_items` | `{item}` | ObserveCounter | Cumulative individual items lost due to `max_age` segment expiry. Non-zero values indicate data aged out before delivery. | `crates/otap/src/durable_buffer_processor/mod.rs` |
 
+### Flush Failures
+
+| Metric name | Unit | Instrument | Description | Produced in file |
+| --- | --- | --- | --- | --- |
+| `otap.processor.durable_buffer.flush_failures` | `{error}` | Counter | Cumulative segment finalization (flush) failures. Non-zero values indicate data at risk — check logs for root cause. Data may still be recoverable via WAL replay on restart. | `crates/otap/src/durable_buffer_processor/mod.rs` |
```
drewrelmas: nit: Remove non-ascii char
```suggestion
| `otap.processor.durable_buffer.flush_failures` | `{error}` | Counter | Cumulative segment finalization (flush) failures. Non-zero values indicate data at risk - check logs for root cause. Data may still be recoverable via WAL replay on restart. | `crates/otap/src/durable_buffer_processor/mod.rs` |
```

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/telemetry.md
```diff
@@ -62,6 +62,35 @@ The effective instrument name is `otap.processor.durable_buffer.<field_name>`.
...
+
+### Per-Signal Dropped Items (DropOldest policy)
+
+| Metric name | Unit | Instrument | Description | Produced in file |
+| --- | --- | --- | --- | --- |
+| `otap.processor.durable_buffer.dropped_log_records` | `{log_record}` | ObserveCounter | Cumulative log records lost due to force-dropped segments (`DropOldest`). Aggregated across all Arrow and OTLP log slots. | `crates/otap/src/durable_buffer_processor/mod.rs` |
+| `otap.processor.durable_buffer.dropped_spans` | `{span}` | ObserveCounter | Cumulative spans lost due to force-dropped segments (`DropOldest`). Aggregated across all Arrow and OTLP trace slots. | `crates/otap/src/durable_buffer_processor/mod.rs` |
+| `otap.processor.durable_buffer.dropped_metric_datapoints` | `{data_point}` | ObserveCounter | Cumulative metric data points lost due to force-dropped segments (`DropOldest`). Aggregated across all Arrow and OTLP metric slots (slots 11–14 for Arrow, slot 62 for OTLP). | `crates/otap/src/durable_buffer_processor/mod.rs` |
```
drewrelmas: nit: Remove non-ascii char
```suggestion
| `otap.processor.durable_buffer.dropped_metric_datapoints` | `{data_point}` | ObserveCounter | Cumulative metric data points lost due to force-dropped segments (`DropOldest`). Aggregated across all Arrow and OTLP metric slots (slots 11-14 for Arrow, slot 62 for OTLP). | `crates/otap/src/durable_buffer_processor/mod.rs` |
```

rust/otap-dataflow/crates/otap/tests/durable_buffer_processor_tests.rs
```diff
@@ -365,8 +365,9 @@ const IDX_QUEUED_LOG_RECORDS: usize = 27;
 const IDX_QUEUED_METRIC_POINTS: usize = 28;
 const IDX_QUEUED_SPANS: usize = 29;
 const IDX_FLUSH_FAILURES: usize = 30;
+// ─── Appended metrics (Do not shift index above) ─────────────────────────────
```
drewrelmas: nit: Remove leftover line
```suggestion
```

rust/otap-dataflow/crates/quiver/src/engine.rs
```diff
@@ -108,10 +108,14 @@ pub struct QuiverEngine {
     force_dropped_bundles: AtomicU64,
     /// Count of individual data items lost due to force-dropped segments.
     force_dropped_items: AtomicU64,
+    /// Pending bundles from force-dropped segments, for per-signal tracking.
+    dropped_bundles_pending: RwLock<Vec<(Vec<crate::record_bundle::SlotId>, u64)>>,
```
drewrelmas: Thanks for the changes, I believe this is a solid improvement over the `.first()` tracking.

One additional thought upon re-review, it may be better to use a `HashMap` here instead of a `Vec`. In a high-throughput scenario, the `Vec` would grow unbounded even for bundles with the same `Vec<SlotId>`.

Something like the following would help keep new allocation to **per-shape** instead of **per-bundle**.
```rust
dropped_items_by_shape: RwLock<HashMap<SmallVec<[SlotId; N]>, u64>>,
```

A value like `N=8` could be a good starting point.

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs
```diff
@@ -1954,6 +2063,100 @@ pub static DURABLE_BUFFER_FACTORY: ProcessorFactory<OtapPdata> = ProcessorFactor
...
+        BundleDescriptor, PayloadRef, RecordBundle, SchemaFingerprint, SlotDescriptor, SlotId,
+    };
+    use std::sync::Arc;
+    use std::time::{Duration, SystemTime};
+
+    struct SimpleBundle {
+        descriptor: BundleDescriptor,
+        batch: RecordBatch,
+        fingerprint: SchemaFingerprint,
+        slot_id: SlotId,
```
drewrelmas: `SimpleBundle` still has only a single `slot_id` - could you please use something a bit more representative of the OTAP bundles we are working with? I would recommend something including shared slots, like `[1, 2, 30]` to prove that the code properly classifies based on the first non-shared slot.

drewrelmas (commented):
Have one suggestion for perf and an ask for additional test coverage on a more complex Bundle. After that, I think this is good to merge!

Dipanshusinghh:
@drewrelmas addressed all feedback:

HashMap<Vec<SlotId>, u64> grouped by slot shape (bounded memory)
SimpleBundle now uses [1, 2, primary_slot] shared slots
storage_utilization clamped with used.min(cap)
telemetry.md file paths corrected to crates/core-nodes/...
match + error = %e on bundle_metadata failures
PTAL, thanks!

drewrelmas (approved):
@Dipanshusinghh thank you for the contribution and the multiple iterations. I think this is in a good shape to merge, please resolve the remaining failing CI checks (fmt and markdownlint) and we can merge this!

Dipanshusinghh:
Hi @drewrelmas, I have resolved the remaining nightly fmt and markdownlint check failures. PTAL when you have a moment. Thanks!

Dipanshusinghh:
Thanks for the thorough review @drewrelmas and @AaronRM! Great working on this together. 
Also sent you both a LinkedIn request — would love to stay connected!🤝
https://www.linkedin.com/in/dipanshu-singh-6295ab28b/

PR #3274, opened 2026-06-11
Add optional `event_name` to parquet exporter logs schema

### LINKED ISSUE #3270: Parquet exporter logs schema is missing the optional `event_name` field defined in the OTAP schema
### Pre-filing checklist

### Component(s)

## Summary
The Parquet exporter maintains its own hand-written logs template schema, and it is missing the `event_name` field that already exists in the OTAP logs schema. As a result, `event_name` is silently dropped when logs are written to Parquet.

This is a schema-drift gap, not a new field proposal: `event_name` is already an **optional** column in the canonical OTAP logs schema. The Parquet exporter template just hasn't kept up with it. Sources that populate `event_name` (e.g. an ETW receiver, or any producer that sets it) lose the value at the Parquet boundary; sources that don't set it should simply produce `NULL`.

## `event_name` is already part of the OTAP schema

- **OTAP schema (Rust `pdata`)** — `required: false`:
  `rust/otap-dataflow/crates/pdata/src/schema/payloads.rs`
  ```rust
  Field {
      name: EVENT_NAME,
      data_type: DataType::Dictionary { min_key_size: DictKeySize::U8, value_type: Utf8 },
      required: false,
  },
  ```

- **Constant already defined**: `rust/otap-dataflow/crates/pdata/src/schema/consts.rs`
  ```rust
  pub const EVENT_NAME: &str = "event_name";
  ```

- **OTAP read-side views already expose it** as `Option<Str<'_>>` (`OtapLogsView` `LogRecordView` impl), so the value is available right up to the Parquet exporter boundary.

## The gap: Parquet exporter logs template

`rust/otap-dataflow/crates/core-nodes/src/exporters/parquet_exporter/schema.rs`

The `LOGS_TEMPLATE_SCHEMA` lists every logs column except `event_name`:

```rust
static LOGS_TEMPLATE_SCHEMA: LazyLock<Schema> = LazyLock::new(|| {
    Schema::new(vec![
        Field::new(consts::ID, DataType::UInt16, true),
        // resource, scope, schema_url, time_unix_nano, observed_time_unix_nano,
        // trace_id, span_id, severity_number, severity_text, body ...
        Field::new(consts::DROPPED_ATTRIBUTES_COUNT, DataType::UInt32, false),
        Field::new(consts::FLAGS, DataType::UInt32, false),
        // <-- no EVENT_NAME
    ])
});
```

The file itself notes that this template is **not** a comprehensive OTAP reference and intentionally omits metadata such as which fields are optional:

```rust
// note: these shouldn't be interpreted a comprehensive reference for OTAP generally. while the
// schemas below do contain every field and their logical datatypes, they lack information such
// as where dictionary encoding is used and other metadata such as which fields are optional.
```
## Proposed change

Add a single nullable field to `LOGS_TEMPLATE_SCHEMA`:

```rust
Field::new(consts::DROPPED_ATTRIBUTES_COUNT, DataType::UInt32, false),
Field::new(consts::EVENT_NAME, DataType::Utf8, true), // NULL when source omits it
Field::new(consts::FLAGS, DataType::UInt32, false),
```

Notes on the choice:
- **`Utf8`, not dictionary** — consistent with how the template treats other strings (e.g. `severity_text`). The OTAP schema marks it `Dictionary8<Utf8>`, but the transform preserves an incoming dictionary column if present (see `test_keeps_original_datatype_for_dicts`), and Parquet applies its own dictionary encoding at the storage layer.
- **Nullable** — gives exactly the desired "null when absent" behavior.
- **Column position is cosmetic** — fields are matched by name and reordered to the template; placing it next to `flags` mirrors the OTAP Arrow ordering.

## Why this is safe

The schema-transform logic already back-fills absent nullable template fields with an all-null column, and `get_all_null_column` already supports `Utf8`:

```rust
None => {
    let new_column = if template_field.is_nullable() {
        get_all_null_column(template_field.data_type(), num_rows)?
    } else {
        get_all_default_value_column(template_field.data_type(), num_rows)?
    };
    ...
}
```

So no new code paths are needed: present → preserved, absent → `NULL`.

## Considerations / open questions

- **Schema evolution for existing files.** Parquet files written before this change won't contain the `event_name` column. Cross-file reads (e.g. DataFusion) must treat the missing column as null. Worth confirming the reader/projection behavior is acceptable.
- **Larger follow-up (out of scope here):** the Parquet template is hand-maintained and duplicates the canonical OTAP schema, which is why this drift occurred. Deriving it from the canonical OTAP schema (Rust `pdata` `payloads.rs`) would prevent future omissions. Suggest tracking that separately.

### Rationale

Because the Parquet exporter is silently dropping data that the rest of the OTAP pipeline already carries.

### Scope

- Update Parquet Exporter

### Acceptance Criteria

- [ ] `LOGS_TEMPLATE_SCHEMA` includes a nullable `event_name` (`Utf8`) field, matching the OTAP schema's optional `event_name`.
- [ ] Logs that carry `event_name` are written to Parquet with the value preserved.
- [ ] Logs without `event_name` produce a `NULL` column (no error, no default empty string).
- [ ] A test covers both the populated and absent cases (mirroring existing schema-transform tests).

### Dependencies or Blockers

_No response_

### Additional Context

_No response_

swashtek:
# Change Summary
Add optional `event_name` to parquet exporter logs schema

* Closes #3270

- Local validation
- Unit tests

### Changelog

### DIFF
--- a/rust/otap-dataflow/.chloggen/parquet-exporter-logs-event-name.yaml
+++ b/rust/otap-dataflow/.chloggen/parquet-exporter-logs-event-name.yaml
@@ -1,0 +1,23 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: pipeline
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add the optional `event_name` field to the Parquet exporter logs schema so it is no longer silently dropped.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [3270]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  The `event_name` column already exists in the canonical OTAP logs schema but was
+  missing from the hand-written Parquet exporter logs template. Logs that carry
+  `event_name` are now written with the value preserved; logs without it produce a
+  NULL column.
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/parquet_exporter/schema.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/parquet_exporter/schema.rs
@@ -386,6 +386,11 @@ static LOGS_TEMPLATE_SCHEMA: LazyLock<Schema> = LazyLock::new(|| {
             true,
         ),
         Field::new(consts::DROPPED_ATTRIBUTES_COUNT, DataType::UInt32, false),
+        // NULL when the source omits it. The OTAP schema marks this as
+        // Dictionary(UInt8, Utf8), but the transform preserves an incoming dictionary column if
+        // present (see `test_keeps_original_datatype_for_dicts`), and the Parquet writer may
+        // dictionary-encode it at write time depending on its settings/heuristics.
+        Field::new(consts::EVENT_NAME, DataType::Utf8, true),
         Field::new(consts::FLAGS, DataType::UInt32, false),
     ])
 });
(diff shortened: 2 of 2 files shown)


cijothomas (approved):
Thanks! 
EventName in LogRecord is one of the *most debated* topics in OTel world! 
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-eventname

albertlockett (approved):

Issue #3068, opened 2026-05-22
[Controller] Expose in-process metrics handle from observer startup callback


sjmsft:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

Commit `32bc2ca5` (#2505) added `run_forever_with_observer` and `run_till_shutdown_with_observer` to `Controller`, which invoke a `FnOnce(ObservedStateHandle)` callback to give embedders in-process access to pipeline liveness and health state.

However, the callback only provides `ObservedStateHandle` - it does not expose the `TelemetryRegistryHandle` that the controller constructs internally during startup. Without access to this handle, hosting processes that embed the engine as a library cannot read internal metrics snapshots (e.g. per-node throughput, queue depths, processing latencies) without enabling the admin HTTP server and querying it over the network.

This is the metrics-access counterpart to the liveness/health access that `ObservedStateHandle` already provides.

#### Scope

- Extend the observer callback (or add a new callback variant) to also provide a metrics handle alongside `ObservedStateHandle`. This could be a combined struct, a second callback parameter, or a new `run_*_with_metrics_observer` method family.
- The exposed handle should allow callers to read the current metrics snapshot without requiring the admin HTTP server to be enabled.

#### Acceptance Criteria

- Embedding binaries can obtain an in-process metrics handle from the controller startup callback.
- Metrics snapshots can be read without enabling or depending on `http_admin`.
- Existing `run_forever` and `run_till_shutdown` behavior is unchanged.

#### Why the HTTP admin API is insufficient

Requiring embedding processes to query the admin HTTP API for metrics that already live in the same process introduces performance overhead.

This is the same reasoning behind the existing `ObservedStateHandle` for liveness/health: it avoids the HTTP path by returning snapshots directly at near-zero cost. The metrics handle should follow the same pattern.

### Additional Context

Commit `32bc2ca5` (#2505)

drewrelmas:
Accepted in principle, but also need to confirm how it relates to long-term internal telemetry pipeline plans. @lquerel to think on this a little more.

PR #3281, opened 2026-06-12
feat: add controller extension hooks

### LINKED ISSUE #3263: Add controller extension hooks for in-process Admin control-plane access
### Pre-filing checklist

### Component(s)

### Objective

Add a controller extension mechanism that allows in-process extensions to use the same semantic control-plane capabilities currently exposed through Admin endpoints.

### Rationale

The controller already exposes useful Admin operations for observing and managing the running engine, but those capabilities are currently oriented around external HTTP clients. This leaves a gap for embedded or in-process integrations that need to react to engine state, reconcile configuration, or trigger administrative operations without going through the HTTP Admin server.

A first-class controller extension hook would make these capabilities available to trusted runtime extensions through an internal API, while preserving the existing Admin endpoint behavior for external clients.

### Scope

- Add a generic engine-level extension configuration surface.
- Add controller startup/run options for registering extension factories.
- Provide an extension context with access to the semantic control-plane interface and observed-state handle.
- Ensure configured extensions are started and stopped with the controller lifecycle.
- Keep the existing Admin HTTP API and Admin Client SDK backward compatible.
- Avoid tying the extension mechanism to any specific protocol, transport, or vendor integration.

### Acceptance Criteria

- Engine configuration can declare controller-level extensions using a generic type plus extension-specific config payload.
- The controller can start with no extensions and preserve existing behavior.
- The controller fails startup with a clear error when configuration references an unknown extension type.
- A registered extension can access the same semantic operations used by Admin endpoints, including state inspection and live-control operations.
- Extension tasks are cancelled and joined during controller shutdown.
- Existing Admin endpoints and SDK behavior remain compatible.
- Tests cover config parsing, extension registration/startup failure, and lifecycle behavior.

### Dependencies or Blockers

_No response_

### Additional Context

_No response_

lquerel:
# Change Summary

Adds a controller extension hooks for trusted in-process integrations.

This PR introduces:
- an `engine.extensions` configuration section keyed by extension ID
- a reusable `ExtensionUserConfig` envelope for engine-level extension declarations
- `ControllerExtensionRegistry` and `ControllerRunOptions` for embedding applications to register extension factories
- `ControllerExtensionContext`, giving extensions access to:
    - their extension ID and config
    - the semantic control-plane handle
    - observed runtime state
    - the shared telemetry registry
    - the initial engine config
- extension lifecycle startup/shutdown handling on dedicated controller extension tasks

The default behavior remains unchanged: if no engine extensions are configured, no extension tasks are started. If an extension is configured without a registered factory, controller startup fails with a clear
error.

Closes https://github.com/open-telemetry/otel-arrow/issues/3263

Tested with:

- `cargo xtask check`

Yes. This adds a new public engine configuration field, `engine.extensions`, and new controller APIs for embedding applications that need to register trusted in-process controller extensions.

Existing configurations remain backward compatible because `engine.extensions` defaults to empty.

### Changelog


rust/otap-dataflow/.chloggen/controller-extension-hooks.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
+component: engine
+note: Add controller extension hooks and an opt-in read-only controller monitor extension.
+issues: []
```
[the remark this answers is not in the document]
lquerel: Fixed in https://github.com/open-telemetry/otel-arrow/pull/3281/commits/a96a5d9817887c7583f1da13a1c64b96de4c2084

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -1541,14 +1744,24 @@ impl<
         drop(metrics_reporter);
 
         let control_plane = runtime.control_plane();
+        let controller_extension_handles = Self::start_controller_extensions(
+            &engine_config,
+            &options,
+            Arc::clone(&control_plane),
```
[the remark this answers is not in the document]
lquerel: Fixed in https://github.com/open-telemetry/otel-arrow/pull/3281/commits/fe1f9a51cb481a8e1d44992d5c12fa10d20a53d1

gouslu:
One thing to mention here is that the plan I have with extensions is that they can be defined at any viable scope, such as engine, pipeline group etc. or possibly even for a specific node and not shared between nodes in the same pipeline. Today they are defined in pipeline only, but when we are able to define them at engine level, the configuration of it would use the same field engine.extensions. We need to be able to distinguish these two when we parse the config. In the back of my head I was thinking that maybe we could use the same extension mechanism but make these controller extensions a special type of privileged/secure extension that has Admin API access with necessary restrictions if needed.

rust/otap-dataflow/crates/controller/src/controller_monitor.rs
```diff
@@ -0,0 +1,396 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Built-in read-only controller monitor extension.
+//!
+//! This extension is a small reference implementation for controller-level
+//! extension hooks and a useful default monitor for the stock `df_engine`
+//! binary. It is enabled only when declared under `engine.extensions` with the
```
cijothomas: can you add an example configuration so users can run the df_engine with the config, and see the controller_monitor in action?

lquerel:
> One thing to mention here is that the plan I have with extensions is that they can be defined at any viable scope, such as engine, pipeline group etc. or possibly even for a specific node and not shared between nodes in the same pipeline. Today they are defined in pipeline only, but when we are able to define them at engine level, the configuration of it would use the same field engine.extensions. We need to be able to distinguish these two when we parse the config. In the back of my head I was thinking that maybe we could use the same extension mechanism but make these controller extensions a special type of privileged/secure extension that has Admin API access with necessary restrictions if needed.

@gouslu I can move the configuration for this type of extension to `engine.controller.extensions`. But it is important to understand that:
- Controller extensions are precisely the kind of extension that should not be accessible at the group, pipeline, or node level. These extensions exist to allow interaction with the controller itself; they are not globally shared extensions intended to provide access to global resources.
- These extensions must be isolated from other extensions because they interact with the control mechanism. Therefore, they cannot share the same event loop.
- The node binding mechanism used by pipeline-level extensions is not applicable, nor desirable, in this context.

EDIT: moved extensions in `engine.controller` in https://github.com/open-telemetry/otel-arrow/pull/3281/commits/579b7867fcc16ef226f77565d48f88fe79aae92e

rust/otap-dataflow/crates/controller/src/controller_monitor.rs
lquerel: It's already documented in `docs/configuration-model.md` but I can complete the comment of the impl.

lquerel: Fixed in https://github.com/open-telemetry/otel-arrow/pull/3281/commits/579b7867fcc16ef226f77565d48f88fe79aae92e

gouslu:
> > One thing to mention here is that the plan I have with extensions is that they can be defined at any viable scope, such as engine, pipeline group etc. or possibly even for a specific node and not shared between nodes in the same pipeline. Today they are defined in pipeline only, but when we are able to define them at engine level, the configuration of it would use the same field engine.extensions. We need to be able to distinguish these two when we parse the config. In the back of my head I was thinking that maybe we could use the same extension mechanism but make these controller extensions a special type of privileged/secure extension that has Admin API access with necessary restrictions if needed.
> 
> @gouslu I can move the configuration for this type of extension to `engine.controller.extensions`. But it is important to understand that:
> 
> * Controller extensions are precisely the kind of extension that should not be accessible at the group, pipeline, or node level. These extensions exist to allow interaction with the controller itself; they are not globally shared extensions intended to provide access to global resources.
> * These extensions must be isolated from other extensions because they interact with the control mechanism. Therefore, they cannot share the same event loop.
> * The node binding mechanism used by pipeline-level extensions is not applicable, nor desirable, in this context.
> 
> EDIT: moved extensions in `engine.controller` in [579b786](https://github.com/open-telemetry/otel-arrow/commit/579b7867fcc16ef226f77565d48f88fe79aae92e)

Thanks for the changes and the clarification. Quite excited about controller extensions!

gouslu (approved):

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -1384,6 +1592,20 @@ impl<
             engine_config.clone(),
         ));
 
+        let control_plane = runtime.control_plane();
+        // Extension factories are invoked before any pipeline thread is spawned
+        // so missing registrations or synchronous config errors fail startup
+        // without leaving already launched pipeline threads behind. The returned
```
cijothomas, line 1532: minor, better handled in a follow up: We don't catch invalid config of controller extensions in the `--validate-and-exit` mode, though we still fail in startup itself. 

not blocker for this PR.

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -1609,6 +1851,75 @@ impl<
...
     }
 
+    fn prepare_controller_extensions(
+        engine_config: &OtelDataflowSpec,
+        options: &ControllerRunOptions,
+        control_plane: Arc<dyn ControlPlane>,
+        observed_state: ObservedStateHandle,
+        telemetry_registry: TelemetryRegistryHandle,
+    ) -> Result<Vec<PreparedControllerExtension>, Error> {
+        let mut prepared_extensions = Vec::new();
```
cijothomas: really nit: start capacity for `engine_config.engine.controller.extensions.len()`

rust/otap-dataflow/crates/controller/src/error.rs
```diff
@@ -38,6 +38,35 @@ pub enum Error {
...
+    #[error("Controller extension `{extension_id}` failed to start: {source}")]
+    ControllerExtensionStartError {
+        /// Configured extension instance identifier.
+        extension_id: String,
+        /// Underlying extension error.
+        #[source]
+        source: Box<dyn std::error::Error + Send + Sync + 'static>,
+    },
+
+    /// A controller extension task returned an error.
```
cijothomas, line 60: what is the behavior if this happens? Would the main engine stops/exits?

rust/otap-dataflow/crates/controller/src/controller_monitor.rs
```diff
@@ -0,0 +1,412 @@
...
+
+struct ControllerMonitor {
+    extension_id: ExtensionId,
+    observed_state: ObservedStateHandle,
+    telemetry_registry: TelemetryRegistryHandle,
+    metrics: MetricSet<ControllerMonitorMetrics>,
+    config: ControllerMonitorConfig,
+}
+
+impl ControllerMonitor {
```
cijothomas, line 214: @sjmsft This requires a lot more code than your PR to get hold of just the telemetry, but this is also much more powerful and enables additional scenarios. I think this is acceptable, but please take a look.

rust/otap-dataflow/docs/configuration-model.md
```diff
@@ -343,6 +343,7 @@ Important behavior:
 - `telemetry`
 - `observed_state`
 - `observability`
+- `extensions`
```
cijothomas:
```suggestion
- `controller/extensions`
```
?

rust/otap-dataflow/crates/controller/src/lib.rs
lquerel: Fixed in https://github.com/open-telemetry/otel-arrow/pull/3281/commits/8227df960ac32d7c2d4168e1133296136522eb2d

rust/otap-dataflow/docs/configuration-model.md
```diff
@@ -386,6 +387,44 @@ Optional observability policies are supported at:
...
 
+### Controller Extensions
+
+Trusted in-process controller extensions can be declared at
+`engine.controller.extensions`. Each entry uses a standard extension envelope
+with a `type` URN and optional extension-specific `config`. These extensions
+run with controller-level handles and are not exposed to pipeline nodes as
+extension capabilities.
+
+Controller extension implementations are discovered from statically linked Rust
```
cijothomas, line 398: don't have better suggestion, but the name "extension" might not be fully conveying the intent, and properly distinguish between the pipeline extension.
Controller hook? Controller integration?

Not blocker.

cijothomas (approved):
LGTM.
Compared to https://github.com/open-telemetry/otel-arrow/issues/3068, this requires the embedding host to write a bit more code to get hold of the runtime handles, since it needs to provide/register a controller extension rather than just receive them from a startup callback. But given the additional capabilities that this enables, I agree with this direction.

I’m also planning to try an example integration using this mechanism to dynamically toggle parts of internal telemetry on/off.  Excited!

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -132,6 +146,153 @@ enum RunMode {
...
+        validate_config: ControllerExtensionValidateFn,
+    ) where
+        F: Fn(
+                ControllerExtensionContext,
+            ) -> Result<ControllerExtensionTaskFactory, ControllerExtensionError>
+            + Send
+            + Sync
+            + 'static,
+    {
+        _ = self.factories.insert(
```
utpilla: This would overwrite existing factory registrations on URN collision. Do we want to allow that? If yes, maybe we should log a warning here when registering a duplicate controller extension type?

rust/otap-dataflow/docs/configuration-model.md
lquerel: We intend to support several types of extension points: pipeline extensions, group extensions, engine extensions, and controller extensions (may be more). So it seems coherent/consistent to me to call this a controller extension.

rust/otap-dataflow/crates/controller/src/lib.rs
lquerel: Fixed in https://github.com/open-telemetry/otel-arrow/pull/3281/commits/9d8048db22183fb9faa32e66a7618a034014af5f

rust/otap-dataflow/docs/configuration-model.md
cijothomas: Makes sense. Resolving!

rust/otap-dataflow/crates/controller/src/lib.rs
```diff
@@ -1609,6 +1873,77 @@ impl<
...
+    fn prepare_controller_extensions(
+        engine_config: &OtelDataflowSpec,
+        options: &ControllerRunOptions,
+        control_plane: Arc<dyn ControlPlane>,
+        observed_state: ObservedStateHandle,
+        telemetry_registry: TelemetryRegistryHandle,
+    ) -> Result<Vec<PreparedControllerExtension>, Error> {
+        let mut prepared_extensions =
+            Vec::with_capacity(engine_config.engine.controller.extensions.len());
+        for (extension_id, extension) in engine_config.engine.controller.extensions.iter() {
```
utpilla: Should we enforce deterministic ordering (for example by extension id), or explicitly document order as undefined, to prevent implicit startup-order assumptions from turning into fragile dependencies across runs.

sjmsft:
> LGTM. Compared to #3068, this requires the embedding host to write a bit more code to get hold of the runtime handles, since it needs to provide/register a controller extension rather than just receive them from a startup callback. But given the additional capabilities that this enables, I agree with this direction.
> 
> I’m also planning to try an example integration using this mechanism to dynamically toggle parts of internal telemetry on/off. Excited!

So the recommendation is for the hosting app to use run_forever to start the engine and then use the controller extension to obtain telemetry_registry and handles? Can we remove run_forever_with_observer as part of this PR then?

rust/otap-dataflow/crates/controller/src/error.rs
lquerel: I updated the behavior to make controller extension runtime failures fail fast at the engine level. If a controller extension task returns an error, the controller logs it, requests a global pipeline shutdown, unparks the main controller loop, completes normal teardown, and exits with ControllerExtensionRuntimeError.
The rationale is that controller extensions run outside the pipeline graph and may own control-plane responsibilities, so an unexpected runtime failure can leave the engine in an inconsistent or unmanaged state. Recoverable/transient failures should be handled inside the extension itself, for example with retry/backoff.

lquerel:
> > LGTM. Compared to #3068, this requires the embedding host to write a bit more code to get hold of the runtime handles, since it needs to provide/register a controller extension rather than just receive them from a startup callback. But given the additional capabilities that this enables, I agree with this direction.
> > I’m also planning to try an example integration using this mechanism to dynamically toggle parts of internal telemetry on/off. Excited!
> 
> So the recommendation is for the hosting app to use run_forever to start the engine and then use the controller extension to obtain telemetry_registry and handles? Can we remove run_forever_with_observer as part of this PR then?

@sjmsft done in https://github.com/open-telemetry/otel-arrow/pull/3281/commits/e9e8427080d1acd6098e34589f0e8417d21216d6

rust/otap-dataflow/crates/controller/src/lib.rs
lquerel: Good point. I kept the existing "latest registration wins" behavior, since the registry can be used as an assembly/override point, but it no longer happens silently. Duplicate controller extension type registration now emits a structured warning with the extension type URN, and I added a regression test to make the behavior explicit.
 
 https://github.com/open-telemetry/otel-arrow/pull/3281/commits/0e5d0f2c97df714900c2dfe6931e1ff75f29581a

rust/otap-dataflow/crates/controller/src/lib.rs
lquerel: I made the startup order deterministic by sorting configured controller extensions by extension ID before invoking their factories. 

https://github.com/open-telemetry/otel-arrow/pull/3281/commits/af8cac2cff5f28e9aa73598faf5f6f54a7d5a111

utpilla (approved):

PR #3285, opened 2026-06-12
engine: pass ExtensionContext to ExtensionFactory.create

gouslu:
# Change Summary

Restores extension factories' access to telemetry-registration APIs by passing `&ExtensionContext` as the first argument to `ExtensionFactory.create`.

After `PipelineContext` was decoupled from extensions, factories had no way to register their own custom metric sets — `MetricSet<T>` has no public constructor, and `EffectHandler` (reached only inside `start()`) can only `report` a pre-built set, not create one. This change re-opens that path with a narrower, extension-scoped context.

* `cargo check -p otap-df-engine --tests` — clean
* `cargo test -p otap-df-engine --test extension_e2e` → 28/28 pass
* `cargo test -p otap-df-engine --lib test_extension_factory` → 2/2 pass
* No new tests added; existing e2e factories exercise the new signature.

Yes — `ExtensionFactory.create` signature changed. All in-tree factories and tests have been updated.

### Changelog

* [x] Added a `.chloggen/*.yaml` entry (`breaking`, component `engine`).

### DIFF
--- a/rust/otap-dataflow/.chloggen/engine-extension-context-factory.yaml
+++ b/rust/otap-dataflow/.chloggen/engine-extension-context-factory.yaml
@@ -1,0 +1,24 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: engine
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Pass `&ExtensionContext` to `ExtensionFactory.create` so extensions can register custom metric sets and extension entities at construction time.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [3285]
+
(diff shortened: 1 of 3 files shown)


drewrelmas (approved):

rust/otap-dataflow/.chloggen/engine-extension-context-factory.yaml
```diff
@@ -0,0 +1,24 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: breaking
```
utpilla, line 4: We don't have to call this "breaking" even though technically it is changing the public API. But we know that there aren't any users whose experience would be broken by this change. The previous PR wasn't merged that long ago either.

If no release has been cut yet, there is no API contract in place. "enhancement" would better capture what this actually is: extending the `ExtensionFactory.create` signature to expose `&ExtensionContext` at construction time.

drewrelmas: Spoke with @gouslu offline, it would also work to simply modify the Changelog of the last PR to treat it together as one 'item'.

drewrelmas: Spoke with @utpilla as well, we have some work to do to properly classify what `breaking` means and to encourage usage of single tracking `chloggen` file for related work across multiple PRs.

PR #3287, opened 2026-06-12
feat(flow_metrics): Add kept/dropped flow metrics recorded per keep/drop decision node

drewrelmas:
# Change Summary

Adds two customer-facing flow metrics — `signals.kept` and `signals.dropped` — so operators can see, per flow, how many signal items each in-pipeline decision node keeps versus drops.

Unlike `signals.incoming` / `signals.outgoing` (recorded at the start/end of a flow range), kept/dropped are recorded by individual **decision-node processors** inside the range. A processor opts in via a capability flag; the engine registers a per-decision-node entity and the processor reports its counts each cycle. A new `flow.node.decision` scope attribute carries the deciding node's name, so a flow may contain multiple decision nodes without conflating their counts.

This is PR1 of the accept/reject telemetry effort (#2859). kept/dropped is scoped to **true decision-point processors only** — processors that can keep or drop whole signal items by a predicate (filter, sampler, and transform/KQL `where`), where `kept + dropped = incoming`. Processors that only route, fan out, buffer, or aggregate are intentionally excluded, since relocating, duplicating, or collapsing items is not a keep/drop decision. Receiver/exporter directional endpoints are a separate follow-up.

## Motivation

`incoming`/`outgoing` only show the net effect across a whole flow range. With multiple processors that each keep/drop data, operators couldn't attribute the loss to a specific node. `flow.node.decision` makes each decision node's contribution observable while the whole-flow aggregate remains available by summing across the attribute.

## What changed

- **Config**: added `SignalsKept` / `SignalsDropped` to the `FlowMetric` enum (`signals_kept` / `signals_dropped`, enabled by default when `metrics` is omitted).
- **Engine**: new `signals.kept` / `signals.dropped` metric sets and a `flow.node.decision` attribute; processors opt in via `ProcessorRuntimeRequirements::with_keep_drop_decisions()` and report through `EffectHandler::record_flow_signals_kept` / `record_flow_signals_dropped`. Supports single-node flow ranges and rejects flows that share an interior decision node.
- **Processors**: `filter_processor` and `log_sampling_processor` wired as the first consumers.
- **Demo config**: `fake-flow-metrics-demo.yaml` shows two interior decision nodes in one flow (`log_sampling` + `filter`) using the synthetic data source.
- **Changelog**: `.chloggen/flow-kept-dropped-metrics.yaml`.

## Design notes

- Capability is a simple bool on `ProcessorRuntimeRequirements`, not the extension capability/DI system.
- kept + dropped share one entity per decision node, keyed by `flow.node.decision`; the empty value belongs to the flow's incoming/outgoing/duration entity.
- No counting cost when disabled: `record_flow_signals_kept` / `_dropped` are cheap no-ops unless the node is a decision node in an enabled flow. The two wired consumers reuse counts they already compute for their own metrics, so this PR adds no extra per-batch work. Future consumers that need *new* counting (e.g. transform summing `num_items` across routed ports) should guard it with `EffectHandler::is_flow_decision()`.
- Aggregating across `flow.node.decision` only makes sense for `signals.dropped` (drops at different nodes are disjoint, summing to `incoming − outgoing`). `signals.kept` must **not** be summed across nodes — series nodes keep nested subsets, so the sum double-counts; the flow-wide kept count is just `signals.outgoing`, and per-node `signals.kept` is only meaningful on its own entity.
- Flow metrics remain aggregate (no `signal.type` dimension) in this PR. Per-signal breakdown is a tracked follow-up.

## Validation

Captured from `/api/v1/telemetry/metrics` with the demo config (flow range `attr1..attr4`, `sampler` and `filter` interior). The `sampler` keeps ~1/3 of log records (ratio 1-in-3); the `filter` then drops records whose `thread.name` matches its exclude rule. Each entity emits only some metrics; blank = not emitted by that entity:

| pipeline position (`flow.node.decision`) | `signals.incoming` | `signals.kept` | `signals.dropped` | `signals.outgoing` | `compute.duration` |
|---|---|---|---|---|---|
| flow start `""` | 340 | | | | |
| `sampler` | | 114 | 226 | | |
| `filter` | | 68 | 46 | | |
| flow end `""` | | | | 68 | 142,892,257 ns |

(The flow start/end rows are the same flow entity — `flow.node.decision = ""` — shown at both ends of the range for pipeline order. `compute.duration` is accumulated across the range and reported on that entity.)

Aggregation across `flow.node.decision`:

- **`dropped` sums** — drops are disjoint across nodes: `226 + 46 = 272 = incoming − outgoing (340 − 68)`.
- **`kept` does not sum** — series nodes keep nested subsets (the sampler produces 114 while the filter produces 68), so a sum double-counts the same items multiple times. Flow-wide kept is just `outgoing = 68`, not `114 + 68 = 182`.

* Progress towards #2859 

Unit tests, run of demo config

Yes, operators can now get kept/dropped metrics from `flow`s.

### Changelog

### DIFF
--- a/rust/otap-dataflow/.chloggen/flow-kept-dropped-metrics.yaml
+++ b/rust/otap-dataflow/.chloggen/flow-kept-dropped-metrics.yaml
@@ -1,0 +1,25 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: engine
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add `signals.kept` / `signals.dropped` flow metrics recorded per keep/drop decision node
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [2859]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Flow metrics declared in the telemetry policy can now enable `signals_kept` and
+  `signals_dropped`. A processor that declares the keep/drop capability and lies
+  within such a flow records its kept/dropped counts, attributed per decision node
+  via the new `flow.node.decision` scope attribute (so a flow may contain multiple
+  decision nodes). `filter_processor` and `log_sampling_processor` are the first
+  consumers. Single-node flow ranges (`start == end`) are now supported.
--- a/rust/otap-dataflow/configs/fake-flow-metrics-demo.yaml
+++ b/rust/otap-dataflow/configs/fake-flow-metrics-demo.yaml
@@ -1,9 +1,10 @@
-# Demonstrates the distributed flow metrics feature.
+# Demonstrates the distributed flow metrics feature, including two keep/drop
+# decision nodes within a single flow range.
 #
-# Five attribute processors form a sequential processing chain, with a 1-in-3
-# log sampler in the middle so that produced < consumed:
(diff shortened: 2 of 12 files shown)


drewrelmas:
Closing, see https://github.com/open-telemetry/otel-arrow/issues/2859#issuecomment-4710114835

PR #3289, opened 2026-06-12
kafka receiver/exporter benchmark scenarios

c1ly:
# Change Summary

Added benchmark scenarios using the kafka receiver + exporter comparing DFE, Go, and Rotel

benchmark tests

no


lquerel:
@JakeDern Could you take a look. Thanks.

tools/comparison_dashboard/suites/dfe/dfe-logs-kafkaotlp-none-baseline.yaml
```diff
@@ -0,0 +1,22 @@
+name: DFE Kafka OTLP Baseline (Logs)
+slug: dfe_logs_kafkaotlp_none_baseline
+description: Dataflow Engine Kafka pipeline with OTLP protobuf encoding for logs, no compression
+orchestrator_template: ../pipeline_perf_test/test_suites/comparison_dashboard/templates/orchestrator/dfe-kafka-single-core-multi-rate.yaml
+
+meta:
+  binary: dfe
+  protocols: [kafkaotlp]
```
JakeDern: It's not necessary to specify protocol for everything, I think we could just leave this empty for kafka

tools/comparison_dashboard/suites/dfe/dfe-logs-kafka-otap-none-baseline.yaml
```diff
@@ -0,0 +1,22 @@
+name: DFE Kafka OTAP Baseline (Logs)
```
JakeDern, line 1: Just curious, how do we store OTAP in kafka?

IIRC OTLP is pretty straightforward and we just serialize the whole OTLP proto-object because they're self-contained.

For OTAP do we convert to OTLP or some other representation?

tools/pipeline_perf_test/test_suites/comparison_dashboard/templates/engine/kafka-consumer-otap.yaml.j2
```diff
@@ -0,0 +1,54 @@
...
+        nodes:
+          receiver:
+            type: urn:otel:receiver:kafka
+            config:
+              brokers: "{{kafka_hostname}}:9092"
+              group_id: "dfe-kafka-consumer"
+              client_id: "dfe-kafka-consumer"
+              auto_offset_reset: earliest
+{%- set signal = signal | default("logs") %}
+{%- if signal == "logs" %}
```
JakeDern: Might consider removing the conditionals for each signal throughout - I assume if the load generator is only sending one signal then it's fine to map all three on both the receiver or exporter end and keeps the templates simpler

tools/pipeline_perf_test/test_suites/comparison_dashboard/reports/report_traces.yaml
```diff
@@ -349,7 +349,7 @@ queries:
         CROSS JOIN observation_window ow
         LEFT JOIN component_core_counts ccc ON cr.component_name = ccc.component_name
         WHERE cr.metric_name IN ('container.cpu.usage', 'process.cpu.usage')
-          AND cr.component_name IN ('df-engine', 'go-collector', 'rotel')
+          AND cr.component_name IN ('df-engine', 'go-collector', 'rotel', 'kafka-producer', 'kafka-consumer')
```
JakeDern, line 352: What's the experimental setup? Is it loadgen -> kafka-producer -> kafka -> kafka-consumer -> backend? Are there two SUTs?

tools/comparison_dashboard/suites/dfe/dfe-logs-kafka-otap-none-baseline.yaml
c1ly: for otap we store the batch arrow record bytes

tools/pipeline_perf_test/test_suites/comparison_dashboard/reports/report_traces.yaml
c1ly: that is correct, we have the kafka broker sitting between the two systems under test

JakeDern: Is the intent to isolate kafka receiver performance from kafka exporter performance? Otherwise one other alternative would be to change the loadgen to export to kafka and then have the SUT read/write to kafka.

Something like:

loadgen (Kafka exporter) -> Kafka -> SUT (Kafka receiver/exporter) -> Kafka -> backend (Kafka receiver)

tools/comparison_dashboard/suites/dfe/dfe-logs-kafka-otap-none-baseline.yaml
JakeDern: How do we manage the reconstruction of IPC Stream State if a consumer restarts?

If the watermark from the consumer points to the middle of some IPC stream, it won't have enough state if it resumes from that message. It needs to replay everything from the start of the IPC stream until the current watermark.

Do we store a separate watermark somewhere for the start of each IPC stream and then resume from there?

PR #3290, opened 2026-06-13
Add integration tests for the ETW receiver

utpilla:
# Change Summary

Adds the first Windows end-to-end integration test for the ETW receiver and wires it into CI as a required check.

The test drives the receiver from a `TestRuntime`, registers real ETW providers, emits TraceLogging events through both Windows producer crates in a single ETW session, and asserts on the decoded `OtapArrowRecords`.

## Coverage

One test, two producers, two events:

- **`tracelogging_dynamic`** (builder API): integers, `f64`, boolean, UTF-8/UTF-16 strings, `FILETIME`, `GUID`, and a nested struct.
- **`tracelogging`** (static macro API): `u32`, `str8`, native `bool8`, and a nested struct.

Both events are validated through shared helpers that pin the log envelope (timestamps, severity, body, `etw.*` attributes) and each user attribute together with its OTAP type discriminator. Nested struct fields appear as flat `Parent.Child` keys in `LogAttrs`.

## CI

New top-level job `etw_windows_smoke` in `.github/workflows/rust-ci.yml`:

- runs on `windows-latest` (GitHub-hosted runners are Administrator, which ETW kernel sessions need),
- runs on every PR / push to `main` / merge_group,
- listed in `rust-required-status-check.needs`,
- mirrors the existing `user_events_linux_smoke` job.

The test is `#[ignore]`; non-elevated local runs skip it.

Towards #2783 

From an elevated PowerShell:

```pwsh
cargo test -p otap-df-contrib-nodes --features etw-receiver `
    etw_receiver_decodes_tracelogging_events_end_to_end `
    -- --ignored --nocapture
```

Without elevation the test fails fast with an explicit message.

No

### Changelog


rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/windows_e2e_tests.rs
```diff
@@ -0,0 +1,1214 @@
...
+    };
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    for line in stdout.lines() {
+        // `logman query -ets` formats each data-collector row as
+        // `<name><whitespace>Trace<whitespace>Running`; take the
+        // leading whitespace-separated token.
+        let Some(name) = line.split_whitespace().next() else {
+            continue;
+        };
+        if !name.starts_with("OtapEtwE2E-") {
```
lalitb, line 1246: This stops all `OtapEtwE2E-*` sessions, not just the session for this test run. If two CI jobs or two local runs share the same Windows host, one run can stop the other run’s active ETW session; can we only clean up the current session name, or otherwise prove the session is stale before stopping it?

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/windows_e2e_tests.rs
```diff
@@ -0,0 +1,1214 @@
...
+        .find_map(|r| find_event_row(r, event_name).map(|row| (r, row)))
+        .unwrap_or_else(|| {
+            panic!(
+                "did not receive an Arrow batch containing event '{event_name}' \
+                 from the ETW receiver within 15s"
+            )
+        })
+}
+
+fn find_event_row(records: &OtapArrowRecords, expected: &str) -> Option<usize> {
```
lalitb: This only matches by event name, but E2eStaticEvent is the same for every run. If two CI runs or two local runs share the same Windows host, this test could pick the other run’s static event and fail even though its own event was received; can we also match on the current process id?

lalitb (approved):
LGTM overall. The ETW smoke coverage looks useful, and the remaining concerns I had are around shared-runner robustness rather than blockers for the current GitHub-hosted Windows CI path, where each job gets its own fresh runner VM.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/windows_e2e_tests.rs
utpilla: Fixed in https://github.com/open-telemetry/otel-arrow/pull/3290/commits/deaacdf8b9c9cd14b244e3d04aa04b874776b416

PR #2646, opened 2026-04-13
Experimental clickhouse exporter

clhain:
# Change Summary

This PR includes a partial POC / Experimental implementation of a clickhouse exporter that is **not suitable for integration** with the dataflow engine in its current state.

It's being made available to accelerate an actual implementation being discussed with clickhouse engineering, but isn't wired in (via feature flags or otherwise) to the engine today.

Please let me know if there is a better spot to put this (either here, or I could just leave it in a public personal repo).

CC: @lquerel, @cijothomas, @jmacd 

## How are these changes tested?

Poorly!

## Are there any user-facing changes?

No, this is currently inactive code.

PR #3291, opened 2026-06-13
Updated Experimental ClickHouse Exporter

SpencerTorres:
extended from @clhain's PR (https://github.com/open-telemetry/otel-arrow/pull/2646)

# Change Summary

This draft PR includes an experimental implementation of a ClickHouse exporter that is an iteration on the previous PR (https://github.com/open-telemetry/otel-arrow/pull/2646). Notably:
- Removed JSON support (It's not recommended at scale yet for OTel data)
- Uses the official ClickHouse Rust client with Arrow extension
- Uses HTTP instead of TCP

Unit tests and an E2E test, but may need better coverage

Not yet


rust/otap-dataflow/configs/trafficgen-clickhouse.yaml
cijothomas, line 1: we renamed "fake" to trafficgen in https://github.com/open-telemetry/otel-arrow/pull/3288 Please adopt it here too.

rust/otap-dataflow/configs/fake-clickhouse.yaml
```diff
@@ -0,0 +1,34 @@
...
+                signals_per_second: 30
+                log_weight: 100
+              registry_path: https://github.com/open-telemetry/semantic-conventions.git[model]
+          exporter:
+            type: "exporter:clickhouse"
+            config:
+              endpoint: "http://127.0.0.1:8123"
+              database: "otap"
+              username: "default"
+              password: "test"
```
cijothomas: is this common to configure password in file ? (We do support reading it from env: , so maybe that is better to show even in example)

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/metrics.rs
```diff
@@ -0,0 +1,94 @@
...
+use otap_df_pdata::proto::opentelemetry::arrow::v1::ArrowPayloadType;
+use otap_df_telemetry::instrument::Counter;
+use otap_df_telemetry_macros::metric_set;
+
+/// Clickhouse exporter metrics.
+/// Grouped under `otap.exporter.clickhouse`.
+#[metric_set(name = "otap.exporter.clickhouse")]
+#[derive(Debug, Default, Clone)]
+pub struct ClickhouseExporterMetrics {
+    /// Total number of log rows written into clickhouse
```
cijothomas:
```suggestion
    /// Total number of log rows written successfully into clickhouse
```

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/arrays.rs
clhain, line 1: For expedience I copied this more or less verbatim from crates/pdata/src/arrays.rs in the original exploration. We should probably figure out a better way to handle this (used mostly for the various *Accessors in transform_batch/attributes). I think @albertlockett had some ideas when we talked about this maybe 4 months ago, but presume a lot may have changed since then.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/README.md
```diff
@@ -0,0 +1,232 @@
...
+```bash
+docker run -it -p 8123:8123 -p 9000:9000 -e CLICKHOUSE_PASSWORD=TODO-TEST \
+  --name clickhouse-server --ulimit nofile=262144:262144 \
+  clickhouse/clickhouse-server
+```
+
+### 2. Run the data plane with the ClickHouse exporter
+
+```bash
+cargo run --features clickhouse-exporter -- --config configs/fake-clickhouse.yaml
```
cijothomas:
```suggestion
cargo run --features clickhouse-exporter -- --config configs/trafficgen-clickhouse.yaml
```

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/config.rs
```diff
@@ -0,0 +1,374 @@
...
+#[derive(Debug, Clone, Default)]
+pub struct Config {
+    /// ClickHouse HTTP(S) endpoint URL (e.g. "http://localhost:8123"). TCP is not supported for now.
+    pub endpoint: String,
+    /// Database to write to (e.g., "otap")
+    pub database: String,
+    /// Clickhouse user name
+    pub username: String,
+    /// Clickhouse password
+    pub password: String,
```
cijothomas: use `SecretString`? (I think the support for it was introduced recently in a PR..)

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/README.md
```diff
@@ -0,0 +1,232 @@
...
+### 1. Start ClickHouse
+
+```bash
+docker run -it -p 8123:8123 -p 9000:9000 -e CLICKHOUSE_PASSWORD=TODO-TEST \
+  --name clickhouse-server --ulimit nofile=262144:262144 \
+  clickhouse/clickhouse-server
+```
+
+### 2. Run the data plane with the ClickHouse exporter
+
```
cijothomas, line 50: nit: mention the directory from which to run this.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/table_snapshots/log_table_map_attrs.snap
```diff
@@ -0,0 +1,34 @@
...
+expression: sql
+---
+CREATE TABLE IF NOT EXISTS otap.otel_logs (
+    `Timestamp` DateTime64(9) CODEC(Delta(8), ZSTD(1)),
+    `TraceId` String CODEC(ZSTD(1)),
+    `SpanId` String CODEC(ZSTD(1)),
+    `TraceFlags` UInt8,
+    `SeverityText` LowCardinality(String) CODEC(ZSTD(1)),
+    `SeverityNumber` UInt8,
+    `ServiceName` LowCardinality(String) CODEC(ZSTD(1)),
```
cijothomas, line 12: curious - how it was determined which fields gets promoted to a dedicated column? ServiceName is from service.name attribute. Is that because ServiceName is considered most important detail from Resource?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/table_snapshots/log_table_map_attrs.snap
```diff
@@ -0,0 +1,34 @@
...
+    `LogAttributes` Map(LowCardinality(String), String) CODEC(ZSTD(1)),
+    `EventName` String CODEC(ZSTD(1)),
+    INDEX idx_trace_id TraceId TYPE bloom_filter(0.001) GRANULARITY 1,
+    INDEX idx_res_attr_key mapKeys(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_res_attr_value mapValues(ResourceAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_scope_attr_key mapKeys(ScopeAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_scope_attr_value mapValues(ScopeAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_log_attr_key mapKeys(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_log_attr_value mapValues(LogAttributes) TYPE bloom_filter(0.01) GRANULARITY 1,
+    INDEX idx_lower_body lower(Body) TYPE tokenbf_v1(32768, 3, 0) GRANULARITY 8
```
cijothomas, line 29: not fully familiar with the implications on the index for body; but I wonder why not do the same for EventName too, which is a low cardinality identifier for the LogRecord.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/README.md
```diff
@@ -0,0 +1,232 @@
...
+| `List<Map<Utf8,Utf8>>` | `Array(Map(LowCardinality(String), String))` | Events.Attributes, Links.Attributes (one map per event/link) |
+
+No special `input_format_arrow_*` settings were required for a clean insert.
+
+## Attribute Representation
+
+Inline attributes are stored as:
+
+```sql
+Map(LowCardinality(String), String)
```
cijothomas, line 166: we are converting all value types for attributes into String intentionally? (Attribute values could be int/double/bool etc.)

rust/otap-dataflow/configs/trafficgen-clickhouse.yaml
```diff
@@ -0,0 +1,35 @@
...
+            pdata: 128
+
+        nodes:
+          receiver:
+            type: receiver:traffic_generator
+            config:
+              traffic_config:
+                max_batch_size: 1000
+                signals_per_second: 30
+                log_weight: 100
```
cijothomas: given we support logs and traces, lets give some weight to traces too, else the table would have nothing to show.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/table_snapshots/log_table_map_attrs.snap
clhain: With the original effort we wanted to make sure we could interoperate with the go clickhouse exporter's table structure, and this is what it does. I assume the original rationale was something like what you mentioned. I think it'd be an awesome feature one day to let people specify sets of attributes they want to promote like this.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/table_snapshots/log_table_map_attrs.snap
clhain: Same deal, just following the go collector structure. IMO we should continue to prioritize interop and any optimizations are opt-in

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/README.md
clhain: go collector has this pattern as the default, plus experimental support for json column type which preserves some of these types. After we talked with Clickhouse they recommended starting with this for performance reasons.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+
+                    let write_batches = match batch_transformer.apply_plan(arrow_records) {
+                        Ok(batches) => batches,
+                        Err(e) => {
+                            self.pdata_metrics.inc_failed(signal_type);
+                            otap_df_telemetry::otel_warn!(
+                                "clickhouse.exporter.transform.error",
+                                message = "Error transforming batch for export.",
+                                error = e.to_string(),
+                                signal_type = format!("{:?}", signal_type),
```
swashtek, line 265: This error arm logs and bumps `inc_failed` and then falls through so batch is dropped. 
Can we add retry with backoff for retriable/recoverable errors from Clickhouse?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/config.rs
```diff
@@ -0,0 +1,376 @@
...
+    pub password: SecretString,
+    /// Use async insert
+    pub async_insert: bool,
+    pub table_defaults: DefaultTableConfig,
+    pub tables: TablesConfig,
+}
+
+impl Config {
+    pub fn from_patch(p: ConfigPatch) -> Self {
+        let async_insert = p.async_insert.unwrap_or(true);
```
swashtek, line 68: `async_insert` defaults to `true` and in `build_client` we are setting `async_insert=1` without wait. 
It returns success once the row is buffered, not when it's persisted. 
Async insert without waiting is durability tradeoff that's made silently, and exporter is counting `inc_exported` which is misleading. 

Either we can default `async_insert` to `false` or when async insert is enabled, also set wait, so end reflects durability. 

Whichever option we pick, worth documenting the durability trade-off. Wdyt?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/writer.rs
```diff
@@ -0,0 +1,321 @@
...
+/// target database without concern.
+pub fn build_client(config: &Config, database: &str) -> Client {
+    let mut client = Client::default()
+        .with_url(config.endpoint.clone())
+        .with_database(database)
+        .with_user(config.username.clone())
+        .with_password(config.password.expose_secret());
+    if config.async_insert {
+        client = client.with_setting("async_insert", "1");
+    }
```
swashtek, line 133: This suggestion is with respect to the [review comment](https://github.com/open-telemetry/otel-arrow/pull/3291/changes#r3556950721) if `async_insert` is defaulted to `true`.

```suggestion
    if config.async_insert {
        client = client
            .with_setting("async_insert", "1")
            .with_setting("wait_for_async_insert", "1");
    }
```

jmacd (commented):
Thanks for pushing this forward — this is a strategically important component and I'm supportive of landing it. This is a well-structured, feature-gated, honestly-documented experimental exporter, and CI is fully green. Everything below is **non-blocking**: notes for follow-ups, not merge conditions.

## What's great
- Clean module boundaries (config → transform plan → batch transform → writer) with good module-level docs and a candid README that documents its own gaps.
- Real SQL-injection defense on all interpolated DDL (`validate_identifier`, `validate_ttl`, `validate_engine_params`, `reject_injection`).
- Secrets handled well: `SecretString` + `${env:VAR}` substitution.
- Solid unit coverage (config, arrays, metrics, transform ops, DDL snapshots) plus live-ClickHouse `e2e_*` roundtrip tests.
- Now on official published crates (`clickhouse` + `clickhouse-ext-arrow`); interop-first schema aligned to the Go collector.

## Non-blocking notes (what's left unfinished)

**Signal coverage**
- Metrics are stubbed. Logs + traces only; the metrics config (gauge/sum/summary/histogram/exp-histogram) is fully modeled but inert. `build_table_sql` returns `Some("SELECT 1;")` for metric kinds, so startup issues ~3 no-op queries via `iter_tables` and creates no metrics tables. `summary` and `exponential_histogram` are also silently dropped by `MetricsTableConfig::iter()`.

**Robustness / panics**
- Two `unimplemented!()` sit on live paths. `transform_plan.rs:223` is startup-only (fires if `SUPPORTED_ARROW_PAYLOAD_TYPES` and the match drift). `transform_batch.rs:263` (`append_list_value`) is data-dependent on the hot path — an unexpected event/link element type would panic the exporter instead of returning the `CoercionError` used everywhere else. Cheap to turn into a recoverable error.

**Delivery semantics**
- Write failures drop data (logged + `inc_failed`) with no retry/backpressure/NACK. It's also inconsistent: `decode_transport_optimized_ids` failure is fatal (`?` tears down the exporter), while convert/transform/write failures are per-message `continue`.
- No cross-message batching — one HTTP `INSERT` per payload-type per message; `async_insert=true` by default, so "exported" is optimistic w.r.t. durability.
- Partial-write non-atomicity within a message (logs can succeed, then spans fail); no idempotency, so a future retry layer could double-write.

**Testing**
- The `e2e_*` roundtrip tests are `#[ignore]` (need a live ClickHouse), so they don't run in CI, and there's no automated schema-drift check vs the Go collector (README acknowledges both).

**Papercuts**
- TTL format trap: doc-comments/examples use `"72h"`/`"12h"` (e.g. `config.rs` `TableConfig.ttl`), but `validate_ttl` requires a full unit keyword (`"72 HOUR"`), so the natural guess fails at table creation.
- Attributes are stringified into `Map(String, String)` (lossy, original OTAP not reconstructable) — documented and intentional.
- `arrays.rs` is copied ~verbatim from the `pdata` crate (already flagged in-thread) — worth de-duplicating later.
- Latent narrowing cast `Dictionary(UInt32) → UInt16` for `parent_id` (fine in practice; OTAP parent_id is 16-bit by design).
- Trivia: cosmetic `Drop {}` vs `Drop`; an `anync` typo in a comment.

## Three cheap, optional follow-ups
1. Turn the two `unimplemented!()` into `CoercionError`/`UnsupportedPayload` so bad data can't panic the exporter.
2. Fix the `"72h"` → `"72 HOUR"` TTL examples in the doc-comments.
3. Skip metric kinds in `iter_tables` (or drop the `SELECT 1;` stub) so startup doesn't issue no-op DDL.

None of these block merge. Happy to see this go in and iterate.

jmacd (approved):
Approving — experimental, feature-gated, CI green, and honestly documented. My separate comment lists non-blocking follow-ups (metrics stub, the two `unimplemented!()` on live paths, drop semantics, TTL doc format, no CI e2e). None block merge; let's land it and iterate.

swashtek (commented):
Thank you for bringing this exporter. I look forward to trying this out. I have few comments for my understanding of the exporter functionality.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/mod.rs
clhain: Ack/Nack integration should probably come in a followup (that's the mechanism for retries rather than exporter specific implementations as far as I understand). We could note this as planned work in the readme.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/config.rs
clhain: Given the thread per core model, I'm not sure waiting on the clickhouse side persist is something that makes sense (at least not without a good deal of thought/documentation around what conditions it'll be useful for that won't result in unexpected backpressure). IMO this should come as followup work in hardening phases along with the ack/nack implementation.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/writer.rs
clhain: For clarity since I didn't really explain in the last message - I'm worried that blocking until clickhouse commits will cause any instance where '>1 batch on the engine side is being generated before a CH write completes' will put the engine in a state where it chokes to death / can't catch up.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/mod.rs
SpencerTorres: I agree we can note this in a follow up

rust/otap-dataflow/crates/contrib-nodes/src/exporters/clickhouse_exporter/writer.rs
SpencerTorres: In the Go exporter these are controllable via the exporter settings, and can be overridden in the user settings on the server side. We can do a follow up that gives more granular control to the exporter.

We typically use async + wait, but some configs take a "fire and forget" approach

clhain (approved):
LGTM!

lquerel (approved):
LGTM

swashtek (approved):

cijothomas (approved):
Awesome work. After merge, will incorporate to nightly benchmarks.

PR #3299, opened 2026-06-15
feat(engine): opt-in extension readiness probe

gouslu:
# Change Summary

Adds an opt-in extension readiness probe so extensions can hold pipeline-node
startup until they're ready (open connection, primed cache, fetched first
token). Per-variant, parallel, fail-fast. Extensions that don't opt in pay
zero cost.

```rust
ExtensionWrapper::builder(name, user_cfg, runtime_cfg)
    .active()                          // or .background()
    .with_readiness_probe()            // default 5s
    .shared(my_ext).build()?;

// From start():
eh.signal_ready();
```

Semantics:
- Per-variant probe pairs (shared / local independent).
- Parallel wait — three 1s exts total ~1s, not 3s; total wait bounded by the
  longest timeout, never the sum.
- Per-extension timeouts; gate trips on whichever expires first (fail-fast).
- The gate also watches extension task completions: an extension that exits or
  fails during the readiness window aborts startup instead of slipping through.
- Drop-detect fast-fail: dropping the `EffectHandler` without firing surfaces
  `ExtensionReadinessSignallerDropped` immediately.
- `signal_ready()` is idempotent across clones, no-op when not opted in.
- The default 5s timeout can be overridden (shorter or longer) with
  `with_readiness_probe_timeout_override(t)`; a zero timeout is rejected at
  build time.

499 unit + 35 e2e, all green.

- Unit: primitive (fire/wait/drop/idempotency), builder wiring + timeout
  override contract (incl. zero rejection), `EffectHandler::signal_ready`
  threading through `wrapper::start()`, and `wait_all_ready` across variants —
  including deterministic (paused-clock) fail-fast and max-wait-bounded-by-
  longest-timeout guarantees, plus task-failure-during-gate.
- E2E (7 tests): signals-in-time, named-laggard timeout, parallel wait across
  3 exts, per-extension timeouts, extension-failure-before-ready fast-fail,
  background variant gating, background timeout.

Yes — new opt-in builder API. Fully backwards compatible (extensions that
don't call `.with_readiness_probe(...)` get unchanged behavior).

### Changelog

* [x] Added `.chloggen/extension-readiness-probe.yaml`.


rust/otap-dataflow/crates/engine/src/extension/readiness.rs
```diff
@@ -0,0 +1,286 @@
...
+pub const DEFAULT_READINESS_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Send-side handle for a variant's readiness signal. Cloneable;
+/// only the first [`ReadinessSignaller::ready`] call delivers.
+pub struct ReadinessSignaller {
+    inner: Arc<ReadinessInner>,
+}
+
+struct ReadinessInner {
+    fired: AtomicBool,
```
utpilla: The `fired: AtomicBool` is redundant with `tx_slot: Mutex<Option<Sender>>`. `None` already means "fired." `Mutex::take()` is itself an atomic test-and-set, so only one caller can ever get `Some(tx)` and send. We can drop the atomic and the related `compare_exchange` operations entirely:

```rust
pub fn ready(&self) {
    if let Some(tx) = self.inner.tx_slot.lock().expect("poisoned").take() {
        let _ = tx.send(());
    }
}
fn is_ready(&self) -> bool {
    self.inner.tx_slot.lock().expect("poisoned").is_none()
}
```

rust/otap-dataflow/crates/engine/src/extension/mod.rs
```diff
@@ -6,26 +6,32 @@
...
-    LocalDecomposed, PassiveClonedCompleteStage, PassiveClonedLocalStage, PassiveClonedSharedStage,
-    PassiveClonedStage, PassiveConstructedCompleteStage, PassiveConstructedLocalStage,
-    PassiveConstructedSharedStage, PassiveConstructedStage, PassiveStage, SharedDecomposed,
+    ActiveCompleteStage, ActiveLocalStage, ActiveSharedStage, ActiveStage, BackgroundCompleteStage,
+    BackgroundEmptyStage, ExtensionBundleBuilder, LocalDecomposed, PassiveClonedCompleteStage,
+    PassiveClonedLocalStage, PassiveClonedSharedStage, PassiveClonedStage,
+    PassiveConstructedCompleteStage, PassiveConstructedLocalStage, PassiveConstructedSharedStage,
+    PassiveConstructedStage, PassiveStage, SharedDecomposed,
+};
+pub use readiness::{
```
utpilla, line 40: `ReadinessSignaller`, `ReadinessProbe`, and `ReadinessProbeError` aren't used anywhere outside the engine crate. Every consumer is in-crate (wrapper.rs, extension_lifecycle.rs, in-crate tests).

The external surface an extension author needs is just `with_readiness_probe()` / `with_extended_readiness_probe_timeout()` and `EffectHandler::signal_ready()`. Suggest dropping these three from the public re-export (keep them `pub(crate)`) to avoid committing to a public API we don't need.

rust/otap-dataflow/crates/engine/src/extension/builder.rs
```diff
@@ -123,52 +125,98 @@ pub struct LocalDecomposed {
...
+
+    /// Opt in with a timeout `>= DEFAULT_READINESS_TIMEOUT`.
+    ///
+    /// # Panics
+    ///
+    /// Panics when `timeout < DEFAULT_READINESS_TIMEOUT`.
+    #[must_use]
+    pub fn with_extended_readiness_probe_timeout(mut self, timeout: Duration) -> Self {
+        let default = super::readiness::DEFAULT_READINESS_TIMEOUT;
+        assert!(
```
utpilla: This would panic when `timeout < DEFAULT_READINESS_TIMEOUT`. The static `validate_config` pass runs before `create`, but the standard `validate_typed_config::<Config>` helper is structural-only. A sub-default `readiness_timeout_ms` deserializes fine abut would panic here. 

I suggest changing the signature to return `Result` and replacing the assert! with an early `Err` (`InvalidUserConfig)`:

```rust
pub fn with_extended_readiness_probe_timeout(
    mut self,
    timeout: Duration,
) -> Result<Self, Error> {
    let default = super::readiness::DEFAULT_READINESS_TIMEOUT;
    if timeout < default {
        return Err(Error::InvalidUserConfig {
            error: format!(
                "readiness timeout {timeout:?} is below the minimum of {default:?} \
                 (extension: {}); use with_readiness_probe() for the default",
                self.parent.name.as_ref(),
            ),
        });
    }
    if timeout > default {
        otel_info!(
            "extension.readiness.timeout_override",
            extension = self.parent.name.as_ref(),
            timeout_ms = u64::try_from(timeout.as_millis()).unwrap_or(u64::MAX),
            default_ms = u64::try_from(default.as_millis()).unwrap_or(u64::MAX),
        );
    }
    self.readiness_timeout = Some(timeout);
    Ok(self)
}
```

rust/otap-dataflow/crates/engine/src/extension/builder.rs
```diff
@@ -536,42 +594,77 @@ impl PassiveConstructedCompleteStage {
...
+    pub fn with_readiness_probe(mut self) -> Self {
+        self.readiness_timeout = Some(super::readiness::DEFAULT_READINESS_TIMEOUT);
+        self
+    }
+
+    /// See [`ActiveStage::with_extended_readiness_probe_timeout`].
+    #[must_use]
+    pub fn with_extended_readiness_probe_timeout(mut self, timeout: Duration) -> Self {
+        let default = super::readiness::DEFAULT_READINESS_TIMEOUT;
+        assert!(
```
utpilla: Same as https://github.com/open-telemetry/otel-arrow/pull/3299#discussion_r3439217005

rust/otap-dataflow/crates/engine/src/extension/builder.rs
```diff
@@ -705,6 +806,26 @@ impl ExtensionBundleBuilder {
         });
     }
 
+    fn set_shared_readiness(&mut self, probe: ReadinessProbe, signaller: ReadinessSignaller) {
+        assert!(
```
utpilla: Let's use `debug_assert!` for this kind of check.

```suggestion
        debug_assert!(
```

rust/otap-dataflow/crates/engine/src/extension/builder.rs
```diff
@@ -705,6 +806,26 @@ impl ExtensionBundleBuilder {
...
+            self.shared_probe.is_none() && self.shared_signaller.is_none(),
+            "readiness probe already registered for the shared variant of extension '{}'",
+            self.name.as_ref()
+        );
+        self.shared_probe = Some(probe);
+        self.shared_signaller = Some(signaller);
+    }
+
+    fn set_local_readiness(&mut self, probe: ReadinessProbe, signaller: ReadinessSignaller) {
+        assert!(
```
utpilla:
```suggestion
        debug_assert!(
```

rust/otap-dataflow/crates/engine/src/extension/mod.rs
gouslu: Good point, but can't be done in this PR. Let me work on visibility on another PR, that might impact multiple types and files.

rust/otap-dataflow/crates/engine/src/extension_lifecycle.rs
```diff
@@ -265,6 +273,47 @@ impl ExtensionLifecycle {
         }
     }
 
+    /// Await every opted-in readiness probe in parallel, with each
+    /// probe's own timeout. Returns on the first failure with a
+    /// named-laggard error
+    /// ([`Error::ExtensionReadinessSignallerDropped`] or
+    /// [`Error::ExtensionReadinessTimeout`]).
+    pub async fn wait_all_ready(&mut self) -> Result<(), Error> {
```
lalitb, line 286: Could `wait_all_ready()` also watch extension task completions while it waits for  readiness? Right now it only polls the readiness probes. If an extension exits or fails during this window, the readiness gate can still return `Ok(())` and let data-path nodes start even though startup already lost an extension. It seems this should use the same task failure handling as `wait_all_spawned()` while readiness is pending.

rust/otap-dataflow/crates/engine/tests/extension_e2e.rs
```diff
@@ -3979,3 +3979,742 @@ connections:
...
+
+#[test]
+fn test_extension_readiness_gate_times_out_with_named_extension() {
+    let receiver_key = "ready-timeout-recv";
+    let ext_key = "ready-timeout-ext";
+
+    let _recv_probe = make_probe(receiver_key, CallSequence::Local);
+    let _ext_probe = make_ready_gate_probe(ext_key, None, false);
+
+    let timeout_ms: u64 = 5_000;
```
lalitb: These timeout e2e tests seem to wait for the full 5s readiness timeout, and there are multiple cases like this. That adds noticeable time to the test suite just to prove timeout behavior. Could we keep the real default behavior covered in unit tests and make the e2e timeout path shorter or less repeated?

rust/otap-dataflow/crates/engine/src/extension/builder.rs
gouslu: I made it so that error is surfaced at "build()", otherwise it wouldn't work well with the fluent pattern. No panic.

rust/otap-dataflow/crates/engine/src/extension/builder.rs
```diff
@@ -536,42 +584,68 @@ impl PassiveConstructedCompleteStage {
...
+    pub fn with_readiness_probe(mut self) -> Self {
+        self.readiness_timeout = Some(super::readiness::DEFAULT_READINESS_TIMEOUT);
+        self
+    }
+
+    /// See [`ActiveStage::with_readiness_probe_timeout_override`].
+    #[must_use]
+    pub fn with_readiness_probe_timeout_override(mut self, timeout: Duration) -> Self {
+        let default = super::readiness::DEFAULT_READINESS_TIMEOUT;
+        if timeout > default {
```
utpilla: This should be if `timeout != default` to match `ActiveStage` (line 157). Now that sub-default overrides are valid, a below-default timeout here is applied silently while the same on an active extension logs `extension.readiness.timeout_override`. 

Since the two bodies are otherwise identical, consider a shared helper so they can't drift again.

gouslu: good catch!

utpilla (approved):

rust/otap-dataflow/crates/engine/src/extension/builder.rs
```diff
@@ -123,52 +140,80 @@ pub struct LocalDecomposed {
...
     parent: ExtensionBundleBuilder,
+    readiness_timeout: Option<Duration>,
 }
 
 impl ActiveStage {
+    /// Opt in with the default timeout
+    /// ([`DEFAULT_READINESS_TIMEOUT`](super::readiness::DEFAULT_READINESS_TIMEOUT), 5 s).
+    /// Must precede [`shared()`](Self::shared) / [`local()`](Self::local).
+    #[must_use]
+    pub fn with_readiness_probe(mut self) -> Self {
```
lalitb, line 166: Small design question: this opt-in applies to the whole extension bundle, so if an extension registers both local and shared variants, both variants need to call `signal_ready()`. That may be intended, but it is easy to miss from the API. Could we document that contract, or consider a way to opt in per variant later?

lalitb (approved):
LGTM  - small design/ergonomics question, nothing blocking, Thanks.

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
```diff
@@ -340,6 +325,14 @@ impl<PData: 'static + Debug + Clone + ReceivedAtNode + Unwindable + FlowMetricHo
             return Err(barrier_err);
         }
 
+        if let Err(readiness_err) =
+            rt.block_on(local_tasks.run_until(extension_lifecycle.wait_all_ready()))
```
lalitb, line 336: Small question: since the readiness timeout can be made longer, should this wait also react to pipeline shutdown? Right now it looks like startup may keep waiting for readiness until the timeout even if the operator stops the pipeline.

rust/otap-dataflow/crates/engine/tests/extension_e2e.rs
```diff
@@ -3979,3 +3979,743 @@ connections:
...
+        ctx,
+        entity_key,
+        ts,
+        ready_after + Duration::from_millis(250),
+    );
+    let elapsed = started_at.elapsed();
+
+    assert!(result.is_ok(), "{result:?}");
+
+    let parallel_upper = ready_after + Duration::from_millis(250) + Duration::from_secs(1);
```
ThomsonTan, line 4453: Nit: this hard-coded duration could still trigger flaky failure on stressed environment.

rust/otap-dataflow/crates/engine/src/runtime_pipeline.rs
gouslu: will try to take care of it in another PR.

rust/otap-dataflow/crates/engine/tests/extension_e2e.rs
gouslu: 1 second seems plenty overhead, so let's leave it like that. I can increase it later if it seems necessary, but 1 second is a lot of time, don't you think?

rust/otap-dataflow/crates/engine/src/extension/builder.rs
gouslu: I documented it, but the builder language signals it. You opt in for readiness signal before you provide local and shared variants. I intended to do it this way to make it simple. If one variant doesn't require to provide signal, they may emit it immediately.

PR #3303, opened 2026-06-16
Add static headers config to OTAP OTLP exporters

timr-dev:
## Description

The Rust OTAP OTLP exporters (HTTP + gRPC) could not set arbitrary static request headers. The only configurable header was `user_agent`, and the client-settings structs are `#[serde(deny_unknown_fields)]`, so a `headers:` key was rejected at config load. gRPC metadata was propagation-only. This blocks any OTLP backend that requires a static request header on ingest (e.g. `Authorization: Basic <b64>`, tenant-routing or tracing-vendor headers).

This PR adds a `headers` map to both exporters' client settings, bringing parity with the Go Collector's `otlphttp`/`otlp` `headers:` option.

## Changes

- **`HttpClientSettings` / `GrpcClientSettings`**: new `headers: IndexMap<String, String>` field, `#[serde(default)]` (empty by default; `deny_unknown_fields` preserved). Validated at config load — valid header name/value; HTTP rejects protocol-reserved names (`content-type` / `content-encoding` / `content-length` / `host`); gRPC validates ASCII metadata keys/values.
- **HTTP exporter**: configured headers applied once at client-pool build, inserted before `Content-Type` / `Accept` so protocol headers always win.
- **gRPC exporter**: the static metadata template is pre-built once at startup; the per-message hot path keeps its zero-allocation fast path (no clone/alloc when there are no static headers and nothing to propagate). Static headers and propagated headers coexist.
- Docs: README "Static request headers" section, `configs/otlp-http-headers.yaml` example, and a changelog entry.

## Backwards compatibility

Additive and fully backwards-compatible: the field defaults to empty, existing configs are unaffected, `deny_unknown_fields` is preserved (the field is now known), and there is no behavior change unless `headers` is set.

## Performance

The gRPC per-message hot path remains zero-allocation when no static headers are configured. The static metadata template is parsed/validated once at config load and built once at startup, then cloned per request only when present. HTTP headers are parsed once and applied per pooled client at startup — never per request.

## Testing

- HTTP: on-the-wire tests asserting configured headers arrive, and that protocol headers win over configured ones.
- gRPC: unit tests for static-metadata construction plus an end-to-end test that drives the real exporter with `headers: { authorization: "Bearer ..." }` through a tonic server interceptor and asserts the token reaches the wire.
- `validate()` rejection tests (invalid name, invalid value, reserved name).
- Full `otap` + `core-nodes` suites pass; `cargo fmt --check` and `cargo clippy --all-targets -D warnings` are clean.

## Example

```yaml
exporters:
  - type: exporter:otlp_http
    config:
      endpoint: https://example.app/api/default
      http:
        headers:
          authorization: "Basic <base64(user:password)>"
          stream-name: "default"
```

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
@@ -13,6 +13,7 @@ use async_trait::async_trait;
 use bytes::Bytes;
 use futures::future::FutureExt;
 use futures::stream::{FuturesUnordered, StreamExt};
+use indexmap::IndexMap;
 use linkme::distributed_slice;
 use otap_df_config::SignalType;
 use otap_df_config::node::NodeUserConfig;
@@ -203,6 +204,11 @@ impl Exporter<OtapPdata> for OTLPExporter {
         let compression = self.config.grpc.compression_encoding();
         let max_in_flight = self.config.max_in_flight.max(1);
 
+        // Pre-build the static gRPC metadata template ONCE, outside the hot loop.
+        // Returns `None` when no static headers are configured, which preserves
+        // the zero-allocation fast path in `build_grpc_metadata`.
+        let static_metadata = build_static_grpc_metadata(&self.config.grpc.headers);
+
         // reuse the encoder and the buffer across pdatas
         let mut logs_proto_encoder = LogsProtoBytesEncoder::new();
         let mut metrics_proto_encoder = MetricsProtoBytesEncoder::new();
@@ -312,9 +318,12 @@ impl Exporter<OtapPdata> for OTLPExporter {
                     let (context, payload) = pdata.into_parts();
                     self.pdata_metrics.inc_consumed(signal_type);
 
-                    // Build gRPC metadata from transport headers if a propagation
-                    // policy is configured. Computed once before signal dispatch.
-                    let metadata = build_grpc_metadata(&effect_handler, &context);
+                    // Build gRPC metadata from configured static headers and
+                    // any propagated transport headers. Computed once before
+                    // signal dispatch; the static template is cloned only when
+                    // present so the no-metadata case stays allocation-free.
+                    let metadata =
+                        build_grpc_metadata(&effect_handler, &context, static_metadata.as_ref());
 
                     // Dispatch based on signal type and the concrete payload representation.
                     match (signal_type, payload) {
@@ -630,63 +639,117 @@ async fn finalize_completed_export(
     client
 }
 
-/// Builds a [`MetadataMap`] from the transport headers attached to a pdata
-/// message, filtered through the exporter's propagation policy.
+/// Pre-builds the static gRPC metadata template from configured headers.
 ///
-/// Returns `None` when either the policy or the transport headers are absent,
-/// or when all headers are dropped by the policy (zero overhead in the common
-/// case).
+/// Built once at exporter startup (outside the per-message hot path). Returns
+/// `None` when no static headers are configured so `build_grpc_metadata` keeps
+/// its zero-allocation fast path. Header names/values are validated at config
+/// load time; any that somehow fail to parse here are skipped with a debug log.
+fn build_static_grpc_metadata(headers: &IndexMap<String, String>) -> Option<MetadataMap> {
+    if headers.is_empty() {
+        return None;
+    }
+
+    let mut metadata = MetadataMap::with_capacity(headers.len());
+    for (name, value) in headers {
+        let Ok(key) = name.parse::<MetadataKey<tonic::metadata::Ascii>>() else {
+            otel_debug!(
+                "otlp.exporter.grpc.header_skip",
+                reason = "invalid ascii metadata key",
+                header_name = name.as_str()
+            );
+            continue;
+        };
+        let Ok(val) = MetadataValue::try_from(value.as_str()) else {
+            otel_debug!(
+                "otlp.exporter.grpc.header_skip",
+                reason = "invalid ascii metadata value",
+                header_name = name.as_str()
+            );
+            continue;
+        };
+        let _ = metadata.insert(key, val);
+    }
+
+    if metadata.is_empty() {
+        None
+    } else {
+        Some(metadata)
+    }
+}
+
+/// Builds the per-request gRPC metadata by merging the pre-built static
+/// `static_metadata` template with any headers propagated from the incoming
+/// transport context.
+///
+/// Hot path: when there is neither static metadata nor a propagation source
+/// this returns `None` without allocating. The static template is cloned only
+/// when present (each tonic request needs its own owned metadata); propagated
+/// headers are appended on top so static and propagated headers coexist.
 fn build_grpc_metadata(
     effect_handler: &EffectHandler<OtapPdata>,
     context: &Context,
+    static_metadata: Option<&MetadataMap>,
 ) -> Option<MetadataMap> {
-    let policy = effect_handler.propagation_policy()?;
-    let transport_headers = context.transport_headers()?;
+    let propagation = effect_handler
+        .propagation_policy()
+        .zip(context.transport_headers());
 
-    let mut metadata = MetadataMap::new();
-    for header in policy.propagate(transport_headers) {
-        match header.value_kind {
-            ValueKind::Text => {
-                // ASCII metadata: parse the header name and value.
-                let Ok(key) = header
-                    .header_name
-                    .parse::<MetadataKey<tonic::metadata::Ascii>>()
-                else {
-                    otel_debug!(
-                        "otlp.exporter.grpc.header_skip",
-                        reason = "invalid ascii metadata key",
-                        header_name = header.header_name
-                    );
-                    continue;
-                };
-                let Ok(value) = MetadataValue::try_from(header.value) else {
-                    otel_debug!(
-                        "otlp.exporter.grpc.header_skip",
-                        reason = "invalid ascii metadata value",
-                        header_name = header.header_name
-                    );
-                    continue;
-                };
-                let _ = metadata.append(key, value);
-            }
-            ValueKind::Binary => {
-                // Binary metadata: gRPC binary metadata keys must end with `-bin`.
-                // Metadata map will error if attempting to insert key without `-bin`.
-                let key_name = if header.header_name.ends_with("-bin") {
-                    header.header_name.to_string()
-                } else {
-                    format!("{}-bin", header.header_name)
-                };
-                let Ok(key) = key_name.parse::<MetadataKey<tonic::metadata::Binary>>() else {
-                    otel_debug!(
-                        "otlp.exporter.grpc.header_skip",
-                        reason = "invalid binary metadata key",
-                        header_name = header.header_name
-                    );
-                    continue;
-                };
-                let value = MetadataValue::from_bytes(header.value);
-                let _ = metadata.append_bin(key, value);
+    // Zero-alloc fast path: nothing static configured and nothing to propagate.
+    if static_metadata.is_none() && propagation.is_none() {
+        return None;
+    }
+
+    let mut metadata = match static_metadata {
+        Some(static_metadata) => static_metadata.clone(),
+        None => MetadataMap::new(),
+    };
+
+    if let Some((policy, transport_headers)) = propagation {
+        for header in policy.propagate(transport_headers) {
+            match header.value_kind {
+                ValueKind::Text => {
+                    // ASCII metadata: parse the header name and value.
+                    let Ok(key) = header
+                        .header_name
+                        .parse::<MetadataKey<tonic::metadata::Ascii>>()
+                    else {
+                        otel_debug!(
+                            "otlp.exporter.grpc.header_skip",
+                            reason = "invalid ascii metadata key",
+                            header_name = header.header_name
+                        );
+                        continue;
+                    };
+                    let Ok(value) = MetadataValue::try_from(header.value) else {
+                        otel_debug!(
+                            "otlp.exporter.grpc.header_skip",
+                            reason = "invalid ascii metadata value",
+                            header_name = header.header_name
+                        );
+                        continue;
+                    };
+                    let _ = metadata.append(key, value);
+                }
+                ValueKind::Binary => {
+                    // Binary metadata: gRPC binary metadata keys must end with `-bin`.
+                    // Metadata map will error if attempting to insert key without `-bin`.
+                    let key_name = if header.header_name.ends_with("-bin") {
+                        header.header_name.to_string()
+                    } else {
+                        format!("{}-bin", header.header_name)
+                    };
+                    let Ok(key) = key_name.parse::<MetadataKey<tonic::metadata::Binary>>() else {
+                        otel_debug!(
+                            "otlp.exporter.grpc.header_skip",
+                            reason = "invalid binary metadata key",
+                            header_name = header.header_name
+                        );
+                        continue;
+                    };
+                    let value = MetadataValue::from_bytes(header.value);
+                    let _ = metadata.append_bin(key, value);
+                }
             }
         }
     }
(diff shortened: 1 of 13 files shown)


rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/mod.rs
```diff
@@ -1010,6 +1022,176 @@ mod test {
...
+        _ = headers.insert("content-type".to_string(), "text/plain".to_string());
+        let settings = HttpClientSettings {
+            headers,
+            ..Default::default()
+        };
+
+        tokio_rt.block_on(async {
+            let mut pool = HttpClientPool::try_new(&settings, NonZeroUsize::new(1).unwrap())
+                .await
+                .expect("failed to build client pool");
```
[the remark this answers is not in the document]
timr-dev: Removed this test in 2d8391901 — good catch. It configured a reserved `content-type` header, which `HttpClientSettings::validate()` rejects, so that scenario can't arise through normal config loading and the test was exercising an unreachable path. The two behaviours it was meant to guard are still covered:

- protocol-header precedence (Content-Type/Accept win) is asserted in `test_configured_headers_sent_on_wire`;
- reserved-name rejection is covered by the `validate()` unit tests in `client_settings.rs`.

The belt-and-suspenders insert order in `try_new` (static headers first, protocol headers last) stays as cheap insurance.

— Reply drafted by an AI assistant (GitHub Copilot CLI) on the author's behalf.

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/mod.rs
```diff
@@ -1010,6 +1022,176 @@ mod test {
...
     }
 
+    /// Runs a minimal HTTP server that records the headers of the first request
+    /// it receives into `captured` and always replies `200 OK`. Used to assert
+    /// what the exporter actually puts on the wire.
+    fn run_header_capture_server(
+        tokio_rt: &Runtime,
+        endpoint_addr: &str,
+        captured: Arc<std::sync::Mutex<Option<HeaderMap>>>,
+    ) -> CancellationToken {
```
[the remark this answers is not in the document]
timr-dev: Switched the capture test server to `parking_lot::Mutex` in 2d8391901 (one of the two options you suggested). In practice the guard was only ever held for a single synchronous assignment and never across an `.await`, so it couldn't block a worker across a yield — but `parking_lot::Mutex` is the crate's convention here, it can't be poisoned, and it drops the `.unwrap()` on the guard. Thanks!

— Reply drafted by an AI assistant (GitHub Copilot CLI) on the author's behalf.

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -245,6 +260,21 @@ impl GrpcClientSettings {
                 ));
             }
         }
+
+        for (name, value) in &self.headers {
+            if name.parse::<MetadataKey<tonic::metadata::Ascii>>().is_err() {
+                return Err(GrpcEndpointError::InvalidConfig(format!(
+                    "header name \"{name}\" is not a valid ASCII gRPC metadata key"
+                )));
+            }
```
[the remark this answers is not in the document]
timr-dev: Reworded the message in 2d8391901. One nuance worth flagging: I checked tonic's behaviour directly, and `MetadataKey::<Ascii>` actually *accepts* mixed-case keys and normalizes them to lowercase (`Authorization` -> `authorization`, `X-Test` -> `x-test`) rather than rejecting them — so "lowercase-only" isn't quite the input constraint. The cases it genuinely rejects are non-token characters and a trailing `-bin` (reserved for binary metadata). The new message reflects that:

> header name "{name}" is not a valid gRPC metadata key (expected an HTTP/2 token: ASCII letters, digits, or `-_.`; the key is sent lowercased and must not end with `-bin`, which is reserved for binary metadata)

Happy to adjust the wording further if you'd prefer something else.

— Reply drafted by an AI assistant (GitHub Copilot CLI) on the author's behalf.

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -153,6 +155,16 @@ pub struct GrpcClientSettings {
...
+
+    /// Static metadata (headers) added to every outbound OTLP/gRPC request
+    /// (e.g. an `authorization` or tenant-routing header). Iteration follows
+    /// the declaration order in the config.
+    ///
+    /// Keys and values must be valid ASCII gRPC metadata; this is enforced by
+    /// [`GrpcClientSettings::validate`]. These coexist with any header
+    /// propagation policy configured on the exporter.
+    #[serde(default)]
+    pub headers: IndexMap<String, String>,
```
lalitb: `GrpcClientSettings` is also used both by both  `exporter:otap`, and `exporter:otlp_grpc`.
                                            
So if someone configures `headers` on `exporter:otap`, the config will pass validation, but those headers are not added to the OTAP streaming requests today. That means an auth header could look valid in config but never actually be sent.
                    
We should either make header application shared so every `GrpcClientSettings` user gets it automatically, or make `exporter:otap` reject `headers` until it supports them.

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
```diff
@@ -630,63 +639,117 @@ async fn finalize_completed_export(
...
+                    };
+                    let Ok(value) = MetadataValue::try_from(header.value) else {
+                        otel_debug!(
+                            "otlp.exporter.grpc.header_skip",
+                            reason = "invalid ascii metadata value",
+                            header_name = header.header_name
+                        );
+                        continue;
+                    };
+                    let _ = metadata.append(key, value);
```
lalitb, line 748: I think this creates a collision case between configured static metadata and propagated metadata.  

For example, if both contain `authorization`, we may send duplicate auth values. That seems risky because propagated headers come from inbound traffic, while configured headers usually represent the backend credential.                                                                            

The Go Collector avoids duplicate configured gRPC headers by only adding them when the key is not already present. We should also have one clear precedence rule here. My preference is that configured static headers win, so propagated headers with the same key are skipped, unless there are other views cc @open-telemetry/arrow-maintainers.

timr-dev: Good catch — fixed in d71a848. `build_grpc_metadata` now drops a propagated header whose key collides with a statically configured one, so static config wins: a configured backend credential like `authorization` can no longer be overridden or duplicated by inbound transport headers. Non-colliding propagated headers still flow through, and duplicate *propagated* values (with no static collision) are still preserved as before. Added `test_build_grpc_metadata_static_wins_over_propagated_collision` and documented the precedence rule in the exporter README.

<sub>↳ Drafted by an AI assistant and posted on @timr-dev's behalf.</sub>

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
timr-dev: Good catch — thanks. Went with your second option: **`exporter:otap` now rejects a non-empty `headers` map** at config load rather than silently dropping it.

Why reject instead of sharing application: `user_agent` is applied at the tonic `Endpoint` level (`build_endpoint`), so both exporters inherit it for free. But `headers` are per-request metadata — the OTLP/gRPC exporter attaches them in `build_grpc_metadata`, whereas the OTAP exporter streams over long-lived Arrow clients (`ArrowLogsServiceClient::new(channel)`, etc.) and never builds per-message metadata. Wiring headers into those streams means either a tonic interceptor (changes the client type and ripples through the `StreamingArrowService` impls/bounds) or a trait-signature change — a larger change on a streaming hot path, beyond this PR's OTLP scope. So failing loudly is the right call for now.

Implementation (commit 73e272a91):
- `reject_unsupported_headers` guard wired into **both** `validate_config` and `from_config`; a non-empty `headers` map → `InvalidUserConfig` that points operators to `exporter:otlp_grpc` / `exporter:otlp_http`.
- Doc caveat on the `headers` field, a note in the OTAP exporter README "Limits", and a changelog line.
- 3 tests: reject via `from_config`, reject via `validate_config`, accept empty/absent (backwards compatible).

Happy to file a follow-up issue for OTAP-native static headers (the interceptor approach) if you'd like it tracked separately.

<sub><i>Drafted by an AI coding assistant on @timr-dev's behalf; reviewed before posting.</i></sub>

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -153,6 +155,21 @@ pub struct GrpcClientSettings {
...
+    /// Keys and values must be valid ASCII gRPC metadata; this is enforced by
+    /// [`GrpcClientSettings::validate`]. These coexist with any header
+    /// propagation policy configured on the exporter.
+    ///
+    /// Note: `GrpcClientSettings` is shared by the OTLP/gRPC exporter and the
+    /// OTAP (Arrow) exporter, but only the OTLP/gRPC exporter applies these
+    /// headers today. The OTAP exporter rejects a non-empty `headers` map at
+    /// config validation rather than silently dropping it.
+    #[serde(default)]
+    pub headers: IndexMap<String, String>,
```
utpilla: Could we use a HashMap instead? I believe we don't have to care about ordering here.

lalitb (approved):
LGTM, thanks for your first contribution and welcome!

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
```diff
@@ -93,6 +109,32 @@ impl HttpClientSettings {
                 ));
             }
         }
+
+        for (name, value) in &self.headers {
```
utpilla, line 116: Since HTTP header names are case -insensitive, let's reject case-insensitive duplicates as well in this method.

timr-dev: Done in 6d1a5c9b3 — `HttpClientSettings::validate` now rejects case-insensitive duplicate header names, keyed on the canonical lowercased name, so e.g. `X-Tenant` and `x-tenant` configured together fail at config load instead of colliding on the wire. I applied the same check to `GrpcClientSettings::validate` (tonic sends metadata keys lowercased, so they'd collide the same way) and added a regression test to each.

> _Drafted by an AI assistant acting on @timr-dev's behalf._

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
```diff
@@ -93,6 +111,43 @@ impl HttpClientSettings {
...
+                return Err(HttpClientError::InvalidConfig(format!(
+                    "header \"{name}\" has a value that cannot be represented as an HTTP header \
+                     value (must be visible ASCII)"
+                )));
+            }
+            // Reject protocol headers the exporter sets itself, so user config can
+            // never collide with or shadow them.
+            if matches!(
+                header_name.as_str(),
+                "content-type" | "content-encoding" | "content-length" | "host"
```
utpilla: Could we add `accept` and `accept-encoding` as well to this list?

Both are response-negotiation headers whose value must match what the client can actually parse/decompress and not something a user can truthfully declare.

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/README.md
```diff
@@ -53,10 +53,34 @@ config:
...
 ```
 
 Shared HTTP client fields include concurrency limit, connect timeout, request
-timeout, TCP keepalive, TLS, and request-body compression.
+timeout, TCP keepalive, TLS, request-body compression, and static request
+`headers`.
+
+### Static request headers
+
+`http.headers` is an ordered map of header name to value applied to every
```
cijothomas: ordering is important and guaranteed?

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -245,6 +267,37 @@ impl GrpcClientSettings {
...
+                .parse::<MetadataKey<tonic::metadata::Ascii>>()
+                .map_err(|_| {
+                    GrpcEndpointError::InvalidConfig(format!(
+                        "header name \"{name}\" is not a valid gRPC metadata key (expected an \
+                         HTTP/2 token: ASCII letters, digits, or `-_.`; the key is sent \
+                         lowercased and must not end with `-bin`, which is reserved for \
+                         binary metadata)"
+                    ))
+                })?;
+            if MetadataValue::try_from(value.as_str()).is_err() {
```
ThomsonTan, line 296: The HTTP exporter rejects reserved header names but the gRPC one accepts anything, like `grpc-timeout` could be passed from the user to the gRPC server which tells it to abort. Suggest mirroring the HTTP reserved-key check in `GrpcClientSettings::validate` and rejecting it too.

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
timr-dev: Good call — switched the static `headers` field to `HashMap<String, String>` in both `GrpcClientSettings` and `HttpClientSettings`. The headers are applied independently per request, so declaration order isn't meaningful; I dropped the ordering language from the docs/tests and removed the now-unused `indexmap` dependency. Done in 5df6bd388.

> _Drafted by an AI assistant and pushed on @timr-dev's behalf._

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
timr-dev: Done — added `accept` and `accept-encoding` to the reserved list in `HttpClientSettings::validate`, alongside the existing `content-type` / `content-encoding` / `content-length` / `host`. Their effective value is dictated by what the client can actually decode/decompress, so a user-declared value would be misleading. Updated the `validate` doc, the reserved-header test, and the HTTP exporter README. Done in 5df6bd388.

> _Drafted by an AI assistant and pushed on @timr-dev's behalf._

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/README.md
timr-dev: No — with the switch to a `HashMap`, ordering is neither important nor guaranteed for static headers (each header is applied independently). I removed the "ordered map" wording from both exporter READMEs and the field docs. Done in 5df6bd388.

> _Drafted by an AI assistant and pushed on @timr-dev's behalf._

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
timr-dev: Good catch — mirrored the HTTP reserved-header check in `GrpcClientSettings::validate`. It now rejects transport-managed metadata (`content-type`, `te`, `user-agent`) and any name using the spec-reserved `grpc-` prefix, so `grpc-timeout`, `grpc-encoding`, etc. can't be set via `headers` to alter call semantics such as the server-side deadline. Added a `validate_rejects_reserved_grpc_metadata` test and documented the reserved names in the gRPC exporter README. Done in 8ffed2b41.

> _Drafted by an AI assistant and pushed on @timr-dev's behalf._

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/README.md
```diff
@@ -53,10 +53,37 @@ config:
...
   http:
     compression: gzip
+
+    # Static headers added to every outbound OTLP/HTTP request (optional).
+    # Useful for authentication or backend routing. Protocol headers
+    # (Content-Type / Content-Encoding / Content-Length / Host) and
+    # response-negotiation headers (Accept / Accept-Encoding) cannot be set
+    # here and are rejected at config load.
+    headers:
+      authorization: "Basic <base64(user:password)>"
```
jmacd:
```suggestion
    # Static headers added to every outbound OTLP/HTTP request (optional).
    # Useful for arbitrary headers (e.g. backend routing). Not recommended for 
    # authorization, use an Auth extension instead. A number of standard HTTP 
    # headers (e.g., Content-Type, Accept-Encoding) cannot be set here and are 
    # rejected at config load.
    headers:
      environment: "production-west"
```

nit: It's more conventional (in OTel) to include basic HTTP auth via an Auth extension, for example [here is the Golang form of basic auth](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/extension/basicauthextension/README.md).

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
jmacd: Collector has a type called [`configopaque.String`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configopaque/opaque.go) that it would use for the user-supplied value in this spot which protects against accidentally leaking sensitive data into the internal telemetry stream.

In Rust, I recommend that we use the [Microsoft Oxidizer `data-privacy` crate for this](https://github.com/microsoft/oxidizer/blob/main/crates/data_privacy/README.md).

Non-blocking, let's file an issue for this.

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/README.md
timr-dev: Done in 40e1e3cfe. Reworked the example comment and prose in both exporter READMEs (and the `configs/` example) along the lines of your suggestion: the headers example now leads with non-secret routing metadata (`x-scope-orgid` / `environment`), and the comment notes that authorization is better handled by a dedicated Auth extension rather than a hard-coded header. Two small deviations from the literal suggestion: I kept the full reserved-header list (`content-type` / `content-encoding` / `content-length` / `host` + `accept` / `accept-encoding`) for precision instead of the abbreviated `e.g.` form, and used "prefer a dedicated Auth extension" rather than "use the X extension" since that extension surface is still being built out. Thanks for the pointer to the Go `basicauthextension`.

> _Drafted by an AI assistant and pushed on @timr-dev's behalf._

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
timr-dev: Filed #3306 to track this. It captures wrapping the user-supplied header values in a privacy/opaque type — referencing Collector's `configopaque.String` and the Microsoft Oxidizer `data_privacy` crate you linked — across both the gRPC and HTTP `client_settings.rs` (with a note to audit `user_agent` and any `Debug`/telemetry paths too). Keeping it as a non-blocking follow-up to this PR, as you suggested.

> _Drafted by an AI assistant and pushed on @timr-dev's behalf._

utpilla (approved):
@timr-dev Thanks for the contribution and welcome!

ThomsonTan (approved):

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
timr-dev: Filed #3314 to track OTAP-native static header support (the interceptor / initial-stream-metadata approach discussed above), so the loud rejection can be removed once it lands. It also notes that header values there should adopt the privacy/opaque type from #3306.

> _Drafted by an AI coding assistant on @timr-dev's behalf._

PR #3311, opened 2026-06-17
Suppress noisy opentelemetry-prometheus MetricValidationFailed

### LINKED ISSUE #2734: Metric description conflict: duplicate pending_sends_buffered metric across runtime-control and completion scopes

andborja:
# Change Summary

Suppress noisy opentelemetry-prometheus MetricValidationFailed warning message.

Issue https://github.com/open-telemetry/otel-arrow/issues/2734

* Closes #2734

- Unit tests added.
- Local tests using the configuration attached to the issue as reproduce.

### Changelog


rust/otap-dataflow/crates/telemetry/src/tracing_init.rs
```diff
@@ -22,10 +22,38 @@ use tracing_subscriber::{EnvFilter, Registry, layer::SubscriberExt};
 /// Otherwise, the level's [`RUST_LOG`-style directive string][env-filter] is
 /// passed directly to [`EnvFilter`].
 ///
+/// In all cases the filter suppresses one specific benign per-scrape warning
```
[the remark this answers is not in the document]
andborja: Improved

rust/otap-dataflow/.chloggen/suppress-prometheus-metric-validation-warning.yaml
```diff
@@ -0,0 +1,22 @@
+change_type: bug_fix
+
+component: observability
+
+note: Suppress noisy per-scrape `MetricValidationFailed` warnings from the Prometheus pull exporter
+
+issues: [2734]
+
+subtext: |
+  The `opentelemetry-prometheus` crate flattens OpenTelemetry
```
cijothomas, line 10: The admin endpoint was modified recently to follow OTel spec for prometheus mapping, so this should now be hit even when OTel sdk is not in the picture. Are we fixing that too?

andborja: This is fixing an issue with the opentelemetry-prometheus integration only.

drewrelmas (approved):
This looks good to me, would request @cijothomas's review as well since he was involved on the issue.

rust/otap-dataflow/.chloggen/suppress-prometheus-metric-validation-warning.yaml
cijothomas: we are planning to remove otel* crate dependency and directly have ITS for metrics too. How would it be fixed?

rust/otap-dataflow/crates/telemetry/src/tracing_init.rs
```diff
@@ -18,14 +18,45 @@ use tracing_subscriber::{EnvFilter, Registry, layer::SubscriberExt};
...
 pub fn create_env_filter(level: &LogLevel) -> EnvFilter {
-    EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new(level.as_str()))
+    let filter =
+        EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new(level.as_str()));
+    // Target matches `env!("CARGO_PKG_NAME")` set by opentelemetry's `otel_warn!`.
+    // The `[{metric_description}]` field filter matches only the benign
+    // description-conflict `MetricValidationFailed` event, leaving the
+    // data-dropping type-conflict variant (field `metric_type`) visible.
+    filter.add_directive(
+        "opentelemetry-prometheus[{metric_description}]=error"
```
cijothomas, line 56: relying on such specific filtering can break if upstream changes.

cijothomas (commented):
Might be okay as temporary fix. The underlying issue needs to be understood and fixed else this will come back when OTel upstream changes, and also - this is likely already an issue with admin/prometheus endpoint too.

rust/otap-dataflow/.chloggen/suppress-prometheus-metric-validation-warning.yaml
andborja: The internal telemetry pipeline will replace the metrics at some point. A new prometheus exporter for otel arrow will be implemented at that time.

rust/otap-dataflow/crates/telemetry/src/tracing_init.rs
andborja: True. Will not likely change as opentelemetry-prometheus is not actively maintained.

andborja:
> Might be okay as temporary fix. The underlying issue needs to be understood and fixed else this will come back when OTel upstream changes, and also - this is likely already an issue with admin/prometheus endpoint too.

Let's make sure we consider this when implementing the otel-arrow prometheus exporter.

drewrelmas:
Merging with understanding that there is more work to do here when we move to our own Internal Telemetry System for metrics and remove dependency on SDK.

PR #3313, opened 2026-06-18
chore(ci): broaden PR size labeler ignore list for lock/sum files

cijothomas:
Follow-up to #2989.

Two adjustments based on PRs opened after that merge:

- `"go.sum"` → `"*.sum"`. The labeler uses bash glob matching (not minimatch), so `"go.sum"` only matched a root-level file and missed `go/go.sum` and `collector/cmd/otelarrowcol/go.sum`. This was also flagged by @ThomsonTan in the original PR review.
- Added `"*.lock.txt"` to catch `requirements.lock.txt` (used under `tools/pipeline_perf_test/`). Without this, renovate/dependabot PRs that touch only lockfiles get inflated labels — e.g. #3312 was labeled `size/M` even though the only non-generated change is 2 lines in `requirements.txt`.

No changelog entry — workflow-only chore.


ThomsonTan (approved):

albertlockett (approved):
nice, thanks @cijothomas

Issue #3306, opened 2026-06-16
OTLP exporters: wrap static exporter header values in a privacy/opaque type to prevent secret leakage into telemetry


timr-dev:
### Background

[PR #3303](https://github.com/open-telemetry/otel-arrow/pull/3303) added a static `headers` config field to the OTAP OTLP exporters (HTTP + gRPC). Header **values** can carry secrets (e.g. an `authorization` credential, a tenant API key). Today those values are plain `String`s in `HttpClientSettings` / `GrpcClientSettings`, so they can accidentally leak into the internal telemetry stream, logs, or `Debug` output.

This was raised by @jmacd as a non-blocking follow-up in PR review: https://github.com/open-telemetry/otel-arrow/pull/3303#discussion_r3423314274

### Problem

User-supplied sensitive config values are not marked as sensitive. There is no redaction at the type level, so any `Debug`/trace/diagnostic that touches the settings struct (or a future one that does) risks emitting the secret in cleartext.

### Prior art

- Go Collector uses [`configopaque.String`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configopaque/opaque.go) for exactly this — the value redacts itself in marshaling/`String()` while still being usable by the component.
- In Rust, @jmacd recommends the Microsoft Oxidizer [`data_privacy`](https://github.com/microsoft/oxidizer/blob/main/crates/data_privacy/README.md) crate.

### Proposal

Wrap user-supplied header values (and consider `user_agent`) in an opaque / self-redacting type so that:

- `Debug`/`Display`/serialization redact the value (e.g. `[REDACTED]`),
- the raw value is only obtainable via an explicit accessor used at the request-construction site,
- behavior is otherwise unchanged.

### Scope

- `rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs` — `HttpClientSettings::headers`
- `rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs` — `GrpcClientSettings::headers`
- Audit any `Debug`/telemetry paths that currently print these settings.

### Notes

Non-blocking follow-up to #3303; the static-headers feature itself is unaffected. Filing per the maintainer's request so the privacy-typing work can be tracked and scoped separately.

timr-dev:
Go ahead and assign this to me, I'll take care of it once the OTAP exporter is merged.

PR #3315, opened 2026-06-18
OTAP exporter: support static request headers natively

### LINKED ISSUE #3314: OTAP exporter: support static request headers natively (remove the config-load rejection)
### Background

[PR #3303](https://github.com/open-telemetry/otel-arrow/pull/3303) added a static `headers` config map to the OTLP **HTTP** and **gRPC** exporters. The native **OTAP (Arrow-protocol) exporter** (`urn:otel:exporter:otap`) does **not** support it yet — instead it **rejects** a non-empty `headers` map at config load (`reject_unsupported_headers`, wired into both `validate_config` and `from_config`) so an operator can't configure a header that would silently never be sent. This issue tracks adding real OTAP-native support so the rejection can be removed.

Raised as a follow-up to the review thread on #3303 (maintainer @lalitb): https://github.com/open-telemetry/otel-arrow/pull/3303#discussion_r3420022222

### Why it was deferred

`GrpcClientSettings` is shared by `exporter:otap` and `exporter:otlp_grpc`, but the two apply request headers very differently:

- **`user_agent`** is applied at the tonic `Endpoint` level (`GrpcClientSettings::build_endpoint`), so **both** exporters inherit it for free.
- **`headers`** are applied as **per-request metadata** inside the OTLP/gRPC exporter's `build_grpc_metadata`. The OTAP exporter streams over long-lived Arrow clients (`ArrowLogsServiceClient::new(channel)`, `ArrowMetricsServiceClient`, `ArrowTracesServiceClient`) and never builds per-message metadata, so there is no existing injection point.

Wiring headers into the Arrow streams is a larger change on a streaming hot path, which is why #3303 (scoped to OTLP) deferred it.

### Proposed approaches

Static headers on a streaming RPC only need to be attached **once, as the initial request metadata when the stream is opened** — they do not need to be re-sent per message. So the hot path (per-`BatchArrowRecords` send) should stay allocation-free; the cost is paid once per stream establishment.

1. **tonic interceptor (recommended to evaluate first).** Wrap the channel in `InterceptedService` with an interceptor that injects the static metadata. Trade-off: changes the client type (`ArrowLogsServiceClient<Channel>` → `ArrowLogsServiceClient<InterceptedService<Channel, F>>`), which ripples through the `StreamingArrowService` impls and their generic bounds.
2. **Set initial metadata on the streaming request.** Build the `tonic::Request` for the stream open with the static `MetadataMap` (mirroring `build_static_grpc_metadata` in the OTLP/gRPC exporter) rather than intercepting at the channel. May be a smaller blast radius if the stream-open path is centralized.
3. **Share header application at the `GrpcClientSettings` layer** so every consumer gets it automatically (the maintainer's first option). Broadest, but couples the shared settings type to a tonic application strategy.

The implementer should benchmark/inspect the per-message path to confirm zero added per-`BatchArrowRecords` allocation, consistent with the OTLP/gRPC zero-alloc fast path.

### Scope / acceptance criteria

- Configuring `headers` on `exporter:otap` sends them as initial stream metadata on the Arrow log/metric/trace streams.
- Remove `reject_unsupported_headers` (and its 3 tests) once supported; replace with positive tests (configured header present on the outbound stream; empty/absent still works — backwards compatible).
- Reuse the existing validation in `GrpcClientSettings::validate()` (reserved-metadata rejection, ASCII checks).
- Preserve the zero-alloc per-message hot path; document where the one-time cost is paid.
- Update the OTAP exporter README "Limits" note and add a changelog entry.

### Files

- `rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs` — stream client construction (`create_req_stream`, `StreamingArrowService` impls); current `reject_unsupported_headers` guard.
- `rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs` — shared `GrpcClientSettings` (`headers`, `validate()`, `build_endpoint`).
- Reference implementation to mirror: `otlp_grpc_exporter/mod.rs` (`build_static_grpc_metadata` / `build_grpc_metadata`).

### Related

- Follow-up to #3303 (the OTLP HTTP+gRPC static-headers feature).
- Complements #3306: when implemented, OTAP header **values** should adopt the same privacy/opaque type proposed there, so secrets don't leak into telemetry.

_Filed by an AI coding assistant on @timr-dev's behalf, per the offer made in the #3303 review thread._

timr-dev:
Closes #3314.

## What

The native OTAP (Arrow-protocol) exporter (`urn:otel:exporter:otap`) now supports static request `headers`, attaching them as the **initial metadata of each Arrow log/metric/trace stream**. The previous config-load rejection (`reject_unsupported_headers`) is removed.

## Approach

- **Shared builder at the `GrpcClientSettings` layer** (the maintainer's preferred option from the #3303 review thread): added `build_static_metadata(&HashMap) -> Option<MetadataMap>` as a single source of truth, and refactored the OTLP/gRPC exporter to delegate to it (removing its private duplicate).
- **Attach once per stream (re)open**: in `stream_arrow_batches`, the pre-built `Option<Arc<MetadataMap>>` template is cloned onto the stream-open `Request` via `into_streaming_request()` + `metadata_mut()`. The `StreamingArrowService` trait and its impls are untouched.
- **Hot path preserved**: `create_req_stream` (the per-`BatchArrowRecords` path) is unchanged; the only added cost is one `MetadataMap` clone per stream establishment (documented in the README "Limits").
- **Validation reused**: header validation stays in `GrpcClientSettings::validate()` (ASCII, reserved gRPC metadata, case-insensitive duplicates), now also invoked from `from_config` as defense-in-depth, and skipped/invalid entries are logged.

## Tests

- Removed the 2 rejection tests; added positive config tests (accept valid headers; reject reserved) plus a `from_config` rejection test.
- End-to-end: configured headers are present as **initial stream metadata on logs, metrics, AND traces**; multiple headers; empty/absent map sends no metadata.
- Deterministic unit test that headers are re-applied on **every** stream (re)open.
- Builder unit tests (empty → `None`, builds map, skips invalid).
- `cargo xtask check` passes (structure, `fmt`, `clippy -D warnings`, workspace tests).

## Notes

Complements #3306: OTAP header **values** should later adopt the same opaque/secret type so credentials don't leak into telemetry.

### DIFF
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
@@ -138,7 +138,7 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
   `GrpcClientSettings` (ASCII-only, no gRPC-reserved metadata).
   - Because the metadata is bound when a stream opens, a changed `headers` value
     (e.g. a rotated `authorization` or tenant header) only takes effect on the
-    next stream open/reconnect — unlike the OTLP/gRPC exporter, which rebuilds
+    next stream open/reconnect, unlike the OTLP/gRPC exporter, which rebuilds
     metadata per unary request. Existing long-lived streams keep using the value
     they were opened with until they are re-established.
 - Compression values are limited to the variants supported by the shared OTAP
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
@@ -136,6 +136,11 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
   allocation-free; the one-time cost is a single metadata clone per stream
   establishment. Header names/values are validated at config load via the shared
   `GrpcClientSettings` (ASCII-only, no gRPC-reserved metadata).
+  - Because the metadata is bound when a stream opens, a changed `headers` value
+    (e.g. a rotated `authorization` or tenant header) only takes effect on the
+    next stream open/reconnect — unlike the OTLP/gRPC exporter, which rebuilds
+    metadata per unary request. Existing long-lived streams keep using the value
+    they were opened with until they are re-established.
 - Compression values are limited to the variants supported by the shared OTAP
   transport layer.
 - End-to-end delivery depends on downstream OTAP stream responses.
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
@@ -136,11 +136,11 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
   allocation-free; the one-time cost is a single metadata clone per stream
   establishment. Header names/values are validated at config load via the shared
   `GrpcClientSettings` (ASCII-only, no gRPC-reserved metadata).
-  - Because the metadata is bound when a stream opens, a changed `headers` value
-    (e.g. a rotated `authorization` or tenant header) only takes effect on the
-    next stream open/reconnect, unlike the OTLP/gRPC exporter, which rebuilds
-    metadata per unary request. Existing long-lived streams keep using the value
-    they were opened with until they are re-established.
+  - Because the metadata template is built once at exporter startup, a changed
+    `headers` value (e.g. a rotated `authorization` or tenant header) only takes
+    effect when the exporter is restarted/reconfigured and not on stream
+    reconnects, which reuse the template captured at startup. This differs from
+    the OTLP/gRPC exporter, which rebuilds metadata per unary request.
 - Compression values are limited to the variants supported by the shared OTAP
   transport layer.
 - End-to-end delivery depends on downstream OTAP stream responses.
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
@@ -53,6 +53,12 @@ config:
   # Number of streams opened for each signal type (default: 1).
   streams_per_signal: 1
 
+  # Optional static gRPC request headers (metadata) attached once as the
+  # initial metadata of each Arrow stream (e.g. auth or tenant routing).
+  headers:
+    authorization: "Basic dXNlcjpwYXNz"
+    x-scope-orgid: "tenant-1"
+
   # Optional RPC timeout.
   timeout: 30s
 ```
@@ -124,11 +130,12 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
 ## Limits
 
 - `stream_queue_capacity` and `streams_per_signal` must be greater than zero.
-- Static request `headers` are not supported yet: the OTAP exporter shares
-  `GrpcClientSettings` with the OTLP/gRPC exporter, but does not attach static
-  metadata to its Arrow streams, so a non-empty `headers` map is rejected at
-  config validation rather than silently dropped. Use `exporter:otlp_grpc` or
-  `exporter:otlp_http` to send static headers.
+- Static request `headers` are attached once, as the initial metadata of each
+  Arrow stream when it is opened (or re-opened after a reconnect). They are not
+  re-sent per `BatchArrowRecords`, so the per-message hot path stays
+  allocation-free; the one-time cost is a single metadata clone per stream
+  establishment. Header names/values are validated at config load via the shared
+  `GrpcClientSettings` (ASCII-only, no gRPC-reserved metadata).
 - Compression values are limited to the variants supported by the shared OTAP
   transport layer.
 - End-to-end delivery depends on downstream OTAP stream responses.
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
@@ -25,7 +25,6 @@ use otap_df_engine::node::NodeId;
 use otap_df_engine::terminal_state::TerminalState;
 use otap_df_otap::OTAP_EXPORTER_FACTORIES;
 use otap_df_otap::metrics::ExporterPDataMetrics;
-use otap_df_otap::otap_grpc::client_settings::build_static_metadata;
 use otap_df_otap::pdata::OtapPdata;
 use otap_df_pdata::Producer;
 use otap_df_pdata::TryIntoWithOptions;
@@ -479,7 +478,7 @@ impl local::Exporter<OtapPdata> for OTAPExporter {
         // shared across all stream workers via `Arc`; each stream attaches a
         // clone as its initial metadata when it opens (see `stream_arrow_batches`),
         // so the per-`BatchArrowRecords` send path is never touched.
-        let static_metadata = build_static_metadata(&self.config.grpc.headers).map(Arc::new);
+        let static_metadata = self.config.grpc.build_static_metadata().map(Arc::new);
 
         // Operational breadcrumb: confirm which static header KEYS were loaded
         // (never the values, which may be credentials). Lets an operator verify
@@ -2693,7 +2692,7 @@ mod tests {
         let cap_logs = captured_tx.clone();
         let cap_metrics = captured_tx.clone();
         let cap_traces = captured_tx;
-        _ = tokio_rt.spawn(async move {
+        let server_handle = tokio_rt.spawn(async move {
             let tcp_listener = TcpListener::bind(listening_addr).await.unwrap();
             let _ = ready_sender.send(());
             let tcp_stream = TcpListenerStream::new(tcp_listener);
@@ -2790,6 +2789,11 @@ mod tests {
             });
 
         _ = shutdown_sender.send("Shutdown");
+        // Await the server task so teardown is deterministic and no background
+        // task is left pending at runtime drop.
+        tokio_rt
+            .block_on(server_handle)
+            .expect("server task should shut down cleanly");
     }
 
     /// Backwards compatibility: with no `headers` configured, the exporter must
@@ -2809,7 +2813,7 @@ mod tests {
         let listening_addr: SocketAddr = format!("{grpc_addr}:{grpc_port}").parse().unwrap();
         let tokio_rt = Runtime::new().unwrap();
 
-        _ = tokio_rt.spawn(async move {
+        let server_handle = tokio_rt.spawn(async move {
             let tcp_listener = TcpListener::bind(listening_addr).await.unwrap();
             let _ = ready_sender.send(());
             let tcp_stream = TcpListenerStream::new(tcp_listener);
@@ -2877,6 +2881,11 @@ mod tests {
             });
 
         _ = shutdown_sender.send("Shutdown");
+        // Await the server task so teardown is deterministic and no background
+        // task is left pending at runtime drop.
+        tokio_rt
+            .block_on(server_handle)
+            .expect("server task should shut down cleanly");
     }
 
     /// Reliability: the configured static headers must be applied to the request
@@ -2908,8 +2917,11 @@ mod tests {
 
         let mut headers = std::collections::HashMap::new();
         let _ = headers.insert(HDR_AUTH.to_string(), HDR_AUTH_VAL.to_string());
-        let static_metadata =
-            otap_df_otap::otap_grpc::client_settings::build_static_metadata(&headers).map(Arc::new);
+        let settings = otap_df_otap::otap_grpc::client_settings::GrpcClientSettings {
+            headers,
+            ..Default::default()
+        };
+        let static_metadata = settings.build_static_metadata().map(Arc::new);
         assert!(
             static_metadata.is_some(),
             "test setup should produce metadata"
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
@@ -25,6 +25,7 @@ use otap_df_engine::node::NodeId;
 use otap_df_engine::terminal_state::TerminalState;
 use otap_df_otap::OTAP_EXPORTER_FACTORIES;
 use otap_df_otap::metrics::ExporterPDataMetrics;
+use otap_df_otap::otap_grpc::client_settings::build_static_metadata;
 use otap_df_otap::pdata::OtapPdata;
 use otap_df_pdata::Producer;
 use otap_df_pdata::TryIntoWithOptions;
@@ -38,7 +39,7 @@ use otap_df_pdata::proto::opentelemetry::arrow::v1::{
 };
 use otap_df_telemetry::instrument::{Gauge, Mmsc};
 use otap_df_telemetry::metrics::MetricSet;
-use otap_df_telemetry::{otel_error, otel_info, otel_warn};
+use otap_df_telemetry::{otel_debug, otel_error, otel_info, otel_warn};
 use otap_df_telemetry_macros::metric_set;
 use serde_json::Value;
 use std::collections::HashMap;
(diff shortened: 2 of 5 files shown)


rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
```diff
@@ -124,11 +130,12 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
 ## Limits
 
 - `stream_queue_capacity` and `streams_per_signal` must be greater than zero.
-- Static request `headers` are not supported yet: the OTAP exporter shares
-  `GrpcClientSettings` with the OTLP/gRPC exporter, but does not attach static
-  metadata to its Arrow streams, so a non-empty `headers` map is rejected at
-  config validation rather than silently dropped. Use `exporter:otlp_grpc` or
-  `exporter:otlp_http` to send static headers.
+- Static request `headers` are attached once, as the initial metadata of each
```
lalitb, line 133: Small doc clarity thought: since these headers are initial metadata on long-lived streams, should we mention the rotation behavior here? A changed auth/tenant header would only take effect on the next stream open/reconnect, unlike OTLP unary requests where metadata is rebuilt per request.

lalitb (approved):
LGTM. Nicely done!

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -241,6 +241,55 @@ fn validate_grpc_endpoint(endpoint: &str) -> Result<(), String> {
...
+/// lets callers keep a zero-allocation fast path when no headers are set.
+///
+/// Header names/values are validated up front by [`GrpcClientSettings::validate`]
+/// (ASCII key/value, no reserved gRPC metadata, no case-insensitive duplicates),
+/// so the per-entry parse below cannot fail for a validated config. Any entry
+/// that somehow fails to parse (e.g. a programmatic caller that bypassed
+/// validation) is skipped defensively and logged at debug level rather than
+/// panicking, so a dropped header still leaves a breadcrumb.
+#[must_use]
+pub fn build_static_metadata(headers: &HashMap<String, String>) -> Option<MetadataMap> {
```
ThomsonTan: `build_static_metadata` is a free `pub fn` taking `&HashMap`, decoupled from the validated `GrpcClientSettings`. Both production callers pass `&self.config.grpc.headers`, so consider making it a method `GrpcClientSettings::build_static_metadata(&self)` to keep the public surface narrow. Tests can construct a settings value (or keep a thin free helper). Not blocking.

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -241,6 +241,55 @@ fn validate_grpc_endpoint(endpoint: &str) -> Result<(), String> {
...
+        };
+        let Ok(val) = MetadataValue::try_from(value.as_str()) else {
+            otap_df_telemetry::otel_debug!(
+                "grpc.client.static_header_skip",
+                reason = "invalid ascii metadata value",
+                header_name = name.as_str()
+            );
+            continue;
+        };
+        let _ = metadata.insert(key, val);
```
ThomsonTan: `metadata.insert(key, val)` is used while iterating a `HashMap`, so two entries differing only in case (`X-Tenant` vs `x-tenant`) collapse to one lowercased gRPC key and the survivor is non-deterministic by hash order. `validate()` already rejects case-insensitive duplicates at config load (and both `validate_config` and `from_config` now call it), so this is unreachable for a validated config. Since the `fn` is `pub` and documented as defensive for unvalidated callers, consider either a debug log on collision or a doc note that de-dup is the caller's (validate's) responsibility.

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
```diff
@@ -2452,4 +2543,405 @@ mod tests {
...
+        let tokio_rt = Runtime::new().unwrap();
+
+        let cap_logs = captured_tx.clone();
+        let cap_metrics = captured_tx.clone();
+        let cap_traces = captured_tx;
+        _ = tokio_rt.spawn(async move {
+            let tcp_listener = TcpListener::bind(listening_addr).await.unwrap();
+            let _ = ready_sender.send(());
+            let tcp_stream = TcpListenerStream::new(tcp_listener);
+            Server::builder()
```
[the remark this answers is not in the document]
timr-dev: Good catch — fixed in 2f3b3fb. Both new gRPC tests now capture the server `JoinHandle` and `tokio_rt.block_on(server_handle)` after sending the shutdown signal, matching the other gRPC tests in this module so teardown is deterministic.

— Reply posted by Copilot (AI assistant) on behalf of @timr-dev. [GS]

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
```diff
@@ -2452,4 +2543,405 @@ mod tests {
...
+        let grpc_port = portpicker::pick_unused_port().expect("No free ports");
+        let grpc_endpoint = format!("http://{grpc_addr}:{grpc_port}");
+        let listening_addr: SocketAddr = format!("{grpc_addr}:{grpc_port}").parse().unwrap();
+        let tokio_rt = Runtime::new().unwrap();
+
+        _ = tokio_rt.spawn(async move {
+            let tcp_listener = TcpListener::bind(listening_addr).await.unwrap();
+            let _ = ready_sender.send(());
+            let tcp_stream = TcpListenerStream::new(tcp_listener);
+            Server::builder()
```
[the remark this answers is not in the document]
timr-dev: Fixed in 2f3b3fb together with the sibling test — the server `JoinHandle` is now awaited via `block_on` after the shutdown signal so no background task is left pending at runtime drop.

— Reply posted by Copilot (AI assistant) on behalf of @timr-dev. [GS]

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
timr-dev: Great point — added to the README `Limits` note in 2f3b3fb. It now calls out that a changed `headers` value (e.g. a rotated `authorization`/tenant header) only takes effect on the next stream open/reconnect, and that existing long-lived streams keep the value they were opened with — contrasting it with the OTLP/gRPC per-request rebuild.

— Reply posted by Copilot (AI assistant) on behalf of @timr-dev. [GS]

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
timr-dev: Done in 2f3b3fb. `build_static_metadata` is now a method on `GrpcClientSettings` (reads `self.headers`); both production callers use `self.config.grpc.build_static_metadata()` and the tests construct a `GrpcClientSettings` value, so the free function is no longer part of the public surface.

— Reply posted by Copilot (AI assistant) on behalf of @timr-dev. [GS]

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
timr-dev: Addressed in 2f3b3fb. The doc comment now states de-duplication is `validate()`'s responsibility (it rejects case-insensitive duplicates, and both `validate_config` and `from_config` call it), and the builder now emits a `grpc.client.static_header_collision` debug log if an `insert` ever overwrites a case-collapsed key, so the otherwise non-deterministic survivor is visible for an unvalidated caller.

— Reply posted by Copilot (AI assistant) on behalf of @timr-dev. [GS]

ThomsonTan (approved):

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
```diff
@@ -625,6 +650,7 @@ fn spawn_stream_workers<T>(
     streams_per_signal: usize,
     pdata_metrics_tx: Sender<PDataMetricsUpdate>,
     shutdown_rx: tokio::sync::watch::Receiver<bool>,
+    static_metadata: Option<Arc<MetadataMap>>,
```
utpilla: We could use an `Rc` instead of `Arc`.

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
```diff
@@ -1434,35 +1481,76 @@ mod tests {
...
+            .expect("valid headers should be accepted");
+        assert_eq!(exporter.config.grpc.headers.len(), 2);
+        assert_eq!(
+            exporter
+                .config
+                .grpc
+                .headers
+                .get("authorization")
+                .map(String::as_str),
+            Some("Basic dXNlcjpwYXNz")
```
utpilla, line 1510: Add a similar assertion for the other header as well.

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
```diff
@@ -1434,35 +1481,76 @@ mod tests {
...
+        let controller_ctx = ControllerContext::new(telemetry_registry_handle);
+        let pipeline_ctx =
+            controller_ctx.pipeline_context_with("grp".into(), "pipeline".into(), 0, 1, 0);
+
+        // `from_config` validates the shared gRPC settings itself (defense in
+        // depth), so a reserved gRPC header is rejected at construction time and
+        // not only via the factory `validate_config` hook.
+        let json_config = json!({
+            "grpc_endpoint": "http://localhost:4317",
+            "headers": { "grpc-timeout": "1S" }
```
utpilla: The test name suggests that it would be checking for invalid headers (e.g. non-ASCII) but it's checking for another reserved header similar to `test_validate_config_rejects_reserved_headers`.

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
```diff
@@ -124,11 +130,17 @@ runtime metric sets may also be attached by the pipeline telemetry policy.
...
+  Arrow stream when it is opened (or re-opened after a reconnect). They are not
+  re-sent per `BatchArrowRecords`, so the per-message hot path stays
+  allocation-free; the one-time cost is a single metadata clone per stream
+  establishment. Header names/values are validated at config load via the shared
+  `GrpcClientSettings` (ASCII-only, no gRPC-reserved metadata).
+  - Because the metadata is bound when a stream opens, a changed `headers` value
+    (e.g. a rotated `authorization` or tenant header) only takes effect on the
+    next stream open/reconnect, unlike the OTLP/gRPC exporter, which rebuilds
+    metadata per unary request. Existing long-lived streams keep using the value
+    they were opened with until they are re-established.
```
utpilla: > a changed headers value ... only takes effect on the next stream open/reconnect

A reconnect won't pick up a new value: `static_metadata` is built once in `start()` and the reconnect loop re-clones that same template without re-reading config (the `Config` control arm is a no-op today). A changed value only takes effect on restart/reconfigure, not on reconnect. Reword it to say "restarted/reconfigured" instead of "stream open/reconnect".

```suggestion
  - Because the metadata template is built once at exporter startup, a changed
    `headers` value (e.g. a rotated `authorization` or tenant header) only takes
    effect when the exporter is restarted/reconfigured and not on stream
    reconnects, which reuse the template captured at startup. This differs from
    the OTLP/gRPC exporter, which rebuilds metadata per unary request.
```

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
```diff
@@ -240,7 +220,18 @@ impl OTAPExporter {
             }
         })?;
 
-        reject_unsupported_headers(&config)?;
+        // Defense-in-depth: validate the shared gRPC settings (including the
+        // ASCII/reserved/duplicate `headers` checks) here too, not only in the
+        // factory `validate_config` hook. This guarantees that any construction
+        // path — including direct/programmatic `from_config` callers — rejects
```
utpilla: It'd be good to remove em dashes and restrict it to ASCII in the doc comments. There are several other places within this file and `client_settings.rs` where an em dash was added.

utpilla: We should update our CI check to catch this.

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
timr-dev: Done in 50b15e93a. `static_metadata` is now `Option<Rc<MetadataMap>>`. The stream workers run on a thread-local set (`spawn_local`), so the worker future is `!Send` and the atomic refcount isn't needed. One shutdown test was driving `stream_arrow_batches` via `tokio::spawn` (which requires `Send`); I switched it to a `LocalSet` + `spawn_local` so it mirrors how the exporter actually runs. fmt/clippy/test all green.

--- Reply posted by Copilot (AI assistant) on behalf of @timr-dev. [GS]

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
timr-dev: Done in 50b15e93a -- `test_from_config_accepts_headers` now also asserts the second header (`x-scope-orgid` => `tenant-1`), not just `authorization`.

--- Reply posted by Copilot (AI assistant) on behalf of @timr-dev. [GS]

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
timr-dev: Good catch. Renamed to `test_from_config_rejects_invalid_header_value` and made it actually exercise an invalid value instead of duplicating `test_validate_config_rejects_reserved_headers`. One subtlety worth flagging: genuine high-byte non-ASCII values (e.g. `cafe` with an accented e) are accepted as obs-text by the http/tonic `HeaderValue` rules, so they are not rejected -- only non-visible bytes are. The test now uses a value containing a control character (a newline, written ASCII-safe as `bad\nvalue`) and asserts the `must be visible ASCII` error. Done in 50b15e93a.

--- Reply posted by Copilot (AI assistant) on behalf of @timr-dev. [GS]

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
timr-dev: Applied your suggestion in 50b15e93a -- the Limits note now says a changed `headers` value only takes effect when the exporter is restarted/reconfigured, not on stream reconnects (which reuse the template captured at startup). Thanks for catching that the reconnect path reuses the startup template rather than re-reading config.

--- Reply posted by Copilot (AI assistant) on behalf of @timr-dev. [GS]

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/mod.rs
timr-dev: Removed the em dashes from the PR-introduced doc comments in 50b15e93a (`mod.rs` and `client_settings.rs`); those files are ASCII-only now apart from two pre-existing `->` arrows in an unrelated TLS test (`client_settings.rs:1268/1286`, on `main` since 2025-12) which I left out of this PR's scope. On the CI suggestion: a repo-wide non-ASCII guard would also flag those pre-existing characters, so it's cleaner as its own change -- I filed #3320 to track adding the check (e.g. extending `tools/sanitycheck.py`) plus the one-time sweep.

--- Reply posted by Copilot (AI assistant) on behalf of @timr-dev. [GS]

rust/otap-dataflow/crates/core-nodes/src/exporters/otap_exporter/README.md
timr-dev: Heads-up: the wording added here was refined further in 50b15e93a following @utpilla's review -- it now says a changed value takes effect on restart/reconfigure rather than `stream open/reconnect`, since reconnects reuse the metadata template captured at startup. The substance (changed headers don't apply to existing long-lived streams) is unchanged.

--- Reply posted by Copilot (AI assistant) on behalf of @timr-dev. [GS]

utpilla (approved):

Issue #3272, opened 2026-06-11
Design and implement logical retained-work memory budgeting


lalitb:
## Description

Track the design and implementation of retained-work memory budgeting in the dataflow engine.

The current process-wide memory limiter remains the outer safety guard. This effort adds a more local logical memory-budgeting model for memory retained inside the engine, such as data or state held across queues, topics, retry buffers, batchers, processor state, pending exporter requests, delayed work, and similar retention points. Retained-work ownership should move with the data as it crosses queues, topics, and runtime boundaries.

The model should treat memory as a property of retained work, not primarily as allocator, thread, or runtime ownership. Allocator statistics may still be used for diagnostics and validation.

Future tenant or pipeline-group memory limits should be kept in mind by allowing retained-work ownership to carry attribution such as pipeline group, pipeline, node, retention site, or future tenant/request context. This issue does not define the full tenant policy model.

## Tracking Items

  Items have implicit dependencies; the design document should define ordering, invariants, APIs, and rollout sequencing.

  - [x] Write a retained-work memory-budgeting design document. See #3316.
  - [x] Define retained-work size estimation contracts and helper APIs. See #3443.
  - [ ] Add generic node-local delayed-work identity. See #3747.
  - [ ] Add runtime-local ownership and exactly-once accounting primitives. See #3756.
  - [ ] Add bounded attribution for pipeline group, pipeline, runtime generation, component, owner, and retention site. See #3815.
  - [ ] Install retained-work accounting in runtime and component scopes with fail-loud wiring.
  - [ ] Export metrics for retained bytes, unknown-size items, abandonment, accounting corruption, and instrumentation coverage.
  - [ ] Add the first observe-only charge site for node-local retry/delayed work.
  - [ ] Add observe-only accounting for pending batch data.
  - [ ] Validate accounting correctness, bounded cardinality, generation cleanup, and enabled/disabled overhead.
  - [ ] Expand accounting coverage to queues, topics, processor state, pending exporter requests, and other retained-work sites.
  - [ ] Add escrow-style ownership for shared channels, topic broker paths, fanout, failed sends, and cross-runtime boundaries.
  - [ ] Add reclaim and drain hooks for stateful processors and queues.
  - [ ] Add operator-facing diagnostic views by runtime, pipeline group, pipeline, node, owner, and retention site.
  - [ ] Design admission and backpressure behavior after accounting coverage and reclaim paths are sufficient.

  ## Gates / Constraints

[...]

PR #3316, opened 2026-06-18
doc: introduce observe-only retained-work memory accounting/budgeting

lalitb:
# Change Summary

  Adds a first-level observe-only design document for retained-work memory accounting in OTAP
  dataflow. The document explains where telemetry can be retained, how local and shared ownership
  should be modeled, and why observe-only accounting is the foundation for later budgeting,
  enforcement, and tenant/group isolation.

  * Relates to #3272

  * `npx markdownlint-cli2 rust/otap-dataflow/docs/observe-only-retained-work-budgeting.md`
  * `python3 tools/sanitycheck.py`

  No runtime or configuration behavior changes. This is documentation only.

  ### Changelog

  * [x] This PR is a docs-only `chore`; no `.chloggen/*.yaml` entry required.


jmacd (commented):
This looks good --

There is a limited form of this accounting/budgeting for memory in the OTel-Arrow Phase-1 otelarrowreceiver's admission controller https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/otelarrowreceiver/README.md#admission-control-configuration, however the Collector's pipeline assumes synchronous behavior, so we can acquire the resource, call the pipeline, and release the resource. 

I like this document, and I think it would be appropriate to sketch the major structs and interfaces for the resource ticket, how it crosses local/shared boundaries, what the envelope looks like, and so on. Can `OtapPdata` be the envelope type? Everywhere we take apart the `OtapPdata` I worry about loss of context, and these are the same sites where we need to be careful about budget accounting.

lalitb:
>  like this document, and I think it would be appropriate to sketch the major structs and interfaces for the resource ticket, how it crosses local/shared boundaries, what the envelope looks like, and so on. Can OtapPdata be the envelope type? Everywhere we take apart the OtapPdata I worry about loss of context, and these are the same sites where we need to be careful about budget accounting.

Yes, that makes sense. I intentionally kept this PR as the simpler observe-only intro instead of bringing in the full Phase 2 design doc, but I agree it should still include a small sketch of the ownership primitives.

I’ll add a short section that names the major shapes: local ticket ownership for runtime-local retained work, sendable escrow ownership for shared boundaries, and how `OtapPdata` can carry shared-boundary ownership without carrying a `!Send` local ticket. I’ll keep the detailed enforcement/lease/reclaim material out of this doc so it stays readable.

lalitb:
> I’ll add a short section that names the major shapes: local ticket ownership for runtime-local retained work, sendable escrow ownership for shared boundaries, and how `OtapPdata` can carry shared-boundary ownership without carrying a `!Send` local ticket. I’ll keep the detailed enforcement/lease/reclaim material out of this doc so it stays readable.

Updated in 36994cfbe.

I added a focused “Concrete ownership sketch” section with the proposed local/shared ownership shapes. It covers
`LocalMemoryTicket`, `EscrowTicket`, `EscrowSlot`, local/shared envelopes, and the `OtapPdata` boundary.

The doc now says `OtapPdata` can be part of the sendable shared-boundary story, but local `!Send` tickets should stay outside `PData` and move beside retained payloads through an envelope or equivalent owner.

rust/otap-dataflow/.chloggen/retained-work-budgeting-design.yaml
```diff
@@ -0,0 +1,4 @@
+change_type: enhancement
+component: engine
+note: Document observe-only retained-work memory accounting and ownership shapes.
```
drewrelmas: Per #3360, chloggen entry is no longer needed for docs PRs, recommend removing

lalitb:
moving it to draft for now, to add some more details and clarify.

jmacd (approved):

lalitb:
> moving it to draft for now, to add some more details and clarify.

un-draft now. good to review/merge.

PR #3321, opened 2026-06-19
fix: correct retry behaviour for OTLP gRPC exporter/receiver

### LINKED ISSUE #1920: OTLP gRPC/HTTP propagate status code, set permanent status

albertlockett:
# Change Summary

Prior to this fix, the OTLP gRPC exporter treated every error as retryable by indiscriminantly creating Nack control messages with `permanent = false`. Similarly, the OTLP receiver would always send an error code that, according to the OTLP spec, is not retryable even when it receives a non-permanent Nack. This PR corrects the issue such that:

a) OTLP gRPC exporter will correctly set the Nack flag based on the retryability of the error code specified in the spec: https://opentelemetry.io/docs/specs/otlp/#failures
b) the receiver will also use a retryable error code for the non-permanent Nack message.

* Closes #1920

unit

 no

### Changelog


rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
```diff
@@ -418,13 +419,11 @@ async fn route_export_result<T>(
...
             let error_msg = e.to_string();
-            effect_handler
-                .notify_nack(NackMsg::new(
-                    &error_msg,
-                    OtapPdata::new(context, saved_payload),
-                ))
-                .await?;
+            let mut nack = NackMsg::new(&error_msg, OtapPdata::new(context, saved_payload));
+            nack.permanent = !retryable;
+            effect_handler.notify_nack(nack).await?;
```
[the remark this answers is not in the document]
albertlockett: meh, permanent is pretty clearly all that needs to be set for this to be permanent. Also, the flag is public for precisely this use case.

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
```diff
@@ -463,6 +435,80 @@ async fn route_export_result<T>(
...
+        | Code::Aborted
+        | Code::OutOfRange
+        | Code::Unavailable
+        | Code::DataLoss => true,
+
+        // RESOURCE_EXHAUSTED is retryable only when the server signals recovery
+        // is possible by including a RetryInfo detail.
+        Code::ResourceExhausted => has_retry_info(status),
+
+        // All other codes (UNKNOWN, INVALID_ARGUMENT, NOT_FOUND,
```
lalitb: One edge case:  the generated OTLP gRPC client can map temporary client-side readiness failures to `UNKNOWN` before the RPC is even sent. Eg:

https://github.com/open-telemetry/otel-arrow/blob/66bf0c4d35d84dd0184ed884b6b3524fcecf3fe0/rust/otap-dataflow/crates/pdata/src/proto/opentelemetry.proto.collector.logs.v1.rs#L126-L133

With this change, that path becomes a permanent NACK and upstream may drop the payload, even though the channel may only be reconnecting. Should we special-case these client-side readiness/transport errors as retryable, similar to how the HTTP exporter treats connect/timeout errors :

  https://github.com/open-telemetry/otel-arrow/blob/66bf0c4d35d84dd0184ed884b6b3524fcecf3fe0/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/mod.rs#L546-L558

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
```diff
@@ -463,6 +435,80 @@ async fn route_export_result<T>(
...
+        Code::Cancelled
+        | Code::DeadlineExceeded
+        | Code::Aborted
+        | Code::OutOfRange
+        | Code::Unavailable
+        | Code::DataLoss => true,
+
+        // RESOURCE_EXHAUSTED is retryable only when the server signals recovery
+        // is possible by including a RetryInfo detail.
+        Code::ResourceExhausted => has_retry_info(status),
```
lalitb: One compatibility concern probably for chained agents:
 
Our OTLP gRPC receiver may return `RESOURCE_EXHAUSTED` for temporary overload. However, this exporter now treats `RESOURCE_EXHAUSTED` without `RetryInfo` as a permanent error.
 
In an `Agent A -> Agent B` setup, if Agent B applies backpressure, Agent A could interpret that as a permanent failure and drop the batch instead of retrying.

Should we align the receiver behavior as well, so temporary backpressure is not exposed as a permanent failure?

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
jmacd: I would just replace the unknown case here with "unavailable" to address this.

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_grpc_exporter/mod.rs
jmacd: Good catch -- I would say ResourceExhausted is always retryable. The RetryAfter annotation is advisory.

jmacd:
@albertlockett please take a look I added a commit :)

albertlockett:
> @albertlockett please take a look I added a commit :)

Thanks @jmacd ! I approve the changes from your commit which address @lalitb 's feedback

jmacd (approved):

PR #3322, opened 2026-06-19
feat(otap-exporter): redact static exporter header values (#3306)

### LINKED ISSUE #3306: OTLP exporters: wrap static exporter header values in a privacy/opaque type to prevent secret leakage into telemetry

timr-dev:
# Change Summary

Wrap `HttpClientSettings::headers` and `GrpcClientSettings::headers` **values** in `secrecy::SecretString` so a header credential (an `Authorization` token, a tenant API key) can no longer leak through the `Debug`/telemetry output of the settings structs. Cleartext is read only via an explicit `expose_secret()` accessor at the request-construction sites (`validate`, `build_static_metadata`, and `HttpClientPool::build_static_headers`); behavior is otherwise unchanged.

**Defense-in-depth:** the constructed wire values are additionally marked sensitive (`HeaderValue::set_sensitive` / `MetadataValue::set_sensitive`), so the credential is redacted in any `HeaderMap`/`MetadataMap` `Debug` output and excluded from HTTP/2 HPACK indexing.

`user_agent` is intentionally left a plain `String` — it is a non-credential client identifier — mirroring the Go Collector's use of `configopaque` for auth/credential headers only.

**Why `secrecy` over the maintainer-suggested Microsoft Oxidizer `data_privacy`:** `secrecy` (and its `zeroize` transitive dependency) was already in the build graph, so this adds **zero net-new crates**. `data_privacy` would have pulled in four pre-1.0 crates plus taxonomy / classified-container / redaction-engine boilerplate for what is, here, just type-level redaction. Runtime cost is a wash — header values are materialized once at exporter startup, never on the per-export hot path — and `secrecy` adds free zeroize-on-drop.

## Scope / known limitation

This protects the **typed** settings (`Debug`/telemetry) and the values placed on the wire. It does **not** redact the raw configuration snapshot returned by the admin config APIs: `NodeUserConfig.config` retains the original `serde_json::Value`, which the admin endpoints (`GET /api/v1/config` and the per-pipeline detail API) serialize as-is, so a credential header still appears in cleartext there. Redacting the raw config snapshot is broader than this exporter-scoped change (it touches a shared, type-erased config value across all node kinds and the admin API contract), so it is tracked separately in #3328. The changelog wording is scoped accordingly.

* Closes #3306

* `cargo xtask check` (structure, fmt, clippy `-D warnings`, full test suite) passes.
* New unit tests assert the secret value is **absent** from each settings struct's `Debug` output while remaining reachable via `expose_secret()`, for both `HttpClientSettings` and `GrpcClientSettings`.
* `build_static_metadata_builds_map` asserts the constructed gRPC `MetadataValue` reports `is_sensitive() == true`; the symmetric `build_static_headers_marks_values_sensitive` asserts the same for every constructed HTTP `HeaderValue`.
* Existing exporter tests were updated for the `SecretString` field type and continue to pass.

Yes. The `headers` values on `HttpClientSettings` and `GrpcClientSettings` change type from `String` to `secrecy::SecretString`. End-user configuration is unaffected — values still deserialize from plain strings — but they no longer appear in `Debug`/serialized output, and any in-process code reading these fields must now go through `expose_secret()`. Wire behavior is unchanged.

### Changelog

Entry: `rust/otap-dataflow/.chloggen/otlp-exporter-opaque-header-values.yaml` (`change_type: enhancement`, `component: pipeline`).

### DIFF
--- a/rust/otap-dataflow/.chloggen/otlp-exporter-opaque-header-values.yaml
+++ b/rust/otap-dataflow/.chloggen/otlp-exporter-opaque-header-values.yaml
@@ -9,12 +9,13 @@ component: pipeline
 
 # A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
 note: >-
-  Static exporter `headers` values are now wrapped in `secrecy::SecretString`,
-  so a header credential (e.g. an `Authorization` token or tenant API key) can
-  no longer leak through `Debug`/telemetry output of `HttpClientSettings` or
-  `GrpcClientSettings`. The cleartext is only read via an explicit
-  `expose_secret()` accessor at the request-construction site; behavior is
-  otherwise unchanged.
+  Static exporter `headers` values are now wrapped in `secrecy::SecretString`
+  and marked sensitive on the wire, so a header credential (e.g. an
+  `Authorization` token or tenant API key) is redacted in the `Debug`/telemetry
+  output of the typed `HttpClientSettings` / `GrpcClientSettings` and excluded
+  from HTTP/2 HPACK indexing. The cleartext is read only via an explicit
+  `expose_secret()` accessor (during validation and when building the request);
+  behavior is otherwise unchanged.
 
 # Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
 issues: [3306]
@@ -23,11 +24,8 @@ issues: [3306]
 # These lines will be padded with 2 spaces and then inserted directly into the document.
 # Use pipe (|) for multiline entries.
 subtext: >-
-  `user_agent` is intentionally left as a plain `String`: it is a
-  non-credential client identifier, mirroring the Go Collector's use of
-  `configopaque` for auth/credential headers rather than the User-Agent.
-  `secrecy` (and its `zeroize` transitive dependency) was already present in
-  the build graph, so this adds no new crate; it was preferred over the
-  Microsoft Oxidizer `data_privacy` crate, which would have pulled in four
-  pre-1.0 crates plus taxonomy/redaction-engine boilerplate for a need that is
-  just type-level redaction.
+  Scope: this protects the typed settings and the values placed on the wire. It
+  does not redact the raw configuration snapshot returned by the admin config
+  endpoints (e.g. `GET /api/v1/config`), which still serialize the original
+  config value; that broader redaction is tracked in #3328. `user_agent` is
+  intentionally left a plain `String` as a non-credential client identifier.
--- a/rust/otap-dataflow/.chloggen/otlp-exporter-opaque-header-values.yaml
+++ b/rust/otap-dataflow/.chloggen/otlp-exporter-opaque-header-values.yaml
@@ -1,0 +1,33 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: pipeline
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: >-
+  Static exporter `headers` values are now wrapped in `secrecy::SecretString`,
+  so a header credential (e.g. an `Authorization` token or tenant API key) can
+  no longer leak through `Debug`/telemetry output of `HttpClientSettings` or
+  `GrpcClientSettings`. The cleartext is only read via an explicit
+  `expose_secret()` accessor at the request-construction site; behavior is
+  otherwise unchanged.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [3306]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: >-
+  `user_agent` is intentionally left as a plain `String`: it is a
+  non-credential client identifier, mirroring the Go Collector's use of
+  `configopaque` for auth/credential headers rather than the User-Agent.
+  `secrecy` (and its `zeroize` transitive dependency) was already present in
+  the build graph, so this adds no new crate; it was preferred over the
+  Microsoft Oxidizer `data_privacy` crate, which would have pulled in four
+  pre-1.0 crates plus taxonomy/redaction-engine boilerplate for a need that is
+  just type-level redaction.
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/mod.rs
@@ -53,6 +53,7 @@ use otap_df_telemetry::metrics::MetricSet;
 use otap_df_telemetry::{otel_debug, otel_info, otel_warn};
 use prost::Message as _;
 use reqwest::{Client, Response};
+use secrecy::ExposeSecret;
 
 use self::config::Config;
 use crate::exporters::otlp_grpc_exporter::InFlightExports;
@@ -749,9 +750,12 @@ impl HttpClientPool {
             let header_name = http::HeaderName::from_bytes(name.as_bytes()).map_err(|e| {
                 HttpClientError::InvalidConfig(format!("invalid header name \"{name}\": {e}"))
             })?;
-            let header_value = HeaderValue::from_str(value).map_err(|e| {
+            let mut header_value = HeaderValue::from_str(value.expose_secret()).map_err(|e| {
                 HttpClientError::InvalidConfig(format!("invalid value for header \"{name}\": {e}"))
             })?;
+            // Mark the value sensitive so it is redacted in any `HeaderMap`
+            // `Debug` output and excluded from HTTP/2 HPACK indexing.
+            header_value.set_sensitive(true);
             _ = default_headers.insert(header_name, header_value);
         }
 
@@ -1094,11 +1098,8 @@ mod test {
         wait_for_port_ready(&endpoint_addr);
 
         let mut headers = HashMap::new();
-        _ = headers.insert(
-            "authorization".to_string(),
-            "Basic dXNlcjpwYXNz".to_string(),
-        );
-        _ = headers.insert("x-test".to_string(), "abc".to_string());
+        _ = headers.insert("authorization".to_string(), "Basic dXNlcjpwYXNz".into());
+        _ = headers.insert("x-test".to_string(), "abc".into());
         let settings = HttpClientSettings {
             headers,
             ..Default::default()
--- a/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/mod.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/mod.rs
@@ -735,17 +735,25 @@ struct HttpClientPool {
 }
 
 impl HttpClientPool {
-    async fn try_new(
+    /// Builds the user-configured static request headers, marking every value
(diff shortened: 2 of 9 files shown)


rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
```diff
@@ -76,12 +77,17 @@ pub struct HttpClientSettings {
     /// Static headers added to every outbound OTLP/HTTP request (e.g. an
     /// `Authorization` or backend routing header).
     ///
+    /// Values are wrapped in [`SecretString`] so a header credential is not
+    /// accidentally leaked through `Debug`/telemetry; the cleartext is only
+    /// reachable via [`ExposeSecret::expose_secret`] at the request-construction
+    /// site.
```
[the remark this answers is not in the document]
timr-dev: Good catch — `HttpClientSettings::validate()` does call `expose_secret()` too. Updated the doc comment in a7256d78c to list both access points (validation and request construction).

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -163,12 +164,17 @@ pub struct GrpcClientSettings {
     /// [`GrpcClientSettings::validate`]. These coexist with any header
     /// propagation policy configured on the exporter.
     ///
+    /// Values are wrapped in [`SecretString`] so a metadata credential is not
+    /// accidentally leaked through `Debug`/telemetry; the cleartext is only
+    /// reachable via [`ExposeSecret::expose_secret`] at the
+    /// metadata-construction site.
```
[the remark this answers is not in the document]
timr-dev: Good catch — `GrpcClientSettings::validate()` also calls `expose_secret()`. Updated the doc comment in a7256d78c to note both validation and metadata construction.

rust/otap-dataflow/.chloggen/otlp-exporter-opaque-header-values.yaml
```diff
@@ -0,0 +1,33 @@
...
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: >-
+  `user_agent` is intentionally left as a plain `String`: it is a
+  non-credential client identifier, mirroring the Go Collector's use of
+  `configopaque` for auth/credential headers rather than the User-Agent.
+  `secrecy` (and its `zeroize` transitive dependency) was already present in
+  the build graph, so this adds no new crate; it was preferred over the
+  Microsoft Oxidizer `data_privacy` crate, which would have pulled in four
```
cijothomas: good content for PR desc, not really for changelog.

cijothomas (approved):
Looks good.

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
```diff
@@ -76,12 +77,17 @@ pub struct HttpClientSettings {
...
+    /// only through an explicit [`ExposeSecret::expose_secret`] call — during
+    /// [`HttpClientSettings::validate`] and at the request-construction site.
+    ///
     /// Protocol headers (`Content-Type` / `Content-Encoding` /
     /// `Content-Length` / `Host`) and response-negotiation headers
     /// (`Accept` / `Accept-Encoding`) cannot be set here and are rejected by
     /// [`HttpClientSettings::validate`]; they are managed by the exporter.
     #[serde(default)]
-    pub headers: HashMap<String, String>,
+    pub headers: HashMap<String, SecretString>,
```
lalitb, line 90: This protects the typed `HttpClientSettings` value, but I think the original config can still expose the secret. `NodeUserConfig.config` keeps the raw `serde_json::Value`, and the admin config endpoint returns the controller-owned config snapshot.

So if a user config has `headers.authorization`, won’t that value still show up through config/admin APIs? Should we redact it there too, or narrow the PR/changelog wording to say this only protects typed settings/debug output?

rust/otap-dataflow/crates/core-nodes/src/exporters/otlp_http_exporter/mod.rs
```diff
@@ -749,9 +750,12 @@ impl HttpClientPool {
             let header_name = http::HeaderName::from_bytes(name.as_bytes()).map_err(|e| {
                 HttpClientError::InvalidConfig(format!("invalid header name \"{name}\": {e}"))
             })?;
-            let header_value = HeaderValue::from_str(value).map_err(|e| {
+            let mut header_value = HeaderValue::from_str(value.expose_secret()).map_err(|e| {
                 HttpClientError::InvalidConfig(format!("invalid value for header \"{name}\": {e}"))
             })?;
+            // Mark the value sensitive so it is redacted in any `HeaderMap`
+            // `Debug` output and excluded from HTTP/2 HPACK indexing.
+            header_value.set_sensitive(true);
```
ThomsonTan, line 766: **Non-blocking (test coverage):** this `set_sensitive(true)` has no test coverage, whereas the symmetric gRPC marking *is* asserted in `build_static_metadata_builds_map`:

https://github.com/open-telemetry/otel-arrow/pull/3322/changes#diff-3b19757ea1f4d9eae7f548d7035c79c6074650826af4e43c985c899e424ef6c1R900

timr-dev: Good call — added `build_static_headers_marks_values_sensitive` in 43223606d, which asserts `is_sensitive()` on every constructed HTTP `HeaderValue`, mirroring the gRPC `build_static_metadata_builds_map`. To make it unit-testable I pulled the static-header construction out of `try_new` into a `build_static_headers` helper — no behavior change (`try_new` still pre-sizes for the two protocol headers and they still win on collision).

rust/otap-dataflow/.chloggen/otlp-exporter-opaque-header-values.yaml
timr-dev: Agreed — trimmed the changelog to a concise note plus a short scope caveat in 43223606d, and moved the `secrecy` vs `data_privacy` dependency rationale to the PR description where it fits better.

rust/otap-dataflow/crates/otap/src/otlp_http/client_settings.rs
timr-dev: You're right, thanks. The `SecretString` change only covers the typed settings and the on-the-wire (HPACK) marking; the raw `NodeUserConfig.config` (`serde_json::Value`) is still serialized as-is by the admin config endpoints (`GET /api/v1/config`, pipeline details), so a credential header is still exposed there. In 43223606d I narrowed the PR/changelog wording to scope the guarantee honestly, and filed #3328 to track redacting the raw config snapshot — a broader change (a shared, type-erased config value across all node kinds plus the admin API contract) than this exporter-scoped PR.

ThomsonTan (approved):

lalitb (approved):

rust/otap-dataflow/crates/otap/src/otap_grpc/client_settings.rs
```diff
@@ -163,12 +164,17 @@ pub struct GrpcClientSettings {
...
+    /// only through an explicit [`ExposeSecret::expose_secret`] call — during
+    /// [`GrpcClientSettings::validate`] and at the metadata-construction site.
+    ///
     /// `GrpcClientSettings` is shared by the OTLP/gRPC exporter and the OTAP
     /// (Arrow) exporter. The OTLP/gRPC exporter applies them as per-request
     /// metadata; the OTAP exporter applies them once as the initial metadata of
     /// each Arrow stream (see [`build_static_metadata`]).
     #[serde(default)]
-    pub headers: HashMap<String, String>,
+    pub headers: HashMap<String, SecretString>,
```
ThomsonTan, line 177: We already have an in-house redaction type `OpaqueString`, is it possible to re-use that instead of introducing `SecretString`?

timr-dev: Good question — I did weigh `OpaqueString` first. They overlap but have different contracts, and for credential header values the difference matters.

`OpaqueString` redacts `Debug`/`Display`, but it **serializes the cleartext** (`#[serde(transparent)]` — its own test asserts `serialize → "super-secret"`) and hands out the cleartext implicitly via `Deref<str>`, with no zeroize.

`secrecy::SecretString`:
- doesn't implement `Serialize` at all — a compile-time guarantee the value can't be folded into a serialized snapshot / log / telemetry;
- exposes cleartext only through an explicit, greppable `expose_secret()`;
- zeroizes on drop.

That serialization difference is the crux here. The #3328 follow-up exists precisely because node configs *do* get serialized (the admin `GET /api/v1/config` snapshot) — `OpaqueString`'s serialize-cleartext behavior is exactly that leak vector, whereas `SecretString` won't compile if someone later adds `Serialize` to the settings rather than silently leaking. The explicit `expose_secret()` also keeps every cleartext read auditable, which is the access model this PR documents. And `secrecy`/`zeroize` are already in the dependency graph (they're in `main`'s `Cargo.lock` independent of this PR), so it's zero net-new crates.

So my preference is to keep `SecretString` for the credential values. That said, if you'd rather standardize on a single in-house type, I'm happy to instead harden `OpaqueString` to credential-grade — skip/redact on serialize, replace the `Deref` with an explicit `expose()` accessor, add zeroize — and use that here; it has no other call sites today, so tightening its contract is low-risk. Let me know which you'd prefer.

ThomsonTan: Thanks for the detailed write-up. It's a good point that `OpaqueString` serializes the cleartext while `SecretString` doesn't implement `Serialize` at all. For credential values that fail-closed behavior is the right call, so let's keep `SecretString`.

timr-dev: Thanks, @ThomsonTan — appreciate you weighing it through. We'll keep `secrecy::SecretString` for the credential header values then; the fail-closed (no `Serialize`) behavior is the deciding factor, and there's no code change needed since that's what's already in the PR. Resolving this thread.

jmacd (approved):

PR #3345, opened 2026-06-23
chore(ci): commit otap-dataflow Cargo.lock and build with --locked

### LINKED ISSUE #3283: Commit Cargo.lock for rust/otap-dataflow workspace
The `rust/otap-dataflow` workspace produces binaries (e.g. `df_engine`) but does not check in a `Cargo.lock`. Every CI run re-resolves dependencies, so an upstream release can break `main` with no source change on our side — as happened today with `time` 0.3.48 (worked around in #3282, mirroring open-telemetry/opentelemetry-rust-contrib#648).

Proposal:
- Commit `rust/otap-dataflow/Cargo.lock`.
- Run CI with `--locked` so dep updates show up as explicit diffs in PRs.
- Let Renovate/Dependabot drive intentional bumps.

AvinashDevX:
# Change Summary

otap-dataflow builds binaries (df_engine) but didn't commit a Cargo.lock, so CI re-resolved dependencies on every run. That is how the time 0.3.48 release broke main with no change on our side. This commits the lockfile and runs the otap-dataflow-only jobs with --locked, so dependency changes land as explicit Cargo.lock diffs instead of silently shifting a build.

Scope is the minimal lockfile + --locked discussed on the issue. resolver = "3" and an MSRV-pinned 1.87 CI build are left as follow-ups.

--locked is added only where otap-dataflow is built on its own: rust-ci.yml, rust-bench.yml (bench-contrib), rust-validation-tests.yml, df-engine-internal-observability.yml, and the Docker path (cross-arch-build.sh, plus removing Cargo.lock from .dockerignore so the lock reaches the build context). 

Matrix jobs shared with experimental/query_engine are left alone, since that workspace has no lockfile yet and --locked would fail there.

* Closes #3283

* `cargo build --locked --bin df_engine` builds against the committed lock.
* `cargo metadata --locked` and `cargo update --locked --dry-run` both report 0 packages to change, so the lock is already consistent and --locked will not trip on a stale lock in CI.
* CI here runs every job the flag was added to, including the Docker build via docker-build-check.

No. CI and build tooling only.

### Changelog


drewrelmas:
I wonder if we need to update the repo's Renovate configuration for Cargo to properly handle lock file maintenance ([reference](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance))? We may not be able to know for sure until we merge this and subsequent Renovate runs either succeed or fail :smile:

I recall we had to do some special configuration for Python lock files:

https://github.com/open-telemetry/otel-arrow/blob/2785c4fdd9b4bb83274b3095948901a21abd1aeb/.github/renovate.json5#L152-L156

AvinashDevX:
> I wonder if we need to update the repo's Renovate configuration for Cargo to properly handle lock file maintenance ([reference](https://docs.renovatebot.com/configuration-options/#lockfilemaintenance))? We may not be able to know for sure until we merge this and subsequent Renovate runs either succeed or fail 😄
> 
> I recall we had to do some special configuration for Python lock files:
> 
> https://github.com/open-telemetry/otel-arrow/blob/2785c4fdd9b4bb83274b3095948901a21abd1aeb/.github/renovate.json5#L152-L156

Agreed, good catch. Right fix is scoping lockFileMaintenance per-manager so cargo and pip-compile refresh on separate branches instead of colliding. Cargo likely won't need the transitive-dep handling we did for pip. Its manager runs cargo update and manages the lock as a unit, so it's mostly branding/scoping. Can't fully validate until it's live though, so I'll land this as-is and own the Renovate follow-up to watch the first run.

drewrelmas (approved):

drewrelmas:
Merging this on Friday as we will likely get some Rust dependency upgrades at the beginning of next week. We can triage Renovate config changes required as they come.

AvinashDevX:
Thanks! I'll watch the first Renovate runs next week and open the config follow-up based on what we see.

AvinashDevX:
Update from watching the first Renovate Rust runs now that the lockfile is in.

The lockfile plus --locked is working with Renovate as intended:

- Compatible bumps pass clean: #3381 (opentelemetry-proto) is all green.
- Incompatible bumps now fail as explicit PRs instead of silently breaking main: #3377 (weaver) on a real weaver_resolver::SchemaResolver import break, and #3380 (object_store) because 0.14.0 dropped the cloud feature we use.

That is exactly the breakage this was meant to catch, so the core goal is holding up.

No cargo lockFileMaintenance PR has fired yet, so the pip-compile branding collision hasn't surfaced. I'll keep the per-manager scoping fix parked until a maintenance run actually lands on the wrong branch, so the follow-up references a real run rather than a guess.

PR #3350, opened 2026-06-24
feat(query-engine): support nested serialized attribute assignment

### LINKED ISSUE #3343: Support nested mutation of serialized Map/Slice attributes in transform processor
### Pre-filing checklist

### Component(s)

Yes. The transform processor can update top-level attributes, but it cannot update a nested field inside a Map/Slice attribute.

For example, users may want to redact one nested value:

```
set(attributes["http"].request.headers.authorization, "redacted")
```

Today this requires replacing the whole top-level attribute instead of updating only the nested field.

### Proposed Solution

Allow the transform processor to update nested fields inside Map/Slice attribute values.

**Example:**

```
  set(attributes["http"].request.headers.authorization, "redacted")
  set(attributes["gen_ai.request"].messages[0].content, "redacted")
```
 
**Expected behavior:**

  - Update nested fields inside existing Map/Slice attributes.
  - Keep existing scalar attribute behavior unchanged.
  - Treat bytes attributes as scalar values, not nested containers.

### Alternatives Considered

Users can replace the whole top-level Map/Slice attribute manually, but that is awkward and error-prone when only one nested field needs to change.

 A separate processor could also handle this, but this is part of transform semantics and fits better in the existing transform processor.

### Additional Context

Complex attribute values such as Map and Slice are stored in the serialized attribute value column. The transform processor already supports top-level attribute mutation, but nested mutation is currently missing for these complex values.

lalitb:
# Change Summary

* Closes #3343

### Changelog

* [ ] Added a `.chloggen/*.yaml` entry, OR this PR is a `chore` (indicated in title).


rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -1322,6 +1487,196 @@ impl PipelineStage for AssignPipelineStage {
...
+        .downcast_ref::<UInt8Array>()
+        .ok_or_else(|| Error::ExecutionError {
+            cause: "attribute type column must be UInt8".into(),
+        })?;
+
+    let ser_col = attrs_record_batch.column(ser_col_index);
+    let ser_values = match ser_col.data_type() {
+        DataType::Dictionary(_, _) => cast(ser_col, &DataType::Binary)?,
+        _ => Arc::clone(ser_col),
+    };
```
albertlockett: Here we expand the dictionary column to a Binary Array, and then we access the bytes for each row calling `serialized_value_bytes` below on 1533. 

I think we could maybe avoid the cast by simply having `serialized_value_bytes` support returning bytes from the dictionary array.

We also a `ByteArrayAccessor`  type that abstracts away the dictionary encoding which may be helpful to use here instead. Currently it is pub crate, but I'd be OK if we just made it `pub`:
https://github.com/open-telemetry/otel-arrow/blob/bdd514ccf95d7fbac61746c699c0ebc295eed900/rust/otap-dataflow/crates/pdata/src/arrays.rs#L335-L349

Then you could simply do:
```rs
let ser_values = ByteArrayAccessor::try_new(ser_col)?;


// and below:
let original = ser_values.slice_at(row);
```

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -1322,6 +1487,196 @@ impl PipelineStage for AssignPipelineStage {
...
+    let mut ser_builder = BinaryBuilder::with_capacity(ser_col.len(), 0);
```
albertlockett: Could we use an `otap_df_pdata::encode::record::array::BinaryArrayBuilder` here instead?

`BinaryArrayBuilder` will dynamically choose to produce an array that has dictionary encoding if the cardinality of the array would not overflow the dictionary.

We can initialize it with the same options that are passed here:
https://github.com/open-telemetry/otel-arrow/blob/bdd514ccf95d7fbac61746c699c0ebc295eed900/rust/otap-dataflow/crates/pdata/src/encode/record/attributes.rs#L245-L249

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -5401,6 +5805,443 @@ mod test {
         test_upsert_attribute_scalar::<KqlParser>().await
     }
```
albertlockett, line 5786: Could we also add a test case where we assign to a resource or scope attribute? 

```
logs | set resource.attributes["complex"].child.name = "after"
```

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
lalitb: Thanks @albertlockett , I updated this to use `ByteArrayAccessor` and `BinaryArrayBuilder` in https://github.com/open-telemetry/otel-arrow/pull/3350/commits/2ca05f940af6b54a06602056d9fb1b92c827babf

One small question: to use `BinaryArrayBuilder` from query-engine, I made `encode::record::array` public, but that exposes more of the adaptive builder module than this PR really needs. Would you prefer keeping the module internal and adding a narrower public re-export/facade for just `BinaryArrayBuilder`, `ArrayOptions`, append traits, and `DictionaryOptions`?

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
```diff
@@ -1322,6 +1487,196 @@ impl PipelineStage for AssignPipelineStage {
...
+            cause: "attribute type column must be UInt8".into(),
+        })?;
+
+    let ser_col = attrs_record_batch.column(ser_col_index);
+    let ser_values = match ser_col.data_type() {
+        DataType::Dictionary(_, _) => cast(ser_col, &DataType::Binary)?,
+        _ => Arc::clone(ser_col),
+    };
+    let mut ser_builder = BinaryBuilder::with_capacity(ser_col.len(), 0);
+    let mut mutation_indices = vec![0; updates.len()];
```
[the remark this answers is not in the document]
lalitb: No longer valid issue - BinaryArrayBuilder is used now.

rust/otap-dataflow/crates/query-engine/src/pipeline/assign.rs
albertlockett: > I made encode::record::array public

This is fine! I actually think it should be public so we can use the accessors in precisely situations like this

albertlockett (approved):

Issue #2252, opened 2026-03-10
Support explicit broadcast Ack/Nack modes for topic hops


lquerel:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

## Summary

Topic Ack/Nack propagation works today for broadcast subscriptions, but the semantics are not very useful: the first broadcast subscriber that Ack/Nacks a message resolves the upstream message, and later Ack/Nack events become orphaned.

This issue proposes an explicit broadcast Ack/Nack policy so broadcast topics can support both the current cheap behavior and a stricter "all subscribers must Ack" mode.

## Problem

Current behavior is effectively:
- ack_propagation: auto enables end-to-end Ack/Nack bridging
- balanced subscriptions have a coherent single-consumer completion model
- broadcast subscriptions do not: the first downstream Ack/Nack wins

That means broadcast completion is ambiguous:
- one subscriber can Ack while another later Nacks
- one subscriber can Ack while another lags out
- publishers cannot distinguish best-effort fan-out from strict fan-out

This is especially problematic for users who want broadcast consumers to act as required validation/audit/replication paths rather than optional taps.

## Proposal

Add an explicit broadcast Ack mode under topic broadcast policies:

```yaml
topics:
  my_topic:
    policies:
      balanced:
        queue_capacity: 256
        on_full: block
      broadcast:
        queue_capacity: 8192
        on_lag: disconnect
        ack_mode: first                   # or all
      ack_propagation: auto
```

### Supported values

- `first`
    - preserve current behavior
    - first broadcast subscriber Ack/Nack resolves the upstream message
- `all`
    - upstream Ack only when all broadcast subscribers eligible for that message  Ack
    - any Nack resolves the upstream message as Nack

Defaults:
- `broadcast.ack_mode = first`
- `ack_propagation = disabled` remains unchanged
- `broadcast.ack_mode` only matters when `ack_propagation = auto`

## Lag policy rules

For v1, `broadcast.ack_mode = all` should only be supported with: `broadcast.on_lag = disconnect`

and rejected with: `broadcast.on_lag = drop_oldest`

Reason: all + drop_oldest has ambiguous correctness semantics, because a subscriber may silently miss a message that is still awaiting its Ack.

### Required all + disconnect semantics

- Eligible broadcast subscribers are snapshotted at publish time
- Subscribers added later do not affect already-published messages
[...]

PR #3352, opened 2026-06-24
chore(engine/topic): engine wiring of `all` ack mode (PR 2/3)

drewrelmas:
# Change Summary

Next PR in sequence for #2252.

#3323 introduced primitives for this implementation. This PR connects those primitives throughout the rest of the `topic` implementation. This feature is still not wired through to end user config, that is the final PR.

### Outcome

An `all`-mode broadcast tracked publish now:

```text
all eligible subscribers Ack ------> Ack
any required Nack/disconnect/lag --> Nack
```

`first` mode keeps the existing path without a subscriber registry.

### Included

- Snapshot eligible subscribers and reserve the ring sequence under the registry lock.
- Feed subscriber Acks and Nacks into per-message consensus.
- Nack only messages a disconnecting or lagging subscriber still owes.
- Remove or barrier the subscriber under the registry lock, then scan pending entries after releasing it.
- Prevent delayed publishers from overwriting newer wrapped ring slots; stale tracked commits resolve immediately as Nack.

Lock order:

```text
subscriber registry -> tracker state -> tracked entry -> permit
```

Ack bookkeeping for unrelated messages remains outside the tracker-wide lock.

### Benchmarks
Added on to the existing `topic` benchmarks to compare `first` and `all` ack mode performance.

#### Steady-state tracked publish

Identical workload: 10k messages, every subscriber receives and Acks every message, averaged across 32/256/4096-byte payloads.

| subscribers | `first` us/msg | `all` us/msg | ratio |
| --- | ---: | ---: | ---: |
| 1 | 0.80-0.84 | 1.02 | 1.22x-1.27x |
| 2 | 1.20-1.24 | 1.57-1.61 | 1.30x-1.31x |
| 4 | 2.43-2.48 | 3.04-3.05 | 1.23x-1.25x |
| 8 | 5.29-5.40 | 6.36-6.37 | 1.18x-1.20x |

Ranges are from two consecutive runs of the final scan-after-unlock design.

#### Cleanup

Setup is excluded. The measurement covers resolving all outstanding messages for one subscriber. The tracker scan runs after releasing the subscriber registry lock.

| pending | disconnect total | `DropOldest` lag total |
| --- | ---: | ---: |
| 1 | 2.34-2.36 us | 2.41-2.44 us |
| 64 | 18.66-19.88 us | 18.09-20.13 us |
| 1024 | 261.70-277.42 us | 259.52-265.68 us |

At 1024 pending messages, cleanup is about 0.25 us per entry. It scans the tracker, but does not block subscriber registration or tracked publishing.

Part of #2252 

Unit tests

No

### Changelog


rust/otap-dataflow/crates/engine/src/topic/topic.rs
```diff
@@ -1271,10 +1418,52 @@ impl InMemoryDeliveryFinalizer {
...
+        // `all`-mode broadcast: contribute this subscriber's Ack to the
+        // consensus. The message stays tracked (Ok) until the set completes.
+        if let (TopicBroadcastAckMode::All, Some(subscriber_id)) =
+            (self.ack_mode, self.subscriber_id)
+        {
+            return match self.outcomes.resolve_ack_from(message_id, subscriber_id) {
+                AckFromResult::Resolved | AckFromResult::StillPending => Ok(()),
+                AckFromResult::NotTracked => Err(MessageNotTracked),
+            };
+        }
```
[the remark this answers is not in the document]
drewrelmas: Good catch, addressed

rust/otap-dataflow/crates/engine/src/topic/topic.rs
```diff
@@ -1134,19 +1212,88 @@ impl<T: Send + Sync + 'static> BroadcastSub<T> {
...
+                    // `DropOldest`: the subscriber stays connected but the
+                    // skipped messages (ring sequence < new_read_seq) were
+                    // overwritten before it read them, so in `all` mode it can
+                    // never ack them. Nack exactly those owed entries; messages
+                    // at or after new_read_seq remain readable. No-op in `first`
+                    // mode.
+                    self.nack_skipped_on_lag(
+                        new_read_seq,
+                        Arc::<str>::from("broadcast subscriber lagged past message"),
+                    );
```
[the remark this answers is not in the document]
drewrelmas: Forgot that the v1 in the issue proposed not supporting this for now, but I feel like my solution actually covers the problem. By keeping track of the `seq` on each message, engine provides a terminal Nack result instead of orphaning the message or assuming ack.

rust/otap-dataflow/crates/engine/src/topic/topic.rs
```diff
@@ -643,9 +679,35 @@ impl<T: Send + Sync + 'static> BroadcastOnlyTopic<T> {
...
+        // `all` mode: snapshot the eligible subscriber set and reserve the ring
+        // sequence atomically under the registry lock so the consensus
+        // membership exactly matches who will receive this message. The heavier
+        // slot write plus waker fan-out happen after releasing the lock.
+        //
+        // Lock order is registry -> tracker (`register_consensus` locks the
+        // tracker); the tracker must never call back into the registry.
+        if let Some(registry) = &self.registry {
+            let subscribers = registry.subscribers.lock();
+            let snapshot = subscribers.clone();
```
jmacd: Here we are cloning a HashSet under a Mutex. I wonder if there are alternatives? Let's analyze the potential, at least.

drewrelmas: That's a good point, I've made a few perf improvements. On the one you found specifically (cloning `HashSet` in a `Mutex`), I've handled by using `Arc` so we only have to clone in the relatively rare cases when a subscriber appears or disconnects.

drewrelmas:
Followed up on the consensus-tracking overhead flagged in review along with some other suggestions from local Copilot review. Three changes to the `all`-mode ack path, all targeting per-publish work under locks:

1. **Copy-on-write subscriber membership** - the registry now holds `Arc<HashSet<SubscriberId>>`. A tracked publish clones the `Arc` (one refcount bump) to snapshot eligible subscribers instead of deep-copying the set under the registry `Mutex`. `subscribe`/`disconnect` use `Arc::make_mut`, so a live entry's frozen snapshot is never mutated retroactively.
2. **Lock-drop in `resolve_ack_from`** - the consensus ack path now locks the entry map only to `get().cloned()` the `Arc<entry>`, drops the map lock, resolves the entry under its own state mutex, then re-locks only to `remove` on resolution. Message ids are monotonic (`fetch_add`, never reused), so the re-lock `remove(id)` can't race onto a different entry. This is the bulk of the win.
3. **Clone-on-resolve nack reason** - nack helpers take `&Arc<str>` and clone only when a nack actually fires, instead of bumping the reason `Arc` for every scanned entry.

### Results

Same methodology as the initial PR description table (10k msgs, every subscriber acks every message, µs/msg averaged across 32/256/4096B). `first` mode isolates the non-consensus baseline; it's unchanged because none of these touch its path.

| subscribers | `all` µs/msg (before → after) | overhead vs `first` (before → after) |
| --- | --- | --- |
| 1 | 1.29 → 0.99 (−23%) | 1.49x → 1.21x |
| 2 | 2.04 → 1.50 (−26%) | 1.61x → 1.26x |
| 4 | 3.67 → 2.99 (−19%) | 1.42x → 1.20x |
| 8 | 8.49 → 6.47 (−24%) | 1.56x → 1.22x |

Consensus overhead drops from ~1.4–1.7x down to ~1.2x an equivalent tracked `first` — roughly halving the extra cost.

jmacd (commented):
Looks good. @lquerel should have a look at synchronization primitives, the Mutex. Note we merged https://github.com/open-telemetry/otel-arrow/pull/3323 as a foundation for this.

rust/otap-dataflow/crates/engine/src/topic/topic.rs
```diff
@@ -643,9 +679,37 @@ impl<T: Send + Sync + 'static> BroadcastOnlyTopic<T> {
...
+        // Lock order is registry -> tracker (`register_consensus` locks the
+        // tracker); the tracker must never call back into the registry.
+        if let Some(registry) = &self.registry {
+            let subscribers = registry.subscribers.lock();
+            let snapshot = Arc::clone(&subscribers);
+            let seq = self.broadcast_ring.reserve_seq();
+            let receipt = self
+                .outcomes
+                .register_consensus(id, timeout, permit, snapshot, seq);
+            drop(subscribers);
```
lalitb, line 714: Thinking out loud: could this cause corruption if multiple processors within the same dataflow engine instance publish concurrently to the shared topic and the ring wraps?

drewrelmas: Good catch, yes that was possible. A publisher paused between reservation and commit could overwrite a newer sequence after the ring wrapped. I updated to always preserve the newest sequence, treating this case as a lag instead.

rust/otap-dataflow/crates/engine/src/topic/topic.rs
```diff
@@ -1134,19 +1226,88 @@ impl<T: Send + Sync + 'static> BroadcastSub<T> {
...
+    /// completes.
+    fn nack_skipped_on_lag(&self, new_read_seq: u64, reason: Arc<str>) {
+        let (Some(registry), Some(subscriber_id)) = (&self.registry, self.subscriber_id) else {
+            return;
+        };
+        let subscribers = registry.subscribers.lock();
+        self.ack_state
+            .outcomes
+            .nack_owed_before(subscriber_id, new_read_seq, reason);
+        drop(subscribers);
```
lquerel: Disconnect and lag handling hold the subscriber registry lock while synchronously scanning and locking every pending tracked publish. The topic wide entry count is not bounded, so this can monopolize the single-threaded runtime core and block publishing, subscriptions, and acknowledgements.

I suggest to make disconnect/lag processing bounded, such as by maintaining a per-subscriber/sequence index or enforcing a topic-wide pending limit. Avoid a full tracker scan while holding the registry lock, and benchmark the worst case path.

rust/otap-dataflow/crates/engine/src/topic/types.rs
```diff
@@ -280,20 +282,66 @@ impl TrackedPublishTracker {
...
+    ) -> NackFromResult {
         let mut entries = self.inner.entries.lock();
         let Some(entry) = entries.get(&message_id).cloned() else {
-            return AckFromResult::NotTracked;
+            return NackFromResult::NotTracked;
         };
-        let result = entry.resolve_ack_from(subscriber_id);
-        let resolved = matches!(result, AckFromResult::Resolved);
-        if resolved {
+        if entry.nack_if_requires(subscriber_id, &reason) {
```
lquerel: `nack_if_requires` returns false for first mode entries, resolved entries, and non-required subscribers. `resolve_nack_from` maps every false to `NotRequired`, despite documenting first-mode and resolved entries as NotTracked.

We should return a richer internal result that distinguishes Resolved, NotRequired, and NotTracked, then map each state to the documented public variant.

rust/otap-dataflow/crates/engine/src/topic/tests.rs
```diff
@@ -1863,6 +1864,529 @@ async fn tracked_publish_broadcast_mode() {
...
+fn recv_message_id(item: Result<RecvItem<u64>, Error>) -> u64 {
+    match item.unwrap() {
+        RecvItem::Message(env) => env.id,
+        RecvItem::Lagged { missed } => panic!("unexpected lag of {missed} messages"),
+    }
+}
+
+// `all` mode resolves the upstream Ack only after *every* eligible broadcast
+// subscriber has acked the message.
+#[tokio::test]
```
lquerel: nit: We should adopt the Scenario: ... Guarantees: ... template for all our new tests.

lquerel (approved):
There are a few improvements to make before merging. Otherwise, it looks good to go.

rust/otap-dataflow/.chloggen/broadcast-ack-all-engine-wiring.yaml
```diff
@@ -0,0 +1,25 @@
...
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Not yet configurable: this wires the engine to honor `all` mode but the config surface lands in a
+  follow-up PR under #2252. An `all`-mode broadcast topic now resolves the upstream receipt only once
+  every eligible subscriber acks, snapshotting the required subscriber set under a registry -> tracker
+  lock order, and nacks fast on subscriber nack/disconnect/lag. Also fixes a stall where a required
+  subscriber lagging past a tracked message under the `DropOldest` policy would leave consensus
+  incomplete forever; skipped messages are now nacked while still-readable ones stay pending.
```
lquerel: I didn't check but there is probably more than 300 characters here :-)

rust/otap-dataflow/crates/engine/src/topic/types.rs
lquerel, line 356: More an optimization idea to explore than an issue per se. 

Every disconnect or DropOldest lag scans the entire tracked-publish map. Repeated lag events can therefore cost O(total pending publishes) each, even when only a small sequence range is affected.

Could we maintain a sequence ordered index for lag processing and/or a per-subscriber index of outstanding messages. Process only entries belonging to that subscriber and, for lag, only sequences below the threshold.

rust/otap-dataflow/crates/engine/src/topic/types.rs
```diff
@@ -312,7 +360,41 @@ impl TrackedPublishTracker {
...
             .filter_map(|(id, entry)| {
                 entry
-                    .nack_if_requires(subscriber_id, Arc::clone(&reason))
+                    .nack_if_requires(subscriber_id, &reason)
+                    .then_some(*id)
+            })
+            .collect();
+        for id in &to_remove {
+            let _ = entries.remove(id);
+        }
```
lquerel: Again another perf optimization suggestion/exploration.

Both scans allocate and populate a temporary `Vec<u64>`, then traverse it again to remove entries.

Could we remove matching entries in one pass with HashMap::retain or an equivalent extraction API, while recording whether anything resolved. This removes the temporary allocation, though it does not solve the larger full-scan cost.

rust/otap-dataflow/crates/engine/src/topic/types.rs
```diff
@@ -524,8 +618,19 @@ enum PendingKind {
...
-        pending: HashSet<BroadcastSubscriberId>,
+        /// Immutable set of subscribers eligible at publish time. Shared via
+        /// `Arc` so publishing only bumps a refcount instead of deep-copying the
+        /// registry's membership on every tracked publish (copy-on-write lives on
+        /// the rarer subscribe/disconnect paths).
+        members: Arc<HashSet<BroadcastSubscriberId>>,
+        /// Subscribers that have acked so far. The entry resolves as Ack once
+        /// `acked` covers every member. Grows lazily as acks arrive, so no
+        /// upfront per-message allocation proportional to the member count.
+        acked: HashSet<BroadcastSubscriberId>,
```
lquerel, line 622: Again another perf suggestion to explore

Every consensus message creates a separate HashSet for acknowledgements, with hashing and a heap allocation on the first ack. Your benchmark shows overhead scaling with subscriber count, so that could indicate a hot path.

I suggest to profile this path and if the problem exists then consider the following guidance from Codex. 

Assign each subscriber an ordinal within the shared membership snapshot. Store acknowledgements as an inline bitmap per message:
  - Snapshot maps BroadcastSubscriberId -> bit index.
  - Each message uses a u64 for up to 64 subscribers.
  - On ack, look up the bit index and test/set the bit.
  - An already-set bit identifies duplicate acks.
  - Resolve when the acknowledged count reaches the snapshot member count.
   - For more than 64 subscribers, fall back to a boxed multiword bitmap.

This moves the ID mapping into the shared snapshot, eliminates the per-message HashSet allocation for common fan-out sizes, and replaces ack-time hashing with a bit operation. Validate it against the existing 1/2/4/8-subscriber benchmark and add larger fan-out cases before adopting it.

rust/otap-dataflow/crates/engine/src/topic/types.rs
drewrelmas: I have done some testing of this idea locally (maintaining a per-subscriber index of outstanding messages) and while it does optimize the disconnect path, I also observe performance regression on the regular hot path because of the extra bookkeeping in the per subscriber index. I think we should optimize on the publish and ack path and be ok with a slower disconnect. I'll publish another iteration with another approach.

rust/otap-dataflow/.chloggen/broadcast-ack-all-engine-wiring.yaml
drewrelmas: Fixed in e01c3e1ae2f0287760354b202efd743cd5901583

rust/otap-dataflow/crates/engine/src/topic/topic.rs
```diff
@@ -192,9 +194,21 @@ impl<T: Send + Sync + 'static> FastBroadcastRing<T> {
...
+        let can_commit = match &*slot {
+            Some((slot_seq, _)) => *slot_seq < seq,
+            None => true,
+        };
+        if can_commit {
+            *slot = Some((seq, envelope));
+        }
+        drop(slot);
         self.waker_set.wake_all();
     }
```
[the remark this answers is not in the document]
drewrelmas: Fixed in e01c3e1ae2f0287760354b202efd743cd5901583. `commit_slot` now reports if the write succeeded. If a tracked message was overtaken before commit, its receipt immediately resolves as `Nack` instead of waiting for timeout.

rust/otap-dataflow/crates/engine/src/topic/topic.rs
drewrelmas: Fixed by separating ordering from cleanup in e01c3e1ae2f0287760354b202efd743cd5901583. Disconnect now removes the subscriber under the registry lock, then releases it before scanning.

rust/otap-dataflow/crates/engine/src/topic/types.rs
drewrelmas: Fixed, it now returns the same `NackFromResult` directly.

rust/otap-dataflow/crates/engine/src/topic/types.rs
drewrelmas: As mentioned, I tried a per-subscriber index. It kept the cleanup very fast, but the hot path overhead for `all` increased from `1.2x` to `2x-3.2x` versus `first`.

I removed the index and instead only scan after releasing the registry lock, bringing it back to the `1.2x` overhead.

rust/otap-dataflow/crates/engine/src/topic/types.rs
drewrelmas: Fixed, now uses `HashMap::retain`

rust/otap-dataflow/crates/engine/src/topic/types.rs
drewrelmas: I think this should be attempted in a follow-up PR.

drewrelmas: FWIW I don't think overhead actually grows with subscriber count, it is a fairly consistent ~1.2x of the `first` path regardless of count.

PR #3353, opened 2026-06-24
chore: bind test/dev servers to loopback to avoid Windows firewall prompts

### LINKED ISSUE #3351: Windows: test servers bind 0.0.0.0, causing repeated Firewall prompts during `cargo test` / `cargo xtask check`
## Summary

On Windows, running the Rust test suite — especially the full `cargo xtask check` — repeatedly triggers the Windows Defender Firewall "allow/deny" prompt. Because cargo names test binaries by content hash (`target\debug\deps\<crate>-<hash>.exe`), any "Allow" granted does not persist across rebuilds, so the prompt recurs on essentially every code change. It's a jarring inner-loop papercut for Windows contributors.

## Root cause

A handful of test/dev code paths bind listening sockets to `0.0.0.0` (all interfaces) instead of loopback. Windows Firewall exempts loopback (`127.0.0.1`) binds from filtering, so a loopback bind never prompts — an all-interfaces bind does. The codebase already binds the large majority of test servers to `127.0.0.1` (e.g. `crates/otap/src/otap_grpc/proxy.rs` and `client_settings.rs` use `127.0.0.1:0`); only a few paths fall back to `0.0.0.0`.

## Concrete examples

- **`crates/otap/src/otap_grpc/server_settings.rs:316`** — `impl Default for GrpcServerSettings` sets `listening_addr: ([0, 0, 0, 0], 0).into()` (i.e. `0.0.0.0:0`). Any test that starts a gRPC server from `GrpcServerSettings::default()` binds all interfaces.
- **`crates/telemetry/src/otel_sdk/meter_provider.rs:222`** — a meter-provider test configures the Prometheus pull exporter with `"host": "0.0.0.0", "port": 9090`; `prometheus_exporter_provider.rs:91` then calls `TcpListener::bind(addr)` on it (fixed port `9090`).

A quick sweep for non-loopback binds in test/dev paths should surface any remaining stragglers.

## Proposed fix

Bind test/dev servers to `127.0.0.1` rather than `0.0.0.0`:

- For tests relying on a `0.0.0.0` default, pass an explicit `127.0.0.1:0` (ephemeral) address.
- For the Prometheus meter-provider test, set `host` to `127.0.0.1`.
- Optionally add a small shared test helper for "loopback ephemeral bind" so this stays consistent, and/or a short note in the contributor docs.

### Scope / non-goals

- This is about **test/dev** binds. Production defaults that intentionally serve external traffic on `0.0.0.0` should stay as-is. (Whether `GrpcServerSettings::default()` *should* default to all-interfaces is a reasonable but separate question — defaulting servers to loopback is a common safer default.)
- Linux/macOS don't surface the firewall prompt, so this is primarily a Windows-contributor DX fix; the change is harmless/positive on all platforms.

## Acceptance criteria

- A full `cargo xtask check` on Windows completes without raising any Windows Firewall prompts (no test server binds a non-loopback interface).

## Environment

- Windows (native), Rust workspace under `rust/otap-dataflow`.

timr-dev:
## What

On Windows, running the Rust test suite — especially a full `cargo xtask check` —
repeatedly triggers the Windows Defender Firewall allow/deny prompt. A handful of
**test/dev** code paths bind listening sockets to `0.0.0.0` (all interfaces);
Windows exempts loopback (`127.0.0.1`) binds from the prompt. Because cargo names
test binaries by content hash, a granted "Allow" does not persist across
rebuilds, so the prompt recurs on essentially every change.

Fixes #3351.

## Changes

- **`crates/telemetry/src/otel_sdk/meter_provider.rs`** (test): the Prometheus
  pull-exporter test configured `host: "0.0.0.0", port: 9090`, which
  `MeterProvider::configure` → `start_async_prometheus_server` actually binds.
  This is the **only test that opens a non-loopback socket during `cargo test`**.
  Host → `127.0.0.1`.
- **`crates/contrib-nodes/examples/mock_la_server.rs`** (dev example): mock Azure
  Monitor Logs Ingestion server, documented for `http://localhost:9999`, bound
  `0.0.0.0:{port}`. Bind + banner → `127.0.0.1`.
- **`CONTRIBUTING.md`**: documents the loopback test/example bind convention so
  this stays consistent.

## Acceptance criterion

A full `cargo xtask check` on Windows now completes with no Windows Firewall
prompts — no test server binds a non-loopback interface. Verified locally:
structure ✅, `cargo fmt` ✅, `cargo clippy --workspace --all-targets -D warnings`
✅, `cargo test --workspace` ✅. The example compiles under
`--features azure-monitor-exporter`.

## Scope / non-goals (deliberate)

- **`GrpcServerSettings::default()` (`server_settings.rs:316`, `0.0.0.0:0`) is left
  unchanged.** It is a production default, and a full sweep confirms **no test
  binds via the bare default** — every gRPC/HTTP/UDP server test passes an
  explicit `127.0.0.1` address. Per the issue's non-goals, production defaults
  that intentionally serve external traffic stay as-is (same for the Prometheus
  pull `default_host()`).
- **Other `0.0.0.0` occurrences in the Rust workspace are left as-is because they
  do not bind a socket:** deserialize/`validate_config` tests that assert the kept
  production default (`readers.rs`, `pull.rs`, `prometheus_exporter_provider.rs`),
  the `otlp_receiver` test that specifically exercises the unspecified-address
  conflict path, the clap arg-parse test, and doc/CIDR/CEF strings.
- **Out of scope:** non-Rust demo/perf assets that intentionally bind all
  interfaces (e.g. the Python `tools/pipeline_perf_test` Docker templates, and the
  bundled demo `configs/*.yaml`). These are not exercised by `cargo xtask check`
  and do not affect the acceptance criterion; a Windows dev who manually runs one
  of those demos may still see a prompt.
- The fixed port `9090` in the meter-provider test is kept (the issue asked only
  to change the host); the spawned bind result is ignored, so a port collision
  cannot fail the test.

## Changelog

Test/dev-example only — no shipped/production behavior change — so this is a
`chore` with no `.chloggen` entry, per the repo convention.

## On a regression guard

An automated lint that rejects new non-loopback binds in test/example code was
considered but **deferred**: it would need an allow-list for the several
legitimate non-binding `0.0.0.0` usages above, making it brittle. The convention
is documented in `CONTRIBUTING.md` instead.

### DIFF
--- a/rust/otap-dataflow/CONTRIBUTING.md
+++ b/rust/otap-dataflow/CONTRIBUTING.md
@@ -31,15 +31,16 @@ not offset the limited parallelism gains.
 ## Test and example server bind addresses
 
 When a test or example starts a listening server, bind it to the loopback
-interface (`127.0.0.1`), not `0.0.0.0` (all interfaces). Prefer an ephemeral
-port (`127.0.0.1:0`) and read the assigned port back from the listener.
+interface (`127.0.0.1`, or `[::1]` for IPv6), not an all-interfaces address
+(`0.0.0.0`, or `[::]`/`[::]:0` for IPv6). Prefer an ephemeral port
+(`127.0.0.1:0`) and read the assigned port back from the listener.
 
 Windows Defender Firewall exempts loopback binds from its allow/deny prompt, so
 a loopback bind avoids the repeated firewall prompts that an all-interfaces bind
 triggers on every `cargo test` / `cargo xtask check` rebuild (test binaries are
 named by content hash, so a granted exception does not persist across rebuilds).
 Production defaults that intentionally serve external traffic may still bind
-`0.0.0.0`.
+`0.0.0.0` (or `[::]`).
 
 ## Changelog entries
 
--- a/rust/otap-dataflow/CONTRIBUTING.md
+++ b/rust/otap-dataflow/CONTRIBUTING.md
@@ -28,6 +28,19 @@ the full check path, likely because many of the longest tests are concentrated
 in a few large integration-style binaries, so the extra runner orchestration did
 not offset the limited parallelism gains.
 
+## Test and example server bind addresses
+
+When a test or example starts a listening server, bind it to the loopback
+interface (`127.0.0.1`), not `0.0.0.0` (all interfaces). Prefer an ephemeral
+port (`127.0.0.1:0`) and read the assigned port back from the listener.
+
+Windows Defender Firewall exempts loopback binds from its allow/deny prompt, so
+a loopback bind avoids the repeated firewall prompts that an all-interfaces bind
+triggers on every `cargo test` / `cargo xtask check` rebuild (test binaries are
+named by content hash, so a granted exception does not persist across rebuilds).
+Production defaults that intentionally serve external traffic may still bind
+`0.0.0.0`.
+
 ## Changelog entries
 
 User-facing Rust changes are recorded in
--- a/rust/otap-dataflow/crates/contrib-nodes/examples/mock_la_server.rs
+++ b/rust/otap-dataflow/crates/contrib-nodes/examples/mock_la_server.rs
@@ -345,7 +345,7 @@ async fn main() {
     let port = cli.port;
 
     println!("[mock-la] Mock Azure Monitor Logs Ingestion API Server");
-    println!("[mock-la] Listening on 0.0.0.0:{port}");
+    println!("[mock-la] Listening on 127.0.0.1:{port}");
 
     if cli.fail_rate > 0.0 {
         if let Some(retry_secs) = cli.retry_after {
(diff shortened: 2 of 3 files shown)


rust/otap-dataflow/CONTRIBUTING.md
```diff
@@ -28,6 +28,19 @@ the full check path, likely because many of the longest tests are concentrated
 in a few large integration-style binaries, so the extra runner orchestration did
 not offset the limited parallelism gains.
 
+## Test and example server bind addresses
+
+When a test or example starts a listening server, bind it to the loopback
+interface (`127.0.0.1`), not `0.0.0.0` (all interfaces). Prefer an ephemeral
```
ThomsonTan: Probably mention `[::]:0` as IPv6 all interfaces?

timr-dev: Good call — done in e7bff6ce4. The note now also calls out `[::]`/`[::]:0` (IPv6 all-interfaces) and `[::1]` (IPv6 loopback) alongside the IPv4 forms.

ThomsonTan (approved):

jmacd (approved):

Issue #3356, opened 2026-06-25
Azure Identity Auth extension (`BearerTokenProvider`)


pritishnahar95:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

Today the Azure Monitor exporter owns its own authentication: it constructs an
Azure credential, acquires a bearer token, and refreshes it inline within its
own event loop. This couples authentication to a single node and has several
drawbacks:

- **Duplication.** Every Azure-authenticating node must re-implement credential
  construction, token caching, refresh scheduling, and failure handling.
- **No sharing.** Two exporters in the same pipeline instance that target the
  same Azure resource each maintain independent token caches and refresh loops,
  doubling calls to IMDS / the token endpoint.
- **Lifecycle mismatch.** Token refresh is a background concern that does not
  belong on the data path; mixing it into an exporter's hot loop complicates
  both.

### Proposed Solution

Add an **Azure Identity Auth extension** (`azure_identity_auth`) that acquires
and refreshes Azure access tokens via the `azure_identity` SDK and exposes them
to data-path nodes through a `BearerTokenProvider` capability. Consumers (e.g.
the Azure Monitor exporter) resolve the capability and read a fresh token from
cache on the hot path, with no per-call credential round-trip.

Key elements:

- **Capability:** `BearerTokenProvider` with `get_token()` (cached fast path /
  single coalesced slow path) and `token_stream()` (refresh subscription). The
  capability trait, `BearerToken`, and `CapabilityError` are engine-owned
  (defined via the engine's `#[capability]` proc macro).
- **Execution model:** `Active + Shared`. A background refresh loop acquires
  tokens ahead of expiry and publishes them to a shared, lock-free `watch`
  cache; all consumers share one `Arc<Inner>` per pipeline instance.
- **Auth methods:** `managed_identity` (system- or user-assigned),
  `development` (local dev tooling), and `workload_identity` (federated
  ServiceAccount token; see #3336 / PR #3337).
- **Consumers:** Azure Monitor exporter, OTLP exporter (on export to Azure Monitor OTLP endpoints)

### Additional Context

- Builds on the Phase 1 extension system (capability proc macro, registry,
  Active/Passive lifecycle, local/shared execution models).
- Prerequisite: the deployed binary must enable exactly one `crypto-*` feature
[...]

PR #3357, opened 2026-06-25
doc: Add design doc for Azure Identity Extension

pritishnahar95:
# Change Summary

Adds a design doc for the **Azure Identity Auth extension**
(`azure_identity_auth`): an `Active + Shared` extension that acquires and
refreshes Azure access tokens via the `azure_identity` SDK and exposes them to
data-path nodes through an engine-owned `BearerTokenProvider` capability
(`get_token()` + `token_stream()`).

This is **documentation only** — no functional code is added in this PR.

* Related to #3356

Not applicable — doc-only change. Verified Markdown rendering and internal links
locally.

No. This PR adds a design document only; no runtime behavior, configuration, or
APIs change.

### Changelog

### DIFF
--- a/rust/otap-dataflow/docs/azure-identity-auth-extension.md
+++ b/rust/otap-dataflow/docs/azure-identity-auth-extension.md
@@ -1,0 +1,515 @@
+# Azure Identity Auth Extension
+
+<!-- markdownlint-disable MD013 -->
+
+**Status:** Draft
+
+**Extension URN:** `urn:microsoft:extension:azure_identity_auth`
+
+**Capability exposed:** `BearerTokenProvider`
+
+**Execution model:** Active + Shared
+
+**Target crate:** `crates/contrib-extensions`
+
+**Target module:** `crates/contrib-extensions/src/azure_identity_auth/`
+
+This document describes the design of the **Azure Identity Auth extension**
+(`azure_identity_auth`) for the OTAP dataflow engine. The extension acquires and
+refreshes Azure access tokens via the `azure_identity` SDK and exposes them to
+data-path nodes through a `BearerTokenProvider` capability.
+
+It builds on the extension system foundations:
+
+- [Extension System Proposal](extension-requirements.md) - the *what* and *why*
+  of the capability-based extension system.
+- [Extension System Architecture](extension-system-architecture.md) - the
+  Phase 1 *how* (capability proc macro, registry, Active/Passive lifecycle,
+  local/shared execution models).
+- [Design Principles and Constraints](design-principles.md) - thread-per-core
+  execution, minimal synchronization, security/privacy first.
+
+## Problem
+
+Several exporters need to authenticate to Azure services (e.g. Azure Monitor /
+Log Analytics) using OAuth bearer tokens. Today the Azure Monitor exporter owns
+its own authentication: it constructs an Azure credential, acquires a token,
+and refreshes it inline. This couples authentication to a single node and has
+several drawbacks:
+
+- **Duplication.** Every Azure-authenticating node must re-implement credential
+  construction, token caching, refresh scheduling, and failure handling.
+- **No sharing.** Two exporters in the same pipeline instance that target the
+  same Azure resource each maintain independent token caches and refresh loops,
+  doubling calls to the Azure Instance Metadata Service (IMDS) / token endpoint.
+- **Lifecycle mismatch.** Token refresh is a background concern that does not
+  belong on the data path; mixing it into an exporter's hot loop complicates
+  both.
+
+The extension system lets us factor authentication out into a **shared,
+cross-cutting capability** that any node can bind to, while keeping the data
+path free of authentication plumbing.
+
+## Goals
+
+- Provide a reusable `BearerTokenProvider` capability backed by Azure identity
+  flows (Managed Identity, developer tooling, and Workload Identity Federation).
+- Refresh tokens **in the background**, ahead of expiry, so consumers read a
+  fresh token from cache on the hot path with no per-call credential round-trip.
+- Coalesce concurrent token acquisitions within a pipeline instance so a cache
+  miss does not stampede the token endpoint. (At pipeline scope this is per
+  core; see [Performance Considerations](#performance-considerations).)
+- Share a single token cache and refresh loop across all consumers bound to one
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/docs/azure-identity-auth-extension.md
```diff
@@ -0,0 +1,515 @@
...
+            LOOP -->|publish| WATCH
+            WATCH --> CAP[BearerTokenProvider capability]
+        end
+        EXP[azure-monitor-exporter] -->|require_local / require_shared| CAP
+        EXP -->|get_token / token_stream| CAP
+    end
+    AUTH -->|OAuth token request| SRC
+```
+
+### Token flow
```
gouslu, line 180: we have readiness probe now, so extension can hold the startup until first token is received.

rust/otap-dataflow/docs/azure-identity-auth-extension.md
```diff
@@ -0,0 +1,515 @@
...
+pub trait BearerTokenProvider {
+    /// Return the current valid token. Fast path reads the cache; slow path
+    /// performs a single credential call on a cache miss.
+    async fn get_token(&self) -> Result<BearerToken, CapabilityError>;
+
+    /// Subscribe to the stream of token refreshes. Yields each newly
+    /// published token for the lifetime of the extension.
+    fn token_stream(
+        &self,
+    ) -> Pin<Box<dyn Stream<Item = Result<BearerToken, CapabilityError>> + Send + 'static>>;
```
gouslu: Could we remove send from the return type and give it an alias such as `TokenStream` to hide this lengthy return type?

rust/otap-dataflow/docs/azure-identity-auth-extension.md
```diff
@@ -0,0 +1,515 @@
...
+  doubling calls to the Azure Instance Metadata Service (IMDS) / token endpoint.
+- **Lifecycle mismatch.** Token refresh is a background concern that does not
+  belong on the data path; mixing it into an exporter's hot loop complicates
+  both.
+
+The extension system lets us factor authentication out into a **shared,
+cross-cutting capability** that any node can bind to, while keeping the data
+path free of authentication plumbing.
+
+## Goals
```
gouslu, line 53: One design goal I'd suggest making explicit: the BearerTokenProvider capability should be compatible with both active and passive providers across both shared and local execution models (active/passive × shared/local) — from both the provider's and the consumer's side. That's the real value of the token_stream() abstraction: a Stream is a per-consumer subscription, so an active provider can back it with a watch channel while a passive one backs it with an on-demand unfold, and consumers don't care which. Worth listing this explicitly under Goals.

I think this unbiased mental model is the right default for all capabilities we design, not just this one. It isn't the only valid approach — sometimes a capability genuinely warrants a more biased design — but wherever we can avoid baking in that bias, we should, since it pushes the active/passive and shared/local choices down to the implementer rather than the contract.

pritishnahar95: good point! added this as a goal

rust/otap-dataflow/docs/azure-identity-auth-extension.md
pritishnahar95: addressed

rust/otap-dataflow/docs/azure-identity-auth-extension.md
pritishnahar95: addressed

rust/otap-dataflow/docs/azure-identity-auth-extension.md
```diff
@@ -0,0 +1,537 @@
...
+
+**Extension URN:** `urn:microsoft:extension:azure_identity_auth`
+
+**Capability exposed:** `BearerTokenProvider`
+
+**Execution model:** Active + Shared
+
+**Target crate:** `crates/contrib-extensions`
+
+**Target module:** `crates/contrib-extensions/src/azure_identity_auth/`
```
drewrelmas, line 15: One additional decision point for us to solve - are `extensions` considered `nodes`? 

If so, this should go under `crates/contrib-nodes/src/extensions/azure_identity_auth`

If not, this should go under `crates/extensions/src/azure_identity_auth` (or as this PR iteration suggests, in `crates/contrib-extensions` if we want to maintain same separation as `nodes`)

@gouslu do you have an opinion here?

Given the fact that pipeline config definition places `extensions` and `nodes` as siblings at the same level, I think they should be split:

```yaml
groups:
  default:
    pipelines:
      main:
        extensions:
          ...
        nodes:
          ...
```

drewrelmas (approved):

rust/otap-dataflow/docs/azure-identity-auth-extension.md
gouslu: we should have contrib-extensions I think, because the goal is to separate extensions from nodes and based on all the separation we have done so far, this seems natural to me.

gouslu (approved):

rust/otap-dataflow/docs/azure-identity-auth-extension.md
jmacd: The Collector uses a structure where `connectors`, `exporters`, `extensions`, `processors`, `receivers` are siblings, where extensions are named like `extensions/authextension`, `extensions/storageextension` (for the record).

jmacd (approved):
Please include in your scope the `object_store` auth extension (for Azure) for the Parquet exporter. (Roughly [here](https://github.com/open-telemetry/otel-arrow/blob/510247bf32178590394681e082fac547cc05cc50/rust/otap-dataflow/crates/otap/src/object_store.rs#L21))

rust/otap-dataflow/docs/azure-identity-auth-extension.md
pritishnahar95: IMO - the way "nodes" are defined in the config, they are meant to be a "graph" node, meaning an entity actually handling pdata. Extensions are more supplementary functions than steps in a graph. In the future, we will have pipeline group / engine scoped extensions, so they may not be siblings of pipeline graph nodes.
It's best to keep extensions in a separate crate, as @gouslu  suggested ->  `contrib-extensions`

PR #3361, opened 2026-06-25
feat(etw_receiver): add telemetry counters for dropped events, decode failures

swashtek:
feat(etw_receiver): add telemetry counters for dropped events, decode failures

# Change Summary

* Closes #NNN

### Changelog


rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -648,6 +949,39 @@ pub struct EtwReceiverMetrics {
...
+
+    /// Number of fatal ETW session errors (`parse_until` / `ProcessTrace`
+    /// failures).
+    #[metric(unit = "{error}")]
+    pub session_errors: Counter<u64>,
+
+    /// Number of currently active ETW session subscriptions owned by this
+    /// receiver core (delta: `+1` on start, `-1` on terminal state).
+    #[metric(unit = "{session}")]
+    pub sessions_active: UpDownCounter<u64>,
```
utpilla: Could we drop this metric and the related `SessionActiveGuard` machinery? Each core holds exactly one subscription for its whole lifetime. This metric would only ever change on start or shutdown. These one-time events are better modeled as logs than metrics.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -648,6 +949,39 @@ pub struct EtwReceiverMetrics {
...
+
+    /// Number of ETW events dropped because the internal per-core channel was
+    /// full (backpressure between the `ProcessTrace` thread and the receiver).
+    #[metric(unit = "{event}")]
+    pub received_events_dropped_queue_full: Counter<u64>,
+
+    /// Number of fatal ETW session errors (`parse_until` / `ProcessTrace`
+    /// failures).
+    #[metric(unit = "{error}")]
+    pub session_errors: Counter<u64>,
```
utpilla: This metric also seems unnecessary. `session_errors` is effectively a 0-or-1 counter: `parse_until` runs for the process lifetime and the detached session never restarts, so it can only fire once, at teardown. Similar to https://github.com/open-telemetry/otel-arrow/pull/3361#discussion_r3488833662, these things better captured in logs than metrics.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -285,8 +290,9 @@ impl EtwReceiver {
 
         // Acquire this core's consumer channel from the per-session-name
         // session.  The first call initializes the session; subsequent calls
-        // pop from the pre-allocated pool.
-        let event_rx = session::subscribe(&cfg, num_cores).map_err(|e| {
+        // pop from the pre-allocated pool.  The shared telemetry handle bridges
+        // the `!Send` ProcessTrace callback to this core's metric set.
```
utpilla: "this core's metric set." is misleading. The producer doesn't write to one fixed core's `MetricSet`. It only bumps the shared `SessionTelemetry` atomics. On each tick, whichever core drains first claims the whole delta (`swap(0)`) into its own per-core set; the registry sums them into one series. There's nothing special about any single core's set.

You could write something like this:

```rust
// The shared telemetry handle bridges
// the `!Send` ProcessTrace callback to the async receivers: the producer
// writes the session-scoped atomics that whichever core drains first
// folds into its own metric set.
``

Comment (swashtek) [reply rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs]: Updated.

File: rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs [hunk 4]
```diff
@@ -430,9 +439,123 @@ fn try_flush_batch(
...
 impl EtwReceiver {
+    /// Fold the per-session atomic counters - written by the `!Send`
+    /// `ProcessTrace` callback - into this core's metric set.
+    ///
+    /// Each atomic is `swap(0)`-ed so the value is consumed exactly once: the
+    /// running total since the last drain becomes a counter `add()`. Call this
+    /// on every `CollectTelemetry` tick and immediately before any terminal
+    /// `snapshot()` so the final report is not lossy.
+    ///
+    /// ## Cross-core aggregation: "first core drains the whole delta"
```
utpilla: > "first core drains the whole delta"

That's not the case. `claim_session_atomics` does three independent `swap(0)`s (one per field), not a single atomic transaction. Since each core runs its own runtime and receives `CollectTelemetry` independently, two cores can tick concurrently and split the fields. For e.g. core A claims `observed` while core B claims `dropped_queue_full`. So the "whichever core reaches this first folds the whole count; the other cores see zero" framing is only true in the uncontended case, not in general.

This doesn't affect correctness, but the stated reason should be per-field rather than whole-bundle:
- each atomic `swap(0)` claims that one field's delta exactly once (no double count, no loss, even under a race), and
- all per-core `MetricSets` share one attribute key, so the registry sums them. Which core claimed which field is irrelevant; the per-field deltas always add back to the producer total.

Update the doc comments to reflect that actual behavior:

```rust
/// ## Cross-core aggregation: per-field claim, summed across cores
///
/// The `session_telemetry` atomics are **shared across all cores** of a
/// `session_name` (one `Arc<SessionTelemetry>` per session). Each field is
/// drained with an independent `swap(0)`, so a given field's accumulated
/// delta is claimed by **exactly one** core. The three swaps are *not* a
/// single transaction: if two cores tick concurrently they may split the
/// fields (one core claims `observed`, another claims `dropped_queue_full`,
/// etc.). In the common, uncontended case a single core claims all three.
///
/// This is correct for the aggregate because:
///
/// * Each `swap(0)` is atomic, so every field's delta is claimed by exactly
///   one core - no double counting and no loss, even when two cores tick
///   concurrently and split the fields between them.
/// * Every per-core `MetricSet` for this receiver shares the *same*
///   attribute key (the receiver registers metrics without a per-core
///   dimension), so the telemetry registry **sums** the per-core snapshots
///   under one series. Which core claimed which field is irrelevant; the
///   session-wide total is therefore exact.
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -620,13 +662,14 @@ fn spawn_etw_session(config: &Config, txs: Vec<mpsc::Sender<EtwEventData>>) -> R
...
             let ancillary = session.ancillary_data();
 
-            // Shared round-robin counter, drop counter, and sender list.
-            // All provider callbacks run on the single `ProcessTrace` thread,
-            // so `Cell` is safe — no atomics or locking needed.  Sharing the
-            // counter ensures uniform core distribution even at startup when
-            // multiple providers would otherwise all start at index 0.
+            // Shared round-robin counter and sender list.  All provider
+            // callbacks run on the single `ProcessTrace` thread, so `Cell` is
+            // safe — no atomics or locking needed.  Sharing the counter ensures
```
utpilla:
```suggestion
            // safe - no atomics or locking needed.  Sharing the counter ensures
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -702,16 +745,24 @@ fn spawn_etw_session(config: &Config, txs: Vec<mpsc::Sender<EtwEventData>>) -> R
...
-                        // defaults in place — future work will add manifest
-                        // decoding with a (Provider, Id, Version) cache key.
-                        if let Ok(result) = decoder.borrow_mut().decode(record) {
-                            event_name = result.event_name.unwrap_or("").to_owned();
-                            decoded_fields = extract_decoded_fields(
-                                result.event_data.format(),
-                                result.event_data.event_data(),
-                            );
+                        // defaults in place and is counted via
+                        // `received_events_invalid` — the event is still
```
utpilla:
```suggestion
                        // `received_events_invalid` - the event is still
```

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -430,9 +439,123 @@ fn try_flush_batch(
...
+/// **exactly one** caller even when several per-core receivers race here
+/// concurrently - there is no double counting. See
+/// [`EtwReceiver::drain_session_atomics`] for the cross-core aggregation and
+/// shutdown-race reasoning.
+fn claim_session_atomics(telemetry: &session::SessionTelemetry) -> SessionAtomicDeltas {
+    SessionAtomicDeltas {
+        observed: telemetry.observed.swap(0, Ordering::Relaxed),
+        dropped_queue_full: telemetry.dropped_queue_full.swap(0, Ordering::Relaxed),
+        invalid: telemetry.decode_failed.swap(0, Ordering::Relaxed),
+    }
```
utpilla: Minor optimization for `claim_session_atomics`: an unconditional `swap(0)` is a read-modify-write that always takes the cache line exclusive, even when it claims 0. For the counters that are usually zero (`dropped_queue_full`, `decode_failed`), we can `load` first (shared, no line bouncing) and only `swap` when there's actually something to claim:

```rust
#[inline]
fn claim_if_nonzero(a: &AtomicU64) -> u64 {
    if a.load(Ordering::Relaxed) == 0 {
        return 0;            // shared load; other cores can read concurrently
    }
    a.swap(0, Ordering::Relaxed)
}
```

Apply it **selectively**, keep the plain `swap` for `observed`:

```rust
SessionAtomicDeltas {
    observed: telemetry.observed.swap(0, Ordering::Relaxed),
    dropped_queue_full: claim_if_nonzero(&telemetry.dropped_queue_full),
    invalid: claim_if_nonzero(&telemetry.decode_failed),
}
```

`observed` is incremented per event, so at any real rate it's non-zero every tick. The pre-load would just be a wasted op before the inevitable `swap`. The other two are typically zero on a healthy provider, so the load lets all cores skip the exclusive RMW in the common case. Correctness is unchanged: if two cores race past the load, the first `swap` claims the delta and the second gets 0.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -624,10 +754,73 @@ impl local::Receiver<OtapPdata> for EtwReceiver {
...
+/// ```text
+/// received_events_observed
+///     = received_events_total            (events pulled off the channel)
+///     + received_events_dropped_queue_full   (events the producer dropped, never enqueued)
+///
+/// received_events_total
+///     = received_events_forwarded        (events in successfully-sent batches)
+///     + received_events_forward_failed   (events in batches that failed to build/send)
+///     + events still buffered in the in-flight builder at snapshot time
+/// ```
```
utpilla: Could we simplify the metric set? I think we could

**1. Drop the redundant consumer-side counter.** We currently track two denominators: `received_events_observed` (ingress, in the `ProcessTrace` callback) and `received_events_total` (per-event, pulled off the channel). The second is fully derivable:

```
received_events_total(consumer) = received_events_observed − received_events_dropped_*
```

The drops already have their own counter, so the consumer-side number adds no information. Removing it also drops a per-event `.inc()` from the hot `recv`/`try_recv` paths.

**2. Rename `received_events_observed` → `received_events_total`.** This is the ingress denominator. The same role as other nodes such as syslog's `received_logs_total` . `observed` is used nowhere else, so this reads consistently across receivers.

**3. Rename `received_events_dropped_queue_full` → `received_events_dropped_slow_worker`.** "Dropped queue full" reads oddly. There are *two* backpressure points here, and the downstream/egress one is already covered by `received_events_forward_failed`. This metric is the *ingress* drop: the consuming per-core worker couldn't drain the internal channel fast enough, so the ETW consumer thread filled it and events never reached the batching pipeline. Suggest a doc line to disambiguate:

> Events dropped because the consuming worker (this core's async receiver loop) couldn't drain the internal channel fast enough (the ETW consumer thread filled it). Contrast `received_events_forward_failed`, which is downstream backpressure.

Resulting set:

```
received_events_total                    // ingress (was: observed)
received_events_forwarded
received_events_forward_failed
received_events_invalid
received_events_dropped_slow_worker      // was: dropped_queue_full
received_events_rejected_memory_pressure
```

This collapses the counter algebra to a single denominator (`total = forwarded + forward_failed + dropped_slow_worker + in-flight`, with `invalid` orthogonal) and lets us drop the "two denominators / #3268" doc block.

swashtek: Fixed.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
swashtek: Good catch, I've added the non-zero claim for sparse counters only.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -546,6 +548,36 @@ fn extract_decoded_fields(
...
+    /// session, counted *before* any per-core channel send is attempted.
+    /// Published as `received_events_total` in the metric set.
+    ///
+    /// This is the producer-side ingress denominator. The slow-worker drop
+    /// rate is computable as `dropped_queue_full / observed`. See the
+    /// counter-algebra note on `EtwReceiverMetrics` for the exact relationships.
+    pub observed: AtomicU64,
+    /// Events dropped because a per-core channel was full (internal backpressure).
+    /// Published as `received_events_dropped_slow_worker` in the metric set.
+    pub dropped_queue_full: AtomicU64,
```
utpilla: Rename the variable name as well for consistency.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -430,9 +439,134 @@ fn try_flush_batch(
...
+/// the shared [`session::SessionTelemetry`] atomics.
+///
+/// Returned by [`claim_session_atomics`] and applied to a metric set by
+/// [`SessionAtomicDeltas::apply`]. Splitting the `swap` (claim) from the metric
+/// `add` (apply) keeps the borrow of the `RefCell<MetricSet>` short and, more
+/// importantly, makes the "claim exactly once" semantics unit-testable without
+/// constructing a full registered `MetricSet` (which requires a pipeline
+/// context). See the boundary-case tests in this module.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+struct SessionAtomicDeltas {
```
utpilla: I think we can improve the naming here:

| Current | Proposed | Why |
|---|---|---|
| `SessionTelemetry` (live store) | `SessionWideMetrics` | The durable session-scoped store the capture thread writes to; "session-wide" marks the scope (vs per-core). |
| `SessionAtomicDeltas` (snapshot) | `EventCountsDelta` | A transient snapshot of the event counts recorded since the last drain. Avoids `SessionCounts`, which reads as "a count of sessions"; the scope already lives on the store, so the snapshot just needs to say "event counts." |
| `claim_session_atomics()` | `take_event_counts()` | `take` matches the `swap(0)` read-and-reset semantics (cf. `mem::take` / `Option::take`). |
| `drain_session_atomics()` | `drain_event_counts()` | Plain verb that already means "take everything out of the source" (cf. `DrainIngress`); internally calls `take_event_counts()` then `.apply()`. Avoids the FP-jargon "fold." |

Reads cleanly end to end: drain an `EventCountsDelta` out of `SessionWideMetrics` via `take_event_counts()`, then `.apply()` it to the receiver's `MetricSet` (`EtwReceiverMetrics`).

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/session.rs
```diff
@@ -740,22 +788,29 @@ fn spawn_etw_session(config: &Config, txs: Vec<mpsc::Sender<EtwEventData>>) -> R
...
+                            // on the next `CollectTelemetry`. `fetch_add`
+                            // returns the previous value, so the running total
+                            // is `+ 1`.
+                            let count =
+                                telemetry.dropped_queue_full.fetch_add(1, Ordering::Relaxed) + 1;
+                            // Keep the rate-limited human-facing log for the
+                            // per-core detail the metric cannot carry:
+                            // first drop, then every 10,000th.
                             if count == 1 || count.is_multiple_of(10_000) {
                                 otel_warn!(
```
utpilla: I don't think this adds much value now that we have a dedicated metric for this. I say that mainly because it would there isn't any actionable knowledge produced by knowing which core was slow at some point in time.

Moreover, this thread's method is the hot-path. We should reduce its work as much as possible. Even though, we don't expect this in the happy path. This thread should get as much time as possible for reading the ETW events.

swashtek: This is added for local validation, it is not needed for production, I will remove it.

rust/otap-dataflow/crates/contrib-nodes/src/receivers/etw_receiver/mod.rs
```diff
@@ -430,9 +439,134 @@ fn try_flush_batch(
...
+    ///   session-wide total is therefore exact.
+    ///
+    /// The trade-off is intentional: the producer-side counters
+    /// (`received_events_total`, `received_events_dropped_slow_worker`, and
+    /// `received_events_invalid`) are **session-scoped, not
+    /// per-core**. They cannot be attributed to an individual core - if they
+    /// were given a per-core attribute they would land on an arbitrary core
+    /// each interval and produce noisy, meaningless per-core series. The
+    /// per-core dimension is preserved only in the human-facing
+    /// `etw.event.dropped` log, which carries the `core` field.
```
utpilla: This could be removed now that we have removed it.

utpilla (approved):
@swashtek Left a comment about stale comment. The CI checks for formatting are failing.

utpilla (approved):

PR #3363, opened 2026-06-26
fix(engine): size default memory-limiter hysteresis to a small recovery band

timr-dev:
## Summary

Fixes a bug in how the memory-limiter's `hysteresis` is defaulted when it is
omitted from a `MemoryLimiterPolicy`. The default was
`min(hard_limit - soft_limit, soft_limit - 1)`, which for a wide soft-to-hard gap
collapses to `soft_limit - 1`. That makes the Soft-to-Normal recovery threshold
(`usage < soft_limit - hysteresis`) land at `usage < 1`, so the limiter latches
in **Soft** pressure until usage falls to nearly zero.

The default is now `min(hard_limit - soft_limit, soft_limit / 10)` -- a recovery
band of at most ~10% of the soft limit -- so the limiter returns to **Normal**
once usage falls modestly below `soft_limit`.

## Scope

This PR is now **only** the hysteresis-default bug fix. The earlier
`soft_action` graduated-shed mechanism has been dropped: its Soft-pressure
admission control overlaps with the direction in #3529 per merged RFC 0002, and
maintainers preferred a single mechanism rather than two overlapping ones. No
`soft_action` config surface is introduced or referenced.

## Correctness

- `soft_limit / 10 <= soft_limit - 1` for every `soft_limit >= 1`, so recovery is
  only ever **earlier or equal** to the old default -- no slower-recovery regression.
- When `soft_limit < 10`, `soft_limit / 10 == 0`; a `hysteresis` of `0` was
  already reachable on `main` (at `soft_limit == 1`) and passes validation
  (`0 >= soft_limit` is false for `soft_limit >= 1`), giving recovery at
  `usage < soft_limit`. In production soft limits are GiB-scale, so the band sits
  well above zero.

## Tests

Two recovery tests drive the real `tick()` -> `classify()` -> publish path (not a
hand-forced level), asserting the Soft-to-Normal transition on both the tick
result and the published pressure state:

- `soft_pressure_recovers_to_normal_with_default_hysteresis`
- `soft_recovery_boundary_uses_strict_below_soft_minus_hysteresis` -- pins the
  strict `<`: with soft=100 / hysteresis=10, usage 90 stays **Soft**, usage 89 ->
  **Normal**.

Both fail if the fix is reverted (old default => hysteresis 99 => threshold `usage < 1`).

## Validation

- `otap-df-engine` memory_limiter: **19/19**
- `otap-df-config`: **332/0**
- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`: clean
- changelog note/subtext within enforced limits; markdownlint clean

## Files

- `crates/engine/src/memory_limiter.rs` -- fix + explanatory comment + 2 tests
- `crates/config/src/policy.rs` -- `hysteresis` field doc (omitted-default behavior)
- `docs/memory-limiter-phase1.md` -- recovery-from-Soft doc (omitted-default behavior)
- `.chloggen/memory-limiter-hysteresis-default.yaml` -- `bug_fix` / `engine`

### DIFF
--- a/rust/otap-dataflow/crates/engine/src/memory_limiter.rs
+++ b/rust/otap-dataflow/crates/engine/src/memory_limiter.rs
@@ -487,10 +487,16 @@ impl EffectiveMemoryLimiter {
             );
         }
 
+        // When `hysteresis` is omitted, derive a small recovery band below the
+        // soft limit: the smaller of the soft->hard gap and 10% of the soft
+        // limit. Capping at `soft_limit / 10` (rather than `soft_limit - 1`)
+        // keeps the band narrow even when the soft->hard gap is wide, so the
+        // limiter recovers to `Normal` once usage falls modestly below
+        // `soft_limit` instead of only after usage collapses toward zero.
         let hysteresis_bytes = policy.hysteresis.unwrap_or_else(|| {
             hard_limit_bytes
                 .saturating_sub(soft_limit_bytes)
-                .min(soft_limit_bytes.saturating_sub(1))
+                .min(soft_limit_bytes / 10)
         });
 
         if hysteresis_bytes >= soft_limit_bytes {
@@ -1154,7 +1160,131 @@ mod tests {
         })
         .expect("limiter should accept omitted hysteresis");
 
-        assert_eq!(limiter.hysteresis_bytes, 99);
+        // Derived band = min(hard - soft = 150, soft / 10 = 10) = 10 bytes, a
+        // narrow recovery margin below the soft limit (not soft_limit - 1).
+        assert_eq!(limiter.hysteresis_bytes, 10);
+    }
+
+    #[test]
+    fn soft_pressure_recovers_to_normal_with_default_hysteresis() {
+        // Regression: with an omitted `hysteresis`, a wide soft->hard gap must
+        // still yield a recovery band small enough to return to `Normal` once
+        // usage falls modestly below `soft_limit` -- not stay latched in `Soft`
+        // until usage collapses toward zero. Any consumer of the Soft-pressure
+        // signal (the pressure gauge, downstream throttling) stays engaged for
+        // the whole latched window, so the Soft->Normal transition timing is the
+        // user-visible contract. Samples are driven through `tick()` so the
+        // policy-derived hysteresis governs classification, exercising the
+        // defaulting and the pressure-state machine together (not a hand-forced
+        // level).
+        let mut limiter = EffectiveMemoryLimiter::from_policy(&MemoryLimiterPolicy {
+            mode: MemoryLimiterMode::Enforce,
+            source: MemoryLimiterSource::Rss,
+            check_interval: Duration::from_secs(1),
+            soft_limit: Some(100),
+            hard_limit: Some(250),
+            hysteresis: None,
+            retry_after_secs: 1,
+            fail_readiness_on_hard: true,
+            purge_on_hard: false,
+            purge_min_interval: Duration::from_secs(5),
+        })
+        .expect("limiter should accept omitted hysteresis");
+
+        // min(hard - soft = 150, soft / 10 = 10) = 10.
+        assert_eq!(limiter.hysteresis_bytes, 10);
+
+        // Feed usage 100 (== soft_limit) then 50 (well below soft_limit).
+        limiter.sampler = MemoryUsageSampler::from_test_probes(
+            Box::new(SequenceProbe {
+                samples: VecDeque::from([
+                    MemorySample {
+                        usage_bytes: 100,
+                        source: MemorySampleSource::Rss,
+                    },
+                    MemorySample {
+                        usage_bytes: 50,
+                        source: MemorySampleSource::Rss,
+                    },
+                ]),
+            }),
+            None,
+        );
+
+        let state = MemoryPressureState::default();
+
+        // A sample at the soft limit enters Soft pressure.
+        let entered = limiter.tick(&state).expect("tick should succeed");
+        assert_eq!(entered.current_level, MemoryPressureLevel::Soft);
+        assert_eq!(state.level(), MemoryPressureLevel::Soft);
+
+        // Usage of 50 is below soft_limit - hysteresis (90), so pressure
+        // recovers to Normal and the shared state reflects it.
+        let recovered = limiter.tick(&state).expect("tick should succeed");
+        assert_eq!(recovered.current_level, MemoryPressureLevel::Normal);
+        assert_eq!(state.level(), MemoryPressureLevel::Normal);
+    }
+
+    #[test]
+    fn soft_recovery_boundary_uses_strict_below_soft_minus_hysteresis() {
+        // Pins the exact Soft->Normal recovery threshold. With soft=100 and the
+        // derived default hysteresis of 10, recovery is a strict `<` at 90:
+        // usage == 90 stays Soft, usage == 89 recovers to Normal. Driven through
+        // `tick()` so the real classify()/publish path is exercised.
+        fn limiter_with(second_sample: u64) -> (EffectiveMemoryLimiter, MemoryPressureState) {
+            let mut limiter = EffectiveMemoryLimiter::from_policy(&MemoryLimiterPolicy {
+                mode: MemoryLimiterMode::Enforce,
+                source: MemoryLimiterSource::Rss,
+                check_interval: Duration::from_secs(1),
+                soft_limit: Some(100),
+                hard_limit: Some(250),
+                hysteresis: None,
+                retry_after_secs: 1,
+                fail_readiness_on_hard: true,
+                purge_on_hard: false,
+                purge_min_interval: Duration::from_secs(5),
+            })
+            .expect("limiter should accept omitted hysteresis");
+            limiter.sampler = MemoryUsageSampler::from_test_probes(
+                Box::new(SequenceProbe {
+                    samples: VecDeque::from([
+                        MemorySample {
+                            usage_bytes: 100,
+                            source: MemorySampleSource::Rss,
+                        },
+                        MemorySample {
+                            usage_bytes: second_sample,
+                            source: MemorySampleSource::Rss,
+                        },
+                    ]),
+                }),
+                None,
+            );
+            (limiter, MemoryPressureState::default())
+        }
+
+        // First tick (usage 100) enters Soft; second tick at usage == 90 stays
+        // Soft because recovery is a strict `<` (90 is not < 90).
+        let (mut at_boundary, state) = limiter_with(90);
+        assert_eq!(
+            at_boundary.tick(&state).expect("tick").current_level,
+            MemoryPressureLevel::Soft
+        );
+        assert_eq!(
+            at_boundary.tick(&state).expect("tick").current_level,
+            MemoryPressureLevel::Soft
+        );
+
+        // Second tick at usage == 89 (soft_limit - hysteresis - 1) recovers.
+        let (mut below_boundary, state2) = limiter_with(89);
+        assert_eq!(
+            below_boundary.tick(&state2).expect("tick").current_level,
+            MemoryPressureLevel::Soft
+        );
+        assert_eq!(
+            below_boundary.tick(&state2).expect("tick").current_level,
+            MemoryPressureLevel::Normal
+        );
     }
 
     #[test]
--- a/rust/otap-dataflow/docs/memory-limiter-phase1.md
+++ b/rust/otap-dataflow/docs/memory-limiter-phase1.md
@@ -277,7 +277,10 @@ stateDiagram-v2
   when the threshold is crossed.
 - **Recovery from Soft** requires usage to drop below
   `soft_limit - hysteresis` before returning to `Normal`. This prevents
-  oscillation when usage hovers near the soft threshold.
+  oscillation when usage hovers near the soft threshold. When `hysteresis` is
+  omitted it defaults to `min(hard_limit - soft_limit, soft_limit / 10)`, a
+  narrow band so recovery happens once usage falls modestly below `soft_limit`
+  rather than only after it collapses toward zero.
 - **Recovery from Hard** requires usage to drop below `soft_limit`
   before returning to `Soft`.
 - Phase 1 does **not** implement cooldown timers. Those are planned for a
--- a/rust/otap-dataflow/.chloggen/memory-limiter-hysteresis-default.yaml
+++ b/rust/otap-dataflow/.chloggen/memory-limiter-hysteresis-default.yaml
@@ -1,0 +1,21 @@
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: engine
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Size the default memory-limiter hysteresis as `min(hard_limit - soft_limit, soft_limit / 10)` so an omitted hysteresis recovers from Soft once usage falls modestly below the soft limit.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [2425]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: |
+  Previously a wide soft-to-hard gap defaulted the band to `soft_limit - 1`, so the
+  limiter reported Soft pressure until usage fell to nearly zero. Policies with an
+  explicit `hysteresis`, and `source: auto` setups where the gap is already narrow,
+  are unaffected.
--- a/rust/otap-dataflow/crates/config/src/policy.rs
+++ b/rust/otap-dataflow/crates/config/src/policy.rs
@@ -554,6 +554,11 @@ pub struct MemoryLimiterPolicy {
     #[schemars(with = "Option<String>")]
     pub hard_limit: Option<u64>,
     /// Bytes below the soft limit required to leave `Soft` pressure.
+    ///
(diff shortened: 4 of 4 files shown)


lalitb:
Nit: Can we update `rust/otap-dataflow/docs/memory-limiter-phase1.md` too? The “Receiver Behavior Under Hard Pressure” section still says Soft never rejects, but with `soft_action: shed`, Soft can now use the same receiver shedding behavior in `enforce` mode.

Also, should we rename `memory-limiter-phase1.md` to something like `global-memory-limiter.md` and replace the “Phase 1” wording in the title/intro with “global/ process-wide memory limiter”? This PR still changes the global limiter, while the retained-work budgeting / tenant/group isolation direction is tracked separately in #3272. A behavior-based doc name may avoid Phase 1/Phase 2 confusion as both efforts evolve.

For the same reason, could we also reword “Phase 2 of the process-wide memory limiter” in the PR description? Maybe “follow-up to the process-wide memory limiter” would be clearer.

lalitb (approved):
Thanks. LGTM. Good to have doc update before merge.

timr-dev:
[timr-dev] Thanks for the review! Addressing the three points:

1. **Doc accuracy** — Done in 5c4873c10. Renamed the section from "Receiver Behavior Under Hard Pressure" to "Receiver Behavior When Shedding Ingress" and reframed it around shedding states: ingress is shed at `Hard` (always, in `enforce` mode) and at `Soft` when `soft_action: shed`, using the same per-receiver protocol-native signals (OTLP `503` / gRPC & OTAP `RESOURCE_EXHAUSTED` / Syslog drops) and the same receiver-level rejection counters. The default `soft_action: observe` still preserves the prior Hard-only behavior, and `observe_only` mode still overrides `soft_action` and accepts all requests. Verified line-by-line against the `shed_decision(mode, level, soft_action)` predicate in `memory_limiter.rs`.

2. **PR description** — Done. Reworded "Phase 2 of the process-wide memory limiter" → "follow-up to the process-wide memory limiter".

3. **Rename `memory-limiter-phase1.md` → `global-memory-limiter.md` + drop the "Phase 1" wording** — I'd prefer to keep this out of this PR and do it as a dedicated follow-up. The rename touches cross-references in `docs/configuration.md` (which this PR otherwise doesn't change), and the "Phase 1 → global / process-wide" reframing is really part of the naming evolution tracked in #3272, so it reads cleaner as its own focused change than mixing a doc-wide rename into the `soft_action` feature PR. I'm happy to open that follow-up right after this merges — target name `global-memory-limiter.md`, intro reworded to "global / process-wide memory limiter". Let me know if you'd rather I bundle it here instead.

lalitb:
> Rename memory-limiter-phase1.md → global-memory-limiter.md + drop the "Phase 1" wording — I'd prefer to keep this out of this PR and do it as a dedicated follow-up. The rename touches cross-references in docs/configuration.md (which this PR otherwise doesn't change), and the "Phase 1 → global / process-wide" reframing is really part of the naming evolution tracked in https://github.com/open-telemetry/otel-arrow/issues/3272, so it reads cleaner as its own focused change than mixing a doc-wide rename into the soft_action feature PR. I'm happy to open that follow-up right after this merges — target name global-memory-limiter.md, intro reworded to "global / process-wide memory limiter". Let me know if you'd rather I bundle it here instead.

Agree on this.

timr-dev:
[timr-dev] Thanks — agreed. I'll open the dedicated follow-up right after this merges: rename `memory-limiter-phase1.md` → `global-memory-limiter.md`, rework the intro to "global / process-wide memory limiter", and update the cross-references in `docs/configuration.md`, tracked under #3272. Keeping this PR scoped to the `soft_action` feature.

jmacd (approved):
Reviewed critically and validated locally (checked out the branch, ran targeted tests + compiled the new bench). Approving — everything below is non-blocking.

## Validation

- `config` 66 passed, `engine::memory_limiter` 24 passed (incl. all 8 new), `memory_admission` bench compiles, clippy clean on `config`.
- All three `should_shed_ingress()` sites and every ingress consumer (otlp/otap/syslog_cef/user_events receivers + the tower layer) route through the new predicate; no stray `== Hard` shed comparisons remain.
- `soft_action` propagation mirrors the existing `mode` pattern exactly (atomic on the process inner, plain field on receiver inners), and `configure()` runs before receivers snapshot.
- Default path is byte-identical (`shed_decision(_, _, Observe)` == old `Enforce && Hard`). Hysteresis is sound: with `soft_action: shed`, the existing Soft→Normal reopen threshold (`soft_limit − hysteresis`) becomes the shed-reopen point, so no oscillation.
- Doc sweep is thorough — grepped untouched receiver `.rs`/README files for stale "Soft is informational / Hard-only" wording; none remain.

## Non-blocking notes

1. **Process-wide `MemoryPressureState::should_shed_ingress()` has no production caller** — receivers only consult their snapshot copies (`Shared`/`LocalReceiverAdmissionState`). The extra relaxed `soft_action` atomic load on that method (and the two paragraphs justifying it) doesn't touch any real hot path. Not wrong, just more analysis than the call graph warrants.

2. **Wording nit in the description:** `soft_action: "throttle"` is rejected because serde rejects unknown *enum variants* by default — `deny_unknown_fields` is a separate mechanism that governs unknown *struct fields*. Code and the reject-unknown test are correct; only the write-up conflates the two.

3. **No end-to-end test that the HTTP/gRPC layer actually rejects at Soft when `soft_action: shed`.** `MemoryPressureLayer` is the middleware that converts a shed decision into a real `503`/`RESOURCE_EXHAUSTED`. Its only existing test, `soft_pressure_remains_advisory`, exercises just the *default* (Soft does not shed). The new shed *decision* is well covered — `all_three_sites_agree_with_oracle_across_full_matrix` asserts `SharedReceiverAdmissionState::should_shed_ingress()` returns `true` at Soft when shed, which is the exact call the layer makes. So the only thing not asserted is the trivial "layer forwards that `true` into a rejection" step. Low value — fine to fold into the receiver-coverage follow-up you already scoped.

4. **The "Baseline" column in the perf table can't be reproduced after merge.** Commit 1 adds the bench harness and commit 2 makes the behavior change, both in *this* PR — so there is no commit on `main` that has the bench but not the change. Anyone who checks out merged `main` and runs `memory_admission` gets only the "This PR" numbers. That's completely fine for the bench's actual purpose (catching *future* regressions against the committed code); the note is just that "Baseline" was a point-in-time dev measurement, not a number a reader can re-derive from `main` later.

rust/otap-dataflow/crates/engine/src/memory_limiter.rs
```diff
@@ -95,6 +98,43 @@ const fn mode_from_u8(val: u8) -> MemoryLimiterMode {
...
+    mode: MemoryLimiterMode,
+    level: MemoryPressureLevel,
+    soft_action: SoftAction,
+) -> bool {
+    if !matches!(mode, MemoryLimiterMode::Enforce) {
+        return false;
+    }
+    match level {
+        MemoryPressureLevel::Hard => true,
+        MemoryPressureLevel::Soft => matches!(soft_action, SoftAction::Shed),
```
lquerel: Could we add a regression test that builds the limiter with soft_action: shed, soft_limit=100, hard_limit=250, and omitted hysteresis, then feeds samples 100 → 50 through tick()? Assert that it first enters Soft/shedding and then recovers to Normal/stops shedding (or reject this configuration). Today the derived hysteresis is 99, so Soft only clears at usage=0; direct set_level_for_tests tests miss this policy-defaulting/classification interaction.

rust/otap-dataflow/docs/memory-limiter-phase1.md
```diff
@@ -290,14 +292,16 @@ Operationally, this means `soft_limit` is the reopening threshold after
 shedding begins. `hard_limit` starts rejection, but recovery does not begin
 until usage has fallen below `soft_limit`.
```
lquerel: Could we make this recovery guidance conditional on soft_action? With the default observe, Hard starts rejection and ingress reopens after transitioning below soft_limit. With soft_action: shed, rejection starts
  > at Soft and continues until the state reaches Normal, which requires usage < soft_limit - hysteresis and may take an additional sample after leaving Hard. As written, operators may incorrectly expect ingress to
  > reopen at soft_limit.

lquerel (commented):
@timr-dev Overall, this looks good to me. I think there's a small issue in the logic around the soft/hard limits combined with hysteresis.

@lalit During the review of this PR, I noticed a few things related to a previous PR on the memory limiter side that seemed suboptimal. I'll file a separate report for those.

lalitb:
> @lalit During the review of this PR, I noticed a few things related to a previous PR on the memory limiter side that seemed suboptimal. I'll file a separate report for those.

Sure @lquerel. Thanks for reviewing that part. Will look forward to your findings.

lalitb:
> > @lalit During the review of this PR, I noticed a few things related to a previous PR on the memory limiter side that seemed suboptimal. I'll file a separate report for those.
> 
> Sure @lquerel. Thanks for reviewing that part. Will look forward to your findings.

@lquerel  - I dug a bit more into the memory limiter after your comment. 

The [docs](https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/memory-limiter-phase1.md#tradeoffs) describe the mechanics: sampling is periodic and enforcement happens only at ingress when the state reaches `Hard`. So the limitation is mostly implied rather than stated directly: this can reduce new memory growth, but it is not a strict cap and it cannot immediately reclaim memory already held inside the process.

 The real issues I found are more specific:

 -    The default hysteresis is derived from the soft-to-hard gap and can become very large. With a wide gap, `Soft` can become sticky because memory must fall far below the soft limit to return to `Normal`. This mainly affects state reporting today because `Soft` does not reject ingress.
 - `source: auto` only checks the current/leaf cgroup limit. In nested cgroups, the real limit may be on a parent, so auto detection can miss it.
 - If memory sampling fails, the old pressure state stays in place. That can leave the limiter wrongly open or wrongly rejecting, with only a warning log.

I’ll anyway track a separate issue for these, but please let me know if you had a different issue in mind.

jmacd (approved):

rust/otap-dataflow/crates/engine/src/memory_limiter.rs
timr-dev: [timr-dev] Done — added exactly that test, and fixed the underlying defaulting so the recovery it checks for actually happens.

Root cause was the default hysteresis derivation `min(hard − soft, soft − 1)`: with a wide soft→hard gap it collapsed to `soft − 1` (99 for soft=100), so `Soft` only cleared at `usage < 1`. It's now `min(hard − soft, soft / 10)` → a 10-byte band, so recovery happens at `usage < 90`.

New tests in `memory_limiter.rs`, both driving samples through `tick()` (so the policy-defaulting + classification interact as in production, per your point about `set_level_for_tests` missing that path):
- `soft_shed_recovers_to_normal_with_default_hysteresis` — `soft_action: shed`, soft=100, hard=250, omitted hysteresis, feeds 100 → 50; asserts it enters Soft + sheds, then recovers to Normal + stops shedding.
- `soft_shed_recovery_boundary_uses_strict_below_soft_minus_hysteresis` — pins the boundary: usage==90 (`soft − hysteresis`) stays Soft; usage==89 recovers.

Fixed in aa58de618.

rust/otap-dataflow/docs/memory-limiter-phase1.md
timr-dev: [timr-dev] Rewrote the recovery guidance to be conditional on `soft_action` (`docs/memory-limiter-phase1.md`):
- Default `observe`: only `Hard` sheds; ingress reopens once usage falls back below `soft_limit` (the `Hard → Soft` transition).
- `soft_action: shed`: `Soft` also sheds; ingress reopens only at `usage < soft_limit − hysteresis` (the `Soft → Normal` transition).

Also documented the extra-sample detail you flagged: recovery from `Hard` first steps down to `Soft` (at `usage < soft_limit`), so a shed limiter keeps shedding for at least one more sample in `Soft` before reaching `Normal`. Fixed in aa58de618.

timr-dev:
[timr-dev] Thanks all — addressed the outstanding feedback in aa58de618.

**@lquerel** (review + inline threads): the sticky-`Soft` hysteresis logic is fixed — the omitted-hysteresis default is now `min(hard_limit − soft_limit, soft_limit / 10)` instead of `soft_limit − 1`, so a `soft_action: shed` limiter recovers modestly below the soft limit instead of latching until usage nears zero. Regression + boundary tests and the `soft_action`-conditional doc are in the two inline threads above.

**@lalitb** (#issuecomment-4963445519): the first item — the default hysteresis growing large from a wide soft→hard gap and making `Soft` sticky — is fixed here. The other two (auto detection only reading the leaf cgroup; stale pressure state when sampling fails) are real and out of scope for this PR; agree they belong in your separate tracking issue — happy to help there.

**@jmacd** (review notes): (1) agreed the process-wide `should_shed_ingress()` has no prod caller today — left in place for state-reporting/consistency, no code change. (2) Fixed the description wording — `throttle`/`shed_low_priority` are rejected as unknown **enum variants**, distinct from `deny_unknown_fields` (which governs unknown struct keys). (3) The layer-forwards-`true`→503 step I'll fold into the receiver-coverage follow-up. (4) Fair — "Baseline" was a point-in-time dev measurement, not re-derivable from merged `main`; the bench's purpose is catching future regressions against the committed code.

lalitb:
@timr-dev - One context update since this PR was first opened: while you were out, #3484 landed the pressure-aware rate throttling design, and #3529 is now implementing that path. That design uses Soft pressure as the trigger for selective receiver-local throttling, where over-rate scopes are throttled while within-rate scopes continue.

Given that newer direction, should we still keep `soft_action: shed` in this PR as a separate global Soft-pressure emergency mode? Or should this PR avoid adding that behavior now, and leave Soft-pressure throttling to the #3529 rate-throttling implementation? I realize this may mean some rework, but I think it is worth checking now so we do not land two overlapping Soft-pressure control models.

timr-dev:
[timr-dev] @lalitb — agreed, and thanks for catching this before it landed. Two Soft-pressure control models keyed off the same signal is exactly what we want to avoid. I talked this through with my operator before replying, so this reflects a reviewed decision rather than an automated one.

Reading RFC 0002, the split is deliberate: the memory limiter stays the **hard**-pressure global backstop (soft = informational), and the rate policy in #3529 owns **soft**-pressure admission control — selectively throttling over-rate scopes while within-rate scopes continue, and reusing the same `503` / `ResourceExhausted` responses. A global `soft_action: shed` is just the blunt, all-or-nothing version of what #3529 does properly, on the identical trigger. No reason to ship both.

So I'll **drop `soft_action: shed`** from this PR and leave Soft-pressure shedding to #3529.

What I'd like to keep is the **hysteresis-default bug fix** (independent of `soft_action`): the old `min(hard − soft, soft − 1)` default collapsed the recovery band, so the limiter only left `Soft` near usage = 0. Per the RFC's *Bursts and Recovery* section it's the limiter's own hysteresis that governs *leaving soft pressure* — the very signal #3529 uses to stop throttling — so fixing it (`min(hard − soft, soft / 10)`) directly benefits the rate-throttling path too.

Net: this PR narrows to "fix memory-limiter hysteresis default (+docs)", ceding all Soft-pressure *action* to #3529. I'll rescope and re-request review. Does that match what you had in mind?

lalitb:
> [timr-dev] @lalitb — agreed, and thanks for catching this before it landed. Two Soft-pressure control models keyed off the same signal is exactly what we want to avoid. I talked this through with my operator before replying, so this reflects a reviewed decision rather than an automated one.
> 
> Reading RFC 0002, the split is deliberate: the memory limiter stays the **hard**-pressure global backstop (soft = informational), and the rate policy in #3529 owns **soft**-pressure admission control — selectively throttling over-rate scopes while within-rate scopes continue, and reusing the same `503` / `ResourceExhausted` responses. A global `soft_action: shed` is just the blunt, all-or-nothing version of what #3529 does properly, on the identical trigger. No reason to ship both.
> 
> So I'll **drop `soft_action: shed`** from this PR and leave Soft-pressure shedding to #3529.
> 
> What I'd like to keep is the **hysteresis-default bug fix** (independent of `soft_action`): the old `min(hard − soft, soft − 1)` default collapsed the recovery band, so the limiter only left `Soft` near usage = 0. Per the RFC's _Bursts and Recovery_ section it's the limiter's own hysteresis that governs _leaving soft pressure_ — the very signal #3529 uses to stop throttling — so fixing it (`min(hard − soft, soft / 10)`) directly benefits the rate-throttling path too.
> 
> Net: this PR narrows to "fix memory-limiter hysteresis default (+docs)", ceding all Soft-pressure _action_ to #3529. I'll rescope and re-request review. Does that match what you had in mind?
> 
> Reply posted by [GS] @timr-dev Copilot AI Assistant.

@timr-dev -  Yes, that matches what I had in mind.

Keeping the hysteresis-default fix is useful independently and directly improves the Soft-pressure signal that #3529 relies on. Removing `soft_action: shed` also keeps the ownership clear:

- the memory limiter provides the process pressure state and remains the global Hard-pressure backstop;
- #3529 owns selective admission control at Soft pressure.

Please keep the hysteresis fix, its boundary/regression tests, and the related documentation, while removing the `soft_action` configuration and behavior. Once the rescope is pushed, I will take another look.

timr-dev:
[timr-dev] Rescope is live. This PR is now **only** the memory-limiter `hysteresis`-default bug fix -- the `soft_action` graduated-shed mechanism has been dropped entirely (grep-clean; no `soft_action` config surface), since its Soft-pressure admission control overlaps with #3529 per RFC 0002. The branch was rebuilt fresh on `main` and rebased onto the current tip, squashed to a single `bug_fix` commit.

The fix itself: the omitted-`hysteresis` default was `min(hard - soft, soft - 1)`, which collapses to `soft - 1` for a wide soft-to-hard gap and latches the limiter in Soft until usage falls to ~0. It's now `min(hard - soft, soft / 10)` -- a <=10% recovery band, provably <= the old default for all `soft >= 1` (no slower-recovery regression). Two recovery tests drive the real `tick()` -> `classify()` -> publish path and pin the strict Soft-to-Normal boundary. Validation: engine 19/19, config 332/0, fmt/clippy/markdownlint clean, changelog note within the enforced 200-char limit.

Thanks for the steer toward a single mechanism.

Issue #3300, opened 2026-06-15
Alignment of Metric naming conventions and instrumentation


drewrelmas:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?
 
Inspired by #3268 and https://github.com/open-telemetry/otel-arrow/pull/3268#discussion_r3416232880

9 nodes add their own **per-signal** counters (logs / metrics / spans) using ~6 naming schemes and several different units for the same concepts. `debug_processor` alone has 10 signal counters. Worse, nodes mix **counting granularities** (per-message vs per-item) without a documented convention, and units are sometimes mislabeled (e.g. a `{log}` counter that actually adds log *records*).

#### Per-signal nodes today

| Node | Logs | Metrics | Spans/Traces |
|------|------|---------|--------------|
| debug_processor (`processor.debug.pdata`) | `log_signals_consumed` `{log}` (adds log_records!), `events_consumed` `{event}`, `logs_consumed` `{msg}` | `metric_signals_consumed` `{metric}`, `metric_datapoints_consumed` `{datapoint}`, `metrics_consumed` `{msg}` | `span_signals_consumed` `{span}`, `span_links_consumed` `{link}`, `span_events_consumed` `{event}`, `traces_consumed` `{msg}` |
| filter_processor (`processor.filter.pdata`) | `log_signals_consumed` `{log}` | `metric_signals_consumed` `{metric}` | `span_signals_consumed` `{span}` |
| log_sampling_processor (`processor.log_sampling.pdata`) | `log_signals_consumed` `{log}` | — | — |
| signal_type_router (`processor.signal_type_router`) | `signals_*_logs` `{msg}` | `signals_*_metrics` `{msg}` | `signals_*_traces` `{msg}` |
| retry_processor (`processor.retry`) | `consumed_items_logs_*` `{item}` | `consumed_items_metrics_*` `{item}` | `consumed_items_traces_*` `{item}` |
| durable_buffer_processor (`otap.processor.durable_buffer`) | `consumed_log_records` `{log_record}` | `consumed_metric_points` `{data_point}` | `consumed_spans` `{span}` |
| otap pdata exporter (`exporter.pdata`) | `logs_consumed` `{msg}` | `metrics_consumed` `{msg}` | `traces_consumed` `{msg}` |
[...]

PR #3369, opened 2026-06-26
docs: Propose datapoint-level enum attributes for metrics

drewrelmas:
# Change Summary

## Datapoint-level enum attributes for metrics

Design-only RFC (doc: `rust/otap-dataflow/docs/telemetry/datapoint-enum-attributes.md`). Tracking issue: #3300.

### What changed from the previous drafts

The prior draft standardized two interoperable export shapes (name-split "granular" and attribute-split "agnostic") and let operators pick. Per review feedback this revision drops the dual-shape approach and, following further discussion, also drops the earlier  SignalCounter / signal_metric  instrument idea:

1. One format. Signal becomes a datapoint attribute (`signal`=`logs/metrics/traces`), matching the Collector's `otelcol.signal`.
2. Generalize. `signal` is one instance of bounded enum attributes that vary per datapoint (e.g.  `outcome`=`success/error`,  http.response.status_code), each with a static/dynamic qualifier.
3. Keep using `metric_set`/`attribute_set` concepts. This is not a new instrument family. Enum attributes are extra dimensions on a set's datapoints, not separate sets or instruments. Sets with no enum attributes keep today's behavior and hot path.

### Enum attributes: static vs dynamic

Attributes are grouped into reusable `#[attribute_set]` structs and referenced by a `metric_set`:

- **Static** (`static_attributes = <Set>`) — value fixed at registration; no per-call argument (e.g. `journald`, a logs-only receiver, binds `signal = logs` once).
- **Dynamic** (`dynamic_attributes = <Set>`) — value supplied per record. `durable_buffer_processor` folds its 8 hand-rolled loss counters into a single `lost_items` metric via `signal` + `outcome`.

```rust
#[derive(Debug, Clone, Copy, AttributeEnum)]
pub enum Signal { Logs, Metrics, Traces }
#[derive(Debug, Clone, Copy, AttributeEnum)]
pub enum LossOutcome { Dropped, Expired }

#[attribute_set(name = "durable_buffer.loss.attrs", dynamic)]
pub struct DurableBufferLossAttributes {
    #[attribute(key = "signal")]  pub signal: Signal,
    #[attribute(key = "outcome")] pub outcome: LossOutcome,
}

#[metric_set(name = "otap.processor.durable_buffer.loss",
             dynamic_attributes = DurableBufferLossAttributes)]
pub struct DurableBufferLossMetrics {
    #[metric(unit = "{item}")] pub lost_items: Counter<u64>,
}
```

Because every attribute is a closed enum, worst-case cardinality is known at compile time (the product of variant counts), so codegen can warn before a set exceeds the SDK's runtime cardinality limit - a guarantee free-form string attributes can't offer.

### Recording API

The recorder supplies dynamic variants through a single struct-bound view, `with(...)`, which returns a bound view of the whole set:

```rust
loss.with(DurableBufferLossAttributes { signal: Signal::Logs, outcome: LossOutcome::Expired })
    .lost_items
    .add(120);
```

### Scope

Design only - no instrumentation is migrated in this PR. Goals, background, worked examples (`durable_buffer_processor`, `journald`), and open questions are in the doc.

* Related to #3300 

N/A

N/A

### Changelog


rust/otap-dataflow/docs/telemetry/signal-metric-schemas.md
```diff
@@ -0,0 +1,368 @@
+# Dual signal-metric schemas (granular vs agnostic)
+
+Status: Draft (for discussion). Tracking issue:
+[open-telemetry/otel-arrow#3300](https://github.com/open-telemetry/otel-arrow/issues/3300).
+Scope: design only - proposes a direction, changes no production code.
```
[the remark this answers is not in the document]
drewrelmas: Fixing guidance in #3371

lquerel:
Sorry, I haven’t had time to read the proposal in detail yet. Could we just make sure that we: (1) keep the embedded web admin interface fully functional, including channels and topics, and (2) preserve the metric_set concept?

drewrelmas:
> Sorry, I haven’t had time to read the proposal in detail yet. Could we just make sure that we: (1) keep the embedded web admin interface fully functional, including channels and topics, and (2) preserve the metric_set concept?

@lquerel no rush - I imagine this would be a great topic to discuss in a SIG meeting before proceeding.

I don't believe the change would greatly change/break the admin UI. With respect to `metric_set`, this only builds on top of it with a `#[signal_metric]` field and the `SignalCounter` instrument.

I envision the schema choice (granular vs agnostic) just affects how a `SignalCounter` is expanded at registration-time.

jmacd (approved):
A1: costs less
B2->B1: same cost, this can be incremental
C1+C2: both options; compile-time chooses >=1 schema, runtime chooses from compiled choices

drewrelmas:
As a further proving ground for this idea, consider the existing `durable_buffer` metrics from: https://github.com/open-telemetry/otel-arrow/blob/41a1d778995fa162a83c6fb4619e5b374f0b211c/rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs#L155

With this proposed ability to control signal-metric schema, the following 8 `metric` declarations would collapse to only 2 `signal_metric`s:

```rust
// Current

/// Total individual items (log records, data points, spans) lost due to expired segments (max_age retention). Non-zero values indicate data aged out before delivery.
#[metric(unit = "{item}")]
pub expired_items: ObserveCounter<u64>,

/// Total individual items (log records, data points, spans) lost due to force-dropped segments (DropOldest policy). Non-zero values indicate data loss.
#[metric(unit = "{item}")]
pub dropped_items: ObserveCounter<u64>,

/// Log records lost due to force-dropped segments (DropOldest policy).
#[metric(unit = "{log_record}")]
pub dropped_log_records: Counter<u64>,

/// Spans lost due to force-dropped segments (DropOldest policy).
#[metric(unit = "{span}")]
pub dropped_spans: Counter<u64>,

/// Metric data points lost due to force-dropped segments (DropOldest policy).
#[metric(unit = "{data_point}")]
pub dropped_metric_datapoints: Counter<u64>,

/// Log records lost due to expired segments (max_age retention).
#[metric(unit = "{log_record}")]
pub expired_log_records: Counter<u64>,

/// Spans lost due to expired segments (max_age retention).
#[metric(unit = "{span}")]
pub expired_spans: Counter<u64>,

/// Metric data points lost due to expired segments (max_age retention).
#[metric(unit = "{data_point}")]
pub expired_metric_datapoints: Counter<u64>,
```

```rust
// Proposed

/// Items lost due to force-dropped segments (DropOldest policy).
#[signal_metric(verb = "dropped")]
pub dropped: SignalCounter<u64>,

/// Items lost due to expired segments (max_age retention).
#[signal_metric(verb = "expired")]
pub expired: SignalCounter<u64>,

// Callsites
m.expired.add(Signal::Logs, n);
m.dropped.add(Signal::Metrics, n);

granular   expired_log_records {log_record} ->   120   (entity attrs only)
           dropped_metric_points {data_point} ->  80   (entity attrs only)

agnostic   expired_items {item}
             -> 120  (entity + signal=logs)
           dropped_items {item}
             -> 80   (entity + signal=metrics)
```

This idea could be taken even further by supporting something like a differentiator attribute and take us down to a single `signal_metric` declaration:

```rust
// Beyond current scope, but possible

#[signal_metric(verb = "lost", differentiator_attribute = "reason")]
pub lost: SignalCounter<u64>,

m.lost.add(Signal::Logs, 120, "expired") // emits with attribute reason=expired
m.lost.add(Signal::Metrics, 80, "dropped") // emits with attribute reason=dropped

granular   lost_reason_expired_log_records   {log_record} -> 120   (entity attrs only)
           lost_reason_dropped_metric_points {data_point} ->  80   (entity attrs only)

agnostic   lost_items {item}
             -> 120  (entity + signal=logs + reason=expired)
             ->  80  (entity + signal=metrics + reason=dropped)
```

rust/otap-dataflow/docs/telemetry/signal-metric-schemas.md
```diff
@@ -0,0 +1,377 @@
...
+  `consumed_spans`.
+- **Agnostic** - a **single** metric distinguished by a `signal` **attribute**
+  (`logs` / `metrics` / `traces`), matching the Collector's `otelcol.signal`
+  attribute, per the Collector [Pipeline Component Telemetry
+  RFC](https://github.com/open-telemetry/opentelemetry-collector/blob/main/docs/rfcs/component-universal-telemetry.md#processors).
+
+This RFC proposes supporting **both** as first-class targets, with node
+instrumentation written **once** and emitted in whichever schema is selected,
+via **codegen** rather than a runtime reshape. It explores the design axes and
+recommends one option each.
```
lquerel: What is the rationale for supporting two export formats for metrics as part of our project? I understand that OTC needs to support both for compatibility reasons, but do we have the same constraint?

I'm wondering whether we should simply adopt the approach using the `signal` attribute, which seems to be the future direction for OTC. Having two options will double the number of dashboards or alert setups, and I'm not sure I understand the benefit in our case.

rust/otap-dataflow/docs/telemetry/signal-metric-schemas.md
```diff
@@ -0,0 +1,377 @@
...
+  generate Rust via templates; the same registry drives `live-check`, so
+  declared and emitted cannot drift. Cost: a net-new codegen pipeline and a
+  second generator beside the proc-macro.
+- **B2 proc-macro generates, Weaver validates** - extend `#[metric_set]` with
+  the signal-split; Weaver only declares + `live-check`-validates (its current
+  role). Cost: schema lives in two places, kept consistent by CI.
+
+**Recommendation: B2 now, B1 as end-state.** The proc-macro already produces the
+granular descriptor; extending it is incremental. Adopt Weaver for declaration +
+validation immediately; revisit full Weaver codegen once the API stabilizes.
```
lquerel: I share the same conclusion, and I think that's what is expressed somewhere in `docs/telemetry`.

rust/otap-dataflow/docs/telemetry/signal-metric-schemas.md
```diff
@@ -0,0 +1,377 @@
...
+| Descriptor       | Static (both prebuilt) | Needs grouping metadata  |
+| Binary size      | Both shapes compiled   | One shape                |
+| Deploy-time swap | Yes (select at reg.)   | Yes                      |
+| New machinery    | Macro branch           | Mapping meta + transform |
+
+**Recommendation: A1.** It keeps the static descriptor and the hot path, and
+still gives Axis C2's deploy-time switching (compile both, select per config) -
+without A2's permanent mapping metadata and per-tick transform. Its only extra
+cost is binary size. (The per-data-point attribute work is needed by agnostic
+either way, so it is not a differentiator.)
```
lquerel: As I mentioned in my previous comment, I'm not convinced there's really a need to choose between these two options. If we go with the `signal` attribute approach only, then options A1 and A2 don't really exist anymore.

rust/otap-dataflow/docs/telemetry/signal-metric-schemas.md
```diff
@@ -0,0 +1,377 @@
...
+- **C2 runtime config** - a `metrics.signal_schema` knob chosen at startup, with
+  an optional per-component override. Deploy-time flexible; one binary serves
+  both audiences; costs binary size (both shapes compiled) and config plumbing.
+
+**C2 does not require A2's reshape engine:** with A1 codegen, each metric set
+picks its prebuilt shape at registration; the hot path is still a direct add.
+
+**Recommendation: C2**, engine-wide default `granular` with optional
+per-component override, paired with A1. Keep C1-style features only as an
+optional build-time prune for size-constrained builds.
```
lquerel: Same comment as for axis A.

lquerel (commented):
I'm in favor of:  
1. simplifying the approach by choosing a single export format, e.g. a `signal` attribute; and  
2. generalizing the approach based on the concept of enum-based dynamic attributes.

In the current proposal, we focus mainly on the `signal` attribute, which has three possible values: `logs`, `metrics`, and `spans`. But we have the same kind of problem with, for example, metrics that measure the outcome of an operation, i.e. `success` or `error`, or sometimes even a more granular error classification.

Proposal:

* Add support for enum-type attributes in attribute sets, and add a qualifier describing whether the attribute is static or dynamic.
* For static enum-based attributes, the value is provided when the metric set is registered (as today).
* For dynamic enum-based attributes, the value is provided at the instrumentation level (reporting methods `inc`, `add`, etc.). The generated code, whether via proc-macro or Weaver, will add to these reporting methods as many enum-typed arguments as there are dynamic attributes declared in the attribute set.
* At the snapshot level, dynamic attribute values are just integers, present only if there are dynamic attributes, so the cost is minimal.
* At the metric registry level, we will obviously need to take these attributes into account for aggregation.

rust/otap-dataflow/docs/telemetry/signal-metric-schemas.md
jmacd: I agree with the sentiment. My position is also to promote an OTel Metric SDK feature, the "Views" support we are familiar with, but implemented at compile time. Views have been (traditionally) implemented at runtime; I want configurable views at compile time. See this current spec proposal:

https://github.com/open-telemetry/opentelemetry-specification/pull/5173

rust/otap-dataflow/docs/telemetry/signal-metric-schemas.md
drewrelmas: Discussed further offline, agreed to take this approach as long as we have ITS metrics system capable of producing the legacy shape as needed.

jmacd (approved):

lquerel:
I spent some time thinking about a slightly different direction, and I think the most promising path is to keep the existing `metric_set` concept and model this as metric-set-level enum attributes rather than introducing a new `SignalCounter` / `signal_metric` instrument family.

The user-facing shape could look like this:

```rust
#[derive(Debug, Clone, Copy, AttributeEnum)]
pub enum TelemetrySignal {
    Logs,
    Metrics,
    Traces,
}

#[derive(Debug, Clone, Copy, AttributeEnum)]
pub enum LossOutcome {
    Dropped,
    Expired,
}

#[attribute_set(name = "durable_buffer.loss.attrs", dynamic)]
#[derive(Debug, Clone, Copy)]
pub struct DurableBufferLossAttributes {
    #[attribute(key = "signal")]
    pub signal: TelemetrySignal,

    #[attribute(key = "outcome")]
    pub outcome: LossOutcome,
}

#[attribute_set(name = "signal.attrs")]
#[derive(Debug, Clone, Copy)]
pub struct SignalAttributes {
    #[attribute(key = "signal")]
    pub signal: TelemetrySignal,
}
```

A metric set can then declare dynamic enum attributes:

```rust
#[metric_set(
    name = "processor.durable_buffer.loss",
    dynamic_attributes = DurableBufferLossAttributes
)]
pub struct DurableBufferLossMetrics {
    #[metric(unit = "{item}")]
    pub lost_items: Counter<u64>,
}
```

Or extra static enum attributes:

```rust
#[metric_set(
    name = "receiver.journald",
    extra_attributes = SignalAttributes
)]
pub struct JournaldMetrics {
    #[metric(unit = "{log_record}")]
    pub consumed: Counter<u64>,
}
```

Registration stays metric-set-oriented:

```rust
let mut loss_metrics =
    pipeline_ctx.register_dynamic_metrics::<DurableBufferLossMetrics>();

let mut journald_metrics =
    pipeline_ctx.register_metrics_with_extra_attributes::<JournaldMetrics>(
        SignalAttributes {
            signal: TelemetrySignal::Logs,
        },
    );
```

Instrumentation stays direct:

```rust
loss_metrics
    .for_dynamic_attributes(DurableBufferLossAttributes {
        signal: TelemetrySignal::Logs,
        outcome: LossOutcome::Expired,
    })
    .lost_items
    .add(expired_log_count);

loss_metrics
    .for_dynamic_attributes(DurableBufferLossAttributes {
        signal: TelemetrySignal::Metrics,
        outcome: LossOutcome::Dropped,
    })
    .lost_items
    .add(dropped_metric_point_count);

journald_metrics.consumed.add(log_count);
```

The important point is that the metric set remains the unit of declaration, registration, registry state, and admin/UI visibility. Dynamic enum attributes are additional dimensions on snapshots/datapoints, not separate metric sets or separate generated instruments.

Metric sets that do not declare extra static attributes or dynamic attributes keep the current behavior. They register with the default node/entity attributes only, expose no additional metric-set-level dimensions, and keep the existing hot path unchanged.
  
For implementation efficiency, I’d suggest avoiding a runtime hashmap on the hot path. Since dynamic attributes are enum-only, codegen can compute a dense bucket index from the enum variant indexes. With multiple dynamic attributes, this becomes a mixed-radix index over the enum cardinalities. For example, the bucket count above is:

```rust
TelemetrySignal::VARIANTS.len() * LossOutcome::VARIANTS.len()
```

By mixed-radix index, I mean the same idea as flattening nested loops into one array index. If TelemetrySignal has 3 variants and LossOutcome has 2 variants, then (signal, outcome) can be mapped into a single dense index such as:

```rust
bucket = signal.variant_index() + outcome.variant_index() * TelemetrySignal::VARIANTS.len();
```

The local dynamic metric set can then store `Vec<M>` buckets plus a compact touched bitset, and reporting can scan only touched buckets. That gives us no hot-path impact for metric sets without dynamic attributes, minimal hot-path cost for metric sets with dynamic attributes, and avoids node-side explosions of hand-written per-signal/per-outcome metric sets.

So my recommendation for a future implementation PR would be to frame this less as "signal metrics" and more as "bounded enum attributes for metric sets", with signal just being the first concrete use case. Static enum attributes handle components whose value is fixed at registration time; dynamic enum attributes handle dimensions whose value is known only when instrumentation records the metric.

drewrelmas:
> I spent some time thinking about a slightly different direction, and I think the most promising path is to keep the existing `metric_set` concept and model this as metric-set-level enum attributes rather than introducing a new `SignalCounter` / `signal_metric` instrument family.
> 
> ...

@lquerel thank you for the detailed feedback! I've incorporated most of these ideas into latest iteration of the PR. I'm open to further feedback on the reporting API. I believe the following are all viable:

```rust
// suggested by you
loss.for_dynamic_attributes(DurableBufferLossAttributes { signal: Signal::Logs, outcome: LossOutcome::Expired })
  .loss_items
  .add(120);

// shorter variant (proposed in the PR)
loss.with(DurableBufferLossAttributes { signal: Signal::Logs, outcome: LossOutcome::Expired })
  .loss_items
  .add(120);

// builder style
loss.signal(Signal::Logs).outcome(LossOutcome::Expired)
  .loss_items
  .add(120);
```

lquerel (approved):
Thanks for these updates. LGTM

PR #3372, opened 2026-06-27
feat(engine): add BearerTokenProvider capability

pritishnahar95:
# Change Summary

Introduce the engine-side capability surface for the Azure Identity Auth extension:

- BearerTokenProvider trait (via #[capability]) with get_token() and token_stream(), plus the secret-redacting BearerToken data type.
- CapabilityError + CapabilityErrorSource to attribute runtime capability failures to the failing (extension, capability) pair.
- Re-export the local/shared trait variants from {local,shared}::capability.

* Related to #3356

Unit tests

No

### Changelog

* [x] Added a `.chloggen/*.yaml` entry

### DIFF
--- a/rust/otap-dataflow/.chloggen/bearer-token-provider-capability.yaml
+++ b/rust/otap-dataflow/.chloggen/bearer-token-provider-capability.yaml
@@ -1,0 +1,15 @@
+change_type: enhancement
+
+component: engine
+
+note: Add the `BearerTokenProvider` capability (with `BearerToken` and `CapabilityError`) so extensions can hand OAuth bearer tokens to data-path nodes.
+
+issues: [3356]
+
+subtext: |
+  Defines the engine-side capability surface: the `#[capability]`-generated
+  `BearerTokenProvider` trait (`get_token` and `token_stream`), the
+  secret-redacting `BearerToken` type, and `CapabilityError` /
+  `CapabilityErrorSource` for attributing runtime capability failures to the
+  failing `(extension, capability)` pair. No provider implementation is
+  included.
--- a/rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
+++ b/rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
@@ -1,0 +1,129 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! The `BearerTokenProvider` capability.
+//!
+//! A small, purpose-built capability that hands out OAuth bearer tokens
+//! to data-path nodes. It is intentionally provider- and execution-model
+//! agnostic: the same trait serves active and passive providers across
+//! both the shared and local execution models. Consumers depend only on
+//! the two methods below, never on how a token is produced or refreshed.
+//!
+//! The `#[capability]` proc macro expands the trait into:
+//!
+//! - `pub mod local::BearerTokenProvider` (`!Send` trait variant)
+//! - `pub mod shared::BearerTokenProvider` (`Send` trait variant)
+//! - A `SharedAsLocalBearerTokenProvider` adapter
+//! - A zero-sized `pub struct BearerTokenProvider` registration handle
+//! - `local_entry::<E>` / `shared_entry::<E>` factory bridges
+//! - A `KNOWN_CAPABILITIES` distributed-slice entry
+
+use super::error::CapabilityError;
+use futures::Stream;
+use otap_df_engine_macros::capability;
+use std::fmt;
+use std::pin::Pin;
+use std::time::Instant;
+
+/// An OAuth bearer token plus its (optional) expiry.
+///
+/// The secret is held in memory only and is never emitted by [`Debug`]
+/// (see the manual impl below) so it cannot leak into logs or telemetry.
+/// `expires_on` is a monotonic [`Instant`] — providers convert the
+/// credential's absolute wall-clock expiry to an `Instant` once, so the
+/// value is immune to wall-clock jumps thereafter. `None` means the token
+/// does not expire (or no expiry was reported).
+#[derive(Clone)]
+pub struct BearerToken {
+    secret: String,
+    expires_on: Option<Instant>,
+}
+
+impl BearerToken {
+    /// Creates a token from its secret and optional monotonic expiry.
+    #[must_use]
+    pub fn new(secret: String, expires_on: Option<Instant>) -> Self {
+        Self { secret, expires_on }
+    }
+
+    /// The bearer token secret, for injection into an `Authorization`
+    /// header or an `object_store` credential.
+    #[must_use]
+    pub fn secret(&self) -> &str {
+        &self.secret
+    }
+
+    /// The monotonic instant at which this token expires, if known.
+    #[must_use]
+    pub fn expires_on(&self) -> Option<Instant> {
+        self.expires_on
+    }
+}
+
+// Manual `Debug` that redacts the secret. Never print `self.secret`.
+impl fmt::Debug for BearerToken {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("BearerToken")
+            .field("secret", &"<redacted>")
+            .field("expires_on", &self.expires_on)
+            .finish()
+    }
+}
+
+/// A per-consumer subscription to token refreshes.
+///
+/// Boxed to hide the concrete stream type so providers can back it
+/// differently (e.g. a `watch` channel or an `unfold`) without changing
+/// the signature. The `Send` bound is intentionally omitted: the
+/// subscription is always consumed on the core that created it
+/// (thread-per-core), so it need not be `Send`. The `#[capability]`
+/// macro emits this signature into both the `local` (`?Send`) and
+/// `shared` (`Send`) trait variants unchanged.
+pub type TokenStream = Pin<Box<dyn Stream<Item = Result<BearerToken, CapabilityError>> + 'static>>;
+
+/// Hands out OAuth bearer tokens to data-path nodes.
+#[capability(
+    name = "bearer_token_provider",
+    description = "Provides OAuth bearer tokens, refreshed in the background"
+)]
+pub trait BearerTokenProvider {
+    /// Returns the current valid token.
+    ///
+    /// The fast path reads a cached token; the slow path performs a
+    /// single (coalesced) credential call on a cache miss. Returns a
+    /// [`CapabilityError`] if no valid token can be produced.
+    async fn get_token(&self) -> Result<BearerToken, CapabilityError>;
+
+    /// Subscribes to the stream of token refreshes.
+    ///
+    /// Yields each newly published token for the lifetime of the
+    /// extension. Each call returns an independent subscription.
+    fn token_stream(&self) -> TokenStream;
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn accessors_round_trip() {
+        let now = Instant::now();
+        let token = BearerToken::new("super-secret".to_owned(), Some(now));
+        assert_eq!(token.secret(), "super-secret");
+        assert_eq!(token.expires_on(), Some(now));
+
+        let non_expiring = BearerToken::new("s".to_owned(), None);
+        assert_eq!(non_expiring.expires_on(), None);
+    }
+
+    #[test]
+    fn debug_redacts_secret() {
+        let token = BearerToken::new("super-secret".to_owned(), None);
+        let rendered = format!("{token:?}");
+        assert!(
+            !rendered.contains("super-secret"),
+            "secret leaked: {rendered}"
+        );
+        assert!(rendered.contains("<redacted>"));
+    }
+}
--- a/rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
+++ b/rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
@@ -21,36 +21,73 @@
 use super::error::CapabilityError;
 use futures::Stream;
 use otap_df_engine_macros::capability;
-use std::fmt;
+use secrecy::{ExposeSecret, SecretString};
 use std::pin::Pin;
-use std::time::Instant;
+use std::sync::Arc;
+use std::time::{Duration, Instant, SystemTime};
 
 /// An OAuth bearer token plus its (optional) expiry.
 ///
-/// The secret is held in memory only and is never emitted by [`Debug`]
-/// (see the manual impl below) so it cannot leak into logs or telemetry.
+/// The secret is wrapped in [`SecretString`], which zeroizes on drop and
+/// masks itself in [`Debug`] output, so it cannot leak into logs or
+/// telemetry. The `SecretString` sits behind an [`Arc`] so cloning a
+/// token (handing it to multiple `token_stream` subscribers, or returning
+/// it from `get_token` on the hot path) is a cheap refcount bump that
+/// shares one plaintext allocation rather than copying the secret bytes.
+///
 /// `expires_on` is a monotonic [`Instant`] — providers convert the
 /// credential's absolute wall-clock expiry to an `Instant` once, so the
 /// value is immune to wall-clock jumps thereafter. `None` means the token
 /// does not expire (or no expiry was reported).
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct BearerToken {
-    secret: String,
+    secret: Arc<SecretString>,
     expires_on: Option<Instant>,
 }
 
 impl BearerToken {
     /// Creates a token from its secret and optional monotonic expiry.
+    ///
+    /// Accepts anything convertible into [`SecretString`] (e.g. a
+    /// `String`), which is then shared behind an [`Arc`].
     #[must_use]
-    pub fn new(secret: String, expires_on: Option<Instant>) -> Self {
-        Self { secret, expires_on }
+    pub fn new(secret: impl Into<SecretString>, expires_on: Option<Instant>) -> Self {
+        Self {
+            secret: Arc::new(secret.into()),
+            expires_on,
+        }
     }
 
-    /// The bearer token secret, for injection into an `Authorization`
-    /// header or an `object_store` credential.
+    /// Creates a token from its secret and an **absolute** wall-clock
+    /// expiry.
+    ///
+    /// Credential services that report an absolute expiry often give it as
+    /// calendar time ([`SystemTime`]), but [`BearerToken`] stores a
+    /// monotonic [`Instant`] (see the field docs for why). Every such
+    /// provider has to perform the same wall-clock-to-monotonic
+    /// conversion; this constructor centralizes it so no provider gets it
+    /// subtly wrong.
+    ///
+    /// It measures how far `expires_on` is from *now* and offsets the
+    /// current `Instant` by that duration. An `expires_on` already in the
+    /// past (or a backwards clock) clamps to "expires immediately"
+    /// (`Instant::now()`) rather than producing a time before now.
     #[must_use]
-    pub fn secret(&self) -> &str {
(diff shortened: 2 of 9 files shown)


rust/otap-dataflow/.chloggen/bearer-token-provider-capability.yaml
```diff
@@ -0,0 +1,15 @@
+change_type: enhancement
+
+component: engine
+
+note: Add the `BearerTokenProvider` capability (with `BearerToken` and `CapabilityError`) so extensions can hand OAuth bearer tokens to data-path nodes.
+
+issues: [3337]
```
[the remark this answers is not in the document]
pritishnahar95: old comment, updated already

gouslu (approved):

pritishnahar95:
@jmacd @lquerel would appreciate it if you can at least take a quick look :)
This is an important capability - which multiple auth extensions in the future might want to implement

drewrelmas (approved):
This looks solid to me! Up to you if worth addressing the Copilot warning about `Acr<str>` on the secret value or if you get another Rust expert opinion on the perf implication.

rust/otap-dataflow/crates/engine/src/capability/mod.rs
```diff
@@ -12,9 +12,13 @@
 //! re-exported from [`local::capability`](crate::local::capability) and
 //! [`shared::capability`](crate::shared::capability).
 
+pub mod bearer_token_provider;
+pub mod error;
 pub mod factory;
 pub mod registry;
 
+pub use bearer_token_provider::{BearerToken, BearerTokenProvider, TokenStream};
+pub use error::{CapabilityError, CapabilityErrorSource};
```
utpilla: Both modules are `pub` and re-exported at `capability::`, so every item ends up reachable by two public paths, e.g.:

- `shared::capability::BearerTokenProvider` (the intended one), and
- `capability::bearer_token_provider::shared::BearerTokenProvider` (redundant)

and likewise for `BearerToken`, the zero-sized handle, and `CapabilityError`. Nothing outside the crate needs the raw module path. The only `bearer_token_provider::` uses are the in-crate `local::capability` / `shared::capability` re-exports. Making both modules `pub(crate) mod` collapses each item to a single canonical public path with no external breakage (the `pub use` re-exports still work from a crate-private module).

This matches the existing `pub mod factory` pattern, so it's consistent as-is; we should do that for `factory` too. `registry` should stay pub since `registry::Capabilities` isn't re-exported at the root.

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
```diff
@@ -0,0 +1,129 @@
...
+///
+/// The secret is held in memory only and is never emitted by [`Debug`]
+/// (see the manual impl below) so it cannot leak into logs or telemetry.
+/// `expires_on` is a monotonic [`Instant`] — providers convert the
+/// credential's absolute wall-clock expiry to an `Instant` once, so the
+/// value is immune to wall-clock jumps thereafter. `None` means the token
+/// does not expire (or no expiry was reported).
+#[derive(Clone)]
+pub struct BearerToken {
+    secret: String,
```
utpilla: We already have started using `secrecy::SecretString` for secrets in OTLP Exporter. Let's us that instead of `String`.

`secrecy::SecretString` gives zeroize-on-drop and `Debug` masking in one type.

- lets us delete the manual `Debug` impl (one less thing to keep correct),
- `secret()` becomes `expose_secret()`, making each plaintext access explicit/greppable.

utpilla: Also related to this, `get_token()` returns an owned `BearerToken`, and the provider will hand out a clone of its cached token on every call. With `secret: String`, each clone is a fresh heap allocation + copy of the token bytes on the export hot path, per call.

We could store the secret behind an `Arc` (ideally `Arc<SecretString>`. `BearerToken: Clone` then becomes a refcount bump plus a small `Option<Instant>` copy. No allocation and only one plaintext copy exists per refresh, shared across all consumers. 

I prefer this over wrapping the whole struct (`Arc<BearerToken>`): both give an alloc-free clone, but an inner `Arc` keeps the sharing an implementation detail, whereas `Arc<BearerToken>` leaks into the contract (`get_token`/`token_stream` return types) and makes every consumer deref through the `Arc`.

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
```diff
@@ -0,0 +1,129 @@
...
+/// A per-consumer subscription to token refreshes.
+///
+/// Boxed to hide the concrete stream type so providers can back it
+/// differently (e.g. a `watch` channel or an `unfold`) without changing
+/// the signature. The `Send` bound is intentionally omitted: the
+/// subscription is always consumed on the core that created it
+/// (thread-per-core), so it need not be `Send`. The `#[capability]`
+/// macro emits this signature into both the `local` (`?Send`) and
+/// `shared` (`Send`) trait variants unchanged.
+pub type TokenStream = Pin<Box<dyn Stream<Item = Result<BearerToken, CapabilityError>> + 'static>>;
```
lalitb: Should this be `Send` for the shared capability path? Right now the same `TokenStream` type is used there too, so a shared exporter that stores/polls this stream can become non-Send.

rust/otap-dataflow/crates/engine/src/capability/error.rs
```diff
@@ -0,0 +1,115 @@
...
+
+/// The error returned by a failed capability method invocation.
+///
+/// Carries the identity of the provider that failed — the extension
+/// instance id and the capability name — alongside the underlying
+/// `source` error, so a consumer can attribute the failure without
+/// knowing the provider's internal error types.
+#[derive(Debug, thiserror::Error)]
+#[error("capability '{capability}' on extension '{extension}' failed: {source}")]
+pub struct CapabilityError {
```
lalitb, line 37: Should this error be cheap to share or clone? `token_stream()` is documented as per-subscriber, but `Result<BearerToken, CapabilityError>` cannot be fanned out with watch/broadcast because only the token side is Clone.

rust/otap-dataflow/crates/engine/src/capability/error.rs
```diff
@@ -0,0 +1,115 @@
...
+use std::marker::PhantomData;
+
+/// The error returned by a failed capability method invocation.
+///
+/// Carries the identity of the provider that failed — the extension
+/// instance id and the capability name — alongside the underlying
+/// `source` error, so a consumer can attribute the failure without
+/// knowing the provider's internal error types.
+#[derive(Debug, thiserror::Error)]
+#[error("capability '{capability}' on extension '{extension}' failed: {source}")]
```
lalitb: Should we avoid printing `{source}` here since it is also marked as `#[source]`? Engine errors often print the error plus source chain, so this can show the same inner error twice.

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
```diff
@@ -0,0 +1,129 @@
...
+/// The secret is held in memory only and is never emitted by [`Debug`]
+/// (see the manual impl below) so it cannot leak into logs or telemetry.
+/// `expires_on` is a monotonic [`Instant`] — providers convert the
+/// credential's absolute wall-clock expiry to an `Instant` once, so the
+/// value is immune to wall-clock jumps thereafter. `None` means the token
+/// does not expire (or no expiry was reported).
+#[derive(Clone)]
+pub struct BearerToken {
+    secret: String,
+    expires_on: Option<Instant>,
```
lalitb, line 45: Is `Instant` the right public type here? Azure gives a wall-clock expiry, so every provider has to convert it correctly. A wall-clock expiry plus helper methods may be easier to use safely.

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
```diff
@@ -0,0 +1,129 @@
...
+
+/// Hands out OAuth bearer tokens to data-path nodes.
+#[capability(
+    name = "bearer_token_provider",
+    description = "Provides OAuth bearer tokens, refreshed in the background"
+)]
+pub trait BearerTokenProvider {
+    /// Returns the current valid token.
+    ///
+    /// The fast path reads a cached token; the slow path performs a
```
lalitb: Can we document the required sharing pattern here? Capability instances are per consumer, so this only stays\ coalesced if the implementation shares cache state internally.

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
```diff
@@ -0,0 +1,129 @@
...
+    /// single (coalesced) credential call on a cache miss. Returns a
+    /// [`CapabilityError`] if no valid token can be produced.
+    async fn get_token(&self) -> Result<BearerToken, CapabilityError>;
+
+    /// Subscribes to the stream of token refreshes.
+    ///
+    /// Yields each newly published token for the lifetime of the
+    /// extension. Each call returns an independent subscription.
+    fn token_stream(&self) -> TokenStream;
```
lalitb, line 149: Can we define what happens after this stream returns `Err`? It is not clear if the stream continues, ends, or if the consumer should call `get_token()` again.

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
```diff
@@ -0,0 +1,129 @@
...
+    name = "bearer_token_provider",
+    description = "Provides OAuth bearer tokens, refreshed in the background"
+)]
+pub trait BearerTokenProvider {
+    /// Returns the current valid token.
+    ///
+    /// The fast path reads a cached token; the slow path performs a
+    /// single (coalesced) credential call on a cache miss. Returns a
+    /// [`CapabilityError`] if no valid token can be produced.
+    async fn get_token(&self) -> Result<BearerToken, CapabilityError>;
```
lalitb, line 140: `get_token()` returns a token, but the trait does not say what audience/scope it is for. Azure tokens are resource specific, so the wrong provider can be bound successfully and only fail later as 401s.

Is the intent one provider instance per scope? If so, can we document that on the trait, or expose the audience so mismatches can be caught at startup?

pritishnahar95: The scopes will be listed in the provider config and the provider will then ensure that the granted token has the right scopes for it. See https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/azure-identity-auth-extension.md

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
gouslu: Both shared and local share the same trait, so at the trait level we need to stick with one. Is there an issue with this being no-send for shared implementers and consumers? I couldn't find an issue with it, but maybe I am missing something?

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
gouslu: I think putting the burden of conversion to provider is a good trade-off, because it simplifies consumer side logic and in general, I think that kind of trade-off is the one I would like to take. I am saying this, because I assume consumers would usually want to convert it to instant, because it is a simple/easy type to work with (?).

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
gouslu: Sharing state via Arc<> or Rc<> across cloned instances of capabilities/extensions should be part of general guidance of developing extensions. I don't think it should be documented per capability. After first extension contribution is complete, I think it will be a good time to write a full tutorial/documentation on extension development; it can be included in there. Does it sound sufficient to you?

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
gouslu: Stream should be alive for the lifetime of the extension. And, maybe contract shouldn't return error at all, what do you think @pritishnahar95 ?

rust/otap-dataflow/crates/engine/src/capability/error.rs
```diff
@@ -0,0 +1,115 @@
...
+use otap_df_config::ExtensionId;
+use std::marker::PhantomData;
+
+/// The error returned by a failed capability method invocation.
+///
+/// Carries the identity of the provider that failed — the extension
+/// instance id and the capability name — alongside the underlying
+/// `source` error, so a consumer can attribute the failure without
+/// knowing the provider's internal error types.
+#[derive(Debug, thiserror::Error)]
```
ThomsonTan, line 34: `CapabilityError` exposes all-`pub` fields, so adding a field later will be a breaking change. Probably add attribute `#[non_exhaustive]` to it can be extended later.

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
lalitb: That’s fair. I was looking at it from the shared API contract angle, not a concrete current compile failure.

If shared consumers are always polled on the same LocalSet today, then `!Send` may be fine. My main concern was that the shared trait looks `Send`, but the returned stream itself is not movable across threads, so future shared consumers may hit that limit unexpectedly.

If that is intentional, I’m okay with it.

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
lalitb: That makes sense for the consumer side. My concern is mostly about making the provider-side conversion hard to get wrong.

Since Azure gives an absolute expiry, every provider has to do the same “wall clock to Instant” conversion and handle already-expired / clock-skew cases correctly. If we keep `Instant`, could we add a small helper constructor like `BearerToken::from_absolute_expiry(secret, expires_on)` so that logic lives in one place?

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
lalitb: Yes, that sounds reasonable for the general extension docs.

For this PR, my only concern is the wording “single coalesced credential call” reads like the capability guarantees it. Maybe we can soften that line a bit, or add one short note that coalescing depends on the provider sharing its internal cache/refresh state across cloned instances. The full Arc/Rc guidance can live in the extension tutorial later.

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
lalitb: Yes, that’s exactly the ambiguity I was trying to point at. If the stream is expected to stay alive for the extension lifetime, then returning `Err` as an item is a bit unclear unless we define it as a recoverable refresh failure.

I’d be fine with either: remove `Err` from the stream item and keep failures on `get_token()`/telemetry, or document that `Err` does not terminate the stream and the next refresh may still yield a token.

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
lalitb: Thanks, that makes sense if the provider config owns the scopes. My only remaining concern is whether a consumer can accidentally bind to a provider configured for the wrong scope and only find out later as 401s. If startup validation is not planned now, can we document that the provider scope must match the consumer?

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
pritishnahar95: Unless I am missing something, `Send` isn't needed for `TokenStream` in our model.

`token_stream(&self)` is called on the consumer's own core, and the watch-backed subscription it returns is born and polled on that same core - it never crosses a core boundary. Even at the future group scope, the values that do cross cores are the extension handle (`Arc<Inner>`, already Send) and `BearerToken` itself (already Send); a Send stream would only matter if a provider shipped subscription handles across cores, which the API doesn't.

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
pritishnahar95: I think dropping `Err` from stream should be fine

rust/otap-dataflow/crates/engine/src/capability/error.rs
pritishnahar95: addressed

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
pritishnahar95: addressed - updated description

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
pritishnahar95: addressed - dropped `Err`

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
pritishnahar95: addressed - updated description

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
pritishnahar95: addressed - added helper method

rust/otap-dataflow/crates/engine/src/capability/error.rs
pritishnahar95: addressed - removed {source} from print

rust/otap-dataflow/crates/engine/src/capability/error.rs
pritishnahar95: addressed - removed `Err`

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
pritishnahar95: addressed according to feedback, thanks!

rust/otap-dataflow/crates/engine/src/capability/mod.rs
pritishnahar95: addressed according to feedback

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
```diff
@@ -0,0 +1,215 @@
...
+/// subscribers via a `watch`/`broadcast` channel.
+///
+/// Boxed to hide the concrete stream type so providers can back it
+/// differently (e.g. a `watch` channel or an `unfold`) without changing
+/// the signature. The `Send` bound is intentionally omitted: the
+/// subscription is always consumed on the core that created it
+/// (thread-per-core), so it need not be `Send`. The `#[capability]`
+/// macro emits this signature into both the `local` (`?Send`) and
+/// `shared` (`Send`) trait variants unchanged.
+pub type TokenStream = Pin<Box<dyn Stream<Item = BearerToken> + 'static>>;
```
lalitb, line 116: nit - also update the doc - https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/docs/azure-identity-auth-extension.md#capability-bearertokenprovider

lalitb (approved):
LGTM, thanks for iterating on this. The updated API shape looks much cleaner now.

rust/otap-dataflow/crates/engine/src/capability/bearer_token_provider.rs
pritishnahar95: addressed

PR #3374, opened 2026-06-28
feat(core-nodes): batch internal telemetry receiver logs by byte size, on by default

### LINKED ISSUE #1902: Internal telemetry receiver: batching option
The Internal Telemetry Receiver (ITR) component in rust/otap-dataflow/crates/otap/src/internal_telemetry_receiver.rs currently emits singleton log events into the pipeline. This was meant as a proof of concept, and while it is a good configuration some of the time, it would be reasonable to include a batching option in the ITR.

We cannot use the existing batch processor to achieve combining of multiple log records into the same ScopeLogs protocol message without converting to and from Arrow records, since the OTLP-bytes batch_processor currently concatenates OTLP bytes meaning it would produce batches of singletons if provided with data from ITR.

The input to ITR is a LogRecord. Currently this struct represents only the log event and callsite detail, however in #1895 (and #1898) we are adding the scope-level information as a small vector of EntityKeys, which are opaque cheaply-copied identifiers that we locate using thread-local variables. After #1898 merges, LogRecord will contain information we can use to sort by Entity keys and generate one InstrumentationScope per distinct set. With batching in the ITR, we should be able to see multiple log records per scope through the `console_exporter`, and the scope attributes will define the entities associated with each event with no more singletons.

After sorting the records, follow the logic in rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs to encode multiple scopes and records into a single `OtlpProtoBytes::ExportLogsServiceRequest` payload.

AvinashDevX:
# Change Summary
Batches log records instead of emitting one ExportLogsRequest per record. batch_size (bytes, default 64 KiB) is the flush threshold, so batching is on by default now. max_batch_duration (200ms) still bounds latency for quiet streams. Batches are split at 2 MiB so no message exceeds the transport limit.

Two things I'd like your read on:
- Byte-budget splitting at 2 MiB... happy to drop it for a leaner first cut if you'd rather.
- Scope grouping keys off the (target, entity-key) context slice, unordered today since the engine only ever emits one key. Will sort if you want that guaranteed.

What issue does this PR close?

- Closes #1902

How are these changes tested?

- Unit tests cover scope grouping, default-on batching, timer flush, byte-split, and small-batch_size immediate flush. cargo test/clippy/fmt clean across the workspace.

Are there any user-facing changes?

Yes... batch_size and max_batch_duration are new config fields. batch_size defaults to 64 KiB, so batching is on out of the box; set it to 1 for the old per-record behavior.

Changelog

- [x] Added a .chloggen/*.yaml entry
- [ ] This PR is a chore (indicated in title)
- [ ] This is a documentation-only PR.


rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
```diff
@@ -588,50 +588,91 @@ fn encode_any_value(
...
 ) {
     buf.clear();
+    if events.is_empty() {
+        return;
+    }
+
+    // Bucket event indices by scope identity = (target, entity-key context).
+    // A BTreeMap is the codebase's grouping idiom (cf. `build_attribute_index`)
+    // and orders scopes by that key: target first, then the entity-key slice.
+    let mut groups: BTreeMap<(&str, &[EntityKey]), Vec<usize>> = BTreeMap::new();
```
lalitb: When batching is off, we still go through this batch path for every single log record. That means the default path now builds a `BTreeMap` and a `Vec` for a one-record batch, even though users did not enable batching.

Can we keep the old single-record fast path when `batch_size` is not set, and only use this grouping path when batching is enabled?

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
lalitb, line 270: This path was fine before, but now batching can leave logs buffered in `batch`, so returning here without flushing can drop up to `threshold - 1` records.

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
```diff
@@ -161,42 +229,198 @@ impl local::Receiver<OtapPdata> for InternalTelemetryReceiver {
...
-        let mut buf = ProtoBuffer::with_capacity(512);
+        let mut start = 0;
+        while start < batch.len() {
+            // Take records until the estimated size would exceed the budget,
+            // always including at least one: a record's pre-encoded body is
+            // size-bounded, so a single one always fits.
+            let mut end = start;
+            let mut chunk_bytes = resource_bytes.len();
+            while end < batch.len() {
+                let record_bytes = batch[end].record.body_attrs_bytes.len() + RECORD_FRAMING_BYTES;
```
lalitb: Is this intended to be only a best-effort size estimate? It counts resource bytes and log body bytes, but does not include scope names, scope attributes, ScopeLogs framing, or protobuf length overhead, so the final encoded message can still exceed `MAX_BATCH_BYTES`.

Is there a reason those bytes are not included in the split calculation?

rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
```diff
@@ -588,50 +588,91 @@ fn encode_any_value(
...
+    for (i, event) in events.iter().enumerate() {
+        let target = event.record.callsite().target();
+        let context = event.record.context.as_slice();
```
lalitb: Is this entity-key ordering guaranteed to be stable for the same logical scope? If not, two records with the same ke s in a different order would end up in separate `ScopeLogs`.

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
```diff
@@ -161,42 +229,198 @@ impl local::Receiver<OtapPdata> for InternalTelemetryReceiver {
...
+                            Self::flush_batch(&effect_handler, &mut batch, &resource_bytes, &mut scope_cache).await?;
                             return Ok(TerminalState::default());
                         }
                     }
                 }
+
+                // Wake the loop when the deadline arrives so the top-of-loop check
+                // flushes (covers the idle case with no incoming logs). Disabled
+                // while the batch is empty, so an idle receiver never wakes.
+                _ = tokio::time::sleep_until(batch_deadline.unwrap_or_else(tokio::time::Instant::now)),
```
lalitb, line 306: Can we add tests for the new timer flush and byte-splitting paths? The current tests cover shutdown flush and default per-record behavior, but these two new paths are where most batching bugs would show up.

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
```diff
@@ -161,42 +229,198 @@ impl local::Receiver<OtapPdata> for InternalTelemetryReceiver {
...
                     match result {
                         Ok(ObservedEvent::Log(log_event)) => {
-                            if let Some(log_tap) = log_tap.as_ref() {
-                                log_tap.record(log_event.clone());
+                            if batch.is_empty() {
+                                batch_deadline = Some(tokio::time::Instant::now() + flush_period);
+                            }
+                            Self::buffer_log(&log_tap, &mut batch, log_event);
+                            if batch.len() >= threshold {
+                                Self::flush_batch(&effect_handler, &mut batch, &resource_bytes, &mut scope_cache).await?;
```
lalitb: Since grouping and encoding the whole batch happen inline in the receiver task, do we want any guardrails on `batch_size` or encoded work per flush? If this is intentionally left flexible, I’m fine with tracking it separately rather than blocking this PR.

rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
AvinashDevX: Good call, fixed in 398632f. The single-record path encodes directly via a shared encode_scope_logs helper now, and the grouping map is only built for 2+ events, so the default flow no longer allocates a map or size vector.

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
AvinashDevX: Fixed in 398632f: the control-channel error arm now does a best-effort flush before returning, so buffered records aren't dropped.

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
AvinashDevX: Yes, intended as best-effort. You're right that a message can run a bit over MAX_BATCH_BYTES, since the estimate leaves out scope attributes, ScopeLogs framing, and length prefixes. Two reasons it's fine: those bytes aren't cheaply available at split time (scope attrs need a cache lookup per record), and MAX_BATCH_BYTES is a soft target sitting ~256x under the actual OTLP encode limit, so the omitted overhead can't realistically breach the real limit. I reworded the comment to call it a heuristic rather than a guarantee. If you'd prefer it exact, I can fold the cached scope-attribute bytes into the total.

rust/otap-dataflow/crates/telemetry/src/self_tracing/encoder.rs
AvinashDevX: In practice the context only carries a single entity key today, so this can't actually split a logical scope right now. You're right that it's a latent risk if multiple keys ever arrive with unstable ordering. I can make it order-independent by sorting the context into the grouping key (cheap, since contexts are tiny) if you'd like that guarantee up front rather than later.

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
AvinashDevX: Added in 398632f: a deterministic chunk_end unit test for the split math, and a timer-flush test that recvs the message before shutdown so it proves the timer fired rather than the drain.

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
AvinashDevX: The byte-split already bounds the per-message work: each ExportLogsRequest stays near 1 MiB regardless of batch_size, so a large value produces more messages rather than one giant encode, with send_message awaiting between them. Grouping is O(n log n) on the batch. So there's no unbounded single encode, though a very large batch_size does mean proportionally more work per flush on the receiver task. I'd rather not pick an arbitrary cap here; happy to add a configurable upper bound as a follow-up if you'd like one.

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
```diff
@@ -24,23 +24,64 @@ use otap_df_otap::pdata::{Context, OtapPdata};
...
+fn default_max_batch_duration() -> Duration {
+    Duration::from_millis(DEFAULT_MAX_BATCH_DURATION_MS)
+}
+
+/// Per-message byte budget for splitting a batch, so each emitted
+/// `ExportLogsRequest` stays well under typical gRPC/OTLP size limits. The split
+/// is a heuristic: it sums each record's pre-encoded body plus framing and does
+/// not count all protobuf overhead (e.g. scope attributes), so it is a safety
+/// margin, not a hard cap. Internal log records are small, leaving wide headroom.
+const MAX_BATCH_BYTES: usize = 1 << 20; // 1 MiB
```
jmacd: Since this is a maximum, I support slightly larger: the logs encoding mechanism uses variable-width placeholders, and 21 bits makes for max-3 varint placeholders.

```suggestion
const MAX_BATCH_BYTES: usize = 1 << 21; // 2 MiB
```

jmacd (approved):
Looks good to me. @lalitb please take another look.
Thanks @AvinashDevX.

jmacd:
@AvinashDevX how would you feel setting the default batch size to 64KiB instead of disabled by default?

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
AvinashDevX: Good call, I hadn't clocked the varint placeholder angle. You're right... Applying it.

AvinashDevX:
> @AvinashDevX how would you feel setting the default batch size to 64KiB instead of disabled by default?

@jmacd 
Sounds good. Since batch_size is a record count today, I'll switch it to a byte size defaulting to 64 KiB, with max_batch_duration still bounding latency for low-volume streams. Shout if you'd rather keep the count and add bytes alongside instead, otherwise I'll take it.

jmacd:
@AvinashDevX thanks for iterating on this. Two items before I merge.

1. Are you still planning to switch `batch_size` from a record count to a byte size defaulting to 64 KiB, as we discussed on July 1? I don't see it in the current diff. I'd like the receiver to batch by default so downstream messages are reasonably sized without extra configuration.

2. Could you confirm the early-return path lalitb flagged in `internal_telemetry_receiver/mod.rs`? His concern is that returning there without flushing can drop up to `threshold - 1` buffered records, and I want to be sure a shutdown or channel close on that branch cannot lose a partial batch. Tests covering the timer-flush and byte-split paths would also help lock this down.

Happy to re-review once those are addressed. Thanks!

AvinashDevX:
@jmacd Pushed in 55f779e50: batch_size is a byte threshold now, defaults to 64 KiB, batches out of the box without any config. max_batch_duration still bounds latency for quiet streams; set batch_size to 1 if you ever want the old per-record flush back.

The control-channel error arm's best-effort flush from 398632f is unchanged and still holds with the byte-based threshold. Added a test proving the byte threshold flushes on its own via a bounded concurrent recv, not just riding the shutdown drain, on top of the existing timer and byte-split tests.

Branch was a couple weeks behind main so I merged that in too. Ready whenever you have time.

drewrelmas:
@AvinashDevX this looks good to me - before merging can you please update the PR title and description to denote batching is now on by default for future viewing in Git history?

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
```diff
@@ -24,23 +24,83 @@ use otap_df_otap::pdata::{Context, OtapPdata};
...
+    /// [`record_bytes`]) before emitting one batched `ExportLogsRequest`,
+    /// grouping records that share a scope. Default 64 KiB. A configured `0`
+    /// is rejected.
+    #[serde(default = "default_batch_size")]
+    batch_size: NonZeroUsize,
+
+    /// How long a record may wait in a partial batch before being flushed.
+    /// Default 200ms.
+    #[serde(with = "humantime_serde", default = "default_max_batch_duration")]
+    max_batch_duration: Duration,
```
drewrelmas, line 90: It isn't immediately clear that `batch_size` refers to bytes.

Could we make this config a little more similar to the batch processor (with `min_size`, `max_size` and `sizer`)? https://github.com/open-telemetry/otel-arrow/tree/main/rust/otap-dataflow/crates/core-nodes/src/processors/batch_processor

AvinashDevX: Done... renamed batch_size -> min_size and exposed max_size (the previously-hardcoded 2 MiB split cap) to match the batch processor, with max_size >= min_size enforced at config validation; skipped sizer since this receiver only emits OTLP bytes, so it'd have a single legal value.

rust/otap-dataflow/crates/core-nodes/src/receivers/internal_telemetry_receiver/mod.rs
```diff
@@ -200,3 +387,303 @@ impl InternalTelemetryReceiver {
...
+
+    #[test]
+    fn small_max_size_splits_batch_into_multiple_messages() {
+        let test_runtime = TestRuntime::new();
+        // min_size never reached, so records accumulate through the shutdown
+        // drain; max_size of 1 byte then forces the final flush to split them,
+        // one record per message, independent of min_size.
+        let config = Config {
+            min_size: NonZeroUsize::new(10_000).unwrap(),
+            max_size: NonZeroUsize::new(1).unwrap(),
```
ThomsonTan: This receiver-level split test bypasses production validation by passing a raw `Config` to `new_with_telemetry`. The public config path rejects `min_size = 10_000, max_size = 1`, so a passing test does not show that a valid receiver configuration can reach the split path.

Probably construct the test config through `parse_config` and use a reachable overshoot case, for example `min_size == max_size == 2 * record_bytes - 1`, then verify that the second record produces multiple messages.

AvinashDevX: Good catch. Reworked it through parse_config with min_size == max_size a byte under two records, so the second one overshoots max_size and splits... reachable from a valid config now.

jmacd:
@AvinashDevX will you please resolve? Thanks!

jmacd:
Thank you, @AvinashDevX. I ported the receiver-side batching design and test strategy from this PR into the broader internal-log prototype in #3882.

The replacement keeps the 64 KiB/200 ms default batching behavior and scope grouping, adapts it to the current ITR metrics and shutdown architecture, adds a fixed 2 MiB retained-size ceiling, and supports batched `arrow_extended` stack tables. Commit `1a6b9a1f6` retains Avinash Sharma as a co-author, and #3882 explicitly credits this PR.

Closing this older branch in favor of #3882 so review can continue on the current codebase and combined end-to-end design.

PR #3382, opened 2026-06-29
feat(ctl): allow library embedders to override CLI branding

### LINKED ISSUE #3370: otap-df-ctl: make CLI identity (binary name + envelope schemaVersion) configurable for library embedders
### Pre-filing checklist

### Component(s)

`otap-df-ctl` is structured as a thin binary over a reusable library
(`run_with_terminal_and_diagnostics`, etc.), which makes it attractive to
**embed the library inside another binary** to expose OTAP dataflow control as a
subcommand of a larger tool.

When embedded, the library's output is hard-coded to the `dfctl` identity in two
user-visible ways:

1. **Binary name** — help text, shell completions, and generated command IDs say
   `dfctl`, even though the user invoked a different binary.
2. **Machine-readable envelopes** — the agent/stream/error JSON envelopes carry
   `"schemaVersion": "dfctl/v1"`, so automation consuming an embedder's output
   keys on a version that misattributes the output to `dfctl`.

A standalone `dfctl` user is unaffected; this only impacts library embedders.

### Proposed Solution

Introduce an **additive, opt-in** `Branding` value (binary name + envelope schema
version) that an embedder can supply, defaulting to the current `dfctl` /
`dfctl/v1` so the standalone binary and existing public API are unchanged. Expose
it via a new branded entry point (e.g.
`run_with_terminal_and_diagnostics_branded(..., branding)`); the existing entry
points delegate with `Branding::default()`.

Initial scope is intentionally minimal — **binary name** and **envelope
`schemaVersion`**. Possible follow-ups (out of scope here): schema-catalog IDs
(`dfctl.*.v1`), schema `$id` URLs/titles, `DFCTL_*` environment-variable
prefixes, and the interactive TUI command hints.

### Alternatives Considered

- **Embedders depend on the typed admin SDK (`otap-df-admin-api`) directly** and
  build their own output layer — viable, but loses the ready-made
  command/render/troubleshoot layers; many embedders just want the CLI surface
  with correct identity.
- **Fork/patch the constant downstream** — brittle and diverges from upstream.
- **Change existing entry-point signatures to take `Branding`** — rejected to
  avoid a breaking change; the additive `*_branded` overload keeps the public API
  stable.

### Additional Context

Happy to open a draft PR implementing the minimal scope for discussion.

daviddahl:
# Change Summary

Add an additive, opt-in `Branding` to `otap-df-ctl` so a downstream binary that
**embeds the library** can emit help, shell completions, and machine-readable
JSON envelopes under its own identity instead of `dfctl`.

- New `Branding` type (`bin_name` + envelope `schema_version`) and a new
  `run_with_terminal_and_diagnostics_branded(..., branding)` entrypoint. The
  branding is installed process-wide (set-once, mirroring the existing
  crypto-provider `OnceLock`); the existing entrypoints leave it unset and use
  `Branding::default()`.
- The binary name and envelope `schemaVersion` reads are routed through the
  active branding (lib diagnostics, `render/output` envelopes, the error
  envelope, and completions).
- **The standalone `dfctl` binary and all existing public entrypoints/signatures
  are unchanged** — the default branding reproduces the current `dfctl` /
  `dfctl/v1` identity exactly.

Scope is intentionally minimal — binary name and envelope `schemaVersion`.
Out of scope here (possible follow-ups): schema-catalog identifiers
(`dfctl.*.v1`), schema `$id` URLs/titles, `DFCTL_*` environment-variable
prefixes, and the interactive TUI command hints.

* Closes #3370

- All existing `dfctl` output/branding tests remain green via
  `Branding::default()`.
- New integration test (`crates/ctl/tests/branding.rs`) drives the branded
  entrypoint with a non-default `Branding` and asserts the emitted
  `schemaVersion` is overridden. It uses a connection-free command
  (`commands --output agent-json`) so no admin server is required, and runs as a
  separate integration-test process so the process-global branding does not
  affect other tests.
- `cargo xtask check` (structure checks, `cargo fmt --all`,
  `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test
  --workspace`) passes.

Yes — a new public library entrypoint (`run_with_terminal_and_diagnostics_branded`)
and a new public `Branding` type for library embedders. No change to the
standalone `dfctl` binary's behavior or output.

### Changelog

* [x] Added a `.chloggen/*.yaml` entry.


rust/otap-dataflow/crates/ctl/src/branding.rs
```diff
@@ -0,0 +1,70 @@
...
+use std::sync::OnceLock;
+
+/// Identity strings used in the CLI's user-visible and machine-readable output.
+///
+/// Defaults reproduce the standalone `dfctl` identity. The fields are
+/// `&'static str` because a binary's identity is fixed for the life of the
+/// process.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct Branding {
+    /// Installed command name used in help, completions, and generated command
```
lalitb: This says `bin_name` applies to generated command metadata, but `commands --output json` still appears to [use](https://github.com/open-telemetry/otel-arrow/blob/bdd514ccf95d7fbac61746c699c0ebc295eed900/rust/otap-dataflow/crates/ctl/src/commands/catalog.rs#L93) fixed `BIN_NAME`/`dfctl` values inside the catalog. Should the catalog use the active branding too, or should this doc/PR scope be narrowed?

Also good to extend the  branding test to validate this.

rust/otap-dataflow/crates/ctl/src/lib.rs
```diff
@@ -55,6 +57,24 @@ pub async fn run_with_terminal(
...
 
+/// Executes a parsed command with a caller-supplied [`Branding`].
+///
+/// This is the entrypoint for library embedders that need the CLI's
+/// user-visible identity and machine-readable output envelopes to reflect their
+/// own binary rather than `dfctl`. The branding is installed process-wide
+/// (set-once) before the command runs; the non-branded entrypoints leave it
+/// unset and therefore use [`Branding::default`].
+pub async fn run_with_terminal_and_diagnostics_branded(
+    cli: ParsedCli,
```
lalitb, line 73: Seems this method receives an already parsed `Cli`, so the branding is installed too late for Clap-generated help, usage text, and parse errors. Should the PR/changelog wording around branded help be narrowed, or do we need a branded parse/command construction path for embedders?

rust/otap-dataflow/crates/ctl/src/branding.rs
daviddahl: Routed the command catalog through branding::active().bin_name:  the binary field, command_id (<name>.pipelines.get), command_line/example prefixes, and the human-render headers now follow the active branding. 

Also added an integration test that runs commands --output json under embedder branding and asserts binary == "embedder" and command IDs are prefixed embedder. 

(Schema identifiers like dfctl-command-catalog/v1 and dfctl.*.v1 are still deferred — separate namespace surface.)

rust/otap-dataflow/crates/ctl/src/lib.rs
daviddahl: Added Cli::command_branded(&Branding) and Cli::try_parse_from_branded(args, &Branding) so an embedder can produce a clap Command with their own binary name before parsing: --help, --version, and parse errors now all carry the embedder identity. 

Tested via command.get_name() and by asserting parse errors contain the branded name. The existing run_*_branded entrypoint installs runtime branding separately (same flow as before).

rust/otap-dataflow/crates/ctl/src/commands/catalog.rs
```diff
@@ -87,10 +88,11 @@ fn build_catalog() -> CommandCatalog {
         collect_command(command, &mut path, &mut commands);
     }
 
+    let bin_name = branding::active().bin_name;
     CommandCatalog {
         schema_version: SCHEMA_VERSION,
         generated_at: humantime::format_rfc3339_seconds(SystemTime::now()).to_string(),
-        binary: BIN_NAME,
+        binary: bin_name,
```
lalitb, line 95: Now that the catalog can use a custom `binary` name, the schema still says binary must always be `dfctl`. Should we update the schema too, or keep the catalog binary field fixed as `dfctl`?

rust/otap-dataflow/crates/ctl/src/args.rs
```diff
@@ -87,6 +87,32 @@ impl Cli {
...
+        itr: I,
+        branding: &crate::Branding,
+    ) -> Result<Self, clap::Error>
+    where
+        I: IntoIterator<Item = T>,
+        T: Into<OsString> + Clone,
+    {
+        Self::command_branded(branding)
+            .try_get_matches_from(itr)
+            .and_then(|m| Self::from_arg_matches(&m))
```
ThomsonTan: This branded parse path seems silently dropping agent-mode defaults, so `--agent` / `DFCTL_AGENT_MODE` becomes a no-op for embedders.

`try_parse_from_branded` finalizes with `from_arg_matches`, whereas every standalone entrypoint (`parse_effective`, `try_parse_effective_from`) finalizes with `from_arg_matches_with_effective_defaults`. The `agent` flag has no direct runtime consumer, its *only* effect is the effective-defaults pass. The runtime also destructures `agent: _` and ignores it, so nothing downstream compensates.

```suggestion
        Self::command_branded(branding)
            .try_get_matches_from(itr)
            .and_then(|m| Self::from_arg_matches_with_effective_defaults(&m))
```

daviddahl: Thank you for finding this gap. Cheers!

rust/otap-dataflow/crates/ctl/src/commands/catalog.rs
daviddahl: Thanks! Relaxed the schema: "binary" is now { "type": "string", "description": "Active binary name at catalog generation time." }

lalitb (approved):
Thanks.

jmacd (approved):

PR #3386, opened 2026-06-30
fix(durable_buffer): Port per-signal loss metrics from `ObserveCounter` to `Counter`

drewrelmas:
# Change Summary

Follow-up to:
- #3269

The recently added `durable_buffer_processor` `dropped_*` and `expired_*` per-signal metrics are declared as `ObserveCounter`s, which export asynchronously as cumulative lifetime totals:

https://github.com/open-telemetry/otel-arrow/blob/bdd514ccf95d7fbac61746c699c0ebc295eed900/rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs#L313-L315

However, in the current self-telemetry path, those instruments are emitted as `Gauge`s even if the metrics reader requests `Delta` temporality:

https://github.com/open-telemetry/otel-arrow/blob/bdd514ccf95d7fbac61746c699c0ebc295eed900/rust/otap-dataflow/crates/telemetry/src/metrics/dispatcher.rs#L137-L140

The result for an operator - it is impossible to get per-interval values for these metrics.

While we consider a larger change for `ObserveCounter` to properly respect exporter temporality (either through a different usage of the OTel SDK or a non-SDK ITS using `temporal_reaggregation_processor`), a short-term fix is simply converting these metrics to use the `Counter` instrument instead.

Per comments, opened #3397 as a follow-up.

* Related to #3117

Unit tests

Yes, the per-signal drop/expire metrics are now reported over interval instead of lifetime.

### Changelog

* [x] Added a `.chloggen/*.yaml` entry


rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs
```diff
@@ -333,6 +334,31 @@ pub struct DurableBufferMetrics {
...
+    #[metric(unit = "{log_record}")]
+    pub expired_log_records_delta: Counter<u64>,
+
+    /// Spans lost to expired segments (max_age) in this interval.
+    #[metric(unit = "{span}")]
+    pub expired_spans_delta: Counter<u64>,
+
+    /// Metric data points lost to expired segments (max_age) in this interval.
+    #[metric(unit = "{data_point}")]
+    pub expired_metric_datapoints_delta: Counter<u64>,
```
utpilla: I agree with the diagnosis in the description: the real issue is that the dispatcher exports cumulative `ObserveCounter`s as gauges even when the reader asks for Delta, so the temporality is ignored. That is the bug and it is what should be fixed (make the dispatcher honor reader temporality, either via observable instruments so the SDK does the cumulative to delta conversion, or by doing the cumulative to delta conversion ourselves in the dispatcher). That single fix makes every `ObserveCounter` delta-capable with no per-metric changes.

Even for an interim fix though, why not update the existing metrics instead of adding new `*_delta` twins:

1. These metrics were recently added, right? So there shouldn't be any concern for breaking changes.
2. The source is already delta-native. These metrics come from `engine.drain_*_pending()` (per-interval counts); the processor just sums them into `total_*` to feed the `ObserveCounter`.
3. You have already written the delta half (`dropped_log_records_delta.add(count)`). Instead of keeping it as a twin, make it the metric: change `dropped_log_records` from `ObserveCounter` to `Counter`, keep the `.add(count)`, and drop the `ObserveCounter` plus the `total_*` bookkeeping. A delta exporter then gives per-interval values and a cumulative exporter gives the total as a proper Sum.

Also, we should track the dispatcher temporality fix as a follow-up.

cijothomas: +1

rust/otap-dataflow/.chloggen/durable-buffer-delta-loss-counters.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: pipeline
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add per-signal delta (point-in-time) data-loss counters to the durable buffer processor.
```
cijothomas: _delta (point-in-time)_ - this wording is not clear.

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs
drewrelmas: Thanks for the advice @utpilla, that makes sense. I'll open another issue for the dispatcher temporality fix and take your advice of converting existing metrics to `Counter` for the time being.

rust/otap-dataflow/.chloggen/durable-buffer-delta-loss-counters.yaml
drewrelmas: Followed the suggestion above, created new changelog entry.

rust/otap-dataflow/crates/core-nodes/src/processors/durable_buffer_processor/mod.rs
drewrelmas: Opened #3397

jmacd: Yes, I agree. It takes memory in the reader, of course, to translate the cumulative back to delta. I see in the source that the value is set by a `self.total_log_records` value, meaning the struct maintains somewhere and does not increment a counter. (I did not dig further to understand why it doesn't count directly.)

rust/otap-dataflow/.chloggen/durable-buffer-loss-counters-delta.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: pipeline
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Report durable buffer data-loss metrics as synchronous Counters so they honor the reader's temporality.
```
jmacd: I still support this PR -- meaning I think we shouldn't report cumulative unless we are required to.
```suggestion
note: Report durable buffer data-loss metrics as synchronous Counters, lowers cost to export delta temporality.
```

jmacd (approved):

Issue #3387, opened 2026-06-30
OpAMP Agent Controller Extension


albertlockett:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Proposed Solution

Implement the ability for the OTAP Dataflow Engine to act as an OpAMP Agent. At a minimum, we should support having engine configuration be pushed by a remote server and the engine will report to this server its health status.

I will create a PR with a design document containing the initial implementation as I feel it may be easier to comment on the that than a github issue.

albertlockett:
link to original slack thread: https://cloud-native.slack.com/archives/C08RRSJR7FD/p1782439413433189

jmacd:
See https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/extension/healthcheckv2extension

PR #3388, opened 2026-06-30
docs: opamp design doc

albertlockett:
# Change Summary

Adds a design document discussing initial implementation of OpAMP Agent controller extension which can be used to receive configuration from a remote OpAMP server and provide status status to the remote server.

The purpose of the document is mostly to align on the overall initial implementation and specifically align on structure of the configuration & messages plus how health/status information is derived.

* Relates to #3387

n/a

no

 

### Changelog

* [x] This is a documentation-only PR.

### DIFF
--- a/rust/otap-dataflow/docs/opamp.md
+++ b/rust/otap-dataflow/docs/opamp.md
@@ -233,8 +233,8 @@ engine:
 The configuration in the `tls` block will use the structure defined in [`otap_df_config::tls::TlsClientConfig`](
 https://github.com/open-telemetry/otel-arrow/blob/f45dc00642f0187020a4a46de796a07bf368d443/rust/otap-dataflow/crates/config/src/tls.rs#L49-L92)
 
-If the key is supplied without the certificate, or the certificate is 
-configured without the key, configuration will be considered invalid at 
+If the key is supplied without the certificate, or the certificate is
+configured without the key, configuration will be considered invalid at
 startup.
 
 If the `tls` section of the config is not supplied, messages will be exchanged
@@ -611,7 +611,13 @@ will use the following rules:
 
 #### Full Pipeline Status
 
-The implementation will advertise a custom capability representing its ability
+The component health module presents an aggregate view of the health of each
+group and pipeline, but the detailed pipeline status contains additional
+information including specifically in which phase each instance of each
+pipeline is in.
+
+For servers that need this level of insight into the status of the pipelines
+the implementation will advertise a custom capability representing its ability
 to provide the full configuration status via a custom message.
 
 Proposed capability FQDN: `io.open-telemetry.arrow-dfe.pipeline-status/v1`.
--- a/rust/otap-dataflow/docs/opamp.md
+++ b/rust/otap-dataflow/docs/opamp.md
@@ -197,7 +197,7 @@ full state, including the following fields:
   - Reports Effective Config
   - Reports Health
   - Reports Remote Config
-  - Reports Heart Beat
+  - Reports Heartbeat
 - `health` - see section below on server health
 - `flags` - `Unset`
 - `custom_capabilities` - list capability to send full state as custom message
@@ -238,7 +238,7 @@ it should handle:
 In any case where a message is ignored (especially in error cases), appropriate
 telemetry should be emitted.
 
-##### Send periodic heart beats
+##### Send periodic heartbeats
 
 The client should send periodic heartbeats to the server at regular intervals.
 
--- a/rust/otap-dataflow/docs/opamp.md
+++ b/rust/otap-dataflow/docs/opamp.md
@@ -86,97 +86,98 @@ The URN used to identify the controller extension will be
 # example DFE config
 engine:
   controller:
-    opamp:
-      type: urn:otel:controller_extension:opamp
-      config:
-        # The URL of the OpAMP server, including the scheme and path.
-        #
-        # The scheme will be used to determine whether communication
-        # is done over WebSocket or HTTP. E.g. use "ws://..." for websocket
-        # and http:// for http.
-        #
-        # Required
-        endpoint: ws://127.0.0.1:4320/v1/opamp
-
-        # mTLS configuration
-        #
-        # Optional: if not provided, messages will be sent as plaintext
-        tls:
-          # Trust settings for verifying the server certificate
-          ca_file: "./my-certs/ca.crt"
-          include_system_ca_certs_pool: false
-
-          # Client identity for mTLS (these make it mTLS instead of just TLS)
-          cert_file: "./my-certs/client.crt"
-          key_file: "./my-certs/client.key"
-
-
-        # Configuration for heart-beat timing.
-        #
-        # Optional (default = 30s)
-        heart_beat_interval: "10s"
-
-        # Options for configuring exponential backoff for initial Websocket
-        # connection.
-        #
-        # Each exponential backoff computed as
-        # min(last_backoff * factor, max) + jitter
-        #
-        # Optional
-        connect_retry:
-          # initial backoff
+    extensions:
+      opamp:
+        type: urn:otel:controller_extension:opamp
+        config:
+          # The URL of the OpAMP server, including the scheme and path.
           #
-          # Optional (default = 250ms)
-          initial: 1s
+          # The scheme will be used to determine whether communication
+          # is done over WebSocket or HTTP. E.g. use "ws://..." for websocket
+          # and http:// for http.
+          #
+          # Required
+          endpoint: ws://127.0.0.1:4320/v1/opamp
 
-          # maximum backoff
+          # mTLS configuration
+          #
+          # Optional: if not provided, messages will be sent as plaintext
+          tls:
+            # Trust settings for verifying the server certificate
+            ca_file: "./my-certs/ca.crt"
+            include_system_ca_certs_pool: false
+
+            # Client identity for mTLS (these make it mTLS instead of just TLS)
+            cert_file: "./my-certs/client.crt"
+            key_file: "./my-certs/client.key"
+
+
+          # Configuration for heart-beat timing.
           #
           # Optional (default = 30s)
-          max: 30s
+          heart_beat_interval: "10s"
 
-          # exponential backoff increase factor
+          # Options for configuring exponential backoff for initial Websocket
+          # connection.
           #
-          # Optional (default = 2.0)
-          factor: 2.0
+          # Each exponential backoff computed as
+          # min(last_backoff * factor, max) + jitter
+          #
+          # Optional
+          connect_retry:
+            # initial backoff
+            #
+            # Optional (default = 250ms)
+            initial: 1s
 
-        # Options for configuring exponential backoff retry for requests.
-        #
-        # Used in the event that the initial request to the server fails or
-        # when the server responds with an error response type Unavailable but
-        # does not specify retry_info in the response details
-        #
-        # Optional
-        request_retry: {
-          #  ... same exponential retry options as connect_retry above
-        }
+            # maximum backoff
+            #
+            # Optional (default = 30s)
+            max: 30s
 
-        # Options for engine reconciliation
-        reconcile:
-          # timeout for reconcile - Optional (default = 30)
-          step_timeout_secs: 10
+            # exponential backoff increase factor
+            #
+            # Optional (default = 2.0)
+            factor: 2.0
 
-          # timeout for pipeline drain - Optional (default = 30)
-          drain_timeout_secs: 10
+          # Options for configuring exponential backoff retry for requests.
+          #
+          # Used in the event that the initial request to the server fails or
+          # when the server responds with an error response type Unavailable but
+          # does not specify retry_info in the response details
+          #
+          # Optional
+          request_retry: {
+            #  ... same exponential retry options as connect_retry above
+          }
 
-          # timeout for pipeline delete - Optional (default = 30)
-          delete_timeout_secs: 10
+          # Options for engine reconciliation
+          reconcile:
+            # timeout for reconcile - Optional (default = 30)
+            step_timeout_secs: 10
 
-        # Manually specify an instance_uuid.
-        #
-        # Optional - if not provided, a UUID v7 will be generated
-        instance_uuid: "..."
+            # timeout for pipeline drain - Optional (default = 30)
+            drain_timeout_secs: 10
 
-        # Attributes describing agent
-        #
-        # Optional
-        agent_description:
-          identifying_attributes:
-            "attr.key": "attr value"
-            # ...
+            # timeout for pipeline delete - Optional (default = 30)
+            delete_timeout_secs: 10
 
-          non_identifying_attributes:
-            "attr.key2": "attr value"
-            # ...
+          # Manually specify an instance_uuid.
+          #
+          # Optional - if not provided, a UUID v7 will be generated
+          instance_uuid: "..."
+
+          # Attributes describing agent
+          #
+          # Optional
+          agent_description:
+            identifying_attributes:
+              "attr.key": "attr value"
+              # ...
+
+            non_identifying_attributes:
+              "attr.key2": "attr value"
+              # ...
 ```
 
 ## Protocol Implementation
--- a/rust/otap-dataflow/docs/opamp.md
+++ b/rust/otap-dataflow/docs/opamp.md
@@ -5,7 +5,7 @@ configured via an embedded OpAMP Agent, implemented as a controller extension.
 
 ## Problem
 
-As the machinations of those desiring to control telemetry flow the become more
+As the machinations of those desiring to control telemetry flow become more
 sophisticated, the concept of building some fleet management becomes
 increasingly appealing.
 
@@ -14,8 +14,8 @@ plane without having to design, from scratch, a new protocol for sending
 configuration or receiving status.
 
 OpAMP provides such a protocol, and the solution to this problem will be to
-add the capability for dataflow engine to act as a OpAMP agent, receiving
-configuration from some remote server while emitting its state & health.
+add the capability for the dataflow engine to act as an OpAMP agent, receiving
+configuration from some remote server while emitting its state and health.
 
 ## Background
 
--- a/rust/otap-dataflow/docs/opamp.md
+++ b/rust/otap-dataflow/docs/opamp.md
@@ -465,6 +465,7 @@ The payload here is the bytes from a JSON serialized status snapshot (e.g., the
 ### Metrics
 
 The controller extension implementation should produce metrics:
+
 - Count successful `AgentToServer` messages transmitted
 - Count failed `AgentToServer` messages transmitted (due to TCP or HTTP errors)
 - Count `ServerToAgent` successfully handled
--- a/rust/otap-dataflow/docs/opamp.md
+++ b/rust/otap-dataflow/docs/opamp.md
@@ -244,7 +244,8 @@ in plaintext.
 
 ### Capability Configuration
 
-
+Capabilities which control including what the agent sends and what fields it
+accepts from the remote server, can be toggled via configuration.
 
 ### Instance UID resolution
 
@@ -264,6 +265,11 @@ Step 3: If the previous steps were not able to resolve the instance_uid, a new
 instance UID will be generated from a generated UUIDv7.
 
 ### Client Behaviour and Message Flow
+
+Note - this section is written in such a way as to assume that all capabilities
+are enabled In each section, if certain capabilities controlling which fields
+a message should contain, the field should be omitted/ignored in the sent/
+received message unless otherwise stated.
 
 #### 1. Initial Message
 
@@ -601,7 +607,9 @@ will use the following rules:
   `running`
 - Otherwise the group status will be `degraded`.
 
-### Custom Messages & Capabilities: Full Pipeline Status
+### Custom Messages & Capabilities: 
+
+#### Full Pipeline Status
 
 The implementation will advertise a custom capability representing its ability
 to provide the full configuration status via a custom message.
@@ -624,6 +632,94 @@ AgentToServer {
 The payload here is the bytes from a JSON serialized status snapshot (e.g., the
 `HashMap<PipelineKey, PipelineStatus>` returned from
 `ObservedStateHandle::status_snapshot`).
+
+#### Imperative Configuration Commands
+
+The primary mechanism for the server pass configuration to the agent will be as
+described above which is to pass the entire configuration declaratively in the
+`remote_config` config map.
+
+However, there may be instances where the server wishes to change the engine
+configuration using imperative commands such as starting, stopping, updating
+or reconfiguring specific pipelines.
+
+For this, a custom message will be supported that can contain multiple
+operations to perform on certain pipelines.
+
+Proposed capability FQDN: `io.open-telemetry.arrow-dfe.operations/v1`
+
+Custom message structure
+
+```rs
+ServerToAgent {
+  custom_message: CustomMessage {
+    capability: "io.open-telemetry.arrow-dfe.operations/v1",
+    r#type: "operations",
+    data: [...]
+  }
+}
+```
+
+The data will will be a serialized message with the following structure:
+
+```rs
+struct Operations {
+  /// Set of operations to apply
+  operations: Vec<Operation>
+
+  /// timeout for set of operations
+  timeout: Option<Duration>
+}
+
+struct Operation {
+  /// Unique identifier of the request. Can be used for observability purposes
+  /// and the reply to the server will contain the request ID.
+  request_id: Vec<u8>,
+
+  /// The command to perform
+  command: Command,
+
+  /// ID of pipeline on which to perform the command
+  pipeline_id: String,
+
+  /// ID of group to which the pipeline being acted upon 
+  pipeline_group_id: String,
+
+  /// timeout for this operation
+  timeout: Option<Duration>
+}
+
+/// The command to perform
+enum Command {
+  /// Start an existing pipeline
+  Start,
+
+  /// Create (and possibly start) a new pipeline
+  Create {
+    config: PipelineConfig,
+
+    /// whether to start the pipeline after it has been created
+    start: bool,
+  },
+
+  /// Update configuration of a pipeline
+  Update {
+    config: PipelineConfig
+  },
+
+  /// Drain & Shutdown a pipeline
+  Shutdown,
+  
+  /// Delete a pipeline. It will be drained and shutdown before it is deleted.
+  Delete,
+}
+```
+
+After attempting agent should reply with a message containing the request ID
+as well as the information about the result of the operation application,
+including whether the operation was able to be successfully applied, and
+if not, a description of the error that prevented the operation from being
+successful.
 
 ## Telemetry
 
@@ -669,15 +765,6 @@ authentication between the agent and server, using inspiration from the OpAMP
 spec (
 [this section](https://opentelemetry.io/docs/specs/opamp/#connection-settings-management)
 ). A followup design can be created for this.
-
-### Imperative Command Capability
-
-This design assumes the server will specify the remote configuration in full.
-E.g. the agent expects to receive the config declaratively.
-
-However, in the future we may want to expose a mechanism for servers to trigger
-the ad-hoc creation/reconfiguration/deletion of some pipeline. This could be
-supported through a custom agent capability / custom server message.
 
 ### Packages
 
--- a/rust/otap-dataflow/docs/opamp.md
+++ b/rust/otap-dataflow/docs/opamp.md
@@ -1,6 +1,6 @@
 # OpAMP Controller Extension
 
-This document describes the mechanism through which dataflow engine can be 
+This document describes the mechanism through which dataflow engine can be
 configured via an embedded OpAMP Agent, implemented as a controller extension.
 
 ## Problem
@@ -21,23 +21,25 @@ configuration from some remote server while emitting its state & health.
 
 ### OpAMP
 
-Full OpAMP documentation is available [here](https://opentelemetry.io/docs/specs/opamp/).
+See [Full OpAMP documentation](https://opentelemetry.io/docs/specs/opamp/) for
+more information.
 
-OpAMP (Open Agent Management Protocol) is a vendor-agnostic network protocol 
+OpAMP (Open Agent Management Protocol) is a vendor-agnostic network protocol
 existing as part of the Open-Telemetry ecosystem. It specifies a mechanism for:
+
 - _Servers_ to configure remote _Agents_
 - Agents to report their status to the Server
 - Management of Agent downloadable _packages_.
 - Connection credential management (including client-side TLS)
 
 Agent to Server communication typically happens over a network where the Agent
-acts as a client, using either 
+acts as a client, using either
 [WebSocket](https://opentelemetry.io/docs/specs/opamp/#websocket-transport) or
 [plain HTTP](https://opentelemetry.io/docs/specs/opamp/#plain-http-transport)
 (plain meaning standard HTTP request-response model, not HTTP over plaintext).
 
 The messages exchanged are protobuf serialized `AgentToServer` and 
-`ServerToAgent` messages. 
+`ServerToAgent` messages.
 Find proto message definitions
 [here](https://github.com/open-telemetry/opamp-spec/blob/main/proto/opamp/v1/opamp.proto).
 
@@ -45,7 +47,7 @@ Find proto message definitions
 
 OTel Controller Extensions allow custom implementation of logic that interacts
 with the engine control API to reconcile configuration and report status. They
-are configurable at runtime and can be linked at compile time (similar to 
+are configurable at runtime and can be linked at compile time (similar to
 processors, exporters, receivers, etc.).
 
 See [#3263](https://github.com/open-telemetry/otel-arrow/issues/3263) and
@@ -56,7 +58,7 @@ See [#3263](https://github.com/open-telemetry/otel-arrow/issues/3263) and
 This design document only covers the implementation of OpAMP for exchanging
 engine configuration and pipeline status.
 
-OpAMP has additional capabilities such as server directed management of 
+OpAMP has additional capabilities such as server directed management of
 connection/credential settings and package management (which could be used
 for dynamic loading of WASM components, for example). These considerations
 are left as future design work.
--- a/rust/otap-dataflow/docs/opamp.md
+++ b/rust/otap-dataflow/docs/opamp.md
@@ -530,6 +530,8 @@ The controller extension implementation should produce metrics:
   mismatch, invalid proto encoding, etc.)
 - Count remote configs successfully reconciled
 - Count remote configs unsuccessfully reconciled
+- Count of times WebSocket connects/reconnects
+- Count of instances where WebSocket was unexpectedly closed by the server.
 
 ## Future Work
 
--- a/rust/otap-dataflow/docs/opamp.md
+++ b/rust/otap-dataflow/docs/opamp.md
@@ -117,7 +117,7 @@ engine:
           # Optional (default = 30s)
           heartbeat_interval: "10s"
 
-          # Options for configuring exponential backoff for initial Websocket
+          # Options for configuring exponential backoff for initial WebSocket
           # connection.
           #
           # Each exponential backoff computed as
--- a/rust/otap-dataflow/docs/opamp.md
+++ b/rust/otap-dataflow/docs/opamp.md
@@ -155,6 +155,13 @@ engine:
             # timeout for pipeline delete - Optional (default = 30)
             delete_timeout_secs: 10
 
+            # Whether to delete any pipelines that are missing from the config
+            # received from the server. When `true`, any currently running 
+            # pipeline not in the received config is drained/deleted. When 
+            # `false` received remote configs are treated as additive/partial
+            # updates.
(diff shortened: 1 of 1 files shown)


rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+
+The URN used to identify the controller extension will be
+`urn:otel:controller_extension:opamp` and it can be configured as follows:
+
+```yaml
+# example DFE config
+engine:
+  controller:
+    opamp:
+      type: urn:otel:controller_extension:opamp
```
lquerel: The config shape here should use `engine.controller.extensions.<id>` consistently, matching the later example in section `Agent Identity` and the controller extension model. Otherwise this looks like a dedicated `engine.controller.opamp` field rather than a registered controller extension.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+        # The scheme will be used to determine whether communication
+        # is done over WebSocket or HTTP. E.g. use "ws://..." for websocket
+        # and http:// for http.
+        #
+        # Required
+        endpoint: ws://127.0.0.1:4320/v1/opamp
+
+        # mTLS configuration
+        #
+        # Optional: if not provided, messages will be sent as plaintext
```
lquerel: It may be useful to specify TLS validation semantics here: server trust roots, whether system roots are included by default, whether `insecure_skip_verify` is allowed, and that partial mTLS config, cert without key or key without cert, must fail validation before startup.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+        # Optional (default = 30s)
+        heart_beat_interval: "10s"
+
+        # Options for configuring exponential backoff for initial Websocket
+        # connection.
+        #
+        # Each exponential backoff computed as
+        # min(last_backoff * factor, max) + jitter
+        #
+        # Optional
```
lquerel: I'd suggest making the retry/backoff contract more explicit: capped exponential backoff, jitter behavior, reset behavior after a successful connection, and how `Unavailable` responses interact with request retry versus full reconnect. This will make implementations more predictable.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+          # timeout for pipeline drain - Optional (default = 30)
+          drain_timeout_secs: 10
+
+          # timeout for pipeline delete - Optional (default = 30)
+          delete_timeout_secs: 10
+
+        # Manually specify an instance_uuid.
+        #
+        # Optional - if not provided, a UUID v7 will be generated
+        instance_uuid: "..."
```
lquerel: For stable fleet identity, should the spec include an optional persisted `instance_uid`/`instance_uuid` **file path**? Generating UUIDv7 is a good default, but persistence across process restarts seems important for control plane state continuity.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+        # Options for engine reconciliation
+        reconcile:
+          # timeout for reconcile - Optional (default = 30)
+          step_timeout_secs: 10
+
+          # timeout for pipeline drain - Optional (default = 30)
+          drain_timeout_secs: 10
+
+          # timeout for pipeline delete - Optional (default = 30)
+          delete_timeout_secs: 10
```
lquerel: Could the reconciliation config also document `delete_missing` behavior? In a full desired state reconciliation model, `delete_missing: true` would mean that any currently running group or pipeline absent from the remote config is drained/deleted. With `delete_missing: false`, omitted groups/pipelines are left running, which makes the remote config behave more like an additive or partial update.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+- If the field `remote_config` is set:
+  - If a valid config can be found in the config map and the hash does not
+    match the last applied hash, reconcile the new config while sending
+    appropriate replies (see section below on Engine Config Reconciliation)
+- If the flags `ReportFullState` and/or `ReportAvailableComponents` are set,
+  the next reply to the server should contain full state or available
+  components. If a new config is being reconciled, these can be included on
+  the message that is sent to the server letting it know config is being
+  applied. Otherwise, send an ad-hoc message immediately reporting these
+  fields.
```
lquerel, line 369: This section could define which message handling errors are non fatal versus session fatal. For example, a valid `ServerToAgent` message with a mismatched `instance_uid` can be safely ignored and counted as a rejected message, because the transport itself is still healthy. In contrast, malformed protobuf bytes or an unsupported WebSocket frame type may indicate a broken protocol exchange, so the client should either close the session and reconnect or clearly document why it continues.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+These should contain:
+
+- `instance_uid`
+- `sequence_num`
+- `capabilities`
+- `health` - computed from current pipeline status (see section below on
+  Health Resolution).
+- `remote_config_status` - computed from current pipeline status (see section
+  below on Status Resolution)
+- `custom_message` - full pipeline status from status snapshot
```
lquerel: The heartbeat already includes `remote_config_status`, could it explicitly state that heartbeats should carry the last known remote config hash and status even when no new config is being applied? That gives the server periodic confirmation that the agent is still running config hash X, not only that the agent is alive.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
+# OpAMP Controller Extension
+
+This document describes the mechanism through which dataflow engine can be
+configured via an embedded OpAMP Agent, implemented as a controller extension.
```
cijothomas: nit: Link to OpAMP spec repo.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+
+## Problem
+
+As the machinations of those desiring to control telemetry flow become more
```
cijothomas: nit: these are good motivations behind OpAMP itself and is better written in the OpAMP repo itself. For this repo, this is more about - OTel already defines a agent management protocol, and df_engine should adopt it.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+
+## Background
+
+### OpAMP
+
+See [Full OpAMP documentation](https://opentelemetry.io/docs/specs/opamp/) for
+more information.
+
+OpAMP (Open Agent Management Protocol) is a vendor-agnostic network protocol
+existing as part of the Open-Telemetry ecosystem. It specifies a mechanism for:
```
cijothomas:
```suggestion
existing as part of the OpenTelemetry ecosystem. It specifies a mechanism for:
```

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+are configurable at runtime and can be linked at compile time (similar to
+processors, exporters, receivers, etc.).
+
+See [#3263](https://github.com/open-telemetry/otel-arrow/issues/3263) and
+[#3281](https://github.com/open-telemetry/otel-arrow/issues/3281).
+
+## Scope
+
+This design document only covers the implementation of OpAMP for exchanging
+engine configuration and pipeline status.
```
cijothomas: could use some rephrasing.
1. Lets state explicitly if this is most one-way - df_engine reporting its status/health to a opamp server, or does it cover accepting config from opamp server and acting on it.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+
+OpAMP has additional capabilities such as server directed management of
+connection/credential settings and package management (which could be used
+for dynamic loading of WASM components, for example). These considerations
+are left as future design work.
+
+## Design Principles
+
+### Control Plane Implementation Agnostic
```
cijothomas, line 63: not sure how this title reflects our design.. Opamp is already agnostics to how the control is implemented - it can be any control plane, as long as they speak OpAMP protocol, we work with them.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+        # Manually specify an instance_uuid.
+        #
+        # Optional - if not provided, a UUID v7 will be generated
+        instance_uuid: "..."
+
+        # Attributes describing agent
+        #
+        # Optional
+        agent_description:
+          identifying_attributes:
```
cijothomas: should we include Resource/Entity of the engine itself by default here instead of expecting users to do it? They have already likely configured it in another section - could opamp get it from there?

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+
+#### 1. Initial Message
+
+The client begins by sending an initial `AgentToServer` message sending its
+full state, including the following fields:
+
+- `instance_uid` - from config if present, otherwise generated
+- `sequence_num` - zero if first message in sequence
+- `agent_description` - from configuration if present, otherwise omit
+- `capabilities` - list of supported capabilities:
```
cijothomas: This should come from config right? The operator setting up df_engine should be able to tell which capabilities to expose - some operators might just want to report health and nothing more. Some want to report effective config. Some want to accept remote config and apply etc.
Some users would want to use mTLS, but in the case it fails, it still would to report the bare minimum health to the opampserver.

Just some thoughts.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+  the message that is sent to the server letting it know config is being
+  applied. Otherwise, send an ad-hoc message immediately reporting these
+  fields.
+
+In any case where a message is ignored (especially in error cases), appropriate
+telemetry should be emitted.
+
+##### Send periodic heart beats
+
+The client should send periodic heartbeats to the server at regular intervals.
```
cijothomas, line 465: are these also `AgentToServer` itself?

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+              attr.key: "attr_value"
+              # ...
+            non_identifying_attributes:
+              attr.key2: "attr_value"
+              # ...
+```
+
+If `instance_uuid` is not specified, a uuid v7 will be created.
+Note that this logic only applies to bootstrap resolution of the
+`instance_uid` - the server may respond with a new agent identification to
```
cijothomas, line 515: I think we should have an option to persist this identity and use that if we crash and restart? Else each start comes up with new id, which is not likely the desired outcome?

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+    }
+}
+```
+
+The `config_hash` computation is performed by the server. The client does not
+depend on any particular implementation or algorithm, but the server should
+endeavour to choose the hash such that each new config has a unique hash value
+without collisions. The agent does not calculate hashes, it only stores and
+compares them. This is
+[recommended by the OpAMP spec as well](https://opentelemetry.io/docs/specs/opamp/#calculating-hashes).
```
lquerel, line 552: In addition to comparing hashes in memory, it may be worth specifying whether the last successfully applied config hash should be persisted. That would let an agent avoid unnecessary reconciliation after restart when the server resends the same config (stretch goal).

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+
+When the OpAMP agent controller extension receives a `ServerToAgent` message,
+if it contains a remote configuration whose `config_hash` does not match the
+last applied config hash, the new config will be applied to the engine.
+
+The agent will try to deserialize the remote config as an `OtelDataflowSpec`.
+If deserialization fails, it will respond to the server with an `AgentToServer`
+message with a `remote_config_status` containing a `FAILED` status.
+
+If deserialization of the new remote_config succeeds, the agent will immediately
```
lquerel, line 564: It may be worth explicitly requiring the client to send an `APPLYING` `remote_config_status` response immediately after the remote config has been parsed and accepted, but before calling `ControlPlane::reconcile_engine_config`. Reconciliation may take time because it can create, update, drain, or delete pipelines. Sending APPLYING first lets the server know the config was accepted and work has started, rather than treating the agent as silent or stalled.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+- Otherwise if all pipelines have status `running`, the group status will be
+  `running`
+- Otherwise the group status will be `degraded`.
+
+### Custom Messages & Capabilities: Full Pipeline Status
+
+The implementation will advertise a custom capability representing its ability
+to provide the full configuration status via a custom message.
+
+Proposed capability FQDN: `io.open-telemetry.otap-dfe.pipeline-status/v1`.
```
lquerel: As we renamed the project OTel Arrow Dataflow Engine (at least for now...), I suggest to use `io.open-telemetry.arrow-dfe.pipeline-status/v1`

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+```rs
+AgentToServer {
+  capabilities: ["io.open-telemetry.otap-dfe.pipeline-status/v1"],
+  custom_message: CustomMessage {
+    capability: "io.open-telemetry.otap-dfe.pipeline-status/v1",
+    r#type: "report",
+    data: [ ... ]
+  }
+}
+```
```
lquerel: [medium term] The custom status capability should probably have a stable JSON schema. Even if the payload mirrors an internal status snapshot initially, documenting the wire shape avoids coupling external controllers to internal Rust model drift.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+
+The payload here is the bytes from a JSON serialized status snapshot (e.g., the
+`HashMap<PipelineKey, PipelineStatus>` returned from
+`ObservedStateHandle::status_snapshot`).
+
+## Telemetry
+
+### Metrics
```
lquerel, line 784: We should consider adding metric attributes and latency metrics here: outcome, error kind, transport, reconnect attempt, reconciliation duration, remote config status, and whether config was skipped because the hash was unchanged.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+). A followup design can be created for this.
+
+### Imperative Command Capability
+
+This design assumes the server will specify the remote configuration in full.
+E.g. the agent expects to receive the config declaratively.
+
+However, in the future we may want to expose a mechanism for servers to trigger
+the ad-hoc creation/reconfiguration/deletion of some pipeline. This could be
+supported through a custom agent capability / custom server message.
```
lquerel: Even if imperative commands remain future work, it may be useful to reserve a custom capability namespace and a minimal command envelope now. For example, a future custom message could carry `{ "command": "pipeline_delete", "request_id": "...", "payload": { "group_id": "tenant-a", "pipeline_id": "ingest", "timeout_secs": 30 } }`, with responses carrying the same `request_id`, a status, optional result, and optional error. That would keep declarative full-config reconciliation as the primary model while avoiding incompatible custom message shapes later.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+
+Agent to Server communication typically happens over a network where the Agent
+acts as a client, using either
+[WebSocket](https://opentelemetry.io/docs/specs/opamp/#websocket-transport) or
+[plain HTTP](https://opentelemetry.io/docs/specs/opamp/#plain-http-transport)
+(plain meaning standard HTTP request-response model, not HTTP over plaintext).
+
+The messages exchanged are protobuf serialized `AgentToServer` and
+`ServerToAgent` messages.
+Find [proto message definitions](https://github.com/open-telemetry/opamp-spec/blob/main/proto/opamp/v1/opamp.proto).
```
lquerel, line 36: We should probably state that generated bindings from the official OpAMP proto should be used where possible, or explicitly list any supported subset. That will reduce the risk of subtle drift from the upstream OpAMP spec.

lquerel (approved):
Overall, this looks like a really strong direction!
Most of my comments are about making the design more implementation ready and interoperable. Thanks

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+`instance_uid` - the server may respond with a new agent identification to
+override the [initial value](https://opentelemetry.io/docs/specs/opamp/#servertoagentagent_identification).
+
+If `agent_description` is not configured, the `agent_description` field will
+not be set on the `AgentToServer` messages.
+
+### Engine Config Representation
+
+The server should supply the engine config as JSON, embedded within the
+`ServerToAgent.remote_config.config.config_map["desired_state"]` which contains
```
cijothomas, line 524: "desired_state" - I don't find such a key in the opamp spec. Can you link to the relevant OpAMP spec line?

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+last applied config hash, the new config will be applied to the engine.
+
+The agent will try to deserialize the remote config as an `OtelDataflowSpec`.
+If deserialization fails, it will respond to the server with an `AgentToServer`
+message with a `remote_config_status` containing a `FAILED` status.
+
+If deserialization of the new remote_config succeeds, the agent will immediately
+reply to the server with an `AgentToServer` message with a `remote_config_status`
+containing an `APPLYING` status, and it will then use the DFE `ControlPlane` to
+reconcile the engine config.
```
cijothomas: does the engine already support changing settings such as downsizing total memory/size etc, without process restart?
I guess I am asking if we plan a supervisor concept too?

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+  `failed`
+- Otherwise if all pipeline has status `stopped`, the group status will be
+  `stopped`
+- Otherwise if all pipelines have status `running`, the group status will be
+  `running`
+- Otherwise the group status will be `degraded`.
+
+### Custom Messages & Capabilities: Full Pipeline Status
+
+The implementation will advertise a custom capability representing its ability
```
cijothomas: nit: Add a small explanation why we cannot leverage the ComponentHealth and need a custom one here.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+
+The payload here is the bytes from a JSON serialized status snapshot (e.g., the
+`HashMap<PipelineKey, PipelineStatus>` returned from
+`ObservedStateHandle::status_snapshot`).
+
+## Telemetry
+
+### Metrics
+
+The controller extension implementation should produce metrics:
```
cijothomas, line 786: nit: something to track if the ws connection gets lost and reconnected?

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+implementation described in this document
+
+### Client Settings
+
+OpAMP has the capability to configure connection settings both between the
+agent & server, and between agent and some external OTLP receiver to which the
+agent should export its own internal telemetry.
+
+In the case of DFE, the settings between agent & OTLP receiver are not useful
+because this can be configured via the internal telemetry pipeline, the
```
cijothomas, line 817: ITS is not quite there for Metrics yet. lets leave a TODO to account for that here.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+looking to manage their own fleet of DFEs. It's not the intention of this
+design to prescribe a solution for this, especially given that its unknown
+how such a theoretical system will manage/generate DFE config.
+
+However, we can still imagine a world where there are some cross-cutting
+primitives that _most_ OpAMP server implementers will need to build. This
+includes an HTTP server, possibly with websocket enabled, cert management,
+instance_uid and other state management, etc.
+
+We could consider making this easier by providing an SDK in rust that abstracts
```
cijothomas: I think the reference implementation in opamp-go is good enough for now. There are some proposals about making a contrib repo for opamp too, for folks to make additional server implementations etc. We could use that, if we chose to (and that contrib exists!).

cijothomas (approved):
Looks good. I left some comments, we should tackle it as we build the capabilities. Don't have to have everything sorted out right in this design.

rust/otap-dataflow/docs/opamp.md
cijothomas: +1. I think I also left this comment somewhere.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+The URN used to identify the controller extension will be
+`urn:otel:controller_extension:opamp` and it can be configured as follows:
+
+```yaml
+# example DFE config
+engine:
+  controller:
+    opamp:
+      type: urn:otel:controller_extension:opamp
+      config:
```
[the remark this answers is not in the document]
albertlockett: fixed in 93f36a0a6

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+  the next reply to the server should contain full state or available
+  components. If a new config is being reconciled, these can be included on
+  the message that is sent to the server letting it know config is being
+  applied. Otherwise, send an ad-hoc message immediately reporting these
+  fields.
+
+In any case where a message is ignored (especially in error cases), appropriate
+telemetry should be emitted.
+
+##### Send periodic heart beats
```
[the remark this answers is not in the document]
albertlockett: c89992539

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+          # Trust settings for verifying the server certificate
+          ca_file: "./my-certs/ca.crt"
+          include_system_ca_certs_pool: false
+
+          # Client identity for mTLS (these make it mTLS instead of just TLS)
+          cert_file: "./my-certs/client.crt"
+          key_file: "./my-certs/client.key"
+
+
+        # Configuration for heart-beat timing.
```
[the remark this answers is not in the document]
albertlockett: e05945190

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+        #
+        # Optional (default = 30s)
+        heart_beat_interval: "10s"
+
+        # Options for configuring exponential backoff for initial Websocket
+        # connection.
```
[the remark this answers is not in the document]
albertlockett: 3d9e23cf7

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+### Agent Identity
+
+The `AgentToServer` message contains an `instance_uid` and an
+`AgentDescription` with identifying attributes. These are used by the server
+to identify the agent.
+
+Different control plane implementation may have different mechanisms for
+identifying the client depending on their unique deployment scenario. As such,
+it is not specified how agents should identify themselves nor how to compute
+agent description - these parameters will be configurable.
```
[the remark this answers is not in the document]
albertlockett: 69879bac0

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+override the [initial value](https://opentelemetry.io/docs/specs/opamp/#servertoagentagent_identification).
+
+If `agent_description` is not configured, the `agent_description` field will
+not be set on the `AgentToServer` messages.
+
+### Engine Config Representation
+
+The server should supply the engine config as JSON, embedded within the
+`ServerToAgent.remote_config.config.config_map["desired_state"]` which contains
+an  `AgentConfigFile` message.
```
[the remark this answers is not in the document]
albertlockett: ffbbd1bcc

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+returned by the control plane. Note, this snapshot contains a `PipelineStatus`
+object, which contains `PipelineRuntimeStatus` for all instances of a pipeline,
+which contains a `PipelinePhase`.
+
+The remote config status will be derived from the snapshot using the following
+rules:
+
+- If any instance of any pipeline has phase is `Pending`, `Starting`,
+  `Draining`, `Updating`, `RollingBack` or `Deleting` the remote config status
+  will be `Applying`
```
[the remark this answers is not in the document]
albertlockett: 9397ae90f

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,532 @@
...
+- Otherwise the status will be `degraded`
+
+When resolving the status for engine's pipeline groups, the resolution logic
+will use the following rules:
+
+- If any pipeline has status `stopping`, the group status will be `stopping`
+- Otherwise if any pipeline has status `failed`, the group status will be
+  `failed`
+- Otherwise if all pipeline has status `stopped`, the group status will be
+  `stopped`
```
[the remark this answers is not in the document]
albertlockett: 2ccb692c8

rust/otap-dataflow/docs/opamp.md
albertlockett: :+1: fixed in 3a5d6ea62

rust/otap-dataflow/docs/opamp.md
albertlockett: 4c0cbd0b4

rust/otap-dataflow/docs/opamp.md
albertlockett: fixed in 4164072c4

rust/otap-dataflow/docs/opamp.md
albertlockett: Even if the control-plane uses OpAMP protocol, it could be implemented in such a way that it makes certain expectations of the agent. One example I was thinking of when I wrote this is that perhaps some control plane implementation expects certain attributes to be available in the `AgentDescription`, or otherwise for the `instance_uid` to be populated in a specific manner. 

My intention with this principle is to state that we'd like to implement the agent in such a way as to be amenable to these expectations. In practical terms it means exposing flexible options to configure how these fields are populated.

rust/otap-dataflow/docs/opamp.md
albertlockett: Yes these heartbeats are `AgentToServer` messages

rust/otap-dataflow/docs/opamp.md
albertlockett: fixed

rust/otap-dataflow/docs/opamp.md
albertlockett: Updated in 6523cdc22

If this were set to `false`, wouldn't the server be unable to signal to the agent that it needs to shutdown a pipeline that it previously?

rust/otap-dataflow/docs/opamp.md
albertlockett: Added a section in 27b8d85ca

In terms of configuration, I figured we'd use the same config struct we use for the OTLP exporters: https://github.com/open-telemetry/otel-arrow/blob/main/rust/otap-dataflow/crates/config/src/tls.rs#L49-L92

rust/otap-dataflow/docs/opamp.md
albertlockett: I reworked the error handling notes in 6b4b426fc

> In contrast, malformed protobuf bytes or an unsupported WebSocket frame type may indicate a broken protocol exchange, 

Currently the implementation in https://github.com/open-telemetry/otel-arrow/pull/3410 actually just ignores these and keeps exchanging messages on the same connection. I suppose it's true this could indicate a broken protocol exchange, so I've updated the docs to say that we should ignore these and will update behaviour in the impl PR as well.

rust/otap-dataflow/docs/opamp.md
albertlockett: Thinking about this a bit more - I guess they would use some commands to shutdown the pipelines.

rust/otap-dataflow/docs/opamp.md
albertlockett: Sounds good, added in 252c65363

rust/otap-dataflow/docs/opamp.md
albertlockett: Yeah sure we could do that.

By "restart",  do you mean that the controller extension would be restarted without the engine restarting? If this were, by contrast, the process restarting, since we wouldn't have persisted the full configuration last received from the server, we'd simply restart from the bootstrap configuration that was passed via arguments at startup right?

rust/otap-dataflow/docs/opamp.md
albertlockett: Yes sounds good. In fact, that's what this paragraph was trying to state should be done, it just wasn't very clear. I added a new sentence in 55cd72c4b to make it clear that the `APPLYING` message should be sent before calling this method.

rust/otap-dataflow/docs/opamp.md
albertlockett: :+1: fixed in e7847ac75

rust/otap-dataflow/docs/opamp.md
albertlockett: Sounds good - documented as a documentation followup  https://github.com/open-telemetry/otel-arrow/issues/3420

rust/otap-dataflow/docs/opamp.md
albertlockett: Sounds good.  Added a note about this in 9f2aedf2e

rust/otap-dataflow/docs/opamp.md
albertlockett: This "desired_state" key isn't part of the OpAMP spec. This `remote_config.config.config_map` field has type `map<string, AgentConfigFile>`, so I figured we need to decide on a key to use in this map, and "desired_state" seemed reasonable.

albertlockett: https://opentelemetry.io/docs/specs/opamp/#configuration-files

rust/otap-dataflow/docs/opamp.md
albertlockett: Sounds good, added in 70345ddf1

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,590 @@
...
+  - Reports Health
+  - Reports Remote Config
+  - Reports Heartbeat
+- `health` - see section below on server health
+- `flags` - `Unset`
+- `custom_capabilities` - list capability to send full state as custom message
+  (see section below on custom capabilities).
+- `remote_config_status` - status field = `Unset`
+- `connection_settings_status` - status field = `Unset`
+- `custom_message` - full pipeline status from status snapshot
```
lalitb, line 340: Should we also list [AcceptsRemoteConfig](https://opentelemetry.io/docs/specs/opamp/#agenttoservercapabilities) here? The design seems to include receiving and applying remote config, so I assume that is part of the initial scope.

albertlockett: Yes, this missing capability was an oversight. Added in a59f4a857 and added AcceptsRestartCommand as well.

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,594 @@
...
+
+#### `RemoteConfigStatus`
+
+The `AgentToServer` message will contain a
+[`RemoteConfigStatus` message](https://opentelemetry.io/docs/specs/opamp/#remoteconfigstatus-message)
+for this purpose which contains high-level information about the applied remote
+config.
+
+Before receiving any configuration from a remote server, the agent will always
+set the status to `APPLYING`.
```
lalitb: Maybe I am missing something, but the initial state should be "UNSET" or "APPLYING" ?

rust/otap-dataflow/docs/opamp.md
albertlockett: sounds good, removed suggestion in 7ac9f3d0d

rust/otap-dataflow/docs/opamp.md
albertlockett: Sure can add a TODO. What's the gap w/ ITS metrics? Live configuration of the internal telemetry pipeline isn't supported or something?

rust/otap-dataflow/docs/opamp.md
albertlockett: UNSET. fixed in 135a44284

rust/otap-dataflow/docs/opamp.md
```diff
@@ -0,0 +1,590 @@
...
+  controller:
+    extensions:
+      opamp:
+        type: urn:otel:controller_extension:opamp
+        config:
+          # The URL of the OpAMP server, including the scheme and path.
+          #
+          # The scheme will be used to determine whether communication
+          # is done over WebSocket or HTTP. E.g. use "ws://..." for websocket
+          # and http:// for http.
```
lalitb, line 93: Since `http://` is supported here, can we document the plain HTTP polling behavior? Per the OpAMP plain HTTP transport [spec](https://opentelemetry.io/docs/specs/opamp/#plain-http-transport), the server can only send messages in response to the agent's poll, so remote config delivery is bounded by the polling interval. Is `heartbeat_interval` also intended to be the HTTP polling interval here?

lalitb (approved):
LGTM. Nice work, and thanks for initiating this feature :)

rust/otap-dataflow/docs/opamp.md
albertlockett: :+1: c3903b4f6 reworked this sentence to make more explicit.

rust/otap-dataflow/docs/opamp.md
albertlockett: You're thinking that we should get this from like the `engine.telemetry.resource` config value eh? Yeah I guess it'd be convenient if the didn't have to make the user configure this twice.

Do you think we should make this behaviour configurable?  e.g. if we did something like:

```yaml
agent_description:
  identifying:
    include_telemetry_resource_attributes: true # defaults to true
    attributes: {
      # ... key/value pairs of custom attrs for overrides/extra attrs
    }
```

rust/otap-dataflow/docs/opamp.md
albertlockett: Added to design doc in 4eec833c5 and added a followup issue to actually implement this behaviour:  #3423

rust/otap-dataflow/docs/opamp.md
albertlockett: :+1: https://github.com/open-telemetry/otel-arrow/pull/3388#discussion_r3537447817

rust/otap-dataflow/docs/opamp.md
albertlockett: Added in 1278df9f0. FWIW, some of this is also in the OpAMP spec https://opentelemetry.io/docs/specs/opamp/#throttling

rust/otap-dataflow/docs/opamp.md
albertlockett: Added to design doc in 122be36bb

rust/otap-dataflow/docs/opamp.md
albertlockett: Interesting -- yeah I hadn't realized that the agent could/should be impelmented in such a way that capabilities should be toggled by config, but that makes sense as a use case.

Indeed, the docs actually recommend specifically that accepting a remote config should be opt-in https://opentelemetry.io/docs/specs/opamp/#opt-in-remote-configuration

Updated the design doc in b38670669 to specify that each capability should have a config toggle and created an issue to track that work: https://github.com/open-telemetry/otel-arrow/issues/3424

rust/otap-dataflow/docs/opamp.md
albertlockett: Sounds good - I added at least a bit more information about this in 4a737d8b4

rust/otap-dataflow/docs/opamp.md
albertlockett: Yes as far as I know we support changing the resource policies. We have tests for this anyway:
https://github.com/lquerel/otel-arrow/blob/5a733bae9089b400a2a484b59740e7f065dd09ee/rust/otap-dataflow/crates/controller/src/live_control_tests.rs#L484-L652

rust/otap-dataflow/docs/opamp.md
albertlockett: sounds good, added in 3246758be

rust/otap-dataflow/docs/opamp.md
albertlockett: comments about http in [7c8b37d](https://github.com/open-telemetry/otel-arrow/pull/3388/commits/7c8b37da23dcedc1f8c7d58da7d95e2e69ea6b09) including that the heartbeat interval is intended to be the polling interval.

PR #470, opened 2025-05-20
[WIP] Add pipeline configuration and validation using Hyper-DAG

lquerel:
Part of #364 

Building a pipeline from a configuration:

- Pipeline configuration represented as a Hyper-DAG data structure. Hyper-edges are used to define dispatch strategies (e.g., round-robin, broadcast, random, least loaded, etc.) when a node connects to multiple downstream nodes.
- Hyper-DAG validation. Check for duplicated nodes or out ports, cycles, etc.
- Pipeline instantiation from the configuration.
-Validation of a simple pipeline.

Issue #3310, opened 2026-06-17
OPL add ability to use batch headers in expressions


albertlockett:
We have the capability to extract headers and keep them as batch level metadata on OTAP batches. It would be nice if we had the ability to use these headers in expressions. For example, we could set them, or use them as an input (for example, to move the value to an attribute).

```
logs | set headers["x"] = "y"
logs | set attributes["x"] = headers["y"]
```

albertlockett:
marking as triage:needs-info because I need to circle back to this and flesh out what the field identifier should be called. I think `headers` is reasonable but need to double check if there's other batch metadata that we keep and factor that into the design decision

Issue #3273, opened 2026-06-11
OTAP query-engine partitioning batches


albertlockett:
### Pre-filing checklist

- [x] I searched existing issues and didn't find a duplicate

### Component(s)

Rust OTAP dataflow (rust/otap-dataflow/)

### Is your feature request related to a problem?

Imagine I'm writing a telemetry component that wants to group telemetry together by some computed value. 

This could be something simple like: group together all telemetry with the same `service.name` resource attribute. The computation may be more complex, something like `sha265(concat(resource.attributes["service.name"], resource.attributes["k8s.namespace.name"]))`

Why might a component want to do this? Imagine an exporter has to choose where to send this data dynamically based on the computed value. E.g. the value could be used choose a kafka topic, postgres schema, or partition in a partitioned parquet file.

Each component _could_ implement its own partitioning logic, but imo it would be better separation of concerns if we had a generic mechanism for partitioning that was upstream of the component.

### Proposed Solution

I propose we add a `partition` operator call to OPL / OTAP query-engine. It will take some expression, partition the batch by the value of the expression, and assign the partition key to the batch metadata.

In order for this to work, the OTAP query-engine's interface will need to change. Currently, it takes in a single OTAP Batch and produces a single OTAP batch as output. This doesn't naturally fit in the case where the batch may be re-partitioned. As a prerequisite to getting this fully working, we'll also need to refactor the interface to the query-engine and the interface between pipeline stages to work on a stream of batches.

However, the refactoring described in the paragraph above doesn't block us from writing a partition-by-expr kernel for the query-engine crate, which should be the first step.

albertlockett:
working on this issue this week

PR #3389, opened 2026-06-30
Multitenant pipeline design

jmacd:
# Change Summary

Adds the outline of a design for request-level tenant descriptors and a common mechanism
for defining conditional rate and resource limiter configurations by tenant.

Fixes #470 

* [x] This is a documentation-only PR.


jmacd:
See github.com/open-telemetry/otel-arrow/issues/3375

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+
+However, where there are context-specific attributes used for
+limiting, a **tenant descriptor** will be used. A tenant descriptor is
+a small vector of key:value identifiers which represent the current
+tenant in some context. Usually, the context is a request, however,
+custom contexts can be defined as long as a tenant descriptor can be
+stored.
+
+Tenant descriptors are multi-dimensional to enable tenant assignment
+in logically independent ways with user control ove rcardinality. For
```
albertlockett:
```suggestion
in logically independent ways with user control over cardinality. For
```

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+These actions do not apply in contexts where multiple resources are
+present. Under this proposal, the options for handling multi-resource
+requests with these resource descriptor actions are:
+
+- Nack the request.
+- Reject the configuration: callers asserts single-reource context or else
+  invalid configuration.
+- Do not resolve the action, the descriptor will be unresolved; receivers
+  and limiters that require this descriptor will reject, otherwise these
+  requests will not match conditions and take the default limit.
```
albertlockett: I think we could use the concept of partitioning described here, if it were placed upstream of the thing that applies these actions, to mitigate the issue of mixed resource batches. 

This issue https://github.com/open-telemetry/otel-arrow/issues/3273 describes the ability to partition based on the result of some OPL expression. In this case the expression could simply be something like `resource.attributes["service.name"]`. 

Note: I imagine the partition-by-expression implementation will need to partition on multiple expressions so we can use the output for multiple actions (this is not called out in #3273, but seems like the right thing to do).

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+      # for premium-tier customers
+      - entries:
+        - key: tier
+          value: premium
+        output: premium_pipeline
+      # for any customer
+      - entries:
+        - key: customer_id
+        output: shared_pipeline
+      default_output: fallback
```
albertlockett: Although https://github.com/open-telemetry/otel-arrow/issues/3310 only calls out headers, I really intended this to be catchall issue where we added the ability to access / manipulate batch metadata in OPL, and I believe this tenant description is something that could be included. If we did that, I imagine that this kind of tenant based routing could also be expressed in OPL and done through transform processor

```
signals |
if (tenant_descriptor["customer_id"] == "bigfish" ) { 
  route to "bigfish_pipeline" 
} |
if (tenant_descriptor["tier"] == "premium" ) {
  route to "premium_pipeline"
} |
if (tenant_descriptor["customer_id"] != null) {
  route_to "shared_pipeline"
} |
route_to "fallback"
```

albertlockett (approved):
Really cool @jmacd !

Like we touched on in SIG on Tuesday, I think there's some overlap between what this PR is trying to acheive and some work we have planned in the query-engine/OPL.

Specifcally, I think this capability to partition by an OPL expression (which could be extracted to a consumable function for easy invocation outside the full query-engine context), could be used to split batches with metadata that could be used to drive some tenant description actions: 
https://github.com/open-telemetry/otel-arrow/issues/3273

I also think it'd be great to add the capability to work with these tenant descriptors into OPL/query-engine (and batch metadata generally), which I'm tracking in:
https://github.com/open-telemetry/otel-arrow/issues/3310

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+
+```yaml
+actions:
+- key: service_name
+  resource_key: service.name
+- resource_key_match: service.namespace
+  value: production
+```
+
+These actions do not apply in contexts where multiple resources are
```
lalitb: I may be missing something, but this feels like a core gap. Header-derived descriptors classify the whole request/batch as one tenant, while `resource_key` is single-resource only. For mixed-tenant batches, could we define the expected behavior here? The options below seem to sit at different levels (config rejection, runtime Nack, or unresolved/default), so it would help to specify which applies when.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+
+### Resource limiters
+
+The built-in thread-local resource limiter is a semaphore that restricts the
+total weight of some concurrent activity, for example the number of
+requests in flight, the total memory in use, or the number of items in
+a queue. The semaphore acts as a gate, limiting the amount of weight
+admitted into a logical-admitted state.
+
+The main resource limiter interface **acquire** returns a reservation
```
lalitb: I think this needs a clearer rule. This section says a resource hold can live in `Context` and follow the request, but later says holds are `!Send`. If the request then crosses a topic or moves to another core, that seems hard to make safe. Can we define whether holds stay out of `Context`, or get released and re-acquired at those boundaries?

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+semaphore:
+  admitted: 100_000_000  # max admitted weight
+  waiting:  20_000_000   # max waiting weight
+  waiters:  100          # max number waiting
+  mode:     lifo         # prioritize freshness
+```
+
+### Rate limiters
+
+The rate limiter interface **limit** returns a reservation object. If
```
lalitb: This seems inconsistent with the earlier rate-limit definition. If rate tokens are consumed immediately and not returned, then an earlier grant cannot really be “cancelled” when a later limiter fails. I think this needs a clear reserve/commit/cancel rule for correctness, not just as an efficiency follow-up. For example, all limiters could first reserve, then commit only after every limiter grants.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+
+### Shared limiters
+
+Sharing limiter extensions at the pipeline-group and engine scope
+(i.e., across threads or CPUs) in the dataflow engine requires careful
+attention to avoid synchronization costs.  The built-in token bucket
+and semaphore limiters support thread-local use only, so they are
+defined with pipeline-scope. The dataflow engine extension mechanism
+supports requesting shared extensions at wider scopes generally, so
+limiter scope can be thread-local (pipeline), CPU-local (pipeline
```
lalitb: This scope model (pipeline / group / engine) seems to overlap with the existing `resources` policy framework in `config/src/policy.rs`, while the current `memory_limiter` is restricted to top-level policy only. Can we clarify whether these new per-scope limiters are meant to extend that framework or replace it?

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+
+### Tenant trust
+
+Pipeline operators are responsible for secure tenant
+configurations. Tenant descriptors can be defined at multiple levels
+of the engine configuration (e.g., global, pipeline group, pipeline)
+to avoid leaking tenant details outside their scope. The dataflow
+engine is responsible for enforcing this discipline automatically.
+
+Pipeline operators are advised not to use unauthenticated request
```
lalitb: Can we make the trust model a bit more explicit here? If descriptors can come from headers, peer address, or resource attributes, it would help to define which sources are trusted and how they relate to authentication.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+Limiters record a primary observability signal with the declared units,
+exposing the current value of the limiter.
+
+- For rate limiters, an OpenTelemetry Counter named `otelcol.rate_limiter.accepted` measuring the accepted weight
+- For resource limiters, an OpenTelemetry UpDownCounter named `otelcol.resource_limiter.admitted` measuring the currently admitted weight
+
+Each of these uses the following dimensions:
+
+- Condition name
+- Key:values in condition entries of matching bucket
```
lalitb: Can we clarify the intended metric-label policy for tenant values? Customer IDs, workspace IDs, and subscription IDs may be sensitive or high-cardinality even when bucket count is bounded, so it would help to say whether these values are emitted, hashed, or disabled by default.

lalitb:
Thanks for putting this together @jmacd. I like the direction and the model is helpful. I left a few questions where I’m trying to understand the intended integration points better. Once those are clear, I think the design will be easier for implementers to follow consistently.

rust/otap-dataflow/docs/agent-multitenancy-design.md
lquerel:
```suggestion
in logically independent ways with user control over cardinality. For
```

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+pipeline where individual nodes will apply specific limiters.
+
+![A flow diagram showing how tenant descriptor values are computed and flow through the pipeline](./agent-multitenancy-diagram.svg)
+
+### Tenant descriptor actions
+
+Tenant descriptors originate in a series of actions, configured at the
+top level, group-level, and pipeline-level of the dataflow engine
+configuration, since these definitions are shared across pipeline
+groups. Descriptor values refer to one or more tenant associations,
```
lquerel: I don't understand why we say that "these definitions are shared across pipeline groups" when, at the beginning of the sentence, we say that a tenant descriptor can be defined at the pipeline level. In that case, one would expect the descriptor not to be shared at the group level.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+each association consisting of multiple keys and values and used as a
+lookup key for evaluating limiter conditions. Tenant descriptor values
+must be erased when they go out of scope to avoid leaking sensitive
+data across unintended boundaries.
```
lquerel: I like this concept of boundaries.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+
+Request actions are applied by receivers (or processors) when they
+originate new request contexts. Descriptor actions support attribute
+extraction in a variety of ways, for example,
+
+- `receiver_id`, `source_node_id`: the first node or preceding node traversed by the request.
+- `remote_address`, `masked_remote_address`: origin network address, optionally CIDR-masked.
+- `generic_key`: a static, hard-coded key-value.
+- `request_header`: copy a request header value into a descriptor key.
+- `request_header_match`: descriptor key value must match a condition.
```
lquerel:
```suggestion
- `request_header_match`: request header value must match a condition.
```

Not 100% sure but that seems more coherent for me.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+| Condition         | An ordered list of entries selecting a bucket; the first matching condition wins.                                                                  |
+| Bucket            | Contains one distinct limiter instance per distinct set of entry keys for the matching condition, up to the cardinality limit.                     |
+
+Descriptor actions are evaluated and the matching results are placed
+in the request context where it passes with the request data in a
+pipeline where individual nodes will apply specific limiters.
+
+![A flow diagram showing how tenant descriptor values are computed and flow through the pipeline](./agent-multitenancy-diagram.svg)
+
+### Tenant descriptor actions
```
lquerel: The term "action" seems very generic here. Why not use the term "extractor" instead, or "definition" that better characterizes what is being done?
Alternatives: "Tenant description extractor", "Tenant descriptor definition", "Tenant descriptor spec", ...

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+`signal` to identify the type of signal in the request (e.g., "logs").
+
+### Tenant descriptor examples
+
+Here is an example of a single descriptor action list based on the
+client address, a hard-coded value, and an extracted request header as
+the keys:
+
+```yaml
+actions:
```
lquerel: Not a big fan of the term `actions`

rust/otap-dataflow/docs/agent-multitenancy-design.md
lquerel: In other parts of the project, we use the term "transport header" instead of "request header", which seems more appropriate to me.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+Tenant descriptors are multi-dimensional to enable tenant assignment
+in logically independent ways with user control ove rcardinality. For
+example, a request may be accompanied by an environment name, a
+project name, and a user name for three forms of tenancy. Choosing
+three forms means a single limiter table with three-dimensions.
+
+Multiple tenant descriptors per request are supported, enabling
+multiple independent limiters with lower dimensions. For example, a
+request may be accompanied by both an end-user tenant identifier and
+an acting-on-behalf-of tenant identifier. These are logically
```
lquerel: I think the term "acting-on-behalf-of” should be clarified somewhere.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+  request_header: x-user-id
+```
+
+Here is an example engine configuration defining three tenant
+descriptors: two end-user forms (modern and legacy) and one
+acting-on-behalf-of service descriptor.
+
+```yaml
+# Engine top-level descriptors, shared across all pipeline groups.
+descriptors:
```
lquerel: I find the name of this section too generic, unless the intention is to use this descriptor concept in a broader context than just tenant descriptors. I suggest renaming this section to `tenant_descriptors`.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+  enduser_tenant_modern:
+    actions:
```
lquerel: Maybe this will be explained later in the document, but I don't understand the purpose of this intermediate `actions` level. Why not have the list of descriptor components directly under the tenant descriptor name?

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+      request_header: x-subscription-id
+groups:
+  # ... pipeline groups reference the descriptors above ...
+```
+
+The Dataflow Engine's pipeline data type (`OtapPdata`) will be
+modified to propagate tenant descriptors in the context. The interface
+used by receivers (and processors) to construct new contexts from
+headers will be extended to evaluate the list of descriptors. Every
+descriptor is evaluated for every request context. When all conditions
```
lquerel: Shouldn't we use the term `OtapPdata` instead of `request`, as we do in the rest of the project?
For example, if we have a tenant descriptor processor, it feels odd to refer to a request.

rust/otap-dataflow/docs/agent-multitenancy-design.md
lquerel: In the medium term, and to handle more complex cases, we should probably support OPL-based tenant descriptors so that OPL functions can be used.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+describes a single producer of telemetry. When a pipeline request
+contains data for a single resource, as typically produced by an OpenTelemetry SDK,
+the request's resource attributes can be made to act like request headers as the input
+for tenant descriptor actions.
+
+Receivers for non-HTTP telemetry protocols that convey single-resource
+requests may apply descriptor actions directly to resource
+attributes. Limiters that are applied in those contexts may refer to
+single-resource descriptor actions `resource_key` and
+`resource_key_match` to extract or conditionally extract tenancy
```
lquerel: Why introducing a new term "resource key" here, while in the rest of the project we use the term "resource attribute" ?

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+information. These actions can only be applied in single-resource
+contexts. For example to extract a `service_name` tenant for the
```
lquerel: What does this "single-resource context" mean exactly? Does it mean we can't apply multi-tenancy to batches containing multiple resources? Or does it mean we first have to partition/split the batches into smaller batches containing only one resource?

EDIT: I saw some partial answers just after this, but I'm still not completely clear about this limitation.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+engine must add support splitting requests in these ways to enable
+routing, shuffling, grouping and load balancing by tenant descriptor.
+
+### Tenant trust
+
+Pipeline operators are responsible for secure tenant
+configurations. Tenant descriptors can be defined at multiple levels
+of the engine configuration (e.g., global, pipeline group, pipeline)
+to avoid leaking tenant details outside their scope. The dataflow
+engine is responsible for enforcing this discipline automatically.
```
lquerel: Nice!

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+    config:
+      optional_descriptors: [enduser_tenant_modern, enduser_tenant_legacy]
+
+      # first-match wins
+      routes:
+      # customer_id=bigfish
+      - entries:
+        - key: customer_id
+          value: bigfish
+        output: bigfish_pipeline
```
lquerel: I'm not sure I’m correctly interpreting what constitutes an entry. Based on this YAML representation, it's a list of key/value pairs plus an output. But based on the definition table at the beginning of this document, it's just the list of key/value pairs.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+
+### Routing/batching by tenant descriptor
+
+Processors (e.g., fanout) and exporters (e.g., topic) will be
+configurable to route by descriptor condition.
+
+```yaml
+nodes:
+  tenant_split:
+    type: processor:tenant_router
```
lquerel: I don't have a strong opinion on this yet. We already have a content router that could support this type of routing configuration. Is there an advantage to having a dedicated tenant router?

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+    config:
+      optional_descriptors: [enduser_tenant_modern, enduser_tenant_legacy]
```
lquerel: What is the purpose of this list?

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+`sending_queue::batch::partition::metadata_keys`](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/exporterhelper/README.md#sending-queue-batch-settings)
+and a configurable cardinality limit. Tenant descriptors and batching
+conditions will be used to this effect in the dataflow engine.
+
+### Load balancing by tenant descriptor
+
+The example above shows to isolate a small number of routes in static
+configuration based on the tenant descriptor. Engines running on
+multiple cores are also required to route and load balance when the
+number of descriptors is large.
```
lquerel: In my view, the use of topics in this context is not solely due to the number of tenants, but more to the isolation mechanisms that can be applied to them.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+
+Load balancing by tenant descriptor will be organized using the
```
lquerel: There is probably a good reason for using the term "tenant descriptor" instead of "tenant ID", but I'm not sure I understand the rationale behind it.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+dataflow engine's topic broker infrastructure based using tenant
+descriptors and conditions to compute a hash value, and then to use
+(hash-value % N) to distribute the request to one of N topic
```
lquerel: We can imagine using topics in this context as you describe. Personally, I had another approach in mind: having one consumer group per tenant ID (load balancing by tenant ID rather than by hash), so that limits can be applied per tenant.

For the dynamic aspect, the idea was to define the concept of a pipeline template that could be automatically instantiated for each new tenant ID, with an explicit maximum limit, of course.

**To me, this load balancing by tenant ID at the topic level was the basic mechanism for managing multi-tenancy and the associated limits.**

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+receivers. The design of the load balancing mechanism is out of scope
+for this document and part of a larger scope. Often, there is a
```
lquerel: Depending on the conclusion from my previous comment, I think the concept of load balancing at the topic level can either be considered out of scope for this first version of the document, or be central to it. I'll wait to read the rest of the document before giving my opinion on the subject.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+  `memory_bytes/second`). For resource limiters omit the rate suffix (e.g., `memory_bytes`). In the yaml configuration, this gives the
+  raw numbers meaning; in the code, a verification step ensures that
+  each limiter is used to limit the correct category of weight.
+- **optional_descriptors**: Optional, a list of the tenant descriptors
+  used by the limiter. These descriptor values are extracted from the
+  request and used to evaluate conditions.
+- **required_descriptors**: Optional, a list of the tenant descriptors
+  used by the limiter that must be present, otherwise the request is
+  immediately failed. These descriptor values are extracted from the
+  request and used to evaluate conditions.
```
lquerel: We have optional `optional_descriptors` and optional `required_descriptors`, which seems a bit ambiguous/confusing on first read. :-)

Can't we derive everything from the `conditions` field?

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+      entries: { ... }
+      specific: { ... } # first condition specifics
+      cardinality: { ... }
+    - name: second
+      entries: { ... }
+      specific: { ... } # second condition specifics
+      cardinality: { ... }
+    specific: { ... }   # default condition specifics
+    cardinality: { ... }
+```
```
lquerel: I must admit I didn't fully understand this configuration and the concept of `specific_limiter`.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+(hash-value % N) to distribute the request to one of N topic
+receivers. The design of the load balancing mechanism is out of scope
+for this document and part of a larger scope. Often, there is a
+simultaneous need to aggregate and load-balance multi-tenant request
+by features in the data other than tenant, such as metric name or
+TraceId. In both cases, the dataflow engine requires a mechanism to
+create multi-tenant, multi-request batches destined for a single
+consumer.
+
+## Limiter extensions
```
lquerel: In terms of implementation, I don't know yet what you have in mind, but one thing I'm sure of is that, in order to maintain the same level of performance, we absolutely need to avoid introducing any global synchronization mechanisms, as that would essentially negate all the work we've done so far.

We already have the concept of a policy in the project, which seems to me like the right term to use for defining rate limits or resource-based limits. We can definitely imagine policy extensions, but ideally I'd like the concept of policy to be used here.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+weight through its reservation interface. The built-in semaphore
+limiter's specific configuration, for example:
+
+```yaml
+semaphore:
+  admitted: 100_000_000  # max admitted weight
+  waiting:  20_000_000   # max waiting weight
+  waiters:  100          # max number waiting
+  mode:     lifo         # prioritize freshness
+```
```
lquerel: Related to my previous comment about using declarative policies, I think we could come up with a resource policy definition that is a bit more explicit and easier to understand than this one :-)

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+example:
+
+```yaml
+token_bucket:
+  allow:    100_000_000  # maximum per interval
+  burst:    20_000_000   # maximum individual weight
+  interval: 60s          # interval duration
+  waiters:  100          # max number waiting
+  mode:     fifo         # prioritize fairness
+```
```
lquerel: If I want to apply rate limiters and/or resource limiters per tenant, how should I proceed?

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+
+The limiters described in previous examples have been declared in
+thread-local scope, making limits per-tenant and per-pipeline. To
+implement engine-wide or group-wide limits on a per-tenant bases,
+there are two options:
+
+1. Use a group- or engine-shared limiter instance.
+2. Route the data by tenant descriptor to a single pipeline, then use a thread-local limiter instance.
+
+Both are reasonable options.
```
lquerel: Is option 1 really reasonable?

I think option 2 is more appropriate for our engine and our constraints. We could imagine a concept of a tenant cell combined with the concept of topics to implement efficient multi-tenancy in our engine. I'll try to define this concept of a tenant cell more clearly later on.

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+              - name: bigfish
+                entries:
+                - key: workspace_id
+                - key: customer_id
+                  value: bigfish
+                token_bucket:
+                  allow: 50_000   # 50KB/s allowance per pipeline PER CORE
+                  burst: 100_000  # 100KB maximum size PER CORE
+                cardinality:
+                  max_count: 1000 # Up to 1000 workspaces (single customer)
```
lquerel: This is a policy :-)

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+              required_descriptors: [onbehalfof_tenant]
+              # ...
+
+        nodes:
+          otlp:
+            type: receiver:otlp
+            capabilities:
+              # Multiple rate limiters are listed with numeric sorting key "/N".
+              rate_limit/0: obo_rate
+              rate_limit/1: customer_rate
```
lquerel: How does that work when the number of tenant is high/dynamic?

rust/otap-dataflow/docs/agent-multitenancy-design.md
```diff
@@ -0,0 +1,877 @@
...
+```
+
+In this example, where multiple limiters are bound to a node as with
+`rate_limit/0` and `rate_limit/1` above, they are evaluated in
+ascending numeric order and all must grant for the request to proceed.
+If a limiter failure causes the request to short-circuit, the granted
+requests will be cancelled. Specific implementation details are
+out-of-scope, see open questions.
+
+### Limiter fairness
```
lquerel: This document/topic is a big piece of work, so I'll stop my review here for now. I'm sure some of my questions/comments will be answered later, but I hope some of them will remain valid even after a full review.

In any case, this document represents a major specification and design effort. I think we should not hesitate to spend time on it to make sure we're aligned on the approach.

lquerel (commented):
Thanks, Josh, for this PR. For now, I'm only doing a partial review, as explained in my last comment. I'll continue tomorrow.

jmacd:
I will re-open after incorporating feedback.

PR #3396, opened 2026-07-01
fix(journald): follow new entries for start_at: end (anchor after seek_tail)

### LINKED ISSUE #3395: journald receiver: `start_at: end` never follows new entries (missing sd_journal_previous after seek_tail)
### Component

`receiver:journald` — `rust/otap-dataflow/crates/core-nodes` (`src/receivers/journald_receiver`).

### What happens

With `start_at: end` and no existing checkpoint, a freshly started journald
receiver emits ~zero records forever (single active journal file) or a stale
one-entry-per-file chunk (rotated / multi-file journals). It never reaches the
live tail and never follows entries appended after startup.

### Root cause

`SdJournalReader::configure()` seeks with `sd_journal_seek_tail()` for
`StartAt::End`, and the worker read loop (`next_entry_with_wait_timeout`) then
only ever calls `sd_journal_next()` + `sd_journal_wait()`.

Per sd-journal semantics:

- `sd_journal_seek_tail(3)` parks the read head *after* the most recent entry
  and does **not** make any entry current.
- `sd_journal_next(3)`: "if the current location is the tail of the journal,
  this is equivalent to `sd_journal_previous()`, and advanced will be ignored."
  So `next()` at the tail moves **backward**, not forward.

A bare `next()`/`wait()` loop therefore never sails off the end onto entries
appended later. Because it never reads an entry, it never Acks, so it never
commits a checkpoint cursor — and the (correct) checkpoint resume path
(`seek_cursor()` + `next()`) never takes over on restart. A fresh
`start_at: end` deployment is stuck at zero permanently.

### Correct idiom

`seek_tail()` → `sd_journal_previous()` (anchor on the last existing entry) →
then the existing `wait()`/`next()` follow loop, so the first `next()` steps
*forward* onto genuinely new entries.

References:

- `sd_journal_next(3)` / `sd_journal_seek_tail(3)` man pages
- systemd/systemd#17662 ("seek_tail, previous, next ==> perfectly working, the
  journal reports EoF")
- coreos/go-systemd `sdjournal` `NumFromTail` (`SeekTail` → `PreviousSkip` →
  `Next`)

The `StartAt::Beginning` branch (`seek_head()` + `next()`) is already correct;
only the tail branch is affected.

### Fix

Add the `sd_journal_previous` binding and call it once after `seek_tail()` in
the fresh-start `StartAt::End` path so the follow loop anchors on the last
existing entry before advancing.

timr-dev:
## Summary

Fixes #3395. A fresh `start_at: end` journald receiver (no checkpoint) emitted
**zero records forever** and never committed a checkpoint.

`SdJournalReader::configure()` positioned `StartAt::End` with only
`sd_journal_seek_tail()`, and the worker follow loop only calls
`sd_journal_next()` + `sd_journal_wait()`. Per sd-journal semantics:

- `sd_journal_seek_tail(3)` parks the read head *after* the most recent entry
  and makes **no** entry current.
- From there a bare `sd_journal_next(3)` advances toward a *following* entry,
  finds none, and returns `0` (the EOF marker) **without anchoring** — it does
  not move backward. (It is `sd_journal_step_one(3)`, not `sd_journal_next(3)`,
  that libsystemd documents as behaving like `sd_journal_previous()` at the
  tail.)

So a bare `next()`/`wait()` loop never advances onto entries appended after
startup. Because it never reads an entry it never Acks, so it never commits a
cursor, and the (correct) checkpoint-resume path never takes over on restart.
`start_at: end` was permanently stuck at zero.

## Fix

Fresh `StartAt::End` now does `seek_tail()` → `previous()` to anchor on the last
existing entry, so the worker's first `next()` steps forward onto genuinely new
entries. When `previous()` returns `0` (empty journal, or a filter that matches
none of the existing entries) it anchors nothing and the head stays parked at
the tail — the same permanent stall — so the head is repositioned with
`seek_head()`; since no entry matched yet, `next()` then follows the first entry
appended after startup with no risk of replaying old records.

The seek/step FFI is abstracted behind a small `JournalSeek` seam so the
positioning logic (`position_for_fresh_start`) — including the empty-journal
path and error propagation — is unit-tested with a recording mock.
`StartAt::Beginning` (`seek_head()` + `next()`) and the checkpoint-resume
(`seek_cursor()` + `next()`) paths are unchanged.

Refs: `sd_journal_next(3)`, `sd_journal_seek_tail(3)`, systemd/systemd#17662,
coreos/go-systemd `sdjournal` tail-follow.

## Testing

**Unit tests** (`position_for_fresh_start` via a recording mock; run on all
platforms): `Beginning → [seek_head]`; `End` populated → `[seek_tail, previous]`;
`End` empty → `[seek_tail, previous, seek_head]`; `previous` error; `seek_tail`
error.

**Verified against a live journald** (mirroring the receiver's exact sequence; a
matching entry is appended *after* the seek):

| sequence | pre-existing matching entries | result |
| --- | --- | --- |
| `seek_tail` only (old behavior) | yes | STUCK — reproduces the bug |
| `seek_tail` + `previous` | yes | follows new only, no replay |
| `seek_tail` + `previous` | none | STUCK |
| `seek_tail` + `previous` + `seek_head` (this PR) | none | follows new |

`cargo check` / `clippy --all-targets -D warnings` / `cargo fmt` are clean on
Linux; unit tests pass.

## Performance

This is a startup-path fix; the per-entry read hot path is unchanged.

- **Hot path untouched.** The diff is confined to the one-time `configure()`
  positioning path, the `JournalSeek` seam, and the changelog. The per-entry
  read loop (`next_entry_with_wait_timeout` → `current_entry`) is byte-for-byte
  identical to `main`, so steady-state throughput and per-entry allocations are
  unaffected.
- **Startup cost.** Fresh `StartAt::End` now issues one extra
  `sd_journal_previous()`, plus one conditional `sd_journal_seek_head()` only
  when no matching entry exists yet. These run once per journal open (worker
  start) and once per rare `Rewind`/Nack — never per entry; the cost is one-time
  and microsecond-scale.
- **No dynamic dispatch.** `position_for_fresh_start` is generic over
  `JournalSeek` and monomorphized (static dispatch), so the seam adds no vtable
  or allocation overhead.

## Caveats (scope)

`seek_tail()` + `previous()` is the accepted best-effort idiom
(systemd/systemd#17662). Across rotated / multi-file journals the tail position
is approximate, and an entry appended in the race window between `seek_tail()`
and `previous()` may be anchored and skipped. This is acceptable under
`start_at: end`, which has no durable resume anchor until the first checkpoint
commit (already documented in the receiver README/design notes).

The empty/no-match handling (`seek_head()` when `previous()` returns `0`) is
replay-free because a well-behaved `previous()` returns `0` only when no
matching entry exists yet. This was verified end-to-end against live journald on
modern systemd (255), including a 20-file rotated journal with archived-only
matches. A very old/buggy libsystemd of the systemd#17662 class that spuriously
reports `0` while matching entries exist would replay that history once at
startup; this was not observed on modern systemd, and the receiver `dlopen`s the
host libsystemd at runtime.

### DIFF
--- a/rust/otap-dataflow/.chloggen/journald-start-at-end-follow-tail.yaml
+++ b/rust/otap-dataflow/.chloggen/journald-start-at-end-follow-tail.yaml
@@ -1,0 +1,36 @@
+# Use this changelog template to create an entry for release notes.
+
+# One of 'breaking', 'deprecation', 'new_component', 'enhancement', 'bug_fix'
+change_type: bug_fix
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: pipeline
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: >-
+  The journald receiver now follows newly appended entries when `start_at: end`
+  and no checkpoint exists — including on an empty journal, or one whose filters
+  match none of the existing entries. Previously it sought the journal tail but
+  never anchored the read head, so the follow loop never advanced onto new
+  entries: a fresh `start_at: end` receiver emitted no records and never
+  committed a checkpoint.
+
+# Mandatory: One or more tracking issues related to the change. You can use the PR number here if no issue exists.
+issues: [3395]
+
+# (Optional) One or more lines of additional information to render under the primary note.
+# These lines will be padded with 2 spaces and then inserted directly into the document.
+# Use pipe (|) for multiline entries.
+subtext: >-
+  `sd_journal_seek_tail()` parks the read head after the most recent entry
+  without making any entry current, and `sd_journal_next()` at the tail is
+  documented to behave like `sd_journal_previous()` (it moves backward). The
+  fresh `StartAt::End` start now issues `sd_journal_seek_tail()` then a single
+  `sd_journal_previous()` to anchor on the last existing entry, so the worker's
+  first `next()` steps forward onto genuinely new entries. When `previous()`
+  finds no entry (empty or fully-filtered journal) the head is repositioned with
+  `sd_journal_seek_head()` to avoid a permanent stall at the tail. This is the
+  accepted best-effort idiom (see systemd/systemd#17662); across rotated or
+  multi-file journals the tail position is approximate, and `start_at: end` has
+  no durable resume anchor until the first checkpoint commit.
--- a/rust/otap-dataflow/.chloggen/journald-start-at-end-follow-tail.yaml
+++ b/rust/otap-dataflow/.chloggen/journald-start-at-end-follow-tail.yaml
@@ -24,9 +24,12 @@ issues: [3395]
 # Use pipe (|) for multiline entries.
 subtext: >-
   `sd_journal_seek_tail()` parks the read head after the most recent entry
-  without making any entry current, and `sd_journal_next()` at the tail is
-  documented to behave like `sd_journal_previous()` (it moves backward). The
-  fresh `StartAt::End` start now issues `sd_journal_seek_tail()` then a single
+  without making any entry current; from there a bare `sd_journal_next()`
+  advances toward a following entry, finds none, and returns `0` (the EOF
+  marker) without anchoring; it does not move backward. (It is
+  `sd_journal_step_one()`, not `sd_journal_next()`, that libsystemd documents
+  as behaving like `sd_journal_previous()` at the tail.) The fresh
+  `StartAt::End` start now issues `sd_journal_seek_tail()` then a single
   `sd_journal_previous()` to anchor on the last existing entry, so the worker's
   first `next()` steps forward onto genuinely new entries. When `previous()`
   finds no entry (empty or fully-filtered journal) the head is repositioned with
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
@@ -30,25 +30,34 @@ trait JournalSeek {
 /// forward from the first entry (the documented `SD_JOURNAL_FOREACH` idiom).
 ///
 /// `StartAt::End` is subtle. `sd_journal_seek_tail()` parks the read head
-/// *after* the most recent entry without making any entry current, and
-/// `sd_journal_next()` at the tail is documented to behave like
-/// `sd_journal_previous()` (it moves backward and ignores the advance). A bare
-/// `next()`/`wait()` follow loop therefore never advances onto entries appended
-/// after startup. So we step back once with `previous()` to anchor on the last
-/// existing entry; the worker's first `next()` then steps forward onto genuinely
-/// new entries.
+/// *after* the most recent entry without making any entry current. From there a
+/// bare `sd_journal_next()` advances toward a *following* entry, finds none, and
+/// returns `0` (the documented EOF marker) without anchoring — so a plain
+/// `next()`/`wait()` follow loop started at the raw tail never advances onto
+/// entries appended after startup (verified against real journald). (It is
+/// `sd_journal_step_one()`, not `sd_journal_next()`, that libsystemd documents as
+/// behaving like `sd_journal_previous()` at the tail.) So we step back once with
+/// `previous()` — documented to seek the closest *preceding* entry — to anchor on
+/// the last existing entry; the worker's first `next()` then steps forward onto
+/// genuinely new entries.
 ///
 /// When the journal is empty — or a filter matches none of the existing
 /// entries — `previous()` returns `0` and anchors nothing, leaving the read head
 /// parked at the tail where `next()` keeps returning `0` even after `wait()`
 /// reports appends (the same permanent stall the tail branch exists to avoid,
 /// verified against real journald). In that case we reposition to the head with
-/// `seek_head()`; because a well-behaved `previous()` returns `0` only when no
-/// matching entry exists yet, `next()` then follows onto the first entry
-/// appended after startup without replaying history. This assumes a reliable
-/// `seek_tail()` + `previous()`; a very old/buggy libsystemd of the
-/// systemd#17662 class that spuriously reports `0` while matching entries exist
-/// would replay that history once at startup (not observed on modern systemd).
+/// `seek_head()` so the follow loop can make progress. This does not replay
+/// history: `sd-journal` merges *all* open journal files (active plus
+/// rotated/archived) into a single view, and with the receiver's matches
+/// installed a well-behaved `previous()` returns `0` only when no matching entry
+/// exists in ANY of them yet — so `seek_head()` + `next()` land on the first
+/// matching entry appended *after* startup, with nothing older to replay. Replay
+/// would require `seek_tail()` + `previous()` to spuriously report `0` while
+/// matching entries actually exist (the multi-file positioning bug of the
+/// systemd#17662 class noted below); on such an old/buggy libsystemd it would
+/// replay that history once at startup (not observed on modern systemd). Bounding
+/// replay even under that bug would need a durable tail-boundary guard, which
+/// `start_at: end` deliberately omits (see the resume-anchor note below).
 ///
 /// Returns `Err((operation, rc))` with the failing call's name and its negative
 /// return code. `seek_tail`/`seek_head` return `0` on success; `previous`
--- a/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
+++ b/rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
@@ -3,13 +3,202 @@
 
 //! Linux `sd-journal` reader abstraction.
 
+#[cfg(any(target_os = "linux", test))]
+use crate::receivers::journald_receiver::config::StartAt;
+#[cfg(any(target_os = "linux", test))]
+use std::ffi::c_int;
+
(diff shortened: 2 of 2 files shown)


rust/otap-dataflow/.chloggen/journald-start-at-end-follow-tail.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: pipeline
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: >-
+  The journald receiver now follows newly appended entries when `start_at: end`
+  and no checkpoint exists — including on an empty journal, or one whose filters
+  match none of the existing entries. Previously it seeked the journal tail but
```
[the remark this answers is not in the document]
timr-dev: Thanks — addressed in `7b6b78b`: changed "seeked" to "sought" in the changelog note.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
```diff
@@ -3,13 +3,202 @@
...
+}
+
+/// Positions a freshly opened journal (no checkpoint cursor) for `start_at`.
+///
+/// `StartAt::Beginning` seeks the head; the worker's `next()` then iterates
+/// forward from the first entry (the documented `SD_JOURNAL_FOREACH` idiom).
+///
+/// `StartAt::End` is subtle. `sd_journal_seek_tail()` parks the read head
+/// *after* the most recent entry without making any entry current, and
+/// `sd_journal_next()` at the tail is documented to behave like
```
lalitb: nit - I think this explanation slightly overstates the sd-journal API contract. The “tail behaves like previous” wording is documented for `sd_journal_step_one()`, not directly for `sd_journal_next()`. The `seek_tail() -> previous()` idiom may still be right, but can we reword this so it does not attribute that behavior to `sd_journal_next()`?

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
```diff
@@ -3,13 +3,202 @@
...
+        if rc < 0 { Err((operation, rc)) } else { Ok(()) }
+    }
+
+    match start_at {
+        StartAt::Beginning => require_nonneg(seek.seek_head(), "sd_journal_seek_head"),
+        StartAt::End => {
+            require_nonneg(seek.seek_tail(), "sd_journal_seek_tail")?;
+            let anchored = seek.previous();
+            require_nonneg(anchored, "sd_journal_previous")?;
+            if anchored == 0 {
```
lalitb, line 87: non-blocking -  Good to add one behavior-level regression test for the empty/no-match `start_at: end` case. The new unit tests cover the seek call sequence, but the original bug is in the `next()`/`wait()` follow behavior after startup, so a test that exercises that path would make this safer.

lalitb, line 87: This fallback worries me for `start_at: end`. If `seek_tail() -> previous()` returns 0 even though older matching entries exist in rotated/archived journal files, `seek_head()` makes the worker replay old logs, which breaks the “new entries only” contract. I realize a full tail-boundary guard may be a larger change, but can we at least avoid this fallback unless we can prove there are no existing matches, or document why replay is acceptable here?

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
timr-dev: Good catch — you're right, fixed in d16bbe4b3. The "tail behaves like `previous()`" wording in the `sd_journal_next(3)` man page is documented for `sd_journal_step_one()` (v254+), not for `sd_journal_next()`. A plain `next()` at the tail just advances toward a *following* entry, finds none, and returns `0` (the EOF marker) — it doesn't move backward. I reworded the comment to describe that actual `next()` behavior (and to attribute the "acts like `previous()` at the tail" semantics to `step_one()`), while keeping the `seek_tail()` -> `previous()` anchoring rationale intact.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
timr-dev: Fair concern, and thanks for pushing on it — replaying history under `start_at: end` would indeed break the "new entries only" contract. Here's the reasoning, which I've now made explicit in the doc comment (d16bbe4b3):

`sd-journal` exposes a single merged view over *all* open journal files — active plus rotated/archived — so with our matches installed, `seek_tail()` -> `previous()` traverses every file, not just the active one. A well-behaved `previous()` therefore returns `0` only when no matching entry exists in *any* of them yet; in that case `seek_head()` -> `next()` lands on the first entry appended *after* startup, with nothing older to replay.

The one path that would replay is `seek_tail()` + `previous()` spuriously reporting `0` while matching entries actually exist — the multi-file positioning bug of the systemd#17662 class (not observed on modern systemd), which the comment already flags. A hard guarantee even under that bug would need a durable tail-boundary guard (capture the tail cursor, refuse to emit at/before it) — the larger change you alluded to; `start_at: end` deliberately omits it since it has no durable resume anchor until the first checkpoint commit.

I went with documenting the boundedness to keep this PR focused on the stall fix, but I'm glad to open a follow-up to track the tail-boundary guard if you'd prefer that hardening captured separately. Happy to go either way.

timr-dev: Agreed, that's a good idea — the current `fresh_start_tests` only assert the seek call sequence (`seek_tail` -> `previous` -> `seek_head`), not the `next()`/`wait()` follow behavior after startup, which is exactly where the original bug lived.

The follow loop itself (`next_entry_with_wait_timeout`) is the real libsystemd FFI path, so a faithful behavior-level test needs either (a) an integration test against a live journald, or (b) making that loop generic over a step/wait seam the way `position_for_fresh_start` is already generic over `JournalSeek`, so an in-memory fake can drive the full position-then-follow sequence (empty start -> append -> assert only the new entry is followed, no replay). I'd lean toward (b) as a small, focused follow-up rather than bolt a possibly-flaky journald dependency onto CI — I can validate it against a real journald locally either way. Since you flagged this as non-blocking, I'll track it separately unless you'd like it folded into this PR.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
lalitb: Thanks. This looks good now. One small follow-up: I think the PR body and changelog still have the older wording saying `sd_journal_next()` is documented to behave like `previous()` at the tail. Could we update those too, so the public PR/ release-note text matches the corrected code comment?

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
lalitb: That sounds good to me. Keeping the behavior-level test as a follow-up is fine, especially if the alternative is adding a live journald dependency to CI.

lalitb: Thanks, that explanation makes sense. I’m okay with keeping the fallback in this PR, since the modern-systemd behavior was validated and the replay risk is now documented.

A follow-up issue for a stronger tail-boundary guard sounds useful, but I don’t think it needs to block this fix.

lalitb (approved):
LGTM - with nit on updating the changelog and PR body . Thanks.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
timr-dev: Done — updated both to match the corrected code comment:

- **Changelog** (`journald-start-at-end-follow-tail.yaml`, commit `aec6ffac7`): the subtext no longer says `sd_journal_next()` is documented to behave like `previous()` at the tail. It now describes a bare `next()` at the tail as advancing toward a following entry, finding none, and returning `0` (EOF) without anchoring, and notes that the "behaves like `previous()`" wording belongs to `sd_journal_step_one()`, not `sd_journal_next()`.
- **PR body**: the Summary bullet is reworded the same way — dropped the "equivalent to `sd_journal_previous()` … moves backward" claim for `sd_journal_next()`.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs
timr-dev: Filed #3399 to track this. It captures the behavior-level regression test for the empty/no-match `start_at: end` follow path — exercising the `next()`/`wait()` path after startup (append a matching entry after the seek, assert it is delivered and nothing older is replayed), via either a Linux-gated live-journald integration test or a fake `sd-journal` backend behind the FFI seam so it stays in the cross-platform unit suite. Thanks for the suggestion.

timr-dev: Thanks. I have opened #3399 to track the stronger tail-boundary guard as a follow-up — bounding the one-time replay that a systemd#17662-class libsystemd could cause when `previous()` spuriously returns `0` — so it can be evaluated independently of this fix.

jmacd (approved):

Issue #3398, opened 2026-07-01
journald receiver: reduce per-entry allocations in current_entry (avoid MESSAGE double-copy)


timr-dev:
### Context

Follow-up from the performance review of #3396 (journald `start_at: end`
follow-tail fix). That fix is startup-path only and does **not** touch the
per-entry hot path. This issue tracks a hot-path allocation optimization
identified during that review, deliberately kept out of #3396 to keep the
correctness fix focused.

### Opportunity

`SdJournalReader::current_entry()`
(`rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/journal.rs`)
runs once per journal entry and allocates:

- `cursor` — one `String` per entry
- `fields` — one `Vec<JournalField>` per entry
- per field: a `name` `String` and a `value` `Vec<u8>`
- **`message_body = Some(value.clone())`** — clones the `MESSAGE` field value
  (typically the largest field) even though the same bytes are also pushed into
  `fields`, so the message payload is allocated **twice** per entry.

### Proposed work

- Eliminate the `message_body` double-copy: hold the `MESSAGE` bytes once (e.g.
  store an index/marker into `fields`, or restructure `JournalEntry`) and have
  the arrow encoder read from that single copy.
- Investigate reusing per-entry buffers (`cursor` / field `name` / field
  `value`) across the read loop to cut steady-state allocation pressure at high
  log volume.

### Testing / methodology

- Build a benchmark harness for the read path (the crate already uses
  `criterion`; the journald reader is FFI + live-journald, so the harness likely
  needs a live journal or a synthetic `JournalEntry` construction seam).
- Measure per-entry allocations + throughput before/after (p50/p95/p99, mean,
  stdev) with count/content validation each iteration.
- Full multi-model review panel before opening the PR.

### Scope / sequencing

Land as a separate, benchmark-driven PR after #3396 merges. No behavior change
to the emitted records — pure allocation/throughput optimization.

PR #3403, opened 2026-07-01
perf(journald): avoid cloning the MESSAGE payload per entry

timr-dev:
## Summary

Addresses the `MESSAGE` double-copy in #3398.

`SdJournalReader::current_entry()` cloned each entry's `MESSAGE` field value into
`JournalEntry.message_body` while the same bytes were also pushed into `fields`,
so the message payload — usually the largest field in an entry — was allocated
and copied **twice per record**.

This stores `message_body_index: Option<usize>` (the index of the chosen
`MESSAGE` field in `fields`) instead of a second owned copy, and reads the body
back through a new `JournalEntry::message_body()` accessor. One `Vec<u8>`
allocation + memcpy of the message payload is removed per entry that has a
`MESSAGE`; no new allocations are introduced (the index is a `usize`).

## Output is unchanged (contract preserved)

`append()` still emits the log body from the selected `MESSAGE` and still emits
`MESSAGE` as an attribute. The body source is the **first non-dropped** `MESSAGE`
chosen at decode time: when the first `MESSAGE` is dropped for size and a later
one survives, the body must stay unset. Storing the exact chosen index preserves
this. The pinned encoder tests, including that edge case
(`leaves_body_unset_when_first_message_was_dropped`), pass unchanged.

## Testable decode + patch coverage

The per-field decode loop lived inside the Linux-only `imp::current_entry` FFI
method, so its changed lines could not be unit-tested (no live journal) and were
left uncovered (failing `codecov/patch`). It is extracted into a pure,
cross-platform `decode` module as a stateful `FieldDecoder` (`new` / `feed` /
`finish`), and `current_entry` is now a thin FFI adapter over it. **13 new unit
tests** cover the drop, truncation, `MESSAGE`-selection, duplicate-`MESSAGE`,
non-UTF-8, field/entry-cap and empty-message paths — moving the previously
uncovered logic into covered code.

## Sound FFI field iteration

`sd_journal_enumerate_data` returns a pointer valid only until the **next**
enumerate call. The reader exposes it through a *lending*
`EnumerateData::next_field(&mut self) -> Option<&[u8]>`: the returned slice
borrows `&mut self`, so the borrow checker forbids fetching (or aliasing) another
field while it is held — modelling the FFI's lifetime exactly. It is deliberately
**not** a `std::iter::Iterator`, whose non-lending `Item` would let safe code
`collect()` the slices into a use-after-free.

## Benchmark (decode-only)

`benches/journald_decode` (criterion, behind the `bench` feature —
`cargo bench -p otap-df-core-nodes --features bench`) isolates the **field-decode
step only** — it is *not* an end-to-end receiver benchmark (after decode the
receiver still copies each kept field into the Arrow builders). A one-time local
comparison (Linux, LTO off) of the decode with vs without the removed `MESSAGE`
clone, per entry:

| `MESSAGE` size | decode (index body) | decode + body clone (pre-PR) | delta |
| --- | --- | --- | --- |
| 256 B | ~0.70 µs | ~0.90 µs | clone **+28 %** |
| 4 KiB | ~0.94 µs | ~1.00 µs | clone **+7 %** |
| 64 KiB | ~2.53 µs | ~4.08 µs | clone **+61 %** (−1.55 µs) |

The saving is the eliminated per-entry allocation + memcpy of the `MESSAGE`
payload; **decode-side** it is near the noise floor for tiny messages and
dominant for large ones (~38 % faster decode at 64 KiB). Per reviewer feedback
the live benchmark now times only the current decode; the comparison above is
preserved as a documented reference result in the bench's module doc.

## Testing

- `cargo test -p otap-df-core-nodes` — 13 new `decode` unit tests + the existing
  encoder output-contract tests (body + attribute projection, dropped-first-
  `MESSAGE` edge case) pass, confirming byte-identical OTLP output.
- Linux (WSL): `cargo clippy --lib --tests --benches --features bench -- -D warnings`
  clean; the full journald suite (61 tests, incl. the Linux-only `imp` FFI tests)
  passes. `cargo build` clean (no dead-code).
- `cargo fmt` clean.

Per-entry buffer reuse (`cursor` / field `name` / `value`) — the secondary item
in #3398 — is deliberately left out; it reworks `JournalEntry`/`JournalField`
ownership more broadly and warrants its own PR, so #3398 stays open to track it.


rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/arrow_records_encoder.rs
```diff
@@ -26,7 +26,11 @@ pub(crate) struct JournalField {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub(crate) struct JournalEntry {
     pub(crate) cursor: String,
-    pub(crate) message_body: Option<Vec<u8>>,
+    /// Index into `fields` of the `MESSAGE` field used as the log body (the
+    /// first non-dropped `MESSAGE`), or `None` when the entry has no usable
+    /// message. Stored as an index so the message payload is held once (in
+    /// `fields`) instead of being cloned into a second buffer.
```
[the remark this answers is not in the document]
timr-dev: [GS] Good catch — reworded. The `message_body_index` doc now states the exact contract: it is the entry's **first** `MESSAGE`, but only if it survived extraction. If that first `MESSAGE` was dropped (e.g. oversized), the body stays `None` even when a later `MESSAGE` remains in `fields`. That matches the decoder (a dropped first `MESSAGE` permanently poisons the body index) and the pinned `leaves_body_unset_when_first_message_was_dropped` test. Fixed in dc4ba1bb.

_Reply posted by [GS] @timr-dev Copilot AI Assistant._

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/arrow_records_encoder.rs
```diff
@@ -45,6 +49,16 @@ impl JournalEntry {
             .find(|field| field.name == key)
             .map(|field| field.value.as_slice())
     }
+
+    /// Returns the log body: the value of the `MESSAGE` field selected during
+    /// decoding (`message_body_index`), or `None` if the entry has no usable
+    /// message. Reads the bytes back from `fields` so the message payload is not
+    /// duplicated.
```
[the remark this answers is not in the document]
timr-dev: [GS] Reworded `message_body()`'s doc to drop the ambiguous "selected" wording: it now says it returns the value at `message_body_index` — the entry's first `MESSAGE` if it was kept, otherwise `None` (including the case where the first `MESSAGE` was dropped but a later one survives in `fields`) — and that it reads the bytes back from `fields`, so the message payload is not duplicated. Fixed in dc4ba1bb.

_Reply posted by [GS] @timr-dev Copilot AI Assistant._

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/arrow_records_encoder.rs
```diff
@@ -45,6 +51,17 @@ impl JournalEntry {
             .find(|field| field.name == key)
             .map(|field| field.value.as_slice())
     }
+
+    /// Returns the log body: the value at `message_body_index` — the entry's
+    /// first `MESSAGE` field if it was kept, otherwise `None` (including the
+    /// case where the first `MESSAGE` was dropped but a later `MESSAGE` survives
+    /// in `fields`). Reads the bytes back from `fields`, so the message payload
+    /// is not duplicated.
+    pub(crate) fn message_body(&self) -> Option<&[u8]> {
```
lalitb, line 60: Small safety guard: since `message_body_index` now depends on the position in `fields`, can we add a `debug_assert` here that the indexed field name is `MESSAGE`? It is correct today because fields are not changed between decode and append, but this would catch future refactors that reorder or drop fields before encoding.

rust/otap-dataflow/crates/core-nodes/benches/journald_decode/main.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! Benchmark proving the per-entry `MESSAGE`-clone elimination in the journald
```
lalitb: Could we mark this as a decode-only benchmark in doc comment here, and store the baseline result here with the run context? Right now the 28–61% numbers in the PR description are easy to read as end-to-end receiver throughput, but this benchmark only isolates the decode-side `MESSAGE` clone removal; the full receiver path still copies kept fields into the Arrow builders afterward.

We already keep this kind of reference result in other benchmark code, for example:
 - https://github.com/open-telemetry/otel-arrow/blob/474998ab250661c7f54ab1819aa0029a06afbb3f/rust/otap-dataflow/crates/contrib-nodes/benches/exporters/azure_monitor_exporter/transformer.rs#L4-L9

lalitb (approved):
LGTM. Thanks for optimizing this. Have couple of nit comments, and should be good to merge then.

rust/otap-dataflow/crates/core-nodes/benches/journald_decode/main.rs
```diff
@@ -0,0 +1,63 @@
...
+    for msg_len in [256usize, 4 * 1024, 64 * 1024] {
+        let owned = build_entry(msg_len);
+        let raw: Vec<&[u8]> = owned.iter().map(Vec::as_slice).collect();
+        let _ = group.throughput(Throughput::Bytes(msg_len as u64));
+        let _ = group.bench_with_input(BenchmarkId::new("index", msg_len), &raw, |b, raw| {
+            b.iter(|| black_box(bench_reference_decode(black_box(raw), false)));
+        });
+        let _ = group.bench_with_input(BenchmarkId::new("clone", msg_len), &raw, |b, raw| {
+            b.iter(|| black_box(bench_reference_decode(black_box(raw), true)));
+        });
```
albertlockett: Before merging this, maybe we should delete this benchmark case, the associated `clone` arg to the `bench_reference_decode`, and any comments referencing "the change"?

I'm happy that this benchmark was created to prove the performance improvement of this change. However, in its current state this benchmark seems like it was only written to make sense in the context of this PR. E.g. having this case that that copies the message only makes sense in the context of this change, but doesn't make sense in the context of the code long-term because after this PR merges the copy will no longer be happening.

lalitb: Good point. The `clone` case was useful to validate this PR, but I agree it does not need to live in the benchmark long-term once the copy path is gone. Removing that case and the related argument/comments sounds right.

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/arrow_records_encoder.rs
timr-dev: [GS] Added — `message_body()` now `debug_assert_eq!`s that the field at `message_body_index` is named `MESSAGE`, so a future refactor that reorders or drops fields between decode and encode trips it in debug/test builds. It's debug-only (no release cost on the per-entry accessor), and `fields.get(index)?` still returns `None` safely if the index is ever out of range. be952efe

_Reply posted by [GS] @timr-dev Copilot AI Assistant._

rust/otap-dataflow/crates/core-nodes/benches/journald_decode/main.rs
timr-dev: [GS] Done. The bench module doc now (a) states it is a **decode-only** microbenchmark and explicitly notes it is *not* end-to-end receiver throughput — after decode the receiver still copies each kept field into the Arrow builders — and (b) stores the clone-vs-index comparison as a documented "Reference result" with run context (Linux, LTO off), matching the `azure_monitor_exporter/transformer.rs` convention you linked. I also reworded the PR description's benchmark section to scope the numbers as decode-only. be952efe

_Reply posted by [GS] @timr-dev Copilot AI Assistant._

rust/otap-dataflow/crates/core-nodes/benches/journald_decode/main.rs
timr-dev: [GS] Agreed and done. Removed the live `clone` benchmark case, dropped the `clone_body` arg from `bench_reference_decode` (it now just times the real decode as it ships), and removed the comments framing it against "the change". The clone-vs-index comparison that validated this PR is preserved as a documented reference result in the bench module doc (per @lalitb's related thread) rather than as a live case. be952efe

_Reply posted by [GS] @timr-dev Copilot AI Assistant._

rust/otap-dataflow/crates/core-nodes/src/receivers/journald_receiver/arrow_records_encoder.rs
ThomsonTan: Probably also add a runtime check after debug_assert_eq!, like below because the debug_assert is removed in release build.

```rs
    if field.name != "MESSAGE" {
        return None;
    }
```

PR #3405, opened 2026-07-01
OTAP Parquet study

jmacd:
Compares the cost of OTAP/IPC transport vs Parquet.

Then, compares a flattened schema compatible with Parquet that we might called OTAP-flat as opposed to OTAP-standard, which is denormalized. Points out that ~10 of the columns in the OTAP logs table convert directly into Parquet, so the cost of OTAP-to-Parquet-to-OTAP can be improved using zero-copy for the bulk of columns. Also considers whether OTAP-flat with RLE instead of materialized is competitive: it is, but it doesn't convert well to Parquet because Parquet is a quite-old format.


rust/otap-dataflow/benchmarks/benches/otap_parquet/README.md
```diff
@@ -0,0 +1,253 @@
...
+| 1,000  | zstd | 24,748  | 13,422  | 11,326 | 19,512    | 0.69x   |
+| 10,000 | zstd | 92,270  | 80,944  | 11,326 | 57,023    | 1.42x   |
+| 50,000 | zstd | 401,966 | 390,640 | 11,326 | 240,211   | 1.63x   |
+
+OTAP-flat single columnar view. `convert` is the cost of turning the four OTAP
+record batches into one, `view-mem` is the in-memory footprint of that view, and
+`pq-write`/`pq-bytes` are the Parquet encode where arrow-rs can write it. The
+`otap-flat-*` rows exploit the fact that OTAP attribute batches are already
+grouped by `parent_id`, so the per-record log attributes are a zero-copy
+`List<Struct>` and only the layout of the shared resource/scope sets differs. The
```
albertlockett, line 138: > the fact that OTAP attribute batches are already grouped by `parent_id`

Not necessarily true - this is a happy accident of how OTAP is produced in most situations (e.g. when converting from OTLP), but I wouldn't advise relying on it systematically

rust/otap-dataflow/benchmarks/benches/otap_parquet/README.md
```diff
@@ -0,0 +1,253 @@
...
+between, because they share ten of the flat table's thirteen columns and only the
+attribute containers change:
+
+| edge                               | log-heavy | resource-heavy |
+| ---------------------------------- | --------: | -------------: |
+| S  -> F   flatten to REE           |       6.1 |            1.7 |
+| F  -> S   unflatten                |      12.9 |            2.9 |
+| S  -> Ws  standard serialize       |      62.9 |           25.8 |
+| Ws -> S   standard deserialize     |       9.9 |            4.5 |
+| F  -> Wf  flat serialize           |      37.0 |           12.2 |
```
albertlockett, line 178: I find the `S -> Ws` and `F -> Wf` numbers surprising here. Isn't this just doing arrow IPC serialization? That should be pretty cheap as its mostly just copying the arrow buffers.

rust/otap-dataflow/benchmarks/benches/otap_parquet/README.md
```diff
@@ -0,0 +1,253 @@
...
+| parquet-nested | 240,211 | 329,298 | 405,397 | 2,118,185  |
+| parquet-wide   | 245,757 | 327,547 | 327,294 | 344,708    |
+
+OTAP/IPC pipeline breakdown, milliseconds:
+
+| comp | t-enc | ipc-ser | enc-tot | ipc-des | t-dec | dec-tot |
+|------|-------|---------|---------|---------|-------|---------|
+| zstd | 13.0  | 5.9     | 18.9    | 3.0     | 1.2   | 4.2     |
+| lz4  | 13.5  | 2.6     | 16.1    | 18.7    | 1.3   | 20.0    |
+| none | 12.6  | 29.6    | 42.2    | 27.1    | 1.2   | 28.3    |
```
albertlockett, line 109: Strange that the no-compression case performs the worst

albertlockett: I see that elsewhere we explain this by there being less data to copy. 

I guess that makes sense because we are copying 400k vs 12mb into this vec here?
https://github.com/open-telemetry/otel-arrow/blob/f74d7114d1209758b3b1c429c54cff06d8713401/rust/otap-dataflow/benchmarks/src/parquet_study/ipc.rs#L43-L51

FYI, the IPC writer also has an number of data copies, but there is work ongoing to resolve clean it up (https://github.com/apache/arrow-rs/pull/10044 and https://github.com/apache/arrow-rs/pull/10128)

rust/otap-dataflow/benchmarks/benches/otap_parquet/OTAP_FLAT_ANALYSIS.md
```diff
@@ -0,0 +1,453 @@
...
+The struct children are the existing `LogAttrs` value columns used as they are,
+with no `take`, and the list offsets come from a single linear scan that walks
+the sorted `parent_id` runs alongside the sequential `Logs.id` column. Every log
+attribute belongs to exactly one record, so no value is copied and no value is
+shared.
+
+Second, the shared resource and scope attributes line up with runs of log rows,
+because the records are already grouped by resource and then by scope. A resource
+therefore spans a contiguous block of rows, which is exactly the structure that
+run-end and dictionary encodings capture without physically repeating anything.
```
albertlockett, line 63: Although this is how the encoder works today, there's nothing that guarantees it has to work like that or that this sort-order needs to be maintained. Indeed, when we insert new attributes for example, we append them all to the end of the attrs record batch which breaks this assumption. 

Also, if the attributes are were transport encoded, even if we remove the delta encoding we don't reorder the attribute table as far as I know. So again this assumption about the sort-order is incorrect.

FWIW when we proto-encode from OTAP, we have a similar problem where we need to group up attributes for some parent ID. In this case, we create these cursors for which rows to visit in order by sorting the ID columns.
https://github.com/open-telemetry/otel-arrow/blob/f74d7114d1209758b3b1c429c54cff06d8713401/rust/otap-dataflow/crates/pdata/src/otlp/common.rs#L1082-L1240

rust/otap-dataflow/benchmarks/benches/otap_parquet/OTAP_FLAT_ANALYSIS.md
```diff
@@ -0,0 +1,453 @@
...
+still depend on cardinality: correlated logs that share a trace id would compress
+better and lower every number together, so read the comparison as one point on a
+spectrum rather than a fixed ratio.
+
+Two arrow-rs 58.3 limits shape the results. The writer cannot serialize run-end
+or nested-dictionary columns, so those layouts are in-memory forms only, and the
+reader cannot read a dictionary-encoded `FixedSizeBinary` such as `trace_id`
+back, so the study materializes those columns before writing Parquet. Both may
+lift as arrow-rs adds support, at which point the run-end view could also become
+a Parquet write target.
```
albertlockett, line 428: REE support in parquet is being actively worked on https://github.com/apache/arrow-rs/pull/10064

jmacd:
I didn't mean to open this PR in this repo!!!!! It was to be in my fork, for saving.

jmacd:
Thank you @albertlockett this was meant to be for discussion really really not ready to open for wide review. Thank you.

It leads me to believe we have a viable flat-Parquet story but also teaches me that Parquet is very expensive to write.

albertlockett:
> I didn't mean to open this PR in this repo!!!!! It was to be in my fork, for saving.

> Thank you @albertlockett this was meant to be for discussion really really not ready to open for wide review. Thank you.
> 
> It leads me to believe we have a viable flat-Parquet story but also teaches me that Parquet is very expensive to write.

Happy to have reviewed - this was interesting to read! 

I'm not too surprised - Parquet is quite aggressive about adding optimizing the storage of the columns. IIRC it tries to apply dictionary encoding + RLE + Bitpacking for most columns by default.

PR #3407, opened 2026-07-02
feat(azure_monitor_exporter): add attribute passthrough mode

gouslu:
## Summary

Adds an **attribute passthrough** mode to the Azure Monitor exporter so users can emit all log record attributes without enumerating each one in the schema.

Passthrough is **explicit and opt-in inside the schema** — set `log_record_mapping.attributes` to the string `passthrough`:

```yaml
schema:
  log_record_mapping:
    attributes: passthrough
```

Each log record attribute is emitted **as-is as its own top-level `"<key>": <value>` column** (the attribute key is the column name). Omitting `schema` entirely still means "map nothing" — dropping the block does not silently switch modes.

## Behavior

- **Flat columns:** attributes are emitted as top-level `"key": value` pairs, not nested under a wrapper object.
- **Composable:** passthrough works alongside explicit `resource_mapping`, `scope_mapping`, and top-level `log_record_mapping` field mappings, which continue to emit their own columns.
- **Innermost wins on collision:** if a passed-through attribute key equals a mapped column name, the attribute value (the most specific / innermost value) wins and the column is emitted exactly once — output stays valid JSON with unique keys.
- **Ingestion note:** attribute keys are emitted verbatim, so a key such as `service.name` must have a matching, validly-named column in the DCR stream schema; otherwise Azure Log Analytics drops that column at ingestion. Documented in the README and example config.

## Implementation / performance

- Pure passthrough and pure mapped records stay on the fast path (build into a reused `Vec`, one bulk copy into a reused `BytesMut`, `split().freeze()` for zero-copy handoff — benchmarked faster than writing directly into `BytesMut`).
- The collision-aware combined path runs only when passthrough is combined with mappings. Its per-attribute collision check compares attribute-key bytes directly (no per-attribute `from_utf8_lossy`) and is gated by a length bitmask, so most attributes are rejected without any byte comparison.

## Changes

- `config.rs`: `attributes` accepts either the `passthrough` marker (string) or the existing per-attribute mapping object; `schema` stays a plain `SchemaConfig` (omitting it maps nothing).
- `transformer.rs`: flat passthrough emission; collision-aware combined path with innermost-wins; reserved-column detection (bytes + length bitmask).
- README, example `otlp-ame-passthrough.yaml`, changelog, and benches (added a `combined` benchmark mode) updated.

## Testing

- `cargo test -p otap-df-contrib-nodes --features azure-monitor-exporter` — 228 passed (incl. passthrough, empty attributes, composition, resource/scope/field collision, and config validation)
- `cargo clippy --all-targets --features azure-monitor-exporter -D warnings`, `cargo fmt --all --check`, `markdownlint`, `python3 tools/sanitycheck.py` — all clean
- `cargo bench --bench azure_monitor_transformer` (1000 records): passthrough ~290us, mapped ~320us, combined ~370us; fast paths unchanged vs baseline


rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
```diff
@@ -0,0 +1,64 @@
+# rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
+# Azure Monitor Exporter Configuration using attribute passthrough mode.
+#
+# This configuration omits the `schema` block entirely. When `schema` is not
+# provided, the exporter runs in attribute passthrough mode: every log record
+# attribute is emitted as a JSON key/value pair using the attribute key as the
```
cijothomas: what about non-attributs? i.e top level fields? I Assume they are als exported as-is? Could that be clarified in the example and in the changelog/readme.

cijothomas: also, please explicitly clarrify in doc what happens to Resource/Instrumentation scope too -are they are passthrough-ed or dropped.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
```diff
@@ -227,16 +236,49 @@ impl Transformer {
...
     }
 
+    /// Passthrough conversion: emit every log record attribute as a JSON key/value
+    /// pair, using the attribute key as the column name. No field, resource, or
+    /// scope mappings are applied.
+    fn convert_passthrough<T: LogsDataView>(logs_view: &T) -> Vec<Bytes> {
+        let mut results = Vec::with_capacity(1024);
+        let mut record_buf = Vec::with_capacity(512);
+
+        for resource_logs in logs_view.resources() {
```
cijothomas: I think this should export all the fields - resource, scope, log-record top-level fields, not just log-record attributes.

cijothomas (dismissed):
https://github.com/open-telemetry/otel-arrow/pull/3407/changes#r3515142097
I think we should export every fields if schema is not specified, not just log attribtues alone.

Let me know if there is a reason why we specifically handle attributes alone and ignore all top-level fields on log.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
```diff
@@ -227,16 +236,49 @@ impl Transformer {
...
+                    record_buf.clear();
+                    record_buf.push(b'{');
+                    let mut has_field = false;
+                    for attr in log_record.attributes() {
+                        if let Some(val) = attr.value() {
+                            if has_field {
+                                record_buf.push(b',');
+                            }
+                            has_field = true;
+                            Self::write_json_string(attr.key(), &mut record_buf);
```
cijothomas: don't we need sanitization of keys? not all OTel keys are valid LA column names I think... Not super sure, need to confirm. (eg: service.name )

cijothomas:
Suggestion for passthrough mode: instead of emitting each attribute as its own column (raw OTel keys like `service.name` aren't valid Log Analytics column names and get silently dropped at ingestion), use a fixed default schema:

- **Top-level fields → pass through with their OTel names** (`severity_text`, `severity_number`, `body`, `trace_id`, `span_id`, `flags`, `event_name`, etc.). They're already valid column names, so no need to remap. The one exception: `time_unix_nano` must populate the required `TimeGenerated` column.
- **Attributes → a single `dynamic` column** (e.g. `Attributes`). Dotted keys are legal as JSON keys inside the value, so no sanitizing/collision issues.
- **Resource and scope attributes → their own dynamic columns** (`ResourceAttributes`, `ScopeAttributes`) so nothing is dropped.

This avoids the column-name sanitizing problem entirely, keeps the required DCR schema small/fixed/documentable, ensures `TimeGenerated` is always populated, and preserves nested attribute values.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
gouslu: We want log attributes passthrough as an option. No other fields or attributes. We could have passthrough as an alternative to schema configuration where you specify what kind of passthrough it is and if there needs to be additional fields or stuff like that, but that's not what we need right now. What kind of changes do you recommend as minimum to enable log attributes passthrough only, if you don't want it to be the default when no schema is provided?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
gouslu: Requirement is to pass log attributes as is, open to talk about what changes you require to make it possible. https://github.com/open-telemetry/otel-arrow/pull/3407#discussion_r3515527238

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
gouslu: We want passthrough mode "as is". We don't want sanitization, but we can drop invalid keys entirely if that's what you want.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
cijothomas: If the intent is purely **log-attribute passthrough as-is**, then making the *entire* `schema` optional is the wrong lever. "Omit the whole schema block => silently export only attributes" is surprising and easy to misconfigure - a user dropping `schema` would reasonably expect *everything* to pass through, not attributes-only.

Passthrough is really just a special case of **attribute mapping** (identity mapping), so let's express it *inside* the schema config rather than by its absence. A couple of options:

1. Keep `schema` required, and add an explicit passthrough marker on the attribute mapping, e.g.:
   ```yaml
   schema:
     log_record_mapping:
       attributes: passthrough   # emit every log attribute as-is, key = column name
   ```
2. Or a dedicated boolean, e.g. `attributes_passthrough: true` under the attribute mapping section.

Either way the behavior is **opt-in and explicit**, resource/scope/top-level fields are clearly untouched (not accidentally dropped), and the config self-documents what "passthrough" means. Omitting the entire `schema` should probably stay an error, or mean "map nothing," rather than silently switching modes.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
cijothomas: As long as the behavior is documented, I am fine. For example, if we are going to sanitize and convert invalid characters to supported ones - that is fine by me. Or if we are going to drop those keys - that is also fine. If the drop occurs at the exporter - then we could do internal log/metric about this, so users will have a way of knowing. Need to check what happens if we send an invalid column name - does LA reject that entire payload? or just that column? and Does it expose something about the data loss?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
```diff
@@ -194,7 +199,13 @@ pub struct ApiConfig {
     /// Data Collection Rule identifier
     pub dcr: String,
 
-    /// Schema mapping configuration
+    /// Schema mapping configuration.
+    ///
+    /// When omitted, no mappings are applied (rows are emitted empty). To emit
+    /// all log record attributes as-is, set `log_record_mapping.attributes` to
```
cijothomas: please move this comment to the most apt place - which'd be the below IMHO:

  /// Log record field mappings
    #[serde(default)]
    pub log_record_mapping: HashMap<String, Value>,

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
```diff
@@ -290,40 +301,57 @@ impl Config {
...
 
-        for (key, value) in &self.api.schema.log_record_mapping {
-            match value {
-                Value::String(s) if !seen.insert(s.clone()) => {
-                    _ = duplicates.insert(s.clone());
-                }
-                Value::Object(map) => {
-                    if key != "attributes" {
+        for (key, value) in &schema.log_record_mapping {
+            if key == "attributes" {
```
cijothomas, line 242: Small usability follow-up (non-blocking): `attributes` is currently either `passthrough` or an explicit mapping - mutually exclusive. A user who wants to rename one attribute (e.g. `user.id` → `UserId` for a valid LA column) but pass the rest through has no clean option; they'd have to enumerate everything.

Fine to ship as-is - a hybrid form can be added later additively without breaking existing configs.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
cijothomas: also, we are now losing the ability to validate the config for possible duplicates - lets document it so users who leverage this new capability is aware and that the behavior of overriding with innermost is also documented.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
```diff
@@ -59,6 +59,9 @@ config:
     dcr_endpoint: "https://my-workspace.eastus-1.ingest.monitor.azure.com"
     stream_name: "Custom-MyLogTable_CL"
     dcr: "dcr-abc123def456"
+    # Schema mapping is optional. To emit all log record attributes as-is,
```
cijothomas: the comment is not suited at this level. Move it to log_record_mapping.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
```diff
@@ -209,6 +212,31 @@ log_record_mapping:
...
+In this mode every log record attribute is written as-is as a top-level
+`"<key>": <value>` pair, using the attribute key as the column name. Passthrough
+composes with resource, scope, and top-level field mappings, which continue to
+emit their own columns. If an attribute key collides with a mapped column name,
+the attribute value wins (the innermost/most-specific value), and the column is
+emitted only once.
+
+Note: Azure Log Analytics only ingests columns that exist in the DCR stream
+schema and whose names satisfy the column-naming rules (letters, digits, and
+underscores; must start with a letter or underscore). Attribute keys are emitted
```
cijothomas: please document/warn that invalid column would result in dataloss.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
```diff
@@ -0,0 +1,78 @@
...
+            type: "urn:microsoft:exporter:azure_monitor"
+            config:
+              # API configuration - Update with your actual values
+              api:
+                dcr_endpoint: "https://my-workspace.eastus-1.ingest.monitor.azure.com"
+                stream_name: "Custom-MyTable_CL"
+                dcr: "dcr-abc123"
+                # Attribute passthrough: emit each log attribute as-is as its own
+                # top-level "key": value column. No resource/scope or top-level
+                # field mappings are applied here.
```
cijothomas: does the mandatory timestamp field gets populated? won't LA reject logs if that is not set?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
```diff
@@ -59,6 +59,9 @@ config:
     dcr_endpoint: "https://my-workspace.eastus-1.ingest.monitor.azure.com"
     stream_name: "Custom-MyLogTable_CL"
     dcr: "dcr-abc123def456"
+    # Schema mapping is optional; only the sections you configure are emitted
```
cijothomas, line 103: its not optional right? do we accept a config that does not include a schema: section?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
```diff
@@ -209,6 +215,41 @@ log_record_mapping:
...
+```yaml
+schema:
+  log_record_mapping:
+    attributes: passthrough
+```
+
+In this mode every log record **attribute** is written as-is as a top-level
+`"<key>": <value>` pair, using the attribute key as the column name. Passthrough
+applies to log-record attributes only: resource attributes, scope attributes,
+and top-level log-record fields are **not** passed through automatically. They
```
cijothomas: why talk about resource or scope here? this is log_record section only right?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
```diff
@@ -209,6 +215,41 @@ log_record_mapping:
...
+Because passthrough column names come from runtime attribute keys, they are not
+validated for duplicates at config load time the way explicit mappings are;
+collisions are resolved at emit time by the innermost-wins rule above.
+
+> **Warning -- silent data loss:** Azure Log Analytics only ingests columns that
+> exist in the DCR stream schema and whose names satisfy the column-naming rules
+> (letters, digits, and underscores; must start with a letter or underscore).
+> Attribute keys are emitted verbatim and are **not** sanitized, so a key such as
+> `service.name` (containing a `.`) or any key without a matching, validly-named
+> DCR column is **silently dropped at ingestion** -- the exporter does not rewrite
```
cijothomas: is it silent drop? or its just silent from this exporter perspective?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
gouslu: right, not after the requested changes.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
gouslu: to reduce ambiguity. is this a nit comment or do you want it to be changed?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
```diff
@@ -0,0 +1,78 @@
...
+                schema:
+                  log_record_mapping:
+                    attributes: passthrough
+
+              # Authentication configuration for AKS
+              auth:
```
ThomsonTan: The exporter-local auth model was removed by #3470. Probably declare an `azure_identity_auth` extension and bind `capabilities.bearer_token_provider` on this exporter.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
gouslu: it seems like a silent drop based on my research.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
gouslu: no, it adds it itself based on my testing.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
```diff
@@ -270,6 +276,56 @@ log_record_mapping:
...
+are emitted (each to its own column). To emit **all** log record attributes
+without enumerating them, set `attributes` to the string `passthrough`:
+
+```yaml
+schema:
+  log_record_mapping:
+    attributes: passthrough
+```
+
+In this mode every log record **attribute** is written as-is as a top-level
```
lalitb: Are we sure flat top-level columns are the right passthrough shape here? Since Azure still requires matching valid DCR columns, this avoids listing attributes in exporter config but not really in the Azure schema. Would one dynamic `Attributes` column, or support for both flat and dynamic modes, be safer?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/transformer.rs
```diff
@@ -227,16 +347,62 @@ impl Transformer {
...
+                // Reject most attribute keys on length alone (no hashing); only
+                // an actual collision (rare) materializes the key as a String.
+                if check_reserved && schema.is_reserved_column(attr.key()) {
+                    overridden.push(String::from_utf8_lossy(attr.key()).into_owned());
+                }
+            }
+        }
+
+        for fm in &schema.field_mappings {
+            if !overridden.is_empty() && overridden.contains(&fm.dest_name) {
```
lalitb, line 433: Should runtime attributes be able to override explicitly mapped columns? With the current “attribute wins” rule, an incoming attribute named `TimeGenerated`, `Body`, or a mapped resource column can replace the configured mapping for that record. That makes part of the stable output schema controlled by incoming data. Would it be safer for explicit mappings to win by default, or for this to be a configurable collision policy?

gouslu: I think innermost wins is a good general rule that we can apply with or without explicit mapping. The data that is fed into the exporter is also fed into the exporter intentionally, which can be manipuated via kql projections other processors or overall intentionally putting data into the attributes bag. I'd like to keep a blanket rule here across all mappings, saying that innermost data wins, which can keep the logic less error prone and prevent mental gymnastics around how the data is transformed should these rules and exceptions get more and more complex.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
gouslu: adding it in transformer if not mapped, it should be a better behavior overall if there are delays and retries.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
```diff
@@ -0,0 +1,93 @@
+# rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
+# Azure Monitor Exporter Configuration using attribute passthrough mode.
+#
+# This configuration sets `schema.log_record_mapping.attributes` to the string
```
cijothomas: The comment is overly verbose.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
```diff
@@ -0,0 +1,93 @@
...
+# sanitize keys.
+#
+# Passthrough applies to log-record attributes only. Resource attributes, scope
+# attributes, and top-level log-record fields are NOT passed through
+# automatically; they are emitted only when explicitly mapped via
+# resource_mapping / scope_mapping / log_record_mapping (which compose with
+# passthrough). Attributes already mapped explicitly are skipped so they are not
+# duplicated.
+#
+# WARNING - silent data loss: Azure Log Analytics only ingests columns defined in
```
cijothomas: what do we mean by **silent** data loss? 
Lets be very clear if we are going to lose entire row or just the column.
silent - silent from the exporter perspective.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
```diff
@@ -0,0 +1,93 @@
...
+      pipeline: 100
+    pdata: 128
+
+engine: {}
+
+groups:
+  default:
+    pipelines:
+      main:
+        # Authentication is provided by an extension, not exporter config: the
```
cijothomas: not sure why is this verbose comment on auth for an config example that is supposed to show passthrough mode?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
```diff
@@ -0,0 +1,93 @@
...
+              # client_id: "YOUR-USER-ASSIGNED-IDENTITY-CLIENT-ID"
+              scope: "https://monitor.azure.com/.default"
+
+        nodes:
+          otlp-receiver:
+            type: "urn:otel:receiver:otlp"
+            config:
+              protocols:
+                grpc:
+                  # Listen on all interfaces in container
```
cijothomas: its better to remove such comments that is not really adding any value.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
```diff
@@ -0,0 +1,93 @@
...
+              # API configuration - Update with your actual values
+              api:
+                dcr_endpoint: "https://my-workspace.eastus-1.ingest.monitor.azure.com"
+                stream_name: "Custom-MyTable_CL"
+                dcr: "dcr-abc123"
+                # Attribute passthrough: emit each log attribute as-is as its own
+                # top-level "key": value column. No resource/scope or top-level
+                # field mappings are applied here.
+                #
+                # TimeGenerated is not mapped here; Log Analytics auto-populates it
```
cijothomas: this does not seem correct as the PR now automatically adds TimeGenerated.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
```diff
@@ -0,0 +1,93 @@
...
+                # with the ingestion time when a record omits it, so records are not
+                # rejected for a missing timestamp. Map time_unix_nano -> TimeGenerated
+                # under log_record_mapping to set it explicitly.
+                schema:
+                  log_record_mapping:
+                    attributes: passthrough
+
+              # Heartbeat configuration (optional, disabled by default)
```
cijothomas: if its disabled by default already, why we configure it again here and keep disabled?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
cijothomas: sorry I am confused - is it optional or not? And is the comment reflecting the actual implementation?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
```diff
@@ -118,6 +121,9 @@ config:
         "time_unix_nano": "TimeGenerated"
         "trace_id": "TraceId"
         "span_id": "SpanId"
+        # `attributes` is either an explicit per-attribute mapping (below) or the
+        # string `passthrough` to emit every log attribute as-is as its own
+        # top-level "key": value column. See "Attribute Passthrough Mode" below.
```
cijothomas: nit: you can link to the specific section

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
```diff
@@ -270,6 +276,76 @@ log_record_mapping:
...
+`"<key>": <value>` pair, using the attribute key as the column name. Passthrough
+applies to log-record attributes only: resource attributes, scope attributes,
+and top-level log-record fields are **not** passed through automatically. They
+are emitted only when you configure them via `resource_mapping`, `scope_mapping`,
+or the other `log_record_mapping` fields, which compose with passthrough and
+continue to emit their own columns. If a runtime attribute key collides with a
+mapped column name, the attribute value wins (the innermost/most-specific value),
+and the column is emitted only once.
+
+This override is **intentional, not accidental**: a runtime attribute can
```
cijothomas: such wordings are good for PR desc, not for user facing readme doc.
Also, this is overly verbose.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
cijothomas: few lines below here, it says data loss can be detected on the Azure side. which one is correct?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
```diff
@@ -270,6 +276,76 @@ log_record_mapping:
...
+> the sender** (no error is returned), but it is **observable to the workspace
+> operator**: ingestion emits a dropped-column count metric and records the
+> unmatched column names, so data loss can be detected on the Azure side even
+> though this exporter cannot surface it.
+
+Note on the mandatory `TimeGenerated` column: a log record that does not carry a
+`TimeGenerated` value does **not** get rejected -- Azure Log Analytics ingestion
+auto-populates `TimeGenerated` with the ingestion time when it is absent. To
+control it explicitly, map a log-record field to it (e.g.
+`log_record_mapping: { time_unix_nano: TimeGenerated }`).
```
cijothomas: this will occur automatically without user having to explicitly list this mapping?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/README.md
gouslu: it means that I agree with you, you are right, it was leftover from the first iteration. it is already removed.

jmacd:
Based on https://github.com/open-telemetry/otel-arrow/pull/3407#issuecomment-4869145412 I would like to see @cijothomas approve.

rust/otap-dataflow/.chloggen/azure-monitor-schema-passthrough.yaml
```diff
@@ -0,0 +1,19 @@
...
+change_type: enhancement
+
+# The name of the component, or a single word describing the area of concern, (e.g. engine, otap).
+# Must be one of the components listed in rust/otap-dataflow/.chloggen/config.yaml.
+component: pipeline
+
+# A brief description of the change.  Surround your text with quotes ("") if it needs to start with a backtick (`).
+note: Add an attribute passthrough mode to the Azure Monitor exporter. Setting `schema.log_record_mapping.attributes` to `passthrough` emits every log record attribute as-is as its own top-level column named by the attribute key, avoiding per-attribute mapping configuration. Only log-record attributes are passed through; resource attributes, scope attributes, and top-level log-record fields are not passed through automatically and are emitted only when explicitly mapped. Passthrough composes with resource, scope, and top-level field mappings; when an attribute key collides with a mapped column, the attribute value wins and the column is emitted once. Attribute keys are emitted verbatim (not sanitized), so keys without a valid matching DCR column are dropped at ingestion.
```
cijothomas: > Only log-record attributes are passed through; resource attributes, scope attributes, and top-level log-record fields are not passed through automatically and are emitted only when explicitly mapped. 


minor: this (^) can be removed.


Suggest to use a shorter writeup:
Add an attribute passthrough mode to the Azure Monitor exporter. Setting schema.log_record_mapping.attributes to passthrough emits every log-record attribute as its own top-level column named by the attribute key. See the exporter README for scoping, collision, and DCR-column-name rules.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
```diff
@@ -105,7 +110,15 @@ pub struct ApiConfig {
...
     pub dcr: String,
 
-    /// Schema mapping configuration
+    /// Schema mapping configuration.
+    ///
+    /// When omitted (or left empty), no mappings are applied: only the
+    /// mandatory `TimeGenerated` column is auto-injected (from the record's
+    /// event time), and every other section is empty. Only the sections you
+    /// configure are emitted; unmapped resource attributes, scope attributes,
+    /// and log-record fields are dropped, not passed through. See
```
cijothomas: too much details. we already state no mapping applied in the very beginning, why repeat same information again?

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/otlp-ame-passthrough.yaml
```diff
@@ -0,0 +1,68 @@
+# Azure Monitor exporter, attribute passthrough mode.
```
cijothomas: nit: Don't think we need a separate example config just to show passthrough. its simple enough.

rust/otap-dataflow/crates/contrib-nodes/src/exporters/azure_monitor_exporter/config.rs
```diff
@@ -137,7 +150,24 @@ pub struct SchemaConfig {
     #[serde(default)]
     pub scope_mapping: HashMap<String, String>,
 
-    /// Log record field mappings
+    /// Log record field mappings.
```
cijothomas, line 150: suggested rephrasing:

```
/// Log record field mappings.
///
/// Each key maps a log-record field (`body`, `severity_text`, `time_unix_nano`,
/// …) to a destination column name. The special key `attributes` accepts either
/// an explicit `{ <attr key>: <column> }` object OR the string `passthrough`;
/// the two forms cannot be combined within `attributes` (all-or-nothing).
///
/// In `passthrough`, every log-record attribute is emitted as-is as a top-level
/// `"<key>": <value>` column. Other `log_record_mapping` fields, plus
/// `resource_mapping` and `scope_mapping`, continue to emit their own columns
/// independently. If a passthrough attribute key collides with a mapped column,
/// the attribute value wins (innermost) and the column is emitted once.
///
/// Passthrough column names are runtime attribute keys, so unlike explicit
/// mappings they are not checked for duplicates at config load; collisions are
/// resolved at emit time by the rule above.
```

cijothomas (approved):
LGTM. left couple of non-blocking suggestions.

cijothomas:
> Based on [#3407 (comment)](https://github.com/open-telemetry/otel-arrow/pull/3407#issuecomment-4869145412) I would like to see @cijothomas approve.

Approved.

rust/otap-dataflow/crates/contrib-nodes/benches/exporters/azure_monitor_exporter/transformer.rs
```diff
@@ -36,31 +67,34 @@ fn create_config() -> Config {
...
-                    (
-                        "attributes".into(),
-                        json!({
-                            "env": "Environment",
-                            "request.id": "RequestId",
-                            "user.id": "UserId"
-                        }),
-                    ),
-                ]),
+                resource_mapping: HashMap::new(),
```
ThomsonTan, line 70: This does reset the baseline for the benchmark. Is it necessary?

PR #3410, opened 2026-07-03
feat: add opamp controller extension

albertlockett:
# Change Summary

Adds initial implementation of OpAMP Agent Controller Extension.

Design doc can be found in #3388 

Adds:
- controller extension
- OpAMP proto definitions & prost generated structs

Currently only works with websocket.

Followups will need to include:
- [plain HTTP transport](https://opentelemetry.io/docs/specs/opamp/#plain-http-transport) (plain meaning traditional HTTP request/response, not necessarily plaintext)
- mTLS
- connection setting management (this is ignored)
- metrics collection

Currently this lives in the controller crate. I wasn't sure where was the right place for this, and hesitated between controller or core-nodes. Happy to move if anyone has feelings/suggestions

* Relates to #3387 

### Changelog


rust/otap-dataflow/crates/controller/src/extension/opamp/config.rs
```diff
@@ -0,0 +1,302 @@
...
+use crate::extension::opamp::util::ExponentialBackoff;
+
+/// Configuration for OpAMP Controller Extension
+#[derive(Deserialize, Serialize)]
+pub struct Config {
+    /// Initial instance_uid.
+    ///
+    /// May be overridden at runtime by server sending a new agent identity.
+    ///
+    /// If not provided, a UUIDv7 will be generated.
```
cijothomas, line 24: wondering if should mandate this and not make it optional.. If we generate new one each time, then "identity" is kinda lost. Or perhaps warn user about this risk?

rust/otap-dataflow/crates/controller/src/extension/opamp/config.rs
```diff
@@ -0,0 +1,302 @@
...
+    pub instance_uid: Option<String>,
+
+    /// Endpoint for OpAMP server.
```
cijothomas, line 27: lets make explicitly clear we support http and ws both, and the url will identify which one is used.
(I assume we support both)

rust/otap-dataflow/crates/controller/src/extension/opamp/config.rs
```diff
@@ -0,0 +1,302 @@
...
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use url::Url;
+
+use crate::extension::opamp::error::Error;
+use crate::extension::opamp::proto::opamp::v1::{AnyValue, any_value::Value};
+use crate::extension::opamp::util::ExponentialBackoff;
+
+/// Configuration for OpAMP Controller Extension
+#[derive(Deserialize, Serialize)]
+pub struct Config {
```
cijothomas, line 19: I don't see any TLS related setting - its covered, or its follow up candidate?

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1854 @@
+// Copyright The OpenTelemetry Authors
+// SPDX-License-Identifier: Apache-2.0
+
+//! OpAMP controller extension.
+//!
+//! This allows the engine to be controlled by a connection to a remote OpAMP server.
+//!
+//! Example:
```
cijothomas, line 8: could we include a readme.md file with instructions to run the OpAMP-Go server too, and have df_engine report to it, so reviewers can run things side by side as they review :) 

I know the PR is marked draft, but too excited to try this out!

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1854 @@
...
+    Ok(Box::new(move |cancellation_token| {
+        Box::pin(
+            async move { run_websocket_connect_loop(context, config, cancellation_token).await },
+        )
+    }))
+}
+
+/// State of session between OpAMP Agent (this client) and control planes OpAMP server
+struct SessionState {
+    /// OpAMP instance UID. This will be initialized from the K8s Pod's UID. If this hasn't
```
cijothomas: nit: Avoid K8s mention.

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1854 @@
...
+
+        match connect_result {
+            Ok((stream, _response)) => {
+                return Some(stream);
+            }
+
+            // backoff and retry connection
+            Err(e) => {
+                let retry_in = backoff.next_delay();
+                otel_error!(
```
cijothomas: given we are retrying, and not giving up, `warn!` might be more apt here.

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1854 @@
...
+                    .run_until_cancelled(tokio::time::sleep(Duration::from_nanos(wait_ns)))
+                    .await
+                    .is_none()
+                {
+                    // cancellation token was cancelled during backoff
+                    return stats;
+                }
+
+                let Some(last_message) = session_state.last_sent_message.as_ref() else {
+                    otel_error!(
```
cijothomas:
```suggestion
                    otel_warn!(
```

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1854 @@
...
+                            );
+                            return stats
+                        };
+
+                        let message = match received_message {
+                            Ok(m) => m,
+                            Err(e) => {
+                                otel_error!(
+                                    "opamp.controller_extension.receive_message.error",
+                                    message = "Error received while awaiting message",
```
cijothomas: message is not adding additional value, and may be omitted.

rust/otap-dataflow/crates/controller/src/extension/opamp/config.rs
albertlockett: I'm not sure about this. The OpAMP docs seem to indicate that a generated UUIDv7 is the suggested approach.

> The Agent SHOULD self-generate this identifier and make the best effort to avoid creating an identifier that may conflict with identifiers created by other Agents. The instance_uid SHOULD remain unchanged for the lifetime of the Agent process. The instance_uid MUST be 16 bytes long and _**SHOULD be generated using the [UUID v7 spec](https://www.ietf.org/archive/id/draft-ietf-uuidrev-rfc4122bis-14.html#name-uuid-version-7).**_

https://opentelemetry.io/docs/specs/opamp/#agenttoserverinstance_uid

In terms of the agent identity getting lost, the `AgentToServer` message also contains an `AgentDescription` message which. I have the impression that the `identifying_attributes` could be used to identify agent in the case where the UUID is not identifying.

rust/otap-dataflow/crates/controller/src/extension/opamp/config.rs
albertlockett: Yes - TLS is a followup candidate

rust/otap-dataflow/crates/controller/src/extension/opamp/config.rs
albertlockett: Updated docs in 7530bab77

This PR only adds support for WebSocket transport, but I think eventually we should support both. HTTP transport support will need to be another followup.

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: fixed in 6ef8bf631

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: good call, fixed in e0aecff8e

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: good call, fixed in b7f0ff162

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: :+1: fixed in 8c1d64a7a

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1853 @@
...
+    }
+}
+
+/// Report supported capabilities for [`AgentToServer::capabilities`] field
+fn capabilities() -> u64 {
+    (AgentCapabilities::ReportsStatus as u64)
+        | (AgentCapabilities::ReportsEffectiveConfig as u64)
+        | (AgentCapabilities::ReportsHealth as u64)
+        | (AgentCapabilities::ReportsRemoteConfig as u64)
+        | (AgentCapabilities::ReportsHeartbeat as u64)
```
lalitb, line 1303: As discussed in the design doc, should we add `AcceptsRemoteConfig` here too? The implementation applies `remote_config`, and a compliant OpAMP server should not send remote config unless the agent advertises that capability.

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1853 @@
...
+                    return stats;
+                }
+
+                stats.messages_sent += 1;
+                session_state.phase = Phase::ExchangeMessages;
+            }
+            Phase::ExchangeMessages => {
+                tokio::select! {
+                    _ = cancellation_token.cancelled() => {
+                        return stats;
```
lalitb, line 375: On cancellation, we return without sending `AgentDisconnect`. This seems different form what specs [say](https://opentelemetry.io/docs/specs/opamp/#websocket-transport-opamp-client-initiated):

```
WebSocket Transport, OpAMP Client Initiated

To close a connection the Client MUST first send an AgentToServer message with agent_disconnect field set. The Client MUST then send a WebSocket [Close](https://datatracker.ietf.org/doc/html/rfc6455#section-5.5.1) control frame and follow the procedure defined by WebSocket standard.
```

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1853 @@
...
+    if let Some(agent_identification) = message.agent_identification {
+        updates.agent_identification = Some(agent_identification)
+    }
+
+    if let Some(remote_config) = message.remote_config {
+        if let Some(config) = remote_config.config {
+            // we expect to find remote a key called "desired_state" in the remote config_map.
+            // All other keys within the config_map will be ignored...
+
+            if let Some(config_file) = config.config_map.get("desired_state") {
```
lalitb: I think the design doc says we only reconcile a remote config when its `config_hash` differs from the last applied hash. While this path seems to reconcile every `remote_config`. Probably  we skip reconcile when `remote_config.config_hash == session_state.last_config_hash` so repeated offers are idempotent?

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1853 @@
...
+impl SessionState {
+    fn try_new(config: &Config) -> Result<Self, Error> {
+        let instance_uid = match &config.instance_uid {
+            Some(instance_uid) => {
+                // safety: this was validated in validate_config
+                Uuid::parse_str(instance_uid).expect("valid UUID")
+            }
+            None => Uuid::now_v7(),
+        }
+        .to_bytes_le();
```
lalitb: I’m not sure we want little-endian UUID bytes here. OpAMP just treats `instance_uid` as 16 UUID bytes, and most UUID tooling expects the normal/canonical byte order. With `to_bytes_le()`, the UUID string from config will not round-trip to the same UUID on the server side.

Should this use `uuid.as_bytes()` instead, so the configured `instance_uid` is the same one the OpAMP server sees?

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1853 @@
...
+                // time to update config
+                let reconcile_result =
+                    context
+                        .control_plane
+                        .reconcile_engine_config(EngineConfigReconcileRequest {
+                            config: engine_config.engine_config,
+                            step_timeout_secs: config.reconcile.step_timeout_secs,
+                            drain_timeout_secs: config.reconcile.drain_timeout_secs,
+                            delete_timeout_secs: config.reconcile.delete_timeout_secs,
+                            delete_missing: true,
```
lalitb: I think this hardcoding to `true` will make every remote config a full replacement and can delete existing pipelines if the server sends a partial update. Can we make this configurable and pass the configured value through?

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: Yes, you're right. Added in [b729b31](https://github.com/open-telemetry/otel-arrow/pull/3410/commits/b729b31847bddfa8822e1e3171d887d1549326b6)

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: Yes you're right - fixed in e12dbf498

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: yes you're right. Fixed in [b729b31](https://github.com/open-telemetry/otel-arrow/pull/3410/commits/b729b31847bddfa8822e1e3171d887d1549326b6)

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: Added in 92ee543d7.  @lquerel actually had similar feedback from the design doc https://github.com/open-telemetry/otel-arrow/pull/3388#discussion_r3521734020

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: added in [b729b31](https://github.com/open-telemetry/otel-arrow/pull/3410/commits/b729b31847bddfa8822e1e3171d887d1549326b6) :)

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: yes good catch. Made this change in [d32026d](https://github.com/open-telemetry/otel-arrow/pull/3410/commits/d32026d78c8e6343e1fbeb7c7a9a9146e2647157)

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1926 @@
...
+                let mut backoff = config.request_retry.clone();
+                if let Some(last_backoff) = session_state.retry_backoff_ns {
+                    backoff.set_current(Duration::from_nanos(last_backoff));
+                }
+                session_state.retry_backoff_ns = Some(backoff.next_delay().as_nanos() as u64);
+            }
+
+            false
+        }
+        HandledMessageAction::Restart => {
```
lalitb: Now that we advertise `AcceptsRestartCommand`, should this be wired to a real agent/process restart?

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1926 @@
...
+                    return stats;
+                }
+
+                stats.messages_sent += 1;
+                session_state.phase = Phase::ExchangeMessages;
+            }
+            Phase::ExchangeMessages => {
+                tokio::select! {
+                    _ = cancellation_token.cancelled() => {
+                        try_send_close_message(&mut ws_sender, config.shutdown_timeout).await;
```
lalitb: Thanks for adding the WebSocket close frame. I think this still misses the OpAMP part though: before the close frame, the client should send an `AgentToServer` message with `agent_disconnect` set. Right now this only sends the WebSocket Close frame, so the server still does not receive the OpAMP disconnect message.

albertlockett: I'm sorry about that Lalit, you're right. I should have read your previous feedback more carefully!

This has been corrected here: 7f7a5c38f

rust/otap-dataflow/crates/controller/src/extension/opamp/config.rs
cijothomas: okay got it.

rust/otap-dataflow/crates/controller/src/extension/opamp/README.md
```diff
@@ -0,0 +1,123 @@
...
+> -   Body: []byte(agent.CustomInstanceConfig),
+> + cfg.Config.ConfigMap["desired_state"] = &protobufs.AgentConfigFile{
+> +   Body:        []byte(agent.CustomInstanceConfig),
+> +   ContentType: "application/json",
+>   }
+> ```
+
+### 2. Configure and run the engine
+
+Save as `opamp.yaml`:
```
cijothomas: lets have a working minimal file in the configs directory.

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1995 @@
...
+    }
+
+    if let Some(agent_identification) = message.agent_identification {
+        updates.agent_identification = Some(agent_identification)
+    }
+
+    if let Some(remote_config) = message.remote_config {
+        if let Some(config) = remote_config.config {
+            if Some(&remote_config.config_hash) != session_state.last_config_hash.as_ref() {
+                // we expect to find remote a key called "desired_state" in the remote config_map.
```
cijothomas: Following up from the design-doc thread on `desired_state`: https://github.com/open-telemetry/otel-arrow/pull/3388#discussion_r3526014319

If we go with a convention of ours, then the reference implementation of OpAMP-Go won't work as-is with df_engine (I patched the opamp-go to make it work temporarily).

If OpAMP does not define the key, then we'd need to make this key come from user-config, so one can override it to match what their control plane provides. Also, I think we can default to empty-string key, so we'd work out of box with opamp go, and if a user desires, they can override the key.

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: Yeah you're right Lalit - for some reason I was thinking this Restart command should be used to restart the OpAMP Agent, but it makes much more sense that it would restart whatever the OpAMP agent is in control of. 

In 965b94201 I removed support for this, as we don't currently have any mechanism in the otel-arrow admin crate to just restart the process. I added an issue (https://github.com/open-telemetry/otel-arrow/issues/3428) to track adding restart command support in the future.

rust/otap-dataflow/crates/controller/src/extension/opamp/README.md
albertlockett: yes, good idea. added in 32cfdfe72

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: yes, corrected in 32cfdfe72 so the key here is configurable and by default we use empty-string. With this change, the implementation will work with opamp go server example without pathing that server implementation

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
```diff
@@ -0,0 +1,1971 @@
...
+                    context,
+                );
+
+                let send_result =
+                    send_message(&reconcile_result_message, cancellation_token, ws_sender).await;
+                if !is_message_send_success(send_result) {
+                    return false;
+                }
+                session_state.last_sent_message = Some(reconcile_result_message);
+                session_state.last_config_hash = Some(engine_config.config_hash.clone());
```
lalitb, line 1030: Sorry for catching this late, but I think we still need to remember the last remote config status, not just the hash. If reconcile fails, we store this hash here, but later heartbeat/full-state messages rebuild `RemoteConfigStatus` from observed pipeline state and can report this same failed hash as `Applied`. That can make the OpAMP server think a failed remote config was successfully applied.

rust/otap-dataflow/configs/opamp-controller-extension.yaml
```diff
@@ -0,0 +1,17 @@
...
+  controller:
+    extensions:
+      opamp:
+        type: "urn:otel:extension:opamp"
+        config:
+          endpoint: "ws://127.0.0.1:4320/v1/opamp"
+          # Fixed UUID so the agent is easy to find in the UI. Omit to
+          # generate a fresh UUIDv7 on each start.
+          instance_uid: "8be4df61-93ca-11d2-aa0d-00e098032b8c"
+          heart_beat_interval: "5s"
```
lalitb:
```suggestion
          heartbeat_interval: "5s"
```

lalitb (approved):
Thanks @albertlockett. LGTM with one pending issue for config-status/hash - https://github.com/open-telemetry/otel-arrow/pull/3410#discussion_r3547613733.

rust/otap-dataflow/configs/opamp-controller-extension.yaml
cijothomas: thanks!

rust/otap-dataflow/configs/opamp-controller-extension.yaml
```diff
@@ -0,0 +1,17 @@
+version: otel_dataflow/v1
```
cijothomas, line 1: nit: update the readme on configs directory to mention this too.

cijothomas (approved):
Looks good. Tested with reference server implementation and works well. 
Great work - and thanks!

Dipanshusinghh (commented):
Looks really good, albert! I just have a couple of thoughts on the remaining issues to help get the PR ready for merge:

### 1. Tracking config application outcome (re: Lalit's comment)
Since `last_config_hash` alone doesn't capture whether the reconciliation succeeded or failed, subsequent heartbeats/status messages can easily misreport a failed config as `Applied`. 

We should track the full outcome in `SessionState` instead of just the hash:
```rust
struct SessionState {
    // ...
    last_config_status: Option<RemoteConfigStatus>,
}

rust/otap-dataflow/crates/controller/src/extension/opamp/mod.rs
albertlockett: no need to apologize - thanks for catching this! Fixed in [f3bb36a](https://github.com/open-telemetry/otel-arrow/pull/3410/commits/f3bb36abd73977a0fe36e580e2dd3baa2f79bd9f)

rust/otap-dataflow/configs/opamp-controller-extension.yaml
albertlockett: Good idea, added in 60d94b9

cijothomas:
@albertlockett 

I pulled the latest PR and tried the example again with an unmodified opamp-go ref implementations. Startup shows this before I push anything from the UI:

```text
ERROR otap-df-controller::opamp.controller_extension.message.invalid_config_json: Could not deserialize JSON encoded engine config [error=Error("EOF while parsing a value", line: 1, column: 0)]

albertlockett:
> @albertlockett
> 
> I pulled the latest PR and tried the example again with an unmodified opamp-go ref implementations. Startup shows this before I push anything from the UI:
> 
> ```
> ERROR otap-df-controller::opamp.controller_extension.message.invalid_config_json: Could not deserialize JSON encoded engine config [error=Error("EOF while parsing a value", line: 1, column: 0)]
> ```

looks like the server was sending an empty body. fixed in [38de3b5](https://github.com/open-telemetry/otel-arrow/pull/3410/commits/38de3b5720f8a5abe3121a4ec5d9398fbf529a6d)

cijothomas:
> > @albertlockett
> > I pulled the latest PR and tried the example again with an unmodified opamp-go ref implementations. Startup shows this before I push anything from the UI:
> > ```
> > ERROR otap-df-controller::opamp.controller_extension.message.invalid_config_json: Could not deserialize JSON encoded engine config [error=Error("EOF while parsing a value", line: 1, column: 0)]
> > ```
> 
> looks like the server was sending an empty body. fixed in [38de3b5](https://github.com/open-telemetry/otel-arrow/pull/3410/commits/38de3b5720f8a5abe3121a4ec5d9398fbf529a6d)

Yes that did the trick!

